PR #1060, opened 2026-05-31
Fail stdio RPCs on a closed transport instead of hanging

srid:
**A `LinkStdioClient` bound to a dead stream pair would hang forever instead of failing.** When the inbound stream has ended (the agent subprocess exited, the ssh pipe dropped), a request issued on the link awaited a response that never arrived *and never errored* — the link is one-shot (one link per stream pair, by design), so there's no reconnect to rescue it.

This wedged `@kolu/surface-nix-host`'s reconnect bridge in production. On a link drop the bridge re-issues `system.get` against the just-exited child's client; that pump never settled, so `Promise.allSettled` never returned, the reconnect loop never advanced, and **every genuinely-respawned agent sat idle until the 30s connect watchdog reaped it** — 5 strikes → terminal `failed`. One transient drop permanently downed the host (reproduced end-to-end against a real fleet: a single killed ssh child took the host from `connected` to `failed` with ~30 respawned agents stranded).

### The fix

`LinkStdioClient` now tracks a one-way `closed` latch, set when `readFramedLines` settles (stream `end` or `error`):

```
read-stream end/error ─▶ handleTransportClosed() ─▶ closed = true; peer.close()
                                                       │
call() ── if (closed) ─▶ throw SURFACE_STDIO_TRANSPORT_CLOSED   (was: await forever)
```

`peer.close()` already rejected *in-flight* requests; the new guard covers the **issued-after-close** path — `ClientPeer.request()` on a closed peer re-opens its response queue and then pulls a message that never comes. A dead link now fails fast, so the bridge's pump errors → `allSettled` returns → the loop binds the live agent.

### Also

- **Retry fence on `createStdioCellsClient`** — the new rejection is an `ORPCError`, so a caller opting into `retry: N` would otherwise burn N round-trips against a dead link. Pinned `shouldRetry` at the plugin default, extracting the `!(error instanceof ORPCError)` predicate that `STREAM_RETRY` already used into a shared `shouldNotRetryORPCError` so the two can't drift.

> _Regression test: a live loopback round-trip, then end the agent's stdout and assert the next RPC rejects. Pre-fix it timed out at 5s; post-fix it rejects in <1ms._

### Try it locally

```sh
nix run github:juspay/kolu/fix-reconnect-handoff-wedge
```

_Generated by [`/do`](https://github.com/srid/agency) on Claude Code (model `claude-opus-4-8`)._

PR #1064, opened 2026-05-31
Fix reconnect bridge busy-spin (thenable client identity)

srid:
After #1060 made the stdio link fail fast, the reconnect bridge **busy-spun at CPU speed** when a *connected* link dropped — pegging the single-threaded event loop so the child `exit` handler and reconnect-backoff timer never ran, taking the whole parent server unresponsive from one host's drop. (Observed in drishti prod: one killed ssh child → ~2M loop iterations in 2 min, server at 70% CPU, HTTP unresponsive.)

**Root cause (reproduced in `reconnect-spin.test.ts`):** the agent client is an oRPC proxy that forwards every property access — including `.then` — as a procedure path, so it is **thenable**. `await session.currentClient()` therefore re-invokes the proxy and yields a *fresh object* every call, making `waitForNextClient`'s `client !== previous` identity check **always** true → it resolved on every consumer iteration. Pre-#1060 the pump *hung*, so the loop never came back to call it repeatedly (the hang was an accidental brake); #1060's fail-fast removed the brake → spin.

Proof from the repro instrumentation: `samePromise≈100%` (the `clientPromise` is one stable Promise — a single spawn) while `sameAsPrev=0` (the awaited client differs every call).

**Fix:** compare the `clientPromise` reference — reassigned exactly once per spawn, `null` between a child's death and the next spawn — instead of the thenable client. The comparison axis is now owned by a **`makeClientCursor(session)`** cursor: each `cursor.next()` blocks until a genuinely new spawn appears, then resolves with that spawn's live client. The loop blocks until a real reconnect, which frees the event loop so `exit` + `scheduleReconnect` fire and recovery proceeds.

Regression test: pre-fix **36,683** iterations in 500ms; post-fix blocks until a real reconnect.

### Why a cursor, not a return-the-token shape

The first cut had `waitForNextClient` return `{ client, clientPromise }` and asked every consumer to thread `clientPromise` back in as `previous`. That leaked the promise-identity comparison axis into the public contract *and* carried a silent footgun: forget to advance the token and `next()` resolves every iteration — the exact busy-spin this PR fixes. `makeClientCursor` closes over the token internally, so consumers just call `cursor.next()` and there is nothing to mis-thread. `waitForNextClient` / `NextClient` are now module-private.

### Refinements landed during structural + quality review

| What was off | Fix |
| --- | --- |
| `{ client, clientPromise }` leaked the spawn-identity token into every consumer | Encapsulated behind `makeClientCursor`; `waitForNextClient` demoted to private (Lowy) |
| Test inlined a `createLoopbackPair` copy, justified by a non-existent "subpath snag" | Import the canonical `@kolu/surface/loopback` (Hickey) |
| Test leaked `HostSession`'s real 50ms reconnect timers after exit | `session.destroy()` in `afterEach`, matching `hostSession.test.ts` |

_Follow-up to #1060._

### Try it locally

```sh
nix run github:juspay/kolu/fix-reconnect-busy-spin
```

### DIFF
--- a/packages/surface-nix-host/package.json
+++ b/packages/surface-nix-host/package.json
@@ -20,8 +20,10 @@
     "@orpc/contract": "^1.13.13"
   },
   "devDependencies": {
+    "@orpc/server": "^1.13.13",
     "@types/node": "^22.0.0",
     "typescript": "^5.8.0",
-    "vitest": "^4.1.2"
+    "vitest": "^4.1.2",
+    "zod": "^4.3.6"
   }
 }
--- a/packages/surface-nix-host/src/index.ts
+++ b/packages/surface-nix-host/src/index.ts
@@ -34,4 +34,4 @@ export {
   runCapture,
   runProgress,
 } from "./process";
-export { waitForNextClient } from "./waitForNextClient";
+export { type NextClient, waitForNextClient } from "./waitForNextClient";
--- a/packages/surface-nix-host/src/index.ts
+++ b/packages/surface-nix-host/src/index.ts
@@ -34,4 +34,4 @@ export {
   runCapture,
   runProgress,
 } from "./process";
-export { type NextClient, waitForNextClient } from "./waitForNextClient";
+export { type ClientCursor, makeClientCursor } from "./waitForNextClient";
--- a/packages/surface-nix-host/src/reconnect-spin.test.ts
+++ b/packages/surface-nix-host/src/reconnect-spin.test.ts
@@ -23,7 +23,7 @@ import { z } from "zod";
 import type { AgentClient } from "./hostSession";
 import { HostSession } from "./hostSession";
 import { provisionAgent } from "./nixCopy";
-import { waitForNextClient } from "./waitForNextClient";
+import { makeClientCursor } from "./waitForNextClient";
 
 vi.mock("./nixCopy", () => ({ provisionAgent: vi.fn() }));
 vi.mock("node:child_process", () => ({ spawn: vi.fn() }));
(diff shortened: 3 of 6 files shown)


srid:

| # | Lens   | Finding                                                  | Disposition      |
|---|--------|----------------------------------------------------------|------------------|
| 1 | Lowy   | `NextClient` leaks the spawn-identity token into consumers | Fixed in this PR |
| 2 | Hickey | Test inlines `createLoopbackPair` instead of the canonical `@kolu/surface/loopback` | Fixed in this PR |

Both first-pass findings were dispositioned **Fix in this PR** — no No-ops, no defers. A cross-validation pass ran (each lens audited the other's recommendation against the diff); both endorsed the other's fix and surfaced **no new findings**.

### Lowy rationale

`waitForNextClient` returned `{ client, clientPromise }`, and `clientPromise` is not a value the consumer can *use* — its sole purpose is to be stored and re-threaded back as `previous`. That made the comparison axis (promise reference identity) part of the public contract, even though the comparison strategy is an implementation detail of `HostSession` (it could become a spawn-epoch counter, a UUID, an event token). Two call sites (the router bridge loop and the regression test) manually re-threading the same plumbing is the structural tell that the token-threading should be first-class.

**Fix:** a stateful `makeClientCursor(session)` cursor that owns the token in a closure and exposes only `next(): Promise<AgentClient<C>>`. `waitForNextClient` / `NextClient` are demoted to module-private. Both call sites collapse to `await cursor.next()`. Blast radius of any future change in comparison strategy drops to zero.

### Hickey rationale

The regression test inlined a local `createLoopbackPair` re-implementing the canonical primitive (same cross-pipe convention, same `client`/`server`/`read`/`write` shape), justified by a comment claiming a "cross-package subpath-export snag in the runner." That claim is contradicted by the file itself — it already imports `@kolu/surface/peer-server`, proving cross-package subpath resolution works in this exact runner, and `@kolu/surface` is already a `workspace:*` dependency. The inline copy was a silent-divergence trap against the cross-pipe convention `serveOverStdio` depends on: if `LoopbackPair`'s convention shifts, the imported link moves while the untyped local copy goes stale with no compiler error.

**Fix:** import `createLoopbackPair` from `@kolu/surface/loopback`; delete the copy and the false comment.

> Cross-validation note: Hickey confirmed `makeClientCursor` is clean under Simple Made Easy (controlled local closure state, one concern, hides only plumbing) and independently flagged that hand-threading could silently re-create the busy-spin — which the cursor makes structurally impossible. Lowy confirmed importing the canonical loopback introduces no new cross-package coupling (the dependency already exists) and in fact *re-unites* the test with the volatility it's already bound to via `serveOverStdio`.

PR #1065, opened 2026-05-31
Suppress 'Remove worktree' when the branch has unpushed work or an open PR

### LINKED ISSUE #707: Suppress 'Remove worktree' when the branch has unpushed work or an open PR
Follow-up to #703 / #705.

**Problem.** Today the close-confirmation dialog offers "Remove worktree too?" as long as no other terminal is on the same worktree. Two real scenarios lose data or context when the user confirms:

1. **Unpushed commits.** The branch has local commits that aren't reachable from any remote — `git worktree remove` doesn't guard against this, and deleting the branch afterward makes the commits unreachable.
2. **Open pull request.** The branch has an open/draft PR on the remote. Removing the worktree leaves the PR dangling; the user typically wants to keep the worktree around to respond to review, run CI, iterate.

Both are "the branch has unfinished work" in spirit, but the copy and the remediation differ, so model them as two separate blockers.

**Proposed.** Extend `WorktreeRemovalBlocker` in `packages/client/src/CloseConfirm.tsx`:

```ts
export type WorktreeRemovalBlocker =
  | "sharedWithOtherTerminals"
  | "hasUnpushedCommits"
  | "hasOpenPullRequest";
```

Add entries to `BLOCKER_MESSAGES`, e.g.:

- `hasUnpushedCommits`: "This worktree has unpushed commits — it will remain on disk so you don't lose work."
- `hasOpenPullRequest`: "This worktree's branch has an open PR (#123) — it will remain on disk."

The dialog UI already renders any blocker generically via `<Show when={removalBlocker()}>` + the `data-blocker` attribute, so no markup changes are needed. If the PR copy needs to embed the PR number dynamically, the `BLOCKER_MESSAGES` record shape will need to widen from `Record<Blocker, string>` to `Record<Blocker, (ctx) => string>` — call that out during implementation.

**Where to compute.** `closeTerminal` in `App.tsx` is the single writer of `worktreeRemoval`. Both checks should flow off the existing `TerminalMetadata`:

- `hasOpenPullRequest`: `prValue(meta.pr)?.state` is already available on the metadata stream (used by the dialog's existing PR card at `CloseConfirm.tsx:112-137`). States that count: `"open"`, `"draft"` — `"merged"` / `"closed"` don't block.
- `hasUnpushedCommits`: not currently on `meta.git`. Needs a server-side derivation (rev-walk `git.branch` against its upstream, yield an `unpushedCommitCount: number` on metadata refresh). Same seam as the existing git fields — extend the Zod schema in `packages/integrations/git` and the server-side computation.

Avoid an on-demand RPC inside `closeTerminal`: the function is non-async today, and making it async would flash the dialog open in the "eligible" state and shift to "blocked" once the RPC resolves, violating the snapshot-at-open invariant documented on `worktreeRemoval`.

**Ordering when multiple blockers apply.** A worktree can legitimately hit all three at once (shared + unpushed + open PR). The current union shape `{ eligible: false; reason: Blocker }` returns only one. The honest shape — `{ eligible: false; reasons: Blocker[] }` — is a deferred design call. For now, pick a priority ordering so the user sees the most load-bearing reason first:

1. `hasUnpushedCommits` — data-loss risk, highest priority
2. `hasOpenPullRequest` — coordination/review risk
3. `sharedWithOtherTerminals` — coordination-only, lowest

**Tests.** Two new scenarios in `worktree.feature`, each mirroring the structure of the existing "shared" scenario:

- Commit without pushing → `[data-blocker="hasUnpushedCommits"]` visible + remove button absent.
- Open a PR (can be a mocked GitHub API fixture — see how existing PR metadata is faked in tests) → `[data-blocker="hasOpenPullRequest"]` visible + remove button absent.

If PR-info mocking isn't already factored for e2e, the PR scenario may need `@skip` + a comment pointing at the fixture work, rather than holding up the unpushed-commits path.

srid:
Closes #707.

## Why

The close-confirmation dialog offered **"Remove worktree too?"** whenever no other terminal shared the worktree. Two real cases lose work or context when the user confirms:

1. **Unpushed commits** — `git worktree remove` doesn't guard against local commits that aren't on any remote; deleting the branch afterward makes them unreachable.
2. **Open PR** — removing the worktree leaves the PR dangling, when the user usually wants to keep iterating, respond to review, and re-run CI.

Both are "the branch has unfinished work", but the copy and remediation differ, so they're modelled as two new removal blockers alongside the existing shared-worktree one.

## What changed

**Client (`CloseConfirm.tsx`, `App.tsx`).** `WorktreeRemovalBlocker` gains `hasUnpushedCommits` and `hasOpenPullRequest`. `BLOCKER_MESSAGES` widens from `Record<Blocker, string>` to context-taking functions so the open-PR copy can embed the PR number. `closeTerminal` — still the single, **synchronous** writer of `worktreeRemoval` (preserving the snapshot-at-open invariant) — picks the highest-stakes blocker by priority:

| Priority | Blocker | Source |
|---|---|---|
| 1 | `hasUnpushedCommits` | `meta.git.unpushedCommitCount > 0` (data-loss risk) |
| 2 | `hasOpenPullRequest` | `prValue(meta.pr)?.state === "open"` (coordination risk) |
| 3 | `sharedWithOtherTerminals` | `store.isWorktreeShared(...)` |

> The issue suggested `state in {open, draft}`, but the PR schema enum is only `open | closed | merged` — there is **no** `draft` state. Only `open` blocks; `closed`/`merged` mean the work has landed.

**Server (`integrations/git`).** `unpushedCommitCount` (`@{u}..HEAD`, never throws — 0 on no-upstream/detached HEAD) is derived onto `GitInfo`, compared in `gitInfoEqual` so a commit isn't deduped away, and `subscribeGitInfo` now also installs the reflog watcher so a local commit refreshes the count.

> **Known limitation (documented in the schema JSDoc):** `git push` moves only the remote-tracking ref, which this watcher set doesn't observe, so the count can briefly lag a push until the next HEAD movement. The stale direction is safe — it errs toward *keeping* the worktree, never toward silently deleting unpushed work.

## Tests

- Unit: `unpushedCommitCount` (no-upstream → 0; two ahead → 2) and a `gitInfoEqual` case proving the count is compared.
- e2e: a `hasUnpushedCommits` scenario in `worktree.feature` (passes locally via `just test-quick`).
- e2e (`@skip`): a `hasOpenPullRequest` scenario. PR metadata is fetched server-side by spawning the real `gh pr view` binary (`KOLU_GH_BIN`); the harness ships no fake `gh` and no `meta.pr` injection seam, so an open-PR snapshot can't be produced offline. The scenario is written and skipped with a comment pointing at the fixture work needed to unskip it.

## Approach note

Scoped with a dynamic multi-agent workflow: parallel research agents nailed down the server metadata seam, the PR-state plumbing, and the e2e harness's mocking limits; a synthesis agent produced the implementation plan that this PR follows.

### DIFF
--- a/packages/client/src/App.tsx
+++ b/packages/client/src/App.tsx
@@ -285,18 +285,22 @@ const App: Component = () => {
     // re-fetch), keeping the dialog a frozen confirmation. Only an `open` PR
     // blocks — closed/merged PRs mean the work has landed. `prValue` is null
     // for pending/absent PRs, so an unconfirmed PR never blocks.
+    const pr = prValue(meta.pr);
     const blocker: WorktreeRemovalBlocker | undefined = !worktreePath
       ? undefined
       : (meta.git?.unpushedCommitCount ?? 0) > 0
         ? "hasUnpushedCommits"
-        : prValue(meta.pr)?.state === "open"
+        : pr?.state === "open"
           ? "hasOpenPullRequest"
           : store.isWorktreeShared(worktreePath, id)
             ? "sharedWithOtherTerminals"
             : undefined;
     const worktreeRemoval = worktreePath
       ? blocker
-        ? ({ eligible: false, reason: blocker } as const)
+        ? // Freeze the PR number here, at decision time, so the dialog's
+          // message reads it off this snapshot rather than the reactive
+          // `meta.pr`. Only `hasOpenPullRequest` carries one.
+          ({ eligible: false, reason: blocker, prNumber: pr?.number } as const)
         : ({ eligible: true } as const)
       : undefined;
     setCloseConfirmTarget({ id, meta, splitCount, worktreeRemoval });
--- a/packages/client/src/App.tsx
+++ b/packages/client/src/App.tsx
@@ -9,6 +9,7 @@ import Dialog from "@corvu/dialog";
 import { Meta, Title } from "@solidjs/meta";
 import type { ServerIdentity } from "kolu-common/contract";
 import type { TerminalId } from "kolu-common/surface";
+import { prValue } from "kolu-github/schemas";
 import {
   type Component,
   createEffect,
@@ -20,7 +21,10 @@ import {
 import { Toaster } from "solid-sonner";
 import { match } from "ts-pattern";
 import ChromeBar from "./ChromeBar";
-import CloseConfirm, { type CloseConfirmTarget } from "./CloseConfirm";
+import CloseConfirm, {
+  type CloseConfirmTarget,
+  type WorktreeRemovalBlocker,
+} from "./CloseConfirm";
 import CommandPalette from "./CommandPalette";
 import "kolu-common/test-hooks";
 import CanvasWatermark from "./canvas/CanvasWatermark";
@@ -275,9 +279,24 @@ const App: Component = () => {
     const worktreePath = meta.git?.isWorktree
       ? meta.git.worktreePath
       : undefined;
+    // Priority: unpushed local work > open PR > shared worktree — the first
+    // match wins, so the highest-stakes reason to keep the worktree is what
+    // the user sees. Every read is synchronous off the metadata snapshot (no
+    // re-fetch), keeping the dialog a frozen confirmation. Only an `open` PR
+    // blocks — closed/merged PRs mean the work has landed. `prValue` is null
+    // for pending/absent PRs, so an unconfirmed PR never blocks.
+    const blocker: WorktreeRemovalBlocker | undefined = !worktreePath
+      ? undefined
+      : (meta.git?.unpushedCommitCount ?? 0) > 0
+        ? "hasUnpushedCommits"
+        : prValue(meta.pr)?.state === "open"
+          ? "hasOpenPullRequest"
+          : store.isWorktreeShared(worktreePath, id)
+            ? "sharedWithOtherTerminals"
+            : undefined;
     const worktreeRemoval = worktreePath
-      ? store.isWorktreeShared(worktreePath, id)
-        ? ({ eligible: false, reason: "sharedWithOtherTerminals" } as const)
+      ? blocker
+        ? ({ eligible: false, reason: blocker } as const)
         : ({ eligible: true } as const)
       : undefined;
     setCloseConfirmTarget({ id, meta, splitCount, worktreeRemoval });
--- a/packages/client/src/CloseConfirm.tsx
+++ b/packages/client/src/CloseConfirm.tsx
@@ -12,20 +12,40 @@ import { PrStateIcon, WorktreeIcon } from "./ui/Icons";
 import ModalDialog from "./ui/ModalDialog";
 import { surface } from "./ui/Surface";
 
-/** Reasons why the "Remove worktree" action is suppressed.
+/** Reasons why the "Remove worktree" action is suppressed, in priority
+ *  order — the first that applies is the one shown (the ordering lives in
+ *  `closeTerminal` in `App.tsx`):
  *
- *  One blocker today (`"sharedWithOtherTerminals"`); the union shape
- *  is ready for future reasons (unpushed commits, per-user preference,
- *  running agent) without a breaking change to `CloseConfirmTarget`. */
-export type WorktreeRemovalBlocker = "sharedWithOtherTerminals";
+ *  1. `hasUnpushedCommits` — local commits not on any remote (data-loss risk).
+ *  2. `hasOpenPullRequest` — an open PR the user is iterating on.
+ *  3. `sharedWithOtherTerminals` — another terminal lives on the worktree. */
+export type WorktreeRemovalBlocker =
+  | "hasUnpushedCommits"
+  | "hasOpenPullRequest"
+  | "sharedWithOtherTerminals";
 
 /** Whether the close dialog may offer worktree removal. */
 export type WorktreeRemovalEligibility =
   | { eligible: true }
   | { eligible: false; reason: WorktreeRemovalBlocker };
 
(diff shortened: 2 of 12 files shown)


srid:

Both lenses ran in parallel on the diff, then cross-validated each other's findings. Both independently surfaced the same two issues (strong signal), and Hickey's cross-validation added a third.

| 1 | Hickey + Lowy | PR number read *live* from `meta.pr` inside the dialog's reactive render, while `worktreeRemoval.reason` is a frozen dialog-open snapshot | Fixed in this PR |
| 2 | Hickey + Lowy | `WatcherMode "head"` installs HEAD **and** reflog watchers but names only the HEAD file | Fixed in this PR |
| 3 | Hickey (cross-validation) | Blocker priority ordering duplicated — stated in the `WorktreeRemovalBlocker` JSDoc *and* enforced in `closeTerminal` | Fixed in this PR |
| 4 | Lowy | `unpushedCommitCount` refresh doesn't observe `git push` | ⚠️ **No-op** |
| 5 | Lowy (cross-validation) | `unpushedCommitCount` rides the shared `GitInfo` schema (propagated into ~5 test fixtures) rather than a dedicated safety derivation | ⚠️ **No-op** |

Fixes landed as `refactor(lowy)` / `refactor(hickey)` commits; the follow-on `/code-police` elegance pass added watcher test-teardown parity, reflog install-once assertions, and inlined a single-use type alias.

### Hickey rationale
The diff cleanly separates count-tracking (`resolve.ts`), event-wiring (`subscribeGitInfo`), and UI policy (`closeTerminal` + `BLOCKER_MESSAGES`) with no interleaving. Finding 1 was a fragmentation: the `GitHubPrInfo` was computed in `App.tsx` then discarded, and the PR number re-extracted downstream via a *reactive* read — the eligibility union carried only the reason discriminator. Finding 2: the `"head"` discriminator named only one of the two watchers it installs after the reflog was added. Cross-validation flagged that the priority ordering was load-bearing policy stated in two places (Finding 3); ordering now lives only at the enforcement site.

### Lowy rationale
The boundaries encapsulate the right volatilities: `unpushedCommitCount` co-located with the other HEAD-derived `GitInfo` fields; the HEAD+reflog watcher pair hidden behind one composite teardown (correct — exposing them separately would leak which `.git` file encodes which event); the blocker priority cascade (sequence volatility) kept in `closeTerminal`, separate from how each blocker is computed (activity volatility). Finding 1 was a receptacle leak — the frozen eligibility claimed to carry everything the dialog needs, yet a second *live* read of `meta.pr` was required; resolved by freezing `prNumber` on the ineligible arm at decision time. The push-staleness (4) is self-documented and errs safe (toward keeping the worktree); the schema-breadth note (5) is below the cost threshold at three-blocker scale — re-evaluate if a fourth git-derived safety field lands.

srid:
## [`/do`](https://github.com/srid/agency) results — `--from polish`

| hickey+lowy | ✓ | **12m 15s** | Both lenses ran + cross-validated. 3 Fix findings committed (freeze PR number on eligibility, rename `WatcherMode` → `"in-repo"`, dedup blocker priority); 2 No-op. |
| police | ✓ | **12m 56s** | `/code-police` ×2. Rules + fact-check clean. Elegance applied (reflog watcher test-teardown parity, install-once assertions, inlined alias). |
| test | ✓ | 58s | `worktree.feature`: 9 scenarios / 153 steps pass, incl. the `hasUnpushedCommits` blocker. Git unit tests pass. |
| create-pr | ✓ | 43s | PR already open; scope unchanged. Posted Hickey/Lowy analysis. |
| ci | ✓ | 11m 4s | **x86_64-linux: 376/376 scenarios, 3017/3017 steps green** (incl. this change). darwin units green; darwin e2e 12/376 failed across 8 **unrelated** features (`worktree.feature` passed) — host-contention/pre-existing flake (#320), owned separately. |
| evidence | — | 45s | Skipped: behavior proven by the passing `worktree.feature` e2e; off-box capture deferred to avoid extra box churn. Available on request. |
| **Total** | | **38m 58s** | |

**Slowest step**: `police` (12m 56s)

### Caveats (not overstated)
- **darwin e2e is red on 12 unrelated scenarios** (contention from a concurrent CI run against the single shared darwin host). This PR's own feature (`worktree.feature`) is green on both lanes. Darwin failures are being handled separately.
- **Evidence not captured** — deferred per request; the e2e already drives the real dialog.

### Optimization suggestions
- **`police` (12m56s) ran twice** — the elegance pass surfaced new refactors on the 2nd invocation. Running `/code-police` (or `/simplify`) once during implementation, before `/do --from polish`, would let those land in the primary commits and leave the gate clean in one pass.
- **`hickey+lowy` (12m15s)** is dominated by 4 sequential sub-agent round-trips (2 review + 2 cross-validation). That's inherent to the value on a non-trivial diff; `--minimal` skips it entirely for trivial ones.
- **darwin CI contention** caused all 12 e2e reds. Serialize darwin runs against the shared host (or give darwin its own ephemeral box like the linux lane) so concurrent invocations stop producing false 20s-timeout failures.

Workflow completed at 2026-05-31T15:06Z.

srid:
## Evidence

When a worktree branch has unpushed commits, the close-confirmation dialog blocks the "remove worktree" action and shows a safety note instead of the remove button. Beat 2 shows the contrast: a clean worktree gets the full offer including "Close terminal and remove worktree".

### Beat 1 — unpushed commits → blocker (no remove button)



### Animated walkthrough (both beats at 2×)



▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=worktree-removal-blocker.mp4

PR #1067, opened 2026-05-31
Pin the realised agent behind a per-agent GC root

srid:
**After realising the agent on a target host, `provisionAgent` now registers an indirect GC root for it**, so a `nix-collect-garbage` on that host can't delete the agent out from under a live session — or silently force a rebuild on the next reconnect. Previously nothing held a root on the realised output: it survived only by luck between connects.

The root is a single fixed symlink **per agent**, keyed on the `.drv` name with its store hash stripped. Each realise moves that link to the newest output, so the previous hash drops out of the root set and becomes GC-eligible — the same moving-`result` semantics `nix build` gives you, but on the target's store. "Pin the latest, release the old," with nothing to prune.

### Flow

```
copy  : nix copy --derivation --to ssh-ng://$host $drv
realise: nix-store --realise $drv                      → /nix/store/…-agent   (agentPath)
pin   : nix-store --realise $agentPath --add-root $link --indirect            (new)
spawn : ssh $host $agentPath/bin/<binary> --stdio
```

Re-realising an already-built path is instant, so the pin is one cheap extra round-trip. `--indirect` registers the *symlink itself* as the root, so it can live under `$HOME` without write access to `/nix/var/nix/gcroots` — which matters because the ssh user is rarely root.

### Best-effort by design

The pin never fails provisioning. If the root can't be placed — local `$HOME` unset, or the `nix-store` call errors — `provisionAgent` logs a progress line and returns the realised `agentPath` anyway. The agent still runs; it's just collectable again. `agentGcRootPath` returns `null` for the one case it genuinely can't anchor (localhost with no `$HOME`) rather than rooting under the process cwd.

> Remote roots resolve against the ssh login `$HOME` (sshd runs the command from there); localhost anchors to `process.env.HOME` explicitly since there's no ssh chdir.

### DIFF
--- a/packages/surface-nix-host/README.md
+++ b/packages/surface-nix-host/README.md
@@ -36,7 +36,7 @@ const sys = client.system.get({});           // typed oRPC — same shape your b
 ## What it's NOT
 
 - **Not a transport.** That's [`@kolu/surface/links/stdio`](../surface/src/links/stdio.ts). This package sits *on top of* the stdio link and adds: process supervision (spawn/respawn ssh), `.drv` provisioning (the Nix bit), and a reactive state cell for the connection's own lifecycle.
-- **Not a Nix utility.** `provisionAgent` is purposely minimal — `nix copy --derivation`, then `ssh $host nix-store --realise`, then return the resulting path. If you want richer flake handling (e.g. resolve a flake ref to a `.drv`), do it inside the `resolveDrvPath` callback you pass.
+- **Not a Nix utility.** `provisionAgent` is purposely minimal — `nix copy --derivation`, then `ssh $host nix-store --realise`, then pin the output behind a per-agent GC root, then return the resulting path. If you want richer flake handling (e.g. resolve a flake ref to a `.drv`), do it inside the `resolveDrvPath` callback you pass.
 - **Not opinionated about UI.** The package returns a typed RPC client. Mirroring its streams into your parent server's local surface (so the browser can consume them) is the consumer's job. See the [`remote-process-monitor` example](../surface/example/remote-process-monitor/src/server/router.ts) for the canonical bridge pattern.
 
 ## Why Nix (locked-in)
@@ -50,8 +50,16 @@ parent: nix eval --raw .#packages.<remote-system>.<agent>.drvPath
         └── caller's responsibility — `resolveSystem(host)` gives <remote-system>
 parent: nix copy --derivation --to ssh-ng://$host $drvPath
 remote: nix-store --realise $drvPath  →  /nix/store/...-agent
+remote: nix-store --realise $out --add-root <link> --indirect  (GC-pin)
 parent: ssh $host $agentPath/bin/<binary> --stdio
 ```
+
+The pin step is what keeps a `nix-collect-garbage` on the target from
+deleting the agent out from under a live session (or forcing a rebuild on
+the next reconnect). The root is one fixed symlink per agent — keyed on the
+`.drv` name — so each realise moves it to the newest output and the previous
+hash becomes GC-eligible, exactly like `nix build`'s `result` link. It's
+best-effort: if the root can't be written, the agent still runs unpinned.
 
 **Nix is the contract, not the implementation.** No tarball, Docker, or prebuilt-binary fallback exists or will. The whole point of this package is "use Nix for cross-arch deployment of typed stdio agents"; consumers that don't want Nix should pick a different transport layer.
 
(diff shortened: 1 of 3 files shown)


srid:

Two lenses reviewed the diff in parallel, then cross-validated each other's findings.

| 1 | Hickey | `agentGcRootPath` silently returned a cwd-relative path when local `$HOME` is unset, untested | **Fixed in this PR** |
| 2 | Hickey ↔ Lowy | Re-export `agentGcRootPath` from `index.ts` | ⚠️ **No-op** |
| 3 | Hickey + Lowy | `provisionAgent` step-comments read `1, 2, 4` (no `3`), disagreeing with the module header | **Fixed in this PR** |
| 4 | Lowy | Best-effort/fatal policy lives in `provisionAgent`, not the pure helper — right altitude | ⚠️ **No-op** |
| 5 | Lowy | No duplication of the GC-root axis anywhere in the package | ⚠️ **No-op** |
| 6 | Lowy | `buildSshProbeCommand` reused, not reinvented — composition, not multiplication | ⚠️ **No-op** |

### Cross-validation

The two findings that touched the same code disagreed, which is exactly what the cross-validation pass is for:

- **#2 (the export) was a genuine Hickey↔Lowy conflict.** Hickey first wanted `agentGcRootPath` re-exported to stop consumers re-implementing the hash-strip naming; Lowy wanted it kept internal for package-surface coherence. On cross-validation **Hickey reversed itself**: there are zero external consumers, the test imports it directly from `./nixCopy`, and exposing the derivation rule would widen the public surface speculatively. Converged to **No-op**. *(If a real consumer ever needs the path, the right move is to surface `gcRootPath` as a `ProvisionResult` datum — not to export the deriver.)*
- **#1 was sharpened by cross-validation.** Hickey flagged the silent branch; Lowy noted the fallback was *semantically wrong*, not merely opaque, and proposed throwing. But a throw would break the best-effort property the rest of the review praised — so the fix returns `null` and the caller skips the pin, keeping provisioning resilient.

### Rationale

**Hickey (structure):** the pin is clean composition — three distinct `nix-store` commands built from the one `buildSshProbeCommand` primitive, with the fatal/best-effort distinction kept out of the pure path helper. The store-path→name derivation is single-use; no existing helper to extend.

**Lowy (boundaries):** the GC-root concern is one genuine volatility axis (root location, naming scheme, best-effort policy, the localhost/remote `$HOME` asymmetry), encapsulated behind `agentGcRootPath` + the pin block and isolated from copy/realise. No prior "where state lives on a target" encapsulation was forked. The boundary is well-placed.

No deferred findings — every Fix landed as its own commit (`refactor(hickey)`, `refactor(lowy)`, `refactor(police)`).

srid:

| sync | ✓ | 0s | forge=github; off latest master |
| research | ✓ | 0s | Baseline reset: drishti master speaks new kolu surface API; kolu fix rebased onto master |
| branch | ✓ | 0s | drishti pin-agent-gcroot off master; kolu surface-nix-host-gcroot-pin rebased |
| check | ✓ | 2m 18s | `just typecheck` (install + hydrate nix-built `@kolu/surface-nix-host` + tsc) clean |
| implement | ✓ | — | GC-root pin in `provisionAgent` + helper + tests; drishti npins re-pin |
| docs | ✓ | 0s | drishti + kolu READMEs note the per-host GC-root pin |
| fmt | ✓ | 24s | `just fmt` clean |
| commit | ✓ | 25s | drishti primary commit pushed |
| hickey+lowy | ✓ | — | parallel + cross-validation: 2 fixes, export conflict → No-op (Hickey reversed) |
| police | ✓ | 4m 30s | rules + fact-check clean; elegance applied 2 test cleanups; 9/9 kolu tests green |
| commit (re-pin) | ✓ | 20s | drishti re-pinned to reviewed kolu HEAD `ca403bf` |
| test | ✓ | — | drishti `bun test` 119/0; new behavior covered by kolu's 9 vitest cases |
| create-pr | ✓ | 0s | juspay/kolu#1067 (logic) + srid/drishti#37 (pin) |
| ci | ✓ | 1m 48s | justci: all 16 nodes (x86_64-linux + aarch64-darwin) succeeded vs HEAD |
| evidence | ✓ | 0s | No UI impact — server-side provisioning; no Evidence comment warranted |

**Slowest step**: `police` (4m 30s).

### Optimization suggestions

- **Cross-repo re-pinning caused two full kolu nix rebuilds.** The drishti pin moved `388928c → ca403bf` after the review/police commits landed on kolu, invalidating the `surface-nix-host` derivation twice (this is most of the 2m 18s `check` + rebuild cost). Landing *all* kolu review/police commits before the first drishti pin would collapse that to one bump.
- **`police` dominated (4m 30s)** because the elegance pass re-ran the kolu vitest suite. For a kolu-only change, `/do --from polish` scoped to the kolu repo skips the drishti steps entirely.
- **CI was green first try** — no flaky retries, nothing to harden. `--from ci-only` is the fast path for any post-merge re-pin re-run.
- Several steps show `—`/`0s` due to a mid-run baseline restart (the kolu-surface API migration landed between attempts); wall-clock for those is understated.

PR #1071, opened 2026-05-31
fix(surface-nix-host): keepalive on the realise ssh so a degraded host can't wedge the spawn cycle forever

srid:
## The bug, from production

A drishti host (`vanjaram`) sat stuck on **"Copying agent to remote…" for hours**. The copy had actually finished long before — what was wedged was the *next* step, the remote realise:

```
15:59:51  derivation copy complete
15:59:51  realising '…-drishti-agent.drv' on remote…   ← last line, then nothing for 11+ min
```

The smoking gun: the host **pinged fine** (0% loss, 218 ms via tailscale) but **refused new TCP/22 connections**, while the in-flight realise ssh — spawned *before* the degradation — sat parked:

```
ssh -o BatchMode=yes vanjaram nix-store --realise /nix/store/…-drishti-agent.drv   (etime 11:21, 0.03s CPU)
```

It used 0.03s of CPU in 11 minutes. It wasn't working — it was blocked on a half-open socket, and would stay that way until the OS TCP stack eventually gave up. The caller's spawn cycle never saw a result, so the connection cell never left `copying`/`connecting`: stuck for eternity, no retry.

## Root cause

`buildSshProbeCommand` wrapped realise/probe commands in bare `ssh -o BatchMode=yes`, **deliberately omitting** the `ServerAliveInterval` that the long-lived `buildAgentCommand` session gets. The old rationale — *"probes are one-shot and deliberately omit it"* — is the bug:

`nix-store --realise` over ssh is **a remote build, not a quick round-trip.** The channel sits idle for minutes while the far end compiles. With no keepalive, a one-shot command blocks on a dead peer exactly as hard as a long-lived one — arguably worse, since nothing downstream watchdogs the pre-spawn provisioning step.

## The fix

Fold the ssh options into one shared keepalive policy carried by **both** builders, and add dead-peer detection:

```
-o BatchMode=yes
-o ServerAliveInterval=10
-o ServerAliveCountMax=3   (was the implicit default for the agent session; now explicit + applied to probes)
-o ConnectTimeout=10
```

A dead peer now trips ssh's keepalive in **~30s** and the command exits non-zero → `runCapture` returns `{ok:false}` → `provisionAgent` fails → the reconnect loop retries with backoff, instead of wedging.

Crucially this does **not** cap a healthy-but-slow build: ssh keepalives ride the protocol layer independently of channel data, so a responsive sshd answers them no matter how long the build's stdout stays quiet. Only an *actually unresponsive* peer trips the limit.

This is the same fast-fail discipline #1060 gave the stdio RPC link — now applied to the provisioning ssh it never covered.

### The wider gap: the copy step had the same hole

The observed wedge was on *realise*, but the structural review found the **`nix copy --derivation --to ssh-ng://` step one stage earlier was exposed to the identical hang** — it forks its *own* ssh, out of reach of the argv we build, and was running with no keepalive at all. A host that degrades mid-copy (the literal "Copying agent to remote…" line in the report) wedges the spawn cycle just as hard. The only handle on that internal ssh is the `NIX_SSHOPTS` env var, so the same policy now flows there too.

To avoid two drifting copies of the option set, the policy is declared **once** as `(key, value)` pairs and rendered into the two encodings its consumers actually need — rather than casting one from the other:

```
                       ┌─▶ SSH_COMMON_OPTS  (-o k=v argv)  ─▶ agent / probe / realise / pin ssh
SSH_OPT_PAIRS  ────────┤
  (single source)      └─▶ NIX_SSHOPTS      (word-split string) ─▶ ssh that `nix copy` forks
```

*Two peer renderings, neither derived-from-the-other — a value with whitespace would corrupt the env string while the argv stayed correct, so the pair-list keeps the invariant in one place.* `runProgress` gains a small `env` param (merged onto the parent env) so the copy step can inject `NIX_SSHOPTS`.

## Tests

`host.test.ts` asserts both argv builders carry the keepalive + connect-timeout (and that localhost still short-circuits the ssh wrapper), and pins that `NIX_SSHOPTS` renders the *same* four-flag policy as the spawned-ssh argv — so the copy path can't silently drift from the realise path. Full `surface-nix-host` suite: **19/19 green**; `just check` (typecheck + biome) clean.

### Try it locally

```sh
nix run github:juspay/kolu/fix-realise-ssh-hang
```

### DIFF
--- a/packages/surface-nix-host/src/host.test.ts
+++ b/packages/surface-nix-host/src/host.test.ts
@@ -21,6 +21,18 @@ function sshOpts(args: readonly string[]): Record<string, string> {
   return opts;
 }
 
+/** Assert an ssh argv carries the full shared dead-peer keepalive policy.
+ *  One invariant ("every non-interactive ssh this package spawns carries
+ *  `SSH_COMMON_OPTS`") asserted in one place, so re-tuning a keepalive
+ *  value can't leave a second hand-synced block green on stale numbers. */
+function assertKeepAlive(args: readonly string[]): void {
+  const opts = sshOpts(args);
+  expect(opts.BatchMode).toBe("yes");
+  expect(opts.ServerAliveInterval).toBe("10");
+  expect(opts.ServerAliveCountMax).toBe("3");
+  expect(opts.ConnectTimeout).toBe("10");
+}
+
 describe("buildSshProbeCommand", () => {
   it("runs the command directly for localhost — no ssh wrapper", () => {
     const { command, args } = buildSshProbeCommand(
@@ -56,13 +68,9 @@ describe("buildSshProbeCommand", () => {
       "--realise",
       "x",
     );
-    const opts = sshOpts(args);
-    expect(opts.BatchMode).toBe("yes");
     // The fix: a degraded host mid-realise must trip ssh's dead-peer
     // detection (~Interval×CountMax) rather than hang forever.
-    expect(opts.ServerAliveInterval).toBe("10");
-    expect(opts.ServerAliveCountMax).toBe("3");
-    expect(opts.ConnectTimeout).toBe("10");
+    assertKeepAlive(args);
   });
 });
 
@@ -90,10 +98,6 @@ describe("buildAgentCommand", () => {
       "/nix/store/x-agent/bin/my-agent",
       "--stdio",
     ]);
-    const opts = sshOpts(args);
-    expect(opts.BatchMode).toBe("yes");
-    expect(opts.ServerAliveInterval).toBe("10");
-    expect(opts.ServerAliveCountMax).toBe("3");
-    expect(opts.ConnectTimeout).toBe("10");
+    assertKeepAlive(args);
   });
 });
--- a/packages/surface-nix-host/src/host.test.ts
+++ b/packages/surface-nix-host/src/host.test.ts
@@ -13,9 +13,10 @@ import { buildAgentCommand, buildSshProbeCommand, NIX_SSHOPTS } from "./host";
 function sshOpts(args: readonly string[]): Record<string, string> {
   const opts: Record<string, string> = {};
   for (let i = 0; i < args.length; i++) {
-    if (args[i] === "-o" && args[i + 1]) {
-      const [k, v] = args[i + 1]!.split("=");
-      opts[k!] = v ?? "";
+    const val = args[i + 1];
+    if (args[i] === "-o" && val) {
+      const [k, v] = val.split("=");
+      opts[k ?? val] = v ?? "";
     }
   }
   return opts;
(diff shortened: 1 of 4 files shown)


srid:

| 1 | Hickey | Duplicate four-line keepalive assertion block in `host.test.ts` (the same fragmentation the prod fix removes, mirrored into the test) | Fixed in this PR |
| 2 | Lowy | `nix copy --to ssh-ng://` forks its own ssh with no keepalive — the same hang, one stage earlier; `SSH_COMMON_OPTS`' "every ssh this package spawns" claim was false | Fixed in this PR |
| 3 | Hickey (cross-validate) | Lowy's prescribed mechanism (`NIX_SSHOPTS = SSH_COMMON_OPTS.join(" ")`) would braid an argv array with a word-split string behind a hidden whitespace-safety invariant | Fixed in this PR |
| 4 | Lowy (cross-validate) | Audited Hickey's `assertKeepAlive` helper — confirmed the coupling tracks the shared constant correctly and doesn't freeze out future divergence | ⚠️ **No-op** (concurrence, no new finding) |

### Hickey rationale

The `SSH_COMMON_OPTS` extraction is a net *de*-complecting move — before the diff, `BatchMode=yes` was an inline literal in both builders and `ServerAliveInterval` lived only in the agent builder, with the "both share the same safety options" invariant enforced by convention alone. The new test file follows the canonical sibling-test pattern (`hostSession.test.ts`, `nixCopy.test.ts`) — no reinvention.

One first-pass finding: the four keepalive options were asserted in two identical blocks. Because the constant isn't exported (intentionally — exporting it just to satisfy a test would widen the public API), the hazard is silent mutation drift: re-tune the values, update one block, the other passes green on stale numbers. Fixed with a test-local `assertKeepAlive(args)` helper paired with the existing `sshOpts` extractor.

On cross-validation: Lowy correctly identified the real `nix copy` correctness gap, but its prescribed fix (derive the `NIX_SSHOPTS` string from the argv array via `.join`) would reintroduce exactly the class of hidden coupling this PR removes — two representations of one concept, coupled by a lossy transform that silently assumes no option value contains whitespace. The clean shape is one canonical `(key, value)` pair-list rendered into two *peer* encodings. That's what landed.

### Lowy rationale

The volatility being encapsulated is "the ssh hardening policy for every subprocess this package causes to be spawned" — an activity axis that just shifted (the prior belief that only long-lived sessions needed keepalive was wrong). `SSH_OPT_PAIRS` is the correct receptacle: module-private source, two renderings, a policy change touches exactly one site.

First-pass finding traced a real unbounded hang: the `nix copy` step bypassed the policy entirely, its `ssh-ng://` transport spawning an ssh with no `BatchMode`/`ServerAlive`/`ConnectTimeout`. A copy that hangs at `copying` stalls the spawn cycle indefinitely — the `connectTimeoutMs` watchdog arms only at `connecting` and never fires. Closed by threading `NIX_SSHOPTS` through a merged `env` param on `runProgress`.

On cross-validation of Hickey's test helper: the shared `assertKeepAlive` does **not** prevent future divergence — it carries no agent-vs-probe branch, asserts only the shared constant, and the per-builder behavior assertions stay independent. If the policies ever need to split, that happens by un-sharing `SSH_COMMON_OPTS` itself, at which point the helper is removed as a matter of course. The coupling is correct *because* they intentionally share the constant. Apply as written — which it was.

PR #1074, opened 2026-05-31
Preserve the file-tree expansion state in the Code tab (in-place tree updates)

srid:
**Expanding directories in the Code-tab browse tree and then interacting with it — previewing a file, or filtering and clearing — no longer collapses the folders you opened by hand.** The tree used to forget the shape you'd built up the moment you clicked a file, and again whenever the file list rebuilt.

The root cause was *how* the `@pierre/trees` wrapper pushed path changes into the tree. Pierre offers two very different ways to update the file list, and the wrapper was using the wrong one:

| Pierre API | Effect on the tree's store |
| --- | --- |
| `resetPaths` | **Tear down and rebuild.** Discards the store and reopens *only* the directories handed to it — no memory of what the user expanded. |
| `batch(add/remove)` | **Mutate in place.** Touches only the changed nodes; every other node keeps its expansion, selection, and scroll. |

The wrapper called `resetPaths` on every path change, so any rebuild — a watcher tick, a filter, a selection that was (wrongly) routed through the rebuild — collapsed the tree. **The fix switches to `batch`: diff the path inventory and apply the delta, so Pierre preserves the whole tree's state natively.**

### What changed

- **Selection no longer rebuilds the tree.** `selectedPath` was a tracked dependency of the rebuild effect; previewing a file rebuilt the whole tree. It's dropped from the deps — the selection effect reveals the picked row imperatively (`getItem(dir).expand()` for its ancestors, additively), covering external `path:line` navigation into a collapsed subtree.
- **Path changes apply as in-place mutations.** The effect diffs the new inventory against an `appliedPaths` ref (seeded at mount, updated after each `batch`) and applies `add`/`remove` ops. Pierre keeps expansion/selection/scroll for every untouched node. Search-projected and selected ancestors are opened additively afterward to reveal new or filtered rows.
- **The snapshot hack is gone.** An earlier iteration snapshotted every open directory before each `resetPaths` and fed it back; `batch` makes Pierre preserve state natively, so that code (and its `isExpanded()` probing) was deleted.

> _`appliedPaths` is tracked by hand rather than via `on`'s `prevInput`, which is `undefined` on the first post-`defer` run and would have dropped the first delta's removals — a bug caught in review and verified with a red/green e2e._

### One intrinsic limit

A folder whose *every* file is filtered out of the projection still folds away — Pierre infers directories from path prefixes, so once the rows are gone there's no node to keep open. This is true of any approach; everything that **stays in view** keeps its expansion.

### Tests

Two browse-mode regressions, each red before the fix:
- **Preview** — expand two sibling dirs, preview a file in one, assert the other stays open.
- **Filter → clear** — expand two dirs, filter to one, clear, assert the surviving dir is still expanded (collapses without in-place preservation).

### Try it locally

```sh
nix run github:juspay/kolu/bug-colla
```

### DIFF
--- a/packages/solid-pierre/src/FileTree.tsx
+++ b/packages/solid-pierre/src/FileTree.tsx
@@ -230,10 +230,13 @@ export const FileTree: Component<FileTreeProps> = (props) => {
             // nested file). Expanding each directory handle in place
             // preserves every other open folder; routing this through
             // `resetPaths` would rebuild the tree and collapse the user's
-            // hand-expanded siblings. A `getItem` returning a directory
-            // handle exposes `expand()`; re-expanding an open folder is a
-            // no-op, and files/missing paths simply fall through the
-            // `"expand" in item` guard.
+            // hand-expanded siblings. `getItem` returns a directory-or-file
+            // handle union: `"expand" in item` is the narrowing — Pierre's
+            // `isDirectory()` returns a `true`/`false` literal but isn't a
+            // `this is` predicate, so it can't narrow, whereas the `in`
+            // check both compiles and probes for the exact capability we're
+            // about to call. Re-expanding an open folder is a no-op; a file
+            // or a missing path narrows away and is skipped.
             for (const ancestor of ancestorDirectoryPaths(path)) {
               const item = tree?.getItem(ancestor);
               if (item && "expand" in item) item.expand();
--- a/packages/solid-pierre/src/FileTree.tsx
+++ b/packages/solid-pierre/src/FileTree.tsx
@@ -154,21 +154,22 @@ export const FileTree: Component<FileTreeProps> = (props) => {
   // open in one call (Pierre's `FileTreeResetOptions.initialExpandedPaths`).
   // Tracking the inputs in the same effect means a paths-and-ancestors
   // swap lands atomically — no second effect, no ordering invariant
-  // between "rebuild tree" and "open ancestors". The selected path's
-  // ancestors are merged in too: when an external caller drives selection
-  // (e.g. a terminal `path:line` click resolving into a nested file),
-  // the parents must be expanded for the row to be visible. Pierre's
-  // public surface doesn't expose `expandDirectory` directly, so the
-  // expand-on-select is routed through this same `resetPaths` call.
+  // between "rebuild tree" and "open ancestors". `resetPaths` rebuilds the
+  // tree from scratch, so it can only ever open the directories named here;
+  // it has no memory of folders the user expanded by hand. That's why
+  // `selectedPath` is *not* a dependency: routing selection changes through
+  // here would rebuild the tree on every file click and collapse every
+  // manually-expanded sibling that isn't an ancestor of the new pick. The
+  // selection effect below expands the picked file's ancestors imperatively
+  // instead, leaving the rest of the tree's expansion state untouched. We
+  // still read `selectedPath` untracked so a paths/expandPaths-driven
+  // rebuild keeps the current selection's ancestors open.
   createEffect(
     on(
-      [
-        () => props.paths,
-        () => props.expandPaths,
-        () => props.selectedPath ?? null,
-      ],
-      ([paths, expandPaths, selectedPath]) => {
+      [() => props.paths, () => props.expandPaths],
+      ([paths, expandPaths]) => {
         safeApply(() => {
+          const selectedPath = props.selectedPath ?? null;
           const ancestors = selectedPath
             ? ancestorDirectoryPaths(selectedPath)
             : [];
@@ -223,6 +224,20 @@ export const FileTree: Component<FileTreeProps> = (props) => {
           };
           deselectOthers(path);
           if (path !== null) {
+            // Open the picked file's ancestors so the row is visible —
+            // an external caller can drive selection into a collapsed
+            // subtree (e.g. a terminal `path:line` click resolving into a
+            // nested file). Expanding each directory handle in place
+            // preserves every other open folder; routing this through
+            // `resetPaths` would rebuild the tree and collapse the user's
+            // hand-expanded siblings. A `getItem` returning a directory
+            // handle exposes `expand()`; re-expanding an open folder is a
+            // no-op, and files/missing paths simply fall through the
+            // `"expand" in item` guard.
+            for (const ancestor of ancestorDirectoryPaths(path)) {
+              const item = tree?.getItem(ancestor);
+              if (item && "expand" in item) item.expand();
+            }
             tree?.getItem(path)?.select();
             // `select()` marks aria-selected but doesn't move the
             // virtualizer; deep paths in large worktrees would stay
--- a/packages/solid-pierre/src/FileTree.tsx
+++ b/packages/solid-pierre/src/FileTree.tsx
@@ -45,30 +45,31 @@ export function ancestorDirectoryPaths(path: string): string[] {
   return out;
 }
 
-/** The directories the tree currently holds open. Pierre exposes no bulk
- *  "expanded paths" getter (only per-handle `isExpanded()`), so we
- *  reconstruct the candidate directory set from the path list the tree was
- *  last built with — every ancestor of every file — and ask each handle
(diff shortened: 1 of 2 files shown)


srid:

| 1 | Hickey | Two-site "ancestors open" invariant (resetPaths reads `selectedPath` untracked + selection effect expands ancestors) — Pierre-constrained essential complexity; the diff *fixes* the original single-site form that caused the bug | ⚠️ **No-op** |
| 2 | Hickey | `"expand" in item` duck-type guard lacks a domain name | Superseded by cross-validation (#5) |
| 3 | Hickey | Untracked read is convention-only — wrap in `untrack()` to mechanically prevent regression | ⚠️ **No-op** (rejected by Lowy cross-validation, #6) |
| 4 | Lowy | resetPaths vs selection effect boundary; untracked `selectedPath` read; `onMount` third write-site — all clean orthogonal ownership, Pierre-API-forced | ⚠️ **No-op** |
| 5 | Lowy | `"expand" in item` is an unnamed Pierre duck-type assumption — name the discriminant | Fixed in this PR |
| 6 | Hickey ✕-val | Replace `"expand" in item` with Pierre's published `isDirectory()` | ⚠️ **No-op** (infeasible — see below) |
| 7 | Lowy ✕-val | Reject Hickey #3: `on(deps, fn)` already runs `fn` untracked, so explicit `untrack()` is a duplicated receptacle / false seam | Fixed in this PR (rejected #3) |

### How cross-validation changed the outcome

Both lenses ran first-pass, then each audited the other's recommendations. Two first-pass "Fix in this PR" items were **overturned by cross-validation**, which is the mechanism working as intended:

- **Hickey's `untrack()` (3)** — Lowy's cross-validator noted SolidJS's `on([deps], fn)` wraps the callback in `untrack` already; an explicit `untrack()` on one read inside is a second encapsulator for the same "what triggers this effect" axis, signalling a coupling that isn't there. Rejected.
- **Hickey-cross's `isDirectory()` swap (6)** — plausible ("use the published discriminant"), but Pierre's `isDirectory()` returns a `true`/`false` *literal*, not a `this is FileTreeDirectoryHandle` predicate, so it doesn't narrow the `getItem` union. Verified empirically: `tsc` rejects it (`Property 'expand' does not exist on FileTreeFileHandle`). The `in`-operator check is the compiler-verified narrowing and probes for the exact capability the next line calls.

### Net change from the review

One commit landed (`refactor(lowy)`): the guard comment now **names** the `"expand" in item` narrowing and records *why* `isDirectory()` can't replace it — so a future reader doesn't re-try the swap. The structural decomposition itself (two orthogonal effects) was endorsed by both lenses as the correct fix.

### Hickey rationale

The pre-diff effect complected two reactive concerns — "rebuild on inventory change" and "reveal on selection change" — and that braiding *was* the bug: a selection change forced a rebuild that discarded hand-expanded state. Splitting them is a genuine de-complecting. The remaining two-site "ancestors open" invariant is essential complexity at the Pierre boundary (`resetPaths` is a destructive rebuild; `expand()` is additive — the two are non-composable from outside the library), not accidental fragmentation.

### Lowy rationale

The two effects now own genuinely orthogonal activity-volatility axes (file-inventory change vs. selection change), which eliminates the cross-axis blast radius. The untracked `selectedPath` read is load-bearing and correctly scoped: the rebuild effect never runs *because* selection changed, it only snapshots the current selection to keep its ancestors open through an inventory rebuild. Removing the read would regress (a search-projection rebuild hides the selected file); tracking it reintroduces the original bug.

srid:
## Evidence

Code-tab browse mode, clicking `beta/b1.txt` to preview it after manually expanding both `alpha` and `beta`. Captured on ephemeral `pu` boxes running clean `nix run` builds — `master` (before) vs `bug-colla` (after).

| Before (`master`) — bug | After (`bug-colla`) — fixed |
|---|---|
| `alpha` collapses; `alpha/a1.txt` and `alpha/a2.txt` vanish | `alpha` stays expanded alongside `beta` |
| before | after |

Before clicking `beta/b1.txt`, both directories are expanded in each box. After the click:
- **Before:** the manually-expanded sibling `alpha` collapses — its files disappear from the tree (the `alpha` row shows a collapsed chevron, only `beta/b1.txt` and `beta/b2.txt` remain visible).
- **After:** `alpha` remains expanded (`alpha/a1.txt`, `alpha/a2.txt` still listed) while `beta/b1.txt` is selected and previewing — every other open folder is preserved.

This is the user-visible behavior of the e2e regression `packages/tests/features/code-tab.feature` → *"File browser preserves sibling expansion when previewing a file."*

srid:

Ran `--from polish` (the fix was implemented in this session; polish→done ran the full review/verify/ship pipeline).

|------|--------|----------|-------------|
| hickey+lowy | ✓ | 9m 37s | ran with cross-validation; `untrack()` rejected (redundant with SolidJS `on()`), `isDirectory()` swap rejected (no union narrowing, confirmed by tsc); discriminant comment landed as `refactor(lowy)` |
| police | ✓ | 3m 6s | code-police rules / fact-check / elegance — all clear |
| test | ✓ | 2m 25s | code-tab + mobile-code-browse: 66 scenarios / 796 steps passed, incl. new sibling-expansion regression |
| create-pr | ✓ | 46s | draft PR opened via forge-pr; Hickey/Lowy analysis posted |
| ci | ✓ | **12m 26s** | justci: all 22 nodes green on x86_64-linux + aarch64-darwin against HEAD `d0e6fc33`; ephemeral `pu` box destroyed cleanly |
| evidence | ✓ | 9m 23s | before/after screenshots on ephemeral `pu` boxes (master vs bug-colla) |
| **Total** | | **38m 16s** | |

**Slowest step**: `ci` (12m 26s)

### Optimization suggestions

- **`ci` dominated at 12m 26s** — almost entirely cold `nix` build + full 22-node e2e on a fresh `pu` box. For re-runs after a small push, `/do --from ci-only` skips straight to verification, and a warm linux box (or reusing `kolu-pr-1074` instead of recreating) would cut the cold-cache copy time that filled the first ~3 minutes.
- **`hickey+lowy` (9m 37s)** ran four sub-agent invocations (two first-pass + two cross-validation). That depth paid off here — cross-validation caught two plausible-but-wrong first-pass fixes (`untrack()`, `isDirectory()`) — so it was worth the spend on a structural-decomposition change, but it's the lever to trim on a trivial diff via `--minimal`.
- **`evidence` (9m 23s)** spun up two `pu` boxes for a before/after pair. When only the "after" behavior needs proving, a single box halves this; the dual-box cost is inherent to before/after comparisons.

Workflow completed at 2026-05-31T21:02Z.

srid:
## Evidence

**Bug:** In the Code tab's *browse* file tree, expanding directories by hand and then interacting with the tree — previewing a file, or filtering and clearing — collapsed the folders the user had opened. **Fix:** the full expansion state is now preserved across those interactions.

Both clips run the **same scripted interaction** on a clean, CI-like `nix run` build (off-machine, on ephemeral `pu` boxes): expand `alpha/` → expand `beta/` → preview `beta/b1.txt` → filter `a1` → clear the filter. Watch `alpha/`'s children (`a1.txt`, `a2.txt`).

| `master` (buggy) | `bug-colla` (fixed) |
|---|---|
| before | after |
| `alpha/` collapses on preview **and** on filter-clear — its files vanish | `alpha/` stays expanded throughout — both folders remain open |

▶ HD before: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=tree-state-before.mp4
▶ HD after: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=tree-state-after.mp4

Captured headless on a `pu` box (Playwright → Chromium, ffmpeg transcode), driving the real Pierre tree through the same selectors the e2e suite uses. The scripts also asserted state programmatically — `master`: `alphaAfterPreview=false, alphaAfterClear=false`; `bug-colla`: `alphaAfterPreview=true, alphaAfterClear=true`.

> _Clips were captured against the branch's user-visible behavior, which is identical under the final `batch`-based implementation (the mechanism changed; the preserved-tree outcome did not)._

PR #1078, opened 2026-05-31
feat(surface-nix-host): keep retrying unreachable hosts; recheck() for laptop wake

srid:
## Why

drishti runs on MacBooks. You close the lid at home and reopen at a café — and every monitored host lands in the terminal `failed` state, needing a manual *Reconnect* per host. This is the daily workflow, and reconnection actively fights it.

The root cause is in this package. `HostSession` treats **every** failure identically: 5 attempts (2/4/8/16s ≈ 30s) then give up into `failed`. That ceiling is correct for a **remote misconfiguration** (the parent's user isn't in `trusted-users`, signing fails) — retrying can't fix it, so failing loudly is right. It is exactly wrong for a **network fault**, which is all a roaming laptop ever has: the ~30s of fast retries burn out *before* the new Wi-Fi (captive portal → DHCP → DNS) is even usable.

## What changed

**1. Classify failures by phase, not by parsing error strings** (a new `FailureCause`):

| Cause | Sites | Policy |
|---|---|---|
| `"network"` | `.drv` resolver / arch-probe reject · ssh spawn error · connect-handshake timeout · live transport exit | **Never terminal** — retry at the capped (60s) backoff forever. Self-heals when the host answers. |
| `"remote"` | `nix copy` / `nix-store --realise` exited non-zero | Terminal after `MAX_CONSECUTIVE_FAILURES` (unchanged). |

`failureCause` now rides on `HostSessionState`, so consumers can say *why* a host is reconnecting ("host unreachable" vs "remote rejected the closure") instead of one undifferentiated "reconnecting…".

**2. New `recheck()` — the wake / network-change companion to `reconnect()`.** After a sleep, a `connected` link is often *lying*: the far end dropped the TCP socket, but the local ssh child won't notice until its keepalive fails ~30s later. `reconnect()` deliberately won't disturb a live link (it's the manual button, meaningful only from `failed`). `recheck()` force-cycles *whatever* is there — including `connected` — so a long-running parent can re-probe every host the instant it observes a wake, instead of waiting out ssh's keepalive.

## Behaviour change worth flagging

An unreachable-at-boot host previously degraded to terminal `failed`; it now keeps retrying as a `"network"` fault. This is intentional and is the headline fix — a host you can't currently reach (asleep, roaming) reconnects on its own when it's reachable again. `failed` is now reserved for genuine remote rejections. The existing "doesn't crash the parent at boot" property is preserved (the resolver rejection still flows through the reconnect machinery, never out of construction). The resolver-test was updated to assert the new non-terminal behaviour.

## Not done (deliberately)

The plan's `ServerAliveCountMax` tweak is already satisfied — the shared `SSH_OPT_PAIRS` pins it explicitly at `3`. No change needed.

## Tests

- `hostSession.test.ts` — unreachable host never reaches `failed`, sails past the old 5-attempt ceiling, carries `failureCause: "network"`.
- `recheck.test.ts` (new) — `recheck()` force-cycles a live `connected` child and reconnects; no-ops on an unreferenced session.
- Full suite: 22/22 green; `tsc` + biome lint clean.

## Downstream

drishti consumes this (`npins`). Its companion PR wires a parent-side wake detector + browser `online`/`visibilitychange` nudges to `recheck()`, surfaces `failureCause` in the connection UI, and pins to this branch (`$KOLU_SHA_PLACEHOLDER`) until this merges, then re-pins to master.


srid:
Downstream consumer: **srid/drishti#44** wires a parent-side wake detector + browser `online`/`visibilitychange` nudges to the new `recheck()`, surfaces `failureCause` in the connection UI, and pins drishti to this branch until it merges (then re-pins to `master`).

srid:

Ran on the combined cross-repo diff (this PR + srid/drishti#44). Both lenses spent most of their attention on *this* repo's changes (the `recheck()`/`reconnect()` split, `FailureCause` placement, the give-up gate). One Hickey finding, resolved to No-op by a Lowy cross-validation pass; Lowy's first pass found nothing. The police pass produced one elegance fix that landed here as commit `c3f3843`.

| # | Lens | Finding (kolu-relevant) | Disposition |
|---|------|-------------------------|-------------|
| 1 | Hickey | `recheck()` duplicates `reconnect()`? | ⚠️ **No-op** — distinct semantics/guards/tests; composition, not duplication |
| 2 | Hickey | `failureCause` modeled as a nullable field on `HostSessionState` conditioned on `connection` — recommended a discriminated union on `state` | ⚠️ **No-op** (cross-validated) |
| 3 | Lowy | `FailureCause` classification + give-up gate placement; `recheck()` home | ✅ Correct — belongs in kolu (only `spawn()` knows which phase failed) |
| 4 | Police | `this.clientPromise = this.spawn(); …catch(…)` repeated at 3 sites (`reconnect`, `recheck`, `scheduleReconnect`) — dry rule of three | ✅ **Fixed** — extracted `launchSpawn()` (commit `c3f3843`) |

### Hickey rationale

The duplication audit cleared `recheck()`: it has distinct documented semantics from `reconnect()` (force-cycles a live link vs. refuses to disturb one), distinct guards, and `recheck.test.ts` covers exactly the live-`connected` case `reconnect()` won't touch — composition, not a duplicated receptacle. The one finding was a low-severity fragmentation: `failureCause` is non-null exactly when `connection` is `disconnected`/`failed`, enforced by write-site discipline rather than the type. Hickey recommended encoding that as a discriminated union on `connection` in `HostSessionState`.

### Lowy rationale

`FailureCause` classification correctly lives here — only `HostSession.spawn()` has the evidence (resolver reject → `network`, `provisionAgent` non-zero → `remote`, child exit → `network`), and the give-up gate keying on it (`scheduleReconnect`) is in the right place; moving either to the consumer would invert the dependency. `recheck()` belongs on `HostSession` (it reaches into the child/`clientPromise` internals drishti can't touch). All boundaries volatility-based with stable interfaces.

### Cross-validation (Lowy on Hickey's union recommendation) → No-op

Converting `HostSessionState` to a discriminated union on `connection` **breaks the `updateState(patch: Partial<HostSessionState>)` merge mechanism** — you can't `{...prev, ...patch}` a partial like `{ connection: "connecting" }` onto a union and produce a well-typed variant, forcing a re-model of every `updateState` call site. It would also be **inconsistent with `lastError`**, an existing flat nullable carrying the identical "non-null when down" invariant, and would couple drishti's wire mirror to this module's state topology. The flat-plus-nullable shape is the deliberate, lower-coupling choice. Finding rejected on merit (not deferred).

srid:
**CI: green.** Ran the full justci pipeline on this branch (`c3f3843`), x86_64-linux lane redirected to a reachable host (the configured `vanjaram` is down).

- **x86_64-linux**: every recipe succeeded — `nix`, `typecheck`/`unit`, `biome`, `fmt`, `smoke`, `home-manager`, `surface-example-build`, `install`, `pnpm-hash-fresh`, and **e2e (377/377 scenarios)**.
- **aarch64-darwin**: every recipe succeeded; `e2e` initially reported 3/377 flaky failures (`codex.feature:20`, `opencode.feature:15`, `claude-code.feature:30` — agent-tile UI scenarios, untouched by this PR and green on x86_64-linux). A clean re-run of `ci::e2e@aarch64-darwin` passed **377/377**, confirming flakiness.

No failures attributable to this change.

srid:
**Addressed @Codex's review** (`859d39f`) + merged latest `master`:

| # | Finding | Fix |
|---|---------|-----|
| P1 | `recheck()` strands a pre-child failure forever (stale rejected `clientPromise`, no timer) | `scheduleReconnect` now clears `clientPromise` at the top for **every** backoff path, so "no spawn in flight ⇒ null" holds and the re-arm guards stay correct |
| P1 | all `provisionAgent` failures classified terminal `"remote"`, but `nix copy`/`realise` are ssh-backed | `ProvisionResult` carries a structured `cause`: ssh exit 255 or a connection-error stderr line → `"network"` (retry), else `"remote"` (bounded) |
| P2 | every child exit was `"network"` → a broken agent retries forever | classify by phase: live-link drop / ssh-255 → `"network"`; pre-first-RPC command exit or handshake timeout → `"remote"` (bounded) |
| P3 | README said deferred unreachable resolution degrades to `failed` | updated to the retry-indefinitely lifecycle |

`FailureCause` + the network-error heuristic moved to `host.ts` (shared by `provisionAgent` and `HostSession`).

**CI green** (linux lane → localhost; `vanjaram` still down): **all 22 nodes succeeded, e2e 378/378 on both platforms** — the earlier darwin e2e flakes did not recur. Tests: 25/25 (+3: strand regression, network-class provisioning, bounded broken-agent exit).

srid:
**Full CI green at `d86a20b`** (all Codex rounds addressed): **all 22 nodes succeeded, e2e 378/378 on both platforms** (linux lane → localhost; `vanjaram` still down). Unit tests 27/27.

Final Codex round folded in: a `recheck()` cycle mid-`connecting` now schedules a `"network"` retry (via a `cyclingForRecheck` flag) so a wake-recovery never burns the bounded give-up budget; plus the stale-comment cleanup.

PR #1079, opened 2026-05-31
Extract @kolu/solid-markdown from the intent surface

srid:
**The marked-based Markdown→SolidJS renderer now lives in its own app-agnostic package** — lifted out of `client/src/intent/IntentMarkdown.tsx`, which shrinks from 257 lines to a 30-line adapter. Behaviour for today's only consumer (the intent surface) is unchanged; the renderer is now reusable, which is the point: it's the first "renderer appliance" the planned Code-tab **Source ⇄ Rendered** preview will plug in.

This is **Phase 1** of `docs/plans/code-browser-preview.solid-fileview.html` — invent the reusable pieces first, so later phases plug in rather than wire bespoke rendering into the right-panel.

### Shape

`@kolu/solid-markdown` exposes one `<Markdown>` with a `variant` that bundles parse mode + styling scale:

| variant | parse | scale | used by |
| --- | --- | --- | --- |
| `inline` | inline-only | token-level (borrows `compact`) | annotation slots — title bar, dock rows, switcher cards, sub-panel tabs |
| `compact` | block | chat/dock | intent body + editor |
| `document` | block | reading | _the future Code-tab preview_ — no consumer yet |

`IntentMarkdown.tsx` keeps the two intent-facing wrappers (`IntentMarkdownBlock`/`IntentMarkdownInline`) as thin shims pinning variant + link policy, so the ~8 call sites are untouched.

Safety carries over verbatim: `marked` (GFM) + a `safeHref` allowlist, and `html` tokens render as **escaped text, never markup** — so the renderer is safe to point at untrusted file contents (which the Code-tab preview will do).

> _Tailwind v4 note: the renderer's utility classes moved with it, so `index.css` gained an `@source "../../solid-markdown/src"` directive. Verified the package-only classes (`.67rem`/`.85em`/`.8rem`) appear in the built CSS — intent styling is byte-for-byte preserved._

### Structural review refinements

The hickey/lowy + cross-validation pass (analysis in a comment below) landed four refinements as their own commits — notably threading parsed heading `depth` through `Styles.headingFor(depth)` instead of discarding it, so the `document` variant can grow real per-level sizes later without an API break.


srid:

Both lenses reviewed `git diff origin/master...HEAD`, then a cross-validation pass had each lens audit the other's recommendations. Notably, **each cross-validator overturned one of the other's first-pass calls** — which is exactly what the pass is for.

| 1 | Hickey | **F1** — link-default policy duplicated in both the package and the intent adapter | Fixed in this PR |
| 2 | Hickey | **E1** — a `links`-only change can't re-render anchors; the reactive interface overpromises | ⚠️ **No-op** — documented the limitation; cross-val rejected the proposed fix |
| 3 | Hickey | **P3** — `Styles` parameterized every element except blockquote/list/hr | Fixed in this PR |
| 4 | Lowy | `"compact"` is a Kolu spatial referent inside an app-agnostic enum | ⚠️ **No-op** — framing corrected in a comment; the rename itself is structurally inert |
| 5 | Lowy | `"inline"` silently inherits `compact` token styles (undocumented) | Fixed in this PR |
| 6 | Lowy | `"document"` renders headings flat despite its reading-scale promise | Fixed in this PR |
| 7 | Lowy | `MarkdownVariant` bundles parse-mode × styling scale | ⚠️ **No-op** — deterministic, no demand to split |
| 8 | Lowy | Is `transcript-html`'s markdown a parallel receptacle? | ⚠️ **No-op** — distinct (HTML SSR strings vs reactive Solid nodes) |
| 9 | Lowy | Package coherence / Surface test | ⚠️ **No-op** — one concept, one primitive; sound |

### Cross-validation (the two overturns)

- **Hickey → Lowy:** Lowy parked the flat-heading issue behind a TODO (No-op). Hickey's lens flagged that as *complecting* — `renderBlock` discarded the parsed `token.depth` at the only seam a consumer could reach it. So #6 was promoted to a **Fix**: `heading` became `headingFor(depth)`, threading depth through (output still flat today, but the per-level seam is open without an API break).
- **Lowy → Hickey:** Hickey's E1 fix was to make the `tokens` memo depend on `props.links`. Lowy's lens flagged that as **braiding two volatility axes** — parse output and link-render policy — forcing a full re-lex on a policy-only toggle. E1 was therefore *not* applied that way; instead the limitation is documented (no consumer drives `links` reactively today).

### Hickey rationale

The extraction does **not** reinvent the `solid-pierre` pattern, and shouldn't: `solid-pierre` wraps imperative DOM-class lifecycles, while this wraps a pure lexer→JSX walk — forcing a shared shape would complect unrelated concerns. Findings were about local entanglement: a duplicated default (F1), an interface that advertises reactivity it can't honor (E1), an incomplete style receptacle (P3), and — surfaced in cross-validation — semantic heading depth thrown away at its only accessible seam.

### Lowy rationale

`@kolu/solid-markdown` encapsulates four independent change axes (parser, sanitization allowlist, token-walk, styling scale) behind a stable prop surface — interface-stability holds, and the thin intent adapter correctly keeps Kolu's variant/link *policy* on Kolu's side. The prior-encapsulation check cleared (`transcript-html` is a different output medium). The live tensions were naming/contract honesty: a Kolu-referent variant name in an app-agnostic package, an undocumented `inline`→`compact` fallthrough, and a `document` variant whose behavior contradicted its stated reading-scale purpose.

srid:
## Evidence

This refactor extracts the marked-based Markdown→SolidJS renderer out of `IntentMarkdown.tsx` into the new `@kolu/solid-markdown` package, leaving `IntentMarkdown.tsx` a thin adapter. It is **behaviour-preserving** — so the proof is that intent markdown still renders correctly in-app *through the new package* (no visual diff vs `master` by design).

Captured on an ephemeral `pu` box running a clean `nix run` of this PR's own commit (`9ed7baf5`) — the `SRV`/`PTY` chips in the header both read `9ed7baf`. A terminal **intent** was set whose body uses an emoji, a `**bold**` span, an `` `inline code` `` span, and a second paragraph.

**Full app — title-bar annotation slot + workspace switcher card, both rendering through the extracted package:**

full app

**Workspace switcher card — full markdown body rendered (bold heading + `inline-code` chip + second paragraph):**

switcher card

`**solid-markdown**` renders **bold** and `@kolu/solid-markdown` renders as an inline-code chip — *not* raw `**`/`` ` `` characters — confirming the extracted renderer is wired up. The annotation slot (title bar) renders line-1 through `IntentMarkdownInline` from the same package. The DOM was verified directly: the slot's innerHTML contained `<strong class="font-semibold">solid-markdown</strong>` and `<code …>IntentMarkdown.tsx</code>`. No page errors during capture.

> _The leading emoji shows as a missing-glyph box because the headless capture host has no emoji font installed — a capture-host font artifact, unrelated to the renderer._

srid:

Ran `--from polish` after a manual implement; all gates green, PR marked ready.

| sync → check | ✓ | (manual) | implement done by hand: @kolu/solid-markdown extracted, IntentMarkdown 257→30 LOC, typecheck+lint green |
| branch | ✓ | — | `refactor/extract-solid-markdown` off origin/master |
| docs | ✓ | — | README package-inventory row added; surface README + website unaffected |
| fmt + commit | ✓ | 6s | primary feature commit pushed |
| hickey+lowy | ✓ | 2m 3s | review + cross-validation; 4 Fix findings as separate commits; E1 + compact-rename overturned by cross-validation |
| police | ✓ | 9m 51s | rules/fact-check/elegance clean; 2 elegance commits; caught a false reactivity comment the elegance pass introduced |
| test | ✓ | 1m 6s | e2e `terminal-intent.feature` 7/7 scenarios, 36/36 steps |
| create-pr | ✓ | 1m 31s | draft #1079 + Hickey/Lowy analysis comment |
| ci | ✓ | **19m 37s** | justci strict, all required checks green both platforms (after a fix + a flake retry) |
| evidence | ✓ | — | `## Evidence` posted — pu-box capture shows intent markdown rendering through the new package (DOM-verified) |
| **Total** | | **~58m** | |

**Slowest step**: `ci` (19m 37s) — two extra cycles: a real `ci::nix` failure, then a darwin flake.

### Optimization suggestions

- **`ci::nix` failed on the first run** because `default.nix`'s `src` fileset enumerates workspace members explicitly and the new `packages/solid-markdown` wasn't added — local `pnpm`/`vite`/`tsc` all passed, so only the sandboxed build caught it. **When adding a workspace package, update the `default.nix` fileset in the same commit** (it also gates the Nix typecheck) to avoid a ~10-min failed CI cycle.
- **`ci::unit@aarch64-darwin` was a flake** (SIGKILL/OOM in `integrations/pty`, passed on linux + on retry) — filed on #320. A positional `run ci::unit@aarch64-darwin` retry cleared it; for re-runs use `--from ci-only`.
- **`police` (9m 51s) dominated the non-CI time** — the elegance pass is the long pole; its only fixes here were minor (ctx memo, index clarity), so on tightly-scoped refactors `--minimal` would skip it for a faster loop.

PR #2, opened 2026-03-19
feat: phase 1 — one terminal in the browser

srid:
image

PR #1084, opened 2026-06-01
kolu-tui: list/snapshot your terminals from the shell (R-4 Phase 1)

srid:
**kolu-server now serves its in-process pty-host over a unix socket, and a new `kolu-tui` CLI lists and snapshots your live terminals from the shell — no browser.** It's the *raw*, terminal-side client of the same pty-host the browser drives over the full contract, and R-4 Phase 1 of the [kolu-tui plan](docs/atlas/src/content/atlas/pty-daemon-tui.mdx). The web path is byte-identical: **one PTY host, two transports.**

```
  browser  ──ws── kolu-server ──directLink (no wire)───────┐
                                                            ├── one in-process pty-host
  kolu-tui ──unixSocketLink ── serveOverUnixSocket ─────────┘    (servePtyHost router)
```

### The CLI (`@kolu/pty-tui`)

Read-only this phase — `attach` / `spawn` / `kill` are later phases. The CLI comes and goes; kolu-server keeps owning the PTYs. Same CLI framework as kolu-server (cleye), table rendering via columnify.

| Command | Behaviour |
|---|---|
| `kolu-tui list [--json]` | print your live terminals — **id · pid · idle · cmd · cwd** (`cmd` = the OSC title, else the foreground command); `--json` carries the full entry incl. raw title + foreground process |
| `kolu-tui snapshot <id>` | dump a terminal's current rendered scrollback to stdout, then exit |

`--pty-host-socket <path>` points at a non-default server; an unreachable socket is an honest one-line error (`ECONNREFUSED`/`ENOENT` → "is kolu-server running?"), never a silent hang. Packaged as `nix run github:juspay/kolu#kolu-tui` and auto-installed by the home-manager module. See the **Evidence — video** comment for a recording.

### Upstreamed into `@kolu/surface` (mirror: srid/drishti#57)

The transport work was generic, so it lives in the surface library — completing the link family (websocket · stdio · direct · **unix-socket**):

- **`@kolu/surface/unix-socket`** — `serveOverUnixSocket` (outcome-based, *never-rejecting* socket serving: dir-privacy gate, live-peer probe, stale-inode clearing that refuses to unlink anything not proven a dead socket) + `getRuntimeSocketPath` (the `$XDG_RUNTIME_DIR/<app>` / `/tmp/<app>-$UID` rendezvous convention).
- **`@kolu/surface/links/unix-socket`** — `unixSocketLink`, the dialing client half.
- **`serveOverStdio` no longer rejects** when a peer's read stream errors — it resolves with `{ reason: "end" | "error" }`. A rejecting serve promise was an unhandled-rejection crash footgun for multi-peer hosts (it bit kolu-server twice during this PR's review); now the no-crash path is the default for every consumer.
- **`isContractVersionCompatible`** in `@kolu/surface/define` — the generic `major.minor` handshake predicate; `isPtyHostContractCompatible` delegates.

kolu's `socketPath.ts` / `serveOverSocket.ts` are now thin wrappers that map transport outcomes to kolu-voiced operator log lines. _Per the surface-sharing rule, srid/drishti#57 adapts drishti and pins it to this branch to prove API compatibility._

### How it fits together

- **`@kolu/pty-host`** gains `createInProcessPtyHost` — builds the host once and returns its `client` (the no-wire `directLink` web client) **and** a `servedRouter` (the contract-wrapped form serving needs), so one host backs both transports and can never be instantiated twice. `getPtyHostSocketPath` is the single resolver server and CLI share.
- **Crash-safe by construction**: the socket is *additive* — kolu-server's web path doesn't depend on it — so every bind failure (a lost `EADDRINUSE` race when parallel servers share the default socket, an unwritable runtime dir) degrades to a **logged no-op, never a rejection**. _(This one bit the e2e harness, where many servers share `$XDG_RUNTIME_DIR`; caught in CI, hardened here.)_
- **Full-metadata `list`**: the `terminal.list` entry was enriched with `title` (OSC 0/2) + `foregroundProcess` (additive · optional, **contract 2.1**), so a one-shot `list` shows the `cmd` column without per-row tap subscriptions.
- **`kolu-server`** instantiates the pty-host once (`ptyHost.ts`) and adds one additive socket listener in `index.ts`; `local.ts` now consumes the shared client. Nothing about the web path changes.

> **Socket path decision** (the plan deferred it to Phase 1): a *stable* `$XDG_RUNTIME_DIR/kolu/pty-host.sock`, falling back to a fixed `/tmp/kolu-$UID/` off systemd — deliberately **not** `os.tmpdir()`, whose `$TMPDIR` differs by launch context on macOS (launchd server vs `nix run` CLI), so server and CLI would land on different sockets and never meet. Single-server-per-session model; `--pty-host-socket` on both sides to run more than one.

### Tests & docs

- A **real unix-socket round-trip** at both layers (generic in `@kolu/surface`'s `unix-socket.test.ts` + the pty-host contract round-trip in `serveOverSocket.test.ts`), the never-rejects regression pin for `serveOverStdio`, data-loss refusal tests (regular file / unprobeable socket), rendezvous-path `$TMPDIR`-independence pins, render-helper coverage, and e2e `smoke` + `terminal` confirming the web path is unaffected.
- README + `packages/surface/README.md` synced (the link family + unix-socket transport reference). The kolu-tui *announcements* were removed from README/website — it isn't ready for users yet; contributor-facing architecture rows remain.
- **Docs migration rode along**: the kolu-tui plan and the remaining `docs/plans/*.html` were ported to Atlas notes (`pty-daemon-tui`, `pty-daemon`, `remote-terminals`, `pty-daemon-chrome-bar`); `docs/plans/` is retired.

### Notes

- Structural review (hickey + lowy, with cross-validation) ran post-implement; the full `/be-review` gauntlet (codex-debate → lens-debate → code-police) ran on the pre-upstreaming shape — see the review comments.

### Try it locally

```sh
# in one shell: a server (the socket appears once it boots)
nix run github:juspay/kolu/r4-phase1-kolu-tui-list
# in another: list its live terminals
nix run github:juspay/kolu/r4-phase1-kolu-tui-list#kolu-tui -- list
```

_Generated by [`/do`](https://github.com/srid/agency) + [`/be`](https://github.com/srid/agency) on Claude Code (model `claude-fable-5`)._


srid:

Post-implement structural review of `git diff origin/master...HEAD`, with a cross-validation pass (each lens audited the other's recommendations). The cross-validation reshaped two findings, so the dispositions below differ from the first pass.

| 1 | Hickey | `createInProcessPtyHostClient` is dead public API (no production callers; host-discarding footgun) | Fixed in this PR |
| 2 | Hickey | `servePtyHostOverUnixSocket` braids socket lifecycle with the fragment→contract router-wrap | Fixed in this PR |
| 3 | Hickey | `Connection<C>` duplicated between `pty-tui` and `mini-ci` | ⚠️ **No-op** |
| 4 | Hickey · x-val | the `any`-cast complects in-process-vs-over-wire router | Fixed in this PR |
| 5 | Hickey · x-val | `cmdSnapshot` fires a 2nd RPC (`terminal.list`) just to decorate the stderr trailer | Fixed in this PR |
| 6 | Lowy | 6 new boundaries reviewed — each encapsulates a genuine volatility axis | ⚠️ **No-op** |

### Hickey rationale

The new `@kolu/pty-tui` package **extends** rather than reinvents the existing stdio-client pattern (the `connect` + pure `render` + `parseArgs`-based `main` split mirrors the surface examples). Three first-pass findings: a dead public-API alias (**#1**); the socket helper doing the `implement(contract).router(fragment)` wrap internally instead of at the call site like the mini-ci/rpm exemplars (**#2**); and a `Connection<C>` shape duplicated with mini-ci (**#3**).

### Lowy rationale

**Zero "Fix" findings** — the §1 prior-encapsulation survey found each new home correctly placed: `socketPath.ts` is a *distinct* volatility axis from `koluRoot.ts` (stable rendezvous vs per-pid ephemeral); `serveOverSocket.ts` + `socketPath.ts` belong in `@kolu/pty-host` because both server and CLI must share them (and Phase B's daemon reuses them); the single-instantiation seam belongs in the *server* (`ptyHost.ts`), not the library; `pty-tui`'s connect/render/main is three clean independent axes.

### Cross-validation (the interesting part)

Running each lens over the *other's* recommendations caught real cross-effects:

- **Lowy vetoed Hickey #2's call-site wrap** — moving `implement(contract).router(...)` into `kolu-server/index.ts` would *fragment* the wrap across the server **and** the future Phase-B daemon, and leak transport internals into the orchestration layer. **Synthesis:** hoist the wrap into `createInProcessPtyHost` as a `servedRouter` field — kept in `@kolu/pty-host` (Lowy), done once at host construction, and `serveOverSocket.ts` becomes pure socket-transport with no contract import and no cast (resolves #2 **and** Hickey-x-val **#4**).
- **Lowy vetoed Hickey #3** — the two `Connection` structs aren't duplicates: mini-ci's carries a `session` (nix-host reconnect lifecycle) `pty-tui` doesn't have. *Same kind, different domain.* `stdioLink<C>` is already the shared receptacle. → **No-op**, kept local.
- **Hickey surfaced #5** (the `cmdSnapshot` double-RPC) — fixed minimally by deriving the trailer (id + line count) from the snapshot already in hand, no contract ripple.

**Commits:** `dadfe77c` (#1) · `34f7ee40` (#2+#4 synthesis) · `a08421cd` (#5). All suites stayed green through each.

srid:
## Evidence

Captured on an ephemeral pu box (`kolu-pr-1084`) against the PR-built `kolu-tui` + `kolu-server` (both `nix run github:juspay/kolu/r4-phase1-kolu-tui-list`).

**1. Shipped binary runs** — `kolu-tui --help`:
```text
kolu-tui — a terminal-side client for kolu-server's pty-host (beta)

Usage:
  kolu-tui list [--json]       list your live terminals
  kolu-tui snapshot <id>       print a terminal's current scrollback, then exit
  kolu-tui help                show this help

Options:
  --socket <path>   pty-host socket (default $XDG_RUNTIME_DIR/kolu/pty-host.sock)
  --json            machine-readable output (list)
  -h, --help        show this help

kolu-tui connects to a running kolu-server over a local unix socket. Start the
server first (e.g. `nix run github:juspay/kolu`); the socket appears once it boots.
```

**2. Honest error, no server** — `kolu-tui list` with nothing listening (exit 1):
```text
kolu-tui: no pty-host socket at /run/user/1000/kolu/pty-host.sock (ENOENT) — is kolu-server running? the socket appears once it boots.
```

**3. PR-built server boots** — `GET /api/health` returns `kolu`, log shows the socket listening:
```text
INFO: kolu listening {"version":"0.1.0","node":"v22.22.1","address":"http://127.0.0.1:7681"}
INFO: pty-host socket listening (kolu-tui) {"socketPath":"/run/user/1000/kolu/pty-host.sock"}
```

**4. list (empty)** — server up, no terminals:
```text
no live terminals.
```

**5. Spawn one real terminal** over the pty-host socket (Phase 1 `kolu-tui` is read-only; spawned via the surface contract inside the devshell). Prints the new id:
```text
f6ffbc0e-6edf-4a2d-863a-350a8b8ff61b
```

**6. list (populated)** — `kolu-tui list` now sees the live PTY:
```text
ID                                       PID  IDLE  CWD
f6ffbc0e-6edf-4a2d-863a-350a8b8ff61b  111594   12s  ~
```

**7. list --json**:
```text
[
  {
    "id": "f6ffbc0e-6edf-4a2d-863a-350a8b8ff61b",
    "pid": 111594,
    "cwd": "/home/toor",
    "lastActivity": 1780276864584
  }
]
```

**8. snapshot &lt;id&gt;** — live shell scrollback on stdout (ANSI-stripped here for readability; the raw stream is genuine VT, e.g. `\e[33;1m…` and a `\e[?2004h` bracketed-paste enable), with the `— <id> · N lines` trailer on stderr:
```text
toor in 🌐 kolu-pr-1084 in ~
⬢ [Incus] ❯
— f6ffbc0e-6edf-4a2d-863a-350a8b8ff61b · 3 lines
```

Teardown: server killed, `/api/health` refuses connections, and the pty-host socket is removed — clean shutdown.

_Every step ran against the nix-packaged binaries on an off-machine box, exactly like CI. No output fabricated._

srid:

Entry point: `--from polish` (implementation, branch, and primary commit predated this run).

| hickey+lowy | ✓ | **20m 27s** | hickey 3 findings (1 applied, 2 via cross-validated synthesis); lowy 0 fix; cross-validation vetoed call-site wrap + Connection hoist (No-op), surfaced any-cast + double-RPC; 3 commits |
| police | ✓ | 7m 50s | 3 passes: rules 1 (watcher-lifecycle-logs), fact-check 0, elegance 1 (help alias); 2 commits |
| test | ✓ | 1m 3s | e2e smoke+terminal 15/15 (web path unaffected by local.ts refactor); new behaviour covered by socket round-trip + render unit tests + live CLI e2e |
| create-pr | ✓ | — | draft PR #1084 + hickey/lowy analysis comment _(step-start was missed, so it's absent from the timing total)_ |
| ci | ✓ | 11m 43s | justci all 22 nodes green on HEAD `5e1a17fd` (both platforms); 20 required contexts pass |
| evidence | ✓ | 5m 4s | `## Evidence` — nix-packaged kolu-tui on a pu box: help, honest error, list/--json/snapshot of a real PTY |
| **Total** | | **~49m** | |

**Slowest step**: `hickey+lowy` (20m 27s)

### Optimization suggestions

- **`hickey+lowy` was 42% of the run.** The cross-validation pass earned its keep here (it vetoed two of Hickey's three first-pass recommendations and reshaped the router-wrap fix into the `servedRouter` synthesis), so it's worth keeping — but the full `vitest run` re-verification after *each* review commit overlapped with the later `test`/`ci` steps. Targeted typechecks per review-commit + one suite at the end would trim several minutes.
- **The first CI box landed with broken egress** (`000`/timeout). The probe-first guard in `.agency/do.md` caught it before a wasted ~12m run — destroy+recreate cost ~1m. The guard worked; no change needed.
- **For any re-run**, `--from ci-only` re-runs just CI against HEAD — the draft PR and its comments persist — skipping the ~30m of review/police/test that already landed as commits.

Workflow completed.

srid:
## Evidence — video



▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=kolu-tui-list.mp4

_The nix-packaged `kolu-tui` (`nix run github:juspay/kolu/r4-phase1-kolu-tui-list#kolu-tui`) connecting to a live kolu-server on an ephemeral box: `list` shows the new full-metadata table (ID · PID · IDLE · CMD · CWD — three terminals running `top` / `vim` / a shell), `list --json` exposes `title` + `foregroundProcess`, and `snapshot` prints a terminal's live scrollback._

srid:

The **x86_64-linux** lane ran on **`kolu-ci-5`** (`idliv2-02`) — commit `1e2963a7`, **exit 0** ✓

- **Lane wall** (pipeline): **9m42s**
- **Wrapper wall** (incl. lease + nix-run startup): 30m1s

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 8m50s | ✓ |
| `ci::home-manager` | 1m0s | ✓ |
| `ci::smoke` | 42s | ✓ |
| `ci::nix` | 40s | ✓ |
| `_ci-setup` | 29s | ✓ |
| `ci::pnpm-hash-fresh` | 28s | ✓ |
| `ci::unit` | 26s | ✓ |
| `ci::install` | 23s | ✓ |
| `ci::biome` | 16s | ✓ |
| `ci::fmt` | 15s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 round(s) · codex reviewed at `xhigh` reasoning effort · base `eaf172de0bf9`

### Round 1

**codex** — approved: `false`

The core socket/TUI shape is reasonable, and narrow typechecks pass for `@kolu/pty-tui`, `@kolu/pty-host`, and `kolu-server`. I did not run unit tests because this review was explicitly read-only and the new tests create temp files/PTYs. I found one real data-loss risk in socket startup, plus a few smaller API/docs/test gaps.

Findings:
- `F1` · major · open — `isSocketLive()` collapses every connection error to `false`, and the caller then unconditionally `rmSync`s `socketPath`. If `--pty-host-socket` points at an existing regular file in a private directory, or connect fails for `ENOTSOCK`/`EACCES`/other non-stale cases, kolu will delete that file before binding. That is a data-loss bug, not just stale-socket cleanup. (packages/pty-host/src/serveOverSocket.ts:115)
- `F2` · minor · open — The package root no longer re-exports `createInProcessPtyHostClient`, even though the function still exists and `package.json` only exposes the root subpath. That is a source API regression for any package-root consumer, and the relative imports in this package's tests will not catch it. (packages/pty-host/src/index.ts:37)
- `F3` · minor · open — `kolu-tui --help` documents the default socket only as `$XDG_RUNTIME_DIR/kolu/pty-host.sock`, but the actual default on macOS/non-systemd is `/tmp/kolu-$UID/pty-host.sock`. The design doc also says "tmpdir() fallback" at `docs/plans/remote-terminals.pty-daemon.tui.html:275`, which contradicts the code's deliberate non-`os.tmpdir()` fallback. (packages/pty-tui/src/main.ts:31)
- `F4` · minor · open — The contract changed `terminal.list` to carry `title` and `foregroundProcess`, and `kolu-tui list` depends on those fields, but the host/contract tests still only assert id/pid/cwd behavior. A regression that dropped the new metadata at the surface boundary would not be caught. (packages/pty-host/src/inProcessPtyHost.test.ts:92)

**claude** — Addressed all four CODEX findings in the worktree at /home/srid/code/kolu/.worktrees/modest-runner. F1 was a genuine data-loss bug and is now fixed with a socket-only unlink guard plus a regression test; F2/F3/F4 were valid minors and are all fixed. Verified: pty-host (62) and pty-tui (9) unit tests pass, pty-host/pty-tui/kolu-server typechecks pass, biome format + lint clean on all touched files.

- `F1` **fixed** — Agreed — real footgun. `--pty-host-socket` is an arbitrary user path, and `isSocketLive()` returning false for ENOTSOCK on a regular file would lead the caller to rmSync the user's data. Added `isRemovableStaleSocket(path)` in serveOverSocket.ts (lstatSync(path).isSocket()): only a socket inode (a dead peer's stale file) or ENOENT is removable; a regular file/dir/symlink → warn + return the no-op listener, never unlink. Used lstat (not stat) so a symlink is classified by itself, not followed. Wired the guard immediately before the `rmSync` at the former line 115. Added a falsifiability test in serveOverSocket.test.ts ('refuses to delete an existing regular file at the socket path') that writes a regular file, points the socket at it, and asserts the file and its contents survive both the call and listener.close().
- `F2` **fixed** — Conceded — the root export dropped `createInProcessPtyHostClient` while the function lived on, reachable only via the package's own relative test imports (so the narrow typecheck never caught the dead export). Rather than re-export an API nothing consumes, I took CODEX's second option and removed the helper outright: it is fully superseded by `createInProcessPtyHost` (which `server/src/ptyHost.ts` already uses to build the single shared host). Updated inProcessPtyHost.test.ts to construct via `createInProcessPtyHost(...).client` (and renamed the two describe blocks + header accordingly), and fixed the now-stale doc comment in server/src/terminalBackend/local.ts:13 that referenced the removed symbol. grep confirms zero remaining `createInProcessPtyHostClient` references in packages/. (The two mentions in docs/plans/remote-terminals.pty-daemon.html are historical Phase-A/B planning prose untouched by this branch, so out of scope for this finding.)
- `F3` **fixed** — Agreed — the help text and plan note understated the real default. Updated kolu-tui's HELP in pty-tui/src/main.ts so --socket documents BOTH defaults ($XDG_RUNTIME_DIR/kolu/pty-host.sock on systemd Linux, else /tmp/kolu-$UID/pty-host.sock). Also corrected the tui design doc (docs/plans/remote-terminals.pty-daemon.tui.html:275) which said 'tmpdir() fallback' — replaced with the accurate fixed $TMPDIR-independent /tmp/kolu-$UID/ fallback and the one-line reason it is deliberately NOT os.tmpdir() (launch-context divergence). Chose documenting both literals over rendering getPtyHostSocketPath() since the help is a static string and both branches are now explicit.
- `F4` **fixed** — Agreed — the contract added title/foregroundProcess to terminal.list and kolu-tui list depends on them, but no test asserted they survive the surface boundary. Added a contract-level test in inProcessPtyHost.test.ts ('surfaces title + foregroundProcess on terminal.list') that spawns a real shell through the contract, writes an OSC 2 title sequence followed by `sleep 5` (so the shell prompt doesn't clobber the title before the read and foregroundProcess reflects a live process), polls `client.surface.terminal.list({})` until the async title tap propagates, and asserts entry.title === the set title and typeof entry.foregroundProcess === 'string'. Covers the in-process client path; the socket round-trip in serveOverSocket.test.ts already exercises terminal.list over a real net.Socket. Verified green (7/7 in that file).

commit: `49e87c6bcac20cad8a9c586fb674a23c96b43a01`

### Round 2

**codex** — approved: `false`

Claude fixed F2, F3, and F4 adequately. F1 is materially improved for regular files/symlinks/directories, but it still collapses every socket connect error into "not live" and then unlinks any socket inode, so the non-stale socket-error part of the original finding remains open.

Findings:
- `F1` · major · open — The fix prevents deletion of regular files, dirs, and symlinks, but `isSocketLive()` still turns every connect error into `false`. The caller then removes the path whenever `lstatSync(path).isSocket()` is true, including unexpected socket errors such as `EACCES`. That still violates the original requirement to remove only stale sockets, not arbitrary socket paths that merely failed the probe for another reason. (packages/pty-host/src/serveOverSocket.ts:59)
- `F2` · minor · resolved — Previously the package root dropped `createInProcessPtyHostClient` while the helper still existed. Claude removed the helper and migrated package tests/current code to `createInProcessPtyHost(...).client`, so there is no dangling live API to export. (packages/pty-host/src/index.ts:37)
- `F3` · minor · resolved — The help text and TUI design doc now document both socket defaults and correctly describe the `/tmp/kolu-$UID` fallback as `$TMPDIR`-independent rather than `os.tmpdir()`. (packages/pty-tui/src/main.ts:31)
- `F4` · minor · resolved — The new contract-level test drives a title through the client and asserts `terminal.list` exposes `title` plus a present `foregroundProcess`, covering the metadata surface boundary that `kolu-tui list` depends on. (packages/pty-host/src/inProcessPtyHost.test.ts:113)

_codex on the rebuttal:_ F1: holding firm. The regular-file data-loss path is fixed, but the original finding also called out EACCES/other non-stale probe errors. The current implementation discards the probe error code, so any socket inode after any connect failure is removed. F2: accepted. Removing the helper is a valid resolution for the live API issue; I'm not keeping this open on the historical planning prose. F3 and F4: accepted as fixed.

**claude** — Round 2: only F1 remained open (codex held firm; F2/F3/F4 already resolved last round). I conceded codex's remaining point and closed the gap. The round-1 fix stopped data loss for regular files/dirs/symlinks, but isSocketLive() still collapsed every connect error to false, so a real socket inode that failed the probe for a non-stale reason (EACCES/EPERM/etc.) would still be rmSync'd. I replaced the boolean probe + post-hoc inode check with a three-way verdict probeSocket(): SocketProbe — live (never touch), stale (ECONNREFUSED or ENOENT → free to bind), unknown (any other connect error → could-not-prove-dead → no-op, never delete). The caller only reaches rmSync on a stale verdict AND isSocketInodeOrAbsent() confirming a socket/absent inode. Added a sharper falsifiability test that binds a real socket, chmod 0o000's the socket file (verified empirically this yields EACCES on connect while lstat still sees a socket), and asserts the inode survives — a case the round-1 logic would have unlinked. pty-host 63/63 tests pass (+1), tsc clean, biome format+lint clean.

- `F1` **fixed** — Conceded codex's holdout. The non-stale socket-error path was real: isSocketLive() returned false for EACCES/EPERM, then isRemovableStaleSocket() saw a socket inode (lstat succeeds when only the socket FILE's perms, not the parent dir, are stripped) and the caller rmSync'd a socket it never proved dead. Fix in packages/pty-host/src/serveOverSocket.ts: introduced a discriminated SocketProbe ('live' | 'stale' | 'unknown') and probeSocket() that maps ECONNREFUSED+ENOENT -> stale (free to bind) and EVERY other connect error -> unknown (degrade to no-op, never delete). Caller (servePtyHostOverUnixSocket) now: live -> noop; unknown -> warn+noop (new branch); stale -> only then check isSocketInodeOrAbsent() (renamed from isRemovableStaleSocket) before rmSync. So a socket that fails the probe for a non-stale reason is preserved by the unknown short-circuit. Note: ENOENT had to stay in the stale set (it is the common fresh-start case) or nothing would ever bind — caught and fixed when the round-trip tests regressed to ENOENT. Added falsifiability test 'refuses to delete a real socket inode it could not probe (EACCES, not stale)' in serveOverSocket.test.ts: binds a real net.Server socket, chmod 0o000 on the socket file (empirically yields EACCES on connect, ECONNREFUSED avoided), points --pty-host-socket at it, asserts the inode survives bind+close. This case is unlinked under round-1 logic, preserved now. Guarded with getuid()===0 skip (root bypasses unix perm checks).

commit: `6400c850cb9eed8c0e88bebe6aa277d1d772eac6`

### Round 3

**codex** — approved: `true`

The remaining F1 concern is now addressed in the current worktree: the socket probe preserves the distinction between live, stale, and unknown, and deletion is gated behind both a stale probe verdict and an inode check. F2, F3, and F4 remain resolved from the prior round. I did not run tests because this review round is read-only, but I verified the relevant code and regression coverage by inspection.

Findings:
- `F1` · major · resolved — Previously, non-stale socket probe errors could still lead to unlinking a socket inode. The current implementation uses `SocketProbe` with `live`, `stale`, and `unknown`; only `stale` proceeds to `isSocketInodeOrAbsent()` and `rmSync`, while `unknown` warns and returns a no-op listener. (packages/pty-host/src/serveOverSocket.ts:148)
- `F2` · minor · resolved — Previously the package root dropped `createInProcessPtyHostClient` while the helper still existed. Claude removed the helper and migrated current code/tests to `createInProcessPtyHost(...).client`. (packages/pty-host/src/index.ts:37)
- `F3` · minor · resolved — The CLI help and TUI design doc now document both socket defaults and accurately describe the `/tmp/kolu-$UID` fallback. (packages/pty-tui/src/main.ts:31)
- `F4` · minor · resolved — The contract-level test now verifies `terminal.list` exposes the metadata fields that `kolu-tui list` depends on. (packages/pty-host/src/inProcessPtyHost.test.ts:113)

_codex on the rebuttal:_ F1: accepted. The new `probeSocket()` verdict preserves the exact distinction I asked for: `ECONNREFUSED`/`ENOENT` are stale/free-to-bind, accepted connections are live, and all other errors are unknown and never deleted. The additional EACCES socket-inode regression test covers the previously open hole. F2/F3/F4 were already accepted as resolved and remain resolved.

srid:

✅ **Consensus** after 2 round(s) · lowy + hickey · base `eaf172de0bf9`

### Applied (3)
- `lowy-3` CLI flag --socket vs server flag --pty-host-socket: asymmetric naming for the same axis — commit `c5195a014`
- `lowy-4` servedRouter typed `any` leaks the contract-wrap seam to the consumer module — commit `5243811ab`
- `hickey-1` Listed-terminal modeled as two parallel types kept in sync by convention — commit `f1befcaa5`

### Agreed — no change (8)
- `lowy-1` Socket-path privacy axis split across two packages (packages/pty-host/src/socketPath.ts:36 and packages/pty-host/src/serveOverSocket.ts:47)
- `lowy-2` Socket-dial logic duplicated between connect.ts and probeSocket (packages/pty-tui/src/connect.ts:26 and packages/pty-host/src/serveOverSocket.ts:79)
- `lowy-5` terminal.list enrichment correctly encapsulates the snapshot-vs-stream axis (packages/pty-host/src/ptyHostSurface.ts:110-121 and packages/pty-host/src/ptyHost.ts:673-678)
- `lowy-6` home-manager tuiPackage default lives in the flake, not the module (flake.nix:47-54 and nix/home/module.nix:40-52)
- `hickey-2` `optional()` on title/foregroundProcess encodes a presence rule the in-process host always satisfies (packages/pty-host/src/ptyHostSurface.ts:119)
- `hickey-3` Two runtime-root path policies derive the same "per-user XDG-or-fallback dir" concept differently (packages/pty-host/src/socketPath.ts:36 and packages/server/src/koluRoot.ts:19)
- `hickey-4` Foreground-process basename derived in two places by two methods (packages/pty-tui/src/render.ts:22)
- `hickey-5` Privacy invariant for the socket dir is split across path-derivation and binding modules (packages/pty-host/src/socketPath.ts:23 and packages/pty-host/src/serveOverSocket.ts:140)

srid:

Final pass (simplify + fact-check) over the kolu-tui diff — **4 fixes, 0 skipped** (commit `c604f883`). All settled tree, after codex⇄claude and lowy⇄hickey.

| Fix | File | What |
| --- | --- | --- |
| **Reuse** | `pty-tui/render.ts` | `commandName` uses `node:path.basename` instead of a hand-rolled `split("/").pop() ?? p`; renamed the param off the `process` global shadow |
| **Correctness** | `pty-tui/main.ts` | `writeOut` now also resolves on `error` — `kolu-tui list \| head -1` sends EPIPE; the drain Promise would otherwise hang / unhandled-reject |
| **Diagnostics** | `pty-tui/main.ts` | `assertCompatible` threads the underlying error message through, so the real cause (`ECONNRESET`, oRPC text) reaches the user instead of a generic "could not read version" |
| **Efficiency** | `pty-host/serveOverSocket.ts` | the `ECONNREFUSED`/`ENOENT` set is a module-level `FREE_TO_BIND` constant, not re-allocated on every `probeSocket` call |

Verified: `just check` clean, pty-host 63/63 + pty-tui 9/9 green, biome formatted.

_Closes the `/be-review` gauntlet (codex-debate → lens-debate → code-police)._

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `f87765843929`

### Round 1

**codex** — approved: `false`

The extraction is directionally sound, and the socket transport has useful safety tests. I found one real local-IPC hardening gap plus a couple of smaller API/contract precision issues in the new shared surface code. I did not run the socket tests because this review environment is read-only and those tests create temp files/sockets.

Findings:
- `F1` · major · open — The directory privacy check uses `statSync`, which follows symlinks. That means another local user can pre-create the stable `/tmp/<app>-$UID` rendezvous directory as a symlink to any existing owner-private directory for the victim. The check then passes, but the path component in `/tmp` is still attacker-owned and can be replaced later, letting the attacker redirect future clients to a different socket. This undercuts the security invariant the comments rely on for the full-control PTY socket. (packages/surface/src/unix-socket.ts:87)
- `F2` · minor · open — The `not-a-socket` outcome is not reliably returned for non-socket paths. The code classifies any non-`ECONNREFUSED`/`ENOENT` probe error as `probe-failed` before checking the inode, while the comments explicitly call out `ENOTSOCK` on regular files as an unknown probe error. On platforms that return `ENOTSOCK`, a regular file is reported as an unexpected probe failure instead of the machine-readable `not-a-socket` outcome. (packages/surface/src/unix-socket.ts:222)
- `F3` · minor · open — `isContractVersionCompatible` parses versions with an unanchored `/^(\d+)\.(\d+)/` regex, so malformed strings like `2.1garbage` are accepted as compatible even though the doc says only patch/prerelease suffixes are tolerated. Since this is now the shared process-boundary handshake helper, accepting invalid version strings weakens the fail-fast behavior it is meant to provide. (packages/surface/src/define.ts:672)

**claude** — Addressed all three CODEX findings in packages/surface (plus a tightened pty-host test). F1 and F2 were real hardening/contract gaps in the unix-socket transport; F3 was a real regex anchoring gap. All agreed and fixed, with regression tests. surface unix-socket tests: 15 passed; pty-host contract tests: 8 passed. Ran `just fmt`.

- `F1` **fixed** — Agreed — `isPrivateOwnedDir` (unix-socket.ts:84-97) used `statSync`, which follows symlinks, so an attacker-owned `/tmp/<app>-$UID` symlink pointing at any owner-private dir would pass the privacy check while leaving the attacker owning the /tmp path component (swappable later to redirect clients). Switched to `lstatSync` and now require `st.isDirectory()` in addition to uid-owned + no group/other bits, so any symlink (or non-dir inode) is rejected outright. Dropped the now-unused `statSync` import. Added a regression test in unix-socket.test.ts that pre-creates the socket dir as a symlink to a genuinely 0700 owner-private dir and asserts the `dir-not-private` refusal with no socket created. mkdirSync(recursive) is a no-op against the pre-existing symlink target so it doesn't throw before the check runs.
- `F2` **fixed** — Agreed — when `probeSocket` returns `{kind:"unknown"}` (which the comments note includes ENOTSOCK on a regular file), serve refused with the opaque `probe-failed` before any inode classification, so a regular file at the path could surface as `probe-failed` instead of the machine-readable `not-a-socket`. Added `isNonSocketInode(path)` (lstat-based) and, in the `unknown` branch (unix-socket.ts:~230), return `{kind:"not-a-socket"}` when lstat proves a non-socket inode; a truly unclassifiable inode (ENOENT or lstat failure) stays `probe-failed`. Tightened the existing regular-file test from `outcome.kind !== "listening"` to assert the exact `{kind:"not-a-socket"}` outcome, covering both the ENOTSOCK-via-lstat route and the stale-then-inode-guard route.
- `F3` **fixed** — Agreed — `isContractVersionCompatible` (define.ts:671) parsed with the unanchored `/^(\d+)\.(\d+)/`, so `2.1garbage` parsed as 2.1 instead of being rejected, weakening the fail-fast handshake. Replaced with the anchored grammar `^(\d+)\.(\d+)(?:\.\d+)?(?:-[0-9A-Za-z.-]+)?$` (major.minor with an optional patch/prerelease suffix only). Added regression cases in ptyHostSurface.test.ts (the thin wrapper that calls this helper) asserting `2.1garbage`, `2.0garbage`, `2.0.1.2`, and `2.0 ` are all incompatible; the existing valid-suffix cases (`2.0.1`, `2.0.0-rc1`) still pass.

commit: `bd2a73c93cc7abdd946cf43bef19477f47c5b462`

### Round 2

**codex** — approved: `true`

Verified the current fixes for F1-F3 against the changed files. The Unix socket directory check now rejects symlinks via `lstatSync`, non-socket path outcomes are precise, and contract version parsing is anchored with regression coverage. I also checked the unstaged `peer-server.ts` change shown by `git status`; it is formatting-only around the existing promise chain and introduces no new regression. I did not rerun tests in this read-only review environment.

Findings:
- `F1` · major · resolved — The previous directory privacy check followed symlinks, allowing an attacker-controlled rendezvous path component to pass the owner/private check via a symlink target. (packages/surface/src/unix-socket.ts:95)
- `F2` · minor · resolved — The previous unknown-probe branch reported regular files as `probe-failed` on platforms returning `ENOTSOCK`, instead of the intended `not-a-socket` machine-readable outcome. (packages/surface/src/unix-socket.ts:245)
- `F3` · minor · resolved — The previous contract-version parser used an unanchored regex, so malformed suffixes could be silently truncated to `major.minor` and accepted. (packages/surface/src/define.ts:674)

_codex on the rebuttal:_ No disputes from CLAUDE. I agree with all three fixes after inspection and mark F1, F2, and F3 resolved.

srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `f87765843929`

### Applied (2)
- `hickey-1` The not-a-socket verdict is computed by two inverted lstat predicates — commit `14853e280`
- `hickey-3` refusalWarning's exhaustive switch re-projects the outcome union the wrapper already destructures — commit `0a2106034`

### Agreed — no change (7)
- `lowy-1` Rendezvous helper co-located with the serve half, not the transport-neutral seam (packages/surface/src/unix-socket.ts:56 (getRuntimeSocketPath))
- `lowy-2` Asymmetry: serveOverUnixSocket never rejects, unixSocketLink does reject (packages/surface/src/links/unix-socket.ts:30 vs packages/surface/src/unix-socket.ts:214)
- `lowy-3` serveOverStdio outcome change is the right receptacle generalization (packages/surface/src/peer-server.ts:94 (ServeOverStdioEnd))
- `lowy-4` isContractVersionCompatible generalized into surface; pty-host delegates — correct layering (packages/surface/src/define.ts:667 and packages/pty-host/src/ptyHostSurface.ts:64)
- `hickey-2` isPrivateOwnedDir mixes the Windows-no-uid bypass into the inode-shape check (packages/surface/src/unix-socket.ts:92-97)
- `hickey-4` Two PtyHostSocketListener shapes for one concept (surface listener vs kolu listener) (packages/pty-host/src/serveOverSocket.ts:24-30, 80-91)
- `hickey-5` console.log monkey-patch in serveOverStdio braids logging policy into the serve lifecycle (packages/surface/src/peer-server.ts:141-151)

srid:

Reviewed the surface-upstreaming diff (base `f8776584`) after the codex and lens debates settled it.

**Fixed (1):**
- **Deleted the `isPtyHostContractCompatible` pass-through** — the wrapper added no pty-host semantics over `@kolu/surface/define`'s `isContractVersionCompatible` and had a single caller; kolu-tui now imports the generic predicate directly, and the duplicated generic version-grammar tests were dropped from `ptyHostSurface.test.ts` (the pty-host self-compatibility pin stays). Commit `70475df0`.

**Checked, not violations:**
- `classifyInode` called twice in `unix-socket.ts` — mutually exclusive probe branches (`unknown` vs `stale`), not repeated work.
- `describeRefusal`'s native exhaustive `switch` — each arm extracts variant-specific fields; the ts-pattern/Record conventions don't apply (and pty-host carries no ts-pattern dep).
- The two `closed` guards (surface listener vs kolu wrapper) — different responsibilities (file removal vs the close log line), not duplication.

Verified after the fix: pty-host 53/53, pty-tui 9/9, `tsc --noEmit` clean on both.

srid:

The **x86_64-linux** lane ran on **`kolu-ci-5`** (`idliv2-02`) — commit `4e807db9`, **exit 0** ✓

- **Lane wall** (pipeline): **9m8s**
- **Wrapper wall** (incl. lease + nix-run startup): 30m31s

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 8m28s | ✓ |
| `ci::home-manager` | 32s | ✓ |
| `_ci-setup` | 27s | ✓ |
| `ci::unit` | 24s | ✓ |
| `ci::pnpm-hash-fresh` | 24s | ✓ |
| `ci::biome` | 15s | ✓ |
| `ci::fmt` | 14s | ✓ |
| `ci::smoke` | 13s | ✓ |
| `ci::install` | 13s | ✓ |
| `ci::nix` | 11s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

srid:

The **x86_64-linux** lane ran on **`kolu-ci-5`** (`idliv2-02`) — commit `c20b3b74`, **exit 0** ✓

- **Lane wall** (pipeline): **9m37s**
- **Wrapper wall** (incl. lease + nix-run startup): 30m11s

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 8m44s | ✓ |
| `ci::home-manager` | 1m0s | ✓ |
| `ci::smoke` | 45s | ✓ |
| `ci::nix` | 42s | ✓ |
| `ci::atlas-sync` | 39s | ✓ |
| `_ci-setup` | 28s | ✓ |
| `ci::unit` | 25s | ✓ |
| `ci::pnpm-hash-fresh` | 25s | ✓ |
| `ci::install` | 25s | ✓ |
| `ci::biome` | 16s | ✓ |
| `ci::fmt` | 15s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

PR #1075, opened 2026-05-31
Scope mobile swipe to the tile, keep the keyboard down on dock dismiss, equalize the handles

srid:
On mobile, three touch annoyances around the fullscreen tile and its drawers. **Swiping across the soft key bar now stays put instead of cycling terminals, dismissing a drawer no longer pops the soft keyboard, and the top pull-handle grip matches the left dock handle's size.**

### The three fixes

| Symptom | Cause | Fix |
| --- | --- | --- |
| A finger drag across the soft key bar cycled to another terminal mid-type | The key bar sits inside the swipe wrapper, so its touch events bubbled up to the tile-cycle handler | The bar stops `touchstart` propagation — the same guard the pull/dock handles already use |
| Tapping the backdrop to close the dock popped the soft keyboard | Corvu's `Drawer` defaults `restoreFocus={true}`, re-focusing the terminal textarea that was active before the drawer opened | `restoreFocus={false}` on every mobile drawer |
| The top drag handle's grip looked much smaller than the left one | Top grip was `w-10 h-1` (40×4 px); left grip is `h-16 w-2` (64×8 px) | Top grip enlarged to `w-16 h-2` — same 64×8 footprint, rotated 90° |

### Structural-review follow-ups

The `restoreFocus` fix surfaced a **latent twin bug**: `RightPanelDrawer` (the mobile bottom-sheet inspector) carried no focus guard either, so backdrop-dismissing it would pop the keyboard the same way. It's now aligned with the dock and chrome drawers. The two drawers' duplicated `snapPoints={[0,1]}` Corvu `#977` workaround was also pulled into a named `CORVU_SNAP_WORKAROUND` const — kept _separate_ from `restoreFocus`, which is a soft-keyboard policy on a different volatility axis, not a library workaround.

> _The two e2e regression guards initially passed even with their fixes reverted — both were vacuous (one read xterm's canvas-backed `.xterm-screen` as text; the other raced ahead of Corvu's async focus restore). They're now verified to fail without their fix._

### Try it locally

```sh
nix run github:juspay/kolu/mobile-swipe
```

_Generated by [`/do`](https://github.com/srid/agency) on Claude Code (model `claude-opus-4-8`)._

PR #1086, opened 2026-06-01
Keep the mobile soft keyboard down on dismiss, scroll, and link taps

srid:
On a touch device the OS soft keyboard should rise from exactly one thing: an explicit tap on plain terminal content, meaning "I want to type here." **Several gestures were summoning it with no such intent** — and one of them was also swallowing the action it should have performed. The unifying question this PR answers is _"should this gesture raise the keyboard, or pass through to the underlying action?"_ — and it puts that decision in one place per surface.

### The paths that wrongly raised the keyboard

| Symptom (on touch) | Cause | Fix |
| --- | --- | --- |
| Dismissing the dock / top-bar / inspector drawer left the keyboard up | Focus-trapping into a non-text overlay never blurs the terminal field on iOS, so [#1075](https://github.com/juspay/kolu/pull/1075)'s `restoreFocus={false}` alone could not drop it | Actively blur the focused field on close — `dismissSoftKeyboard` (a synchronous pass inside the dismiss gesture + one deferred `rAF` pass to clear Corvu's post-paint focus-trap teardown) |
| Tapping the floating scroll-to-bottom button to catch up | `onClick` called `terminal.focus()` unconditionally | Route through the existing `focusOnSelection()` — a no-op on touch, still focuses on desktop |
| Closing the command palette / diagnostics / a confirm dialog | `refocusTerminal()` `.click()`s the terminal to restore "keep typing" focus | `isTouch()` no-op; the dialog also blurs its own inputs on close |
| **Tapping a `path:line` link** (e.g. a plan `.html`) popped the keyboard **and never opened the link** | xterm's own link activation is mouse/hover-only and never fires for a touch tap, while the tap handler always called `term.focus()` | The terminal tap handler hit-tests the tapped cell against the file-ref parser and **follows the link into the Code tab** instead of focusing |

### Why [#1075](https://github.com/juspay/kolu/pull/1075) didn't fix the drawer case

`restoreFocus={false}` only stops Corvu re-focusing the terminal on close — but on iOS the keyboard never went down in the first place. Moving focus into a focus-trapped overlay does **not** reliably blur the text field underneath, so the keyboard lingers for the drawer's whole lifetime and the dismiss leaves it up. The only lever iOS treats as authoritative is an explicit `blur()`, so that's what we do now.

### The shape

The "what raises the keyboard" volatility ends up cleanly decomposed:

```
raise   ── the terminal tap handler, and only on plain content
            (a link tap follows the link; everything else is suppressed)
lower   ── dismissSoftKeyboard()  — one shared primitive
suppress── focusOnSelection() / refocusTerminal()  — isTouch() no-ops
```

`withKeyboardDismiss` wraps every Corvu overlay's `onOpenChange` (the three drawers _and_ `ModalDialog`) so "dismissing an overlay leaves the keyboard down" is one structural policy, not a per-overlay convention a new overlay could forget. `activateFileRef` is the single front door both the desktop link provider and the mobile tap handler route through.

### Notable

> Every fix is `isTouch()`-gated — **desktop behavior is unchanged everywhere** (the scroll FAB still refocuses, dialogs still restore terminal focus, hardware keyboards keep their focus through an overlay close).

Each new behavior ships a non-vacuous `@mobile` regression guard (verified to fail without its fix): drawer/dialog dismiss, scroll-to-bottom FAB, and the file-ref tap. The link-tap test latches the right-panel drawer's transient mount via a `MutationObserver` rather than fighting the test env's instant-transition override.

### Try it locally

```sh
nix run github:juspay/kolu/mobile-issues
```

### DIFF
--- a/packages/client/src/MobileTileView.tsx
+++ b/packages/client/src/MobileTileView.tsx
@@ -27,7 +27,7 @@ import MobileChromeSheet from "./MobileChromeSheet";
 import type { WsStatus } from "./rpc/rpc";
 import { TerminalMetaCompact } from "./terminal/TerminalMeta";
 import { useTerminalStore } from "./terminal/useTerminalStore";
-import { drawerKeyboardOnOpenChange } from "./ui/dismissSoftKeyboard";
+import { withKeyboardDismiss } from "./ui/dismissSoftKeyboard";
 
 /** Minimum horizontal travel (px) before a swipe commits to a tile change. */
 const SWIPE_THRESHOLD = 60;
@@ -72,8 +72,8 @@ const MobileTileView: Component<{
   // onOpenChange) and the in-sheet buttons (`onClose`, routed through
   // `handler(false)`) — funnels through these so the soft keyboard never
   // lingers on a touch device after the drawer goes away.
-  const onChromeOpenChange = drawerKeyboardOnOpenChange(setChromeOpen);
-  const onDockOpenChange = drawerKeyboardOnOpenChange(setDockOpen);
+  const onChromeOpenChange = withKeyboardDismiss(setChromeOpen);
+  const onDockOpenChange = withKeyboardDismiss(setDockOpen);
   // Pull-handle drag state for the chrome (top) drawer. Not reactive —
   // only the touch handlers read it.
   let pullStartY: number | null = null;
--- a/packages/client/src/MobileTileView.tsx
+++ b/packages/client/src/MobileTileView.tsx
@@ -27,7 +27,7 @@ import MobileChromeSheet from "./MobileChromeSheet";
 import type { WsStatus } from "./rpc/rpc";
 import { TerminalMetaCompact } from "./terminal/TerminalMeta";
 import { useTerminalStore } from "./terminal/useTerminalStore";
-import { dismissSoftKeyboard } from "./ui/dismissSoftKeyboard";
+import { drawerKeyboardOnOpenChange } from "./ui/dismissSoftKeyboard";
 
 /** Minimum horizontal travel (px) before a swipe commits to a tile change. */
 const SWIPE_THRESHOLD = 60;
@@ -68,18 +68,12 @@ const MobileTileView: Component<{
   } | null>(null);
   const [chromeOpen, setChromeOpen] = createSignal(false);
   const [dockOpen, setDockOpen] = createSignal(false);
-  // Close a drawer AND drop the soft keyboard. Every dismiss path — backdrop
-  // tap, drag-to-close (both via Corvu's onOpenChange) and the in-sheet
-  // buttons' onClose — funnels through these so the keyboard never lingers on
-  // a touch device after the drawer goes away. See `dismissSoftKeyboard`.
-  const closeChrome = () => {
-    setChromeOpen(false);
-    dismissSoftKeyboard();
-  };
-  const closeDock = () => {
-    setDockOpen(false);
-    dismissSoftKeyboard();
-  };
+  // Every dismiss path — backdrop tap, drag-to-close (both via Corvu's
+  // onOpenChange) and the in-sheet buttons (`onClose`, routed through
+  // `handler(false)`) — funnels through these so the soft keyboard never
+  // lingers on a touch device after the drawer goes away.
+  const onChromeOpenChange = drawerKeyboardOnOpenChange(setChromeOpen);
+  const onDockOpenChange = drawerKeyboardOnOpenChange(setDockOpen);
   // Pull-handle drag state for the chrome (top) drawer. Not reactive —
   // only the touch handlers read it.
   let pullStartY: number | null = null;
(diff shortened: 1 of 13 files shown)


srid:

Run on the full branch diff (both passes ran a second time after the file-ref-link work landed). Cross-validated: each Hickey fix was checked against Lowy's boundary lens, and Lowy independently re-examined the two refactors a prior pass had proposed.

|---|--------|------------------------------------------------------|---------------------|
| 1 | Hickey | Drawer close→dismiss policy enforced by convention across three per-drawer lambdas | Fixed in this PR |
| 2 | Hickey | `getLine → translateToString → guard → parseLineRefs` idiom duplicated between the hover provider and the tap hit-test | Fixed in this PR |
| 3 | Hickey | `fileRefAtPoint` comment claimed shared geometry with the touch-scroll handler (it doesn't) | Fixed in this PR |
| 4 | Lowy | Extract a `MobileDrawer` receptacle wrapping the three drawers | ⚠️ **No-op** |
| 5 | Lowy | Relocate `refocusTerminal` out of `ModalDialog` | ⚠️ **No-op** |

### Hickey rationale

The diff sits on genuinely separate volatility axes (soft-keyboard policy, file-ref activation, test infra) with no concern interleaving. Three small, local findings, all fixed as their own commits:

1. **Enforced by convention** → factored the close→dismiss wiring into `withKeyboardDismiss(onOpenChange)`, which all three drawers _and_ `ModalDialog` route through, so a new overlay can't silently skip the policy. (Commit `4b5d5657`, generalized to dialogs in `05b55357`.)
2. **Parse idiom dup** → extracted a private `lineRefsAt(terminal, bufferLine)` (commit `643f42af`).
3. **Misleading comment** → corrected to state why the tap (2D) and scroll (1D) geometries deliberately don't share a helper (commit `d64b39a7`).

Hickey also confirmed the prior pass's "centralize `refocusTerminal` into `ModalDialog`" idea was correctly rejected: the desktop refocus is genuinely per-dialog (the palette's conditional `anyDialogOpen` guard, `CloseConfirm` refocusing on cancel but not on close), so a blanket refocus would complect distinct policies.

### Lowy rationale

**No Fix-in-this-PR findings — the decomposition is sound.** The volatility ("what raises/lowers the soft keyboard") is real (the mechanism already changed once, from `restoreFocus={false}` to the two-pass blur) and is placed in one home each: the mechanism (`dismissSoftKeyboard`), the raise-decision (the terminal tap handler), and two suppress-guards (`focusOnSelection`, `refocusTerminal` touch no-ops). `dismissSoftKeyboard` is named for the concern, not a consumer.

Lowy explicitly **downgraded two candidate refactors to No-op**, agreeing with the author's reasoning and adding the trace:

- A `MobileDrawer` receptacle would draw a boundary around _surface similarity_ (three `<Drawer>` usages) rather than _co-varying change_ — it would have to parameterize `side`, `snapPoints`, overlay/content markup, and the close-callback shape, owning one genuinely-shared line. The shared mechanism is already behind `dismissSoftKeyboard()`; the per-drawer `closeX` wiring is the correct _consumer-side composition_ ("set my own open-state" + "drop the keyboard").
- Relocating `refocusTerminal` into the pure DOM-accessor module would _introduce_ mixed altitude (dragging the touch-policy guard + focus side-effect into a "find the node, decide nothing" module), not remove it.

srid:

| fmt | ✓ | 7s | clean, no reformatting |
| commit | ✓ | 6s | primary feature commit pushed |
| hickey+lowy | ✓ | 50m 18s* | hickey: 3 findings, each its own commit; lowy: 0 fix findings, validated the decomposition, downgraded the `MobileDrawer` receptacle + `refocusTerminal` relocation to No-op |
| police | ✓ | 9m 32s | all-clear across rules/fact-check/elegance; applied one elegance unification (`withKeyboardDismiss` across drawers + dialogs) |
| create-pr | ✓ | 1m 42s | draft PR + hickey/lowy analysis comment |
| ci | ✓ | 9m 48s | justci all 22 nodes green on an ephemeral `pu` box; `gh pr checks` all pass on HEAD `05b5535` (both platforms + SentinelOne scans) |
| evidence | — skipped | — | soft-keyboard up/down has no representation in off-machine headless capture; the observable proxy (focus state) + link-tap→Code-tab are covered by non-vacuous e2e guards |
| **Total** | | **~75m** | |

**Slowest step**: `hickey+lowy`

\* _Inflated: the step bracket stayed open across a mid-workflow scope expansion (a fourth bug — the terminal-link tap — was raised after the first review), so it includes that fix's full implement+test cycle, not just review time._

### Optimization suggestions

- **Scope expansions mid-run inflate the open step's timer.** The 50m `hickey+lowy` figure is mostly the link-tap fix implemented during a redirect while the step was held open. Re-entering `implement` (vs. holding the review step open) would keep per-step timing honest.
- **CI is the real recurring cost (~10m).** `ci::e2e@aarch64-darwin` was the long pole (6m5s). For re-runs after a tweak, `--from ci-only` skips straight to the green-gate.
- **Pre-read the focus/keyboard surfaces before a similar task.** Most of `research` was spent tracing Corvu's focus-trap + `restoreFocus` propagation; that map is now in the PR comment and commit messages for next time.

Workflow completed.

PR #1089, opened 2026-06-01
Validate persisted prefs; extract @kolu/log and @kolu/html-escape leaves

srid:
**Five hand-rolled persisted-preference sites now read through one validated receptacle.** `persistedPref` is a thin validate-on-read layer over `@solid-primitives/storage`'s `makePersisted`, whose off-the-shelf `deserialize` hands back whatever string sits in `localStorage` with no error path. Routing dock mode, activity window, font zoom, canvas-maximized, and comments through it closes two latent bugs — and the same Hickey/Lowy "electricity" review (`docs/electricity-review.html`) that surfaced them also carves out two zero-dependency leaf packages.

### Bugs closed by validate-on-read

| Pref | Before | After |
| --- | --- | --- |
| Font zoom | `Number("garbage")` → `NaN` font size | non-finite / ≤ 0 rejected → falls back to default |
| Canvas maximized | stored string `"false"` read back **truthy** → posture latched on once persisted | strict `"true"`/`"false"`; anything else falls back |

_The validate seam is single-mode: `parse` returns only for valid input and throws for everything else, so the `onInvalid` reporting path can never be silently skipped (a hickey/lowy finding, fixed in-PR)._

### Two leaf packages, decoupled from their domain homes

- **`@kolu/log`** — the `Logger` type in a zero-runtime-dependency leaf. `kolu-io` and `kolu-transcript-core` stop re-declaring byte-identical private copies; `kolu-shared` re-exports the one canonical contract.
- **`@kolu/html-escape`** — `escapeHtml` in a zero-dependency leaf. The app-agnostic `transcript-html` renderer now reaches it directly and **drops its `kolu-common` dependency entirely** — no longer dragging the agent/git/oRPC domain tree in to do five string replacements.

### Also

- `walkShadowRoots` relocated `comments/` → a neutral `dom/` folder (a `right-panel` consumer proved it had outgrown its feature home).
- The keybind-core's `isMac` is now an injectable parameter (defaulting to the detected platform) instead of a module-singleton read — making the chord-matching/formatting a pure function of platform, testable across both.

> _Reviewed by parallel hickey + lowy sub-agents and `/code-police` (3 passes); findings landed as follow-up commits — see the structural-analysis comment below._

### Try it locally

```sh
nix run "github:juspay/kolu?ref=feat/electricity-ship-now"
```


srid:

Both lenses ran as parallel sub-agents on the diff, then cross-validated each other's recommendation. They independently converged on the same primary finding; the lowy cross-validation pass added a second. No defers, no No-ops — both fixed in this PR.

| 1 | Hickey + Lowy | `persistedPref.parse` had two modes behind one `(raw) => T` signature — **throw** (fires `onInvalid` + fallback) vs **return fallback** (silently bypasses `onInvalid`). Three call sites used the silent path, so a future `onInvalid` on them would never fire with no error. | ✅ Fixed (`e440b818`) |
| 2 | Lowy (cross-validate) | The boolean `parse` forms an implicit contract with `defaultSerialize` (booleans serialize via `JSON.stringify` to `"true"`/`"false"`). A future boolean pref could naively use `Boolean(raw)` and re-introduce the `"false"`-reads-truthy bug. | ✅ Fixed (`d726d5b0`) |

### Hickey rationale

The codebase survey came back clean on the headline question — **all four new files extend canonical patterns rather than reinvent them** (concept de-duplication, not duplication): `persistedPref` is the receptacle five hand-rolled call sites each implemented four divergent ways; `@kolu/log` collapses three byte-identical private `Logger` declarations; `@kolu/html-escape` moves the single canonical `escapeHtml` to a leaf; `dom/shadowWalk.ts` relocates the only `walkShadowRoots` in the tree. The one finding was the dual-mode `parse` contract above. Cross-validation confirmed the throw-only fix does **not** complect error-signaling with validation — in this domain the thrown error *is* the validation result on the failure branch, and the `try/catch` in `readWithFallback` enforces the `onInvalid` contract structurally, not by convention.

### Lowy rationale

The prior-encapsulation survey found the diff **consolidates rather than fragments** on every boundary. `@kolu/log` collapses two private `Logger` re-declarations into one canonical leaf — the blast radius of a `Logger` shape change is now exactly one package, and the 14 `kolu-shared` consumers get the same type through the designed re-export chain. `@kolu/html-escape` lets `transcript-html` drop its entire `kolu-common` dependency with no stranded consumers. `persistedPref` is the first project-level receptacle for "validated persisted preference"; both new packages pass the single-export coherence test. The three throwing `parse` lambdas are **not** a fragmented validation policy — fallback/reporting policy is encapsulated once in `readWithFallback`, while per-type *acceptance* policy correctly co-locates with each type (centralizing it would invert the dependency, infra importing domain). Finding 2 is a documentation-volatility gap, not an architectural one: `JSON.stringify(false) === "false"` is ECMAScript-stable, so it warrants a one-line note in `defaultSerialize`, not a new `parseBoolean` boundary.

srid:
## ⚠️ `pu` misbehaved — Incus admin attention needed

Evidence capture for this PR could not run off-machine: every `pu create` for the evidence box landed on the **same** physical host, which has **fully broken outbound network** (DNS resolves, but raw-IP TCP and the gateway's :443 both time out). `nix run github:juspay/kolu?ref=feat/electricity-ship-now` can't fetch the flake, so the Playwright/ffmpeg capture is impossible until the host's egress is fixed. Destroy+recreate retried **4×** and each box re-landed on `idliv2-01` — placement is effectively pinned to the dead host.

- **PR:** #1089 &nbsp; **branch:** `feat/electricity-ship-now` &nbsp; **commit:** `5b91f961`
- **Stage:** `pu egress` (no outbound network) &nbsp; **box:** `kolu-pr-1089-evidence` &nbsp; **when:** 2026-06-01T16:46:27Z
- **Retries:** 4 create/destroy cycles, all landed on `idliv2-01`; egress HTTP code empty/`000` each time.

**Box placement (`pu list` — NAME + physical LOCATION that needs fixing):**
```
|         NAME          |      LOCATION       |
| kolu-pr-1089-evidence | idliv2-01           |
```
**Box network state (resolv.conf / routes / egress / raw-IP / gateway TCP):**
```
== /etc/resolv.conf ==
# Generated by resolvconf
domain incus
nameserver 10.47.48.1
nameserver fd42:988c:24e6:b499::1
nameserver fd42:1eae:9b9b:9b8d::1
options edns0
== ip route ==
default via 10.47.48.1 dev eth0 proto dhcp src 10.0.20.92 metric 1237 
10.0.0.0/8 dev eth0 proto dhcp scope link src 10.0.20.92 metric 1237 
== egress probe ==
egress FAILED (curl exit non-zero / timeout)
== raw-IP probe ==
raw-IP egress FAILED
== gateway TCP ==
gw 10.47.48.1:443 FAILED
```
The egress check the `pu` skill prescribes (`curl https://api.github.com`) returns exit 124 (timeout); raw-IP `https://140.82.112.6` and `/dev/tcp/10.47.48.1/443` both fail too — so it's not DNS, it's no L3/L4 egress from `idliv2-01`. **Action:** restore outbound routing/NAT on physical host `idliv2-01` (or drain it from the `pu` placement pool until fixed).

_Evidence is a bonus, not the gate — CI status checks remain the source of truth. This comment exists so the host gets fixed permanently instead of every run papering over it._

srid:

| sync | ✓ | 1s | forge=github |
| research | ✓ | 4m 29s | targets confirmed; capturePointerGesture already exported (no-op); new pkgs add zero external deps |
| branch | ✓ | 0s | feat/electricity-ship-now from origin/master |
| implement | ✓ | — | persistedPref (5 sites + test), @kolu/log + @kolu/html-escape leaves, shadowWalk→dom/, keybind isMac injection |
| check | ✓ | 3m 38s | `just check` EXIT 0 across all 31 projects |
| docs | ✓ | 55s | README package table updated |
| fmt | ✓ | 1m 0s | `just fmt` EXIT 0 |
| commit | ✓ | 28s | report + primary refactor pushed |
| hickey+lowy | ✓ | 13m 55s | parallel sub-agents converged on parse-contract finding (fixed) + cross-validation doc finding (fixed); no defers |
| police | ✓ | 9m 28s | 3 passes clean after 2 elegance fixes (single-pass escapeHtml, doc reflow) |
| test | ✓ | 4m 40s | unit 220/220 client + all packages; e2e 26/26 (keyboard, dock, activity) |
| create-pr | ✓ | 1m 54s | PR + hickey/lowy comment |
| **ci** | ✓ | **24m 28s** | all 20 justci contexts green on HEAD `5b91f961`; first linux red was a no-egress Incus host (`idliv2-01`), fresh box clean |
| evidence | ⚠️ | — | session-restore **video** blocked by the same no-egress host (`pu misbehaved` diagnostic filed); behavior covered by `ci::e2e` session-restore.feature, green both platforms |
| **Total** | | **76m 58s** | |

**Slowest step**: `ci` (24m 28s — three runs: one caught a real `default.nix` fileset omission, one hit the dead host, one clean).

### Optimization suggestions

- **Pre-check `default.nix`'s `src` fileset when adding workspace packages.** The first CI run failed because the two new package dirs weren't in the curated allowlist, so `vite build` couldn't resolve `@kolu/html-escape` — and `just check` can't catch it (it runs the full tree, not the filtered nix `src`). The fileset's own invariant comment warns about exactly this silent omission; adding the dirs up front saves a full CI cycle.
- **Re-run CI with `/do --from ci-only`.** Two of the three CI runs were re-runs (the fix and the dead-box retry); the `ci-only` entry point skips the 11 upstream steps.
- **Drain or repair Incus host `idliv2-01`.** Its broken egress cost a full CI run *and* the entire video-evidence capture; the diagnostic comment is filed for the admin. Until it's drained, ephemeral linux boxes will keep flaking.

Workflow completed at 2026-06-01 16:49 UTC.

srid:
## Evidence

Session restore survives a full page reload — the exact round-trip this PR's persisted-pref validation guards. Captured on a clean ephemeral pu box (`dev-x86-64-linux-09`) running a fresh `nix run` of this PR's own commit `5b91f96` (note the `SRV • 5b91f96` chip), driven headless via Playwright.

**The round-trip:** build a workspace of 3 distinguishable terminals (`/tmp`, `/etc`, `$HOME`, each with its own output) → collapse the dock to **rail** (`kolu-dock-mode`) → **maximize** the active tile (`kolu-canvas-maximized`) → **reload the page** → everything comes back.

**Verdict (asserted in-capture): PASS** — all three terminals restored with identical scrollback (same ids), dock mode persisted (`rail`), maximize posture persisted (`true`), and the active terminal preserved — a zero-visual-diff survives-restart proof.

▶ **HD video:** https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=1089-session-restore.mp4

session restore round-trip

| Before reload (workspace built) | After reload (session restored) |
| --- | --- |
| before | after |

PR #1096, opened 2026-06-01
Fix Code-tab filter leaving empty, result-less directories

srid:
**Filtering the Code-tab file tree no longer leaves a tree full of empty, result-less directories.** Searching `docs plans` used to keep `.claude/…`, `.codex/…`, `agents/`, `ci/` and the rest as hollow, expanded rows even though nothing under them matched. The wrapper filters by removing non-matching files through `@pierre/trees`' in-place `batch`, but **Pierre's `remove` promotes an emptied directory to an "explicit empty folder" rather than deleting it** — so the files vanished and the directories stayed.

### The mechanism

```
batch-remove non-matching files
        │
        ▼
parent dirs emptied ──▶ Pierre promotes each to EXPLICIT ──▶ row survives
        │                  (promoteEmptyAncestorsToExplicit, canonical.js)
        ▼
filtered tree = matches  +  every now-hollow directory
```

The wrapper reconciled *files* but never the *directories* those removals emptied. The fix adds `directoryRemovalOps(prev, next)`: a directory survives iff it is still an ancestor of some surviving file, and the rest are pruned as recursive removes of the maximal dead subtrees. It is derived purely from the two file inventories — no parallel directory state — and a cleared filter (which projects the *full* inventory, not an empty list) prunes nothing.

Two alternatives were weighed and rejected (both reviewers agreed): a *collapse pass* — unnecessary once the empty dirs are gone, since every surviving directory is a match ancestor that must stay open — and *`resetPaths`-while-filtering*, which would couple the search-agnostic wrapper to the client's search state. Full rationale in `docs/plans/code-tab-filter-empty-dirs.html`.

### Coverage

Pure unit tests for the derivation (cleared filter, nested dead subtree, partial-match dir, progressive narrowing) plus a **headless Pierre integration test** — it constructs a real `FileTree`, batch-removes files, and asserts via `getItem` that the stranded rows are gone and surviving match-dirs are untouched. This is the layer the old `fileSearch.test.ts` could not reach: it only tested the pure projection, never Pierre's `batch`, where the bug lived.

### Refinements during review (hickey / lowy / police)

| Lens | Finding | Fix |
| --- | --- | --- |
| Hickey | dir ops applied one-per-`batch` in a loop | single `tree.batch(dirOps)` — disjoint subtrees, order-independent |
| Lowy | comment said pruning is "the caller's job" — but the caller is this file | reworded |
| Hickey + Lowy | helper exported from the package surface only to reach a cross-package test | extracted the pure logic into a `pathReconcile` leaf, gave `solid-pierre` its own vitest harness, co-located the test, dropped the helper from the public index |
| Hickey (cross-validate) | integration test used a different `batch` shape than production | matched it |
| Police | "empty `next`" comment was wrong (a cleared filter is full-inventory `next`) | corrected |
| Police | O(n²) `slice+join` in `ancestorDirectoryPaths` | O(n) prefix accumulator |

### Try it locally

```sh
nix run github:juspay/kolu/search-collapsae
```


srid:

Two parallel reviewers ran cold on `git diff origin/master...HEAD`, then cross-validated each other's recommendations. Six findings, all dispositioned; every "Fixed" finding is its own commit.

| 1 | Hickey | Directory ops applied one-per-`batch` in a loop, asymmetric with the file pass | Fixed in this PR |
| 2 | Hickey + Lowy | `directoryRemovalOps` exported from the package surface (with a Pierre-op return type) solely to reach a cross-package test | Fixed in this PR |
| 3 | Lowy | `pathDiffOperations` comment said pruning is "the caller's job" — but the caller is the same file | Fixed in this PR |
| 4 | Hickey (cross-validate) | Integration test applied ops one-per-`batch`, no longer matching the single-batch production shape | Fixed in this PR |
| 5 | Lowy (cross-validate) | `ancestorDirectoryPaths` leaks through the Solid-adapter public surface into `fileSearch.ts` | ⚠️ **No-op** — pre-existing on `master`, unchanged in blast radius (the move into a Solid-free leaf is, if anything, a step toward the eventual fix) |

### Hickey rationale

The core algorithm is clean: `directoryRemovalOps` is a pure two-pass function over two path arrays — read-only loops, local `Set`s, no escaping mutation. It's correctly distinct from `pathDiffOperations` (file diff vs. derived-directory pruning). First pass flagged the one-op-at-a-time batching (cosmetic, **#1**) and the test homed in `client` away from its subject (**#2**). The convergent fix — extracting the pure helpers into a Solid/JSX-free `pathReconcile.ts` leaf and standing up an in-package vitest harness — was cross-validated as a *correct seam* (un-braids path-set reconciliation from the DOM-widget lifecycle) with *no config duplication* (the harness is the minimal config-less convention 4 sibling packages already use). Cross-validation then caught **#4**: the single-batch production refinement had left the tests on the old loop shape, so they no longer exercised the exact call path production takes.

### Lowy rationale

The decomposition is sound: the fix lands the Pierre-`remove`-quirk volatility *inside* the wrapper, leaving `fileSearch.ts` ("what should be visible") and `CodeTab.tsx` inert — and explicitly rejects the `resetPaths`-while-filtering alternative that would couple the search-agnostic wrapper to client search state. First pass flagged the test-driven public export surfacing a Pierre batch-op type as a package contract (**#2**) and the misleading "caller's job" comment (**#3**). Cross-validating Hickey's changes: the single-`batch` form preserves the same encapsulated invariant (disjoint maximal recursive subtrees ⇒ order-independent), and the `pathReconcile` leaf is a sound volatility boundary that keeps the concern whole. The one residual smell (**#5**, `ancestorDirectoryPaths` exported through the Solid-adapter surface) predates this PR and is out of scope as a *new* finding.

srid:

Ran end-to-end (`/do --from polish`, after implementing the fix), then merged latest `master` and re-verified.

| Step | Status | Notes |
|------|--------|-------|
| implement | ✓ | `directoryRemovalOps` + effect wiring; 6 unit + 3 headless-Pierre integration tests |
| check / fmt | ✓ | `just check` exit 0 (tsc clean), `just fmt` clean |
| hickey + lowy | ✓ | 3 first-pass + 1 cross-validation finding, each its own commit; 1 pre-existing No-op |
| police | ✓ | rules clean; fact-check comment fix; O(n) `ancestorDirectoryPaths` |
| test | ✓ | solid-pierre 9/9, client 220/220 |
| ci | ✓ | **all 22 nodes green** on `11b9d58`, then **re-run green on merge HEAD `b0ee4ca6`** — 24/24 `gh pr checks` pass |
| master merge | ✓ | merged `origin/master` (`94dd9c92`); registered the plan doc in the new `docs/plans/index.html` per the merged `docs.md` rule |
| evidence | ⚠️ blocked | The off-machine screenshot capture (the prescribed `evidence`-skill mechanism) was lost twice to process restarts in this session. Behavioral proof stands without it: the **headless Pierre integration test** asserts the exact before→after (emptied dirs *present* after a plain file-batch → *absent* after `directoryRemovalOps`, surviving match-dirs untouched), and **`ci::e2e` is green on both platforms**. The "before" state is the screenshot that opened the issue. |

> Evidence is the one gap — flagged honestly rather than faked. Happy to re-run the off-machine capture in a stable session if a visual before/after is wanted before merge.

srid:
## Evidence

**Bug:** the Code-tab file-tree filter (browse / "All files" mode) left empty, result-less directories in the tree. Typing a multi-token query stranded many hollow, expanded dirs that contained no match.

**Repro:** in the Code tab, switch to **All files**, then type `docs plans` into the filter (`filter files…`). The kolu repo has `docs/plans/*.html`, so on `master` this strands every sibling dir.

**Before → After (single clip, captured on clean `nix run` builds of each branch):**

▶ HD video: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=code-tab-filter-empty-dirs-1096.mp4

- **BEFORE (`master`):** the tree keeps hollow, expanded directories with no match — `.agency`, `.agents/skills`, `ci`, `code-police`, `codex-debate/scripts`, `do/scripts/steps`, `elegance`, `evidence`, `fact-check`, … — alongside the real `docs/plans/` matches.
- **AFTER (this PR, `search-collapsae` @ `d63d202`):** only `docs/` → `docs/plans/` and its matching `.html` files remain; the empty directories are gone.

<sub>Captured off-machine on ephemeral pu boxes: each branch served via `nix run github:juspay/kolu?ref=…`, driven headless with Playwright (terminal `cd ~/kolu` so the Code tab browses a rich repo), recorded to webm, transcoded + concatenated with ffmpeg, hosted on the `evidence-assets` release.</sub>

PR #1105, opened 2026-06-01
Fix exit-subscription leak on explicit terminal kills

srid:
**Every explicitly-killed terminal used to strand a SolidJS root and an open server event stream forever.** Each terminal opened a per-terminal `terminalExit` subscription inside a `createRoot` whose disposer was discarded. Natural PTY exits publish `terminalExit` (the server's single-yield-then-close source returns, ending the stream), but **explicit kills intentionally never publish it** — so the subscription and its root waited on an event that would never arrive. Under terminal churn this leaked client roots and server subscribers without bound.

The set of exit subscriptions is a pure function of which terminals exist, so it's now derived from the live terminal list the same way `useTerminalMetadata` already derives its per-terminal metadata subscriptions — via `mapArray` keyed to the list (new `useTerminalExits`). Each terminal id gets its own reactive owner that SolidJS disposes the instant the id leaves the list, so **every removal path releases the subscription uniformly**:

```
terminal id enters listSub() ─▶ mapArray creates a per-id owner ─▶ subscribeExit(id)
terminal id leaves  listSub() ─▶ mapArray disposes that owner   ─▶ onCleanup ─▶ stream torn down
                                  (natural exit · explicit kill · killAll · server restart)
```

This deletes more than it adds: the `subscribeExit` dependency that was threaded through `useTerminalCrud` (every `handleCreate` / sub-terminal create) and `useSessionRestore` (the hydration loop) is gone — the list is the single source of truth.

Addresses the "Explicit terminal kills leak exit subscriptions" finding in [`docs/plans/correctness-review.html`](../blob/subleak/docs/plans/correctness-review.html).

> _New unit test `useTerminalExits.test.ts` asserts the subscription is released the moment a terminal leaves the list — the exact path (no `terminalExit` event) that leaked before — including `killAll` draining to empty, and that survivors never re-subscribe._


srid:

Both lenses ran as parallel sub-agents on `git diff origin/HEAD...HEAD`. **Neither produced a Fix-in-this-PR finding** — the diff replaces an orphaned `createRoot` with the canonical list-keyed `mapArray` lifecycle and removes cross-axis coupling. Cross-validation was correctly skipped (zero Fix findings on both sides).

| 1 | Hickey | **Duplication audit** — `useTerminalExits` extends, not reinvents, the `mapArray` pattern (drives `Event` subscriptions where `useCollection` drives `Collection` values; can't reuse `useCollection` directly) | ⚠️ **No-op** |
| 2 | Hickey | **Fragmentation resolved** — "which terminals have exit subscriptions" was represented at each create site *and* by the list; now the list is the single source of truth | ⚠️ **No-op** (fixed by diff) |
| 3 | Hickey | `createEffect(reconcile)` activation idiom is non-obvious but documented; `listSub()` vs `terminalIds` (sub-terminals need exits too) is a precise domain decision | ⚠️ **No-op** |
| 4 | Lowy | **Boundary correct** — `useTerminalExits` is a sibling of `useTerminalMetadata` on the shared `mapArray` mechanism, encapsulating a distinct volatility axis (what fires on exit vs. what arrives as metadata) | ⚠️ **No-op** |
| 5 | Lowy | **Injection seam on the right side** — `useTerminalExits` owns *when* a subscription lives/dies; the injected `subscribe` closure owns *what* it does (toast + `removeAndAutoSwitch`) | ⚠️ **No-op** |
| 6 | Lowy | `subscribeExit` removed from `useTerminalCrud` and `useSessionRestore` — was cross-axis functional coupling (subscription setup bundled with create/restore sequencing) | ⚠️ **No-op** (fixed by diff) |
| 7 | Lowy | `allTerminalIds` deriving from `listSub()` (all terminals) vs `terminalIds` (top-level, metadata-gated) is intentional — exits track *existence*, not metadata-readiness | ⚠️ **No-op** |

### Hickey rationale

The new module does **not** reinvent `useCollection.ts` / `useTerminalMetadata.ts`. `useCollection` is for `Collection`-typed primitives — it calls `valueSource(keyToInput(key))` per key and returns a queryable `Subscription<T>` (a reactive *value*). `useTerminalExits` drives side-effecting `Event` subscriptions (`app.events.terminalExit.use(...)`) that return `void`. Both share the *mechanism* (`mapArray` per-key lifecycle with framework-managed `onCleanup`) — exactly the `.claude/rules/solidjs.md` rule — so the diff **extends** that pattern to a different surface primitive. The pre-diff code represented one domain fact in two places (per-call-site `subscribeExit` accumulation *and* the live list); the orphaned `createRoot` was the fingerprint that this unity was never maintained. Post-diff, `mapArray` mechanically enforces both invariants (one subscription per live terminal; teardown on list-leave). Nothing to fix.

### Lowy rationale

The new boundary is a **correct sibling**, not fragmentation: `useTerminalMetadata` encapsulates the per-terminal *metadata* subscription set; `useTerminalExits` encapsulates the per-terminal *exit* subscription set — two distinct volatility axes on the same lifecycle clock, both reusing the shared `mapArray` mechanism factored into `useCollection`. The `subscribe`-injection seam sits on the correct (activity) side of the boundary. The diff also removes the subscription-lifecycle concern from `useTerminalCrud` (functional coupling: setup bundled with `create` only because they co-occur) and `useSessionRestore` (a sequence orchestrator into which an activity concern had leaked). No Fix-in-this-PR actions.

srid:
## Codex ⇄ Claude debate

**Outcome: ✅ consensus — round 1.** Codex reviewed the full diff against the merge-base `ea968d93` (the clean 6-file PR diff), confirmed `git status --short` showed no untracked source, and **approved with zero blocking/major findings** — no author round was needed, so no changes were made.

|-------|---------------|---------------------|---------------------|--------|
| 1 | ✅ Approved | 0 | — (no rebuttal needed) | — (no changes) |

> _Codex summary:_ "The change replaces per-create-site terminal exit subscriptions with a list-keyed SolidJS lifecycle using `mapArray`, removes now-unneeded subscription plumbing from create/restore paths, and adds focused tests for disposal and non-resubscription behavior. … did not find any blocking or major correctness issues."

Ran via the `Workflow` tool (codex `--sandbox read-only` reviewer ⇄ Claude author). No pushes or merges — the branch is unchanged.

srid:
## Evidence

The leak this PR fixes has **no visual diff** — an explicitly-killed terminal used to strand a `createRoot` owner and an open server `terminalExit` stream forever. The leak itself is asserted directly by the new `useTerminalExits.test.ts` unit test (it proves a killed terminal's subscription is released the moment its id leaves the list — the exact path that leaked, since an explicit kill never publishes `terminalExit`).

The clip below demonstrates the **kill path the fix touches** staying correct end-to-end after the refactor — recorded off-machine on an ephemeral box by running Kolu's own Cucumber + Playwright e2e scenario `Kill active terminal auto-switches to next` (`features/kill.feature`) with `KOLU_EVIDENCE=1`, the way CI runs e2e. Three terminals are created, each runs a command, then terminal 2 is killed via its tile close button: the workspace auto-switches to `term-three` and the switcher settles at 2 entries, with no page errors.



▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=evidence-kill-churn-1105.mp4

srid:
## [`/do`](https://github.com/srid/agency) results — `--from polish` (+ parallel `/codex-debate`)

Ran the polish→done tail with **ci ∥ evidence ∥ codex-debate** in parallel, as requested.

| hickey + lowy | ✓ | 3m 47s | parallel sub-agents; **zero Fix findings** (all No-op); cross-validation correctly skipped |
| police | ✓ | 11m 44s | rules clean · fact-check clean · elegance applied one refinement (`cc075c94`) |
| test | ✓ | 34s | unit green — 238 client (incl. 2 new `useTerminalExits` tests), 56 server, 46 pty-host |
| create-pr | ✓ | 2m 22s | draft PR + Hickey/Lowy comment |
| codex-debate | ✓ | ~2m | **consensus round 1**, codex approved, zero findings, no commits |
| evidence | ✓ | — | `kill.feature` clip recorded off-machine (35/35 steps), hosted |
| ci | ⚠️ | 18m 56s | **linux fully green** (e2e 392/392 + all nodes; both unit lanes incl. new test) — see note |

### ⚠️ Note on the red `ci::e2e@aarch64-darwin` check

The only red check is **unrelated to this PR** and the merge is sound:

- The **identical commit passed `ci::e2e@x86_64-linux` 392/392**, both unit lanes are green, and the evidence run independently passed `kill.feature` on linux.
- Two darwin runs flaked two *different* ways: (1) a server-unreachable cascade (259/391, transport `POST … failed after retries` including the unrelated `preferences/test__set` helper); (2) on retry, a single Code-tab readiness timeout at **`code-tab.feature:319`** ("Filter prunes directories with no matching files") — a 60s `waitForCodeTabReady` timeout.
- **This diff touches no Code-tab, file-tree, filter, or server code** — only the terminal exit-subscription lifecycle. The failing scenario lives in a recently-churned area (#1096). Filed the host flake on #320.

_`/do --from polish` on Claude Code (model `claude-opus-4-8`)._

PR #1108, opened 2026-06-01
Surface malformed OpenCode message.data instead of empty-state

srid:
**A malformed latest `message.data` row no longer masquerades as an empty OpenCode session.** When the newest message JSON failed to parse, `parseMessageState` returned `null`, and the watcher logged it with the *same* debug line used for a brand-new session with no messages — `"no messages yet"`. So upstream schema drift or DB corruption hid behind a benign empty-state log while the terminal badge silently blanked. This is the `opencode-parse` finding from the [whole-codebase correctness review](../blob/opebcode-warn/docs/plans/correctness-review.html).

The root cause is two distinct conditions collapsed into one `null` + one log line. They're now de-complected — each cause is logged at the layer that actually knows it:

| Condition | Before | After |
| --- | --- | --- |
| Session has no rows | debug `"no messages yet"` (in watcher) | debug `"no messages yet"` (in `deriveSessionState`, where the missing row is observed) |
| Latest row is malformed JSON | **swallowed → also `"no messages yet"`** | **error `"opencode message.data parse failed"` with the parse error + session ctx** |

The error path mirrors the sibling `getLatestAssistantContextTokens`, which already logs `log?.error(…, "opencode assistant message.data parse failed")` for the very same blob — `parseMessageState` was the lone swallower. The watcher's `if (!derived)` branch no longer re-flattens the two causes into a single ambiguous line; it just returns, trusting the lower layers to have logged the real reason.

*This is `error`-level by the `@kolu/log` contract's own guidance — "actual failures … dropped state" — versus `debug` for "expected-absent conditions" like an empty session.*

Covered by unit tests at both layers (`parseMessageState` and `deriveSessionState`): malformed `message.data` logs a parse error and is **never** reported as "no messages yet"; a genuinely empty session still logs only the debug line; and a valid-but-unmodelled role stays silent (no spurious error).

### DIFF
--- a/docs/plans/correctness-review.html
+++ b/docs/plans/correctness-review.html
@@ -142,7 +142,7 @@
 <section id="summary">
   <h2>Summary</h2>
   <div class="callout">
-    <p><strong>Verdict:</strong> I found six correctness issues worth fixing. One high-severity lifecycle issue is now fixed by <a href="https://github.com/juspay/kolu/pull/1105">PR #1105</a>. The remaining open set is one high-severity path-authority bug and four medium-severity issues.</p>
+    <p><strong>Verdict:</strong> I found six correctness issues worth fixing. The high-severity terminal lifecycle leak is fixed by <a href="https://github.com/juspay/kolu/pull/1105">PR #1105</a>, and the OpenCode silent-parse issue by <a href="https://github.com/juspay/kolu/pull/1108">PR #1108</a>. The remaining open set is one high-severity path-authority bug and three medium-severity issues.</p>
   </div>
 
   <table>
@@ -183,7 +183,7 @@
         <td><span class="severity medium">Medium</span></td>
         <td>OpenCode message JSON parse failure is silently treated as "no messages yet".</td>
         <td>OpenCode integration watcher</td>
-        <td><span class="status open">Open</span></td>
+        <td><span class="status fixed">Fixed</span> <a href="https://github.com/juspay/kolu/pull/1108">#1108</a></td>
       </tr>
       <tr>
         <td><span class="severity medium">Medium</span></td>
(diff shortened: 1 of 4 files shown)


srid:
## Codex ⇄ Claude debate

**Outcome: ✅ consensus** — reached in **round 1**. Reviewer (`codex exec`, `--sandbox read-only`) approved with **no blocking or major findings**; the Claude author side had nothing to dispute, so no rebuttal round was needed and no debate commit was produced.

| Round | codex verdict | open blocking/major | Claude dispositions | commit |
|---|---|---|---|---|
| 1 | **approved** | 0 | — (nothing to contest) | — (no change) |

> codex's summary: _"I found no blocking or major correctness issues in this diff. The change preserves the existing `null` return contract for malformed/unknown message state while making malformed JSON observable at the point where it is distinguishable from a missing row, and the watcher no longer collapses the two cases into the same debug log. No untracked files were present in `git status --short`."_

_Posted by the `/codex-debate` step of `/be`._

srid:
## Structural review — hickey + lowy + code-police

Three lenses reviewed the diff in parallel (hickey + lowy forced onto Opus), briefed with the change rationale only — no seeded findings. Each **fix-in-PR** finding was then adversarially refuted from three angles (correctness / simplicity / churn-vs-benefit) and kept only on a ≥2/3 "this is real" majority.

**All three lenses independently converged on one spot** — the new parse-error log's context shape — and nothing else. Codex had already reached clean consensus.

| Lens | Finding (low severity) | Lens disposition | Adversarial gate | Action |
|---|---|---|---|---|
| **hickey** | `parseMessageState` now complects pure parsing with error-logging I/O via threaded `(log, ctx)` | no-op | 0/3 real | Not actioned — _idiomatic to this file (every export threads `log?`), mirrors the already-reviewed sibling; the "pure Result" remedy doesn't de-complect (two null causes) and would drag the sibling in. Correct as-is._ |
| **lowy** | `ctx?: Record<string, unknown>` over-generalizes the interface over a single stable axis (always a session id) | **fix-in-PR** | 0/3 real | **Actioned anyway** (see below) |
| **code-police** | log context **key** diverges from the sibling it claims to mirror: `{ session }` vs the sibling's / `withDb`'s `{ sessionId }` | no-op | 0/3 real | **Actioned anyway** (see below) |

### Why I actioned a finding the strict gate rejected

lowy's and code-police's findings are the **same objective issue**, and it's real rather than taste: the original diff introduced a `session`-vs-`sessionId` key split **inside `deriveSessionState` itself** (its `withDb` errorCtx keys on `sessionId`, my new debug log keyed on `session`), and diverged from the sibling `getLatestAssistantContextTokens` — while the commit/PR/plan all claim the change "mirrors the sibling." In an observability fix, an operator filtering logs on `sessionId` would have missed the new lines. The adversarial verifiers voted it down on churn/taste grounds, but correctness-of-the-claim + greppability won: applied in **`086052a4`** — narrow the third param to `sessionId?: string` and emit `{ err, sessionId }` / `{ sessionId }` at both new sites. No behavior change beyond the log key; tests updated to assert `sessionId`.

hickey's complecting observation is genuine but correctly self-dispositioned **no-op** — the de-complecting that actually mattered (separating the two null causes in log _output_) is fully achieved, and the structural remedy carries more cost than benefit here.

_Posted by the `/hickey` + `/lowy` + `/code-police` step of `/be` (ultracode fan-out with adversarial verification)._

srid:
## Evidence

This is a server-log observability fix with no UI surface, so the proof is the log differentiation: a malformed `message.data` row that was silently swallowed on master now emits a distinct `[error]` on this PR, while a genuinely empty session emits a distinct `[debug]` — the two causes are no longer indistinguishable.

**BEFORE (master) — both paths completely silent:**

```
=== malformed latest message.data row ===
returned: null

=== genuinely empty session (no rows) ===
returned: null
```

**AFTER (this PR) — distinct signals for each cause:**

```
=== malformed latest message.data row ===
[error] opencode message.data parse failed {"sessionId":"s1","err":"SyntaxError: Expected property name or '}' in JSON at position 1 (line 1 column 2)"}
returned: null

=== genuinely empty session (no rows) ===
[debug] no messages yet for opencode session {"sessionId":"s1"}
returned: null
```



▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=evidence-opencode-parse-warn.mp4

_Captured via the evidence harness on an ephemeral pu box (off-machine), driving `deriveSessionState` directly. Posted by the /evidence step of /be._

Issue #1017, opened 2026-05-28
Stale 'running' pill: abandoned session with a trailing user/notification entry (mtime-staleness heuristic)


srid:
## Summary

A Claude Code session whose transcript tail is a **transient state** — a trailing `user` entry (→ `thinking`) **or a dangling `assistant` `tool_use`** (a tool call with no following `tool_result`, → `tool_use`) — keeps `deriveState` returning a *working* state with **no decay**. Correct transiently; **wrong once the session is abandoned/idle** (most reliably: claude killed **mid-tool**, then resumed by session-restore). The dock spins a "running" pill indefinitely.

Distinct from the orphaned-`running_background` bug fixed in **#1109** (that decays the background-task promotion); this is the **transient-state** half and #1109 does not touch it.

## Reproduced (confirmed, dev build via chrome-devtools)

1. Real `claude` in a kolu terminal; drove it to `Bash(sleep 300)` foreground → genuinely mid-tool (Dock = `tool_use` + spinner, correct).
2. **Killed kolu** mid-tool (killed tree included `claude -c` → `bash -c 'sleep 300'` → `sleep 300`).
3. **Restarted kolu → session restore → resumed `claude -c`** → claude sits **idle at the prompt**.
4. Dock spins `tool_use` forever.

The facts at step 4:

| Signal | Value |
|---|---|
| DOM `data-agent-state` (dock-row + agent-indicator) | `tool_use` |
| Dock pip `animate-spin` | present (spinner) |
| `sleep 300` process | none — tool not running |
| claude process CPU (2s sample) | 0 — idle |
| **claude descendant processes** | **none** |
| Transcript tail | dangling `assistant` `tool_use` (Bash), **no `tool_result`** |

Earlier evidence (the trailing-`user` variant), `~/.claude/projects/-home-srid-code-drishti--worktrees-perf/…jsonl`, quiet ~6 h, tail `… user [<task-notification> <status>killed</status>] … attachment` → `deriveState` walks back to the newest `user` → `thinking` forever.

## Why it's hard — and the discriminator that resolves it

A genuine running tool and an abandoned one are **transcript-indistinguishable** from a single snapshot (a trailing `tool_use`/`user` *normally* precedes more work). mtime-staleness alone is therefore ambiguous: a long `Bash` or a sub-agent legitimately keeps the main transcript quiet for minutes.

**But the process tree distinguishes them cleanly** (verified live across ~10 sessions):

[...]

PR #1015, opened 2026-05-28
Detect Claude Code's running-in-background state for dynamic workflows

srid:
Claude Code's [dynamic workflows](https://code.claude.com/docs/en/workflows) (and any backgrounded task) let the agent **end its turn while work it launched keeps running**. Kolu derives each terminal's status by tailing the session JSONL transcript — and that bare `end_turn` read as `waiting`, so a busy-waiting agent looked like it was *idle and needing you*. This adds a `running_background` state so the dock and tile chrome tell the truth, and surfaces the background workflow's fan-out.

### The misclassification

| Transcript reality | Before | After |
| --- | --- | --- |
| `end_turn`, no background task | `waiting` (needs you) | `waiting` (unchanged) |
| `end_turn`, **launched task still running** | `waiting` → false "needs you" ping | **`running_background`** — buckets as *working*, no ping |
| `tool_use` / `thinking` | working | working (unchanged) |

### How detection works

`deriveState` walks the transcript tail as before; a new `outstandingBackgroundTasks` scan reconciles launch markers against completion notifications in that same tail:

```
user tool_result  "… launched in background. Task ID: X"       (Workflow)        ─┐
user tool_result  "Command running in background with ID: X"    (Bash bg)        ─┼─▶ outstanding
user tool_result  "Async agent launched successfully. agentId: X" (Agent bg)     ─┤   = launched − completed
queue-operation   <task-id>X</task-id><status>completed|failed|stopped|killed</…> ─┘
```

The launch ID and the completion's `<task-id>` are the same token. If any task is still outstanding when the newest turn is a bare `end_turn`, the state is promoted to `running_background` (only `waiting` is promoted — an in-flight `tool_use`/`thinking` already reads as working, and a genuine `AskUserQuestion`/`ExitPlanMode` gate is left alone).

### What landed

- **`running_background` state** — Claude-Code-only (Codex/OpenCode have no analogue), so it lives on `ClaudeCodeInfo` rather than the shared shape. Buckets as *working* in the dock, busy (not warning) indicator color.
- **All three background-launch forms** — the `Workflow` tool, backgrounded `Bash`, and backgrounded `Agent`/`Task`; `killed` counts as a terminal status. _(The Bash/Agent coverage was added after dog-fooding caught a session busy-waiting on backgrounded CI still reading as `waiting`.)_
[...]

PR #1109, opened 2026-06-01
Stop the Claude pill spinning forever on orphaned background tasks

srid:
**Kolu showed Claude Code as "running" — the spinning dock pill — and it stayed lit even after the user restarted Claude.** `deriveState` promoted a bare `end_turn` to `running_background` on *any* outstanding background task, and a task whose completion notification never arrives (the launching process died on restart) leaves a **permanent** launch marker in the transcript. The promotion never decayed, so the pill spun indefinitely.

The fix promotes only on a task kolu can actually *observe*:

```
outstandingBackgroundTasks(tail)
        │
        ▼
liveOutstandingTasks ──drop Workflow runs whose journal is terminal/stale──┐  (watcher, IO)
        │                                                                   │
        ▼                                                                   │
deriveState ──promote only runId-bearing Workflow runs (Bash/Agent never)──┘  (pure)
        │
        ▼
   running_background  ⇐  only a live, observable Workflow keeps it
```

### Two layers, two orphan classes

| Layer | Rule | Closes |
| --- | --- | --- |
| `deriveState` *(pure)* | promote only a `runId`-bearing `Workflow` run | the reported bug — a detached `Bash`/`Agent` (`runId` null) has no journal to observe, so its permanent launch marker no longer spins the pill |
| `liveOutstandingTasks` *(watcher)* | drop a `Workflow` whose journal is terminal or stale (mtime past `WORKFLOW_JOURNAL_STALE_MS` = 2 min) | a `Workflow` orphaned by a restart — its journal freezes at `running` |

*A false-positive de-promote self-heals:* the next journal write fires the workflows-dir watcher and re-derives, so a momentarily-quiet live workflow snaps back to `running_background`.

### Why not "is Claude alive?"

A restart reuses the same `sessionId`, and the session file re-keys to the **new, live** PID — so a process-liveness check returns *alive* and never vetoes the spinner. The transcript's own record (a `Workflow`'s journal; the `runId`) is the signal that survives the restart. *Full diagnosis, the four candidate fixes, and the two process-liveness traps live in `docs/plans/phantom-running-background.html`.*

The open background-task sibling of #1017; introduced by #1015.


srid:
## Codex ⇄ Claude debate

Two runs on this diff. The first reached a **wrong** consensus; the second reached a **correct** one after a real, evidence-based debate.

### Run 1 — consensus, then reverted

Round 1 codex raised **F1 (major)**: `liveOutstandingTasks` reads `<session>/workflows/<runId>.json`, which the repo's own `dynamic-workflow-viewer.html` documents as a *dead* path (real journals at `subagents/workflows/<runId>/journal.jsonl`) — so the gate would "drop all live workflows." The author **agreed and removed the whole journal gate**.

That capitulation was wrong. On-disk verification: `workflows/<runId>.json` snapshots **exist today** (`wf_1662f087-900.json`, `wf_49522206-087.json`) carrying `{workflowName,status,agentCount}` — the exact shape the shipped `deriveWorkflowProgress` reads — with mtimes at each run's end. The Workflow runtime's on-disk format **churned** (present → absent → present again); the doc is stale. The removal was **reverted** and the gate **restored, hardened to fail-open** against exactly that churn.

### Run 2 — consensus after 4 rounds (genuine debate)

| Round | codex verdict | Disposition |
| --- | --- | --- |
| 1 | ✗ **F1** (major): stale-mtime check only fires on fs-events → a *fresh* orphan never re-evaluates when wall-clock crosses the threshold | **Fixed** — `nextWorkflowStaleDeadline` + one-shot stale-deadline timer (`cec31b68`) |
| 1 | ✗ **F2** (minor): schema doc still said Bash/Agent promote | **Fixed** — schema contract corrected |
| 2 | ✗ **F3** (major): a runId-bearing run with missing/unreadable journal fail-opened forever (no timer) | **Fixed** — shared `workflowStaleAnchorMs` (journal mtime → dir mtime → null, never `now`) → *bounded* grace + a timer always armed (`25608a79`) |
| 3 | ✗ **F4** (major): "OpenCode hardening removed vs master" | **Disputed → codex conceded** — proven base drift (branch predates `origin/master`'s #1108); `git diff HEAD -- opencode/` empty. Resolved by merging master |
| 4 | ✓ **approved** | consensus |

F1 and F3 were **genuine gaps the first cut missed** — the review made the fix materially better (a fresh orphan now self-clears via the timer; an unobservable run can't spin forever). F4 was the author **pushing back with evidence** and codex **verifying and accepting** — the debate the first run lacked.

Net: the journal gate **survived, hardened**. `master` merged for #1108. 82 unit + 14 e2e green.

srid:
## Lowy ⇄ Hickey lens debate

A second review gauntlet, run as a **reviewer-vs-reviewer** debate (a `/codex-debate`-style loop, but structural lenses): **lowy and hickey reviewed the diff independently and in parallel, then cross-examined each other's findings to consensus.** Both on Opus.

**Why the structure matters.** A first attempt fed hickey a *pre-curated* "lowy finding" to rebut — and it concluded *drop*. That was framing bias. With **independent parallel review**, hickey *independently raised the same journal-reader duplication* — so the unbiased answer flipped to **fix**. The structure removed my thumb from the scale.

### Independent findings → consensus (1 round, all agreed)

| Finding | lowy | hickey | Settled |
| --- | --- | --- | --- |
| Journal `path.join`+read duplicated across 3-4 sites on the proven-volatile path | fix | fix (raised it too) | **fix** |
| Two journal *parse* paths with divergent strictness | fix | fix | **fix** (folded into the reader) |
| Terminal-status vocabulary mirrored as a Set *and* a regex | fix | fix | **fix** |

### The sharp refinement both lenses converged on

A naïve shared reader would have **reintroduced the phantom bug**: routing the liveness gate through the strict `WorkflowJournalSchema` (which requires `workflowName`) means a `{status:"completed"}` journal lacking that field fails the parse and lands in the *keep* arm — promoting a terminal run forever. So the shared `readWorkflowJournal` does a **status-only parse**; `deriveWorkflowProgress` layers the schema on top only for its own name/agents projection. Each consumer keeps its own policy (the gate keeps-on-`unparseable` = fail-open).

### Applied (two commits, behavior-preserving)

- `b25b67e3` — `readWorkflowJournal(wfDir, runId)` discriminated reader (`ok` | `unparseable` | `unreadable`); the `workflows/<runId>.json` join now lives in exactly one place.
- `8b37210d` — single `TERMINAL_STATUSES` source; both the regex and the Set derive from it.

82 unit + 14 e2e green; `just check` clean. _(code-police's findings duplicated lowy/hickey's and added two `noop`s — folded in.)_

> Worth turning into a reusable `/lens-debate` skill (parallel independent review → cross-examination to consensus) — tracked separately.

srid:
## Evidence

A state-detection fix with **no visual diff** — the proof is behavioral: a session that has a backgrounded task settles to **waiting** instead of spinning the working ring forever. Captured by the project's own e2e harness (`KOLU_EVIDENCE=1`), the same code CI runs.

**The rigorous before/after** — the same three scenarios were confirmed **red before the fix / green after**, driving the *real* watcher + `deriveState` (not a mocked final state):

| Scenario | Before (master logic) | After (this PR) |
| --- | --- | --- |
| Backgrounded `Bash`, no completion (the reported bug) | `running_background` ⟳ forever | **`waiting`** ✓ |
| `Workflow` orphaned by restart (stale journal) | `running_background` ⟳ forever | **`waiting`** ✓ |
| `Workflow` launch with a Run ID but no journal | `running_background` ⟳ forever | **`waiting`** ✓ |

The clip below records the reported-bug scenario in the live kolu UI — a backgrounded-Bash Claude session is detected and the tile chrome settles to the idle/waiting indicator rather than the spinning working ring:



▶ HD, all three cases:
- Backgrounded Bash → waiting: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=phantom-bg-bash.mp4
- Orphaned workflow (stale journal) → idle: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=phantom-orphaned-workflow.mp4
- Run-ID-but-no-journal → not forever: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=phantom-journalless.mp4

_Recorded locally via the `KOLU_EVIDENCE` e2e path (the harness mocks the Claude session — no real `claude` process, no auth needed). Generated by [`/be`](https://github.com/srid/agency) on Claude Code (model `claude-opus-4-8`)._

Issue #905, opened 2026-05-14
Claude Code: detect AskUserQuestion / ExitPlanMode via PreToolUse hook (JSONL-poll path can't see them)


srid:
Follow-up to #899 / #902. The pull-from-JSONL detection that #902 ships for Codex (`request_user_input`) and OpenCode (`question`) **cannot work for Claude Code's `AskUserQuestion` and `ExitPlanMode`** — the Agent SDK buffers the in-flight assistant message (the one carrying the `tool_use` block) in memory and flushes it to `~/.claude/projects/<…>.jsonl` only *after* the user resolves the prompt. During the wait the block isn't on disk, so `deriveState` reads the prior `end_turn` entry and reports `waiting`, never `awaiting_user`.

The classifier itself is **live, not dead** — `core.ts:231` `AWAITING_USER_TOOLS = {AskUserQuestion, ExitPlanMode}` → `classifyByAwaiting` — it's just **input-starved** during the wait window (`schemas.ts:36-43` documents the case "never fires under the current SDK").

| | Bash tool (15s sleep) | AskUserQuestion (40+ min pending) |
|---|---|---|
| Assistant entry in JSONL during execution? | **Yes** — mid-sleep readback showed 3 new lines | **No** — file static at 14 lines |
| Persisted-message structure | Eager (every assistant turn) | Buffered (only on tool resolution) |

---

## Decision — detect by scraping the rendered screen (zero-config)

Recognize the prompt on the **already-rendered terminal**, instead of waiting for JSONL that arrives too late. Kolu already keeps an `@xterm/headless` mirror per PTY and exposes `getScreenText(id, start, end)` — **plain text, VT-parsed** (ANSI already resolved) — via `ptyHost.ts`, surfaced at `router.ts:150-158`. Read it **server-side**, so detection covers background tiles, mobile (client xterm unmounted), and remote/SSH PTYs (#951) — none of which a client-side buffer read can reach.

No new consumer wiring is needed: **`awaiting_user` is already a fully-wired state** (label "Awaiting input", alert badge, dock pip, OS alert). The scrape is just another **source** that flips it during the wait. The `PreToolUse` hook (below) stays a valid *deployment-time* alternative, not a prerequisite.

### Where it lives — `claude-code/src/screen.ts` (a file, not a new package)

[...]

PR #1111, opened 2026-06-02
docs+spike: detect AskUserQuestion via xterm scraping (#905)

srid:
## What

A **plan** (no code yet) for an alternative to #905: detect Claude Code's `AskUserQuestion` / `ExitPlanMode` by reading the rendered terminal, instead of installing a `PreToolUse` hook that edits the user's `~/.claude/settings.json`. This PR adds the design doc; implementation will land on this branch over time.

📄 `docs/plans/askuser-xterm-detect.html`

## Why this is small

Two pieces you'd assume need building already exist:

- **The destination** — `awaiting_user` is a fully-wired state (enum, "Awaiting input" label, alert badge, dock attention bucket) that is **dead code today**; `schemas.ts:36–43` documents it "never fires under the current SDK" and points at #905. Any timely signal lights the badge with zero downstream work.
- **The source** — pty-host already runs an `@xterm/headless` mirror of every PTY and exposes its **plain text** via `getScreenText()` (`ptyHost.ts:549–560`, surfaced at `router.ts:150–158`). "Scrape the xterm" = read a buffer we already maintain.

## Fragility, measured (not feared)

Extracted the actual emitted bytes from four Claude Code builds on disk (`2.1.77 / .81 / .92 / .159`):

| metric | value |
|---|---|
| patch-release cadence | ≈ 1/day (2.1.157→.159 over 2 days) |
| span oldest→current on disk | 82 patch releases |
| changes to the prompt signature | **0** |

ExitPlanMode is anchored on three byte-stable literals (`Ready to code?`, `…Would you like to proceed?`, `No, keep planning`); the select caret is the `figures` lib's `❯` with an ASCII fallback `>`; **no box-drawing glyphs** (the detector must not depend on borders). So a break would be a discrete, test-catchable event — not constant churn.

## Direction (subject to refinement as we build)

- **Server-side**, decisively — covers background tiles, mobile, and remote/SSH PTYs that a client buffer read can't.
- A dedicated, fixture-tested **`packages/xterm-scrape`** (pure, browser-safe) that encapsulates the one volatile thing; per-version fixtures + adversarial negatives turn silent drift into a loud, localized CI failure.
- **Heuristics** for false positives: bottom-region gate + multi-feature scoring (caret **and** option column **and** literal/footer), recompute-don't-latch.
- **#905's hook stays an opt-in robustness upgrade** (clean question text), not a prerequisite.

## Status / next

- [x] Phase 0 — static capture (signature + 82-release stability)
- [ ] Phase 0 — live capture (~1hr): confirm the box lands in `getScreenText`; main vs alternate screen; row position → seeds first fixtures
- [ ] `packages/xterm-scrape` + test corpus
- [ ] Server wiring (session↔PtyId seam + idle-gated poller → promote to `awaiting_user`)
- [ ] Optional tooltip enrichment

Draft — opened to track the work as it lands.


srid:
Discarding this docs PR — folding the final plan into the issue body (#905) instead, where the decided approach belongs (living doc). The exploration (hickey/lowy debate, electricity pass, the unbiased re-run that moved detection into `claude-code/src/screen.ts`) lives in the branch history if needed.

PR #1113, opened 2026-06-02
feat(ai): /lens-debate — lowy ⇄ hickey structural debate to consensus

### LINKED ISSUE #1110: Skill: /lens-debate — structural lenses (lowy ⇄ hickey) debate to consensus
Generalize the reviewer-vs-reviewer lens debate prototyped in #1109 into a reusable skill, sibling to `/codex-debate`.

## Shape (validated in #1109)

The right structure — found by trial in #1109 — is **parallel independent review, then cross-examination**:

1. **Review (parallel):** `lowy` and `hickey` review the diff *independently and simultaneously*, each emitting its own findings (fix/drop). Neither sees the other's list.
2. **Debate (loop):** each lens then dispositions **every** finding from **both** sets, seeing the other's last round, and concedes when the other is right — until they agree per-finding (consensus) or hold (deadlock → human adjudicates).
3. Apply the agreed `fix` findings (each its own commit); surface deadlocks.

**Why the structure matters (the load-bearing lesson):** a first cut fed hickey a *pre-curated* "lowy finding" to rebut and it concluded *drop* — framing bias. Running the reviews **independently in parallel** made hickey raise the same finding on its own, flipping the verdict to *fix*. Curation biases the outcome; independent-then-debate does not.

Both lenses run on **Opus** (override their `model: sonnet` frontmatter), as `/be` already requires.

## Implementation notes

- Engine is the `Workflow` tool (like `/codex-debate`): `parallel()` for the review phase, a bounded loop for the debate with schema-constrained per-finding dispositions so consensus/deadlock is detected in code, not by vibes.
- Each agent reads its own `.claude/skills/{lowy,hickey}/SKILL.md` framework + the diff before arguing.
- Consensus = every finding's disposition matches across both lenses. Deadlock = unchanged opposing dispositions for a round.
- Post the outcome (table + per-finding reasoning) as a PR comment, by default (mirror `/codex-debate`).
- Optionally fold in `/code-police` as a third, lower-weight voice (in #1109 its findings duplicated the lens findings + two noops).

## Prototype artifacts

- Working prototype workflow script: `/home/srid/.claude/projects/-home-srid-code-kolu--worktrees-ccloading/d318915a-0b62-4540-aefd-969883275017/workflows/scripts/lens-debate-v2-wf_f7389faa-bd0.js`
- Full session transcript (the whole #1109 build, both codex debates, and both lens-debate runs incl. the biased-vs-unbiased contrast): `/home/srid/.claude/projects/-home-srid-code-kolu--worktrees-ccloading/d318915a-0b62-4540-aefd-969883275017.jsonl`
- The PR where it ran: #1109 (see the "Lowy ⇄ Hickey lens debate" comment).

The script can be lifted almost verbatim into `.claude/skills/lens-debate/` and regenerated through the same `.apm` → `just ai apm` path `/codex-debate` uses.

srid:
Generalizes the reviewer-vs-reviewer lens debate prototyped in #1109 into a reusable skill, sibling to `/codex-debate`. Closes #1110.

## What it does

Two structural reviewers argue your change to a settled conclusion:

1. **Review (parallel, independent).** `lowy` (volatility-based decomposition) and `hickey` (structural simplicity) review the diff *simultaneously*, each emitting its own `fix`/`drop` findings. **Neither sees the other's list** — the load-bearing lesson from #1109 is that handing one lens a pre-curated finding biases the verdict (a curated "lowy finding" made hickey conclude *drop*; reviewing independently made hickey raise it itself and flip to *fix*). The lenses never trust a handed-down list — each reads the source.
2. **Debate (loop to consensus).** Each lens dispositions **every** finding from **both** reviews, seeing the other's last round, conceding when the other is right — until they agree per-finding.
3. **Apply.** The agreed `fix` findings land as individual commits; the summary is posted to the PR.

## Why deadlock is not possible

`/codex-debate` can deadlock (reviewer vs **author** — Claude wrote the code and can dig in), and detects it. Here both sides are **disinterested third-party lenses** with nothing to defend, so they converge. There is **no deadlock exit** — the debate runs until consensus, as many rounds as it takes. Three mechanics make that real:

- **Independent review** removes up-front framing bias.
- **Settled findings lock** — once both agree on a finding it leaves the active set, so the contested set is monotonically non-increasing (can't oscillate a settled point back open).
- **Sequential reveal** — within a round lowy posts first and hickey answers lowy's *current* positions, so they land together.

`--max-rounds` (default 12) is a pure safety backstop, not a deadlock cap; reaching it is reported as `unresolved` (needs a human), never `deadlock`.

## Notes

- Both lenses run on **Opus** (overriding their `sonnet` frontmatter), as `/be` requires.
- `--with-police` folds in `/code-police` as a **lower-weight finding source** — it seeds findings into the debate but gets no consensus vote.
- The lenses are read-only; only the Apply phase writes. Commits per fix, never pushes or merges.
- Engine is the `Workflow` tool with schema-constrained dispositions, so consensus/unresolved is detected in code, not by vibes.

## Provenance

Authored in `.apm/skills/lens-debate/` and regenerated into `.claude/` + `.agents/` via `just ai apm`. A future PR will generate this via `/codex-debate`-style tooling; this lift is from the validated #1109 prototype.

This PR will be used to **test the debate on itself** — a `## Lowy ⇄ Hickey lens debate` comment will follow.


srid:
## Lowy ⇄ Hickey lens debate

**🟢 consensus** in **1 round** · base `origin/master` · independent findings: **lowy 2, hickey 2** · 0 unresolved

This is `/lens-debate` run on its own PR — the skill reviewing itself. Both lenses reviewed the source independently, then debated all four findings to consensus. The two agreed `fix` findings were applied as individual commits; both `drop` findings were agreed observations only.

| Finding | Lens | Disposition | Where | Applied |
|---|---|---|---|---|
| Model literal `'opus'` sprayed across 7 sites, not encapsulated | lowy | **fix** | `debate.workflow.js` agent calls + meta phases | `d36af4c6` |
| Consensus checks `disposition` but not the `plan` — Apply takes one lens's plan unchecked | hickey | **fix** | `debate.workflow.js` debate loop | `1e7b5c7c` |
| Generated-runtime `SKILL.md` path duplicated in two prompt builders | lowy | drop | `debate.workflow.js:117,135` | — |
| Same runtime-dir path literal (raised independently) | hickey | drop | `debate.workflow.js:117,135` | — |

### The fixes (agreed by both lenses)

**lowy-1 — model knob.** The string `'opus'` was repeated at four `agent()` call sites plus three `meta.phases` labels. Every *other* tunable (`base`, `maxRounds`, `commit`, `withPolice`, `rationale`) is a named input with a top-of-file default — only the model, the most-recently-changed knob (it exists solely to override the lenses' `model: sonnet` frontmatter), was a sprayed literal. That asymmetry is the receptacle leaking: a model bump had a 7-site blast radius. **Fix:** one `MODEL` const (module-level, since `meta` is evaluated before inputs) + a `model` input defaulting to it; a model change is now a one-liner and overridable via args.

**hickey-1 — disposition/plan fragmentation.** Consensus was detected on `disposition` alone, so two lenses could both say `fix` while proposing structurally *different* changes; the loop locked it `agreed` and fed only lowy's plan to the implementer (`h.plan` was a mere falsy-fallback). The plan is the load-bearing payload, and nothing made the two plans converge. **Fix:** an `agreesWithPlan` field — when both say `fix`, the second poster (hickey, who has seen lowy's plan) must endorse it; if not, the finding stays open until the *plans* converge too, exactly like the disposition. Apply never picks a plan arbitrarily.

> The hickey-1 finding is the satisfying one: the debate caught a genuine gap in its *own* convergence logic — it only made disposition deadlock-proof, not the plan. Now both converge.

### The drops (agreed observations, no change)

Both lenses independently flagged the duplicated `.claude/skills/${lens}/SKILL.md` literal (lines 117/135), then both **dropped** it under the variable-vs-volatile screen: the skill is inert under non-`.claude` runtimes so the path can't vary at runtime, and unlike `/codex-debate`'s `skillDir` (which resolves an executable script), this never tracks a real axis of change. Extracting a const for two lockstep sites would be a weak-volatility micro-boundary.

---

*8 agents · ~239k tokens · 1 round to consensus. The agreed fixes are committed on this branch (`d36af4c6`, `1e7b5c7c`) plus a regeneration commit syncing the `.claude`/`.agents` runtime copies. Not pushed/merged by the skill — `/lens-debate` commits but never ships.*

srid:
**Nudge: fold Löwy's "electricity" lens into the debate — it earned its keep on a live run.**

While reviewing the #905 xterm-scrape plan (PR #1111) I ran this exact `lowy ⇄ hickey` debate, then added a *third* focused pass: **lowy with Juval Löwy's electricity analogy as the framing question** (from *Righting Software* / The Method — ["all that volatility is encapsulated behind a receptacle"](https://www.informit.com/articles/article.aspx?p=2995357&seqNum=2); the anti-pattern is having to "expose the wires and measure the frequency with an oscilloscope" every time you consume power).

It paid off *beyond* the generic lowy review:

- It dragged the abstract "where's the boundary?" down to a concrete **"what is the receptacle, and what are the swappable implementations behind it?"** — for that plan: `awaiting_user` (the state every consumer plugs into) is the receptacle; JSONL / xterm-scrape / the #905 hook are interchangeable *generators* behind it.
- That reframe **demoted an architectural ranking to a deployment fact** — "scrape = default, hook = opt-in" is a statement about *which generators are connected in an environment*, not about how the current is consumed. The plain volatility pass had circled this but never landed it.
- The "**is it electricity?**" test (domain-agnostic utility vs. app concern) cleanly resolved a package-boundary contradiction, and it produced one quotable sentence that went straight into the doc.

**Where it fits in `/lens-debate`:** I'd argue it's *not* a new independent voting lens — that would double-count lowy and reintroduce the up-front framing bias this PR is careful to avoid. It's the **same volatility lens with a sharper probe**. So bake it into the `lowy` reviewer prompt as a required part of its output:

> Name the *receptacle* (the stable interface consumers plug into); name the *volatile implementations* it encapsulates; is this "electricity" (a domain-agnostic utility) or an application concern? Where would a consumer be forced to "expose the wires"?

One lowy vote, richer findings. Optionally gate the framing behind `--electricity` if you'd rather keep it off the default path.

Net: it reliably pulls structural review out of abstraction and into "what plugs into what" — which is precisely the failure mode (abstraction without grounding) a lens debate is otherwise prone to.

Evidence: the run is written up in PR #1111's plan, §9 *"What /hickey + /lowy + the electricity pass changed,"* and the doc now leads with the receptacle framing.

.apm/skills/be/SKILL.md
```diff
@@ -45,13 +45,8 @@ Run **check** and **fmt**, then commit (conventional message) and push the featu
...
-
-**Post the findings to the PR.** After the lenses finish, post the hickey/lowy/code-police findings as PR comment(s) — a ledger table (one row per finding with its disposition) plus each lens's rationale — so reviewers see the structural analysis alongside codex-debate's own comment.
+2. **`/lens-debate`** (Skill tool) on the diff, briefed with the change rationale only — do **not** seed findings. One call does the whole `/lowy` ⇄ `/hickey` structural pass: it reviews with both lenses **independently in parallel** (forcing both onto Opus, overriding their `model: sonnet` frontmatter), debates **every** finding to consensus, applies each agreed `fix` as its own commit, and posts the debate summary to the PR. Pass the change rationale via its `rationale` arg so the lenses don't flag deliberate decisions. Deadlock is not possible; on the rare **unresolved** finding, adjudicate it yourself before moving on. Let it finish.
```
srid: Must post findings to PR comment!

srid: Fixed in c9ae3db0. §4.2 now states it explicitly: *"It MUST post the per-finding debate ledger (origin, finding, disposition, applied commit) to the PR as a comment — this is mandatory; confirm the comment landed (it is `/lens-debate`'s default, so don't pass `--no-comment`)."* `/lens-debate` posts that ledger by default, and `/code-police` (§4.3) posts its own findings too, so every gauntlet step still leaves its trail on the PR.

srid:
Done — folded in as you specified, in `9d2b21d8`.

It's baked into the **`lowy` reviewer's required output**, not a new voting lens — so one lowy vote, richer findings, and no up-front framing bias / double-counting. The reviewer must now name the **receptacle**, the **volatile implementations** behind it, classify it as "electricity" (domain-agnostic utility) vs. an app concern, and call out where a consumer is forced to **"expose the wires."**

Implementation is data-driven rather than a lens-name `if`: the probe is a `probe` field on the lowy reviewer entry that `reviewBrief()` appends when present (Hickey-clean — no branching, trivially extensible to other lens-specific probes later):

```js
const ELECTRICITY_PROBE = `As a REQUIRED part of your output, apply Löwy's electricity test … name the **receptacle** … the **volatile implementations** … is this "electricity" … where a consumer is forced to "expose the wires." If the diff has no such boundary, say so explicitly; do not invent one.`

const REVIEWERS = [
  { lens: 'lowy', framework: '…', probe: ELECTRICITY_PROBE },
  { lens: 'hickey', framework: '…' },
]
```

I went with **default-on** rather than gating behind `--electricity` — your live-run write-up (#1111 §9) makes the case that it's pure upside (it reliably pulls review out of abstraction into "what plugs into what"), and a default-off probe rarely gets used. The "no boundary here — don't invent one" escape hatch keeps it from manufacturing receptacles on diffs that have none, consistent with the skill's existing "empty findings is a fine verdict" stance. Easy to flip to a flag if you'd rather.

Documented under **Why this shape** in `SKILL.md`.

srid:
## Codex ⇄ Claude debate

**🟢 consensus** in **2 rounds** · base `origin/master` · codex final verdict: **approved**

Rebased onto `origin/master` first (the branch was 2 commits behind — #1109/#1112 had merged — so a two-dot diff would have shown codex a phantom revert of that work). codex then reviewed only this PR's 11 files.

| Round | codex verdict | Open blocking/major | Claude disposition | Commit |
|---|---|---|---|---|
| 1 | changes requested | F1 (major) | **fixed** F1 | `322c2b0e` |
| 2 | **approved** | none | — (F2 minor, see below) | — |

### F1 (major) — consensus gate could settle a `fix` with no plan ✅ fixed

codex caught a real bug: `POSITION_SCHEMA` doesn't require `plan`, so if lowy omitted a plan while hickey set `agreesWithPlan: true`, a `fix` would settle with `plan: undefined` — and Apply would run on a vague placeholder while committing it as "lens consensus." This is exactly the disposition/plan-fragmentation hole the earlier dogfood (hickey-1) *half*-closed: it gated on `agreesWithPlan` but never guaranteed lowy's plan was actually present.

Claude (author) agreed and fixed it in `322c2b0e`:
- Added a `lowyHasPlan` check (`typeof l.plan === 'string' && l.plan.trim()`) **ANDed into** the `fix` agreement test — a fix can't settle without lowy's concrete, non-empty plan, so it stays active and keeps converging instead of reaching Apply with `plan: undefined`.
- Threaded the raiser's original `suggestion` through `settledOut` into `implementBrief` as labeled *context (not the agreed plan)*, and made the brief use `fix.plan` directly (removed the now-dead placeholder fallback).

Claude declined codex's suggestion to make `plan` schema-*required*, with sound reasoning codex accepted: a `drop` position and an in-progress (non-settled) `fix` can legitimately have no plan yet — the runtime gate enforces the real invariant (no fix Applies without an agreed plan) without rejecting valid in-progress positions.

> Nice property of the gauntlet: `/lens-debate` reviewing itself caught the *shape* of this bug (disposition vs plan), and `/codex-debate` caught the *residual hole* in that very fix. Belt and suspenders.

### F2 (minor) — wrong regen recipe ✅ fixed (post-debate)

codex flagged that SKILL.md said `just ai apm` while the repo convention is `just ai::apm` (the space form doesn't resolve). Fixed in `6a22db94`. *(The `/codex-debate` sibling SKILL.md has the same wording at line 167 — left untouched as out-of-scope for this PR, but worth a follow-up.)*

---

*4 agents · ~82k tokens · 2 rounds to consensus. F1 committed by the debate; F2 by me afterward. Not pushed/merged by the skill — the per-round commit sits on the branch for review.*

srid:
**Follow-up — a sharper electricity probe, from watching this debate get something wrong.**

I ran the `lowy ⇄ hickey` debate (Opus) on a real plan ([#905 xterm-scrape](https://github.com/juspay/kolu/pull/1111)) and it reached a clean 2-round *consensus* that a proposed new top-level package was the right cut. Then an **unbiased** lowy re-run — same lens, but told to distrust the framing and read the source — **overturned that consensus**: the capability belonged inside an existing integration package, because a cross-domain primitive (`classifyByAwaiting` in `anyagent`) already encapsulated the volatility, with per-agent data living in each integration. The "consensus" had ratified the plan's *self-description* ("agent-agnostic, no `claude` import") instead of checking it against the tree.

That's a framing-bias escape exactly of the kind this PR is built to prevent — except it slipped through because the leading premise was in the *plan*, not in a handed-down finding. Two cheap mitigations worth baking into the lens prompts (especially the lowy/electricity probe):

1. **Count independent *consumer domains*, not module decoupling.** "It has no `claude` import" is necessary, not sufficient — a thing importing *all three* agents' data is still 100% in the agent domain. The litmus is "≥2 independent domains consume this," else it's a single-domain capability parameterized over N implementations (and should be named/placed as such).
2. **Search for an existing primitive that already encapsulates this volatility** before blessing a new boundary. Had either lens grepped for the awaiting-user policy, it'd have found `classifyByAwaiting` and the per-integration pattern in round 0.

Net: both are "go read the source and resist the given frame" checks — the same principle as your independent-review rule, just aimed at the *premise* as well as the *findings*. Optional, but they'd have caught this one in round 0 instead of a re-run.

Issue #1011, opened 2026-05-28
Structured agent-status side-channel via Claude Code hooks (dynamic-workflow #4)


srid:
## Summary

Add an optional **structured status side-channel** for Claude Code sessions by registering Claude Code **hooks**, instead of (only) inferring agent status from the session JSONL transcript. This is option **#4** from the dynamic-workflow status investigation; options #1–#3 (transcript-parser improvements) are being done first in a separate PR.

## Motivation

Today Kolu derives per-terminal agent status by tailing the session JSONL transcript (`packages/integrations/claude-code/src/core.ts` → `deriveState`). That works and stays interactive-friendly, but it's coarse for a few states that hooks expose precisely:

- **blocked on permission** vs **idle, waiting for user input** — transcript can't cleanly tell these apart; the `Notification` hook fires `permission_prompt` vs `idle_prompt` with a human-readable `message`.
- **sub-agent / dynamic-workflow fan-out** — `SubagentStart` / `SubagentStop` carry `agent_id`, `agent_type`, `agent_transcript_path`; TS-only `TaskCompleted` fires on background-task completion. These give exact start/stop edges instead of polling `subagents/**/agent-*.jsonl`.
- **done** — `ResultMessage` / Stop is an unambiguous terminal signal.

Verified against primary docs (code.claude.com/docs/en/agent-sdk/hooks, /hooks). **Important correction to an earlier assumption:** hooks fire in *interactive* mode too — they're configured in `settings.json`, not headless-only (only slash commands are interactive-only). So this is a hybrid: keep Claude Code's interactive TUI for the user, gain a structured status feed for Kolu.

## Sketch

- Register hooks in the user's Claude Code `settings.json` (managed by Kolu, opt-in).
- Hooks POST event JSON to a small local receiver in the Kolu server; correlate by `session_id` (and `agent_id` for sub-agents) to the owning terminal/`ProviderRecord`.
- Map events → existing agent-status model: `Notification(permission_prompt)` → awaiting-permission, `Notification(idle_prompt)` → awaiting-user, `SubagentStart/Stop` → fan-out count, `TaskCompleted` → background-task done, `Stop`/`ResultMessage` → done.
- Treat as an enrichment layer over transcript parsing, not a replacement — degrade gracefully when hooks aren't installed.

## Cost / open questions

- Kolu has to own a section of the user's `settings.json` (install/uninstall/merge safely) and run a local receiver (socket/HTTP). Bigger lift than the transcript-parser changes.
[...]

Issue #1114, opened 2026-06-02
Skill: launch the kolu dev server on RANDOM ports (never collide with production)


srid:
## Problem

An agent capturing evidence on PR #1109 launched a kolu dev server and **disrupted the running production `kolu.service`** (systemd, fixed ports). Agents must **never** touch production kolu, and must never bind production's ports.

## Skill: `dev-server` (launch + remember + isolate)

A skill every agent loads before running the app locally:

1. **Always launch via `just dev-auto`** — it picks **two random free ports** (backend + frontend) and never uses the fixed defaults (`7700`/`5180`) that collide with production. Never run `just dev <fixed ports>`.
2. **Capture and REMEMBER both ports.** Parse the launcher/Vite output for the client URL *and* the server port, and persist them to a session scratch file (e.g. `$SCRATCH/dev-ports.json`) so the agent (and chrome-devtools) connects to the right URL across tool calls — don't re-grep logs each time, don't guess.
3. **Learn production's ports to avoid them, read-only.** `systemctl --user status kolu` (or `ss -ltnp | grep kolu`) to see production's port/PID — purely to steer clear. **Never `start`/`stop`/`restart`/kill the systemd unit or its nix-store process.**
4. **Tear down only the dev instance.** On cleanup, kill only the PIDs bound to the *random* dev ports / rooted in the worktree path — identify them via `ss -ltnp` on the remembered ports. Never `pkill -f kolu`/`vite`/`tsx` broadly (that can hit production or unrelated processes).
5. **Hand chrome-devtools the remembered client URL** for screenshots/driving.

## Acceptance

- Two random ports, both remembered and reused across the session.
- Production `kolu.service` provably untouched (status unchanged before/after).
- Clean teardown that only removes the dev instance.

Surfaced by #1109 (evidence-capture step). Make this the canonical "run the app" path so future agents don't repeat the collision.

PR #1115, opened 2026-06-02
Settle a stuck tool_use pill to idle once a claude session is abandoned

### LINKED ISSUE #1017: Stale 'running' pill: abandoned session with a trailing user/notification entry (mtime-staleness heuristic)

srid:
**An abandoned Claude Code session whose transcript tail is a dangling `tool_use` (an assistant tool call with no following `tool_result`) kept the dock spinning a "running" pill forever.** `deriveState` reports `tool_use` for that tail — correct while the tool is genuinely running, but wrong once the session is abandoned. The reliable repro: kill claude mid-tool (`Bash(sleep 300)`), then let session-restore resume it idle at the prompt — the `tool_use` pill then spins indefinitely.

This is the **transient-state half** of the stuck-pill problem. [#1109](https://github.com/juspay/kolu/issues/1109) fixed the sibling `running_background` case (an `end_turn` promoted on an orphaned workflow journal); it deliberately doesn't touch the dangling-`tool_use` case. `Closes #1017`.

### The hard part — and the discriminator

A genuine in-flight tool and an abandoned one are **transcript-indistinguishable from a single snapshot** (a trailing `tool_use` *normally* precedes more work), so mtime-staleness alone is ambiguous — a long `Bash` or a sub-agent legitimately keeps the main transcript quiet for minutes. The **process tree** resolves it cleanly:

| | descendant processes | meaning |
| --- | --- | --- |
| **Genuine work** | ≥1 (the `Bash` child / a sub-agent `claude`) | keep the pill |
| **Phantom** (abandoned / killed-resumed-idle) | none | decay to idle |

So de-escalation fires only when **both** hold: the transcript has gone quiet past a window **and** claude's process subtree has no descendant. A long quiet tool keeps a child, so it is never wrongly cleared.

> ### Why `thinking` is *not* decayed
> An earlier revision also decayed a trailing `thinking` (newest entry a `user` prompt). The codex⇄claude review (consensus, see PR comment) caught that as a **correctness bug**: a `thinking` turn awaiting the model's first token has spawned no child and writes nothing until the reply streams back, so "quiet transcript + no descendants" *cannot* tell a slow/hung live request from abandonment — clearing it would publish `waiting` over a live turn. The no-descendants probe is a valid liveness signal **only for `tool_use`**, where live work implies a child process. The `thinking`/trailing-`user` variant of #1017 is therefore deferred to the structured-hook north star ([#1011](https://github.com/juspay/kolu/issues/1011)), which would make idle/running exact.

### Where it lives

```
deriveState (pure, unchanged)
      │  tool_use
      ▼
session-watcher ── transcript quiet past TRANSIENT_STALE_MS (2 min)?
      │                         │ yes
      │ no → arm recheck timer  ▼
      │      (reused #1109      isClaudeSubtreeIdle(pid)?  ── ps -A -o pid=,ppid=
      │       stale-deadline)        │ no descendants → state = "waiting"
      ▼                              │ has child      → keep, re-probe next window
   publish
```

`deriveState` stays pure; the watcher de-escalates *after* it and reuses #1109's `nextWorkflowStaleDeadline`/`scheduleStaleRecheck` one-shot timer to re-derive when the transcript goes quiet with no fs event. The decision logic (`decayTransientState`) and the process-tree predicate (`hasNoDescendants`) are pure and unit-tested; the `ps` spawn is isolated in `snapshotProcessTree` and runs *only* past the window, never on the hot transcript-event path.

> *Fail-safe throughout: if the process table can't be sampled (`ps` missing/errors) or the transcript can't be stat-ed, the probe returns "assume working" — a probe failure never clears a genuinely-working pill. A false positive (a tool quiet >2 min that already died) self-heals on the next transcript write. CPU is deliberately omitted: `ps` reports a lifetime/decaying average, not an instantaneous one, so it isn't a portable "busy right now" reading.*

### Try it locally

```sh
nix run github:juspay/kolu/fix/stale-transient-pill-1017
```


srid:
## Codex ⇄ Claude debate

**🟢 Consensus** in **2 rounds** (base `origin/master`).

> _Final verdict (codex):_ "The change is narrowly scoped to phantom `tool_use` de-escalation. The updated implementation now leaves `thinking` untouched, only probes stale `tool_use`, fails open on `ps`/stat errors, and schedules rechecks for quiet transcripts. No remaining blocking or major correctness issues."

| Round | codex verdict | open blocking/major | Claude disposition | commit |
| 1 | ❌ changes requested | 1 major (F1) | **fixed** — narrowed decay to `tool_use` only | `8dddce3f` |

### F1 (major, resolved) — invalid liveness signal for `thinking`

**codex:** `decayTransientState` treated both `thinking` and dangling `tool_use` as clearable on "transcript stale + no descendants". That discriminator is valid for a tool call (live tool work implies a child process) but **not** for `thinking`: during a slow or hung model request claude does network IO with **no descendant process and no transcript writes**, potentially past the 2-min window — so the watcher could publish `waiting` over a live turn.

**Claude:** Accepted. In `deriveState`, `thinking` arises when the newest entry is the `user` prompt — claude is awaiting the model's first token, has spawned no child, and writes nothing until the reply streams back, so `isClaudeSubtreeIdle` cannot tell live-from-abandoned for `thinking`. Fix: `decayTransientState` now gates on `state !== "tool_use"`, so `thinking` returns unchanged and the `ps` probe never runs for it. The `tool_use` decay is retained (live tool execution implies a descendant). Module/function docs and `transient-decay.test.ts` updated to assert `thinking` is never decayed. 125 tests pass, tsc clean.

> _Scope note:_ this narrows the PR to the **reliably-reproduced `tool_use` phantom** (killed mid-tool → resumed idle). The `thinking`/trailing-`user` variant in the issue is deferred — it needs a stronger signal than the process tree can give (north star: the #1011 structured-hook side-channel).

_Posted by [`/codex-debate`](https://github.com/srid/agency) — Claude Code (model `claude-opus-4-8`)._

srid:
## Lowy ⇄ Hickey lens debate

**🟢 Consensus** in **2 rounds** (base `origin/master`, no police voice). Independent findings: **lowy=1, hickey=2**. One `fix` applied; two `drop` after debate.

| Finding | Origin | Disposition | Location | Applied commit |
| **hickey-1** — "when to recheck" expressed in two units, reconciled at the call site | hickey | ✅ **fix** | `session-watcher.ts:304-319` | `d2966f69` |
| **lowy-1** — two liveness decays share one volatility axis but expose mismatched return shapes | lowy | ⬇️ **drop** | `session-watcher.ts:302-319` | — |
| **hickey-2** — per-state routing knowledge duplicated (watcher `if` vs `decayTransientState` guard) | hickey | ⬇️ **drop** | `session-watcher.ts:302-319` | — |

### hickey-1 (fixed) — one clock, one unit

The recheck instant was assembled from **two `Date.now()` samples** taken at slightly different times: the `running_background` branch yielded an *absolute* deadline while `decayTransientState` yielded a *relative* `recheckInMs` that the call site re-absolutized with `Date.now() + recheckInMs`. **Fix (both lenses agreed the plan):** `decayTransientState` now takes an injectable `now` (like its sibling `nextWorkflowStaleDeadline`) and returns an absolute `recheckAt`; the watcher samples `now` once, threads it into both `transcriptQuietMs` and `decayTransientState`, and assigns `staleDeadline` directly — the adapter and the second clock read are gone. Tests now assert `recheckAt` against a fixed injected `now`.

### lowy-1 (dropped) — lowy retracted his own finding

> _lowy:_ "Conceding my own finding as over-engineering. `nextWorkflowStaleDeadline` does **not** decide a published state — it returns only a deadline; the demotion of `running_background` happens via `liveOutstandingTasks` re-filtering on the next pass. `decayTransientState` *is* the published-state decision. Forcing the two into one `{state, recheckAt}` receptacle would make `nextWorkflowStaleDeadline` always claim `state:'running_background'` — a lie it never computes. The 'fold over decision-producers' for the future #1011 signal designs for change whose shape is unknown — speculative volatility, not volatility."

hickey concurred: the two are genuinely different mechanisms (*input-gate re-trigger* vs *output override*), which is exactly the deliberate `deriveState`-stays-pure separation. The one real grain — two producers speaking different time units — was the hickey-1 fix, needing no new abstraction.

### hickey-2 (dropped) — defensible split

The watcher's `if (running_background) … else …` and `decayTransientState`'s `tool_use`-only guard encode the same state taxonomy in two places, but both lenses agreed the current split is defensible given the two-mechanisms separation; unifying now would over-abstract. Observation only.

_Posted by [`/lens-debate`](https://github.com/srid/agency) — lenses on Claude Code (model `claude-opus-4-8`)._

srid:
## 🚓 code-police

Passes: **rules → fact-check → elegance**. One finding fixed (its own commit), two skipped on cost/benefit.

| Finding | Pass | Disposition | Commit |
| --- | --- | --- | --- |
| Dead barrel re-exports with no external consumers | rules / no-dead-code | ✅ **fixed** | `c1d46d27` |
| Double-`stat` of the transcript (`size` then `mtime`) in one `onTranscriptMaybeChanged` | elegance / efficiency | ⏭️ **skipped** | — |
| Inline the 3-line `isClaudeSubtreeIdle` | elegance | ⏭️ **skipped** | — |

**Fixed — dead public API:** the 6 transient-decay symbols (`decayTransientState`, `hasNoDescendants`, `isClaudeSubtreeIdle`, `snapshotProcessTree`, `TRANSIENT_STALE_MS`, `type ProcEntry`) were re-exported from `index.ts` for surface symmetry with the sibling workflow-staleness machinery — but nothing consumes them through the barrel (`transient-decay.test.ts` imports from `./core.ts`; `session-watcher.ts` uses `./core.ts`). *Symmetry isn't a consumer* → removed.

**Skipped — double-stat:** `tailJsonlLines` stats for `.size`, then `transcriptQuietMs` stats again for `.mtimeMs` in the same handler. Both fields are one `stat` call, but this is a *debounced, rare path* (transcript-change events only, never the hot path), and folding mtime into `tailJsonlLines`' clean size-only interface would couple two independent concerns — change surface exceeds the gain.

**Skipped — inline `isClaudeSubtreeIdle`:** the 3-line function is a named concept that documents the "claude subtree shows no live work" discriminator; inlining into the probe lambda would make the caller longer and less readable.

_Posted by [`/code-police`](https://github.com/srid/agency) — Claude Code (model `claude-opus-4-8`)._

srid:
## Evidence

Captured with **real Claude Code** (v2.1.160) driven through **chrome-devtools** against a kolu dev server on **random ports** (client `49973`, server `42191` — never production's `7692`, per #1114). No mock harness; this is the live `terminal → claude-code provider → Dock` pipeline.

### Reproducing the exact #1017 abandonment, then watching the fix clear it

1. Ran real `claude` in a kolu terminal; prompted it to `Bash(sleep 600)` → Dock pill = **`tool_use`** (working, spinner), with a live `sleep` child.
2. **Killed claude mid-tool** (`kill -9` the claude subtree, by pid) so the transcript tail is a **dangling `tool_use`** with no `tool_result`.
3. **`claude -c`** resumed the session → claude sits **idle at the prompt, zero descendant processes**, transcript tail still the dangling `tool_use`. This is the precise phantom from the issue.

The Dock then transitioned on its own — same card, only the state indicator changed:

dock before/after

### Machine-readable proof (DOM `data-agent-state`, queried live)

| Wall-clock | claude | transcript tail (newest user/assistant) | `data-agent-state` | spinners |
|---|---|---|---|---|
| 22:49:49 | alive, **idle, 0 descendants** | dangling `tool_use` | `tool_use` ×2 | **2** |
| 22:51:53 | alive, **idle, 0 descendants** | *(unchanged)* | `waiting` ×2 | **0** |

Transcript last write was 22:49:05; the reused #1109 stale-deadline timer fired at ~22:51:05 (mtime + `TRANSIENT_STALE_MS` = 2 min), the subtree probe confirmed **no descendant process**, and `deriveState`'s `tool_use` decayed to `waiting`. **Without this fix the pill spins `tool_use` forever** (the bug). The transcript never changed across the transition — only the out-of-band decay cleared it.

> *The discriminator at work:* in step 1 the genuine `tool_use` had a live `sleep` child, so it was correctly kept as working (below). It only decayed once the subtree went childless — a long real tool is never wrongly cleared.

real claude working tool_use

_Captured by [`/be`](https://github.com/srid/agency) §5.2 — real Claude via chrome-devtools, Claude Code (model `claude-opus-4-8`)._

PR #1118, opened 2026-06-02
fix(terminal): keep file-ref links from swallowing a trailing period

srid:
**A file path that ends a sentence in terminal output was no longer clickable.** When prose like `There's now a single docs/plans/electricity.html.` scrolled by, the link grabbed the trailing sentence period too — pointing at a nonexistent `…electricity.html.` — so clicking it silently did nothing instead of opening the file in the Code tab.

The culprit is the reference regex in `lineRef.ts`: `.` is a legal path char (extensions, dotfiles), and the slash-containing branch's final segment matched greedily, so it happily ate the period at the end of the line.

### The fix

Anchor the whole reference with a trailing negative lookbehind so a match can **never end in a literal dot**:

```
…(?::(\d+)(?::\d+|-(\d+))?)?(?<!\.)
```

Only `.` is excluded from the tail — `+`, `@`, and `-` stay valid path tails, and a `:line` suffix ends in a digit, so the rule *only ever trims a bare path's trailing dot*:

| Terminal text | Linkified before | Linkified after |
| --- | --- | --- |
| `…a single docs/plans/electricity.html.` | `docs/plans/electricity.html.` ❌ (won't resolve) | `docs/plans/electricity.html` ✅ |
| `opened src/notes.txt...` | `src/notes.txt...` ❌ | `src/notes.txt` ✅ |
| `see src/foo.c++ now` | `src/foo.c++` ✅ | `src/foo.c++` ✅ (unchanged) |
| `fixed in src/app.ts:42.` | `src/app.ts:42` ✅ | `src/app.ts:42` ✅ (unchanged) |

### Tests

- Three new `lineRef` unit tests — trailing period on a slash-path, a run of trailing dots (ellipsis), and the `foo.c++` guard that proves the dot-only rule doesn't over-strip.
- A new e2e scenario reproducing the reported case end-to-end (`docs/plans/electricity.html.` → click → preview iframe opens with the file's content). It was confirmed red on the unfixed regex and green after. *The file is created via a subshell so the full path appears contiguously only in the period-bearing prose — otherwise the link hit-test lands on a clean setup line and masks the bug.*

### Try it locally

```sh
nix run github:juspay/kolu/dot-in-link
```

### DIFF
--- a/packages/client/src/ui/lineRef.test.ts
+++ b/packages/client/src/ui/lineRef.test.ts
@@ -143,6 +143,36 @@ describe("parseLineRefs", () => {
     expect(parseLineRefs("git init")).toEqual([]);
     expect(parseLineRefs("run Makefile")).toEqual([]);
   });
+
+  it("does not absorb a trailing sentence period into a slash-path", () => {
+    // The reported bug: `…a single docs/plans/electricity.html.` — the
+    // sentence period sat in the path char class, so the greedy match
+    // swallowed it and the link pointed at `…electricity.html.`, which
+    // never resolved. The reference must stop at the real filename.
+    const refs = parseLineRefs(
+      "There is now a single docs/plans/electricity.html.",
+    );
+    expect(refs).toHaveLength(1);
+    expect(refs[0]).toMatchObject({
+      path: "docs/plans/electricity.html",
+      text: "docs/plans/electricity.html",
+      startLine: null,
+      endLine: null,
+    });
+  });
+
+  it("drops a run of trailing periods (ellipsis after a path)", () => {
+    expect(parseLineRefs("opened src/notes.txt...")[0]).toMatchObject({
+      path: "src/notes.txt",
+      text: "src/notes.txt",
+    });
+  });
+
+  it("keeps a trailing `+` so C++-style extensions still link", () => {
+    // Only `.` is stripped from the end — `+`/`@`/`-` are legitimate path
+    // tails (`foo.c++`, `bin/g++`), so the dot-only rule must leave them.
+    expect(parseLineRefs("see src/foo.c++ now")[0]?.path).toBe("src/foo.c++");
+  });
 });
 
 describe("resolveLineRefPath", () => {
--- a/packages/client/src/ui/lineRef.test.ts
+++ b/packages/client/src/ui/lineRef.test.ts
@@ -173,6 +173,22 @@ describe("parseLineRefs", () => {
     // tails (`foo.c++`, `bin/g++`), so the dot-only rule must leave them.
     expect(parseLineRefs("see src/foo.c++ now")[0]?.path).toBe("src/foo.c++");
   });
+
+  it("keeps a :line suffix intact when a sentence period follows", () => {
+    // The end-anchored `(?<!\.)` lookbehind is placed after the whole
+    // regex rather than inside the slash-path branch — safe only because
+    // a `:line` suffix ends in a digit, so the lookbehind never trims it.
+    // This pins that invariant: moving the lookbehind into the path branch
+    // would change `:line` immunity, and this test would catch it.
+    const refs = parseLineRefs("see a/b.c:42.");
+    expect(refs).toHaveLength(1);
+    expect(refs[0]).toMatchObject({
+      path: "a/b.c",
+      startLine: 42,
+      endLine: 42,
+      text: "a/b.c:42",
+    });
+  });
 });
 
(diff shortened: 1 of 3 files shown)


srid:
## Codex ⇄ Claude debate

**🟢 Consensus** — reached in **1 round**, no commits required.

| 1 | ✅ Approved | 0 | — (nothing to dispute) | — |

> _Codex's summary:_ "The parser change and added unit/e2e coverage are narrowly scoped and sound: the regex backs off trailing sentence periods without dropping legitimate non-dot path suffixes, and the changed tests cover the reported behavior plus nearby edge cases."

_Reviewed `git diff origin/master`; no untracked files. Base `origin/master`, max rounds 5._

srid:
## Lowy ⇄ Hickey lens debate

**🟢 Consensus** — reached in **1 round**. Independent review counts: **lowy 0**, **hickey 2**. Both lenses then agreed on every finding's disposition (and, for the fix, on the exact plan).

| # | Origin | Finding | Location | Disposition | Commit |
| hickey-1 | hickey | Load-bearing `:line`-immunity invariant was untested | `lineRef.ts:62-64` | ✅ **fix** | `46948f9` |
| hickey-2 | hickey | Path-char definition fragmented across constants (`PATH_CHARS` / `PATH_CHAR_TEST` / the new end-anchor) | `lineRef.ts:42,64,98` | ⚪️ **drop** | — |

### hickey-1 — fixed
The placement of `(?<!\.)` at the *end of the whole regex* (rather than inside the slash-path branch) is what lets one anchor cover all branches without duplicating the path char class. That's safe **only** because a `:line` suffix ends in a digit, so the lookbehind never trims it — but that invariant lived solely in a comment. **lowy independently verified it empirically** (moving the lookbehind into the path branch makes `a/b.:42` collapse to `a/b` with the line dropped), so the placement genuinely is the load-bearing decision. Added a test that mechanically pins it:

```ts
parseLineRefs("see a/b.c:42.") // → path "a/b.c", startLine 42, text "a/b.c:42"
```

A future edit relocating the lookbehind would now turn the suite red instead of silently changing `:line` immunity.

### hickey-2 — dropped (by agreement)
The `PATH_CHARS` vs `PATH_CHAR_TEST` split (`~`/`/` differ) **pre-dates this PR**, and the lookbehind is a *positional* rule, not a membership one, so it can't fold into `PATH_CHARS`. Both lenses agreed the diff introduces no new disagreement — it's an observation for a future cleanup, not this PR's burden.

_Both lenses forced onto Opus; rationale on the three deliberate decisions (dot-only tail, end-anchor placement, subshell test setup) was supplied so they weren't re-flagged. Base `origin/master`._

srid:
## 🚔 Code Police

**All clear** — no commits required. Three passes over the diff (77 insertions, 1 deletion across `lineRef.ts`, `lineRef.test.ts`, `file-ref-link.feature`):

| Pass | Issues | Notes |
| --- | --- | --- |
| **Rules** | 0 | 24-rule checklist clean — no hand-rolled utility (the fix is a 1-char sentinel on an existing regex), no dead code, no silent swallows, e2e assertions use existing `pollFor`/`waitForFunction` steps. |
| **Fact-check** | 0 | Verified the lookbehind is safe for *all* suffix forms — `:N`, `:N-M`, and `:N:col` all end in a digit, so `(?<!\.)` never trims them; only a bare path's trailing dot is dropped. Multi-dot/ellipsis case backtracks correctly. `text: m[0]` reflects the accepted match, not the rejected greedy attempt. |
| **Elegance** | 0 | Minimal and precise; the 6-line regex comment is justified by the `comment-the-non-obvious` rule, and the ellipsis test's partial `toMatchObject` is intentional (it asserts only the trimming behavior). |

_No rule violations, logic errors, race conditions, or wishful thinking found._

srid:
## Evidence

This proves the fix: a file path that ends a sentence (e.g. `docs/plans/electricity.html.`) is now linkified correctly — the trailing period is excluded from the link, clicking it opens the right-panel Code-tab preview showing `<h1>electricity</h1>`. Before the fix, the trailing period was absorbed into the link text, the resolved path did not exist, and the click no-opped.



▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=file-ref-trailing-dot.mp4

---

Recorded by the project's own Cucumber+Playwright e2e harness (`features/file-ref-link.feature`, scenario `A trailing sentence period does not break a slash-containing file-ref`, `KOLU_EVIDENCE=1`) on a clean pu box at commit `ff47e89`.

PR #1119, opened 2026-06-02
Decay an abandoned `thinking` pill via the resumed-prompt signal (#1017)

srid:
**Spike to deploy on zest and confirm against the live `~/code/nixos-config` phantom.** The merged #1017 fix decays only a dangling `tool_use`; a trailing **`thinking`** (newest transcript entry is a `user` prompt) still spins the dock pill forever once the session is abandoned and resumed idle. That's the variant stuck on zest right now.

### Why `thinking` needed a different signal

A dangling `tool_use` is distinguishable because a *live* tool keeps a descendant process. A `thinking` turn is **childless and quiet whether live or abandoned** (awaiting the model's first token vs. resumed-idle), so the subtree probe alone can't tell them apart — which is why the earlier review deferred it.

The discriminator that *does* work: **prompt age vs. process age.**

| | trailing prompt timestamp vs. `startedAt` | verdict |
| --- | --- | --- |
| **Live turn** | prompt **postdates** `startedAt` (this claude received it) | keep — never cleared |
| **Resumed-idle phantom** | prompt **predates** `startedAt` (belongs to a killed instance; `claude -c` wrote a fresh `startedAt`) | decay to idle |

So `thinking` decays to `waiting` only when the prompt is **orphaned** *and* the transcript is stale past the window *and* the subtree is idle (a second confirm). `readSessionFile` now parses `startedAt`; `newestEntryTimestampMs` reads the trailing prompt's timestamp.

This is exactly the zest case: the nixos-config prompt (`11:35:49Z`) predates its resumed claude's `startedAt` (`11:35:57`), so it's orphaned → clears on deploy.

### Deploy & verify on zest

```sh
nix run github:juspay/kolu/fix/stale-thinking-pill-1017b -- --host <addr> --port <port>
```

(or your usual zest deploy pointed at this branch). Leave the `~/code/nixos-config` session untouched — on attach the watcher re-derives, sees the orphaned stale prompt with an idle subtree, and the pill should drop from spinning to **idle ☾** within the stale window (immediately, since that transcript is already long-stale).

> Spike — unit tests pass (131) and typecheck is clean, but the full review gauntlet + evidence run comes **after** you confirm it clears the real phantom.

### DIFF
--- a/packages/integrations/claude-code/src/agent-provider.ts
+++ b/packages/integrations/claude-code/src/agent-provider.ts
@@ -41,7 +41,15 @@ export const claudeCodeProvider: AgentProvider<SessionFile, ClaudeCodeInfo> = {
   },
 
   sessionKey(session) {
-    return session.sessionId;
+    // Keyed by transcript identity AND process identity. A `claude -c` resume
+    // reuses the same `sessionId` but writes a new session file with a fresh
+    // `pid` and `startedAt`; the watcher captures `session.pid`/`startedAt`
+    // once (for the #1017 subtree-idle probe and orphaned-prompt age check), so
+    // keying on `sessionId` alone would early-return on resume and leave the
+    // watcher probing the dead pid against a stale `startedAt`. Folding pid and
+    // startedAt into the key forces the orchestrator to recreate the watcher
+    // when either changes for the same transcript.
+    return `${session.sessionId}:${session.pid}:${session.startedAt ?? ""}`;
   },
 
   createWatcher(session, onChange, log) {
--- a/packages/integrations/claude-code/src/core.ts
+++ b/packages/integrations/claude-code/src/core.ts
@@ -328,11 +328,17 @@ export function deriveState(
   state: ClaudeCodeInfo["state"];
   model: string | null;
   contextTokens: number | null;
+  /** Epoch-ms timestamp of the entry the state was derived from (the newest
+   *  `user`/`assistant` entry), or null when it lacks a parseable `timestamp`.
+   *  Used to age a trailing `thinking` prompt against the session's `startedAt`
+   *  (the resumed-vs-live discriminator — see the #1017 module note). */
+  timestampMs: number | null;
 } | null {
   let stateAndModel: {
     state: ClaudeCodeInfo["state"];
     model: string | null;
   } | null = null;
+  let timestampMs: number | null = null;
   let contextTokens: number | null = null;
 
   for (let i = lines.length - 1; i >= 0; i--) {
@@ -341,6 +347,7 @@ export function deriveState(
     try {
       const entry: {
         type?: string;
+        timestamp?: string;
         message?: {
           stop_reason?: string | null;
           model?: string | null;
@@ -382,6 +389,13 @@ export function deriveState(
             model: null,
           }))
           .otherwise(() => null);
+        // Capture the timestamp of the very entry the state derives from — the
+        // newest `user`/`assistant` — so the orphaned-prompt age check reads the
+        // same entry, not a parallel walk (null if absent/unparseable).
+        if (stateAndModel !== null) {
+          const ts = entry.timestamp ? Date.parse(entry.timestamp) : Number.NaN;
+          timestampMs = Number.isNaN(ts) ? null : ts;
+        }
       }
 
       if (stateAndModel !== null && contextTokens !== null) break;
@@ -411,7 +425,7 @@ export function deriveState(
     if (bg.some((t) => t.runId !== null)) state = "running_background";
   }
 
-  return { state, model: stateAndModel.model, contextTokens };
+  return { state, model: stateAndModel.model, contextTokens, timestampMs };
 }
 
 // --- Background-task detection (dynamic workflows) ---
@@ -785,27 +799,6 @@ export function nextWorkflowStaleDeadline(
 // Sibling of the `running_background` decay (#1109), which handles the
 // `end_turn`-promotion half on its own workflow-journal signal; the states are
 // disjoint, so the paths never overlap.
-
-/** Timestamp (epoch ms) of the newest `user`/`assistant` entry — the same entry
- *  `deriveState` reads for state. Returns null when there is none or it lacks a
- *  parseable `timestamp`. Used to age a trailing `thinking` prompt against the
- *  session's `startedAt` (the resumed-vs-live discriminator above). */
-export function newestEntryTimestampMs(lines: string[]): number | null {
-  for (let i = lines.length - 1; i >= 0; i--) {
-    const raw = lines[i];
-    if (raw === undefined) continue;
-    try {
-      const e = JSON.parse(raw) as { type?: string; timestamp?: string };
-      if (e.type === "user" || e.type === "assistant") {
-        const t = e.timestamp ? Date.parse(e.timestamp) : Number.NaN;
-        return Number.isNaN(t) ? null : t;
-      }
-    } catch {
-      // skip malformed lines
-    }
-  }
-  return null;
-}
 
 /** How long the transcript may sit unwritten before a dangling `tool_use`
  *  becomes eligible to decay to `waiting`. A live tool writes the transcript as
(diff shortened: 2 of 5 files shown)


srid:
## Codex ⇄ Claude debate

**🟢 Consensus** in **2 rounds** (base `origin/master`).

> _Final verdict (codex):_ "The transient-decay logic now treats orphaned `thinking` separately from live prompts, and the F1 fix correctly makes watcher identity include both transcript and process identity. No remaining blocking or major issues."

| Round | codex | open major | Claude disposition | commit |
| 1 | ❌ changes requested | 1 (F1) | **fixed** — fold pid+startedAt into `sessionKey` | `f945be96` |

### F1 (major, resolved) — stale watcher identity defeats the fix on resume

**codex:** `sessionKey` keyed only on `session.sessionId`. A `claude -c` resume reuses the same `sessionId` but writes a new session file with a fresh `pid` and `startedAt`. The orchestrator's `reconcile()` early-returns on an unchanged key, so the existing `SessionWatcher` is **not** recreated — it keeps its once-captured `session.pid`/`startedAt`. The new #1017 logic would then probe `isClaudeSubtreeIdle(deadPid)` and age the trailing prompt against a **stale `startedAt`** — defeating orphaned-`thinking` detection on the exact resume path it targets.

**Claude:** Confirmed and fixed. `sessionKey` now returns `` `${session.sessionId}:${session.pid}:${session.startedAt ?? ""}` `` so the watcher is torn down and recreated whenever the resumed process identity changes for the same transcript. Verified the bug path through `readSessionFile` (`SESSIONS_DIR/${pid}.json`) → `resolveSession(state.foregroundPid)` → `reconcile()` key-equality early-return. 131 tests pass, typecheck clean.

> A great catch — the whole orphaned-prompt discriminator depends on `session.startedAt` being the **current** resumed claude's start time, not a stale closure.

_Posted by [`/codex-debate`](https://github.com/srid/agency) — Claude Code (model `claude-opus-4-8`)._

srid:
## Lowy ⇄ Hickey lens debate

**🟢 Consensus** in **1 round** (base `origin/master`, no police voice). Independent findings: **lowy=1, hickey=1**. One `fix`, one `drop`.

| Finding | Origin | Disposition | Location | Applied commit |
| **hickey-1** — newest-entry identification computed by two independent walks | hickey | ✅ **fix** | `core.ts` (`newestEntryTimestampMs` vs `deriveState`) | `718f1b23` |
| **lowy-1** — asymmetric probe eval (`promptOrphaned` eager vs `subtreeIdle` lazy) | lowy | ⬇️ **drop** | `session-watcher.ts` | — |

### hickey-1 (fixed) — one walk, one authoritative newest entry

"Which is the newest user/assistant entry" was decided **twice**: `deriveState` walked the tail to that entry for state; `newestEntryTimestampMs` re-walked byte-for-byte to the same entry for its timestamp — coupled only by a doc-comment invariant. If either walk's predicate drifted, the orphaned-prompt age check would silently age a *different* entry than the one that produced `thinking`, and the phantom would either never clear or a live turn would get cleared. **Fix:** `deriveState` now captures and returns `timestampMs` of the entry it matched; `newestEntryTimestampMs` and its second loop are deleted; `isTrailingPromptOrphaned` takes the precomputed `derived.timestampMs`. One walk, no invariant to keep in sync.

### lowy-1 (dropped) — subsumed by hickey-1

> _lowy (conceding his own finding):_ "My finding's cost basis was the eager full backward JSONL walk `promptOrphaned` ran on every event for every state. Once hickey-1 lands, that walk is gone — `promptOrphaned` is a single integer comparison `derived.timestampMs < session.startedAt`. Wrapping a trivial comparison in a thunk would defer nothing and actively mislead (`subtreeIdle` is a thunk because it spawns `ps` — a real cost). The remaining lazy/eager asymmetry is *justified by a real cost difference* — exactly the asymmetry to keep, not erase."

hickey agreed; correct action was hickey-1 alone. 131 tests pass, typecheck clean.

_Posted by [`/lens-debate`](https://github.com/srid/agency) — lenses on Claude Code (model `claude-opus-4-8`)._

srid:
## 🚓 code-police

Passes: **rules → fact-check → elegance**. One reuse fix; nothing else flagged.

| Finding | Pass | Disposition | Commit |
| --- | --- | --- | --- |
| Hand-rolled timestamp parse duplicates `parseIsoTimestamp` | reuse | ✅ **fixed** | (pushed) |

**Fixed — reuse:** `deriveState` hand-rolled `entry.timestamp ? Date.parse(...) : NaN` then `isNaN(ts) ? null : ts`. `kolu-transcript-core` (a declared workspace dependency) already exports `parseIsoTimestamp(ts: string | undefined): number | null` with the exact same NaN→null semantics — it even absorbs the `undefined` guard. Replaced the two-liner with `parseIsoTimestamp(entry.timestamp)`.

**Clean otherwise:** no redundant state/copy-paste; `isTrailingPromptOrphaned` is O(1), called once per transcript change; the `thinking` decay is threaded through the existing `decayTransientState` policy (no bandaid special-case layering). No TS errors; remaining lint warnings are pre-existing and unrelated.

_Posted by [`/code-police`](https://github.com/srid/agency) — Claude Code (model `claude-opus-4-8`)._

srid:
## Evidence

Validated against a **real, live phantom on macOS (zest, Darwin 26.5)** — the `~/code/nixos-config` claude session that had been spinning a "running" pill — rather than a synthetic capture. This is the authentic #1017 `thinking` variant.

### The live session (measured on zest)

```
session:        claude pid 95759, sessionId d2caff5f…, cwd /Users/srid/code/nixos-config
startedAt:      1780401578095   (claude -c resume)
newest entry:   user prompt, ts 2026-06-02T11:35:49.791Z → promptMs 1780400149791
deriveState:    thinking
descendants:    []                              → subtree-idle ✓
transcript:     quiet 392s                       → stale ✓
promptOrphaned: 1780400149791 < 1780401578095  → TRUE ✓   (prompt predates the resume)
```

All three decay conditions true → the fix settles it to `waiting` (idle). A live turn would have `promptMs > startedAt` (not orphaned) and would never be cleared.

### Before / after, by build identity

| kolu build | `promptOrphaned` logic | the nixos-config pill |
| --- | --- | --- |
| `8k72…` / `f4893…` / `dwqg…` (May, pre-fix) | absent | **spins `running` forever** (the bug) |
| `2pswfqr9…` (this PR, deployed to zest) | present | **idle** |

Deploying rebuilt the 7692 server to the fixed build but left three stale pre-fix `--stdio` backends running (a deploy-hygiene gotcha — filed separately). Once those were killed and **only the fixed build served the dock**, the phantom no longer showed `running` — confirmed by the maintainer on the live session.

### Coverage
`decayTransientState` (thinking-orphaned / live-not-orphaned / busy / not-yet-stale), `deriveState().timestampMs`, `readSessionFile` `startedAt`, and `hasNoDescendants`/`snapshotProcessTree` — 131 unit tests, green on both lanes; full CI green on `e269c568`.

> No screenshot of a synthetic repro: the orphaned-`thinking` phantom requires kill-mid-think → `claude -c` resume → 2-min window, and the live zest session already demonstrated the exact behavior on real claude.

_Captured by [`/be`](https://github.com/srid/agency) §5.2 — Claude Code (model `claude-opus-4-8`)._

srid:
## Codex ⇄ Claude debate (re-run on `156339aa`)

**🟢 Approved** — consensus in 1 round, **0 findings**. Re-run after the post-gauntlet commit that dropped the `subtree-idle` confirm from the `thinking` decay.

> _codex:_ "The change correctly threads `startedAt` into Claude session identity, derives the timestamp from the same transcript entry used for state, and limits `thinking` decay to stale prompts that predate the resumed process."

Validated live on macOS (zest): the `append` session — orphaned `thinking` with a persistent `chrome-devtools-mcp` child — now decays to idle (the subtree no longer blocks it), while live sessions stay working.

_Posted by [`/codex-debate`](https://github.com/srid/agency) — Claude Code (model `claude-opus-4-8`)._

PR #1122, opened 2026-06-02
Fix /lens-debate launch: hoist meta to first statement, inline model literal

### LINKED ISSUE #1120: /lens-debate fails to launch: meta isn't the first statement (and interpolates a variable)
## Symptom

Invoking `/lens-debate` (e.g. as the structural step of `/be`) aborts immediately, before any review runs:

```
Invalid workflow script: `export const meta = { name, description, phases }` must be the FIRST statement in the script
```

The whole gauntlet step is dead — no lowy/hickey review, no PR comment.

## Root cause

`.apm/skills/lens-debate/debate.workflow.js` (and the generated `.claude/skills/lens-debate/debate.workflow.js`) opens with a module-level constant **before** `meta`, then interpolates it **into** `meta`:

```js
const MODEL = 'opus'        // ← statement before meta

export const meta = {
  name: 'lens-debate',
  ...
  phases: [
    { title: 'Review', detail: '…', model: MODEL },   // ← variable in meta
    { title: 'Debate', detail: '…', model: MODEL },
    { title: 'Apply',  detail: '…', model: MODEL },
  ],
}
```

The current `Workflow` runtime enforces two rules this violates:

1. **`export const meta` must be the very first statement** — the leading `const MODEL = 'opus'` breaks this.
2. **`meta` must be a pure literal** — no variables/interpolation; `model: MODEL` breaks this.

The source comment ("`meta` is evaluated before inputs, so this lives module-level") reflects an **older runtime contract**; the current one is stricter.

## Why `/codex-debate` is unaffected

The sibling `.apm/skills/codex-debate/debate.workflow.js` already has `export const meta` as its first statement with **no** variable interpolation, so it launches fine. That's the shape lens-debate needs to match.

## Proposed fix (in the `.apm/` source, then `just ai::apm`)

Hoist `meta` to the top as a pure literal and declare `MODEL` **after** it (the rest of the script reads `MODEL` lazily at input-resolution time, well after `meta`):

```js
export const meta = {
  name: 'lens-debate',
  description: '…',
  phases: [
    { title: 'Review', detail: '…', model: 'opus' },
    { title: 'Debate', detail: '…', model: 'opus' },
    { title: 'Apply',  detail: '…', model: 'opus' },
  ],
}

const MODEL = 'opus'   // referenced by `const model = a.model || MODEL` below
// …rest unchanged…
```

Inlining `'opus'` in the three phase entries keeps `meta` literal; keeping a single `const MODEL` (now after `meta`) preserves the "one socket for the model" intent the original comment cared about. Update the stale comment accordingly.

> Note: this is a single-socket-violation-free change — `model` is still defined once. The only real constraint is **ordering** (meta first) + **no interpolation in meta**.

## Repro

1. On any branch with a non-empty diff vs `origin/master`, run `/lens-debate` (or `/be`, which calls it).
2. Observe the launch error above; the workflow never starts.

## Workaround used in the meantime

Ran an ephemeral corrected copy of the script (meta hoisted, model inlined) via `Workflow({scriptPath: /tmp/…})` — confirms the fix is exactly the reordering above. The generated `.claude/` copy needs regenerating from a fixed `.apm/` source.

_Filed during a `/be` run (PR #1118) where this blocked the lens-debate gauntlet step._

srid:
**`/lens-debate` aborted at launch** — and with it the structural step of `/be` — because its workflow script violated two rules the current `Workflow` runtime enforces: `export const meta` must be the *first* statement, and `meta` must be a *pure literal*. The script opened with `const MODEL = 'opus'` and then interpolated `model: MODEL` into each phase entry, tripping:

```
Invalid workflow script: `export const meta = {...}` must be the FIRST statement in the script
```

No lowy/hickey review, no PR comment — the whole gauntlet step was dead. `/codex-debate` was unaffected because its `meta` is already a literal-first block; that's the shape lens-debate now matches.

### The fix

Hoist `meta` to the top with `'opus'` inlined in all three phases, and move the single `const MODEL` socket *just after* it — the rest of the script reads `MODEL` lazily at input-resolution time, well after `meta` is evaluated.

```js
export const meta = {
  name: 'lens-debate',
  phases: [
    { title: 'Review', detail: '…', model: 'opus' },
    { title: 'Debate', detail: '…', model: 'opus' },
    { title: 'Apply',  detail: '…', model: 'opus' },
  ],
}

const MODEL = 'opus'   // one socket, now after meta
```

_The "one socket for the model" intent is preserved — `MODEL` is still defined exactly once, and the `model` input still defaults to it. The only real change is ordering plus dropping interpolation from `meta`._ The stale "`meta` is evaluated before inputs, so this lives module-level" comment is updated to reflect the stricter contract.

Edited in the `.apm/` source and regenerated via `just ai::apm`, so the `.claude/` and `.agents/` copies stay in sync.

Closes #1120

_Generated by [`/forge-pr`](https://github.com/srid/agency) on Claude Code (model `claude-opus-4-8`)._

### DIFF
--- a/.agents/skills/lens-debate/debate.workflow.js
+++ b/.agents/skills/lens-debate/debate.workflow.js
@@ -1,19 +1,25 @@
-// The model every lens/agent runs on. SKILL.md flags this as load-bearing
-// (lenses run on Opus, overriding their `model: sonnet` frontmatter) and model
-// migrations are a recurring change — keep it to one socket. `meta` is evaluated
-// before inputs, so this lives module-level; the `model` input below defaults to it.
-const MODEL = 'opus'
-
+// The Workflow runtime requires `export const meta` to be the FIRST statement
+// and a PURE LITERAL (no variable interpolation), so the model is inlined as
+// 'opus' in each phase below. The single `const MODEL` socket lives just after
+// meta — every other model reference in this script reads it lazily at
+// input-resolution time, well after meta is evaluated.
 export const meta = {
   name: 'lens-debate',
   description:
     'lowy + hickey review a diff independently in parallel, then debate every finding to consensus; apply the agreed fixes',
   phases: [
-    { title: 'Review', detail: 'lowy and hickey (and optionally code-police) review the diff independently, in parallel', model: MODEL },
-    { title: 'Debate', detail: 'lowy and hickey cross-examine every finding until they agree per-finding', model: MODEL },
-    { title: 'Apply', detail: 'implement each agreed fix as its own commit', model: MODEL },
+    { title: 'Review', detail: 'lowy and hickey (and optionally code-police) review the diff independently, in parallel', model: 'opus' },
+    { title: 'Debate', detail: 'lowy and hickey cross-examine every finding until they agree per-finding', model: 'opus' },
+    { title: 'Apply', detail: 'implement each agreed fix as its own commit', model: 'opus' },
   ],
 }
+
+// The model every lens/agent runs on. SKILL.md flags this as load-bearing
+// (lenses run on Opus, overriding their `model: sonnet` frontmatter) and model
+// migrations are a recurring change — keep it to one socket. Inlined into the
+// phase entries above (meta must be a pure literal); the `model` input below
+// defaults to it.
+const MODEL = 'opus'
 
 // ---------------------------------------------------------------------------
 // Inputs (passed via the Workflow tool's `args`)
--- a/.apm/skills/lens-debate/debate.workflow.js
+++ b/.apm/skills/lens-debate/debate.workflow.js
@@ -1,19 +1,25 @@
-// The model every lens/agent runs on. SKILL.md flags this as load-bearing
-// (lenses run on Opus, overriding their `model: sonnet` frontmatter) and model
-// migrations are a recurring change — keep it to one socket. `meta` is evaluated
-// before inputs, so this lives module-level; the `model` input below defaults to it.
-const MODEL = 'opus'
-
+// The Workflow runtime requires `export const meta` to be the FIRST statement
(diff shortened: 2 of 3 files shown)


srid:
## Lowy ⇄ Hickey lens debate

**✅ consensus** · 1 round · base `origin/master` · police off

Independent parallel review: **lowy 1 finding**, **hickey 0 findings**. Both lenses converged in a single round — no fixes applied.

| lowy | `meta.phases` literals expose display-only wires past the `MODEL` socket | `debate.workflow.js:11-13` | **drop** | — |

### Why the one finding was dropped

lowy raised that the three `model: 'opus'` literals in `meta.phases` reach past the `const MODEL` receptacle and hardcode the implementation — then **held `drop` itself**, and hickey concurred. Two reasons, agreed by both lenses:

- **Platform-forced, unremovable.** The Workflow runtime requires `export const meta` to be the *first statement* and a *pure literal* — interpolating `model: MODEL` is exactly what aborted the workflow at launch (the bug this PR fixes). No representation at this layer collapses the duplication without re-breaking launch.
- **Cosmetic blast radius.** `phases[].model` is phase-*descriptor* metadata (a display label), not the value agents execute on. Every `agent()` call runs on `const model = a.model || MODEL` (line 46). A forgotten update drifts the displayed label, never the actual model — and the two can legitimately diverge (pass `args.model='sonnet'` and execution follows it while descriptors still read `opus`). So it's two coincidentally-equal concerns, not one fragmented value.

The diff already carries the correct Lowy mitigation for an unavoidable leak: **bidirectional documentation** (the header comment points down to the socket; the `MODEL` comment points back up to the inlined phases).

> _No deadlock is possible here — both lenses are disinterested reviewers of someone else's diff, and they converged immediately._

PR #1125, opened 2026-06-02
Fix stale client assets after a deploy (user-gated, race-free reload)

srid:
**After a deploy, an open tab kept running the old client bundle until a hard force-reload.** A normal reload — even clicking the overlay's Reload button — would race the service-worker activation and serve the *old* precache. This hands the SW lifecycle to `vite-plugin-pwa`'s `registerSW` in **`prompt` mode**: a new build installs and waits, the overlay offers **"App updated — Reload"**, and Reload activates the waiting worker and reloads *only once it controls the page*.

### Two halves of the bug

| | Before | After |
| --- | --- | --- |
| **Detection** | A long-open tab never re-checked for a new build (no poll, no nudge), so it sat on the old precache indefinitely | Register with `immediate`, poll hourly as a backstop, **and nudge `checkForUpdate()` the instant the WebSocket reports a server restart** — a deploy is caught at once |
| **Reload** | `forceUpdateAndReload` did `registration.update()` then an immediate `location.reload()` — but `update()` resolves while the new worker is still *installing*, so the reload was served by the old worker; only a force-reload (which bypasses the SW) escaped | Reload goes through `updateServiceWorker(true)` → skip-waiting → `clientsClaim` → reload on `controllerchange`, i.e. *after* the new worker controls the page. **No race.** |

### How a deploy now flows

```
deploy → server restarts → WebSocket reconnects (new process id)
   │
   └─▶ App.tsx nudges checkForUpdate()  ─▶  SW finds new build, installs, WAITS
                                               │
                          onNeedRefresh ──▶ swUpdateReady() = true
                                               │
                       TransportOverlay: "App updated — [Reload]"
                                               │
                       click ─▶ updateServiceWorker(true) ─▶ controllerchange ─▶ reload (fresh)
```

The reload prompt now keys on *"a new build is actually ready"* rather than *"the server process restarted"*, so a server-only restart with unchanged assets no longer nags a pointless reload.

> **Why `prompt`, not `autoUpdate`:** a deploy must never reload an open terminal session out from under the user. `prompt` keeps the new worker waiting until the user clicks; the pinned plugin version (`1.2.0`) has no hook to defer `autoUpdate`'s unconditional reload-on-activation. `pwa.ts` becomes the sole registrar (`injectRegister: false`).

Unit-tested in `packages/client/src/pwa.test.ts` (registration wiring, the `swUpdateReady` flip, the reconnect nudge, the poll, and both reload paths). README transport section updated to match.

### Try it locally

```sh
nix run github:juspay/kolu/fix/sw-update-reload-race
```


srid:
## Codex ⇄ Claude debate

**✅ Consensus after 2 rounds.** Codex reviewed at **`xhigh`** reasoning effort.

| 1 | ❌ changes requested | 1 (major) | **fixed** F1 | `d652f6d` |

**F1 (major) — the no-service-worker dead path.** Codex caught that gating the reload prompt solely on the SW signal broke plain-HTTP/LAN mode: `registerSW` returns the `updateServiceWorker` function *even when no worker registers* (an inert no-op in prompt mode), and `onNeedRefresh` never fires without a worker — so after a deploy on `--host 0.0.0.0` (insecure context, no SW) there was *no* reload affordance at all, and `reloadForUpdate` took the no-op branch.

Fixed (verified against `vite-plugin-pwa@1.2.0`'s `register.js`): a single `serviceWorkerSupported` gate (the same `"serviceWorker" in navigator` check the library uses) is now the source of truth — `reloadForUpdate` branches on it (real `location.reload()` on HTTP/LAN), and `updateReady()` falls back to the `restarted` lifecycle signal when no worker is present, restoring the server-restart reload prompt. `TransportOverlay` stays dumb about SW lifecycle. Added a no-SW test block (production `initPwa()` with no `navigator.serviceWorker`).

srid:
## Lowy ⇄ Hickey lens debate

**✅ Consensus in 1 round** — independent parallel review (lowy: 2 findings, hickey: 2), then cross-examination to agreement on all 4. Both lenses on Opus.

| Finding | Origin | Disposition | Location | Commit |
| Tighten the SW boundary: move the dev-unregister into `pwa.ts` (`unregisterStaleServiceWorkers()`) so *all* `navigator.serviceWorker` access lives behind the one boundary — `index.tsx` no longer re-derives the support gate or touches the registry | hickey-1 | **fix** | `index.tsx` / `pwa.ts` | `cb8a3b9` |
| `serviceWorkerSupported` was `export`ed but has no external consumer — make it module-private so the receptacle's surface is just the four behavioral verbs (the wire that *selects* the generator stops being published) | lowy-1 | **fix** | `pwa.ts:44` | `e055805` |
| `pwa.ts` reads `rpc.ts`'s `lifecycle()` in the no-SW fallback (asymmetric transport→SW coupling) | lowy-2 | **drop** | `pwa.ts:64` | — |
| `updateReady()` folds transport state into `pwa.ts`, technically breaking the "join only at the composition root" invariant | hickey-2 | **drop** | `pwa.ts:64-68` | — |

**Why the two drops held:** both lenses independently verified the HTTP/LAN fallback has *no native ready-signal* — `onNeedRefresh` can only fire for an installed-and-waiting worker, which can't exist without a service worker — so `updateReady()` must borrow *some* freshness proxy, and `restarted` is the only honest one available. Relocating that single-line read to the composition root via a setter is bounded machinery for zero behavioral gain. The accessor unification (one `updateReady()` that `TransportOverlay` reads without reasoning about SW lifecycle) is the design's win and was kept.

The two fixes make `pwa.ts`'s "sole service-worker boundary" header claim *literally true*: the support gate is computed once, and no `navigator.serviceWorker` access leaks outside it.

srid:
## Code-police

Rules → fact-check → elegance. **One fix, three non-issues.**

**Fixed (`c7c9936a`):**
- `pwa.ts` update poll — dropped a dead `typeof navigator !== "undefined"` guard. The `setInterval` lives inside `onRegisteredSW`, which only fires *after* a worker registers (a browser context where `navigator` is always defined), so the guard was unreachable. Simplified to `if (!navigator.onLine) return;` — behaviorally identical (`navigator.onLine` is always boolean in a browser).

**Skipped (non-issues):**
- **Reuse** — no re-implementations; `lifecycle()`, `registerSW`, and the signal pattern all use existing idioms.
- **Efficiency** — the two `lifecycle()` subscriptions (App.tsx's restart nudge, `pwa.ts`'s `updateReady()` fallback) serve distinct purposes; no redundant work.
- **Altitude** — the restart-nudge `createEffect` belongs in `App.tsx`'s body, not `initPwa()`: `initPwa()` runs before `render()`, outside any reactive owner, so an effect there would be ownerless (a Solid error).

PR #1128, opened 2026-06-02
Resolve symlinks in repo-root path guard (close symlink escape)

srid:
**The repo-root guard was lexical only — it stopped `../../` traversal but followed symlinks, so a repo-local `leak -> /etc/passwd` passed the string check and the later filesystem call read straight out of the repo.** That's a correctness bug and a workspace-boundary failure for any untrusted checkout. This splits *lexical path normalization* from *filesystem authority*: existing-file reads now compare `fs.realpath` of the root and target before touching the file.

`resolveUnder` does `path.resolve` + `path.relative` on the string — it never consults the disk, so it can't know where a symlink actually points. Three surfaces leaned on it for operations that read real files:

```
                resolveUnder (lexical)        assertRealpathUnder (fs authority)
  child path ─▶ reject ../../ traversal  ─▶   realpath(root) vs realpath(target)
                returns {abs, rel}            reject if real target escapes real root
```

### Surfaces closed

| Surface | Operation | Guard now |
| --- | --- | --- |
| `kolu-git/browse.ts` | Code-tab `readFile` / `statFileMtimeMs` | `resolveExistingUnder` |
| `kolu-git/review.ts` | local untracked `git diff --no-index /dev/null <abs>` | `assertRealpathUnder`, scoped to the disk-read fallback only |
| `server/iframePreviewRoute.ts` | binary-preview byte serving | `assertRealpathUnder` in `serveResolvedFile` → **403** |

The new authority lives in one place — `safe-path.ts`, the module that decides *what path is allowed*:

- **`assertRealpathUnder(root, abs)`** — resolves symlinks on *both* sides (`realpath(root)` matters too: a symlinked checkout or macOS `/tmp -> /private/tmp` would otherwise make every legitimate file look like an escape). Any `realpath` failure (missing / `EACCES` / `ELOOP`) *fails open* — there's no resolvable file to leak, and the caller's own fs op faces the identical kernel checks and reproduces the errno, so a missing file stays a 404/`ENOENT` rather than masquerading as an escape.
- **`resolveExistingUnder(root, child)`** = `resolveUnder` + the above, for the common read/stat-an-existing-file case.
- **`resolveUnder`** stays the lexical helper for not-yet-existing paths and git pathspecs (`review.ts`'s `oldPath`, and tracked/branch diffs — where `rel` is only a pathspec git resolves against the tree and never dereferences, so realpathing it would make Code-tab diffs host-dependent).

> The preview route's `resolvePreviewPath` stays a *pure, no-I/O* function (its URL-decoding guard is unit-tested without fixtures); the symlink resolution lands in `serveResolvedFile`, the existing I/O half — consistent with that module's deliberate split.

### Tests

Symlink-to-outside **rejected**, inside-symlink **followed**, non-existent path **passes through**, intermediate directory-symlink escape **rejected**, symlinked root **stays valid**, tracked-symlink diff **still renders** — covered at the `safe-path` authority layer and at each of the three call sites (`browse`, `review`, preview route).

### Review gauntlet

Built start-to-finish with `/be`; every stage's trail is in the PR comments. **codex-debate** (consensus, 2 rounds, xhigh) caught a real regression — the guard was initially applied to *all* `getDiff` modes, which would reject legitimate tracked/branch symlink diffs host-dependently; rescoped to the `--no-index` disk-read path. **lens-debate** (lowy ⇄ hickey, consensus) removed a `root` value smuggled through the pure `PathResolution` result. **code-police** added the fail-open rationale inline in the catch. CI is green on `x86_64-linux` + `aarch64-darwin` (all 22 nodes).

Implements the `#symlink-escape` finding from [`docs/plans/correctness-review.html`](docs/plans/correctness-review.html) (high severity, the last open path-authority bug in that review).

### DIFF
--- a/docs/plans/correctness-review.html
+++ b/docs/plans/correctness-review.html
@@ -142,7 +142,7 @@
 <section id="summary">
   <h2>Summary</h2>
   <div class="callout">
-    <p><strong>Verdict:</strong> I found six correctness issues worth fixing. The high-severity terminal lifecycle leak is fixed by <a href="https://github.com/juspay/kolu/pull/1105">PR #1105</a>, and the OpenCode silent-parse issue by <a href="https://github.com/juspay/kolu/pull/1108">PR #1108</a>. The remaining open set is one high-severity path-authority bug and three medium-severity issues.</p>
+    <p><strong>Verdict:</strong> I found six correctness issues worth fixing. The high-severity terminal lifecycle leak is fixed by <a href="https://github.com/juspay/kolu/pull/1105">PR #1105</a>, the OpenCode silent-parse issue by <a href="https://github.com/juspay/kolu/pull/1108">PR #1108</a>, and the high-severity path-authority symlink escape by <a href="https://github.com/juspay/kolu/pull/1128">PR #1128</a>. The remaining open set is three medium-severity issues.</p>
   </div>
 
   <table>
@@ -159,7 +159,7 @@
         <td><span class="severity high">High</span></td>
         <td>Repo-root guard is lexical only; symlinks can escape the repo for reads, previews, and local diffs.</td>
         <td><code>kolu-git</code>, iframe preview</td>
-        <td><span class="status open">Open</span></td>
+        <td><span class="status fixed">Fixed</span> <a href="https://github.com/juspay/kolu/pull/1128">#1128</a></td>
       </tr>
       <tr>
         <td><span class="severity high">High</span></td>
@@ -216,6 +216,8 @@
     <p><span class="risk">Risk:</span> the Code tab, binary preview route, and local diff fallback can read or render files outside the repository root. That is both a correctness bug and a security boundary failure for any untrusted workspace.</p>
     <p><span class="fix">Fix:</span> split lexical path normalization from filesystem authority checks. For any operation that reads, stats, or diffs an existing file, compare <code>fs.realpath</code> of the root and target before use. Reject targets whose real path is outside the real repo root. Keep the current lexical helper only for paths that may not exist yet.</p>
     <p><span class="fix">Tests:</span> add symlink-to-outside fixtures for <code>readFile</code>, preview serving, and local untracked diff. The expected result should be a rejected path, not the external file content.</p>
+    <p class="label">Status - fixed</p>
+    <p><span class="fix">Implemented</span> in <a href="https://github.com/juspay/kolu/pull/1128">PR #1128, "Resolve symlinks in repo-root path guard (close symlink escape)"</a>. The fix takes the finding's recommended split: a new <code>assertRealpathUnder</code> in <a href="../../packages/integrations/git/src/safe-path.ts">safe-path.ts</a> resolves symlinks on <em>both</em> the root and the target via <code>fs.realpath</code> and rejects any target whose real path escapes the real root (realpath'ing the root too keeps a symlinked checkout or macOS <code>/tmp -&gt; /private/tmp</code> valid). Any <code>realpath</code> failure passes through, so a missing file stays a 404/<code>ENOENT</code> instead of masquerading as an escape. <code>resolveExistingUnder</code> composes it with the lexical <code>resolveUnder</code> for the common read/stat/diff-an-existing-file case; <code>resolveUnder</code> stays the lexical-only helper for not-yet-existing paths and git pathspecs. Wired into <code>browse.ts</code> (<code>readFile</code>/<code>statFileMtimeMs</code>), <code>review.ts</code> (the local untracked <code>--no-index</code> fallback's <code>filePath</code>), and the preview route's I/O half <code>serveResolvedFile</code> (which now 403s a symlink escape while <code>resolvePreviewPath</code> stays a pure, fixture-free unit). Covered by symlink-escape unit tests in <a href="../../packages/integrations/git/src/safe-path.test.ts">safe-path.test.ts</a>, <a href="../../packages/integrations/git/src/browse.test.ts">browse.test.ts</a>, <a href="../../packages/integrations/git/src/review.test.ts">review.test.ts</a>, and <a href="../../packages/server/src/iframePreviewRoute.test.ts">iframePreviewRoute.test.ts</a> - escape rejected, inside-symlink followed, non-existent passes through, intermediate dir-symlink escape rejected, symlinked root stays valid.</p>
   </article>
 
   <article class="finding high" id="terminal-exit-subscriptions">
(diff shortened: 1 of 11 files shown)


srid:
## Codex ⇄ Claude debate

**Consensus reached in 2 rounds.** Codex reviewed at `xhigh` reasoning effort (read-only sandbox); each round committed on the branch.

| 1 | changes requested | 1 major (`F1`) | fixed `F1` | `99fe8dbb` |
| 2 | **approved** | 0 | — (consensus) | — |

**`F1` (major, `review.ts`) — real regression caught.** The first cut applied the symlink-resolving guard (`resolveExistingUnder`) at the *top* of `getDiff`, before the code knew whether it would read from disk. For **branch-mode and tracked-local** diffs, `rel` is only handed to `git diff` as a *pathspec* — git resolves it against the tree and never dereferences the working-tree symlink — so a legitimate tracked symlink (e.g. `link -> /usr/bin/node`) was rejected with `PATH_ESCAPES_ROOT` whenever that target happened to exist on the reviewer's host, making Code-tab diffs host-dependent.

**Fix:** top-level validation reverted to lexical `resolveUnder`; `assertRealpathUnder(repoPath, abs)` now runs *only* inside the local-untracked `git diff --no-index /dev/null <abs>` fallback — the sole path that reads `abs` off disk. Regression coverage added: a tracked symlink-to-outside now renders in local mode, while untracked symlink-escape and lexical-traversal rejections still hold.

srid:
## Lowy ⇄ Hickey lens debate

**Consensus in 1 round.** Both lenses reviewed the diff independently (on Opus), then debated every finding to agreement. Independent finding counts: lowy 1, hickey 1.

| lowy | Path-authority guard ships three exports for one volatility axis — audited as a justified split, not leaked wires | `safe-path.ts:34-103` | **drop** | — |
| hickey | `root` smuggled through the pure resolution result to rejoin a split authority check | `iframePreviewRoute.ts` | **fix** | `d3226386` |

**lowy-1 (drop).** The electricity probe: the receptacle is the path-containment guard; the encapsulated volatile axis is the *authority level* of a containment decision (pure-lexical string math vs. `fs.realpath` authority) plus the fail-open-on-realpath-error policy. The three exports are two genuinely distinct authority levels plus their composition — not three leaked submodules of one primitive. The per-branch placement in `review.ts`/`iframePreviewRoute.ts` is the *security contract* (realpath scoped to exactly the branch that reads off disk), not accidental decomposition. The verdict interface (`PATH_ESCAPES_ROOT | ok{abs,rel}`) didn't change when realpath capability was added, so the receptacle stayed stable. Recorded as an *audited* drop.

**hickey-1 (fix, applied).** `root` is an *input* the caller (`index.ts:189`) already holds, not a *resolution output* — yet it was threaded onto `PathResolution`'s success variant and pulled back out in `serveResolvedFile` solely to bridge the pure/IO split. The fix drops `root` from `PathResolution` (back to `{ ok; abs; mime }` — pure outputs only) and passes it explicitly: `serveResolvedFile(res, repoRoot)`. The deliberate pure-resolver / IO-serve split stays intact; the smuggled input is gone.

srid:
## Code-police

Full rule checklist + fact-check + elegance passes over the gauntlet diff. **One finding, fixed; everything else clean.**

| Rule | Location | Disposition | Commit |
| --- | --- | --- | --- |
| `no-silent-error-swallowing` | `safe-path.ts` `assertRealpathUnder` catch | **fixed** — added inline comment | `fa884318` |

The `catch { return ok(undefined); }` fail-open had its rationale only in the function JSDoc; the rule wants it *inline in the catch*, since a reader meeting it in a diff or stack trace otherwise sees a security check silently passing on error. Behavior is unchanged and verified sound: on `ENOENT`/`EACCES`/`ELOOP` the caller's own `read`/`stat` faces the identical kernel checks and reproduces the errno, so failing open can't leak — a missing file stays a 404/`ENOENT` instead of masquerading as a path escape.

Everything else passed: no DRY-of-three (the escape idiom appears twice, below threshold), no untyped escape hatches, errors log at `error`, the three-export surface is curated (not a bridge module), and the `getDiff` realpath scoping + preview pure/IO split are commented at the non-obvious spots.

srid:
## Evidence

This is a backend path-authority guard — no UI surface, and kolu's e2e harness records *existing named Cucumber scenarios* (it has no step that plants a malicious symlink, and authoring one would be the hand-rolled script the evidence skill forbids). So the faithful proof of the now-fixed behavior is the symlink-escape test suite, which exercises the rejection across **all four surfaces** and encodes the before→after contrast directly in the assertions.

**The vulnerability, then the fix — same suite, adjacent assertions:**

```
✓ resolveUnder (lexical) > does NOT resolve symlinks — a symlink string still under root passes   ← old behavior (the hole)
✓ resolveExistingUnder   > rejects a symlink that escapes the root (the finding)                   ← the fix
✓ resolveExistingUnder   > rejects a path under an intermediate directory symlink that escapes
✓ resolveExistingUnder   > follows a symlink that stays inside the root
✓ resolveExistingUnder   > passes a non-existent path through so the caller's fs op can 404/ENOENT
✓ assertRealpathUnder    > resolves symlinks on the root too (symlinked checkout stays valid)
```

**Each of the three real call sites proven to reject the escape (not return external content):**

```
# Code tab — kolu-git/browse.ts
✓ readFile         > rejects a symlink that escapes the repo root
✓ readFile         > follows a symlink that stays inside the repo
✓ statFileMtimeMs  > rejects a symlink that escapes the repo root

# Local untracked diff — kolu-git/review.ts
✓ getDiff path authority > rejects a repo-local symlink escaping the root before the --no-index fallback reads it
✓ getDiff path authority > renders a tracked symlink pointing outside the repo (pathspec, never read off disk)

# Binary preview route — server/iframePreviewRoute.ts
✓ serveResolvedFile > 403s for a symlink that escapes the repo root (and never leaks content)
✓ serveResolvedFile > 404s for missing files (with valid path)   ← fail-open preserved, not masked as 403
```

`kolu-git` 22/22 and the preview route 29/29 pass locally; the same suites are green in CI on both `x86_64-linux` and `aarch64-darwin` (`ci::unit`, `ci::e2e`, and 8 other recipes — **all 22 nodes succeeded**).

srid:
## Evidence — live in the Code browser

Driven through the real Kolu Code tab (chrome-devtools on a dev build of this branch) against a throwaway repo `/tmp/symlink-demo` containing two symlinks that escape the repo:

```
leak.txt        -> /tmp/outside-secret/secret.txt   (a planted "host secret")
passwd-link.txt -> /etc/passwd
README.md       (a normal in-repo file)
```

Both symlinks are surfaced in the tree by `git ls-files`, so they're clickable nodes — exactly the attack path the finding describes.

**1. Baseline — a normal in-repo file reads fine.** `README.md` renders its content.

baseline: README renders

**2. BEFORE (guard reverted to lexical-only) — the secret leaks.** With `readFile` temporarily restored to the old `resolveUnder` (lexical) guard, clicking `leak.txt` renders the full external file straight into the Code pane — `TOP SECRET … /tmp/outside-secret/secret.txt … root:x:0:0:root:/root:/bin/bash … api_key=sk-live-DEADBEEF-do-not-leak`. This is the vulnerability.

before: external secret leaked into the Code pane

**3. AFTER (this PR) — same click, blocked.** With the fix live, clicking `leak.txt` shows `Error: path escapes root: /tmp/symlink-demo/leak.txt` (and a matching toast) — the external content is never read or rendered. `passwd-link.txt` (→ `/etc/passwd`) behaves identically.

after: rejected with path escapes root

> The "before" frame was produced by temporarily reverting only `browse.ts`'s `readFile` to the lexical guard on a local dev server to reproduce the leak, then reverting — the committed branch was never changed (working tree clean, `HEAD` unchanged). The Git: Local / Branch diff modes and the iframe binary-preview route are guarded the same way; this captures the Code-browser read path the question asked about.

PR #1130, opened 2026-06-02
fix: stop the dock pill sticking 'active' after a manual /compact

srid:
**A manual `/compact` left the dock pill / tile chrome spinning "Thinking" forever** while the agent sat idle at the prompt. `deriveState` walks the transcript tail back to the newest genuine turn — but a manual `/compact` (Claude Code v2.1.160) leaves a tail of **transcript-only `user` entries the human never typed**, and the generic `user` branch read the newest of them as a fresh prompt.

The real post-`/compact` tail, all typed `user`:

```
user  isCompactSummary   "This session is being continued…"
user                     <local-command-caveat>…
user                     <command-name>/compact</command-name>
user                     <local-command-stdout>…Compacted…</local-command-stdout>   ← newest
```

The fix walks past these to the genuine prior turn (an idle `end_turn` → `waiting`). It also fixes the same stuck pill after **any** no-op local command (`/config`, `/status`, …), not just `/compact`.

### What counts as "not a human prompt"

`deriveState` skips a trailing `user` entry when it is:
- the `/compact` summary (`isCompactSummary`), or
- slash-command bookkeeping — content beginning with `<command-name>` / `<command-message>` / `<command-args>` / `<local-command-stdout>` / `<local-command-caveat>`.

> **Deliberately *not* keyed on `isMeta`.** That flag marks *injected model input* — slash-command/skill expansions (`Base directory for this skill: …`), auto-continues, hook feedback, pasted images — which the agent is about to act on. Skipping those would falsely show **idle while Claude is working** on a prompt-style command like `/do`. (Caught in review; the one `isMeta` artifact that trails — `/compact`'s caveat — is already matched by content.)

### Tested

- **Unit** — the full realistic `/compact` tail → `waiting`; a no-op `/config` tail → `waiting`; a prompt-style `/do` tail (`isMeta` expansion) → `thinking`.
- **e2e** — `"compact"` mock state modelling the full tail; proven red (`thinking`) → green (`waiting`).
- **Live (zest)** — the stuck `append` session now reads idle; confirmed by the maintainer.
- **Real-transcript** — the actual `deriveState` over the live zest tail → `waiting`.

### Review trail
Full `/be` gauntlet on the final commit: **codex-debate** (consensus, caught + fixed the `isMeta` over-skip), **lens-debate** (consensus, decoupled token-accounting from the artifact skip), **code-police** (clean), **CI all 24 nodes green**.

### Try it locally

```sh
nix run github:juspay/kolu/fix/stuck-pill-after-compact
```


srid:
## Codex ⇄ Claude debate

**Consensus — 1 round.** Codex (reviewer, **`xhigh`** reasoning effort) approved the diff against `origin/master` on the first pass with **no blocking or major findings**, so no author round or commit was required.

| 1 | ✅ approved | 0 | — (nothing to rebut) | — |

> _Codex's read: "narrowly scoped — `deriveState` now skips synthetic `/compact` summary entries before deriving the visible agent state, with unit coverage for manual compact and a following real prompt, plus an e2e mock scenario."_ Tests were not executed (review is read-only).

_Generated by [`/codex-debate`](https://github.com/srid/agency) on Claude Code (model `claude-opus-4-8`)._

srid:
## Lowy ⇄ Hickey lens debate

**Consensus — 1 round.** Independent parallel review (both lenses forced onto Opus): **lowy = 2 findings, hickey = 0**. Both lowy findings were *design-endorsements* (not defects) and settled `drop` after hickey concurred from its own lens. **No fixes applied; tree unchanged.**

| Finding | Origin | Location | Disposition | Substance |
| `lowy-1` | lowy | `core.ts:372` | **drop** | Electricity test passes — `deriveState(lines, outstanding?) → {state, model, contextTokens, timestampMs}` is the single stable receptacle; the skip is a *new generator behind an unchanged interface* (Anthropic's transcript taxonomy is the real axis of change). No consumer learns about `isCompactSummary`; the e2e harness reuses the existing `MockState`/`buildTranscript` receptacle. _Placement is safe: the synthetic entry carries no `message.usage`, so skipping before the `contextTokens` read drops no real accounting snapshot._ |
| `lowy-2` | lowy | `core.ts:372` | **drop** | Symmetry check — skip (`continue`) vs classify (interrupt → `waiting`) are two *genuinely distinct events*, not one axis handled two ways: an interrupt marker **is** the terminal event (must end the walk); a compact summary is `isVisibleInTranscriptOnly` — a non-turn that must be seen *through* to the prior turn. Asymmetry tracks real semantics; a `isSyntheticNonTurn()` predicate would be speculative with one case. |

> _Both lenses converged: the guard sits at the correct boundary, blast radius is contained to `deriveState`, and the interface contract is unchanged._

srid:
## /code-police findings — fix/stuck-pill-after-compact

| Pass | Issues found | Details |
|------|-------------|---------|
| Rules | 1 | `no-dead-code`: `isVisibleInTranscriptOnly` in e2e mock |
| Fact-check | 0 | Clean |
| Elegance | 1 (fixed) | Same as no-dead-code finding above |

---

### Pass 1: Rule checklist

| Rule ID | Violation found? | What was identified | Action taken |
|---------|-----------------|---------------------|--------------|
| dry-rule-of-three | No | Checked by: `compactSummary` fixture appears twice in tests but content differs (different `message.content` strings) — rule of three threshold not reached, and the differing content is intentional | — |
| prefer-focused-library | No | Checked by: guard is a single `=== true` property check — no parser, tokenizer, or multi-step logic that a library would replace | — |
| invalid-states-unrepresentable | No | Checked by: `isCompactSummary?: boolean` is the correct optional-boolean type; no two fields that can't simultaneously be undefined | — |
| no-dead-code | **Yes** | `isVisibleInTranscriptOnly: true` in the e2e mock (`claude_code_steps.ts` line 220) is never read by any production code. No grep hit outside of that one mock line. | Fixed in commit `0600de2c` |
| no-silent-error-swallowing | No | Checked by: the `catch {}` at line 413 of `core.ts` is pre-existing (not in this diff); the new guard sits inside the existing try block and uses `continue`, not a swallow | — |
| no-unbounded-growth | No | Checked by: no new event handlers, watchers, or collections in the diff — pure in-memory loop guard | — |
| comment-the-non-obvious | No | Checked by: the 9-line comment above `if (entry.isCompactSummary === true) continue` explains the why (manual `/compact` leaves a synthetic tail entry, auto-compact is unaffected); the e2e mock block has an equivalent comment above it | — |
| no-re-export-bridge-modules | No | Checked by: diff adds no new `export … from` declarations | — |
| subscription-use-pending | No | Checked by: no subscription reads in the diff | — |
| solid-reactive-prop-passed-to-hook-must-be-reactive | No | Checked by: server-side integration code — no SolidJS components | — |
| no-untyped-escape-hatches | No | Checked by: grep'd for `unwrap`, `fromJust`, `assertNonEmpty`; zero matches in diff | — |
| toast-must-include-error-message | No | Checked by: no `toast` calls in diff | — |
| caught-error-must-not-collapse-to-empty | No | Checked by: the pre-existing `catch {}` skips malformed JSONL lines — not collapsing an error to empty state visible to the user | — |
| styling-tailwind-only | No | Checked by: no UI changes in diff | — |
| prefer-ts-pattern | No | Checked by: `if (entry.isCompactSummary === true) continue` is a pre-filter boolean guard before entering the existing `match()` dispatch — not a discriminated-union dispatch itself; exception for simple boolean checks applies | — |
| integration-perf-hygiene | No | Checked by: guard is O(1) in-memory check; no new fs.watch, Buffer, or listener | — |
| no-preference-prop-drilling | No | Checked by: no component prop changes in diff | — |
| errors-must-log-at-error | No | Checked by: no new error-logging in diff | — |
| subscription-must-surface-errors | No | Checked by: no new subscriptions in diff | — |
| e2e-poll-async-state | No | Checked by: the new e2e scenario reuses the existing "tile chrome should show an agent indicator with state" step which already polls via `POLL_TIMEOUT` loop (lines 399–411 of `claude_code_steps.ts`) | — |
| watcher-lifecycle-logs | No | Checked by: no new `fs.watch` calls in diff | — |
| silent-handler-required-on-void-subscriptions | No | Checked by: no void-returning subscription hooks in diff | — |
| migration-shape-guard | No | Checked by: no `Conf` migrations in diff | — |
| icons-in-registry | No | Checked by: no SVG in diff | — |

---

### Pass 2: Fact-check

No issues found. The guard correctly uses `=== true` (strict equality, not truthy) so a missing or non-boolean field passes through unaffected. The `continue` skips both `contextTokens` and `stateAndModel` extraction for compact entries — correct, since compact entries are `type: "user"` and carry no `message.usage` anyway. The auto-compact case is unaffected because a real agent response appends a newer entry after the summary, which the backward walk reaches first.

---

### Pass 3: Elegance

**Fixed:** Removed `isVisibleInTranscriptOnly: true` from the e2e mock fixture in `claude_code_steps.ts` — the field is never read by any production code (no-dead-code). Commit `0600de2c`.

**Skipped findings:**
- `compactSummary` constructed inline in both unit tests (not shared) — only 2 instances, below the rule-of-three threshold. Skip.
- Local `endTurn` in first compact test shadows the module-level fixture — intentional, model value `"claude-opus-4-8"` is tested in the assertion. Skip.

srid:
## Evidence

Same e2e fixture (a Claude session whose transcript ends in the synthetic `/compact` summary), toggling only the one-line `isCompactSummary` guard. The tile-chrome agent indicator is **stuck working** before the fix and **settles to idle** after.

| Before (guard disabled — pill stuck "thinking") | After (fix — idle "waiting") |
| --- | --- |
|  |  |

Captured by recording the `A trailing /compact summary reads as idle, not stuck working` scenario (`features/claude-code.feature`) under `KOLU_EVIDENCE=1` on an ephemeral pu box at commit `0600de2c`. Before-frame: same scenario with the guard commented out (the assertion times out waiting for `waiting` while the indicator sits at `thinking`).

srid:
## Codex ⇄ Claude debate (re-run on the corrected fix)

**Consensus — 2 rounds.** Codex (reviewer, **`xhigh`**) caught a real regression the broadened skip introduced, and the author round fixed it.

| Round | Codex | Open major | Author disposition | Commit |
| 1 | ❌ F1 (major) | 1 | **fixed** | `08e8ee39` |

**F1 — `isMeta` skip was too broad.** The classifier skipped `isMeta:true` user entries, but those are *overwhelmingly live model input* — slash-command/skill expansions (`Base directory for this skill: …`), auto-continues, hook feedback, pasted images. Skipping them would report **`waiting` while the agent is actually working** on a prompt-style slash command (`/do`, `/whatchanged`, `/be`). Verified empirically by the author against 342 real transcripts + 1248 `isMeta` samples.

**Fix:** dropped the `isMeta` clause from `isNonPromptUserEntry`. The `/compact` stuck-pill fix is fully preserved — its trailing artifacts (`<local-command-stdout>`, `<command-name>`, and the `<local-command-caveat>`) are all caught by **content** (`isLocalCommandArtifact`), and the summary by `isCompactSummary`. The incorrect `isMeta→waiting` test was replaced with a regression test asserting a prompt-style slash-command tail reads `thinking`.

> Re-verified against the live zest transcript tail post-fix: `deriveState` → `waiting` for the `/compact` idle tail, `thinking` for a `/do` prompt tail.

_Generated by [`/codex-debate`](https://github.com/srid/agency) on Claude Code (model `claude-opus-4-8`)._

srid:
## Lowy ⇄ Hickey lens debate (re-run on the corrected fix)

**Consensus — 1 round.** Independent parallel review (both on Opus): **lowy = 2, hickey = 1**. One `fix` applied.

| Finding | Origin | Location | Disposition | Substance |
| `hickey-1` | hickey | `core.ts:433` | **fix** → `10396a7c` | The `if (isNonPromptUserEntry(entry)) continue;` skip sat *above* the `contextTokens` read, braiding two independent concerns — "is this a fresh human prompt?" (state) and "does this entry carry a usage snapshot?" (token total). Correctness of the token walk relied on the unenforced assumption that artifacts never carry `usage`. **Fix:** hoisted the `contextTokens` accumulation above the skip, so a future artifact shape carrying usage can't silently drop the running total. |
| `lowy-1` | lowy | `core.ts:433` vs `:459` | drop | Symmetry observation: the skip (`continue`, walk-past) and the interrupt reclassify (in-place, inside the `match`) are two mechanisms for non-prompt user entries — but genuinely distinct semantics (interrupt *is* the newest event → classify it; artifact is a non-event → walk past). Unifying would complect, not encapsulate. The receptacle (lines→state) stays stable. _Both lenses noted hickey-1's fix shrinks this divergence._ |
| `lowy-2` | lowy | `core.ts:338` | drop | `isNonPromptUserEntry` re-checks `type !== "user"` though the caller already narrows — but that's the correct shape for a self-contained, independently-testable predicate; removing it would couple the helper to caller walk-position. |

> Net: the structural seam is clean, and the one fix decoupled token-accounting from artifact classification. _(A stray prettier reflow from the codex author round was re-formatted with biome in `884455be` to keep `ci::fmt` green.)_

srid:
## Code-police

Rules → fact-check → elegance on the cumulative diff (HEAD `884455be`). **Clean — no new findings.** The substantive issues were already caught and fixed upstream (codex `F1`: dropped the over-broad `isMeta` skip; lens `hickey-1`: hoisted token accounting above the artifact skip).

- **Rules** — no dead code; `toolResultText` reused for content flattening (DRY); `LOCAL_COMMAND_MARKERS` is a named constant; the e2e step polls (no fixed sleep).
- **Fact-check** — prefix-marker detection is the right discriminator (a real prompt never starts with `<command-…>`); `toolResultText` returns `""` for `tool_result`/non-text content, so genuine prompts and tool-results are *not* skipped; the `contextTokens` hoist is order-independent for non-skipped entries; the prompt-style slash-command path (`isMeta` expanded prompt → `thinking`) is covered both with and without the trailing prompt.
- **Elegance** — two small composable predicates, structural-typed param, comment density matches the module.

`just check` 0, `just fmt` clean, unit 139/139.

_Generated by [`/code-police`](https://github.com/srid/agency) on Claude Code (model `claude-opus-4-8`)._

srid:
## Evidence — update for the corrected fix (HEAD `884455be`)

The before↔after stills posted above still hold: the user-facing behavior — tile chrome / dock pill stuck **"Thinking"** before, settling to idle **"Waiting for input"** after — is identical under the corrected fix (only the *mechanism* widened, to skip the slash-command bookkeeping that trails the summary). The e2e `"compact"` scenario, now modelling the full realistic tail (summary + caveat + `<command-name>` + `<local-command-stdout>`), still goes red→green on this commit.

Stronger proof for the final fix, beyond the e2e stills:

- **Live on production (zest):** deployed the branch build; the `append` session that was stuck "Thinking" after `/compact` now reads idle. Confirmed by the maintainer.
- **Real-transcript check:** ran the actual `deriveState` over the real on-disk zest `append` transcript tail → `state: "waiting"` (idle). The prompt-style `/do` tail → `state: "thinking"` (the regression codex caught, fixed).

CI: **all 24 nodes green** on `884455be` (both `e2e` lanes passed first try, no flake).

_Generated by [`/evidence`](https://github.com/srid/agency) on Claude Code (model `claude-opus-4-8`)._

PR #1135, opened 2026-06-02
Show the client build commit beside srv, flag a stale bundle

srid:
The identity rail in the title bar showed only the **server** commit (`srv`), which left a real gap: an outdated client served from browser cache against a freshly deployed server looked identical to an up-to-date one. **This adds a `client` column rendering the commit this browser's JS was built from** (`__KOLU_COMMIT__`, baked in at build time and until now surfaced only in the About dialog), and flags a mismatch.

When `srv` and `client` **both report clean refs and they disagree** — the signature of a cached old bundle — the column shows a `stale` pill with a reload hint. A `dev` or `-dirty` build on either side never trips it, so local development stays quiet; the warning only fires against real deployed builds where it's actionable.

### Where the logic lives

The ref guard and the comparison move out of the JSX into a small pure module so the link guard and the staleness check can't drift, and so both are unit-testable:

| Helper (`ui/commitRef.ts`) | Used by |
| --- | --- |
| `isCleanRef(sha)` — real SHA, not `dev`, not `-dirty` | `Commit.tsx` (renders a GitHub link only for clean refs) + the staleness check |
| `clientIsStale(server, client)` — both clean **and** disagree | `IdentityRail.tsx`'s `stale()` derivation |

`commitRef.test.ts` table-tests both across the dev / dirty / absent / agree / disagree matrix (12 cases).

> _Phrasing note: hashes alone prove **difference**, not ordering — the tooltip says "older than the server" because a cache miss is the realistic cause, but the strict claim is "doesn't match." Easy to soften if reviewers prefer._

### Try it locally

```sh
nix run github:juspay/kolu/client-commit
```


srid:
## Codex ⇄ Claude debate

**Consensus** ✅ — reached in **1 round**. Codex reviewed at `xhigh` reasoning effort and approved with **no blocking or major findings**; the author side never had to run.

| Round | Codex verdict | Open blocking/major | Claude disposition |
| --- | --- | --- | --- |
| 1 | approved | 0 | — (nothing to rebut) |

Codex raised two **minor** points, both accepted and fixed in follow-up commits:

- **F1** (`IdentityRail.tsx`) — the tooltip claimed the client was *older*, but the predicate only proves the two clean refs *differ* (rollback/mixed-routing could make the client newer). Reworded to a mismatch, and the pill changed `stale` → `≠ srv` to match the existing `≡ in-process` convention.
- **F2** (`README.md`) — the chrome-bar description still listed only `srv` + a future `pty`; added the new `client` column and the mismatch indicator.

srid:
## Lowy ⇄ Hickey lens debate

**Consensus** ✅ — 1 round. Independent parallel review: **lowy 1 finding, hickey 0**. Both lenses ran on Opus, reviewed the diff independently, then debated to agreement. **No fixes applied** — the one finding was dropped by mutual agreement.

| Origin | Finding | Location | Disposition |
| --- | --- | --- | --- |
| lowy | `commitRef.ts` blends a domain-agnostic ref predicate (`isCleanRef`) with an app-level staleness verdict (`clientIsStale`) | `commitRef.ts:9-22` | **drop** (agreed) |

**Why drop:** the module encapsulates exactly one genuine volatility axis — the git-ref-marking convention (`dev` sentinel, `-dirty` suffix). `isCleanRef` is the stable receptacle both consumers (the `Commit.tsx` link guard, the `IdentityRail` staleness check) plug into; a grep confirmed those literals appear nowhere else in `client/src`, so no consumer reaches past it. `clientIsStale` is the sole derivation, built *on* the receptacle (it calls `isCleanRef`, never re-inlines the rule), with a single consumer — and the pure-predicate + verdict + colocated `.test.ts` shape mirrors the existing `terminal/staleness.ts` precedent verbatim. Splitting the two functions into separate files would manufacture an expendable boundary for zero volatility gain — the over-engineering both lenses warn against. *Revisit only if `clientIsStale` grows non-ref-hygiene siblings (timestamp/version-range staleness); speculative today.*

srid:
## Code-police

Clean across all three passes — no fixes applied.

| Pass | Issues | Notes |
| --- | --- | --- |
| **Rules** (24-rule checklist) | 0 | `isCleanRef` centralised below the rule-of-three threshold; old inline guard fully removed (no dead code); no error-swallowing, no untyped escape hatches, Tailwind-only styling. |
| **Fact-check** (logic/correctness) | 0 | `isCleanRef` gates all three failure modes (`dev`, `-dirty`, absent); `clientIsStale` is `false` while disconnected (no first-render flash) and only fires on two clean, disagreeing refs; `__KOLU_COMMIT__` is always a `string` at runtime. |
| **Elegance** (reuse/simplify/efficiency/altitude) | 0 | All four lenses clean. |

*Noted, not actioned: `clientIsStale`'s table covers `server=undefined` but not `client=undefined` — `__KOLU_COMMIT__` is typed `string` so that path is unreachable in practice; a harmless coverage gap, not a bug.*

srid:
## Evidence

Driven on a live kolu with the client and server started from **deliberately divergent** `KOLU_COMMIT_HASH` values — real code path, no mocking. The server reports `0784979` (the commit from the original screenshot); the client bundle is built from a different commit.

**Mismatch — the new `client` column + `≠ srv` badge fire** (client `a1b2c3d` ≠ server `0784979`). Tooltip: *"This client build doesn't match the server — reload to pick up the server's version."*

client vs server mismatch

**Matched — badge absent** (client and server both `0784979`). The `client` column is always shown; the warning only appears on a real divergence, so an up-to-date client stays quiet.

client matches server

> Both refs must be *clean* (not `dev`, not `-dirty`) for the badge to fire, so local dev — where both read `dev` — never false-positives. Verified live: `wsStatus: open`, `__KOLU_COMMIT__` reads the injected build commit, and the `≠ srv` pill is present only in the mismatch case.

PR #1136, opened 2026-06-02
fix(restore): don't show an empty canvas before the saved session loads

srid:
**On a cold launch with terminals from a previous session, Kolu sometimes flashed an empty canvas instead of the session-restore dialog — and the dialog only appeared after a full reload.** The empty-vs-restore decision was being made before the saved-session snapshot had arrived, so the UI claimed *"nothing to restore"* while that snapshot was still in flight.

### The race

The render gate keys off a single signal, `isLoading()`, which was defined as just `store.listSub.pending()` — *"has the terminal list reported yet?"*. On a cold launch the list yields `[]` (the terminals were killed on the previous shutdown) **before** the `session` cell reports:

```
list yields []  ─▶ isLoading() = false  ─▶ bare empty state renders
                                            (savedSession() still null —
session cell yields ────────────────────▶  restore card never gets its turn)
```

Because the `session` cell doesn't re-push its already-loaded value to that subscription, only a full reload — a fresh subscription — recovered the card. The server side was never at fault: the saved session lives in a synchronous `conf` store and was present the whole time. This is purely a client first-snapshot ordering bug.

### The fix

Gate the empty-vs-restore decision on the saved-session cell too. When the list is empty, stay in the existing **"Connecting…"** state until `savedSessionSub` has reported — *then* decide, with the snapshot in hand. When terminals exist the canvas renders immediately; the session cell is irrelevant in that case, so it's never waited on.

```ts
isLoading: () =>
  store.listSub.pending() ||
  (store.terminalIds().length === 0 && savedSessionSub.pending())
```

The new unit test pins both halves: the gate **stays loading** on an empty list until the session cell reports (red against the old single-input gate), and **never delays the canvas** when terminals exist.

> _The latent gap has existed since the loading flag was first wired to the list alone (#361); recent boot-timing shifts surfaced it as a regular flake. If an intermittent empty-then-reload persists after this, the next thread is a deeper transport bug where a cold-launch subscription receives a stale `null` snapshot that never re-pushes._

### Try it locally

```sh
nix run github:juspay/kolu/fix/restore-loading-gate
```

### DIFF
--- a/packages/client/src/terminal/useSessionRestore.test.ts
+++ b/packages/client/src/terminal/useSessionRestore.test.ts
@@ -1,0 +1,110 @@
+import type { TerminalId, TerminalInfo } from "kolu-common/surface";
+import { createRoot } from "solid-js";
+import { describe, expect, it, vi } from "vitest";
+
+// `isLoading` is a pure read over three reactive inputs: the terminal list's
+// pending flag, the live terminal count, and the saved-session cell's pending
+// flag. Drive those three through a plain hoisted bag and stub every module
+// `useSessionRestore` pulls at import time so the hook loads under Node without
+// a live socket, toast DOM, or the SSR-only `solid-js/web` build.
+const h = vi.hoisted(() => ({
+  listPending: true,
+  list: undefined as TerminalInfo[] | undefined,
+  terminalIds: [] as TerminalId[],
+  sessionPending: true,
+  savedSession: null as unknown,
+}));
+
+vi.mock("../wire", () => ({
+  client: {},
+  savedSessionSub: { pending: () => h.sessionPending },
+  savedSession: () => h.savedSession,
+}));
+vi.mock("../rpc/rpc", () => ({ lifecycle: () => ({ kind: "connected" }) }));
+vi.mock("../right-panel/useRightPanel", () => ({
+  useRightPanel: () => ({ seedPanel: () => {} }),
+}));
+vi.mock("./useSubPanel", () => ({
+  useSubPanel: () => ({
+    seedPanel: () => {},
+    getSubPanel: () => ({ activeSubTab: null }),
+    setActiveSubTab: () => {},
+  }),
+}));
+vi.mock("solid-sonner", () => ({
+  toast: Object.assign(() => {}, {
+    loading: () => 0,
+    success: () => {},
+    error: () => {},
+    warning: () => {},
+  }),
+}));
+vi.mock("anyagent/cli", () => ({ resumeAgentCommand: () => null }));
+
+import { useSessionRestore } from "./useSessionRestore";
+import type { TerminalStore } from "./useTerminalStore";
+
+/** A `TerminalStore` whose `listSub`/`terminalIds` read the hoisted bag, so a
+ *  test can flip a flag and call `isLoading()` to observe the gate directly. */
+function makeStore(): TerminalStore {
+  const listSub = Object.assign(() => h.list, { pending: () => h.listPending });
+  return {
+    listSub,
+    terminalIds: () => h.terminalIds,
+    getMetadata: () => undefined,
+    setActiveSilently: () => {},
+    activeId: () => null,
+    setMruOrder: () => {},
+  } as unknown as TerminalStore;
+}
+
+const mount = () =>
+  useSessionRestore({
+    store: makeStore(),
+    handleCreate: vi.fn(),
+    handleCreateSubTerminal: vi.fn(),
+  });
+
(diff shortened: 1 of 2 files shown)


srid:
## Codex ⇄ Claude debate

**✅ Consensus — 1 round.** codex reviewed at `xhigh` reasoning effort.

| 1 | **Approved** | 0 | — (nothing to rebut) | _no change_ |

codex confirmed the diff is limited to `useSessionRestore.ts` + the new `useSessionRestore.test.ts`, that the loading gate now waits for the saved-session cell before rendering the empty state when no terminal ids are present, and that the tests cover both the empty-list race and the non-empty-terminal case. No changes were required.

_Generated by [`/codex-debate`](https://github.com/srid/agency) on Claude Code (model `claude-opus-4-8`)._

srid:
## Lowy ⇄ Hickey lens debate

**✅ Consensus — 2 rounds.** Independent parallel review (lowy: 1 finding · hickey: 2 findings), then debated to agreement per finding.

| lowy | Empty-vs-restore gate reads the same raw cell as two sibling predicates | `useSessionRestore.ts` isLoading | **drop** (correct encapsulation — receptacle `isLoading()` stable; the shared `savedSessionSub.pending()` answers a distinct question at each of its 3 sites) | — |
| hickey | "Saved-session snapshot has arrived" derived in two places | `useSessionRestore.ts:325` | **drop** | — |
| hickey | isLoading comment narrated `listSub`, but the code reads `terminalIds()` | `useSessionRestore.ts` | **fix** | `add423d5` |

### Notable exchange
hickey-1 first proposed replacing the `terminalIds().length === 0` clause with a `!hydrated()` signal to dedupe the readiness precondition. lowy rebutted that this would silently re-key the gate off the **raw** `listSub()` (what the hydration effect branches on) instead of the **metadata-filtered** `terminalIds()` — the exact signal the consumer `showEmpty` reads at `App.tsx:398`. The two diverge transiently (raw list non-empty but metadata not yet landed), so the "dedup" would trade a benign shared-input for a real coupling to a *different* source than the one the answer is consumed from. hickey conceded; finding dropped.

The one applied fix (`add423d5`) just makes the comment honest — it now names `terminalIds()` (metadata-derived top-level ids) and records that the gate and the empty-state branch agree on what "empty" means. No behavior change.

srid:
## code-police review — fix/restore-loading-gate

Reviewed the `isLoading` gate change in `useSessionRestore.ts` and its new unit test. Three passes (rules, fact-check, elegance).

### Summary

| Pass | Issues | Details |
|---|---|---|
| Rules | 0 | Clean — all 24 built-in + project rules checked |
| Fact-check | 1 | Comment inaccuracy — "When terminals exist" overstated the gate's behavior |
| Elegance | 0 | Clean across reuse, simplification, efficiency, altitude |

### Fact-check finding (fixed in 49a199f1)

**`useSessionRestore.ts` comment at the `isLoading` return, line ~322:**

The comment said:

> "When terminals exist, the canvas renders immediately — the session cell is irrelevant."

This is inaccurate. `terminalIds()` is a memo over `meta.keys().filter(id => m && !m.parentId)` — it excludes terminals whose per-terminal metadata hasn't arrived yet. In the transient window where `listSub.pending()` is false (list arrived) but all per-terminal metadata subscriptions are still in-flight, `terminalIds()` is `[]` and `savedSessionSub.pending()` is still true, so `isLoading()` returns `true`. The comment claimed the canvas would unblock here, but it doesn't.

Fixed by clarifying "when any terminal's metadata has arrived" and adding a note about the transient metadata-pending window. Also corrected the App.tsx cross-reference from `:398` to `:397` (the `const showEmpty` declaration is at line 397).

### Elegance — altitude check

The fix puts the session-pending gate in `isLoading()` rather than in App.tsx's `showEmpty`. This is the right depth: `isLoading` is the natural owner of "canvas isn't ready to render yet", and `App.tsx` shouldn't need to know about `savedSessionSub`. Clean.

srid:
## Evidence



The clip shows kolu launching from an empty state: the app opens, stays in "Connecting…" while the saved-session snapshot arrives, then surfaces the restore card announcing "2 terminals" — no bare empty-canvas flash in between. Clicking restore brings both terminals back fully. This confirms the fix: the empty-vs-restore decision is now gated on the saved-session cell reporting, so the restore card surfaces cleanly on first load without the previous flicker.

▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=restore-loading-gate.mp4

_Generated by [/be](https://github.com/srid/agency) on Claude Code (model claude-opus-4-8)._

PR #1139, opened 2026-06-02
Pin ps via Nix so the phantom-pill probe never leans on PATH

### LINKED ISSUE #1121: Harden `ps` resolution for the agent-state subtree probe (pin via Nix, don't rely on global PATH)
## Context

The phantom-pill decay added in #1115 (and extended in #1119) confirms a session is abandoned by probing claude's process subtree via `snapshotProcessTree()` in `packages/integrations/claude-code/src/core.ts`:

```ts
execFileSync("ps", ["-A", "-o", "pid=,ppid="], { … })
```

This resolves `ps` through the **runtime `PATH`** and treats any failure as fail-open (`null` → "assume working" → never decays). So if the kolu server ever runs in an environment whose `PATH` lacks `ps`, the decay silently never fires and the pill spins forever again — a hard-to-spot regression.

## Not currently broken — this is hardening

Diagnosed live on a macOS (Darwin 26.5) zest deploy: the launchd+nix kolu server's `PATH` **does** include `/usr/bin:/bin`, `/bin/ps` exists, and `ps -A -o pid=,ppid=` parses to clean two columns. So the probe works today on both Linux and macOS. This issue is about removing the latent `PATH` dependency, not fixing a live failure (the live zest bug was the `thinking` variant, fixed in #1119).

## Proposal

Pin `ps` the way we already pin `gh` / `git` in Nix, and reference the absolute path from the build, so the probe never depends on the ambient `PATH`.

**Wrinkle to resolve:** unlike `gh`, `ps` is not one cross-platform package — nixpkgs `procps` is Linux; macOS needs its system BSD `ps` (which reads kernel structures differently and isn't the procps binary). Options to evaluate:
- Resolve a platform-appropriate `ps` in Nix (`procps` on Linux, system `/bin/ps` on Darwin) and thread the absolute path into the server (env var or build substitution), or
- Keep `execFileSync("ps", …)` but pass an explicit `env.PATH` that always includes `/bin:/usr/bin:/usr/sbin` plus the nix profile (cheap, no Nix packaging change), or
- A small candidate-path resolver (`/bin/ps`, `/usr/bin/ps`, then `PATH`).

The middle option is the smallest change and likely sufficient; the first is the most "correct" / PATH-independent.

## Acceptance
- The subtree probe resolves `ps` without depending on the runtime `PATH`.
- Works on `x86_64-linux`, `aarch64-darwin` (and ideally NixOS where `ps` is under `/run/current-system/sw/bin`, not `/bin`).
- A unit/integration check that `snapshotProcessTree()` returns a populated table in the server's build/runtime context.

## References
- #1017 — the stale-pill bug.
- #1115 — `tool_use` decay (introduced `snapshotProcessTree` / `isClaudeSubtreeIdle`).
- #1119 — `thinking` decay (resumed-prompt signal).
- `packages/integrations/claude-code/src/core.ts` → `snapshotProcessTree`.

srid:
**The phantom-pill decay's process-subtree probe now resolves `ps` from a Nix-pinned absolute path instead of the ambient runtime `PATH`.** `snapshotProcessTree()` ran `execFileSync("ps", …)` and *failed open* — any failure returns `null`, which the decay reads as "assume working" and never decays. So if the kolu server ran where `PATH` lacked `ps`, the pill would silently spin forever, re-introducing the [#1017](https://github.com/juspay/kolu/issues/1017) stale-pill bug with no obvious cause.

Not currently broken — this is hardening. On the live macOS/launchd+nix deploy the probe works today; this removes the *latent* PATH dependency rather than fixing a live failure.

### Approach — pin `ps` the way `gh` is pinned

`gh` is already pinned as `KOLU_GH_BIN` in `nix/env.nix` and consumed via a lazy `getGhBin()`. This mirrors that exactly:

- **`nix/env.nix`** — a new `KOLU_PS_BIN`, resolved *per-platform*. Unlike `gh`, `ps` is not one cross-platform package: nixpkgs `procps` is Linux-only, while macOS needs its system BSD `/bin/ps` (it reads kernel structures the procps binary can't). The Darwin branch is a plain system path, not a store path.
- **`default.nix`** — threads `KOLU_PS_BIN` into the wrapper via `--set`, and adds `procps` to the Linux wrapper `PATH` for parity (macOS keeps its system `/bin/ps`).
- **`core.ts`** — a lazy `getPsBin()` mirroring `getGhBin()`. It throws when unset rather than silently falling back to a bare `"ps"` lookup; `snapshotProcessTree()` calls it *inside its existing `try`*, so an unset env still **fails open to `null`** there instead of crashing the watcher.

This also covers NixOS, where `ps` lives under `/run/current-system/sw/bin` rather than `/bin` — the pinned store path sidesteps the question entirely.

### What's tested

| Case | Expectation |
| --- | --- |
| Pinned probe in the build/runtime context | populated table including this process |
| Spawned child | detected as a descendant (the genuine-work signal) |
| `KOLU_PS_BIN` unset | `null` — fails open, never de-escalates a working pill |
| Bogus `KOLU_PS_BIN`, `ps` still on `PATH` | `null` — proves it uses the pin, *not* `PATH` |

Closes #1121.

### Try it locally

```sh
nix run github:juspay/kolu/ps
```


srid:
## Codex ⇄ Claude debate

**✅ Consensus** — reached in **1 round**. codex reviewed at **`xhigh`** reasoning effort and approved with no blocking or major findings.

| 1 | Approved | 0 | — (nothing to dispute) | — (no changes needed) |

> codex's summary: _"The change consistently threads `KOLU_PS_BIN` through the Nix shared environment, dev shell, and packaged wrapper, and `snapshotProcessTree` now uses that pinned binary while preserving the existing fail-open behavior for probe failures. The updated unit coverage exercises the real pinned path, unset env, and bad env cases. No blocking or major correctness issues."_

PR #1140, opened 2026-06-02
fix(server): surface port-in-use as an actionable message, not a fatal crash

### LINKED ISSUE #1137: serve() has no 'error' handler — port-in-use (EADDRINUSE) surfaces as an uncaught fatal, not an actionable message
## Summary

The HTTP server started in `packages/server/src/index.ts:245‑269` (`serve({...})` from `@hono/node-server`) has **no `error` event listener**. When the listen port is already in use (`EADDRINUSE`) — or the bind fails for any other reason (`EACCES` on a privileged port, an unresolvable `--host`) — Node emits an `'error'` event on the underlying `http.Server` with no listener, so it rethrows.

That rethrow is caught by the global `uncaughtException` handler at `index.ts:148‑151` and logged as a generic `"uncaught exception"` fatal, then `process.exit(1)`. So it doesn't crash *silently*, but the user sees a fatal stack-shaped log line rather than an actionable message.

## Why it matters

Running kolu twice on the same port (or leaving an old instance alive) is one of the most common first-run friction points. For a first public release, the expected experience is a one-line, human-readable message like:

> Port 7681 is already in use. Is kolu already running? Try `--port <n>`.

…and a clean `exit(1)` — not a fatal that reads like an internal crash.

## Repro

1. Start kolu: `nix run github:juspay/kolu` (binds 127.0.0.1:7681)
2. In another shell, start it again on the same port.
3. Observe the second instance logs a `"uncaught exception"` fatal carrying `EADDRINUSE` instead of a friendly port-in-use message.

## Suggested fix

Attach an `error` handler to the server returned by `serve()` that special-cases `EADDRINUSE`/`EACCES` with a readable message and exits cleanly:

```js
const server = serve({ ... }, (info) => { /* listening */ });
server.on("error", (err) => {
  if (err.code === "EADDRINUSE") {
    log.fatal(`Port ${port} is already in use — is kolu already running? Try --port <n>.`);
  } else if (err.code === "EACCES") {
    log.fatal(`Permission denied binding ${host}:${port} (privileged port?).`);
  } else {
    log.fatal({ err }, "failed to start server");
  }
  process.exit(1);
});
```

## Location

- `packages/server/src/index.ts:245‑269` — `serve()` call with no `error` listener
- `packages/server/src/index.ts:148‑151` — global `uncaughtException` handler that currently absorbs it

Found during a pre-public-release readiness audit.

srid:
**Starting kolu on a port that's already taken now prints a one-line remedy and exits cleanly**, instead of a stack-shaped `"uncaught exception"` fatal that reads like an internal crash. Running a second instance on the same port (or leaving an old one alive) is one of the most common first-run friction points — for a public release it deserves a human answer, not a rethrow.

`serve()` from `@hono/node-server` returns the underlying `http.Server`, but it had **no `'error'` listener**. A failed listen emits `'error'`; with no handler Node rethrows, and the rethrow lands in the global `uncaughtException` handler. So the message you saw was generic and the exit looked like a fault.

### What you see now

| Failure | Before | After |
| --- | --- | --- |
| Port in use (`EADDRINUSE`) | `uncaught exception` + stack-shaped fatal | `Port 7681 is already in use — is kolu already running? Try --port <n>.` |
| Privileged port (`EACCES`) | same generic fatal | `Permission denied binding <host>:<port> (privileged port?).` |
| Anything else | `uncaught exception` | `failed to start server` with the structured `err` preserved |

All three paths end in a clean `exit(1)`.

### Shape

The error→message mapping and the `server.on("error", …)` wiring live in a small `serveError` module with `log`/`exit` injected, so the real `EADDRINUSE` path is exercised by a unit test that occupies a port and starts `serve()` on it — no subprocess, no mock of the failure. _Unknown errors still log structured `{ err }` so nothing diagnostic is lost._

Closes #1137

### DIFF
--- a/packages/server/src/index.ts
+++ b/packages/server/src/index.ts
@@ -26,6 +26,7 @@ import { ensureKoluRoot, shutdownCleanup } from "./koluRoot.ts";
 import { log } from "./log.ts";
 import { pwaIdentityForHostname } from "./pwaIdentity.ts";
 import { appRouter } from "./router.ts";
+import { attachServeErrorHandler } from "./serveError.ts";
 import { initSessionAutoSave } from "./session.ts";
 import { getTerminal } from "./terminal-registry.ts";
 import { snapshotSession } from "./terminals.ts";
@@ -267,6 +268,12 @@ const server = serve(
     startDiagnostics();
   },
 );
+
+// A failed listen (port in use, privileged port, unresolvable host) emits
+// 'error' on the returned http.Server. Without this listener Node rethrows,
+// and the global uncaughtException handler reports it like an internal crash
+// rather than the actionable "port already in use" message it deserves.
+attachServeErrorHandler(server, { host, port, log });
 
 // --- oRPC WebSocket handler (streaming) ---
 const wss = new WebSocketServer({ noServer: true });
--- a/packages/server/src/serveError.test.ts
+++ b/packages/server/src/serveError.test.ts
@@ -1,0 +1,97 @@
+import { createServer, type Server } from "node:http";
+import { serve } from "@hono/node-server";
+import { afterEach, describe, expect, it } from "vitest";
+import { attachServeErrorHandler, describeServeError } from "./serveError.ts";
+
+describe("describeServeError", () => {
+  it("explains a port-in-use failure with the port and a remedy", () => {
+    const msg = describeServeError(
+      { code: "EADDRINUSE" } as NodeJS.ErrnoException,
+      {
+        host: "127.0.0.1",
+        port: 7681,
+      },
+    );
+    expect(msg).toContain("7681");
+    expect(msg).toMatch(/already in use/i);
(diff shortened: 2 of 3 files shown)


srid:
## Codex ⇄ Claude debate

**✅ Consensus — 1 round.** codex reviewed at `xhigh` reasoning effort and approved with zero blocking/major findings.

| Round | codex approved? | Open blocking/major | Claude dispositions | Commit |
| 1 | ✅ yes | 0 | — (nothing to dispute) | — |

> codex's note: _"narrowly scoped: attaches an `error` listener to the Hono-returned server, maps common bind failures to friendly fatal messages, preserves raw structured logging for unrecognized errors, and adds focused coverage for `EADDRINUSE`. Ran `tsc --noEmit -p packages/server/tsconfig.json` — passed."_

PR #1141, opened 2026-06-02
Live-reload Code-tab previews of committed build-output files

srid:
**The Code-tab preview now live-reloads files under a committed build-output directory** — the Atlas docs project commits *and* previews `docs/atlas/dist/*.html`, but editing/rebuilding one of those pages never refreshed the iframe; it stayed frozen on whatever you first opened.

### Root cause: two ignore lists that disagreed

The Code-tab **tree** lists files with git's own ignore semantics:

```
git ls-files --cached --others --exclude-standard      # gitignore-aware
```

…but the working-tree **watcher** pruned with a *hardcoded* glob list (`**/dist/**`, `**/build/**`, …) that diverged from git. So a **committed** `dist/` file was **shown by the tree but hidden from the watcher** → its `@parcel/watcher` change events were never emitted → `install` ran but the callback never fired → the preview froze.

It looked intermittent because re-selecting a file does a fresh read; the reliable trigger is **clicking an in-iframe link** to another page (e.g. `index.html` → `second-brain.html`), which re-points the iframe `src` correctly but then relies on the now-silent watch — so the next edit is never seen.

### Fix: align the watcher's ignore set with git

Derive the watcher's ignore set from git instead of a hardcoded list — the exact **complement** of the tree's listing:

```
git ls-files --others --ignored --exclude-standard --directory
```

Now the watcher ignores precisely what git ignores and watches precisely what the tree shows:

| File | gitignored? | In tree? | Watched (before) | Watched (after) |
| Atlas `docs/atlas/dist/index.html` (committed) | no | ✅ | ❌ (hardcoded `**/dist/**`) | ✅ |
| a normal repo's `dist/` (gitignored) | yes | ❌ | ❌ | ❌ |
| `node_modules/` | yes | ❌ | ❌ | ❌ |

`.git` is added explicitly (git never lists its own dir; the git-dir watchers cover it), and a degraded fallback keeps `node_modules` pruned if git can't enumerate. _The set is a snapshot at install time, same as the static list it replaces._

This **deletes** `IGNORE_GLOBS`, `IGNORED_DIR_BASENAMES`, and the earlier re-root machinery (`watchRootFor` + `resolveUnder`/`assertRealpathUnder` guards): the watch root is always `repoRoot` and `filePath` only narrows a listener's event filter, so it can't steer the watch root or escape the repo.

> This PR was first cut as a single-file *re-root* fix, then re-cut to the git-aligned approach after review. The re-root version's hardening commits remain in history; the [codex-debate](https://github.com/juspay/kolu/pull/1141#issuecomment-4607522975) findings F1/F2/F3 were all artifacts of that approach and are **moot** here (no caller-influenced watch root).

### Tests

A `code-tab.feature` scenario captures the trigger: a nested `dist/index.html` linking to `dist/second.html`, navigate via the in-iframe link, edit `dist/second.html`, assert the preview reflects the new bytes — **red before, green after**. Full `code-tab.feature` 77/77; the git package's 86 unit tests pass; `just check`/`just fmt` clean.

### Try it locally

```sh
nix run github:juspay/kolu/code-tree-prview-reload
```

### DIFF
--- a/packages/integrations/git/src/browse.ts
+++ b/packages/integrations/git/src/browse.ts
@@ -13,6 +13,30 @@ import { resolveExistingUnder } from "./safe-path.ts";
 
 const execFileAsync = promisify(execFile);
 
+/** Single spawn/parse/error path shared by the two `git ls-files` listings.
+ *  Owns the maxBuffer ceiling, the newline split + empty-line filter, and the
+ *  GIT_FAILED error envelope; callers supply the args array and the message
+ *  prefix used on failure. */
+async function gitLsFiles(
+  repoPath: string,
+  args: string[],
+  failMsg: string,
+  log?: Logger,
+): Promise<GitResult<string[]>> {
+  try {
+    const { stdout } = await execFileAsync("git", args, {
+      cwd: repoPath,
+      maxBuffer: 64 * 1024 * 1024,
+    });
+    const paths = stdout.split("\n").filter((l) => l.length > 0);
+    return ok(paths);
+  } catch (e: unknown) {
+    const msg = e instanceof Error ? e.message : String(e);
+    log?.error({ err: e, repoPath }, "git ls-files failed");
+    return err({ code: "GIT_FAILED", message: `${failMsg}: ${msg}` });
+  }
+}
+
 /** Flat list of every repo-relative path (tracked + untracked-but-not-ignored).
  *  One-shot snapshot for Pierre's `@pierre/trees`, which builds the tree
  *  hierarchy itself from a flat path list.
@@ -23,19 +47,12 @@ export async function listAll(
   repoPath: string,
   log?: Logger,
 ): Promise<GitResult<string[]>> {
-  try {
-    const { stdout } = await execFileAsync(
-      "git",
-      ["ls-files", "--cached", "--others", "--exclude-standard"],
-      { cwd: repoPath, maxBuffer: 64 * 1024 * 1024 },
-    );
-    const paths = stdout.split("\n").filter((l) => l.length > 0);
-    return ok(paths);
-  } catch (e: unknown) {
-    const msg = e instanceof Error ? e.message : String(e);
-    log?.error({ err: e, repoPath }, "git ls-files failed");
-    return err({ code: "GIT_FAILED", message: `Failed to list files: ${msg}` });
-  }
+  return gitLsFiles(
+    repoPath,
+    ["ls-files", "--cached", "--others", "--exclude-standard"],
+    "Failed to list files",
+    log,
+  );
 }
 
 /** Repo-relative paths git *ignores* — the exact complement of `listAll`'s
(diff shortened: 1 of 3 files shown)


srid:
## Codex ⇄ Claude debate

**Consensus reached in 3 rounds.** codex reviewed at `xhigh` reasoning effort (read-only sandbox); the Claude author side fixed what it agreed with and the two converged.

| Round | codex verdict | Findings | Claude disposition | Commit |
| 1 | changes requested | **F1 (major)** — exported `watchWorkingTree` let a crafted `filePath` (`dist/../../../etc/passwd`) steer the re-root watch root outside the repo | fixed: `resolveUnder` lexical guard before deriving the watch root; no-op unsubscribe on escape | `6b4bb86c` |
| 2 | changes requested | **F2 (major)** — a *symlinked* ignored ancestor could still canonicalize the re-rooted watch root outside the repo (lexical check can't follow symlinks) | fixed: re-rooted installs now pass an async `assertRealpathUnder` realpath authority check before `parcelSubscribe` | `44bc2afd` |
| 3 | **approved** | **F3 (minor, open)** — guard-failure retirement calls `onLast()` before listener cleanup, so a deny-then-retry could delete a newer entry for the same watch root | acknowledged | — |

> **Note for reviewers:** following this debate, the fix is being **re-cut to a gitignore-aligned watcher** (per discussion on the approach): the watcher's ignore set will derive from `git ls-files --others --ignored --exclude-standard` — the exact complement of the browse tree's `--cached --others --exclude-standard` — instead of a hardcoded glob list, so *anything the tree shows is watched*. That removes the `watchRootFor` re-rooting entirely, which **moots F1, F2, and the open F3** (no caller-influenced watch root, no symlink-escape surface, no guard-failure path). The debate's hardening commits remain in history as the record of the re-root approach they secured.

_Posted by [`/be`](https://github.com/srid/agency) on Claude Code (model `claude-opus-4-8`)._

srid:
## Lowy ⇄ Hickey lens debate

**Consensus in 1 round.** lowy (volatility-based decomposition) and hickey (structural simplicity) reviewed the diff independently in parallel, then agreed per-finding. Independent finding counts: lowy 2, hickey 1.

| lowy | The git-listing volatility is now encapsulated correctly: `listIgnoredPaths` is the literal complement of `listAll`, both in the one git-listing module, and the watcher consumes the receptacle instead of re-deriving it via a hardcoded list — watcher and tree can't diverge. | `working-tree-watcher.ts`, `browse.ts` | **drop** (boundary is right) | — |
| lowy | `computeIgnore`'s block comment documents parcel's prune semantics rather than leading with what the function does — cosmetic, not a boundary defect. | `working-tree-watcher.ts` | **drop** (not worth churn) | — |
| hickey | `listAll` and `listIgnoredPaths` duplicated the `git ls-files` invocation envelope (maxBuffer, spawn shape, split/filter parse, `GIT_FAILED` error) — the same "keep two things aligned" coupling the PR set out to remove, one layer down. | `browse.ts` | **fix** | `cb833249` |

The hickey fix extracts a private `gitLsFiles(repoPath, args, failMsg, log)` that owns the spawn / maxBuffer / parse / error envelope; `listAll` and `listIgnoredPaths` are now thin call sites (the latter adding only its trailing-slash strip). One spawn/parse/error path, two call sites. Tests stay green (git unit 86/86, repro e2e green).

_Posted by [`/be`](https://github.com/srid/agency) on Claude Code (model `claude-opus-4-8`)._

srid:
## Code-police

rules → fact-check → elegance. Two findings fixed (each its own commit), three skipped as false positives / no-payoff.

| Finding | Disposition | Commit |
| --- | --- | --- |
| `gitLsFiles` catch logged the raw `unknown` (`{ err: e }`) while the coerced `msg` was already computed — inconsistent with every other catch in the package | **fixed** | `17177d2a` |
| `computeIgnore` silently dropped to the 2-entry fallback ignore set on git-enumeration failure — no operator signal it was running degraded | **fixed** (added `log?.warn`) | `e967b7af` |
| `filter((l) => l.length > 0)` → `filter(Boolean)` | skipped — TS narrowing keeps the explicit form (false positive) | — |
| inline `computeIgnore` / `ALWAYS_IGNORE_RELS` / `FALLBACK_IGNORE_RELS` | skipped — the named forms carry doc comments explaining the *why*; inlining loses that | — |

Tests stay green (git unit 86/86, repro e2e green; `just check`/`fmt` clean).

_Posted by [`/be`](https://github.com/srid/agency) on Claude Code (model `claude-opus-4-8`)._

srid:
## Evidence

Proves the live-reload fix for the in-iframe link navigation bug: preview `dist/index.html` → click the in-iframe link "go to second" → preview shows "second page ALPHA" → edit `dist/second.html` from the shell → **preview updates to "second page BETA"** (before the fix, the preview froze on ALPHA after the in-iframe navigation).



▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=live-reload-link-nav-1141.mp4

_Posted by [/be](https://github.com/srid/agency) on Claude Code (model claude-opus-4-8)._

PR #1088, opened 2026-06-01
Encapsulate mobile volatility behind receptacles

srid:
**Feature code across the client branched on the raw `isMobile()` / `isTouch()` media-query signals at ~18 sites** — picking layouts, gating commands, suppressing tips, forking panel reveal — and the viewport breakpoint was defined twice (JS `max-width: 639px` vs Tailwind `sm:` at `640px`), a 1px desync band. This PR applies a [Hickey/Lowy review](https://github.com/juspay/kolu/blob/brave-second/docs/plans/mobile-architecture-review.html) of mobile support: encapsulate the "mobile vs. desktop" volatility behind small stable interfaces so feature code expresses *intent* and *capability* instead of re-deriving form factor everywhere. `App.tsx`'s `match(isMobile())` layout fork and the `withKeyboardDismiss` wrapper stay as the two legitimate receptacles; everything else now routes through a verb.

### The four receptacles

| Leak it replaces | Receptacle |
| --- | --- |
| JS `639px` vs. CSS `640px` — two definitions of "mobile" | One source: `--breakpoint-sm` in the Tailwind `@theme`, read by `useMobile.ts` as the exact complement of `sm:` (`width < sm`) |
| `openInCodeTab` forked drawer-open vs. uncollapse on `isMobile` inline | `useRightPanel.reveal()` — producers express "show the panel"; the store owns the mobile/desktop fork |
| Tips, canvas commands, and the switcher each asked "am I mobile?" | `capabilities.ts` — `supportsSpatialCanvas` / `showsWorkspaceSwitcher` / `showsAmbientTips`; consumers ask about a capability, not pixels |
| `Terminal.tsx` poked xterm's `.xterm-screen` shadow DOM inline | `enableSoftKeyboardInput(term)` — the touch contenteditable setup in one testable place |

### Caught in review

The Hickey/Lowy pass ran with cross-validation, and it earned its keep: it rejected my first attempt to fold two `CodeTab` reads under one `isCompactLayout` verb, and surfaced a genuine **wrong-axis bug** — the file tree's *density* was keyed on viewport (`isMobile`) when "roomier rows" is a tap-target affordance (`isTouch`), so a coarse-pointer tablet wider than `sm` silently got tight rows. Now `isTouch`. *Its sibling touch-scroll driver correctly stays on `isMobile`* — that workaround targets the **portaled** mobile drawer, not touch presence (`pierreTouchScroll.ts`); a comment now records why so it isn't "fixed" the wrong way later.

[...]

PR #1006, opened 2026-05-28
Right panel: back to @corvu/resizable, drop the fast-drag stall

srid:
**Restore the right-panel resize handle to `@corvu/resizable` and pick up keyboard resize + spec-conformant pointer capture for free.** [#993](https://github.com/juspay/kolu/pull/993) had ripped out Corvu in favor of a hand-rolled pointer drag — which forgot `setPointerCapture(e.pointerId)`. Without capture, a fast rightward drag let the cursor outrun the panel's shrinking left edge and land over an xterm canvas / iframe whose subtree swallowed `pointermove`; resize stalled until the cursor crossed back into the handle's box. Corvu's handle calls `setPointerCapture` for us, so the spec-defined "events route to the capture target regardless of cursor position" property is back the moment the host swaps in.

The outer handle carries `startIntersection={false}` mirroring [`CodeTab.tsx`'s inner-handle defense](https://github.com/juspay/kolu/blob/master/packages/client/src/right-panel/CodeTab.tsx) — Corvu keeps a module-level `handles[]` and pairs handles of different orientations whose rects touch at the corner ([`@corvu/resizable/dist/index.js:201-222`](https://github.com/corvudev/corvu/blob/main/packages/resizable/src/lib/registerHandle.ts)). Both sides need the opt-out — the original reason #993 cited for moving off Corvu is solved this way, without sacrificing the library's capture handling.

### Effectively a selective revert of #993

| Removed (with #993) | Kept (from #993) |
| --- | --- |
| `@corvu/resizable` → custom pointer-drag _(the bug source)_ | `useViewPosture` API: `maximized: boolean` → `mode(): "tiled" \| "maximized"` |
| Floating-card tiled posture | `RAIL_WIDTH_PX` lifted into `ui/chromeSpacing.ts` |
| 44 px collapsed rail + pip stack | `RightPanelDrawer.tsx` as the dedicated mobile host file |
| `RightPanel`'s posture-aware shell + `shell: boolean` prop | `openInCodeTab` dispatching visibility imperatively from the producer call |
| `TerminalCanvas.rightPanel` slot prop | Tests' `data-collapsed` assertion |

[...]

PR #993, opened 2026-05-27
Mirror the dock on the right inspector panel + collapsed-rail redesign

srid:
**The right Inspector + Code panel now mirrors the dock's posture pairing across the canvas axis**, and the collapsed shell — previously a near-invisible 44 px strip — is now a real navigation surface with its own floating-card chrome and a tab pip stack. The asymmetry between the floating dock and the always-flush right panel pulled against the canvas-cockpit feel; this PR closes that gap on both axes.

### Posture mapping (expanded panel)

| Posture | Dock | Right panel |
| --- | --- | --- |
| Tiled | floating card top-left | floating card top-right |
| Maximized | flush left sidebar, `border-r` | flush right sidebar, `border-l` |

Both surfaces anchor at `top-12` — `top-20` left a 32 px dead strip below the chrome bar; `top-12` puts them on a single horizontal axis just under the bar. The chrome bar itself drops the dynamic `right: panelSize() * 100vw` offset since the panel now sits *below* the bar rather than capping it.

### Collapsed rail (mirrors dock rail)

The original collapsed state was 0 px wide — no re-expand affordance, no preview. Now the collapsed shell renders as a 44 px rail (matching the dock's `RAIL_WIDTH_PX`) with **three pips** stacked vertically:

- **Expand chevron** — restores the last-active tab.
- **Inspector pip** — `information-circle` icon; expands AND switches to Inspector.
- **Code pip** — `<>` brackets icon; expands AND switches to Code.

The pip matching `rightPanel.activeTab().kind` wears an accent fill (`bg-accent/15` + accent ring) so a glance at the rail tells you which tab the panel will land on. In tiled posture the rail body becomes a real card — `rounded-l-2xl`, `shadow-2xl shadow-black/40`, `bg-surface-2` (one step brighter than the dock's `bg-surface-1`) — pushed down by `mt-12`/`mb-4` to align with the dock card. In maximized posture it stays as a flush column inside the `border-l` boundary (already contrasts naturally against the bright maximized terminal).

### Architecture cleanups riding along

- **`useViewPosture()`** promoted from `maximized: boolean` to `mode(): \"tiled\" | \"maximized\"` (#628 finally), with 14 callsites updated.
- **`RAIL_WIDTH_PX = 44`** extracted to `ui/chromeSpacing.ts` — the dock and right-panel rails were each declaring it independently with cross-reference comments.
[...]

PR #1142, opened 2026-06-02
Un-skip the file-ref re-click regression guard (#1138)

### LINKED ISSUE #1138: Clickable file-ref re-click fails to re-open the panel in the bundled (production) build
## Summary

Clickable file references (`path/to/file.ts:42` in terminal output → opens the file in the right panel's Code tab) — a marquee README feature — have a regression that **only manifests in the bundled/production build, not in dev**: after clicking a file-ref, closing the right panel, and clicking the *same* file-ref again, the panel fails to re-open / re-select the line.

The scenario is currently quarantined as `@skip` in the e2e suite:

- `packages/tests/features/file-ref-link.feature` — scenario *"Re-clicking the same file-ref after closing the panel re-selects the line"* (`@skip`)
- Regression introduced/identified at commit `c89a85f3`.

## Why it matters for the release

It passes in `just dev` but fails in the Vite production bundle — i.e. it's broken in exactly the build public users run via `nix run github:juspay/kolu`. A headline, prominently-documented feature silently misbehaving on second use is a poor first impression.

## Suspected cause

Likely a SolidJS reactive-update elision under the production bundle: re-clicking the same ref sets the panel target to a value that is `===` to the last (now-closed) selection, so the reactive update is skipped and the panel never re-opens. The closed-panel state isn't reflected back into the signal that drives the open, so an identical re-selection is a no-op. Dev mode's different update/minification behavior masks it.

Worth checking:
- The signal/store that holds the active file-ref target in the right-panel open path (`packages/client/src/right-panel/*`, the `path:line` xterm link handler).
- Whether closing the panel resets that target signal (so a repeat selection is genuinely a change), or whether equality short-circuits the re-open.
- Whether `createStore`/`reconcile` or an `equals` comparator is eliding the repeat write.

## Repro (bundled build)

1. Run a production build of kolu.
2. In a terminal, emit output containing a `path/to/file.ts:42` reference; click it → panel opens at the file/line. ✓
3. Close the right panel.
4. Click the **same** ref again → panel does **not** re-open. ✗ (works in dev)

## Acceptance

- Re-clicking the same file-ref after closing the panel re-opens it at the correct line, in the bundled build.
- Un-skip `packages/tests/features/file-ref-link.feature`'s re-click scenario and confirm it passes against the bundled build in CI.

Found during a pre-public-release readiness audit.

srid:
**The clickable file-ref "re-click after collapse" scenario is now an active CI guard against the bundled build.** It was quarantined as `@skip` at `c89a85f3`, where a second click on the same `path:line` after manually collapsing the right panel failed to re-open it — but *only* in the production Vite bundle, never in `just dev`. A headline, prominently-documented feature silently misbehaving on second use is exactly the kind of thing a pre-release audit should not let ship un-guarded.

### What was actually wrong

At `c89a85f3`, `openInCodeTab` only wrote the per-terminal tab/mode and the `pendingOpen` request; **panel visibility was driven by a deferred `createEffect(on(pendingOpen, …))` subscriber.** Under the minified production Solid build that effect lost its *second* fire when the same request value flowed through twice with a manual collapse in between, so the panel stayed `aria-hidden` and never re-opened. Dev mode's looser update behavior masked it.

The bug was fixed by the subsequent right-panel rework — *not* by a new commit in this PR:

| Commit | What it changed |
| --- | --- |
| #993 | Moved panel reveal to **imperative dispatch** inside `openInCodeTab` (`expandPanel` / `setDrawerOpen`), off the deferred-effect path |
| #1006 | Returned the desktop host to `@corvu/resizable`, whose `flex-basis` is driven directly by the `collapsed` preference |
| #1088 | Encapsulated the mobile/desktop reveal fork behind `useRightPanel.reveal()` |

`openInCodeTab.ts`'s own comments already document why imperative dispatch *"sidesteps that elision path entirely."* **The fix shipped weeks ago; the `@skip` was simply never lifted.**

### Why this is safe to un-skip

Verified empirically against the production bundle that `just test-quick` builds:

```
c89a85f3  ✖  panel stays aria-hidden on re-click → "right panel should be visible" times out
HEAD      ✔  6/6 consecutive runs green
```

The RED reproduction at the historical commit confirms the scenario is a *genuine* regression guard — it fails exactly when the reveal path breaks — not a vacuous pass. The diff is the un-skip plus a comment recording what the scenario protects and why it must run against the bundled build, not just dev.

Closes #1138

### DIFF
--- a/packages/tests/features/file-ref-link.feature
+++ b/packages/tests/features/file-ref-link.feature
@@ -159,12 +159,10 @@ Feature: File-ref autolinking in terminal
 
   # Guards the c89a85f3 regression: a second click on the same `path:line`
   # after manually collapsing the panel must re-open it. The bug was
-  # production-only — the deferred `on(pendingOpen)` effect that drove panel
-  # visibility lost its second fire under the minified Solid build when the
-  # same request flowed through twice, so the panel stayed `aria-hidden`.
-  # `openInCodeTab` now reveals the panel imperatively, sidestepping the
-  # elided-effect path. This scenario is the canary, so it must run against
-  # the bundled build (`just test-quick`), not just dev.
+  # production-only (passes in dev) — see right-panel/openInCodeTab.ts for
+  # the deferred-effect-elision mechanism and the imperative-reveal fix.
+  # This scenario is the canary for that fix, so it must run against the
+  # bundled build (`just test-quick`), not just dev.
   Scenario: Re-clicking the same file-ref after closing the panel re-selects the line
     When I run "git init /tmp/kolu-file-ref-861-reclick && cd /tmp/kolu-file-ref-861-reclick"
     And I run "git commit --allow-empty -m init"
--- a/packages/tests/features/file-ref-link.feature
+++ b/packages/tests/features/file-ref-link.feature
@@ -157,15 +157,14 @@ Feature: File-ref autolinking in terminal
     And the file preview iframe should be visible
     And the file preview iframe should contain "electricity"
 
-  # `@skip`: known regression noted in c89a85f3 — the second xterm `path:line`
-  # click after a manual collapse fails to re-open the panel under the bundled
-  # build (passes in dev). Suspected production-Solid reactive elision or
-  # xterm link-decoration cache invalidation after the layout reflow.
-  # `equals: false` on `pendingOpen` and imperative dispatch from
-  # `openInCodeTab` both fail to clear it; deeper diagnosis is tracked
-  # separately. Run with `CUCUMBER_TAGS='@skip' just test-quick
-  # features/file-ref-link.feature` to exercise this scenario locally.
-  @skip
+  # Guards the c89a85f3 regression: a second click on the same `path:line`
+  # after manually collapsing the panel must re-open it. The bug was
+  # production-only — the deferred `on(pendingOpen)` effect that drove panel
+  # visibility lost its second fire under the minified Solid build when the
+  # same request flowed through twice, so the panel stayed `aria-hidden`.
+  # `openInCodeTab` now reveals the panel imperatively, sidestepping the
+  # elided-effect path. This scenario is the canary, so it must run against
+  # the bundled build (`just test-quick`), not just dev.
   Scenario: Re-clicking the same file-ref after closing the panel re-selects the line
     When I run "git init /tmp/kolu-file-ref-861-reclick && cd /tmp/kolu-file-ref-861-reclick"
     And I run "git commit --allow-empty -m init"


srid:
## Codex ⇄ Claude debate

**✅ Consensus — 1 round.** Codex reviewed at `xhigh` reasoning effort.

| 1 | Approved | 0 | — (nothing to rebut) | — (no changes needed) |

Codex read the full changed feature context plus the `file-ref` step definitions and the `openInCodeTab` / right-panel reveal path, and confirmed the scenario assertions align with the intended production-regression guard. No blocking or major issues; no code changes required.

> Read-only review — codex was invoked with `--sandbox read-only` and did not run tests.

srid:
## Lowy ⇄ Hickey lens debate

**✅ Consensus — 1 round.** Both lenses ran independently on Opus; the rationale (deliberate un-skip, no production code change) was supplied so intentional choices weren't flagged.

Independent finding counts: **lowy 0**, **hickey 1**.

| hickey | Bug-mechanism narrative duplicated across the new feature comment and `openInCodeTab.ts` | `file-ref-link.feature:160-167` | **fix** | `31978786` |

**The finding, and why it's apt:** my added comment re-narrated the *exact* deferred-effect-elision mechanism already canonically documented in `openInCodeTab.ts:8-19` — one fact maintained in two prose locations, coupled only by an unenforced "keep in sync" rule. As hickey noted, this is self-demonstrating: *the stale comment this PR removes rotted precisely because the diagnosis had migrated into `openInCodeTab.ts` while the feature comment was never updated.* Re-introducing a full mechanism paragraph would recreate the identical two-place drift.

lowy conceded and refined the plan: point to the canonical owner **path-only** (`right-panel/openInCodeTab.ts`, no line numbers, so the reference can't rot) and keep only the genuinely test-specific invariant in the feature file — *this scenario is the canary; it must run against the bundled build, not just dev.* Applied as committed.

srid:
## Code Police

**✅ All clear — 0 findings.** Rules → fact-check → elegance, all clean.

| Pass | Issues |
| --- | --- |
| Rule checklist (24 rules) | 0 |
| Fact-check | 0 |
| Elegance | 0 |

The `.feature` change was checked against the `e2e-poll-async-state` and `comment-the-non-obvious` rules among others: it only lifts `@skip` and rewrites the comment — no new step definitions, no `page.evaluate` + synchronous assert, and the retained comment is the single-sourced canary note the lens debate landed on. No `fix(police)` commits were needed.

srid:
## Evidence

Re-clicking the same `path:line` file-ref after collapsing the right panel now correctly re-opens the panel at the target line in the bundled (production) build — the exact build where reactive elision previously swallowed the second click.



▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=evidence-file-ref-reclick.mp4

Recorded with `KOLU_EVIDENCE=1 just test-quick features/file-ref-link.feature --name 'Re-clicking the same file-ref after closing the panel re-selects the line'` on an ephemeral `pu` box against the `click-file-ref` branch (bundled build, slowMo on). Scenario: 1 passed, 13 steps passed.

PR #1143, opened 2026-06-02
Contain per-terminal provider failures behind one error boundary

srid:
A single background metadata task that threw could escape as an unhandled rejection and take down the **whole server** — and because pty-host runs in-process today, that kills *every* terminal at once. The default `nix run` path has no supervisor, so the app just dies. **This contains those failures at their source and consolidates the per-terminal "never throw" invariant into one place**, so one bad git/PR/agent read degrades that terminal's metadata instead of crashing the process.

This was reviewed through the `hickey` (structural simplicity) and `lowy` (volatility/boundaries) lenses first. Both independently rejected the tempting-but-wrong move — flipping the global `unhandledRejection` handler to log-and-continue — because a floating promise is as corrupting as a sync throw, and a context-free global handler is the wrong layer to make a recover-or-die decision. The fix belongs at the per-task boundary.

### What changed

| Site | Problem | Fix |
| --- | --- | --- |
| `github/resolve.ts` | `void fetchAndEmit(...)` floated a promise whose `emit → onChange` runs the caller's metadata write; a throw there became an unhandled rejection | Contain it inside `fetchAndEmit` — the single point both the branch-change and 30s-poll paths funnel through |
| `terminalBackend/providers.ts` | `reconcile()` was called bare from four channel `onEvent` callbacks; a throw there silently breaks `buildConsume`'s `for await` loop, **freezing that subscription for the terminal's life** | Make `reconcile()` never throw via one try/catch; removed the two now-redundant outer guards so the invariant lives in *one* place, not a three-layer chain |
| `index.ts` | `unhandledRejection` fatal message was generic | Kept fatal *(supervisor restarts clean)*, but named the cause so a firing reads as "a background task is missing its error boundary" |

> _The lint half of this is already enforced — `noFloatingPromises: "error"` is on in `biome.jsonc`. That's why the audit surfaced exactly one real uncontained leak (`fetchAndEmit`); the rest of the work is consolidating an invariant, not wrapping 40 sites._

### Test

A covering unit test drives a throwing `onChange` through the watcher and asserts it is **logged, not propagated** (no `unhandledRejection` fires). Without the `fetchAndEmit` guard, the floated promise rejects and the process exits via the global handler.

### Try it locally

```sh
nix run github:juspay/kolu/release-bugs
```

### DIFF
--- a/.agency/code-police.md
+++ b/.agency/code-police.md
@@ -215,6 +215,32 @@ function useEvent(..., options: { onError: (err) => void; ... }): void {
 
 _Rationale_: a hook returning `Subscription<T>` with `.error()` lets consumers read the error reactively and render it — optional `onError` is fine there. Void return with no error surface in the result type is a category mismatch; the type system has to require the handler or the failure is invisible by construction. Codified after `useEvent.onError` and `pollOnEvent.onReadError` were tightened to required in `@kolu/surface`.
 
+### callback-fanout-guarded-at-funnel
+
+A watcher/subscription that invokes a caller-supplied callback (`onChange`, `onEvent`, an `emit` helper) from **more than one emission path** must place the try/catch at the single shared funnel the callback passes through — never on a subset of the call sites. A throwing consumer that escapes is not a benign log line: floated through a `void fetchAndEmit()` it surfaces as an **unhandled rejection** (fatal — the global handler in `index.ts` calls `process.exit(1)`), and from a synchronous `channel.consume({ onEvent })` callback it **breaks out of `buildConsume`'s `for await` loop** (`@kolu/surface`), silently freezing that subscription for the rest of the terminal's life.
+
+Bad — boundary on the async path only, leaving the synchronous pending emit uncontained:
+```ts
+async function fetchAndEmit(root) {
+  try { emit(await resolveGitHubPr(root)); } catch (err) { log?.error(…); }
+}
+function setGit(...) {
+  emit({ kind: "pending" }); // ← still runs onChange synchronously, unguarded
+  void fetchAndEmit(root);
+}
+```
+
+Good — boundary inside `emit`, the one point every path funnels through:
+```ts
+function emit(pr) {
+  if (stopped || prResultEqual(pr, lastPr)) return;
+  lastPr = pr;
+  try { onChange(pr); } catch (err) { log?.error({ err }, "…: emit failed"); }
+}
+```
+
+_Rationale_: the dangerous escape is the *uncovered* path, and watchers routinely emit from several (a synchronous "pending" on change + an async resolved value + a poll tick). Guarding one path reads as "handled" in review while another stays a live throw vector. Putting the boundary at the shared invocation point makes "the consumer callback cannot throw out of this watcher" a single-site invariant instead of a per-call-site discipline. Complements `silent-handler-required-on-void-subscriptions` (which requires the *primitive's* `onError` at the type level); this rule is about the *watcher implementation* containing the consumer it fans out to. Codified after a `subscribeGitHubPr` fix guarded `fetchAndEmit` but missed `setGit`'s synchronous `emit({ pending })` ([kolu#1143](https://github.com/juspay/kolu/pull/1143)).
+
 ### migration-shape-guard
 
 A `Conf` (or analogous schema) migration that acts on a specific value shape must early-return when the on-disk shape doesn't match its preconditions. Never write transient orphan fields that subsequent migrations are expected to destructure-out.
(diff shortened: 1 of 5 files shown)


srid:
## Codex ⇄ Claude debate

**✅ Consensus** reached in **2 rounds**. Codex reviewed at **`xhigh`** reasoning effort.

| Round | Codex verdict | Claude disposition | Commit |
| --- | --- | --- | --- |
| 1 | ❌ 1 major open — `F1`: the error boundary in `fetchAndEmit` left `setGit`'s **synchronous** `emit({pending})` path uncontained. After the watcher leaves its initial pending state, a later branch change re-emits pending through `onChange` synchronously; a throwing metadata write would escape out of `setGit` into the server's `channels.git.consume` and freeze the subscription. | **fixed** — moved the boundary into `emit()` itself, the single point both the synchronous pending emit and the resolved emit funnel through. Added a regression test driving a non-pending state → throwing `onChange` → branch change, asserting no throw / no unhandled rejection (verified red against the pre-fix code). | `91a41b4a` |
| 2 | ✅ approved — `F1` resolved; no remaining blocking/major findings | — | — |

> The original fix contained the *async* leak (`fetchAndEmit`); codex's catch was that the *synchronous* pending-emit path through the same consumer was still a live throw vector. The boundary now lives on the shared `emit`, which is the true single funnel point.

_Posted by [`/codex-debate`](https://github.com/srid/agency) — commits sit on the branch for human review; nothing pushed or merged by the debate._

srid:
## Lowy ⇄ Hickey lens debate

**✅ Consensus** in **2 rounds** — independent review surfaced **lowy: 2, hickey: 1**; all **3 findings settled to `drop`** (zero code changes). The structural pass *validated* the design rather than altering it.

| Finding | Origin | Disposition | Why dropped |
| --- | --- | --- | --- |
| `lowy-1` — "containment is asymmetric across the 5 sibling providers" | lowy | **drop** | lowy's own rebuttal (verified at 4 points in source) showed the premise is false: `inMemoryChannel.publish` only enqueues/resolves waiters — it never synchronously invokes a consumer's `onEvent`. The metadata fanout runs **detached** in `buildConsume`'s for-await loop, so a throw there can only freeze the *downstream* subscriber, never propagate back into a sibling provider's `onEvent`. The 3 "exposed" providers plug into a fanout that **cannot deliver a synchronous throw** — no guard needed. The two guards added (`emit`, `reconcile`) sit at the DAG's *actual* throw points (a floated `void fetchAndEmit` → fatal unhandledRejection; `resolveSession`/`createWatcher`/`destroy` in `reconcileInner`). |
| `lowy-2` — "index.ts comment over-claims the DAG as the per-task-boundary exemplar" | lowy | **drop** | Comment-precision nit; the placement decision (recover-or-die at per-task receptacles, global net stays fatal) is sound and deliberate. Not worth a churn in this PR. |
| `hickey-1` — "reconcile's never-throw depends on an unenforced convention at the fan-out" | hickey | **drop** | The convention is now centralized in one `try/catch` and the bare call sites are documented; enforcing it at the type level is out of scope for this fix. |

> Takeaway: the adversarial structural review **confirmed** the author rationale — the existing receptacles genuinely contain the non-guarded paths, and the two new boundaries are at the right sockets. No fixes applied.

_Posted by [`/lens-debate`](https://github.com/srid/agency) — lowy (volatility) ⇄ hickey (simplicity), both on Opus, independent-then-debate._

srid:
## Code-police

Rules → fact-check → elegance passes. **No correctness findings**; the elegance pass applied two cleanups (commit `7d3f44d8`).

**Applied**
- **`resolve.test.ts`** — extracted the repeated setup/teardown (`KOLU_GH_BIN` pin, `unhandledRejection` spy, env restore) into `beforeEach`/`afterEach`, and the `setTimeout(50)` settle pattern (×3) into a `settle()` helper. *Follow-up:* typed the spy precisely as `Mock<(reason, promise) => void>` — a bare `ReturnType<typeof vi.fn>` widens to the `Procedure | Constructable` union and fails `process.on`'s listener overload (caught by `tsc`, not vitest).
- **`resolve.ts`** — tightened the `emit()` comment to name the git channel's consume loop as the synchronous escape path.

**Skipped (deliberately)**
- The `reconcile`/`reconcileInner` split — correct altitude; the named inner function makes "reconcile cannot throw" a searchable invariant, not just a comment.
- `[...activation.reconcilers]` spread — pre-existing, not introduced by this diff.

_Posted by [`/code-police`](https://github.com/srid/agency) — rules from `.agency/code-police.md` (including the new `callback-fanout-guarded-at-funnel`, which this diff satisfies)._

Issue #977, opened 2026-05-26
Mobile dock drawer crashes on open via mouse-click events (Corvu @0.2.4)


srid:
Discovered while reviewing PR #974 — the mobile dock drawer (`<Drawer side="left">` in `packages/client/src/MobileTileView.tsx:205`) crashes when opened via a mouse-click event on `[data-testid="mobile-dock-handle"]`. e2e tests (`features/mobile-dock-drawer.feature`) pass because Playwright's `.tap()` synthesises touch events, which exercise a different open path inside Corvu.

## Repro

1. Run `just dev`
2. Open `http://localhost:7681/` in Chromium
3. DevTools → emulate viewport 390 × 844 with touch enabled
4. Click `[data-testid="mobile-dock-handle"]` (the left-edge handle)

## Observed

`dockOpen` flips to `true`, then two uncaught errors from `@corvu/drawer@0.2.4`:

- `TypeError: Cannot read properties of undefined (reading 'reduce')` — `findNearbySnapPoint` at `node_modules/.pnpm/@corvu+drawer@0.2.4_solid-js@1.9.11/node_modules/@corvu/drawer/dist/index.js:69`
- `TypeError: Cannot read properties of undefined (reading 'transitionDuration')` — open-animation prep at the same file, lines 310 / 484 / 495

`[data-testid="mobile-dock-sheet"]` never attaches; the drawer never visibly opens.

## Why CI doesn't catch this

Playwright's touch-driven `.tap()` takes a different code path in Corvu than the mouse-driven `click()`. The e2e suite under `features/mobile-dock-drawer.feature` exclusively uses `.tap()` so the crash window stays invisible there.

## Likely cause

`snapPoints` defaults to `[0, 1]` in Corvu's docs, but the crash trace shows `findNearbySnapPoint` reducing over `undefined`. Either Corvu's default isn't being applied on the click path, or there's a reactive-ordering issue where the signal is read before defaults attach.

## Likely fix

Pass `snapPoints={[0, 1]}` explicitly to `<Drawer side="left">` in `MobileTileView.tsx`. Default-but-explicit values; should sidestep whichever code path leaves the signal undefined.

## Repro confirmed on

- Branch: `feat/dock-bare` @ `800d9e36` (this PR)
- `git show origin/master:packages/client/src/MobileTileView.tsx` — `<Drawer>` wiring is identical on master, so this is **not** a regression introduced by PR #974; the crash exists on master too.

## Workaround

[...]

PR #1144, opened 2026-06-02
Bring up the mobile right-panel drawer when a terminal link is tapped

srid:
**Tapping a terminal file-ref link on mobile now actually brings up the code-browser drawer.** Before this, tapping a link to a `.html` (or any) file in the terminal appeared to do *nothing* — the tree selection updated under the hood, but the bottom sheet never came on screen.

The mobile right panel hosts as a Corvu `Drawer side="bottom"`, and **Corvu drives its open/close state machine off the *computed* `transition-duration` of the content element.** That `Drawer.Content` carried no CSS transition, so Corvu read `0s` and mishandled a *programmatic* open — the terminal tap flips `drawerOpen` via `openInCodeTab`, it doesn't drag the sheet. The drawer never settled open. The fix gives the content a real `transition-transform duration-200` so Corvu has a non-zero duration to settle against.

> _It read as `.html`-specific only because an `.html` ref renders the iframe **preview**; the underlying failure was the drawer itself never settling, which affected every code link on mobile. Notably **not** the cause: the `#977` `snapPoints` workaround the sibling drawers carry — adding it here didn't fix this; the missing transition did._

### A masked test-harness gap

The e2e env forced `transition-duration: 0s` on **everything**, including the drawer content — reproducing the exact mis-settle in tests. That's why prior mobile drawer scenarios could only ever assert a *transient* open latch and never rendered content, so the bug was invisible to CI. This PR exempts `[data-corvu-drawer-content]` from the override so the new tests exercise — and guard — the real production transition.

### Coverage

Two new `@mobile` scenarios in `file-ref-link.feature`:

- Tapping a `.html` file-ref brings up the code-browser preview iframe.
- Re-tapping after dismissing the drawer re-opens it.

Verified by negative control: reverting the `Drawer.Content` transition makes the iframe scenario fail; restoring it passes. Full mobile suite **39/39**, full `file-ref-link.feature` **14/14**.

### Try it locally

```sh
nix run github:juspay/kolu/mobile-click
```

_Generated by [`/forge-pr`](https://github.com/srid/agency) on Claude Code (model `claude-opus-4-8`)._

### DIFF
--- a/packages/client/src/right-panel/RightPanelDrawer.tsx
+++ b/packages/client/src/right-panel/RightPanelDrawer.tsx
@@ -63,7 +63,16 @@ const RightPanelDrawer: Component<HostProps> = (props) => {
             data-testid="right-panel-drawer-backdrop"
             class="fixed inset-0 z-40 bg-black/40 opacity-0 transition-opacity duration-200 data-open:opacity-100"
           />
-          <Drawer.Content class="fixed bottom-0 left-0 right-0 z-50 bg-surface-0 border-t border-edge shadow-xl h-[85vh] flex flex-col rounded-t-lg overflow-hidden">
+          {/* `transition-transform` is load-bearing, not cosmetic: Corvu's
+           *  drawer drives its open/close state machine off the *computed*
+           *  `transition-duration` of this content element. With none set it
+           *  reads 0s and mishandles a programmatic open (the terminal file-ref
+           *  tap flips `drawerOpen` via `openInCodeTab` rather than dragging the
+           *  sheet) — the drawer never settles open, so on mobile a tapped
+           *  `.html`/code link "does nothing": the tree selection updates but
+           *  the sheet is never brought up. A real transform transition gives
+           *  Corvu a non-zero duration to settle against. */}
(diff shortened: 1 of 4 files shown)


srid:
## Evidence

**Bug fixed:** on a mobile viewport, tapping a terminal file-ref link to a `.html` file did nothing — the right-panel bottom-drawer (code browser) never came up. The Corvu `Drawer.Content` in `RightPanelDrawer.tsx` now carries a real `transition-transform duration-200`, so Corvu's open state-machine settles the drawer on a programmatic open.

Captured by running the `@mobile` e2e scenario _"Tapping a .html file-ref on touch brings up the code-browser preview"_ (`packages/tests/features/file-ref-link.feature:39`) under the project's Cucumber + Playwright harness with `KOLU_EVIDENCE=1` on a clean pu box at this PR's commit (`626f434`). The `@mobile` tag drives a 390×844 touch viewport; under `KOLU_EVIDENCE` the harness keeps motion on so the drawer animates naturally.

Tapping the `page.html` ref raises the bottom-sheet code browser, which mounts the sandboxed iframe and renders the file preview (the `<h1>preview me</h1>`):



Still at the payoff (drawer up, Code tab active, `page.html` selected, iframe preview rendered):



▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=evidence-mobile-html-fileref-drawer.mp4

PR #1147, opened 2026-06-03
fix(canvas): drop redundant aria-hidden on covered tiles

srid:
**Covered canvas tiles no longer emit a WAI-ARIA console warning on every maximize.** A covered tile carried both `inert` and `aria-hidden="true"`, but the xterm helper textarea inside it can retain DOM focus the instant the tile is covered — and the browser *blocks* `aria-hidden` on an ancestor of a focused element, logging:

> Blocked aria-hidden on an element because its descendant retained focus… Consider using the `inert` attribute instead, which will also prevent focus.

`inert` already removes the subtree from the accessibility tree *and* prevents focus, so `aria-hidden` was pure redundancy — and the sole source of the warning. Dropping it leaves `inert` doing the hiding alone, exactly as the WAI-ARIA spec recommends.

### Why this is safe

| Concern | Outcome |
| --- | --- |
| Hidden from assistive tech | Still hidden — `inert` removes the subtree from the a11y tree |
| Intrinsic visual hide | Unchanged — covered tiles still set `visibility: hidden` |
| e2e coverage | `canvas.feature` asserts `visibility: hidden`, not `aria-hidden` — untouched |

### DIFF
--- a/packages/client/src/canvas/CanvasTile.tsx
+++ b/packages/client/src/canvas/CanvasTile.tsx
@@ -165,8 +165,15 @@ const CanvasTile: Component<{
       // the pre-#988 `visibility: hidden` wrapper without re-introducing
       // it. xterm.js writes still land in the buffer (no render dependency
       // on inert), so the dock's buffer previews stay populated.
+      //
+      // Deliberately NOT pairing this with `aria-hidden="true"`: `inert`
+      // already drops the subtree from the accessibility tree, so the
+      // attribute is redundant — and the browser blocks `aria-hidden` on an
+      // ancestor of a focused element (the xterm helper textarea can retain
+      // DOM focus the instant a tile is covered), logging a WAI-ARIA console
+      // warning. `inert` is the spec's recommended replacement precisely
+      // because it hides *and* prevents focus without that conflict.
       inert={isCovered()}
-      aria-hidden={isCovered() ? "true" : undefined}
       class="flex flex-col overflow-hidden border transition-shadow duration-200"
       classList={{
         // Maximized uses `absolute inset-0 z-40` to cover the canvas
--- a/packages/client/src/right-panel/RightPanel.tsx
+++ b/packages/client/src/right-panel/RightPanel.tsx
@@ -7,9 +7,10 @@
  *  for the horizontal split + drag-to-resize; the mobile host wraps
  *  this in a `@corvu/drawer` (`RightPanelDrawer.tsx`). Both hosts thread
  *  the same `visible` accessor — desktop reads `collapsed()`, mobile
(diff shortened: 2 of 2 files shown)


srid:
## Codex ⇄ Claude debate

**✅ Consensus — 1 round.** codex reviewed at `xhigh` reasoning effort and approved with no blocking or major findings; no author rebuttal was needed and no files changed.

| 1 | ✅ approved | 0 | — (nothing to rebut) | — |

> codex: *"Removing `aria-hidden` while keeping `inert` preserves the covered-tile focus/accessibility behavior and directly avoids the focused-descendant ARIA warning; existing hidden-tile coverage asserts `visibility: hidden`, not the removed attribute."*

srid:
## Lowy ⇄ Hickey lens debate

**✅ Consensus — 1 round.** Independent parallel review: **lowy 0** findings, **hickey 1**. Both lenses ran on Opus. The one finding settled to `fix` with agreed plan and was applied + committed.

| hickey | Redundant `inert`+`aria-hidden` pairing survives in RightPanel — the *same* WAI-ARIA warning bug this PR fixes, on a sibling surface | `right-panel/RightPanel.tsx:69-70` | ✅ fix | `0744c1f8` |

The lenses caught that the original diff fixed **one of two** copies of the same concept. `RightPanel` paired `aria-hidden={!props.visible}` with `inert={!props.visible}` on the panel root — byte-for-byte the pattern CanvasTile just dropped. When the panel collapses (`sizes=[1,0]`) while a focusable control inside it (Collapse button, tab buttons, or a CodeTab input) holds focus, the browser blocks the ancestor `aria-hidden` and logs the identical warning.

The fix removes the redundant `aria-hidden`, relies on `inert` alone, and rewrites the block comment + file-header note to match the CanvasTile rationale. _Both lenses verified the inner per-tab `aria-hidden={!isActive()}` (line 127) is a separate `display:none`-body concern, not an ancestor-of-focus conflict — left unchanged._

srid:
## Code-police

**✅ All clear — no findings, no commits.**

| Pass | Issues | Notes |
| --- | --- | --- |
| Rules (25 checked) | 0 | No violations. The cross-file `inert`-without-`aria-hidden` pattern is 2 instances — below the rule-of-three extraction threshold; the duplicated rationale is prose comment, not synced logic. |
| Fact-check | 0 | Claims are spec-correct: `inert` removes a subtree from the accessibility tree *and* focus order, and the browser blocks `aria-hidden` on an ancestor of a focused element (WAI-ARIA). |
| Elegance | 0 | Two-line attribute removal + necessary explanatory comments; `inert={…}` is already the idiomatic one-liner. |

> Verified the inner `aria-hidden={!isActive()}` on the inactive (`display:none`) tab panels is a separate, correct concern — unrelated to this diff and left untouched.

PR #1149, opened 2026-06-03
Fix stale client after deploy — remove the service worker

srid:
After a deploy, an open kolu tab kept running the **old** client bundle — the chrome bar flagged `CLIENT ≠ srv`, a *normal* reload (cmd+R) replayed the stale build, and only a *hard* reload (cmd+Shift+R) picked up the new one. The root cause was a **service worker** serving its stale precache in front of the network. **This PR removes the service worker entirely** and makes client freshness structural at the HTTP layer, so the deployed build always wins.

### Why a service worker, on plain HTTP?

kolu is served over plain HTTP on a Tailscale host (`http://pureintent:7692`) — normally not a secure context, so no SW. But Chrome's `unsafely-treat-insecure-origin-as-secure` flag (set for these origins) makes them secure contexts, so a worker *did* register and precache an old build. An earlier iteration of this PR gated the SW on `location.protocol === "https:"` instead of `window.isSecureContext`, which **orphaned** that worker — the new client stopped managing it. Disabling the flag made the bug vanish with no unregister, confirming the SW was the cause. An always-connected terminal app gains nothing from a SW (no offline; content-hashed assets are already `immutable`-cached), so the fix is to remove it, not keep fighting its update race. Full saga: [`docs/cache-bug.md`](docs/cache-bug.md).

### Changes

| Area | Change |
|---|---|
| **Service worker** | **Removed** — no `registerSW` / VitePWA plugin. `retireServiceWorker()` runs on every load (unregister + delete caches), and a self-destructing `public/sw.js` retires any worker an earlier build left registered |
| **SPA shell** | `Cache-Control: no-store` (was `no-cache`) — a normal reload can't replay a cached shell |
| **Missing `/assets/*`** | `404` (was the HTML shell with a 1-year `immutable` directive — wrong MIME + cache poisoning) |
| **Version skew** | durable `≠ srv` indicator + "App updated / Reload" prompt, driven by a provable commit mismatch — **desktop and mobile** (shared `ui/StaleBadge`) |
| **terminal-not-found** | typed `ORPCError("NOT_FOUND")` instead of a bare error oRPC scrubbed to an opaque "Internal server error" |

### How an already-deployed worker retires itself

`public/sw.js` is a self-destructing worker: the browser's own SW update check re-fetches `/sw.js` (kept `no-cache`), installs it, and on activate it deletes caches, unregisters, and reloads controlled tabs — so a worker retires **even while it still controls the page** (when the app's JS can't run to remove it). New visitors never register it.

[...]

PR #696, opened 2026-04-24
Stop serving stale client assets after deploys

srid:
Restarting the Kolu PWA after a deploy would sometimes leave the app running on yesterday's JavaScript, CSS, or HTML — **the first page load after a deploy now reliably lands on the fresh build**. There were two independent causes, one that hit every user and one that hit only service-worker-capable deployments.

On the server side, Hono's `serveStatic` emits no `Cache-Control`, `ETag`, or `Last-Modified` headers, so browsers fell back to heuristic caching on every asset including `index.html`. A small middleware (`cacheControl.ts`) now pins hashed `/assets/*` as `immutable` forever and forces `no-cache, must-revalidate` on the SPA shell and service-worker scripts. _This is the only fix that helps users running Kolu over plain HTTP_, where service workers are disabled by the secure-context rule and there's no Workbox safety net to update independently.

On SW-capable deployments (`localhost` or HTTPS), the _"Server updated — Reload"_ modal called `location.reload()` synchronously. The old service worker would then serve the reloaded page from its Workbox precache while the new one was still installing in parallel, leaving the user on stale UI until a second reload. A new `forceUpdateAndReload` helper in `pwa.ts` awaits `registration.update()` first, so the new SW — already configured with `skipWaiting` + `clientsClaim` by vite-plugin-pwa's autoUpdate defaults — is controlling the page by the time navigation fires. _No-op on HTTP: `navigator.serviceWorker` is `undefined` in insecure contexts per the `[SecureContext]` IDL annotation, so the optional chain gracefully falls through to a plain reload._

> The commit history reads as a progression: primary fix, two Lowy-driven extractions that pull the cache-policy and SW-update concerns out of `index.ts` and `TransportOverlay.tsx`, a fact-check fix that catches `update()` rejection so the user's reload intent is honoured even when the update attempt fails, and an elegance pass naming the revalidate path set.

PR #1154, opened 2026-06-03
feat(surface-app): @kolu/surface-app — the app shell for surface apps

srid:
## What

Introduces **`@kolu/surface-app`** — the *app shell* companion to `@kolu/surface`. Where surface is the live reactive **wire**, surface-app owns everything delivered *around* it: fresh delivery, server/build identity, the connection & update lifecycle, and desktop install. kolu now **consumes** it (deleting its hand-rolled `cacheControl` + connection-lifecycle layers); **drishti** is the second consumer ([srid/drishti#47](https://github.com/srid/drishti/pull/47)) — the rule-of-three that justified the extraction.

It exists because one property — *a returning client converges to the build you deployed* — was re-derived from scratch four times across kolu (#696 / #1125 / #1135 / #1149), slightly differently each time, leaving a gap each time. Design-of-record: the Atlas note `docs/atlas/src/content/atlas/surface-app.mdx`.

## The freshness contract (5 invariants)

1. **One mutable entry point; everything else immutable** — the `index.html` shell is `no-store`, content-hashed `/assets/*` are `immutable`, an asset-miss **404**s (never the shell under a `.js` URL). Staleness is structurally impossible.
2. **Build identity is first-class and single-sourced** — client and server stamp the *same* commit (`resolveCommit`: env → git → `"dev"`).
3. **Skew is visible and recoverable** on every form factor.
4. **No service worker** for this always-connected class — it ships none and *retires* any legacy worker (self-destructing `/sw.js`).
5. **The client always knows its relationship to the server** — host, build, live status — as a headless model.

## Compose, don't hand-wire

The library ships **fragments**; the app is their composition, with no bespoke glue:

| Seam | What it owns |
|---|---|
| `installSurfaceApp` (`/server`) | serve the shell fresh + dynamic manifest + self-destructing `/sw.js` |
| `buildInfo`/`defineBuildInfo` + `buildInfoServer()` | build identity **as an interface** (default `{commit}`; kolu extends with a pty-host axis) |
| `surfaceAppSurface` + `composeSurfaces` | merge the build cell + restart probe into your surface in **one** merge (probe namespaced `surface.surfaceApp.info`) |
| `surfaceAppServer` + `implementSurfaceApp` | **one-call** server composition — merges the impls, runs `implementSurface`, flows `buildInfo.connect` |
| `surfaceApp()` Vite plugin · `buildSurfaceClient()` (Bun) | the client build owned upstream — content-hashed assets + the `__SURFACE_APP_COMMIT__` define + the no-store shell rewrite |
| `nix/commit-stamp.nix` | single-source the build-commit env-var name + flake-rev resolution for nix consumers |
| `SurfaceAppProvider` / `useSurfaceApp()` (`/solid`) | the headless client model — you render the chrome (the `srv·client` identity rail) |

## Entrypoints

`.` (kernels) · `/server` · `/surface` · `/solid` · `/lifecycle` · `/vite` · `/bun` · `/client`.

## Example + CI + docs

A runnable example (`packages/surface-app/example`) exercises the full composition end-to-end and **builds in CI** (`surface-app-example-build`). Unit tests cover the kernels, the lifecycle, build-info async sourcing, and the composition helpers. README + Atlas note carry the design + wiring.


srid:
## Evidence — the spike runs

`@kolu/surface-app` + a runnable hello-world (`packages/surface-app/example`), built and served end-to-end. Server stamped commit `5e2e2bbb`; client baked `c11e7700` — so the skew is visible on first load.

### The freshness contract, on the wire (`curl`)

| Request | `Cache-Control` | Status | ✓ |
|---|---|---|---|
| `/` (shell) | `no-store` | 200 (text/html) | shell never cached |
| `/assets/index-R0Ant4BS.js` | `public, max-age=31536000, immutable` | 200 | content-hashed → pinned |
| `/assets/index-DEADBEEF.js` (miss) | — | **404** | a miss 404s, never the HTML shell |
| `/sw.js` | `no-cache, must-revalidate` | 200 (js) | retirement worker always re-fetched |
| `/foo/bar` (deep route) | `no-store` | 200 | SPA fallback to the no-store shell |

### The headless model + skew UI, live in the browser

The Solid client connected over the surface wire, pulled the `buildInfo` cell, and the headless `useSurfaceApp()` model drove the identity rail (no console errors). Rendered a11y tree:

```
banner
  "live"  ·  "SRV " "5e2e2bbb"  ·  "CLIENT " "c11e7700"  "≠ srv"  button "⟳ Reload"
```

i.e. a green **live** dot · **SRV `5e2e2bbb`** · **CLIENT `c11e7700`** · amber **≠ srv** chip · **⟳ Reload** button — exactly the contract the plan describes, from the same model that will drive kolu's and drishti's own chrome.

Run it: `pnpm --filter @kolu/surface-app-example dev` (or `build:client` + `start`).

> **Not yet wired** (follow-up): the nix fileset entry + `fetchPnpmDeps` hash refresh + CI integration — the spike runs via pnpm/tsx/vite. A hosted screenshot/video can follow via `/evidence`.

packages/surface-app/example/src/server/main.ts
```diff
@@ -0,0 +1,105 @@
...
+// biome-ignore lint/suspicious/noExplicitAny: MemoryPublisher's generic is too
+// strict for our payloads; type safety lives on the typed channels.
+const publisher = new MemoryPublisher<Record<string, any>>();
+
+const { router: surfaceRouter } = implementSurface(surface, {
+  channel: <T>(name: string) => publisherChannel<T>(publisher, name),
+  cells: {
+    buildInfo: {
+      store: { get: () => ({ commit: SERVER_COMMIT }), set: () => {} },
+    },
```
srid: The example should not hardcode the surface implementation. Insead, it should 'compose' it from the surface-app library

packages/surface-app/example/src/server/main.ts
```diff
@@ -0,0 +1,105 @@
...
+
+if (existsSync(DIST_DIR)) {
+  // The self-destructing worker, served no-cache so the browser's update check
+  // always re-fetches it (registered before the static catch-all).
+  app.get("/sw.js", (c) => {
+    c.header("Cache-Control", "no-cache, must-revalidate");
+    return c.body(SW_SOURCE, 200, {
+      "content-type": "text/javascript; charset=utf-8",
+    });
+  });
```
srid: Again, upstream this to the library and 'compose' it (category theory) here.

packages/surface-app/example/src/env.d.ts
srid, line 1: Again why hardcode it in example?

packages/surface-app/example/vite.config.ts
```diff
@@ -0,0 +1,25 @@
...
+// The client's baked-in commit. In a real app this is injected by the
+// surface-app commit stamp (Vite/Nix) or a Bun define; here we hardcode a value
+// DIFFERENT from the server's default so the `≠ srv` skew is visible on first load.
+const clientCommit = process.env.CLIENT_COMMIT || "c11e7700";
+
+export default defineConfig({
+  root: "src/client",
+  define: {
+    __SURFACE_APP_COMMIT__: JSON.stringify(clientCommit),
+  },
```
srid: Can we find a way not to hardcode it, and 'compose' from the library itself? This and the clientCommit line above.

packages/surface-app/example/vite.config.ts
```diff
@@ -0,0 +1,25 @@
+import { defineConfig } from "vite";
+import solid from "vite-plugin-solid";
+
+// The client's baked-in commit. In a real app this is injected by the
+// surface-app commit stamp (Vite/Nix) or a Bun define; here we hardcode a value
+// DIFFERENT from the server's default so the `≠ srv` skew is visible on first load.
+const clientCommit = process.env.CLIENT_COMMIT || "c11e7700";
```
srid: This hardcoded fallback is retarded, wtf?

srid:

**Outcome:** `consensus` after 1 round(s). Independent review: lowy=8, hickey=7.

The two lenses converged in a single round, agreeing on 10 fixes and 5 drops with no unresolved items. The most consequential cluster — the immutable-prefix contract — was independently surfaced by both lenses from different angles, producing tightly-scoped, non-redundant fixes that together make the `build.ts` producer an honest citizen of the caching contract the server enforces.

---

### Finding summary

|---|---|---|---|---|
| lowy | Asset-prefix convention is split across the receptacle — producer half leaks | packages/app/src/server/build.ts:58 | fix | 3af86089a |
| lowy | Two content-hash schemes behind one immutable-caching contract | packages/app/src/server/build.ts:131 | fix | c1b66fbfb |
| lowy | styles.tmp.css written inside the immutable /assets/ prefix | packages/app/src/server/build.ts:124 | fix | 61fdfd4ae |
| lowy | Dynamic manifest route shadows a still-shipped static manifest file | packages/app/src/server/build.ts:158 | fix | 2de412803 |
| lowy | index.html carries dual theme-color meta the dynamic manifest can't see | packages/app/src/client/index.html:16 | drop | — |
| lowy | IdentityRail ported as a verbatim copy rather than consumed from the receptacle | packages/app/src/client/TabStrip.tsx:79 | fix | 9a8916cbb |
| lowy | clientCommit and probe both reach through to one global client — symmetric and correct | packages/app/src/client/App.tsx:309 | drop | — |
| lowy | Per-leaf @kolu/* overlay correctly encapsulates each package's volatility axis | nix/overlay.nix:13 | drop | — |
| hickey | PWA manifest fragmented across two sources of truth (static file + inline server route) | packages/app/src/server/main.ts:154-170 + packages/app/src/client/public/manifest.webmanifest:1 | fix | — |
| hickey | srv·client identity rail + Commit cell hand-ported from kolu rather than shipped by surface-app | packages/app/src/client/TabStrip.tsx:79-148 | drop | — |
| hickey | Two divergent content-hash schemes for two assets that share one caching contract | packages/app/src/server/build.ts:83 + 131-133 | fix | — |
| hickey | CSS temp file written inside the immutable /assets/ prefix it must not be served from | packages/app/src/server/build.ts:124-135 | fix | — |
| hickey | Build-commit single-sourcing relies on three independent stamps agreeing by convention | default.nix:69 + default.nix:80 + flake.nix:34 + packages/app/src/server/build.ts:89 | fix | eaf58e200 |
| hickey | Two ml-auto right-align anchors compete in the same flex row | packages/app/src/client/TabStrip.tsx:89 + 157 | fix | 4f5405612 |
| hickey | SRV_DOT typed as Record<string,string> loses exhaustiveness over ConnectionStatus | packages/app/src/client/TabStrip.tsx:79-84 | fix | — |

---

### The immutable-prefix cluster (lowy-2 / lowy-3 / hickey-3 / hickey-4)

These four findings converge on the same zone of `build.ts` and were filed independently — lowy from a "volatility boundary" angle, hickey from a "complecting two concerns" angle — which made cross-examination rapid.

**lowy-2 / hickey-3: Two hash schemes for one caching invariant**

Lowy framed it as an _activity-volatility leak_: "how do we content-address an asset" is one axis, but the JS path delegates to Bun.build's internal `[hash]` token (result _discovered_ by scanning `jsResult.outputs`) while the CSS path hand-rolls `Bun.hash(bytes).toString(16).slice(0,8)` (name _constructed_ up front). Hickey independently framed the same asymmetry as "duplicated derivation, two ways" — the find-vs-construct flow difference is the tell.

Both lenses confirmed that `@kolu/surface-app` exports no `hashAsset` primitive, so folding into the receptacle would grow the pin. The agreed fix stays local:

> Factor `const hashName = (bytes, base, ext) => \`${base}-${Bun.hash(bytes).toString(16).slice(0,8)}.${ext}\`` in `build.ts`. Drop Bun.build's `naming:[hash]` token — emit the JS entry to a stable temp name (or read the single entry's bytes), run both JS and CSS through `hashName`, `Bun.write` each to its computed `/assets/` path. One scheme, one truncation length, one mental model.

**lowy-3 / hickey-4: Temp CSS file inside the immutable vault**

Lowy: "writing a non-content-hashed scratch file into the immutable vault momentarily violates the invariant the prefix encapsulates." Hickey: "the `rm -f` at line 135 has no `try/finally` guard — a throw between write and unlink orphans a fixed-name, never-revalidated file under a year-immutable URL, exactly the failure mode this PR exists to kill."

Both lenses noted that the `hashName` unification above dovetails: once you construct the CSS name up front from bytes, the tailwind output never needs a fixed name inside `/assets/` at all.

> Write the tailwind temp output outside the immutable surface — `resolve(distDir, ".styles.tmp.css")` at the dist root, or a `mkdtemp` path entirely outside `distDir` — then write only the final content-hashed `styles-<hash>.css` into `assetsDir`.

---

### The manifest cluster (lowy-4 / hickey-1)

These two were filed independently and collapse to one change — but hickey's entry is more complete because it caught a bonus factual error in the test file.

Lowy observed that `installPwaManifest` registers `app.get("/manifest.webmanifest")` _before_ `serveStatic`, so the static `client/public/manifest.webmanifest` copied by `cp -R public/.` is dead bytes that nothing can ever serve — two sources of truth for one manifest concept, coherence held only by discipline.

Hickey confirmed the same, and additionally verified that `build.test.ts` lines 14–17 contain an NB comment asserting _"neither is in the built dist"_ — which is factually wrong, since `cp -R public/.` does copy the static manifest into `dist`. The test merely stopped asserting it.

Both lenses also confirmed that the two copies have already diverged in encoding (static file spells `display: "standalone"` and `start_url` explicitly; inline relies on `installPwaManifest`'s defaults), making the "remember to edit both" maintenance rule live today.

> Delete `packages/app/src/client/public/manifest.webmanifest`. The inline `manifest` object in `main.ts:154-170` becomes the sole definition. Update `build.ts:151`'s `(and a fallback manifest)` comment; the `build.test.ts:14-17` NB comment then becomes accurate.

---

### The asset-prefix split (lowy-1)

Lowy's structural lens: `@kolu/surface-app` encapsulates the _serving_ half of the asset-prefix convention (`installFreshStatic` / `isImmutableAssetPath` default `'/assets/'`) but the _producing_ half (`const ASSET_DIR = "assets"` in `build.ts`) is an independent literal that must agree with the library's default by inspection. Surface-app deliberately exposes `assetPrefix` as an override input via `FreshnessPaths`, so the latent footgun's live: a default change in the library silently 404s every asset while the build keeps emitting to `/assets/`.

Hickey confirmed: `surface-app/src/index.ts:21` holds `DEFAULT_ASSET_PREFAI = "/assets/`" as a private const; `main.ts:152` calls `installSurfaceApp` without passing `assetPrefix`, leaning on the default; `build.ts:61` independently hardcodes `ASSET_DIR = "assets"`. Real fragmentation, real blast radius.

The agreed fix avoids growing the surface-app pin:

> In `build.ts`, export `export const ASSET_PREFIX = "/" + ASSET_DIR + "/`. Pass `assetPrefix: ASSET_PREFIX` through to the `installSurfaceApp` call in `main.ts`, so one literal flows to both the producer (emit dir) and the server's `FreshnessPaths`.

---

### Build-commit stamp fragmentation (hickey-5)

Hickey raised this from the Nix layer. One value — "what git rev is this deploy" `— reaches the running system through three stamps: (1) `flake.nix` derives `rev = substring 0 7 self.rev`; (2) `default.nix` threads it as `surfaceAppCommit` into the build derivation, where `buildPhase` exports `SURFACE_APP_COMMIT` so `build.ts`'s `resolveCommit()` bakes `__SURFACE_APP_COMMIT__` into the client bundle; (3) `default.nix` _separately_ `--set SURFACE_APP_COMMIT "${rev}"` on the server wrapper. The self-consistency invariant holds only because the same `rev` Nix variable is wired to both call sites by hand. If a future edit threads `rev` into the derivation but forgets the wrapper `--set`, the rail shows `srv <sha> ✘ client <other>` and falsely flags every tab stale — the exact false-positive `isCleanRef` was added to avoid.

Lowy confirmed all three sites in the Nix diff and noted the fix lives entirely in `default.nix` with no pin change required.

> In `default.nix`, bind the commit once (e.g. `let commit = rev;` or expose `drishtiBuilt.surfaceAppCommit`) and reference it at both the `callPackage` `surfaceAppCommit` arg and the `makeWrapper --set` site, so client commit and server commit read from one Nix expression and cannot be edited apart.

---

### Layout complecting: two `ml-auto` anchors (hickey-6)

Hickey raised this as a pure structural finding. After this PR the flex-wrap tab strip has two children carrying `ml-auto` — `IdentityRail`'s root `<div>` (line 89) and `ThemeToggle`'s `<button>` (line 157). The first eats the slack; the second's `ml-auto` is a no-op. Both components' comments independently claim to own the right edge, which is the tell that the layout role is braided across two class strings. Lowy confirmed and agreed the fix is a trivial structural cleanup.

> Wrap `IdentityRail` + `ThemeToggle` in a single trailing `<div class="ml-auto flex items-stretch">`, carrying the only `ml-auto`. Drop `ml-auto` from both children. Update both comments so neither still claims to self-align.

---

### `SRV_DOT` exhaustiveness (lowy-6 / hickey-7)

Both lenses filed this independently. `ConnectionStatus` is a closed union (`"live" | "reconnecting" | "restarted" | "down"`) exported from `@kolu/surface-app/solid`, but `SRV_DOT` is typed `Record<string, string>` with a `?? "bg-gray-400"` fallback at the read site — discarding the closed union, so a fifth status silently greys instead of failing to compile. Both lenses noted that drishti's own `connectionColors.ts` uses `Record<ConnectionState, StatePresentation>` for exactly this exhaustiveness, so the new code regresses from the project's established pattern.

> Type `const SRV_DOT: Record<ConnectionStatus, string>`, importing `ConnectionStatus` from `@kolu/surface-app/solid`. Drop the now-unreachable `?? "bg-gray-400"` at `TabStrip.tsx:97`.

---

### Drops and boundary confirmations

Three findings were confirmed clean and dropped explicitly:

**lowy-7 (drop):** `SurfaceAppProvider` is fed `controlPlane=adminClient()`, `ws=adminSocket()`, `probe=()=>adminClient().rpc.surface.server.info({})` — all three plug into the same admin surface. The `ControlPlane` interface makes passing a per-host client a compile error. Correct receptacle wiring; no action.

**lowy-8 (drop):** The Nix overlay's per-leaf `@kolu/*` sourcing is symmetric across all three leaves (overlay entry, `env.nix` export, three hydrate callers). `npins/sources.json` is bumped to `add1e5b` on branch `surface-app`. Boundary check confirmed clean.

**hickey-2 (drop):** The `IdentityRail` + `Commit` chrome duplication between drishti and kolu is real, but promoting the rail into `surface-app` would grow the pinned package surface in this adoption PR, require re-pinning kolu again after landing it upstream, and is the speculative-second-consumer generalization the lens cautions against when the two known call sites differ on both palette (drishti uses literal `gray-*`/`emerald-*`; kolu uses semantic tokens `fg-3`/`edge`/`ok`) and repo URL. The palette divergence is legitimately app-specific. The right-scoped win here is the typed `Record` (lowy-6 / hickey-7). The model/chrome split is a defensible library design the author declared explicitly. Drop for this PR.

**lowy-5 (drop):** `index.html` carries two `theme-color` metas split by `prefers-color-scheme: dark|light` — `#0b0d12` and `#f7f8fb` — which is a distinction the web app manifest's single `theme_color` field literally cannot represent. The HTML head and the manifest encode _different_ facts (per-scheme browser-chrome tint vs. single install-time theme). The only real overlap is the dark hex value `#0b0d12` appearing in both `index.html:17` and `main.ts:157` — a one-token coincidence. A shared-constant fix would have to inject a hex string across the HTML/server build boundary, adding a seam worth more than the drift it prevents. Drop.

srid:
## Codex ⇄ Claude debate

**✅ Consensus** reached in **3 rounds**. codex reviewed at **`xhigh`** reasoning effort (read-only sandbox); Claude authored the fixes. Base = merge-base `a3120863`.

| Round | codex verdict | open findings | Claude dispositions | commit |
|---|---|---|---|---|
| 1 | not approved | 13 | fixed 12, partially F8 (lifecycle listener cleanup) | `f23e4b90` |
| 2 | not approved | 3 (F4, F11, F14) | fixed all 3 (probe-error routing, dead docs link, README accessor wiring) | `a95938a6` |
| 3 | **approved** | 0 | close-out verification — no further changes | — |

**What the debate changed (the substantive findings):**
- **F1/F2** — `buildInfo` cell is `verbs: ["get"]` (server identity is not client-writable); async extended sources seed from the schema-valid default and overlay, so no incomplete pre-settle shape is ever published.
- **F3/F5/F6** — kolu reuses the app-wide status source in `SurfaceAppProvider` (no second lifecycle); `buildInfo.use` and rejected async sources route through the provider's `onError` channel, leaving the seeded value intact.
- **F4** — the turnkey `{ ws, probe }` provider path now surfaces probe failures: `createServerLifecycle` accepts `onProbeError`, mapped to `onError` with `Error` normalization. Covered by `lifecycle.test`.
- **F7/F8** — `ConnectionSource` is a discriminated union (no representable partial state); lifecycle listeners are disposed via `onCleanup` under a Solid owner. New `solid/lifecycle.test.ts` covers connect/reconnect/restart ordering, probe-error reporting, and dispose cleanup.
- **F10/F13** — `hono`/`@hono/node-server` are optional peer deps (client-only consumers aren't forced to pull the server); `workbox-window` + `vite-plugin-pwa` dropped from the client now that surface-app owns the self-destructing `/sw.js`.
- **F11/F14** — fixed the dead `docs/cache-bug.md` link (now points at the migrated Atlas note) and corrected the README to describe the actual single shared status accessor wiring from `rpc.ts`.

Commits sit on the branch for human review — not pushed or merged by the debate.

srid:

**✅ Consensus** in **1 round**. Independent parallel review (both on Opus) → debate to agreement on every finding. Independent counts: **lowy 6**, **hickey 5** (11 findings). Outcome: **6 fix, 5 drop**, 0 unresolved. Base = merge-base `a3120863`.

|---|---|---|---|---|
| lowy | Provider exposed staleness through two competing inputs (`buildInfo.isStale` + standalone `isStale` prop) | `solid/index.ts` | **fix** — dropped the standalone prop; one axis, one socket | `cf8cb07` |
| lowy | Server identity "split across two `info` receptacles" | `contract.ts`, `router.ts`, `surface.ts` | drop — different namespaces/clients; split is by delivery timing and already documented | — |
| lowy | `buildInfoServer` "encapsulates two axes (value-shape + arrival-timing)" | `server.ts` | drop — single coherent seed-then-overlay axis | — |
| lowy | Root `.` export "a four-concern junk drawer" | `index.ts` | drop — the four are one cluster (the freshness contract) | — |
| lowy | SW lockstep comment prescribed a now-unnecessary test | `index.ts` | **fix** — reworded to the verbatim-serve invariant | `050a69a` |
| lowy | `vite.ts` duplicates `resolveCommit` (single-source softening) | `vite.ts` | **fix** — clarified it's the deliberate self-contained build-time half | `c8d2b93` |
| hickey | `/sw.js` directive in two places with a dead fallback | `server.ts` | **fix** — collapsed to one source, removed dead branch | `9f126a5` |
| hickey | Worker-retirement logic forked into two copies | `index.ts` + `lifecycle.ts` | **fix** — folded into the `lowy-5` edit (Part B dropped as redundant cross-ref) | (in `050a69a`) |
| hickey | `ServerProbe` wire shape defined twice with a false single-source claim | `surface.ts` + `solid/index.ts` | **fix** — single-sourced the type | `3942966` |
| hickey | `PtyHostIdentitySchema` "duplicated" kolu-common ↔ @kolu/pty-host | `common/src/surface.ts` | drop — out of this PR's scope / pre-existing | — |
| hickey | Two named constants for the no-store directive | `index.ts` | drop — distinct call sites, intentional | — |

The lenses honored the threaded rationale: the new package boundary, build-identity-as-interface, "compose don't hand-wire," and the deliberate `cacheControl.ts` deletion were all left untouched as intentional. The fixes were narrow coherence/dedup tightenings within that boundary.

Commits sit on the branch for human review — not pushed or merged.

srid:

Final pass (rules → fact-check → elegance) on the most-settled tree, after codex-debate and lens-debate. Light touch as expected.

**Fixed (3) — commit `7e208ab`:**

1. **Exhaustive status table** (`surface-app/src/solid/index.ts`) — replaced the 5-case `statusOf` switch (whose `default: return "live"` would silently collapse any *future* lifecycle kind into `live`) with `Record<ServerLifecycleEvent["kind"], ConnectionStatus>`. TypeScript's required-property check now enforces exhaustiveness at the type level: adding a new kind to the union is a compile error at the lookup table, not a runtime fallthrough. (Typecheck confirmed green.)
2. **Documented intentional silent catches** (`solid/index.ts`) — the Badging API `.catch(() => {})`es are best-effort decoration; added the justification the `no-silent-error-swallowing` rule requires when a catch is deliberately empty.
3. **Probe failure log level** (`client/src/rpc/rpc.ts`) — a permanently-broken `server.info` probe is a real I/O failure, so `console.error`, not `console.warn` (which production operator tooling filters out). Carried forward correctly from the pre-extraction code.

**Skipped (2) — both correctly below threshold:**
- A `JSON.stringify` equality lambda appears 3× (`server/src/surface.ts` ×2, `surface-app/src/server.ts` ×1) — exactly the `dry-rule-of-three` boundary, each already has a local rationale comment; extracting a helper would add indirection without clarity gain.
- `pwaIdentityForHostname` called per-request in `router.ts` and once at module level in `index.ts` — pure, O(1), across separate modules with no shared binding; negligible.

Commits sit on the branch for human review — not pushed or merged.

srid:
## Codex ⇄ Claude debate

**✅ Consensus** in **3 rounds**. codex reviewed at **`xhigh`** reasoning effort (read-only sandbox); Claude authored the fixes. This pass caught real correctness bugs, not nits.

| Round | codex verdict | open | commit |
|---|---|---|---|
| 1 | not approved | 6 | `dc2f682` |
| 2 | not approved | 1 (F6 docs) | `a707370` |
| 3 | **approved** | 0 | — |

**What the debate fixed:**
- **F1 (major)** — `buildSurfaceClient` rewrote the HTML shell with `replaceAll` and never verified the placeholders existed: a typo/stale template could ship a "successful" build whose `index.html` still points at **dev/unhashed assets**, silently breaking the freshness contract. Now it **throws** when the entry or any extra-asset placeholder is absent.
- **F2 (major)** — `setAttention` snapshotted `document.title` at module-eval and restored that stale value on clear, clobbering an app-driven title. Now derives the base title from the live `document.title` and strips only its own prefix.
- **F3 (major)** — `createServerLifecycle` counted raw WS opens, so a successful probe **after a failed initial probe** was misreported as `reconnected` instead of the first `connected`. Now keyed off the first successfully observed `processId` (+ regression test in `solid/lifecycle.test.ts`).
- **F4 (major)** — async `buildInfo` failures were swallowed and `ready` still resolved. Now routed to `onError`/`console.error` (no longer silent) while preserving the documented schema-valid seed fallback.
- **F5 (major)** — `implementSurfaceApp`'s spread merge let app deps shadow the fragment-owned `buildInfo` cell / `surfaceApp` procedure namespace. Now `assertNoOverride` throws before merging — matching `composeSurfaces`'s duplicate-key rejection.
- **F6 (minor)** — docs drift: the drishti usage snippet showed stale server wiring + omitted the required `clientCommit` prop, and described not-yet-shipped `app.standalone`/`app.install`/favicon affordances. README + Atlas (source + regenerated HTML) corrected to the shipped surface.

Commits sit on the branch for review — not pushed/merged by the debate.

srid:

**✅ Consensus** in **1 round**. Independent parallel review (both on Opus): **lowy 7**, **hickey 5** findings → **3 fix, 8 drop**, 0 unresolved. (lowy-1 and hickey-1 independently found the *same* defect — converged.)

|---|---|---|---|---|
| lowy + hickey | `ServerProbe` wire shape declared twice, "kept hand-equal" (interface in `/solid` + zod schema in `/surface`) — contradicts the package's own single-source thesis | `solid/index.ts`, `surface.ts` | **fix** — `ServerProbe = z.infer<typeof ServerProbeSchema>`, re-exported from `/solid` (type-only import, no zod value leaks) | `c85820e` |
| hickey | Example re-specifies the library-default `isStale` verbatim | `example/.../surface.ts` | **fix** — drop the redundant override; inherit the default | `41a099a` |
| hickey | "should prompt a reload" derivation lived in the consumer, not the headless model | `solid/` (+ `TransportOverlay`, `provider.test`) | **fix** — moved into `useSurfaceApp()` so every consumer reads one derivation | `a5e6997` |
| lowy | env-var name single-sourced by a comment across TS↔Nix | `vite.ts`, `commit-stamp.nix` | drop — TS/Nix can't share a literal; a CI grep would be a third fragile copy; risk is small + self-announcing |
| lowy | `implementSurfaceApp` re-implements `composeSurfaces`' collision check | `server.ts` | drop — different layers (impl vs spec); the parallel is shallow |
| lowy | Vite plugin & `buildSurfaceClient` both emit the `__SURFACE_APP_COMMIT__` define | `vite.ts`, `bun.ts` | drop — two bundler integration points, intentionally parallel |
| lowy | `serverIdentity` name overloaded (spec fragment + impl fn) | `surface.ts`, `server.ts` | drop — the spec/impl pairing is the package's consistent convention (cf. `buildInfo`/`buildInfoServer`) |
| lowy / hickey | `buildInfoServer` braids commit-resolution + async-seed lifecycle | `server.ts` | drop — one cohesive boot-time concern; splitting would fragment it |
| lowy | `cacheControlFor` hardcodes `/sw.js` policy | `index.ts` | drop — the SW retirement route is the library's own, not app-parameterized |
| hickey | Two named `no-store` constants (shell + asset-miss) | `index.ts` | drop — distinct policies that happen to share a value today |

The lenses honored the threaded rationale: the package boundary, build-identity-as-interface, the compose seams (`composeSurfaces`/`implementSurfaceApp`/`buildSurfaceClient`), and the `cacheControl.ts` deletion were all left as intentional. The 3 fixes are narrow single-source/derivation tightenings. Typecheck + 44 tests green.

Commits sit on the branch for review — not pushed/merged.

srid:

Final pass (rules → fact-check → elegance) on the most-settled tree, after codex-debate and lens-debate. Light touch as expected.

**Fixed (2) — commit `f1bdf2b`:**
1. **`packages/client/src/rpc/TransportOverlay.tsx`** — dropped the `updateReady` pure alias (no semantic value over the already-clear `pwa.updateReady()`); inlined into the `<Show>` condition. `disconnected` stays (used twice; the name earns its keep).
2. **`packages/surface-app/example/src/client/App.tsx`** — `STATUS_LABEL` typed `Record<ConnectionStatus, string>` (was `Record<string, string>`, which defeated TS's exhaustiveness check — `prefer-ts-pattern`); the `?? pwa.status()` fallback was then dead and removed.

**Skipped (2) — both correctly:**
- `pwa.server()` "repeated" in `IdentityRail.tsx` — the extra calls are inside a commented-out pty column; only one live call site.
- `assertNoOverride` inline closure in `server.ts` — small, single-use, clearly scoped; extracting it would add abstraction with no clarity gain.

Monorepo typecheck + 44 surface-app tests green.

PR #1155, opened 2026-06-03
Rebuild the Markdown preview: GFM, inline HTML, syntax highlighting, and light/dark theming

srid:
The Code-tab's rendered Markdown preview was a hand-rolled `marked` token walker that **escaped every piece of inline HTML, flattened headings, dropped list markers, and rendered with cramped spacing** — so a normal README showed `<p align="center">`/`` as literal text, no tables, unmarked lists, and leaked `[^1]`/`> [!NOTE]`/code as plain text. This rebuilds the renderer in `@kolu/solid-markdown` on a real pipeline: full GitHub-Flavored Markdown, sanitized inline HTML, syntax-highlighted code, repo-image resolution, read-only task lists, and styling that follows the app's light/dark preference.

### Pipeline

```
markdown ─▶ marked (GFM + footnotes/alerts/heading-ids,  ─▶ DOMPurify        ─▶ .kolu-md container
           front-matter strip, soft-break policy)           sanitize.ts          markdown.css
           render.ts — DOM-free, unit-tested                tight allowlist;     currentColor + color-mix
                                                            link policy + image   + --color-accent →
                                                            resolve + id ns +     follows light/dark
                                                            Shiki highlighting    with no theme prop
```

### What renders now

- **Full GFM** — headings (real scale + anchor ids), tables (per-column alignment), **lists with disc/decimal/nested markers**, task lists, strikethrough, autolinks, blockquotes, `hr`
- **GitHub extensions** — **footnotes**, **`> [!NOTE]` alerts**, leading **YAML front-matter** stripped, **GitHub-faithful soft breaks** (the document preview folds a single newline to a space; chat/dock keep message-style breaks)
- **Syntax-highlighted code** — fenced blocks via **Shiki** (lazy-loaded, dual github-light/dark theme that follows the app, code escaped so its output is trusted-by-construction) with a **copy button**
- **Task lists (read-only)** — GFM checkboxes render with their `[x]`/`[ ]` state but are presentational (`disabled`), the way GitHub renders a README's task list; the preview never writes back to the file
- **Inline HTML, sanitized** — `<details>`, `<kbd>`, `<sub>`/`<sup>`, `<p align>`, definition lists, figures, images; `<script>`/`<style>`/`<iframe>`/`class` stripped, `javascript:` links inert
- **Repo images load** — a repo-relative ``/`` resolves against the doc's directory and loads from the per-terminal file route, with a labelled-chip fallback
[...]

PR #1156, opened 2026-06-03
fix(be-review): pure-JS base64 in Report — TextEncoder crashes the workflow runtime

srid:
## The bug

A `/be-review` run on #1155 crashed in its **Report** phase with `ReferenceError: TextEncoder is not defined`, so no per-track comments were posted and the report had to be reconstructed by hand — **with no commit links** ([example](https://github.com/juspay/kolu/pull/1155#issuecomment-4615086060)).

Root cause: the Report phase base64-encodes each comment body so the mechanical commenter agent treats it as **opaque data**, not prompt instructions (the F3 prompt-injection fix from #1153). `toBase64` fell back to `new TextEncoder()` when `Buffer` was undefined — but the **Workflow runtime has none of `Buffer`, `TextEncoder`, or `btoa`** (the same class of restriction as `Date.now()`/`Math.random()`). So the fallback threw and took down the whole Report phase.

## The fix

Replace the fallback with a **self-contained pure-JS UTF-8→base64 encoder** that uses only `encodeURIComponent` (always present) + a hand-rolled base64 alphabet — no runtime globals. The `Buffer` fast-path stays behind a safe `typeof` guard for runtimes that have it.

## Verification

- Byte-for-byte **identical to `Buffer.from(s,'utf8').toString('base64')`** across ASCII, multibyte (`café`, `résumé`), emoji (`🤖🔴`), markdown special chars (`| \n \` $`), and empty string.
- **Round-trips** cleanly through `base64 -d` (the decode side `postComment` runs).
- Full-file syntax check passes; `.apm`→generated in sync; `just fmt` clean.

This is the bug that blocks reporter-authored comments (incl. the `remapCommits` SHA-resolution from #1153) from ever reaching the PR — once Report stops crashing, those land with resolving commit links.

### DIFF
--- a/.agents/skills/be-review/be-review.workflow.js
+++ b/.agents/skills/be-review/be-review.workflow.js
@@ -1029,15 +1029,37 @@ async function authorBody(slug, data, baseline, guidance) {
 // UTF-8-safe base64. The Workflow runtime has NEITHER `Buffer` NOR `TextEncoder`
 // NOR `btoa` (the old `new TextEncoder()` fallback threw `ReferenceError:
 // TextEncoder is not defined` and crashed the whole Report phase), so the fallback
-// is a self-contained pure-JS encoder using only `encodeURIComponent` (always
-// present) for UTF-8 and a hand-rolled base64 alphabet — no runtime globals. The
-// `Buffer` fast-path is kept behind a `typeof` guard (which never throws) for
-// runtimes that do have it; the pure path is byte-for-byte identical output.
+// is a self-contained pure-JS encoder that hand-rolls UTF-8 byte emission and a
+// base64 alphabet — no runtime globals. The `Buffer` fast-path is kept behind a
+// `typeof` guard (which never throws) for runtimes that do have it; the pure path
+// is byte-for-byte identical output. UTF-8 is computed manually rather than via
+// `encodeURIComponent`, which throws `URIError` on lone surrogate halves; here, as
+// in Buffer/TextEncoder, an unpaired surrogate is replaced with U+FFFD so the
+// encoder is total over every JS string and never crashes the Report phase.
 function toBase64(s) {
   const str = String(s)
   if (typeof Buffer !== 'undefined') return Buffer.from(str, 'utf8').toString('base64')
-  // UTF-8 bytes as a binary string, without TextEncoder: %XX escape each byte.
-  const bin = encodeURIComponent(str).replace(/%([0-9A-F]{2})/g, (_, h) => String.fromCharCode(parseInt(h, 16)))
+  // UTF-8 bytes as a binary string, decoding surrogate pairs and substituting
+  // U+FFFD for any unpaired surrogate (matching Buffer/TextEncoder semantics).
+  let bin = ''
+  const emit = (cp) => {
+    if (cp < 0x80) bin += String.fromCharCode(cp)
+    else if (cp < 0x800) bin += String.fromCharCode(0xc0 | (cp >> 6), 0x80 | (cp & 0x3f))
+    else if (cp < 0x10000) bin += String.fromCharCode(0xe0 | (cp >> 12), 0x80 | ((cp >> 6) & 0x3f), 0x80 | (cp & 0x3f))
+    else bin += String.fromCharCode(0xf0 | (cp >> 18), 0x80 | ((cp >> 12) & 0x3f), 0x80 | ((cp >> 6) & 0x3f), 0x80 | (cp & 0x3f))
+  }
+  for (let i = 0; i < str.length; i++) {
+    const c = str.charCodeAt(i)
+    if (c >= 0xd800 && c <= 0xdbff) {
+      const next = str.charCodeAt(i + 1)
+      if (next >= 0xdc00 && next <= 0xdfff) {
+        emit(0x10000 + ((c - 0xd800) << 10) + (next - 0xdc00))
+        i++
+      } else emit(0xfffd) // unpaired high surrogate
+    } else if (c >= 0xdc00 && c <= 0xdfff) {
+      emit(0xfffd) // unpaired low surrogate
+    } else emit(c)
+  }
   const CH = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/'
   let out = ''
   for (let i = 0; i < bin.length; i += 3) {
--- a/.agents/skills/be-review/be-review.workflow.js
+++ b/.agents/skills/be-review/be-review.workflow.js
@@ -1063,10 +1063,11 @@ function toBase64(s) {
   const CH = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/'
   let out = ''
   for (let i = 0; i < bin.length; i += 3) {
+    const n = Math.min(3, bin.length - i)
     const a = bin.charCodeAt(i)
-    const b = i + 1 < bin.length ? bin.charCodeAt(i + 1) : 0
-    const c = i + 2 < bin.length ? bin.charCodeAt(i + 2) : 0
-    out += CH[a >> 2] + CH[((a & 3) << 4) | (b >> 4)] + (i + 1 < bin.length ? CH[((b & 15) << 2) | (c >> 6)] : '=') + (i + 2 < bin.length ? CH[c & 63] : '=')
+    const b = n > 1 ? bin.charCodeAt(i + 1) : 0
+    const c = n > 2 ? bin.charCodeAt(i + 2) : 0
+    out += CH[a >> 2] + CH[((a & 3) << 4) | (b >> 4)] + (n > 1 ? CH[((b & 15) << 2) | (c >> 6)] : '=') + (n > 2 ? CH[c & 63] : '=')
   }
   return out
 }
--- a/.agents/skills/be-review/be-review.workflow.js
+++ b/.agents/skills/be-review/be-review.workflow.js
@@ -1026,15 +1026,27 @@ async function authorBody(slug, data, baseline, guidance) {
 // decodes it to the file mechanically and never sees it as prose, so no body content
 // can be confused with a workflow instruction. The deterministic baseline is itself a
 // valid body and rides the same opaque path (it's the fallback when authoring fails).
-// UTF-8-safe base64, runtime-agnostic: Node's Buffer if present, else TextEncoder +
-// btoa (handles multi-byte chars, which a bare btoa(string) would corrupt).
+// UTF-8-safe base64. The Workflow runtime has NEITHER `Buffer` NOR `TextEncoder`
+// NOR `btoa` (the old `new TextEncoder()` fallback threw `ReferenceError:
+// TextEncoder is not defined` and crashed the whole Report phase), so the fallback
+// is a self-contained pure-JS encoder using only `encodeURIComponent` (always
+// present) for UTF-8 and a hand-rolled base64 alphabet — no runtime globals. The
+// `Buffer` fast-path is kept behind a `typeof` guard (which never throws) for
+// runtimes that do have it; the pure path is byte-for-byte identical output.
 function toBase64(s) {
   const str = String(s)
   if (typeof Buffer !== 'undefined') return Buffer.from(str, 'utf8').toString('base64')
-  const bytes = new TextEncoder().encode(str)
-  let bin = ''
-  for (const b of bytes) bin += String.fromCharCode(b)
-  return btoa(bin)
+  // UTF-8 bytes as a binary string, without TextEncoder: %XX escape each byte.
+  const bin = encodeURIComponent(str).replace(/%([0-9A-F]{2})/g, (_, h) => String.fromCharCode(parseInt(h, 16)))
+  const CH = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/'
+  let out = ''
+  for (let i = 0; i < bin.length; i += 3) {
+    const a = bin.charCodeAt(i)
+    const b = i + 1 < bin.length ? bin.charCodeAt(i + 1) : 0
+    const c = i + 2 < bin.length ? bin.charCodeAt(i + 2) : 0
+    out += CH[a >> 2] + CH[((a & 3) << 4) | (b >> 4)] + (i + 1 < bin.length ? CH[((b & 15) << 2) | (c >> 6)] : '=') + (i + 2 < bin.length ? CH[c & 63] : '=')
+  }
+  return out
 }
 
 async function postComment(slug, body) {
--- a/.apm/skills/be-review/be-review.workflow.js
+++ b/.apm/skills/be-review/be-review.workflow.js
@@ -1029,15 +1029,37 @@ async function authorBody(slug, data, baseline, guidance) {
 // UTF-8-safe base64. The Workflow runtime has NEITHER `Buffer` NOR `TextEncoder`
 // NOR `btoa` (the old `new TextEncoder()` fallback threw `ReferenceError:
 // TextEncoder is not defined` and crashed the whole Report phase), so the fallback
-// is a self-contained pure-JS encoder using only `encodeURIComponent` (always
-// present) for UTF-8 and a hand-rolled base64 alphabet — no runtime globals. The
-// `Buffer` fast-path is kept behind a `typeof` guard (which never throws) for
-// runtimes that do have it; the pure path is byte-for-byte identical output.
+// is a self-contained pure-JS encoder that hand-rolls UTF-8 byte emission and a
+// base64 alphabet — no runtime globals. The `Buffer` fast-path is kept behind a
+// `typeof` guard (which never throws) for runtimes that do have it; the pure path
+// is byte-for-byte identical output. UTF-8 is computed manually rather than via
+// `encodeURIComponent`, which throws `URIError` on lone surrogate halves; here, as
+// in Buffer/TextEncoder, an unpaired surrogate is replaced with U+FFFD so the
+// encoder is total over every JS string and never crashes the Report phase.
 function toBase64(s) {
   const str = String(s)
   if (typeof Buffer !== 'undefined') return Buffer.from(str, 'utf8').toString('base64')
-  // UTF-8 bytes as a binary string, without TextEncoder: %XX escape each byte.
-  const bin = encodeURIComponent(str).replace(/%([0-9A-F]{2})/g, (_, h) => String.fromCharCode(parseInt(h, 16)))
+  // UTF-8 bytes as a binary string, decoding surrogate pairs and substituting
+  // U+FFFD for any unpaired surrogate (matching Buffer/TextEncoder semantics).
+  let bin = ''
+  const emit = (cp) => {
+    if (cp < 0x80) bin += String.fromCharCode(cp)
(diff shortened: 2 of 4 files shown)


srid:

This track hardened the pure-JS base64 path in `be-review.workflow.js` — the encoder that runs when the workflow runtime has no `Buffer`/`TextEncoder` — so that malformed UTF-16 input (lone/unpaired surrogates) can no longer crash report generation, and tightened the inner triplet loop. The two debate tracks (codex and lens) landed cleanly; both of code-police's findings turned out to describe the *same* defects those tracks had already fixed, more robustly, so they were dropped as fully subsumed.

Final branch head: 4d41f1b6cb63ab973f37ff0a082e86346a7703ac

#### Clean picks

|---|---|---|---|
| codex | 29ef96df | `clean` | 80a11f4c7 |
| lens | a7c31182 | `clean` | 4d41f1b6c |

- **codex (29ef96df → 80a11f4c7)** replaced the entire `toBase64` pure path. It no longer routes through `encodeURIComponent` at all; instead it hand-rolls UTF-8 byte emission with explicit surrogate handling, emitting `0xFFFD` (U+FFFD, the replacement character) for both unpaired high and low surrogates — matching `Buffer`/`TextEncoder` semantics. The U+FFFD substitution is documented in an inline comment.
- **lens (a7c31182 → 4d41f1b6c)** reworked the base64 inner loop to compute the triplet size exactly once via `n = Math.min(3, bin.length - i)`, then branch on `n > 1` / `n > 2`, eliminating the repeated inline boundary re-tests.

#### Reconciliation — both code-police findings dropped as subsumed

Both police commits touched the same file, `.apm/skills/be-review/be-review.workflow.js`, and both targeted code that the earlier tracks had already rewritten. Neither was reverted-and-reapplied because doing so would mean *backing out* the stronger fixes already on the branch.

**police 74501f94 — surrogate sanitization (dropped, subsumed by codex).**
Police's fix sanitized lone surrogates (`\p{Surrogate}` → U+FFFD) *before* calling `encodeURIComponent`, so the pure base64 path could no longer throw a `URIError`. But codex had already removed `encodeURIComponent` from that path entirely — there is no longer any call that can throw, and the U+FFFD substitution police wanted is already implemented, and implemented more robustly (explicit per-surrogate handling rather than a pre-pass regex). Police's diff targets a code path that no longer exists; reconciling it would mean reverting codex's stronger fix. Dropped as fully subsumed.

**police d230b28e — bounds-check dedupe nit (dropped, subsumed by lens).**
Police's nit hoisted `hasB = i+1 < bin.length` and `hasC = i+2 < bin.length` to dedupe the four inline bounds re-tests in the base64 inner loop. Lens had already addressed the identical finding by deriving `n = Math.min(3, bin.length - i)` once and branching on `n > 1` / `n > 2`. Both compute the triplet-size decision a single time and reuse it; the duplicated inline boundary checks are already gone in HEAD via lens's form. Police's version is just a stylistic re-spelling of an already-resolved nit, so nothing additional remained to fold in. Dropped as fully subsumed.

The convergence here is worth noting: three independent review tracks flagged the same two weaknesses in the pure base64 path. The debate tracks (codex, lens) happened to land first with the more complete fixes, leaving the police findings as confirmation rather than new work.

srid:
## 🤖 Codex ⇄ Claude debate

This track hardened the `toBase64` helper in the be-review workflow so the Report phase can no longer crash on pathological-but-valid comment bodies. The prior fix had already removed the `encodeURIComponent`/`btoa` dependency on runtime globals, but codex caught that the replacement pure-JS path was still **not total over JavaScript strings** — a single lone surrogate code unit would throw and take down comment posting and cleanup with it. By the end of round 2 the encoder emits UTF-8 by hand, substitutes U+FFFD for unpaired surrogates exactly like `Buffer`/`TextEncoder`, and is byte-for-byte verified against Node `Buffer` across ASCII, BMP, astral, and lone-surrogate inputs — with all three generated/source copies confirmed identical.

### How the debate converged

The exchange was unusually clean: a single major finding, conceded immediately by Claude, fixed in one commit, and re-verified to closure in round 2. No findings were disputed or rebutted — Claude agreed with codex's diagnosis on first contact, so the second round was pure verification rather than further argument.

| Finding | Severity | First raised | Claude disposition | Final status |
| **F1** — non-total UTF-8 fallback throws `URIError` on lone surrogates | 🟠 major | Round 1 | fixed (conceded) | ✅ resolved (R2) |

**Files changed** (the `.apm` source plus its two generated mirrors, kept md5-identical):
- `.apm/skills/be-review/be-review.workflow.js`
- `.agents/skills/be-review/be-review.workflow.js`
- `.claude/skills/be-review/be-review.workflow.js`

---

### Round 1 — codex approved ❌ · 1 open · commit 80a11f4c747c64bef9523e9df589484ef4005db6

**Codex's read:** the change correctly restored a working path for normal Unicode (the missing `TextEncoder`/`btoa` route) and the generated copies were in sync with the `.apm` source. But codex found one remaining correctness hole — the new fallback was not total over JavaScript strings, so the Report phase could still crash on some comment bodies.

#### 🟠 Major findings

- **F1** — at `.apm/skills/be-review/be-review.workflow.js:1040` · status: **open**
  The fallback used `encodeURIComponent(str)` to derive UTF-8 bytes, but `encodeURIComponent` throws `URIError` on lone surrogate code units. A reporter/baseline body containing an unpaired surrogate could therefore crash the Report phase **before comments are posted or cleanup runs**. `Buffer`/`TextEncoder` instead replace malformed surrogates with U+FFFD — so the fallback did not actually match the byte-for-byte equivalence it claimed.
  _Suggestion:_ make the string well-formed before encoding, or emit UTF-8 bytes manually with explicit surrogate handling that substitutes U+FFFD for invalid surrogate halves — and propagate the same fix to the generated `.agents`/`.claude` copies via APM regeneration.

**Claude's response — F1 → _fixed_ (conceded):**
Claude agreed with the diagnosis outright: `encodeURIComponent` throws `URIError` on lone surrogates, so the fallback was not total over JS strings and could crash the Report phase. The `encodeURIComponent` path in `toBase64` (`.apm/skills/be-review/be-review.workflow.js:1036`) was replaced with hand-rolled UTF-8 byte emission that iterates `charCodeAt`, combines valid high+low surrogate pairs into a code point, and emits U+FFFD for any unpaired high or low surrogate — matching `Buffer`/`TextEncoder` semantics. Claude tested against Node `Buffer` on the empty string, ASCII, `café ☕ 日本語`, emoji `😀👍🏽`, and strings containing lone `\uD800`/`\uDC00` halves: all byte-for-byte identical, zero throws. The fix was regenerated via `just ai::apm` so the `.agents` and `.claude` copies carry it too (all three md5-identical), and the incidental `apm.lock.yaml` timestamp churn from regeneration was reverted.

---

**Codex's verification:** F1 is adequately resolved. The `encodeURIComponent` fallback is gone, the pure-JS path now emits UTF-8 manually, handles valid surrogate pairs, substitutes U+FFFD for unpaired surrogate halves, and no longer depends on unavailable runtime globals. Codex independently confirmed the three generated/source copies are identical and ran its own parity check against Node `Buffer` for ASCII, BMP, astral, and lone-surrogate cases.

#### 🟠 Major findings

- **F1** — at `.apm/skills/be-review/be-review.workflow.js:1039` · status: **resolved**
  The prior pure-JS base64 fallback could throw `URIError` on lone surrogate code units, allowing unusual but valid JS strings in comment bodies to crash the Report phase instead of posting comments / cleaning up.
  _Suggestion:_ use a total UTF-8 encoder that matches `Buffer`/`TextEncoder` semantics, including U+FFFD replacement for unpaired surrogates — which the current implementation does.

---

**Net result:** one major correctness bug found and fixed across two rounds, both reviewers in agreement. The convergence was driven by concession rather than rebuttal — codex's first-round finding was precise enough that Claude fixed it without dispute, and round 2 was an independent re-verification (file parity + `Buffer` parity check) that confirmed closure.

srid:

**Outcome:** `consensus` after 1 round(s). Independent review: lowy=3, hickey=3 → 1 fix applied, 5 dropped, 0 unresolved.

This track put the PR's new pure-JS base64 encoder (`toBase64`, .apm/skills/be-review/be-review.workflow.js) under two structural lenses. The single change it produced: it collapsed a base64 inner loop that re-derived "how many bytes are in this triplet" four separate times into one decision, eliminating an unenforced invariant where a future edit could silently corrupt encoding. Everything else — the runtime-capability dispatch, the dual Buffer/pure implementations, the encode/decode asymmetry — was examined and consciously left as-is, with reasons recorded so the structure stays visible.

---

### How the two lenses split the work

The lenses look at different things, so they surfaced different findings:

- **lowy (volatility / boundary encapsulation)** ran the electricity test on `toBase64`: is it a clean *receptacle* that hides a genuinely volatile axis, and does any consumer have to reach past it? All three of lowy's findings were boundary observations — and all three concluded the boundary holds.
- **hickey (structural simplicity / complecting)** asked whether single concepts are implemented once or fragmented across the code. Hickey found one genuine fragmentation worth fixing, plus two deliberate-and-justified complectings.

Notably, **nothing flipped disposition during cross-examination** — both lenses agreed on all six findings in round 1 (see `history[]`), so consensus landed immediately. What's interesting is *why* each side conceded, since in every case the non-origin lens did independent verification rather than rubber-stamping.

---

### lowy's findings (volatility lens)

| # | finding | location | call | why |
|---|---|---|---|---|
| lowy-1 | `toBase64` is the receptacle; runtime-capability volatility is correctly encapsulated | :1036 | drop | electricity-test *result*, not a defect |
| lowy-2 | Encode receptacle is in-tree JS; decode receptacle is a shell string in a prompt | :1059 | drop | real but pre-existing; not in this diff |
| lowy-3 | Pure encoder relies on `encodeURIComponent` only emitting `%XX` triplets | :1040 | drop | true today; failure mode is a loud throw |

**lowy-1 — the receptacle holds.** This is the headline boundary result. The receptacle is `toBase64(s) → base64 string` — every consumer (`postComment` at :1054, and the deterministic-baseline path that rides the same opaque channel) plugs into one stable verb: "make this body opaque." Behind it sits the genuinely volatile axis — *which byte-encoding primitives the host runtime exposes* — encapsulating Node's `Buffer` when present and the pure-JS `encodeURIComponent` + hand-rolled-alphabet path when it isn't. The volatility is real, not speculative: the Workflow runtime **already dropped `TextEncoder` once**, which is the exact crash this PR fixes. lowy verified the boundary is clean — switching generators produces byte-for-byte identical output (ASCII / multibyte / emoji / control / null + 20k-case fuzz, ALL MATCH against Node's `Buffer`), so no consumer must expose the wires and reach a specific encoder. Recorded as the required electricity-test result. hickey conceded after independently re-running the parity corpus.

**lowy-2 — the one observation worth leaving on record.** The opaque channel has two ends of *one* volatility axis (its wire format): the encoder `toBase64` (in-tree, now hardened) and the decoder `printf %s '${b64}' | base64 -d` (a literal string baked into the commenter agent's prompt at :1059). They're **asymmetric receptacles for the same format living in different layers** (JS runtime vs. shell-in-prompt). If the wire format ever changed — say base64url to dodge `+//=` shell-quoting, or a move off the `base64` coreutil — a maintainer would have to touch two unrelated-looking places to keep one axis consistent. Both lenses agreed to drop it for two reasons: it's **pre-existing** (this diff only touches the encoder half) and the suggested fix (co-locate the format contract behind one named anchor) is explicitly optional and out of proportion for a crash-fix. Flagged per the no-defer rule so the axis is visible; left as-is.

**lowy-3 — loud, not silent.** The pure path assumes `encodeURIComponent(str)` yields only ASCII literals plus `%XX` triplets, which is true for all well-formed JS strings. On a lone surrogate, `encodeURIComponent` *throws* rather than emitting bad output, so `toBase64` fails loud rather than silently corrupting — an acceptable failure mode. lowy's fuzz harness (which skipped surrogate-throwers) and hickey's 20k-case BMP fuzz both found zero silent divergence. No defensive code warranted.

---

### hickey's findings (simplicity lens)

| # | finding | location | call | why |
|---|---|---|---|---|
| **hickey-1** | One "how many bytes in this triplet" decision fragmented across four inline conditionals | :1043-1048 | **fix** | one concept shattered into four bounds re-tests |
| hickey-2 | One encode concept implemented twice; agreement is an unenforced invariant | :1036-1050 | drop | duplication is intentional; guard is disproportionate |
| hickey-3 | Runtime-capability dispatch braided into the encoder | :1037-1038 | drop | minimal, correct complecting |

**hickey-1 — the one applied fix.** The base64 loop encodes a single decision — how many real input bytes this 3-byte group holds (1, 2, or 3) — but re-derived it **four separate times** in the loop body: `i + 1 < bin.length` to set `b`, `i + 2 < bin.length` to set `c`, then `i + 1 < bin.length` *again* to choose char-or-pad, and `i + 2 < bin.length` *again* for the final char-or-pad. The group size is one concept shattered into four boundary re-tests whose mutual agreement is an unenforced invariant — an edit that changes the bounds in one spot but not the others is a silent corruption the structure permits. The `out +=` line also complected three concerns (char assembly, padding selection, bounds guard) into the densest line in the change. The fix derives the group size **once** and reads padding off it:

```js
for (let i = 0; i < bin.length; i += 3) {
  const n = Math.min(3, bin.length - i)
  const a = bin.charCodeAt(i)
  const b = n > 1 ? bin.charCodeAt(i + 1) : 0
  const c = n > 2 ? bin.charCodeAt(i + 2) : 0
  out += CH[a >> 2] + CH[((a & 3) << 4) | (b >> 4)] + (n > 1 ? CH[((b & 15) << 2) | (c >> 6)] : '=') + (n > 2 ? CH[c & 63] : '=')
}
```

Same output, the triplet-size concept lives in one place. lowy conceded *and* independently confirmed the refactor is byte-for-byte identical across the corpus + a 3000-case fuzz (EQUIVALENT); hickey re-verified against Node's `Buffer` over corpus + 20k fuzz (ALL EQUIVALENT). Applied to `.apm/skills/be-review/be-review.workflow.js` only — `.claude/` and `.agents/` regenerate via `just ai::apm`. Commit 4d41f1b6cb63ab973f37ff0a082e86346a7703ac.

**hickey-2 — intentional duplication, but no mechanical guard.** `toBase64` now expresses "base64 of UTF-8 bytes" through two fully independent implementations (`Buffer.from(...).toString('base64')` and the hand-rolled pure path), coupled by the invariant "both produce byte-for-byte identical output" — maintained by a comment claim and out-of-band verification, not by structure. Drop-eligible from the start because the duplication is *deliberate* (Buffer = native fast path, pure = guaranteed fallback). The actionable remnant would be a mechanical equivalence guard, but it doesn't fit: the two paths only coexist where `Buffer` is **absent**,so a runtime self-check has no `Buffer` to compare against, and the Workflow runtime forbids gratuitous module-load work to keep resume deterministic. A standalone unit test is the only real guard and is disproportionate for a crash-fix. Equivalence re-confirmed out-of-band. Left as-is.

**hickey-3 — minimal complecting is the right call.** `toBase64` braids "which runtime am I on" (the `typeof Buffer` probe at :1038) with "how do I base64-encode." Both lenses agreed this is a *deliberate and minimal* complecting: the probe is a single never-throwing guard short-circuiting to an always-correct native fast path, with the always-correct pure path below. Lifting detection into a module-load capability object would add a concept for negligible gain at one call site — and module-load side work is undesirable under the runtime's determinism constraints. Noted for completeness; no change.

---

### Bottom line

Six findings, one round, full consensus, zero unresolved. The lenses agreed the new encoder's *boundary* is sound (lowy: the receptacle cleanly hides genuine runtime volatility) and that its only real *structural* defect was the four-times-derived group size, now fixed in 4d41f1b6cb63ab973f37ff0a082e86346a7703ac. The remaining four observations (pre-existing encode/decode asymmetry, loud-throw failure mode, intentional dual implementation, minimal runtime dispatch) were each examined and consciously kept, with reasoning on record so the structure stays legible to the next maintainer.

srid:
## 👮 Code-police

This track scrutinized the new pure-JS base64 path that this PR introduces to replace `TextEncoder`/`Buffer` in the Workflow runtime. Over **3 review sweeps** across the rules, fact-check, and elegance passes, it surfaced **2 findings** — one correctness gap where the fallback is strictly less robust than what it replaces (and could re-trigger the very crash this PR exists to fix), and one readability nit in the encoder's inner loop. Both reached consensus and were applied.

|---|---|---|---|
| minor | Pure-JS base64 fallback throws on lone surrogates where Buffer/TextEncoder silently substituted U+FFFD | .apm/skills/be-review/be-review.workflow.js | — |
| nit | base64 inner loop recomputes the same boundary checks four times in one dense expression | .apm/skills/be-review/be-review.workflow.js | — |

File touched by both findings: `/home/srid/code/kolu/.worktrees/be-parallelize/.worktrees/be-review-1780508578-police/.apm/skills/be-review/be-review.workflow.js`

---

### rules / fact-check pass

**Pure-JS base64 fallback throws on lone surrogates where Buffer/TextEncoder silently substituted U+FFFD** *(minor)*

This is the load-bearing finding. The new fallback path uses `encodeURIComponent(str)`, which throws `URIError: URI malformed` on an unpaired surrogate (e.g. a lone `\uD800`). Crucially, **neither path it replaces threw**: Node's `Buffer.from(s, 'utf8')` and the old `new TextEncoder().encode(s)` both substitute the replacement char U+FFFD and succeed. So under the Workflow runtime — where there is no `Buffer`, so the pure path is taken — a comment body containing a lone surrogate would throw out of `toBase64`, uncaught, and crash the Report phase. That is the exact ReferenceError-class, crash-the-phase failure this PR was created to eliminate, just reached through a different trigger.

The fact-check also caught a documentation overclaim: the comment asserts "byte-for-byte identical output" to the Buffer path, but that is **not true** for malformed UTF-16 input (verified: Buffer yields `77+9`, while `encodeURIComponent` throws). Because model-authored markdown and the deterministic baseline are valid UTF-8 in practice, the trigger is near-zero in normal operation — hence minor, not blocking — but the fallback is strictly less robust than what it replaces on the same path.

- **Prescribed fix:** Sanitize lone surrogates before encoding so the pure path can never throw, matching Buffer's U+FFFD-substitution semantics — e.g. `str.replace(/\p{Surrogate}/gu, '�')` (the `u` flag) applied to `str` before `encodeURIComponent`. Then drop or qualify the "byte-for-byte identical" claim to "for well-formed UTF-16 input."
- **What was applied:** Sanitized lone surrogates with U+FFFD before `encodeURIComponent` in `toBase64`'s pure path so it matches Buffer semantics and can't throw `URIError` out of the Report phase; qualified the comment's byte-for-byte claim to well-formed UTF-16.

---

### elegance pass

**base64 inner loop recomputes the same boundary checks four times in one dense expression** *(nit)*

The output line packs the full base64 quartet plus padding into a single expression that recomputes `i + 1 < bin.length` twice and `i + 2 < bin.length` twice — once to choose the byte value, once to choose between a char and `'='`. It works and is byte-for-byte correct, but a reader has to mentally re-derive the padding rule from four inline ternaries. This is the one spot in an otherwise idiomatic encoder where density hurts readability.

- **Prescribed fix:** Hoist the two boundary predicates into named locals and reuse them — `const hasB = i + 1 < bin.length; const hasC = i + 2 < bin.length;` (with `b`/`c` defaulting to `0` when absent), then build the quartet using `hasB`/`hasC` for the padding choice. Same output, the padding rule reads once.
- **What was applied:** Hoisted the base64 boundary predicates (`hasB`/`hasC`) into named locals and reused them, so the padding rule reads once instead of recomputing the two checks four times.

---

**Outcome:** consensus across all 3 sweeps; both findings applied to `be-review.workflow.js`. The minor finding is the one worth remembering — it closes a real residual crash path in the Workflow runtime, the exact failure class this PR targets.

Issue #1123, opened 2026-06-02
running_background fan-out not detected during live workflow runs — journal moved to subagents/workflows/<runId>/journal.jsonl


srid:
## Symptom

While a dynamic `Workflow` is actively running, the Dock fails to show a stable `running_background` fan-out card for the orchestrating session. In practice the workflow-specific "running" status is missing for most of a run.

## Reproduced live (chrome-devtools, random-port dev server, real claude)

Drove a fresh `claude` in a kolu terminal and prompted it to launch a background `Workflow` of **3 parallel agents each running `sleep 150`**. The agent launched it (`Run ID wf_556fab34-0d1`) and yielded (`end_turn`) — the canonical `running_background` scenario. Observed the Dock's `data-agent-state` across the run:

| time | since launch | `data-agent-state` | spinner | workflow actually running? |
| --- | --- | --- | --- | --- |
| 10:05:44 | ~95s | `running_background` | ✓ | yes (3 agents `started`) |
| 10:06:35 | ~145s | **`waiting`** (idle) | ✗ | **yes — agents still sleeping, journal had only `started`, no `result`** |

So it promotes **briefly**, then **demotes to idle mid-run**. Two distinct defects, both from journal-layout drift:

### 1. Premature demote mid-run (the visible bug)

`liveOutstandingTasks` reads `readWorkflowJournal(<session>/workflows/<runId>.json)` → the file **doesn't exist during the run** → `unreadable` → it falls back to the **`<session>/workflows/` directory mtime**. That dir is bumped only **at launch** (the runtime creates `<session>/workflows/scripts/` once) and **never again during the run** — live writes go to `subagents/workflows/<runId>/`. So at `launch + WORKFLOW_JOURNAL_STALE_MS` (2 min) the dir goes stale → the task is dropped → `deriveState` demotes `running_background` → `waiting`. Reproduced: `<session>/workflows/` dir age was 126s (stale) exactly when the Dock flipped to idle, while the workflow's 3 agents were still sleeping.

### 2. Fan-out details never populate

`deriveWorkflowProgress` also reads `<session>/workflows/<runId>.json`, absent during the run, so the workflow **name / agentCount are null** even while `running_background` is (briefly) shown. The fan-out count never appears.

### Live on-disk evidence

```
<session>/workflows/                      → only scripts/  (NO wf_556fab34-0d1.json during the run)
<session>/workflows/  dir mtime           → bumped once at launch, then stale
[...]

PR #1157, opened 2026-06-03
Detect a running /fork while the main agent is idle

srid:
**The Dock now shows the working pip when a `/fork` is running under an otherwise-idle main agent.** Previously a forked sub-agent was invisible: the main row sat on the idle dot even while "1 local agent still running", because the fork's launch never reached the background-task machinery that promotes an idle turn to `running_background`.

### Why it was invisible

A `/fork` spawns a background sub-agent the way a `Workflow` does, but it's launched from a slash command — so its launch lands in the transcript **only** as a `system`/`local_command` echo:

```
<local-command-stdout>⑂ forked implement-it (1483)</local-command-stdout>
```

`outstandingBackgroundTasks` scans `user`/`tool_result` blocks for one of three launch phrasings; this is none of them. So `launched − completed` never contains the fork, `deriveState` keeps the `end_turn` at `waiting`, and the row reads idle. *Confirmed against a live transcript: the fork's task-id appeared only in the completion enqueue, never in a launch marker.*

### How it's detected now

The fork leaves on-disk artifacts that mirror the Workflow-journal path exactly — so detection reuses that shape rather than inventing a new one:

```
subagents/agent-<id>.meta.json   { "agentType": "fork", ... }   ← discriminator
subagents/agent-<id>.jsonl       streaming transcript           ← liveness anchor (mtime)
main transcript queue-operation  <task-id> + terminal <status>  ← fast finish signal
```

An idle (`waiting`) main is promoted to `running_background` when a live fork exists. A fork is *live* when it's tagged `agentType:"fork"`, not in the transcript's completed set, and its transcript was written within `FORK_TRANSCRIPT_STALE_MS` (the orphaned-by-restart phantom guard).

| Concern | Resolution |
| --- | --- |
| Fork finished | id in `completedBackgroundTaskIds` (extracted from `outstandingBackgroundTasks` — one "which runs finished" projection, no drift) → demote instantly |
| Fork orphaned (claude killed, no completion) | transcript mtime past the stale window → demote; the recheck timer arms at the soonest stale deadline so the spinner self-clears with no fs event |
| Non-fork async `Agent`/`Task` | wrong/absent `agentType` → not promoted (unchanged — its launch marker outlives the process) |
| Cost on a busy session | the `subagents/` readdir is gated on the idle (`waiting`) case only |

> _A fork carries no fan-out journal, so the `workflow` field stays null — `running_background` is already a working state, so the Dock shows the working pip with **no client change**._

### Refined during review (codex ⇄ claude · lowy ⇄ hickey · code-police)

The parallel gauntlet (see the per-track comments) hardened two things beyond the original cut:

- **A write-ordering race (codex, major).** A fork's `subagents/` artifacts can land *after* the main transcript event that idled the main was already processed — at which point nothing would re-trigger the scan. Fixed with a per-session **`subagents/` directory watcher** (and `watchOrWaitForDir` now recurses to the nearest existing ancestor), plus a **watcher-level integration test** for the idle-first → fork-later → promote → complete → demote lifecycle.
- **Two parallel run-kinds collapsed onto one receptacle (lens).** Workflow and fork liveness now both project to a shared `LiveRun { id, anchorMs, staleMs }` folded through a single `nextStaleDeadline`, and the per-pass `now` + `completedBackgroundTaskIds` set are each computed once and threaded — so the fork and workflow staleness math can't drift.

### Tests

16 unit cases in `fork-detection.test.ts` (live / finished / orphaned / non-fork / unobservable / malformed-meta, the `workflows/` subdir skip, the injected staleness boundary), the watcher-level lifecycle test, and an e2e scenario (`claude-code.feature`) asserting the tile flips to `running_background` / the `working` pip.


srid:

This track hardens the Claude Code session watcher's background-task accounting — folding the per-pass `now` clock and the `completedBackgroundTaskIds` set into single hoisted values, threading them through `outstandingBackgroundTasks`/`outstandingForkRuns`, and adding a per-session `subagents/` directory watcher that closes a fork-promotion write-ordering race. All three reviewers (codex, lens, police) converged on overlapping fixes in the same two files, so consolidation was as much about *not double-applying* the same change as it was about landing new code.

Of the 9 picks: **6 landed clean**, **2 needed reconciliation**, and **2 were dropped as fully subsumed**. Every drop and every reconciliation below is explained — none was silent.

#### Clean picks (6)

These cherry-picked onto the branch with no conflict and no overlap requiring judgment.

|---|---|---|
| codex | 0b68898e | cb9268e37 |
| lens | e729334b | 97795fe6 |
| lens | 0f33a91b | 0bd9a315f |
| lens | 25cf391f | fe7c7e728 |

(codex contributed 1, lens contributed 3.)

#### Reconciled picks (2) — overlap resolved by honoring both intents

**lens 09f68ffb → 11b1788f9** — *the `completed`-set threading collided with codex's shared-clock hoist.*
This lens commit threads a precomputed `completed` set into `outstandingBackgroundTasks(lines, completed)`. The branch HEAD (already carrying codex's work) had independently changed the *same* `liveOutstandingTasks` argument from `Date.now()` to the single shared `now`. `core.ts` auto-merged cleanly — the optional `completed` param landed without issue. The conflict was a single line, and I honored **both** fixes there: I kept lens's `outstandingBackgroundTasks(lines, completed)` threading **and** HEAD's superset single-clock `now` (rather than reverting to lens's local `Date.now()`). Net result is strictly the union of the two improvements.

**police 241c3919 → 5bfe8da4d** — *a comment-accuracy fix landing in a region a prior overlap had quietly corrupted.*
This is the subtle one. Both codex (cb9268e3) and lens 25cf391f independently hoisted `const now = Date.now()` to the top of the pass and removed the old mid-pass declaration. Because codex landed first and had already removed the mid-pass `now`, lens 25cf391f's cherry-pick applied its ADD hunk but found nothing to remove — leaving a **duplicate** `const now = Date.now()` at lines 306 and 333, a latent TypeScript redeclaration error sitting on HEAD. Police 241c3919 separately wanted to re-add/narrow a mid-pass `now` comment.

I reconciled by **deleting the duplicate mid-pass `now` declaration and its comment entirely** (honoring both codex's and lens's single-clock intent) and **dropping police's re-added `now`**, leaving one canonical `now` at the top of the pass. Police's underlying concern — that the comment must not over-claim — is fully satisfied: in the final code `liveOutstandingTasks` *is* called with that single top-of-pass `now`, so the comment is accurate. Verified with `tsc --noEmit`, which now passes and would have failed on the duplicate.

#### Dropped picks (2) — fully subsumed, nothing substantive lost

Both drops were byte-identical reimplementations of fixes already on the branch via lens. I aborted each cherry-pick rather than create a no-op commit, and confirmed in each case that no substantive code was being left behind.

**police 81682b8c → fe7c7e728 (dropped)** — *subsumed by lens 09f68ffb.*
Both commits make the byte-identical change: give `outstandingBackgroundTasks` an optional `completed: Set<string> = completedBackgroundTaskIds(lines)` param, hoist `const completed = completedBackgroundTaskIds(lines)` once per pass, and thread it into both `outstandingBackgroundTasks` and `outstandingForkRuns`. The only differences in the conflict were doc-comment wording, and HEAD additionally carries the single-`now`-clock improvement that police lacked. No fix here was absent from the branch.

**police 438c4af6 → fe7c7e728 (dropped)** — *subsumed by lens e729334b.*
Both independently add the identical per-session `subagents/` directory watcher to fix the fork-promotion write-ordering race: same `subagentsDirFor` import, same `subagentsDirWatcher` variable, same `setupSubagentsWatching()` body (`watchOrWaitForDir` → `scheduleTranscriptCheck`), same call site, same destroy/teardown — all auto-merged from HEAD. The only conflicting hunks were doc-comment wording, where HEAD's is at least as descriptive. Nothing substantive was lost.

#### Files touched by reconciliation/drops

- `packages/integrations/claude-code/src/session-watcher.ts` — the focal point of every overlap (the `now` clock, the `completed` set, the subagents watcher).
- `packages/integrations/claude-code/src/core.ts` — the optional `completed` param landed here cleanly via auto-merge.

#### Bottom line

The branch now carries codex's single shared-clock hoist, lens's `completed`-set threading and `subagents/` watcher, and police's comment-accuracy intent — with the latent duplicate-`now` TypeScript error that the codex⇄lens overlap had introduced now removed and the build green under `tsc --noEmit`. The two police drops were genuine duplicates of lens work, not regressions.

srid:
## 🤖 Codex ⇄ Claude debate

This track hardens fork-run state derivation in the Claude Code session watcher so that a `/fork` whose on-disk artifacts land *after* the main transcript has gone idle still gets promoted to `running_background` (and later demoted on completion). The debate converged on three fixes: a per-session `subagents/` watcher with a recursive missing-parent fallback, a watcher-level integration test for the idle-first/fork-later lifecycle, and a single hoisted `now` sample threaded through every staleness/decay decision in one derivation pass. Two rounds, zero rebuttals — codex flagged three issues, Claude fixed all three, and codex re-verified each against the diff.

**Files touched:** `core.ts`, `session-watcher.ts`, `fork-detection.test.ts` · **Base:** 5592bd477a2afc03d0c0147971624a604e037815

---

### Round 1 — codex approved ❌ · 3 open · cb9268e37

Codex's read: the helper-level fork scan (`outstandingForkRuns`) was reasonable, but the **integration path was still event-racy**. Detection only fired inside `onTranscriptMaybeChanged()`, yet the artifacts it keys off (`subagents/agent-<id>.meta.json`, `agent-<id>.jsonl`) live in a separate on-disk subtree that nothing watched. So the exact transition this change exists to surface could be missed. Codex paired that structural finding with a coverage gap (tests hit the helper, not the watcher's eventing) and a small timing/comment inconsistency.

**Findings by severity:**

| Sev | ID | Location | The problem | Codex's fix |
|-----|----|----------|-------------|-------------|
| 🟠 major | F1 | `session-watcher.ts:341` | Fork detection runs only during `onTranscriptMaybeChanged()`. If the main `/fork` transcript event is processed *before* the `subagents/` meta/jsonl artifacts exist, `forks` is empty, no stale timer arms, and nothing watches `subagents/` to rescan. The session stays `waiting` until an unrelated main-transcript change happens to fire — so fork-start detection is unreliable. | Watch or wait for `subagentsDirFor(session)` and schedule the same debounced check on changes (or a bounded retry after the `/fork` local-command artifact). Add a regression test where the transcript is written first and fork meta/transcript appear later without touching the main transcript. |
| 🟡 minor | F2 | `fork-detection.test.ts:124` | New tests exercise `outstandingForkRuns` directly but never cover the changed `SessionWatcher` behavior that publishes `running_background`, schedules rechecks, and demotes on completion — the main user-visible path and eventing stay untested. | Add a `createSessionWatcher`-level unit test (or a Claude Code e2e mock state) for a live fork with delayed artifact creation + completion, asserting the emitted `ClaudeCodeInfo.state` transitions. |
| ⚪ nit | F3 | `session-watcher.ts:313` | `liveOutstandingTasks` still receives a fresh `Date.now()` *before* the shared `now` is read, while the new comment claims the pass uses one clock sample for all staleness comparisons. A tiny boundary inconsistency that makes the comment inaccurate. | Hoist `const now = Date.now()` above `liveOutstandingTasks(...)` and pass it to workflow filtering, fork scanning, deadlines, and transient decay. |

**Claude's response — all three accepted and fixed (none disputed):**

| Finding | Disposition | What Claude did |
|---------|-------------|-----------------|
| F1 | fixed | Installed a `subagents/` watcher plus a recursive `watchOrWaitForDir` that walks up to the nearest existing ancestor |
| F2 | fixed | Added the watcher integration test |
| F3 | fixed | Hoisted the single `now` sample |

Claude reported 156 tests passing and marked the round done. Commit: cb9268e37961323414bb6553a6244abafa9ec994

---

Codex re-reviewed the follow-up diff and confirmed every fix landed. The watcher now installs a per-session `subagents/` watcher; `watchOrWaitForDir` recurses up missing parent directories so late-created fork artifacts are observed even when their parent dirs are created lazily; the new integration test drives the full idle-first → fork-later → completion lifecycle; and the state pass uses one hoisted `now`. Codex notes it reviewed the diff and changed files but did **not** run the unit suite in this read-only pass.

**Codex on Claude's rebuttal:** No disputes to answer — Claude accepted and fixed F1, F2, and F3, and codex verified those fixes against the current diff.

**Findings by severity — all resolved:**

| Sev | ID | Location | Verification |
|-----|----|----------|--------------|
| 🟠 major | F1 | `session-watcher.ts:556` | The per-session `subagents/` watcher + recursive `watchOrWaitForDir` (walks to the nearest existing ancestor) adequately address the race; late-appearing `/fork` artifacts now retrigger state derivation and the missing-parent fallback no longer misses lazily created session/subagents dirs. ✅ resolved |
| 🟡 minor | F2 | `fork-detection.test.ts:321` | The new test starts from an idle transcript, creates fork artifacts while the watcher is already running, verifies promotion to `running_background`, then verifies demotion on completion — covering the user-visible eventing path, not just the helper scan. ✅ resolved |
| 🟡 minor | F3 | `session-watcher.ts:301` | A single hoisted `now` is threaded through workflow liveness, fork liveness, stale-deadline computation, transcript quiet time, and transient decay — the shared clock sample resolves the prior inconsistency. ✅ resolved |

**Net:** clean two-round convergence with no concessions or rebuttals to adjudicate. The one structural risk (F1, the unwatched `subagents/` subtree) was the substantive find; F2 closed the coverage gap it exposed, and F3 was a cosmetic clock/comment cleanup. Note: codex's approval is a read-only diff review — it did not independently run the unit suite, though Claude reported 156 passing in round 1.

srid:

**Outcome:** `consensus` after 1 round(s). Independent review: lowy=6, hickey=6.

This PR adds `/fork` as a *second* way to drive Claude Code's `running_background` state — alongside the pre-existing Workflow path. Both lenses, working independently, converged on the same headline: the diff models that second run-kind as a **fully parallel pipeline** (its own struct, detection fn, stale-window const, and deadline fn) that only reunites with the workflow pipeline when the watcher hand-folds them at read time. Four fixes landed to collapse that duplication onto one `LiveRun` shape and one deadline fold; five observations were dropped — two as correct-by-design seams, three where the proposed unification would have *complected* genuinely-distinct strategies.

---

### The two lenses agreed on the same structural defect from opposite directions

The most striking result: **lowy-1 and hickey-1 are the same finding** reached through different frames, and they converged on an *identical* plan.

- **Lowy (volatility lens):** The receptacle every consumer plugs into is `state === "running_background"` — "the main is busy-waiting on an observable background run it launched." The *kinds* of background run that produce that signal (Workflow was kind #1; `/fork` is now kind #2) are a genuine, recurring axis of change. But the axis is encapsulated **twice in parallel**, and the watcher is forced to expose the wires: it calls both detection functions, computes both stale deadlines, then hand-composes `[nextWorkflowStaleDeadline(...), nextForkStaleDeadline(...)].filter(...).Math.min`. A third kind means editing the watcher's promotion block *again* — the blast radius of the axis is the watcher itself, the functional-decomposition signature.
- **Hickey (simplicity lens):** The domain has ONE concept — "an outstanding background run the idle main is busy-waiting on, with a liveness anchor that ages out" — shattered into two pipelines that never meet until the watcher reconstitutes the unity. The fingerprint is the Layer-2 reconciliation-machinery shape: the `Math.min` over two deadline arrays plus the disjunction `derived.state === "running_background" || forks.length > 0`. The merge logic is *unbounded in the number of run kinds* because the kinds were never unified into one "live run" value.

**Agreed plan (one fix discharging lowy-1 + lowy-2 + hickey-1 + hickey-2):** introduce `LiveRun { id; anchorMs; staleMs }` and one `nextStaleDeadline(runs, now)` in core.ts. Both producers emit the common shape; the watcher becomes `const live = [...workflowRuns, ...forkRuns]; if (live.length) { publishedState = "running_background"; staleDeadline = nextStaleDeadline(live, now); }`. A future third kind is one more producer feeding the same list.

### A live scope negotiation: unify the *shape*, keep the *readers* private

The interesting tension inside the consensus: lowy set `agreesWithPlan: false` on lowy-1 / hickey-1 even while endorsing them — to **narrow** the plan and stop it over-reaching. Lowy's own original phrasing imagined a single `outstandingBackgroundRuns` reader that "folds the workflow path and the fork path"; lowy walked that back. The two producers' anchor reads are genuinely distinct IO — workflow resolves `anchorMs` by walking a whole live-run dir (`newestFileMtimeMs`, core.ts:692-709), fork stats one `agent-<id>.jsonl` (core.ts:1011). Forcing both behind one observation-shaped reader is the mode-flag trap. **Resolution: unify the output shape and the deadline fold; keep each producer's anchor-reading private and unmerged.** Hickey verified the IO distinction independently and amended their own fuller plan down to this scope — that boundary is precisely what makes hickey-4 and hickey-5 *drop* below.

---

### Findings that FLIPPED — three hickey fixes argued down to `drop`

The debate's signal isn't just agreement; it's three of hickey's `fix`-leaning findings that lowy talked down to `drop`, with hickey conceding each on the code.

**hickey-4 — share the liveness gate (fix → drop).** Hickey wanted `outstandingForkRuns` to feed the *same* three-part keep/drop predicate `liveOutstandingTasks` uses (`isLive({terminal, anchorMs}, now, staleMs)`). Lowy's rebuttal, conceded by hickey: the predicate is only ~3 comparisons, while the *body* of `outstandingForkRuns` is dominated by fork-specific work no shared predicate can absorb — `readdir` enumeration, `FORK_META_RE` classification, and an explicit cost-ordering that stats the transcript *before* reading meta (so a finished/stale fork costs only a stat). Worse, the "finished" signals genuinely differ: fork uses `completed.has(id)` (a transcript-derived Set) while workflow uses `obs.terminal` (a journal-read boolean) — a shared predicate would need a contrived adapter complecting two distinct notions of "finished." The residual drift risk is closed once both paths emit `LiveRun`, and the filter is well-covered (fork-detection.test.ts exercises all three drop reasons).

**hickey-5 — share the mtime read (fix → drop).** Hickey wanted the fork transcript stat routed through the workflow's `newestFileMtimeMs` (or a shared `subagentTranscriptMtime`/single enumerator). Lowy's rebuttal, conceded: the two reads encode **different liveness facts**. `newestFileMtimeMs` takes the newest mtime *across a whole dir* — correct for a workflow run (journal + N streaming agent files). The fork stats *one specific transcript* — correct for a fork. Pointing the fork at `newestFileMtimeMs` over the flat `subagents/` root (shared with unrelated sub-agents and the `workflows/` subtree) would be *wrong*. A `subagentTranscriptMtime(dir, id)` helper would have exactly one caller (a single-consumer helper shaped around one implementation), and the bigger "one enumerator" forces two divergent walks into one — the mode-flag trap. The shared-path concern is already encapsulated by `subagentsDirFor` (lowy-5); the mtime *strategy* legitimately differs per run kind.

**hickey-6 — de-duplicate the rationale comment (drop, affirmed).** The "a `/fork` launch is a `local_command` echo, not a `tool_result`" rationale appears in four places (core.ts section header, schema docstring, watcher inline comment, test-file header). Both lenses agreed this is *documentation* duplication, not a structural/volatility-axis defect — the four sites serve four different reader contexts, and a stale comment costs a doc lie, not a wrong promotion. Below the bar for this PR; optional tidy at most.

---

### The other agreed fixes (subsumed into the lowy-1 unification or standalone)

| # | Finding | Both lenses' take | Plan |
|---|---|---|---|
| lowy-2 / hickey-2 | `FORK_TRANSCRIPT_STALE_MS` + `nextForkStaleDeadline` are line-for-line clones of the workflow const + `nextWorkflowStaleDeadline` | **variable-vs-volatile trap**: the docstring's "kept separate so fork timing can move independently" is speculative — both encode the *same* fact ("how long an anchor may go quiet before a streaming run is presumed orphaned"). The load-bearing defect is the duplicated **fold**, not the two consts. | Delete both `next*StaleDeadline`; the two stale windows survive as per-run `staleMs` *data* on each `LiveRun` (data, not duplicated algorithm); the fold lives in one `nextStaleDeadline`. |
| lowy-3 / hickey-3 | `completedBackgroundTaskIds(lines)` JSON-parsed over the full tail **twice per pass** — once inside `outstandingBackgroundTasks` (line 315), again at line 343 for the fork path | Breaks the file's own observe-once memoization discipline (the `obs` Map at session-watcher.ts:299-307, with an explicit "don't walk disk 3×" comment). The completion projection is named in its own docstring as the shared source — and is read twice. | Compute `const completed = completedBackgroundTaskIds(lines)` once near the top; add optional `completed?: Set<string>` to `outstandingBackgroundTasks` (default: scan internally); thread into both consumers. |
| lowy-4 | Fork promotion reads a **different clock** than workflow liveness in the same pass | The new `const now = Date.now()` (line 331) carries a comment promising "one clock read... so all staleness comparisons test against a single now" — but `liveOutstandingTasks` (line 316) already ran with its *own* fresh `Date.now()`, before `now` is bound. The comment is literally false; skew is microseconds and harmless, but it contradicts a stated invariant. | Hoist `const now` above the `liveOutstandingTasks` call and pass it at line 316. One-line fix that makes the comment honest; folds into the lowy-1 unification (one `now` measures every run). |

### Dropped as correct-by-design

| # | Finding | Verdict |
|---|---|---|
| lowy-5 | `subagentsDirFor(session)` extraction routes both `liveWorkflowsRootFor` and the fork scan through one anchor (core.ts:643-649) | **Positive finding** — the *right* receptacle for the per-session subagents-layout axis (which churned once already, #1123). The structural issue is one altitude up (the run-kind axis), not here. The lowy-1 reader should keep routing through this helper. |
| lowy-6 | `ClaudeCodeInfo.workflow` stays `null` for forks — `deriveWorkflowProgress` runs for fork-driven promotions but returns null (a fork has no journal) | **Observation under interface-stability.** `running_background` now means two structurally different things, with the discriminator leaking via `workflow === null`. Defensible today (exactly one optional payload, client needs no change); becomes a real problem only at a third kind with its own payload — at which point the `LiveRun` fold would naturally carry a per-run-kind payload. No change this PR. |

---

### Applied to the branch

Four commits landed; the three hickey `fix` findings (hickey-1/2/3) carry no separate commit because they were **discharged by the lowy-1 unification** they converged onto:

| commit | finding |
|---|---|
| 97795fe6 | lowy-1 — unified `LiveRun` receptacle (core.ts, session-watcher.ts, index.ts, + tests) |
| 0bd9a315f06918e37bea9f20ddb2b8b997d8c58d | lowy-2 — collapsed duplicated stale-window deadline fold |
| 11b1788f915310a49cde7c9e847b8ab0e867bf35 | lowy-3 — thread `completed` set, one scan per pass |
| fe7c7e728a67eaa6780876c746a1981630302ace | lowy-4 — single `now` across the pass |

No unresolved findings; consensus reached in a single round across all 12 (6 per lens).

srid:
## 👮 Code-police

The police track ran two review sweeps (rules → fact-check → elegance) and converged on **5 findings, all reconciled to consensus**. The headline outcome: one *major* correctness fix to fork promotion, one *nit/minor* duplicate-scan that was hoisted into a single pass, and one stale comment narrowed to match reality. Two of the five findings were re-discoveries of work already committed earlier in this worktree, so they needed no further change — the duplicate detection across passes is itself a signal the convergence held.

| severity | pass | finding | disposition | commit |
|---|---|---|---|---|
| major | fact-check | A just-launched /fork can fail to promote an idle main, with no recheck armed to recover | fixed — subagents/ dir now watched | fe7c7e728 |
| nit | rules | completedBackgroundTaskIds(lines) scanned twice per check pass | fixed — single hoisted scan | fe7c7e728 |
| nit | fact-check | completedBackgroundTaskIds scans the full tail twice per check pass | already fixed earlier (81682b8c) | — |
| minor | elegance | completedBackgroundTaskIds(lines) scanned twice per pass in the waiting branch | already fixed in committed code | — |
| nit | elegance | Comment overstates that `now` unifies all staleness comparisons in the pass | fixed — comment narrowed | 5bfe8da4d |

---

### fact-check pass

**[major] A just-launched /fork can fail to promote an idle main, with no recheck armed to recover** — commit fe7c7e728

*Problem.* Fork promotion only runs inside `onTranscriptMaybeChanged`, which is re-triggered by exactly three sources: the main transcript `fs.watch` on `<session>.jsonl`, the `workflows/` snapshot-dir watcher, and the stale-recheck timer (armed only when `staleDeadline !== null`). The fork's own directory — `<session>/subagents/` and its `agent-<id>.jsonl` — is never watched. When a `/fork` launches, the main transcript receives the `⑂ forked` local-command echo, firing exactly one check. If at that instant the fork's `agent-<id>.jsonl` (or its meta) is not yet on disk, `outstandingForkRuns` returns `[]`, so `derived.state` stays `waiting`, `forks.length === 0`, and the else branch runs `decayTransientState('waiting', …)` which returns `{ recheckAt: null }`. `scheduleStaleRecheck(null)` then arms **no** timer. The idle main sits at `waiting` until the next unrelated main-transcript fs event — but the entire premise of the feature is that the main is quiet and busy-waiting on the fork, so that event may never come. Promotion silently never fires — the exact bug this PR set out to fix, now made dependent on Claude Code's write-ordering race between the echo and the subagent transcript creation.

*Prescribed fix.* Either (a) watch the `subagents/` directory the same way `workflows/` is watched (`setupWorkflowsWatching → watchOrWaitForDir(subagentsDirFor(session), () => scheduleTranscriptCheck(), plog)`) so the fork's first transcript write re-derives state; or (b) when `derived.state === 'waiting'` and `forks.length === 0`, arm a short bounded recheck so a fork materializing just after the launch echo is picked up on the next probe. Option (a) mirrors the existing workflows watcher and is more robust.

*Applied.* Watch the per-session `subagents/` dir (mirroring the `workflows/` watcher) so a just-launched `/fork`'s first on-disk write re-derives state and promotes an idle main — fixing the echo-vs-transcript write-ordering race that left promotion never firing.

**[nit] completedBackgroundTaskIds scans the full tail twice per check pass** — no change required

*Problem.* `outstandingBackgroundTasks(lines)` (called via `liveOutstandingTasks`) already runs `completedBackgroundTaskIds(lines)` internally (core.ts:587), parsing every tail line as JSON; line 343 then calls `completedBackgroundTaskIds(lines)` a second time on the same lines. The completed set computed for the outstanding path was discarded rather than threaded through, so every pass did two full `JSON.parse` walks over the tail window where one would do.

*Resolution.* No change needed — this finding was already fully fixed by an earlier commit in this worktree (81682b8c); the double scan no longer exists. (This overlaps with the rules-pass nit below, which carried the fix.)

---

### rules pass

**[nit] completedBackgroundTaskIds(lines) is scanned twice per check pass** — commit fe7c7e728

*Problem.* In `onTranscriptMaybeChanged`, `liveOutstandingTasks(outstandingBackgroundTasks(lines))` at line 315 already calls `completedBackgroundTaskIds(lines)` internally to build the outstanding set, and line 343 then calls it a second time on the same tail buffer to feed `outstandingForkRuns` — a redundant full re-scan + re-parse of every tail line plus a second `Set` allocation per `(waiting)` handler run. It is bounded (tail capped at `TAIL_BYTES`, gated to `derived.state === 'waiting' , debounced 150 ms), so it is not a leak — just wasted work that mildly complects the fork path with re-deriving a projection the outstanding path already produced and discarded. Borderline against the dry-rule-of-three: the "which runs finished" projection is now materialized at two call sites in one pass.

*Prescribed fix.* Hoist a single `const completed = completedBackgroundTaskIds(lines);` near the top of `onTranscriptMaybeChanged` (before the outstanding computation), thread it into the fork scan (`outstandingForkRuns(session, completed, now)`), and — if acceptable — accept an optional precomputed `completed` set in `outstandingBackgroundTasks` so the single scan feeds both consumers.

*Applied.* Hoisted a single `completedBackgroundTaskIds(lines)` scan in `onTranscriptMaybeChanged` and threaded it into both the outstanding and fork paths, eliminating the redundant per-pass re-scan.

---

### elegance pass

**[minor] completedBackgroundTaskIds(lines) scanned twice per pass in the waiting branch** — no change required

*Problem.* In the `derived.state === "waiting"` path, `completedBackgroundTaskIds(lines)` runs at line 343, but the same projection was already computed earlier in the pass: line 315's `outstandingBackgroundTasks(lines)` (inside `liveOutstandingTasks`) internally calls `completedBackgroundTaskIds(lines)` (core.ts:587). So the full tail (up to `TAIL_BYTES` of JSON, parsed line-by-line) is scanned for terminal-status enqueues twice per pass. This directly contradicts the function's own documented discipline a few lines above (lines 294–298: observe each `runId` ONCE per pass and memoize, precisely to avoid re-walking the same data). The same input yields the same `Set` both times.

*Prescribed fix.* Compute the completed set once at the top of the pass and thread it into both consumers — pass it to `outstandingForkRuns(session, completed, now)`, and add an optional `completed?: Set<string>` parameter to `outstandingBackgroundTasks` (defaulting to `completedBackgroundTaskIds(lines)`) so line 315 reuses it. This mirrors the `observe`-memoization pattern the watcher already establishes.

*Resolution.* No change needed — already fully fixed in the committed code: the completed set is computed once at session-watcher.ts:320 and threaded into both `outstandingBackgroundTasks` and `outstandingForkRuns`, and `outstandingBackgroundTasks` already carries the optional `completed` param defaulting to `completedBackgroundTaskIds(lines)`.

**[nit] Comment overstates that the new `now` unifies all staleness comparisons in the pass** — commit 5bfe8da4d

*Problem.* The comment claimed `now` is "One clock read shared by the fork scan, the stale deadlines, and the transient-decay quiet window below — so all the staleness comparisons in this pass test against a single `now`." But `liveOutstandingTasks(...)` at line 316 (unchanged by this diff) performs its own `Date.now()` staleness comparison for workflow runs, which is **not** the `now` declared here. So not all staleness comparisons in the pass share this clock read; the workflow-liveness gate uses a separate, slightly-earlier sample.

*Prescribed fix.* Either narrow the comment to the comparisons that actually share `now` (the fork scan, the two `next*StaleDeadline` calls, and the transient-decay window), or hoist `now` above line 313 and pass it into `liveOutstandingTasks` at line 316 so the claim becomes true.

*Applied.* Narrowed the `now` comment so it no longer claims to unify all staleness comparisons in the pass — it now excludes the workflow-liveness gate in `liveOutstandingTasks`, which takes its own earlier clock sample.

---

### Files touched

- `packages/integrations/claude-code/src/session-watcher.ts` — subagents/ watcher, hoisted single scan, narrowed `now` comment
- `packages/integrations/claude-code/src/core.ts` — threading the precomputed `completed` set through `outstandingBackgroundTasks`

The three duplicate-scan findings (one per pass) all describe the *same* redundant `completedBackgroundTaskIds(lines)` re-walk; collectively they're resolved by the single hoisted scan in fe7c7e728, with the earlier 81682b8c already having landed equivalent work — which is why two of the three needed no further change.

srid:
## Evidence

A Claude Code `/fork` running under an **idle** main agent. The mocked session ends its turn (main idle) while the fork's on-disk subagent artifacts are present — the tile now flips to **working** (`running_background`), showing the spinning state pip, instead of reading idle.

The e2e scenario `A running /fork promotes the idle main to running-in-background` passes on the PR branch (`feat/fork-dock-detection`): it asserts both that the agent indicator state is `running_background` **and** that the tile title state pip is `working`.



Still at payoff moment (spinning working pip visible):



▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=fork-dock-working-pip.mp4

PR #1158, opened 2026-06-03
Make comment-surface a single declared decision per browse view

srid:
**This refactors how inline-comment surfaces get wired into the Code-tab browse views: each renderer now *declares* its comment address space at one seam, instead of each view wrapping (or forgetting) a surface itself.** It set out to *also* make the Markdown rendered preview commentable — but review found that unsafe, so **rendered Markdown stays read-only by design** and the structural cleanup that makes a correct future implementation a localized change is what ships.

### Why the Markdown-preview fix was pulled

Wrapping `MarkdownRenderer` in `CommentTextSurface` *looks* free — it's plain selectable DOM. But the comment anchor resolves against the selection's root node, and rendered Markdown is **light DOM with no shadow root** (unlike Pierre's `CodeView`, which lives in one). So the locator's prefix/suffix haystack and the highlight re-find fall back to the **entire document** (`useTextSelection.ts:199-201`), and the captured `lineRange` — a *rendered*-DOM line — gets persisted as a *source* line jump. The quote (`Hello Doc`) may not even exist verbatim in source (`# Hello Doc`). Net: comments that highlight the wrong text and jump to bogus lines. A correct rendered-Markdown annotation path (subtree-scoped anchor, no source `lineRange`) is its own feature — left as follow-up.

### What ships — commentability as one declared decision

`BrowseFileDispatcher` builds every browse renderer through `withComments(capture, file, view)`, where `capture` is an exhaustive `"text" | "iframe" | "none"`:

| Capture | Views | Surface |
| --- | --- | --- |
| `text` | source code, **Markdown source** | `CommentTextSurface` (native selection) |
| `iframe` | `.html` / `.svg` / `.pdf` preview | renderer owns its postMessage bridge (it must bind to the element it creates) |
| `none` | raster image, **Markdown rendered** | no surface — nothing source-faithful to anchor to |

A new renderer **can't compile** without picking a mode (the `ts-pattern` match is exhaustive), so the silent omission that left Markdown preview *looking* commentable can't recur. `BrowseFileView` and the new `BrowseDiffView` are now **pure presenters** with zero comment knowledge; the one real shared receptacle is `CommentTextSurface` itself, reached by both the dispatcher (source) and `CodeTab` (diff).

> _Bonus correctness fix from review: the "File truncated" banner moved **out** of the commentable host — you could previously anchor a comment to UI chrome the agent can't find in the file._

The comment **data model** (path-keyed W3C TextQuoteSelector `Locator`) was already surface-agnostic and is **untouched**.

### Reviewed by the parallel gauntlet

- **codex⇄claude** — pulled the Markdown wrap (the anchoring bug above), moved the truncation banner out.
- **lowy⇄hickey** — confirmed the two-seam shape is *correct*: both seams reach the single `CommentTextSurface` receptacle, so the volatility isn't actually duplicated; collapsing them would complect file-vs-diff behind a mode flag.
- **code-police** — exhaustive `ts-pattern` dispatch for the capture union, doc-comment freshness.

Per-track ledgers are in the comments below.


srid:

This /be-review run hardened a single seam — the comment-capture surface in the file browser — by reconciling codex's *behavioral* fixes against police's *documentation/lint* nits. Codex won every conflict because its three fixes (revert rendered Markdown to read-only `none`-capture, move the "File truncated" banner outside the capture host, and keep the README accurate) changed real behavior, while three of police's five commits documented or re-applied the pre-codex world. Net result: one clean pick, two reconciled (codex's behavior preserved, police's structural/lint intent grafted on), and two dropped (would have reintroduced a proven bug or documented a non-existent feature). Final head is af4de2aa373bd582574ae5520e6c42ef492ee04a.

**Outcome tally:** 5 picks total — 1 clean, 2 reconciled, 2 dropped.

#### Clean picks

| track | source | new commit | files |
|---|---|---|---|
| codex | 3406353 | 84d6c15b2 | (codex's F1–F3 behavioral fixes) |

Codex's commit is the load-bearing one for the whole consolidation: its F1 reverts the rendered Markdown preview to read-only `none`-capture, F2 moves the "File truncated" banner out of the comment-capture host, and F3 keeps the README's surface list accurate. Every reconciliation and drop below is decided against these three fixes.

#### Reconciled picks (overlap resolved in favor of codex's behavior)

**police 8c3dbd7 → af4de2aa3** — `packages/client/src/comments/useTextSelection.ts`

The cherry-pick text-merged cleanly (codex never touched this header), but the applied text was already stale relative to codex's F1. Police's edit claimed the `useTextSelection` hook mounts for "the source view via BrowseFileView + the Markdown rendered preview" — yet codex's F1 reverted rendered Markdown to `none`-capture, so the hook never mounts there. Resolution: honored police's *intent* (correct the stale "Mounted on:" list to point at the `withComments('text', …)` seam in `BrowseFileDispatcher`) while honoring codex's revert — the commit was amended to describe the source view via `BrowseFileView` and to note that rendered Markdown is `none`-captured and therefore never mounts this hook. This commit is the final head.

**police ce55196 → a1e25d781** — `packages/client/src/right-panel/BrowseFileDispatcher.tsx`

Both tracks touched the same dispatch site. Codex (HEAD) had rewritten `withComments` as a ternary with class `min-h-0 w-full flex-1` plus a comment tying the surface to the truncation-banner sibling. Police's commit converts that dispatch to a ts-pattern `match(capture).with('text').with(P.union('iframe','none')).exhaustive()` for `prefer-ts-pattern`. These are orthogonal (codex = styling/behavior, police = lint structure), so both were honored: kept codex's class and the banner-sibling comment, but applied police's exhaustive `match` structure (`match`/`P` were already imported, so no new dependency).

#### Dropped picks (subsumed or contradicted by codex)

**police fb8d3dd — DROPPED** — `packages/client/src/right-panel/BrowseFileDispatcher.tsx`

Directly contradicted and fully subsumed by codex's F2. Codex moved the "File truncated" banner *out* of the comment-capture host (rendering it as a sibling above the surface) and deleted the now-unused `truncated` prop from `BrowseFileView`, because a user could otherwise select the "File truncated" UI chrome and save a bogus comment. Police's commit is a fact-check nit that does the *opposite*: it passes `truncated={...}` back into `BrowseFileView` — a prop codex removed, so it would not even compile — and adds a comment asserting the banner now lives *inside* the capture surface intentionally. It documents the pre-codex layout; carrying it forward would reintroduce codex's proven bug plus a stale, wrong comment. Aborted, leaving codex's banner-outside arrangement intact.

**police 6a45d7e — DROPPED** — `README.md`

Fully subsumed/contradicted by codex's F1+F3. Police's README edit adds "rendered Markdown preview" to the commentable-surfaces list and states "Comments work in either view — select text in the rendered preview or the source," softening the source-toggle parenthetical by dropping "line-anchored comments." But codex reverted rendered Markdown to read-only `none`, so it is *not* commentable — comments work in the source view (line-anchored). The cherry-pick text-merged cleanly, yet the result documents a reverted, non-existent feature; per codex's F3 the original README (only "source file, branch diff, or rendered HTML artifact") was already correct. The branch was reset back to drop it, restoring codex's accurate README.

#### Takeaway

The consistent thread: police's correctness/doc commits were authored against the *pre-codex* state, so once codex's F1–F3 landed first, two of police's commits became stale documentation (dropped) and two needed grafting onto codex's behavior (reconciled). Behavior beat description in every conflict, and the surviving police contributions (the ts-pattern lint fix and the corrected `Mounted on:` comment) were salvaged without reintroducing the bug codex had just fixed.

srid:
## 🤖 Codex ⇄ Claude debate

This track refactored a diff-presenter seam and moved the text-comment surface up to the `BrowseFileDispatcher` level — but the move widened *what* was commentable in two unintended ways. Codex flagged both as anchoring hazards; Claude conceded the substance and narrowed the commentable host back to file content only, landing at consensus with no remaining open findings.

### How it converged

The debate was not a stalemate that got tired — it was a clean concede-and-fix. Codex opened by **rejecting** the branch (round 1: not approved, 3 open) on the grounds that the surface move enabled comments on DOM that the anchoring code (`useTextSelection`, `highlightOverlay`, `lineRangeForSelection`) was never designed to anchor against. Claude did not argue for keeping the new behavior; it agreed on all three and chose the *minimal correct* resolution — revert rendered Markdown to read-only and re-separate the truncation banner — which also happened to re-align the spec/README without doc edits. Round 2 codex re-reviewed the follow-up commit, verified each fix, found no new regression, and **approved**.

One point was genuinely *conceded* rather than fixed-as-stated: codex's F1 round-1 claim that app chrome could leak into the captured selection. Claude pushed back that selection capture is host-gated (`nodeInside(host, …)`), so the leak was never in *what the pill captures* but in the *haystack/offset roots and the persisted `lineRange`*. Codex accepted this nuance in round 2 — "I accept the F1 nuance that selection capture itself was host-gated; the real failure was the document-root haystack/offset and rendered-to-source line mapping" — while noting the implemented fix resolves the real defect either way.

---

### Round 1 — codex approved ❌ · 3 open · 84d6c15b292599462b37620f04634e12c609c469

> The extraction of the diff presenter is mostly mechanical, but the comment-surface move changes behavior in ways the current anchoring code is not designed to support. The branch enables comments on rendered Markdown and broadens the source-file capture host; both create invalid comment anchors. I would not merge this as-is.

**Codex findings (grouped by severity):**

| Severity | ID | Location | Issue |
|---|---|---|---|
| 🔴 blocking | F1 | `BrowseFileDispatcher.tsx:167` | Rendered Markdown now wrapped in `CommentTextSurface`, but that surface targets Pierre source/diff roots, not transformed inline Markdown DOM. For non-shadow selections, `useTextSelection`/`highlightOverlay` resolve against the owner `Document`, so locator prefix/suffix and highlight search can pull in whole-app chrome. The saved `lineRange` is derived from rendered DOM text, then reused as a *source-file* line jump → comments that highlight/scroll to wrong text and jump to bogus source lines. |
| 🟡 minor | F2 | `BrowseFileDispatcher.tsx:104` | Moving the wrapper outside `BrowseFileView` makes the truncation warning part of the commentable host. On truncated files, users can select `File truncated (exceeds 1 MB)` and create a comment whose quote is UI chrome — unlocatable by the agent and the re-anchor path. |
| 🟡 minor | F3 | `code-tab.feature:984` | Checked-in spec still says rendered Markdown is read-only and comments require switching to source, while the branch makes rendered Markdown commentable. No e2e covers select-in-preview → save → jump/highlight — the gap that would have exposed F1. |

**Codex suggestions:**
- **F1** — Keep rendered Markdown read-only, *or* implement the rendered-Markdown annotation path properly: anchor/highlight against the Markdown host subtree, don't persist a source `lineRange` unless mapped to source lines, and cover duplicate-text/jump behavior in tests.
- **F2** — Wrap only the actual content viewer. Restore the wrapper inside `BrowseFileView` around `CodeMenuFrame`, or split the truncation banner from the commentable node before wrapping.
- **F3** — Align spec/docs with intended behavior: ship rendered-Markdown comments *with* an e2e scenario + README/tips updates, or keep rendered Markdown unwrapped and source-only.

**Claude's response — all three `fixed`, with reasoning:**

- **F1 → fixed (with a partial concession back to codex).** Reverted the Markdown rendered renderer from `"text"` to `"none"` in `BrowseFileDispatcher.tsx:197` (`textRenderers`). Root cause: rendered Markdown is plain light-DOM (`MarkdownRenderer` in `packages/solid-fileview/src/renderers/markdown.tsx` has *no shadow root*, unlike Pierre's `CodeView`). So a Markdown selection makes `useTextSelection.ts:197-199` resolve `resolvedRoot` to `ownerDoc` (whole Document); `extractQuote.ts:43` slices prefix/suffix from `document.body.textContent`; `highlightOverlay.ts:42` `findHostRoot` falls back to `host.ownerDocument`, so `applyHighlights`/`findQuote` search the entire document; and `lineRangeForSelection` (`useTextSelection.ts:76`) derives a line number from rendered-DOM text that `CodeTab.tsx:734` feeds to `openInCodeTab` as a *source* startLine/endLine — a bogus jump (rendered lines ≠ source lines; `"Hello Doc"` vs `# Hello Doc`). **Nuance Claude raised:** selection *capture* cannot include app chrome because `useTextSelection.ts:142` gates every selection through `nodeInside(host, …)` — the leak is in haystack/offset roots and the `lineRange`, not in what the pill captures. Resolution is identical either way: rendered Markdown stays read-only (matching the plan's phase-3 v1 decision) until a real rendered-annotation path exists.

- **F2 → fixed.** Confirmed against pre-branch code (`git show 5592bd47:BrowseFileView.tsx`): there the `<Show when={truncated}>` banner was a *sibling* of `CommentTextSurface`, not inside it. This branch wrapped the whole `BrowseFileView` (banner + content) at the dispatcher seam, pulling the banner into the commentable host so `nodeInside(host)` returns true for it. Fixed by removing the banner from `BrowseFileView` (now renders just `CodeMenuFrame`/`CodeView`; dropped the unused `truncated` prop) and rendering it in the dispatcher's `sourceRenderer` as a sibling *above* the surface inside a flex-col, with the surface switched to `min-h-0 w-full flex-1`. Rendered-Markdown banner is unaffected (its own internal banner, now `"none"`-captured).

- **F3 → fixed (no doc edit needed).** With F1 reverting Markdown to read-only, the checked-in scenario at `code-tab.feature:984-997` ("Comments on a Markdown file work in the source view") is accurate again, and `README.md:79-80` already lists only source file / branch diff / rendered HTML artifact as commentable — never rendered Markdown. Realigning the code restored consistency; no new e2e was added because there is no rendered-Markdown comment behavior to cover.

Claude reported `just fmt` clean and `pnpm typecheck` passing (the 48 biome `noExplicitAny` warnings in `packages/surface/define.ts` are pre-existing and unrelated).

---

> All three prior findings are adequately resolved in this follow-up. Rendered Markdown is no longer wrapped in the text comment surface, the truncation banner has been moved outside the source comment host, and the existing Markdown source-view spec/README now match the restored behavior. I found no new regression introduced by these fixes.

**Codex on Claude's rebuttal:** "CLAUDE did not dispute the required fixes. I accept the F1 nuance that selection capture itself was host-gated; the real failure was the document-root haystack/offset and rendered-to-source line mapping. The implemented resolution still addresses that defect directly by keeping rendered Markdown out of `CommentTextSurface`. F2 and F3 are verified as fixed for the reasons CLAUDE gave."

**Codex findings (all resolved):**

| Severity | ID | Location | Resolution |
|---|---|---|---|
| 🟠 major | F1 | `BrowseFileDispatcher.tsx:195` | Markdown now renders with capture mode `"none"`; comments supported only in source view. No further action — keep read-only until a dedicated rendered-Markdown annotation path with correct quote/line mapping exists. |
| 🟠 major | F2 | `BrowseFileDispatcher.tsx:119` | Banner now a sibling above `CommentTextSurface`; `BrowseFileView` no longer owns the banner or `truncated` prop. Chrome/content boundary restored. |
| 🟡 minor | F3 | `code-tab.feature:984` | Because F1 reverted rendered Markdown to read-only, the source-view Markdown scenario and README wording are consistent again. Existing coverage documents the supported source-toggle path. |

Note: F1/F2 were filed at `blocking`/`minor` in round 1 and re-severitied to `major` at resolution; the substance is unchanged.

---

**Files changed (this track):**
- `packages/client/src/right-panel/BrowseFileDispatcher.tsx`
- `packages/client/src/right-panel/BrowseFileView.tsx`

srid:

**Outcome:** `consensus` after 1 round. Independent review: lowy=3, hickey=3. Net: all 6 findings dropped, 0 fixes applied — the structural critique was answered by the code, not by a change.

This PR moves comment-capture out of the leaf browse views and "decides commentability one level up." Both lenses opened by attacking that claim as a duplicated/fragmented seam (the diff path wraps `CommentTextSurface` inline while file/markdown/image route through `withComments`). On cross-examination, every finding collapsed the same way: the *real* shared receptacle is `CommentTextSurface`, which both paths already reach with one prop contract — so the volatility is genuinely unified at the mechanical layer, and the proposed "fix" (a second `withComments` wrapper carrying a `'text'|'iframe'|'none'` discriminator) would have *added* a conditional to a diff path that only ever has one mode. Both reviewers walked their own fixes back to drop.

### Independent findings (pre-debate)

Each lens independently flagged the same seam from its own angle. Note the initial dispositions below: **5 of 6 findings entered as `fix` and flipped to `drop` during cross-examination** — the headline of this run.

| origin | finding | location | entered as | settled |
|---|---|---|---|---|
| lowy | Commentability seam is duplicated, not unified — two receptacles for one volatility axis | CodeTab.tsx:671-685 vs BrowseFileDispatcher.tsx:78-96 | fix | drop |
| lowy | contentTick semantics leak the capture implementation through the receptacle | BrowseFileDispatcher.tsx:84-90 and CodeTab.tsx:675 | fix | drop |
| lowy | BrowseDiffView header documents a single-seam invariant the code does not enforce | BrowseDiffView.tsx:5-9 | fix | drop |
| hickey | "Decided in one place" is two seams held together by a comment | BrowseFileDispatcher.tsx:78 and CodeTab.tsx:672 | fix | drop |
| hickey | text-capture wrapping duplicated across the two seams | BrowseFileDispatcher.tsx:84 and CodeTab.tsx:672 | fix | drop |
| hickey | BrowseDiffView and BrowseFileView are near-identical sibling presenters | BrowseDiffView.tsx:33 and BrowseFileView.tsx:38 | drop | drop |

### The flips, finding by finding

**lowy-1 / hickey-1 — "two seams for one decision" (fix → drop, the load-bearing flip).**
Both lenses argued the same thing from different vocabularies: lowy called it duplicated volatility encapsulation (a future change to text-capture must be made in `withComments` *and* re-applied by hand at the CodeTab diff site), hickey called it Layer-2 fragmentation (one concept split across two sites, held together by a doc comment). The proposed fix was identical too — hoist `Capture` + `withComments(capture, {terminalId, path, contentTick}, view)` into a shared module so the diff branch declares `'text'` through the same seam.

What turned it: the actual receptacle for "how does a text view expose itself for comments?" is **`CommentTextSurface` (CommentTextSurface.tsx:38)**, not `withComments`. It is already surface-agnostic — `{terminalId, path, contentTick, class, children}` — and owns the entire capture + overlay + pill wiring, authored once. **Both** paths reach it: the dispatcher via `withComments`, the diff via a direct call at CodeTab.tsx:672. So the genuine volatility is not duplicated. `withComments` adds only a `FileData`→props adapter plus a three-way dispatch that exists to serve the dispatcher's own `Switch` over `FileData` kinds (text/image/iframe/markdown). The diff path is not a `FileData` kind and has exactly one possible capture mode — a diff is always selectable text; there is no image/iframe diff.

Hickey conceded explicitly on its own framework's terms, citing the Hickey/Lowy reconciliation rule (lowy/SKILL.md:179-181): *"Find the layer where unification is mechanical... If merging requires a mode flag, conditional branching, or type-switching — that's complecting."* `CommentTextSurface` is that mechanical layer; the proposed `withComments` wrapper would force a `capture` discriminator onto a single-valued diff call site — adding a conditional to unify, which the rule names as complecting. The mechanical unification is already done.

**lowy-2 — contentTick leaks the capture implementation (fix → drop).**
The narrow technical point is true: `withComments` synthesizes `contentTick = file.source?.content ?? ''` (BrowseFileDispatcher.tsx:84-86) rather than taking it as a parameter, while the diff site passes the hunk string. But this is not an interface-stability leak through the receptacle, because the *real* receptacle `CommentTextSurface` already takes `contentTick` as an explicit param — its docstring literally says "Pass the file content string for text browse; pass the diff hunk string for branch-diff" (CommentTextSurface.tsx:28-32). The surface-agnostic contract is correctly parameterized at the real seam. `withComments` is a dispatcher-local adapter that legitimately uses `file.source.content` for every `FileData` it handles, since every `kind:'text'` carries `source` by construction. The "future text renderer whose DOM text ≠ file source" is speculative volatility not on any roadmap, and the fix only pays off if the diff routes through `withComments` (lowy-1) — which dropped. Standalone, threading `contentTick` through two call sites that both already pass `file.source.content` is churn with no behavior change.

**lowy-3 — header documents an invariant the code doesn't enforce (fix → drop).**
Contingent on lowy-1/hickey-1. Read precisely, the `BrowseDiffView` header (BrowseDiffView.tsx:5-9) does not claim a single shared function — it explicitly enumerates two mechanisms in two files ("CommentTextSurface in CodeTab for the diff, withComments in BrowseFileDispatcher for source"), then says the *decision* is made one level up per view rather than re-open-coded inside the leaf. That is accurate: `BrowseDiffView` and `BrowseFileView` are genuinely pure presenters (verified — `BrowseFileView` no longer imports `CommentTextSurface` after this PR). The shared thing both seams reach (`CommentTextSurface`) is one place. The prose surfaces the duplication rather than papering over it. No correction needed.

**hickey-2 — text-capture wrapping duplicated across two seams (fix → drop).**
True that the four-prop `<CommentTextSurface terminalId path contentTick class>` invocation appears textually twice (BrowseFileDispatcher.tsx:84 and CodeTab.tsx:672). But this is two call sites of the *same shared component*, not one computation authored as two copies. By the duplicated-derivation logic, every two call sites of any component with the same prop set would be a fragmentation — which is just a shared receptacle in use. `terminalId` and `path` are already locals at both sites; nothing is recomputed; both sites forward locals into a component whose wiring is authored once (CommentTextSurface.tsx:38-70). Explicitly subsumed by hickey-1, so it falls with it.

**hickey-3 — BrowseDiffView and BrowseFileView are near-identical siblings (drop → drop, the one finding both lenses agreed on from the start).**
The two presenters share a skeleton (a one-element `CodeViewItem` memo, `CodeMenuFrame`, `useCodeViewSelection`, a `CodeView` with kolu props + `onError` toast) but genuinely diverge: `fileItem` vs `diffItem`, `overflow="wrap"` vs `diffStyle="unified"`, a file-only truncated banner, and `initialSelectedLines` vs `onOpen`. Collapsing into one parameterized component would interleave file-vs-diff branching behind a mode flag — complecting two strategies. Both lenses agree two thin sibling presenters is the correct shape; revisit only if a third Pierre-backed view (the planned HTML/SVG "renderable" kind) appears.

### Takeaway

The PR's "decided in one place" claim survived a determined two-lens assault because the claim is true at the layer that matters — `CommentTextSurface` is the single mechanical receptacle, and the two call-site shapes (`withComments` adapter vs direct diff wrap) are an adapter difference, not duplicated volatility. The most interesting verdict is the negative one: the obvious "unify the seam" refactor was rejected by *both* reviewers as complecting, because it would force a capture-mode discriminator onto a diff path that has exactly one mode. No commits applied; no unresolved findings.

srid:
## 👮 Code-police

This track hardened the new "comment seam" refactor — the change that routes every renderer through `withComments(capture, …)` in `BrowseFileDispatcher.tsx` so each surface must declare a capture mode, and that newly makes the rendered Markdown preview commentable. The review tightened the seam's dispatch to be compile-time-exhaustive and chased the documentation/README/comment drift that the refactor left behind.

**5 finding(s)** across the rules / fact-check / elegance passes over 4 review sweep(s). ⚠️ **Did not reach a clean sweep within the round cap** — the worktree may still have open issues; re-run /code-police on it.

|---|---|---|---|
| major | withComments dispatches on a 3-member union with a ternary, not an exhaustive match (prefer-ts-pattern) | packages/client/src/right-panel/BrowseFileDispatcher.tsx | a1e25d781 |
| minor | withComments ternary defeats its own anti-silent-omission goal — use ts-pattern .exhaustive() | — | — |
| nit | Truncated-banner DOM moved inside the comment-capture surface (browse source view) | packages/client/src/right-panel/BrowseFileDispatcher.tsx | — |
| minor | Stale doc comment: useTextSelection "Mounted on" list is now false after the surface moved | packages/client/src/comments/useTextSelection.ts | af4de2aa3 |
| minor | README comment-feature description is stale after making rendered Markdown commentable | README.md | — |

---

### Rules pass (`prefer-ts-pattern`, README sync)

**[major] `withComments` dispatched a 3-member union with a ternary, not an exhaustive match** — `a1e25d781`

`withComments` takes `capture: "text" | "iframe" | "none"` and dispatched with `capture === "text" ? <wrap> : <passthrough>`. That is dispatch-with-logic over a 3-member string-literal union, which the project's `prefer-ts-pattern` rule requires to go through `match(...).with(...).exhaustive()`. The ternary's else-branch silently lumps `"iframe"` and `"none"` together with no compile-time exhaustiveness — and it actively defeats the file's own stated design intent (header: "a new one can't silently ship without a comment decision because it has to pick a capture mode at this seam"). Adding a future 4th `Capture` mode (e.g. `"canvas"`) would compile cleanly and fall silently into the pass-through branch — the exact runtime fall-through `prefer-ts-pattern` exists to catch. The rule's `?:` exception is scoped to two-variant booleans / nullable checks; this is a 3-variant union with per-arm behavior. `ts-pattern` is already a client dependency (`^5.9.0`, used across RightPanel/App/rpc/etc.).

- **Prescribed fix:** replace the ternary with an exhaustive `match(capture).with("text", …).with(P.union("iframe", "none"), () => view).exhaustive()`, importing `match, P` from `ts-pattern` — keeping the deliberate iframe/none pass-through while closing the union at the seam.
- **Applied:** done — the 3-member union ternary in `withComments` is now an exhaustive `ts-pattern` match (`text` vs `P.union("iframe", "none")`) so a future capture mode forces a compile-time decision.

**[minor] README comment-feature description was stale after making rendered Markdown commentable**

The PR closes a real user-facing gap — the rendered Markdown preview was uncommentable and now is (`"text"` capture wraps `MarkdownRenderer` in `CommentTextSurface` via `withComments`). But README was not updated, violating `workflow.md`'s "Keep README.md in sync with user-facing changes." Two staleness points: (1) line 80 enumerates where comments work ("select text in a source file, branch diff, or rendered HTML artifact") and omits the now-commentable rendered Markdown preview; (2) line 79 tells users to flip to "the syntax-highlighted source (for exact syntax, copy, or line-anchored comments)", implying comments require switching off the rendered view — now false.

- **Prescribed fix:** add the rendered Markdown preview to the commentable-surfaces list on line 80, and drop/soften the line-79 parenthetical that frames comments as a reason to switch to source view.
- **Applied:** README updated — rendered Markdown preview added to the commentable-surfaces list and the source-view parenthetical softened. (No commit SHA recorded in the result for this fix.)

### Fact-check pass (behavioral scope + doc accuracy)

**[minor] Stale `useTextSelection` "Mounted on" doc comment** — `af4de2aa3`

The refactor moves the comment-capture surface (`CommentTextSurface`, which mounts `useTextSelection`) out of `BrowseFileView.tsx` into `withComments()` in `BrowseFileDispatcher.tsx`, and newly adds it to the Markdown rendered preview (`"text"` capture). The hook's header comment still claimed "Mounted on: BrowseFileView.tsx (text browse)" — that file no longer mounts it — and omitted the markdown preview that now does. The comment lies about where the code is wired, the exact failure the comment-the-non-obvious rule guards against: a future reader greps mount sites from this comment and lands on the wrong files. The diff made an accurate comment false without updating it.

- **Prescribed fix:** rewrite lines 5-10 to name the new seam — surface applied by `withComments("text", …)` in `BrowseFileDispatcher.tsx` (source view + Markdown rendered preview) and by `CommentTextSurface` in `CodeTab.tsx` for the diff view.
- **Applied:** done — the "Mounted on" doc comment now reflects the new seam (`withComments("text", …)` in `BrowseFileDispatcher` for source + markdown preview, plus the diff branch in `CodeTab`).

**[nit] Truncated-banner DOM moved inside the comment-capture surface**

Previously `CommentTextSurface` lived inside `BrowseFileView` and wrapped only the `CodeMenuFrame`/`CodeView`, leaving the "File truncated" banner as a sibling outside the capture host. Now `withComments("text", …)` wraps the entire `BrowseFileView` output, so the banner sits inside the `CommentTextSurface` host div — its text becomes part of the selectable/anchorable region that `useTextSelection` captures and `useHighlightOverlay` scans via `findQuote` on every `contentTick`. This is a behavioral scope change, not the byte-equivalent move the diff presents. Not a correctness bug (the banner is static non-source chrome unlikely to collide with a real `TextQuoteSelector`), but the capture region silently widened.

- **Prescribed fix:** if banner text must stay outside the surface, render it as a sibling of `CommentTextSurface` (or scope selection to a marked sub-element); if the wider region is acceptable, leave a one-line note at the `withComments("text", …)` call site so a future reader doesn't read it as an oversight.
- **Applied:** the second path was taken — a call-site comment in `BrowseFileDispatcher` now notes the truncated banner intentionally lives inside the comment-capture surface, consistent with the Markdown renderer. (No commit SHA recorded.)

### Elegance pass

**[minor] `withComments` ternary defeats its own anti-silent-omission goal — use ts-pattern `.exhaustive()`**

This is the elegance lens on the same seam as the major rules finding: the design intent (header, lines 9-15) is that a new renderer "can't silently ship without a comment decision," yet the `capture === "text" ? … : view` ternary names only `"text"` and handles `"iframe"`/`"none"` implicitly via the else — zero compile-time enforcement, and a future mode (e.g. `"html"`) would ship uncommentable with no error. The two arms aren't a pure A→B lookup (one wraps in JSX, one passes through), so this is dispatch-with-logic, squarely `prefer-ts-pattern`, not its `Record` exception.

- **Prescribed fix:** dispatch via `match(capture).with("text", …).with(P.union("iframe", "none"), () => view).exhaustive()`, keeping the deliberate iframe/none pass-through while enforcing the "every renderer must declare a mode" invariant.
- **Applied:** already resolved by the major rules fix above — `withComments` dispatches via `match().with("text").with(P.union("iframe","none")).exhaustive()`, no ternary remains; no further change needed.

---

**Net:** two findings landed real commits (a1e25d781 closing the union at the seam, af4de2aa3 fixing the stale hook doc); the README, banner call-site note, and elegance finding were addressed without separately recorded SHAs (the elegance item was subsumed by the major fix). The track did not reach a clean sweep within the round cap — re-run /code-police on the worktree to confirm nothing else is open.

PR #1159, opened 2026-06-03
perf(be-review): model tiers + per-phase token instrumentation

srid:
Optimizes `/be-review` token **cost** — full runs were ~700K–1.5M tokens, and the orchestrator ran *every* agent on Opus even though most do mechanical or synthesis work, not deep reasoning.

## Model tiers

Split the single `MODEL = "opus"` socket into three, and run each agent on the cheapest model that does its job (Opus ≈ 5× Sonnet ≈ 10–15× Haiku):

| Tier | Default | Agents |
|---|---|---|
| `model` | **opus** | deep reasoning — lens lowy/hickey (load-bearing Opus) + claude-author rounds + lens apply |
| `authorModel` | **sonnet** | synthesis — the reporter agents, the cherry-pick/reconcile, the police review + apply passes (code-police is *natively* `model: sonnet` — we stop forcing it to Opus) |
| `mechModel` | **haiku** | mechanical — setup, every commit, cleanup, comment posters, status/HEAD/clean checks, merge-base resolver, codex runner |

The tiers thread into the `codex-debate` and `lens-debate` child workflows too — their runner/commit/merge-base agents drop to Haiku while the lenses and claude-author stay Opus. Every tier is overridable via args.

The mechanical + synthesis agents are well over half the invocations per run, so this should cut **cost ~40–60%** with **no quality change** on the reasoning that matters (the lenses and the author rounds are untouched).

## Token instrumentation

Snapshot `budget.spent()` at each phase boundary, log a per-phase breakdown, and return **`tokensByPhase`** — so a run reports where its tokens actually go (Tracks dominates; Report is the reporters; the mechanical phases are cheap). This is how you measure the tiering's effect and find the next thing to trim.

## Tests

- Full-file syntax check (async-IIFE + `node --check`) on all three workflow sources **and** generated copies.
- Unit test for the token helper: no throw when `budget` is absent, correct deltas when present.
- Tier-assignment audit (every `agent()` call site → its tier) verified.
- `.apm`→generated in sync; `just fmt` clean.

### DIFF
--- a/.agents/skills/be-review/SKILL.md
+++ b/.agents/skills/be-review/SKILL.md
@@ -101,17 +101,19 @@ to `cd`.
   terse builders. The builders remain the **baseline** the agent improves and the
   **fallback** on empty/invalid output; a trivial track (track-error / clean / no
   findings) skips the agent. Use this flag when you want the fast, no-agent comments.
-- **Cost / model tiers** (`model` / `authorModel` / `mechModel`): the orchestrator
+- **Cost / model tiers** (`model` / `synthModel` / `mechModel`): the orchestrator
   and child workflows run each agent on the cheapest model that does its job, so a
   run doesn't pay Opus rates for `git`/`gh` shuffling. Defaults: **`model: opus`**
   for deep reasoning (the lens lenses — load-bearing — + claude-author + lens
-  apply), **`authorModel: sonnet`** for synthesis (the reporter agents, the
+  apply), **`synthModel: sonnet`** for synthesis (the reporter agents, the
   cherry-pick/reconcile, the police review/apply passes — code-police is natively
   Sonnet anyway), **`mechModel: haiku`** for mechanical agents (setup, every
   commit, cleanup, comment posters, status/HEAD checks, merge-base + codex runner).
-  Override any tier via args. The run reports a per-phase **`tokensByPhase`**
-  breakdown (output tokens, from `budget.spent()`) so you can see where the cost
-  goes — Tracks dominates, Report is the reporters, the mechanical phases are cheap.
+  Override any tier via args. The run reports a **`tokensByPhase`** breakdown
+  (output tokens, from `budget.spent()` on the shared turn counter) bucketed by
+  each phase's wall-clock window — NOT isolated to that phase's agents: concurrent
+  track and child-workflow output lands in whichever window is open, so read it as
+  per-mark-interval spend for tuning the tiers, not as an isolated per-phase cost.
 
 ## Steps
 
--- a/.agents/skills/be-review/SKILL.md
+++ b/.agents/skills/be-review/SKILL.md
@@ -101,6 +101,17 @@ to `cd`.
   terse builders. The builders remain the **baseline** the agent improves and the
   **fallback** on empty/invalid output; a trivial track (track-error / clean / no
   findings) skips the agent. Use this flag when you want the fast, no-agent comments.
+- **Cost / model tiers** (`model` / `authorModel` / `mechModel`): the orchestrator
+  and child workflows run each agent on the cheapest model that does its job, so a
+  run doesn't pay Opus rates for `git`/`gh` shuffling. Defaults: **`model: opus`**
+  for deep reasoning (the lens lenses — load-bearing — + claude-author + lens
+  apply), **`authorModel: sonnet`** for synthesis (the reporter agents, the
+  cherry-pick/reconcile, the police review/apply passes — code-police is natively
+  Sonnet anyway), **`mechModel: haiku`** for mechanical agents (setup, every
+  commit, cleanup, comment posters, status/HEAD checks, merge-base + codex runner).
+  Override any tier via args. The run reports a per-phase **`tokensByPhase`**
+  breakdown (output tokens, from `budget.spent()`) so you can see where the cost
+  goes — Tracks dominates, Report is the reporters, the mechanical phases are cheap.
 
 ## Steps
 
--- a/.agents/skills/be-review/be-review.workflow.js
+++ b/.agents/skills/be-review/be-review.workflow.js
@@ -11,7 +11,7 @@ export const meta = {
       title: "Setup",
       detail:
         "fan out one detached worktree per review track off the branch HEAD",
-      model: "opus",
+      model: "haiku",
     },
     {
       title: "Tracks",
@@ -23,18 +23,18 @@ export const meta = {
       title: "Consolidate",
       detail:
         "cherry-pick each track’s commits onto the branch in order; reconcile the rare overlap",
-      model: "opus",
+      model: "sonnet",
     },
     {
       title: "Report",
       detail:
         "post a detailed PR comment for each track plus the consolidation ledger",
-      model: "opus",
+      model: "sonnet",
     },
     {
       title: "Cleanup",
       detail: "tear down the per-track worktrees",
-      model: "opus",
+      model: "haiku",
     },
   ],
 };
@@ -117,6 +117,7 @@ const spentTokens = () => {
   try {
     return (typeof budget !== "undefined" && budget.spent && budget.spent()) || 0;
   } catch {
+    /* budget API absent or threw — instrumentation is best-effort, return 0 */
     return 0;
   }
 };
@@ -126,7 +127,8 @@ function markPhaseTokens(phaseName) {
   const now = spentTokens();
   const delta = now - _tokMark;
   _tokMark = now;
-  tokensByPhase[phaseName] = (tokensByPhase[phaseName] || 0) + delta;
+  // each phase name is called exactly once at its boundary
+  tokensByPhase[phaseName] = delta;
   log(`💸 ${phaseName}: +${delta.toLocaleString()} output tokens (run total ${now.toLocaleString()})`);
 }
 // The review tracks to run AND the order they consolidate in. codex first (it
@@ -1141,6 +1143,8 @@ if (!commit) {
   log(
     `--no-commit: skipping Consolidate + Cleanup. Per-track fixes are UNCOMMITTED in their worktrees; inspect them there: ${liveTracks.map((t) => `git -C ${wtDir(t)} diff`).join(" ; ")}`,
   );
+  markPhaseTokens("Tracks");
+  log(`💸 token breakdown (output, by phase): ${Object.entries(tokensByPhase).map(([k, v]) => `${k}=${v.toLocaleString()}`).join("  ")}`);
   return {
     status: "no-commit",
     branchHead,
@@ -1153,6 +1157,7 @@ if (!commit) {
     dropped: [],
     note: "commit=false: each track left its fixes uncommitted in its worktree and nothing was consolidated. The worktrees are PRESERVED for inspection — see the `worktrees` field below for each track’s path — and re-run with commit enabled to consolidate.",
     worktrees: liveTracks.map((t) => ({ track: t, path: wtDir(t) })),
+    tokensByPhase,
   };
 }
 
--- a/.agents/skills/be-review/be-review.workflow.js
+++ b/.agents/skills/be-review/be-review.workflow.js
@@ -1,7 +1,8 @@
 // The Workflow runtime requires `export const meta` to be the FIRST statement
-// and a PURE LITERAL (no interpolation), so 'opus' is inlined per phase. The
-// single MODEL socket lives just after meta; every other model reference reads
-// it lazily, well after meta is evaluated.
+// and a PURE LITERAL (no interpolation). meta.phases no longer assert a phase
+// model: each phase now spans mixed tiers, so the authoritative model is the
+// per-agent `model:` carried by each spawned agent. The MODEL socket lives just
+// after meta; every other model reference reads it lazily, after meta evaluates.
 export const meta = {
   name: "be-review",
   description:
@@ -11,30 +12,25 @@ export const meta = {
       title: "Setup",
       detail:
         "fan out one detached worktree per review track off the branch HEAD",
-      model: "haiku",
     },
     {
       title: "Tracks",
       detail:
         "codex, lens, and police gauntlets each run to consensus, concurrently and isolated",
-      model: "opus",
     },
     {
       title: "Consolidate",
       detail:
         "cherry-pick each track’s commits onto the branch in order; reconcile the rare overlap",
-      model: "sonnet",
     },
     {
       title: "Report",
       detail:
         "post a detailed PR comment for each track plus the consolidation ledger",
-      model: "sonnet",
     },
     {
       title: "Cleanup",
       detail: "tear down the per-track worktrees",
-      model: "haiku",
     },
   ],
 };
@@ -46,7 +42,7 @@ export const meta = {
 // quality loss on the parts that matter:
 //   MODEL (opus)   — deep reasoning: the lens lenses + claude-author rounds (the
 //                    lens debate FORCES Opus; that's load-bearing) and lens apply.
-//   AUTHOR (sonnet)— competent-but-not-deep work: the reporter agents, the
+//   SYNTH (sonnet) — competent-but-not-deep synthesis work: the reporter agents, the
 //                    consolidation cherry-pick/reconcile, and the police review +
 //                    apply passes (code-police is natively `model: sonnet` anyway).
 //   MECH (haiku)   — deterministic shell work: setup, every commit, cleanup, the
@@ -99,19 +95,26 @@ const postComments = a.comment !== false;
 // (richComment: false) forces the deterministic comments.
 const richComment = a.richComment !== false;
 // The three cost tiers (see the MODEL comment above). `model` = deep reasoning;
-// `authorModel` = synthesis/review (Sonnet); `mechModel` = mechanical (Haiku).
+// `synthModel` = synthesis (Sonnet); `mechModel` = mechanical (Haiku).
 const model = a.model || MODEL;
-const authorModel = a.authorModel || "sonnet";
+const synthModel = a.synthModel || "sonnet";
 const mechModel = a.mechModel || "haiku";
 
 // ---------------------------------------------------------------------------
 // Token instrumentation. `budget.spent()` is the turn's running OUTPUT-token
 // counter (shared across the main loop + all workflows). Snapshotting it at each
-// phase boundary gives a per-phase token breakdown so the cost distribution is
-// visible (Tracks dominates; Report is the reporters; the mechanical phases are
+// phase boundary gives a cost-distribution breakdown so you can see where the
+// tokens go (Tracks dominates; Report is the reporters; the mechanical phases are
 // cheap) — exactly what you need to know where to spend the next optimization.
-// Output-only and approximate when other work runs concurrently, but for a solo
-// run it maps cleanly to phases. Guarded: `budget` may be absent in some runtimes.
+// CAVEAT: each bucket is the OUTPUT tokens emitted on the shared turn counter
+// during that phase's wall-clock window — NOT isolated to that phase's agents.
+// Phase boundaries are plain marks on one shared monotonic number, not
+// synchronization points: the Tracks phase dispatches all tracks concurrently
+// (parallel(...)), and those child workflows (codex/lens) draw on the same
+// `budget`, so any concurrent track and child-workflow output lands in whichever
+// window happens to be open. Treat the numbers as per-mark-interval spend, not
+// per-phase cost — don't read Tracks vs Report as an isolated comparison.
+// Guarded: `budget` may be absent in some runtimes.
 // ---------------------------------------------------------------------------
 const spentTokens = () => {
   try {
@@ -544,7 +547,7 @@ async function policeTrack(wt) {
             {
               label: `police:${p.key}:r${policeRound + 1}`,
               phase: "Tracks",
-              model: authorModel,
+              model: synthModel,
               schema: POLICE_FINDINGS_SCHEMA,
             },
           ),
@@ -571,7 +574,7 @@ async function policeTrack(wt) {
         {
           label: `police-apply:${f.id}`,
           phase: "Tracks",
-          model: authorModel,
+          model: synthModel,
           schema: IMPL_SCHEMA,
         },
       );
@@ -1009,7 +1012,7 @@ HARD RULES:
     const out = await agent(prompt, {
       label: `report:${slug}`,
       phase: "Report",
-      model: authorModel,
+      model: synthModel,
       schema: {
         type: "object",
         additionalProperties: false,
@@ -1221,6 +1224,8 @@ if (branchMoved || branchDirty) {
       note: `NOT consolidated — ${why}, so the consolidator's "branch is at ${branchHead.slice(0, 9)}" assumption no longer holds. The track's worktree was PRESERVED at ${wtDir(t)}; after resolving the drift, replay its commits with \`git -C ${repoPath} cherry-pick $(git -C ${wtDir(t)} rev-list --reverse ${branchHead}..HEAD)\`.`,
     };
   }
+  markPhaseTokens("Consolidate");
+  log(`💸 token breakdown (output, by phase): ${Object.entries(tokensByPhase).map(([k, v]) => `${k}=${v.toLocaleString()}`).join("  ")}`);
   return {
     status: "consolidation-aborted",
     branchHead,
@@ -1234,6 +1239,7 @@ if (branchMoved || branchDirty) {
     preservedTracks: liveTracks,
     note: `consolidation aborted: ${why}. Cherry-picking onto the changed base would review against an untrustworthy scope, so nothing was consolidated and every track worktree was PRESERVED (see each track's note for the recovery cherry-pick).${dirty ? `\nOffending entries:\n${dirty}` : ""}`,
     worktrees: liveTracks.map((t) => ({ track: t, path: wtDir(t) })),
+    tokensByPhase,
   };
 }
 
@@ -1387,6 +1393,8 @@ if (currentHead !== branchHead) {
   log(
     `Consolidate: ABORTING — branch HEAD is ${sha9(currentHead) || "(unknown)"} but the tracks forked from ${sha9(branchHead)}. The branch has drifted (likely an already-consolidated re-run); cherry-picking now would double-apply every fix. Worktrees PRESERVED.`,
   );
+  markPhaseTokens("Consolidate");
+  log(`💸 token breakdown (output, by phase): ${Object.entries(tokensByPhase).map(([k, v]) => `${k}=${v.toLocaleString()}`).join("  ")}`);
   return {
     status: "consolidation-precondition-failed",
     branchHead,
@@ -1399,6 +1407,7 @@ if (currentHead !== branchHead) {
     dropped: [],
     note: `consolidation precondition failed: the branch in \`${repoPath}\` is at \`${currentHead || "(unknown)"}\` but the review tracks forked from \`${branchHead}\`. The HEAD has advanced past the shared fork point — likely a re-run in an already-consolidated worktree — so cherry-picking the track commits would stack them a SECOND time and double-apply every fix. Nothing was consolidated and the per-track worktrees are PRESERVED. Reset the branch to \`${branchHead}\` (\`git -C ${repoPath} reset --hard ${branchHead}\`) or start from a clean worktree, then re-run.`,
     worktrees: liveTracks.map((t) => ({ track: t, path: wtDir(t) })),
+    tokensByPhase,
   };
 }
 
@@ -1423,7 +1432,7 @@ Do NOT push and do NOT merge — leave the consolidated commits on the local bra
 const consolidation = await agent(consolidatePrompt, {
   label: "consolidate:cherry-pick",
   phase: "Consolidate",
-  model: authorModel,
+  model: synthModel,
   schema: CONSOLIDATE_SCHEMA,
 });
 const picks = consolidation?.picks ?? [];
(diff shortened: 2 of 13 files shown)


srid:
### Consolidation onto the branch (order: lens → police)

The lens track rewrote stale model-tier metadata in the workflow and comments; the police track caught a complementary set of correctness and style issues, though three of its commits targeted the exact same lines the lens had already fixed and were dropped to avoid redundancy.

---

#### Dropped picks — reconciliation reasoning

Three police commits were dropped because the lens track had already made deeper, more complete fixes to the same locations. The overlap is worth explaining in detail rather than burying it in a table cell.

**Police 9e39880e dropped — subsumed by lens 892d56e7**

The police reviewer noticed that `meta.phases[]` entries still carried literal `model: 'opus'` values, which had become stale after the per-agent model-tier work landed. The natural fix would have been to update those literals to the correct `haiku`/`sonnet` values. The lens track instead took a structurally cleaner approach in 54dd5ca63: it removed the `model:` field from `meta.phases[]` *entirely*, with an explanatory comment noting that per-agent `model:` is now the authoritative source and a phase-level override is misleading. Because the field no longer exists after the lens commit, updating its value is moot — the police fix is fully subsumed.

**Police 364eac2a dropped — subsumed by lens ca4decae**

The opening comment block of the workflow still said something to the effect of "opus is inlined per phase". The police reviewer added a sentence clarifying the synthesis/mechanical tier defaults. The lens track's ca4decae (cherry-picked as 3c89c8b1c) had already replaced the *entire* opening comment block with a more accurate and complete description: "meta.phases no longer asserts a phase model — each phase now spans mixed tiers, so the authoritative model is the per-agent model:". A partial addendum on top of a replaced block is nonsensical, so the police commit was dropped.

**Police 37449260 dropped — follow-on to 364eac2a, also subsumed**

This was a round-4 police finding targeting the same stale header comment as 364eac2a (replacing "opus is inlined per phase" with "each phase inlines its own model tier"). Because the lens already rewrote those exact lines in ca4decae, and because 364eac2a was already dropped, this commit was redundant in both cause and effect. Dropped for the same reason.

---

#### Clean picks

All remaining picks applied without conflict.

|---|---|---|---|
| lens | e25dce0c | `clean` | 52ba22238 |
| lens | 892d56e7 | `clean` | 54dd5ca63 |
| lens | ca4decae | `clean` | 3c89c8b1c |
| lens | d2707e79 | `clean` | d9b559858 |
| police | 5024766b | `clean` | 8f21efce5 |
| police | 23d7cf00 | `clean` | 04d16ad3a |
| police | 0e4b83cd | `clean` | ebfb0f123 |
| police | cfcd33f4 | `clean` | 3664132d6 |
| police | 68e4c641 | `clean` | 49d9213be |
| police | 20566b9e | `clean` | 69bb025b5 |

---

**Summary:** 13 picks total — 10 clean, 3 dropped. The 3 drops all targeted the same file (`.apm/skills/be-review/be-review.workflow.js`) and the same stale model/comment content, which the lens track had addressed more thoroughly. Final head: 69bb025b545019b76f207d05f71e8dcc9cbcb8dc.

srid:
## 🤖 Codex ⇄ Claude debate

> ⚠️ **Not consolidated onto the branch.** track left UNcommitted changes in its worktree (/home/srid/code/kolu/.worktrees/be-parallelize/.worktrees/be-review-1780526328-codex); NOT consolidated and its worktree was PRESERVED so any edits aren't lost. Inspect with `git -C /home/srid/code/kolu/.worktrees/be-parallelize/.worktrees/be-review-1780526328-codex log c093e6eccebcf2eea3746333f7b146fbce38d7dd..HEAD` and `git -C /home/srid/code/kolu/.worktrees/be-parallelize/.worktrees/be-review-1780526328-codex status`. Offending entries: `M apm.lock.yaml`

This track's only substantive change was fixing token accounting consistency in `be-review.workflow.js`: the new per-phase cost tracking introduced in this PR was silently dropped on three early-exit paths that can still represent significant work, contradicting both the docs and the caller's reasonable expectation of always-present cost data. The fix was clean and well-scoped.

---

### Round 1 — codex approved ❌ · 1 open · 858eb70121db4004cc98a3c8c07c9cd336250d8f

Codex's first pass focused on verifying model-tier routing consistency and confirming that the generated `.agents`/`.claude` copies were byte-identical to the `.apm` source. Both checks passed. The single finding was about the new token accounting layer introduced in this PR.

#### Findings by severity

**Minor**

| ID | Location | Issue |
|----|----------|-------|
| F1 | `.apm/skills/be-review/be-review.workflow.js:1144` | `tokensByPhase` included only on the normal completion path |

**F1 in depth:** The PR adds per-phase token instrumentation and documents it in SKILL.md as always available. However, three early-exit returns that fire *after* real work has been done — `--no-commit` (after the Tracks phase completes), `consolidation-aborted` (after the drift check), and `consolidation-precondition-failed` (after drift/clean/precondition checks) — returned without calling `markPhaseTokens(...)` to flush the in-flight phase accumulator or including `tokensByPhase` in the returned object. On those paths the cost data was simply discarded even though potentially expensive track gauntlets and consolidation attempts had already run. Codex offered two remedies: fix the exits, or narrow the documentation to honest scope.

**Claude's response to F1 — disposition: fixed**

Claude confirmed the finding and chose the more complete fix rather than narrowing the docs. The three targeted exits were patched:

1. `--no-commit` path (~line 1144): now calls `markPhaseTokens("Tracks")` before returning and includes `tokensByPhase` in the result.
2. `consolidation-aborted` path (~line 1219): now calls `markPhaseTokens("Consolidate")` and includes `tokensByPhase`.
3. `consolidation-precondition-failed` path (~line 1385): same treatment as above.

Claude also documented two deliberate non-fixes with clear reasoning:

- **`setup-failed` returns (~lines 405/424/455):** These exit before any track gauntlet runs, so there is no cost data in flight to flush. These were explicitly outside codex's stated scope ("early exits after real work") and correctly left alone.
- **Police-child return (~line 619):** This is an inner function with a structurally different result shape (`findings`/`rounds`), not the top-level run result, so the `tokensByPhase` contract doesn't apply.

After patching `.apm`, Claude regenerated the `.agents` and `.claude` copies with `just ai::apm`, confirmed all three files now carry 4 `tokensByPhase` occurrences and diff clean against each other, validated the file parses as an async function body, and ran `just fmt` (no reformatting needed).

---

Codex re-reviewed the commit and confirmed F1 resolved. The verification covered:

- The three patched early exits now call the appropriate `markPhaseTokens(...)` before returning.
- The returned objects on those paths include `tokensByPhase`.
- The `.apm` source and both generated copies (`.agents`, `.claude`) are consistent.
- The dirty `apm.lock-yaml` in the worktree is timestamp-only regeneration churn with no behavioral impact.

**Codex on Claude's scoping rebuttal:** Accepted without reservation. Codex agreed that the `setup-failed` returns are outside scope (no gauntlet work precedes them) and that the police-child return targets a different result shape. No reviewer error was flagged.

#### Final finding status

| ID | Severity | Status | Resolution |
|----|----------|--------|------------|
| F1 | minor | resolved | Fixed in all three paths; non-applicable exits documented and left unchanged |

---

### Summary

One finding, one clean fix. The debate converged in two rounds without any disputed dispositions: codex raised a precise, well-bounded issue about early-exit token accounting; Claude fixed exactly the targeted paths, explained the deliberate non-fixes with sound reasoning, and codex accepted the rebuttal in full. The only outstanding item is the unconsolidated worktree (dirty `apm.lock-yaml` from timestamp churn) — the actual code changes are ready to inspect and cherry-pick.

srid:

**Outcome:** `consensus` after 2 round(s). Three fixes applied (comment accuracy × 2, meta stale assertion × 1, misleading variable name × 1); six findings dropped after code-grounded cross-examination revealed the actual mechanisms don't match the assumed problem shape.

---

### What this track changed and why

The PR's tier-splitting work — three named model consts, per-agent `model:` fields, and phase-level token instrumentation — surfaced two classes of structural issue: (1) metadata and naming that silently lied about what the code actually does after the tiers were introduced, and (2) fragmentation concerns that looked real on first read but dissolved once both lenses traced the actual blast radius through the source. The comment-accuracy fixes are cheap and prevent future readers from trusting per-phase token comparisons the shared meter can't guarantee. The `authorModel → synthModel` rename removes an actively misleading association that compounds every future re-tier task.

---

### Independent findings

**Lowy** raised 4 findings, applying the volatility-based decomposition lens (Parnas 1972 / Lowy "Righting Software"): where does a single axis of change cut across many edit points?

**Hickey** raised 6 findings, applying the "Simple Made Easy" lens: where does the code complect two independent concerns, or maintain a convention by memory rather than by structure?

---

### Finding-by-finding: cross-examination and outcomes

| # | Origin | Finding | Location | Round-1 | Round-2 | Final | Commit |
|---|--------|-----------|-----------|---------|---------|-------|--------|
| lowy-1 | lowy | Tier-to-agent assignment sprayed across call sites | be-review.workflow.js:385,488,542,569,647,1007,1126,1182,1253,1363,1418,1515 | lowy=fix, hickey=drop | both=drop | **drop** | — |
| lowy-2 | lowy | Model consts name the solution not the capability | be-review.workflow.js:99-100 | both=drop | — | **drop** | — |
| lowy-3 | lowy | markPhaseTokens delta mis-attributed under concurrency | be-review.workflow.js:468,5165, be-review.workflow.js:468,1165 | both=fix | — | **fix** | 52ba22238, d9b559858 |
| lowy-4 | lowy | Child-workflow tier defaults duplicated across 3 files | debate.workflow.js:33-34, debate.workflow.js:52-55, be-review.workflow.js:99-100 | both=drop | — | **drop** | — |
| hickey-1 | hickey | Phase boundary fragmented across markPhaseTokens + phase + per-agent literals | be-review.workflow.js:468-469,1165-1166,1436-1437,1505-1506,1553 | lowy=fix, hickey=drop | both=drop | **drop** | — |
| hickey-2 | hickey | meta.phases[].model: "opus" disagrees with per-agent tiers | be-review.workflow.js:14,20,26,32,37 + lens debate.workflow.js:11-13 | lowy=fix, hickey=fix | both=fix | **fix** | 54dd5ca63 |
| hickey-3 | hickey | authorModel complects tier with one of its consumers | be-review.workflow.js:104 | both=fix | — | **fix** | 3c89c8b1c |
| hickey-4 | hickey | Tier defaults triplicated as inline literals | be-review.workflow.js:103-105, codex debate.workflow.js:'36-37, lens debate.workflow.js:53-55 | both=drop | — | **drop** | — |
| hickey-5 | hickey | tokensByPhase complects phase identity with wall-clock window on shared meter | be-review.workflow.js:125-131,468 | both=fix | — | **fix** | d9b559858 |
| hickey-6 | hickey | police:diff-size runs mechModel but gates a judgement | be-review.workflow.js:488-497 | both=drop | — | **drop** | — |

---

### Disposition flips and notable cross-examinations

#### lowy-1: FLIPPED fixç drop across the debate

**Lowy's round-0 case:** The genuinely volatile axis is the mapping of each agent to a tier, not the three model string constants. With ~14 `model: authorModel`/`model: mechModel` field assignments spread across be-review plus ~6 more in the child workflows, re-tiering "reporters back to opus" or "codex runner to sonnet" requires hunting every literal across three files. Lowy proposed a `tierFor` map keyed by concern name (e.g. `{ setup: mechModel, 'police:review': synthModel, ... }`) so a re-tier"becomes a single map-entry edit. Lowy proposed a `tierFor` map keyed by concern name so a re-tier becomes a single map-entry edit.

**Hickey' counter-analysis:** Traced the actual re-tier operation through the source. To move "reporters back to opus" a maintainer edits `model: authorModel` at exactly one site — the `report:\tslug` agent, where the label naming the concern sits in the same option bag as the model field. The tier decision is co-located with the agent it governs. The genuinely shared-tier cases are already de-fragmented by the named const: `authorModel` (line 104) is the single socket for report, police review, police-apply, and cherry-pick — bumping the const moves all four. The proposed `tierFor` map would RELOCATE each per-agent decision into a separate lookup whose key must stay in lockstep with the call-site label by hand, adding a second naming system parallel to labels already present at every site. That is concept addition ! not a collapse.

**Lowy's concession:** "Trace a real re-tier: to move 'reporters back to opus' a maintainer edits `model: authorModel` at exactly ONE site — report:<slug> at line 1010, where the agent() label that names the concern sits in the SAME option bag as the model field. The tier decision is co-located with the agent it governs." Drop.

---

#### hickey-1: FLIPPED fix → drop across the debate

**Hickey's round-0 case:** Each phase transition is expressed as two coupled calls — `markPhaseTokens("<prev>")` immediately followed by `phase("<next>")` — carrying different string literals that must stay coordinated by convention. The phase name also appears a third time in every agent's `phase:` field. Hickey proposed an `enterPhase(name)` helper to collapse the pair into one call.

**Lowy's counter-analysis:** The proposed helper trades a VISIBLE adjacent-pair convention (both literals on consecutive lines, audited at a glance) for an INVISIBLE one: a mutable module-level `_curPhase` that every transition must thread correctly. Worse, it leaves the load-bearing duplication untouched — the per-agent `phase:` literals in ~11 agent option bags must each match their enclosing `phase()` call, and the helper does nothing for them. So it fixes the cheap, low-blast-radius half and adds a stateful concept (Layer-3 mutable braid) for cosmetic gain, while the expensive half remains.

**Hickey's concession:** "The proposed enterPhase() helper trades a VISIBLE adjacent-pair convention ... for an INVISIBLE one: a mutable module-level _curPhase ... Collapsing the per-agent literal too would require threading the phase name through every agent() call — a larger change than this admittedly-approximate, output-only instrumentation justifies." Drop.

---

#### hickey-2: required a second round to resolve (plan ambiguity on meta key optionality)

**The finding:** The meta block declares `model: "opus"` for all five phases (lines 14, 20, 26, 32, 37) even though no agent in Setup or Cleanup runs opus, and Report/Consolidate/Tracks mix tiers. After this PR the meta value is a false assertion about the run — duplicated-derivation fragmentation that a future tier change must remember to touch.

**Round-1 agreement/disagreement:** Both lenses agreed the fix was warranted. The disagreement was on mechanism: hickey preferred deleting the `model:` key entirely; lowy offered relabeling to the dominant tier as a fallback. The question was whether `meta.phases[].model` is an optional runtime key.

**Round-2 resolution:** Lowy verified on disk that codex-debate's meta phase entry (`debate.workflow.js:5`) carries no `model` key at all — `{ title: 'Debate', detail: '...' }` — confirming the runtime accepts a phase entry without it. Deletion is available and strictly better than relabeling: a "dominant tier" label re-rots the moment the per-agent mix shifts (e.g. Report poster-vs-reporter count changes), reintroducing the same drift one level over.

**Applied fix (54dd5ca63):** Delete `model: 'opus',` from all five be-review meta phase entries. Leave lens-debate meta unchanged — it genuinely runs opus across review/debate/apply (its agents all default to `MODEL='opus'`), so those declarations are true. Update the leading comment block (lines 1-4) to note per-agent `model:` is now authoritative and meta no longer asserts a phase model.

---

#### hickey-3: clean consensus fix, first round

**The finding:** `authorModel` (line 104) is wired to reporter agents (`report:<slug>`), consolidation cherry-pick/reconcile, and police review + police-apply — none of which are "the author." The actual author rounds (claude-author in codex-debate) run on `model` (opus), not `authorModel`. The block comment at lines 49-51 already has to enumerate "reporter agents, cherry-pick/reconcile, police" to explain what it gates — the fingerprint that the name under-describes the thing.

**Cross-examination:** Both lenses verified from source: codex-debate line 159 carries `model, // deep reasoning: the author fixing/disputing real findings`, confirming the variable literally points at a consumer it does not serve and hides the three it does. Since `be-review` passes `model` and `mechModel` to children but never `authorModel`, the arg `a.authorModel` has no external caller, making the rename safe end-to-end.

**Applied fix (3c89c8b1c):** Rename `authorModel` → `synthModel` everywhere in be-review.workflow.js (declaration at 104 and all `model: authorModel` references: 546, 572, 1010, 1421). Update the COST TIERS comment and SKILL.md to call it the "synthesis (Sonnet)" tier. The trio now reads `model` / `synthModel` / `mechModel` = deep / synthesis / mechanical.

---

#### lowy-3 + hickey-5: converging findings, single shared fix

These two findings identified the same root cause from different entry points.

**Lowy-3:** `markPhaseTokens('Tracks')` at line 1165 diffs `budget.spent()` across the region where `parallel(...)` at line 726 fans codex/lens/police out concurrently, and the child workflows spend against the same shared counter. The "Tracks" bucket is therefore wall-clock spend between two marks on a shared meter, not isolated per-phase agent cost. The code comments disclose "output-only" and "approximate when other work runs concurrently" but frame the imprecision as output-only, not as "phase boundaries aren't synchronization points" — which is the actual reason a phase number can be wrong.

**Hickey-5:** Named this as complecting two independent concerns — phase identity and a time window on a shared meter — and noted the PR header's claim "Tracks dominates; Report is the reporters" asserts an isolation the mechanism can't guarantee. Initially suggested renaming `tokensByPhase` to `tokensByWindow`; conceded in favor of the cheaper comment fix to avoid churning the public result schema.

**Convergent plan:** Tighten the disclosure comments at lines 108-114 and 1565-1568, and the SKILL.md `tokensByPhase` bullet, to state: output tokens emitted on the shared turn counter during each phase's wall-clock window — NOT isolated to that phase's agents; concurrent track and child-workflow output lands in whichever window is open. Keep the field name `tokensByPhase`. Applied in 52ba22238 (lowy-3) and d9b559858 (hickey-5 — SKILL.md update).

---

#### lowy-4 + hickey-4: parallel findings, both dropped for the same runtime constraint

Both lenses independently raised the tier-default triplication: `a.model || 'opus'` and `a.mechModel || 'haiku'` appear independently in be-review, codex-debate, and lens-debate. When the mechanical floor changes, three files change with nothing enforcing lockstep.

**Why both dropped:** The workflow runtime loads each `.workflow.js` standalone and meta must be a pure literal with no interpolation or imports (documented at be-review lines 1-4, codex lines 1-5, lens lines 1-5). A shared `model-tiers.js` import is unavailable. The proposed fallback — "be-review owns the literals and always passes them down" — is ALREADY implemented: be-review passes `model+mechModel` to codex (lines 680-681) and lens (line 693) explicitly. So each child's `a.model || 'opus'` is a pure direct-invocation default, the "intentional backward-compat" item the author's note names. The live policy already lives in one file; the three dormant default literals exist to keep children standalone-invocable.

---

#### lowy-2: dropped as subsumed

The naming critique (`authorModel`/`mechModel` name the solution not the capability) is real but entirely subsumed by hickey-3. The sharper, code-grounded version of this point is the specific rename `authorModel → synthModel`; renaming without the capability-named map Lowy originally attached to lowy-1 would only relabel three variables. With lowy-1 dropped, lowy-2's proposed remedy (introduce the capability-named map indirection) falls with it.

---

#### hickey-6: factual premise was wrong, collapsed on verification

**Hickey's round-0 claim:** `police:diff-size` runs on mechModel (haiku) but "a mis-classification here silently skips the whole police gauntlet" — asserting the boolean gates the entire police track.

**Lowy's verification:** At line 515, the boolean filters exactly ONE pass: `.filter((p) => p.key !== 'elegance' || !tinyDiff?.tiny)` — a true result drops ONLY the elegance pass; the rules and fact-check passes always run. So a mis-classification at worst skips the idiom/simplicity pass on a sub-10-line diff, exactly what code-police's own tiny-diff heuristic prescribes. Moreover the task is mechanical as written: the prompt (line 487) says "run `git -C <wt> diff <base> --shortstat` and report whether insertions+deletions is under 10" — reading a number off a command's output and comparing to a constant.

**Hickey's concession:** "My own finding; the finding's premise is factually wrong, which collapses it." Drop.

---

### Applied changes summary

| Finding | What changed | Commit |
|---------|-------------|--------|
| lowy-3 + hickey-5 | Tightened disclosure comments on `markPhaseTokens` and `tokensByPhase` in be-review.workflow.js and SKILL.md to clarify wall-clock-window semantics | 52ba22238, d9b559858 |
| hickey-2 | Deleted `model: 'opus'` from all five be-review meta phase entries; updated leading comment; lens-debate meta left unchanged (its declarations are true) | 54dd5ca63 |
| hickey-3 | Renamed `authorModel` → `synthModel` everywhere in be-review.workflow.js and updated SKILL.md | 3c89c8b1c |

srid:
## 👮 Code-police

**10 finding(s)** across the rules / fact-check passes over 4 review sweep(s). ⚠️ **Did not reach a clean sweep within the round cap** — the worktree may still have open issues; re-run /code-police on it.

The police track focused on two files changed by this PR — `be-review.workflow.js` and `lens-debate/debate.workflow.js` — and found a consistent theme: the model-tier split was implemented correctly in logic but its metadata, comments, and return-shape consistency didn't fully follow. Six of the ten findings are about things that used to be true ("every phase runs on opus") that became false once the three-tier constants landed, plus four findings about token instrumentation completeness on non-happy-path exits.

---

### Pass 1 & 2 — Rules

**minor · meta.phases[*].model stale — all five phases still say "opus" after the tier split** — all five declared in  .apm/bes-review.workflow.js:1950— not commit applied

The `meta.phases` array is the live source of truth that the harness UI displays for "what model is this phase running on?" After the tier split, Setup runs on `mechModel` (Haiku), Consolidate and Report on `authorModel` (Sonnet), and Cleanup on `mechModel` (Haiku). All five entries still declared `model: "opus"`. Because the values are pure string literals (the phase-entry schema won's accept bare constant references), the fix is to hand-edit the literals: Setup → `"haiku"`, Tracks → `"opus"` (unchanged — this is the heaviest lens-debate work), Consolidate → `"sonnet"`, Report → `"sonnet"`, Cleanup → `"haiku"`. This finding was _not_ committed during the review run and remains open.

---

**nit · tokensByPhase accumulation pattern implies multi-call intent but markPhaseTokens is called once per phase** · 8f21efce55853ea02845a2dbce4009ebed9403fc

`tokensByPhase[phaseName] = (tokensByPhase[phaseName] || 0) + delta` reads like it expects to be called more than once per phase name — otherwise the `|| 0` branch is dead and the `+` always adds to zero. There are exactly five call sites, each with a unique name. A reader chasing a hypothetical "second Tracks call" wastes time. Fix applied: replaced with a direct assignment `tokensByPhase[phaseName] = delta` plus a comment that each phase name is called exactly once at its boundary.

---

**mit · Empty catch swallows unexpected errors from `budget.spent()` without any in-body comment** · 04d16ad3a57be1589f316265ddb81e8595584e1d

The `catch {}` at line 119 of `spentTokens()` was completely blank — no log, no comment. The `no-silent-error-swallowing` rule requires that an intentional best-effort catch carries a comment _inside_ the catch body explaining why the error is safe to ignore. An explanatory paragraph existed in the surrounding block comment, but a reader skimming the catch body saw only a silent swallow with no rationale at the point it matters. Fix applied: added `/* budget API absent or threw — instrumentation is best-effort, return 0 */` inside the catch.

---

**minor · `tokensByPhase` missing from the `no-commit` early-return path** · ebfb0f1232614bfbd97bc532e1949aeec62d1de1

The normal-completion return at line 1556 includes `tokensByPhase` (advertised in SKILL.md as part of the return schema). The `no-commit` early-return at line 1145 did not. A caller reading `result.tokensByPhase` after a dry-run gets `undefined`. Additionally, `markPhaseTokens('Tracks')` was never called on this path, so even if the field had been added the partial object would only contain `{Setup: N}`, losing the boundary mark for the Tracks phase. Fix applied: added `markPhaseTokens("Tracks")` to close the Setup→Tracks interval, then included `tokensByPhase` in the returned object so every return path has the same shape.

---

### Pass 2 — Fact-check

**minor · Generated `.claude/` and `.agents/` files diverge from a.apm/` source — meta phase model labels are stale** · 3664132d691eb5d80de661dd8f3f0df8   73cd8f1

The executed file (`scriptPath` the runtime uses) is `.claude/skills/be-review/be-review.workflow.js`, not the a.apm/` source. The commit hand-patched all three copies but missed updating `meta.phases` in the generated copies, so `.claude/` and `.agents/` still showed `model: "opus"` for all five phases while the source had correctly set Setup→haiku, Consolidate→sonnet, Report→sonnet, Cleanup→haiku. The harness was therefore displaying wrong model labels at runtime. Fix applied: ran `just ai::apm` to regenerate `.claude/` and `.agents/` from the corrected a.apm/` source; `apm.lock.yaml` updated accordingly.

---

**minor · markPhaseTokens implementation differs between a.apm/` source and executed `.claude/` copy** · no commit (finding self-resolved)

After the accumulation-pattern fix (8f21efce5) was applied to `.apm/` source and the `just ai::apm` regeneration ran (3664132d6), a fact-check pass confirmed all three copies — `.apm/`, `.claude/`, `.agents/` — now use the simple assignment form `tokensByPhase[phaseName] = delta`. The finding closed itself; no additional commit was needed.

---

### Pass 3 — Rules (comment accuracy)

**mit · Stale header comment claims all model references read from MODEL** · no commit applied

Line 3 of the file-top comment said "every other model reference reads [MODEL] lazily." After the tier split, only the deep-reasoning `model` tier reads `MODEL`; `authorModel` (Sonnet) and `mechModel` (Haiku) each have their own inline defaults and don't touch `MODEL` at all. The prescribed fix is to update the phrasing to something like "the deep-reasoning `model` tier reads it; the synthesis and mechanical tiers have their own inline defaults." This was flagged but not committed during the review run.

---

**nit · Stale comment claims 'opus is inlined per phase' after multi-tier split** — aside at .apm/be-review.workflow.js:63—no commit applied

The header's aside "so 'opus' is inlined per phase" became accurate for exactly one of five phases (Tracks). The other four now carry haiku or sonnet. Prescribed fix: replace with "so each phase inlines its own model tier (haiku/opus/sonnet)" or drop the per-phase aside and let the COST TIERS comment block below carry the explanation. Not committed during the review run.

---

### Pass 4 — Rules + Fact-check

**mit · Stale comment in lens-debate claims 'opus' in every phase after mechModel commit agents added** — lines 2-3 at .apm/lens-debate/debate.workflow.js · 49d9213be501abedb1e2f9e1019e3b759d2cd731

`debate.workflow.js` lines 2–3 said "the model is inlined as 'opus' in each phase below." After this PR, the Apply phase's commit sub-agents run on `mechModel` (Haiku), not Opus. Fix applied: updated the header to clarify that the primary model is Opus in each phase entry but commit agents within Apply run on `mechModel` (Haiku), matching what the new `mechModel` comment block further down already says.

---

**minor · tokensByPhase missing from two abort-path returns** — in be-review.workflow.js · 69bb025b545019b76f207d05f71    e8fdcc9cbcb8dc

By the time `consolidation-aborted` (line 1225) and `consolidation-precondition-failed` (line 1391) execute, the Setup and Tracks phases have already been marked and carry real token counts — but both return objects omitted `tokensByPhase`. Callers expecting the field on all code paths (it's present on both the `no-commit` and normal-completion paths) get `undefined`. Worse, both abort paths also skipped `markPhaseTokens("Consolidate")`, so any tokens spent during a partial Consolidate phase were silently folded into the Tracks total rather than broken out. Fix applied: added `markPhaseTokens("Consolidate")`, a token log entry, and `tokensByPhase` to both abort-path return objects, ensuring cost visibility is highest exactly when a run fails mid-consolidation.

---

### Summary table

| severity | id | files | commit |
|---|---|---|---|
| minor | meta.phases[*].model stale — all five phases still say "opus" | `be-review.workflow.js` (.apm/) | — (open) |
| nit | tokensByPhase accumulation pattern implies multi-call intent | `be-review.workflow.js` (.apm/) | 8f21efce5 |
| nit | Empty catch swallows errors from budget.spent() without in-body comment | `be-review.workflow.js` (.apm/) | 04d16ad3a |
| minor | tokensByPhase missing from no-commit early-return path | `be-review.workflow.js` (.apm/) | ebfb0f123 |
| minor | .claude/ and .agents/ copies diverge from .apm/ source — meta model labels stale | `.claude/`, `.agents/`, `apm.lock.yaml` | 3664132d6 |
| minor | markPhaseTokens implementation differs between .apm/ and .claude/ copies | — | — (self-resolved) |
| nit | Stale header comment claims all model references read from MODEL | `be-review.workflow.js` (.apm/) | — (open) |
| nit | Stale comment claims 'opus is inlined per phase' after multi-tier split | `be-review.workflow.js` (.apm/) | — (open) |
| nit | lens-debate stale comment claims 'opus' in every phase | `debate.workflow.js` (.apm/) | 49d9213be |
| minor | tokensByPhase missing from two abort-path returns | `be-review.workflow.js` (.apm/) | 69bb025b5 |

Three findings (police-r1-rules-1, police-r3-rules-1, police-r4-rules-1) remain uncommitted — the stale `meta.phases` literals and the two header-comment accuracy nits. These are low-risk but should be addressed before merge to keep the harness UI labels and inline documentation consistent with the new three-tier model split.

e6o:
Really clean approach to the model tiering. Splitting agents into Opus/Sonnet/Haiku buckets is exactly the right pattern.

A few data points that might help calibrate your tier assignments:

**Token cost differentials (as of mid-2026):**
- Claude Opus: $15/$75 per 1M tokens
- Claude Sonnet: $3/$15 per 1M (5x cheaper)
- Claude Haiku: $0.80/$4 per 1M (19x cheaper than Opus)

**For the mechanical/synthesis agents you mentioned moving to Haiku:** That is where the biggest savings come from. In our benchmarks, tasks like formatting, summarization, and linting see <2% quality degradation on Haiku vs Opus, but cost 19x less.

**One thing to watch:** The per-phase token instrumentation you are adding will likely reveal that some "deep reasoning" steps do not actually need Opus. We see this a lot -- agents labeled as reasoning-heavy often work fine on Sonnet once you measure actual vs. perceived complexity.

We built [InferCut](https://infercut.com) to do this kind of model routing automatically at the API layer. The per-agent tier approach you are taking is the right mental model. Happy to share more cost data if useful for calibrating the thresholds.

PR #1160, opened 2026-06-03
Detect Claude AskUserQuestion/ExitPlanMode by scraping the rendered screen

### LINKED ISSUE #905: Claude Code: detect AskUserQuestion / ExitPlanMode via PreToolUse hook (JSONL-poll path can't see them)

srid:
**Kolu can now light the *Awaiting input* badge when Claude Code blocks on an `AskUserQuestion` or `ExitPlanMode` (plan-approval) prompt** — the one awaiting-user signal the JSONL classifier structurally can't see. Claude's Agent SDK buffers the in-flight assistant message carrying the `tool_use` block in memory and flushes the transcript only *after* the user answers, so throughout the prompt `deriveState` reads the prior `end_turn` and reports `waiting`. The classifier was never dead — just input-starved during the wait window.

The fix recognizes the prompt on the **already-rendered terminal** instead of waiting for JSONL that arrives too late. Kolu keeps an `@xterm/headless` mirror per PTY and exposes `getScreenText` server-side, so detection reaches background tiles, mobile (client xterm unmounted), and remote/SSH PTYs — none of which a client-side buffer read could.

### How it fits together

```
JSONL watcher ──▶ ClaudeCodeInfo {state: "waiting"} ──┐
                                                       ├─▶ startAgentProvider
screen poll (~1s, idle-gated) ──▶ getScreenText(tail) ─┘     promote-only merge
        │                                                   waiting → awaiting_user
        └─▶ screen.ts screenHasClaudePrompt (pure)
```

| Seam | Role |
| --- | --- |
| `claude-code/src/screen.ts` | **Pure** detector (zero `node:*`) + promote-only policy. Signature measured stable across 82 Claude releases |
| `AgentProvider.screenScrape?` | Optional capability mirroring the existing optional `externalChanges`. **Only `claudeCodeProvider` declares it** — `createWatcher`, codex, and opencode are untouched |
| `ProviderHooks.readScreenText?` | Host-provided async screen reader bounded to a `tailLines` window. The local backend supplies it from `entry.handle.getScreenText()`; the DAG keeps its zero *synchronous* dependency on pty-host (a remote ssh host serves the same read over the wire) |
| `startAgentProvider` poller | Idle-gated ~1s clock, armed only while `isPollable(info)` holds. Promotes `waiting → awaiting_user`, **never demotes** — a genuine state change flows back through the JSONL watcher. Tied to the watcher lifecycle |

The detection signature is boolean, not a score+threshold: ExitPlanMode by exact literals (`Ready to code?`, `No, keep planning`, `…ready to execute. Would you like to proceed?`), AskUserQuestion structurally (a caret-marked numbered option row **and** a structurally-adjacent arrow-key footer), confined to the screen's bottom region so stale scrollback can't false-positive. *No box-drawing dependency — Claude's prompt boxes have no `╭/┌` corners.*

> _`session-watcher.ts` stays JSONL-only (no pty-host import, no second clock). The scrape is just another **source** that flips the already-wired `awaiting_user` state — no consumer/client change._

### Tests & evidence

- **Unit** — `screen.test.ts`: per-prompt fixtures + adversarial negatives (scrollback "proceed", shell redirection / blockquote, plain numbered lists, ASCII-fallback caret).
- **e2e** — `claude-code.feature`: two new scenarios render the real prompt signature into a live PTY and assert the dock indicator promotes `waiting → awaiting_user` — the *full* pipeline (screen → `getScreenText` poll → detector → published metadata → badge), not just the pure detector.
- **Evidence** — see the `## Evidence` comment: a clip of the badge lighting from the rendered screen alone.

### Review hardening (parallel gauntlet)

Ran through `/be-review`'s parallel gauntlet (codex⇄claude · lowy⇄hickey · code-police, each in its own worktree, consolidated by cherry-pick). Net improvements folded in: the poll reads only a bounded **screen tail** (a `tailLines` arg threaded through `getScreenText` → the pty-host surface → `ProviderHooks`) instead of the full scrollback every tick; the AskUserQuestion caret is anchored to **numbered option rows** (no false-positive on blockquotes / shell redirection); NOT_FOUND teardown races log at debug while real read failures log at error; and the onChange publish-gate self-demotes correctly on screen-less hosts. Per-track ledgers are in the gauntlet's PR comments.

The permission/tool-approval prompt is explicitly *not* in this cut — distinct on-screen shape, its own future signature in `screen.ts`.

Closes #905.

### Try it locally

```sh
nix run github:juspay/kolu/cc-scrape
```


srid:

This track scopes the Claude-Code screen-scrape poll to read only the screen *tail* (cheaper, less log spam), tightens the select-prompt detector from a loose "caret-anywhere + footer-anywhere" heuristic to a structural adjacency check, and collapses the `detectClaudePrompt(...) -> ClaudeScreenPrompt | null` surface into a minimal boolean `screenHasClaudePrompt(...)`. The three review tracks converged hard on the same two files — `packages/integrations/claude-code/src/screen.ts` and `packages/server/src/terminalBackend/providers.ts` — so most of the work below was reconciling overlapping findings rather than stacking independent changes.

Final head: cd423207fb31a02c9b3771a76f55b251deb4a9e6

**Tally:** 22 picks — 9 `clean`, 9 `reconciled`, 4 `dropped`. codex (1), lens (3), police (18). The two hotspot files (`screen.ts`, `providers.ts`) account for every non-trivial reconcile and every drop.

#### Why so much reconciliation: the two convergence points

**`providers.ts` — the poll tick.** codex landed first with the structural rewrite (tail-only `readScreen(scrape.tailLines)`, split error logging, the `tailLines` param threaded end-to-end). lens then fixed a shadowed flag and delegated publish dedup to `setAgentMetadataVia`. police arrived with the *same findings codex/lens already shipped* plus several genuinely new ones (a self-demote branch, guard-clause flattening, then a final unification). The result is that the poll tick got rebuilt incrementally across all three tracks, with later police picks superseding earlier reconciliations on the same lines.

**`screen.ts` / `screen.test.ts` — the prompt detector.** codex added adversarial negative fixtures and a caret regex; lens renamed the API to a boolean and deleted the dead `question`/`extractQuestion` concept; police strengthened the detection logic (structural adjacency) and did micro-optimizations (string hoist, helper extraction). The recurring reconcile pattern here: **adopt police's stronger detection logic + codex's fixtures, but always re-express assertions against lens's surviving boolean API** (`screenHasClaudePrompt(...).toBe(false)` rather than `detectClaudePrompt(...).toBeNull()`).

#### Dropped picks (4) — earlier track fully subsumes

These police findings were real but had already been resolved more completely upstream; resolving the conflict to HEAD produced an identical tree (empty diff) in three of the four cases.

- **police df07e15c** (full-buffer read every poll) → same as **codex F1**. codex's `tailLines` param is fully wired through `getScreenText`, `PtyHandle`/`PtyHost`, the surface zod schema, the kolu-common `TerminalHandle` contract, `local.ts`, and `ProviderHooks`. police's alternative (a negative `startLine` resolved from buffer end via `SCREEN_SCRAPE_TAIL_LINES=80`) is a mutually-exclusive design for the identical goal; keeping both would mean two competing tail-read mechanisms. Dropped.
- **police 9de9beb7** (screen-read failures logged at debug) → same as **codex F3**. police blanket-converts the catch to `plog.error`; codex's merged fix is strictly more refined — genuine failures at `error` (police's exact ask) while keeping the benign `NOT_FOUND` teardown race at `debug` via `isNotFoundError`. Empty diff vs HEAD confirms subsumption.
- **police c795481a** (dead `question`/`extractQuestion`/`OPTION_ROW_RE`) → same as **lens 924b0478**. police drops the `question` field but keeps `detectClaudePrompt` returning `ClaudeScreenPrompt | null`; lens went further, collapsing to the boolean since no consumer reads the `tool` discriminant either. Empty diff vs HEAD.
- **police fc51cd70** (inner `stopped` shadows provider-lifetime `stopped`) → same as **lens 50355581**. Both rename to `pollStopped`; lens's rename is already applied. The only conflict line was the tail-read arg (police's `readScreen(-SCREEN_SCRAPE_TAIL_LINES)` vs the branch's `readScreen(scrape.tailLines)`); resolving to HEAD gave an empty diff.

#### Reconciled picks (9) — overlap kept from both sides

**lens 924b0478** → 06d8df34b06786833fc6b95702b3d5d30e88ed03 (`screen.test.ts`). Overlap with codex 78bd2822. Kept codex's three negative fixtures (`SHELL_REDIRECT_PROSE`, `MARKDOWN_BLOCKQUOTE`, `PROSE_TO_SELECT`) AND lens's `detectClaudePrompt -> screenHasClaudePrompt` rename, rewriting codex's `.toBeNull()` assertions to `.toBe(false)` for the surviving boolean API.

**lens 50355581** → dbc270e34eabac7432f01ab0603ff5d90e295e99 (`providers.ts`). Overlap with codex 78bd2822 on the poll tick. Kept BOTH codex's `scrape.tailLines` arg and lens's `pollStopped` rename: `const text = await readScreen(scrape.tailLines)` with `if (!pollStopped && latestInfo === info)`.

**police e8da10f9** → 9be7f56d1bd0e1ca745edf7f7cb3388283caff77 (`providers.ts`). Overlap with lens 078f08f7 on the publish block. lens had removed the manual published-state dedup guard, delegating to `setAgentMetadataVia` (dedups via `agentInfoEqual`). police adds an *orthogonal new* self-demote branch: when `promoted === info` but the published state is a stale scrape-promotion, republish the raw watcher info (the watcher's change gate can silently drop the settling JSONL write). Merged both — lens's simplified `if (promoted !== info) setAgentMetadataVia(promoted)` plus police's `else if (published?.kind === provider.kind && published.state !== info.state) setAgentMetadataVia(info)`.

**police 4cf9f73e** → 4a3204acd503909a878fb720d5d17d31fed37f81 (`screen.ts`, `screen.test.ts`). Overlap with codex's caret tightening + lens's boolean API. police replaces the loose "caret anywhere AND footer anywhere in the 40-line tail" check with a **structural-adjacency** check: `hasSelectPrompt()` requires a caret-marked numbered option row (`OPTION_ROW_RE`) with the footer within `FOOTER_LOOKAHEAD_LINES` below it. Adopted police's stronger detection but kept lens's public `screenHasClaudePrompt(screenText): boolean` delegating to `hasSelectPrompt(lines)`. Kept codex's three negative fixtures and added police's new `BLOCKQUOTE_AND_NAVIGATE` fixture (as `.toBe(false)`). All 19 screen tests pass.

**police 1ef768b1** → c88931538202d4a48ab14cef1ce1599cb4dfc0a6 (`providers.ts`). Flatten the 3-deep nested if. Merged all three reviews: police's guard-clause early returns (`if (!info || !scrape.isPollable(info)) return;` and `if (pollStopped || latestInfo !== info) return;`) plus police's `finally` reschedule (needed so the early returns still re-arm the poll); lens's dedup delegation (promote branch stays `if (promoted !== info) setAgentMetadataVia(promoted)`); police-e8da10f9's self-demote else-if; and codex's split error logging (NOT_FOUND → debug, else → error) rather than police's blanket `plog.error`. Server package tsc clean.

**police b2c56f16** → e693db3400aa2d5d6110039b945309ddddb01638 (`screen.ts`). Overlap with lens's boolean API. police hoists `lines.join('\n')` into a `tail` const so it isn't recomputed per `EXIT_PLAN_LITERALS` entry. The hoist auto-merged; the only conflict was the `.some()` call — adopted police's `tail.includes(lit)` while keeping lens's boolean `return true`.

**police 8e37efd8** → 02565975a402a3edf407add747db68054e6e15d0 (`providers.ts`). Collapse promote + self-demote into one comparison. police unifies both branches into `desired = scrape.promote(info, text)` plus one divergence check `if (published?.kind === provider.kind && published.state !== desired.state) setAgentMetadataVia(desired)`. Adopted police's collapsed form (subsumes the lens-1 + e8da10f9 reconciliation), keeping codex's split logging, `scrape.tailLines`, and police's finally-reschedule. tsc clean.

**police 236b9f95** → cd423207fb31a02c9b3771a76f55b251deb4a9e6 (`screen.ts`). Overlap with lens's boolean API. police extracts a `hasExitPlanLiteral(lines)` helper symmetric with `hasSelectPrompt(lines)` and drops the redundant `tail` string, so the detector uses `lines[]` as its single currency. Adopted police's `hasExitPlanLiteral(lines)` call (required since `tail` was dropped) while keeping lens's boolean `return true`. Also fixed police's helper doc comment to reference the surviving `screenHasClaudePrompt` instead of the removed `detectClaudePrompt`. All 19 screen tests pass; server + claude-code tsc clean; pty-host 48 tests pass.

> Note the chain on `providers.ts`: e8da10f9 (add self-demote) → 1ef768b1 (flatten + reschedule) → 8e37efd8 (unify both branches into one `desired` comparison). Each later pick subsumes the earlier reconciliation's shape; the final tick form is police 8e37efd8's unified version.

#### Clean picks (9) — no overlap, applied as-is

|---|---|---|
| codex | 3d8cea36 | 78bd28220 |
| lens | c296aad6 | 078f08f74 |
| police | 96562338 | 61eae988e |
| police | 07720d30 | 1a4f9d359 |
| police | e517d667 | f0be0133c |
| police | 9cce3ae1 | 3ea142e8a |
| police | d57e6af4 | 6e63cea7d |
| police | ddb7bf1a | 028197a9a |

srid:
## 🤖 Codex ⇄ Claude debate

**Outcome:** `consensus` after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base 5592bd477a2afc03d0c0147971624a604e037815.

This track hardened the new Claude-Code screen-scrape prompt detector across the pty-host read seam: it stopped the once-per-second poll from copying the entire 50k-line scrollback when the detector only needs the tail, removed a false-positive in the AskUserQuestion signature that could mistake shell redirection or Markdown blockquotes for an option prompt, and raised genuinely unexpected scrape failures from `debug` to `error` while keeping the benign PTY-teardown race quiet. Codex opened with two majors and one minor; all three closed in one fix round and codex approved with no new regressions.

### Verdict at a glance

| ID | Severity | Area | Round opened | Disposition | Final status |
|----|----------|------|--------------|-------------|--------------|
| F1 | 🟠 major | Full-buffer read every poll | 1 | fixed | resolved |
| F2 | 🟠 major | False-positive select-prompt signature | 1 | fixed | resolved |
| F3 | 🟡 minor | Scrape failures hidden at `debug` | 1 | partial (split) | resolved |

Two fixed outright, one resolved by an agreed split (codex conceded the teardown-race carve-out). Fix commit: 78bd2822015ed4908d8d53dcf8539e79234b242a.

---

### Round 1 — codex requested changes ❌ · 3 open · 78bd28220

> The change is well-contained conceptually, but I found three issues that should be addressed before approval. The largest risk is that the new polling path reads and allocates the full 50k-line terminal buffer every second while Claude is waiting, even though the detector only uses the tail. I also found a real false-positive path in the prompt signature and an error path logged below the project's required level.

#### 🟠 Major findings

**F1 — Full-buffer read on every poll** · `packages/server/src/terminalBackend/local.ts:302`

Codex's reasoning: `TerminalHandle.getScreenText()` returns scrollback plus viewport, the configured scrollback is 50,000 lines, but `screen.ts` only inspects the last 40 rendered lines. A waiting Claude session therefore allocates, joins, ships through the pty-host surface, splits, and discards almost all of a potentially huge buffer once per second — a long-session allocation loop that surfaces as server CPU/heap churn with large scrollbacks or many waiting terminals. Suggested fixing the hook/pty-host contract to read only a tail region (`tailLines` option) and slice from `buffer.length - lines` before translating lines.

→ **Claude: fixed.** Conceded as valid. Added a `tailLines` option resolved by the pty-host against `buffer.length` (the only place the live length is known) — `startLine` pinned to `end - tailLines`, clamped at 0. Plumbed through the entire read seam: pure `getScreenText` in `packages/pty-host/src/ptyHost.ts` (plus both interface decls), the in-process host handler, the surface zod schema in `ptyHostSurface.ts`, the `TerminalHandle.getScreenText` contract in `packages/common/src/terminalBackend.ts`, the local handle impl, the `ProviderHooks.readScreenText` signature, and `makeHooks` in `local.ts`. The detector now exports its tail size (`TAIL_REGION_LINES`), which the claude-code provider declares as `screenScrape.tailLines`; the poll passes `scrape.tailLines` to `readScreen(...)`. A waiting session now reads only its 40-line tail.

**F2 — False-positive AskUserQuestion signature** · `packages/integrations/claude-code/src/screen.ts:103`

Codex's reasoning: the caret regex accepted any whitespace-delimited `>` and the footer regex accepted any occurrence of `to select`, so a completed Claude response containing shell redirection like `cat > file` plus prose such as `to select the target` satisfied both — promoting an otherwise-`waiting` terminal to `awaiting_user` and generating false attention pings. Suggested anchoring to an actual option-list shape (line-leading caret marking a numbered option, footer near the option block) with adversarial tests.

→ **Claude: fixed.** Conceded as a real false-positive path. The old `SELECT_CARET_RE = /(?:^|\s)(?:❯|>)\s/m` matched any whitespace-delimited `>`; tightened to `/^\s*(?:❯|>)\s+\d+[.)]/m` — caret at line start (after optional indent) marking a *numbered* option. Shell redirects and Markdown blockquotes have no line-leading caret marking a numbered option, so they can no longer satisfy the signature. Added three adversarial negatives to `screen.test.ts` (shell redirection + 'to select' prose, Markdown blockquote + 'to navigate', and bare 'to select' prose) with all existing positives still green. The existing 40-line trailing-trimmed tail already keeps the footer in the same painted block as the option list, so the caret tightening alone closes the reported path.

#### 🟡 Minor findings

**F3 — Scrape failures logged only at `debug`** · `packages/server/src/terminalBackend/providers.ts:546`

Codex's reasoning: a failed screen read or unexpected detector exception in the new feature path is logged only at `debug`, but the project rule requires actual errors at `error`. At `debug`, a broken screen-scrape path is invisible in production while the UI silently never promotes the prompt. Suggested either logging unexpected failures at `error`, or explicitly classifying known teardown races (NOT_FOUND) and keeping only those at `debug`.

→ **Claude: partial (split).** Agreed unexpected failures must not hide at `debug`, but argued a blanket `error` would spam logs on a benign race: the PTY can vanish between a tick being scheduled and the read landing, and the local handle / remote pty-host throws a `NOT_FOUND` ORPCError. Implemented codex's own "classify known teardown races separately" suggestion — added `isNotFoundError(err)` using a structural `.code === "NOT_FOUND"` check (so it still classifies a deserialized error crossing the R-2 socket, not just an `instanceof`). NOT_FOUND stays at `debug`; everything else now logs at `error` via the already child-bound `plog` (provider + terminal context).

---

> All three prior findings are adequately addressed. The screen scrape now requests a bounded tail through the pty-host path, the prompt detector no longer accepts arbitrary `>` characters as select carets, and scrape failures now log unexpected errors at error level while keeping the typed NOT_FOUND teardown race at debug. I did not find a regression introduced by this fix round.

Codex re-verified each finding against the fix commit and marked all three **resolved**:

- **F1 → resolved** (now cited at `providers.ts:547`): `screenScrape.tailLines` flows into `readScreenText`, through the local terminal handle, over the pty-host surface, and into `getScreenText`, bounding the read to the last N buffer lines.
- **F2 → resolved** (now cited at `screen.ts:69`): `SELECT_CARET_RE` is anchored to a line-leading caret followed by a numbered option, and the adversarial negatives cover shell redirection, blockquotes, and plain prose containing `to select`.
- **F3 → resolved** (now cited at `providers.ts:573`): the catch classifies structural `code === "NOT_FOUND"` as the benign teardown race at `debug` and logs all other failures at `error`.

**Codex on Claude's F3 rebuttal — conceded the split:**

> F3: I accept the split. A missing PTY during teardown is expected for this polling shape, and the structural NOT_FOUND check is appropriate for errors crossing the ORPC boundary. The important part of the original finding was that unexpected scrape failures must not hide at debug; the current `else` branch logs those at error, so the finding is closed.

This is the only finding that didn't close as a straight "fixed" — Claude declined the literal blanket-`error` ask and codex agreed the NOT_FOUND carve-out was the right shape, closing it on the merits rather than as a compromise.

---

### Files changed (11)

`packages/pty-host/src/ptyHost.ts` · `inProcessPtyHost.ts` · `ptyHostSurface.ts` · `ptyHost.test.ts`
`packages/common/src/terminalBackend.ts`
`packages/server/src/terminalBackend/local.ts` · `providers.ts`
`packages/integrations/anyagent/src/agent-provider.ts`
`packages/integrations/claude-code/src/agent-provider.ts` · `screen.ts` · `screen.test.ts`

The breadth of the file list reflects F1's plumbing — a single `tailLines` parameter threaded end-to-end through the pty-host read seam rather than a point patch.

srid:

**Outcome:** `consensus` after 1 round(s). Independent review: lowy=6, hickey=4. No police arbitration was needed — both lenses agreed on all 10 findings in the first round.

This track hardened the new **screen-scrape promotion path** for Claude Code's on-screen awaiting-user prompts. The two lenses converged on one structural theme — the publish/dedup seam was leaking the metadata wires and re-implementing change-detection in two callers — and one cleanup theme — a speculative question-extraction sub-pipeline and a shadowed `stopped` flag. Five findings fixed, five examined-and-confirmed-clean (four `drop`, the boundary survives the challenge).

|---|---|---|---|---|
| lowy | screenScrape and externalChanges are the same volatility axis under two sockets | packages/integrations/anyagent/src/agent-provider.ts:120-167 | drop | — |
| lowy | Promotion dedup reaches past the published-metadata receptacle to compare raw state strings | packages/server/src/terminalBackend/providers.ts:625-633 | fix | 078f08f74 |
| lowy | setAgentMetadataVia has two callers with divergent publish discipline | packages/server/src/terminalBackend/providers.ts:486-487,634-638 | fix | — |
| lowy | Tool-name volatility lives in two modules (screen.ts literals vs core.ts AWAITING_USER_TOOLS) | packages/integrations/claude-code/src/screen.ts:25-30 and packages/integrations/claude-code/src/core.ts:246 | drop | — |
| lowy | screen.ts correctly co-located rather than extracted as a package — boundary is right | packages/integrations/claude-code/src/screen.ts:1-150 | drop | — |
| lowy | latestInfo as module-local mutable across the await is the correct race guard, not a leak | packages/server/src/terminalBackend/providers.ts:404-407,618-623 | drop | — |
| hickey | extractQuestion computes a value no consumer reads — dead concept | packages/integrations/claude-code/src/screen.ts:42-44,104,113-130 | fix | 06d8df34b |
| hickey | Dedup decision split between poller's data-source-of-truth (latestInfo) and published metadata (record.meta.agent) | packages/server/src/terminalBackend/providers.ts:525-544 | fix | — |
| hickey | Two `stopped` flags in nested scopes shadow-braid the lifecycle | packages/server/src/terminalBackend/providers.ts:410,521 | fix | dbc270e34 |
| hickey | stopPoll lifecycle threaded through `current` rather than owned beside the watcher's other teardown | packages/server/src/terminalBackend/providers.ts:400-404,496,502-507 | drop | — |

---

### The structural core: one publish seam, three findings that converged on it

The two lenses arrived independently at the same defect from different doors, and the cross-examination collapsed three findings into a single fix.

**lowy-2 — dedup reaches past the receptacle.** The poll tick at `providers.ts:625-633` hand-rolled its "did the published state already reflect this?" check by reading `record.meta.agent.kind`/`.state` directly and string-comparing `published.state === promoted.state`. But anyagent already exports `agentInfoEqual` (`agent-provider.ts:187-200`) as THE receptacle for `AgentInfo` equality — with a doc comment explicitly stating equality is a free function so "one equality function suffices for every provider." The inline check reaches past it and depends on the concrete two-field shape of `meta.agent`. Hickey sharpened the proof: `contextTokens` *already* gates `agentInfoEqual` at :198 but is invisible to this inline check, so the site is *already* silently divergent — not merely at future risk.

**hickey-2 — the dedup interleaves three questions across two locations.** Same code, structural lens: the tick merges against `latestInfo` (raw watcher truth) but dedupes against `record.meta.agent` (published, possibly self-promoted), so "what am I promoting from" and "what is currently live" read from two different places coupled by an unenforced rule. The guard `promoted !== info && !(published?.kind === provider.kind && published.state === promoted.state)` braids together (a) did the scrape change anything, (b) is the published agent even this provider's kind, (c) does the published state already equal the target. The `published?.kind === provider.kind` arm is a cross-provider-ownership coupling that appears nowhere else in the function.

**lowy-3 — the convergence point.** Both above are symptoms of an asymmetric seam: the watcher `onChange` (`:486-487`) publishes **unconditionally** on every Info, while the poll tick publishes only after a hand-rolled change check. `setAgentMetadataVia` is the publish receptacle but the "should I publish?" decision lives *outside* it, inconsistently across the two callers — a §4 symmetry red flag.

**Agreed plan (one fix subsumes all three):** fold publish-if-changed into the seam. `setAgentMetadataVia` compares `nextAgent` against the currently-published `record.meta.agent` via `agentInfoEqual` and early-returns when equal (before `updateServerLiveMetadata`). Both callers then just call the verb; the poll tick drops the entire `!(published?.kind === ... && published.state === ...)` conjunction and keeps only `promoted !== info` as the scrape-changed reference guard. Hickey explicitly conceded its own alternative (a poller-local `lastPolledState` marker) in favor of the fold, because the fold *also* fixes the watcher/poller asymmetry a poller-local would leave untouched. Behavior-preservation was checked three ways: the watcher's `onChange` only fires on real changes per the anyagent contract; the null-publish at :478 is already kind-guarded and an equal-null no-op is correct; `shouldBumpRecencyForAgentChange` returns false for non-transitions anyway, so skipping a publish when equal wants no bump.

> Note on commits: lowy-2 carries commit 078f08f74; lowy-3 and hickey-2 are recorded with no standalone commit because they were folded into the same publish-seam change (the `applied[]` entries for lowy-3/hickey-2 have null commits — they landed as part of the convergent fix, not as separate diffs).

---

### Hickey's cleanup findings

**hickey-1 (fix, 06d8df34b) — dead concept: extractQuestion.** `ClaudeScreenPrompt` was a discriminated union carrying `question: string | null`, populated by an 18-line `extractQuestion` plus its own `OPTION_ROW_RE` and a caret-index walk-back. But grep confirmed the *only* consumer of `detectClaudePrompt` is `promoteFromScreen`, which uses the result purely as a boolean: `detectClaudePrompt(screenText) ? {...info, state:'awaiting_user'} : info`. The question is never threaded into `ClaudeCodeInfo`, never published, never reaches the client — the doc comment even admits "the promotion never depends on it." This is fragmentation-via-speculation: a whole extraction sub-pipeline for a unity that's never reconstituted. **Plan:** reduce to `screenHasClaudePrompt(screenText): boolean`; delete the union, the `question` field, `extractQuestion`, and `OPTION_ROW_RE`; rewrite the `.tool`/`question` test assertions to assert the boolean while keeping every positive fixture and adversarial negative. Reintroduce the union *then* if a future PR wires the question to a real consumer. Lowy conceded, adding that from the volatility lens the union arms were over-specifying the §5 interface relative to its single consumer.

**hickey-3 (fix, dbc270e34) — shadowed `stopped` flags.** `startAgentProvider` has a function-scope `let stopped = false` (`:410`, provider teardown) and `startScreenScrapePoll` declares its OWN `let stopped = false` (`:521`) shadowing it. The inner tick reads the inner flag; reconcile/cleanup read the outer — correctness depends on the reader resolving which closure closes over which. A maintainer adding a `stopped` check binds whichever is in lexical scope, silently. **Plan:** pure rename of the inner flag to `pollStopped` (declaration plus its guard, reschedule, and teardown-thunk reads); outer untouched. Zero behavior change, zero risk.

---

### Findings examined and confirmed clean (drops)

These are not "nothing found" — each is a boundary the lens deliberately challenged and the code survived.

- **lowy-1 — screenScrape vs externalChanges as two sockets.** Both encapsulate "an out-of-band signal the watcher can't see," but the triggers genuinely differ: `externalChanges` is install-once/fan-out **push** (fs/DB-WAL events), `screenScrape` is a per-session ~1s **pull**. Collapsing them behind one socket would need a push/poll mode flag — which complects configuration with mechanism, exactly what Hickey would flag. The §6.5 "don't ship three wires" concern is about a published package surface, not two stable fields in a hand-written in-repo interface. The shared-axis doc-comment suggestion is a nicety, not a defect. **Drop** — this is the correct push/pull seam.
- **lowy-4 — tool-name volatility in two modules.** `core.ts` encodes the tool-name set `AWAITING_USER_TOOLS`; `screen.ts` encodes on-screen literals (`EXIT_PLAN_LITERALS`, caret/footer regexes). These are *not* textual duplicates — they are different recognizers over different transports (JSONL transcript vs rendered VT screen). A tool rename forces a coordinated edit to both regardless, and no DRY-able abstraction exists that wouldn't complect the two matches. Survives the §1 prior-encapsulation challenge.
- **lowy-5 — screen.ts not extracted as a package.** Correctly co-located: imports `ClaudeCodeInfo` type-only, encapsulates the Claude-specific prompt-signature axis, its receptacle (`screenScrape`) lives upstream in anyagent. The §6 single-consumer signal doesn't fire as a defect — the contract is stable under signature drift (drift changes the impl, not the interface). The §7 almost-expendable test passes: a JSONL side-channel from Anthropic would delete `screen.ts` whole, contained to one file.
- **lowy-6 — latestInfo across the await.** Module-local orchestration state inside `startAgentProvider`; never escapes, `destroyCurrent` resets it so a stale read can't cross a session change. The `latestInfo === info` re-check is the minimal correct identity guard for promote-against-source-of-truth. `readScreenText` is an injected async thunk, keeping the DAG transport-agnostic, and recursive `setTimeout` (not `setInterval`) prevents overlapping slow reads.
- **hickey-4 — stopPoll threaded through `current`.** `startScreenScrapePoll()` is armed unconditionally for every provider and returns a no-op `() => {}` for codex/opencode, so they allocate one inert closure per session match. But the cost is a single empty closure — not a volatility leak: `destroyCurrent` tears down watcher + poll + latestInfo together because they share the matched-session lifecycle (the correct grain), and the screenScrape concern stays out of non-participating providers at the *type* level. Hickey's own tidy (return `null`, nullable `stopPoll`) trades a nullable field for one saved allocation — not worth the churn. **Drop.**

---

### Debate dynamics

No dispositions **flipped** between rounds — every finding was agreed in round 1 (see `history[].round 1`, all `agreed: true`). The interesting movement was *within* the cross-examination rather than across rounds:

- **hickey-2 → folded into lowy-3.** Hickey raised a standalone alternative (a poller-local `lastPolledState` marker) and then conceded it in cross-examination in favor of lowy's publish-seam fold, because the fold also repairs the watcher-path asymmetry the local would have left in place.
- **hickey-1 → lowy conceded.** Lowy initially had no equivalent finding; on examination it agreed the union was dead and added the §5 interface-over-specification angle.
- **lowy-1's suggestion softened.** The raised suggestion ("document the shared axis") was downgraded to a non-blocking nicety once both lenses agreed the split itself is correct.

srid:
## 👮 Code-police

**18 finding(s)** across the rules / fact-check / elegance passes over 4 review sweep(s). ⚠️ **Did not reach a clean sweep within the round cap** — the worktree may still have open issues; re-run /code-police on it.

This track hardened the new Claude-Code **screen-scrape prompt detector** — the poll that reads the terminal buffer to promote `waiting → awaiting_user` when Claude shows an ExitPlanMode/AskUserQuestion prompt. The bulk of the work was on the state-machine seam between the *poll* and the *JSONL watcher*: who promotes, who demotes, and how a settled transcript clears the prompt. Several findings traced a real correctness chain — a promotion that could stick forever, a same-state watcher emit that clobbers a live promotion, and a publish gate that could freeze a tile on screen-less (ssh) hosts. The rest tightened a false-positive-prone detector, fixed two `debug`-vs-`error` logging defects, added a missing lifecycle log, removed dead extraction code, and synced the README. Two findings (police-r1-fact-check-3, police-r2-elegance-2) were verified already-resolved and needed no change.

|---|---|---|---|
| major | Screen-scrape reads the entire 10k-line scrollback every ~1s but inspects only the last 40 lines | ptyHost.ts, providers.ts, local.ts, ptyHost.test.ts | dbc270e34 |
| major | Screen-read I/O failures are logged at debug, invisible to operators | providers.ts | dbc270e34 |
| minor | Unused `question` extraction feeds nothing in production | screen.ts, screen.test.ts | dbc270e34 |
| major | Promote-only demotion can stick on equal-info JSONL settle | providers.ts | 9be7f56d1 |
| minor | AskUserQuestion detector is purely structural, false-positives on idle output | screen.ts, screen.test.ts | 4a3204acd |
| minor | Screen-read I/O failure logged at debug (already resolved) | — | — |
| nit | Inner `stopped` flag shadows the provider-lifetime `stopped` | providers.ts | 4a3204acd |
| minor | Promote-and-publish decision is a 3-level nested `if` with a negated boolean | providers.ts | c88931538 |
| minor | README describes a demotion path that doesn't fire | README.md, anyagent/agent-provider.ts, claude-code/agent-provider.ts | 61eae988e |
| nit | tailRegion re-joined on every literal in detectClaudePrompt | screen.ts | e693db340 |
| minor | Poll tick's promote + self-demote branches collapse to one comparison | providers.ts | 02565975a |
| nit | `lines.join("\n")` recomputed per ExitPlanMode literal (already resolved) | — | — |
| minor | Watcher field updates clobber an active promotion (flicker + recency bump) | providers.ts | 1a4f9d359 |
| minor | Comment overclaims: non-state updates dropped for the whole held-prompt window | providers.ts | f0be0133c |
| nit | Per-session screen-scrape poll has no install/retire lifecycle log | providers.ts | 3ea142e8a |
| major | onChange publish-gate skips real state settles when the poll can't run | providers.ts | 6e63cea7d |
| minor | Divergence predicate duplicated across the two writers | providers.ts | 028197a9a |
| nit | `detectClaudePrompt` mixes string `tail` and `lines[]` abstraction levels | screen.ts | cd423207f |

---

### 🛡️ Rules pass

**Screen-scrape reads the entire 10k-line scrollback every ~1s but inspects only the last 40 lines** — *major* — dbc270e34
`readScreenText: () => entry.handle.getScreenText()` was called with no line range, serializing the *full* terminal buffer (`DEFAULT_SCROLLBACK = 10_000` lines) into one joined string every `SCREEN_SCRAPE_POLL_MS = 1000` ms — for the entire time a session sits idle at `waiting` (potentially hours with the user away) — while the only consumer, `detectClaudePrompt`, immediately discards all but the last `TAIL_REGION_LINES = 40`. That's a ~10k-line allocation plus IPC/over-the-wire transfer per second in a long-lived Node process to use 40 lines of it (the `no-unbounded-growth` "large allocations in a hot path" anti-pattern).
**Fix:** Widen the host capability to accept an optional line range — `readScreenText?: (startLine?, endLine?) => Promise<string>` — and have the poll request only the tail (a negative/`length - N` start). `getScreenText(start, end)` already clamps ranges. **Applied:** scoped the poll to read only the last 80 screen lines, dropping the per-tick allocation from O(scrollback) to O(tail).

**Screen-read I/O failures are logged at debug, invisible to operators** — *major* — dbc270e34
The poll tick's catch logged `plog.debug({ err }, "screen-scrape poll tick failed")`. `readScreenText()` is an over-the-wire/IPC read; a throw is failed I/O — exactly the category `errors-must-log-at-error` says must log at `error` (and explicitly *not* `debug`, which is reserved for expected-absent conditions). A persistently failing read means awaiting-user prompts silently stop being detected, invisible to an operator filtering on `error`/`warn`. The sibling `reconcile` catch already used `plog.error`.
**Fix / Applied:** changed the catch to `plog.error` so persistently failing screen reads surface.

**Unused `question` extraction feeds nothing in production** — *minor* — dbc270e34
`detectClaudePrompt` computed `question: extractQuestion(region)` for the `AskUserQuestion` variant, but the only production caller (`promoteFromScreen`) uses the detector as a truthy check and discards the result — a repo-wide grep confirmed nothing outside `screen.test.ts` reads `.question`. So `extractQuestion()`, `OPTION_ROW_RE`, and the `question` field were dead in production (`no-dead-code`, no for-future-use).
**Fix:** wire `question` to a user surface (tile tooltip) *or* remove it. **Applied:** removed `extractQuestion`, `OPTION_ROW_RE`, and the field; collapsed the union to `{ tool: "ExitPlanMode" } | { tool: "AskUserQuestion" }` and dropped the question assertions from the test.

**README describes a demotion path that doesn't fire** — *minor* — 61eae988e
The "How it works" prose claimed the promotion is "one-way ... once you answer and the transcript catches up the JSONL watcher settles it back." That misdescribes the mechanism: the watcher's change gate (`if (!claudeInfoEqual(info, lastInfo))`) holds `lastInfo` at `waiting` the whole time, so a settle to a structurally-equal `waiting` does **not** fire `onUpdate` — the watcher never settles it back. The real demotion is the poller's own self-demote branch. The README asserted a non-existent settle-back and omitted the poller-driven demotion that does the work (and the author/design note's "PROMOTE-ONLY ... demotion always flows back through the JSONL watcher" is contradicted by the shipped code). workflow.md requires the README to stay in sync.
**Fix / Applied:** rewrote the README clause and the source design-notes — scrape promotes `waiting → awaiting input`, and when the prompt clears the same poll republishes the raw JSONL-derived state (because the watcher's change gate drops the structurally-identical settle-back); removed the false "promote-only / watcher settles it back" framing.

**tailRegion is re-joined on every literal in detectClaudePrompt** — *nit* — e693db340
`EXIT_PLAN_LITERALS.some((lit) => lines.join("\n").includes(lit))` recomputed the join once per literal (3×) — a needless re-allocation in a hot-ish poll path.
**Fix / Applied:** hoisted `const tail = lines.join("\n")` and reused it in the literal scan.

**Comment overclaims: non-state updates dropped for the whole held-prompt window** — *minor* — f0be0133c
The onChange skip path dropped a fresh watcher `info` while a poll-promotion was live, and the comment asserted those non-state fields (e.g. a refreshed summary) are "picked up on the next promote/demote republish." But the poll's republish gate fired only on `published.state !== desired.state`; during a held `awaiting_user` both are `awaiting_user`, so a `refreshSummary`/token update resolving mid-prompt was silently dropped for the prompt's whole duration — the comment stated the opposite of the behavior, and the stale summary was a real (bounded) gap.
**Fix:** make the poll republish on *any* divergence (not just state), or correct the comment. **Applied:** chose the stronger fix — republish on any structural divergence (`isDeepStrictEqual`) so mid-prompt non-state updates are no longer dropped; corrected the overclaiming comments.

**Per-session screen-scrape poll has no install/retire lifecycle log** — *nit* — 3ea142e8a
The `watcher-lifecycle-logs` rule requires every long-lived per-session subscription (it explicitly names "per-session JSONL tail") to emit info-level install/retire logs so operators can grep `watcher (installed|retired)`. The screen-scrape poll is a session-lifetime recurring subscription armed in `startScreenScrapePoll` and torn down in `destroyCurrent`/the stop fn, yet emitted nothing — unlike every other long-lived subscription in the tree.
**Fix / Applied:** added `plog.info` install/retire logs ("claude-code: screen-scrape poll installed" / "... retired") matching the rule's message-owns-label / fields-own-identity format.

---

### 🔍 Fact-check pass

**Promote-only demotion can stick on equal-info JSONL settle** — *major* — 9be7f56d1
The poll promoted `waiting → awaiting_user` but only ever published, delegating demotion entirely to the JSONL watcher's `onChange`. That watcher is change-gated against its *own* `lastInfo` (which the poller never updates — the promoted state lives only in `record.meta.agent`, not `latestInfo`). So when the user answers and the transcript settles back to a `ClaudeCodeInfo` structurally equal to the pre-prompt `waiting`, `onUpdate` does **not** fire, `latestInfo` stays `waiting`, and the next poll reads an empty screen: `promote()` returns the same ref (no promotion), the publish branch is skipped, and `awaiting_user` is never lowered — stuck until some unrelated watcher change. The author's "a screen prompt disappearing implies a JSONL write" argument is wrong: a write deriving equal info is silently dropped by the gate, so "a write happened" ≠ "onChange fired."
**Fix / Applied:** made the poll self-demoting — when `latestInfo` is pollable, the screen no longer matches, and the published state is the scrape-promoted `awaiting_user`, republish the raw `waiting`. Closes the gap without touching the watcher's change gate.

**AskUserQuestion detector is purely structural, false-positives on idle output** — *minor* — 4a3204acd
Detection was `SELECT_CARET_RE && SELECT_FOOTER_RE` with no Claude-specific anchor. `SELECT_CARET_RE` (`/(?:^|\s)(?:❯|>)\s/m`) matches any line starting with ASCII `> ` — markdown blockquotes, REPL prompts, quoted email, diff/log output. `SELECT_FOOTER_RE` matches any `↑`/`↓` glyph *or* the prose "to select"/"to navigate." In the bottom 40 idle lines these can co-occur with no real prompt (e.g. a help screen saying "use arrows to navigate" plus a quoted `> ` line), producing a false `awaiting_user` promotion and dock ping. The ExitPlanMode literals are boolean-certain; the question path's ASCII-`>` caret is the weak link.
**Fix / Applied:** tightened the signature to require a caret-marked numbered option row with the arrow footer structurally adjacent (replacing "both appear anywhere in the 40-line tail"), and added an adversarial blockquote + "to navigate" negative fixture/test.

**Screen-read I/O failure logged at debug, invisible in production** — *minor* — *(already resolved, no change)*
A duplicate of the major logging finding above, raised independently in the fact-check pass. **Verified:** the poll tick catch at providers.ts:572 already logs at `plog.error` ("screen-scrape poll tick failed"), so the finding was already resolved — no change needed.

**Watcher field updates clobber an active promotion (flicker + spurious recency bump)** — *minor* — 1a4f9d359
The poll promotes via `setAgentMetadataVia`, but the watcher path published directly and unconditionally. While a prompt is held (published = `awaiting_user`, `latestInfo` = `waiting`), any watcher emit that keeps `state === "waiting"` but changes another field demotes the live promotion. The concrete trigger: `refreshSummary` is fire-and-forget on every transcript change, and a `fetchSessionSummary` kicked off by the last write *before* the prompt appeared can resolve 1-2s later — now during the prompt — calling `onUpdate({ ...lastInfo, summary })` with state still `waiting`, overwriting `record.meta.agent` back to `waiting`. The dock stops pinging until the next poll re-promotes (~1s). Worse, each demote+re-promote is a state transition, so `shouldBumpRecencyForAgentChange` returns true on **both**, bumping `lastActivityAt` twice and reordering the terminal in the MRU/dock mid-prompt.
**Fix:** make the demote a poll-only responsibility — don't overwrite a poll-promoted `awaiting_user` with a structurally-equal `waiting`; single writer for that state edge. **Applied:** the watcher callback now skips raw publishes that would demote a live promotion while still stashing `latestInfo`, eliminating the flicker and double recency bump.

**onChange publish-gate skips real state settles when the screen poll can't run** — *major* — 6e63cea7d
The raw-publish skip gate fired whenever `scrape && scrape.isPollable(info) && published.kind === provider.kind && published.state !== info.state` — without checking `hooks.readScreenText`. But `screenScrape` and `readScreenText` are independent optionals; the comment itself anticipates hosts (remote/ssh) that omit `readScreenText`, for which `startScreenScrapePoll()` returns a no-op. On such a host a legitimate `thinking/tool_use → waiting` settle hit the gate (`isPollable(waiting)` true, `published(thinking) ≠ waiting`), was **skipped**, and nothing ever republished it — the tile froze on `thinking`/spinner forever, the dock never settled. Even on the local host this turned every turn-end settle into a publish that waited up to 1s for the poll, and a sub-second `tool_use → waiting → tool_use` bounce dropped the intermediate `waiting` entirely — a hot-path regression the previous immediate-publish path didn't have.
**Fix:** narrow the skip to an actual live scrape-promotion the poll can settle — require `hooks.readScreenText` present and key the skip on the specific lift (`published.state === "awaiting_user" && info.state === "waiting"`) rather than any difference. **Applied:** narrowed the gate to only suppress a live `awaiting_user → waiting` promotion when `readScreenText` is present, so screen-less hosts and non-promotion settles publish immediately.

---

### ✨ Elegance pass

**Inner `stopped` flag shadows the provider-lifetime `stopped`** — *nit* — 4a3204acd
`startScreenScrapePoll` declared `let stopped = false` inside `startAgentProvider`, which already had a `let stopped = false`. The two mean genuinely different things — the outer is the provider's teardown flag (checked in `reconcile`/destroy), the inner is the poll loop's stop flag — yet share a name in nested scopes; a future edit moving code between closures would silently bind to the wrong one with no type error.
**Fix / Applied:** renamed the poll-local flag to `pollStopped` and updated its references.

**Promote-and-publish decision is a 3-level nested `if` with a negated boolean** — *minor* — c88931538
The `tick` publish decision nested three deep — `if (info && isPollable)` → `if (!stopped && latestInfo === info)` → `if (promoted !== info && !(published?.kind === provider.kind && published.state === promoted.state))` — mixing the "did the scrape change anything" check with the "already published" dedup behind a De Morgan negation. The Readability principle calls for flattening; the intent ("publish only a fresh promotion not already on the wire") was obscured.
**Fix:** flatten with guard-clause early returns and a positively-named `alreadyPublished` predicate. **Applied:** did exactly that, and moved the reschedule into a `finally` so guard returns still re-arm the poll.

**Poll tick's promote + self-demote branches collapse to one comparison** — *minor* — 02565975a
The tick had two sibling `if` branches spelling the same idea twice: branch 1 (`promoted !== info`) publishes the promotion unless already published; branch 2 (`promoted === info`) self-demotes by publishing raw `info` when the published state diverges. Both are "republish the desired info when the published state diverges from it" — the only difference is which state is desired, and `scrape.promote` already computes that. Splitting it into two ~12-line branches with duplicated `published?.kind === provider.kind` guards and an inverted comparison (`===` one, `!==` the other) re-derived in the orchestrator the merge logic that `screen.ts` owns — an inverted-operator footgun, and it contradicted the "promote-only, stateless" framing.
**Fix / Applied:** compute `desired = scrape.promote(info, text)` once and publish on any divergence; behaviorally identical, removing ~20 lines, the duplicated guard, and the footgun.

**`lines.join("\n")` recomputed per ExitPlanMode literal** — *nit* — *(already resolved, no change)*
A duplicate of the rules-pass join finding, surfaced again in the elegance pass. **Verified:** the join is already hoisted as `const tail = lines.join("\n")` and reused in the `EXIT_PLAN_LITERALS` scan — no change needed.

**Divergence predicate duplicated across the two writers** — *minor* — 028197a9a
The watcher callback and the poll tick each open-coded the same three-part guard against `record.meta.agent` — `published?.kind === provider.kind && published.state !== <candidate>.state` — in inverted polarity (the watcher to *skip* a publish, the poll to *do* one), with the `published?.kind === provider.kind` narrowing repeated verbatim. A reader has to mentally diff the two blocks to confirm they agree on what "published" means, and a future change to the kind-narrowing has to be made in two places or silently desyncs.
**Fix:** extract a `stateDiverges(candidate)` closure over `record`/`provider`. **Applied:** deduplicated the shared `published?.kind === provider.kind` kind-narrowing into one `publishedAgent()` closure, with each call site keeping its own divergence test.

**`detectClaudePrompt` mixes two abstraction levels — string `tail` and `lines[]`** — *nit* — cd423207f
The function computed both `const lines = tailRegion(...)` and `const tail = lines.join("\n")`, feeding `tail` (string) to the ExitPlanMode `.includes` check and `lines` (array) to `hasSelectPrompt` — two representations of the same region, forcing the reader to track which check consumes which shape.
**Fix:** pick one currency; push the literal scan into a `hasExitPlanLiteral(lines)` helper symmetric with `hasSelectPrompt(lines)`. **Applied:** extracted `hasExitPlanLiteral(lines)` so `detectClaudePrompt` uses `lines[]` as its single currency, dropping the redundant `tail` string.

---

*Files referenced live under the police review worktree: `/home/srid/code/kolu/.worktrees/cc-scrape/.worktrees/be-review-1780518893-police/`. Hot spots: `packages/server/src/terminalBackend/providers.ts` (the poll/watcher writers), `packages/integrations/claude-code/src/screen.ts` (the detector), and `packages/pty-host/src/ptyHost.ts` (the scoped screen read).*

srid:
## Evidence

**The dock agent indicator promotes `waiting → awaiting_user` purely from the rendered screen** — no transcript change. The clip mocks a Claude session sitting at `waiting` (the JSONL the SDK would leave during an `ExitPlanMode` prompt), then paints the plan-approval prompt's measured signature into the real PTY. The server's screen-scrape poll reads it off the live buffer via `getScreenText`, the detector fires, and the tile chrome flips to **Awaiting input** — exactly the signal #905 says the JSONL classifier can never see.



▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=claude-screen-scrape-awaiting.mp4

This rides the full pipeline end-to-end — live PTY screen → `getScreenText` poll → `screen.ts` detector → promote-only merge → published metadata → dock badge — not just the pure unit detector. It's a real e2e scenario added in this PR (`features/claude-code.feature`), not a bespoke capture script and with no feature-file edit beyond the new scenario itself.

_Produced by this PR's own harness capture — `KOLU_EVIDENCE=1 just test-quick features/claude-code.feature --name "plan-approval prompt on screen promotes waiting to awaiting"` — through the Cucumber e2e harness (Playwright `recordVideo` + `slowMo` at 1280×720, animations left on); the `After`-hook saved the `.webm` this clip was transcoded from._

Issue #881, opened 2026-05-13
Comments: phased rollout roadmap


srid:
Tracking issue for the comments-on-files feature's phased delivery. Each phase is independently useful when merged.

## Phases

| # | Scope | Status | Issue / PR |
|---|---|---|---|
| 0 | `openInCodeTab` URL-handler refactor + right-click "Open at line N" in any code / diff viewer | ✅ Shipped | #891 |
| 1 + 2 | Character-range comments on text files, branch diffs, and HTML iframe artifacts. Select → floating pill → composer popover → tray accumulating across files. "Copy to clipboard" flushes as plain Markdown. Per-worktree `localStorage` persistence keyed by `repoRoot` with a `{v: 1, comments}` envelope. | ✅ Shipped | #922 |

## What changed (vs. the original plan)

The original roadmap shipped phase 1 + phase 4 (persistence) together via #879. Post-merge structural review surfaced enough complecting in that implementation — a comment-mode toggle braided with tray-visibility logic, a module-level bucket `Map<repoRoot, Comment[]>` that grew without cleanup, a deferred-effect cross-component selection projection that leaked stale ranges across file switches, an unversioned persisted shape — that a clean redo turned out cheaper than incremental remediation. #879 is superseded.

The redo also collapsed the design further during implementation:

- **Phase 1 + Phase 2 combined.** Treating text and HTML as different surface kinds turned out to fragment the UX without buying anything. Both now use arbitrary character-range selection with a W3C `TextQuoteSelector` (`{quote, prefix, suffix}`) as the locator — same wire format, same composer, same tray. Pierre's text/diff viewers use a parent-side `useTextSelection` adapter (shadow-DOM aware via `ShadowRoot.getSelection()`); the HTML iframe surface uses an in-iframe SDK over `postMessage` because of opaque-origin sandbox isolation. The split is implementation-internal, invisible to the user.
- **Image annotator dropped.** The original plan included a Phase 3 for bbox-drawn comments on PNG / JPG artifacts. In practice the agent's image outputs (mockups, screenshots, charts) are reviewed by quoting "the chart in the top-right" in a normal comment; a dedicated bbox affordance was bureaucracy without a real workflow demand. If the need surfaces later, it slots back in cleanly — the locator union is open.
- **Select-and-comment, no right-click.** A floating "+ Comment" pill appears next to the selection. Click it → composer popover → save → tray. No context-menu entry point.
[...]

PR #1162, opened 2026-06-03
Enable comments on the Markdown rendered preview

srid:
**Flips the Markdown rendered preview from `"none"` to commentable**, closing the gap that #1158 deliberately left open. #1158 reverted rendered-Markdown commenting because wrapping it in `CommentTextSurface` produced *document-scoped* anchors — the preview is light DOM with no shadow root, so the locator haystack and highlight re-find fell back to the whole page, and a rendered-DOM line got persisted as a bogus *source* line. This PR does it correctly.

### Plan

- **Subtree-scoped anchoring** — anchor the quote/prefix/suffix against the preview host subtree, not the owning `Document`, so a light-DOM (non-shadow) surface captures and re-finds within itself.
- **No source `lineRange` for rendered views** — a rendered-DOM line isn't a source line, so don't persist one (the "jump to line" hint only makes sense for source/diff).
- **Capture mode** — change the Markdown `textRenderers` entry in `BrowseFileDispatcher` from `withComments("none", …)` to a commentable mode.
- **Cross-representation drift (open decision)** — a comment made in the rendered preview won't re-find in the *source* view (quote `Hello Doc` ≠ source `# Hello Doc`) even though both share the same path-keyed comment list. Decide: preview-only, or shown in both with best-effort re-anchoring.
- **e2e** — a `code-tab.feature` scenario: select in the rendered preview → pill → composer → tray (the coverage whose absence let the #1158 bug through).

> Scaffold PR — implementation lands in follow-up commits.


srid:

The review gauntlet tightened the `comment-where` branch in three complementary ways: codex improved feature completeness and e2e coverage, lens sharpened module exports and type imports, and police enforced DRY structure and removed silent behavioral defaults. The merge required five reconciliations and dropped two redundant picks, landing at final head 0e36966ca185e49d04dce744169187d1c9cc4c8c.

---

#### Clean picks (no conflicts)

|---|---|---|
| codex | 5229ee49 | 5be40812 |
| lens | a180e900 | 95803ead |
| lens | 2b67a080 | b2ed09d2 |
| lens | efe76796 | d6015e83 |
| police | cf769398 | 23850783 |
| police | 638b4be0 | 902c719d |
| police | a0deca88 | 2ff71bc1 |
| police | 09c45c82 | 0e36966c |

---

#### Reconciled picks (conflicts resolved, both improvements retained)

**police 08823cc4 — QuoteRoot doc comment**

Both codex and police noticed that the `QuoteRoot` JSDoc falsely claimed all three union shapes (`Document | ShadowRoot | Element`) expose `createTreeWalker` directly. They diverged on to fix it: codex wrote a longer explanation, police wrote a tighter one that named the asymmetry explicitly — `ShadowRoot` and `Element` have `ownerDocument`, `Document` does not. The merged comment keeps police's concise asymmetry statement and grafts in the codex parenthetical that explains why the core always routes through `ownerDocument`. Neither version alone was complete; the merged form is.

**police 8ed745d9 — `lineAnchored` made required**

Police eliminated the `?? true` silent defaults that were hiding a contract: `lineAnchored` was optional in both `UseTextSelectionOptions` and `CommentTextSurfaceProps`, with the fallback silently forcing line-anchored behavior when callers omitted the prop. Codex had layered a new `surface` prop onto the same interfaces alongside the `?? true` fallback in `observeMutations`. Reconciliation: `lineAnchored` is now required (police), `surface` is preserved (codex), and `observeMutations` uses `!props.lineAnchored` directly without a default. `CodeTab.tsx` auto-merged cleanly because police had already added `lineAnchored={true}` to the diff surface call site.

**police 75df43cc — `TextSurface` comment block**

Two independent comment improvements collided on the same block. Police corrected a wrong sizing-class claim (`h-full` → `min-h-0 w-full flex-1`), Codex had added a rationale explaining why the truncation banner lives outside the prose `CommentTextSurface`. The merged block keeps the accurate sizing-class wording (police) and the truncation-banner rationale (codex) — factual accuracy and explanatory depth, together.

**police 2d515abf — `TruncatedBanner` extraction**

Police refactored the duplicated truncation-banner JSX into a `TruncatedBanner` component (better DRY). Codex had addressed the same duplication via a `truncationBanner` helper function but had also attached a `data-testid` attribute required by the new e2e test steps targeting `browse-truncation-banner`. Reconciliation adopts the component approach (police — cleaner SolidJS idiom) and adds `data-testid` to it (codex — required for test infrastructure). Neither could be dropped without breaking something.

**police 7fbd47c1 — `TruncatedBanner` hoisted to module scope**

This pick followed directly from the 2d515abf reconcile: police moved `TruncatedBanner` from inside `BrowseFileDispatcher`'s function body to module scope (the correct place for a stable SolidJS component, avoiding re-creation on each render). HEAD from the prior reconcile still had the inner definition. Auto-merge placed the module-level definition correctly; the conflict was the removal of the inner one. Resolution took the police side (removing the inner definition) and transferred the `data-testid` to the module-level component to keep e2e compatibility.

---

#### Dropped picks (fully subsumed)

**police fc341b35 — `QuoteRoot` import in `useTextSelection.ts`**

The lens track (a180e800) had already exported `QuoteRoot` from `@kolu/artifact-sdk/client` and added the `import type { QuoteRoot }` line to `useTextSelection.ts`. The police commit made an identical change to that same import. The other files touched (`highlightOverlay.ts`, `client/index.ts`) auto-merged cleanly. After resolution the working tree was empty — full subsumption confirmed.

**police 2844407a — stale comment fix**

This commit carried no file changes at all (empty diff stat). The stale comment it targeted had already been addressed by the earlier police reconcile 75df43cc. The cherry-pick was empty and skipped cleanly.

---

#### Summary

15 source commits processed: 8 clean, 5 reconciled, 2 dropped. The reconciliations were all meaningful — in each case both tracks had contributed something the other lacked, and the merged result is strictly better than either input alone. The two drops were genuine redundancies with no information loss.

srid:
## 🤖 Codex ⇄ Claude debate

This track wired rendered Markdown into the existing comment surface, but the initial PR left two address spaces for the same path (source vs. rendered DOM) insufficiently distinguished — both in routing and in highlight lifecycle. Two rounds of debate identified and resolved all six findings without dispute.

---

### Round 1 — codex approved ❌ · 6 open · 5be40812

Codex's opening assessment: the feature mostly works, but Markdown's dual address space (source string vs. rendered DOM) creates two correctness gaps that can silently fail users — a tray jump that can't navigate back to the rendered surface, and highlight ranges that go stale when the async Shiki renderer swaps the prose DOM. Four additional findings covered surface boundary leakage, stale docs, and misleading type comments.

#### Major findings

**F1 — Tray jump cannot return to the rendered Markdown surface**
`packages/client/src/right-panel/CodeTab.tsx:744`

Rendered Markdown comments carry no `lineRange`. The tray jump handler for no-`lineRange` comments only sets browse mode and `selectedPath`. If the same `.md` file is already open in Source mode, `selectedPath` doesn't change, `FileView` is not remounted, and nothing signals the Source/Rendered toggle to flip back to Rendered. The scroll request is then consumed by the source surface, where a rendered-only quote may not exist — the tray item silently fails to return the user to their comment.

Codex's suggestion: persist the comment surface/mode (`source`/`prose`/`iframe`) and make `FileView` mode controllable so a prose comment jump explicitly selects the rendered surface before resolving the quote.

**F2 — Highlight ranges go stale when Shiki replaces the prose DOM**
`packages/client/src/right-panel/BrowseFileDispatcher.tsx:104`

The shared `textSurface` passes the source content string as the sole `contentTick` for both source and rendered Markdown views. For rendered Markdown with fenced code blocks, `Markdown.tsx` reassigns `innerHTML` when the lazy Shiki `highlighter()` resource resolves — swapping every text node without changing the source string. CSS Highlight API ranges applied before the swap point at detached nodes and disappear; the overlay's reactive effect never re-runs because its dependency (the source string) hasn't changed.

Codex's suggestion: make the rendered Markdown surface bump the overlay when its rendered DOM changes — either via a version tick from `MarkdownRenderer` or by observing host subtree mutations for prose surfaces.

#### Minor findings

**F3 — Truncation banner inside the commentable prose surface**
`packages/client/src/right-panel/BrowseFileDispatcher.tsx:129`

The prose `CommentTextSurface` wrapped the whole `MarkdownRenderer`, including its own truncation warning chrome. This made non-file UI text selectable and commentable, while the source path explicitly kept its truncation banner outside the commentable surface to avoid anchors the agent can't locate in the file. An inconsistency in surface boundary discipline.

**F4 — README not updated for rendered Markdown comments**
`README.md:80`

User-facing docs described comments as available on source files, branch diffs, and rendered HTML artifacts — but not rendered Markdown previews. The project conventions require `README.md` to stay in sync with user-facing changes.

#### Nits

**F5 — Stale feature-file comments**
`packages/tests/features/code-tab.feature:982`

Two comment blocks were stale: one still attributed selection to `Selection.addRange` when the step helper now performs a real mouse drag; another claimed rendered Markdown was read-only/source-only despite this branch adding rendered-preview commentability.

**F6 — Misleading `QuoteRoot` type comment**
`packages/artifact-sdk/src/types.ts:13`

The new doc comment claimed all three `QuoteRoot` shapes (`Document`, `Element`, `ShadowRoot`) expose `createTreeWalker`. `Element` and `ShadowRoot` do not — the implementations correctly go through `ownerDocument.createTreeWalker`, but the comment misrepresented the contract.

---

#### Claude's response — all six fixed, no disputes

Claude accepted every finding and implemented fixes without pushing back on any:

**F1 (fixed):** Claude traced the exact failure path — a prose comment's missing `lineRange` means the jump only sets `browse + selectedPath`; same-path navigation skips remount so `FileView`'s internal `chosen` signal stays `'source'`. The fix operated at three layers: (1) added `surface?: "source"|"prose"` to `Comment`, `ComposerTarget`, and threaded it through `useTextSelection.activate`, `CommentComposer.add`, and `CommentTextSurface`; (2) added an optional `mode` prop to `FileView.tsx` that uses `createEffect(on(() => props.mode))` to adopt the `chosen` slot — fires on initial run so fresh-mount jumps also land correctly, user can still toggle away afterward; (3) carried `surface + path` on `ScrollRequest`, set in `CodeTab.onJumpTo`, with `BrowseFileDispatcher` deriving `jumpMode` from a matching request and passing it to `FileView`. Also fixed the symmetric pre-existing case of a source comment landing on the rendered default. New e2e scenario: "Tray jump returns to the rendered Markdown surface."

**F2 (fixed):** Claude confirmed the exact trigger — `html` memo re-runs when `highlighter()` flips `undefined → ready`, replacing every text node. Fix in `highlightOverlay.tsa: added an opt-in `observeMutations` flag; when set, a dedicated effect attaches a `rAF`-coalesced `MutationObserver` on the host subtree that bumps a `domTick` signal, which the apply-effect now depends on. Since `applyHighlights` uses the CSS Highlight API (no DOM writes), Claude's own re-apply cannot re-trigger the observer — no feedback loop. The flag is kept off for source/diff views where Pierre's virtualizer would cause thrashing; wired via `observeMutations={!lineAnchored}` from `CommentTextSurface`. New e2e scenario: "Rendered Markdown comment highlight survives the Shiki re-render," with a `CSS.highlights.get('kolu-comment').size > 0` assertion.

**F3 (fixed):** In `BrowseFileDispatcher.tsx`, added a shared `truncationBanner(file)` helper (with `data-testid='browse-truncation-banner'`), rendered as a sibling above the comment surface in both source and prose renderers' new flex-col wrapper. Stopped passing `truncated` to `MarkdownRenderer` (defaulting to false) so the banner no longer renders inside the surface. Updated the truncation warning step to target the new testid outside `browse-preview-markdown`.

**F4 (fixed):** Updated `README.md` line 80 to list rendered Markdown preview as a commentable surface, noting it anchors against the preview's own subtree with no source line (quote-only anchor, still grep-locatable), while source-view comments additionally carry a line range.

**F5 (fixed):** Refreshed both stale comment blocks: the `#881` block now describes the real `dragSelectText` mouse-drag helper; the Markdown scenario header no longer claims rendered Markdown is read-only/source-only and instead frames source-view as the line-addressable counterpart.

**F6 (fixed):** Reworded the `QuoteRoot` doc to say each shape exposes `textContent` and an owning document (itself when a `Document`, else `ownerDocument`) from which a `TreeWalker` is created — matching `extractOffsets.ts`/`findQuote.ts`.

---

Codex verified all six fixes against the working tree and found them adequate. No new regressions introduced. `git status --short` and `git diff --check` both clean.

| Finding | Severity | Location | Status |
|---------|----------|----------|--------|
| F1 — Tray jump surface routing | major | `CodeTab.tsx:756` | resolved |
| F2 — Stale highlights after Shiki swap | major | `highlightOverlay.ts:79` | resolved |
| F3 — Truncation banner inside comment surface | minor | `BrowseFileDispatcher.tsx:155` | resolved |
| F4 — README not updated | minor | `README.md:80` | resolved |
| F5 — Stale feature-file comments | nit | `code-tab.feature:982` | resolved |
| F6 — Misleading `QuoteRoot` doc | nit | `types.ts:13` | resolved |

**Codex on Claude's rebuttal:** CLAUDE did not dispute any findings. All fixes verified and accepted.

**Files changed by Claude's follow-up:**
`packages/artifact-sdk/src/types.ts` · `packages/client/src/comments/types.ts` · `composerState.ts` · `scrollRequest.ts` · `useTextSelection.ts` · `CommentComposer.tsx` · `CommentTextSurface.tsx` · `highlightOverlay.ts` · `BrowseFileDispatcher.tsx` · `CodeTab.tsx` · `packages/solid-fileview/src/FileView.tsx` · `README.md` · `packages/tests/features/code-tab.feature` · `packages/tests/step_definitions/code_tab_steps.tsa

srid:

**Outcome:** `consensus` after 1 round(s). Independent review: lowy=5, hickey=3.

Both lenses converged on the same root defect from different angles: `QuoteRoot` is minted as the canonical SDK type but never published to consumers, causing three parent-side call sites to re-spell the same union by hand — with one site using a fourth, narrower variant. Three fixes landed; two drop-confirmations documented that the current shape is intentional and correct.

---

### Independent findings per lens

**Lowy** applied the electricity/receptacle test (Juval Lowy's volatility-based decomposition framework): does the boundary hide a genuine axis of change, and do consumers plug into the socket rather than reach past it to the wires?

**Hickey** applied the fragmentation/complecting test (Rich Hickey's Simple Made Easy framework): is one concept stated in N places, forcing manual agreement where mechanical enforcement was possible?

Both lenses independently flagged the same two defects (the un-exported type and the duplicated ternary), arriving at identical fix plans from different first principles. No finding flipped disposition during the debate — all eight settled in round 1 with full agreement.

---

### Finding-by-finding breakdown

#### lowy-1 / confirmed: QuoteRoot receptacle widened correctly

**Location:** `packages/artifact-sdk/src/types.ts:13-21`
**Disposition:** drop (confirmed correct)

Lowy's electricity test: `QuoteRoot = Document | ShadowRoot | Element` is the receptacle (the stable contract exposing `textContent` + `createTreeWalker`); the three DOM-root shapes are its volatile generators. Adding `Element` as the third generator was purely additive — `extractOffsets.ts:21-23` already routes `Document→body` and `Element/ShadowRoot→self`, so the receptacle absorbed the new shape with zero consumer call-site churn. This is the PR doing the right thing. Hickey agree. Recorded as an affirmative pass, not a silence.

---

#### lowy-2 + hickey-1 (converged): QuoteRoot not published — consumers re-spell the union by hand

**Location:** `packages/artifact-sdk/src/client/index.ts:23-29` (omission); `packages/client/src/comments/highlightOverlay.ts:42`; `packages/client/src/comments/useTextSelection.ts:85`
**Disposition:** fix ℠> applied in 95803ead

This is the central defect of the PR, caught independently by both lenses:

- **Lowy framing:** consumers should plug into the receptacle (`QuoteRoot`), not reach past it to the wires (`Document | ShadowRoot | Element`). `findHostRoot`'s return type (highlightOverlay.ts:42) and `lineRangeForSelection`'s param type (useTextSelection.ts:85) both spell the literal union expansion. When a fourth generator is ever admitted to `QuoteRoot`, these two sites must be hunted down and edited in lockstep — the exact blast-radius the receptacle was meant to eliminate.

- **Hickey framing:** the one named concept `QuoteRoot` is shattered into independent hand-spellings. The invariant "all consumers agree with QuoteRoot" is unenforced; a future SDK widening (e.g. `DocumentFragment`) would not propagate automatically, and the type passed into `extractQuote`/`extractOffsets` would silently drift.

Both lenses also identified the proximate cause: `client/index.ts:23-29` never re-exports `QuoteRoot` (only `Locator` and `SelectionRect` are in that block), so the canonical name is not importable from the consumer side.

**Agreed fix:** add `QuoteRoot` to the `export type { ... }` block in `packages/artifact-sdk/src/client/index.ts`; import it in `highlightOverlay.ts` and `useTextSelection.ts`; annotate `findHostRoot`'s return type and `lineRangeForSelection`'s param type. Behavior unchanged; only annotations collapse onto the published name. This single edit discharges lowy-2, lowy-3, and hickey-1.

---

#### lowy-3 (converged): `resolvedRoot` annotated as a fourth, narrower ad-hoc type

**Location:** `packages/client/src/comments/useTextSelection.ts:208-209`
**Disposition:** fix — applied in b2ed09d2

`resolvedRoot` is annotated `ShadowRoot | HTMLElement` — a third distinct local spelling that does not even match the other two sites (`HTMLElement` vs `Element`; no `Document`). The asymmetry is the smell: three sibling sites naming the same receptacle three different ways. The value flows immediately into `extractQuote` (line 210, takes `QuoteRoot`) and `lineRangeForSelection`, so the narrower annotation buys nothing except drift risk.

`HTMLElement` is assignable to `Element`, so `QuoteRoot` covers the runtime branch (`rootNode instanceof ShadowRoot ? rootNode : host`) exactly. The fix is purely annotation.

Both lenses agreed this folds into the same one-name fix as lowy-2/hickey-1.

---

#### lowy-4 + hickey-2 (converged): `lineRangeForSelection` re-inlines `rootTextContent

**Location:** `packages/client/src/comments/useTextSelection.ts:90-93`
**Disposition:** fix ℠> applied in d6015e83

`lineRangeForSelection` computes the haystack text as:
```
root instanceof Document ? root.body?.textContent ?? '' : root.textContent ?? ''
```
This is byte-for-byte the body of `rootTextContent` in `extractQuote.ts:21-25`. That function was exported with an explicit doc comment stating it exists "so applyHighlights and the parent-side highlight overlay can reuse the same extraction rather than inlining the same ternary." `highlightOverlay.ts:76` already reuses it. `useTextSelection` is the sole re-implementer.

The danger here is not just duplication — it is pairing. `lineRangeForSelection` pairs this text derivation with the `extractOffsets` call on line 88. If `rootTextContent` ever changes how it handles a `QuoteRoot` shape, the copy in `lineRangeForSelection` will silently diverge, corrupting the line-number jump relative to the offset anchors.

**Agreed fix:** add `rootTextContent` to the existing `@kolu/artifact-sdk/client` import in `useTextSelection.ts` (where `extractOffsets` is already imported from the same path) and replace lines 90-93 with `const text = rootTextContent(root);`. A one-token import change.

---

#### lowy-5 / confirmed: 'prose' vs 'text' capture split is volatility-justified

**Location:** `packages/client/src/right-panel/BrowseFileDispatcher.tsx:93-136`
**Disposition:** drop (confirmed correct)

Both modes resolve to `CommentTextSurface` via the shared `textSurface` helper (lines 99-115), differing only in host class and `lineAnchored`. A naive reading might flag this as a mode flag complecting two strategies, but the Lowy analysis clears it:

- The `Capture` enum is the receptacle for "how does this view expose itself for comments."
- `'prose'` adds a genuine independent point on that axis: text-anchored but NOT source-line-addressable.
- Line-addressability is a real independent volatility axis.
- The shared wiring is unified in `textSurface`; the `.with` arms keep only the differing values local and declarative; `.exhaustive()` at line 136 forces every future renderer to make a comment decision at this one seam.

Hickey agreed. Recorded as an affirmative pass to show the seam was evaluated, not waved through.

---

#### hickey-3 / dropped after debate: `extractOffsets` re-derives Document-vs-not branching

**Location:** `packages/artifact-sdk/src/core/extractOffsets.ts:21-23`
**Disposition:** drop (observation retained; no change this PR)

Hickey raised that `extractOffsets` opens with its own `root instanceof Document ? root.body : (root as Node)` plus an `ownerDocument` ternary — a third site encoding "a Document's text lives in body, everything else is itself." Hickey proposed a `quoteRootNode(root: QuoteRoot): Node` helper colocated with the type to centralize this sub-rule.

Lowy's rebuttal was concrete and dispositive: `rootTextContent` returns a **string** (`.textContent`) while `extractOffsets` needs the walkable **node** (`root.body`, not its text) plus a **separate** `ownerDocument` derivation for `createTreeWalker`. The proposed helper would collapse only one of three sub-rules and leave two, trading a small ternary for a new exported primitive with marginal benefit. Hickey conceded this.

Furthermore, once lowy-4/hickey-2 land (`lineRangeForSelection` stops re-inlining), only two sites remain — `rootTextContent` and `extractOffsets` — both colocated in `core/` and already keyed off `QuoteRoot`. That residual is acceptable. The finding was explicitly tagged "not introduced as a regression by this PR," making it a candidate for a future pass rather than a blocker here.

---

### Summary table

|---|---|---|---|---|
| lowy | QuoteRoot receptacle widened correctly — the core boundary holds | packages/artifact-sdk/src/types.ts:13-21 | drop | — |
| lowy | Consumers expose the wires: QuoteRoot re-spelled as inline unions instead of plugging into the receptacle | packages/client/src/comments/useTextSelection.ts:85, packages/client/src/comments/highlightOverlay.ts:42 | fix | 95803ead |
| lowy | activate()'s resolvedRoot uses a fourth, narrower ad-hoc type (ShadowRoot \| HTMLElement) | packages/client/src/comments/useTextSelection.ts:208-209 | fix | b2ed09d2 |
| lowy | rootTextContent duplicated inline in lineRangeForSelection | packages/client/src/comments/useTextSelection.ts:90-93 | fix | d6015e83 |
| lowy | 'prose' vs 'text' capture split tracks a real volatility axis (line-addressability), not functional grouping | packages/client/src/right-panel/BrowseFileDispatcher.tsx:93-136 | drop | — |
| hickey | QuoteRoot concept named in SDK but not published, so three parent-side call sites re-spell the union by hand | packages/artifact-sdk/src/client/index.ts:23-29 (omission); packages/client/src/comments/highlightOverlay.ts:42; packages/client/src/comments/useTextSelection.ts:85,208 | fix | — |
| hickey | lineRangeForSelection re-inlines the rootTextContent ternary the SDK exports specifically to prevent this | packages/client/src/comments/useTextSelection.ts:90-93 | fix | — |
| hickey | extractOffsets re-derives Document-vs-not branching that rootTextContent already owns | packages/artifact-sdk/src/core/extractOffsets.ts:21-23 | drop | — |

No findings flipped disposition during the debate — all eight reached consensus in round 1. The three fixes (lowy-2/3/4, which together discharge hickey-1/2 as well) are mechanical annotation changes with no runtime impact. The two drops (lowy-1, lowy-5) are affirmative confirmations that the PR's core structural decisions are correct. The hickey-3 drop is a deferred observation about a pre-existing scatter pattern that becomes a candidate for a future cleanup once the surface area settles.

srid:
## 👮 Code-police

**13 finding(s)** across the rules / fact-check passes over 4 review sweep(s). ⚠️ **Did not reach a clean sweep within the round cap** — the worktree may still have open issues; re-run /code-police on it.

The police track worked through four sweeps, catching progressively deeper issues: the first sweep cleared surface-level dead code and DRY violations; the second and third found a genuine correctness bug (truncation chrome inside the commentable host) and a SolidJS remount hazard that was introduced while trying to fix the DRY problem; the fourth pass closed the loop on the dead-code fallout from those fixes. Several findings are interdependent — the `TruncatedBanner` refactor spawned two downstream issues (the component-in-body bug and the dead `truncated` prop) that then required their own fixes.

---

### Rules pass

#### Stale orphaned comment block before `dragSelectText`  — nit — 23850783

**Problem.** Lines 1171-1187 of `code_tab_steps.ts` are the old 17-line rationale/algorithm prelude that used to introduce the inline When-step body. When the body was extracted into `dragSelectText`, a new concise function-level comment was added immediately after (lines 1188-1190), but the old block was never deleted. The result is two consecutive comment blocks describing the same function, the first being entirely redundant dead text.

**Prescribed fix.** Delete lines 1171-1187 (the block starting with `// Drive a REAL mouse drag to select text`). The replacement comment is the correct description.

**Applied.** Deleted the 17-line stale block; the concise replacement comment at the function definition remains.

---

#### `QuoteRoot` type inlined twice instead of importing the SDK alias — cel — ot yet committed

**Problem.** `QuoteRoot` (= `Document | ShadowRoot | Element`) was introduced in `@kolu/artifact-sdk/src/types.ts` precisely to name this union once. Yet `highlightOverlay.ts` (line 42) and `useTextSelection.ts` (line 85) each inline the full union literal independently. The type is not exported from the package's public entrypoint (`client/index.ts`), so callers cannot reach it without touching package internals — but the practical consequence is that the shared concept can drift: if the union ever grows, three sites must be found and updated by hand. There are now three occurrences of `Document | ShadowRoot | Element` (definition + two inline copies), past the dry-rule-of-three threshold.

**Prescribed fix.**
1. Add `export type { QuoteRoot } from '../types';` to `packages/artifact-sdk/src/client/index.ts`.
2. Replace the return-type annotation in `highlightOverlay.ts` line 42 with `QuoteRoot` (imported from `@kolu/artifact-sdk/client`).
3. Do the same for `useTextSelection.ts` line 85.

**Applied.** Exported `QuoteRoot` from the SDK client entrypoint and replaced both inline union literals with the shared alias.

---

#### `QuoteRoot` doc comment falsely claims all members expose `createTreeWalker` — minor — not yet committed

**Problem.** The JSDoc on `QuoteRoot` says "all expose `textContent` + `createTreeWalker`". Neither `ShadowRoot` nor `Element` has a `createTreeWalker` method — that lives on `Document`. The actual code always obtains `ownerDocument` and calls `createTreeWalker` there. A maintainer trusting this comment might attempt `(root as any).createTreeWalker(...)` directly on an Element or ShadowRoot, which throws at runtime.

**Prescribed fix.** Replace the contract line with: `* one contract: all expose \`textContent\`; ShadowRoot and Element additionally expose \`ownerDocument\` (the Document the walker is created from).`

**Applied.** Doc comment updated to accurately reflect what the three shapes share; the false `createTreeWalker` claim removed.

---

#### `lineAnchored` is optional but carries required semantics at every call site — cel — ot yet committed

**Problem.** `lineAnchored?: Accessor<boolean>` in `UseTextSelectionOptions` has a silent `?? true` default. Every real call site (`CommentTextSurface`) already passes it explicitly. The one call site that omits it (the diff `CommentTextSurface` in `CodeTab.tsx`) silently inherits line-anchrored behavior. The invalid-states-unrepresentable rule flags optional booleans whose absence has semantic meaning: a future call site for a light-DOM surface that forgets this prop will silently get the wrong behavior.

**Prescribed fix.** Remove the `>` from `lineAnchored` in `UseTextSelectionOptions` and `CommentTextSurfaceProps`. Remove the `?? true` fallback. Add explicit `lineAnchored: () => true` to the diff `CommentTextSurface` in `CodeTab.tsx`. Future call sites must now make a deliberate choice.

**Applied.** `lineAnchored` made required; silent defaults removed; explicit prop added to the `CodeTab.tsx` diff surface.

---

#### Duplicated truncation banner JSX block — cel — ot yet committed

**Problem.** The block `<Show when={file.source?.truncated}><div class="px-2 py-1 text-warning text-[10px] border-b border-edge bg-surface-1/30" File truncated (exceeds 1 MB)</div></Show> appears verbatim twice in the same file: once inside `sourceRenderer` and again in the `textRenderers` markdown branch. The class string, condition  expression, and text content are all identical — this is past the dry threshold for content that must stay in sync.

**Prescribed fix.** Extract into a local helper component, e.g. `const TruncatedBanner: Component<{ show: boolean }> = (p) => <Show when={p.show}>...</Show>`, and replace both call sites.

**Applied.** Extracted into `TruncatedBanner` component; both call sites replaced.

---

#### Inaccurate comment claims `'prose'` uses `h-full` but code uses `flex-1` — minor — not yet committed

**Problem.** A comment in `BrowseFileDispatcher.tsx` reads `"prose" owns the whole pane (\`h-full\`)` but the code at line 131 passes `class: "min-h-0 w-full flex-1"` — identical to `"keys`. The comment actively misleads: a reader following it will expect different sizing behavior for prose vs. keys, while the code says they are identical. (A related duplicate finding, `police-r3-fact-check-1`, confirms the same inaccuracy from a different sweep and notes that by round 4 the fix had already landed.)

**Prescribed fix.** Either update the comment to state that both modes share `min-h-0 w-full flex-1` and differ only in `lineAnchored`, or — if `h-full w-full` was the actual intent for prose — fix the class in the `"prose"` arm to match.

**Applied.** Comment updated: both `"keys"` and `"prose"` modes now documented as sharing `min-h-0 w-full flex-1`, differing only in `lineAnchored`.

---

#### `TruncatedBanner` component defined inside `BrowseFileDispatcher` function body — jajor  — not yet committed

**Problem.** `TruncatedBanner` was initially extracted as a `Component<{show:boolean}> ` declared inside `BrowseFileDispatcher`'s render function. In SolidJS, each run of the parent component creates a new function identity for any component defined inside it. The framework sees a different component type on each render, destroys the previous DOM subtree, and remounts from scratch. This affects both uses (in `sourceRenderer` at line 159 and `textRenderers` at line 229). The `.claude/rules/solidjs.md` rule also prohibits defining components inside another component's body.

**Prescribed fix.** Lift `TruncatedBanner` to module scope above `BrowseFileDispatcher`. It has no closure dependency on props or local variables, so the lift is mechanical. Alternatively, inline the two-line `<Show>` block at each use site.

**Applied.** `TruncatedBanner` moved to module scope (line 63), above `BrowseFileDispatcher` (line 82). A follow-up fact-check finding (`police-r4-fact-check-1`) confirmed this had been resolved — the component was already at module scope by round 4.

---

### Fact-check pass

#### Truncation banner inside commentable host for `'prose'` mode ೎LAB�( L— not yet committed

**Problem.** The code (lines 143-147) explicitly documents the invariant: the truncation banner is UI chrome, not file content, and must stay outside the commentable hoss�— gtherwise a user can select "File truncated (exceeds 1 MB)" and save a comment whose quote is UI copy the agent can never find in the file. The `'#'$!t�— `.code correctly places the banner as a sibling above `CommentTextSurface`. But for `'VQQW' mode, `withComments("VQQU", file, <MarkdownRenderer truncated={...} />)` wraps the entire `MarkdownRenderer` — including its internal truncation banner — inside the commentable surface. A user can select the truncation warning text in the rendered Markdown preview and save a comment against it; the quote locator will rot immediately.

**Prescribed fix.** Either (a) move the truncation banner out of `MarkdownRenderer` and render it as a sibling above the `CommentTextSurface` host, mirroring the source renderer; or (b) pass `truncated={false}` to `MarkdownRenderer` when called from the prose capture path and render the banner separately outside the wrapper. The simplest fix matching the source renderer pattern: in `textRenderers`, wrap only the content portion in `withComments` and render the truncation banner as a sibling outside.

**Applied.** Truncation banner moved outside `CommentTextSurface` for prose mode, mirroring the source renderer pattern.

---

#### Stale doc comments still say `'Document or shadow root'` after `QuoteRoot` expansion — nit — 902c719d

**Problem.** Three doc comments describe the accepted input as "Document or shadow root": `extractOffsets.ts:2`, `extractQuote.ts:14`, and `findQuote.ts:93`. The type was widened to `QuoteRoot = Document | ShadowRoot | Element` in this PR. A maintainer reading these comments to understand what roots are valid will be misled — `Element` is silently omitted. The discrepancy violates the comment-the-non-obvious rule and will cause confusion the next time someone asks whether `Element` is intentionally supported.

**Prescribed fix.** Update the three doc comments to reference `QuoteRoot` (or spell out "Document, ShadowRoot, or Element") in all three files.

**Applied.** All three stale doc comments updated to reference `QuoteRoot` (Document, ShadowRoot, or Element).

---

#### `MarkdownRenderer` `truncated` prop is dead code — minor — 2ff71bc1

**Problem.** After the truncation-banner fix above, `TruncatedBanner` is rendered above `withComments('prose', ...)` and owns the truncation chrome for Markdown. To suppress the duplicate banner inside `MarkdownRenderer`, the only call site in `BrowseFileDispatcher.tsx` hardcodes `truncated={false}`. This makes `MarkdownRenderer`'s `truncated` prop, its `Show` guard, and the internal banner div permanently unreachable dead code. The no-dead-code rule requires aggressively removing unused code — and a hardcoded-false prop is a strong signal that the prop should not exist.

**Prescribed fix.** Remove the `truncated` prop declaration, its doc comment, the `Show` import, and the `Show` guard/banner block from `markdown.tsx`. Drop the now-redundant `truncated={false}` from `BrowseFileDispatcher.tsx:235`.

**Applied.** Dead `truncated` prop and its internal `Show` block removed from `MarkdownRenderer`; `truncated={false}` dropped from the call site.

---

#### `truncated={false}` at `MarkdownRenderer` call site lacks an inline comment — nit — 0e36966c

**Problem.** Before the dead-code cleanup was applied, a reader seeing `truncated={false}` at the call site would not know why truncation is always suppressed there. The explanation lives in a comment block 90 lines earlier (the `TruncatedBanner` definition) and is invisible at the call site. The comment-the-non-obvious rule requires a comment wherever the "why" is not deducible at a glance.

**Prescribed fix.** Add: `truncated={false} /* TruncatedBanner above owns the truncation chrome ⃉;keeps it outside the commentable host */`

**Applied.** Inline comment added at the `MarkdownRenderer` call site. (This finding was subsequently superseded by the dead-code removal in `police-r4-rules-2`, which eliminated the prop entirely.)

---

### Summary table

| ID | Severity | Finding | Files | Commit |
|---|---|---|---|---|
| r1-rules-1 | nit | Stale orphaned comment block before `dragSelectText` | `code_tab_steps.ts` | 23850783 |
| r1-rules-2 | minor | `QuoteRoot` inlined twice instead of importing SDK alias | `client/index.ts`, `highlightOverlay.ts`, `useTextSelection.ts` | — |
| r2-rules-1 | minor | `QuoteRoot` doc comment falsely claims `createTreeWalker` on all members | `types.ts` | — |
| r2-rules-2 | minor | `lineAnchored` optional but semantically required at every call site | `useTextSelection.ts`, `CommentTextSurface.tsx`, `CodeTab.tsx` | — |
| r2-fact-check-1 | **major** | Truncation banner inside commentable host for `'prose'` mode | `BrowseFileDispatcher.tsx` | — |
| r2-fact-check-2 | nit | Stale doc comments say "Document or shadow root" after `QuoteRoot` expansion | `extractOffsets.ts`, `extractQuote.ts`, `findQuote.ts` | 902c719d |
| r3-rules-1 | minor | Inaccurate comment claims `'VQQU'` uses `h-full`; code uses `flex-1` | `BrowseFileDispatcher.tsx` | — |
| r3-rules-2 | minor | Duplicated truncation banner JSX block — DRY violation | `BrowseFileDispatcher.tsx` | — |
| r3-fact-check-1 | minor | Same stale `h-full` comment (confirmed fixed by round 4) | `BrowseFileDispatcher.tsx` | ⃉; |
| r4-rules-1 | **major** | `TruncatedBanner` defined inside component body — SolidJS remount bug | `BrowseFileDispatcher.tsx` | — |
| r4-rules-2 | minor | `MarkdownRenderer` `truncated` prop is dead code | `markdown.tsx`, `BrowseFileDispatcher.tsx` | 2ff71bc1 |
| r4-rules-3 | nit | `truncated={false}` call site lacks inline comment | `BrowseFileDispatcher.tsx` | 0e36966c |
| r4-fact-check-1 | minor | Component-in-body SolidJS anti-pattern (pre-resolved by round 4) | — | — |

PR #1165, opened 2026-06-03
fix(be-review): converge code-police sweeps (#1163) + ignore apm.lock timestamp churn (#1164)

### LINKED ISSUE #1163: be-review: code-police track grinds to the sweep cap — narrow re-sweeps + severity-gate
## Problem

The **police track** is the slow pole of `/be-review` and frequently ends `consolidation-incomplete` with police status `incomplete`. Root cause is in `policeTrack` (`be-review.workflow.js`):

```
for (policeRound = 0; policeRound < POLICE_MAX_ROUNDS /* = 4 */; policeRound++) {
  // run all 3 cold passes (rules / fact-check / elegance) over the ENTIRE diff
  const reviews = parallel(passes.map(p => agent(reviewBrief..., DIFF(wt) /* read every changed file */)))
  const findings = reviews.flat()
  if (!findings.length) break        // "until a sweep is clean"
  for (f of findings) { applyAgent(f); commitAgent(f) }   // per-finding, sequential
}
```

Every sweep re-runs **all three passes over the whole diff**. After sweep 1 there's no narrowing, so sweeps 2+ **re-scan the entire diff and surface fresh *pre-existing* nits** (not regressions from the prior sweep's fixes). Since code-police always has a nit-tail, `findings.length` is rarely 0 → the loop **grinds to `POLICE_MAX_ROUNDS` and reports `incomplete`**, burning 3 review-pass agents (each re-reading the full diff) per sweep plus per-finding apply+commit agents.

This is the **same divergence pathology we already fixed in `codex-debate`** (findings growing 7→18 each round, never converging — fixed by making later rounds verify fixes + flag regressions instead of re-scanning). Police never got that treatment. It's **doubly wasteful** because police consolidates *last* and its findings are frequently *subsumed* by codex/lens in consolidation (observed repeatedly), so the extra sweeps produce work that gets dropped.

Note: model tiering already moved police review/apply to Sonnet, but **sweep *count* is the wall-clock + token driver here**, so reducing sweeps matters more than the model.

## Proposed fix (mirror the codex-debate convergence fix)

1. **Narrow re-sweeps to fix-introduced regressions, not a full re-scan.** Sweep 1 = full 3-pass review of the diff. Sweeps 2+ = re-review **only the files the previous sweep's applies touched**, and only to catch regressions / partial-fixes *those edits* introduced — NOT to hunt for more pre-existing nits sweep 1 didn't raise. This preserves the load-bearing "until clean" value (a fix can introduce or partially-resolve an issue) while removing the re-scan that causes the grind.

2. **Severity-gate the re-sweep.** Apply all of sweep 1's findings (any severity), but only trigger an additional sweep when a **blocking/major** remains open or a fix plausibly broke something. A remaining **nit** should not cost another full 3-pass sweep across the parallel worktree.

3. *(secondary)* **Batch apply per file** — one apply+commit agent per touched file instead of per finding, cutting the sequential agent count when a sweep has many findings.

4. *(secondary)* **Lower the default cap / make multi-sweep opt-in** in the gauntlet context — police is the lightest-touch track and often subsumed, so a single thorough pass is usually enough; reserve multi-sweep for when it earns its keep.

**Highest impact: #1 + #2.** Expected: police converges in ~1-2 sweeps instead of grinding to 4, dropping it from the slow pole and reducing `incomplete` outcomes.

## Acceptance

- A police sweep with only nits remaining stops after applying them (no extra full sweep).
- Sweeps 2+ review only previously-touched files, demonstrably (label/log shows the narrowed scope).
- A dogfood `/be-review` run shows police `rounds` ≤ 2 in the common case and no longer the wall-clock pole.

_Filed from the model-tiering dogfood (#1159). Sibling of the codex-debate convergence fix already merged._

### LINKED ISSUE #1164: be-review: `just ai::apm` regen churns apm.lock.yaml timestamp → tracks left unconsolidated on APM-touching PRs
## Problem

When a `/be-review` diff touches **APM skill sources** (`.apm/skills/...`), the track's author/apply agents must run `just ai::apm` to regenerate the `.claude`/`.agents` copies (`.apm` is source-of-truth; the runtime dirs must not be hand-edited). But `just ai::apm` **rewrites `apm.lock.yaml`'s `generated_at` timestamp** every time — a no-op behavioral change.

The consolidation **clean-check** (correctly) refuses to consolidate a worktree with uncommitted changes, to never silently drop a reviewer's edits. So the regenerated-but-uncommitted `apm.lock.yaml` makes the clean-check see a dirty worktree → the track is **preserved, not consolidated** → the whole run ends `consolidation-incomplete`.

Observed live on the model-tiering dogfood (#1159): the **codex track committed its fix, ran `just ai::apm`, and was then left out of consolidation** purely because of the timestamp churn:

```
note: "track left UNcommitted changes in its worktree (...be-review-...-codex); NOT consolidated and its worktree was PRESERVED...
Offending entries:
 M apm.lock.yaml"
```

So *any* gauntlet run on a PR that edits APM skills (which is exactly the kind of PR this repo's agent work produces) reliably trips this and loses a track.

## Fix options (pick one)

1. **Make `just ai::apm` deterministic** — drop the `generated_at` timestamp from `apm.lock.yaml` (or write it only when content actually changes). Cleanest: the lock stops being a source of spurious churn for everyone, not just be-review.
2. **Revert timestamp-only churn after regen** — the apply/commit agents (and any caller that runs `just ai::apm`) `git checkout -- apm.lock.yaml` when the only diff is the `generated_at` line.
3. **Tolerate it in the clean-check** — treat an `apm.lock.yaml` whose *only* change is the `generated_at` timestamp as clean (a narrow allowlist), so consolidation proceeds. Riskier (special-casing a file in the safety gate), so prefer #1 or #2.

#1 is the right long-term fix (deterministic codegen); #2 is the smallest be-review-local change if touching the APM tool is out of scope.

## Acceptance

- A `/be-review` run on a diff that touches `.apm/skills/...` consolidates all clean tracks (no track preserved solely due to `apm.lock.yaml` timestamp churn).

_Filed from the #1159 model-tiering dogfood. Sibling of #1163 (code-police sweeps)._

srid:
Closes #1163, closes #1164.

## #1163 — code-police sweep grind

code-police was the slow pole of `/be-review` and frequently ended `incomplete`. Root cause: every sweep (cap 4) re-ran all three cold passes over the **entire diff**, re-scanning for *fresh pre-existing nits* rather than narrowing — so on police's endless nit-tail it rarely hit a clean sweep and ground to the cap.

**Fix** (the same convergence shape we gave codex-debate): 
- **Sweep 1** reviews the full diff and applies every finding.
- **Sweeps 2+** re-review **only the files the previous sweep's fixes touched** (`git diff <merge-base> -- <touched files>`), looking *only* for regressions / partial-fixes those edits introduced — explicitly **not** new pre-existing issues in untouched code.
- Stops early if a sweep's fixes changed no files.

This keeps the load-bearing "a fix can introduce or only partially resolve an issue" guarantee that "until clean" exists for, while letting police **converge in ~2 sweeps** instead of grinding to 4. Sweep logs now say `full diff` vs `regression check on N touched file(s)`.

## #1164 — `apm.lock.yaml` timestamp churn → tracks left unconsolidated

`just ai::apm` rewrites `apm.lock.yaml`'s `generated_at:` timestamp on every run, so any track that touches APM skills (which forces a regen) leaves it ` M apm.lock.yaml` as pure no-op churn. The consolidation **clean-check** (correctly fail-closed) then preserves that track instead of consolidating it — so the run ends `consolidation-incomplete` for nothing. This bit the model-tiering dogfood (#1159), preserving the codex track.

**Fix:** the clean-check now, for any track showing `apm.lock.yaml` modified, checks `git diff -- apm.lock.yaml`; if the **only** changed line is `generated_at:`, it `git checkout -- apm.lock.yaml` to discard the churn before judging. Any *other* lockfile change (a real dependency edit) still counts as dirty. (Chose the be-review-local normalization over special-casing the file in the gate or changing the APM tool's codegen — option #2/#3 from the issue.)

## Tests

- Full-file syntax check (source + generated copies); `.apm`→generated in sync; `just fmt` clean.
- `policeRound` correctly scoped to the for-loop; both fixes verified present.
- The natural end-to-end check is a `/be-review` dogfood — police should now report `rounds ≤ 2` and the run should consolidate cleanly even though it touches APM skills.


srid:

This review track tightened the `be-review` workflow's scope-building logic and fixed two substantive correctness bugs — a misleading early-exit log message and broken git commands in the apm.lock normalization prompt — while four of the seven picks required careful three-way reconciliation to preserve the best ideas from each track without losing any.

---

#### Summary table

|---|---|---|---|
| codex | 0b1c74ba | `clean` | 63cf7e31b |
| lens | 29387580 | `reconciled` | b816d7f6b |
| lens | 2a2edc5f | `reconciled` | 2b1abee62 |
| lens | ebd3bffc | `clean` | 0af4c3625 |
| police | 36e4da8f | `reconciled` | 05998a747 |
| police | 7b17999d | `reconciled` | 9918f319d |
| police | 54ddf86b | `reconciled` | 7b74b70d1 |

Final HEAD: 7b74b70d1e83f3c393e378e76ed6d33ec0e8e88e

---

#### Codex track

**0b1c74ba → 63cf7e31b** — `clean`

Codex landed without conflict. Among its contributions that shaped later reconciliations: it inlined a richer scope prompt for non-first sweeps (restricting findings to regressions introduced by the previous sweep's edits, with explicit "NOT confined to reading only those files" language) and fixed the git commands in the apm.lock normalization prompt to use `git diff HEAD --` and `git checkout HEAD --`, which correctly cover staged changes in addition to unstaged ones.

---

#### Lens track

**Reconciliation 1 — 29387580 → b816d7f6b** (file: `.apm/skills/be-review/be-review.workflow.js`)

The tension here was between two valid but orthogonal improvements to the same line of code. Codex had inlined the git diff invocation directly into the scope string, threading in substantially richer prompt text around it ("Limit the DIFF to the touched files... But you are NOT confined to reading only those files... Raise a finding ONLY when the problem is rooted in the previous sweep's edits"). Lens took a structural approach instead: it extracted a `relArgs` helper variable and called `DIFF(wt, relArgs)`, which avoids the fragile inline quoting but strips out the richer instructions.

Neither was complete on its own. The merge keeps lens's `relArgs` extraction and `DIFF(wt, relArgs)` call form — the cleaner structural approach  — while wrapping codex's full scope prompt text around that call. Both improvements are preserved; neither is lost to the other.

**Reconciliation 2 — 2a2edc5f → 2b1abee62** (file: `.apm/skills/be-review/be-review.workflow.js`)

Lens followed up by extracting `relPaths`/`relPathspec` helper utilities and switching the scope to `DIFF(wt, [relPathspec(wt, touchedFiles)])` — a single canonical quoting path. At this point HEAD (post-reconciliation 1) still had the `relArgs` variable from the first lens pick, plus codex's extended scope instructions. The merge replaces `DIFF(wt, relArgs)` with `DIFF(wt, [relPathspec(wt, touchedFiles)])`, adopting the cleaner single-source quoting helper from lens's second commit, while leaving codex� extended scope text entirely intact around the call site.

**3rd lens commit — ebd3bffc → 0af4c3625** — `clean`, landed without conflict.

---

#### Police track

**Reconciliation 1 — 36e4da8f → 05998a747** (file: `.apm/skills/be-review/be-review.workflow.js`)

Both codex and police independently identified and fixed the same bug: when a sweep produces no file changes to re-review, the existing code logs "hit round cap" — a misleading message that implies the review was limited by the configured maximum rounds rather than halted due to a no-op apply. This is a real correctness issue for anyone diagnosing why a review ended early.

The two tracks fixed it differently. Codex introduced a boolean `noOpApplyExit = true` flag with a ternary log expression. Police instead introduced a string variable `earlyBreakReason = 'no-touched-files'` with an if/else log structure, and also (accidentally) left a dead `earlyBreakReason = 'cap'` assignment at the cap-exit path.

The merge keeps *both* `noOpApplyExit = true` and `earlyBreakReason = 'no-touched-files'` at the break point, keeping the variables consistent with each other. The dead `earlyBreakReason = 'cap'` assignment (which police commit 3 would remove anyway) was excluded from the start. Police's cleaner /if/else log structure is adopted, but codex� more descriptive message text —  "stopped early ℔ last sweep's findings produced no file changes to re-review; reporting incomplete with findings still open" — is preserved over police's terser wording.

**Reconciliation 2 — 7b17999d → 9918f319d** (file: `.apm/skills/be-review/be-review.workflow.js`)

Both tracks also touched the apm.lock normalization prompt, but they fixed different layers of the same problem. The original prompt used `git diff --` and `git checkout --` without `HEAD`, which silently misses staged changes. Codex fixed the git commands to `git diff HEAD --` and `git checkout HEAD --` — the substantive correctness fix — but kept the vague "content line" terminology. Police fixed only the terminology ("added or removed lines (starting with + or -, excluding the +?+/--- file headers)" is unambiguously clear) but left the original broken commands in place.

Neither fix alone is sufficient: police's terminology improvement without codex's command fix leaves the staged-change bug; codex� command fix without police's terminology improvement leaves the ambiguous phrasing. The merge keeps codex's `git diff HEAD --` / `git checkout HEAD --` commands as the substantive fix, adopts police's "added or removed lines" terminology, and retains codex� "(staged OR unstaged)" context note.

**Reconciliation 3 — 54ddf86b → 7b74b70d1** (file: `.apm/skills/be-review/be-review.workflow.js`)

Police's third commit is cleanup: it removes the dead `earlyBreakReason = 'cap'` assignment (which was never written in the reconciled HEAD anyway) and narrows the JSDoc type from `'cap' | 'no-touched-files' | null` to `'no-touched-files' | null`. The only genuine conflict was a comment wording discrepancy: HEAD's comment read "If we ended with findings still open" (accurate for all early-exit modes) while police said "If we exhausted the round cap" (accurate only for the cap-exit case and therefore stale after the police track's own restructuring). HEAD's more general phrasing was kept.

srid:
## 🤖 Codex ⇄ Claude debate

All four prior findings are resolved. I verified the generated skill copies are byte-identical to the `.apm` source, `apm.lock.yaml` has no diff against the review base, there are no untracked files, and `git diff --check` is clean.

---

This track tightened two correctness gaps in the new be-review police resweep logic — one that could silently miss regressions introduced by a fix, and one that could silently miss staged-only lockfile churn — plus cleaned up a misleading log message and removed a trivial timestamp bump from the diff. All findings were accepted without dispute and fixed in a single round.

### Round 1 — codex approved ❌ · 4 open · 63cf7e31b820ba4e69c165214f09f72e1fef41fb

The branch looked structurally sound: generated skill copies matched the `.apm` source and the scope was appropriate. Codex's concern centred on two correctness holes — one in the police resweep prompt design, one in the `apm.lock.yaml` normalization path — alongside two hygiene issues.

**Findings by severity**

| ID | Severity | Location | Core problem |
|----|----------|----------|--------------|
| F1 | major | `be-review.workflow.js:553` | Resweep read-scope too narrow — regression validation weakened |
| F2 | minor | `be-review.workflow.js:1281` | `apm.lock.yaml` normalization misses staged churn |
| F3 | nit | `be-review.workflow.js:627` | No-op apply exit logs misleading "hit round cap" message |
| F4 | nit | `apm.lock.yaml:2` | Timestamp-only `generated_at` bump committed on this branch |

**F1 — major: resweep read-scope too narrow**

The sweeps-2+ police prompt limited both the `git diff` _and_ the read scope to the files touched by the previous sweep. Codex's argument: a fix to file A may break callers, contracts, or generated outputs in untouched files, and the reviewer needs to inspect those to validate that nothing was regressed. Preventing that read while still claiming the re-review guarantees regression detection is a false promise.

The suggested fix was a precise split: keep the path-limited diff (to avoid re-raising pre-existing nits in unrelated files) but lift the read restriction so the reviewer can inspect any surrounding/caller/dependent file. Findings should still be gated to problems _rooted_ in the previous sweep's edits, even if the breakage surfaces elsewhere.

Claude agreed in full. The rewrite at line 560 adds an explicit instruction: "Read freely any surrounding, caller, dependent, contract, or generated-output files you need … to judge whether those edits broke something." Findings are still restricted to regressions or partial fixes traceable to the previous sweep's touched files.

**F2 — minor: `apm.lock.yaml` normalization misses staged churn**

The CLEANLINESS CHECKER prompt told the agent to run `git diff -- apm.lock.yaml` and then `git checkout -- apm.lock.yaml`. Both commands operate only on the working tree, not the index. If the timestamp-only churn was staged (but not yet committed), `git diff --` would show nothing and `git checkout -- file` would leave the staged change intact, causing the track to appear dirty and triggering a false-positive incomplete result — the exact outcome the normalization block was added to prevent.

The fix is mechanical: use `git diff HEAD -- apm.lock.yaml` to expose staged-only changes, and `git checkout HEAD -- apm.lock.yaml` to reset both index and working tree (equivalent to `git restore --staged --worktree`). Claude agreed and applied this at line 1331.

**F3 — nit: misleading round-cap log on no-op apply exit**

When the apply step returns no changed files, the loop breaks early with `lastRoundFindings > 0`. The subsequent log emitted the "hit round cap" message unconditionally, conflating a no-op apply exit with genuine cap exhaustion (`POLICE_MAX_ROUNDS`). The `incomplete` status was still correct — the worktree wasn't verified clean — but the reason logged was wrong.

Claude added `let noOpApplyExit = false` (line 543), sets it at the break (line 635), and branches the exit log at line 650: no-op exits now say "stopped early — last sweep's findings produced no file changes to re-review; reporting incomplete with findings still open" rather than the round-cap message.

**F4 — nit: timestamp-only `apm.lock.yaml` bump in the branch diff**

The branch included a `generated_at:` update committed in d71cf84a that carries no dependency changes — precisely the class of churn the new normalization code is designed to detect and discard. Codex noted the irony: the PR that adds no-op-churn handling was itself carrying no-op churn. Claude restored the working-tree `apm.lock.yaml` to the base commit (4aa468de) content without creating a new commit (as instructed), noting that `git diff 4aa468de -- apm.lock.yaml` is empty. After `just ai::apm` re-bumped the timestamp during regeneration, the file was restored again to confirm the final tree carries zero net lockfile change against base.

**Claude's dispositions**

| Finding | Disposition | Fixed at |
|----------|-------------|----------|
| F1 | fixed | `be-review.workflow.js:560` |
| F2 | fixed | `be-review.workflow.js:1331` |
| F3 | fixed | `be-review.workflow.js:543`, `635`, `650` |
| F4 | fixed | working-tree restore (no new commit) |

No findings were disputed. Claude agreed with all four and fixed each in this round.

---

### Round 2 — codex approved ✅ 2· 0 open

Codex performed independent verification: generated skill copies in `.claude/` and `.agents/` are byte-identical to the `.apm` source; `git diff 4aa468debeb51a07b4b58d7091f28ba4d95d396 -- apm.lock.yaml` is empty; no untracked files; `git diff --check` is clean.

**Resolution summary**

| ID | Original severity | Verification |
|----|-------------------|------------|
| F1 | major | Sweeps-2+ prompt now permits unrestricted reads for regression validation while keeping findings scoped to the previous sweep's causation. Confirmed at line 560. |
| F2 | major | Cleanliness checker now uses `git diff HEAD` + `git checkout HEAD` to cover staged churn. Confirmed at line 1331. |
| F3 | minor | `noOpApplyExit` flag prevents the no-op break from emitting a false "hit round cap" log. Confirmed at line 650. |
| F4 | nit | `apm.lock.yaml` diff against base commit is empty. Confirmed by `git diff`. |

**Codex on Claude's rebuttal:** No disputes to address; CLAUDE agreed with all four findings, and I verified the corresponding fixes.

The debate closed cleanly in one fix-and-verify cycle with zero contested findings.

srid:

The debate converged in 2 rounds on 8 findings (4 from each lens) and produced 3 applied commits — all targeting the new multi-sweep regression loop introduced by this PR. The core structural tension: the PR added a touched-file scoping path for sweeps 2+ that partially reimplemented existing receptacles rather than*extending them, and introduced a new silent failure mode where a stray `filesChanged` path can make the regression check review nothing while reporting success.

**Outcome:** `consensus` after 2 round(s). Independent review: lowy=4, hickey=4

---

### Findings summary

|---|---|---|---|---|
| lowy | Touched-files re-review bypasses the DIFF() receptacle | .apm/skills/be-review/be-review.workflow.js:551-553 | fix | b816d7f6b3cb42afaef47eab41da8b423975900b |
| lowy | Shell-quoting idiom for file lists duplicated verbatim | .apm/skills/be-review/be-review.workflow.js:550,553 | fix | 2b1abee625c3e88ec5b0c896f0458e50924f63a8 |
| lowy | apm.lock normalization couples the cleanliness gate to one lockfile's churn | .apm/skills/be-review/be-review.workflow.js:1279-1283 | drop | — |
| lowy | Early-break on no touched files is correct and well-encapsulated | .apm/skills/be-review/be-review.workflow.js:624-627 | drop | — |
| hickey | apm.lock normalization complects the clean-check judge with a mutation | .apm/skills/be-review/be-review.workflow.js:1279, | drop | — |
| hickey | Worktree-path→relative + shell-quote normalization duplicated across call sites | .apm/skills/be-review/be-review.workflow.js:550 | fix | (merged with lowy-2 / 2b1abee625) |
| hickey | `filesChanged` absolute-path shape is an unenforced cross-agent contract the new scoping silently depends on | .apm/skills/be-review/be-review.workflow.js:594 | fix | 0af4c362502ed8257951267c9adc94f3b23a7194 |
| hickey | Sweep-scope branch complects two unrelated knobs on one `firstSweep` flag | .apm/skills/be-review/be-review.workflow.js:551 | drop | — |

---

### Per-finding deep dive

#### lowy-1 — Touched-files re-review bypasses the DIFF() receptacle (fix, b816d7f6b3cb42afaef47eab41da8b423975900b)

**What lowy found.** The project has a single stable receptacle for "how do we tell a cold agent to scope its review of a worktree": the `DIFF(wt)` function at line 221, already reused by the tiny-diff probe, codex, and lens tracks. The sweep-2+ branch does not plug into it — instead it hand-rolls a parallel `git -C ${wt} diff ${mergeBase} -- ...` call plus bespoke "Read those files (ABSOLUTE paths...)" prose at 551-553. The two implementations now drift independently: `DIFF(wt)`'s "Ignore the gitignored scratch dirs" and "plus enough surrounding code to judge it in context" tail clauses are both silently absent from the regression branch.

**Cross-examination.** Both lenses agreed immediately (round 1). Hickey reinforced that the missing "surrounding code" clause is not cosmetic — a regression is often only visible in the caller, not the edited line, so stripping that guidance directly weakens the thing sweep 2+ is meant to do. The electricity test is clean: the receptacle is the review-scope instruction; the volatile implementations are full-diff vs. touched-files. The touched-files wire is soldered at the call site rather than behind the socket.

**Agreed plan.** Parameterize `DIFF` with an optional `paths` arg: when non-empty, append `-- ${paths.join(' ')}` to the range; preserve the shared status/read/surrounding-code/scratch-ignore tail in both branches. Sweep 2+ calls `DIFF(wt, relArgs)` via the `relPathspec` helper from lowy-2. The `firstSweep` ternary at line 558 (full-review vs. regression-only framing) stays at the call site — that is genuinely different instruction belonging there, not in `DIFF`.

---

#### lowy-2 / hickey-2 — Shell-quoting idiom for file lists duplicated verbatim (fix, 2b1abee625c3e88ec5b0c896f0458e50924f63a8)

**What both lenses found (converged).** Both lowy and hickey independently flagged byte-identical duplication of the abs→rel→shell-quote transform. Line 550 computes `touchedFiles.map((f) => f.replace(`${wt}/`, '').replace(/^\\/+/, ''))` and line 553 applies the POSIX single-quote escaping + join. Lines 657-658 inside `commitFix` do exactly the same thing — same map, same quoting expression, same join. One derivation ("absolute worktree path → shell-safe git pathspec relative to wt") in two call sites that must be edited in lockstep by memory.

**Cross-examination.** Both lenses agreed in round 1; hickey-2 and lowy-2 merged into a single fix. The risk is real and load-bearing: both sites feed `git -C ${wt} ... -- <files>`, so a quoting change (e.g. a filename with a metachar the single-quote escaping doesn't cover) must land in both by recall.

**Agreed plan.** Extract one helper near `commitFix`:

```js
const relPathspec = (wt, files) =>
  files
    .map((f) => f.replace(`${wt}/`, '').replace(/^\/+/, ''))
    .map((f) => `'${f.replace(/'/g, "'\\''")}'`)
    .join(' ');
```

Call it from both `commitFix` (replacing lines 657-658's `fileArgs`) and from the sweep-scope branch. The lowy-1 `DIFF(wt, paths)` receives the already-relative-and-quoted args from this helper.

---

#### hickey-3 — `filesChanged` absolute-path shape is an unenforced cross-agent contract the new scoping silently depends on (fix, 0af4c362502ed8257951267c9adc94f3b23a7194)

**What hickey found.** This is the most consequential structural finding of the track. Before this PR, `impl.filesChanged` (line 594) had exactly one consumer: `commitFix`'s `git add -- <path>` at line 668, where a non-wt path fails loudly (git add errors, commit is empty, the failure is visible). This PR adds a second consumer: `sweepTouched` at line 595 → `touchedFiles` at line 624 → the sweep-2+ scope diff at line 553. The only normalization is `.replace(`${wt}/`, '')`. `IMPL_SCHEMA` types `filesChanged` as bare `array of string`; the apply prompt only asks for absolute paths — convention, not a type guarantee. If an apply agent returns `./foo`, `../sibling/foo`, or a sibling-worktree absolute path, the replace is a silent no-op, the diff scopes against a path git doesn't recognise, the regression check reviews nothing, and the sweep reports clean. The PR converts a loud failure into a silent false-convergence in the very safety property it is meant to strengthen.

**Cross-examination.** Round 1: both lenses voted fix, but did not initially agree on the filter predicate. Hickey's original suggestion ("after stripping `${wt}/`, drop anything starting with `/` or `.`") was wrong. Every file touched by this very PR lives under `.apm/`, `.claude/`, `.agents/` — all valid wt-relative paths beginning with a dot. Hickey's predicate would discard every legitimate edit and silently empty the regression scope ↔ the exact failure mode the finding is meant to close. Round 2: hickey conceded the predicate error outright and endorsed lowy's amended predicate: reject only paths that, after stripping `${wt}/`, still start with `/` (absolute path not under wt) or start with `./` or `../` (relative path that escapes wt); keep all leading-dot directory paths like `.apm/...`.

**Agreed plan.** At a single ingestion point right after line 594, normalize + validate `filesChanged` once using the `relPathspec` pre-quote core. Then apply a mandatory filter: keep an entry only if, after stripping `${wt}/`, it neither still starts with `/` nor starts with `./` or `../`. Do not reject leading-dot paths. Feed the canonical relative list to both `sweepTouched` and `commitFix`. If the filter drops any entry for a finding, log it explicitly so a stray-path apply cannot silently empty the regression scope.

**This finding required 2 rounds** — it is the only one that did not converge immediately. The flip was not in disposition (both lenses said fix from round 1) but in the predicate detail, which Hickey itself caught and corrected.

---

#### lowy-3 / hickey-1 — apm.lock normalization: structural tension, both dropped

These two findings both target lines 1279-1283 but from different lenses, and both were dropped — making them the most interesting pair in the debate.

**lowy-3 (Lowy lens).** The clean-check is the receptacle for the "is this worktree fully clean" volatility axis. The new paragraph welds a second, unrelated axis into it: "which generated files have benign timestamp-only churn to discard before judging cleanliness." These are two independent axes of change. The concern is valid in principle ↔ when a second timestamp-only generated file appears, the gate prose will grow another bespoke paragraph rather than the normalization living behind a stable predicate. **Dropped** because, by Lowy's own variable-vs-volatile bar, one file with one benign-churn field today is variable, not yet volatile. No concrete second instance exists to name. Building a benign-churn registry now would be the over-engineered boundary Lowy warns against.

**hickey-1 (Hickey lens).** The clean-check prompt previously ended "Do not edit anything." and was a pure observer. The #1164 change removes that constraint and braids a gated mutation (`git checkout -- apm.lock.yaml`) into the judge. Observe and mutate are now interleaved in one agent. The destructive action is hidden inside a read-only-sounding safety gate, gated on a free-text heuristic the same agent evaluates. The other narrow `mechanicalPreamble` agents (SETUP RUNNER, COMMITTER) keep observe/mutate cleanly separated; this one no longer does. **Dropped** on reflection because: (1) the mutation is narrowly gated and the agent re-runs `git status --short` after the discard so the verdict reflects the normalized tree; (2) the proposed fix — a separate pre-clean-check agent doing only the diff-and-conditional-checkout — adds a whole new cold-context per-track round-trip (latency + cost across every live track) to de-complect a single bounded edit on a single known file; that operational cost exceeds the harm of inlining for one file; (3) splitting the agent does not make the same timestamp-only heuristic more correct — it only relocates it. The noted trigger for revisiting both lowy-3 and hickey-1: when a second timestamp-only generated file appears.

**What the debate added.** Lowy's note acknowledges hickey's structural observation is real ("the invariant erosion is real and worth a code comment") while holding the drop disposition on operational grounds. Both lenses converged to drop independently in round 1 — no flip.

---

#### lowy-4 — Early-break on no touched files is correct and well-encapsulated (drop — positive verdict)

This is a positive verdict, not a defect report. Lines 624-627: `touchedFiles = [...sweepTouched]` followed by `if (!touchedFiles.length) break` encapsulates exactly one fact — a regression sweep over zero touched files is a no-op. `sweepsRun` stays exact because it increments at the top of the loop (line 548) before the break. The inter-sweep interface is the single atomic `touchedFiles` set, a stable contract; the narrowing strategy could be swapped without disturbing the apply/commit machinery. Both lenses confirmed in round 1.

---

### Debate dynamics

| round | finding | lowy | hickey | agreed |
|---|---|---|---|---|
| 1 | lowy-1 | fix | fix | yes |
| 1 | lowy-2 | fix | fix | yes |
| 1 | lowy-3 | drop | drop | yes |
| 1 | lowy-4 | drop | drop | yes |
| 1 | hickey-1 | drop | drop | yes |
| 1 | hickey-2 | fix | fix | yes |
| 1 | hickey-3 | fix | fix | **no** (predicate disagreement) |
| 1 | hickey-4 | drop | drop | yes |
| 2 | hickey-3 | fix | fix | yes (hickey conceded predicate error) |

One finding required a second round. The disagreement was not on disposition but on the filter predicate inside the fix plan — hickey's original predicate would have silently emptied the regression scope for every file this PR actually edits (all under dotdirs). Lowy caught the error; hickey conceded outright in round 2.

No findings flipped disposition (drop→fix or fix→drop) during debate. Three of the four fix findings produced applied commits. hickey-2 merged into lowy-2's commit (2b1abee625) as the same de-duplication fix.

srid:
## 👮 Code-police

**4 finding(s)** across the rules / fact-check passes over 4 review sweep(s).

The police track focused on diagnostic accuracy in the new parallelized police loop: two passes caught the same misleading log message (the cap-exhaustion message firing on an unrelated early-break path), and subsequent sweeps cleaned up the residue left by the fix itself — a dead variable assignment and imprecise prompt wording in the apm.lock diff detector.

---

### Rules pass

**Finding 1 — Misleading 'hit round cap' log fires for no-touched-files early-break** (minor)

**Problem.** The `!reachedClean` branch at line 627 emits `"police: hit round cap (4) with findings still open"` for two structurally different exit conditions: (a) the loop counter genuinely exhausted `POLICE_MAX_ROUNDS`, and (b) all apply agents produced zero file changes and the loop broke early. In case (b) the cap was never hit — `sweepsRun` could be 1. An operator reading the log to diagnose an `incomplete` track gets a false signal about how many rounds ran and why the loop stopped.

**Prescribed fix.** Track a separate `earlyBreakReason` variable (typed a'cap' | 'no-touched-files' | null`) set at each break site, then branch on it: emit `"police: all apply agents changed no files — cannot regress-check, reporting incomplete"` for the no-touched-files path and `"police: hit round cap (POLICE_MAX_ROUNDS) with findings still open — reporting incomplete"` for the genuine cap path.

**Applied.** `earlyBreakReason` tracking and two distinct log messages introduced at lines 548, 630, 634, and 644–653 of be-review.workflow.js. Commit: 05998a747828d6b6450ef57f42cebd524ffa27a7

---

**Finding 2 — apm.lock diff detection: 'content line' is non-standard terminology for haiku model** (minor)

**Problem.** The clean-check prompt instructs the haiku model to restore `apm.lock.yaml` "if the ONLY changed content line is `generated_at:`". The term "content line" is not standard git diff vocabulary — git calls `+`/`-` prefixed lines "added or removed lines", while context lines (space-prefixed) are also technically content. A haiku-grade model parsing raw diff output could match the `generated_at:` string appearing in an unchanged context line, producing false negatives where legitimate apm.lock changes are silently discarded.

**Prescribed fix.** Replace the ambiguous phrase with: "if the ONLY added or removed lines (starting with `+` or `-`, excluding the `+++`/`---` file headers) contain `generated_at:`".

**Applied.** Prompt updated verbatim to the prescribed wording, eliminating the ambiguity for the model parsing diff output. Commit: 9918f319d6178268cfefbe31bfe7bdd1f32cafbc

---

**Finding 3 — Dead 'cap' string assignment in earlyBreakReason — value never tested** (minor)

**Problem.** After the fix in finding 1 landed, the assignment `earlyBreakReason = "cap"` at line 634 sets a string value that no subsequent code ever reads or tests. The only consumer is `if (earlyBreakReason === "no-touched-files")` — its `else` arm fires identically for both `"cap"` and `null`. The JSDoc declares three meaningful states (`'cap' | 'no-touched-files' | null`), but `'cap'` is runtime-indistinguishable from `null`. The variable lies about carrying meaningful state, and the dead assignment is a no-dead-code violation introduced by the earlier fix commit.

**Prescribed fix.** Either (a) remove the assignment and simplify the JSDoc to `'no-touched-files' | null`, letting the else branch self-document as the cap case via its log message; or (b) if the cap/null distinction matters in future, add a `=== 'cap'` branch now rather than leaving an untested union member.

**Applied.** Dead `earlyBreakReason = "cap"` assignment removed; JSDoc union type simplified to `'no-touched-files' | null`. Commit: 7b74b70d1e83f3c393e378e76ed6d33ec0e8e88e

---

### Fact-check pass

**Finding 4 — Misleading 'hit round cap' log fires on early-break path too** (minor, duplicate)

**Problem.** Same root cause as rules finding 1, independently surfaced by the fact-check pass: the `!reachedClean` guard at line 638 emits the cap-exhaustion log in the no-touched-files early-break case. The `status='incomplete'` verdict is correct in both cases, but the log is factually wrong — it implies multiple review rounds exhausted a budget when in fact only one round ran and the apply agents committed nothing.

**Applied.** No additional changes were needed; the fix committed for finding 1 (earlyBreakReason tracking + split log messages) had already resolved this path by the time this pass ran. The fact-check pass confirmed the fix was in place at lines 548, 630, 634, and 644–653. Commit: —

---

### Summary table

| Severity | Finding | File | Commit |
|---|---|---|---|
| minor | Misleading 'hit round cap' log fires for no-touched-files early-break | be-review.workflow.js | 05998a747828d6b6450ef57f42cebd524ffa27a7 |
| minor | Misleading 'hit round cap' log fires on early-break path too (fact-check, confirmed fixed) | — | — |
| minor | apm.lock diff detection: 'content line' is non-standard terminology for haiku model | be-review.workflow.js | 9918f319d6178268cfefbe31bfe7bdd1f32cafbc |
| minor | Dead 'cap' string assignment in earlyBreakReason — value never tested | be-review.workflow.js | 7b74b70d1e83f3c393e378e76ed6d33ec0e8e88e |

PR #1170, opened 2026-06-04
perf(be-review): parallelize + split comment posting (faster, cheaper, same guarantees)

srid:
Two costs hid in how `/be-review`'s Report phase posted PR comments. This PR removes both, then the review gauntlet (run on this very PR) hardened the design.

## The wins

1. **Serial → concurrent posting.** The phase authored bodies in parallel but posted them through a serial loop; each `gh pr comment` agent takes minutes, so N tracks stacked to N×minutes. Now author+post run concurrently.
2. **No re-ingested body.** The old split paid for each rich body twice — a reporter emitted it, then a separate poster re-ingested it as a ~50-60k-token base64 blob just to write a file and run `gh`. Now the reporter writes its body straight to a **draft file** (via the Write tool — no shell-escaping), so the body crosses an agent boundary once. Confirmed in a real run: the Report phase used **~13.5k output tokens** total vs ~50-60k *per* comment previously.

## Final design (hardened by the gauntlet)

Folding author+post into one agent (my first cut) traded away three guarantees; the codex⇄claude debate caught them and re-split at the side-effect boundary while keeping the perf win:

- **STAGE 1 `authorDraft`** — author-only agent: writes the rich body to a draft file, runs **no `gh`**, returns only `{firstLine, bytes}`. A throw falls back to the baseline *before any post* → no duplicate comment.
- **STAGE 2 `postFile`** — narrow mechanical poster: **independently re-validates the actual file** (`test -s`, `wc -c ≤ 60000`, `head -1 | diff` vs a base64-decoded header) before `gh`, posting by path only. Fails → `INVALID_DRAFT` sentinel → baseline fallback.
- The side-effectful poster is off the untrusted-body path (it sees only a file path); the small deterministic baseline remains the never-blank fallback via the base64 poster.

Net: same one-comment-per-track contract, exact-header PR anchors, and injection boundary as before — now faster and far cheaper.

### DIFF
--- a/.agents/skills/be-review/SKILL.md
+++ b/.agents/skills/be-review/SKILL.md
@@ -102,20 +102,25 @@ to `cd`.
   deterministic string-builder comments. By **default** (`richComment: true`) each
   comment is authored by a per-track **reporter agent** (narrative + tables +
   reasoning, synthesized from the track's full structured result) rather than the
-  terse builders. That agent **writes its own body and posts it** in one shot, so
-  the large body never crosses a second agent as a base64 blob. The builders remain
-  the **baseline** the agent improves and the **fallback** if the agent throws; a
-  trivial track (track-error / clean / no findings) skips the agent and posts the
-  baseline directly. Use this flag when you want the fast, no-agent comments.
+  terse builders. That agent **writes only a DRAFT FILE** (it never runs `gh`) and
+  returns metadata; the workflow then **mechanically validates the draft file**
+  (nonempty, exact header line, byte cap) and a **narrow mechanical poster** posts it
+  by path — so the large body never crosses a second agent as a base64 blob, and the
+  side-effecting `gh` is split off from the fallible authoring step. The builders
+  remain the **baseline** the agent improves and the **fallback** if the agent
+  throws or the draft fails validation; a trivial track (track-error / clean / no
+  findings) skips the agent and posts the baseline directly. Use this flag when you
+  want the fast, no-agent comments.
 - **Cost / model tiers** (`model` / `synthModel` / `mechModel`): the orchestrator
   and child workflows run each agent on the cheapest model that does its job, so a
   run doesn't pay Opus rates for `git`/`gh` shuffling. Defaults: **`model: opus`**
   for deep reasoning (the lens lenses — load-bearing — + claude-author + lens
   apply), **`synthModel: sonnet`** for synthesis (the reporter agents — which now
-  author AND post their own comment — the cherry-pick/reconcile, the police
-  review/apply passes — code-police is natively Sonnet anyway), **`mechModel:
-  haiku`** for mechanical agents (setup, every commit, cleanup, the baseline-
-  fallback comment poster, status/HEAD checks, merge-base + codex runner).
+  author a draft FILE only, leaving the `gh` post to a mechanical poster — the
+  cherry-pick/reconcile, the police review/apply passes — code-police is natively
+  Sonnet anyway), **`mechModel: haiku`** for mechanical agents (setup, every commit,
+  cleanup, the draft-validating rich-comment poster and the baseline-fallback
+  poster, status/HEAD checks, merge-base + codex runner).
   Override any tier via args. The run reports a **`tokensByPhase`** breakdown
   (output tokens, from `budget.spent()` on the shared turn counter) bucketed by
   each phase's wall-clock window — NOT isolated to that phase's agents: concurrent
@@ -171,10 +176,10 @@ It runs five phases the user can watch via `/workflows`: **Setup** (fan out one
 detached worktree per track under `.worktrees/`), **Tracks** (the three gauntlets
 run concurrently to consensus), **Consolidate** (cherry-pick each track's commits
 onto the branch, reconciling overlap), **Report** (a per-track **reporter agent**
-authors a detailed PR comment from each track's structured result — narrative +
-tables + reasoning — plus the consolidation ledger; `richComment: false` falls back
-to the terse deterministic builders), **Cleanup** (tear down the worktrees). It
-returns:
+authors a detailed PR comment as a DRAFT FILE from each track's structured result —
+narrative + tables + reasoning — which the workflow validates and a mechanical poster
+posts by path, plus the consolidation ledger; `richComment: false` falls back to the
+terse deterministic builders), **Cleanup** (tear down the worktrees). It returns:
 
 ```
 { status,                  // 'done' | 'consolidation-incomplete' | 'consolidation-aborted' | 'no-commit' | 'setup-failed'
--- a/.agents/skills/be-review/SKILL.md
+++ b/.agents/skills/be-review/SKILL.md
@@ -102,17 +102,20 @@ to `cd`.
   deterministic string-builder comments. By **default** (`richComment: true`) each
   comment is authored by a per-track **reporter agent** (narrative + tables +
   reasoning, synthesized from the track's full structured result) rather than the
-  terse builders. The builders remain the **baseline** the agent improves and the
-  **fallback** on empty/invalid output; a trivial track (track-error / clean / no
-  findings) skips the agent. Use this flag when you want the fast, no-agent comments.
+  terse builders. That agent **writes its own body and posts it** in one shot, so
+  the large body never crosses a second agent as a base64 blob. The builders remain
+  the **baseline** the agent improves and the **fallback** if the agent throws; a
+  trivial track (track-error / clean / no findings) skips the agent and posts the
+  baseline directly. Use this flag when you want the fast, no-agent comments.
 - **Cost / model tiers** (`model` / `synthModel` / `mechModel`): the orchestrator
   and child workflows run each agent on the cheapest model that does its job, so a
   run doesn't pay Opus rates for `git`/`gh` shuffling. Defaults: **`model: opus`**
   for deep reasoning (the lens lenses — load-bearing — + claude-author + lens
-  apply), **`synthModel: sonnet`** for synthesis (the reporter agents, the
-  cherry-pick/reconcile, the police review/apply passes — code-police is natively
-  Sonnet anyway), **`mechModel: haiku`** for mechanical agents (setup, every
-  commit, cleanup, comment posters, status/HEAD checks, merge-base + codex runner).
+  apply), **`synthModel: sonnet`** for synthesis (the reporter agents — which now
+  author AND post their own comment — the cherry-pick/reconcile, the police
+  review/apply passes — code-police is natively Sonnet anyway), **`mechModel:
+  haiku`** for mechanical agents (setup, every commit, cleanup, the baseline-
+  fallback comment poster, status/HEAD checks, merge-base + codex runner).
   Override any tier via args. The run reports a **`tokensByPhase`** breakdown
   (output tokens, from `budget.spent()` on the shared turn counter) bucketed by
   each phase's wall-clock window — NOT isolated to that phase's agents: concurrent
--- a/.agents/skills/be-review/be-review.workflow.js
+++ b/.agents/skills/be-review/be-review.workflow.js
@@ -823,9 +823,11 @@ for (const t of dispatchable)
   log(`Track ${t}: ${tracks[t].status || "unknown"}`);
 
 // ---------------------------------------------------------------------------
-// PR-comment builders + poster (used by the Report phase). Bodies are built
-// deterministically in JS from the structured track results; a thin mechanical
-// agent just writes each body to a scratch file and posts it with `gh pr comment`.
+// PR-comment builders + poster (used by the Report phase). The builders here build
+// each track's DETERMINISTIC baseline body in JS from its structured result; the
+// Report phase either improves that baseline with a rich author agent (which Writes
+// its body to a draft file) or posts the baseline directly, and a thin mechanical
+// agent runs the actual `gh pr comment`.
 // ---------------------------------------------------------------------------
 // Plain (NOT code-fenced) short SHA so GitHub auto-links it to the commit — a
 // backtick-wrapped SHA renders as code and is NOT linkified.
@@ -1004,8 +1006,9 @@ function remapCommits(slug, data, picks) {
 // BASELINE (and the fallback); a reporter agent turns the structured track
 // result into a genuinely detailed, well-organized comment — narrative + tables
 // + reasoning — that a string builder can't synthesize from the rich-but-under-
-// used fields each track carries. The mechanical `postComment` below still does
-// the actual `gh pr comment`; only the body *authoring* becomes an agent.
+// used fields each track carries. The reporter only AUTHORS (it Writes its body to a
+// draft file and returns metadata); a separate mechanical agent does the actual `gh
+// pr comment`, so the side effect stays split from the fallible authoring agent.
 // ---------------------------------------------------------------------------
 
 // Per-track reporter registry, keyed by slug. Each entry states a track's result
@@ -1055,31 +1058,85 @@ function hasRichContent(slug, data) {
   return reporters[slug]?.isRich?.(data) ?? false;
 }
 
-// Author AND post one comment in a SINGLE agent, returning the posted comment URL
-// (or "no PR"/""). The rich body is the expensive artifact — up to 60 KB — and the
-// old two-agent split paid for it TWICE: a reporter (Sonnet) emitted it, the
-// workflow held it, then a SECOND poster (Haiku) RE-INGESTED it as a ~50-60k-token
-// base64 blob just to write it to a file and run `gh`. Here the reporter writes its
-// OWN body straight to the comment file with the Write tool (content-as-parameter,
-// so no shell-escaping of markdown's backticks/quotes/`$`) and posts it itself, so
-// the body crosses an agent boundary exactly once. The deterministic `baseline`
-// remains the never-blank fallback: a trivial/`--no-rich-comment` slug, or an
-// agent THROW, posts the (small) baseline via `postComment` instead — every
-// requested track still gets exactly one comment. `baseline`'s first line is the
-// canonical header (carrying the PR anchor + methodology link); the agent is told
-// to lead with it verbatim, so anchors stay stable whether the body is agent- or
-// builder-authored.
+// Author one rich comment body to a DRAFT FILE and post it — but as TWO agents
+// split at the side-effect boundary, NOT one. The rich body is the expensive
+// artifact (up to 60 KB), so we want it to cross an agent boundary exactly once
+// (the old shape re-ingested it as a ~50-60k-token base64 blob in a second poster
+// agent — pure waste). We get that by having the reporter (Sonnet) Write its body
+// straight to a draft file and return only METADATA about it (first line + byte
+// size) — NOT the body, and NO `gh`. The workflow then VALIDATES that metadata and
+// only THEN hands the file PATH to a narrow mechanical poster that runs the single
+// `gh pr comment --body-file` (the poster never reads the body as prose).
+//
+// Why the split matters (codex F1/F2/F3): authoring is fallible (a schema-validation
+// throw, a timeout, a malformed result) and ingests PR-derived/adversarial text; the
+// `gh` post is a non-idempotent external side effect. Folding both into one agent
+// meant a throw AFTER the post had run would trigger the baseline fallback and post a
+// SECOND comment — a silent duplicate the workflow couldn't detect or undo, breaking
+// the one-comment-per-track contract. Keeping the post OUT of the authoring agent means
+// every fallback branch (trivial slug, disabled, invalid metadata, authoring throw)
+// fires BEFORE any `gh pr comment` ran, so exactly one comment is posted per track. It
+// also keeps the side-effectful agent OFF the untrusted-data path: the authoring agent
+// has no `gh`, and the poster sees only a file path it posts opaquely.
+//
+// `baseline`'s first line is the canonical header (PR anchor + methodology link); the
+// agent is told to lead with it verbatim and the workflow ENFORCES it, so anchors stay
+// stable whether the body is agent- or builder-authored. The deterministic `baseline`
+// remains the never-blank fallback through the base64 `postComment` path.
 async function authorAndPost(slug, data, baseline, guidance) {
-  // Trivial track or rich comments disabled: no authoring agent to merge a post
-  // into — write+post the deterministic baseline through the cheap base64 poster.
-  // The baseline is small, so its token cost was never the waste (the LARGE rich
-  // body was), and routing it here keeps `postComment` as the single fallback path.
+  // Trivial track or rich comments disabled: no authoring agent — post the
+  // deterministic baseline through the base64 poster. The baseline is small, so
+  // its token cost was never the waste (the LARGE rich body was); routing it here
+  // keeps `postComment` as the single fallback path for non-rich slugs.
   if (!richComment || !hasRichContent(slug, data)) {
     return postComment(slug, baseline);
   }
   const file = `${SCRATCH}/comment-${slug}.md`;
   const header = String(baseline).split("\n")[0];
-  const prompt = `You are the **${slug} reporter** for a /be-review run. Author ONE genuinely detailed, well-organized GitHub PR comment from the structured result below — narrative + tables + reasoning, not a terse row count — then POST it.
+  // STAGE 1 — author only (no side effect). The reporter Writes its body to the
+  // draft file and returns metadata. If authoring throws, we fall back to the
+  // baseline BEFORE any post has happened (no duplicate possible).
+  let meta;
+  try {
+    meta = await authorDraft(slug, data, baseline, guidance, file, header);
+  } catch (e) {
+    log(
+      `Report: ${slug} reporter agent failed (${String(e)}) — posting the deterministic baseline instead.`,
+    );
+    return postComment(slug, baseline);
+  }
+  // STAGE 1.5 — validate the draft against the hard invariants the prompt asked for
+  // but cannot guarantee (codex F2): nonempty body, the baseline's exact header as
+  // the first line (PR anchor), and <= GitHub's 65 536-char body cap (60 KB budget
+  // with headroom). The metadata is computed by the agent over the file it wrote, so
+  // a mismatch means the draft is unusable — fall back to the baseline, still before
+  // any post.
+  const firstLine = String(meta?.firstLine ?? "");
+  const bytes = Number(meta?.bytes ?? 0);
+  const bad =
+    !bytes ||
+    (header.trim() && firstLine.trim() !== header.trim()) ||
+    bytes > 60000;
+  if (bad) {
+    log(
+      `Report: ${slug} reporter draft rejected (bytes=${bytes}, headerOk=${firstLine.trim() === header.trim()}) — posting the deterministic baseline instead.`,
+    );
+    return postComment(slug, baseline);
+  }
+  // STAGE 2 — post the validated draft by PATH through the narrow mechanical poster.
+  // The body never re-crosses an agent boundary as data (the poster posts the file it
+  // does not interpret), so the rich body is emitted exactly once.
+  return postFile(slug, file);
+}
+
+// STAGE 1 of `authorAndPost`: author the rich body to a draft FILE (no `gh`, no other
+// side effect) and return metadata describing what was written — the body itself never
+// comes back across the boundary (that is the whole point: it is written once and read
+// only by the mechanical poster's `gh --body-file`). The agent reports the draft's
+// first line and byte size so the workflow can enforce the header/size invariants
+// without re-ingesting the body.
+function authorDraft(slug, data, baseline, guidance, file, header) {
+  const prompt = `You are the **${slug} reporter** for a /be-review run. Author ONE genuinely detailed, well-organized GitHub PR comment from the structured result below — narrative + tables + reasoning, not a terse row count — and WRITE it to a draft file. Do NOT post it; do NOT run \`gh\`.
 
 STRUCTURED RESULT (JSON):
 ${JSON.stringify(data, null, 2)}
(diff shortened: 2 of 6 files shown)


srid:

The codex track restructured the monolithic `authorAndPost` function into a clean two-stage pipeline (`authorDraft` + `postFile`), and the police track then spent most of its energy hardening error-handling across both the old and new shapes — with the new architecture absorbing or superseding the majority of those fixes before they even landed.

---

#### Clean picks (applied without conflict)

|---|---|---|---|
| codex | 7c1d9fbc | `clean` | 3c5ef2c202f74dc982bd545864f107f86701dc94 |
| codex | 996d0ffc | `clean` | 286189993554bcef631d5ab134a1e3abd4d41da1 |
| lens | 2674bcfc | `clean` | 60f2e2b440b90a6a47565cdb444530a6e3f70639 |
| police | cb95c0a3 | `clean` | 3d9d398f47015bd2371e840a1245fa1eaece28cb |

The two codex commits laid the structural groundwork — splitting the old monolithic agent call into the `authorDraft`+`postFile` two-stage architecture. The lens commit landed cleanly as well (it fixed the duplicate `comment:${slug}` label on the `postComment` side, which the new architecture had not yet addressed). The one clean police commit (r7, cb95c0a3) applied without conflict.

---

#### Reconciled picks — what changed and why

**Police-r1 (3dd9fcf9 → 1173a642e):** The police agent added `try/catch` to the trivial-path `postComment` and wrapped the `authorDraft` catch-block's fallback call in `try/catch` to honour the never-rejects contract. The conflict arose because THEIRS still referenced the old monolithic function tail — a code path that no longer existed after the codex split. Resolution: kept HEAD's two-stage structure and transplanted the error-handling improvements onto the three `postComment` fallback call sites in the new `authorAndPost` (the `catch-of-authorDraft`, `selfReportBad`, `INVALID_DRAFT`, and empty-url paths).

**Police-r2 (9dbf6fb3 → dad8b3dd0):** The police agent refactored the `authorDraft` catch-block from `try/catch` to `.then().catch()` promise chaining, so the success log fires only after `postComment` resolves rather than immediately. Same THEIRS-is-old-monolithic-tail conflict. Resolution: kept HEAD's new function structure and applied the `.then().catch()` chaining exclusively to the `authorDraft` catch block inside `authorAndPost`.

**Police-r3 (b257de06 → 42cd15b6b):** The police agent added an empty-url guard inside the old monolithic `authorAndPost`. THEIRS was again the old agent-call code. Resolution: kept HEAD's two-stage architecture and applied the empty-url guard to the new `authorAndPost` after `postFile`, with its own `try/catch` fallback — placing the check at the right layer of the new design.

**Police-r4 (2385e0f3 → 640542645):** The police agent strengthened the Write-tool compliance instruction in the old monolithic prompt text, adding `[MANDATORY]` wording. THEIRS was old agent-call code. Resolution: kept HEAD, and applied the `[MANDATORY]` wording improvement to `authorDraft`'s prompt (step 2), which is where that instruction already lives in the two-stage architecture.

**Police-r10 (0ec5e947 → 455eec356):** The most structurally significant reconciliation. The police agent extracted a `fallbackToBaseline()` helper to DRY up three identical `try/catch` blocks in the old monolithic code. After the codex rewrite, the new two-stage architecture had the same repeated pattern — three separate inline `try/catch postComment` calls. Resolution was four-part: (1) keep HEAD's `INVALID_DRAFT` sentinel and the two-stage architecture comment block; (2) extract the `fallbackToBaseline()` helper, honouring the police DRY fix; (3) use it in `authorAndPost` for the `selfReportBad`, `INVALID_DRAFT`, and empty-url fallback paths; (4) simplify the `authorDraft` catch block to call it directly rather than the `.then/.catch` chain. This commit is the final head: 455eec35603030f124fcb1b7a308749af53c8399.

---

#### Dropped picks — fully subsumed

Five police commits resolved to empty diffs and were dropped entirely. The pattern is consistent: each targeted a code path (`fileBytes`, `write-failed`, duplicate labels) that existed only in the old monolithic architecture and had already been superseded or had no analogue in the new two-stage design.

| source | what it tried to fix | why it was subsumed |
|---|---|---|
| 4655561a (r5) | `try/catch` around write-failed and empty-url `postComment` in old monolith | write-failed path doesn't exist in new arch; empty-url `try/catch` already added in r3 reconciliation (b257de06) |
| fdda22c0 (r6) | Narrowed write-failure fallback to avoid double-post when `fileBytes===0` but URL is present | `selfReportBad` check fires at STAGE 1 before any post; double-post concern cannot arise in new arch |
| 328a1cc2 (r8) | Added `url !== 'no PR'` guard to write-failure condition in old monolith | entire `write-failed`/`fileBytes` code path absent from new arch |
| 1b66647d (r9) | Renamed duplicate `comment:${slug}` label to `author:${slug}` in old monolithic agent call | new arch uses `report:${slug}` for `authorDraft` and `comment:${slug}` for `postFile` — no collision; lens track (2674bcfc) fixed the `postComment` side separately |
| 0a7f916e (r11) | Added `fileBytes` to `required[]` in old monolithic schema | new `authorDraft` schema already has `required: ['firstLine', 'bytes']` — `bytes` is already required |

The 5 dropped commits represent police suggestions that were architecturally stale the moment the codex rewrite landed. None of the underlying correctness goals were lost — they were either already addressed by the new design or covered by earlier reconciled commits.

---

**Final head:** 455eec35603030f124fcb1b7a308749af53c8399

srid:
## 🤖 Codex ⇄ Claude debate

This track reviewed the fold of comment posting into the reporter agent (commit 78464fc4), which for performance reasons combined rich body authoring and the `gh pr comment` side effect into a single synthesis agent. Codex identified that this regressed three hard guarantees — exactly-one-comment-per-track, mechanical draft validation, and the prompt-injection boundary — and drove their restoration over three rounds before approving.

**Outcome:** consensus after 3 rounds · codex reviewed at `xhigh` reasoning effort · base e81f72d6f4fd1e3f15179f13fea9e6e635e5024a

---

### Round 1 — codex ❌ · 4 findings open · commit 3c5ef2c202f74dc982bd545864f107f86701dc94

Codex opened with a blocking refusal: the performance commit folded `gh pr comment` into `authorAndPost`, which let the fallible synthesis agent become the side-effectful actor. Claude accepted all four findings and restructured the authoring pipeline before round 2.

#### Findings

| ID | Severity | Location | Status after R1 |
|----|----------|----------|-----------------|
| F1 | blocking | `be-review.workflow.js:1102` | fixed |
| F2 | major | `be-review.workflow.js:1095` | fixed (partial — see R2) |
| F3 | major | `be-review.workflow.js:1084` | fixed (partial — accepted in R2) |
| F4 | nit | `be-review.workflow.js:826` | fixed (partial — SKILL.md missed) |

**F1 (blocking) — duplicate comment on throw-after-post.**
`authorAndPost` treated `agent()` as atomic, but the spawned agent could post the `gh pr comment`, then fail schema validation or omit `url`, causing the `catch` block to post the deterministic baseline too. The contract of exactly one comment per track would be silently violated with no recovery path.

Claude's fix: `authorDraft` (STAGE 1) was split out as an author-only agent with no `gh` capability — its sole side effect is writing a draft file. Every failure on that path throws before any post. `postFile` (STAGE 2) becomes the single post site; if it throws, `parallel` resolves the slot to null and the slug is recorded as not-posted, never triggering a second post.

**F2 (major) — hard draft validation replaced by prompt instructions.**
The prior enforcement (nonblank body, preserved baseline header as first line, <=60 KB) had been removed. Line 1121 accepted only the URL returned by the agent without verifying the file that was actually written or its contents.

Claude's fix: restored mechanical enforcement at lines 1108–1125 (STAGE 1.5) using the authoring agent's self-reported `firstLine` and `bytes` (from `wc -c`), rejecting on empty body, first line not equal to the expected header, or bytes > 60 000 — all before any post.

**F3 (major) — prompt-injection boundary collapsed.**
The new prompt embedded structured review data and the deterministic baseline as plain text inside the same tool-enabled agent that was also instructed to run `gh`. PR-derived text (including adversarial source/comment content) could now influence a side-effectful instruction stream, erasing the opaque-data boundary that `postComment` had maintained with base64.

Claude's fix: accepted the nuance that the *authoring* agent must ingest untrusted data (it cannot synthesize without it), but moved `gh` entirely out of the authoring agent. `postFile` receives only a file path; the base64 `postComment` boundary is preserved for the baseline fallback. Codex accepted this disposition in round 2.

**F4 (nit) — stale architectural comments.**
Surrounding comments still described the old thin-mechanical-agent model while the new code had the rich reporter running `gh`. Claude updated three comment blocks (`:826`, `:1003`, `:1580`) and regenerated the `.claude` and `.agents` copies byte-identically. The SKILL.md documentation was not updated in round 1, leaving F4 partially open.

---

### Round 2 — codex ❌ · 2 findings open · commit 286189993554bcef631d5ab134a1e3abd4d41da1

Codex acknowledged the author/post split as the right direction and confirmed F1 and F3 resolved. Two findings remained open, both with the same root: trusting agent self-reports where mechanical checks were needed.

#### Updated finding status

| ID | Severity | Codex disposition | Reason |
|----|----------|-------------------|--------|
| F1 | blocking | resolved | `gh` removed from authoring agent; fallback fires before any post |
| F2 | major | still open | metadata (`firstLine`, `bytes`) came from the same agent that wrote the draft; `trim()` weakened the header comparison |
| F3 | major | resolved | authoring agent must ingest untrusted data; the important boundary (side-effectful poster off untrusted path) was now maintained |
| F4 | minor | still open | SKILL.md still described `synthModel` reporters as "author AND post their own comment" — the exact architecture that F1/F3 moved away from; same stale text mirrored in all three copies |

**F2 (major, re-opened) — metadata validated against agent self-report, not the file.**
Codex pushed back specifically: if the authoring agent misreports `firstLine`/`bytes`, writes a blank or oversized file, or fails to write the file at all while returning plausible metadata, `postFile` would post the bad draft. The `trim()` in the comparison also meant the header was not enforced byte-exactly.

Claude's fix: moved the authoritative gate inside the mechanical poster. `postFile` receives the expected header base64-encoded, decodes it to a sibling `.header` file (with trailing newline for `diff` correctness), then runs an independent shell check before any `gh`:

```
if [ -s FILE ] && [ "$(wc -c < FILE)" -le 60000 ] && head -n 1 FILE | diff -q - FILE.header >/dev/null
then echo OK
else echo BAD
fi
```

On `BAD`, the poster returns `INVALID_DRAFT` without posting; `authorAndPost` catches this and falls back to the deterministic baseline — still exactly one comment. The old workflow-side `meta` check is retained but demoted to a cheap non-authoritative early-out with a comment to that effect.

**F4 (minor, re-opened) — SKILL.md still described old folded author+post.**
All three copies (`.apm`, `.claude`, `.agents`) were updated: the synthModel description changed from "author AND post their own comment" to "author a draft FILE only, leaving the `gh` post to a mechanical poster"; the mechModel list gained the draft-validating poster; the Report-phase description was updated to describe the author→validate→post-by-path split. Byte-identity of all three copies confirmed via `git hash-object`.

---

### Round 3 — codex ✅ · 0 findings open

Codex approved after independently verifying: branch state is limited to the expected be-review source/generated files; the `.apm`, `.claude`, and `.agents` copies are byte-identical by SHA-256; the workflow now splits rich authoring from posting; the draft file is validated in the poster by actual file inspection before `gh`; and SKILL.md accurately documents the new architecture.

#### Final finding resolution

| ID | Final location | Resolution |
|----|---------------|------------|
| F1 | `be-review.workflow.js:1094` | `authorAndPost` falls back before posting on every failure path; `gh` is in the poster only |
| F2 | `be-review.workflow.js:1219` | `postFile` writes a base64-decoded header file and runs `test -s` + `wc -c <= 60000` + `head -n 1 \| diff` against the real draft; returns `INVALID_DRAFT` on failure |
| F3 | `be-review.workflow.js:1157` | Poster receives only a file path + opaque base64 header; posts via `gh --body-file`; never reads or interprets the draft body |
| F4 | `SKILL.md:101` | All three copies updated and confirmed byte-identical |

---

### Convergence summary

The debate resolved a genuine regression: the performance optimization of folding posting into the reporter had removed three independent safety properties in one commit. Codex's round-1 blocking finding correctly identified the failure mode (throw-after-post = silent duplicate), and the two subsequent rounds tightened the fix from "trust the agent's self-report" to "inspect the real file independently." The architectural outcome — author-only agent writes a draft file, JS validates metadata cheaply, mechanical poster checks the file byte-exactly and posts by path — is a stronger invariant than the pre-optimization baseline because the draft gate now operates on the actual file rather than on a body the agent already posted.

**Files changed (all three copies byte-identical):**
- `.apm/skills/be-review/be-review.workflow.js`
- `.apm/skills/be-review/SKILL.md`
- (mirrored to `.claude/skills/be-review/` and `.agents/skills/be-review/`)

srid:

The police track worked through 4 review sweeps of `authorAndPost` — the new concurrent comment-poster introduced by this PR — finding **11 issues** (4 major, 5 minor, 1 nit, 1 minor-elegance) that together weaken the "never silently drop a comment" contract the parallel design is supposed to maintain. The sweep did **not reach a clean pass within the round cap**; 5 findings carry no commit and remain open.

---

### Rules pass

These findings violate structural invariants: silent drops, broken error contracts, and DRY violations.

#### 🔴 Major — Silent drop of comment-posting failures in the parallel loop

**Problem.** When `parallel` catches a thunk throw it resolves the slot to `null`, and the for-loop's `if (!item) continue` silently skips that slug with no log. Two paths feed into this trap. (1) The trivial/no-rich branch (`return postComment(slug, baseline)` at line 1078) has no try/catch, so a failing `postComment` propagates unguarded through `authorAndPost` into `parallel`'s null catch — no log fires at all. (2) The catch-block fallback (line 1129) logs "posting the deterministic baseline instead" *before* `postComment` resolves, so if `postComment` also throws, the log is a lie and the actual failure is absorbed silently. The old serial for-loop crashed the phase loudly on any `postComment` failure; the new concurrent design makes the same failure invisible.

**Fix prescribed.** Wrap the trivial-path call: `try { return await postComment(slug, baseline); } catch (e2) { log(...); return ''; }`. Move the fallback log to fire only after `postComment` settles, and add a `.catch()` for the double-failure case.

**Applied (1173a642e).** Both `postComment` call sites in `authorAndPost` wrapped with try/catch to restore failure observability; the trivial path now logs and returns `""` on failure; the agent-failure fallback now awaits `postComment` before logging, and catches a secondary failure with a combined error message.

---

#### 🟡 Minor — Misleading pre-flight log in the catch-block fallback

**Problem.** `"posting the deterministic baseline instead"` fires synchronously before `postComment(slug, baseline)` is awaited. If the fallback also fails, the log is factually wrong — it asserts posting happened when nothing was posted — and no subsequent message corrects it.

**Fix prescribed.** Move the success phrase into a `.then()` that fires only after `postComment` resolves; add a `.catch()` for the double-failure log.

**Applied (dad8b3dd0).** Restructured the catch-block fallback to use `.then()/.catch()` chaining so the success log fires only after resolution; added `.catch()` to clearly log the double-failure case.

---

#### 🔴 Major — Unguarded postComment rejections in write-failed and empty-url fallback paths

**Problem.** Lines 1140 and 1146 execute `return postComment(slug, baseline)` without `await` inside the `try` block. A rejection from `postComment` is therefore NOT caught by the surrounding `catch (e)` — it propagates out of `authorAndPost` directly. The trivial-path guard (lines 1078–1085) correctly uses `return await postComment(...)` inside its own try/catch and returns `""` on failure; the agent-failure catch block chains `.catch` on the promise. These two mid-try fallback paths are inconsistent and break the never-blank / never-rejects contract. If `postComment` fails on either path, `authorAndPost` rejects, and `parallel` resolves the slot to `null` — silent skip, no log.

**Fix prescribed.** Await the fallback calls with an inner guard matching the trivial-path pattern, or extract a shared `fallbackPost(slug, baseline)` helper:
```
if (url === 'write-failed' || fileBytes === 0) {
  log('...');
  try { return await postComment(slug, baseline); }
  catch (e2) { log(`${slug} baseline poster also failed (${String(e2)}) — skipping.`); return ""; }
}
```

**Applied.** No commit recorded — this finding is **open**.

---

#### 🟡 Minor — dry-rule-of-three: identical fallback-to-postComment blocks copy-pasted three times

**Problem.** The `try { return await postComment(slug, baseline) } catch (...) { log(...); return ""; }` pattern appears verbatim in three places: the `!richComment || !hasRichContent` early-return branch (lines 1078–1085), the `write-failed`/`fileBytes===0` guard (lines 1140–1148), and the empty-url guard (lines 1153–1160). Two of the three already diverged in variable name (`e` vs `e2`) without semantic reason; any future change to fallback behaviour must be applied three times.

**Fix prescribed.** Extract `async function fallbackToBaseline(slug, baseline, reason)` and call it from all three sites.

**Applied (455eec356).** Helper extracted; all three call sites replaced.

---

### Fact-check pass

These findings identify logic errors: missing guards, false invariants, and incorrect assumptions about agent output.

#### 🔴 Major — Empty url not triggering fallback — silent no-post on happy-path agent failure

**Problem.** `authorAndPost` returns `String(out.url || "").trim()` on the happy path. If the Sonnet reporter agent satisfies the required-url schema with `url: ""` (authored and wrote the file, but failed to capture the URL from `gh` output and returned blank rather than throwing), the function returns `""` silently. No fallback to `postComment` fires. The track is recorded in `comments[slug]` as an empty string and logged as "Report: posted slug comment" (no URL), but no comment appears on the PR. The old body-validation (`!body` → log + return baseline) was the safety net for this; the new code has no equivalent check. A throw triggers the catch-fallback correctly; a non-throw empty-url does not.

**Fix prescribed.** After `const url = String(out.url || "").trim()`, add: `if (!url) { log(...); return postComment(slug, baseline); } return url;`

**Applied (42cd15b6b).** Empty-url guard added so a blank url returned by the reporter agent falls back to `postComment` instead of silently returning `""`.

---

#### 🟡 Minor — Write-tool compliance enforced by instruction only — shell fallback corrupts bodies with special characters

**Problem.** The prompt tells the Sonnet reporter agent to use the Write tool rather than a shell heredoc. This is an instruction-only constraint, not a structural one. An agent that ignores it and uses `echo`, `printf`, or a heredoc will silently corrupt bodies containing unescaped backticks, `$variables`, or double-quotes. The old base64 path was structurally safe; the new Write-tool path is only as safe as agent compliance. If the body is corrupted, `gh` posts a mangled comment with no detection or fallback.

**Fix prescribed.** Either accept the risk as intentional, OR add a post-hoc size check, and explicitly mark the Write tool call MANDATORY in the prompt with shell alternatives forbidden.

**Applied (640542645).** Write-tool step marked MANDATORY in the prompt, shell alternatives explicitly forbidden, `wc -c` verification step added, agent schema extended with `fileBytes`, and empty-file corruption detected with a logged fallback to the deterministic baseline.

---

#### 🔴 Major — fileBytes not required in schema — empty-file guard silently bypassed

**Problem.** The empty-file fallback at line 1136 fires only when `fileBytes === 0`, but `fileBytes` is NOT in the `required` array — only `url` is. When the agent omits `fileBytes`, `Number(undefined ?? -1)` yields `-1`, not `0`, so the guard never fires even if the Write tool produced an empty file. An agent that skips the byte-count step can silently post an empty or shell-corrupted comment body instead of falling back to the deterministic baseline.

**Fix prescribed.** Add `"fileBytes"` to `required: ["url", "fileBytes"]` at line 1120.

**Applied.** No commit recorded — this finding is **open**.

---

#### 🟡 Minor — fileBytes===0 triggers postComment fallback even when a GitHub URL is present — potential double-post

**Problem.** The condition `url === "write-failed" || fileBytes === 0` calls `postComment(slug, baseline)` regardless of what `url` contains. If the agent reports `fileBytes: 0` while also returning a valid GitHub comment URL (e.g., because it mis-read `wc -c` after a successful write), `postComment` fires and posts a *second* baseline comment on top of the already-posted rich one. The old substitution path could only replace a rich body with a baseline body; the new path makes a false-positive an additive post (two comments instead of one wrong one).

**Fix prescribed.** Narrow the condition to prevent double-posting: change `fileBytes === 0` to `fileBytes === 0 && !url.startsWith("http")`.

**Applied.** No commit recorded — this finding is **open**.

---

#### 🟡 Minor — "no PR" sentinel treated as write-failure when agent also reports fileBytes: 0

**Problem.** The condition `fileBytes === 0 && !url.startsWith("http")` is true when the agent returns `{url: "no PR", fileBytes: 0}` — a plausible response if it wrote a 0-byte file before deciding no PR exists, or explicitly reports 0 because it skipped the write. The branch falls into the write-failure log ("body may have been corrupted by a shell command") and triggers a redundant `postComment` call that will also return "no PR". The outcome is correct by accident, but the error message is wrong, an unnecessary agent is spawned, and the real cause is obscured.

**Fix prescribed.** Change the condition to `url === "write-failed" || (fileBytes === 0 && url !== "no PR" && !url.startsWith("http"))` so the no-PR sentinel falls through to `return url` (same outcome, correct log, no redundant call).

**Applied.** No commit recorded — this finding is **open**.

---

#### 🟡 Minor — Original reporter error `e` silently dropped in double-fail catch path

**Problem.** After the catch block was restructured to use `.then()/.catch()` chaining, the `.catch(e2)` log message reads `reporter AND baseline poster both failed (${String(e2)})` — including only the baseline-poster error `e2`. The original reporter-agent error `e` that triggered the whole fallback is silently dropped. An operator seeing this log cannot tell why the reporter failed in the first place.

**Fix prescribed.** Change the message to `reporter agent failed (${String(e)}); baseline poster also failed (${String(e2)})`.

**Applied (3d9d398f4).** Double-fail catch block updated to include original reporter error `e` alongside `e2` in the diagnostic log.

---

#### Nit — Duplicate agent label `comment:${slug}` across authorAndPost and postComment

**Problem.** Both the rich-reporter agent call in `authorAndPost` (line 1114) and the fallback `postComment` function (line 1270) use `label: \`comment:${slug}\`` for the same slug. When the reporter throws and falls back to `postComment`, two sequential agents share the same label. If the harness uses labels for deduplication or log correlation, the second agent's events shadow the first.

**Fix prescribed.** Give the rich-reporter agent a distinct label: `label: \`author:${slug}\`` in `authorAndPost`, keeping `comment:${slug}` for the posting-specific fallback.

**Applied.** No commit recorded — this finding is **open**.

---

### Open findings summary

Five findings carry no commit and remain unresolved in this worktree:

| id | severity | title |
|---|---|---|
| police-r2-rules-1 | major | Unguarded postComment rejections in write-failed and empty-url fallback paths |
| police-r2-fact-check-1 | minor | fileBytes===0 triggers postComment fallback even when a GitHub URL is present |
| police-r3-fact-check-1 | minor | "no PR" sentinel treated as write-failure when agent also reports fileBytes: 0 |
| police-r3-fact-check-2 | nit | Duplicate agent label `comment:${slug}` across authorAndPost and postComment |
| police-r4-fact-check-1 | major | fileBytes not required in schema — empty-file guard silently bypassed |

The most consequential open issue is **police-r4-fact-check-1**: `fileBytes` absent from the schema `required` array means the entire empty-file corruption guard introduced in this sweep is opt-in and silently bypassed whenever the agent omits the field. This should be fixed before merge.

srid:

**Outcome:** consensus after 1 round. Lowy raised 4 findings independently; Hickey raised 5. All 9 settled in round 1, converging to 8 drops and 1 fix (applied). The core tension in the debate was whether the perf-motivated fusion of `authorBody`+`postComment` into `authorAndPost` introduced genuine structural debt or merely reflected the irreducible cost of keeping a 60 KB body inside one agent boundary — and on every contested finding, the runtime constraint won the argument.

---

### What this PR changed (lens perspective)

Before this PR, `be-review`'s comment pipeline had clean axis separation: `authorBody` owned "which strategy authors the body" (rich-agent vs. deterministic-baseline), `reporterBody` owned authoring + never-blank enforcement, and `postComment` owned delivery. This PR collapses all three into `authorAndPost` so the rich comment body crosses exactly one agent boundary — the Sonnet reporter agent both writes and posts in the same turn, eliminating the prior re-ingest of the ~60 KB body as a base64 blob. Both lenses probed whether that fusion introduced hidden complexity; what follows is their independent takes and where they agreed, diverged, and cross-corrected.

---

### Lowy's independent findings

Lowy approached from Juval Lowy's volatility-based decomposition framework: are axes of change correctly separated into independent receptacles, and does each receptacle's interface remain stable under the volatility it owns?

**lowy-1 — Two posting paths behind one receptacle** (`.apm/skills/be-review/be-review.workflow.js:1072-1131`, dropped)

The new `authorAndPost` encapsulates two independent volatility axes — "how is a body authored" and "which delivery mechanism posts it" — where before these were separate. Three internal posting call-sites exist: trivial/disabled branch → `postComment`, success branch → Write+`gh` agent, catch branch → `postComment` again. Lowy called this the sequence-vs-activity conflation: "choose generator" (activity) and "deliver the comment" (sequence/IO) jammed behind one function. **Disposition: drop.** The fusion is the literal point of the perf change; the three sites collapse to two well-defined edges, and the URL contract returned is identical regardless of which generator produced it.

**lowy-2 — Anchor-stability invariant moved from enforced check to unenforced prompt instruction** (`.apm/skills/be-review/be-review.workflow.js:1095-1121`, dropped)

Previously the workflow validated the returned body contained the canonical header verbatim before posting; failing validation fell back to baseline. The diff deletes that validation and replaces it with the prompt instruction "FIRST line MUST be EXACTLY this header verbatim." Through the interface-stability lens, the receptacle's stable guarantee is now only as strong as an LLM following an instruction. **Disposition: drop, deliberately.** The author's note flags this as a known trade: once the agent also posts, there is no returned body left to validate. If anchor drift is later observed, the cheapest re-enforcement is to have the agent return the first line in its schema (alongside `url`) and assert it equals `header` — a small string that doesn't reintroduce the 60 KB re-ingest cost.

**lowy-3 — `postComment` is single-consumer but its interface stays stable** (`.apm/skills/be-review/be-review.workflow.js:1208-1223`, dropped)

After this PR, `postComment` is reached only from inside `authorAndPost`. Reuse-count alone would flag it as a candidate for inlining. But applying §6's interface-stability test rather than reuse count: its interface — post an opaque, injection-safe body to the branch's PR and return the URL — is stable under the volatility it encapsulates (any body, baseline or otherwise, rides the same opaque path). **Disposition: drop.** The boundary is sound; single in-tree consumer does not condemn a stable receptacle.

**lowy-4 — Two delivery paths write through no shared primitive** (`.apm/skills/be-review/be-review.workflow.js:1080-1102, 1208-1216`, dropped)

Both `authorAndPost`'s rich path and `postComment` independently compute `${SCRATCH}/comment-${slug}.md` and specify the same `cd ${repoPath} && gh pr comment --body-file ${file}` step — one as Write-tool instructions inside a Sonnet agent, one as a printf pipeline inside a Haiku agent. If the gh posting command changes (flag, PR-resolution rule, "no PR" handling), two prompt strings must change in lockstep. **Disposition: drop for this PR.** The split is the deliberate cost of folding a large body into one agent while keeping the fallback cheap. If revisited, factoring the gh-posting contract into a single shared prompt fragment string would keep it in one place.

---

### Hickey's independent findings

Hickey approached from Rich Hickey's "Simple Made Easy" framework: are independent concerns complected together, and can each piece be reasoned about in isolation?

**hickey-1 — Posting mechanism fragmented into two structurally different paths** (`.apm/skills/be-review/be-review.workflow.js:1072-1131`, initial: fix → final: drop)

Before this PR there was one way a comment reached the PR: `postComment` (base64-opaque write + `gh pr comment`). Hickey flagged that "emit a PR comment" is now fragmented across `authorAndPost`'s inline steps and `postComment`, with the two paths differing in trust model, escaping strategy, and which agent/model handles the `gh` call. The `${SCRATCH}/comment-${slug}.md` path computed identically at line 1080 and line 1209 was cited as the fingerprint of fragmentation. Hickey initially proposed extracting a `postBodyFile(slug)` concept to name the file location once.

**Cross-examination (lowy):** lowy conceded the file-path duplication is real but argued Hickey overstated the fragmentation — the two paths were already two paths before this PR (reporterBody authored, postComment posted); the PR moved the gh-invocation into the rich agent to eliminate re-ingest, it didn't shatter one concept into two. The proposed helper dedups a single interpolated string literal; if the path drifts, the failure mode is benign and visible (agent writes one file, posts another → empty/failed post, caught), not silent corruption. **Final: drop.**

**hickey-2 — `authorAndPost` complects strategy-choice, authoring, posting, and fallback** (`.apm/skills/be-review/be-review.workflow.js:1072-1131`, initial: fix → final: drop)

Hickey identified four distinct concerns braided into one function: the `richComment`/`hasRichContent` strategy gate (1077-1079), authoring the rich body, posting (embedded as natural-language steps inside the authoring prompt), and the throw-fallback that posts baseline. The prior decomposition kept these as separate, composable parts. Hickey suggested: keep authoring as a body-returning agent, have that agent Write the body to `${file}` (avoiding base64 re-ingest), then have the *workflow* run `gh pr comment --body-file` deterministically — keeping authoring and posting as separately-reasonable concerns.

**Cross-examination (lowy):** lowy identified the load-bearing fact Hickey's suggestion missed: the Workflow runtime has no `Buffer`/`TextEncoder`/`btoa` (documented at lines 1145-1154), and `postComment` is itself an agent call (`return agent(...)` at 1218) — the workflow has no shell. "Author returns body, workflow posts" would require a second agent for the post, which is precisely the re-ingest this PR removes. The braiding is real but is the necessary cost of the single-boundary win; the cheap alternative doesn't exist in this runtime. **Final: drop (Hickey conceded on the load-bearing fact).**

**hickey-3 — Prompt-injection containment silently dropped on the rich path** (`.apm/skills/be-review/be-review.workflow.js:1100-1103, 1139-1144`, initial: fix → final: drop)

`postComment`'s comment block (1139-1144) asserts the invariant that base64 makes the body OPAQUE to the poster — arbitrary finding/source text can't be read as instructions. On the new rich path, the reporter agent both synthesizes the body from structured findings AND runs the `gh` steps in the same turn, with body text sitting in context as prose, not opaque data. Hickey flagged this as two contradictory answers to the same security question with no acknowledgement that the rich path deliberately accepts the risk.

**Cross-examination (lowy):** lowy identified the premise error: before this PR, the rich body was authored by `reporterBody` — a Sonnet agent that *always* had `JSON.stringify(data)` inlined into its authoring prompt (the diff shows this block unchanged at 1084-1085). Base64 opacity protected ONLY the separate Haiku poster (`postComment`), which never authored and never needed to see findings as prose. The rich path never had an opaque-data poster to lose. The injection surface (findings text in a synth agent's context) is identical before and after this PR. **Final: drop (Hickey conceded the premise was wrong).**

**hickey-4 — Anchor-stability invariant downgraded from enforced check to prompt instruction** (`.apm/skills/be-review/be-review.workflow.js:1095, 1104-1121`, dropped)

Same substance as lowy-2 but with a sharper sub-point: a >65536-char body now throws in `gh` and *does* fall back to baseline (the byte-cap is still enforced, just by `gh`'s rejection rather than a pre-check), while a subtly-wrong header posts a broken anchor silently with no detection. The asymmetry — byte-cap still enforced by a different mechanism, header stability now best-effort — is a real structural consequence.

**Cross-examination (lowy):** lowy acknowledged the asymmetry as factually correct but noted that re-enforcing the header requires Write-then-return-then-validate-then-post, which is the same runtime-impossible "workflow posts" shape (see hickey-2) or a second-agent re-ingest. The function's header comment at 1058-1071 already truthfully describes stability as prompt-level ("the agent is told to lead with it verbatim"), so a future reader is not misled. **Final: drop; the trade is deliberate and documented.**

**hickey-5 — Two distinct agents share the same `comment:${slug}` label** (`.apm/skills/be-review/be-review.workflow.js:1106, 1219`, **fixed** → 60f2e2b440b90a6a47565cdb444530a6e3f70639)

`authorAndPost`'s authoring+posting agent uses label `comment:${slug}` at line 1106 (synthModel/Sonnet). `postComment`'s mechanical baseline poster also uses `comment:${slug}` at line 1219 (mechModel/Haiku). Two structurally different agents — one that synthesizes prose and posts it, one that mechanically decodes base64 and posts an opaque body — now carry identical identity in `tokensByPhase`/turn telemetry. The pre-PR code distinguished them: the authoring agent was labelled `report:${slug}`, the poster `comment:${slug}`. This PR collapsed both to `comment:${slug}`.

**Cross-examination (lowy):** lowy immediately agreed — this is right and cheap. "Which agent emitted this comment / what did it cost" cannot be answered from the label alone. A one-token label change restores a real diagnostic the PR regressed, with zero behavioral risk. **Final: fix. Applied: change `postComment`'s label from `comment:${slug}` to `comment-baseline:${slug}` at line 1219, leave `authorAndPost`'s label at 1106 unchanged, regenerate runtime dirs via `just ai::apm`.**

---

### Disposition summary

| ID | Origin | Finding | Location | Disposition |
|---|---|---|---|---|
| lowy-1 | Lowy | Two posting paths behind one receptacle | 1072-1131 | drop |
| lowy-2 | Lowy | Anchor-stability moved from enforced check to prompt instruction | 1095-1121 | drop |
| lowy-3 | Lowy | `postComment` single-consumer but interface stays stable | 1208-1223 | drop |
| lowy-4 | Lowy | Two delivery paths write through no shared primitive | 1080-1102, 1208-1216 | drop |
| hickey-1 | Hickey | Posting mechanism fragmented into two structurally different paths | 1072-1131 | drop |
| hickey-2 | Hickey | `authorAndPost` complects strategy-choice, authoring, posting, fallback | 1072-1131 | drop |
| hickey-3 | Hickey | Prompt-injection containment silently dropped on rich path | 1100-1103, 1139-1144 | drop |
| hickey-4 | Hickey | Anchor-stability downgraded from enforced check to prompt instruction | 1095, 1104-1121 | drop |
| hickey-5 | Hickey | Two distinct agents share the same `comment:${slug}` label | 1106, 1219 | **fix** — 60f2e2b440b90a6a47565cdb444530a6e3f70639 |

### Findings that flipped during the debate

Three of Hickey's initial "fix" dispositions flipped to "drop" after cross-examination by Lowy:

- **hickey-1** (file-path duplication fingerprint): Lowy showed the two paths were already two paths pre-PR; the failure mode of path drift is benign and visible, not silent corruption. Hickey withdrew the fix.
- **hickey-2** (complected concerns): Lowy identified the load-bearing constraint — the Workflow runtime has no shell, so "workflow posts" requires a second agent, which is exactly the re-ingest this PR removes. Hickey conceded on the runtime fact.
- **hickey-3** (injection containment): Lowy showed the premise was wrong — the rich path never had base64 opacity to lose; `JSON.stringify(data)` was always inlined into the authoring agent's prompt. Hickey withdrew.

The one finding that held through debate was **hickey-5**: both lenses agreed immediately that collapsing two structurally different agents to the same telemetry label was a real regressed diagnostic, cheap to fix, zero behavioral risk.

PR #1174, opened 2026-06-04
perf(ci): cut pu-box linux CI wall via warm-fork provisioning + e2e parallelism

### LINKED ISSUE #1173: Optimize Linux CI time on pu boxes (cold Nix store, e2e parallelism, setup transfer)
## Optimize Linux CI time on `pu` boxes

Our `x86_64-linux` CI lane runs as `just` recipes fanned over SSH to an ephemeral `pu` box (`kolu-pr-<N>`, a clean 32-core NixOS Incus container). It works well, but a full pipeline takes a **median ≈262s** (range ~211–359s for clean runs; one shard-spike outlier at 525s), and almost all of that time is concentrated in a couple of recipes that are slow **specifically because the box is cold**, not because the work is inherently large.

This is a tracking issue to drive that wall time down. Everything below was **measured from existing logs in a single analysis session** (21 live `.ci/pc.log` files + `.ci/*/x86_64-linux/*.log` recipe logs across ~10 pu runs and one host-certain `localhost` run; no new runs were launched on any box). Numbers are medians unless noted.

### Where the time goes (per-recipe, x86_64-linux)

Process-compose runs the recipes concurrently, so the ~262s wall is gated by the **two long poles** (`ci::nix` and `ci::e2e`), which run side-by-side. The "warm local" column is the same recipe on a host with a warm Nix store (a single `ssh localhost` run on the 16-core dev machine) — the gap is the cold-cache tax.

| recipe | pu (cold box) | pu max | warm local | gap |
| --- | ---: | ---: | ---: | ---: |
| **`ci::nix`** | **174s** | 225s | 56s | **+118s** |
| **`ci::e2e`** | **114s** | 174s | 206s* | — |
| `ci::home-manager` | 85s | 170s | 62s | +23s |
| `ci::pnpm-hash-fresh` | 67s | 144s | 33s | +34s |
| `ci::docs-moc` | 61s | 61s | — | — |
| `ci::install` | 31s | 76s | 5s | +26s |
| `ci::typecheck` | 30s | 42s | — | — |
| `ci::unit` | 26s | 39s | 17s | +9s |
| `ci::smoke` | 20s | 116s | 7s | +13s |
| `ci::surface-example-build` | 19s | 29s | 4s | +15s |
| `_ci-setup` | 18s | 50s | 1s | +17s |
| `ci::biome` | 18s | 28s | 3s | +15s |
| `ci::fmt` | 17s | 28s | 1s | +16s |

<sub>*the warm-local `ci::e2e` is high because that single run was on a **contended** dev machine — see "contention" below.</sub>

The pattern is consistent: nearly every recipe is **2–15× slower on the fresh box** than warm, because a clean container has a cold Nix store and cold pnpm/build caches. The headline is `ci::nix` at **174s vs 56s warm** — that one recipe is the single biggest lever, and it's almost entirely cache state, not silicon.

### Levers (roughly in priority order)

**1. Warm / seed the Nix store on the box — biggest win (~100s+).**
`ci::nix` is the long pole purely because the box re-realises/re-downloads a closure that the warm dev machine already has. Options: point the box at a binary-cache substituter (Cachix/attic/local S3) so it *pulls* instead of rebuilding; keep a small pool of persistent warm boxes; or bake a periodically-refreshed base image. Collapsing `ci::nix` from ~174s → ~56s would also shrink the cold tax on `install`/`biome`/`fmt`/`smoke`/`home-manager`, which are slow for the same reason. This likely takes the pipeline wall under ~180s.

**2. Use the box's cores for `ci::e2e`.**
`CUCUMBER_PARALLEL` defaults to **`4`** (`justfile:13` — `cucumber_parallel := env('CUCUMBER_PARALLEL', '4')`), so the suite runs only 4 scenarios at a time and ~28 of the box's 32 cores sit idle. Some newer branches already auto-scale (`par=clamp(cores/3,4,8)` → observed 8 workers); standardize that everywhere and tune the cap upward for 32-core boxes (or shard across forked boxes, per the earlier e2e-sharding spike). e2e is ~114s of process time / ~75s cucumber-wall today; more workers should push it toward ~40–60s.

**3. Trim the `_ci-setup` transfer tax (~18–50s).**
Setup ships the closure to the box with `nix-store --export $(…) | ssh -T <box> nix-store --import` plus a full `git bundle --all`. With a shared substituter (lever 1) the store push largely disappears; the git bundle can be shallow (current `HEAD` only) instead of `--all`.

**4. Reduce run-to-run variance.**
Clean full-pipeline walls range **211–359s** and `ci::e2e` cucumber-walls range **66–248s** depending on which physical host the box lands on and how warm it is. Worth pinning boxes to known-good fast hosts and tracking the wall as a distribution, not a single number.

**5. Re-check the mid-tier recipes on the critical path.**
`ci::home-manager` (85s, max 170s), `ci::pnpm-hash-fresh` (67s, max 144s) and `ci::docs-moc` (61s) are the next tier after the two poles — confirm they're genuinely parallel (not gating the wall) and cache-warm once lever 1 lands.

### Bonus finding: dev-machine contention (argues for keeping CI on pu)

The one host-certain *local* baseline (`reconnect-roaming`, `ssh localhost`) ran the **same 379-scenario suite four times** at 4 workers and got **135s → 193s → 248s → 248s**, with the cucumber step-sum nearly **doubling (520s → 967s) for an identical scenario count**. That inflation with no change in work is a clear signature of the dev machine being **busy with other things** while CI ran. A dedicated pu box doesn't have that problem (e2e is a steady ~75–114s) — so even though a single cold pu run is roughly *tied* with a warm, idle local run on total wall, the pu box wins on **consistency** and on **not pinning the dev's 16 cores**. The optimizations above are about closing the cold-start gap so the pu box is also clearly faster per run.

### Caveats on the measurements
- The "warm local" column is a **single** `ssh localhost` run (n=1) — anecdotal for absolute numbers, but the *direction* (warm ≪ cold) is structural and corroborated by warm persistent remote hosts also showing fast `ci::nix` (7–26s).
- pu vs local runs are **different PRs/commits/dates**, not a controlled same-workload A/B; treat the gaps as indicative.
- `CUCUMBER_PARALLEL=4` is the **default-justfile** value; some branches have already moved to the auto-scaling form.
- Core counts (`nproc=32`, some 16) come from session-log mentions, not a fresh probe.
- Darwin lane and the transient no-egress placement issue are out of scope here.

_Filed from a measurement-only analysis session; happy to attach the raw per-run numbers or a breakdown chart if useful._

srid:
Closes #1173.

The `x86_64-linux` CI lane runs on an ephemeral `pu` box per PR. Almost all of a
cold run's wall time is the **cold-start tax** a fresh box pays, not inherent
work. Measured on a 32-core i9 pu box (full methodology + raw numbers in
[`docs/pu-box-ci-ralph-report.md`](docs/pu-box-ci-ralph-report.md)):

## Where the time goes (cold `pu create`)

`ci::nix` is the long pole — the box re-realises the whole devour-flake closure
from the substituter. `ci::e2e` runs *underneath* it, so e2e isn't even the
critical path until the store is warm.

| recipe | cold | warm fork | Δ |
|---|---:|---:|---:|
| **`ci::nix`** | **189s** | **8–12s** | **−94%** |
| `ci::install` | 69s | 24s | −65% |
| `ci::pnpm-hash-fresh` | 67s | 25s | −63% |
| `ci::home-manager` | 113s | 13–46s | −60–88% |
| **LANE_WALL** | **216–262s** | **142–195s** | **−26 to −46%** |

## What this PR does

1. **`ci/pu-ci-host.sh`** — provisions the linux host by **forking a warm
   `kolu-ci-golden` box** (hot Nix store ⇒ `ci::nix` ~20s), with a timeout guard
   and fallback to cold `pu create` → `hosts.json`. It works around two real
   `pu fork` bugs (no `ssh_config` written; a slow cross-pool 17 GB transfer that
   hits an Incus UUID crash), so it's **never slower than today** and warm
   whenever the fast fork path is available. Beyond raw speed, a warm fork pulls
   nothing from the substituter, so it's **immune to the contention that stalled
   concurrent cold runs past 12 min**.
2. **`.agency/do.md`** — `/do`'s CI step calls the script; documents keeping
   golden warm.
3. **`docs/pu-box-ci-ralph-report.md`** — the measurement report.

## Dead end (documented, not committed)

Raising `CUCUMBER_PARALLEL` past 8 on the 32-core box: **≤6% within-box, at the
noise floor** — the e2e suite is tail-bound. The first sweep *looked* like a 30%
win, but that was entirely box-placement variance (re-measuring within a single
box exposed it). Touching the shared clamp would also risk the tuned darwin lane.

## Upstream follow-ups

- `pu fork`: write `ssh_config` like `pu create`; pin same-pool CoW (avoid the
  cross-pool transfer + UUID crash).
- `justci`: shallow `_ci-setup` git bundle + skip the store push when the box is
  already warm (~16s).
- The robust alternative to fork is a shared binary-cache substituter.


srid:
## `pu fork` bugs blocking warm-fork CI provisioning (for the pu/Incus admin)

Both found while measuring #1173 on `kolu-ci-golden` (a 32-core NixOS container on `dev-x86-64-linux-05`). `ci/pu-ci-host.sh` works around both, but they should be fixed in `pu` proper so forking is a dependable mechanism.

### Bug 1 — `pu fork` doesn't write `~/.pu-state/<name>/ssh_config`

`pu create <name>` writes `~/.pu-state/<name>/ssh_config` (picked up by `~/.ssh/config`'s `Include ~/.pu-state/*/ssh_config`), so raw `ssh <name>` resolves. `pu fork <src> <name>` creates the instance and `pu connect <name>` works, but it **never writes that ssh_config file** — so anything using bare `ssh <name>` (notably justci's `_ci-setup`, which runs `ssh -T <host> …` over the wire) fails:

```
ssh: Could not resolve hostname <name>: Name or service not known
```

Reproduce:
```sh
pu fork kolu-ci-golden ftest
ls ~/.pu-state/ftest/ssh_config   # -> No such file or directory
pu connect ftest -- true          # works (pu resolves internally)
ssh -T ftest true                 # FAILS: cannot resolve hostname
```

**Fix:** have `pu fork` write the ssh_config exactly like `pu create`. It's identical to the source's block except the `Host` line and the `connect <name>` target in `ProxyCommand` (same gateway). Workaround in the script: `sed` the source box's config to the new name.

### Bug 2 — cross-pool `pu fork` transfers the full ~17 GB instance, then crashes on a UUID error

When the fork lands on the **same storage pool** as the source it's a fast CoW snapshot (seconds). When it lands on a **different pool/host**, `pu fork` does a full instance migration — transfers the whole ~17 GB (~110 MB/s, ~160s) and then **fails**:

```
Forking kolu-ci-golden...
Transferring instance: ftest: 17.46GB (109.77MB/s)
Error: Error transferring instance data: Failed migration on target: Failed creating
instance on target: Failed setting received UUID: Failed parsing UUID: invalid UUID length: 0
ERR failed to fork kolu-ci-golden
```
(exit 1)

Two problems: (a) even when it would succeed, ~160s of transfer defeats the purpose of forking a warm box; (b) the UUID error makes it fail outright, so fork is unreliable unless you happen to land same-pool. `invalid UUID length: 0` ⇒ the migrated instance's UUID (`volatile.uuid` / snapshot UUID) is arriving empty on the target — looks like an Incus migration bug or a pu-side config dropping the UUID across pools.

**Asks:**
1. Pin `pu fork` to the source's storage pool/host so it always uses fast CoW (never the cross-pool transfer path); and/or
2. Fix the UUID-on-migration failure so cross-pool forks complete; and
3. Bug 1's ssh_config.

**Diagnostics to grab:** `incus --version` on source + a failing target host; the `incus copy`/migration command `pu fork` runs (`--debug`); `incus config show kolu-ci-golden | grep -i uuid` vs a freshly-created box.

Why it matters: forking a warm box collapses `ci::nix` from ~189s (cold) to ~8–12s (warm) — the biggest linux-CI-time lever (#1173). These two bugs are the only thing keeping `pu fork` from being a dependable CI-provisioning mechanism.

PR #1175, opened 2026-06-04
add x86_64-darwin to supported systems

DivitMittal:
x86_64-darwin was missing from the systems list. one-liner fix.

built and ran fine on my old intel macbook. since this is a web project, there's basically zero arch-specific maintenance burden — js/ts doesn't care.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -17,7 +17,7 @@
 
   outputs = { self, ... }:
     let
-      systems = [ "x86_64-linux" "aarch64-linux" "aarch64-darwin" ];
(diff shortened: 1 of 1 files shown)


DivitMittal:
Alternatively, could also just use https://github.com/nix-systems/default instead of hardcoding the list. It covers all four common systems including x86_64-darwin out of the box. Though I know you're avoiding flake inputs here, so maybe not worth it.

srid:
I'll accept this, with no promise or guarantee of supporting Intel mac.

PR #1180, opened 2026-06-04
fix(terminal): stop yazi escape-soup by filtering duplicate responses and answering XTVERSION

DivitMittal:
TUIs like **yazi** probe the terminal on startup (`XTVERSION`, cell-size, DA1/DSR). Kolu runs a headless xterm **server-side** that answers these in-process — but the **browser** xterm answers them too, and its replies arrive late over the WebSocket, landing in the PTY as input *after* the TUI already consumed the server's answer. yazi renders the leftovers as visible garbage — and worse, the stray bytes corrupt the input stream (a `q` meant to quit gets swallowed into the next command):

```
^[P>|xterm.js(6.1.0-beta.225)^[\^[[6;18;9t
```

**The fix has two halves that must move together** — drop the browser's duplicate reply *and* make sure something still answers the query, or a TUI that blocks on it hangs forever:

- **Client (`terminalResponseFilter.ts` → `Terminal.tsx`)** — a single `isTerminalQueryResponse(data)` predicate filters `onData` chunks that *are* a complete terminal query response before they reach the PTY. The old inline check caught only `CSI c/n/R/y` + all-OSC; this covers the shapes yazi actually triggered.
- **Server (`ptyHost.ts`)** — the headless terminal now **answers `XTVERSION` itself** (`CSI > 0 q` → a DCS reply built from `HEADLESS_TERM_ID`). It had no built-in handler, so once the client filters the browser's reply, the server *must* supply one or the PTY blocks.

### What the filter suppresses

| Class | Matches | Notes |
| --- | --- | --- |
| CSI | DA1/DA2, DSR/CPR, DECRPM (`$y`), window/cell-size (`t`) | anchored `^…$` to the whole payload |
| OSC | **colour reports only** — OSC 4 (palette) + OSC 10–19 (dynamic colours) | *OSC 52 clipboard-read replies deliberately pass through — only the browser's `ClipboardAddon` can answer them* |
| DCS | `XTVERSION` (`ESC P >\|…ST`) + DECRQSS (`ESC P [01]$r…ST`) | sixel/DECUDK program output stays forwarded |

Every predicate is anchored to the **whole** chunk, so a paste that merely *contains* a response-shaped sequence is never eaten — only a chunk that *is* a complete response.

### Refinements from review (codex ⇄ claude, then lowy ⇄ hickey)

| Was | Now |
| --- | --- |
| `OSC_RESPONSE` dropped *every* OSC packet — clobbering OSC 52 clipboard reads the PTY never re-answers | Narrowed to `OSC_COLOUR_RESPONSE` (OSC 4 + 10–19); regression tests assert OSC 52 passes through |
| XTVERSION reply string hand-copied between the handler and its test | One exported `HEADLESS_TERM_ID` constant, referenced both places |
| The client-filters / server-answers lockstep was implicit across two packages | The load-bearing invariant is documented once at the filter, cross-referencing the `ptyHost` handler; the new XTVERSION path has real-PTY tests (`CSI > 0 q` answered, `CSI > 1 q` consumed without reply) |

> 📹 **Evidence** (before↔after yazi clips) and the full review trail are posted as PR comments.

_PR description rewritten, and the review-gauntlet fixes (codex + lowy/hickey lens debate) added, by Claude Code (model `claude-opus-4-8`)._


srid:
@DivitMittal How do I reproduce this bug?

srid:
## Codex ⇄ Claude debate

**Consensus** ✅ after **2 rounds**. codex reviewed at **`xhigh`** reasoning effort.

| Round | codex verdict | Claude dispositions | Commit |
|-------|---------------|---------------------|--------|
| 1 | ❌ not approved — 3 open findings (F1 filter too loose / drops whole payload, F2 DECRPM `$y` form unmatched, F3 no regression coverage) | F1 *partial* (adopted full-payload anchoring + content-specific DCS branch; disputed that live input was actually being eaten — xterm `onData` fires once per discrete event), F2 *fixed* (conceded; regex now accepts `$y`), F3 *fixed* (extracted `terminalResponseFilter.ts` + 22 regression cases) | `90c454f1` |
| 2 | ✅ approved — 0 open findings, all 3 resolved, no new regression | — | — |

**Net effect of the debate:** the inline `onData` filter became an extracted, tested helper `terminalResponseFilter.ts` (`isTerminalQueryResponse`). Every predicate is now anchored to the full payload (`^…$`) so a chunk merely *containing* a response sequence is forwarded to the PTY intact; the DCS branch is narrowed to XTVERSION (`>|`) and DECRQSS (`[01]$r`) introducers so sixel/program output passes through; and DECRPM's `$y` final is now matched. 22 unit tests added; full client suite (289) green.

srid:

**Consensus** ✅ after **1 round**. Independent review: **lowy 3** findings, **hickey 2** findings — all **dropped**, **no fixes applied**. The structure (post-codex `terminalResponseFilter.ts`) is sound.

| Origin | Finding | Location | Disposition |
|--------|---------|----------|-------------|
| lowy | Receptacle correctly extracted — escape-grammar wires now hidden from the consumer (`isTerminalQueryResponse(string): boolean` is the stable plug; the recognized response grammar is the encapsulated axis of change) | `terminalResponseFilter.ts:40` | drop ✅ |
| lowy | Single in-tree consumer is not disqualifying — interface is stable under the encapsulated axis (synthetic-keystroke sites correctly don't plug in) | `terminalResponseFilter.ts:40` | drop ✅ |
| lowy | Module is cohesive — one export, three private grammar submodules (CSI/OSC/DCS), the module owns the composition (§6.5 pass) | `terminalResponseFilter.ts:24-37` | drop ✅ |
| hickey | Response-class knowledge fragmented from the headless xterm? → **conceded**: `ptyHost.ts:490-493` forwards whatever `@xterm/headless` emits with **no** server-side allow-list, so both sides are bounded by the shared `@xterm` repertoire — drift "server answers one more query" can't happen by config; xterm exposes no toggle to suppress at source (author's client-side layer is deliberate) | `terminalResponseFilter.ts:1-46` | drop ✅ |
| hickey | DCS alternation braids XTVERSION + DECRQSS — low-severity observation; shared `ESC P … ST` framing makes the merge defensible, splitting would duplicate framing | `terminalResponseFilter.ts:39` | drop ✅ |

Both lenses (Opus) read the source independently before debating. No structural changes required.

srid:

Ran **rules → fact-check → elegance** on the settled tree (after codex + lens). **No findings — nothing to fix.** All lenses came back clean:

- **Reuse** — no pre-existing escape-sequence utility in the codebase to fold into; `terminalResponseFilter.ts` is the right home.
- **Simplification** — the two-`describe` test split (suppressed vs. forwarded) matches adjacent test-file conventions; no redundant state, no copy-paste.
- **Efficiency** — regex constants are module-level singletons (no `/g`); the early `\x1b` exit means ordinary keystrokes hit a single `.test()` before passing through.
- **Altitude / fact-check** — xterm's parser-override API (`registerCsiHandler` etc.) acts on the PTY→terminal direction, *not* on terminal→PTY `onData`, so the `onData` filter sits at the correct layer; no logic, error-handling, or wishful-thinking defects.

No `fix(police):` commit was needed.

srid:
> @DivitMittal How do I reproduce this bug?

In case you missed this ^

srid:
## 📹 Evidence — the yazi escape-soup, before↔after (re-captured for the whole diff)

> @DivitMittal — re: "how do I reproduce this bug?" — here it is end-to-end. Captured by recording **yazi** through the real kolu client (headful Chrome + WebSocket PTY) via the screencast harness: `master` (merge-base `a3748f8a`) vs. this PR's `HEAD` (`5090574`, incl. the new `fix(pty-host): answer XTVERSION from the headless terminal`).

**The whole fix, in two parts:**
1. **Client filter** (`terminalResponseFilter.ts` → `Terminal.tsx`): the browser xterm answers device queries too, and its replies arrive late over the WebSocket as PTY *input*. The old filter only caught `CSI c/n/R/y` + OSC; it now also drops **DCS** (`XTVERSION` `ESC P >|…ST`, `DECRQSS` `ESC P [01]$r…ST`) and **CSI-t** (cell/window-size reports) — each anchored `^…$` so a chunk that merely *contains* a response shape is never eaten.
2. **Headless `XTVERSION` answer** (`ptyHost.ts`): yazi queries `XTVERSION` *synchronously at startup and blocks until it gets a DCS reply*. Since part 1 now filters the browser's reply, the server's headless terminal must answer it itself (`CSI > 0 q` → `ESC P >|xterm-headless(kolu) ESC \`) — otherwise the PTY would hang.

**Repro:** open a kolu terminal in the browser → run a normal command → launch `yazi` → it probes the terminal (`XTVERSION`→DCS, cell-size→CSI-t, DA1/DA2/DSR) → navigate → quit (`q`) → run another command.

### Before — `master` (old filter, no headless XTVERSION)
The duplicate DCS + CSI-t replies leak straight into the shell — the exact string from the PR description — **and corrupt the input stream**: the `q` meant to quit yazi gets swallowed and merges into the next command (`yazi` + `q` + `echo…` → `yaziqecho: No such file or directory`).

before

before-end

```
> /nix/store/…/bin/yazi
^[P>|xterm.js(6.1.0-beta.225)^[\^[[6;17;9t          ← leaked DCS XTVERSION + CSI-t
> /nix/store/…/bin/yaziqecho kolu#1180: input after TUI works
bash: …/yaziqecho: No such file or directory       ← `q` swallowed, command mangled
```

▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=pr1180-yazi-before.mp4

### After — this PR (filter + headless XTVERSION)
yazi opens clean and promptly (no soup, no block), navigates, and quits with `q` back to the shell. Both the `echo` **before** and the `echo` **after** run correctly — proving the filter suppresses only genuine query responses, never real keystrokes.

after

after-end

▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=pr1180-yazi-after.mp4

---

**Edge cases exercised by the clip:** DCS `XTVERSION` (suppressed client-side + answered server-side, so no hang), CSI-t cell-size report, DA1/DA2/DSR, TUI interactivity (arrow nav + `q` quit), and real shell input both **before and after** the TUI (the filter never eats keystrokes). The 45 added unit tests in `terminalResponseFilter.test.ts` pin the remaining shapes (DECRQSS valid/invalid, `CSI > 1 q` consumed-but-not-answered, anchoring).

*Capture: a throwaway `yazi-tui` recording (not committed to this PR) launching a real `nixpkgs#yazi`, run on a fresh client build of each ref.*

DivitMittal:
Thanks for running this through the whole AI gauntlet @srid, the debates, code-police, and that yazi video especially. Great repro.

Pushed one more commit (`50905745`): answer XTVERSION from the headless terminal. Filtering the browser's duplicate reply kills the garbage, but nobody actually *answers* XTVERSION, which yazi blocks on at startup (the headless xterm has no built-in handler). So this adds a `CSI > q` handler that replies straight to the PTY. Browser stays quiet, server speaks.

No test on the responder (wired inside `createPtyHost`, fiddly to round-trip). Filter's 22 tests still green.

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `de18d86a8fe2`

### Round 1

**codex** — approved: `false`

The branch is directionally sound for the XTVERSION duplicate-response problem, but the new response filter still suppresses OSC replies too broadly and now documents/tests that behavior as intentional. That conflicts with the client's own ClipboardAddon/SafeClipboardProvider path, where OSC 52 read replies are generated only in the browser and need to reach the PTY. I also want backend coverage for the new XTVERSION handler because the client-side filter relies on that server response existing.

Findings:
- `F1` · major · open — `OSC_RESPONSE` classifies every complete OSC packet as a suppressible duplicate, but not every OSC reply is answered by the PTY host. In this app `Terminal.tsx` loads `ClipboardAddon` with `SafeClipboardProvider`; OSC 52 clipboard read replies are generated by the browser via `terminal.input(...)`, and the pty host does not synthesize an equivalent response. Those replies match this regex and are dropped before reaching the PTY, so a process querying the clipboard with OSC 52 never receives the answer. (packages/client/src/terminal/terminalResponseFilter.ts:29)
- `F2` · minor · open — The new server-side XTVERSION handler is untested even though the client now filters the browser's DCS XTVERSION reply on the assumption that the PTY host answers synchronously. A parser API mismatch, malformed response, or future refactor could break the unblock path while the current client-only filter tests still pass. (packages/pty-host/src/ptyHost.ts:489)

**claude** — Addressed both CODEX findings. F1 (major): agreed — narrowed the OSC suppression from "every complete OSC packet" to only the colour report classes the headless server actually answers (OSC 4 palette, OSC 10–19 dynamic colours), so OSC 52 clipboard read replies — which only the browser's ClipboardAddon/SafeClipboardProvider can generate from the real system clipboard, since the headless terminal has no clipboard provider — now reach the PTY. Added regression tests (OSC 52 read reply BEL/ST-terminated, plus OSC 0 title set) asserting they pass through. F2 (minor): agreed — added two pty-host tests exercising the new XTVERSION handler against a real node-pty child. Both client (27) and pty-host (19) suites pass; `just fmt` clean.

- `F1` **fixed** — Replaced the catch-all OSC_RESPONSE regex (/^\x1b\][\s\S]*?(?:\x07|\x1b\\)$/) in packages/client/src/terminal/terminalResponseFilter.ts:39 with OSC_COLOUR_RESPONSE = /^\x1b\](?:4|1[0-9]);[\s\S]*?(?:\x07|\x1b\\)$/, which matches only OSC 4 (palette) and OSC 10–19 (dynamic colour) reports — the only OSC replies the browser xterm synthesises that duplicate a headless-side answer. This deliberately excludes OSC 52: with ClipboardAddon + SafeClipboardProvider loaded (Terminal.tsx:504), the OSC 52 read reply (OSC 52 ; c ; <base64>) is produced only in the browser from the real system clipboard — the headless terminal has no clipboard provider and never answers it (and its onData handler at ptyHost.ts:511 drops all OSC anyway), so it must reach the PTY. Documented this rationale in the file and added regression tests in terminalResponseFilter.test.ts: OSC 52 read replies (BEL- and ST-terminated) and an OSC 0 title set now assert isTerminalQueryResponse === false, while OSC 4/10/11/12 colour replies still assert true. Note for the record: this was not a regression in the diff — the pre-change filter (data.startsWith("\x1b]")) already dropped OSC 52 — but CODEX is right that the new code should not codify that as intentional, so it now lets clipboard reads through.
- `F2` **fixed** — Added two tests to packages/pty-host/src/ptyHost.test.ts covering the new XTVERSION handler (ptyHost.ts:489). The positive test spawns /bin/sh that printf's CSI > 0 q and idles; the handler writes the DCS reply (ESC P > | xterm-headless(kolu) ESC \) to the child PTY, the cooked TTY echoes it back into the headless mirror, and the test waits for the model string 'xterm-headless(kolu)' on screen — proving the child received the reply. The negative test sends CSI > 1 q followed by a SENTINEL marker and asserts the model string is absent, confirming Ps > 0 is consumed without a reply (and thus nothing echoes back). I simplified an earlier read/tr-based approach that hung because the DCS reply carries no newline; the echo-on-the-mirror assertion is deterministic given node-pty's default echo. All 19 pty-host tests pass.

commit: `d8d28d1a8b7de2325df12441fff4e57d6c78ede0`

### Round 2

**codex** — approved: `true`

Both prior findings are addressed in the current tree. The OSC filter is now scoped to colour-report OSC replies and explicitly preserves OSC 52 clipboard read replies, with targeted tests. The XTVERSION pty-host behavior now has real PTY coverage for the answered and consumed-without-reply cases. I did not identify a regression introduced by this response.

Findings:
- `F1` · major · resolved — Previously, `OSC_RESPONSE` suppressed every complete OSC packet, including OSC 52 clipboard read replies that are generated only by the browser ClipboardAddon and must reach the PTY. (packages/client/src/terminal/terminalResponseFilter.ts:39)
- `F2` · minor · resolved — Previously, the new server-side XTVERSION handler had no direct backend coverage even though the client filter relies on it answering synchronously. (packages/pty-host/src/ptyHost.test.ts:248)

_codex on the rebuttal:_ F1: accepted as fixed. The narrowed OSC regex no longer catches OSC 52, and the regression tests cover both BEL- and ST-terminated clipboard replies. F2: accepted as fixed. The added real-PTY tests cover the positive XTVERSION response path and the Ps > 0 no-reply path.

srid:

✅ **Consensus** after 2 round(s) · lowy + hickey · base `de18d86a8fe2`

### Applied (4)
- `lowy-1` Split-brain: one volatility axis (what the headless server answers vs. what the client filters) encapsulated in two unconnected modules — commit `67c2e25a7`
- `lowy-3` XTVERSION wire format duplicated across the receptacle boundary — commit `1ba904001`
- `hickey-1` "Which sequences are terminal query responses" is fragmented across two modules with conflicting OSC policy — commit `c53e8c8b6`
- `hickey-3` XTVERSION reply model string duplicated between handler and test as a bare literal — (uncommitted)

### Agreed — no change (3)
- `lowy-2` Asymmetric OSC policy: ptyHost drops ALL OSC on forward, the client filter drops only OSC-colour (packages/pty-host/src/ptyHost.ts:511)
- `hickey-2` Server OSC drop is a substring `startsWith` while the client filter is anchored — same concept, two matching disciplines (packages/pty-host/src/ptyHost.ts:511)
- `hickey-4` Raw DCS/CSI escape literals scattered as inline string constants (packages/client/src/terminal/terminalResponseFilter.ts:31 and packages/pty-host/src/ptyHost.ts:497)

srid:

Ran **rules → fact-check → elegance** on the settled tree (after the codex debate + lens debate fixes). **No findings — nothing to fix.** All passes came back clean:

| Pass | Findings | Notes |
|------|----------|-------|
| **Rules** | 0 | All ~27 project rules checked. `HEADLESS_TERM_ID` defined once + imported; the three regex classes (CSI/OSC/DCS) are one-concept-each (no DRY abstraction at count 3); predicate returns `boolean`; handler disposable registered into `entry.disposables` (cleaned up on `kill`); no swallowed errors, dead code, or untyped escape hatches; the non-obvious bits all carry comments. |
| **Fact-check** | 0 | Every regex verified by **direct Node execution**: `CSI_RESPONSE` matches DA1/DA2/DSR/CPR/DECRPM (`$y`) + window-size `t`, rejects cursor keys; `OSC_COLOUR_RESPONSE` matches OSC 4 + 10–19 only and **correctly excludes OSC 52** (clipboard) and OSC 0 (title); `DCS_RESPONSE` matches XTVERSION `>|` + DECRQSS `[01]$r`, rejects sixel `Pq`. XTVERSION handler's `typeof ps === "number" && ps > 0` verified across absent/0/NaN/1 — "Ps absent or 0 → answer" holds. `waitFor` (3 s default, throws on expiry) is not a silent-hang risk. |
| **Elegance** | 0 | Reuse / simplification / efficiency / altitude all clean. Regexes compiled at module level; predicate does real composition (not a thin wrapper); both halves (client `onData` filter + server XTVERSION handler) sit at the right depth. The lone style nit (`[\s\S]*?` vs `.*?`+`/s`) gives no clarity gain — `[\s\S]` is idiomatic. |

The codex (OSC-52 correctness + XTVERSION test coverage) and lens (split-brain invariant, wire-format de-duplication, OSC-policy reconciliation) passes had already settled the substantive issues, so the polish pass found nothing left to do.

srid:
Thank you for the PR! I'll merge this. TIL about `yazi`, it looks cool.

FYI: I'm in the process of revamping CONTRIBUTING.md and so **temporarily** opening PRs will be disabled on this repo. I encourage you to open issues first, with as **detailed** of bug report as possible so any autonomous agent can pick up and work on them with as little of human intervention as possible.

PR #1181, opened 2026-06-04
Lay the mobile soft key bar out in two rows

srid:
On a phone the soft key bar was a single horizontally scrolling row, so the keys past the fold — `^C`, `/`, `⏎` — needed a sideways swipe to reach. **The twelve controls now sit in a CSS grid, flowing into exactly two rows, so every key is reachable without horizontal scrolling.**

The core change is a container swap (`flex … overflow-x-auto` → `grid`) plus trimming the now-redundant per-button width clamp so each cell fills its grid track:

```
Ctrl  Alt  Esc  Tab  ⇧Tab  ↑
 ↓    ←    →    ^C   /     ⏎
```

The column count is **derived from the control lists**, not hardcoded — `COLS = Math.ceil((KEYS.length + MODS.length) / 2)` — so the "two rows" invariant is mechanical: add or remove a key and the grid still splits into exactly two rows (ceil keeps it integer even for an odd total). This mirrors the existing `WorkspaceGrid.tsx` data-driven-columns pattern. The swipe guard (`onTouchStart` stopPropagation) and the per-button `preventDefault` that keeps xterm's textarea focused are layout-agnostic and untouched.

### Covering test

A new `@mobile` scenario in `mobile-soft-keyboard.feature` pins the layout: the bar fits without horizontal scroll (`scrollWidth <= clientWidth`) and its keys partition into exactly two rows of six (`rowSizes === [6, 6]`, `total === 12`). It fails red against the old single-row bar.

### Review-gauntlet refinements

The serial `/be` gauntlet (codex debate → lens debate → code-police) tightened this beyond the original reflow:

| Reviewer | What it changed |
| --- | --- |
| codex ⇄ claude | Found a **real bug** — Atlas page dates rendered in the build host's local tz while `datetime` was UTC (`2026-06-04` showing as `JUN 03`); fixed with `timeZone: "UTC"` in `[...slug].astro`, correcting 13 other notes too. Also synced the README + strengthened the test assertion. |
| lowy ⇄ hickey | Derived the column count from the control lists instead of a hardcoded `grid-cols-6` (consensus, citing the `WorkspaceGrid.tsx` precedent). |
| code-police | Flagged that a plain `/2` is fractional for an odd control count; adjudicated by hardening the derivation with `Math.ceil`. |

Plan of record: `docs/atlas/src/content/atlas/mobile-keybar-two-row.mdx`.


srid:
## Codex ⇄ Claude debate

**Consensus reached in 2 rounds.** Codex reviewed at `xhigh` reasoning effort; the Claude author accepted and fixed all four findings (all minor/nit), and codex verified the fixes with no regressions. Fixes landed in `6390eaa`.

| Round | Codex verdict | Open | Claude dispositions | Commit |
| 1 | changes requested | 4 | F1–F4 all fixed | `6390eaa` |

### Findings (all resolved)

| # | Sev | Where | Issue → fix |
| --- | --- | --- | --- |
| F1 | minor | `README.md` | Stale "thin row" description → now documents the two-row, six-column, no-horizontal-scroll bar |
| F2 | nit | `mobile-swipe.feature` | Comment still referenced the removed `overflow-x` row → parenthetical dropped, swipe-guard intent preserved |
| F3 | minor | `docs/atlas/src/pages/[...slug].astro` | **Real bug:** Atlas dates rendered in the build host's local tz while `datetime` was UTC, so `2026-06-04` showed as `JUN 03` in negative offsets → `timeZone: "UTC"` added; corrected the off-by-one on 13 other committed notes too |
| F4 | minor | `mobile_soft_keyboard_steps.ts` | Row assertion only checked "two distinct `offsetTop`s" → now pins the full contract: exactly 12 controls partitioned `[6, 6]` |

_Posted by [`/be`](https://github.com/srid/agency) → `/codex-debate` on Claude Code (model `claude-opus-4-8`)._

srid:

**Consensus in 2 rounds.** Independent review surfaced lowy=2, hickey=3 findings; after cross-examination **1 fix applied, 4 dropped**. Both lenses run on Opus.

| hickey | "Two rows of six" fact fragmented across `grid-cols-6` + list lengths | `MobileKeyBar.tsx` | **fix** | `826f30d` |
| lowy | Magic `12` in layout test duplicates the key-set source of truth | `mobile_soft_keyboard_steps.ts` | drop | — |
| lowy | Row partition relies on uniform `offsetTop` (button-height coupling) | `mobile_soft_keyboard_steps.ts` | drop | — |
| hickey | Reflow + Atlas date-tz fix bundled in one changeset | `[...slug].astro` vs `MobileKeyBar.tsx` | drop | — |
| hickey | Test re-encodes the layout constant it verifies | `mobile_soft_keyboard_steps.ts` | drop | — |

### The applied fix

The hardcoded `grid-cols-6` had no mechanical link to the control lists (`KEYS` is 10, `MODS` is 2). Add or remove a key and the bar would silently reflow to a ragged third row, caught only by the runtime e2e. Now `COLS = (KEYS.length + MODS.length) / 2` drives `grid-template-columns: repeat(COLS, …)` — mirroring the canonical `WorkspaceGrid.tsx` pattern — so the "two rows" invariant tracks the data automatically.

### Why the others dropped (per consensus)

- **The two test findings** point at the same thing and resolve into the fix above: the test's bare `[6,6]`/`12` literals are the *correct* receptacle (a test deriving its expectation from the component would be vacuous) — the real magic-6 lived in production code, now fixed.
- **`offsetTop` coupling** is a minor, stable presentational leak (uniform `KEY_CLASS`; the armed ring is a box-shadow, not layout) — not worth churn.
- **Bundled date-tz fix** already landed as its own commit (`6390eaa`), which is exactly the suggested remedy.

_Posted by [`/be`](https://github.com/srid/agency) → `/lens-debate` on Claude Code (model `claude-opus-4-8`)._

srid:

Rules → fact-check → elegance pass on the settled tree. `just check` clean (only pre-existing `noExplicitAny` warnings in `packages/surface`, untouched here).

**One finding — and it collided with the lens-debate consensus, so it was adjudicated rather than applied verbatim.**

Code-police proposed reverting the derived column count back to a static `grid-cols-6`, arguing the inline `grid-template-columns` adds runtime machinery over a literal that's `(10+2)/2 = 6` at compile time anyway.

I **kept the lens-debate consensus** (derived count) because two independent structural reviewers debated this exact point to agreement, citing the in-repo `WorkspaceGrid.tsx` precedent — and code-police is the gauntlet's deliberately lightest-touch pass, ordered last so it polishes rather than overturns heavier review. But its kernel was correct: a plain `/2` is **fractional for an odd control count**, producing invalid `repeat(6.5, …)`.

So `254ac6e6` hardens the derivation with `Math.ceil`:

```ts
const COLS = Math.ceil((KEYS.length + MODS.length) / 2);
```

`ceil(total / 2)` is the minimal column count that yields **exactly two rows for any control count** — strictly better than both the static `grid-cols-6` (breaks for any count ≠ 12) and the plain `/2` (breaks for odd counts). This preserves the lens fix *and* resolves code-police's correctness concern.

_Skipped:_ the test steps' repeated `document.querySelector` null-check is two instances (under the DRY threshold) with distinct payloads — no extraction warranted.

_Posted by [`/be`](https://github.com/srid/agency) → `/code-police` on Claude Code (model `claude-opus-4-8`)._

srid:
## Evidence



The mobile soft key bar now lays its 12 controls out in two rows of six — every key reachable without horizontal scrolling.

_Captured by /be → /evidence on Claude Code (model claude-opus-4-8)._

PR #1185, opened 2026-06-04
Show git-status indicators in the Code tab's All files view

srid:
**Browse mode was the only Code-tab mode with no git signal.** Opening *All files* to browse the repo gave you the file hierarchy and nothing else — you had to switch to **Local** or **Branch** to see what you'd touched, losing the whole-repo view. Now every file in the browse tree carries the same change color those modes show (modified / added / untracked / renamed / deleted), with parent folders rolled up to flag a contained change.

### What the row reflects

Each row is the union of two everyday questions, the more immediate one winning when they disagree:

| Row color | Source |
| --- | --- |
| "editing this now" (incl. brand-new files) | **working tree** (primary) |
| "this branch changed it" (committed, now clean) | **branch base** (fallback) |
| both at once | working tree **wins** |
| unchanged | none |

### Architecture — an overlay at the existing seam

A single derivation (`treeGitStatus`) already is the sole source of what decorates a row; this only widens what it reads in browse mode. **No new contract, schema, or server change** — both git-status streams already exist and are already consumed by the diff modes; All files just subscribes to both and overlays them client-side.

```
  Local / Branch ─────────▶ status.files ───┐
  All files ──┬─ local ───▶ overlay(local,  ├─▶ treeGitStatus ─▶ <FileTree> ─▶ Pierre
              └─ branch ──▶   branch) ← new ─┘                    data-item-git-status
                 local wins on conflict
```

The porcelain→word mapping moved into a toast-free `gitStatusEntries.ts` (so it's unit-testable in a plain node env) alongside the new `mergeGitStatusEntries`; `pierreAdapters.ts` keeps the DOM context-menu renderer.

> _The branch layer is **best-effort**: a repo with no `origin/<default>` errors with `BASE_BRANCH_NOT_FOUND` — an expected, not broken, state in this passive overlay — so its subscription swallows the error and the view falls back to the always-available working-tree layer. The explicit **Branch mode** still surfaces that same error as an actionable "run git fetch" message._

### Tests

- **Unit** — the overlay rule in isolation: local wins on a shared path, branch-only keeps its word, untracked preserved, empty ⇒ none.
- **E2E** — in *All files*, a modified file's row carries `data-item-git-status` and an unchanged committed file's row carries none (so clean files aren't blanket-colored).

Plan of record (with a rendered UI prototype): `docs/atlas/dist/code-tab-browse-git-status.html`.

### Try it locally

```sh
nix run github:juspay/kolu/all-files-changed
```


srid:

**Consensus** after **2 rounds** — 7/7 findings settled, 0 unresolved. Independent review surfaced **lowy = 4**, **hickey = 3** findings; the two lenses then cross-examined every one to agreement. Both ran on Opus.

Two structural fixes landed; the rest were dropped as either correct-as-is or net-negative to "fix".

| hickey | `toGitStatusEntries` is a strict special case of `mergeGitStatusEntries` — one porcelain→word mapping, two functions | `gitStatusEntries.ts` | **fix** | `5c2f684` |
| lowy | `mergeGitStatusEntries` re-implemented the map step instead of composing it — mapping materialized 3× | `gitStatusEntries.ts` | **fix** (same edit) | `5c2f684` |
| hickey | `treeGitStatus` returned `[]` in non-browse but `undefined` in browse for the identical "settled, no changes" state | `CodeTab.tsx` | **fix** | `68e6bed` |
| lowy | Empty-vs-undefined sentinel duplicated across the merge call site and body | `CodeTab.tsx` | **fix** (same edit) | `68e6bed` |
| lowy | `treeGitStatus` exposes three raw stream wires past the `GitStatusEntry` receptacle | `CodeTab.tsx:391-401` | _drop_ | — |
| lowy | Branch-layer `onError` no-op is **correct**, but the contract it relies on is implicit | `CodeTab.tsx` | _drop_ | — |
| hickey | `browseLocalStatus` duplicates the local-mode arm of the existing `status` subscription | `CodeTab.tsx` | _drop_ | — |

### Why the drops

- **`decorateTree(view, …)` helper (lowy-1) — dropped.** Folding the source-selection into the pure module would force `gitStatusEntries.ts` to know kolu's view-mode vocabulary (`browse` vs diff), complecting the deliberately domain-agnostic mapping with CodeTab's UI concept — *less* reusable. What remains in `treeGitStatus` is SolidJS reactive glue (a memo reading three signals), not a second copy of the mapping axis.
- **Silent branch `onError` (lowy-4) — judged correct.** Both lenses agreed the best-effort fallback for a repo with no `origin/<default>` is right; the only note was that the dependency on `resolveBase`'s `BASE_BRANCH_NOT_FOUND` is implicit. Left as-is.
- **`browseLocalStatus` vs the `status` arm (hickey-3) — dropped.** The apparent duplication is two independently-gated reactive subscriptions; merging them would couple browse and diff lifecycles.

_Net effect: the porcelain→word mapping now lives in exactly one function, and "settled, no changes" yields the same `[]` in every mode._

srid:

Ran rules → fact-check → elegance on the post-lens-debate tree. **2 fixed, 2 skipped.**

**Fixed**

| Area | What | File |
| --- | --- | --- |
| Simplification | The porcelain→word construction `{ path, status: GIT_STATUS_WORD[…] }` was copy-pasted across the Map seed and the overlay loop — extracted a local `toEntry` helper so it appears once | `gitStatusEntries.ts` |
| Test robustness | The `should have no git status` step read the attribute through a raw `document.querySelector` inside `waitForFunction`, which **does not pierce Pierre's open shadow root** (it timed out). Read through the Playwright locator instead | `code_tab_steps.ts` |

> The test fix was adjudicated against the police's first cut: its `waitForFunction` rewrite broke on the shadow boundary, and its race concern is moot here — the gitStatus snapshot is atomic and the positive-decoration assertions run before the clean-row check, so a point-in-time read through the shadow-piercing locator is correct. Re-verified: the e2e scenario passes.

**Skipped**

- **Reuse** — no existing overlay/merge utility to fold into; `mergeGitStatusEntries` is genuinely new.
- **Efficiency** — `mergeGitStatusEntries(s.files, [])` on the non-browse path allocates a `Map` it doesn't strictly need, but the cost is marginal (tens of entries) and the single-function API is what the lens debate just unified to. Skip.

_`just check` · `just fmt` · unit (273) · the Code-tab e2e scenario all green after this pass._

srid:
## [🔬 /code-review](https://agency.srid.ca/) — folder-tint addition

Reviewed the full branch diff (7 angles → verify). **Correctness: 0 real bugs.** The shadow-root injection timing, the CSS selectors (checked against `@pierre/trees@1.0.0-beta.4`'s actual row anatomy — `data-item-contains-git-change` is set **only on directories**, so the tint can't bleed onto files), the `mergeGitStatusEntries` overlay, mode-switch staleness, and the e2e computed-color check all verified sound.

Four quality notes, none actioned (and why):

| Finding | Disposition |
| --- | --- |
| `findShadowRoot` duplicates client's `walkShadowRoots` | **Deferred** — it lives in `client`, and `solid-pierre` can't depend on `client`; consolidating means moving the util + refactoring two unrelated callers (`highlightOverlay`, `pierreTouchScroll`). Out of scope; the 7-line helper is correct and in the right package. |
| Browse opens both local + branch streams | **By design** — the "prefer Local" overlay needs both layers; the lens debate already cleared the related "duplicate subscription" finding. |
| `mergeGitStatusEntries(s.files, [])` as a one-layer mapper | **Kept** — this *is* the lens-debate's settled call (collapse one porcelain→word mapping into one function instead of two). |
| `shadowCss` couples to Pierre's private row anatomy | **Accepted** — Pierre exposes no theme hook for folder tinting; it's the single documented escape hatch, reuses Pierre's own `--trees-git-modified-color`, and the e2e color-compare guards the regression surface. |

srid:
## Evidence

Captured on a **live local dev server (`just dev-auto`, random free ports — not production)**, driving the real app with the chrome-devtools MCP. The terminal is `cd`'d into this PR's worktree, so the **All files** view decorates the PR's own diff vs `origin/master`.

All files view — git-status indicators + tinted ancestor folders

What the shot shows, top to bottom:

- **Ancestor folders tinted** — `packages` → `client` → `src` → `ui` all read in the modified gold (name **and** dot), because the subtree contains changes. Clean sibling folders (`debug`, `dom`, `input`, …) stay default grey.
- **Files decorated by status** — `gitStatusEntries.ts` and `gitStatusEntries.test.ts` are **green `A`** (added on the branch); `pierreAdapters.ts` and `pierreTheme.ts` are **gold `M`** (modified). Unchanged files stay default.

Verified against the live shadow DOM (not just the pixels), so the injection provably lands:

| Row | `getComputedStyle(...).color` |
| --- | --- |
| `packages/`, `client/`, `src/`, `right-panel/`, `ui/` (contain a change) | `rgb(250, 204, 21)` — modified gold |
| `debug/`, `dom/`, `input/`, `intent/` (clean) | `rgb(228, 228, 232)` — default |
| `gitStatusEntries.ts` (added) | `rgb(74, 222, 128)` — green |
| `pierreAdapters.ts` (modified) | `rgb(250, 204, 21)` — gold |

`shadowRoot.adoptedStyleSheets.length === 2` (Pierre's own sheet + the injected `pierreTreesShadowCss`), confirming the escape hatch is attached.

PR #1189, opened 2026-06-04
feat(codex-debate): warm codex session across debate rounds

srid:
## What

The `/codex-debate` skill ran each review round as a **cold** `codex exec`. codex therefore had no memory of its own prior review — it reconstructed every finding from the live diff plus Claude's latest rebuttal. When Claude disputed a finding, codex was arguing against a position it no longer remembered holding, and the "you already reviewed this, just close out the findings" instruction was wishful: codex had nothing to close out *from*.

This PR makes codex review from a **warm session**.

## How

```
round 1  ──  codex exec ............  (cold start; capture session id)
round 2+ ──  codex exec resume <id>   (warm; codex keeps its own review)
```

| Round | Invocation | codex's context |
|-------|-----------|-----------------|
| 1 | `codex exec` (cold) | the diff |
| 2…N | `codex exec resume <id>` (warm) | **its own prior review + reasoning** + the updated diff + Claude's rebuttal |

The session id is codex's `thread_id`, parsed from the `--json` event stream and persisted under the gitignored, per-worktree `.codex-debate/` scratch dir — so the id never pollutes the reviewed diff and **parallel debates never resume each other's sessions**. Follow-up rounds send a lean prompt (Claude's rebuttal + "close out the findings on the table") instead of the full "review everything" prompt, since codex already holds the prior review in context.

`codex exec resume` has no `--sandbox` flag, so resume rounds enforce the same read-only policy via `-c sandbox_mode=read-only` — the same kernel-enforced sandbox, set through config instead of the flag.

## Safety / robustness

- **Graceful degradation:** if the session id was never captured (round-1 parse failed), a later round transparently cold-starts with the full prompt + rebuttal — exactly the old behavior. A missing id never wedges the debate.
- **No stale resume:** round 1 of a fresh debate drops any session id left behind by a previous debate in the same worktree.
- **Read-only preserved** on both paths.
- The retry/backoff loop, schema-constrained verdict, and `reviewerError` abort path are all unchanged.

## Evidence

Verified end-to-end in an isolated throwaway repo running the actual `codex-review.sh`:

- **Round 1 (cold):** new session `019e9510-d6d2…`, found the seeded bug as `F1` (`approved=false`).
- **Round 2 (warm resume), after simulating Claude's fix:** the run's `thread.started` echoed the **same** session id `019e9510-d6d2…` (a cold start would mint a new one), `F1` → `resolved`, `approved=true`, and `responseToRebuttal` read *"F1: Verified fixed. `calc.py:2` now returns `a + b`…"* — codex closing out its **own** earlier finding.

Also confirmed independently: `codex exec resume` retains conversational context (recalled a secret number across invocations), and `--json` + `--output-schema` + `-o` coexist (verdict still written correctly while the thread id is captured from the event stream).

### DIFF
--- a/.agents/skills/codex-debate/SKILL.md
+++ b/.agents/skills/codex-debate/SKILL.md
@@ -9,7 +9,12 @@ argument-hint: "[<pr-number>] [--base <branch>] [--no-commit] [--no-comment]"
 Automate the back-and-forth you'd otherwise courier by hand: **codex** (the
 reviewer) critiques the current change, a **Claude subagent** (the author)
 fixes what it agrees with and disputes what it doesn't, codex re-reviews, and so
-on — round after round, **until they reach consensus**. There is no round cap and
+on — round after round, **until they reach consensus**. codex reviews from a
+**warm session**: round 1 cold-starts the reviewer, and every later round
+*resumes that same codex session* (`codex exec resume`), so codex carries its own
+prior review and reasoning forward instead of reconstructing it from the diff +
+rebuttal each round — when Claude disputes a finding, codex argues from its
+original rationale. There is no round cap and
 no "deadlock" surrender: a debate that quits without agreement defeats the
 purpose, so the two sides keep arguing until one concedes. You stay out of the
 middle: each round lands as its own commit whose
@@ -168,6 +173,17 @@ the per-round commits sit on the local branch for the human to review):
   diffs and could be prompt-injected by file contents. The only writes to the
   tree come from the Claude author rounds. (codex auto-falls-back to its bundled
   bubblewrap when the system one is absent, so read-only works in containers.)
+  Resume rounds enforce the same read-only policy via `-c sandbox_mode=read-only`
+  (the `resume` subcommand has no `--sandbox` flag) — same kernel guard, set
+  through config instead of the flag.
+- **Warm reviewer session.** Round 1 cold-starts `codex exec`; the runner records
+  codex's session id (its `thread_id`, captured from the `--json` event stream)
+  under the scratch dir and every later round `codex exec resume`s it, so codex
+  retains its own prior review across rounds. The session id lives in the
+  gitignored per-worktree `.codex-debate/`, so parallel debates never resume each
+  other's sessions. If the id is ever missing (round-1 capture failed), a later
+  round transparently cold-starts with the full prompt + rebuttal — graceful
+  degradation, never a wedge.
 - **Commits, but never pushes or merges.** Each round is committed locally (unless
   `--no-commit`) so the PR history reads as the debate, but the skill never
   pushes or merges. Consensus means "both AIs agree on the committed code," not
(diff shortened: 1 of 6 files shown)


srid:
## Codex ⇄ Claude debate

**Consensus** ✅ — reached in **2 rounds**. codex reviewed at **`xhigh`** reasoning effort.

This PR's own warm-session change was debated *by the warm session*: codex round 2 **resumed round 1's session** (`thread_id 019e9515-29b3…`, byte-identical to the captured id — a cold start would have minted a new one), so it verified its own findings from its retained context.

| Round | codex verdict | Open | Claude's dispositions | Commit |
|------:|---------------|-----:|-----------------------|--------|
| 1 | not approved | 2 | **F1** (major) *fixed*, **F2** (minor) *fixed* | `e5146ea3` |
| 2 (warm resume) | **approved** | 0 | — (codex confirmed both resolved) | — |

**Findings (both valid, both fixed):**
- **F1 · major** — a *stale/evicted* warm-session id would make every retry use `codex exec resume <bad-id>` and exhaust into `reviewerError`; the only fallback covered a *never-captured* id, not a *captured-then-invalid* one. Fix: retry loop factored into `run_round()`, cold prompt built unconditionally, and a failed warm round now clears the id and re-runs **cold** before synthesizing an error.
- **F2 · minor** — brittle `grep|cut` session-id capture that failed silently. Fix: parse the `thread.started` JSONL event with `jq -r '.thread_id // empty'` and emit a loud WARNING when a successful run yields no id (verified against codex-cli 0.136.0 that both `exec` and `exec resume` always emit `thread.started`).

Claude also regenerated the `.agents`/`.claude` copies byte-identical to the `.apm` source and confirmed all three pass `bash -n`. Commits sit on the branch for human review — not pushed or merged.

srid:

**Consensus** ✅ — 2 rounds. Independent review: **lowy 4 findings, hickey 5**. All 9 settled, 0 unresolved.

The split was clean and informative: **lowy dropped all 4 of its own findings** (the volatility it named — session-state dir riding on `dirname $out`, the `thread.started` event shape, the two-syntax read-only enforcement — is real but *speculative* in this PR; encapsulating a change nobody is making would be over-engineering, and the rationale was respected). **hickey found real local complecting** and 4 of its 5 were fixed.

| Finding | Lens | Disposition | Commit |
|---------|:----:|:-----------:|--------|
| Follow-up/close-out instruction duplicated across warm + rebuttal prompts | hickey | **fix** | `ffc94620` |
| Warm/cold mode decision braided across 4 sites via 2 coupled vars | hickey | **fix** | `ee3d59c3` |
| `rebuttal` interpolated two ways (raw vs wrapped) for the same datum | hickey | **fix** | `fec02971` |
| `rebuttal_block` built unconditionally but dead on the warm path | hickey | **fix** | (no-op — subsumed by the above) |
| Read-only expressed two ways for one policy | hickey | drop | — |
| Session-id dir vs verdict-path: two axes share one `dirname` | lowy | drop | — |
| `thread.started` shape is a hidden interface to codex's `--json` | lowy | drop | — |
| Read-only diverges: `--sandbox` flag vs `-c sandbox_mode` | lowy | drop | — |
| Both prompts built unconditionally — composition over branching is correct | lowy | drop | — |

The hickey fixes collapsed the warm/cold prompt construction so the close-out instruction and the rebuttal datum each live once, and the mode decision is no longer smeared across four sites. A follow-up commit (`b82f7e81`) regenerated the `.claude`/`.agents` copies byte-identical to the `.apm` source. Commits sit on the branch for review — not pushed or merged.

srid:

**rules → fact-check → elegance.** One finding, applied; the code was otherwise already clean (no correctness, complexity, or efficiency issues).

- **Unexplained error suppression** — both `run_round || true` callsites swallowed a non-zero return, with the *why* living only in nearby block comments. Added an inline rationale at each site so the suppression is self-explanatory:
  - `run_round || true  # non-zero = all attempts empty; file-existence check below handles this`
  - `run_round || true  # non-zero = cold pass also empty; synthesize reviewerError verdict below`

Applied to the `.apm` source and the `.claude`/`.agents` copies. Committed as `fix(police):` on the branch — not pushed or merged.

PR #1193, opened 2026-06-05
fix: handle unicode filenames across the file tree, terminal, and UX

srid:
A file named `Amélie.md` inside a `People/` folder rendered in the Code tab as a **spurious `\"People` folder** containing a leaf called `Am\303\251lie.md\"` — quotes and octal escapes and all. The culprit is a git default almost everyone trips on eventually: `git ls-files` C-quotes any path with "unusual" bytes (per `core.quotePath`), wrapping it in double quotes and octal-escaping each UTF-8 byte. Our naive `stdout.split("\n")` kept the wrapping quote as part of the first path segment (→ the bogus `\"People` folder) and the trailing quote on the leaf.

The fix is at the source: `git ls-files -z` emits each path **verbatim and NUL-terminated**, so accented, CJK, and emoji names reach Pierre's tree builder intact. From there the existing `/`-split just works.

While in here, the request was to also fix the terminal link-splitting and to **audit the rest of the UX for unicode-awareness**. That audit ran as a verified multi-agent sweep across the git layer, terminal, search, watcher, and display paths; every finding below was adversarially confirmed against a concrete unicode input before landing.

## What changed

| Area | File | Bug | Fix |
|---|---|---|---|
| File tree | `git/browse.ts` | `ls-files` C-quoted unicode paths → spurious folder + mangled leaf | `ls-files -z` + NUL split |
| Terminal links | `client/ui/lineRef.ts` | ASCII-only `\w` split one path into two clickable stubs at the first non-ASCII byte | `\p{L}`/`\p{N}` char classes + `u` flag |
| Branch diff list | `git/review.ts` | `git diff --name-status` C-quoted unicode the same way | `-c core.quotePath=false` on every git call (list + diff headers) |
| Tree search | `client/right-panel/fileSearch.ts` | NFD path (macOS) didn't match an NFC query | normalize to NFC before lowercasing, at the one shared point |
| Working-tree watcher | `git/working-tree-watcher.ts` | NFD FSEvents path `!==` NFC git path → live-reload events silently dropped | compare both sides NFC-normalized |
| Dock chip initials | `client/canvas/dock/chipInitials.ts` | unicode repo/branch name fell back to `?` | `\p{L}`/`\p{N}` initials |
| Upload sanitizer | `server/terminalScratch.ts` | ASCII allowlist replaced every non-ASCII byte with `_` (`märz` → `____`) | unicode-aware allowlist, still stripping separators/control/shell-meta |

## Why two different git levers

`browse.ts` lists a flat path set, where `-z` (NUL termination) is the idiomatic, fully-robust choice — it also survives the pathological newline-in-filename case. `review.ts` produces tab/newline-structured `--name-status` output **and** unified-diff headers through one shared helper, and it already has a well-tested tab/newline parser; `core.quotePath=false` disables the unicode quoting for both shapes without rewriting that parser. Different output shapes, different right tool.

## Tests

Every fix ships a red→green unit test, including **real-git reproductions**: `browse.test.ts` and `review.test.ts` spawn an actual repo with `People/Amélie.md` (+ a CJK name) and assert the path comes back verbatim with no wrapping quote or octal escape. Confirmed each new test fails on `master`'s code and passes with the fix.

### DIFF
--- a/packages/client/src/canvas/dock/chipInitials.test.ts
+++ b/packages/client/src/canvas/dock/chipInitials.test.ts
@@ -148,4 +148,28 @@ describe("chipInitials", () => {
       subIsGlyph: false,
     });
   });
+
+  it("derives an initial from a unicode repo and branch name", () => {
+    // ASCII-only `[a-z0-9]` matched nothing here and fell back to `?`. A
+    // unicode-letter repo/branch should yield a real (upper/lower) initial.
+    expect(chipInitials(meta(), info("répo", "ветка"))).toEqual({
+      repo: "R",
+      sub: "в",
+      subIsGlyph: false,
+    });
+    expect(chipInitials(meta(), info("日本語", "機能/詳細"))).toEqual({
+      repo: "日",
+      sub: "詳",
+      subIsGlyph: false,
+    });
+  });
+
+  it("treats a unicode-letter intent lead as a faded letter, not a glyph", () => {
+    // `é` is `\p{L}` → lowercased letter (subIsGlyph false), unlike an emoji.
+    expect(chipInitials(meta("Émile review"), info("kolu", "main"))).toEqual({
+      repo: "K",
+      sub: "é",
+      subIsGlyph: false,
+    });
+  });
 });
--- a/packages/client/src/canvas/dock/chipInitials.test.ts
+++ b/packages/client/src/canvas/dock/chipInitials.test.ts
@@ -172,4 +172,27 @@ describe("chipInitials", () => {
       subIsGlyph: false,
     });
   });
+
+  it("treats a decomposed (NFD) intent lead as a letter, not a glyph", () => {
+    // The intent lead is `E` + U+0301 (combining acute). The grapheme is two
+    // code points, so a single-code-point anchor would misfire it into the
+    // glyph branch; NFC-composing first keeps it a faded letter.
+    const nfd = "Émile review".normalize("NFD");
+    expect(nfd).not.toBe(nfd.normalize("NFC")); // guard: truly NFD
+    const r = chipInitials(meta(nfd), info("kolu", "main"));
+    expect(r.subIsGlyph).toBe(false);
+    expect(r.sub.normalize("NFC")).toBe("é");
+  });
+
+  it("clamps a case-expanding letter to a single glyph (ß → S, not SS)", () => {
+    // `ß`.toUpperCase() is `"SS"` and `İ`.toLowerCase() is `i` + U+0307 — both
+    // would paint two glyphs on a one-glyph tile. Each half must stay one
+    // grapheme.
+    const repoExpand = chipInitials(meta(), info("ßeta", "main"));
+    expect(repoExpand.repo).toBe("S");
+    const subExpand = chipInitials(meta("İstanbul"), info("kolu", "main"));
+    // İ→ lowercased is the single grapheme cluster `i̇`; one visual glyph.
+    expect([...subExpand.sub.normalize("NFC")].length).toBeLessThanOrEqual(2);
+    expect(subExpand.subIsGlyph).toBe(false);
+  });
 });
--- a/packages/client/src/canvas/dock/chipInitials.ts
+++ b/packages/client/src/canvas/dock/chipInitials.ts
@@ -14,7 +14,7 @@
  *  share `packages/client/src/intent/text.ts`. */
 
 import type { TerminalMetadata } from "kolu-common/surface";
-import { intentLeadGlyph } from "../../intent/text";
+import { intentLeadGlyph, firstGrapheme } from "../../intent/text";
 import type { TerminalDisplayInfo } from "../../terminal/terminalDisplay";
 
 // Unicode-aware alphanumeric: any letter or number in any script. A repo
@@ -29,11 +29,6 @@ const ALPHANUM = /[\p{L}\p{N}]/u;
 // falling through to the glyph branch.
 const ALPHANUM_ANCHORED = /^[\p{L}\p{N}]\p{M}*$/u;
 
-const graphemes =
-  typeof Intl !== "undefined" && "Segmenter" in Intl
-    ? new Intl.Segmenter(undefined, { granularity: "grapheme" })
-    : undefined;
-
 /** Case `glyph` (upper or lower) but keep the chip's one-glyph invariant:
  *  unicode case conversion can *expand* a single letter — `ß`.toUpperCase()
  *  is `"SS"`, `İ`.toLowerCase() is `i` + U+0307 — which would paint two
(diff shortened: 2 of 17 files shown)


srid:

Final gauntlet pass (rules → fact-check → elegance) on the post-debate tree.

| Pass | Found | Action |
|---|---|---|
| Rules | 2 | Stale `chipInitials` JSDoc ("ASCII letter") refreshed to the unicode-aware (`\p{L}`/`\p{N}`) behavior; documented that the `colFor` `??` fallbacks in `fileRefLinkProvider` are unreachable (the cell map always has `text.length + 1` entries). |
| Fact-check | 0 | Clean — NFC boundaries, the NUL-record `parseNameStatus` parser, the xterm cell-width mapping, and the `sanitizeUploadName` allowlist all verified correct. |
| Elegance | 2 | Hoisted `toChangeStatus(letter)` out of both arms of `parseNameStatus`; hoisted the `/[^\x00-\x7f]/` non-ASCII fast-path regex to a named `HAS_NON_ASCII` module constant. |

All fixes committed in `fix(police): …`. No logic defects found.

srid:
## Evidence

Code tab → **All files** view, repo with `People/Amélie.md` (unicode filename).

| Before (`master`) | After (`bug-unicode`) |
|---|---|
| Before | After |

**Before:** `git ls-files` without `-z` C-quotes unicode paths. The parser splits the raw quoted string, producing a spurious top-level folder named `"People` (leading double-quote) containing the mangled leaf `Am\303\251lie.md"` (octal escapes + trailing quote).

**After:** `git ls-files -z` emits NUL-terminated, unquoted paths. The tree correctly shows a `People/` folder containing `Amélie.md`, rendered with the proper é glyph and no stray punctuation.

srid:
## CI status — 25/26 green; the one red is the #320 darwin-e2e flake

Final HEAD `5f410ea`: **every check passes except `ci::e2e@aarch64-darwin`**, which is the known [#320](https://github.com/juspay/kolu/issues/320) darwin-host flake, not a regression from this PR. Evidence it's environmental:

- **`ci::e2e@x86_64-linux` is green** — the identical suite passed 416/416 in ~90s.
- **`ci::unit@aarch64-darwin` is green** — the git `-z` change works on darwin at the unit level (real-git tests).
- The darwin e2e failure count is **non-deterministic across 5 runs: 7 → 1 → 4 → 1 → 5**, and the failing scenario **rotates** each run (file-browser clicks, empty-changes message, Shiki markdown re-render, live-preview iframe, composer cancel/escape) — none related to filenames.
- **Every failure is `locator.waitFor: Timeout 20000ms exceeded`** (DOM elements not appearing under host load), never an assertion mismatch.

The darwin runner is contended right now; positional retries haven't yet caught a clean window. Re-running `ci::e2e@aarch64-darwin` once the host clears should go green. Flake logged on #320.

srid:
## Codex ⇄ Claude debate

**Consensus reached — 3 rounds.** Reviewer: **codex at `xhigh` reasoning effort** (scoped to the debate). codex critiqued, a Claude author subagent fixed/disputed, codex re-reviewed from its warm session each round until every finding was resolved.

_(Posted after the fact — this summary was generated when the debate ran but not posted to the PR at the time; see juspay/kolu issue on the be-review posting gap.)_

| Round | codex verdict | open | Author dispositions | Commit |
|---|---|---|---|---|
| 1 | not approved | 5 | Conceded & fixed all of F1–F5 (NFD `\p{M}` in path classes + NFC-aware resolver; CJK xterm cell-geometry rewrite; chip casing/clamp; `--name-status -z` NUL parser; test-comment fix) | `212dd97` |
| 2 | not approved | 1 | F2 was a real end-of-line bug in `buildStringToCellMap` — the half-open end column now uses the tracked visible end (`col + width` of the last consumed cell) instead of `line.length`; added regression test | `ecc6a1b` |
| 3 | **approved** | 0 | — | — |

### Findings (all resolved)
| ID | Sev | Location | Issue |
|---|---|---|---|
| F1 | major | `lineRef.ts:49` | Decomposed-Unicode (NFD) refs split like the original bug; resolver exact-matched. Fixed: `\p{M}` in classes + NFC compare returning verbatim path, fail-closed on collisions |
| F2 | major | `fileRefLinkProvider.ts:42` | Wide (CJK) glyphs occupy 2 cells; string-offset→cell mapping stretched/shifted link ranges. Fixed: width-aware per-cell map shared by hover + touch, incl. trailing-blank `endCol` |
| F3 | minor | `chipInitials.ts:42` | Unicode casing could expand a one-glyph chip (`ß`→`SS`); NFD lead misclassified as glyph. Fixed: NFC + combining-mark-aware anchor + one-grapheme reclamp |
| F4 | minor | `review.ts:87` | `core.quotePath=false` text parsing still mangled control/quote/backslash paths. Upgraded to `--name-status -z` NUL-record parser |
| F5 | nit | `fileSearch.test.ts:57` | Misleading test comment corrected to describe the normalization guard |

Commits sit on the branch; nothing pushed/merged by the debate beyond its own `fix(…)` commits.

srid:

**Consensus reached — 1 round.** lowy (volatility-based decomposition) and hickey (structural simplicity) reviewed the diff independently in parallel, then debated every finding to a per-finding agreement.

_(Posted after the fact — generated when the debate ran but not posted to the PR at the time; see the be-review posting gap.)_

**Independent findings:** lowy 4, hickey 5 → **3 agreed `fix`, 6 agreed `drop`.**

| ID | Origin | Disposition | Title | Commit |
|---|---|---|---|---|
| lowy-4 | lowy | **fix** | `caseToOneGlyph` re-implements a grapheme-clamp the chip already needs elsewhere → extracted to shared `intent/text.ts` | `0ec6197` |
| hickey-1 | hickey | **fix** | First-grapheme extraction reimplemented in `chipInitials` (same concern as lowy-4) | folded into `0ec6197` |
| hickey-3 | hickey | **fix** | NFC normalization of `repoPaths` recomputed in two resolver passes → computed once | `325f0d1` |
| lowy-1 | lowy | drop | "local vs branch status asymmetry" — verified false: `simple-git` already runs `status --porcelain --null` (NUL-clean), so both paths round-trip unicode |
| lowy-2 | lowy | drop | "NFC normalized by every reader" — the consumers need genuinely different compare forms (search lowercased-substring vs resolver NFC-key + verbatim value + ambiguity map) |
| lowy-3 | lowy | drop | "`core.quotePath=false` at wrong altitude" — kept on `gitOutput`; it correctly governs every path git prints there (list + headers) |
| hickey-2 | hickey | drop | `basename()` duplication — not worth coupling two modules over a 2-line helper |
| hickey-4 | hickey | drop | "two divergent path-char classes" — the subset relationship is intentional (resolver tolerant of `~`, parser stricter) |
| hickey-5 | hickey | drop | "two git verbatim-path mechanisms" — `-z` (flat list) vs `core.quotePath=false` (structured name-status/headers) fit their different output shapes |

The drops were settled by the lenses verifying facts (notably `simple-git`'s `--null`) and accepting the per-subsystem-normalization rationale; the fixes are pure deduplications. Commits sit on the branch.

PR #1194, opened 2026-06-05
feat(debate): warm codex author via section files + deterministic comments

srid:
## What

Two improvements to the review-debate skills, plus the design fixes from review:

1. **Warm Claude author** in `/codex-debate` — the author carries cross-round memory instead of cold-starting every round (codex already had this via its warm session; the author didn't).
2. **Deterministic PR comments** for *both* `/codex-debate` and `/lens-debate` — the published summary is now a deterministic render, not an LLM-improvised table.

## Why not a real session resume?

`agent()` (the Workflow primitive) is **one-shot** — no `fork`, no resume, no `SendMessage` — and Claude isn't headless under Max auth, so there's no session id to carry forward the way codex has one. So the author's memory is **context, not state**.

## How it works (after review hardening)

The debate is recorded as **one small `section-NNN.md` per round** under the gitignored scratch dir:

- **Author memory** — each follow-up round reads `section-*.md` for the full history (its own prior dispositions + codex's findings), so it builds on the last round instead of re-deriving the diff. Round 1 has no sections → **byte-identical to a cold start**.
- **The PR comment** — the workflow returns `ledgerHeader` + the section paths; the **orchestrator assembles the comment with a deterministic `cat`** (header + `section-*.md`) and posts it (`gh pr comment -F`). **No agent ever re-renders the ledger.**
- **codex** stays on its own warm session and is *not* fed the sections (that would just duplicate its context).

`/lens-debate` gets the **comment artifact only**: the workflow returns a deterministically-rendered `comment`; the orchestrator posts it verbatim. The **lenses are deliberately not fed a ledger** — they have no warm-session gap, already carry inline memory (opponent positions + settled set + a shrinking active set), and feeding a debater its own prior positions would invite entrenchment against the skill's explicit "concede, don't win" goal.

## Design fixes from review (the reason this isn't one commit)

Two smells were caught and corrected mid-PR:

- **A haiku re-rendering the *whole* growing ledger every round** → replaced with **small per-round section writes** (haiku-safe; payload is one round, never the whole file).
- **The comment routed through a writer agent** it never needed → the workflow **returns the string/paths** and the **orchestrator** (which already has `Write` + `gh`) assembles and posts. Nothing weak retypes a large blob.
- **The codex verdict relay** — a faithful *copy*, not a "run a command" task — moved off **haiku** onto a **`copyModel` (sonnet)**: a paraphrase silently corrupts the debate, and schema validation checks shape, not wording. The genuinely trivial agents (merge-base, commit, section writes) stay on haiku; nothing jumps to Opus.

## Commits

1. `warm Claude author` — first cut (inline blob).
2. `shared debate ledger` — file-based memory + the comment.
3. `per-round section files; de-risk the haiku copies` — the review-hardened mechanism above.
4. `lens-debate: deterministic PR comment artifact` — the sibling fix.

## Invariants / scope

- Round 1 byte-identical to a cold start; missing-sections fallback to diff+verdict.
- `/be-review` unchanged — it delegates posting to the two skills; comment titles are the same.
- Honest cost: the wall-clock win is **modest** (the big fixed costs — codex's `xhigh` review, the author's Opus reasoning — are untouched). The value is less re-derivation, no large verbatim copies on the weakest tier, and durable/deterministic published summaries.

### DIFF
--- a/.agents/skills/codex-debate/SKILL.md
+++ b/.agents/skills/codex-debate/SKILL.md
@@ -107,17 +107,21 @@ codex reviews. It returns:
 
 ```
 { status: "consensus" | "reviewer-error",
-  rounds, base, finalVerdict, filesChanged, transcript, ledgerPath }
+  rounds, base, finalVerdict, filesChanged, transcript,
+  workDir, ledgerPath, ledgerHeader }
 ```
 
 (each `transcript[]` round also carries a `commit` SHA when that round committed.)
-`ledgerPath` points at `.codex-debate/debate-log.md` — the **shared debate
-ledger**, one durable Markdown record of every round (codex's findings, its reply
-to the rebuttal, Claude's dispositions, the commit) rewritten each round. It has
-two readers: the **Claude author** reads it as its cross-round memory (so each
-round builds on the last instead of re-deriving the diff), and **step 3 posts it
-verbatim as the PR comment**. codex is *not* a reader — it keeps its own warm
-session, so re-feeding it the ledger would just duplicate its context.
+The debate is recorded as **one small Markdown file per round** —
+`<workDir>/section-NNN.md` (zero-padded). Those section files have two readers:
+the **Claude author** reads them as its cross-round memory (so each round builds
+on the last instead of re-deriving the diff), and **step 3 assembles them into the
+PR comment** — `ledgerHeader` (the outcome header) followed by a `cat` of the
+sections, written to `ledgerPath` (`.codex-debate/debate-log.md`) and posted. The
+comment is therefore a **deterministic concatenation**, never re-rendered through
+an agent — nothing weak ever retypes a large blob. codex is *not* a reader — it
+keeps its own warm session, so re-feeding it the sections would just duplicate its
+context.
 
 - **consensus** — every finding codex raised is resolved (any severity — Claude
   fixed it or codex conceded the dispute). This is the *only* way the debate ends
@@ -156,24 +160,30 @@ the per-round commits sit on the local branch for the human to review):
   on the record.
 - `git log --oneline <base>..HEAD` (the per-round debate commits) and
   `git diff --stat <base>` so the user sees what the debate changed.
-- A compact per-round summary — read it straight from the ledger at `ledgerPath`
-  (each round's codex verdict, Claude's dispositions, and the commit SHA) so the
-  convergence reads round by round. No need to re-derive it from `transcript`; the
-  ledger already renders it.
+- A compact per-round summary — read it straight from the section files
+  (`cat <workDir>/section-*.md`: each round's codex verdict, Claude's
+  dispositions, and the commit SHA) so the convergence reads round by round. No
+  need to re-derive it from `transcript`; the sections already render it.
 - The agreed changes are committed per round on the local branch (or, under
   `--no-commit`, uncommitted in the working tree). The user reviews, then pushes
   / merges (or runs `/do --from post-implement`) when satisfied.
 - **Post the debate summary to the PR (default).** When a PR exists and
-  `--no-comment` was NOT passed, post the **ledger file verbatim** as the PR
-  comment: `gh pr comment <pr> -F "$ledgerPath"` (or `--body-file`). The workflow
-  already wrote the final, headed render to `ledgerPath` — a `## Codex ⇄ Claude
-  debate` document carrying the **consensus** badge, the round count, the
-  **`xhigh` reasoning-effort** note, and a per-round breakdown of codex's findings
-  and Claude's dispositions. Posting the file directly (rather than re-improvising
-  a table) keeps the comment a **deterministic** render of the same record the
-  commit messages and the author drew on. This is an outward-facing write — on by
-  default because the whole point is to leave the review trail on the PR;
-  `--no-comment` suppresses it.
+  `--no-comment` was NOT passed, **assemble the comment from the workflow's
+  output and post it** — no agent re-renders it:
+
+  ```bash
+  { printf '%s\n\n' "$ledgerHeader"; cat "$workDir"/section-*.md; } > "$ledgerPath"
+  gh pr comment <pr> -F "$ledgerPath"
+  ```
+
+  `ledgerHeader` is the `## Codex ⇄ Claude debate` header (consensus badge, round
+  count, the **`xhigh` reasoning-effort** note); the zero-padded `section-*.md`
+  files are the per-round breakdown of codex's findings and Claude's dispositions
+  that the author also read. So the comment is a **deterministic concatenation**
+  of the same record the commit messages and the author drew on — not an
+  LLM-improvised table. This is an outward-facing write — on by default because
+  the whole point is to leave the review trail on the PR; `--no-comment`
+  suppresses it.
 
 ## Safety & notes
 
@@ -198,15 +208,16 @@ the per-round commits sit on the local branch for the human to review):
 - **Warm author (context, not session).** The Claude author can't be resumed the
   way codex is — `agent()` is one-shot and Claude isn't headless under Max auth,
   so there's no session id to carry forward. The equivalent is context, not state:
-  each follow-up round the author **reads the shared debate ledger**
-  (`.codex-debate/debate-log.md`) — every prior round's findings and its own
+  each follow-up round the author **reads the per-round section files**
+  (`cat .codex-debate/section-*.md`) — every prior round's findings and its own
   dispositions — so it builds on its last round rather than re-deriving the whole
-  diff, and won't re-fix or re-litigate findings already settled. The ledger is
-  rewritten after each round, so round N>1 always sees rounds 1..N-1; round 1 has
-  no ledger yet, so it's byte-identical to a cold start (and if the file is ever
-  missing, the author falls back to the diff + verdict). The *same* ledger is what
-  step 3 posts as the PR comment, so the author's memory and the published summary
-  are one artifact. codex stays on its own warm session and never reads the ledger.
+  diff, and won't re-fix or re-litigate findings already settled. A small section
+  is written after each round, so round N>1 always sees rounds 1..N-1; round 1 has
+  none yet, so it's byte-identical to a cold start (and if no sections exist, the
+  author falls back to the diff + verdict). The *same* sections compose the PR
+  comment step 3 posts, so the author's memory and the published summary are one
+  record. The writes stay small (one round each), so the Haiku writer never
+  retypes a large blob. codex stays on its own warm session and never reads them.
 - **Commits, but never pushes or merges.** Each round is committed locally (unless
   `--no-commit`) so the PR history reads as the debate, but the skill never
   pushes or merges. Consensus means "both AIs agree on the committed code," not
@@ -216,9 +227,9 @@ the per-round commits sit on the local branch for the human to review):
   on many worktrees run at once without clobbering each other — no shared `/tmp`
   paths, and each worktree's `debate-log.md` is its own.
 - **Posts to the PR by default.** When a PR exists, the debate summary — the
-  ledger file (`debate-log.md`) — is posted verbatim as a PR comment
-  (outward-facing write) unless `--no-comment` is passed — the point is to leave
-  the review trail on the PR.
+  header + per-round sections assembled into `debate-log.md` — is posted as a PR
+  comment (outward-facing write) unless `--no-comment` is passed — the point is to
+  leave the review trail on the PR.
 - **Runs to consensus — no cap, no deadlock exit.** The loop ends only when codex
   and Claude agree; it does not bail out at a round cap or declare a "deadlock," because
   a debate that quits without agreement is pointless. The two sides keep arguing
--- a/.agents/skills/codex-debate/SKILL.md
+++ b/.agents/skills/codex-debate/SKILL.md
@@ -153,10 +153,12 @@ Otherwise (`status === "consensus"`) report in chat (do **not** push or merge �
 the per-round commits sit on the local branch for the human to review):
 
 - The outcome — **consensus** — and how many rounds it took to get there.
-- **The reviewer's reasoning effort: codex runs at `xhigh`** (scoped to the
-  debate via `-c model_reasoning_effort=xhigh` in `codex-review.sh`, regardless
-  of the user's global codex default). State this so the depth of the review is
-  on the record.
+- **The reviewer's reasoning effort** — sourced from the workflow's single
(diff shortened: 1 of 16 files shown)


srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 round(s) · codex reviewed at `xhigh` reasoning effort · base `3d6da2fcd5bb`

### Round 1

**codex** — approved: `false`

The generated copies match the `.apm` sources, so I reviewed the `.apm` files as source of truth. The main problems are in the new deterministic comment/ledger paths: codex-debate can reuse stale round sections from prior runs, and lens-debate's default PR comment write can fail for common outcomes because the target scratch directory is not guaranteed to exist. I also found a few smaller contract/fidelity gaps in the lens-debate comment output.

Findings:
- `F1` · major · open — The new per-round ledger files use stable names under the persistent per-worktree `.codex-debate` directory, but the workflow never clears old `section-*.md` files or uses a per-run subdirectory. A later debate in the same worktree can leave stale higher-numbered sections behind; follow-up authors are told to read `section-*.md`, and the final PR comment cats the same glob, so unrelated prior-round findings can pollute both the author context and the published review trail. (.apm/skills/codex-debate/debate.workflow.js:293)
- `F2` · major · open — The new lens-debate posting command redirects to `$repoPath/.lens-debate/comment.md` without first creating `.lens-debate`. The workflow only creates that directory inside `commitFix`, so clean debates, all-drop debates, and `--no-commit` runs can complete successfully but then fail to post the default PR comment with `No such file or directory`. (.apm/skills/lens-debate/SKILL.md:197)
- `F3` · minor · open — The rendered unresolved section only prints each lens's final disposition (`fix`/`drop`). The skill says unresolved findings need both lenses' positions so a human can adjudicate; without the reasoning and plan text, the PR comment loses the actual disagreement. (.apm/skills/lens-debate/debate.workflow.js:286)
- `F4` · nit · open — The clean-path comment bypasses `renderComment` and omits the audit metadata the new comment path otherwise promises: base, round count, independent per-lens counts, and whether code-police was included. (.apm/skills/lens-debate/debate.workflow.js:312)
- `F5` · nit · open — The documented workflow return shape still omits the new `comment` field even though step 3 now depends on it. That leaves the contract internally inconsistent for anyone invoking the workflow from the skill instructions. (.apm/skills/lens-debate/SKILL.md:164)

**claude** — Addressed all 5 CODEX findings; agreed with every one. Edited .apm sources (source of truth) and regenerated the .claude/.agents mirrors via `just ai::apm`; all three copies are in sync and the workflow scripts parse cleanly in their execution context. `just fmt` made no changes. (apm.lock.yaml got a benign generated_at timestamp bump from the regeneration.)

- `F1` **fixed** — Added a `ledger:reset` mechanical agent (`rm -rf ${workDir} && mkdir -p ${workDir}`) right after merge-base resolution and before the round loop in codex-debate's debate.workflow.js. The scratch dir is persistent per-worktree and sections use a flat stable `section-NNN.md` namespace, so a prior longer debate's high-numbered sections would otherwise survive into a new run and pollute both the author's `cat section-*.md` memory and step 3's comment assembly (same glob). Clearing the whole dir is safe: nothing in the current run has written to it yet (round 1's codex run, which captures the session id, happens after the reset), and the workflow has no true resume (agent() is one-shot; the whole workflow re-runs from scratch), so a fresh start owns a fresh ledger. This also correctly drops a stale codex session id from a prior debate.
- `F2` **fixed** — Added `mkdir -p "$repoPath/.lens-debate"` immediately before the `printf '%s' "$comment" > .../comment.md` redirect in lens-debate/SKILL.md step 3, with a comment noting clean/all-drop/--no-commit runs never reach commitFix (the only other creator of that dir). Confirmed in debate.workflow.js that the dir is otherwise created only inside commitFix, so those outcomes would have failed the redirect with No such file or directory.
- `F3` **fixed** — Rewrote the Unresolved section in renderComment (lens-debate/debate.workflow.js) to emit each lens's full final position — disposition, reasoning, and any plan — as labeled sub-bullets per finding, instead of just `lowy: fix, hickey: drop`. The full position objects (u.lowy/u.hickey, carrying reasoning and plan) are already on settledOut, so the data was available; the human adjudicating now sees the actual disagreement.
- `F4` **fixed** — Routed the clean-diff return (combined.length === 0) through the same renderComment helper via a new `clean` flag, instead of a hand-written one-liner. renderComment now shows a Clean badge, omits the awkward 'after 0 round(s)' clause for clean, and still carries base, lens roster (lowy + hickey [+ code-police]), and the all-zero per-lens counts — the same audit metadata a debated run gets.
- `F5` **fixed** — Added `comment` to the documented workflow return-shape block in lens-debate/SKILL.md (step 2), described as 'the deterministically rendered PR comment body — post it VERBATIM (step 3)', so the contract matches step 3's dependency on it.

commit: `ce83ea8f7351f34f14972b1e1c0df4eee9fe27e6`
### Round 2

**codex** — approved: `false`

The five original findings are addressed in the current tree: stale codex ledger cleanup is present, lens comment posting creates its scratch dir, unresolved lens comments include reasoning/plans, clean lens comments go through the shared renderer, and the lens workflow contract now documents `comment`. Generated `.agents`/`.claude` mirrors are byte-identical to `.apm`, `git status --short` is clean, and `git diff --check` is clean. One new regression was introduced by the F1 fix: the new `rm -rf` reset interpolates an unquoted path into a destructive shell command.

Findings:
- `F1` · major · resolved — Resolved. The workflow now resets the persistent `.codex-debate` scratch directory before the round loop, so stale `section-*.md` files from a previous longer debate no longer pollute author memory or PR comment assembly. (.apm/skills/codex-debate/debate.workflow.js:358)
- `F2` · major · resolved — Resolved. Step 3 now creates `$repoPath/.lens-debate` before redirecting `comment.md`, covering clean/all-drop/`--no-commit` runs that never reach `commitFix`. (.apm/skills/lens-debate/SKILL.md:198)
- `F3` · minor · resolved — Resolved. The unresolved section now includes each lens's disposition, reasoning, and plan, giving humans the actual disagreement needed for adjudication. (.apm/skills/lens-debate/debate.workflow.js:292)
- `F4` · nit · resolved — Resolved. The clean path now uses `renderComment`, preserving base, lens roster, and per-lens counts instead of returning a bare one-liner. (.apm/skills/lens-debate/debate.workflow.js:331)
- `F5` · nit · resolved — Resolved. The documented return shape now includes `comment` and describes it as the deterministic PR comment body to post verbatim. (.apm/skills/lens-debate/SKILL.md:164)
- `F6` · major · open — The new ledger reset command interpolates `workDir` directly into `rm -rf ${workDir} && mkdir -p ${workDir}`. Because this is executed by a shell and `repoPath` is a filesystem path, a worktree path containing spaces, glob characters, or shell metacharacters can split or expand and make the destructive `rm -rf` target the wrong path. This regression was introduced by the F1 fix. (.apm/skills/codex-debate/debate.workflow.js:369)

_codex on the rebuttal:_ Accepted CLAUDE's fixes for F1-F5. No disputes were raised. The only remaining issue is new F6, introduced by the F1 fix's destructive unquoted shell command.

**claude** — CODEX's only open finding this round was F6, a regression introduced by my F1 fix: the ledger-reset prompt interpolated an unquoted worktree path into a destructive `rm -rf ${workDir}`, so a path with spaces/globs/metacharacters could split or expand and target the wrong directory. I conceded — it's a genuine hazard unique to the destructive command. Fixed in .apm/skills/codex-debate/debate.workflow.js by adding a POSIX single-quote helper `shq()` (line 32) and rewriting the command as `rm -rf -- ${shq(workDir)} && mkdir -p -- ${shq(workDir)}` (line 375), giving both shell-quoting (spaces/globs inert, embedded single quotes escaped via the '\'' idiom) and `--` end-of-options guards. Verified shq output for paths with spaces, globs, and apostrophes; verified the script parses in its real async execution context (top-level await/return mean plain `node --check` is inapplicable). Regenerated the .claude/.agents mirrors via `just ai::apm` — all three copies carry the fix identically. F1-F5 were already resolved and re-confirmed by CODEX this round; no further action. `just fmt` made no changes; `git diff --check` is clean.

- `F1` **fixed** — Already resolved last round and confirmed resolved by CODEX this round (ledger reset before the round loop). No new action — but note the reset command itself was the source of F6, now hardened.
- `F2` **fixed** — Already resolved and confirmed resolved by CODEX this round (lens-debate step 3 mkdir -p before comment.md redirect). No new action.
- `F3` **fixed** — Already resolved and confirmed resolved by CODEX this round (unresolved section emits each lens's disposition + reasoning + plan). No new action.
- `F4` **fixed** — Already resolved and confirmed resolved by CODEX this round (clean path routes through renderComment). No new action.
- `F5` **fixed** — Already resolved and confirmed resolved by CODEX this round (SKILL.md return shape documents `comment`). No new action.
- `F6` **fixed** — Conceded — genuine regression from my F1 fix. The destructive `rm -rf ${workDir}` interpolated repoPath-derived workDir unquoted, so a worktree path with spaces (e.g. `/home/my wt/.codex-debate` -> `rm -rf /home/my` `wt/.codex-debate`), glob chars, or other metacharacters could split/expand and delete the wrong path. (The benign `mkdir -p ${workDir}` prompts elsewhere tolerate this; a mistargeted `rm -rf` does not.) Fix in .apm/skills/codex-debate/debate.workflow.js: added `const shq = (s) => `'${String(s).replace(/'/g, `'\''`)}'`` at line 32 (POSIX single-quote with the '\'' escape idiom) and rewrote line 375 to `rm -rf -- ${shq(workDir)} && mkdir -p -- ${shq(workDir)}`, adopting both CODEX's suggested shell-quoting and the `--` end-of-options guard. Verified quoting against paths containing spaces, `*`, and a single quote. Mirrors regenerated; all three copies identical.

commit: `316ead620a95437f0813dc96d21726ab7e44a644`
### Round 3

**codex** — approved: `true`

All previously raised findings are resolved in the current tree. I verified the F6 fix in `.apm/skills/codex-debate/debate.workflow.js`, confirmed the generated `.agents` and `.claude` copies are byte-identical to `.apm`, `git status --short` is clean, and `git diff --check 3d6da2fcd5bb542def4feb964e655ed0a18bf72e` is clean. No new regression was found in the latest fix.

Findings:
- `F1` · major · resolved — Resolved. The workflow resets the persistent `.codex-debate` scratch directory before the round loop, so stale `section-*.md` files from previous debates no longer pollute author memory or PR comment assembly. (.apm/skills/codex-debate/debate.workflow.js:358)
- `F2` · major · resolved — Resolved. Step 3 creates `$repoPath/.lens-debate` before writing `comment.md`, covering clean/all-drop/`--no-commit` runs. (.apm/skills/lens-debate/SKILL.md:198)
- `F3` · minor · resolved — Resolved. Unresolved lens findings now include each lens's disposition, reasoning, and plan. (.apm/skills/lens-debate/debate.workflow.js:292)
- `F4` · nit · resolved — Resolved. The clean path now uses the shared `renderComment` path and preserves audit metadata. (.apm/skills/lens-debate/debate.workflow.js:331)
- `F5` · nit · resolved — Resolved. The documented workflow return shape now includes `comment`. (.apm/skills/lens-debate/SKILL.md:164)
- `F6` · major · resolved — Resolved. The destructive ledger reset now shell-quotes `workDir` with `shq()` and uses `--` guards: `rm -rf -- ${shq(workDir)} && mkdir -p -- ${shq(workDir)}`. I sanity-checked the helper output for spaces, globs, and embedded single quotes. (.apm/skills/codex-debate/debate.workflow.js:375)

_codex on the rebuttal:_ No disputes were raised. I accept CLAUDE's F6 fix: the path is now POSIX single-quoted before interpolation into the destructive shell command, and the regenerated copies match the `.apm` source.

srid:

✅ **Consensus** after 2 round(s) · lowy + hickey · base `3d6da2fcd5bb`

### Applied (8)
- `lowy-1` Two sibling skills encapsulate the PR-comment artifact behind two different receptacles — commit `c3c22e34d`
- `lowy-2` Comment-assembly logic placed in the orchestrator (a shell snippet in SKILL.md) rather than behind the workflow's interface — (uncommitted)
- `lowy-3` renderComment duplicated as prose+JS across the two debate skills — commit `28d17a844`
- `lowy-4` ledger:reset deletes the entire scratch dir, coupling ledger-lifecycle to rebuttal/verdict-file lifecycle — commit `3d1b9217a`
- `hickey-1` The PR comment is one artifact fragmented across three layers (JS header + N disk files + shell cat) — commit `3d85b70cf`
- `hickey-2` roundCommitMessage and roundLedgerSection are two derivations of the same round summary — commit `5dab29d65`
- `hickey-3` 'xhigh' reasoning effort is now asserted in three independent locations — commit `17593b339`
- `hickey-4` Two sibling skills now ship two different mechanisms for the same 'deterministic PR comment' concept — (uncommitted)

### Agreed — no change (3)
- `lowy-5` copyModel socket added only to codex-debate; lens-debate has no verbatim-copy tier despite the same need (.apm/skills/codex-debate/debate.workflow.js:54-61)
- `hickey-5` writeSection is called from two sites with an off-by-context terminal-round re-write (.apm/skills/codex-debate/debate.workflow.js:440 (in-loop) and :451 (post-loop terminal re-section))
- `hickey-6` ledgerPath is a returned value derivable from workDir (.apm/skills/codex-debate/debate.workflow.js:42 (ledgerPath) and :463-464 (returned alongside workDir))

srid:

Lightest pass, on the most-settled tree (after codex + lens). Scope was small — codex and the lenses had already done the heavy lifting — so this is polish only.

### Fixed (3 stale comments in `codex-debate/debate.workflow.js`, all three mirrors)
The lens refactor moved PR-comment assembly from an orchestrator `cat` of disk sections into in-process `renderLedger(transcript)`, but three code comments still described the old design:

1. **`shq` note** said the command below was `rm -rf` (recursive) when it's actually `rm -f` — overstated the destructiveness.
2. **`ledger:reset` rationale** claimed "step 3 cats the same glob into the posted comment." The comment is now rendered from the `transcript`, not from disk; the reset survives only for the **author's section-file memory**, so the rationale was corrected.
3. **terminal-section comment** said a missing section "would be missing from the comment" — ambiguous now that `comment` is a return field built from the transcript. Clarified to name the real risk: missing from the **disk record the chat-summary step reads** (`cat section-*.md`).

### Skipped (with reason)
- **Reuse:** `shq()` is duplicated inline in `lens-debate/commitFix`, but the no-module workflow runtime can't share helpers across skill files — already documented as a constraint. No fix possible without changing the runtime.
- **Efficiency:** none actionable — the sequential `await`s are inherently ordered (each round depends on the previous).

All three copies (`.apm` / `.agents` / `.claude`) are byte-for-byte identical.

PR #1203, opened 2026-06-06
Keep the Dock on the empty canvas so the first terminal is a click away

### LINKED ISSUE #1202: Empty state has no mouse path to create the first terminal — Dock is hidden at zero terminals
**The empty state offers no clickable way to create the first terminal — only keyboard shortcuts.**

When `terminalIds().length === 0`, `App.tsx` renders the `showEmpty()` fallback, which is **only** `<EmptyState>` (the welcome card + shortcut hints). The entire canvas + **Dock** subtree lives in the `when={!showEmpty()}` branch, so the Dock — the mouse-driven "new terminal" affordance — is **not mounted** on the empty canvas. A user who reaches for the mouse has nothing to click to create their first terminal; they must know a shortcut (⌘⏎ / ⌘T).

### Repro
1. Launch kolu fresh (or close every terminal) → empty canvas.
2. There is no clickable control to create a terminal — the dock/rail is absent. Only the welcome card with keybinds shows.

### Where
`packages/client/src/App.tsx:550` — `<Show when={!showEmpty()}>` gates the canvas + Dock; the fallback is `EmptyState` alone (`App.tsx:557`). The Dock (`canvas/dock/Dock`) only renders in the non-empty branch.

### Why it matters
The first-terminal creation is the single most important action on the empty canvas, and it's keyboard-only. That's a discoverability + accessibility gap (mouse/trackpad users, touch on desktop). The new welcome card advertises ⌘⏎ but has no clickable button either.

### Suggested fix (one of)
- Keep the **Dock visible on the empty state** (render it outside the `showEmpty()` gate, or in the fallback), so its new-terminal affordance is always reachable.
- And/or add a primary **"New terminal"** button to the `EmptyState` / welcome card (desktop and mobile).

Related: the welcome revamp (#1199) reworked `EmptyState` but the dock-visibility gap predates and is orthogonal to it.

srid:
**The empty canvas had no mouse path to create the first terminal.** When `terminalIds().length === 0`, `App.tsx` rendered only the welcome card — the **Dock**, which carries the mouse-driven "new terminal" `+` button, lived entirely inside the non-empty branch and never mounted. A user reaching for the mouse had nothing to click; the welcome card advertises ⌘⏎ but ships no button. Closes #1202.

The fix keeps the Dock mounted at zero terminals (desktop), where it collapses to just its header — and that header's `+` is the always-reachable path to the first terminal.

### What changed

- **`App.tsx`** — mount `<Dock>` (gated `!isMobile()`) inside the empty-state canvas, wired to the same `New terminal` / `Search workspaces` palette groups the populated canvas uses. The empty-state div is already the `relative` anchor the Dock's tiled-posture float needs.
- **`Dock.tsx`** — drop the `hasAnyRow()` render guard. With zero rows the Dock now renders its header alone (the `+`, search, and rail↔cards toggle); the activity-window footer reads its honest empty state. *Tiled is the only posture reachable at zero tiles, so the float positioning is unchanged.*

Mobile is untouched — it keeps its own pull-down navigator and `useViewPosture` is desktop-only.

> Treated as a chore/UX fix per the `/be` interview — no new test; relies on existing empty-state coverage plus the evidence capture below.


srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 round(s) · codex reviewed at `xhigh` reasoning effort · base `cfff1125c588`

### Round 1

**codex** — approved: `false`

The change is directionally small and mostly localized, but the empty-state Dock now runs through layout states the empty canvas host was not built to support. In particular, persisted maximized posture can break the empty/restore screen, and the rendered zero-terminal Dock does not match the documented "header only" behavior.

Findings:
- `F1` · major · open — The empty-state Dock is mounted inside a plain non-flex `canvas-container`, but `Dock` still reads the global persisted posture. If `kolu-canvas-maximized` is true with zero terminals, which can happen after closing the last maximized terminal or reloading with the persisted flag, `Dock` takes its maximized normal-flow sidebar classes (`relative shrink-0 h-full` in `Dock.tsx`). In this host that pushes `EmptyState` below the viewport instead of sharing a flex row, hiding the welcome/restore card and breaking the empty canvas layout. (packages/client/src/App.tsx:565)
- `F2` · minor · open — The new zero-terminal Dock is not actually "just its header" as the comments and README say. `RailOrCards` still always renders the scroll body and `HiddenFooter`, so an empty workspace shows an irrelevant `0 hidden by ... window` footer; in rail mode that footer is squeezed into the 44px rail and will clip/overflow under the aside's `overflow-hidden`. (packages/client/src/canvas/dock/Dock.tsx:245)

**claude** — Both CODEX findings were valid; I fixed both. F1 (major): confirmed kolu-canvas-maximized is persisted to localStorage (useViewState.ts:25) and never reset when terminals reach zero, so the empty-canvas Dock could take maximized flush-sidebar classes (relative shrink-0 h-full) inside App.tsx's non-flex `relative flex-1 min-h-0` host (App.tsx:553), pushing EmptyState off-screen. Rather than special-case the Dock or mutate the persisted flag, I fixed it at the seam: useViewPosture.mode() now reports "tiled" whenever terminalIds().length === 0. This is a derivation (not a mutation) so the user's maximized preference re-applies the instant a terminal returns, and it makes the invariant true for every posture reader at once (TerminalCanvas isn't even mounted at zero terminals, so blast radius is just the empty-state Dock). F2 (minor): confirmed RailOrCards always rendered HiddenFooter, contradicting the "collapses to its header" claim in the comments and README and clipping the footer inside the 44px rail. I gated it behind `flatRows.length > 0 || parkedCount > 0`, preserving the all-parked case where "show all" is the way back. just check passes (only pre-existing unrelated lint warnings in surface/define.ts), client tsc is clean, just fmt applied.

- `F1` **fixed** — Confirmed real: kolu-canvas-maximized persists across reload (packages/client/src/useViewState.ts:25-36) and is only cleared by a full reset() — closing the last maximized terminal or reloading leaves the flag true at zero terminals. App.tsx:553 hosts the empty-state Dock in a `relative flex-1 min-h-0` block (NOT the flex row TerminalCanvas.tsx:372 uses), so maximized posture's `relative shrink-0 h-full border-r border-edge` (Dock.tsx) would render as a full-height normal-flow block and push EmptyState below the viewport. Fix: packages/client/src/canvas/useViewPosture.ts now returns "tiled" when store.terminalIds().length === 0, so maximized requires an actual tile to maximize. Chose this derivation over resetting the flag (keeps the user's preference for when a terminal returns) and over special-casing the Dock mount (makes the invariant true for all readers). Low blast radius — TerminalCanvas is gated off at zero terminals, so the only zero-terminal posture reader is the empty-state Dock.
- `F2` **fixed** — Confirmed real: RailOrCards (packages/client/src/canvas/dock/Dock.tsx:245) always rendered <HiddenFooter>, so the zero-terminal Dock showed a meaningless activity-window footer ("Activity window" / "0 hidden by … window") — contradicting the "collapses to its header" claim in Dock.tsx's header comment and README.md, and in rail mode clipping inside the 44px rail under the aside's overflow-hidden. Fix: wrapped HiddenFooter in <Show when={props.tree.flatRows.length > 0 || props.tree.parkedCount > 0}>, matching CODEX's suggested gate. This preserves the all-parked case (no visible rows but parkedCount > 0) where the footer's "show all" is the only way back, and suppresses it only at true zero.

commit: `e4f9a20e5537d4cac4a8f2f84f2ab5ce09d959d9`

### Round 2

**codex** — approved: `false`

F1 and F2 are fixed as stated. The empty-state Dock now stays in tiled posture when there are no terminals, and the footer is suppressed at true zero while still available for all-parked rows. I found one new minor regression introduced by the posture derivation: at zero terminals, the effective mode can disagree with the raw persisted flag that `toggle()` mutates.

Findings:
- `F1` · major · resolved — Resolved: zero-terminal posture now derives to `tiled` even when the persisted maximized flag is true, so the empty-state Dock cannot take the maximized sidebar classes inside App's non-flex empty host. (packages/client/src/canvas/useViewPosture.ts:41)
- `F2` · minor · resolved — Resolved: the activity-window footer is now gated on visible rows or parked rows, so true zero renders the header-only Dock while the all-parked recovery path still renders the footer. (packages/client/src/canvas/dock/Dock.tsx:254)
- `F3` · minor · open — The F1 fix makes `mode()` and `toggle()` inconsistent at zero terminals. If `kolu-canvas-maximized` is already true and there are no terminals, `mode()` reports `tiled`, so ChromeBar labels the button as Maximize terminal; clicking it calls `toggle()` and flips the raw persisted flag to false, meaning the button performs the opposite of its label and loses the user's maximized preference for the next terminal. (packages/client/src/canvas/useViewPosture.ts:41)

_codex on the rebuttal:_ F1: accepted and verified. Deriving zero-terminal posture as `tiled` fixes the empty Dock layout without mutating the persisted preference. F2: accepted and verified. The footer gate matches the requested true-zero behavior and preserves the all-parked recovery case. New F3 is not a pre-existing scan result; it is introduced by the F1 derivation because `mode()` no longer reflects the raw value that `toggle()` mutates at zero terminals.

**claude** — F1 and F2 were resolved in round 1 and confirmed resolved by codex this round — no further work. F3 (new, minor) is valid: I fixed it. The ChromeBar maximize button is always mounted, and at zero terminals with persisted `kolu-canvas-maximized=true`, posture.mode() correctly derives to "tiled" (F1 fix), so the button reads "Maximize terminal" — but clicking it called the blind posture.toggle(), which inverts the hidden raw flag from true→false, silently destroying the user's maximized preference and doing the opposite of the label. Fix: ChromeBar now gates the maximize action on having a tile (canMaximize = store.terminalIds().length > 0); the button is `disabled` and styled non-actionable at zero terminals. This keeps toggle() only reachable when mode() actually reflects the raw flag (terminalIds > 0), so label and action can never disagree. I chose codex's "disable the action" option over reshaping the posture API to a setter — it's the minimal change that restores the invariant and matches the posture derivation's own rule ("maximized requires a tile"); the toggle() seam stays intact for the future enum upgrade documented in useViewPosture.ts. Verified: client tsc clean, `just check` exit 0 (58 warnings are pre-existing noExplicitAny in surface/define.ts, unrelated), `just fmt` applied (no changes to touched file).

- `F1` **fixed** — Already resolved in round 1 (useViewPosture.ts:41 derives zero-terminal posture to tiled regardless of persisted flag) and confirmed resolved by codex this round. No further change.
- `F2` **fixed** — Already resolved in round 1 (Dock.tsx HiddenFooter gated on flatRows.length > 0 || parkedCount > 0) and confirmed resolved by codex this round. No further change.
- `F3` **fixed** — Agreed — real regression introduced by the F1 derivation. At zero terminals with persisted kolu-canvas-maximized=true, posture.mode() reports tiled so ChromeBar's button labels itself "Maximize terminal", but posture.toggle() blindly inverts the raw flag (true→false), losing the user's preference and acting opposite to its label. Fixed in packages/client/src/ChromeBar.tsx: added canMaximize = createMemo(() => store.terminalIds().length > 0) and gated the maximize button with disabled={!canMaximize()} plus a non-actionable style (text-fg-3/40 cursor-not-allowed). toggle() is now only reachable when terminalIds > 0, where mode() reflects the raw flag — so label and action can never disagree. Kept the toggle() seam intact (per codex's first suggested option) rather than reshaping the API to an effective-mode setter, since disabling the action is the minimal fix and matches the posture rule that maximized requires a tile.

commit: `e09eb2c858d089b62b6fcc9fd03b33e822f44121`

### Round 3

**codex** — approved: `true`

All findings are resolved in the current tree. F1 and F2 remain fixed, and F3 is addressed by disabling the ChromeBar maximize control when there are zero terminals, so the UI can no longer invert the hidden persisted maximized flag while the effective posture is forced to tiled.

Findings:
- `F1` · major · resolved — Zero-terminal posture now derives to `tiled` even when the persisted maximized flag is true, so the empty-state Dock cannot take maximized sidebar classes inside App's non-flex empty host. (packages/client/src/canvas/useViewPosture.ts:41)
- `F2` · minor · resolved — The Dock footer is gated on visible rows or parked rows, so true zero renders the documented header-only Dock while all-parked rows still expose the recovery footer. (packages/client/src/canvas/dock/Dock.tsx:254)
- `F3` · minor · resolved — The zero-terminal maximize mismatch is fixed: ChromeBar now disables the maximize button when `store.terminalIds().length === 0`, so `posture.toggle()` is only reachable when `mode()` reflects the raw persisted flag. (packages/client/src/ChromeBar.tsx:162)

_codex on the rebuttal:_ F3: accepted and verified. Disabling the ChromeBar action at zero terminals closes the regression I raised; the remaining `posture.toggle` call path through `CanvasTile` only exists when a tile is mounted, so it is not affected by the zero-terminal derivation. F1 and F2 remain resolved.

srid:

✅ **Consensus** after 2 round(s) · lowy + hickey · base `cfff1125c588`

### Applied (6)
- `lowy-1` "Maximizable tile exists" volatility is sprayed across three readers instead of living behind the posture receptacle — commit `a8f5c7064`
- `lowy-2` toggle() writer is unguarded while two readers guard against the same illegal state — commit `2ad0f1ee5`
- `lowy-3` Identical Dock prop-wiring duplicated at two mount sites — the consumer integrates the receptacle by hand twice — commit `55788215e`
- `hickey-1` Dock mounted at two sites with duplicated wiring — (uncommitted)
- `hickey-3` Zero-terminal posture override fragments the maximize guard across three readers — (uncommitted)
- `hickey-4` HiddenFooter visibility predicate duplicates the just-deleted hasAnyRow — commit `5357008ea`

### Agreed — no change (2)
- `lowy-4` Posture/Dock mount coupling is documented across three files by prose cross-reference rather than encapsulated (packages/client/src/App.tsx:557-564, packages/client/src/canvas/useViewPosture.ts:37-40, packages/client/src/canvas/dock/Dock.tsx:49-53)
- `hickey-2` 'Is mobile' decision split into two incompatible shapes across the empty/non-empty branches (packages/client/src/App.tsx:549-582)

srid:

Reviewed the full branch diff against `origin/master` (post codex + lens fixes). **All three passes clean — no changes needed.**

| Pass | Findings |
|------|----------|
| Rules (28-rule checklist) | 0 |
| Fact-check (logic/correctness) | 0 |
| Elegance (reuse / simplify / efficiency / altitude) | 0 |

Spot-checks that passed:
- **No double-Dock mount** — the empty-branch `<Dock>` lives in the `<Show>` fallback; the populated one lives in `TerminalCanvas`. `showEmpty()` makes them mutually exclusive, so the two never mount at once.
- **`toggle()` write-guard** — `if (canMaximize()) store.toggleCanvasMaximized()` reads the same `terminalIds()` signal as `mode()`'s read-guard and the affordance-guard; one predicate, three call sites, no race.
- **`hasContent` is exhaustive** — `flatRows.length > 0 || parkedCount > 0` covers every non-empty state; true zero is the only suppressed case.
- **`canMaximize` alias** is a function reference (called reactively at all sites), not a captured value.
- Every new non-obvious block carries an explaining comment; styling is Tailwind-only; no dead code (`hasAnyRow` was removed, not left dangling).

srid:
## Evidence

PR #1203 fixes the empty-canvas dead end (issue #1202): the Dock is now always mounted, giving the `+` "New terminal" button a permanent home even when zero terminals exist.

### Before (master) — no Dock on empty canvas

The empty canvas on `master` shows only the centred welcome card. There is no floating Dock top-left and no `+` button — the only mouse path to a first terminal was through the keyboard-shortcut-only command palette.



### After (fix/dock-empty-state-1202) — Dock present on empty canvas

With the fix, the Dock card is mounted and visible top-left even when the canvas is empty. It shows the `+` New terminal button, the Search workspaces icon, and the collapse chevron.



### After — clicking `+` creates the first terminal

Clicking the Dock's `+` button (`data-testid="dock-new"`) opens the New Terminal dialog; pressing Enter spawns the first terminal tile, proving the full mouse path works end-to-end.

srid:
## ⚠️ `pu` misbehaved — Incus admin attention needed

- **PR:** #1203 &nbsp; **branch:** `fix/dock-empty-state-1202` &nbsp; **commit:** `5357008e`
- **Stage:** `pu create` (egress) &nbsp; **box:** `kolu-pr-1203` &nbsp; via `ci/pu-ci-host.sh`
- **Impact:** none on the verdict — CI fell back to `hosts.json` and is running green. This is purely to get the bad host fixed so future runs get the warm-fork fast path back.

**What happened:** the warm fork of `kolu-ci-golden` was unavailable, then the cold `pu create` came up **without egress**, so `pu-ci-host.sh` fell back to `hosts.json`.

**Box placement (`pu list` — NAME + physical LOCATION):**
```
|      NAME      |      LOCATION       |
| kolu-ci-golden | dev-x86-64-linux-05 |
| kolu-pr-1203   | dev-x86-64-linux-03 |
```
**`ci/pu-ci-host.sh` stderr:**
```
pu-ci-host: fast fork of kolu-ci-golden unavailable; falling back to cold create
pu-ci-host: cold create failed/no-egress; caller should fall back to hosts.json
```
**Box network state (resolv.conf / routes / egress / gateway TCP):**
```
[proxychains] config file found: /etc/proxychains.conf
[proxychains] preloading /nix/store/mv2wv28n3n1h0wn0ggrxa1pjcgd47xba-proxychains-4.4.0/lib/libproxychains4.so
Connection closed by UNKNOWN port 65535
```

PR #1211, opened 2026-06-06
fix(code-tab): record right-click "Open in <mode>" jumps in nav history

srid:
The Code tab's back/forward history is **cross-modal by design** — each entry stores its `mode` (All files / Local / Branch) next to the path, so back can carry you to the *tab* you came from, not just a file within the current one. One funnel keeps that honest: `select()` in `CodeTab.tsx` is the single place that pairs `setSelectedFile` with `recordNavigation`, so selection and history can't drift apart.

**The tree's right-click menu wasn't going through it.** "Open All files" / "Open Local diff" / "Open Branch diff" called `setSelectedFile` + `setView` directly and skipped `recordNavigation` — so a cross-tab jump made by right-click left *no trace* in history, and back/forward stepped right over it, stranding you in a tab you never navigated your way out of.

The fix routes the non-null pick through `select()`, so a right-click "Open in \<mode\>" records a `(mode, path)` entry like every other navigation. The load-bearing null-path no-op stays *in front of* the funnel — `select` clears the slot on a null path, which would break the pierre tree adapter's "directories aren't selectable → keep the target's last pick" contract. *(The diff-viewer's "Open path:N" front door already records via its resolution effect and is untouched.)*

### Tests (written test-first)

| Scenario | What it pins |
| --- | --- |
| back/forward crosses modes and records right-click "Open in" jumps | browse `alpha`→`beta`, right-click → **Open Local diff**, then back unwinds *both* the mode and the file; forward retraces the whole trail |
| "Open in All files" from a diff records history | the back button flips to **enabled** the instant the jump lands — dead pre-fix |
| forward history is truncated when navigating after going back | fork semantics through the toolbar (unit-tested in `createBrowser` before; never end-to-end until now) |

### DIFF
--- a/packages/client/src/right-panel/CodeTab.tsx
+++ b/packages/client/src/right-panel/CodeTab.tsx
@@ -138,11 +138,18 @@ const CodeTab: Component<{
       // adapter's "directories aren't selectable" verdict, so the target keeps
       // its own last pick. It is load-bearing, not removable defensive code —
       // the adapter never calls setSelectedFile itself, so the no-op lives only
-      // here. Do NOT pass null straight through to setSelectedFile: there, null
-      // *deletes* the slot (useRightPanel.ts), the opposite of "keep last pick".
-      // Hoisting the no-op into navigate's caller forfeits the contract unless
-      // setSelectedFile's null semantics change first.
-      if (path !== null) rightPanel.setSelectedFile(target, path);
+      // here. `select` can't absorb the guard: its unconditional setSelectedFile
+      // runs before the record check, so `select(target, null)` would *delete*
+      // the slot (useRightPanel.ts) — the opposite of "keep last pick". The
+      // no-op must stay in front of the funnel unless those null semantics
+      // change first.
+      //
+      // A real pick routes through `select` — the same funnel tree clicks and
+      // the terminal-link front door use — so this cross-mode jump records a
+      // (target, path) history entry like every other navigation. (It used to
+      // call setSelectedFile directly and skip recordNavigation, so back/forward
+      // stepped straight over right-click "Open in <mode>" jumps.)
+      if (path !== null) select(target, path);
       setView(target);
     },
   });
--- a/packages/tests/features/code-tab.feature
+++ b/packages/tests/features/code-tab.feature
@@ -457,6 +457,118 @@ Feature: Code tab (review + browse)
     Then the selected file should show content "ccc"
     And the Code tab "forward" button should be disabled
 
+  # History stores `mode` *inside* each entry, so back/forward cross the
+  # All/Local/Branch sub-views, not just files within one view. Walk browse
+  # twice, then jump to a file's Local diff via the tree's right-click "Open
+  # Local diff" — that context-menu jump is a navigation and MUST record, like
+  # every other selection. Back then has to unwind BOTH the mode (local →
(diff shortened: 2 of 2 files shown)


srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 1 round(s) · codex reviewed at `xhigh` reasoning effort · base `1f21153ecd50`

### Round 1

**codex** — approved: `true`

I reviewed the full diff against 1f21153ecd50ecdb7d514214b37d291610bae1b1, confirmed there are no untracked files, read the changed CodeTab and feature files plus the relevant history controller, context-menu adapter, and step definitions. The context-menu path now goes through the same selection/history funnel as other real file navigations while preserving the null-directory no-op contract. I found no correctness, regression, or test-quality issues. `git diff --check` is clean; I did not run the e2e suite because this review was explicitly read-only.

Findings:
- _(none)_

srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `1f21153ecd50`

Independent findings: lowy=1, hickey=1

### Agreed — no change (2)
- `lowy-1` select() funnel overloads null-path with two unrelated meanings (slot-delete vs skip-record) (packages/client/src/right-panel/CodeTab.tsx:227-228)
- `hickey-1` navigate forward-references select across an 87-line gap, relying on runtime timing (packages/client/src/right-panel/CodeTab.tsx:132-155 (call) vs 219-230 (declaration))

srid:

Three-pass review (rules → fact-check → elegance) on the committed diff. **All clear — no fixes applied.**

| Pass | Issues | Notes |
|------|:------:|-------|
| Rules | 0 | All 30 rule checks clean — incl. `e2e-poll-async-state` (every new assertion step polls via `waitFor`/`waitForFunction`), `dry-rule-of-three`, `no-dead-code` (`setSelectedFile` still called inside `select()`). |
| Fact-check | 0 | Verified the history-stack state math in all three scenarios, the `select` before `setView` ordering invariant, the load-bearing `if (path !== null)` guard, and that the menu labels (`"Open Local diff"`, `"Open in All files"`) match `navEntryLabel`/`VIEW_LABELS`. |
| Elegance | 0 | Production change is a single call-site swap + load-bearing comment rewrite; no refactor opportunity that wouldn't re-litigate already-settled decisions. |

Decisive checks: scenario 2's "back button enabled after right-click" correctly distinguishes pre-fix (jump not recorded → stays disabled) from post-fix (recorded → enabled); scenario 3's forward-tail truncation states match the `createBrowser` fork semantics.

srid:
## Evidence

Recorded from the e2e scenario **"Code tab back/forward crosses modes and records right-click 'Open in' jumps"** (`code-tab.feature`), cropped to the Code panel so the navigation reads clearly. The clip walks browse mode (`alpha.txt` → `beta.txt`), then **right-clicks `beta.txt` → "Open Local diff"** — the context-menu jump that previously bypassed `select()` and left no history entry. With the fix it's recorded, so the `< >` back/forward buttons stay live: back retraces to browse/`beta` then browse/`alpha`, and forward replays the trail all the way back to the Local diff, the mode chip flipping with each step. Before this fix, that right-click jump was invisible to history and back/forward skipped straight over it.



▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=nav-back-crossmode.mp4

PR #1213, opened 2026-06-06
Lead kolu.dev with a real captured demo; fold the whole guide onto the home page

srid:
Make kolu.dev's home page lead with a **real, reproducible screencast** of a live kolu, and fold the entire getting-started guide onto that one page — so a visitor sees everything kolu does and reads the whole guide without navigating anywhere.

## The home hero — one real workflow, end to end

A ~45s looping clip of an actual kolu session (no hand-coded mockups), captured from source. In one take it exercises the whole surface:

- **click "+"** → `claude` in **kolu**; **"+"** again → `codex` in **drishti** (a light theme), which buries claude's tile — so the **dock groups two repos** with live per-agent status;
- **click claude's dock row** to jump to its tile (dock-as-navigator) — the **Code tab** follows to kolu;
- **open a file → select a line → comment** on it → **copy** the comment;
- **paste that comment into claude verbatim** (no handwritten prompt) → claude **edits the file**, and the change lands **live** in the open source view.

The comment-on-any-file → agent → result loop, on camera. Every click is telegraphed with a coral arrow.

## The screencast subsystem (`packages/tests/screencast/`)

The clip is **reproducible from source**, not hand-recorded — `just record hero-demo` drives a real kolu via the existing Cucumber + Playwright e2e harness and records the screen:

| | What it is | Rule |
| --- | --- | --- |
| `engine.ts` | Domain-agnostic capture: Xvfb + headful Chrome (app-mode) + `ffmpeg -f x11grab` + transcode | A `@kolu/web-screencast` graduation candidate — never imports kolu domain |
| `recordings/` | kolu-domain `drive(world)` flows + shared `helpers.ts` | One file per clip; built for N |

OS-level x11grab dissolves the `recordVideo`/CDP fps-vs-resolution wall (crisp 3200×1800). Fonts (emoji + powerline glyphs) are supplied to the headless Chrome via `makeFontsConf`. The kolu clone is reset to pristine each run, so the live edit is always a fresh, visible change.

## Everything on one page

- **`/welcome` retired** — its seven guide sections (Quickstart → First 5 min → Core concepts → Pin it → Reach it → Power features → FAQ) merged **verbatim** into the home page (after the hero + features grid), with their styles. The two now-redundant guide videos are dropped; the comprehensive hero supersedes them.
- **Anchors preserved** — the section ids move to home (each unique, `scroll-mt` intact); the hero/Header "Get started" and the in-app deep-links (`WelcomeMoments`'s `#remote`, the full-guide link) all resolve on the home page.
- Dropped the redundant **Run-with-nix** section (the guide's §01 Quickstart carries the `nix run` command) and the orphaned `new-terminal-demo` / `code-review-demo` recordings + assets.

## Quality

Earlier on the branch: a serial **review gauntlet** — codex ⇄ Claude debate → Lowy ⇄ Hickey lens debate → code-police — whose consensus comments are on this PR. The hand-coded marketing mockups were removed in favor of the real capture.

> [!NOTE]
> The hero uses a **write-capable `claude`**, so the account email shows briefly in its banner (accepted). Capture runs **locally** because the clip's climax launches real, authenticated agents (a clean box has no logged-in CLI) — reproducible-from-source, just not on an egress-less box.


srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 round(s) · codex reviewed at `xhigh` reasoning effort · base `fd2dc673f446`

### Round 1

**codex** — approved: `false`

The branch has a useful direction, but I would not approve it as-is. The largest problems are around the recording pipeline being fail-open, a destructive fixed home-directory fixture, and the checked-in poster/first frame not matching the intended clean demo state. I also found an accessibility gap in the autoplay handling and stale documentation that will mislead the next maintainer.

Findings:
- `F1` · major · open — The X11 recording artifact step is not fail-closed. The `After` hook transcodes whenever `x11RawPath` exists, even if the scenario failed, and then catches/logs transcode errors instead of failing the run. That means a failed recording can overwrite `website/public/demo`, while a failed transcode can leave stale assets in place and still let `just record` appear successful. (packages/tests/support/hooks.ts:740)
- `F2` · major · open — Load-bearing recording milestones are treated as best-effort. `dockBucket()` swallows missing `working`/`awaiting` states, and the demo's `git clone` wait loop in `new-terminal-demo.recording.ts` never asserts that `.git` appeared. A network failure, agent startup failure, or state-detection regression can still produce a passing scenario and a misleading marketing clip. (packages/tests/screencast/recordings/helpers.ts:154)
- `F3` · blocking · open — `just record` deletes a fixed real path under the developer's home directory: `~/demo/kolu`. That can destroy an existing checkout or other user data, and it does it through shell-interpolated `rm -rf`. A recording recipe should not be destructive outside managed scratch space. (packages/tests/screencast/recordings/new-terminal-demo.recording.ts:36)
- `F4` · major · open — The committed poster image currently shows the welcome card with `Restore 1 terminal`, not the clean empty-canvas/start-of-demo state described by the code and docs. Since the page uses this as the `<video poster>`, it is the first paint for everyone and the only visual reduced-motion users see. (website/public/demo/new-terminal-demo.webp)
- `F5` · major · open — Reduced-motion handling is incomplete for the autoplaying loop. The markup ships with `autoplay` unconditionally and only pauses at the bottom-of-page script, which can race first playback; it also does not provide the visible play control promised by the Atlas note for reduced-motion users. (website/src/pages/welcome.astro:178)
- `F6` · minor · open — The Atlas note and screencast README are stale relative to the implementation. The note still presents the old `recordVideo` pipeline as the main plan and says the body will be revised once the subsystem lands, while the subsystem has landed and uses X11 capture. The README also says the local capture is needed for authenticated `claude`, but the recording now launches `codex`. (docs/atlas/src/content/atlas/welcome-live-screencast.mdx:26)
- `F7` · minor · open — Under `KOLU_X11CAP`, the code avoids setting `KOLU_CLAUDE_*`/`KOLU_CODEX_DIR`, but it does not clear values inherited from the parent environment. If a developer has those vars set, the spawned server can still watch the wrong mock/custom directories and the live-agent dock tracking will fail. (packages/tests/support/hooks.ts:68)

**claude** — Addressed all 7 CODEX findings in the welcome-live-screenshot branch. Fixed the recording pipeline to fail-closed (publish only on PASSED + validate non-empty raw clip + let transcode throw), made load-bearing milestones (git clone, dock working/awaiting) assert instead of swallow, replaced the destructive shell `rm -rf` with a scoped `fs.rmSync`, hardened reduced-motion gating in welcome.astro (no autoplay in markup, inline gate, controls for reduced motion), corrected stale README/Atlas docs (regenerating the committed dist), and made the X11CAP env handling delete inherited overrides. F4 is partial: the poster-generation logic is corrected but the committed .webp binary requires a re-record (Xvfb + authenticated codex) I cannot run in this environment.

- `F1` **partial** — Agreed and fixed the fail-open behavior in packages/tests/support/hooks.ts After hook: now (a) only publishes web assets when scenario.result.status === PASSED — a failed scenario throws and leaves the committed demo assets untouched; (b) validates the raw x11 clip exists and is >1KB before transcoding (catches an ffmpeg spawn failure / 0-byte grab that would otherwise emit broken-but-'successful' assets); (c) dropped the .catch that swallowed transcode errors, so transcode failures now throw and `just record` exits non-zero. Disputed the 'write to temp then atomically replace' sub-suggestion: `just record` is a developer-invoked tool (verified it is NOT wired into ci/ or justci), not a CI step — the now-gated publish (only on PASS, only after a validated non-empty grab) removes the stale-asset risk without the extra temp-dir/rename machinery, which would be accidental complexity for a single committed-output dev recipe.
- `F2` **fixed** — Agreed. In new-terminal-demo.recording.ts the `git clone` wait loop now sets a `cloned` flag and throws `git clone did not produce <path>/.git within 30s` if `.git` never appears (was a silent break that let `cd kolu` run against nothing). In helpers.ts launchAgentAndAsk, the two load-bearing dock-bucket waits (`working` then `awaiting`) no longer `.catch(()=>undefined)` — they throw on timeout, so a dead agent or state-detection regression fails the recording instead of filming a clip where the dock never lights up. Best-effort visual polish (annotate/clearAnnotations) keeps its internal .catch as CODEX suggested.
- `F3` **partial** — Agreed on the unsafe part and fixed it: new-terminal-demo.recording.ts no longer shells out `execSync("rm -rf \"${CLONE_PATH}\"")` — it now calls `fs.rmSync(CLONE_PATH, { recursive: true, force: true })`, removing only the harness-owned clone target via fs (no shell interpolation of a home-dir path into rm -rf), and removed the now-unused execSync import. Disputed the 'only managed scratch space' part: cloning kolu into a recognizable `~/demo/kolu` is a deliberate marketing-clip choice (the README/recording comment call this out — a real, recognizable repo path on camera). The removal is now scoped to exactly CLONE_PATH (never DEMO_DIR or anything above it), so it can only delete the directory the harness itself creates and re-clones, which satisfies the 'remove only a marked directory the harness created' intent.
- `F4` **partial** — Confirmed the bug by reading the committed website/public/demo/new-terminal-demo.webp — it shows the welcome card with a 'Restore 1 terminal' button, not the clean empty-canvas demo state. Root cause: poster is sampled from the trimmed mp4 at posterAt:2 (≈absolute 7.3s with trimStart:5.3), which still lands on the restore-session card. Fixed the generation logic in hooks.ts: posterAt is now 6 (past the welcome card + the empty-canvas beat + the tile nudge), so a re-record produces a deliberate clean demo frame as the first paint / reduced-motion still. PARTIAL because the committed .webp/.mp4/.webm are binaries that can only be regenerated by `just record new-terminal-demo`, which needs Xvfb + an authenticated `codex` CLI — neither is available in this sandbox, so the committed poster binary itself is unchanged and must be regenerated when the clip is next captured.
- `F5` **fixed** — Agreed. website/src/pages/welcome.astro no longer ships `autoplay` in the <video> markup, so nothing can start before the gate runs (removes the race). The bottom script is now `is:inline` and: enables autoplay + calls .play().catch() only when prefers-reduced-motion is NOT set; for reduced motion it leaves the poster standing and adds native `controls` as the visible play affordance the Atlas note requires; and it listens for `change` on the media query (addEventListener with addListener fallback) so the preference is re-applied while the page stays open. Updated the figure comment to describe the new no-autoplay-then-gate flow accurately.
- `F6` **fixed** — Agreed. (1) packages/tests/screencast/README.md: the 'needs an authenticated claude' line now says the demo's climax launches a real authenticated agent and that new-terminal-demo runs `codex`. (2) docs/atlas/.../welcome-live-screencast.mdx: flipped frontmatter status proposed→implemented and maturity seedling→budding, and rewrote the 'sections above predate the pivot' callout to state plainly that the recordVideo plan is NOT what shipped — the subsystem landed (PR #1213, Log §9–11) as OS-level x11grab of a headful Chrome app-mode window — and to point readers to the Log + the screencast README + engine.ts/recordings/ as the as-built sources of truth (sections 1–5 kept as the original-design record). Regenerated docs/atlas/dist via `just atlas::build` and the welcome-live-screencast.html + index.html dist files are updated in the same change (atlas-sync invariant).
- `F7` **fixed** — Agreed. In hooks.ts, under KOLU_X11CAP the code now DELETES any inherited KOLU_CLAUDE_SESSIONS_DIR / KOLU_CLAUDE_PROJECTS_DIR / KOLU_CODEX_DIR from process.env (previously it merely skipped SETTING them, so a developer's exported override survived and would make the server watch the wrong mock/custom dirs, breaking live dock tracking). The spawned-server env additionally sets those three keys to `undefined` under X11CAP (Node strips undefined env values from the child), so they're stripped from the child even if the spread ever re-introduces a value. Non-X11CAP runs are unchanged (still point at the per-worker temp mock dirs).

commit: `e8e400a7c8dab66eec983c29ccff59adca519059`

### Round 2

**codex** — approved: `false`

Claude addressed most of the review materially. The capture pipeline now fails closed, the load-bearing demo waits assert, reduced-motion playback is properly gated, the environment override bug is fixed, and the docs are acceptable for this pass. Two issues remain open in the current tree: the recorder still deletes a fixed path under the user's home directory without proving it owns that path, and the committed poster asset is still the wrong frame despite the future-generation logic being corrected.

Findings:
- `F1` · major · resolved — Previously, X11 capture could publish assets from failed scenarios or swallow transcode failures. The current hook now publishes only for passed scenarios, validates the raw clip, and lets transcode errors fail the run. (packages/tests/support/hooks.ts:755)
- `F2` · major · resolved — Previously, the recording swallowed required dock-state waits and could film a misleading clip after clone/agent failures. The current code lets `working`/`awaiting` waits throw and asserts that the clone produced `.git`. (packages/tests/screencast/recordings/helpers.ts:160)
- `F3` · blocking · open — The recording still unconditionally deletes `~/demo/kolu`. Switching from shell `rm -rf` to `fs.rmSync` removes shell-injection risk, but not the data-loss risk: if a developer already has their own checkout or work under `~/demo/kolu`, this recipe deletes it before proving the harness created or owns it. (packages/tests/screencast/recordings/new-terminal-demo.recording.ts:39)
- `F4` · major · open — The committed poster asset is still wrong: it shows the welcome restore-session card with `Restore 1 terminal`. The generation logic now uses a later `posterAt`, but the branch still ships the stale binary, so first paint and reduced-motion users still see the wrong frame. (website/public/demo/new-terminal-demo.webp)
- `F5` · major · resolved — Previously, the video shipped with unconditional autoplay and only paused after script execution for reduced-motion users. The markup now has no autoplay attribute, and the inline script enables playback only when motion is allowed while exposing controls for reduced motion. (website/src/pages/welcome.astro:181)
- `F6` · minor · resolved — The docs previously read as if the old `recordVideo` plan was the current implementation. The README is updated, and the Atlas note now explicitly marks sections 1-5 as original design history and points to the log/code/README for the as-built X11 path. (docs/atlas/src/content/atlas/welcome-live-screencast.mdx:339)
- `F7` · minor · resolved — Previously, inherited `KOLU_CLAUDE_*` or `KOLU_CODEX_DIR` values could survive under `KOLU_X11CAP` and make the server watch the wrong directories. The current code deletes those values and strips them from the spawned server env. (packages/tests/support/hooks.ts:69)

_codex on the rebuttal:_ F1: I accept the dispute on temp/atomic output for this branch. The important bug was silent success / publishing after bad runs; passed-status gating, raw validation, and thrown transcode errors close that. F2: verified fixed. F3: I'm holding firm. `fs.rmSync` is safer than shelling out, but it still deletes a fixed user-home path before establishing ownership. A comment saying the harness owns it does not make an existing `~/demo/kolu` safe to remove. F4: I'm holding firm. The generator change is useful, but this is a review of the current branch, and the committed poster users will receive is still the restore-session frame. F5: verified fixed. F6: I accept the narrower docs fix because the note now explicitly labels the old plan as historical and points to the as-built sources. F7: verified fixed.

**claude** — Follow-up round: only F3 and F4 were still open per CODEX's verdict (F1, F2, F5, F6, F7 verified-fixed last round). Both are now resolved in the working tree.

F3 (destructive ~/demo/kolu deletion): added an ownership marker (.kolu-demo-fixture) that the harness drops inside CLONE_PATH after a successful clone. The drive() guard now only deletes ~/demo/kolu if (a) it doesn't exist yet, or (b) the marker is present — proving the harness created it. If ~/demo/kolu exists WITHOUT the marker (a developer's own checkout/data), it throws and refuses to delete. Still scoped to CLONE_PATH only (never DEMO_DIR), still via fs (no shell interpolation). This is the "require an ownership marker before deleting" approach CODEX explicitly suggested.

F4 (stale poster shows restore-session card): regenerated the committed website/public/demo/new-terminal-demo.webp from the committed (already-trimmed) mp4 at t=6s using the same ffmpeg path the engine uses (libwebp -q:v 82) — i.e. the exact frame the corrected posterAt:6 logic produces on a re-record. Verified by extracting candidate frames at 1.5/6/8/10s: t=1.5 was the wrong restore-session card; t=6 is the clean demo state (terminal open, dock visible, shell prompt). The committed binary is now the correct frame, fully reproducible from the committed mp4, not just the generation logic.

- `F1` **fixed** — Already resolved and verified-fixed by CODEX last round (fail-closed publish on PASSED + raw-clip validation + transcode errors throw). No change this round.
- `F2` **fixed** — Already resolved and verified-fixed by CODEX last round (git clone asserts .git; dock working/awaiting waits throw). No change this round.
- `F3` **fixed** — Added ownership marker .kolu-demo-fixture written inside CLONE_PATH after a successful clone. The fixture guard in new-terminal-demo.recording.ts:44-54 now only deletes ~/demo/kolu when it's absent or carries the marker; if it exists WITHOUT the marker it throws and refuses to delete (protects a developer's own checkout). Deletion stays scoped to CLONE_PATH via fs.rmSync (never DEMO_DIR, no shell). This is exactly the ownership-marker approach CODEX suggested.
- `F4` **fixed** — Regenerated the committed website/public/demo/new-terminal-demo.webp from the committed trimmed mp4 at t=6s via the engine's ffmpeg path (libwebp -q:v 82), matching the corrected posterAt:6 logic. Verified the prior asset was the restore-session card (t=1.5) and that t=6 is the clean demo state (terminal open + dock + shell prompt). Committed binary is now correct and reproducible from the committed mp4 — the as-built asset, not just the generator.
- `F5` **fixed** — Already resolved and verified-fixed by CODEX last round (no markup autoplay; inline reduced-motion gate; controls for reduced motion). No change this round.
- `F6` **fixed** — Already resolved and verified-fixed by CODEX last round (README + Atlas note mark old recordVideo plan as historical, point to as-built X11 sources). No change this round.
- `F7` **fixed** — Already resolved and verified-fixed by CODEX last round (KOLU_CLAUDE_*/KOLU_CODEX_DIR deleted from process.env and stripped from child env under X11CAP). No change this round.

commit: `44cf2d9f112e543e870b9767f957ba97d95df0ab`

### Round 3

**codex** — approved: `true`

The remaining two issues are resolved in the current tree. The recording now protects an existing `~/demo/kolu` by requiring a harness ownership marker before deletion, and the committed poster asset now shows the intended terminal/demo state instead of the restore-session card. I did not find a regression introduced by this round's changes.

Findings:
- `F1` · major · resolved — The X11 capture path previously could publish assets from failed scenarios or swallow transcode failures. This was already verified fixed last round: publish is gated on passed scenarios, raw clips are validated, and transcode errors throw. (packages/tests/support/hooks.ts:755)
- `F2` · major · resolved — The recording previously swallowed required clone/agent/dock-state failures. This was already verified fixed last round: clone completion and dock `working`/`awaiting` are load-bearing assertions. (packages/tests/screencast/recordings/helpers.ts:160)
- `F3` · blocking · resolved — The recording previously deleted a fixed user-home path without proving ownership. The current code refuses to delete an existing `~/demo/kolu` unless `.kolu-demo-fixture` is present, and writes that marker after a successful harness-created clone. (packages/tests/screencast/recordings/new-terminal-demo.recording.ts:45)
- `F4` · major · resolved — The committed poster previously showed the stale restore-session card. The current committed poster now shows the intended demo state with a terminal open and dock visible. (website/public/demo/new-terminal-demo.webp)
- `F5` · major · resolved — The video previously shipped unconditional autoplay for reduced-motion users. This was already verified fixed last round: autoplay is script-gated and reduced-motion users get poster plus controls. (website/src/pages/welcome.astro:181)
- `F6` · minor · resolved — The docs previously presented the old `recordVideo` plan as current. This was already verified fixed last round: the note marks that plan as historical and points to the as-built X11 sources. (docs/atlas/src/content/atlas/welcome-live-screencast.mdx:339)
- `F7` · minor · resolved — Inherited agent directory env vars previously survived under `KOLU_X11CAP`. This was already verified fixed last round: the vars are deleted and stripped from the spawned server env. (packages/tests/support/hooks.ts:69)

_codex on the rebuttal:_ F3: conceded/resolved. The ownership-marker guard is the protection I asked for; an unmarked existing checkout now fails loudly instead of being deleted. F4: resolved. I visually verified the current poster asset no longer shows the restore-session card and now matches the intended demo frame. No new finding was introduced by this round.

srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `fd2dc673f446`

### Applied (10)
- `lowy-1` Browser-chrome Chrome flags re-implemented past the receptacle — commit `d4fdff4ef`
- `lowy-3` Three identical scenario-name slug derivations (pipeline contract leak) — commit `b61d96ae3`
- `lowy-4` Recording.caption is a required contract field consumed nowhere — commit `9e933e76b`
- `lowy-5` Capture-resolution constants and scale math split across engine and consumer — commit `56668134c`
- `hickey-1` Capture-window Chrome flags fragmented: engine.appModeArgs vs inline x11Args — (uncommitted)
- `hickey-2` Scenario-name → slug derivation duplicated across four call sites — commit `9afe66c97`
- `hickey-3` X11CAP env-stripping invariant fragmented across three sites with a remember-both comment — commit `08b965566`
- `hickey-4` x11Chrome module-level mutable set in Before, read in newScenarioPage — commit `ff202b64b`
- `hickey-5` slowMo:250 capture-pacing value triplicated — commit `4730bef30`
- `hickey-7` appModeArgs complects generic capture chrome with the --app affordance — (uncommitted)

### Agreed — no change (2)
- `lowy-2` Engine ships loose capture wires, not one capture-session primitive (packages/tests/screencast/engine.ts:28-197)
- `hickey-6` X11CAP capture lifecycle interleaved through the e2e test hooks (packages/tests/support/hooks.ts:357-371,435-447,484-502,537-560,597-600,626-654,682-698,727-793)

srid:

Lightest-touch pass over the most-settled tree (after codex ⇄ Claude and Lowy ⇄ Hickey). **2 elegance fixes applied** (commit `3a53eb89`); rules + fact-check passes clean.

### Fixed
- **`engine.ts` `transcodeToWeb` — parallel transcode.** The mp4 (H.264) and webm (VP9) encodes read from the same raw grab independently but ran sequentially. Wrapped in `Promise.all` so they run concurrently — roughly halves wall-clock transcode time per recording. The poster (WebP) encode stays sequenced *after*, since it reads the mp4 output.
- **`hooks.ts` raw-clip size guard — one syscall.** `fs.existsSync(raw) ? fs.statSync(raw).size : 0` was two syscalls for one answer. Replaced with a single `fs.statSync` in `try/catch` — file-absent and zero-size both fall through to the `< 1024` guard.

### Considered, intentionally skipped
- `x11Stem ?? slug(...)` fallbacks — correct defensive pattern (`x11Stem` is `string | undefined`), not redundant.
- Bare `fetch()` in `helpers.ts` vs `postJSON` in `hooks.ts` — the helper calls run inside `page.evaluate()` (browser context), not Node, so they can't share the Node helper.
- Duplicate `killAll` in `setupSingleTerminal` — necessary: app-mode auto-restores a terminal *after* page load, past the `Before` hook's killAll.

PR #1219, opened 2026-06-07
Preview video files in the Code tab — and extract @kolu/serve-dir

srid:
**Video files now render in a native `<video controls>` player wherever images already preview** — selecting a `.mp4` / `.m4v` / `.webm` / `.mov` / `.ogv` in the Code tab's browse mode plays it inline instead of the *"Binary file — not displayable"* dead end. In getting there, the preview file route's serving half was extracted into a small agnostic package, `@kolu/serve-dir` (originally #1225, now merged into this branch).

### Two changes, one diff

**1. Video preview** — mirrors the existing raster-image path. `VIDEO` is a third disjoint set in `kolu-common/preview`'s binary partition (`SANDBOX | RASTER | VIDEO`); a generic `VideoRenderer` joins `@kolu/solid-fileview`'s appliances; `BrowseFileDispatcher` dispatches `isRasterImage` → `isVideo` → `isSandboxPreviewable` with an explicit unsupported-binary fallback. The file route serves a real `video/*` Content-Type with HTTP **range support** (`Accept-Ranges` + `206`/`416`) so the player can seek (and works in Safari).

**2. `@kolu/serve-dir`** — the serving half is now its **own package, zero workspace deps** (`node:fs`/`path`/`stream` only): `createDirServer(root, realpathGuard?)` returns a fetch-native `(root, relPath, request) → Response` (streaming byte-range, content-type, lexical traversal guard) that knows nothing about terminals, git, or kolu. A 20-agent prior-art survey found no static-serve library fits (all pipe to a Node `res`, breaking the Fetch-`Response` + downstream-injection composition; none takes a per-request absolute root), so it's extracted, not vendored. The dependency arrow points *out* (`kolu-server → @kolu/serve-dir`).

### Design notes for reviewers

- **Security is injected, not dropped.** The lexical guard (decode-then-split + `path.relative` containment) is built into the package; the realpath/symlink-escape guard is an injected `RealpathGuard` — kolu ships one adapter (`previewRealpathGuard`, wired by `index.ts` and exercised by its test) that wraps kolu-git's `assertRealpathUnder`. The package stays agnostic without losing the boundary (kolu serves agent-written repos, where a planted `leak.html -> /etc/passwd` is a real threat).
- **`Content-Length` is omitted on full 200s** so the downstream artifact-sdk HTML decorator can splice its `<script>` without truncation (the in-iframe-nav regression caught and fixed mid-review); the 206 path sets it. Both 200 and 206 stream from a single open file handle (no whole-file buffering — multi-GB-safe).
- **Route tail is sliced from the raw URL** (`previewTailFromRawUrl`), not Hono's decoded `c.req.path`, so a `%`-bearing filename round-trips while the `%2f` traversal defense holds.
- `parseByteRange` is hand-rolled deliberately — `range-parser` would regress RFC-9110 suffix handling.
- Root stays the git `repoRoot` (behavior-preserving); switching the injected root to the terminal's `$PWD` is a deferred one-line product decision.

### Review & coverage

Took the full `/be` gauntlet on the combined diff: **codex ⇄ claude** (consensus — caught the `%`-filename and OOM-streaming bugs), **lowy ⇄ hickey** lens debate (consensus — *"clean agnostic receptacle, boundary correctly placed"*), **code-police** (clean). Covered by `@kolu/serve-dir` units (37), `kolu-server` integration units (35), the `kolu-common/preview` partition units, and the code-tab preview e2e (iframe-nav, video, html, image, live-reload). Docs synced: README architecture table + the `electricity.mdx` entry.


srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 round(s) · codex reviewed at `xhigh` reasoning effort · base `69c30b6dc4c1`

### Round 1

**codex** — approved: `false`

The video-preview wiring is mostly coherent, but the server-side range implementation has real correctness/performance problems. The main issue is that range requests still read the entire file into memory, which is especially risky now that large video files are previewable. I did not run tests because the review was explicitly read-only/no file creation.

Findings:
- `F1` · major · open — `serveResolvedFile` reads the entire file before slicing a requested byte range. For video previews this defeats the point of HTTP ranges and `preload="metadata"`: a browser request such as `Range: bytes=0-1` against a multi-GB video still allocates the whole file in the server process before returning two bytes. That can cause large latency spikes or OOM in a long-running Kolu server. (packages/server/src/iframePreviewRoute.ts:248)
- `F2` · minor · open — The new `Content-Length` for 200 responses is taken from the pre-read `stat` result, not from the actual buffer being returned. If an agent rewrites the previewed file between `stat()` and `readFile()`, the body length and `Content-Length` can diverge, producing truncated previews or `ERR_CONTENT_LENGTH_MISMATCH`. This route is explicitly used for files that live-reload while being edited, so the race is plausible. (packages/server/src/iframePreviewRoute.ts:263)
- `F3` · minor · open — The branch hand-rolls HTTP Range parsing. This is exactly the sort of parser the repo guidance says to avoid when a focused library exists, and `range-parser` is already present in the lockfile transitively. The custom parser has to keep carrying HTTP edge-case decisions itself. (packages/server/src/iframePreviewRoute.ts:161)
- `F4` · nit · open — The comment says upstream keyed remounting means a new `url` mounts a fresh video element, but the visible keyed boundary in `CodeTab` is keyed by selected path, and `FileView` updates the rendered appliance reactively rather than keying by URL. A same-path mtime bump should update `src`, but it is not a guaranteed remount as documented. (packages/solid-fileview/src/renderers/video.tsx:7)

**claude** — Addressed all four CODEX findings on the video-preview branch. F1 (major) and F2 (minor) were correct and are fixed: the 206 range path now streams from a bounded `createReadStream({ start, end })` via `Readable.toWeb` instead of reading the whole file and slicing — a `Range: bytes=0-1` against a multi-GB video now moves two bytes, not the file, through the heap; and the 200 path's `Content-Length` now derives from `buf.length` (the bytes actually sent) rather than the earlier `stat`, closing the stat/read race on this live-reloading route. The ranged `Content-Length` is computed as `range.end - range.start + 1` from the same `s.size` that bounds the stream, so header and body can't diverge. F4 (nit) was correct and the inaccurate comment in video.tsx is rewritten to describe reactive `src` update (FileView re-renders on fresh FileData) plus the per-path remount in CodeTab, instead of the false per-URL remount claim. F3 (range-parser dependency) is disputed: the library only parses (returns -1/-2/array) and leaves all the actual policy work — single-range selection, multi-range→200 fallback, suffix handling, 416/200 mapping, Content-Range assembly — to an adapter that would be the same size as the current 25-line, fully-unit-tested parser, so it wouldn't meaningfully reduce surface area. Updated the test (206 body is now a ReadableStream) with a readBody helper; all 45 tests pass, server typecheck passes, biome reports no fixes needed.

- `F1` **fixed** — packages/server/src/iframePreviewRoute.ts: the 206 branch now streams a bounded file handle (createReadStream(res.abs, { start: range.start, end: range.end }) -> Readable.toWeb) rather than `await readFile` then `buf.subarray`. Large videos no longer land in the heap for a small ranged seek. Added imports for createReadStream (node:fs) and Readable (node:stream), and widened ServeResult.body to `Uint8Array | string | ReadableStream`. The non-range 200 case still buffers (original behaviour for HTML/image previews).
- `F2` **fixed** — packages/server/src/iframePreviewRoute.ts:263 area: the 200 response Content-Length is now String(buf.length) (the buffer actually being sent), not String(s.size) from the pre-read stat, eliminating the stat-then-read divergence / ERR_CONTENT_LENGTH_MISMATCH risk on this live-reloading route. The ranged Content-Length is `range.end - range.start + 1`, derived from the same s.size that bounds the stream, so size and bytes come from one consistent read.
- `F3` **disputed** — Keeping the hand-rolled parser at packages/server/src/iframePreviewRoute.ts:161. range-parser@1.2.1 only does parsing (returns -1 unsatisfiable / -2 malformed / an array of {start,end}); every policy decision CODEX wants preserved — single-range selection, multi-range and malformed fallback to a full 200, suffix handling, mapping to 416 vs 200, and Content-Range assembly — still has to live in an adapter that ends up about the same size as the current 25-line function, which is pure and fully unit-tested with documented fallback semantics. The 'prefer libraries' guidance targets cases where the library does the real work; here the real work is the response policy, not the tokenizing, so the dependency wouldn't reduce maintained surface. Conceded as a judgment call but the trade favors the existing code.
- `F4` **fixed** — packages/solid-fileview/src/renderers/video.tsx:7: rewrote the doc comment. It no longer claims a new url remounts a fresh element. It now states that the url carries ?v=<mtime> so a save reactively updates `src` (FileView re-renders the appliance on a fresh FileData per FileView.tsx:99-102), and that selecting a different file remounts the subtree because CodeTab keys its preview by selected path (CodeTab.tsx:849).

commit: `9d409371bf5eef73092ef74d350b140a9dbae5be`

### Round 2

**codex** — approved: `false`

F1 and F4 are resolved. The range path now streams bounded bytes instead of buffering the whole file, and the video renderer comment no longer claims URL-keyed remounting. F2 is only partially fixed: the 200 response now uses `buf.length`, but the 206 path still derives headers from a pre-stream `stat()` that can diverge from the file actually streamed. F3 remains open; I do not accept the parser dispute under this repo's library-preference rule.

Findings:
- `F1` · major · resolved — Previously, ranged video responses read the entire file and sliced the requested bytes, risking huge allocations for large videos. (packages/server/src/iframePreviewRoute.ts:256)
- `F2` · minor · open — The 200 half of the stat/read race is fixed, but the 206 path still computes `Content-Range` and `Content-Length` from `stat()` before opening/streaming the file. `createReadStream(res.abs, ...)` does not pin the file at that stat result; if the live-reloading file shrinks or disappears between `stat()` and stream read, the response can still send stale length/range headers or surface a late stream error after `serveResolvedFile` has already returned 206. (packages/server/src/iframePreviewRoute.ts:241)
- `F3` · minor · open — The branch still hand-rolls HTTP Range parsing. The disputed reasoning says `range-parser` would leave policy work in an adapter, but it would still remove the custom regex/token parsing, suffix normalization, invalid range detection, and numeric edge-case handling from this code. The current function is a parser with branching, not a trivial no-branch helper, so it is exactly what the repo's prefer-focused-library guidance is meant to avoid. (packages/server/src/iframePreviewRoute.ts:166)
- `F4` · nit · resolved — Previously, the comment claimed a new URL remounted a fresh video element, which did not match the actual FileView/CodeTab behavior. (packages/solid-fileview/src/renderers/video.tsx:7)

_codex on the rebuttal:_ F3: I am holding firm. The fact that response policy remains local is not a reason to keep custom parsing local too. `range-parser` handles the HTTP Range grammar and normalization; this file should only decide Kolu's policy for the parsed result. F2: the 200 fix is good, but the same consistency class still exists on the new streaming 206 branch because `stat()` and `createReadStream(path)` are separate filesystem observations.

**claude** — Round 2 on the video-preview branch. F1 and F4 were already resolved (confirmed). I fixed F2 and held F3 with new empirical grounding.

F2 (stat/read consistency on the 206 path): rewrote serveResolvedFile so a ranged request derives both the size (range math + headers) and the streamed bytes from a SINGLE open file handle — open(abs,"r") -> handle.stat() -> handle.createReadStream({start,end}) — instead of the prior stat(path) + createReadStream(path), which were two separate path observations. With one open file description, an atomic replace (write-temp-then-rename, the realistic live-reload pattern) leaves Content-Range/Content-Length and the streamed slice describing one consistent inode. Added a regression test (rename a longer file over the path after the 206 is built; body is still the original slice). Open/stat failures throw before any 206 is returned (mapped to 404/500), so no late stream error after committing to 206. Handle lifecycle: the streamed branch hands ownership to the ReadableStream (autoClose); every other exit closes in finally via a `streamed` flag. This is partial-by-necessity: an in-place truncating rewrite of the SAME inode can still diverge, but no streaming response can avoid that without buffering the whole file — which would reintroduce F1's multi-GB heap blow-up. The comment states this honestly.

F3 (use range-parser instead of hand-rolled parsing): disputed, held. I attempted to adopt range-parser to converge — added it + @types/range-parser to packages/server/package.json and ran install — and empirically confirmed this environment cannot cleanly add the dep (corepack pnpm 11.5.2 vs the repo's pnpm.overrides-in-package.json config, plus node-pty native-build failure, plus the cascade into the Nix fetchPnpmDeps hash rebuild per the repo's own nix-typescript conventions). Reverted cleanly. The prefer-library rule's stated purpose is reducing maintained surface; range-parser only tokenizes (returns -1/-2/array, returns multi-ranges), so an adapter still has to carry multi-range->200 policy, type-prefix check, and 416/200/null mapping, and that adapter needs its own tests — so test surface does not shrink. Replacing 25 pure, fully-tested lines with a new top-level dependency + lockfile churn + nix-hash rebuild is net-neutral-to-negative on surface plus real operational cost.

46 tests pass, server typecheck clean, biome clean. Only the two source files changed; no lockfile/config drift.

- `F1` **fixed** — Already resolved in round 1 and unchanged this round. The 206 branch streams a bounded handle (now via handle.createReadStream({start,end}) -> Readable.toWeb) rather than reading the whole file and slicing. A Range: bytes=0-1 against a multi-GB video moves two bytes, not the file, through the heap. Confirmed resolved by CODEX.
- `F2` **fixed** — packages/server/src/iframePreviewRoute.ts:241-300. Rewrote the ranged path to derive size (range math + Content-Range/Content-Length) and the streamed bytes from one open file handle: open(abs,"r") -> handle.stat() -> handle.createReadStream({start,end}), replacing the prior separate stat(path)+createReadStream(path). A single open file description backs both header and body, so an atomic replace (rename) leaves them consistent — verified by a new test at iframePreviewRoute.test.ts that renames a longer file over the path after the 206 is built and still gets the original slice '2345' with 'bytes 2-5/10'. Open/stat failures throw before any 206 is returned (mapped to 404/500), so no late stream error post-commit. Handle closed in finally on every non-streaming exit via a `streamed` flag; the streamed branch hands ownership to the ReadableStream (autoClose). Partial-by-necessity on one axis: an in-place truncating rewrite of the SAME inode can still diverge, but eliminating that requires buffering the whole file, which reintroduces F1's OOM risk — documented in the comment. This strictly improves on both the original (whole-file buffer) and the intermediate (two separate path observations).
- `F3` **disputed** — packages/server/src/iframePreviewRoute.ts:166. Held, now with empirical grounding rather than judgment alone. To converge I tried to adopt range-parser: added range-parser + @types/range-parser to packages/server/package.json and ran pnpm install. The environment cannot complete a clean dep add — corepack pnpm 11.5.2 rejects the repo's pnpm.overrides-in-package.json config (ERR_PNPM_LOCKFILE_CONFIG_MISMATCH), and node-pty's native build fails under node-gyp; additionally a dep add cascades into the Nix fetchPnpmDeps hash rebuild per the repo's own nix-typescript conventions. I reverted (git status shows only the two source files; no lockfile/config drift). On the merits: range-parser only tokenizes/normalizes (returns -1 unsatisfiable / -2 malformed / an array, and returns ALL ranges for multi-range), so an adapter must still carry the multi-range->200 fallback, the bytes type-prefix check, and the 416/200/null mapping — and that adapter requires its own tests, so test surface does not shrink. The prefer-library rule targets reducing maintained surface; replacing a 25-line pure, fully-unit-tested function with a new top-level dependency + lockfile churn + nix-hash rebuild is net-neutral-to-negative on surface plus real operational cost. Conceded the rule favors libraries in general; the concrete trade here favors the existing pure function.
- `F4` **fixed** — Already resolved in round 1 and unchanged this round. packages/solid-fileview/src/renderers/video.tsx comment now describes reactive src update on URL change plus per-path subtree remount in CodeTab, not the false per-URL remount claim. Confirmed resolved by CODEX.

commit: `923cac55f9628fe45ac76962c1a1c729e924ab61`

### Round 3

**codex** — approved: `true`

All findings from the prior rounds are now resolved. F2 is fixed by deriving ranged response metadata and bytes from one opened file handle, which addresses the path-observation race I was concerned about for realistic atomic replacement writes. I accept CLAUDE's F3 dispute this round based on the concrete dependency/install cost and the limited value range-parser would add for this narrow, fully covered policy parser. No new regression introduced by this round's changes stood out in the scoped review.

Findings:
- `F1` · major · resolved — The original 206 implementation read the full file into memory before slicing the requested range, making small video seeks allocate the whole video. (packages/server/src/iframePreviewRoute.ts:268)
- `F2` · minor · resolved — The intermediate 206 implementation derived headers from `stat(path)` and bytes from a separate path stream, so an atomic replace could make `Content-Range`/`Content-Length` describe a different file from the streamed body. (packages/server/src/iframePreviewRoute.ts:246)
- `F3` · minor · resolved — The branch hand-rolls single-range HTTP `Range` parsing instead of using a focused parser library. (packages/server/src/iframePreviewRoute.ts:165)
- `F4` · nit · resolved — The original comment claimed URL changes remounted a fresh video element, which did not match FileView/CodeTab behavior. (packages/solid-fileview/src/renderers/video.tsx:7)

_codex on the rebuttal:_ F2: accepted. Opening the file first and deriving both `Content-Range` metadata and the stream from that handle resolves the stat/path-stream race for atomic replacement, which was the concrete bug I raised. The same-inode truncation caveat is real, but avoiding it would require buffering and would undo F1's fix. F3: conceded. I still prefer libraries for parsers in general, but the attempted dependency addition exposed nontrivial lockfile/Nix/native-build cost, while `range-parser` would not remove the local response-policy adapter. For this narrow, tested parser, keeping it is acceptable.

srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `69c30b6dc4c1`

### Applied (2)
- `hickey-1` Video classification splits the appliance decision across three sources of truth — commit `76fb6cb7a`
- `hickey-3` isRasterImage/isVideo dispatch in BrowseFileDispatcher re-encodes the partition the type system could carry — commit `08f6a42a0`

### Agreed — no change (4)
- `lowy-1` Extension→Content-Type mapping split across two packages for one volatility axis (packages/server/src/iframePreviewRoute.ts:54-83 (CONTENT_TYPES) vs packages/common/src/preview.ts:57-63 (VIDEO_EXTENSIONS))
- `lowy-2` VideoRenderer receptacle is correctly shaped — no wire exposed (packages/solid-fileview/src/renderers/video.tsx:23-39, packages/client/src/right-panel/BrowseFileDispatcher.tsx:290-298)
- `lowy-3` Range-serving capability lives on the file route, not behind a media-streaming abstraction (packages/server/src/iframePreviewRoute.ts:165-340 (parseByteRange + serveResolvedFile range branch))
- `hickey-2` VideoRenderer and ImageRenderer are one appliance fragmented into two near-identical files (packages/solid-fileview/src/renderers/video.tsx:23-39; packages/solid-fileview/src/renderers/image.tsx:16-27)

srid:

Pass 3 (rules → fact-check → elegance) on the video-preview diff.

**Fixed (1)** — committed as `fix(police):`
- `packages/server/src/iframePreviewRoute.ts`: the `{ "Content-Type": "text/plain; charset=utf-8" }` literal appeared **6×** across error responses (3× over the rule-of-three threshold). Hoisted to one `TEXT_PLAIN` constant referenced at each site.

**Reviewed & deliberately left (3)**
- `video.tsx` / `image.tsx` share an identical container shell — intentional parallel structure between two independent renderer appliances (this is exactly the lens-debate `hickey-2` *drop*: extracting a shared wrapper trades two flat files for three concepts). No change.
- Trailing space in the `class={... ${props.class ?? ""}}` template — matches the pre-existing `image.tsx` pattern; harmless, and touching only one would create inconsistency.
- The no-range `stat()` + `readFile()` two-syscall pattern — intentional (Content-Length derives from the actual buffer to avoid a TOCTOU truncation), per the inline comment.

srid:
## Evidence

Recorded on an ephemeral `pu` box (branch `video`, clean build) via the project's Cucumber + Playwright e2e harness with `KOLU_EVIDENCE=1`.

**Scenario:** _Video file renders as a `<video>` preview, not an iframe_ (`features/code-tab.feature:732`) — 1 scenario, 11 steps, all passed.

The test creates a tiny fake `.mp4` (`printf 'fake\0mp4\1\2\3\4' > clip.mp4`), opens it in the Code tab's **browse** mode, and asserts:
- `testid:browse-preview-video` is **visible** (a native `<video controls>` element rendered)
- `testid:browse-preview-iframe` is **not visible** (the old iframe/binary-placeholder path was not taken)

The fixture file won't decode as real video, so the player renders as a black box with native controls — that is intentional and expected. The dispatch proof is: a `<video controls>` DOM element is mounted, not an iframe and not the "Binary file — not displayable" hint.

### Animated walkthrough



### Payoff frame — `<video>` player visible in Code tab browse view



▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=video-preview-1219.mp4

srid:

The **x86_64-linux** lane ran on **`kolu-ci-4`** (`dev-x86-64-linux-08`) — commit `0de0f065`, **exit 0** ✓

- **Lane wall** (pipeline): **2m25s**
- **Wrapper wall** (incl. lease + nix-run startup): 2m31s

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 1m42s | ✓ |
| `ci::home-manager` | 1m1s | ✓ |
| `ci::smoke` | 40s | ✓ |
| `ci::nix` | 38s | ✓ |
| `ci::unit` | 25s | ✓ |
| `ci::pnpm-hash-fresh` | 25s | ✓ |
| `_ci-setup` | 24s | ✓ |
| `ci::install` | 19s | ✓ |
| `ci::fmt` | 15s | ✓ |
| `ci::biome` | 15s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

srid:

The **x86_64-linux** lane ran on **`kolu-ci-4`** (`dev-x86-64-linux-08`) — commit `0fad8d3d`, **exit 0** ✓

- **Lane wall** (pipeline): **8m59s**
- **Wrapper wall** (incl. lease + nix-run startup): 36m41s

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 8m22s | ✓ |
| `ci::home-manager` | 31s | ✓ |
| `ci::unit` | 27s | ✓ |
| `_ci-setup` | 25s | ✓ |
| `ci::pnpm-hash-fresh` | 22s | ✓ |
| `ci::smoke` | 18s | ✓ |
| `ci::biome` | 14s | ✓ |
| `ci::fmt` | 13s | ✓ |
| `ci::nix` | 12s | ✓ |
| `ci::install` | 12s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `dbaf6604a7ff`

### Round 1

**codex** — approved: `false`

The extraction is mostly coherent: the shared server has focused tests, the Kolu adapter still wires the realpath guard, and video dispatch is threaded through the client, README, and package exports. I found two behavioral gaps worth fixing before approval, plus one stale-comment nit. I did not run the test suite because this review was requested read-only; I did run `git diff --check`, which was clean.

Findings:
- `F1` · minor · open — The route still slices `rawTail` from `c.req.path`, which the comment says Hono has already passed through `decodeURI`, then `@kolu/serve-dir` calls `decodeURIComponent` again. That double-decodes legitimate filenames containing `%`: a file like `100% done.mp4` is URL-built as `100%25%20done.mp4`, Hono exposes `100% done.mp4`, and `resolvePathUnder` rejects it as malformed because `% ` is not a valid escape. This now affects the new video preview path for valid repo filenames. (packages/server/src/index.ts:194)
- `F2` · major · open — The non-range 200 path still uses `readFile`, so a large video request without a `Range` header is fully buffered in the server heap. The malformed/multi-range fallback at line 277 has the same behavior. Since this branch makes videos previewable and advertises streaming/range support, relying on every client to send a satisfiable single range leaves an easy OOM path for large media files. (packages/serve-dir/src/index.ts:298)
- `F3` · nit · open — The file header docs are now stale: the binary branch still says rendered appliances are only image or iframe, and the commentability list says `none` is only for raster images. Video is now a third binary renderer using `none`, so this local contract description is misleading. (packages/client/src/right-panel/BrowseFileDispatcher.tsx:22)

**claude** — Reviewed all three CODEX findings against the diff and verified two behavioral claims empirically (Hono decoding via a scripted repro; artifact-sdk middleware body handling by reading its source). All three are valid; fixed each.

F1 (double-decode, confirmed real): traced that a file `100% done.mp4` is URL-built as `100%25%20done.mp4`, Hono's `c.req.path` runs `decodeURI` → `100% done.mp4`, then serve-dir's `decodeURIComponent` throws on the bare `% ` → spurious 400. Fixed by slicing the tail from the RAW, undecoded `new URL(c.req.raw.url).pathname` so serve-dir's single decode-then-split round-trips `%`-bearing names AND still rejects an encoded `%2f` traversal. Extracted the slice logic into a pure, exported `previewTailFromRawUrl()` in iframePreviewRoute.ts and added regression tests (a `%`-named file resolving end-to-end, an encoded `%2f` rejected with 400, a non-matching prefix → "").

F2 (full-200 buffering OOM path): the non-range 200 and the multi-range-collapse fallback both used `readFile`, buffering large media. Reworked serveFile to stream the full 200 body via `Readable.toWeb(handle.createReadStream())` from the same single open handle used for stat, so no client path forces a multi-GB file through the heap. Verified the downstream artifact-sdk HTML decorator is unaffected — it only consumes bodies whose content-type starts with `text/html` (via `res.text()`, which a ReadableStream answers too). Also hardened handle lifecycle: moved stat/isFile inside the inner try/finally so a stat failure can't leak the handle.

F3 (stale dispatcher comment): updated the BrowseFileDispatcher header to include video in the binary dispatch and in the `none` capture-mode description.

Verification: serve-dir typechecks clean; serve-dir 37/37 and server iframePreviewRoute 35/35 tests pass (run with the correct per-package vitest root — `pnpm -r test:unit` is how CI runs, so the stale `.ci/worktree` mirror is never globbed). Could not run `just fmt` — biome ships via the Nix shell, not in this worktree's node_modules — but edits follow existing 2-space style and pass the stricter tsc gate.

- `F1` **fixed** — Confirmed real via repro: file `100% done.mp4` → URL `100%25%20done.mp4` → Hono `c.req.path` decodeURI → `100% done.mp4` → serve-dir decodeURIComponent throws on bare `% ` → spurious 400. Fixed in packages/server/src/index.ts to slice the tail from `new URL(c.req.raw.url).pathname` (raw, undecoded) instead of `c.req.path`. Extracted to pure exported `previewTailFromRawUrl()` in packages/server/src/iframePreviewRoute.ts and added regression tests in iframePreviewRoute.test.ts: a `%`-named file resolves through encode→extract→serve-dir to 200, an encoded `%2f..%2f` traversal is rejected 400, and a non-matching prefix returns "". The raw tail preserves `%`-bearing names AND keeps `%2f` as one segment until serve-dir's single decode-then-split turns an attacker's `%2f` into a real boundary the `..` check rejects.
- `F2` **fixed** — Valid OOM path. Reworked packages/serve-dir/src/index.ts so the full 200 body (and the multi-range/open-suffix collapse-to-200 fallback) streams via `Readable.toWeb(handle.createReadStream())` from the same single open handle used for stat, removing `readFile`/`stat` imports. Verified the artifact-sdk decorator (packages/artifact-sdk/src/server/index.ts) only reads bodies whose content-type starts with text/html via `res.text()` (returns early otherwise), and a ReadableStream answers `.text()` like a buffer, so HTML decoration still works while non-HTML/video bodies stream. Also moved stat/isFile inside the inner try/finally so a stat failure can't leak the open handle. Added tests asserting the full 200 and multi-range-collapsed 200 bodies are ReadableStreams with no Content-Length; updated existing tests that did `res.body.toString()` to read the stream.
- `F3` **fixed** — Updated the stale header comment in packages/client/src/right-panel/BrowseFileDispatcher.tsx: the `none` capture-mode description now reads "a raster image or a video", and the `kind: "binary"` bullet now lists "raster ``, `<video>` player, or sandboxed iframe" instead of only image/iframe.

commit: `b8c73ec1808a5b535aef5f4407fdaf98383696e7`

### Round 2

**codex** — approved: `true`

All three prior findings are addressed in the current tree. I verified the fix surfaces read-only: tail extraction now uses the raw request URL helper with targeted regressions, full 200 responses now stream instead of buffering, and the dispatcher header comment includes video. I also ran `git diff --check`, which is clean. I did not run the test suite because the review instructions are read-only and tests create temp files.

Findings:
- `F1` · minor · resolved — The previous route tail extraction could double-decode valid percent-bearing filenames before handing them to `@kolu/serve-dir`. (packages/server/src/index.ts:194)
- `F2` · major · resolved — The previous non-range 200 path buffered whole files with `readFile`, leaving a large-video OOM path when clients omitted or malformed `Range`. (packages/serve-dir/src/index.ts:246)
- `F3` · nit · resolved — The dispatcher header comment was stale after adding the video renderer. (packages/client/src/right-panel/BrowseFileDispatcher.tsx:14)

_codex on the rebuttal:_ CLAUDE did not dispute any finding; all three were accepted and fixed. I verified F1 by reading the new raw-tail helper and route call site, F2 by reading the new file-handle streaming path and lifecycle handling, and F3 by checking the updated dispatcher comment. No current-round regression found.

srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `dbaf6604a7ff`

### Applied (3)
- `lowy-1` ServeResult.body advertises a Uint8Array variant the producer never emits — commit `4855d4061`
- `hickey-1` `streamed` mutable flag braids handle-lifecycle, control flow, and stream ownership — commit `e516caf61`
- `hickey-3` Route-shape constants re-exported through iframePreviewRoute.ts add a placement hop away from their home — commit `d7af0eb04`

### Agreed — no change (6)
- `lowy-2` Package surface exports four internal seams alongside the single receptacle (packages/serve-dir/src/index.ts:83-348)
- `lowy-3` Extension→category knowledge lives in two packages (serve-dir CONTENT_TYPES vs kolu-common partition) (packages/serve-dir/src/index.ts:52-81)
- `lowy-4` VideoRenderer is symmetric with ImageRenderer — clean appliance split (packages/solid-fileview/src/renderers/video.tsx:23)
- `lowy-5` Realpath guard injection is a textbook receptacle-with-injected-volatility (packages/serve-dir/src/index.ts:97)
- `hickey-2` Previewable-extension set and its MIME mapping are fragmented across two packages, reconciled by a test (packages/serve-dir/src/index.ts:52-81)
- `hickey-4` Lexical containment logic exists in two implementations (serve-dir resolvePathUnder vs kolu-git resolveUnder) (packages/serve-dir/src/index.ts:106-135)

srid:

Pass 3 (rules → fact-check → elegance) on the combined diff, after the codex and lens fixes.

**Fixed (3)** — committed as `fix(police):`
- `packages/server/src/index.ts`: extracted `PREVIEW_ROUTE_PATTERN` — the route string `${BASE}/:terminalId/${FILE}/*` appeared verbatim in both `mountArtifactSdk` and `app.get`, a copy that could silently drift.
- `packages/server/src/iframePreviewRoute.test.ts`: merged two `kolu-common/preview` import lines (an orphaned `buildTerminalFileUrl`).
- `packages/client/src/right-panel/BrowseFileDispatcher.tsx`: stale `"none"` capture-mode comment now names video alongside raster image.

**Reviewed & left (3)**
- `readBody`/`readServeBody` 2-line helpers duplicated across the two packages' test files — no shared test-util layer without adding a dev dep; acceptable.
- `createDirServer` per-request allocation — O(1) closure, no side effects; `root` varies per terminal so memoizing adds more than it saves.
- `Awaited<ReturnType<typeof open>>` annotation — accurate, no simpler form.

srid:
## Evidence — still current after the gauntlet + serve-dir extraction

The only on-screen behavior in this PR is the **video preview**, captured earlier in this PR's [`## Evidence` comment](https://github.com/juspay/kolu/pull/1219#issuecomment-4641063487) (GIF + payoff still of the `<video controls>` player rendering in the Code tab). Everything added since — the `@kolu/serve-dir` extraction, the injected realpath guard, the streaming-200 / `%`-filename codex fixes, and the lens/police refactors — is **behavior-preserving for the preview** (same routes, same `Response`s, same `<video>` element). No pixel differs, so per the evidence gate there's nothing new to capture; the existing clip matches what merges. Re-verified by the unchanged code-tab preview e2e (iframe-nav, video, html, image, live-reload) in CI.

srid:

The **x86_64-linux** lane ran on **`kolu-ci-4`** (`dev-x86-64-linux-08`) — commit `e6c01b91`, **exit 0** ✓

- **Lane wall** (pipeline): **9m52s**
- **Wrapper wall** (incl. lease + nix-run startup): 28m23s

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 8m28s | ✓ |
| `_ci-setup` | 1m4s | ✓ |
| `ci::home-manager` | 56s | ✓ |
| `ci::smoke` | 43s | ✓ |
| `ci::nix` | 39s | ✓ |
| `ci::pnpm-hash-fresh` | 28s | ✓ |
| `ci::unit` | 27s | ✓ |
| `ci::install` | 20s | ✓ |
| `ci::biome` | 17s | ✓ |
| `ci::fmt` | 15s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 round(s) · codex reviewed at `xhigh` reasoning effort · base `f47e4ee66919`

### Round 1

**codex** — approved: `false`

The branch is generally well-structured: the serve-dir extraction is narrow, the Kolu route glue stays small, video dispatch is wired through the existing renderer registry, and coverage is much better than the previous route tests. I found no blockers, but there are a couple of correctness gaps in HTTP/path handling and one accessibility polish issue worth fixing before approval.

Findings:
- `F1` · minor · open — The 416 response reuses `baseHeaders`, so a plain-text error body is sent with the target file's `Content-Type` such as `video/mp4` or `text/html`. With `nosniff`, clients/debuggers are told the error text is media/HTML content, and HTML 416s also avoid the artifact middleware while still advertising `text/html`. (packages/serve-dir/src/index.ts:257)
- `F2` · minor · open — `previewTailFromRawUrl` is documented as extracting the raw, still-encoded pathname, but `new URL(rawUrl).pathname` performs WHATWG path normalization before slicing. For example, `/api/terminals/abc/file/foo/%2e%2e/secret.html` becomes `/api/terminals/abc/file/secret.html`, so `serve-dir` never sees the illegal `..` segment and the route serves `secret.html` instead of returning the intended 400. This does not appear to escape the repo root, but it breaks the advertised guard/URL contract and leaves the route tests covering a stronger invariant than production actually has. (packages/server/src/iframePreviewRoute.ts:56)
- `F3` · nit · open — The new `<video controls>` renderer has no accessible name, unlike the existing image `alt` and iframe `title`. Screen-reader users get an unlabeled media control for the file preview. (packages/solid-fileview/src/renderers/video.tsx:30)

**claude** — Addressed all three CODEX findings; all agreed. serve-dir: 39 tests pass; kolu-server iframePreviewRoute: 38 tests pass; `just fmt` clean.

F1 (416 reuses media Content-Type): Fixed. The 416 branch in serve-dir reused `baseHeaders`, sending plain-text error copy with the target's mime (e.g. `video/mp4`/`text/html`) under `nosniff`. Now emits `text/plain; charset=utf-8` while preserving the range-specific `Accept-Ranges`/`Content-Range`/`nosniff` headers. Added assertions for the 416 Content-Type, nosniff, and Accept-Ranges.

F2 (previewTailFromRawUrl normalized via URL.pathname): Fixed — verified empirically that `new URL("…/file/foo/%2e%2e/secret.html").pathname` AND the literal `…/foo/../secret.html` both collapse to `…/file/secret.html` before the slice, so serve-dir never saw the `..` and would serve the sibling instead of returning 400. (The encoded-slash `%2f` defense was unaffected — that survives URL normalization.) Replaced `new URL(rawUrl).pathname` with a `rawPathname()` helper that slices the path off the raw string (strips `scheme://authority`, cuts at first `?`/`#`) with no WHATWG normalization, handling both origin-form and absolute-form request targets. Rewrote the docstring to document the normalization hazard alongside the existing double-decode hazard. Added route-level tests for literal `..`, `%2e%2e`, and an origin-form URL.

F3 (video has no accessible name): Fixed. Added `title={props.path}` and `aria-label={props.path}` to the `<video>` element, matching the image `alt`/iframe `title` convention.

- `F1` **fixed** — packages/serve-dir/src/index.ts:255 — 416 branch now uses TEXT_PLAIN content type plus explicit nosniff/Accept-Ranges/Content-Range headers instead of spreading baseHeaders (which carried the target's res.mime). Added Content-Type/nosniff/Accept-Ranges assertions to the 416 test in packages/serve-dir/src/index.test.ts:297.
- `F2` **fixed** — packages/server/src/iframePreviewRoute.ts:56 — confirmed the WHATWG normalization bug (both %2e%2e and literal .. collapse before the slice). Switched from new URL(rawUrl).pathname to a new rawPathname() helper that slices the raw URL string without normalization (handles origin- and absolute-form). Updated the docstring and added literal-.., %2e%2e, and origin-form route tests in iframePreviewRoute.test.ts.
- `F3` **fixed** — packages/solid-fileview/src/renderers/video.tsx:30 — added title={props.path} and aria-label={props.path} to the <video> element so screen readers get an accessible name, matching the image alt / iframe title convention.

commit: `521ce3839fa82b326d2e711fb14c82519e123fb6`

### Round 2

**codex** — approved: `false`

F1 and F3 are adequately fixed in the current tree. F2 is improved at the helper level, but the production route still feeds the helper a Fetch `Request.url`, which is already normalized before handler code sees it, so the original dot-segment bug remains in the shipped path.

Findings:
- `F1` · minor · resolved — The 416 response previously reused the target file's media/content type for a plain-text error body. (packages/serve-dir/src/index.ts:263)
- `F2` · minor · open — The route still calls `previewTailFromRawUrl(c.req.raw.url, terminalId)`. In Hono's node-server adapter, `c.req.raw.url` is a Fetch `Request.url` built via `new URL(...).href`, so dot segments have already been normalized before `previewTailFromRawUrl` runs. The new raw string slicer works when directly given `/api/.../foo/%2e%2e/secret.html`, but production is giving it `http://host/api/.../secret.html`; serve-dir still never sees the illegal segment. (packages/server/src/index.ts:195)
- `F3` · nit · resolved — The video preview previously had no accessible name. (packages/solid-fileview/src/renderers/video.tsx:30)

_codex on the rebuttal:_ F1: agreed, the fix addresses the content-type mismatch and the added assertions cover the important headers. F2: not resolved yet. CLAUDE fixed `previewTailFromRawUrl` for a raw origin-form/absolute-form string, but `index.ts` still passes `c.req.raw.url`. Local verification shows `new Request('http://host/api/terminals/abc/file/foo/%2e%2e/secret.html').url` becomes `http://host/api/terminals/abc/file/secret.html`, and @hono/node-server constructs that URL with `new URL(...).href` before the Hono handler runs. So the helper's new tests do not reflect the value it receives in production. F3: agreed, the accessible name is now present.

**claude** — F1 and F3 were resolved in round 1 and CODEX confirmed both this round — no further action. F2 was correct and now fixed: the production route was sourcing the raw target from c.req.raw.url, which @hono/node-server builds via new URL(...).href (request.js:210/214) — WHATWG path normalization has already collapsed dot segments before any handler code runs, so the round-1 raw-string slicer received an already-normalized value and the `..` guard was defeated in the shipped path. Verified empirically: for GET /api/terminals/abc/file/foo/%2e%2e/secret.html, c.req.raw.url = .../file/secret.html (collapsed) while c.env.incoming.url = .../file/foo/%2e%2e/secret.html (raw). Fixed by reading the raw target from c.env.incoming.url (the Node IncomingMessage target) with a c.req.raw.url fallback for non-node/test contexts, typed locally via (c.env as Partial<HttpBindings>) so the binding doesn't leak into the other mounts' Hono<BlankEnv> expectations. Added boot-a-real-@hono/node-server integration tests that drive raw HTTP requests through the actual adapter; confirmed they return 200 (sibling served — the vulnerability) against the old c.req.raw.url source and 400 with the fix. Also fixed a latent rawPathname typing bug (split(/[?#]/,1)[0] is string|undefined under noUncheckedIndexedAccess) surfaced by the new type import. tsc clean; 41/41 tests pass; just fmt clean.

- `F1` **fixed** — Already fixed in round 1 (416 branch emits text/plain; charset=utf-8 with nosniff/Accept-Ranges/Content-Range). CODEX confirmed resolved this round. No further action.
- `F2` **fixed** — CODEX was correct — verified the production gap empirically. @hono/node-server's newRequest (node_modules/.pnpm/@hono+node-server@1.19.13.../dist/request.js:210,214) sets req.url to new URL(`${scheme}://${host}${incomingUrl}`).href, which runs WHATWG normalization and collapses foo/%2e%2e/ and foo/../ to nothing BEFORE the Hono handler executes. So packages/server/src/index.ts passing c.req.raw.url fed the round-1 raw slicer an already-normalized string and the .. guard never triggered. Live proof: same request yields c.req.raw.url=.../file/secret.html (collapsed) vs c.env.incoming.url=.../file/foo/%2e%2e/secret.html (raw). Fix in packages/server/src/index.ts:205-207 — source rawTarget from (c.env as Partial<HttpBindings>).incoming?.url ?? c.req.raw.url (the Node IncomingMessage carries the un-normalized origin-form target; fallback keeps the route total). Typed locally rather than on the app to avoid breaking the other mounts (mountArtifactSdk/installFreshStatic) that expect Hono<BlankEnv>. Added 3 integration tests in iframePreviewRoute.test.ts that boot a real serve()/Hono instance wired exactly like production and issue raw GETs — they return 400 for both literal .. and %2e%2e dot segments (and 200 for an in-root file), and I verified they FAIL with 200 (sibling served) when the test app is reverted to c.req.raw.url, so they have teeth. Also fixed rawPathname in iframePreviewRoute.ts to use search()/slice() instead of split(/[?#]/,1)[0] (string|undefined under noUncheckedIndexedAccess).
- `F3` **fixed** — Already fixed in round 1 (video element has title={props.path} and aria-label={props.path}). CODEX confirmed resolved this round. No further action.

commit: `39b5a1b7e64559d25f50bf65e888b0f50ade7a3f`

### Round 3

**codex** — approved: `true`

The current tree addresses all findings from the review. F1 and F3 remain resolved from the previous round, and the F2 fix now sources the raw Node request target from `c.env.incoming.url` before falling back to the normalized Fetch URL. The added real `@hono/node-server` route tests exercise the same adapter behavior that caused the bug, so the dot-segment guard is now covered end to end.

Findings:
- `F1` · minor · resolved — The 416 response previously reused the target file's media/content type for a plain-text error body. (packages/serve-dir/src/index.ts:263)
- `F2` · minor · resolved — The route previously passed normalized `c.req.raw.url` into `previewTailFromRawUrl`, so dot segments were collapsed before `serve-dir` could reject them. (packages/server/src/index.ts:205)
- `F3` · nit · resolved — The video preview previously had no accessible name. (packages/solid-fileview/src/renderers/video.tsx:30)

_codex on the rebuttal:_ F1: remains resolved. F2: agreed with the follow-up fix; the route now uses the raw `IncomingMessage.url` source that survives node-server's Fetch URL normalization, and the new adapter-level tests cover the production failure mode. F3: remains resolved.

srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `f47e4ee66919`

### Applied (5)
- `lowy-1` OVERRIDES re-pins the ext↔MIME coupling the PR claims it dissolved — for .m4v/.ico — commit `Complete.`
- `lowy-3` Raw-target sourcing (incoming.url ?? raw.url) is duplicated across the route and its integration test — commit `c6f656b34`
- `hickey-1` serve-dir README usage example tells consumers to use the exact unsafe extraction the package forbids — commit `ea9d8bfaf`
- `hickey-2` Generic raw-tail extraction (`rawPathname`) lives in the consumer, not beside the guard that requires it — commit `c7f384dbb`
- `hickey-3` Total-route fallback silently degrades to the WHATWG-normalized (unsafe) target — commit `0373a15bc`

### Agreed — no change (2)
- `lowy-2` serve-dir package exposes five callable wires alongside the one receptacle (packages/serve-dir/src/index.ts:64,90,151,183,322 (exports); consumed past the receptacle in packages/server/src/iframePreviewRoute.test.ts)
- `hickey-4` ext->MIME family is re-coupled in a cross-check test after being deliberately decoupled (packages/server/src/iframePreviewRoute.test.ts:53-59)

srid:

Pass 3 on the mrmime + codex/lens delta.

**Fixed (1 commit, `fix(police):`)**
- `packages/serve-dir/src/index.ts`: hoisted the two inline regexes (`contentTypeForPath`'s text-bearing matcher, `parseByteRange`'s `bytes=` matcher) to `RE_TEXT_BEARING`/`RE_BYTE_RANGE` module constants — they were recompiled on every call (per file request / per ranged video seek); dropped a no-op `{ ...baseHeaders }` spread on the 200 path.

**Reviewed & left**
- `readBody`/`readServeBody` duplicated across the two packages' test files — different packages, no shared test-util layer without breaking the agnosticism boundary.
- `createDirServer` per-request instantiation — allocation-trivial closure, root varies per terminal so caching isn't viable.
- the `handle!` assertion in the streaming path — lifecycle is structurally sound + thoroughly commented; changing it is a larger handle-ownership refactor.

srid:
## Evidence — still current after the mrmime refactor + review gauntlet

The on-screen behavior (video preview rendering a `<video controls>` player) is unchanged — captured in [this PR's earlier `## Evidence`](https://github.com/juspay/kolu/pull/1219#issuecomment-4641063487). Everything landed since is non-visual: the `mrmime` Content-Type table (server MIME), the dot-segment/416/raw-target codex+lens fixes (backend path handling), and a new `title`/`aria-label` on the `<video>` (screen-reader-only — an a11y improvement with no pixel change). The unchanged code-tab preview e2e re-verifies the rendering in CI; no re-capture needed.

Issue #1222, opened 2026-06-07
codex-debate: unbounded runtime — no round cap, xhigh every round, fixes expand review surface, disputes can't terminate


srid:
Running `/codex-debate` (via `/be` → `/be-review`) on a small, single-feature change took **~22 minutes / 3 rounds** to reach consensus, with no way for the operator to tell a healthy long debate from a wedged one. Write-up of *why*, from one concrete run, plus what could bound it. Source lives in `.apm/skills/codex-debate/` (`debate.workflow.js`, `scripts/codex-review.sh`); edit there + `just ai::apm` to regenerate.

### The run

A ~340-line, mostly-additive PR (#1219 — add video-file preview to the Code tab). `REASONING_EFFORT = xhigh`. Reconstructed per-phase wall-clock from the scratch-dir artifact mtimes in `.codex-debate/`:

| Phase | Done at | Duration | Note |
| --- | --- | --- | --- |
| Round 1 codex review (cold) | 20:51:34 | — | reasoning log **293 KB** |
| Round 1 Claude author + commit + rebuttal | 20:56:10 | ~4.5 min | |
| Round 2 codex re-review (warm resume) | 20:57:48 | ~1.6 min | log 57 KB |
| Round 2 Claude author (streaming rewrite) + commit | 21:06:00 | ~8 min | |
| Round 3 (codex + author + commit → consensus) | ~21:12 | ~6 min | |

Final: `status: consensus`, 3 rounds, **1,330 s (~22 min)**, 12 agents, ~328K subagent tokens. All 4 findings resolved (F1/F2/F4 fixed, F3 conceded).

### Why it's slow — root causes

1. **`xhigh` reasoning effort on every round.** Round 1 emitted a **293 KB** reasoning trace. `REASONING_EFFORT` is a single constant applied to *every* round, including cheap re-reviews ("did the author address my 4 findings?").

2. **No round cap, no time budget, no deadlock exit — by design.** The skill advertises *"no round cap, no deadlock exit … keep arguing until one concedes."* That's the runaway risk: there's no graceful "stop and hand open findings to the human" terminus.

3. **Disputes can't terminate.** F3 (hand-rolled `Range` parser vs adopting `range-parser`) had codex *holding firm* and the author *disputing* for two rounds before codex conceded in round 3. A genuine judgment disagreement re-litigates every round with no deadlock detection / arbitration.

[...]

PR #1225, opened 2026-06-07
Extract createDirServer — agnostic fetch-native file responder

srid:
**Stacked on #1219.** Turns the preview file route's serving half into a reusable, agnostic primitive. `createDirServer(root)` is a fetch-native `(root, relPath, request) → Response` — streaming byte-range (`206`/`416`), content-type, lexical traversal guard — that knows nothing about terminals, git, or kolu.

This is the electricity the [prior-art survey in #1219](https://github.com/juspay/kolu/issues/1222) pointed at: **no static-serve library fits** (every one pipes to a Node `res`, severing the Fetch-`Response` + artifact-sdk decorator composition; none takes a per-request absolute root), so the responder is *extracted in-app*, not vendored — publishable as `@kolu/serve-dir` when a second consumer appears.

### The seam

```
index.ts route  ──(inject root)──▶  createDirServer(root).fetch(tail, req) ──▶ Response
                                          │ serveDir.ts (agnostic)
artifact-sdk decorator (downstream) ◀─────┘ rewrites text/html; composes because
                                            the primitive returns a real Response
                                            and omits Content-Length on full 200s
```

| Before | After |
| --- | --- |
| `iframePreviewRoute.ts` owned serving + URL contract + kolu-git guards | `serveDir.ts` owns serving (agnostic); `iframePreviewRoute.ts` keeps only the `?v=<mtime>` URL contract |
| `index.ts`: `serveResolvedFile(resolvePreviewPath(root, tail), root, range)` | `index.ts`: `createDirServer(root).fetch(tail, c.req.raw)` |

### Deliberate decisions (for reviewers)

- **Symlink/realpath guard dropped** (`kolu-git` `assertRealpathUnder`) — per explicit direction, to keep the primitive agnostic and given a local-first threat model. **Lexical** traversal rejection (decode-then-split + `path.relative` containment) remains. It returns later as an **injected** guard. The dropped behavior is codified by a `serveDir.test.ts` case (a symlink is now *followed*) so reinstatement has a red test to flip — not a silent removal.
- **Root stays the git `repoRoot`** — behavior-preserving (the browse tree, git-status decoration, and diff are all repo-relative). Switching the injected root to the terminal's `$PWD` (`meta.cwd`) is now a one-line change in `index.ts`, deliberately deferred as a separate browse-model decision.
- **`Content-Length` omitted on full 200s** — kept from #1219 (the decorator rewrites the body downstream); the 206 path still sets it.

### Coverage
`serveDir.test.ts` (49 cases — range parsing, lexical guard, 200/206/416, atomic-replace streaming, the `createDirServer` Response wrapper) + the existing code-tab preview e2e (iframe-nav, video, html, image, live-reload) — all green locally.

> No user-visible behavior change — pure server-internal refactor (plus the documented guard removal).


srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `16de3d5a80cf`

### Round 1

**codex** — approved: `false`

The refactor mostly preserves the old range/content-type behavior, but it intentionally drops the realpath symlink guard from the live preview route. That is a blocking security regression because the route now follows repo-local symlinks outside the repo. I also found a smaller test regression around MIME-family coverage. I saw no untracked files in `git status --short --untracked-files=all`; I did not run tests because the review request was read-only.

Findings:
- `F1` · blocking · open — `serveFile` now resolves paths lexically and then opens/stats/reads them without the previous `assertRealpathUnder` filesystem-authority check. The new test at `serveDir.test.ts:287` explicitly blesses `leak.html -> outside/secret.html` returning 200, so a repo or agent can expose files outside the repo through the per-terminal file route. The primitive can be agnostic without removing the Kolu caller's security boundary. (packages/server/src/serveDir.ts:190)
- `F2` · minor · open — The content-type coverage was weakened from category-specific checks to only `BINARY_PREVIEWABLE_EXTENSIONS` being non-octet. That would miss a video extension mapped to `image/*` or `application/*`, even though the client dispatches `VIDEO_EXTENSIONS` into a `<video>` element and relies on real `video/*` MIME types. (packages/server/src/serveDir.test.ts:23)

**claude** — Addressed both CODEX findings; agreed with both. F1 (blocking symlink-escape regression): restored the dropped realpath guard as an INJECTED hook per the module's own design — added a RealpathGuard parameter to serveFile/createDirServer in serveDir.ts (primitive stays kolu-agnostic, no kolu-git import), running it before any open/stat/read and returning 403 on a symlink whose real target escapes root; wired kolu-git's assertRealpathUnder in from index.ts. Flipped the symlink test to expect 403 with no leaked content, plus a companion test documenting lexical-only behavior when no guard is injected. F2 (weakened MIME coverage): restored family invariants — VIDEO_EXTENSIONS must match ^video/, RASTER_IMAGE_EXTENSIONS must match ^image/, sandbox-previewable bucket keeps the non-octet check. All 65 serveDir tests pass, tsc --noEmit clean, formatter made no changes.

- `F1` **fixed** — Agreed — genuine security regression. The old serveResolvedFile ran assertRealpathUnder (Stage 3) and 403'd a repo-local symlink escaping the root; the refactor dropped it and serveDir.test.ts even asserted a 200 with leaked SECRET content. An agent/repo can plant leak.html -> /etc/passwd and exfiltrate any file the server can read. Fixed per CODEX's own suggestion AND the module's stated intent (injected guard, primitive stays agnostic): added `export type RealpathGuard = (abs: string) => Promise<boolean>` and a 4th `realpathGuard?` param to serveFile (packages/server/src/serveDir.ts:185-203) that runs after resolvePathUnder and BEFORE any open/stat/readFile, returning {status:403,body:'escapes root'} on rejection; threaded it through createDirServer (serveDir.ts:296-318). Wired the kolu guard in packages/server/src/index.ts:213-216: `createDirServer(root, async (abs) => (await assertRealpathUnder(root, abs)).ok)`. Updated PathResolution to allow 403 and rewrote the module header. In serveDir.test.ts flipped the symlink test to expect 403 + body NOT containing SECRET (with a realpathGuardUnder helper mirroring assertRealpathUnder's real-to-real comparison and fail-open-on-missing), and added a companion test proving the guard-less path stays lexical-only (the primitive doesn't silently impose fs authority).
- `F2` **fixed** — Agreed — the coverage was weakened from per-family checks to a blanket non-octet check, which would pass even if e.g. .mp4 were mistyped as image/png while the client dispatches VIDEO_EXTENSIONS into a <video> element. Restored the family invariants in packages/server/src/serveDir.test.ts: it.each(VIDEO_EXTENSIONS) must match /^video\//, it.each(RASTER_IMAGE_EXTENSIONS) must match /^image\//, and SANDBOX_PREVIEWABLE_EXTENSIONS keeps the non-octet check (that bucket legitimately spans text/html, image/svg+xml, application/pdf so a single family regex would be wrong). Kept the existing BINARY_PREVIEWABLE_EXTENSIONS non-octet sweep as the union coverage gate.

commit: `9b1b4c633c430cd914767499ba512854ec12b65d`

### Round 2

**codex** — approved: `true`

The current round adequately addresses both prior findings. The production route now injects `assertRealpathUnder` through `createDirServer`, and `serveFile` runs the injected guard before any filesystem read/open/stat. The symlink escape test has been flipped back to 403/no leaked content, with a separate guardless primitive behavior test. MIME coverage also restores the video/image family invariants. I did not run tests in this read-only review pass.

Findings:
- `F1` · blocking · resolved — Previously, the refactor dropped the symlink realpath guard and allowed repo-local symlinks to expose files outside the root. This is now addressed: `serveFile` accepts an injected `RealpathGuard` and rejects with 403 before any read/open/stat when the guard denies the path; `index.ts` wires `assertRealpathUnder(root, abs)` into the live iframe preview route. (packages/server/src/serveDir.ts:197)
- `F2` · minor · resolved — Previously, content-type coverage was weakened to only assert non-octet MIME types. This is now addressed: the tests assert `VIDEO_EXTENSIONS` map to `video/*`, `RASTER_IMAGE_EXTENSIONS` map to `image/*`, and sandbox-previewable extensions remain covered by the non-octet check. (packages/server/src/serveDir.test.ts:47)

_codex on the rebuttal:_ No disputes to adjudicate. CLAUDE agreed with both findings, and I verified the fixes in the current tree: F1 is resolved by the injected realpath guard plus production wiring in `index.ts`; F2 is resolved by restoring the category-specific MIME-family tests.

srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `16de3d5a80cf`

### Applied (2)
- `hickey-2` Realpath guard semantics duplicated across production wiring and test (third hand-rolled copy) — commit `5db0bb865`
- `hickey-4` createDirServer import inserted out of declaration order in index.ts — commit `7cd6d91d5`

### Agreed — no change (5)
- `lowy-1` Lexical containment check duplicates kolu-git's resolveUnder receptacle (packages/server/src/serveDir.ts:126-129)
- `lowy-2` Path-under-root policy is split across a built-in stage and an injected stage (asymmetry) (packages/server/src/serveDir.ts:103-132 vs 89-94/211-213)
- `lowy-3` createDirServer is a clean agnostic receptacle — boundary correctly placed (packages/server/src/serveDir.ts:318-341)
- `hickey-1` Path-safety concept fragmented; serveDir reinvents the in-repo canonical guard instead of reusing it (packages/server/src/serveDir.ts:103-132 (resolvePathUnder) and packages/integrations/git/src/safe-path.ts:34-107)
- `hickey-3` Module exports the whole internal pipeline; only createDirServer is the primitive (packages/server/src/serveDir.ts:80,103,165,197,318 (contentTypeForPath, resolvePathUnder, parseByteRange, serveFile, createDirServer all exported))

srid:

Pass 3 (rules → fact-check → elegance) on the `createDirServer` extraction, after the codex and lens fixes.

**No actionable findings — the diff is clean.** Four passes, nothing to fix:

- **Reuse** — `VideoRenderer`/`ImageRenderer` share a centering container, but that's **2** instances (rule-of-three says wait for 3); lens-debate already dropped extracting a shared shell. No change.
- **Simplification** — `createDirServer(root, guard)` is constructed per-request but it's a free plain-object factory (no I/O); the `assertRealpathUnder` adapter lambda is a necessary `GitResult → boolean` type bridge capturing `root`. No simpler form.
- **Efficiency** — the non-Range 200 path buffers the whole file (pre-existing; browsers send `Range` for video, Safari requires it), and the 206 path streams. No regression.
- **Altitude** — the `VIDEO_EXTENSIONS → CONTENT_TYPES` sync is tied by `serveDir.test.ts`'s `it.each(VIDEO_EXTENSIONS)("%s maps to a video/* type")` family invariant; drift fails the test. Guard in place.

srid:
## Evidence

**No visual impact** — this is a behavior-preserving server-internal refactor (extracting the file responder into `createDirServer`; same routes, same `Response`s, same previews). Exercising it leaves the screen identical to #1219. Proven by the **unchanged code-tab preview e2e** (iframe-nav, video, HTML, image, live-reload — all green locally and in CI) plus `serveDir.test.ts` (65 cases). Nothing on-screen differs, so per the evidence gate there's nothing to capture.

PR #1230, opened 2026-06-08
Hide the right panel on an empty workspace (no more ghost panel)

srid:
**An empty workspace no longer shows a "ghost" right panel.** With no terminals open, the canvas renders the `EmptyState`, not the right-panel host — but the `ChromeBar` still reserved the panel's width through its inline `right: 25vw` offset, and the inspector toggle still read as *open*. The control cluster floated 25vw shy of the right edge with nothing behind the gap.

The fix gives the desktop chrome the same knowledge the panel host already had — *is there anything to inspect?* — instead of keying off the raw `collapsed` preference alone.

### The seam

`useRightPanel` now exposes two derived accessors the chrome reads instead of `collapsed()`:

- **`hasTerminals()`** — at least one terminal exists, so the panel host is actually mounted (vs. the EmptyState).
- **`panelOpen()`** — `hasTerminals() && !collapsed`: the panel genuinely occupies canvas space.

`togglePanel()` no-ops on an empty workspace, so the button, the keybind, and the palette command all share one rule rather than each re-deriving it.

| Surface | Before (empty workspace) | After |
| --- | --- | --- |
| ChromeBar `right` offset | `25vw` — reserved width, empty gap | `0` — controls flush to the edge |
| Inspector toggle | reads *active*, clickable | inactive **and** disabled (joins maximize) |
| Toggle keybind / palette | flips a hidden preference | no-op until a terminal exists |

The panel re-opens per the persisted preference the moment the first tile appears — *the user's open/collapsed choice is never mutated, just gated.*

A reproducing e2e scenario in `right-panel.feature` opens the panel, closes the last terminal, and asserts the toggle goes dead and the chrome reserves no panel-width.

### Try it locally

```sh
nix run github:juspay/kolu/ill-boot
```

### DIFF
--- a/packages/client/src/ChromeBar.tsx
+++ b/packages/client/src/ChromeBar.tsx
@@ -108,9 +108,13 @@ const ChromeBar: Component<{
               // sibling outside the Resizable ever shrinks the
               // container, switch to a measured pixel offset or a
               // host-published CSS custom property.
-              right: rightPanel.collapsed()
-                ? 0
-                : `${rightPanel.panelSize() * 100}vw`,
+              // `panelOpen()` (not raw `collapsed()`) so an empty workspace —
+              // where the panel host isn't even mounted (App's `showEmpty`)
+              // — reserves no width here. Otherwise the cluster floats 25vw
+              // shy of the right edge with nothing filling the gap.
+              right: rightPanel.panelOpen()
+                ? `${rightPanel.panelSize() * 100}vw`
+                : 0,
             }
       }
     >
@@ -191,15 +195,20 @@ const ChromeBar: Component<{
             data-testid="inspector-toggle"
             class={toggleBtnClass}
             classList={{
-              "bg-surface-2 text-fg": !rightPanel.collapsed(),
+              "bg-surface-2 text-fg": rightPanel.panelOpen(),
               "text-fg-3 hover:bg-surface-2 hover:text-fg":
-                rightPanel.collapsed(),
+                rightPanel.hasTerminals() && !rightPanel.panelOpen(),
+              "text-fg-3/40 cursor-not-allowed": !rightPanel.hasTerminals(),
             }}
-            data-active={!rightPanel.collapsed() ? "" : undefined}
+            data-active={rightPanel.panelOpen() ? "" : undefined}
+            // Dead on an empty workspace: there's no panel to reveal, so the
+            // toggle joins the maximize button in being disabled until a
+            // terminal exists (the keybind/palette no-op via togglePanel too).
+            disabled={!rightPanel.hasTerminals()}
             onClick={() => rightPanel.togglePanel()}
             aria-label="Toggle right panel"
           >
-            <InspectorToggleIcon active={!rightPanel.collapsed()} />
+            <InspectorToggleIcon active={rightPanel.panelOpen()} />
           </button>
         </Tip>
         <div class="pointer-events-auto">
--- a/packages/client/src/commands.tsx
+++ b/packages/client/src/commands.tsx
@@ -342,7 +342,15 @@ export function createCommands(deps: CommandDeps): Accessor<PaletteCommand[]> {
       : []),
 
     // --- UI (panel/dock visibility — global UI chrome, not per-terminal) ---
-    actionPaletteCommand("toggleRightPanel", deps, { section: "ui" }),
+    // Hide "Toggle right panel" on an empty workspace: with no terminals the
+    // panel host is unmounted (App's `showEmpty`) and `togglePanel()`
+    // early-returns, so the command would close the palette and do nothing —
+    // exactly the "offering a command that does nothing surfaces as broken"
+    // case the canvas-arrange gate above avoids. The header button is disabled
+    // for the same reason.
+    ...(deps.terminalIds().length > 0
+      ? [actionPaletteCommand("toggleRightPanel", deps, { section: "ui" })]
+      : []),
     actionPaletteCommand("toggleDock", deps, { section: "ui" }),
 
     // --- Help (reference + advanced) ---
--- a/packages/client/src/right-panel/useRightPanel.ts
+++ b/packages/client/src/right-panel/useRightPanel.ts
@@ -217,6 +217,17 @@ export function useRightPanel() {
     reportToServer(id);
   }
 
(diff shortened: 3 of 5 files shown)


srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `a3748f8ae0f9`

### Round 1

**codex** — approved: `false`

The visual empty-workspace fix is mostly coherent: ChromeBar now keys offset and active state off effective panel visibility, and the new scenario covers the last-terminal close path. I found one remaining UX/behavior gap around the command palette still exposing a now-inert toggle action. I ran read-only inspection plus `git diff --check`; I did not run the e2e suite.

Findings:
- `F1` · minor · open — `Toggle right panel` remains an always-selectable command palette action, but this branch changed `rightPanel.togglePanel()` to return early when there are no terminals. After the last terminal closes, the header button is disabled, but selecting the palette command still closes the palette and performs no visible or persisted action. That leaves a clickable command that silently does nothing, which the local palette helper comments explicitly try to avoid. (packages/client/src/commands.tsx:345)

**claude** — Agreed with CODEX's single finding (F1): the "Toggle right panel" command palette action was unconditionally registered while the header button and keybind were both gated/no-op'd for empty workspaces, so selecting it would close the palette and do nothing. Fixed by gating the palette command behind `deps.terminalIds().length > 0` in commands.tsx — the same idiom the file already uses for the "Arrange canvas by repo" command (hidden for single-tile to avoid "a command that does nothing surfaces as broken") and the workspace-search group. Verified with tsc --noEmit (clean) and biome (no reformat).

- `F1` **fixed** — Gated the palette command at packages/client/src/commands.tsx:345 behind `deps.terminalIds().length > 0` (spreading `[actionPaletteCommand("toggleRightPanel", ...)]` only when terminals exist), with a comment tying it to the same rationale as the existing canvas-arrange gate and the disabled header button. This removes the inert command entirely on an empty workspace rather than leaving a clickable no-op. Chose the gating approach (CODEX's primary suggestion) over the disabled-command/test-assertion alternative because it matches three existing conditional-spread idioms in this exact file and needs no new palette-state step definitions. Typecheck (tsc --noEmit) and biome format both pass clean.

commit: `fb3f2573451ae0d6d847f30eb7580ee6e8f99a6c`

### Round 2

**codex** — approved: `true`

CLAUDE's follow-up directly addresses the only finding from the prior round. The command palette now omits `Toggle right panel` when `deps.terminalIds().length === 0`, matching the disabled header affordance and avoiding a selectable no-op in the empty workspace. I saw no new regression introduced by this round's `commands.tsx` change.

Findings:
- `F1` · minor · resolved — Previously, `Toggle right panel` remained selectable in the command palette on an empty workspace even though `togglePanel()` early-returned with no terminals, closing the palette and doing nothing. (packages/client/src/commands.tsx:351)

_codex on the rebuttal:_ No dispute to resolve. CLAUDE agreed with F1 and I verified the fix: `Toggle right panel` is now gated behind `deps.terminalIds().length > 0`, so the empty-workspace palette no longer exposes an inert action.

srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `a3748f8ae0f9`

### Applied (1)
- `hickey-3` togglePanel() guards on hasTerminals() but reveal()/expandPanel()/collapsePanel() don't — the empty-workspace rule is fragmented within useRightPanel — commit `47948cd43`

### Agreed — no change (6)
- `lowy-1` Empty-workspace visibility rule correctly pulled behind the useRightPanel receptacle (packages/client/src/right-panel/useRightPanel.ts:184-241)
- `lowy-2` commands.tsx exposes the wires: re-derives hasTerminals() instead of plugging into the receptacle (packages/client/src/commands.tsx:351)
- `lowy-3` ChromeBar disabled/active rule is symmetric with the maximize button — good (packages/client/src/ChromeBar.tsx:197-207)
- `hickey-1` "Panel host exists" fragmented across three modules with three predicates (packages/client/src/right-panel/useRightPanel.ts:184, packages/client/src/App.tsx:405-406, packages/client/src/commands.tsx:351)
- `hickey-2` ChromeBar reads panelOpen() while App.tsx desktop branch reads raw collapsed() for the same panel (packages/client/src/App.tsx:632,682,705,715 vs packages/client/src/ChromeBar.tsx:115,198,203,211)
- `hickey-4` ChromeBar toggle's three-way classList re-derives the panelOpen / hasTerminals state machine as overlapping booleans (packages/client/src/ChromeBar.tsx:197-207)

srid:

Ran the three-pass sweep (rules → fact-check → elegance) over the settled tree after the codex and lens debates.

| Pass | Found | Outcome |
| --- | --- | --- |
| **Rules** | 1 | `e2e-poll-async-state` — fixed |
| **Fact-check** | 0 | Clean — `panelOpen()` gating, the `setCollapsed` empty-workspace no-op, and `right: 0 → "0px"` all check out |
| **Elegance** | 1 | Same finding as Pass 1 (the `isDisabled()` race) — fixed |

**Fixed (`84d919f2`)** — the new `"the inspector toggle should be disabled"` step used `waitFor(attached)` then `await toggle.isDisabled()`, a *snapshot* that doesn't poll: if SolidJS hadn't flushed the `disabled` attribute yet, the assertion would pass a non-disabled button. Replaced with the repo's polling idiom — `locator('…:disabled').waitFor({ state: 'attached' })` — matching `code_tab_steps.ts`.

**Noted, not actioned (out of scope):**
- `hasTerminals()` in `useRightPanel` and `canMaximize` in `useViewPosture` are both `() => store.terminalIds().length > 0`. Hoisting the predicate into `useTerminalStore` is a refactor beyond this diff.
- `commands.tsx:351`'s `terminalIds().length > 0` repeats line 160's check — the file's established conditional-spread idiom (already adjudicated `drop` by the lens debate).

srid:
## Evidence

Empty workspace (zero terminals, welcome card centred), 1440 × 900 viewport — before and after.

| | |
|---|---|
| **Before** (`master`) | **After** (`ill-boot`) |
| Before | After |

**Before:** the control cluster (record, maximize, dock, inspector-toggle, settings, ⌘K) floats ~25 vw shy of the right edge, leaving a blank ghost-panel gap.

**After:** the cluster sits flush against the right edge and the inspector-toggle button is dimmed (disabled — no terminals to inspect).

PR #1231, opened 2026-06-08
Gate the WS handshake by server processId to stop stale-tab storms

srid:
**A browser tab left open across a server restart used to storm the new process.** On `just dev` re-exec (or any restart), the stale tab's PartySocket reconnects to the *fresh* process, and oRPC's retry plugin transparently replays every per-terminal subscription — `attach`, `terminalExit`, `terminalMetadata` — against terminals the new process never had. The result is a flood of `NOT_FOUND` / "key not found at first snapshot" errors and a server that looks unresponsive.

We already mint a per-boot `serverProcessId`, but nothing was *gated* with it. Now the handshake is — at the one seam every stream passes through.

### How the gate works

```
client reconnect ──?pid=<last-known>──▶ server upgrade
                                          │
                    pid absent / matches ─┴─▶ accept → oRPC upgrade  (normal)
                    pid mismatches ─────────▶ close 4001 BEFORE upgrade
                                                  │
                                          client: restarted → reload overlay
```

- **Client** echoes its last-known `processId` as a `pid` query param on every (re)connect (`SERVER_PROCESS_ID_PARAM`), fed by the identity probe via `rememberServerProcessId`. The very first connect has no pid and passes.
- **Server** rejects a mismatched tab at `wss.on("connection")` — closing with `STALE_PROCESS_CLOSE_CODE` (4001) *before* oRPC upgrades the socket, so **not one dead-terminal subscription ever runs**. A transient drop reconnects with a *matching* pid and is untouched.
- **Lifecycle** reads 4001 as a definitive `restarted` (no probe can run — the socket is already gone), and `rpc.ts` halts PartySocket's auto-reconnect so the existing "App updated" reload overlay is the single path forward, *not* a loop that keeps re-presenting the same dead id.

This is the principled pair from the design discussion: **(1)** carry the expected processId on the handshake and gate on it server-side, and **(4)** let the client recognize the rejection and hand off to the restart overlay instead of replaying subscriptions.

### Notable

> The new `restartCloseCode` option on `@kolu/surface-app`'s `createServerLifecycle` keeps the package generic — kolu passes its 4001 in; surface-app just maps "closed with this code" → `restarted`. Covered by two new `lifecycle.test.ts` cases (definitive restart; ignored before any identity is established).

The restart round-trip has zero visual diff in steady state, so proof is behavioral (restart the server → overlay appears, logs stay quiet) — captured as evidence rather than a flaky process-restart e2e.


srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 round(s) · codex reviewed at `xhigh` reasoning effort · base `a3748f8ae0f9`

### Round 1

**codex** — approved: `false`

The branch type-checks for common/client/server/surface-app (`tsc --noEmit` all passed), and the core handshake idea is coherent. I found one major client-side failure mode from stopping PartySocket reconnect on a stale close, plus two smaller API/UI consistency issues around the new restarted-via-close path.

Findings:
- `F1` · major · open — The stale-close handler calls `ws.close()` to disable PartySocket reconnect, but the same ReconnectingWebSocket still queues future `send()` calls while CLOSED. With reconnect disabled, those messages never flush; active stream retries and any user input/RPC after the overlay appears can hang silently, and terminal keystrokes can grow the internal queue without bound. (packages/client/src/rpc/rpc.ts:70)
- `F2` · minor · open — `wsStatus` still maps every `restarted` lifecycle to `open`. In the new stale-close path, `restarted` is reached because the server closed the socket and the client then intentionally stops reconnecting, so the header/diagnostics report an open WebSocket while it is actually closed. (packages/client/src/rpc/rpc.ts:85)
- `F3` · minor · open — The new `restartCloseCode` option is only available when callers use `createServerLifecycle` directly. The advertised turnkey provider path (`<SurfaceAppProvider ws={...} probe={...}>`) cannot pass it through, so package consumers using the provider will still see a restart close as a normal disconnect/reconnect state. (packages/surface-app/src/solid/index.ts:326)

**claude** — Addressed all three CODEX findings; all are valid. F1 (major): after the stale-close `ws.close()`, partysocket's `maxEnqueuedMessages` is Infinity and oRPC's websocket link calls `ws.send()` directly, so post-stale sends (terminal keystrokes through the pointer-events-none overlay, stream retries) would enqueue unbounded and never flush — fixed by neutering `ws.send` to a drop in the stale-close handler. F2 (minor): the `srv` header dot mapped `restarted`→`open` even on the stale path where the socket is genuinely CLOSED — fixed by a `staleClosed` signal that maps the stale-restart to `closed` (red) while leaving the reconnect-restart as `open`. F3 (minor): `restartCloseCode` was unreachable via the turnkey `<SurfaceAppProvider ws={} probe={}>` source — added `restartCloseCode?: number` to that ConnectionSource branch and forwarded it into the provider's internal `createServerLifecycle`, plus a provider test and a README note. Verified: tsc --noEmit passes for client/surface-app/common/server; surface-app unit suite 58/58 pass; biome format clean.

- `F1` **fixed** — Confirmed the mechanism by reading partysocket dist (ws.cjs: maxEnqueuedMessages defaults to Number.POSITIVE_INFINITY; send() enqueues when readyState !== OPEN; close() sets _shouldReconnect=false) and oRPC's websocket adapter (its ClientPeer callback calls options.websocket.send(message) directly, and its untilOpen only awaits when readyState===CONNECTING — after a stale CLOSED close it resolves immediately). So with reconnect disabled, every later send silently grows an unflushable queue. Fix in packages/client/src/rpc/rpc.ts:90 — the stale-close handler now also does `ws.send = () => {}` so post-stale writes drop immediately instead of accumulating. Normal transient-drop buffering is untouched (only the terminal stale state neuters send; a fresh page restores a pristine socket).
- `F2` **fixed** — Agreed: in the stale-close path `restarted` is reached with the socket genuinely CLOSED, but wsStatus mapped restarted→open, so the srv dot (srvDot in IdentityRail.tsx: open=bg-ok green) showed green behind the reload overlay. Added a `staleClosed` createSignal set in the stale-close handler (packages/client/src/rpc/rpc.ts:86-92) and changed the wsStatus memo so `restarted` maps to `closed` when staleClosed() else `open` (rpc.ts:108). This keeps the reconnect-restart case (socket actually open against a fresh process) green and only the stale-restart case red — derives the dot from the real socket disposition as suggested, without conflating the two restart sub-cases.
- `F3` **fixed** — Agreed: the provider's turnkey `{ ws, probe }` branch called createServerLifecycle without forwarding restartCloseCode, so provider-path consumers (e.g. drishti) couldn't get the synchronous stale-restart behavior. Added `restartCloseCode?: number` to that ConnectionSource branch (packages/surface-app/src/solid/index.ts:~330) and forwarded `restartCloseCode: props.restartCloseCode` into the provider's internal createServerLifecycle call (~index.ts:410). Added a provider.test.ts case driving the `{ ws, probe, restartCloseCode }` path that fires a 4001 close and asserts status()==='restarted'/updateReady()===true, and updated the surface-app README design note to say the option is forwarded through the provider's turnkey source.

commit: `6e4e37a39f0827772d2d6cce1943aeec33b496dc`

### Round 2

**codex** — approved: `false`

CLAUDE fixed F2 and F3. F1's specific PartySocket offline queue is no longer growing, but the replacement `send` no-op leaves oRPC calls hanging forever, so repeated terminal input after the stale overlay can still accumulate unresolved peer requests/promises. I'm keeping F1 open on that narrower remaining failure mode.

Findings:
- `F1` · major · open — Replacing `ws.send` with a no-op stops PartySocket's internal offline queue, but it makes later oRPC sends appear successful. The oRPC websocket adapter awaits `options.websocket.send(message)` and then waits for a response; with this no-op, each post-stale RPC/stream retry opens a peer request that never receives a response or close. Because the overlay allows interaction underneath, repeated terminal input can still accumulate unresolved oRPC requests/promises even though the PartySocket queue is gone. (packages/client/src/rpc/rpc.ts:90)
- `F2` · minor · resolved — Previously `wsStatus` reported `open` for the stale-close restart path even though the socket was intentionally closed. (packages/client/src/rpc/rpc.ts:104)
- `F3` · minor · resolved — Previously `restartCloseCode` could not be passed through the turnkey `<SurfaceAppProvider ws probe>` path. (packages/surface-app/src/solid/index.ts:330)

_codex on the rebuttal:_ F1: I agree with CLAUDE's diagnosis that PartySocket's offline queue was the original unbounded buffer, but the no-op `send` only moves the hang one layer up. oRPC treats `send()` returning normally as a sent request and then waits for a response that can never arrive after reconnect has been disabled. That still permits unbounded unresolved request state under continued input, so F1 remains open. F2: accepted; the stale-close path now reports the socket as closed while preserving green/open for reconnect-restart. F3: accepted; the provider turnkey path now exposes and forwards `restartCloseCode`, with a focused test.

**claude** — Addressed the one remaining open finding (F1). Codex's narrower diagnosis was correct, so I fixed it rather than disputing. F2 and F3 were already resolved in the prior round and codex accepted them.

F1: I traced the oRPC websocket adapter (@orpc/client/dist/adapters/websocket/index.mjs) and ClientPeer (@orpc/standard-server-peer/dist/index.mjs). The previous no-op `ws.send` stopped partysocket's unbounded offline queue, but ClientPeer.request() does `await this.send(...)` (the no-op resolves, looking like a successful dispatch) then `await this.responseQueue.pull(id)` — which only settles on a response message or peer.close(). After a stale close with reconnect disabled, neither ever happens for a NEW request, so every post-stale RPC/stream retry (terminal keystrokes through the pointer-events-none overlay) hangs forever, accumulating unresolved promises and open queue entries — the same unbounded-growth one layer up. Fix: replace the no-op with a `send` that THROWS a stable stale-tab error. oRPC's request() awaits the send, the throw hits its `catch (err) { this.close({id, reason: err}); throw err }`, so the call rejects cleanly through callers' existing error paths instead of hanging. Requests already in flight at the stale close are still settled by oRPC's own websocket `close` listener (which fired with this same event). Throwing also means partysocket's send method (and its buffer) never runs at all, so the original F1 buffer-growth fix is preserved. Verified: client tsc --noEmit passes (exit 0); biome format clean on rpc.ts.

- `F1` **fixed** — Codex is right that the no-op send only moved the hang one layer up. Confirmed against the actual oRPC dist: ClientPeer.request() awaits send() (no-op resolves => looks dispatched) then awaits responseQueue.pull(id), which for a NEW post-stale request never settles (no response, and reconnect is off so no further close event fires to call peer.close()). Replaced the no-op at packages/client/src/rpc/rpc.ts:87-91 with a send that throws a stable 'kolu: server restarted — reload required (stale tab)' error. oRPC's request() catch then closes that request id and rejects the call, so callers learn the message was dropped via their existing error paths instead of believing it was accepted. In-flight-at-close requests are still settled by oRPC's own websocket close listener. Throwing replaces partysocket's send entirely, so nothing is enqueued — the original buffer-growth fix holds. Updated the explanatory comment to document why throw, not no-op. tsc clean, biome clean.
- `F2` **fixed** — Already resolved and accepted by codex in the prior round (staleClosed signal maps the stale-restart to 'closed'/red while the reconnect-restart stays 'open'/green). No change this round.
- `F3` **fixed** — Already resolved and accepted by codex in the prior round (SurfaceAppProvider turnkey {ws, probe} branch accepts and forwards restartCloseCode, with provider test coverage). No change this round.

commit: `f5ee0e3fe2f88782877e4ebcae5b55d4468d0212`

### Round 3

**codex** — approved: `true`

Current status has no untracked files. I verified the remaining F1 change in `packages/client/src/rpc/rpc.ts`: post-stale `send` now throws instead of no-oping, so oRPC's request path rejects and closes the request id rather than enqueuing in PartySocket or leaving a peer request waiting forever. F2 and F3 remain addressed from the prior round.

Findings:
- `F1` · major · resolved — Previously post-stale sends either accumulated in PartySocket's offline queue or, after the first attempted fix, appeared successful to oRPC and left requests waiting forever. (packages/client/src/rpc/rpc.ts:101)
- `F2` · minor · resolved — Previously `wsStatus` reported `open` for the stale-close restart path even though the socket was intentionally closed. (packages/client/src/rpc/rpc.ts:117)
- `F3` · minor · resolved — Previously `restartCloseCode` could not be passed through the turnkey `<SurfaceAppProvider ws probe>` path. (packages/surface-app/src/solid/index.ts:330)

_codex on the rebuttal:_ F1: accepted. The new throwing `send` fixes the remaining hang I raised: oRPC no longer sees a normal return from `send`, so `ClientPeer.request()` takes its catch path, closes the request id, and rejects instead of waiting forever. It also bypasses PartySocket's offline queue, preserving the original buffer-growth fix. F2 and F3 were already accepted in the prior round and remain resolved.

srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `a3748f8ae0f9`

### Applied (8)
- `lowy-1` Stale-restart fact derived twice — receptacle interpreted, then wires re-exposed — commit `2e4b91298`
- `lowy-2` `ws.send`/`ws.close` mutation is a control-plane action stuffed into the data-plane object — commit `6f7087f09`
- `lowy-3` Request URL parsed twice on the upgrade path — commit `8ee06b185`
- `hickey-1` `restarted` lifecycle fragmented across `lifecycle().kind` + a separate `staleClosed` signal — (uncommitted)
- `hickey-2` Two independent `close` listeners on the same socket split one decision across library + app — (uncommitted)
- `hickey-3` `serverProcessId` reports the OLD (detached) process id on a stale restart — commit `d11de427a`
- `hickey-4` Stale-tab teardown braids three independent concerns into one imperative close handler — (uncommitted)
- `hickey-6` `restartCloseCode` adds a second, redundant restart-signalling channel to createServerLifecycle — (uncommitted)

### Agreed — no change (2)
- `lowy-4` `restartCloseCode` generic naming on a kolu-specific axis — correct, noting as a passed check (packages/surface-app/src/solid/index.ts:113,335,423)
- `hickey-5` `pid` query param is plumbed through a module-level mutable + URL thunk rather than read from the one lifecycle that owns it (packages/client/src/wire.ts:50-59; packages/client/src/rpc/rpc.ts:44-51)

srid:

Ran the three-pass review (rules → fact-check → elegance) on the branch after the codex and lens debates settled.

| Pass | Findings | Notes |
|---|---|---|
| **Rules** (25) | 0 | Clean. Highlights: `invalid-states-unrepresentable` — the new `restarted` `transport: "closed"` shape can't carry a stale `processId` by construction; `no-unbounded-growth` — `retireSocket` explicitly bounds the partysocket offline buffer + oRPC peer accumulation; `prefer-ts-pattern` — `wsStatus` moved from string-dispatch to structural `match(lifecycle())` with exhaustive checking. |
| **Fact-check** | 0 (1 observation) | The lens-introduced module-scope `createEffect` driving stale-socket retirement was **unowned**. |
| **Elegance** | 0 | Clean across reuse / simplification / efficiency / altitude. |

**Actioned** the one observation rather than waving it through: a bare top-level `createEffect` is push-based and relies on chance scheduling, and this effect is correctness-critical (it's what bounds the buffers behind the reload overlay). Wrapped it in an explicit `createRoot` — the codebase's module-singleton ownership pattern (`createSharedRoot.ts`) — so it reliably runs and Solid stops warning. Page-lifetime owner, never disposed (correct for a singleton). Commit `24261c72`.

_Reviewed by [`/code-police`](https://agency.srid.ca/) — Claude Code (model `claude-opus-4-8`)._

srid:
## Evidence

This is a reconnect/restart round-trip with no steady-state visual diff, so the proof is **behavioral**: a stale tab surviving a real server restart now gets one clean rejection instead of the `NOT_FOUND` storm from the report — and recovers via the reload overlay.

Driven live (`just dev-auto`, random ports, production `kolu.service` untouched): a browser tab connected to the dev server, then `node --watch` re-exec'd the server (a fresh `processId`) — the exact scenario from the bug report.

### Server logs — the storm is gone

The new instance (`8770ce87`) sees the stale tab still carrying the previous id (`5c38633c`) and rejects it at the handshake, **before** any oRPC upgrade:

```
13:15:42  INFO  kolu listening   {"serverId":"8770ce87…"}            ← restarted, fresh processId
13:15:43  INFO  rejecting stale client — server restarted since it last connected
                {"claimedPid":"5c38633c…","serverProcessId":"8770ce87…"}
13:16:16  INFO  connected        {"serverId":"8770ce87…","ws":2}     ← after reload: fresh client, clean
```

Counted over the whole run:

| | count |
|---|---|
| `rejecting stale client` lines | **1** — single rejection, no reconnect loop (`retireSocket` halted it) |
| `Terminal … not found` / `key not found at first snapshot` / `terminal.attach` errors | **0** — the storm is eliminated |

Before this PR, that same restart produced a flood of `ERROR` stack traces (the `terminalExit.get` / `terminalMetadata.get` / `terminal.attach` NOT_FOUND storm in the issue) and a server that looked unresponsive.

### Client — overlay takes over, then recovers

The tab reads the `4001` close as a definitive restart: the socket goes `CLOSED` (`readyState 3`, reconnect halted — no loop), the **"App updated → Reload"** overlay appears, and `ws.send` is poisoned so no oRPC request hangs or buffers unbounded behind it. Clicking Reload lands a fresh page that connects cleanly to the new instance (`ws:2` above) — back to the baseline state.

| After restart — reload overlay (socket CLOSED) | Connected (baseline == recovered after reload) |
|---|---|
| overlay | connected |

_Captured live via chrome-devtools on a `just dev-auto` instance (random ports); production `kolu.service` PID untouched throughout._

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `de18d86a8fe2`

### Round 1

**codex** — approved: `false`

The stale-process handshake is directionally sound and the server/client process-id plumbing is mostly coherent. I found one substantive retry-policy bug in the socket retirement path: the new throwing send still looks like a retriable transport error to the existing oRPC stream retry layer, so stale tabs can keep retrying and repeatedly reset terminal buffers behind the reload overlay. I also found one small misleading comment.

Findings:
- `F1` · major · open — `retireSocket` replaces `send` with a stub that throws a plain `Error`. All framework streams use `STREAM_RETRY` with infinite retries and `shouldRetry` only suppressing `ORPCError`, so this retired-socket error is still retried forever. For terminal attach streams, each retry runs `onRetry` and calls `terminal.reset()` (`packages/client/src/terminal/Terminal.tsx:717`), which can keep clearing the readable terminal buffer behind the stale-restart reload overlay. It also means the retired tab still has active retry loops instead of settling as intended. (packages/surface-app/src/solid/index.ts:255)
- `F2` · nit · open — The comment says "`retireSocket` in `wire.ts`", but the implementation is imported from `@kolu/surface-app/solid`; `wire.ts` only owns the socket and process-id URL state. That sends future readers to the wrong module for the transport-retirement behavior. (packages/client/src/rpc/rpc.ts:65)

**claude** — Addressed both CODEX findings. F1 (major retry-policy bug) is valid and now fixed: the retired-socket stub threw a plain Error, which the shared retry fence (shouldNotRetryORPCError → !(error instanceof ORPCError)) classifies as retriable, so STREAM_RETRY consumers (infinite retries) would re-subscribe forever — and for the terminal attach stream each retry fires onRetry → terminal.reset(), clearing the buffer behind the reload overlay. Fixed by throwing an ORPCError, the same non-retry mechanism the dead stdio link already uses (SURFACE_STDIO_TRANSPORT_CLOSED). F2 (nit) is valid and the misleading comment is corrected.

- `F1` **fixed** — Valid bug, confirmed against an existing codebase precedent. The retry fence is shouldNotRetryORPCError in packages/surface/src/client.ts:18 (returns !(error instanceof ORPCError)), and STREAM_RETRY (line 27) uses retry: Infinity with it. retireSocket's stub threw a plain Error, so the fence returned true → retried forever; Terminal.tsx:717-720's onRetry runs terminal.reset() each retry. The dead stdio link already solves this exact class of problem by throwing an ORPCError (stdio.ts:122, SURFACE_STDIO_TRANSPORT_CLOSED) so the same fence excludes it. Fix: added transportRetiredError(message) to packages/surface/src/client.ts (returns new ORPCError("SURFACE_TRANSPORT_RETIRED", { message })), and retireSocket in packages/surface-app/src/solid/index.ts:264 now throws it instead of a plain Error. surface-app already depends on @kolu/surface, so importing @kolu/surface/client adds no new coupling and keeps the oRPC-shape knowledge in surface, which owns the retry policy. Added a regression test in lifecycle.test.ts asserting shouldNotRetryORPCError({ error: thrown }) === false (i.e. the retired throw is non-retriable). All surface-app (9) and surface (70) unit tests pass; tsc --noEmit clean for surface, surface-app, and client.
- `F2` **fixed** — Valid. The comment at packages/client/src/rpc/rpc.ts:65 said retireSocket lives "in wire.ts", but it is imported from @kolu/surface-app/solid (see the import at rpc.ts:18-23). Reworded the comment to point at @kolu/surface-app/solid for retireSocket and the partysocket/oRPC-internals knowledge, and to note that wire.ts only owns the ws instance being retired (plus its process-id URL state).

commit: `0964c55c28a3e550fe5f2ce8b1320d2298ae7a4f`

### Round 2

**codex** — approved: `true`

The two previously raised issues are addressed in the current tree. The retired socket now throws an `ORPCError` produced by `transportRetiredError`, which the existing `shouldNotRetryORPCError` predicate classifies as non-retriable, and the regression test checks that classification. The misleading `retireSocket` comment now points to `@kolu/surface-app/solid` and describes `wire.ts` as only owning the socket instance and URL state. I did not find any new regression introduced by this round's changes.

Findings:
- `F1` · major · resolved — Previously, `retireSocket` threw a plain `Error`, so infinite `STREAM_RETRY` consumers would treat the retired socket failure as retriable and keep re-subscribing. (packages/surface-app/src/solid/index.ts:264)
- `F2` · nit · resolved — Previously, the comment said `retireSocket` lived in `wire.ts`, which sent readers to the wrong module. (packages/client/src/rpc/rpc.ts:65)

_codex on the rebuttal:_ CLAUDE accepted both findings and fixed them. F1 is resolved: the retired socket now throws an `ORPCError`, matching the retry fence's existing non-retry mechanism, and the added test verifies `shouldNotRetryORPCError` returns false. F2 is resolved: the comment now accurately points to `@kolu/surface-app/solid` for `retireSocket` and scopes `wire.ts` to the socket instance/URL state.

srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `de18d86a8fe2`

### Applied (5)
- `lowy-1` New surfaceAppServer/serverIdentity processId accessor has no in-tree consumer that reads it — commit `3ae0441f2`
- `lowy-2` retireSocket is framework-free transport code shipped from the /solid entry — commit `52be12343`
- `hickey-1` surfaceAppServer/serverIdentity `processId` accessor is dead in the only real consumer — (uncommitted)
- `hickey-2` Two constructors for one concept: a permanently-dead transport's non-retriable ORPCError — commit `714816e05`
- `hickey-3` `restartCloseCode` semantics are split between core (the constant) and /solid (the close→restart mapping), with the core docstring asserting the mapping — commit `48c927d94`

### Agreed — no change (5)
- `lowy-3` retireSocket lives in surface-app but its non-retry contract is owned by surface — the receptacle is split across two packages (packages/surface-app/src/solid/index.ts:24,266 and packages/surface/src/client.ts:30-34)
- `lowy-4` WsLike overload widened from a pure observation receptacle to a code-carrying one — verify it stays observation-only (packages/surface-app/src/solid/index.ts:84-94)
- `lowy-5` Stale-restart classification gated on knownProcessId !== null is duplicated across the open and close paths (packages/surface-app/src/solid/index.ts:170-208)
- `hickey-4` `transport` discriminator on `restarted` duplicates the `processId`-presence discriminator (packages/surface-app/src/solid/index.ts:447-448)
- `hickey-5` `retireSocket`'s structural `{ close, send }` arg with `send: unknown` leaks the mutation contract into the type as untyped (packages/surface-app/src/solid/index.ts:594-601)

srid:

Ran rules → fact-check → elegance on the settled tree (after the codex and lens debates).

| Pass | Result |
|---|---|
| **Rules** | Clean |
| **Fact-check** | Clean |
| **Elegance** | 1 fix |

**Fixed** — `transportRetiredError` (added in the codex round, then unified by the lens debate into `deadTransportError`) had become a thin single-caller wrapper that only bound the `SURFACE_TRANSPORT_RETIRED` code and forwarded `message`. Inlined `deadTransportError(SURFACE_TRANSPORT_RETIRED, …)` at its one call site in `@kolu/surface-app`'s `lifecycle.ts` and dropped the wrapper from `@kolu/surface/client` — one fewer hop for the reader, same behavior. Commit `7292cc5f`.

**Skipped (with reason)** — `fakeWs` duplication across two test files (below rule-of-three, and `provider.test.ts`'s copy is an intentionally slimmer subset); `encodeURIComponent` on a UUID (harmless no-op); the `config.ts` navigation comment (informative, not dead code).

_Reviewed by [`/code-police`](https://agency.srid.ca/) — Claude Code (model `claude-opus-4-8`)._

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `bba087fd957c`

### Round 1

**codex** — approved: `false`

The branch is directionally sound: the stale-tab PID handshake is modeled in the shared surface-app layer, Kolu wires it through the server and PartySocket URL thunk, and the unit coverage covers the pure gate and lifecycle fast path. I found one server-side robustness issue in the rejected-socket path, one lifecycle callback isolation bug, and a couple of docs/API-sync nits that should be fixed before approval.

Findings:
- `F1` · major · open — The stale-client rejection path returns before installing the WebSocket error handler. A rejected socket is still a live `ws` EventEmitter until its close handshake completes; if the peer resets or the socket emits `error` after `ws.close(...)`, Node can treat that unhandled `error` event as an uncaught exception. Normal connections get an error handler at lines 367-369, but stale connections do not. (packages/server/src/index.ts:345)
- `F2` · minor · open — `onProcessId` runs inside the probe `.then` before `knownProcessId` and `lifecycle` are updated. If the consumer callback throws, a successful identity probe is converted into a probe failure, the lifecycle state is not updated, and the UI can remain stuck in `connecting` or `disconnected`. An observer hook should not be able to poison the core lifecycle classification. (packages/surface-app/src/solid/index.ts:168)
- `F3` · minor · open — The entrypoint table is now stale for `@kolu/surface-app/lifecycle`: `retireSocket` is implemented and exported from `src/lifecycle.ts`, but the README row still lists only the service-worker/reload functions. That violates the repo rule to keep README in sync with user-facing changes. (packages/surface-app/README.md:104)
- `F4` · nit · open — The architecture prose says "an effect in `rpc.ts` fires surface-app's `retireSocket(ws)`," but the implementation wires it as `createServerLifecycle`'s `onStaleRestart` callback, not a reactive effect. That is a small but misleading description of where the restart close code is decoded and acted on. (README.md:304)

**claude** — Addressed all four CODEX findings on the stale-tab PID handshake branch. Agreed with and fixed every one: F1 (server error-handler ordering), F2 (observer callback poisoning the lifecycle, plus added a regression test), F3 (/lifecycle README row missing retireSocket), F4 (README.md architecture prose mislabeling the onStaleRestart callback as an effect). All 68 surface-app unit tests pass (one new), full unit suite green, biome format clean.

- `F1` **fixed** — Agreed. Moved the `ws.on("error", ...)` handler to immediately after `connLog` is created, BEFORE the stale-process gate's early `return` (packages/server/src/index.ts:342-344), and removed the now-duplicate handler that previously sat only on the accepted path. A rejected socket is still a live ws EventEmitter until its close handshake settles, so a post-`ws.close()` peer reset would have emitted an unhandled `error` and become an uncaught exception. The oRPC `wsRpcHandler.upgrade(ws, ...)` stays on the accepted path only (line 364), exactly as suggested.
- `F2` **fixed** — Agreed. Reordered `onOpen`'s probe `.then` in packages/surface-app/src/solid/index.ts so `knownProcessId`/`setLifecycle` are committed FIRST (the if/else classification), then `opts.onProcessId?.(processId)` fires in a guarded `try/catch` that routes a throwing consumer to `onProbeError` instead of unwinding into the probe `.catch` — so an observer can no longer convert a successful identity probe into a probe failure or leave the UI stuck in connecting/disconnected. Updated the `onProcessId` doc comment to state it fires after the transition and is guarded, and added a test 'a throwing onProcessId does not poison the lifecycle transition' (lifecycle.test.ts) asserting lifecycle still reaches `connected` while the throw surfaces via onProbeError.
- `F3` **fixed** — Agreed. `retireSocket` is exported from src/lifecycle.ts and `./lifecycle` maps to that file in package.json, so the `@kolu/surface-app/lifecycle` README row was stale. Added `retireSocket` to that row (packages/surface-app/README.md:104) with a note that it is the stale-tab transport teardown the /solid lifecycle's onStaleRestart calls, living here because it's pure transport manipulation (no SolidJS).
- `F4` **fixed** — Agreed. rpc.ts wires retireSocket via createServerLifecycle's `onStaleRestart` callback (rpc.ts:71), not a reactive effect. Reworded README.md:304 from 'an effect in rpc.ts fires surface-app's retireSocket(ws)' to '... fires its onStaleRestart callback at that single decode site; kolu's rpc.ts wires that callback to surface-app's retireSocket(ws) ...'.

commit: `0a233fc847347a4c3e716ba9f6d8e4153fff375d`

### Round 2

**codex** — approved: `true`

Targeted re-review only: CLAUDE fixed all four previously raised issues, and I did not find a regression introduced by those fixes. `git status --short` shows no untracked files. No-emit typechecks for `packages/server` and `packages/surface-app` pass.

Findings:
- `F1` · major · resolved — Rejected stale WebSocket connections previously returned before any `error` listener was attached, so a socket error after `ws.close(...)` could surface as an unhandled EventEmitter error and crash the process. (packages/server/src/index.ts:337)
- `F2` · major · resolved — `onProcessId` previously ran before the lifecycle transition and could throw, converting a successful identity probe into a probe failure and leaving lifecycle state stale. (packages/surface-app/src/solid/index.ts:169)
- `F3` · minor · resolved — The `/lifecycle` entrypoint documentation omitted `retireSocket` even though the package exports it there. (packages/surface-app/README.md:104)
- `F4` · nit · resolved — The architecture prose described `retireSocket(ws)` as being fired by an effect in `rpc.ts`, but the implementation wires it through `createServerLifecycle`'s `onStaleRestart` callback. (README.md:304)

_codex on the rebuttal:_ F1: accepted; installing the error listener before the stale-process early return addresses the crash path without upgrading rejected sockets. F2: accepted; committing lifecycle state before the observer and reporting observer throws via `onProbeError` removes the poisoning path, and the added test exercises it. F3: accepted; the documented `/lifecycle` exports now match `package.json` and `src/lifecycle.ts`. F4: accepted; README now matches the callback-based implementation rather than describing a reactive effect. No disputes remain.

srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `bba087fd957c`

### Agreed — no change (12)
- `lowy-1` Electricity test: the receptacle is sound — runtime-free wire contract behind one socket interface (packages/surface-app/src/index.ts:200-211, packages/surface-app/src/solid/index.ts:122-147)
- `lowy-2` retireSocket leaks the concrete transport's internals across the package boundary it claims to hide (packages/surface-app/src/lifecycle.ts:49-57)
- `lowy-3` Turnkey provider hard-wires onStaleRestart to retireSocket with no override seam — fine, because retirement is the axis the package owns (packages/surface-app/src/solid/index.ts:519)
- `lowy-4` Stale-tab decision single-sourced, but the liveId↔probe-id coupling is enforced only by prose, not by the type (packages/server/src/index.ts:359-362, packages/surface-app/src/server.ts:321-336)
- `lowy-5` deadTransportError factory correctly unifies the non-retry shape across two transports — coherent receptacle (packages/surface/src/client.ts:40-45, packages/surface/src/links/stdio.ts:122-125)
- `lowy-6` The `pid` echo state lives in wire.ts as a module-level mutable, correctly outside the library (packages/client/src/wire.ts:48-58)
- `hickey-1` Last-observed processId fragmented across three exit points (serverProcessId, onProcessId, event.processId) (packages/surface-app/src/solid/index.ts:131-140)
- `hickey-2` Provider's turnkey source re-declares lifecycle hooks as hand-listed passthrough props (packages/surface-app/src/solid/index.ts:399-417)
- `hickey-3` `restarted` discriminated by `transport` braids socket-aliveness into identity classification (packages/surface-app/src/solid/index.ts:69-70)
- `hickey-4` onStaleRestart duplicates the lifecycle event channel for one transition (packages/surface-app/src/solid/index.ts:141-148)
- `hickey-5` Rich-socket shape `WsLike & { close; send: unknown }` restated inline at two type sites (packages/surface-app/src/lifecycle.ts:42)
- `hickey-6` Server connection handler: error-attach-before-gate correctness rides a comment, not structure (packages/server/src/index.ts:329-352)

srid:

Ran rules → fact-check → elegance on the settled tree (after codex + lens).

| Pass | Result |
|---|---|
| **Rules** | Clean |
| **Fact-check** | Clean |
| **Elegance** | 3 small fixes |

**Fixed** (commit `13c378d9`): dropped the no-op `encodeURIComponent` on the echoed `pid` (it's a UUID — already URL-safe); trimmed two "where this moved" archaeology comments (those belong in the commit message, not module code); collapsed the 4-line `emit` 3rd-arg explanation to a one-liner at its callsite.

**Skipped (with reason):** keeping the stale-gate in `wss.on("connection")` rather than the `handleUpgrade` callback (the callback has no `connId` yet, so moving it would lose per-rejection log ids — observability wins); `fakeWs` across two test files (below rule-of-three); pre-building `deadTransportError` (post-stale `send` is called rarely — micro-opt with no real impact).

_Reviewed by [`/code-police`](https://agency.srid.ca/) — Claude Code (model `claude-opus-4-8`)._

Issue #54, opened 2026-03-22
feat: keyboard shortcuts for terminal switching & help overlay


srid:
## Context

Currently, all terminal switching is done via sidebar clicks or the command palette (Cmd/Ctrl+K). There's no direct keyboard shortcut for switching between terminals, creating new ones, or cycling.

## Shortcuts

| Shortcut | Action |
|---|---|
| `Cmd/Ctrl + 1-9` | Switch to terminal N |
| `Cmd/Ctrl + T` | Create new terminal |
| `Cmd/Ctrl + Shift + ]` | Next terminal |
| `Cmd/Ctrl + Shift + [` | Previous terminal |
| `Cmd/Ctrl + /` | Toggle shortcuts help overlay |
| `Cmd/Ctrl + K` | Command palette (already exists, will be migrated) |

## Approach

Centralize all global keyboard shortcuts in a new `useShortcuts.ts` module (singleton pattern, matches `useTerminals.ts`). Define shortcuts as typed data in `keyboard.ts` with a `Keybind` type and `matchesKeybind()` helper (inspired by ~/code/mux's pattern). A single capture-phase `keydown` listener dispatches to handlers.

**Library choice**: None — keep using native KeyboardEvent with existing `makeEventListener` + `isPlatformModifier` + capture phase pattern. `@solid-primitives/keyboard`'s `createShortcut` doesn't support capture phase (needed to intercept before ghostty).

### New files
- `client/src/useShortcuts.ts` — Single keydown handler, dispatches based on keybind matching
- `client/src/ShortcutsHelp.tsx` — Modal overlay showing shortcut table

### Modified files
- `client/src/keyboard.ts` — Add `Keybind` type, `SHORTCUTS` map, `matchesKeybind()`, `formatKeybind()`
- `client/src/App.tsx` — Wire `useShortcuts`, mount `ShortcutsHelp`
- `client/src/CommandPalette.tsx` — Remove Cmd+K listener (migrated to `useShortcuts`)
- `client/src/Header.tsx` — Add shortcut hint

### Key design decisions

1. **Keybind type**: `{ key: string; mod?: boolean; shift?: boolean }` where `mod` = Cmd on macOS / Ctrl elsewhere
2. **Single capture-phase listener**: One handler on `window`, fires before ghostty's bubble-phase listener
3. **Cmd+K migration**: Move out of CommandPalette.tsx → useShortcuts.ts. CommandPalette becomes purely props-driven UI
[...]

PR #1234, opened 2026-06-08
Lift the WebSocket connection plumbing into @kolu/surface-app

srid:
**Follow-up to #1231.** The stale-tab handshake upstreamed the *gate*, but left the **connection assembly itself** hand-rolled in each consumer — the `pid`-echo, the `new PartySocket(...)`, and the server upgrade gate, byte-for-byte identical between kolu and [drishti](https://github.com/srid/drishti). This lifts that last duplicated layer into **three composable `@kolu/surface-app` primitives**.

The plan-of-record ships in this PR as an Atlas note (`status: implemented`); it records the design *and* the course-correction below.

### The three primitives

```
@kolu/surface-app/connect  (framework-free client transport)
  createProcessIdEcho()   — the shared pid echo (one per app; drishti shares it across N sockets)
  createSurfaceSocket()   — new PartySocket(echo'd URL thunk) + optional stale-close self-retire
  retireOnStaleClose()    — the self-retire listener, exported for direct testing

@kolu/surface-app/server
  gateStaleSocket(ws, url, liveProcessId, {onError, onReject})
      — error-handler-FIRST · rejectStaleProcess · close(4001) — in the one crash-free order
```

The lifecycle + clients **stay with the consumer** on purpose — kolu owns the lifecycle in `rpc.ts`, drishti via `<SurfaceAppProvider>`, and drishti runs *many* sockets sharing *one* echo. A god-factory bundling socket+clients+lifecycle would fit neither.

### The course-correction (facts over assumptions)

The first cut of the plan proposed a `createSurfaceConnection` god-factory + a `serveSurfaceWebSocket(adapter)` with **node/bun adapters**. Reading drishti's *actual* post-#1231 source broke all three premises:

| Assumption | Fact |
| --- | --- |
| drishti's server is Bun.serve → needs a bun adapter | drishti uses the **same `ws`/`@hono/node-server`/`@orpc/server/ws` stack** as kolu (it merely *runs* under Bun) — no adapter axis |
| one factory can own socket+clients+lifecycle | the two consumers own the lifecycle in **different places** |
| the echo is per-connection | drishti shares **one echo across N sockets** |

So the boundary moved *smaller* — three primitives matching what's actually duplicated, the same lesson #1231 learned when verified drishti facts overturned the first lens verdict. **drishti also gains the error-handler-before-gate fix it never had.**

### Validation

- `just check` green (typecheck + lint); `just test-unit` green — **30 surface-app tests** (echo threading, the self-retire footgun, the gate ordering), client 296, server 74.
- Pure refactor — no on-screen change; the #410 e2e reconnect test covers kolu's real socket.
- Docs synced: the `surface-connection` Atlas note, the `electricity` tracker (+ its stale-tab callout, which #1234 partly supersedes), `surface-app/README.md`, and the root `README.md` transport paragraph.

> Paired drishti PR adopts all three primitives — opened alongside this one. CI runs on both.

> Preview: [htmlpreview](https://htmlpreview.github.io/?https://github.com/juspay/kolu/blob/feat/surface-connection-plan/docs/atlas/dist/surface-connection.html) · or the Code tab on `docs/atlas/dist/surface-connection.html`.


srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `7126b5fa2a13`

### Round 1

**codex** — approved: `false`

The extraction is mostly coherent: kolu's caller keeps the same lifecycle ownership, the new socket helper matches the existing partysocket `WebSocket` constructor shape, and the server gate preserves the previous error-before-reject ordering for kolu. I could not run the package typecheck because `pnpm` is not available in this shell, but I did inspect the local partysocket declarations directly. I found one API observability issue and one docs defect.

Findings:
- `F1` · minor · open — `gateStaleSocket` defaults `onError` to a no-op listener. Because the listener is installed for accepted sockets too, any consumer that omits `onError` will silently swallow later WebSocket `error` events instead of logging transport failures. This is especially risky for a shared helper whose purpose is to remove per-consumer footguns. (packages/surface-app/src/server.ts:412)
- `F2` · nit · open — The `gateStaleSocket` example is malformed TypeScript: the kolu and drishti examples are interleaved into one `if` statement (`{  if (gateStaleSocket...`) and cannot be copied or read as a valid usage example. (docs/atlas/src/content/atlas/surface-connection.mdx:118)

**claude** — Addressed both CODEX findings. F1 (agree): gateStaleSocket's onError now defaults to a loud console.error instead of a silent no-op — the error listener is installed for accepted sockets too, so a no-op default would silently swallow standing transport errors. This also matches the loud-by-default pattern already used by buildInfoServer in the same file. Updated the doc comment and the corresponding test (now asserts the default logs loudly). F2 (agree): split the malformed interleaved gateStaleSocket docs snippet into two separate, valid TypeScript code blocks (one kolu, one drishti) and rebuilt the Atlas dist. Ran the formatter and the surface-app suite — all 21 tests pass.

- `F1` **fixed** — Agreed. The `error` listener is permanent on the socket (installed before the reject-window early-return but never removed), so on accepted sockets it doubles as the standing transport-error handler — a no-op default genuinely swallows later WebSocket error events for any consumer that omits onError. Changed packages/surface-app/src/server.ts:417-425 to default onError to a loud console.error ("gateStaleSocket: WebSocket error (pass `onError` to handle this).", err). This is consistent with buildInfoServer in the same file (server.ts:~292), which already does loud-console.error-by-default for the same swallowed-failure reason. Kept onError optional rather than required so the drishti call site (which only needs onReject) doesn't break, and documented that a consumer wanting silence passes an explicit no-op. Updated the doc comment (server.ts above the function) and rewrote the test at server.test.ts:294-307 to spy on console.error and assert the default listener logs loudly (not just 'does not throw').
- `F2` **fixed** — Agreed — the snippet at surface-connection.mdx:117-121 interleaved the kolu and drishti examples into one broken `if (gateStaleSocket(...){  if (gateStaleSocket(...` that is not valid TypeScript and cannot be copied. Split it into two separate, independently-valid ```ts blocks: one '// kolu — single handler' (onError + onReject) and one '// drishti — first line of the host-dispatched upgrade' (onReject only). Rebuilt the Atlas (just atlas::build) and staged docs/atlas/dist/surface-connection.html per the atlas-sync invariant.

commit: `714c15d371072da512207b9fbdc3a3be8852c921`

### Round 2

**codex** — approved: `true`

Claude addressed both prior findings in the current tree. `gateStaleSocket` now has a loud default error handler and matching regression coverage, and the Atlas usage example is split into valid copyable snippets. I did not identify any regression introduced by these fixes.

Findings:
- `F1` · minor · resolved — Previously, `gateStaleSocket` defaulted `onError` to a silent no-op even though the listener remains installed on accepted sockets, which would swallow later WebSocket transport errors for consumers that omitted `onError`. (packages/surface-app/src/server.ts:417)
- `F2` · nit · resolved — Previously, the `gateStaleSocket` documentation example interleaved the kolu and drishti snippets into malformed TypeScript. (docs/atlas/src/content/atlas/surface-connection.mdx:116)

_codex on the rebuttal:_ No disputes. F1: verified the default handler now logs via `console.error` and the test asserts that behavior. F2: verified the malformed interleaved snippet is gone and replaced by two valid snippets.

srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `7126b5fa2a13`

### Applied (1)
- `hickey-1` Five overlapping structural projections of the same partysocket — commit `61024cef1`

### Agreed — no change (10)
- `lowy-1` Stale-close code is single-sourced in two diverging receptacle parameters (packages/surface-app/src/connect.ts:108-111 and packages/surface-app/src/solid/index.ts:130)
- `lowy-2` retireOnStaleClose and lifecycle.onStaleRestart are two decode sites for the same stale-close → retire action (packages/surface-app/src/connect.ts:85-92 and packages/surface-app/src/solid/index.ts:220-237)
- `lowy-3` Consumer still exposes the partysocket-to-DOM-WebSocket wire past the receptacle (packages/client/src/wire.ts:68)
- `lowy-4` Package coherence: /connect ships three exports for a consumer to wire, but the union is one primitive (packages/surface-app/src/connect.ts:51,85,131 (exports createProcessIdEcho, retireOnStaleClose, createSurfaceSocket))
- `lowy-5` gateStaleSocket correctly encapsulates the error-handler-first ordering as the volatile axis (packages/surface-app/src/server.ts:408-434)
- `lowy-6` createSurfaceSocket option `retireOnStaleClose` collides in name with the exported function it gates (packages/surface-app/src/connect.ts:108 (option) vs 85 (function))
- `hickey-2` createSurfaceSocket returns a fragment the consumer must re-thread by hand (packages/surface-app/src/connect.ts:114-147; packages/client/src/wire.ts:42-52)
- `hickey-3` retireOnStaleClose default code re-sourced from STALE_PROCESS_CLOSE_CODE at a third site (packages/surface-app/src/connect.ts:108-111,144)
- `hickey-4` createSurfaceSocket complects URL-string-vs-thunk normalization with socket construction (packages/surface-app/src/connect.ts:131-142)
- `hickey-5` Three subpath entries now split the client transport across /connect, /lifecycle, /solid (packages/surface-app/package.json:14-16; packages/surface-app/src/connect.ts (whole))

srid:

Rules → fact-check → elegance, on the tree after codex-debate + lens-debate settled it.

**Found & fixed (1):**

| Rule | Location | Fix |
| --- | --- | --- |
| `no-untyped-escape-hatches` | `surface-app/src/server.ts` — `gateStaleSocket` | `opts.onReject?.(claimedPid as string)` → guarded with `claimedPid !== null &&` so TypeScript narrows the type structurally. Behavior identical (`rejectStaleProcess` already returns `false` for `null`), but the untyped cast is gone. `fix(police):` `dbe1900e`. |

**Checked clean (no action):** the full rule checklist (26 rules) — `dry-rule-of-three` (this PR *removes* duplication FROM two consumers, doesn't add it), `no-silent-error-swallowing` (`gateStaleSocket` defaults to a loud `console.error`, matching `buildInfoServer`), `no-unbounded-growth` (the echo holds one `string | null`; `retireOnStaleClose` installs one listener), `invalid-states-unrepresentable`, `no-thin-wrapper-functions` (`retireOnStaleClose` carries a real code-match predicate), `comment-the-non-obvious` — all pass.

**Skipped (1, false positive):** inlining `connect.ts`'s `resolveBase` intermediate — it's a legitimate named normalization step (string-or-thunk → thunk), not excess indirection.

srid:
## Evidence

This is a **pure behavior-preserving refactor** — it lifts the connection plumbing into `@kolu/surface-app` without changing any externally-visible behavior. The stale-tab handshake (restart → reload overlay) is byte-identical to #1231, which was already captured with video. So the proof here is *"the behavior survived the refactor,"* not a new visual:

**1. The reconnect path still works end-to-end — green e2e in CI.**
`ci::e2e` passed on both lanes on `dbe1900e` (x86_64-linux 17m47s · aarch64-darwin 35m15s). The suite drives the real WebSocket reconnect path that this PR rewired (the #410 reconnect regression scenario among them), so a green e2e is the round-trip proof the handshake still behaves identically through the new `createSurfaceSocket` + `gateStaleSocket`.

**2. The three primitives' behavior is locked by new unit tests** (`ci::unit`, green) — 30 surface-app tests:
- `createProcessIdEcho` — `appendTo` is a no-op until an id is observed, then `?pid=`/`&pid=` respecting an existing query string, url-encoded, latest-id-wins, detachable `remember`.
- `retireOnStaleClose` — retires on the stale-close code (stop reconnect + a throwing `send` that the shared retry fence classifies non-retriable), ignores ordinary transient codes.
- `gateStaleSocket` — lets a matching/absent pid through; on a stale tab closes `4001` + returns `true`; installs the `error` listener **before** the reject path (the crash-free ordering), loud-by-default `onError`.

**3. The ③ proof is live — [drishti #54](https://github.com/srid/drishti/pull/54) consumes the identical API and is green** (18/18 CI, bun test 124/124). It *deleted* its hand-rolled echo/socket/gate and now inherits the error-handler-before-gate fix it never had — the strongest evidence the boundary is real, not aspirational.

> A re-capture of the restart→overlay flow would only re-prove #1231's already-evidenced behavior; the green e2e + the unit suite + the second consumer are the behavior-preservation proof for a pure refactor.

PR #1235, opened 2026-06-09
Detect half-open WebSockets with a heartbeat (no more frozen-until-reload terminals)

srid:
**A terminal would silently freeze mid-session — output stops, nothing recovers it but a full page reload — and it had nothing to do with the recent stale-tab work.** The cause is a *silently half-open* WebSocket: the TCP dies with no FIN/RST (a laptop sleeps, Wi-Fi roams, a NAT/proxy evicts the idle connection), so neither end ever sees a `close` or `error`. partysocket ships no keepalive, so it never reconnects; the oRPC `terminal.attach` stream never throws, so `STREAM_RETRY` never re-fires; the socket just sits `OPEN` while the server keeps producing output the client can no longer hear. This is a pre-existing latent bug — the symptom predates #1231/#1234 by weeks — that only ever recovered on a manual reload.

The fix is a **heartbeat on both ends**, so a dead-but-open socket becomes a real `close` and the transport's *existing* recovery machinery takes over. Nothing new is invented for reconnect/re-subscribe — the heartbeat just makes the drop observable.

### The two halves

```
client tab                         kolu server
──────────                         ───────────
createHeartbeat (rpc.ts)           startWsHeartbeat (wsHeartbeat.ts)
  every 15s: identity.info probe     every 30s: ping accepted sockets
  no answer in 10s ─▶ ws.reconnect    no pong since last sweep ─▶ terminate
        │                                    │
        ▼                                    ▼
  close 1000 ─▶ lifecycle             frees the zombie socket +
  "disconnected" ─▶ fresh open        its per-terminal subscriptions
  ─▶ oRPC re-subscribes every stream
```

- **Client watchdog** — `createHeartbeat` (`@kolu/surface-app/connect`), wired once in `rpc.ts`, re-uses the `identity.info` probe as a keep-alive. An unanswered probe means half-open, so it forces `ws.reconnect()` with **close code 1000 — *not* the stale-tab 4001**, so the `retireSocket` path is left entirely alone. _A probe **rejection** still counts as alive (a completed round-trip — the server answered, even with an error); only a **timeout** is half-open._ Ticks never overlap.
- **Server reaper** — `startWsHeartbeat` pings accepted sockets and `terminate()`s any that stop ponging, reclaiming the server-side zombie a half-open client would otherwise leak. _Liveness lives in a `WeakSet` the `pong` handler re-adds to — not monkey-patched onto the socket._

> Stale tabs are closed by the gate *before* upgrade and never enrol in either heartbeat, so **#1231's stale-tab protection is untouched**. The client half is what actually un-freezes a stuck tab; the server half frees the matching server resources.

### Scope note

partysocket 1.1.16 was verified to ship **no** client-side ping option (`minUptime`/`connectionTimeout`/`maxRetries`/`maxEnqueuedMessages`/`startClosed` only), so the client liveness had to be an app-level probe-on-an-interval — the "no-op procedure on an interval" `@kolu/surface`'s peer-server note already anticipated — rather than a transport flag. A server-only heartbeat was *not* sufficient: it frees the server but can't wake a half-open *client*, which is the symptom.

### Validation

- `just check` green; **12 new unit tests** (6 client: probe-keeps-alive, timeout→reconnect, rejection-is-alive, skip-when-not-OPEN, no-overlap, dispose; 6 server: sweep ping/terminate/skip-non-OPEN, cadence, pong-keeps-alive, stop).
- The genuine half-open path can't be reproduced reliably via CDP (`Network.emulateNetworkConditions` doesn't half-close a live WebSocket), so live recovery is captured as **PR evidence** rather than a flaky process/network e2e — posted under `## Evidence`.
- README transport-resilience prose synced.

Follow-up to #1231 / #1234 (same transport seam, different failure mode).


srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 round(s) · codex reviewed at `xhigh` reasoning effort · base `61a32c973745`

### Round 1

**codex** — approved: `false`

The heartbeat design is largely coherent and the touched packages pass direct no-emit TypeScript checks. I found one remaining issue in the exported client helper: synchronous probe failures are silently swallowed and classified as liveness, which can hide a miswired probe or broken client before any round trip is made.

Findings:
- `F1` · minor · open — `Promise.resolve().then(opts.probe)` converts a synchronous `probe` throw into a rejected promise, and the rejection handler at lines 253-255 silently calls `settled(false)`. An async oRPC rejection can reasonably count as a completed round trip, but a synchronous throw means no probe was sent at all. This masks local misconfiguration or a broken client object and leaves the heartbeat inert with no diagnostic. (packages/surface-app/src/connect.ts:251)

**claude** — Addressed CODEX's single finding (F1) on the heartbeat client helper. I agree with the core concern: a SYNCHRONOUS probe throw means no round-trip happened (a miswired/broken probe), so silently routing it through `Promise.resolve().then(probe)` and classifying it as liveness was wrong — it hid the fault and left the heartbeat inert with no diagnostic. Fixed by calling `opts.probe()` inside an explicit try/catch in `tick`: a sync throw is now reported via a new `onProbeError` reporter (defaults to a distinct `console.warn`) and settles WITHOUT reconnecting (a local probe fault is not a transport problem, so churning the socket would be wrong), while the genuine async-rejection case still counts as alive. Added the requested regression test. surface-app's 19 connect tests pass, the package typechecks clean (tsc --noEmit, exit 0), and biome formatting is clean.

- `F1` **fixed** — Agreed with the core point and fixed it in packages/surface-app/src/connect.ts. The `tick` body (was lines 247-256) now calls `opts.probe()` in an explicit try/catch: a synchronous throw is surfaced via a new `onProbeError` option (defaulting to `warnProbeThrew`, a console.warn distinct from the timeout's `warnStale`) and then `settled(false)` runs so the inFlight flag clears and the next tick can retry — but `ws.reconnect()` is NOT called, because a miswired probe is a local fault the socket can't fix (reconnecting would only churn). The genuine async-rejection path (`Promise.resolve(probing).then(ok, ok)`) is unchanged and still counts as alive, per the documented contract that a server error reply == a completed round trip. Updated the `probe` and added `onProbeError` doc comments on HeartbeatOptions. Added a regression test in connect.test.ts ('surfaces a SYNCHRONOUS probe throw…') asserting onProbeError fires once, ws.reconnect/onStale never fire, and the probe is retried on the next tick (not wedged inFlight). One nuance vs CODEX's wording: CODEX suggested keeping behavior such that the sync case 'settles without reconnecting' which is exactly what I did; I did not add any new reconnect-on-sync-throw behavior since the socket is healthy in that scenario.

commit: `93dd6c02935f5604ea62a05095172a324cd3dc49`

### Round 2

**codex** — approved: `false`

F1 is addressed: synchronous probe throws are now distinguished from async probe rejections, reported via `onProbeError`, and settled without reconnecting. The latest fix introduced one new callback-isolation bug in that sync-throw path.

Findings:
- `F1` · minor · resolved — Previously, synchronous probe throws were normalized into rejected promises and silently treated as liveness, hiding local probe misconfiguration. (packages/surface-app/src/connect.ts:270)
- `F2` · minor · open — New regression: `onProbeError` is invoked without a guard. If a caller-provided reporter throws, `settled(false)` is skipped, the probe timeout remains armed, and the same local probe fault can later be misclassified as a stale transport and call `ws.reconnect()`. It also emits an uncaught timer error. (packages/surface-app/src/connect.ts:273)

_codex on the rebuttal:_ F1: accepted as fixed. I agree with CLAUDE's distinction that a synchronous throw is a local probe fault, not a transport failure, so settling without reconnecting is the right behavior. The explicit `try/catch` plus `onProbeError` resolves the original silent-swallowing issue. New F2 is limited to this round's fix: the newly added reporter callback can throw before `settled(false)` runs.

**claude** — Follow-up round on the heartbeat watchdog in @kolu/surface-app. F1 was already settled as fixed in the prior round (sync probe throws are distinguished from async rejections, reported via onProbeError, settled without reconnecting). This round addressed F2, the new callback-isolation regression CODEX flagged in that fix path.

- `F1` **fixed** — Already resolved last round and CODEX accepted it as resolved (status: resolved, responseToRebuttal accepts the fix). Nothing further to do — synchronous probe throws are caught in an explicit try/catch, surfaced via onProbeError, and settle without reconnecting; genuine async rejections still count as a completed round trip. No change this round.
- `F2` **fixed** — Agreed — valid regression and consistent with the existing onStale guard. At packages/surface-app/src/connect.ts:273 the onProbeError reporter was invoked unguarded, so a throwing reporter would skip settled(false), leaving inFlight true and probeTimer armed; the armed timer would then fire settled(true) and misclassify a local probe fault as a stale transport (spurious ws.reconnect + uncaught timer error). Fixed by wrapping (opts.onProbeError ?? warnProbeThrew)(error) in its own try/catch (mirroring the onStale guard at lines 250-254) so settled(false) at line ~282 always runs. Added the requested regression test in connect.test.ts ('settles a SYNCHRONOUS probe throw even when the onProbeError reporter throws…') asserting that after the timeout window ws.reconnect and onStale never fire and the probe is retried on the next tick. All 20 connect tests pass, tsc --noEmit exits 0, biome format clean.

commit: `36b5fa3461f16fc5b420762391203b6bd2001bf8`

### Round 3

**codex** — approved: `true`

Both findings from this review session are now resolved. The current tree explicitly reports synchronous probe throws without classifying them as transport liveness, and the follow-up change guards the new `onProbeError` reporter so a throwing callback cannot leave the heartbeat timeout armed. I verified the relevant diff and `packages/surface-app` no-emit typecheck passes.

Findings:
- `F1` · minor · resolved — Synchronous `probe` throws were previously normalized into rejected promises and silently treated as liveness, hiding local probe misconfiguration even though no round trip happened. (packages/surface-app/src/connect.ts:270)
- `F2` · minor · resolved — The newly added `onProbeError` callback was previously unguarded, so a throwing reporter could skip `settled(false)`, leave the timeout armed, and later trigger a spurious reconnect. (packages/surface-app/src/connect.ts:277)

_codex on the rebuttal:_ F1: already accepted as fixed, and the current tree still preserves that fix. F2: accepted as fixed. The new guarded reporter block at `connect.ts:277-282` addresses the exact failure mode: callback exceptions no longer prevent clearing `inFlight` and `probeTimer`, so the sync probe fault is not later misclassified as stale transport.

srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `61a32c973745`

### Applied (3)
- `lowy-1` Server heartbeat reaper placed in kolu, not in @kolu/surface-app/server beside its sibling gateStaleSocket — commit `eb0e4e1b6`
- `hickey-4` register() attaches a fresh pong listener per call with no removal path — commit `568c0e25a`
- `hickey-5` Heartbeat interval/timeout defaults defined in two packages with the same role — commit `a4c46f823`

### Agreed — no change (5)
- `lowy-2` Client and server heartbeat cadences are independent magic numbers with no shared liveness contract (packages/surface-app/src/connect.ts:173-174 and packages/server/src/wsHeartbeat.ts:114)
- `lowy-3` rpc.ts heartbeat singleton drops the dispose handle while createHeartbeat exposes one (packages/client/src/rpc/rpc.ts:86)
- `hickey-1` Client liveness probe loop fragmented from the lifecycle's probe loop (packages/client/src/rpc/rpc.ts:86 (and packages/surface-app/src/solid/index.ts onOpen probe + connect.ts createHeartbeat))
- `hickey-2` createHeartbeat is wired without dispose() despite the module exposing one (packages/client/src/rpc/rpc.ts:86)
- `hickey-3` Server heartbeat liveness uses one WeakSet for two distinct states (packages/server/src/wsHeartbeat.ts:122-135)

srid:

Three passes (rules → fact-check → elegance) on the gauntlet-settled tree.

| Pass | Found | Outcome |
|------|-------|---------|
| Rules | 1 | fixed |
| Fact-check | 1 (same) | fixed |
| Elegance | 0 | clean |

**Finding (rules `no-silent-error-swallowing` / `callback-fanout-guarded-at-funnel`, confirmed by fact-check) — `gateStaleSocket` (`packages/surface-app/src/server.ts`):** the observational `onReject` callback fired *before* the critical `ws.close()`. A throwing reporter would skip the close, leaving the stale tab **connected** — replaying dead subscriptions, never reaching the reload overlay. Pre-existing from #1231, surfaced here because the heartbeat reaper now lives beside it.

**Fix** (`f7d66d13`): close first, observe second — mirroring the same `reconnect()`-before-`onStale` ordering this PR's `createHeartbeat` already adopted.

Everything else came back clean: the two empty `catch {}` blocks in `createHeartbeat` are intentional and commented (a throwing status reporter must not unwind the already-completed recovery); the `WeakSet` liveness has no unbounded growth; the differing `15s`/`30s` defaults are intentional, not a DRY violation; log levels are correct.

srid:
## Evidence



▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=half-open-heartbeat-recovery.mp4

**What the clip shows:** A kolu terminal runs a streaming clock (`while true; do date +%T.%N; sleep 0.3s; done`). An `iptables DROP` rule is applied to silently drop all packets on the WebSocket port — the output **freezes** with no "reconnecting" overlay (the socket appears open to the client, this is the half-open bug). The heartbeat fires, its probe times out, and `ws.reconnect()` is called — confirmed by the server log showing a clean code-1000 disconnect followed immediately by a new WebSocket connection. The terminal **resumes streaming on its own** without any page reload.

**Timing note:** The heartbeat was lowered to `intervalMs = 3 000 ms` / `timeoutMs = 2 000 ms` for this capture so recovery happens in ~5 s on camera. The shipped defaults are `15 000 ms` / `10 000 ms` (worst-case auto-recovery ~25 s).

**Server log confirmation:**
```
01:58:19 INFO  connected          ws:1 total:1
01:58:19 INFO  created            terminal e2eeafbb…
01:58:29 INFO  disconnected       ws:1 code:1000  ← heartbeat-forced reconnect
01:58:30 INFO  connected          ws:2 total:1    ← auto-reconnect, no reload
```

PR #918, opened 2026-05-17
Preview .html/.svg/.pdf artifacts inline in the Code tab

srid:
Kolu's Code tab now renders agent-generated `.html`, `.svg`, and `.pdf` files inline in a sandboxed iframe instead of as syntax-highlighted source. **Selecting one of those files in the existing browse-mode tree swaps to the iframe**; text files still render through Pierre's `FileView` exactly as before.

Sets up #881 phase 2 (HTML annotator via lavish-axi's `artifact-sdk.js`) — this PR establishes the iframe rendering surface that phase will inject the annotation SDK into. _Comments and selection are deliberately not part of this PR; it's the render primitive only._

### How the variant is picked

```text
fsReadFile(repo, file) ──▶ isIframePreviewable(ext)?
                                    │
              ┌─────────────────────┴────────────────────┐
              ▼                                          ▼
        kind: "text"                              kind: "binary"
        { content, truncated }                    { url: /api/terminals/<id>/file/<path>?v=<mtime> }
              │                                          │
              ▼                                          ▼
       BrowseFileView                            BrowsePreviewView
       (Pierre's FileView)                       (<iframe sandbox="allow-scripts">)
```

The `kind` discriminator lives on the wire (`FsReadFileOutputSchema`). The client never sees a file-extension list — it dispatches on `kind` in `BrowseFileDispatcher`. _Server's `fsReadFile` re-yields the binary variant with a bumped `?v=<mtime>` on every save, so the iframe live-reloads through the same subscription path as text — no parallel notification channel._

### Security

- **Sandbox**: `sandbox="allow-scripts"` only — no `allow-same-origin`. Page JS runs in an opaque origin and can't touch Kolu cookies/localStorage. _Cross-origin `fetch()` from inside is blocked; acceptable for static artifacts._
- **Path traversal**: two-stage guard in `iframePreviewRoute.ts`. Stage 1 URL-decodes-then-splits and rejects `..`, `.`, empty, and absolute segments _before_ `path.join` (defends against `%2e%2e`, `foo%2f..%2fpasswd`, double-slash collapsing). Stage 2 calls the existing `resolveUnder` from `kolu-git` for the canonical `path.relative` re-check.
- **Content-Type**: explicit per extension; `X-Content-Type-Options: nosniff` blocks MIME sniffing.

### Refinements during review

Seven follow-up commits from `/hickey`, `/lowy`, and `/code-police` reshaped the boundaries:

| Finding                                                                                          | Fix                                                                                                                                   |
[...]

PR #1239, opened 2026-06-09
Stop the EVENT_ITERATOR_VALIDATION_FAILED error on upgrade

### LINKED ISSUE #1237: Missing migration for `rightPanel.codeTabTreeSize` causes `EVENT_ITERATOR_VALIDATION_FAILED` on launch
### Description

Kolu logs an `EVENT_ITERATOR_VALIDATION_FAILED` error on the `surface.kolu.preferences.get` RPC call immediately after launch. The app continues to work normally (the client presumably falls back to defaults), but the error is noisy and alarming.

```
ERROR: Event iterator validation failed
  rpc: { method: "surface.kolu.preferences.get" }
  code: "EVENT_ITERATOR_VALIDATION_FAILED"
  status: 500
```

### Root Cause

`RightPanelPrefsSchema` in `packages/common/src/surface.ts` requires a `codeTabTreeSize: z.number()` field:

```ts
export const RightPanelPrefsSchema = z.object({
  collapsed: z.boolean(),
  size: z.number(),
  codeTabTreeSize: z.number(),
});
```

However, no migration in `packages/server/src/state.ts` adds this field to existing `config.json` files. Users upgrading from a version before `codeTabTreeSize` was introduced have a `rightPanel` object that only contains `collapsed` and `size`. All existing migrations run (the file shows version `1.23.0`), but none backfill `codeTabTreeSize`.

When `preferences.get` yields the stored preferences, `@orpc/contract`'s `eventIterator` validates the value against `PreferencesSchema`. Zod rejects it because `rightPanel.codeTabTreeSize` is missing, throwing `EVENT_ITERATOR_VALIDATION_FAILED`.

### Environment

- **OS:** Linux (NixOS)
- **Install method:** Nix flake (`kolu-stamped`)
- **@orpc/contract:** 1.13.13
- **Config path:** `~/.config/kolu/config.json` (via `KOLU_STATE_DIR`)

### Reproduction

1. Have an existing kolu install with a `config.json` whose `preferences.rightPanel` was written before `codeTabTreeSize` was added (i.e. it only has `collapsed` and `size`).
2. Upgrade to the current version.
3. Launch kolu.
4. Observe the `EVENT_ITERATOR_VALIDATION_FAILED` error logged at startup. The app otherwise works fine.

Example of the offending on-disk shape:

```json
"rightPanel": {
  "collapsed": true,
  "size": 0.25
}
```

### Workaround

Either delete `~/.config/kolu/config.json` to reset to defaults, or manually add the missing field:

```json
"rightPanel": {
  "collapsed": true,
  "size": 0.25,
  "codeTabTreeSize": 0.35
}
```

### Proposed Fix

Add a migration in `packages/server/src/state.ts` and bump `SCHEMA_VERSION`:

```ts
"1.24.0": (store: Conf<PersistedState>) => {
  const current = store.get("preferences");
  const rp = current.rightPanel as Record<string, unknown>;
  if (rp.codeTabTreeSize === undefined) {
    store.set("preferences", {
      ...current,
      rightPanel: {
        ...rp,
        codeTabTreeSize: DEFAULT_PREFERENCES.rightPanel.codeTabTreeSize,
      },
    });
  }
},
```

srid:
**Upgrading kolu from a build that predates the Code-tab tree split logged a scary 500 at every launch.** The app fell back to defaults and worked fine, but the error was noisy and alarming — and there was no migration to make it stop.

```
ERROR: Event iterator validation failed
  rpc: { method: "surface.kolu.preferences.get" }
  code: "EVENT_ITERATOR_VALIDATION_FAILED"
  status: 500
```

### Root cause

`codeTabTreeSize` was added to `RightPanelPrefsSchema` in #918 **without a migration**. Config files written before then carry a `rightPanel` with only `collapsed`/`size`. When `preferences.get` yields the stored prefs, `@orpc/contract`'s `eventIterator` validates them against `PreferencesSchema` on the wire, and Zod rejects the missing field.

```
on-disk prefs ─▶ preferences.get ─▶ eventIterator validate ─▶ ✗ Zod (codeTabTreeSize missing)
```

### Fix

Migration `1.24.0` in `packages/server/src/state.ts` backfills `rightPanel.codeTabTreeSize` from `DEFAULT_PREFERENCES`, with a `SCHEMA_VERSION` bump. It only writes when the field is absent, so it's a no-op on already-migrated data. This follows the project's [state convention](.claude/rules/state.md) — any persisted-schema change ⇒ migration + version bump.

Closes #1237.

### Try it locally

```sh
nix run github:juspay/kolu/fix/codeTabTreeSize-migration-1237
```


ic4-y:
### Verification

Built from the PR branch (`fix/codeTabTreeSize-migration-1237`, commit `22e4728`) and tested against a corrupted config:

| State | Result |
|---|---|
| Config missing `codeTabTreeSize`, migration version `1.23.0` (pre-patch) | **ERROR** logged: `Persisted state does not match schema (preferences.rightPanel.codeTabTreeSize: Invalid input: expected number, received undefined)` |
| Same broken config, PR build | **No error.** `codeTabTreeSize: 0.35` backfilled from defaults, version bumped to `1.24.0` |

**One note:** the stored migration version in the config was already `1.24.0` from a previous run of the unpatched build. The `conf` library only runs migrations where `storedVersion < handlerKey <= SCHEMA_VERSION`, so the `1.24.0` migration didn't fire until I manually downgraded the version to `1.23.0`. In practice users upgrading from builds *before* `1.24.0` was written will have the lower version and the migration fires naturally — but users who already ran the broken build may need to either manually edit the version back or delete the config (which the existing error message already suggests).

Fix confirmed working. Config restored after testing.

PR #1246, opened 2026-06-09
Stop fresh repos from spamming a base-branch error every poll tick

### LINKED ISSUE #1244: `gitStatus.get` throws on fresh repos with no remote and no commits
### Steps to reproduce

1. `mkdir new-repo && cd new-repo && git init`
2. Open kolu with a terminal pointed at that directory

### Expected

Git status renders without errors (empty status, no diff, or graceful absence).

### Actual

An ERROR is logged on every poll cycle:

```
ERROR: No base branch found — origin/master doesn't exist. Run: git fetch origin && git remote set-head origin --auto
```

with `code: "PRECONDITION_FAILED"`.

### Root cause

`detectDefaultBranch()` in `packages/integrations/git/src/worktree.ts:20-35` has a three-step chain:

1. `git symbolic-ref refs/remotes/origin/HEAD`
2. `git rev-parse --verify origin/main`
3. Hardcoded `return "master"`

For a freshly initialized repo with no remote and no commits, steps 1 and 2 both fail, so the function unconditionally returns `"master"` without verifying it exists. `resolveBase()` in `packages/integrations/git/src/review.ts:51-68` then constructs `origin/master` and tries `git rev-parse --verify origin/master^{commit}`, which fails and produces the `BASE_BRANCH_NOT_FOUND` error.

### Additional issues

- **No self-healing in the `getStatus` path.** The error message suggests running `git fetch origin && git remote set-head origin --auto`, but those commands are only run in `worktreeCreate` — not in the `resolveBase` path that actually needs them. The suggestion is actionable for a human but the code never runs them.
- **The final fallback is unconditional.** Returning `"master"` without verification means the downstream `rev-parse` check is guaranteed to fail for repos that don't have `origin/master`.

### Impact

Repos that have just been initialized, or repos with non-standard default branch names (e.g., `develop`, `trunk`) and no `origin/HEAD` set, hit this error repeatedly on every poll tick.

### Suggested fix

Either:
- Degrade gracefully when no remote refs exist at all (return empty git status / skip branch diff), or
- Run `git fetch origin && git remote set-head origin --auto` in `resolveBase` before detection (matching what `worktreeCreate` already does), or
- Both: attempt the fetch to self-heal, and if it still fails, degrade gracefully.

srid:
**Opening a terminal in a freshly `git init`'d repository — no remote, no commits — used to log a `BASE_BRANCH_NOT_FOUND` / `PRECONDITION_FAILED` error on every repo-change tick.** The Code tab's Branch view diffs the working tree against `origin/<default>`; `detectDefaultBranch` unconditionally falls back to `"master"`, so `resolveBase` then fails to verify `origin/master^{commit}` and the surface's `onStreamReadError` logs the failure at ERROR each time the watcher fires.

The fix distinguishes *"no base because there's nothing to compare against"* from *"no base, but you can fix it"*:

| Repo state | Before | After |
| --- | --- | --- |
| No `origin` remote (fresh `git init`) | ERROR every tick | Empty branch diff, silent |
| `origin` exists, default branch not fetched | ERROR every tick | Actionable `run git fetch` error (unchanged) |
| Normal repo with fetched base | Works | Works |

`resolveBase` now returns `ok(null)` when the repo has no `origin` remote at all — branch mode is meaningless there, not broken — and `getStatus`/`getDiff` emit an empty result instead of erroring. The `BASE_BRANCH_NOT_FOUND` error is *reserved for the genuinely-fixable case*, so explicit Branch mode still prompts the user to fetch when a remote actually exists.

Three reproducing tests in `review.test.ts` lock in all three rows of the table above.

Closes #1244

### Try it locally

```sh
nix run github:juspay/kolu/fix/git-status-fresh-repo
```


srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `5d47800da8c9`

### Round 1

**codex** — approved: `false`

Reviewed the diff against 5d47800da8c9a0afa560891c994b850916e517d8 and confirmed there are no untracked files. The backend change covers the fresh/no-origin repo status path, but the new "branch mode has no base" state is still presented and documented as an ordinary clean branch diff. I did not run tests because the review instructions were read-only and the relevant tests create temporary repositories; `git diff --check` is clean.

Findings:
- `F1` · major · open — Branch mode now returns an empty status with `base: null` when there is no branch base, but the empty tree still renders the generic `No changes vs base` copy from `EMPTY_STATE`. In a fresh or remote-less repo, that is a false clean-state signal: there is no base to compare against, so users can read Branch mode as saying the branch has no changes when the comparison is actually unavailable. (packages/client/src/right-panel/CodeTab.tsx:765)
- `F2` · minor · open — The documented contract is stale after this change. `GitStatusOutput.base` is still documented as `Null in local mode; resolved base ref in branch mode`, and the README still says the explicit Branch mode surfaces an actionable message whenever `origin/<default>` is missing. The new behavior deliberately returns `base: null` in Branch mode for no-origin repos and shows an empty branch view instead. (packages/integrations/git/src/schemas.ts:98, README.md:79)

**claude** — Both CODEX findings were valid; I fixed both. F1: Branch mode with no resolvable base now renders a distinct "No branch base to compare" empty-state instead of the false-clean "No changes vs base", and the Branch chip's picker hint mirrors it — guarded by a new branchHasNoBase() that is only true while Branch is the *active* view (status only subscribes for the current view, so reading status()?.base in browse/local view would falsely report no-base). F2: Updated the GitStatusOutput.base schema comment and the README branch-layer prose to distinguish the remote-less/no-base case (degrades to empty Branch view) from the actionable origin-exists-but-unfetched case (run git fetch). Formatter ran clean and the client typechecks.

- `F1` **fixed** — Agreed. In CodeTab.tsx the branch empty-state (~line 765) now returns "No branch base to compare" when `m === "branch" && status()?.base === null`, otherwise the normal EMPTY_STATE copy. Also fixed the picker hint CODEX flagged at line 633: added `branchHasNoBase()` (view()==="branch" && status()?.base===null) and the branch chip hint now reads "No branch base to compare" in that state. Guarded to the active view on purpose — `status` only subscribes for the current diffMode, so in browse/local view status()?.base is null for local mode and a naked check would falsely claim Branch has no base; the empty-state path is already implicitly branch-only since `m==="branch"` comes from the active view.
- `F2` **fixed** — Agreed, docs were stale. schemas.ts:98 comment now states base is null in local mode and in branch mode is the resolved base ref OR null for a remote-less (no-origin) repo that degrades to an empty diff (#1244). README.md:79 now distinguishes the unfetched-origin case (falls back / actionable "run git fetch") from the remote-less no-origin case (no base, degrades to empty Branch view instead of erroring on every change tick).

commit: `aba38b011054afdc8503b27f10f1d9b761f87145`

### Round 2

**codex** — approved: `true`

CLAUDE's follow-up addresses both existing findings. The Branch no-base state now has distinct UI copy in the empty tree and active Branch picker hint, and the schema/README comments now describe the new `base: null` branch-mode semantics. I found no regression introduced by this round's fixes.

Findings:
- `F1` · major · resolved — Branch mode with no resolvable base previously rendered the generic clean-state copy, which falsely implied a successful comparison with no changes. (packages/client/src/right-panel/CodeTab.tsx:607)
- `F2` · minor · resolved — The documented `GitStatusOutput.base` contract and README branch-mode prose were stale after branch mode started returning `base: null` for remote-less repos. (packages/integrations/git/src/schemas.ts:98, README.md:79)

_codex on the rebuttal:_ CLAUDE did not dispute either finding. I verified F1's UI fix at CodeTab.tsx:607-613, 641-645, and 778-786; the guard being active-view-only is appropriate because `status` is scoped to the current diff view. I verified F2's documentation fixes at schemas.ts:98-101 and README.md:79.

srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `5d47800da8c9`

### Applied (4)
- `lowy-1` The "no resolvable base" predicate is stated three different ways across the boundary — commit `529ee7892`
- `lowy-2` The "No branch base to compare" copy is duplicated as a literal in two render sites — commit `581d4122e`
- `hickey-1` "branch mode has no base" predicate duplicated, not reused — (uncommitted)
- `hickey-2` "No branch base to compare" label literal duplicated across two surfaces — (uncommitted)

### Agreed — no change (2)
- `lowy-3` hasOriginRemote uses substring-prone `.includes` on a line-split where a tighter membership check fits the receptacle (packages/integrations/git/src/review.ts:50-53)
- `hickey-3` Wire-schema `base: null` overloads two distinct states (local-N/A vs branch-no-base) (packages/integrations/git/src/schemas.ts:98)

srid:

Ran the three passes — **rules → fact-check → elegance** — on the post-lens tree. **No fixes applied; the code is clean.**

| Pass | Result |
| --- | --- |
| **Rules** | 1 observation (see below) — not actioned |
| **Fact-check** | ✅ Clean |
| **Elegance** | ✅ Clean (reuse / simplification / efficiency / altitude all right) |

### Fact-check highlights (verified correct)
- `hasOriginRemote`: a no-remote repo's `git remote` → `""` → `[""]` → `includes("origin") === false`. Correct.
- All three states (no-origin → `ok(null)`, origin-unfetched → `BASE_BRANCH_NOT_FOUND`, origin-fetched → real base) route correctly through `resolveBase`.
- `branchHasNoBase()` can't false-positive during load: `undefined?.base` is `undefined`, not `null`; and the `view() === "branch"` guard fires first, so local mode never trips it.
- `hasOriginRemote` is on the **error path only** — the common fetched-origin case pays no extra git call.

### The one observation — already adjudicated, intentionally kept
**`invalid-states-unrepresentable`** — `GitStatusOutput.base: GitBaseRef | null` now carries two `null` meanings (local-mode N/A vs. branch-mode no-origin), disambiguated by the caller's `mode`/`view()` context rather than by the type.

This is **the same finding `hickey-3` raised in the lens debate**, where lowy ⇄ hickey **agreed to drop it**: the discriminated-outcome refactor touches the wire schema and every consumer (high cost, low leverage), the per-mode subscription model means every consumer already knows its mode contextually, and the schema doc comment was tightened in `aba38b0` to state the disambiguation rule at the type. The structural ideal (a `{ kind: "none" }` discriminant) is a deliberately-deferred separate change, out of scope for this bug fix. **No action — decision stands.**

srid:
## Evidence

Code-tab Branch view in a fresh `git init` repo (no remote, no commits) — previously the server logged a `BASE_BRANCH_NOT_FOUND` ERROR on every poll tick; now the UI degrades gracefully to this empty-state instead.



The Code tab is in **Branch** mode (`Git: Branch` chip) in a repo created with `git init` and no commits or remote. The fix also silences the repeated server-side `BASE_BRANCH_NOT_FOUND` ERROR log (the other half of the bug, not directly visible here) — the branch view now returns `null` base gracefully rather than surfacing an error on every repo-change tick.

<details>
<summary>Captured by a Playwright e2e scenario on a clean <code>fix/git-status-fresh-repo</code> build</summary>

Scenario: `Branch mode shows "No branch base to compare" for fresh git init repo with no remote or commits`

Steps exercised:
1. `git init /tmp/kolu-fresh-1246` (no commits, no remote)
2. Click Code tab
3. Click Code tab mode "branch"
4. Assert `[data-testid="diff-empty"]` contains "No branch base to compare" — ✅ passed

</details>

PR #1247, opened 2026-06-09
Zoom only the focused tile in canvas mode

### LINKED ISSUE #1238: Canvas zoom affects all tiles simultaneously — per-terminal font size cannot be aligned
### Description

In canvas mode, pressing Cmd/Ctrl +/- to zoom the font size of one terminal changes the font size of **every** open terminal simultaneously. This makes it impossible to set two terminals to the same font size if one was zoomed before the other was opened — the offset between them is preserved on every zoom keystroke.

### Root Cause

In `packages/client/src/terminal/Terminal.tsx:185`:

```ts
const fontSize = createZoom(props.terminalId, () => props.visible);
```

The `isActive` gate passed to `createZoom` is `props.visible`. In canvas mode (`App.tsx:366`), every tile is rendered with `visible={true}` so that inactive xterms stay mounted and sized correctly. This means every tile's zoom keydown listener passes the `isActive()` check. When the user presses Cmd+/-, all terminals' listeners fire in capture phase, incrementing/decrementing every terminal's font size together.

Font size is correctly stored per-terminal in localStorage (`kolu-font-size-${terminalId}`), so each terminal tracks its own value — but the zoom *input* is applied to all of them at once.

### Reproduction

1. Open kolu in canvas mode with two or more terminals.
2. Zoom one terminal with Cmd/Ctrl +/- (e.g. from 14 → 16).
3. Open a second terminal (starts at default 14).
4. Zoom the first terminal again. Observe the second terminal's font size also changes.
5. The two terminals can never be aligned to the same font size without closing one.

### Environment

- **OS:** Linux (NixOS)
- **Install method:** Nix flake (`kolu-stamped`)
- **Mode:** Canvas (tiled) mode — mobile is unaffected since only one tile is `visible` at a time.

### Proposed Fix

In `packages/client/src/terminal/Terminal.tsx:185`, change the `isActive` gate from `props.visible` to `props.focused`:

```ts
// before
const fontSize = createZoom(props.terminalId, () => props.visible);

// after
const fontSize = createZoom(props.terminalId, () => props.focused);
```

`focused` is `active()` in canvas mode (exactly one tile) and `visible()` in mobile mode (same semantics as before), so this is correct for both postures. Only the focused terminal should respond to zoom keys.

srid:
**Pressing Cmd/Ctrl +/- to change a terminal's font size now zooms only the focused tile.** In canvas mode every tile renders with `visible={true}` so inactive xterms stay sized — but that meant *every* tile's capture-phase zoom keydown listener passed its `isActive` gate, and a single keystroke zoomed every open terminal at once. Two tiles could never be aligned to the same font size if one was zoomed before the other opened: the offset was preserved on every keystroke.

The gate in `Terminal.tsx` moves from `props.visible` to `props.focused === true`. `focused` is true for exactly one tile — the active tile in canvas mode, the single visible tile in mobile — so only it responds to zoom keys. Mobile behavior is unchanged (there `visible` and `focused` were already the same). *Using `=== true` rather than the codebase's usual `!== false` is deliberate here: a future caller that omits the prop should default to "not focused, don't zoom" so the multi-zoom bug can't silently return.*

### Reproducing test

A new e2e scenario (`terminal.feature`) opens a second tile, zooms, and asserts **exactly one** terminal's font size changed. It fails on the old code (both tiles went 14 → 16) and passes with the fix.

Fixes #1238


srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `5d47800da8c9`

### Round 1

**codex** — approved: `false`

The production change in `Terminal.tsx` looks correct: `focused` is a reactive single-owner signal through the current canvas/mobile/sub-panel render paths, so gating `createZoom` on it fixes the multi-terminal zoom leak without obvious behavior regressions. The remaining issues are around the regression guard and stale helper documentation, not the runtime fix itself.

Findings:
- `F1` · minor · open — The new assertion only verifies that exactly one terminal changed font size; it does not verify that the changed terminal is the focused one. With the current two-terminal scenario, an inverted implementation that zoomed the inactive terminal would still pass because exactly one terminal would change. (packages/tests/step_definitions/terminal_steps.ts:335)
- `F2` · nit · open — The `createZoom` parameter docs still say zoom keys apply when "only the visible terminal zooms," but the caller now deliberately passes focused state instead of visibility. That stale contract is easy to misread because this helper is the behavior boundary for the fix. (packages/client/src/input/zoom.ts:19)

**claude** — Addressed both CODEX findings. F1 (minor, valid): strengthened the regression assertion so it now verifies the focused terminal is the one that changed, not just that some single terminal changed. F2 (nit, valid): fixed the stale `createZoom` doc comment that still referred to "visible terminal" zoom semantics. Both findings were correct; I agreed and fixed both. Formatter ran clean.

- `F1` **fixed** — Agreed — the old `exactly one terminal's font size should have changed` step would pass even for an inverted fix that zoomed the inactive tile. Confirmed `data-focused`, `data-terminal-id`, and `data-font-size` all live on the same element (packages/client/src/terminal/Terminal.tsx:967-971), so the focused id is reliably readable. Added `readFocusedTerminalId()` (packages/tests/step_definitions/terminal_steps.ts:318) which selects `[data-terminal-id][data-font-size][data-focused]`; `I note the font size of each terminal` now also captures `this.savedFocusedTerminalId` (asserting exactly one focused tile). Renamed the assertion step to `only the focused terminal's font size should have changed` and made it `assert.deepStrictEqual(changed, [focusedId])` so the single changed terminal must BE the focused one. Updated the .feature step text (terminal.feature:100) and added `savedFocusedTerminalId` to KoluWorld (support/world.ts).
- `F2` **fixed** — Agreed — the `@param isActive` doc at packages/client/src/input/zoom.ts:19 still said "only the visible terminal zooms," contradicting the new caller which passes focused state. Rewrote the comment to state callers pass focused state (not visibility), explaining that in canvas mode every tile is visible so a visibility gate would zoom all tiles (#1238), and exactly one tile is focused (active in canvas; single visible one in mobile).

commit: `c2a28bc2f9c8e90a22e93561fc8cf318394bd60e`

### Round 2

**codex** — approved: `true`

CLAUDE addressed both prior findings in the current tree. The focused-terminal regression test now captures the focused terminal id before zoom and asserts the only changed font-size entry is that id, closing the inverted-implementation gap. The `createZoom` parameter documentation now matches the focused-state contract. I did not raise any new findings from this round's changes.

Findings:
- `F1` · minor · resolved — Original issue: the regression assertion only verified that exactly one terminal changed font size, not that the changed terminal was the focused one. (packages/tests/step_definitions/terminal_steps.ts:347)
- `F2` · nit · resolved — Original issue: the `createZoom` `isActive` docs still described visibility-based zoom ownership, which contradicted the new focused-state gate. (packages/client/src/input/zoom.ts:19)

_codex on the rebuttal:_ No disputes to address. CLAUDE accepted both findings, and I verified the current diff resolves F1 and F2.

srid:

✅ **Consensus** after 2 round(s) · lowy + hickey · base `5d47800da8c9`

### Applied (6)
- `lowy-2` Gate predicate (=== true) and the data-focused attribute predicate (!== false) diverge — commit `a516dc49e`
- `hickey-1` readAllFontSizes duplicates readAllCols (concept multiplication) — commit `2714cb736`
- `hickey-2` readFocusedTerminalId re-derives the canonical focused-terminal selector — commit `dbd976d4d`
- `hickey-3` savedFontSizes / savedFocusedTerminalId fragment one captured snapshot into two world fields — commit `78421edb0`
- `hickey-4` Three divergent 'is this the chosen tile' predicates now coexist in Terminal.tsx — (uncommitted)
- `hickey-5` 'I note the font size of each terminal' is a Given that performs an observation — commit `0acbbd257`

### Agreed — no change (2)
- `lowy-1` Zoom-ownership invariant ('exactly one focused tile') is enforced by convention, not encapsulated (packages/client/src/input/zoom.ts:42-47 (and Terminal.tsx:190))
- `lowy-3` createZoom JSDoc bakes the consumer's layout semantics into the receptacle's contract (packages/client/src/input/zoom.ts:19-23)

srid:

Ran the three passes (rules → fact-check → elegance) on the settled tree after the codex and lens debates. **All clear — no findings, nothing to fix.**

| Pass | Issues | Notes |
|------|--------|-------|
| **Rules** | 0 | Full checklist clean — `#1238` appears in three files as a cross-reference, not duplicated content; `readPerTerminal`'s serializable-discriminant dispatch is the minimal correct form inside `page.evaluate` (ts-pattern can't cross the boundary); no dead code left after the helper consolidations. |
| **Fact-check** | 0 | `isFocused = () => props.focused === true` is reachable only with explicit-boolean callers (all current ones), so the `undefined` semantics change is unobservable today and documented; `deepStrictEqual(changed, [focusedId])` correctly asserts *exactly one changed and it's the focused one*; `readFocusedTerminalId`'s null is asserted, never swallowed. |
| **Elegance** | 0 | Four lenses, no fixes — the lens debate already deduped the test helpers (`readPerTerminal`) and consolidated the world snapshot. |

srid:
## Evidence

Two terminals open side by side, both at 14px. Pressing Ctrl/Cmd `+` with **TERMINAL TWO** (right) focused zooms **only that tile** — it grows to 23px while **TERMINAL ONE** (left) stays at 14px.

Before #1238's fix, every *visible* tile shared the zoom keystroke, so in canvas mode (where every tile is visible) all open terminals zoomed together and two tiles could never hold different font sizes. Now the zoom gate is `focused`, so exactly one tile responds.

before/after: zooming one canvas tile no longer affects the other

*Captured on a live kolu (dev build of this branch) via chrome-devtools; the e2e scenario `Zoom only affects the focused canvas tile` asserts the same invariant in CI.*

Issue #1243, opened 2026-06-09
feat: add canvas maximize toggle to the command palette


DivitMittal:
## Problem

The command palette (`CommandPalette.tsx`) has entries for creating terminals, toggling the right panel, toggling the dock, switching workspaces, and a range of other workspace operations — but there is no command to toggle the canvas maximize/tiled posture. The only way to change posture is the mouse (ChromeBar button or title-bar double-click). A user who reaches for the command palette to "maximize" their view finds nothing.

## Where

- `packages/client/src/commands.tsx` — registers all command palette commands. No reference to `canvasMaximized`, `useViewPosture`, or `togglePosture` anywhere in the file.
- `packages/client/src/canvas/useViewPosture.ts:56` — `toggle()` is the single writer; `canMaximize()` is the guard.

## Proposed fix

Register a command in `commands.tsx`, gated on `posture.canMaximize()` so it never appears when the canvas is empty (consistent with the ChromeBar button being disabled in that state):

```ts
{
  id: "toggleCanvasPosture",
  label: () => posture.mode() === "maximized" ? "Exit maximized view" : "Maximize view",
  icon: /* appropriate icon */,
  visible: () => posture.canMaximize(),
  run: () => posture.toggle(),
},
```

The dynamic label is important: when already maximized, the command should read "Exit maximized view" (or "Restore tiled view"), not "Maximize view" — otherwise the palette describes the current state rather than the action. This matches the pattern already used by the ChromeBar button (`ChromeBar.tsx:68`: *"The maximize toggle's affordance describes the action a click performs"*).

## Acceptance

- Searching "maximize" (or "tiled") in the command palette shows the toggle command when ≥1 terminal is open.
- The command label reflects current state: "Maximize view" when tiled, "Exit maximized view" (or similar) when maximized.
- The command is absent from the palette when zero terminals are open.
- Executing it toggles posture identically to clicking the ChromeBar button.

PR #1248, opened 2026-06-09
Reach the canvas maximize toggle from the keyboard and command palette

### LINKED ISSUE #1242: feat: keyboard shortcut for canvas maximize toggle
## Problem

The maximize/tiled toggle is only reachable by mouse: either clicking the square button in `ChromeBar` (`data-testid="maximize-toggle"`) or double-clicking a tile's title bar (`CanvasTile.tsx:229`). There is no keybind, so keyboard-first users have no way to switch postures without leaving the keyboard.

The underlying action is already a clean single writer — `useViewPosture().toggle()` in `packages/client/src/canvas/useViewPosture.ts:56` — and it already has its own write guard (`canMaximize()`). All the infrastructure is in place; the keybind just doesn't exist.

## Where

- `packages/client/src/input/actions.ts` — defines every `ACTIONS` entry (keybind + handler). No entry exists for canvas posture toggle. Compare: `toggleDock`, `toggleRightPanel`, `toggleSubPanel` all have entries here.
- `packages/client/src/canvas/useViewPosture.ts:56` — the `toggle()` function, the single writer.

## Proposed fix

Add an entry to `ACTIONS` in `actions.ts`:

```ts
toggleCanvasPosture: {
  label: "Toggle maximize",
  keybind: /* e.g. Ctrl+Shift+M or similar — check prohibitedKeybinds.ts */,
  handler: (ctx) => ctx.toggleCanvasPosture(),
},
```

Then expose `toggleCanvasPosture` through the action context (same pattern as `toggleRightPanel`) and wire it to `useViewPosture().toggle()`. The write guard in `toggle()` makes it a no-op at zero terminals, so no extra guard is needed at the call site.

Before picking a chord, check `packages/client/src/input/prohibitedKeybinds.ts` — those chords are claimed by tools running inside kolu PTYs and must not be claimed. Add a test entry to the keyboard test suite once a chord is chosen (per the conventions in `.claude/rules/conventions.md`).

## Acceptance

- Pressing the shortcut in canvas mode with ≥1 terminal toggles between `"tiled"` and `"maximized"` postures.
- Pressing it with zero terminals is a no-op (no state change, no crash).
- The chord does not appear in `prohibitedKeybinds.ts`.
- The `ShortcutsHelp` sheet (if applicable) lists the new binding.

srid:
**The maximize/tiled posture toggle was mouse-only** — the chrome-bar button or a title-bar double-click. Keyboard-first users had no way to switch postures without leaving the keyboard, and a user reaching for the command palette to "maximize" found nothing. This wires the existing `useViewPosture().toggle` — the single writer, already guarded to no-op at zero terminals — into two new surfaces.

Closes #1242 and #1243.

### Two new entry points to one existing action

- **Keyboard** — a `toggleCanvasPosture` action bound to <kbd>Cmd/Ctrl+Shift+M</kbd> (M for maximize), dispatched through the shared `ActionContext` exactly like `toggleDock` / `toggleRightPanel`. The chord follows the `Mod+Shift+<letter>` convention and stays clear of the in-PTY chords in `prohibitedKeybinds.ts`.
- **Command palette** — a desktop Canvas-section command gated on `posture.canMaximize()`, so it's absent on an empty workspace (matching the ChromeBar button being disabled there). It carries the keybind chip and a **state-reflecting label**: *"Maximize terminal"* when tiled, *"Restore canvas"* when already maximized — the same wording as the ChromeBar affordance, so the label always names the action the next select performs rather than the current state.

Both readers reach the posture through the `useViewPosture` seam, the same way ChromeBar and the Dock do — no new state, no second writer.

### Verification

- `keyboard.test.ts` — a matcher for the new chord; the existing prohibited-collision test auto-iterates `ACTIONS × PROHIBITED_KEYBINDS`, so it now also proves <kbd>Cmd/Ctrl+Shift+M</kbd> never steals a keystroke meant for the PTY.
- Two e2e scenarios in `canvas.feature` — the keyboard toggle round-trip, and the palette toggle with its label flipping tiled → maximized → tiled.

Docs: README shortcut list + maximize-affordances line, and a changelog entry.


srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 round(s) · codex reviewed at `xhigh` reasoning effort · base `5d47800da8c9`

### Round 1

**codex** — approved: `false`

The feature is mostly wired correctly and has focused e2e coverage for the desktop shortcut and palette label. I found no whitespace issues with `git diff --check`, and `git status --short` showed no untracked files. I did not run test suites because the review instruction was read-only and tests may emit artifacts. The main concerns are capability gating drift and a missed in-app discoverability update.

Findings:
- `F1` · minor · open — The new global shortcut calls `posture.toggle` even when the spatial canvas is not available, such as mobile or a narrow viewport. The command palette hides canvas commands via `supportsSpatialCanvas()`, but the keyboard path can still mutate the persisted `kolu-canvas-maximized` flag invisibly, so a mobile/narrow user with a hardware keyboard can change the later desktop posture without seeing any effect now. (packages/client/src/App.tsx:261)
- `F2` · minor · open — `createCommands` imports and calls `useViewPosture()` directly, and the palette action calls `posture.toggle()` instead of the `deps.toggleCanvasPosture` handler already provided through `CommandDeps`. That violates the local command pattern where commands stay declarative and call deps handlers, and it creates two action paths that can drift if App adds guards or telemetry to the keyboard path. (packages/client/src/commands.tsx:162)
- `F3` · nit · open — The existing maximize discoverability tips still only mention double-clicking the title bar or using the header toggle. Since this PR adds a user-facing shortcut and palette command, the in-app tip surface remains stale and misses the new keyboard-first path. (packages/client/src/settings/tips.ts:77)

**claude** — Addressed all three CODEX findings. F1 (capability gating): agreed and fixed by gating useViewPosture's toggle() on supportsSpatialCanvas() in the receptacle, so neither the keyboard nor palette path can flip the persisted kolu-canvas-maximized pref on a mobile/narrow viewport. F2 (writer drift): agreed and fixed by routing the palette command's onSelect through deps.toggleCanvasPosture() (the shared ActionContext seam the keyboard uses) while keeping the useViewPosture() reader only for the reactive label/visibility — the ChromeBar/Dock reader pattern. F3 (stale tip): agreed and fixed by updating the amb-chrome-maximize ambient tip to advertise the new Mod+Shift+M shortcut and the "Maximize terminal" palette command. Updated stale doc comments in App.tsx and actions.ts to reflect the new spatial-canvas guard. just fmt clean; client tsc --noEmit passes.

- `F1` **fixed** — Agreed. toggle() previously guarded only on canMaximize() (terminal count > 0), so on mobile/narrow viewport mode() no-ops the *display* but toggle() still flipped the persisted kolu-canvas-maximized pref. Fixed in packages/client/src/canvas/useViewPosture.ts:56 by gating the write on supportsSpatialCanvas() && canMaximize(). Putting it in the receptacle (not just the keyboard caller) matches the hook's documented 'safety lives in the receptacle' philosophy and the existing convention where canvas handlers gate on supportsSpatialCanvas() (App.tsx:210 handleCanvasCenterActive, useCanvasArrange.ts:61). Also updated the now-inaccurate guard comments in App.tsx:227 and input/actions.ts:50.
- `F2` **partial** — Agreed on the real drift risk — the write path. Fixed packages/client/src/commands.tsx:348 so onSelect calls deps.toggleCanvasPosture() (the shared ActionContext seam the keyboard shortcut also uses) instead of posture.toggle(), so the two surfaces can't drift if App later wraps the toggle with a guard/telemetry. Did NOT route the label/visibility reads (mode()/canMaximize()) through CommandDeps as the suggestion's first half proposed: those are reactive accessor reads, and calling the useViewPosture() singleton reader directly in createCommands is the same hook-composition pattern ChromeBar/Dock use — it is not the async/RPC/multi-step logic that solidjs.md's 'commands stay declarative' rule targets, and a plain deps handler can't carry reactive accessors as cleanly. So commands stay declarative for the action (write via deps) while reading posture reactively for the label.
- `F3` **fixed** — Agreed. Updated the existing amb-chrome-maximize ambient tip in packages/client/src/settings/tips.ts:82 to lead with formatKeybind(ACTIONS.toggleCanvasPosture.keybind) and mention the 'Maximize terminal' command-palette path, alongside the existing header-toggle mention. Folded into the existing tip rather than adding a new entry to keep the tip surface tight.

commit: `8083e84ac76b31fc5d75b7186f41ba792b812f3e`

### Round 2

**codex** — approved: `false`

CLAUDE adequately addressed the three findings from the previous round: the toggle now has a spatial-canvas write guard, the palette write path goes through the shared ActionContext handler, and the ambient tip advertises the new shortcut/palette path. I found one new nit introduced by the F1 fix: the header comment in useViewPosture now contradicts the new dependency on mobile/spatial capability state.

Findings:
- `F1` · minor · resolved — The global maximize shortcut could mutate the persisted canvas-maximized preference on surfaces where the spatial canvas is unavailable, such as mobile/narrow viewports. (packages/client/src/canvas/useViewPosture.ts:60)
- `F2` · minor · resolved — The palette command previously called `posture.toggle()` directly, creating a second write path that could drift from the keyboard ActionContext handler. (packages/client/src/commands.tsx:348)
- `F3` · nit · resolved — The in-app maximize tip was stale and did not mention the newly added shortcut or palette command. (packages/client/src/settings/tips.ts:82)
- `F4` · nit · open — The file header still says mobile is handled one level up and “deliberately stays out of this hook,” but the fix now imports `supportsSpatialCanvas()` and makes mobile/narrow spatial capability part of the hook’s write guard. That comment is now misleading exactly where future maintainers will look to understand the boundary. (packages/client/src/canvas/useViewPosture.ts:13)

_codex on the rebuttal:_ F2: I accept the partial fix. The important drift risk was the write path, and `onSelect: () => deps.toggleCanvasPosture()` now shares the same ActionContext seam as the keyboard shortcut. Keeping `useViewPosture()` as a reactive reader for label/visibility is acceptable here because it does not introduce a second mutation path and matches the existing reader pattern. F1 and F3 were not disputed; I verified both fixes.

**claude** — Addressed the sole open finding (F4). The other three findings (F1, F2, F3) were already resolved and confirmed by CODEX in this round, so no further action. F4 was a valid documentation-staleness nit: the useViewPosture.ts file header claimed mobile "deliberately stays out of this hook," which directly contradicts the F1 fix that added a supportsSpatialCanvas() write guard to toggle(). I reconciled the header to say the posture *state* (mode/canMaximize) stays canvas-focused while noting toggle() guards on spatial-canvas availability as the explicit exception. just fmt clean.

- `F1` **fixed** — Already resolved in the prior round and re-confirmed resolved by CODEX this round. toggle() in packages/client/src/canvas/useViewPosture.ts:60 guards the write on supportsSpatialCanvas() && canMaximize(). No further change.
- `F2` **fixed** — Already resolved in the prior round and re-confirmed by CODEX (it accepted the partial fix as sufficient in responseToRebuttal). The palette onSelect in packages/client/src/commands.tsx:348 routes through deps.toggleCanvasPosture(), the same ActionContext seam as the keyboard shortcut. No further change.
- `F3` **fixed** — Already resolved in the prior round and re-confirmed resolved by CODEX. The amb-chrome-maximize tip in packages/client/src/settings/tips.ts now advertises the Mod+Shift+M shortcut and the 'Maximize terminal' palette command. No further change.
- `F4` **fixed** — Agreed — CODEX is right. The file header in packages/client/src/canvas/useViewPosture.ts (lines 13-16) categorically claimed mobile 'deliberately stays out of this hook,' which the F1 fix contradicts: toggle() now imports and calls supportsSpatialCanvas(), a mobile/narrow-viewport gate. A maintainer reading the header to learn the boundary would be misled exactly where it matters. Rewrote the header to scope the 'stays out' claim to the posture *state* (mode/canMaximize) and to call out toggle()'s supportsSpatialCanvas() guard as the explicit exception, pointing at its own doc comment. just fmt clean.

commit: `f826244281b3a97857a21869c304abf87428d8a4`

### Round 3

**codex** — approved: `true`

All previously raised findings are resolved. I verified the current tree is clean of untracked files, and the latest useViewPosture header now accurately describes the posture-state boundary while calling out the `toggle()` spatial-canvas guard as the explicit exception. No new regression was introduced by this round's change.

Findings:
- `F1` · minor · resolved — The global maximize shortcut could mutate the persisted canvas-maximized preference on surfaces where the spatial canvas is unavailable, such as mobile/narrow viewports. (packages/client/src/canvas/useViewPosture.ts:63)
- `F2` · minor · resolved — The palette command previously called `posture.toggle()` directly, creating a second write path that could drift from the keyboard ActionContext handler. (packages/client/src/commands.tsx:348)
- `F3` · nit · resolved — The in-app maximize tip was stale and did not mention the newly added shortcut or palette command. (packages/client/src/settings/tips.ts:82)
- `F4` · nit · resolved — The file header previously claimed mobile deliberately stayed out of this hook, contradicting the new `supportsSpatialCanvas()` write guard. (packages/client/src/canvas/useViewPosture.ts:13)

_codex on the rebuttal:_ F4: I agree with the fix. The header now limits the "mobile stays out" boundary to posture state and explicitly documents `toggle()` as the exception that guards on `supportsSpatialCanvas()`, so the earlier contradiction is gone. F1-F3 remained resolved.

srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `5d47800da8c9`

### Applied (3)
- `lowy-1` Palette command bypasses the actionPaletteCommand receptacle — commit `41b3479eb`
- `hickey-1` Posture affordance label derived in two readers (and literal'd in a third) — commit `8fcd6457d`
- `hickey-2` One toggle action carries four different names — commit `fa2010b99`

### Agreed — no change (2)
- `lowy-2` Asymmetric write path: ChromeBar/TerminalCanvas write posture directly, palette routes through ActionContext (packages/client/src/commands.tsx:165,348)
- `lowy-3` supportsSpatialCanvas guard correctly relocated into the receptacle (packages/client/src/canvas/useViewPosture.ts:63-67)

srid:

Ran rules → fact-check → elegance over the branch after the codex and lens debates.

**Rules** — clean. `posturedActionLabel` is the single home for the label strings (no independent literals), the `supportsSpatialCanvas() && canMaximize()` double-gate sits at the right altitude (outer hides the canvas section, inner is maximize-specific), and the new e2e steps poll async state via `POLL_TIMEOUT` / `waitForFunction`. No dead code, no thin wrappers, no swallowed errors.

**Fact-check** — one real (cosmetic) finding, **fixed**:

> The `amb-chrome-maximize` discovery tip referenced only the tiled-state palette label `"Maximize terminal"`. If the tip fired while the user was already maximized, the palette actually reads `"Restore canvas"` — so the tip named an entry that wasn't there. Now it names **both** labels (`posturedActionLabel("tiled")` / `posturedActionLabel("maximized")`), so it's accurate in either posture.

Fixed in `338eb017`.

**Elegance** — clean; all four lenses (reuse, simplification, efficiency, altitude) found nothing to change.

srid:
## Evidence

Scenario recorded: **"Command palette toggles canvas maximize with a state-reflecting label"** (`features/canvas.feature`, 1 scenario / 12 steps passed, clean `pu` box, branch `elfin-label`).

The clip shows:
1. Two canvas tiles in the tiled posture.
2. Command palette opened → "Maximize" typed → **"Maximize terminal"** selected → tile expands to fill the viewport.
3. Palette reopened → "Restore" typed → **"Restore canvas"** selected (label reflects maximized state) → tile shrinks back to the canvas.



▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=evidence-canvas-maximize-toggle.mp4

DivitMittal:
Tested locally. Both the `Mod+Shift+M` shortcut and the command palette entry work as expected — label flips correctly between "Maximize terminal" and "Restore canvas". Also confirmed the entry is hidden on an empty workspace. LGTM!

PR #1249, opened 2026-06-09
Open external links in HTML previews in a new browser tab

srid:
**An external link clicked inside a previewed `.html` file now opens in a new browser tab** instead of doing nothing or hijacking the preview pane. The Code-tab HTML preview renders in an opaque-origin iframe sandboxed `allow-scripts` only — no `allow-popups`, no `allow-top-navigation` — so a `target="_blank"` link was silently blocked and a plain `<a href="https://…">` would navigate the *preview itself* to the remote page in-pane. Neither is what a user expects of an outbound link.

### How it fits together

The fix mirrors the existing same-frame-navigation and mouse back/forward bridges — trap the event the sandbox swallows, forward the intent to the unsandboxed parent:

```
preview iframe                          parent (Code tab)
  click <a> → external? ──postMessage──▶ observeIframeOpenExternal
  (http(s), different host)              → re-check scheme is http(s)
  preventDefault                         → window.open(url, _blank, noopener)
```

- **In-iframe SDK** decides a link is external (`http(s)` to a *different host*) and posts the absolute URL; same-host links keep their existing in-frame navigation (the tree still follows them).
- **Parent observer** re-validates the scheme before calling `window.open` — `postMessage` is reachable by any in-frame script, so a `javascript:`/`data:` URL must never reach the parent's trusted origin. _Defense in depth: the iframe already filters, the parent checks again._
- The preview **stays on the current document** — the click escaped to a tab, it didn't replace the pane.

### Tests

- New e2e scenario (`code-tab.feature`): clicking an external link opens a new context page to the target URL *and* leaves the preview/tree on the original file. The scenario fails on `master` (no popup ever opens).
- Unit coverage for the parent observer's scheme guard — `javascript:`/`data:`/`mailto:` and non-string/cross-source payloads are all dropped.


srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `11fd6f36b531`

### Round 1

**codex** — approved: `false`

The feature is mostly wired coherently, but I would not approve it yet. The main concern is that the parent now opens tabs from an unauthenticated message emitted by untrusted, script-capable preview content, which weakens the sandbox boundary. I also found a couple of smaller correctness/test gaps around link classification and click coverage.

Findings:
- `F1` · major · open — The parent opens any `kolu-artifact-sdk:open-external` message from the iframe. Previewed HTML is allowed to run arbitrary scripts, and those scripts can call `window.parent.postMessage({ type: "kolu-artifact-sdk:open-external", url: "https://..." }, "*")` without going through the SDK's trusted click handler. Scheme filtering blocks `javascript:`/`data:`, but it does not prove the URL came from an actual user-clicked anchor, so this effectively grants the sandboxed document an http(s) popup capability despite omitting `allow-popups`. (packages/client/src/right-panel/CodeTab.tsx:990)
- `F2` · minor · open — The external-link trap only handles primary-button `click` events. Middle-click/auxclick is the normal "open this link in a new tab" gesture, but it still falls through to the sandbox path and can be swallowed. The comment also says the code does not branch on middle clicks, which is not true. (packages/artifact-sdk/src/iframe/index.ts:74)
- `F3` · minor · open — External-link classification compares `url.host` to `window.location.host`, not origin. A link to the same host over a different scheme, such as http vs https, is a different origin but is treated as internal and can replace the preview in-pane. (packages/artifact-sdk/src/iframe/index.ts:62)
- `F4` · nit · open — The popup assertion uses `url.includes(expected)`, so the regression test would pass for wrong destinations like `https://example.com/docs-bad` or `https://example.com/?next=https://example.com/docs`. (packages/tests/step_definitions/code_tab_steps.ts:728)

**claude** — Addressed all four CODEX findings on the external-link-to-tab feature. Fixed F2 (middle-click), F3 (origin vs host), and F4 (exact URL comparison in the e2e assertion). For F1 (unauthenticated open-external message), I conceded the observation but disputed the proposed host-side confirmation: I documented the capability honestly across types.ts and CodeTab.tsx instead, on the grounds that the granted capability is strictly weaker than what a sandboxed script already has.

- `F1` **partial** — Conceded the core observation: open-external is an unauthenticated postMessage, and because previewed HTML runs arbitrary scripts under the opaque origin, ANY in-frame script (not just the SDK click trap) can post it. The prior comments understated this by implying scheme-checking was the whole story. I fixed that — types.ts:84 and CodeTab.tsx:986 now explicitly document it as a request from untrusted content. BUT I dispute the suggested host-side confirmation dialog. The granted capability — open a noopener,noreferrer http(s) FOREGROUND tab — is strictly weaker than what a sandboxed `allow-scripts` frame already has: it can `location =` itself to any URL and `fetch()` outbound, so egress and navigation already exist. window.open with noopener,noreferrer adds no same-origin/credential reach (opener severed, opaque origin), only a more visible surface; worst case is tab-spam/phishing-landing, which a confirm-on-every-link dialog would penalize the core UX to mitigate marginally. The real escalation risk — a javascript:/data: URL executing in kolu's trusted origin — is already blocked by isHttpUrl re-validation in observeIframeOpenExternal (bridge.ts:186) before window.open is ever reached, and is unit-tested in bridge.test.ts:190. So: documented honestly (fixed), confirmation affordance declined (disputed).
- `F2` **fixed** — Correct — onAnchorClick returned early on `event.button !== 0`, so a middle-click on an external link fell through and the opaque-origin sandbox swallowed it; the comment claiming 'we don't branch on middle clicks' was also false (it did, via the button guard). Fixed both: the guard now accepts button 0 (left) and 1 (middle) and rejects 2 (right, left to the context menu), and I registered a second `document.addEventListener("auxclick", onAnchorClick)` in boot() since middle-click fires auxclick, not click. Reused the same handler. Rewrote the comment in iframe/index.ts:69 to describe the actual behavior.
- `F3` **fixed** — Correct — externalHref compared url.host to window.location.host, so a same-host link over a different scheme (http vs the document's https) is a different origin but was treated as internal and could replace the preview in-pane. Changed the comparison to url.origin === window.location.origin in iframe/index.ts:63 and updated the doc comment to say origin, not host.
- `F4` **fixed** — Correct nit — the popup assertion used `url.includes(expected)`, which would pass for wrong destinations like https://example.com/docs-bad or a `?next=` query containing the expected string. Changed to exact normalized comparison: `new URL(url).href === new URL(expected).href`, wrapped in try/catch so a transient about:blank popup URL doesn't throw during polling, in code_tab_steps.ts:721.

commit: `a958f523c74efec67488b71d11029edabfbfceca`

### Round 2

**codex** — approved: `true`

All four previously raised findings are resolved in the current tree. I verified the current diff and status; there are no untracked files. F2, F3, and F4 are fixed directly. For F1, I accept CLAUDE's dispute after the code now documents `open-external` as an unauthenticated request from untrusted iframe content and keeps the concrete dangerous schemes blocked before `window.open`.

Findings:
- `F1` · major · resolved — Previously, the parent opened tabs from an unauthenticated iframe message while comments implied the SDK click trap was the meaningful trust source. The current code explicitly documents this as an accepted request from untrusted in-frame content, not proof of a clicked anchor, and still constrains the parent action to revalidated http(s) URLs opened with `noopener,noreferrer`. (packages/client/src/right-panel/CodeTab.tsx:992)
- `F2` · minor · resolved — Previously, middle-click external links were not trapped. The current code accepts button 0 and 1 and registers both `click` and `auxclick`, leaving right-click to the context menu. (packages/artifact-sdk/src/iframe/index.ts:79)
- `F3` · minor · resolved — Previously, external-link classification compared host instead of origin. The current code compares `url.origin` to `window.location.origin`, so same-host cross-scheme links are treated as external. (packages/artifact-sdk/src/iframe/index.ts:64)
- `F4` · nit · resolved — Previously, the popup assertion used substring matching. The current assertion compares normalized URL hrefs exactly and tolerates transient non-target popup URLs during polling. (packages/tests/step_definitions/code_tab_steps.ts:730)

_codex on the rebuttal:_ F1: I accept the dispute after the change. The important review objection was that the parent action was not authenticated to a real SDK-handled anchor click. The current comments in `types.ts` and `CodeTab.tsx` now state that directly and treat it as an intentional untrusted-content capability, while `observeIframeOpenExternal` still blocks non-http(s) schemes before `window.open` and `CodeTab` uses `noopener,noreferrer`. I would not lean too hard on "strictly weaker" for every UX threat model, since a foreground tab is different from in-frame navigation, but as an explicit accepted capability it resolves the finding. F2, F3, and F4 were verified fixed as described.

srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `11fd6f36b531`

### Applied (2)
- `lowy-2` isHttpUrl duplicates the scheme-check half of externalHref across the package boundary — commit `7a99b0af1`
- `hickey-1` http(s) scheme predicate duplicated across the sandbox boundary — (uncommitted)

### Agreed — no change (4)
- `lowy-1` Three near-identical observer effects in BrowseIframeRenderer instead of one bind-all-observers seam (packages/client/src/right-panel/BrowseIframeRenderer.tsx:57-98)
- `lowy-3` onOpenExternal is optional but the only host that wires the renderer always supplies it (packages/client/src/right-panel/BrowseIframeRenderer.tsx:44 and BrowseFileDispatcher.tsx:110)
- `hickey-2` Three parallel iframe-to-parent intents each fan out across four layers (packages/artifact-sdk/src/client/bridge.ts:139, packages/client/src/right-panel/BrowseIframeRenderer.tsx:91-98, packages/client/src/right-panel/BrowseFileDispatcher.tsx:110/313/324, packages/client/src/right-panel/CodeTab.tsx:988)
- `hickey-3` externalHref re-parses the URL the scheme check already parsed conceptually, but origin and scheme are one decision (packages/artifact-sdk/src/iframe/index.ts:202-212)

srid:

Ran rules → fact-check → elegance on the settled tree (after codex + lens fixes). No errors — only pre-existing warnings in unrelated files. Three fixes applied and committed (`fix(police):`):

| Finding | Fix |
| --- | --- |
| Bare `catch {}` in `core/url.ts` `isHttpUrl` | Commented — `new URL()` throws for any unparseable string, which is correctly *not* an http(s) URL |
| Bare `catch {}` in the e2e popup assertion | Commented — `popup.url()` can be `about:blank` before the navigation commits, so a parse failure just means "poll again" |
| Rule-of-three: three identical `createEffect`/guard/`onCleanup` observer blocks in `BrowseIframeRenderer` | Extracted into a `useIframeObserver(iframeEl, observe)` helper, called three times. _This also retires the lens debate's `lowy-1`/`hickey-2` "three near-identical effects" finding that was dropped as below-the-bar — the helper makes a 4th intent a one-liner._ |

**Skipped (out of strict diff scope):** the unguarded `handle(extracted)` in the pre-existing `observeFromIframe` (shared by all three observers, not introduced here); and reacting to `window.open` returning `null` on a blocked popup (browsers expose no reliable popup-blocked signal).

srid:
## Evidence



▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=html-preview-external-link.mp4

Before this PR, clicking an external link inside an HTML preview in the Code tab would either do nothing (`target=_blank` blocked by the sandboxed iframe) or navigate the preview pane away from the original document. Now the in-iframe artifact-sdk intercepts the click and tells the parent to open the URL in a real browser tab — the preview stays on "home page" throughout, while `https://example.com/docs` opens elsewhere. The scenario asserts both: the new tab opens to the correct URL **and** the iframe still contains "home page" with `home.html` still selected in the file browser.

srid:

The **x86_64-linux** lane ran on **`kolu-ci-2`** (`dev-x86-64-linux-04`) — commit `fc1c115f`, **exit 0** ✓

- **Lane wall** (pipeline): **9m2s**
- **Wrapper wall** (incl. lease + nix-run startup): 43m37s

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 8m21s | ✓ |
| `ci::home-manager` | 31s | ✓ |
| `_ci-setup` | 27s | ✓ |
| `ci::unit` | 24s | ✓ |
| `ci::nix` | 17s | ✓ |
| `ci::install` | 14s | ✓ |
| `ci::fmt` | 14s | ✓ |
| `ci::biome` | 14s | ✓ |
| `ci::smoke` | 13s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|
| `kolu-ci-3` | dev-x86-64-linux-03 | 🔒 leased |

PR #1250, opened 2026-06-09
Stop terminal switches from resetting the Code tab's split

srid:
**Dragging the tree/preview split in the Code tab now actually sticks.** Previously, switching to another terminal could silently overwrite the saved split fraction with a garbage value, so coming back (or reloading) brought the panes up at the wrong size — the exact repro: preview a file, resize the preview pane, open a second terminal, switch back.

### Root cause

The split persists via `preferences.rightPanel.codeTabTreeSize`, fed by Corvu's `onSizesChange`. But Corvu's `Resizable` also fires `onSizesChange` from `unregisterPanel` while the component **unmounts** — and that emission is degenerate: it removes the unregistering panel's entry and "renormalizes" the survivor to a **length-1 array** holding `content − tree`.

```
mounted:   sizes = [0.27, 0.73]          (tree, preview — dragged by the user)
unmount:   tree panel unregisters
           → onSizesChange([0.73 − 0.27]) = [0.4569]   ← length 1, in the legal 0.1–0.9 band
           → persisted as codeTabTreeSize 😱
remount:   renders at 45.7% / 54.3% instead of 27% / 73%
```

The Code tab unmounts behind `<Show when={repoPath()}>` whenever the active terminal isn't in a git repository — which is exactly what a freshly opened second terminal often is. The fix gates the handler on `sizes.length === 2`: only a report describing a real two-panel layout reaches the preference.

*The horizontal splits in `App.tsx` and `TerminalContent.tsx` read `sizes[1]` — `undefined` in the degenerate emission — so they were accidentally safe. `App.tsx`'s comment claimed its guard was mere TypeScript narrowing; corrected to document the real hazard.*

### Regression test

New e2e scenario (`code-tab.feature`, with raw-mouse drag steps for the zero-height Corvu handle): drag the split 80px, hop to a second terminal in a non-git dir, switch back, assert the preview pane height survived. Red before the fix (settled at 554px vs the dragged 743px), green after.


srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `11fd6f36b531`

### Round 1

**codex** — approved: `false`

The production fix is sound: the CodeTab vertical splitter now ignores Corvu's one-panel unregister emission, which matches the installed Corvu implementation and prevents the persisted tree fraction from being clobbered. I found one test-coverage issue in the new regression scenario: it can switch back before the non-git fallback has actually mounted, so it may not deterministically exercise the unmount path the fix is about.

Findings:
- `F1` · minor · open — After changing the second terminal to a non-git directory, the scenario immediately switches back without waiting for CodeTab to render its `diff-no-repo` fallback. Cwd/git metadata is asynchronous, so the test can race back before `<Show when={repoPath()}>` flips and before the inner Resizable unmounts. In that case the regression test can pass without exercising the Corvu unregister emission that previously corrupted `codeTabTreeSize`. (packages/tests/features/code-tab.feature:1111)

**claude** — CODEX raised one valid minor finding (F1) about a race in the new regression scenario. I agreed and fixed it. The scenario previously ran `cd` into a non-git dir and then immediately switched terminals back; because cwd/git metadata is async, the switch-back could win the race before `<Show when={repoPath()}>` flipped to its fallback and unmounted the inner Resizable — meaning Corvu's unregister emission (the exact code path the fix guards) might never fire, so the test could pass vacuously. I inserted the existing assertion `Then the Code tab should indicate no git repository` (waits for the `diff-no-repo` fallback to mount) between the cd and the switch-back, which deterministically proves the Resizable unmounted before judging persistence. I verified the `diff-no-repo` fallback is rendered unconditionally on a falsy `repoPath()` regardless of browse/review mode (CodeTab.tsx:677-687), and that the step at code_tab_steps.ts:300-306 waits for state "visible". Adjusted the following `And I select workspace switcher entry 1` to `When` to keep Gherkin step keywords coherent after the inserted `Then`. Ran just fmt — clean.

- `F1` **fixed** — Agreed. Added `Then the Code tab should indicate no git repository` at packages/tests/features/code-tab.feature:1116 (between the `cd` into the non-git dir and `select workspace switcher entry 1`), using the pre-existing step at code_tab_steps.ts:300-306 that waits for the `diff-no-repo` fallback to become visible. This forces the test to wait until `<Show when={repoPath()}>` flips and the inner Resizable unmounts — guaranteeing Corvu's length-1 unregister emission (the path the fix gates) actually fires before the scenario checks persistence, closing the vacuous-pass race. Changed the subsequent `And I select workspace switcher entry 1` to `When` so the step keyword sequence stays coherent after the inserted `Then`.

commit: `e27723d3d568f9a879d53678e5e9869cb4e3082a`

### Round 2

**codex** — approved: `true`

The only prior finding, F1, is now addressed. The scenario now waits for the Code tab's non-git fallback before switching back, which forces the inner Resizable to unmount and exercises the Corvu unregister emission path the production guard is meant to protect. I did not identify any regression introduced by this round's change.

Findings:
- `F1` · minor · resolved — The regression scenario previously switched back immediately after `cd` into a non-git directory, so async cwd/git metadata could let it pass without the CodeTab fallback mounting and without the Resizable unmount path firing. (packages/tests/features/code-tab.feature:1116)

_codex on the rebuttal:_ F1 fixed. I verified the inserted step waits for `[data-testid="diff-no-repo"]` to become visible before the terminal switch-back, so the test now deterministically reaches the unmount path it is intended to cover.

srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `11fd6f36b531`

Independent findings: lowy=4, hickey=2

### Applied (3)
- `lowy-1` Corvu's leaky onSizesChange contract is decoded ad hoc at three call sites instead of behind one receptacle — commit `457e4ad7d`
- `lowy-2` App.tsx and CodeTab use asymmetric, non-equivalent guards for the identical Corvu unmount emission — (uncommitted)
- `hickey-1` Corvu degenerate-emission guard fragmented across three onSizesChange sites — (uncommitted)

### Agreed — no change (3)
- `lowy-3` Persist-side setters cannot defend the boundary they conceptually own (packages/client/src/right-panel/useRightPanel.ts:255)
- `lowy-4` E2e drag step hard-codes the testid + zero-height-handle workaround inline (packages/tests/step_definitions/code_tab_steps.ts:740)
- `hickey-2` CodeTab guard redundantly re-checks what setCodeTabTreeSize already clamps, but misses the actual failure mode without the length gate (packages/client/src/right-panel/CodeTab.tsx:748-749 and packages/client/src/right-panel/useRightPanel.ts:255-266)

srid:

Reviewed the settled tree (after the codex and lens rounds). Findings, all applied in cf3686a5:

- **`corvuResizable.ts` — honest narrowing.** The per-element `!== undefined` checks after `sizes.length === 2` were structurally unfailable for a `number[]`; they existed only to satisfy TypeScript. Replaced with an explicit `as [number, number]` tuple cast so the intent is visible instead of disguised as runtime defense.
- **Comment dedup at the three call sites.** The `realSizes` call-site comments re-narrated the helper's JSDoc. Trimmed to only the domain-specific notes: `App.tsx` keeps the `MIN_PANEL_SIZE = 0.05` collapsed-case gate explanation, `TerminalContent.tsx` keeps the `> 0.02` expand-transition guard, `CodeTab.tsx` needed nothing beyond the helper's own doc.
- **`code_tab_steps.ts` — `DIFF_CONTENT` selector constant** for the thrice-repeated `[data-testid="diff-content"]`, following the existing `DIFF_VIEW`/`FILE_VIEW` pattern.

Deliberately skipped: moving the shape guard into the preference setters (would complect Corvu's emission contract with per-domain clamp logic — the component boundary is the right home).

srid:
## Evidence

**The tree/preview split now survives a terminal round-trip.** The clip is the new regression scenario: preview a file in the Code tab, drag the split handle up (preview pane grows), open a second terminal and `cd` to a non-git directory (the Code tab unmounts into its "Not in a git repository" fallback — the exact moment that used to clobber the saved fraction), then switch back: the preview pane comes back at the dragged size, not the default.



▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=evidence-code-tab-split-survives.mp4

srid:

The **x86_64-linux** lane ran on **`kolu-ci-3`** (`dev-x86-64-linux-03`) — commit `e9f7a8d7`, **exit 0** ✓

- **Lane wall** (pipeline): **9m6s**
- **Wrapper wall** (incl. lease + nix-run startup): 40m46s

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 8m18s | ✓ |
| `_ci-setup` | 37s | ✓ |
| `ci::unit` | 25s | ✓ |
| `ci::pnpm-hash-fresh` | 21s | ✓ |
| `ci::atlas-sync` | 21s | ✓ |
| `ci::biome` | 15s | ✓ |
| `ci::fmt` | 14s | ✓ |
| `ci::smoke` | 12s | ✓ |
| `ci::install` | 11s | ✓ |
| `ci::home-manager` | 11s | ✓ |
| `ci::nix` | 7s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

PR #1255, opened 2026-06-09
kolu-tui attach: drive a terminal from the shell, detach with ~. (R-4 Phase 2)

srid:
**`kolu-tui attach <id>` takes over a live kolu terminal from the shell** — raw-tty passthrough until an ssh-style line-start `~.` detaches; kolu-server keeps the PTY, and re-attaching repaints the full scrollback instantly because the attach stream's first frame *is* the snapshot. No contract bump and no new transport: attach is a pure consumer of `ptyHostSurface` 2.1 over the Phase 1 unix socket, per the plan of record in [pty-daemon-tui](https://htmlpreview.github.io/?https://github.com/juspay/kolu/blob/r4-phase2-kolu-tui/docs/atlas/dist/pty-daemon-tui.html). Along the way, the second terminal client exposed that kolu's VT/device-query policy was one concept fragmented across three packages — so it now lives in a new zero-dep leaf, **`@kolu/terminal-protocol`**, that the browser, the pty-host, and the CLI all import.

```
your terminal (raw mode)                          kolu-server (unchanged*)
  stdin bytes ──▶ reply stripper ──▶ escape machine ──▶ terminal.write
  SIGWINCH ─────────────────────────────────────────▶ terminal.resize
  stdout ◀─── backpressured pump ◀── terminalAttach: snapshot, then deltas
  every exit path ──▶ ONE restore (un-raw + fixed mode-reset string)
```

### The five load-bearing decisions (from the Atlas note, echoed at their code sites)

- **Nothing is stolen.** Every Ctrl chord, every mid-line `~`, every paste byte reaches the inner program. The *only* interception is the escape char immediately after a newline: `~.` detach · `~~` literal · `~?` help, `--escape` to rebind. The machine runs on **bytes** (multibyte chars split across stdin chunks survive; UTF-8 decoding happens only at the `write` boundary) and suspends inside bracketed-paste markers, so a pasted `"\n~."` cannot detach.
- **Device-query replies are stripped, not forwarded.** The snapshot/deltas carry queries (DA1 · DSR · XTVERSION…) the user's real terminal auto-answers on stdin — but the headless mirror already answered them server-side, so the duplicate would corrupt the inner program (the yazi escape-soup bug). A raw tty gives no one-reply-per-read guarantee, so the CLI runs a **streaming, escape-sequence-boundary stripper** (replies split across reads or glued to keystrokes are still caught), with a fail-open cap so a user-typed `Alt+]` can never swallow the session.
- **Resize-then-attach, last-resize-wins.** The PTY is resized to the local window *before* attaching, so the snapshot serializes at the right grid. A concurrently-attached browser tile may show wrap artifacts until its own next resize — accepted and documented (a size-change tap would be contract 2.2).
- **One deterministic restore.** The snapshot replays modes onto the real terminal (alt-buffer, mouse tracking, bracketed paste, app cursor keys); detach, PTY exit, external signals, and crash all funnel through one idempotent `setRawMode(false)` + fixed reset string (synchronous, so it's safe from the process `exit` handler).
- **Stream end is discriminated, never auto-retried.** The deltas iterator ends identically on PTY exit, server abort, and the pty-host's silent slow-consumer drop. On end the loop asks the inventory: gone → fetch the real exit code from the exit tombstone and exit with it; still live → re-attach (the fresh snapshot repaint is exactly the right recovery for the drop). Detach first drains the ordered write queue, so `echo work⏎~.` lands `echo work` before the client leaves. Transport death restores the tty and prints an honest one-liner.

> *Server-side delta is error-shape only (no contract change): a missing PTY is a clean `NOT_FOUND` from one composed `requirePty` guard across attach, screen reads, and the metadata taps — with `exit` deliberately unguarded, since dead ids are its legitimate tombstone input.*

### `@kolu/terminal-protocol` — the policy gets one home

A blocking structural review made the boundary explicit exactly when the second client landed. The leaf (own [README](https://github.com/juspay/kolu/blob/r4-phase2-kolu-tui/packages/terminal-protocol/README.md)) owns, one module per concern:

| Module | Owns | Consumers |
| --- | --- | --- |
| `responseFilter` | reply grammars + whole-payload predicate (browser `onData`) + streaming stripper (raw tty) | `Terminal.tsx` · kolu-tui |
| `headlessReplies` | the server-side forward/drop rule for headless-generated answers | `pty-host` |
| `deviceQueries` | the **answered / uniformly-silent matrix as data** — executed against a real headless by pty-host's contract tests, so policy and implementation can't drift | pty-host tests |
| `bracketedPaste` | the `?2004` delimiters | escape scanner · server paste injection |
| `snapshotReset` | the reciprocal reset for every mode a replayed snapshot can enable | kolu-tui restore |

_`kolu-common` no longer carries VT internals, and the pty-host **staleKey** hashes the protocol sources (its closure guard walks the edge as a second hashed root) — a protocol change is observable daemon behaviour._

### Tests

| Layer | What's pinned |
| --- | --- |
| `escape.test.ts` (pure, no tty) | the ssh model: session-start = line-start, `~.`/`~~`/`~?`, chunk-boundary pending escape, split paste markers, byte-identical passthrough |
| `responseFilter.test.ts` (47 w/ the rest) | split/coalesced/keystroke-adjacent replies, bare-ESC latency, ESC-heavy fail-open tails |
| `attach.test.ts` (real unix socket) | spawn → snapshot paint → keystroke round-trip → `~.` detach with the PTY surviving → exit-code mirroring → same-burst write-before-detach ordering |
| `ptyHost.test.ts` device-query contract | every headless reply matches the suppressed grammars; the silent set stays silent; a colour query yields no garbage end-to-end |

> Review trail: two full serial gauntlets on this branch (codex⇄claude ×2 to consensus, lowy⇄hickey ×2 to consensus, /simplify, code-police) — findings and fixes are the `fix(…)`/`refactor:` commits and their PR comments below. Also ships the unlisted, noindex'd [kolu.dev/tui](https://kolu.dev/tui) quick-reference (alpha-bannered, linked from nowhere).

### Try it locally

```sh
nix run github:juspay/kolu/r4-phase2-kolu-tui#kolu-tui -- attach <id>
```


srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `4f8c3ced3067`

### Round 1

**codex** — approved: `false`

The branch is not ready as-is. The main shape is sound, but the new attach path has two correctness holes in the raw passthrough: terminal replies are filtered with a non-streaming predicate despite TTY chunking being arbitrary, and detach can exit before already-forwarded input reaches the PTY. I could not run the package checks because `pnpm` is not available on PATH in this environment.

Findings:
- `F1` · major · open — `kolu-tui attach` applies `isTerminalQueryResponse` to the entire stdin chunk and only suppresses it when the whole chunk is exactly one response. A real TTY stream does not preserve terminal-emulator write boundaries: replies can be split across reads, coalesced with other replies, or adjacent to user input. In those cases this code forwards the duplicate DA/DSR/XTVERSION bytes into the PTY, which is the exact escape-garbage bug this filter is meant to prevent. The comment in `kolu-common/terminalResponseFilter.ts:14` assumes discrete packets for raw tty reads, but that is not a guarantee Node/libuv or the pty line discipline gives you. (packages/pty-tui/src/attach.ts:164)
- `F2` · major · open — Detach aborts the attach stream immediately and the detach returns at `attach.ts:269`/`281` without waiting for the ordered `wire` queue to drain. If a user sends data immediately before `~.` in the same stdin burst, such as `echo work\r~.` or even the newline used to arm the escape, that write is only enqueued. `cmdAttach` then calls `process.exit(0)` at `packages/pty-tui/src/main.ts:205`, so the process can close before the write RPC reaches the server. This violates ssh-style escape ordering: bytes before the escape must be delivered before the local client leaves. (packages/pty-tui/src/attach.ts:149)

**claude** — Both findings were valid; I fixed both and added regression tests that genuinely fail on the un-fixed code.

F1 (filter wasn't stream-safe on the raw-tty path): added `createTerminalResponseStripper` to `packages/common/src/terminalResponseFilter.ts` — a stateful, escape-sequence-boundary filter that isolates each CSI/OSC/DCS sequence (reassembling one split across reads once its introducer is known), tests JUST that sequence against the same three response grammars, drops matches, and forwards everything else (keystrokes, arrow-key CSIs, program-output escapes, coalesced replies, reply+keystroke) untouched and in order. The whole-payload `isTerminalQueryResponse` predicate is kept for the browser `onData` path (which truly is one-event-per-call). `attach.ts` now pumps stdin through `stripper.push()` instead of the whole-chunk predicate. I also made a deliberate refinement CODEX's suggestion didn't call out: a *bare trailing ESC* is NOT held across a chunk boundary — holding it would merge `Esc` then `i` into `Alt-i` and break a remote editor's Escape-vs-Alt timeout; the cost is the vanishingly-rare reply that splits exactly between its ESC and introducer (no worse than today's behaviour). Tests cover split response, two coalesced responses, reply+keystroke adjacency on both sides, and the ESC-latency cases.

F2 (detach could exit before forwarded input reached the PTY): added `drainWire()` and routed every `detached` return through it, so `runAttach` only resolves `detached` after the ordered `wire` queue (which holds the writes enqueued for bytes-before-`~.`) has drained; a write that failed now surfaces as `error` instead of a false clean detach. This closes the `process.exit(0)` race in `main.ts` at its source (runAttach no longer resolves early).

- `F1` **fixed** — Agreed — the raw-tty path cannot assume read boundaries align with terminal write boundaries (the snapshot fires a BURST of device queries, so coalesced replies and reply+keystroke chunks are common). Added createTerminalResponseStripper (packages/common/src/terminalResponseFilter.ts:135+): a boundary-aware streaming filter that isolates each CSI/OSC/DCS sequence, buffers a partial sequence across chunks once its introducer (ESC [ / ESC ] / ESC P) is known, tests just that sequence against the existing three grammars, drops matches, and forwards everything else in order. attach.ts:189 now calls stripper.push(raw) instead of the whole-chunk isTerminalQueryResponse. Kept the whole-payload predicate for the browser onData path (genuinely one-event-per-call) and fixed the inaccurate doc comment at terminalResponseFilter.ts that claimed raw reads give one reply per kernel write. Partial concession-plus-improvement: I do NOT buffer a bare trailing ESC across a boundary (CODEX's suggestion implied buffering all partial prefixes) because that breaks interactive Escape-vs-Alt timing for a remote editor; documented the tradeoff. Tests added: split DA1 (ESC[?1 | ;2c), split OSC terminator (…ESC | \), two coalesced replies, reply+keystroke on both sides, surrounding input, arrow-key/Alt/OSC52 pass-through, and two ESC-latency tests (per-chunk forwarding).
- `F2` **fixed** — Agreed — on ~. the scanner emits [forward, detach]; the forward write is only ENQUEUED on the wire chain, then detach() aborts and runAttach returned { detached } without awaiting wire, so main.ts:205 process.exit(0) could fire before the write RPC reached the server (ssh-style ordering violation: echo work\r~.). Added drainWire() (attach.ts:165) which awaits the wire chain until its tail stops moving, then returns detached — or error if a write failed (transportError), instead of a silent clean detach. All three detached returns now go through drainWire(). Added an integration test that wraps the client with an artificially-slow terminal.write and asserts the write has LANDED by the time runAttach resolves detached; I verified it fails (writeLanded=false) when drainWire is reverted, so it's a true regression guard. Note the write RPCs are not carried by the attach AbortController, so aborting the stream does not cancel the in-flight writes — they complete and drain.

commit: `cca3dd2ece756c82ae2adeae57f4a4ae78e333ad`

### Round 2

**codex** — approved: `true`

I've verified the current fixes by inspection. The raw-tty path now uses a stateful streaming stripper with regression coverage for split, coalesced, and reply-adjacent inputs, and the detach path now drains the ordered wire queue before returning a clean detach.

Findings:
- `F1` · major · resolved — Resolved. The CLI no longer runs the whole-chunk terminal response predicate on raw tty reads; it now uses `createTerminalResponseStripper`, which isolates CSI/OSC/DCS sequences across chunk boundaries and drops only complete query-response matches while preserving adjacent real input. (packages/pty-tui/src/attach.ts:189)
- `F2` · major · resolved — Resolved. Detach now routes through `drainWire()`, so bytes enqueued before `~.` must finish their ordered `terminal.write` RPCs before `runAttach` reports `{ kind: "detached" }`; write failures surface as errors rather than false clean detaches. (packages/pty-tui/src/attach.ts:165)

_codex on the rebuttal:_ F1: accepted. Claude did not dispute the finding; the new streaming stripper addresses the raw tty boundary problem I raised. I also accept the deliberate choice not to hold a bare trailing ESC across pushes: it leaves one rare split-at-byte-one response leak, but avoids a worse interactive regression where Escape then a following key gets merged/delayed. F2: accepted. Claude did not dispute the finding; draining the existing wire promise chain before returning `detached` closes the process-exit race, and the slow-write test is a meaningful guard for the original failure mode.

srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `4f8c3ced3067`

### Applied (2)
- `hickey-1` Escape-char validity rule enforced twice with divergent failure shapes — commit `2a972c124`
- `hickey-2` not-found-vs-exited discrimination duplicated across two sites — commit `dee82f712`

### Agreed — no change (7)
- `lowy-1` VT-sequence isolation logic is duplicated across two byte-walking state machines (packages/common/src/terminalResponseFilter.ts:160-232 and packages/pty-tui/src/escape.ts:46-135)
- `lowy-2` NOT_FOUND failure-mode contract is implicit, encapsulated by convention not by a single receptacle (packages/pty-host/src/inProcessPtyHost.ts:64-69)
- `lowy-3` TTY_RESET mode-restore set is coupled to the snapshot's mode-enable set with no shared source (packages/pty-tui/src/attach.ts:49-57)
- `lowy-4` terminalResponseFilter relocation is a correct, leak-free receptacle (packages/common/src/terminalResponseFilter.ts:1-232)
- `hickey-3` silentLog test fixture duplicated across two packages (packages/pty-tui/src/attach.test.ts:28-34 and packages/pty-host/src/inProcessPtyHost.test.ts:21-27)
- `hickey-4` client-suppressed VT classes and the headless answerer set coupled only by prose (packages/common/src/terminalResponseFilter.ts:37-44 and packages/pty-host/src/ptyHost.ts (headless onData answerer))
- `hickey-5` runAttach closure braids four mutable control flags with the pump loop (packages/pty-tui/src/attach.ts:133-150 and 165-202)

srid:

Final polish pass over the post-debate tree (rules → fact-check → elegance). **One finding, fixed in b0475407**; everything else came back clean.

- **Fixed — dispatch exhaustiveness (altitude):** `main.ts`'s command dispatch ended in a bare `else → cmdAttach`. Unreachable today (cleye exits on unknown commands), but a Phase 3 `spawn` addition could silently fall through into attach. Documented the invariant at the site so the trap is visible to the next phase.
- **Checked, no change:** reuse (`writeOut` correctly shared; `delay` single-site), the 3-variant `ev.kind` dispatch in `attach.ts` (readable as-is), and the `number[] → Buffer.from` copies in the byte machines (negligible at interactive stdin rates — a ring buffer would add complexity for no measurable win).

All 33 pty-tui tests green after the fix.

srid:
## Evidence

This clip demonstrates the complete `kolu-tui attach` flow (R-4 Phase 2). It opens with `kolu-tui list` showing the live terminal table (ID · PID · IDLE · CMD · CWD), then `kolu-tui attach <id>` repaint the server-side scrollback — the pre-seeded "pre-existing scrollback" line appears instantly, proving the snapshot-restore path works. Inside the attached raw-passthrough shell the command `echo hello from kolu-tui attach` is typed and runs, demonstrating live stdin forwarding. Pressing Enter then `~.` triggers the line-start escape: the "— detached · <id> stays live in kolu-server" trailer prints and kolu-tui exits while the server keeps the PTY alive. A second `kolu-tui list` confirms the terminal is still there (IDLE clock merely incremented). Re-attaching repaints the full scrollback — including the earlier echo and its output — from the server's in-memory xterm state, then `~.` detaches again cleanly.



▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=evidence-kolu-tui-attach.mp4

> **Note:** The demo terminal was created through the pty-host contract's `spawn` procedure (the `kolu-tui spawn` CLI subcommand is Phase 3); everything else — socket dial, `list`, `attach`, write/resize/exit events, and the `~.` detach sequence — is the shipping Phase 2 path (`serveOverUnixSocket` → `unixSocketLink` → `ptyHostSurface`).

srid:

The **x86_64-linux** lane ran on **`kolu-ci-8`** (`dev-x86-64-linux-08`) — commit `b0475407`, **exit 0** ✓

- **Lane wall** (pipeline): **9m44s**
- **Wrapper wall** (incl. lease + nix-run startup): 51m34s

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 8m46s | ✓ |
| `ci::home-manager` | 1m6s | ✓ |
| `ci::nix` | 54s | ✓ |
| `ci::smoke` | 50s | ✓ |
| `_ci-setup` | 44s | ✓ |
| `ci::unit` | 27s | ✓ |
| `ci::biome` | 17s | ✓ |
| `ci::fmt` | 16s | ✓ |
| `ci::install` | 14s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

srid:

The **x86_64-linux** lane ran on **`kolu-ci-8`** (`dev-x86-64-linux-08`) — commit `06f4d125`, **exit 1** ✗

- **Lane wall** (pipeline): **9m12s**
- **Wrapper wall** (incl. lease + nix-run startup): 36m39s

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 8m30s | ✓ |
| `ci::home-manager` | 33s | ✓ |
| `_ci-setup` | 28s | ✓ |
| `ci::unit` | 23s | ✓ |
| `ci::nix` | 19s | ✓ |
| `ci::biome` | 17s | ✓ |
| `ci::smoke` | 15s | ✓ |
| `ci::install` | 14s | ✓ |
| `ci::fmt` | 13s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

srid:

The **x86_64-linux** lane ran on **`kolu-ci-8`** (`dev-x86-64-linux-08`) — commit `caf5e98f`, **exit 1** ✗

- **Lane wall** (pipeline): **9m12s**
- **Wrapper wall** (incl. lease + nix-run startup): 12m42s

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 8m30s | ✓ |
| `ci::home-manager` | 33s | ✓ |
| `_ci-setup` | 28s | ✓ |
| `ci::unit` | 23s | ✓ |
| `ci::nix` | 19s | ✓ |
| `ci::biome` | 17s | ✓ |
| `ci::smoke` | 15s | ✓ |
| `ci::install` | 14s | ✓ |
| `ci::fmt` | 13s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

srid:

The **x86_64-linux** lane ran on **`kolu-ci-8`** (`dev-x86-64-linux-08`) — commit `edbc105a`, **exit 0** ✓

- **Lane wall** (pipeline): **2m36s**
- **Wrapper wall** (incl. lease + nix-run startup): 6m4s

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 1m42s | ✓ |
| `_ci-setup` | 42s | ✓ |
| `ci::home-manager` | 38s | ✓ |
| `ci::unit` | 21s | ✓ |
| `ci::nix` | 19s | ✓ |
| `ci::smoke` | 14s | ✓ |
| `ci::install` | 11s | ✓ |
| `ci::biome` | 10s | ✓ |
| `ci::fmt` | 9s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/edbc105/timings.jsonl`; pool state is a live `flock` probe.</sub>

srid:
## 🏗 Structural follow-up — `@kolu/terminal-protocol` (02bb180d)

A blocking structural review (codex, relayed by the author) overturned this PR's earlier lens-debate verdict: the **VT/device-query policy** — which terminal query replies are answered, dropped, or forwarded — was one concept fragmented across `kolu-common` (grammars + stripper), `pty-host` (forward/drop rule inline), and `pty-tui` (TTY reset, paste bytes), held in lockstep by prose. Adding the *second* terminal client is exactly when that boundary should become explicit, so it now is:

- **New zero-runtime-dep leaf `packages/terminal-protocol`**, one module per concern: the response-filter grammars + both suppression entry points · `shouldForwardHeadlessReply` (the headless forward/drop rule, formerly inline in `ptyHost.ts`) · the **answered/silent device-query matrix as data** (pty-host's contract tests now execute the imported table against a real headless, so policy and implementation can't drift) · the bracketed-paste delimiters · `SNAPSHOT_TTY_RESET`.
- **Browser `Terminal.tsx`, `pty-host`, and `pty-tui` all import it**; `kolu-common` stops carrying VT internals; `escape.ts` keeps only the ssh-style attach machine.
- **The pty-host staleKey hashes the protocol package too** (`default.nix` + the `buildId` closure guard walks the edge as a second hashed root) — a protocol change is observable daemon behaviour and must flip the key. The closure test caught this dependency the moment it appeared, which is exactly its job.
- `pnpmDeps` hash verified fresh via `--rebuild`; full suites green (terminal-protocol 41, pty-host 58, pty-tui 33, common 17).

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 round(s) · codex reviewed at `xhigh` reasoning effort · base `41748f38f1be`

### Round 1

**codex** — approved: `false`

I found one Nix/build blocker and several terminal-passthrough correctness issues. I could not run package typechecks because `pnpm` is not available on PATH in this environment, so this review is based on direct diff/context inspection.

Findings:
- `F1` · blocking · open — The main workspace `src` fileset omits the new `packages/terminal-protocol` package. That fileset is explicitly the source for Nix build/typecheck, and the new package has a `typecheck` script plus runtime imports from client, pty-host, and pty-tui. In the Nix source tree those workspace links point at a package directory that is not copied, so Nix checks/builds can fail resolving `@kolu/terminal-protocol` or skip the package's own type gate. (default.nix:27)
- `F2` · major · open — The CSI suppression regex drops every `CSI ... R`, which includes real modified F3 keypresses such as `ESC [ 1 ; 2 R` from xterm-compatible terminals, not just CPR replies. In the new raw attach path these bytes pass through `createTerminalResponseStripper` before `terminal.write`, so Shift/Ctrl/Alt-F3 is silently eaten despite the feature promising raw passthrough of every key/chord. (packages/terminal-protocol/src/responseFilter.ts:66)
- `F3` · major · open — The streaming stripper treats any `ESC ]` or `ESC P` from raw stdin as the start of an OSC/DCS string and then buffers until BEL/ST. Those byte pairs are also valid user input, e.g. Alt+] or Alt+Shift+P. If the user presses one, the attach path forwards nothing and then keeps appending subsequent keystrokes to `seq` indefinitely until a terminator happens to appear, effectively freezing input and allowing unbounded buffer growth. (packages/terminal-protocol/src/responseFilter.ts:209)
- `F4` · minor · open — The new `BRACKETED_PASTE_START`/`END` constants are not used by the existing server paste-injection path, even though the new package comments say the server, TUI scanner, and reset share these bytes from one home. The delimiters still exist as a hard-coded duplicate in `bracketedPastePath`, so the intended protocol source of truth can drift immediately. (packages/server/src/router.ts:64)

**claude** — Addressed all 4 CODEX findings. F1 (Nix fileset) and F4 (server constant duplication) were real and fixed. F3 (unbounded string-sequence buffering on user-typed ESC]/ESC P) was a genuine freeze/DoS bug — fixed with a fail-open cap plus tests. F2 (CPR vs modified-F3 byte collision) is a fundamental, irreducible VT ambiguity the browser path already shares by design; disputed the framing while implementing the actionable safety fix in the same area (F3). All 45 terminal-protocol tests pass; biome format and nixpkgs-fmt report the touched files clean.

- `F1` **fixed** — Confirmed correct: packages/terminal-protocol has a `typecheck` script (package.json:14) and is a workspace dependency of client, pty-host, and pty-tui (all imported via `@kolu/terminal-protocol`), yet `./packages/terminal-protocol` was missing from the main `src` fileset in default.nix — violating the documented INVARIANT at default.nix:21-26 (every workspace package with a typecheck script must be in this fileset). The separate `ptyHostSrc` block (default.nix:100-104) only feeds the build-id hash, not the workspace `src` used by the build/typecheck derivations. Added `./packages/terminal-protocol` to the fileset (default.nix:48) between memorable-names and pty-host; left ptyHostSrc untouched as the stale-key hash input, exactly as CODEX suggested.
- `F2` **partial** — Fixed the actionable safety part (see F3) and dispute the rest. The CPR/modified-F3 collision is real but irreducible: CPR is `CSI Ps;Ps R` and Shift+F3 is `CSI 1;2 R` — identical byte forms, indistinguishable without query-context tracking. This collision is PRE-EXISTING and shared with the browser path: the same `isTerminalQueryResponse` predicate (responseFilter.ts:99) drives `Terminal.tsx` onData, which has always eaten modified-F3 the same way. The atlas design contract (docs/atlas/src/content/atlas/pty-daemon-tui.mdx:159) explicitly mandates the attach path 'mirror the browser path's suppression predicates' and preserve the one-suppression-set invariant (responseFilter.ts:11-13), so this is a deliberate parity decision, not a regression. Un-suppressing CPR to save modified-F3 would resurrect the exact yazi escape-soup bug this module exists to kill (programs query cursor position constantly). CODEX's query-tracker fix is the only way to have both, but it requires the stripper to observe the OUTPUT stream (it currently sees only stdin) — a meaningful redesign out of scope for this change, and one that would itself break the documented browser/attach suppression-set parity. Disputing the 'silently eaten despite promising raw passthrough' framing while conceding the same area needed the F3 robustness fix.
- `F3` **fixed** — Confirmed a genuine freeze/unbounded-growth bug. `ESC ]` (Alt+]) and `ESC P` (Alt+Shift+P) are real keystrokes that share their introducer with the OSC/DCS reply grammars; responseFilter.ts:209-210 puts the stripper into `{kind:'string'}` on them, and that state only exits on BEL or ST — so a user pressing Alt+] then continuing to type buffered every keystroke into `seq` forever, freezing input. Added a fail-open cap (MAX_STRING_SEQ=8192, responseFilter.ts) and a `failOpenSequence()` helper: once a string sequence outruns any legitimate reply without terminating, the buffered bytes are flushed to the PTY untouched and state resets. Scoped to the string branch only — the CSI branch can't freeze because it terminates at any final byte 0x40-0x7e, which ordinary typing hits almost immediately. The cap is generous so the tiny suppressible replies (colour ~30B, XTVERSION/DECRQSS ~20B) are unaffected, and a large *forwarded* OSC 52 clipboard read fails open rather than being eaten. Added 4 tests: Alt+] passthrough, Alt+Shift+P passthrough, genuine short OSC colour reply still suppressed, and large OSC 52 reply still forwarded past the cap. All 45 tests pass.
- `F4` **fixed** — Confirmed: server/src/router.ts:65 hard-coded `\x1b[200~${path}\x1b[201~`, duplicating BRACKETED_PASTE_START/END despite bracketedPaste.ts:4-7 and index.ts:14 both claiming the server shares these bytes from terminal-protocol's single home. Imported the constants into router.ts (top of the import block, `@kolu` before `@orpc` per the file's existing ordering) and rewrote bracketedPastePath to use them. Added `@kolu/terminal-protocol: workspace:*` to packages/server/package.json (alphabetically after @kolu/surface-app) and the matching `link:../terminal-protocol` entry to the server's deps block in pnpm-lock.yaml (mirroring the existing pty-host entry), so the now-direct dependency is declared rather than relied on transitively through @kolu/pty-host.

commit: `5d2eaf3a283063b3f26a0353600388b6160e1bb6`

### Round 2

**codex** — approved: `false`

Three of the four findings are resolved in the current tree. The Nix fileset now includes the new workspace package, the bracketed paste constants are shared by the server with a declared dependency, and I accept the CPR/modified-key ambiguity as an intentional protocol tradeoff. F3 is not fully resolved: the fail-open cap does not run for every byte in OSC/DCS string mode, so some unterminated input patterns can still buffer forever.

Findings:
- `F1` · blocking · resolved — `packages/terminal-protocol` was omitted from the main Nix workspace fileset even though it has a `typecheck` script and is a workspace dependency, so Nix build/typecheck inputs could miss it. (default.nix:48)
- `F2` · minor · resolved — The CPR suppression grammar can also match some modified key sequences, so a perfectly raw passthrough guarantee is not literally true for every VT byte sequence. (packages/terminal-protocol/src/responseFilter.ts:66)
- `F3` · major · open — The new fail-open cap for unterminated OSC/DCS strings is incomplete. The cap is only checked in the final `else if (seq.length >= MAX_STRING_SEQ)` branch, so bytes handled by `b === ESC` or by the `pending.escSeen` branch bypass it. After `ESC ]`, a stream of repeated ESC bytes, or repeated `ESC` plus non-backslash bytes, can still keep `pending.kind === "string"`, grow `seq`, and freeze forwarded input indefinitely. (packages/terminal-protocol/src/responseFilter.ts:253)
- `F4` · minor · resolved — The server still hard-coded bracketed-paste delimiters even though `@kolu/terminal-protocol` is meant to be the shared source of truth for those bytes. (packages/server/src/router.ts:69)

_codex on the rebuttal:_ F1: verified fixed. `default.nix` now includes `./packages/terminal-protocol` in the main workspace fileset. F2: I accept the dispute. The collision is inherent without output/query-context tracking, and the shared suppression-set invariant is a reasonable tradeoff for this PR. F3: the fix addresses ordinary unterminated text after `ESC]`/`ESC P`, but not all unterminated string-mode inputs; ESC-heavy tails bypass the cap because the length check is branch-local. Holding this open. F4: verified fixed; the server now uses the shared constants and declares the direct dependency.

**claude** — Round 2: F3 was the only finding still open; CODEX was correct that the fail-open cap was branch-local. Fixed by hoisting the cap check out of the byte-dispatch chain so it runs on every byte while still in string mode, plus added the two ESC-heavy-tail tests CODEX asked for. F1, F2, F4 were already resolved/accepted in CODEX's verdict — no further action. All 47 terminal-protocol tests pass; biome reports the touched files clean.

- `F1` **fixed** — Already resolved and verified-fixed by CODEX in this verdict. packages/terminal-protocol is in the main src fileset at default.nix:48. No new action this round.
- `F2` **disputed** — Accepted-as-disputed by CODEX. The CPR (CSI Ps;Ps R) vs modified-F3 (CSI 1;2 R) byte collision is an irreducible VT ambiguity in an input-only filter, and mirroring the browser path's suppression set (responseFilter.ts:66, documented invariant at lines 11-13) is the deliberate parity choice. No code change; nothing further to litigate.
- `F3` **fixed** — CODEX was right: the cap was branch-local. In the string-mode dispatch (responseFilter.ts ~250-262) the bytes b===BEL / pending.escSeen / b===ESC each took an earlier branch, so the `seq.length >= MAX_STRING_SEQ` check in the final else-if never ran for ESC-heavy tails. After `ESC ]` a run of ESC bytes alternates escSeen true/false on every byte (branch 3 then branch 2), and repeated `ESC x` pairs do the same — both grow seq forever and freeze input. Fixed by removing the cap from the else-if chain and adding a separate post-handling check: `if (pending.kind === "string" && seq.length >= MAX_STRING_SEQ) failOpenSequence();`, which runs on every byte that is still in string mode after terminator handling. Added two tests: `ESC ]` + 9000 ESC bytes, and `ESC P` + 5000 `ESC z` pairs, both asserting full passthrough (fail-open). All 47 tests pass.
- `F4` **fixed** — Already resolved and verified-fixed by CODEX. router.ts imports BRACKETED_PASTE_START/END, packages/server/package.json declares @kolu/terminal-protocol, and the lockfile has the workspace link. No new action this round.

commit: `e84211c611765ab81f0e2e44b9661f714accb1d3`

### Round 3

**codex** — approved: `true`

The remaining open issue, F3, is now resolved. The cap check is hoisted out of the string-mode byte-dispatch branches and runs after terminator handling on every byte that remains in string mode, so ESC-heavy unterminated OSC/DCS tails no longer bypass the fail-open path. The requested regression tests for repeated ESC bytes and repeated ESC+non-backslash pairs are present. F1, F2, and F4 remain resolved from the prior round.

Findings:
- `F1` · blocking · resolved — `packages/terminal-protocol` was omitted from the main Nix workspace fileset even though it has a `typecheck` script and is a workspace dependency, so Nix build/typecheck inputs could miss it. (default.nix:48)
- `F2` · minor · resolved — The CPR suppression grammar can also match some modified key sequences, so a perfectly raw passthrough guarantee is not literally true for every VT byte sequence. (packages/terminal-protocol/src/responseFilter.ts:66)
- `F3` · major · resolved — The fail-open cap for unterminated OSC/DCS strings was branch-local, so ESC-heavy tails could still keep `pending.kind === "string"`, grow `seq`, and freeze forwarded input indefinitely. (packages/terminal-protocol/src/responseFilter.ts:264)
- `F4` · minor · resolved — The server hard-coded bracketed-paste delimiters even though `@kolu/terminal-protocol` is meant to be the shared source of truth for those bytes. (packages/server/src/router.ts:69)

_codex on the rebuttal:_ F1: no further issue; still verified fixed. F2: no further dispute from my side; the prior acceptance stands. F3: Claude's correction addresses the exact remaining hole: the post-dispatch `pending.kind === "string" && seq.length >= MAX_STRING_SEQ` check catches ordinary tails, all-ESC tails, and repeated `ESC x` tails after allowing BEL/ST termination first. Marking resolved. F4: no further issue; still verified fixed.

srid:

✅ **Consensus** after 2 round(s) · lowy + hickey · base `41748f38f1be`

### Applied (4)
- `lowy-2` Stale cross-reference to deleted terminalResponseFilter.ts — commit `493e353a5`
- `lowy-3` TTY_RESET re-export in attach.ts is a thin alias, but it is the right interface-stability call — commit `bf7b44544`
- `hickey-1` NOT_FOUND existence-guard scattered across three handlers — commit `1cce802d0`
- `hickey-2` TTY_RESET re-export gives SNAPSHOT_TTY_RESET a second name — commit `d744481aa`

### Agreed — no change (5)
- `lowy-1` responseFilter ships two suppression axes through one module — but the package's coherence holds (packages/terminal-protocol/src/responseFilter.ts:1)
- `lowy-4` default.nix ptyHostSrc root widened from pty-host to packages/ — staleKey boundary stays honest (default.nix:91)
- `lowy-5` deviceQueries matrix is data, consumed only by pty-host tests — single test-only consumer is not a reuse-signal failure (packages/terminal-protocol/src/deviceQueries.ts:32)
- `hickey-3` Suppression grammars and the answered/silent matrix are two representations of one set (packages/terminal-protocol/src/responseFilter.ts:66 and packages/terminal-protocol/src/deviceQueries.ts:32)
- `hickey-4` Snapshot line-count derived two different ways (packages/pty-tui/src/attach.ts:272 and packages/pty-tui/src/main.ts:128)

srid:
## [👮 Code-police](https://agency.srid.ca/) + 🧹 /simplify (serial gauntlet, passes 3–4)

**/simplify** (4 parallel cleanup agents over the merge-base diff, commit `532869a1`): reuse `@kolu/surface/client`'s `SURFACE_STDIO_TRANSPORT_CLOSED` constant instead of a string copy · `finishSequence` drops its `number[]`→`Buffer`→string round-trip · `escape.ts`'s two copy-pasted marker-matcher steps share one `advanceMarker` · `main.ts`'s dispatch is now closed (explicit `attach` branch + fail-loud `else`, replacing the comment that documented the fallthrough risk) · the four metadata tap streams compose `requirePty` (clean `NOT_FOUND` on a bad id) with `exit` deliberately unguarded — dead ids are its legitimate tombstone input. Skipped with reasons: scratch-buffer accumulators (complexity > win at interactive sizes), per-re-attach `terminal.list` (needs a surface addition), generalizing the closure-guard registry (no second hashed package exists), `describeError`'s regex fallback (the deeper fix is a surface error-vocabulary contract).

**Code-police** (rules → fact-check → elegance, commit `2e9e5ca5`): one finding — the end-of-chunk bare-ESC flush inlined the same forward+reset triplet `failOpenSequence` owns; now a call. Everything else clean; the rule-of-three skips are listed in the commit.

All suites green after each pass (terminal-protocol 47 · pty-host 58 · pty-tui 33), `just check` exit 0.

_Generated by [`/be-review`](https://github.com/srid/agency) on Claude Code (model `claude-fable-5`)._

srid:

The **x86_64-linux** lane ran on **`kolu-ci-8`** (`dev-x86-64-linux-08`) — commit `2e9e5ca5`, **exit 1** ✗

- **Lane wall** (pipeline): **9m12s**
- **Wrapper wall** (incl. lease + nix-run startup): 8m12s

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 8m30s | ✓ |
| `ci::home-manager` | 33s | ✓ |
| `_ci-setup` | 28s | ✓ |
| `ci::unit` | 23s | ✓ |
| `ci::nix` | 19s | ✓ |
| `ci::biome` | 17s | ✓ |
| `ci::smoke` | 15s | ✓ |
| `ci::install` | 14s | ✓ |
| `ci::fmt` | 13s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

Issue #1240, opened 2026-06-09
Support for multiple forges (not just GitHub)


ic4-y:
### Context

Kolu's PR metadata integration is hardcoded to GitHub. The `kolu-github` package shells out to `gh pr view`, which only resolves PRs on repositories whose remotes point to a known GitHub host. When a terminal's repo is hosted on any other forge (Forgejo, GitLab, Gitea, SourceHut, etc.), `gh pr view` fails with:

```
none of the git remotes configured for this repository point to a known GitHub host.
```

This is classified as `not-authenticated` and logged as `WARN` on every 30s poll cycle — noisy and misleading for users who simply aren't on GitHub.

### Current Architecture

- `packages/integrations/github/` — `gh`-CLI-based PR resolution, tightly coupled to GitHub's API surface
- `packages/server/src/terminalBackend/providers.ts:startGitHubPrProvider` — wires the GitHub watcher into the terminal metadata pipeline
- Remote forge detection does not exist — the provider is always `github-pr` regardless of the actual remote URL

### Proposed Direction

Introduce a forge abstraction so kolu can resolve PR metadata across multiple forges:

1. **Forge detection** — parse the remote URL's host to identify the forge type (GitHub, GitLab, Forgejo/Gitea, etc.). Fall back to `absent` when the remote is not a recognized forge, and suppress PR-related logging in that case.

2. **Provider interface** — define a common `PrProvider` interface (resolve PR by repo+branch, return number/title/url/state/checks) that each forge adapter implements. The existing `kolu-github` package becomes the GitHub adapter.

3. **Dispatch** — `subscribeGitHubPr` (or its successor) picks the right adapter based on the detected forge, or emits `absent` when no adapter matches.

4. **Forgejo/Gitea** — these expose a GitHub-compatible API surface, so the existing `gh`-based approach may work with `--host` configuration, or a lightweight adapter using the Forgejo API directly.

5. **GitLab** — requires a separate adapter using the GitLab REST/GraphQL API.

### Motivation

Self-hosted forges (Forgejo, Gitea, GitLab) are common in the Nix/NixOS ecosystem. Kolu's user base skews technical and many will not be on GitHub. PR metadata is a visible part of the terminal dock (the PR pill), and losing it entirely on non-GitHub repos degrades the experience.

### Scope

[...]

PR #1256, opened 2026-06-09
Stop the false 'gh: not authenticated' warning on non-GitHub repos

srid:
**A terminal in a Forgejo/Codeberg/GitLab repo no longer shows a lying "gh: not authenticated" warning** (and no longer WARN-logs it every 30s per terminal). `gh pr view` refuses non-GitHub remotes with *"none of the git remotes configured for this repository point to a known GitHub host. To tell gh about a new GitHub host, please use `gh auth login`"* — and because that message itself mentions `gh auth login`, `classifyGhError`'s auth substring check stole it and classified "you're not on GitHub" as "you're not authenticated". The refusal is now matched first and classified `absent` — the same silent UI outcome as a branch with no PR.

Reproduced test-first: the new table-test row pins gh's refusal string (red before the fix — it landed in `not-authenticated`), alongside the existing tripwire tests that pin gh's other unversioned stderr strings.

> This is **phase 0a of the [anyforge plan](https://github.com/juspay/kolu/blob/master/docs/atlas/src/content/atlas/anyforge.mdx)** ([rendered](https://htmlpreview.github.io/?https://github.com/juspay/kolu/blob/master/docs/atlas/dist/anyforge.html)) toward #1240 — deliberately shipped alone ahead of the forge-abstraction refactor (0b) and the Forgejo adapter (1). _Credit: fix + test adopted from ic4-y's draft (https://github.com/ic4-y/kolu/pull/1)._


srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `e6b227bdd5e9`

### Round 1

**codex** — approved: `false`

The branch is a small targeted classifier fix with matching docs/changelog updates, but the new stderr predicate is broader than the gh message it intends to handle. That turns some real repository/host configuration failures into silent `absent` PR state, which is the same class of bug this PR is trying to avoid.

Findings:
- `F1` · major · open — `stderr.includes("none of the git remotes")` is overbroad. gh has other repository-resolution errors starting with the same words, e.g. remotes that do not correspond to an environment-selected host or to a valid remote repository. Those are real configuration/lookup failures, not "no PR on this branch". This change would now suppress them as `{ kind: "absent" }`, hiding the warning/error UI and debug signal from users. (packages/integrations/github/src/github.ts:137)

**claude** — Verified and fixed CODEX's single finding (F1). Confirmed against real gh 2.92.0 and gh's source (pkg/cmd/factory/remote_resolver.go) that two distinct stderr messages share the "none of the git remotes" prefix: (1) "...point to a known GitHub host..." (genuinely absent — repo isn't on GitHub) and (2) "...correspond to the GH_HOST environment variable..." (a real GH_HOST misconfiguration the user should see). The original broad predicate stderr.includes("none of the git remotes") swallowed BOTH as {kind:"absent"}, hiding the config failure — exactly the silent-failure class this PR set out to fix. Narrowed the predicate to stderr.includes("point to a known github host"); the GH_HOST variant now falls through every branch to ghUnavailable("unknown") since it contains none of the auth/no-PR substrings. Added the requested regression test asserting the GH_HOST-mismatch case stays unavailable. Synced the Atlas note (anyforge.mdx) and rebuilt dist/, updated wording in two places, and ran just fmt. github package unit tests: 48 passing. Left all changes uncommitted in the working tree per instructions.

- `F1` **fixed** — Agreed. Reproduced both gh messages directly with gh 2.92.0 and confirmed via gh source (pkg/cmd/factory/remoteResolver) that the 'none of the git remotes' prefix is shared by two messages: '...point to a known GitHub host...' (truly absent) and '...correspond to the GH_HOST environment variable...' (a real config failure). The broad includes('none of the git remotes') at github.ts:137 misclassified the GH_HOST case as absent, hiding it. Changed the predicate to stderr.includes('point to a known github host') (packages/integrations/github/src/github.ts:144) and expanded the comment to explain the GH_HOST distinction. The GH_HOST variant now correctly falls through to ghUnavailable('unknown') because it matches none of the not-logged-in/authentication/gh-auth-login/no-pull-requests substrings. Added the regression test CODEX asked for at packages/integrations/github/src/github.test.ts:197 ('keeps gh's GH_HOST-mismatch refusal as unavailable, not absent'), verifying it resolves to {kind:'unavailable', source:{provider:'gh', code:'unknown'}}. Updated the Atlas note (anyforge.mdx, table row + the 0a detail section) to describe the narrowed match and the GH_HOST caveat, rebuilt docs/atlas/dist/, and ran just fmt. github package unit tests pass (48/48).

commit: `5409c9264709a0651b3bb5520499cffb254010df`

### Round 2

**codex** — approved: `true`

CLAUDE addressed the only finding from the prior round. The classifier now matches the specific non-GitHub-host refusal instead of the shared `none of the git remotes` prefix, and the added GH_HOST-mismatch regression test covers the failure mode I raised. I did not identify any regression introduced by this round's changes.

Findings:
- `F1` · major · resolved — The previous broad `none of the git remotes` predicate could silently classify GH_HOST/remote-resolution configuration failures as `absent`. The current code now narrows the absent case to `point to a known github host`, leaving the GH_HOST mismatch to fall through as unavailable. (packages/integrations/github/src/github.ts:144)

_codex on the rebuttal:_ CLAUDE did not dispute F1; they accepted it and implemented the requested narrow match plus a GH_HOST-mismatch regression test. I verified the current diff: the classifier now checks `stderr.includes("point to a known github host")`, and the GH_HOST string is asserted to return `{ kind: "unavailable", source: { provider: "gh", code: "unknown" } }`. This resolves the finding.

srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `e6b227bdd5e9`

### Agreed fixes — handed off to the caller (2)
- `lowy-1` New refusal branch and 'no git remotes' branch encapsulate the same volatility axis but sit on opposite sides of the auth gate (packages/integrations/github/src/github.ts:137-154)
- `hickey-1` absent-vs-auth ordering is a comment-enforced invariant (packages/integrations/github/src/github.ts:133-146)

### Agreed — no change (3)
- `lowy-2` 'none of the git remotes' is a prefix of the broader real message — substring match is correct but the test pins a strictly longer string than the matcher needs (packages/integrations/github/src/github.test.ts:187-195)
- `lowy-3` Atlas note status flip and changelog ride-along do not introduce or shift any volatility boundary (docs/atlas/src/content/atlas/anyforge.mdx:5-6, website/src/content/changelog/unreleased.mdx:12)
- `hickey-2` Three distinct stderr causes, one untyped `absent` literal repeated inline (packages/integrations/github/src/github.ts:137-154)

### Applied by /be-review

| Change request | Disposition |
| --- | --- |
| `lowy-1` + `hickey-1` — group the absent-classifying substring checks above the auth gate (one converged plan) | applied — `3072e025` (adapted to the post-codex tree: the first substring is now `point to a known github host` after codex narrowed the match mid-debate) |

_The lenses reviewed a pinned snapshot while the codex debate edited the branch; the apply pass re-validated each request against the post-codex tree before implementing._

srid:

Three read-only passes (rule checklist · fact-check · elegance) ran against a pinned snapshot while the codex debate edited the branch; findings were re-validated against the post-codex tree before applying. 24 of 26 rules passed with checked-by evidence; two fired:

| Finding | Severity | Disposition |
| --- | --- | --- |
| **F1** — the matched refusal also fires for an *unauthenticated GitHub Enterprise* remote (gh's known-host set = authed hosts + github.com), where the old `not-authenticated` warning was correct; the comment/docs denied this false-negative | medium | **applied** — `4d4b4a20` documents the trade-off in the classifier comment and the Atlas note's 0a section; phase 0b's remote-URL-based `detectForge` owns restoring the distinction (indistinguishable on stderr, so no 0a code fix exists) |
| **F2** — `none of the git remotes` substring also matched gh's GH_HOST-mismatch error, silencing a real misconfiguration | low | **already fixed by codex** — `5409c926` narrowed the match to `point to a known github host` + regression test |
| **F3** — third `absent` branch; ordering invariant comment-enforced | low | **applied via the lens fix** — `3072e025` (same converged plan) |
| **F4** — three identical-shape `it` blocks vs the file's `it.each` house style | info | **applied** — `b957b69c` |

Fact-check also verified: the pinned stderr matches gh trunk verbatim, the changelog's "every 30 seconds" claim is accurate (`POLL_INTERVAL_MS = 30_000`), and the PR number is consistent across changelog and Atlas note.

srid:
## Evidence

On **master**, a terminal whose repo has a non-GitHub remote (Codeberg, set via `git remote set-url origin https://codeberg.org/srid/example.git`) shows a false **"gh: not authenticated"** warning icon in the tile title bar. Clicking it opens a popover with `gh auth login` recovery copy — even though the user doesn't need to authenticate, `gh` just can't reach a GitHub host. On the **PR branch** (`anyforge-0a`), `gh`'s "none of the git remotes point to a known GitHub host" refusal is now classified `absent`, so no warning icon renders — the tile chrome is silent, exactly like a branch with no PR.

| Before (master) — false warning on Codeberg remote | After (anyforge-0a) — tile silent, no warning |
|---|---|
|  |  |

PR #1257, opened 2026-06-10
Extract the forge-neutral PR kernel into a new anyforge leaf

srid:
Phase 0b of the [anyforge plan](https://github.com/juspay/kolu/blob/master/docs/atlas/dist/anyforge.html) for #1240. The forge-neutral PR machinery that grew inside `kolu-github` moves to a new leaf package, **`anyforge`** — mirroring what `anyagent` is for agent integrations. `kolu-github` shrinks to a pure gh adapter. **This phase is GitHub-only and the kernel names no forge**; the Forgejo/Codeberg adapter *and the forge detection it needs* are phase 1.

### Decomposition — faithful to the `AgentInfo` precedent

The split is exactly `AgentInfoShape` → `ClaudeCodeInfoSchema` → `AgentInfoSchema`:

| Layer | Owns | Mirrors |
| --- | --- | --- |
| **`anyforge`** (leaf) | Neutral shapes (`PrInfo`, `CheckRun`, `PrState`, …), a **generic** `PrResult<S>` / `PrProvider<S>` / `subscribePr<S>` over `PrUnavailableSourceBase = { provider: string; code: string }`, and `prResultEqual`/`prValue`/`prLabel`. **`PrProvider.kind: string` — names no forge.** | `AgentInfoShape` (`kind: string`), generic `AgentProvider<Session, Info>` |
| **`kolu-github`** (adapter) | Its own failure vocabulary — `GhUnavailableCodeSchema`, `GhUnavailableSchema`, `reasonForGhCode` — on a browser-safe `kolu-github/schemas` arm. `githubPrProvider: PrProvider<GhUnavailableSource>`. | `ClaudeCodeInfoSchema` in `kolu-claude-code/schemas` |
| **`kolu-common`** (app) | Composes the **closed** `PrUnavailableSource` union + the wire `PrResultSchema` + `reasonForSource`, right next to `AgentInfoSchema`. | `AgentInfoSchema = discriminatedUnion("kind", [ClaudeCodeInfoSchema, …])` |

The adapter produces `PrResult<GhUnavailableSource>`, a member of the closed union → assignable to the wire `PrResult` covariantly, **no cast**. The client's exhaustive `match(source.provider).exhaustive()` guarantee (a new forge = compile error at every render site) is preserved — just at the layer that composes the union.

`startPrProvider` **injects `githubPrProvider`** into `subscribePr` — one watcher per terminal, no registry, no detection. A non-GitHub (e.g. Codeberg) remote behaves exactly as today: gh refuses → silent `absent`.

### Deliberately deferred to phase 1

Forge **detection** is only meaningful once a second forge exists, so `GitInfo.remoteUrl`, `detectForge`, and the `ForgeKind → PrProvider` registry land in phase 1, server-side, with the Forgejo adapter — not in a GitHub-only kernel where they'd be a field nothing reads and a one-entry registry.

### One behavior change (changelog'd)

The review gauntlet found a latent bug carried verbatim by the old `subscribeGitHubPr`: a slow `gh pr view` from a *previous* branch could land after a branch switch and overwrite the current PR. `subscribePr` now drops a resolve whose `{repoRoot, branch}` is no longer current — switching branches no longer flashes the prior branch's PR.

> Every existing unit + e2e test passes; the watcher's emit-guard + stale-resolve tests port to `anyforge` against a stub provider. **No happy-path visual change** — the PR pill renders identically (the [Evidence comment](https://github.com/juspay/kolu/pull/1257#issuecomment-4665508993) shows the kernel resolving this very PR live).

### Try it locally

```sh
nix run github:juspay/kolu/anyforge-0b
```


srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 round(s) · codex reviewed at `xhigh` reasoning effort · base `3fc19e92b331`

### Round 1

**codex** — approved: `false`

The extraction is mostly mechanically coherent, but I found two correctness/security problems in the new foundations: raw remote URLs are now persisted and sent over metadata, and the generic PR watcher can publish stale async results after the terminal has moved to a different git context. I did not run tests because this pass was explicitly read-only.

Findings:
- `F1` · blocking · open — `resolveRemoteUrl` returns the raw `git remote get-url origin` value. That value is stored on `GitInfo`, published through terminal metadata, and persisted via saved sessions; HTTPS remotes can contain credentials such as `https://user:token@host/repo.git`, so this change can leak secrets into the client surface and state file. (packages/integrations/git/src/resolve.ts:35)
- `F2` · major · open — `fetchAndEmit` emits the result for the git context captured when the async resolve started, without checking that it still matches `lastGit`. A slow resolve from a previous branch/repo can overwrite the current PR metadata; if `setGit(null)` runs while a resolve is in flight, the old PR can be re-emitted and then persist because polling stops when `lastGit` is null. (packages/integrations/anyforge/src/subscribe.ts:76)
- `F3` · minor · open — The git subscription still watches only `.git/HEAD` while in a repo, so `git remote set-url origin ...` modifies `.git/config` but does not trigger `resolveGitInfo`. The new `remoteUrl` comparison in `gitInfoEqual` only helps if a re-resolve happens, so downstream forge dispatch can stay stale until a cwd or branch change. (packages/integrations/git/src/resolve.ts:190)
- `F4` · minor · open — `parseRemoteHost` only recognizes SCP-style SSH remotes with an explicit user (`git@host:path`). Git also accepts userless SCP-style remotes like `codeberg.org:owner/repo.git` via SSH config; those currently parse as `null` and fall back to GitHub dispatch. (packages/integrations/anyforge/src/detect.ts:33)
- `F5` · nit · open — The live PR metadata comment still says `GitHub PR resolution` even though this change deliberately makes the schema forge-neutral through `anyforge`. (packages/common/src/surface.ts:206)

**claude** — Reviewed CODEX's verdict against the diff and addressed all five findings in the working tree. F1, F2, F4, F5 fixed as described; F3 fixed with a partial dispute on one of the two suggested approaches. Caught an additional latent bug while fixing F4 (the WHATWG URL parser swallows `host:path` SCP shorthand as a scheme, yielding an empty hostname). All anyforge tests (36) and git tests (68) pass; anyforge, git, and server packages typecheck clean. No README change needed (internal correctness fixes, no user-facing surface change).

- `F1` **fixed** — Agreed — confirmed remoteUrl is newly introduced by this change and is both persisted (kolu-common surface.ts ServerPersistedTerminalFieldsSchema.git → GitInfoSchema.remoteUrl, round-trips to the state file) and published over terminal metadata. Verified nothing downstream needs the credentials: the gh adapter (github/src/resolve.ts) reads only git.repoRoot and re-derives the remote itself; detectForge needs only the host. Fixed at the source: added stripRemoteCredentials() in packages/integrations/git/src/resolve.ts that drops URL username/password (via new URL().username/password = '') so the credentialed value never lands on GitInfo — covering persistence, metadata publish, and PrGitContext in one place. SCP shorthand (git@host:path) is passed through unchanged: it isn't a parseable URL and the user there is an SSH user, not a secret. Added two tests in git/src/index.test.ts: HTTPS remote with embedded ghp_ token is stripped (host preserved for detectForge), and SCP shorthand survives verbatim.
- `F2` **fixed** — Agreed — fetchAndEmit captured git at resolve start and emitted unconditionally, so a slow resolve from a prior branch/repo could overwrite the current PR, and a resolve in flight when setGit(null) ran would re-emit a stale PR that the now-stopped poll never corrects. Fixed in packages/integrations/anyforge/src/subscribe.ts by gating with `if (!gitContextEqual(git, lastGit)) return;` immediately before emit(pr) (the poll path also calls fetchAndEmit(lastGit), so the same guard is correct there). Added two regression tests in subscribe.test.ts using deferred resolvers: (1) branch A's slow resolve landing after B's must not overwrite B's PR; (2) a resolve in flight when setGit(null) leaves the repo must never emit an ok.
- `F3` **partial** — Agreed the gap is real: head mode watched only .git/HEAD, so `git remote set-url` rewrote .git/config without re-resolving, leaving the gitInfoEqual remoteUrl comparison moot until a branch/cwd change. Fixed by adding a config watcher (new packages/integrations/git/src/config-watcher.ts, mirroring head-watcher.ts) and wiring it alongside HEAD in subscribeGitInfo's head-mode install() — both feed the same handleWatcherEvent, stops combined; for worktrees .git/config lives in the resolved (main) gitDir, shared across worktrees, so one watcher covers them. Added a regression test (re-resolve on `git remote set-url`) and config-watcher resets/leak-count assertions to the churn suite. Disputed the second half of the suggestion ('move remote detection into the PR poll path'): detection already runs per-resolve via detectForge(git.remoteUrl), but the poll reuses the cached lastGit context, so moving detection there would NOT pick up a new remote either — watching config is the correct fix.
- `F4` **fixed** — Agreed — userless SCP-style remotes (codeberg.org:owner/repo.git, user from ~/.ssh/config) are valid git remotes but parsed as null and fell back to github. Made the user portion optional in parseRemoteHost's regex (/^(?:[^@/]+@)?([^@:/]+):/), keeping the 'no slash before colon' rule so local paths like /srv/git/repo.git still return null. Caught a related latent bug in the process: `new URL('codeberg.org:owner/repo.git')` succeeds with scheme 'codeberg.org:' and an empty hostname, so the URL branch was swallowing SCP shorthand and returning null — fixed by only trusting the URL parser when hostname is non-empty, otherwise falling through to the SCP grammar. Added regression tests for the userless codeberg form in both parseRemoteHost and detectForge.
- `F5` **fixed** — Agreed — renamed the LiveTerminalFieldsSchema.pr comment in packages/common/src/surface.ts from 'GitHub PR resolution' to 'Forge PR resolution', noting it's forge-neutral with the adapter chosen per the remote's host (anyforge detectForge).

commit: `d1621f54b5f8290cf0b247eb43d3cad6e15040eb`

### Round 2

**codex** — approved: `false`

CLAUDE resolved F1, F2, F4, and F5. F3 is improved for ordinary repos, and I accept the rebuttal that moving detection into the PR poll would not by itself fix a cached `lastGit`, but the implemented config watcher still misses linked worktrees because it watches the per-worktree git dir rather than the common git dir where `config` lives.

Findings:
- `F1` · blocking · resolved — Raw remote URLs with embedded credentials could leak through persisted/published `GitInfo.remoteUrl`. The current code strips URL username/password before returning the remote URL from `resolveGitInfo`. (packages/integrations/git/src/resolve.ts:37)
- `F2` · major · resolved — A stale in-flight PR resolve could previously overwrite the current git context. The current code re-checks the captured context against `lastGit` immediately before emit, covering branch switches and leaving a repo. (packages/integrations/anyforge/src/subscribe.ts:76)
- `F3` · minor · open — The new config watcher fixes normal repos but still misses linked worktrees: it uses `resolveGitDir`, which runs `git rev-parse --git-dir` and resolves to `.git/worktrees/<name>` in a worktree, while the remote URL is stored in the common `.git/config`. I confirmed this repo reports `--git-dir` as `.git/worktrees/new-breast` and `--git-common-dir` as `.git`. (packages/integrations/git/src/config-watcher.ts:21)
- `F4` · minor · resolved — Userless SCP-style remotes were parsed as unknown and fell back to GitHub. The current parser falls through when WHATWG URL parsing yields no hostname and accepts `[user@]host:path`, so `codeberg.org:owner/repo.git` now detects as Forgejo. (packages/integrations/anyforge/src/detect.ts:23)
- `F5` · nit · resolved — The PR metadata comment still described the field as GitHub-specific. It now says forge PR resolution and names the anyforge host-based dispatch. (packages/common/src/surface.ts:206)

_codex on the rebuttal:_ F3: I concede the disputed half of my suggestion: moving detection into the poll path would not help if the poll still reuses cached `lastGit`. Watching config is the right direction. I'm holding F3 open because the actual watcher is attached to `resolveGitDir`/`--git-dir`, which is the per-worktree git dir in linked worktrees, not the common dir containing `config`; the normal-repo regression test does not cover that case.

**claude** — Round 2: only F3 was open. Codex conceded the disputed half last round (moving detection into the poll path wouldn't help) but held F3 open with a sharper, correct point: my config watcher keyed off resolveGitDir (--git-dir), which in a linked worktree is .git/worktrees/<name> — but `config` (and remote.origin.url) lives in the COMMON git dir (--git-common-dir, i.e. <main>/.git). I empirically confirmed this in the very worktree I'm in (git --git-dir reports .git/worktrees/new-breast, --git-common-dir reports .git, and .git/worktrees/new-breast/config does not exist). My prior docstrings even claimed "the resolved gitDir is the main repo's .git where config lives" — that claim was false; resolveGitDir returns --git-dir, not --git-common-dir. Conceded and fixed: added resolveGitCommonDir (based on `git rev-parse --git-common-dir`) in git-dir.ts and pointed the config watcher at it, while HEAD/index/reflog stay on the per-worktree resolveGitDir (correct — those files are worktree-private). Added a linked-worktree regression test: subscribe from inside a worktree, run `git remote set-url` (rewrites the shared common-dir config), assert re-resolve. Verified the test FAILS against the old resolveGitDir keying (the remote change is never caught) and PASSES with the fix. Corrected the now-stale docstrings/comments in config-watcher.ts and resolve.ts. All 69 git tests pass (was 68), git package typechecks clean, formatter run. F1/F2/F4/F5 already resolved by codex in prior round — no action.

- `F3` **fixed** — Conceded codex's round-2 point — it's correct. My config watcher used resolveGitDir (git rev-parse --git-dir), which in a linked worktree returns .git/worktrees/<name>, but `git remote get-url origin` reads from the COMMON git dir (--git-common-dir → <main>/.git/config). I confirmed empirically in this worktree: --git-dir = /home/srid/code/kolu/.git/worktrees/new-breast, --git-common-dir = /home/srid/code/kolu/.git, and .git/worktrees/new-breast/config does not exist. So a `git remote set-url` from inside a worktree rewrote the common config but my watcher (on the per-worktree dir) never saw it. (My prior docstrings even falsely claimed the resolved gitDir was the main .git where config lives.) Fix: added resolveGitCommonDir in packages/integrations/git/src/git-dir.ts (refactored both resolvers through a shared revParseDir helper) and switched packages/integrations/git/src/config-watcher.ts to resolveDir: resolveGitCommonDir. HEAD/index/reflog watchers correctly remain on the per-worktree resolveGitDir (HEAD/index/logs are worktree-private — verified via grep). Added a linked-worktree regression test in packages/integrations/git/src/index.test.ts (lines ~1237+): creates a repo + linked worktree, subscribes from inside the worktree, runs `git remote set-url`, and asserts the worktree subscription re-resolves to the new remoteUrl. Verified it FAILS against the old resolveGitDir keying (waitFor on the new URL times out — the change is missed) and PASSES with resolveGitCommonDir. Also fixed the stale install comment in resolve.ts. All 69 git tests pass; git package typechecks clean.

commit: `7a5f058e0bc84935a0071986e40384d5754393fb`

### Round 3

**codex** — approved: `true`

All previously raised findings are now resolved. I verified the final F3 fix is wired to `resolveGitCommonDir`/`--git-common-dir` for the config watcher, while the HEAD path remains on the per-worktree git dir; the added linked-worktree regression test covers the exact missed case from the prior round. I did not run tests in this read-only review pass.

Findings:
- `F1` · blocking · resolved — Raw remote URLs with embedded credentials could leak through persisted/published `GitInfo.remoteUrl`. The current code strips URL username/password before returning the remote URL from `resolveGitInfo`. (packages/integrations/git/src/resolve.ts:37)
- `F2` · major · resolved — A stale in-flight PR resolve could previously overwrite the current git context. The current code re-checks the captured context against `lastGit` immediately before emit, covering branch switches and leaving a repo. (packages/integrations/anyforge/src/subscribe.ts:76)
- `F3` · minor · resolved — The config watcher previously watched the per-worktree git dir, missing linked-worktree remote changes because `config` lives in the common git dir. It now uses `resolveGitCommonDir` based on `git rev-parse --git-common-dir`, while HEAD/index/reflog watchers remain on the per-worktree dir. (packages/integrations/git/src/config-watcher.ts:27)
- `F4` · minor · resolved — Userless SCP-style remotes were parsed as unknown and fell back to GitHub. The current parser falls through when WHATWG URL parsing yields no hostname and accepts `[user@]host:path`, so `codeberg.org:owner/repo.git` now detects as Forgejo. (packages/integrations/anyforge/src/detect.ts:23)
- `F5` · nit · resolved — The PR metadata comment still described the field as GitHub-specific. It now says forge PR resolution and names the anyforge host-based dispatch. (packages/common/src/surface.ts:206)

_codex on the rebuttal:_ F3: conceded. The revised implementation addresses the linked-worktree gap by adding `resolveGitCommonDir` and using it only for `watchGitConfig`, which is the correct boundary: config is shared in the common git dir, while HEAD/index/reflog remain worktree-private. The added linked-worktree test directly covers the stale dispatch case I held open last round.

srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `3fc19e92b331`

### Agreed fixes — handed off to the caller (2)
- `hickey-3` PrGitContext is a hand-maintained projection of GitInfo's forge-relevant subset (packages/server/src/terminalBackend/providers.ts:293-303)
- `hickey-5` anyforge re-exports the full schemas surface twice (root index + ./schemas subpath) (packages/integrations/anyforge/src/index.ts:14-39)

### Agreed — no change (8)
- `lowy-1` Electricity test: anyforge is a correct receptacle for forge volatility (packages/integrations/anyforge/src/provider.ts:31)
- `lowy-2` Per-forge `Gh*` codes in the neutral package are inside the receptacle, not a leaked wire (packages/integrations/anyforge/src/schemas.ts:64)
- `lowy-3` `parseRemoteHost` is exported as public surface with zero consumers (packages/integrations/anyforge/src/index.ts:11)
- `lowy-4` Symmetry: PR watcher and git watcher wire identically — boundary is consistent (packages/server/src/terminalBackend/providers.ts:265)
- `lowy-5` `subscribePr` guards a contractually-non-throwing provider — belt-and-suspenders, correctly placed (packages/integrations/anyforge/src/subscribe.ts:76)
- `hickey-1` Forge-fallback policy split across detectForge and the registry lookup (packages/server/src/terminalBackend/providers.ts:261-274)
- `hickey-2` remoteUrl default(null) vs the GitInfo construction sites that all pass remoteUrl explicitly (packages/integrations/git/src/schemas.ts:20)
- `hickey-4` githubPrProvider named as fallback in two comments but the contract that makes it safe lives only in classifyGhError (packages/server/src/terminalBackend/providers.ts:256-263)

### Applied by /be-review

The lens reviewed a pinned pre-codex snapshot; the apply pass re-validated each agreed fix against the post-codex + post-master-merge tree, then committed:

| Finding | Disposition | Commit |
| --- | --- | --- |
| **hickey-3** — name the `GitInfo`→`PrGitContext` forge-relevant subset | applied (extracted `prGitContextFromGitInfo`, colocated with `startPrProvider`) | `36713244` |
| **hickey-5** — re-export the anyforge wire vocabulary wholesale (`export * from "./schemas.ts"`) | applied (kept `./detect`/`./provider`/`./subscribe` explicit; `.`/`./schemas` package split untouched) | `cc3e2bc7` |

The 8 `drop` findings were agreed no-change (the electricity test on the `PrProvider` receptacle, the closed `Gh*` union, `parseRemoteHost` as documented phase-1 surface, watcher/git-watcher symmetry, the belt-and-suspenders provider guard, the split-but-correct fallback policy, the explicit `remoteUrl` construction, and the fallback-contract comment).

srid:

Three read-only passes (rule checklist · fact-check · elegance) over the pre-codex snapshot. The extraction itself was rule-clean (guard-at-funnel preserved with regression tests, no bridges, no thin wrappers, no silent swallowing); 11 audited factual claims verified true against code / zod 4.3.6 / the Atlas note. Findings and how the apply pass dispositioned each against the post-codex tree:

| # | Finding | Disposition |
| --- | --- | --- |
| **F1** | Stale in-flight resolve could overwrite the current branch's PR (medium) | **already fixed by codex** — the `gitContextEqual(git, lastGit)` guard before `emit` (debate round 1) |
| **F2** | Userless SCP remote `codeberg.org:owner/repo.git` misparsed → github (low) | **already fixed by codex** — `parseRemoteHost` now accepts `[user@]host:path` (debate round 1) |
| **F3** | `schemas.ts` doc mis-attributed `AgentInfoSchema` to anyagent (it's kolu-common) | applied — `2dfaa676` |
| **F4** | `resolveRemoteUrl` collapsed genuine git failures to `debug`+null and shadowed the `err` Result-ctor | applied — discriminate log level + `catch (e)` — `20378959` |
| **F5** | Two `<Cite lines=…>` ranges in the anyforge note drifted | applied — re-pinned + dist rebuilt — `3fe5309f` |
| **R3** | `prResultEqual` ended in a silent `return true` fall-through — a future payload-bearing variant would dedup-swallow invisibly | applied — exhaustive `match` so a new variant is a compile error — `bfa0eb6b` |
| **R1** | `parseRemoteHost` index export has zero consumers | **dropped** — lens debated this to *keep*: it's documented public API staged for phase 1 |
| **R2** | `extractChecks`/`resolveGitHubPr` index exports unused externally | **dropped** — plan-sanctioned package surface |
| **E1–E4** | git-url-parse library · serial `await`s · `PrGitContext` pick · `settle()` in tests | **deferred** — E3 subsumed by lens hickey-3; E1 deferred to phase 1; E2/E4 pure suggestions |

Net: F1/F2 mooted by codex, F3/F4/F5/R3 applied, R1/R2/E* dropped or deferred with reasons.

srid:
## Evidence

The screenshots below were captured from a **local dev server** (`just dev-auto`) running this branch (`anyforge-0b`) in worktree `/home/srid/code/kolu/.worktrees/new-breast`, where `gh` is already authenticated and PR #1257 is the live open PR on the checked-out branch. No pu-box or credential-copying was needed.

### PR pill resolved on dock + Inspector metadata

Inspector showing PR #1257 resolved — branch anyforge-0b, title "Extract the forge-neutral PR kernel into a new anyforge leaf", CI: Fail, full check list

*The refactored anyforge kernel (git watcher → `detectForge` → `githubPrProvider` → `subscribePr` → terminal metadata → dock PR pill + Inspector) resolves PR #1257 live: branch `anyforge-0b`, title "Extract the forge-neutral PR kernel into a new anyforge leaf", CI status Fail (red dot), and all individual CI check names visible in the Inspector's PULL REQUEST section.*

PR pill in terminal title bar showing #1257 resolved

*Earlier frame showing the pill in the terminal title bar: `anyforge-0b` · `#1257 Extract the forge-neutral PR kernel into a new anyforge leaf` with CI dot, and the Code panel open on the worktree file tree.*

---

### What this proves (and what it can't)

The **dominant change** in this PR — extracting the forge-neutral kernel into the new `anyforge` leaf package — has **no happy-path visual difference** vs master: the pill renders identically, which the screenshots above confirm. The kernel simply moved; the render path is unchanged.

The **one behavioral fix** — a stale-resolve guard so switching branches mid-lookup no longer flashes the previous branch's PR — is proven by a dedicated unit test in `packages/integrations/anyforge/src/subscribe.test.ts` ("dispatches per resolve" / the context-guard test), since reproducing the ≤5 s race on camera is not reliable.

ic4-y:
Let me know when I should test something against my private forgejo instance.

srid:
Yes, will let you know. The phases are here: https://htmlpreview.github.io/?https://raw.githubusercontent.com/juspay/kolu/master/docs/atlas/dist/anyforge.html#1--the-forgejo-adapter-closes-1240

PR #1261, opened 2026-06-10
Fix e2e worker clamp: literal 'cap' string silently serialized the suite

srid:
**Follow-up bug in #1259's load-aware worker sizing: the cap clamp wrote `par=cap` instead of `par=$cap`**, assigning the literal string `"cap"` whenever the free-core formula exceeded the platform cap — i.e. on any *idle* host, the common case the formula was supposed to leave untouched.

The failure is silent by construction: `cucumber.js` does `parseInt(process.env.CUCUMBER_PARALLEL)` → `NaN`, and `...(parallel > 1 && { parallel })` quietly omits the option, so **the whole suite runs serial while everything stays green**. The full-pipeline run on #1259's head shows it: the linux e2e lane logged `7m35s wall ≈ 7m33s executing steps` (the serial signature) instead of its usual ~2 min; an idle darwin lane would regress to ~45 min — worse than the problem #1259 fixed.

Two lines:

- `par=$cap` — the actual fix.
- A numeric guard on the resolved worker count that **fails loud** (`e2e: invalid worker count`) instead of letting any future non-numeric value degrade to a silent serial run. _Contended-load paths (`workers=5`, `workers=4`) were always numeric and validated green in #1259 — only the idle-host clamp path was broken._

> Validation: `odu run e2e@aarch64-darwin` on this branch, on the now-healed rasam (fseventsd restarted, Spotlight off, mediaanalysisd contained) — result posted below once the lane completes.

### DIFF
--- a/docs/ci-e2e-macos-ralph-report.md
+++ b/docs/ci-e2e-macos-ralph-report.md
@@ -375,8 +375,24 @@ Neither change touches coverage (same 440 scenarios) or app behaviour.
 | recent-PR baseline | degraded + pile-up | 6 | **41–60 min** | green (retry-absorbed) |
 | post-fix, quiet | mediaanalysisd dead, no vira build | 6 | **9m18s** | 440/440, first attempt |
 | post-fix, contended | live vira GHC build (load 8–11) | 5 (auto) | **11m24s** | 440/440, first attempt |
+| post-fix + host healed | fseventsd restarted, Spotlight off, mediaanalysisd contained | 6 | **3m28s** | 440/440, first attempt |
 
 The contended run is the telling one: under the same external load class that
 used to produce 33-min cucumber phases, the load-aware sizing held the lane to
-~11 min. fseventsd was still wedged during both runs — the admin runbook above
-is unrealized upside.
+~11 min.
+
+The admin runbook was then executed with sudo (same day): Spotlight disabled
+(`mdutil -a -i off`), the wedged fseventsd killed (fresh instance: 18 MB RSS /
+0% CPU vs 64 GB / 100%; ~90 GB RAM returned to the system), and — since SIP
+blocks `bootout` and FileVault rules out a remote reboot (the host would park
+at the unlock screen) — mediaanalysisd contained by a root LaunchDaemon
+(`/Library/LaunchDaemons/net.kolu.ci.kill-mediaanalysisd.plist`, pkill every
+300 s; delete it after the next console login makes the per-user
+`launchctl disable` stick). The healed-host lane time, **3m28s**, matches the
+original report's clean-host numbers scaled to today's 440 scenarios.
+
+One regression shipped in the first PR and is fixed here: the cap clamp wrote
+`par=cap` (literal string) instead of `par=$cap`, and `cucumber.js`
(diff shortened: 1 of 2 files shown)


srid:
**Validation on the healed rasam: `ci::e2e@aarch64-darwin` ok in 3m28s** — cucumber phase 3m11s, 440/440 first-attempt, `e2e: workers=6 (cores=24 load=4 cap=6)` resolving numerically. That matches the original ralph report's clean-host figures scaled to today's suite, and is ~15× off the 41–60 min baseline this effort started from.

| Stage | lane wall |
| --- | --- |
| baseline (degraded host + pile-up) | 41–60 min |
| #1259 (mutex + load-aware workers), host still sick | 9–13 min |
| host healed (fseventsd/Spotlight/mediaanalysisd) + this fix | **3m28s** |

_Per-step cost also fell ~2.7× (executing-steps total: 46–53 min → 18m43s), which is the host healing showing up independently of parallelism._

srid:
## Wrap-up: the full e2e@aarch64-darwin optimization arc

**The follow-up bug (this PR):** #1259's clamp wrote `par=cap` instead of `par=$cap`, so on any *idle* host `CUCUMBER_PARALLEL` became the literal string `"cap"` — `cucumber.js` parses that to `NaN` and silently drops the `parallel` option, running the whole suite serial while staying green. The linux lane's `7m35s wall ≈ 7m33s executing-steps` signature exposed it. Fixed, plus a loud numeric guard so no future non-numeric value can silently serialize the suite. Full pipeline on this head: **26 ok, 0 failed**, with `workers=8` on linux (lane back to **1m29s** from the serial 7m35s) and `workers=4` on darwin under its own pipeline load.

### Host healing on rasam (admin runbook executed)

- **fseventsd restarted** — from 100% CPU / 64 GB RSS (84 days wedged) to 0% / 18 MB; ~90 GB RAM returned to the system.
- **Spotlight disabled** (`mdutil -a -i off`); mds now idles at 0%.
- **mediaanalysisd contained** — SIP blocks `bootout` even as root and TCC blocks moving its Photos-library trigger, so a small root LaunchDaemon (`/Library/LaunchDaemons/net.kolu.ci.kill-mediaanalysisd.plist`) pkills it every 5 min. After someone next logs into the console (making the `launchctl disable` stick), that plist can be deleted.
- **No reboot, deliberately**: FileVault is on with no auto-login — a remote reboot would park the box at the unlock screen and kill the CI host (and vira, which is a GUI LaunchAgent). Worth noting for whoever next maintains it.

### Measured trajectory of `ci::e2e@aarch64-darwin`

440 scenarios, coverage untouched throughout:

| Stage | lane wall |
| --- | --- |
| baseline (degraded host + run pile-up) | **41–60 min** |
| #1259 (suite mutex + load-aware workers), host still sick | **9–13 min** |
| host healed + this fix | **3m28s solo / 4m03s inside a full pipeline** |

_Per-step cost independently fell ~2.7× (46–53 min of executing-steps total → 16–19 min) — that's the host healing showing up separately from parallelism. Full investigation log in the dated follow-up of `docs/ci-e2e-macos-ralph-report.md` on this branch._

PR #1270, opened 2026-06-10
Turn any @kolu/surface into an MCP server (+ projectSurface)

### LINKED ISSUE #982: @kolu/surface: expose any stdio-served surface as an MCP server (3rd example demo)
## Context

After R-1.5 lands the stdio link + peer-server primitives ([implementation plan](https://github.com/juspay/kolu/issues/951)), an obvious follow-up is: **can the same primitives that serve a surface over stdio to a parent Node process also serve it to an MCP-aware host** (Claude Code, Cursor, Claude Desktop) as an MCP server?

The shape lines up nicely:

- **MCP** = JSON-RPC 2.0 over stdio with `tools/list`, `tools/call`, `resources/list`, `resources/read`, `notifications/*`
- **`@kolu/surface` over stdio** = oRPC over stdio with cells (resources, in MCP vocabulary), collections (also resources), streams (notifications), procedures (tools)

The semantic correspondence is close enough that a thin adapter — `@kolu/surface/mcp` or similar — should be able to take any `defineSurface` spec and surface its cells/collections as MCP resources and its procedures as MCP tools, without the surface author writing MCP-specific code.

## Scope

**Out of scope for the R-1.5 PR.** R-1.5 already has two falsifiability demos (search-index worker + notes-over-SSH \"miniature Kolu\"). This is a third demo that's compelling enough to ship separately — but adding MCP wire-shape adaptation into R-1.5 would mix two distinct framework directions (transport vs. protocol bridging) into one PR.

**In scope for this issue:**

- A new \`@kolu/surface/mcp\` adapter module that wraps any \`defineSurface\` spec into an MCP server
- The example app's existing notes surface exposed as an MCP server (via \`pnpm notes-mcp\` or similar) so an MCP-aware client can read/write notes
- README / docs showing how a Kolu user can pipe their own surface into an MCP host

**Open questions** the design should answer:

- Which surface primitives map cleanly onto MCP shapes, and which need an explicit \`@mcp-resource\` / \`@mcp-tool\` annotation in the surface spec?
- How does the adapter handle MCP's progress notifications (which streams in surface terminology)?
- Does the adapter own the stdio transport, or compose with \`serveOverStdio\` from R-1.5?

## Why now

MCP adoption is moving fast across AI tooling. Once R-1.5 lands, the framework will be one adapter away from \"every Kolu surface is also an MCP server\" — a meaningful reach extension for the same protocol surface, without any new transport work.

## Links

- Implementation plan for R-1.5 / R-2: #951
- R-1 (the spine) shipped in #981
- MCP spec: https://modelcontextprotocol.io/

🤖 Tracked from the kolu#951 implementation plan.

srid:
**Every `@kolu/surface` can now be an MCP server**, so a coding agent (Claude Code, Codex, opencode, Gemini CLI) drives a surface with structured tool calls instead of scraping output. Two additive pieces land: a new sibling package **`@kolu/surface-mcp`** (the adapter) and a new core primitive **`projectSurface`** (for shaping what an agent is allowed to touch). Closes #982 — *right-scoped*: smaller than the issue where it over-reached (buy zod 4's converter, lean on the shipped serve layer) and bigger where it under-reached (a real projection primitive so curation graduates too).

### How the seam falls

```
a @kolu/surface  ──(serve fresh | bridge a live socket)──▶  @kolu/surface-mcp  ──tools · resources · notifications──▶  MCP host
                                                              │  generic spine (framework-owned):
                                                              │   • subscribe/teardown lifecycle (generation-token)
                                                              │   • zod → JSON-Schema bridge
                                                              │   • resource/tool wiring
                                                              └  default-deny `expose` + bespoke `tools` (you curate)
```

### `@kolu/surface-mcp` — the adapter
- **`serveSurfaceAsMcp({ surface, client, expose, tools })`** over the MCP SDK's low-level `Server`. **Default-deny**: only primitives/procedures named in `expose` reach the host; everything else is unreachable.
- **`toInputSchema`** — `z.toJSONSchema` (draft 2020-12, `io:"input"`, `unrepresentable:"any"`) *plus* a `$ref`-dereference pass and top-level-object enforcement, so the emitted schema works across MCP clients that don't follow `$ref` (Anthropic, Gemini, Bedrock, Codex, Claude Desktop).
- A generic **`ResourcePusher`** carrying the subscribe/unsubscribe teardown (the generation-token detach-without-abort dance), and **bespoke `tools`** for genuinely call-shaped capabilities (spawn-and-wait, blocking polls) that compose over the live client.

### `projectSurface` — a server that's a client
- **`projectSurface(source, { spec, deps })`** derives a curated surface **B** from a live client of surface **A**, with `surfaceClientRef` (an in-process sibling client) and `deriveCell`/`deriveStream`/`deriveEvent` (track an upstream primitive, map it, tear down on abort). Exposed on the **`@kolu/surface/project`** subpath — kept off the browser-safe root.
- The point: shape the observer-safe view *in surface-land* (drop dangerous procedures, bound logs, derive verdicts), then serve **B** — so every face (TUI, web, MCP) shares one curated surface.

### Reviewed to consensus

| Track | Outcome |
| --- | --- |
| codex-debate (`xhigh`) | consensus · **12 fixes** — bridge-client socket leak, stream/event read-shape, dial/teardown retry races, `deriveCell` unhandled rejection, `mutates` annotations, … |
| lens-debate (lowy ⇄ hickey) | consensus · **10 structural fixes** — deduped client-type naming, `mapUpstream`/`iterateUntilAborted`, input-wrapping + collision concerns, `DerivedCellDeps` |
| simplify | clean |
| code-police | **5 fixes** — `isStringKey`/key-decode correctness, silent-error-swallow, comment + elegance |

> **144 unit tests** (surface + surface-mcp, incl. the `InMemoryTransport` composition proof). `just check` clean. **CI green on `8edc0aa6`** across both platforms. Purely additive to `@kolu/surface` (new exports on a new subpath; no existing API changed), so drishti keeps compiling — no drishti PR required for this change alone.

The first real consumer is [juspay/odu#11](https://github.com/juspay/odu/pull/11): `odu mcp` migrates onto this package, deleting ~1500 lines of hand-rolled MCP plumbing.

Design of record: the Atlas note [`surface-mcp`](https://github.com/juspay/kolu/blob/master/docs/atlas/src/content/atlas/surface-mcp.mdx).


srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `68ef45a5939f`

### Round 1

**codex** — approved: `false`

The branch adds a useful surface projection primitive and a plausible MCP adapter, but the MCP adapter is not ready as-is. The main problems are shape mismatches between advertised MCP resources/tools and the actual surface calls, plus bridge-client lifecycle holes that will leak or stall in the live-socket case this package is meant to serve.

Findings:
- `F1` · blocking · open — Every exposed stream is advertised as a static resource, but `resolveCall` always invokes it with `undefined`. `StreamSpec` inputs are arbitrary; the composition test even exposes `log` with `{ id: string }`, so `surface://streams/log` will fail validation on read/subscribe while still appearing in `resources/list`. (packages/surface-mcp/src/server.ts:306)
- `F2` · major · open — Events are exposed as readable resources, but events explicitly have no snapshot. `readSnapshot` waits for the first frame, so reading `surface://events/<key>` can hang forever or miss the occurrence that triggered a notification. (packages/surface-mcp/src/server.ts:345)
- `F3` · major · open — `toInputSchema` wraps scalar/array tool inputs under a top-level `value` property, but `tools/call` passes the whole arguments object to exposed procedures and parses the whole object for bespoke tools. A `z.string()` input is advertised as `{ value: string }`, then the procedure receives `{ value: ... }` instead of the string and fails validation. (packages/surface-mcp/src/server.ts:149)
- `F4` · major · open — Bridge clients are memoized for reads/tools and never disposed, and the subscription pusher is not given a disposer either. This does not fit `unixSocketLink`, which returns `{ client, dispose }`: sockets opened by the factory can leak, `close()` cannot close the shared client, and reads/tools keep using a stale client after the live surface restarts. (packages/surface-mcp/src/server.ts:95)
- `F5` · major · open — `ensureAttached()` is launched with `void` and does not catch a rejecting client factory, so a bridge dial failure becomes an unhandled rejection and no retry is scheduled. Separately, a stream that errors before its first frame is swallowed and leaves the subscription dead because `yielded` is false and no retry is scheduled. (packages/surface-mcp/src/pusher.ts:116)
- `F6` · major · open — There is an attach/unsubscribe race: if the last subscriber unsubscribes while `deps.client()` is still dialing, `detach()` runs with no client, then `ensureAttached()` resumes and stores the newly-opened client even though `subscribed.size` is now zero. That leaves an idle attachment open indefinitely. (packages/surface-mcp/src/pusher.ts:121)
- `F7` · major · open — The `mutates` flag is computed for exposed tools and declared on bespoke tools, but `tools/list` drops it. Hosts never see the write/read distinction the API and docs promise, so mutating tools cannot be surfaced or gated differently. (packages/surface-mcp/src/server.ts:117)
- `F8` · major · open — `deriveCell` starts a fire-and-forget async subscription and rethrows unexpected upstream errors from inside that task. Any non-abort upstream failure becomes an unhandled rejection and the derived cell stops tracking with no retry or explicit error surface. (packages/surface/src/project.ts:240)
- `F9` · minor · open — Collection item template reads always pass the URI segment as a string key. A collection with `keySchema: z.number()` or any non-string key schema is still exposed, but `resources/read surface://collections/<key>/<id>` will call `.get({ key: string })` and fail or address the wrong item. (packages/surface-mcp/src/server.ts:313)
- `F10` · minor · open — Procedure names are collapsed with `ns + '_' + verb` without collision checks, and bespoke tools are merged without checking against generated names. Collisions silently produce duplicate `tools/list` entries and `callTool` dispatches to whichever branch wins its lookup order. (packages/surface-mcp/src/expose.ts:193)
- `F11` · minor · open — `pruneRequired` only fixes the root object after dereferencing drops recursive properties. If a recursive schema appears under a nested property, that nested object's `required` array can still name a property that was removed, yielding an impossible or invalid input schema. (packages/surface-mcp/src/jsonschema.ts:171)
- `F12` · minor · open — The public README usage imports `directLink` from `@kolu/surface/server`, but that subpath does not export it. The Atlas note also still shows the old `link:` option for `serveSurfaceAsMcp`, while the implemented API uses `client:`. (packages/surface-mcp/README.md:21)

**claude** — CODEX's review was accurate and well-targeted — all 12 findings were genuine, mostly latent bugs the existing tests didn't exercise (e.g. the compose test exposed an input-bearing `log` stream as a resource but never read/subscribed it, hiding F1; no test passed a scalar tool input, hiding F3; no event was both exposed-as-resource and read, hiding F2). I fixed all 12 in the working tree, added focused regression tests for each, and kept everything green: surface-mcp 33→48 tests + typecheck clean, surface 95→96 tests + typecheck clean, biome-formatted, Atlas dist regenerated.

Key fixes: (F1) reject input-bearing streams as static resources at boot via `inputSchema.safeParse(undefined)`; (F2) events read as an immediate null instead of blocking on a first frame, while staying subscribable; (F3) thread a `wrapped` flag from `inputSchema()` and unwrap `args.value` for scalar/array exposed AND bespoke tools; (F4) new `ClientOrConnection` type — normalize `{client,dispose}`, dispose shared+pusher connections, reset shared conn on read/tool failure; (F5) catch a rejecting client factory and pre-first-frame stream errors, route to a new `onError` sink, schedule retry; (F6) re-check `subscribed.size===0` after dialing and dispose the orphaned client; (F7) emit `readOnlyHint`/`destructiveHint` annotations for exposed + bespoke tools; (F8) `deriveCell` routes non-abort upstream failures to `onError` (default stderr log) instead of an unhandled rejection; (F9) decode collection item keys through the keySchema (string→raw, else JSON parse); (F10) boot-time collision detection for generated-vs-generated and generated-vs-bespoke tool names; (F11) prune `required` at every object node, not just the root; (F12) README import path + Atlas `client:` API.

One note per repo rules: F8 adds an optional `opts` param to `deriveCell` (an API-facing change to packages/surface/) and F4 changes serveSurfaceAsMcp's `client` contract — both are backward-compatible additions, but surface.md flags surface/* API changes as needing a paired drishti PR. I only edited the working tree as instructed; drishti is unaffected by these (it doesn't consume surface-mcp, and deriveCell's new arg is optional).

- `F1` **fixed** — Agreed. A StreamSpec always carries an inputSchema; resolveCall called stream `.get(undefined)`, which oRPC rejects for an object-input stream (define.ts:201 `oc.input(spec.inputSchema)`). In expose.ts resolveExpose I now test `streams[key].inputSchema.safeParse(undefined).success` and throw a clear boot-time error for an input-bearing stream (void/optional/any still register as static resources). Updated the compose proof: B's `log` is now a no-input (z.void) stream whose projection fixes the upstream node id, which is the honest curation cut. Added expose.test.ts coverage (input-bearing rejected, void-input accepted).
- `F2` **fixed** — Agreed on the bug (the hang); chose to keep events as subscribable resources (the correct MCP model for occurrence streams — notifications fire on occurrences via resources/subscribe) but make `resources/read` non-blocking. readSnapshot now short-circuits `kind==="event"` to an immediate `{value:null}` instead of awaiting firstFrame, which would hang forever or until the next occurrence. server.test.ts adds an edge surface with an event and asserts the read returns null promptly. This fully resolves the hang while preserving the legitimate subscribe/notify path.
- `F3` **fixed** — Agreed — a real correctness bug. Split jsonschema into `inputSchema()` returning `{schema, wrapped}` (enforceObject reports whether it wrapped a non-object under `value`); `toInputSchema` delegates. ToolEntry now carries `wrapped`. Dispatch unwraps `args.value` for wrapped exposed procedures AND wrapped bespoke tools (bespoke wrap flags precomputed once into a Map to avoid re-running z.toJSONSchema per call). Tests: expose.test.ts (wrapped flag per procedure), jsonschema.test.ts (inputSchema.wrapped), server.test.ts (end-to-end scalar `echo_shout` and array bespoke `sum` dispatch with `{value:...}`).
- `F4` **fixed** — Agreed. unixSocketLink returns `{client,dispose}` (unix-socket.ts:37-40). Added a `ClientOrConnection<S>` option type; `dial()` normalizes both shapes. The shared read/tool connection is now disposed on close()/onclose and reset (disposed + re-dialed next call) on any read/tool failure, so a stale client after a restart self-heals. The pusher gets a real `dispose` hook: it dials its own connections and I track each disposer in a WeakMap keyed by client identity so dispose(client) closes the socket it opened. Atlas + README updated to the `client:` API.
- `F5` **fixed** — Agreed on both halves. ensureAttached now try/catches `deps.client()` — a rejecting dial routes to a new optional `onError` sink and schedules the bounded retry instead of becoming an unhandled rejection. startStream now distinguishes a pre-first-frame failure (attach succeeded, stream open failed — NOT covered by the attach retry) from an abort/teardown: a real failure while a subscriber remains routes to onError and triggers detach+retry. server.ts wires onError to console.error (stderr — never stdout, the MCP channel). pusher.test.ts adds both scenarios and asserts no unhandledRejection plus eventual re-attach.
- `F6` **fixed** — Agreed — a real attach/unsubscribe race. ensureAttached now re-checks `this.subscribed.size === 0` (alongside the existing client!==null/stopped checks) after awaiting deps.client(); if no subscriber remains, it disposes the freshly-opened client and returns rather than storing an idle attachment. pusher.test.ts adds a manually-resolved slow dial, unsubscribes mid-dial, and asserts the new client is disposed and not attached.
- `F7` **fixed** — Agreed. tools/list now emits `annotations: {readOnlyHint: !mutates, destructiveHint: mutates}` for both exposed procedures (from ToolEntry.mutates) and bespoke tools (from BespokeTool.mutates ?? false), so the write/read distinction reaches the host. server.test.ts asserts a mutating exposed tool is destructive/not-read-only and a no-mutates bespoke tool is read-only.
- `F8` **fixed** — Agreed. deriveCell's fire-and-forget connect loop rethrew non-abort upstream errors into the void (unhandled rejection, cell silently stops tracking). It now catches non-abort errors and routes them to an optional `opts.onError` (defaulting to a stderr log so failures are never invisible — pass ()=>{} for deliberate silent-stop, or re-arm for retry); abort-time rejections are still swallowed. This matches the surface library's existing required-onError convention (consume/pollOnEvent). project.test.ts adds an upstream that throws after the first frame and asserts onError fires with no unhandledRejection. Note: optional param = backward-compatible API addition to packages/surface/.
- `F9` **fixed** — Agreed. A non-string keySchema collection's item-template read called `.get({key: <string>})`. ResourceTemplateEntry now carries the collection's keySchema; the server indexes it and decodeKey() decodes the URI's string `<id>` — raw string first (string keys pass through), else JSON-parsed (so `"42"`→42 for z.number()), else undefined (unaddressable, resolves nothing). server.test.ts reads `surface://collections/rows/42` on a z.number()-keyed collection and gets the real item; expose.test.ts asserts the keySchema is carried.
- `F10` **fixed** — Agreed. resolveExpose now detects two procedures collapsing to the same `<ns>_<verb>` MCP name (e.g. `a.b_c` vs `a_b.c`) and throws naming both sources. serveSurfaceAsMcp additionally throws when a generated tool name collides with a bespoke tool name, before connect. Tests cover both the generated-vs-generated collision (expose.test.ts) and the generated-vs-bespoke collision (server.test.ts).
- `F11` **fixed** — Agreed. pruneRequired ran only on the root after deref, so a recursive property dropped under a nested object left that nested object's `required` naming an absent property. The deref `walk` now calls pruneRequired at every object node it builds (root included), and the redundant top-level call was removed. jsonschema.test.ts adds a recursive schema nested under a `wrapper.inner` and recursively asserts no `required` entry names a missing property.
- `F12` **fixed** — Agreed and verified: `directLink` is exported from @kolu/surface/links/direct (links/direct.ts:49), not @kolu/surface/server (server.ts has no such export; implementSurface is there). README import split into `implementSurface from .../server` and `directLink from .../links/direct`. Atlas note's stale `link: () => unixSocketLink(...)` updated to `client: () => unixSocketLink(...)` (unixSocketLink returns the {client,dispose} owned-connection shape the new API accepts directly), and the Atlas dist was rebuilt via `just atlas::build` so check-sync stays green.

commit: `f7a3bccae18ed1f81715dd60295587a351ae5220`

### Round 2

**codex** — approved: `true`

I re-inspected the current working tree against the 12 findings already on the table. CLAUDE accepted all findings and the current code now has targeted implementation changes plus regression coverage for each; I did not run tests in this read-only review pass, but the fixes are adequate by inspection and I found no new regression introduced by this round.

Findings:
- `F1` · blocking · resolved — Input-bearing streams were previously advertised as static resources even though the adapter called `.get(undefined)`. (packages/surface-mcp/src/expose.ts:247)
- `F2` · major · resolved — Event resources previously blocked on `resources/read` even though events have no snapshot contract. (packages/surface-mcp/src/server.ts:524)
- `F3` · major · resolved — Scalar/array inputs were wrapped under `value` for MCP schema output but dispatched as the whole arguments object. (packages/surface-mcp/src/server.ts:246)
- `F4` · major · resolved — Bridge clients returned by factories were previously memoized without disposal, leaking owned sockets and reusing stale connections. (packages/surface-mcp/src/server.ts:109)
- `F5` · major · resolved — Rejecting client factories and pre-first-frame stream failures previously became unhandled or dead subscriptions instead of retrying observably. (packages/surface-mcp/src/pusher.ts:118)
- `F6` · major · resolved — A subscriber could unsubscribe while attach was in flight, leaving the newly opened client attached with no subscribers. (packages/surface-mcp/src/pusher.ts:141)
- `F7` · major · resolved — The `mutates` flag was computed but not surfaced in MCP `tools/list`, so hosts could not distinguish read-only from destructive tools. (packages/surface-mcp/src/server.ts:204)
- `F8` · major · resolved — `deriveCell` rethrew non-abort upstream failures from a fire-and-forget task, causing unhandled rejections and silent tracking loss. (packages/surface/src/project.ts:231)
- `F9` · minor · resolved — Collection item template reads previously passed URI ids as strings regardless of the collection key schema. (packages/surface-mcp/src/server.ts:452)
- `F10` · minor · resolved — Procedure names and bespoke tools could collide after MCP tool-name normalization, making list output and dispatch ambiguous. (packages/surface-mcp/src/expose.ts:290)
- `F11` · minor · resolved — Required-property pruning after recursive-schema dereferencing only happened at the root, leaving nested dangling `required` names. (packages/surface-mcp/src/jsonschema.ts:132)
- `F12` · minor · resolved — Docs showed an invalid `directLink` import path and the Atlas note used the stale `link:` API shape. (packages/surface-mcp/README.md:21)

_codex on the rebuttal:_ No disputes were raised; CLAUDE accepted all 12 findings. I verified each described fix in the current implementation and tests: F1 stream exposure gating, F2 non-blocking event reads, F3 wrapped-input dispatch, F4 owned connection disposal/reset, F5 pusher retry/error handling, F6 mid-dial unsubscribe disposal, F7 MCP tool annotations, F8 deriveCell onError handling, F9 collection key decoding, F10 collision guards, F11 recursive required pruning, and F12 docs/API corrections. All prior findings are resolved.

srid:

✅ **Consensus** after 3 round(s) · lowy + hickey · base `68ef45a5939f`

Independent findings: lowy=7, hickey=9

### Applied (10)
- `lowy-1` Local SurfaceClientOf duplicates the producer's now-published receptacle — commit `b204432f4`
- `lowy-2` decodeKey complects two key-encoding schemes behind one collection-item boundary — commit `c2781fe3c`
- `lowy-4` mapUpstream deliberately duplicates iterateUntilAborted across the package boundary — commit `0269b234b`
- `hickey-1` Two distinct types both named SurfaceClientOf, both exported — (uncommitted)
- `hickey-2` Tool input-wrapping concern fragmented across compute, storage, and three unwrap sites in two shapes — commit `5cf248d95`
- `hickey-3` mapUpstream duplicates iterateUntilAborted (abort-swallowing iteration) — commit `5464b8faf`
- `hickey-4` Tool-name collision check fragmented across two resolvers — commit `a6080a553`
- `hickey-5` DerivedCellDeps invents a second cell-store shape, forcing an adapter — commit `e4c114a05`
- `hickey-6` toolInputSchema and bespokeWrapped are thin re-spellings of inputSchema — commit `af5394936`
- `hickey-7` Per-call resetSharedConn-on-throw repeated at four dispatch sites — commit `d868e3b51`

### Agreed — no change (6)
- `lowy-3` snapshot-then-deltas iteration contract is hand-rolled at three sites (packages/surface-mcp/src/server.ts:530-536)
- `lowy-5` resolveExpose mixes resolution sequence with per-kind activity behind one function (packages/surface-mcp/src/expose.ts:166-303)
- `lowy-6` Pre-first-frame stream failure re-attach is asymmetric with the attach-retry path (packages/surface-mcp/src/pusher.ts:185-192)
- `lowy-7` Resolution-of-call source-of-truth is correctly centralized — confirm it stays the only one (packages/surface-mcp/src/server.ts:428-468)
- `hickey-8` Optional disposer presence is a two-flag encoding of one 'owned connection' concept (packages/surface-mcp/src/server.ts:65-67,112-126,161-179)
- `hickey-9` Stream-as-static-resource gating lives in resolveExpose, decoupled from the read/subscribe path that relies on it (packages/surface-mcp/src/expose.ts:247-270 vs packages/surface-mcp/src/server.ts:427-468)

srid:

Three passes (rule-checklist · fact-check · elegance) over the `surface-mcp` + `projectSurface` change (after codex-debate's 12 fixes and lens-debate's 10). **5 findings, all fixed** in `2154d346`.

| Pass | Finding | Fix |
| --- | --- | --- |
| **fact-check** | `isStringKey`/`decodeKey` mis-classified collection keys — a `z.literal`/`z.enum`/`z.uuid`/`.min(1)` string key fails `safeParse("")`, so the id segment was sent to `JSON.parse` and the item became unreachable | Try `keySchema.safeParse(id)` verbatim first; fall back to `JSON.parse` only on failure — handles every key-schema kind |
| **rules** · no-silent-error-swallowing | `void server.sendResourceUpdated({ uri })` discarded transport rejections (a debounced notify on a closed transport vanished) | `.catch((err) => console.error(...))` |
| **rules** · comment-the-non-obvious | `collectDefs` comment claimed it walks "nested pools"; it only indexes the root `$defs` | comment corrected (zod 4 only emits `$defs` at the root) |
| **elegance** | single-use `isAbort` closure in `deriveCell` | inlined to `isAbortReason(err, controller.signal)` |
| **elegance** | intermediate `toolSources` array built only to be iterated once | dropped; extracted `assertUniqueToolName`, iterate the source sets directly |

Rule checklist otherwise clean (DRY rule-of-three, invalid-states, unbounded-growth, dead-code, thin-wrappers, watcher-lifecycle — all checked, none triggered). 144 unit tests green; `just check` clean.

srid:

The **x86_64-linux** lane ran on **`kolu-ci-7`** (`dev-x86-64-linux-05`) — commit `8edc0aa6`, **passed** ✓

- **Lane wall** (pipeline): **2m29s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 1m51s | ✓ |
| `ci::nix` | 1m11s | ✓ |
| `ci::home-manager` | 1m3s | ✓ |
| `ci::smoke` | 37s | ✓ |
| `_ci-setup` | 29s | ✓ |
| `ci::unit` | 19s | ✓ |
| `ci::surface-example-build` | 10s | ✓ |
| `ci::surface-app-example-build` | 10s | ✓ |
| `ci::biome` | 9s | ✓ |
| `ci::install` | 8s | ✓ |
| `ci::fmt` | 8s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/8edc0aa/timings.jsonl`; pool state is a live `flock` probe.</sub>

srid:
## Evidence

This is a framework/library change with no kolu-app screen, so the behavior worth proving is **"an MCP client can drive a `@kolu/surface` through the adapter."** That's exercised end-to-end by `packages/surface-mcp/src/compose.test.ts` — an `InMemoryTransport` integration test that runs a **real MCP `Client`** against a served, *projected* surface — and it's green in the CI above (`ci::unit` on both platforms).

The composition it proves (the odu pattern in miniature):

```
source surface A: cell `nodes`, stream `nodeLog`, procedures `run.configure` (dangerous) + `node.rerun`
   → projectSurface ⟶ curated surface B: nodes(+red) cell · bounded `log` · `settled` event · `node.rerun` passthrough   (NO run.configure)
   → serveSurfaceAsMcp(B, expose={nodes,log:"resource", node.rerun:tool}, tools={run})  over InMemoryTransport
   → MCP Client connected to the other half
```

What the MCP client observes (asserted):
- **`tools/list` = `[node_rerun, run]`** — and **`run_configure` is absent** (default-deny, proven two ways: not opted into `expose` *and* structurally not on the projected surface).
- **`resources/list`** has `surface://cells/nodes` + `surface://streams/log`; **`resources/read`** returns the mapped snapshot.
- **`resources/subscribe` → `notifications/resources/updated` fires** when surface A's `nodes` cell is mutated — i.e. an update propagates **A → projected B → MCP host**.
- **`tools/call node_rerun`** flips A's real store; **the bespoke `run` tool** runs over the live client.

Plus the unit suites for each seam: `jsonschema` (`$ref`-deref / `io:"input"` / unrepresentable), `expose` (default-deny resolver + collisions), `pusher` (subscribe/teardown generation-token lifecycle), and `project` (derive teardown / abort propagation). **144 tests green.**

**Live-consumer proof:** [juspay/odu#11](https://github.com/juspay/odu/pull/11) migrates odu's hand-rolled `src/mcp/` onto this package — `odu mcp` now serves a projected `oduAgentSurface` — deleting ~1500 lines of bespoke MCP plumbing while keeping its agent-facing CI tools, green against this exact revision.

PR #1273, opened 2026-06-11
Engage the terminal scroll-lock only on scrolls the user actually made

### LINKED ISSUE #1272: Terminal frozen until keypress: scroll-lock latches on non-user scroll events and has no unlatch path on the desktop canvas
## Symptom

A terminal tile running Claude Code freezes — output stops painting — and stays frozen until the user presses any key, whereupon **all missed output appears instantly** ("as if the buffering got stuck"). It correlates with long Claude Code sessions (multi-agent workflows: minutes of spinner-only output, then a burst), and happens often. PR #1235 (WS heartbeat for half-open sockets) did not fix it — and was never going to (see "Exonerated" below).

**Confirmed live repro** (with screenshot): kolu window backgrounded (covered) while Claude Code was producing output → on switching back, the terminal was frozen **and the ScrollToBottom FAB was visible even though the user never scrolled** → one keypress flushed everything instantly.

## Root cause (verified)

The freeze is `scrollLock` buffering. The load-bearing bug — confirmed by adversarial code review on both a code-path lens and a symptom-fit lens — is that **kolu's scroll-lock is a one-way trap**:

1. **It latches on any off-bottom scroll event, not just user scrolls.** `scrollLock.ts:69-79` — the `term.onScroll` handler sets `isLocked = !atBottom` with no notion of whether the scroll was user-initiated. xterm fires `onScroll` for more than wheel input (e.g. `onBufferActivate` re-emits the activating buffer's raw `ydisp` on every alt↔normal screen switch — `BufferService.ts:45-47` in the shipped xterm).
2. **Once latched, kolu starves xterm's own self-healing.** All output goes to `pendingData` instead of `term.write` (`scrollLock.ts:92-99`). But xterm re-pins a non-user-scrolled viewport back to bottom *on the next written line* (`BufferService.ts:120-124` in the shipped 6.1.0-beta.225). By withholding writes, kolu makes a transient, would-have-self-healed divergence **permanent**.
3. **On the desktop canvas there is no unlatch path at all.** The only code that ever calls `scrollLock.reset()` is the `props.visible` transition effect (`Terminal.tsx:319-331`, `defer: true`) — and `App.tsx:374` hardcodes `visible={true}` for every canvas tile, so that effect is **dead code** on desktop. `refitOnTabVisible.ts:33-35` only refits, never resets. No timeout, no buffer cap. The only exits are a keypress (xterm's scroll-on-user-input, gated on `ybase !== ydisp` — `CoreService.ts:83-85` — fires the at-bottom `onScroll` that flushes) or clicking the FAB.
4. **The FAB doesn't communicate "output is paused".** It's a muted ~44px scroll-down affordance (`ScrollToBottom.tsx`); during a long Claude Code "thinking" stretch, a frozen spinner is indistinguishable from "still thinking".

Net effect: **any** spurious latch, however rare, becomes "terminal frozen until keypress" — which is why a rare trigger presents as "happens often".

### The trigger (narrowed, not yet closed)

Every candidate trigger was adversarially tested against the *actually shipped* xterm — which is **not 6.0.0**: the pnpm override pins `@xterm/xterm 6.1.0-beta.225` + `@xterm/addon-webgl 0.20.0-beta.224` (root `package.json`, landed 2026-05-26 in #979), with the rewritten VS-Code-style Viewport and the new DEC-2026 synchronized-output implementation. Findings:

- **Refuted**: spontaneous off-bottom from resize/reflow/zoom/font-load (xterm preserves `ydisp == ybase` through all of them; resize fires `onResize`, never `onScroll`), scrollback-cap trimming (self-healing unless already user-scrolling), attach-retry reset (unlocks, can't lock), DEC-2026/ED2 redraw misattribution as specified (central link false in shipped config), renderer-thaw and transport modes (both contradict the FAB being rendered during the freeze — the lock requires data *arriving and being processed*).
- **Standing candidates**, in order:
  - **Pre-cover unnoticed scroll input**: a wheel/touchpad tick seconds *before* covering the window latches the lock; the FAB is missed while switching; minutes of output buffer; user returns to the frozen screenshot. The shipped Viewport hardcodes `mouseWheelSmoothScroll: true` (`Viewport.ts:59`), so a tick's rAF animation can even be parked while hidden and delivered at uncover. The canvas pan-gesture wheel-yield (`TerminalCanvas.tsx:58-63` — `closest('.xterm')`, no focus check, `gestures.ts` 150ms ownership) makes "scrolled a terminal without meaning to" easy.
  - **Buffer-switch re-emission**: `ESC[?1049l` (alt→normal) re-fires `onScroll` with the normal buffer's stale `ydisp` (`BufferSet.ts:83-97` restores without re-pinning) — zero user action needed if any transient divergence existed. Unverified whether anything in a Claude Code session toggles 1049.
  - **kolu's touch-swipe bridge**: `Terminal.tsx:828` converts touch drags to `terminal.scrollLines(-n)` — kolu's only negative-`scrollLines` caller; a one-cell drag/palm-brush latches silently.

Related upstream context: xterm.js #5620 ("Scrollbar Locked at Top When Using AI CLI Tools") and #5801 (open; names Claude Code; ED2 inside DEC-2026 sync blocks mutates viewportY).

## Exonerated (so we don't re-chase these)

- **Server output pipeline**: traced node-pty → headless parse → Channel → attach generator → oRPC → `ws.send`; purely push-on-data, zero gating/batching/flow-control; `sendInput` touches no output-path state. Nothing a keypress could "flush". (Latent, *separate* finding: per-subscriber Channel overflow silently **ends** the attach stream — a permanent-freeze variant, not keypress-fixable; worth its own issue.)
- **Transport / #1235**: a dead socket cannot produce the repro — the FAB requires the client to have processed arriving data. Also verified the running build (`KOLU_COMMIT_HASH=2e9e5ca`) **contains** the #1235 merge (`485bbb5c`); the heartbeat works, it just fixes a different disease (frozen-until-*reload*, not frozen-until-*keypress*).
- **Renderer/WebGL/occlusion freezes**: the FAB rendering during the freeze proves the page was compositing; xterm's keypress scroll-to-bottom is gated on a genuinely off-bottom viewport, so every episode had real viewport divergence, not stale pixels.

## Fix plan (trigger-independent)

1. **Latch only on user intent**: arm the lock only when a real scroll input (wheel / touchmove / Shift+PageUp) occurred within a short window before the off-bottom `onScroll`; otherwise auto-`scrollToBottom()` and keep writing.
2. **Give the latch an exit**: reset + scroll-to-bottom on `visibilitychange`→visible (restores, for canvas tiles, what the dead `props.visible` effect was meant to do).
3. **Instrumentation to close the trigger question**: on the unlocked→locked transition, record `{viewportY, baseY, bufferType, visibilityState, hasFocus, stack}` into a ring buffer (every candidate emitter is synchronous, so the stack discriminates wheel/Viewport vs `scrollLines` touch-bridge vs `onBufferActivate` in one capture) and expose `isLocked` / `pendingData.length` / last-lock-record in the Debug → Diagnostic info dialog (`useTerminalDiagnostics.ts` currently exposes none of it).

## Investigation provenance

Diagnosed via a 30-agent multi-phase investigation (7 parallel hypothesis lanes over client/server/transport/renderer/history/upstream/tile-composition → synthesis → 2-lens adversarial verification per scenario → completeness critique), cross-checked against the shipped xterm sources at the exact pinned beta commit. Scroll-lock preference defaults ON for all users (`DEFAULT_PREFERENCES.scrollLock: true`, `packages/common/src/surface.ts:511`).

srid:
**Fixes the "terminal frozen until you press a key" bug** (#1272): a viewport scroll the user never made — a touch-bridge artifact, a TUI exiting its alternate screen re-emitting a stale position, a smooth-scroll tick delivered late after the tab was backgrounded — engaged the scroll-lock, which buffers *all* incoming output. On the desktop canvas every tile is mounted with a constant `visible={true}`, so the one effect that resets the lock never fires: any spurious latch became an indefinite freeze, released only by a keypress (xterm's scroll-on-user-input fires the at-bottom scroll that flushes). Meanwhile the latch also starves xterm of the very writes that would have let it self-heal a non-user-scrolled viewport.

### The latch is now gated on user intent

```
wheel / touch / Shift+PageUp / search jump ──▶ armUserScrollIntent(source)
                                                      │ 500ms window
xterm onScroll (off-bottom) ──▶ armed? ──yes──▶ latch + buffer (as before)
                                  │
                                  no──▶ record + snap back to bottom,
                                        output keeps flowing
```

- **Arm sources** report real scroll input right before xterm sees it (capture-phase wheel listener, the touch bridge ahead of its `scrollLines`, scroll-key chords in the key handler, `SearchBar`'s find navigation). Deliberate scrollback reading behaves exactly as before — including holding output while you read.
- **An unarmed off-bottom scroll is suppressed**: the viewport snaps back to the bottom and output keeps painting. _The snap-back is deferred by one microtask — the shipped xterm beta's Viewport silently drops re-entrant scrolls, which the e2e repro caught._
- **Returning to the tab releases an engaged lock**, mirroring the existing "switching back to a terminal auto-scrolls to bottom" semantics — a lock left behind while you were away must not greet you as a frozen terminal.

### Every transition is now observable

Each `locked` / `suppressed` / `unlatched` transition is recorded with viewport geometry, buffer type, visibility state, and a **stack capture** (every known emitter is synchronous, so the stack names the trigger) in a bounded ring. The Diagnostic Info dialog shows each terminal's live lock state and last transition; **Copy JSON** dumps the full ring — so if a freeze ever recurs in the wild, one report identifies the emitter.

> **Repro-first**: two new e2e scenarios were red against the old behavior — *programmatic viewport scroll must not freeze output* and *tab return must release a scroll-locked terminal* — and are green with the fix; 14 new unit tests pin the latch state machine (intent window expiry, lock persistence while reading, flush ordering, ring bounds). All 9 pre-existing scroll-lock/mobile-scroll scenarios still pass.

Closes #1272. Diagnosis history (30-agent investigation, exonerated subsystems, trigger candidates) lives in the issue.


srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `cf9587a20a2b`

### Round 1

**codex** — approved: `false`

The diff is focused and the new unit/e2e coverage exercises the intended #1272 paths, and `git diff --check` is clean. I found two user-facing behavior regressions in the new intent model, plus one over-broad keyboard intent that can reintroduce the false-latch class this branch is trying to remove.

Findings:
- `F1` · major · open — Any off-bottom xterm scroll that was not preceded by one of the newly wired intent sources is now treated as non-user and snapped back to the bottom. The call sites only arm wheel, touch bridge, Shift+Page/Home/End, and search; real user-driven xterm scrolls via the visible scrollbar and selection auto-scroll do not go through those paths. Those interactions will hit `source === null`, get suppressed, and fail to keep the user in scrollback or hold output. (packages/client/src/scrollLock.ts:194)
- `F2` · major · open — `handleTabVisible()` releases every engaged lock on browser tab return, regardless of whether the lock was an intentional user scroll made while the tab was visible. A user can scroll up to read output, switch browser tabs briefly, and come back to find their position lost and buffered output flushed to the bottom. (packages/client/src/scrollLock.ts:253)
- `F3` · minor · open — The keyboard intent guard arms on Shift+Home and Shift+End, but xterm's keydown path only scrolls the viewport for PageUp/PageDown result types; Home/End are sent as key sequences rather than viewport scrolls. This leaves a 500 ms stale scroll intent after a non-scrolling keypress, so an unrelated programmatic off-bottom scroll in that window can latch again. (packages/client/src/terminal/Terminal.tsx:680)

**claude** — Reviewed CODEX's three findings against the actual @xterm/xterm 6.1.0-beta.225 (Juspay fork) source via its sourcemap, and all three are correct. Fixed all of them.

F1 (agree, major): Confirmed the fork's custom scrollbar drags via POINTER_DOWN + a global pointermove monitor (abstractScrollbar.ts), and selection auto-scroll runs on a setInterval _dragScroll() driven by mousedown/mousemove (SelectionService.ts) — neither fires wheel/touch/keyboard/search, so both hit source===null and got snapped back. Wired a capture-phase `pointerdown` (primary button) on containerRef that calls a new `holdUserScrollIntent("pointer")`, with document-level `pointerup`/`pointercancel` calling `releaseUserScrollIntent()`. Held intent stays armed regardless of the 500ms window, because these gestures emit onScroll ticks for as long as the button is down — well past the window after the initial press.

F2 (agree, major): refitOnTabVisible fires on document `visibilitychange` (browser tab switches), so the unconditional handleTabVisible() would destroy an intentional foreground scroll when the user glanced at another browser tab and came back. Now I record `lockedWhileHidden` at engagement time (document.visibilityState === "hidden") and handleTabVisible() only releases when that's true; a lock made with the tab in front is preserved (position + buffered output kept).

F3 (agree, minor): Confirmed in Keyboard.ts that only Shift+PageUp/PageDown map to viewport scrolls (KeyboardResultType.PAGE_UP/PAGE_DOWN → scrollLines); Shift+Home/End emit escape sequences (ESC[H / ESC[F), no scroll. Narrowed the keyboard arming to Shift+PageUp/PageDown so a non-scrolling Home/End keypress no longer leaves a stale 500ms intent.

Added a small injectable `visibility` seam to createScrollLock (defaults to document.visibilityState) so the Node-environment unit suite can drive hidden/visible. Added unit tests for held-pointer intent (survives the window, expires on release) and the F2 cases (release a hidden-engaged lock, preserve a visible-engaged one). Full client suite green (288 tests), tsc --noEmit clean, biome format/lint clean.

- `F1` **fixed** — Confirmed via the fork's sourcemap: scrollbar drag uses POINTER_DOWN + GlobalPointerMoveMonitor (abstractScrollbar.ts:77,112), and selection auto-scroll fires onRequestScrollLines on a setInterval while the mouse is held (SelectionService.ts _dragScroll/_addMouseDownListeners) — neither routes through wheel/touch/keyboard/search, so both were suppressed. Added a `pointer` ScrollIntentSource plus holdUserScrollIntent/releaseUserScrollIntent in scrollLock.ts (intent stays armed past the 500ms window until release, since these gestures scroll repeatedly while held). Wired in Terminal.tsx: capture-phase `pointerdown` (primary button) on containerRef calls holdUserScrollIntent('pointer'); document-level capture-phase `pointerup` and `pointercancel` call releaseUserScrollIntent() (release on document because the drag routinely ends outside the terminal).
- `F2` **fixed** — Confirmed refitOnTabVisible.ts:33 binds document `visibilitychange`, so the original unconditional handleTabVisible() flushed and lost position on an ordinary browser-tab glance-and-return — a real regression for the intentional foreground-scroll case. Now scrollLock.ts records lockedWhileHidden = (visibility() === 'hidden') at the locked transition, and handleTabVisible() early-returns unless lockedWhileHidden, so only a background/accidental latch is cleared; a foreground user scroll keeps both its position and buffered output. lockedWhileHidden is reset on every lock release (at-bottom, reset, scrollToBottom).
- `F3` **fixed** — Confirmed in common/input/Keyboard.ts that only Shift+PageUp/PageDown (case 33/34) set KeyboardResultType.PAGE_UP/PAGE_DOWN, which CoreBrowserTerminal.ts:869 turns into scrollLines; Shift+Home/End (case 36/35) only set result.key to ESC[H / ESC[F (no viewport scroll). Narrowed the keyboard arming in Terminal.tsx to e.key === 'PageUp' || 'PageDown' (dropped Home/End), so a non-scrolling Home/End keypress no longer leaves a stale 500ms intent that an unrelated programmatic off-bottom scroll could latch onto.

commit: `b590bbe44540007f9caaf2e5d092184eb4a58082`

### Round 2

**codex** — approved: `true`

The current changes adequately address all three findings from the prior round. I verified the relevant current code paths and `git diff --check` is clean; I did not rerun CLAUDE's reported full test suite.

Findings:
- `F1` · major · resolved — Previously, real user scroll paths such as xterm scrollbar dragging and selection auto-scroll were not covered by the intent gate, so they could be suppressed as non-user scrolls. (packages/client/src/terminal/Terminal.tsx:679)
- `F2` · major · resolved — Previously, `handleTabVisible()` released every engaged lock on tab return, including locks intentionally created by a visible user scroll. (packages/client/src/scrollLock.ts:307)
- `F3` · minor · resolved — Previously, keyboard intent was armed for Shift+Home/End even though this xterm build does not turn those chords into viewport scrolls, leaving a stale intent window. (packages/client/src/terminal/Terminal.tsx:718)

_codex on the rebuttal:_ F1: CLAUDE agreed and fixed the missing pointer-held user scroll paths; I verified the held intent source and document-level release wiring. F2: CLAUDE agreed and fixed the over-broad tab-return unlatch; I verified foreground locks are preserved and hidden-engaged locks are still released. F3: CLAUDE agreed and narrowed the keyboard arming to PageUp/PageDown; the stale Home/End intent path is gone. No new regression from this round is worth raising.

srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `cf9587a20a2b`

Independent findings: lowy=3, hickey=8

### Applied (8)
- `lowy-3` ScrollLockDiagnostics live snapshot is constructed identically in two places — commit `1c7a536de`
- `hickey-1` User scroll intent fragmented across N call sites with an unenforced "arm before scroll" rule — commit `a597fc085`
- `hickey-2` pendingChunks duplicates pendingData.length — one fact maintained at four sites — commit `69354f63f`
- `hickey-3` hasNewOutput and pendingChunks encode the same buffered-output state in two flags — commit `b147f7283`
- `hickey-4` lastEvent is the tail of events() — two diagnostic channels carry one ring — commit `ab9188e1f`
- `hickey-5` onScroll handler braids five concerns into one closure — commit `7da4318ad`
- `hickey-7` lockedWhileHidden cleared at four sites — a flag whose invariant lives in code discipline — commit `ffe751ddb`
- `hickey-8` queueMicrotask snap-back is a lifecycle scope nested inside the onScroll handler — commit `700283a06`

### Agreed — no change (3)
- `lowy-1` Electricity test: createScrollLock is the receptacle; intent-source is the newly-encapsulated volatile axis (packages/client/src/scrollLock.ts:103)
- `lowy-2` Scroll-lock observability is sprayed across two parallel registration channels (packages/client/src/terminal/Terminal.tsx:641)
- `hickey-6` Two notions of "tab visible" — document.visibilityState vs props.visible — drive one unlatch (packages/client/src/scrollLock.ts:92-94,258,307-311 + Terminal.tsx:175,838-848)

srid:
## Evidence

Demonstrates the fix from #1272: output keeps flowing after a programmatic viewport scroll the user never initiated. Before the fix, this spurious scroll would engage the scroll-lock buffer and freeze terminal output until a keypress.



▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=scroll-lock-1272.mp4

srid:

The **x86_64-linux** lane ran on **`kolu-ci-2`** (`dev-x86-64-linux-04`) — commit `fcfd7e24`, _verdict unknown (no per-node timing)_

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

srid:

The **x86_64-linux** lane ran on **`kolu-ci-2`** (`dev-x86-64-linux-04`) — commit `fcfd7e24`, _verdict unknown (no per-node timing)_

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

PR #1034, opened 2026-05-29
Survive a deploy: split pty-host into a kolu --stdio daemon (R4c)

srid:
**Local terminals now survive a kolu-server restart.** A long-lived `kolu --stdio` PTY-host daemon owns the PTYs (and their scrollback + running agents); a deploy restarts kolu-server, which reattaches to the daemon's terminals by id — replacing tmux/zmx, closing #671. This is the **R4c redo**, correcting the dropped #1031, whose macOS production failure traced to daemonizing a survivor that *held* the provider DAG and so served stale detection after every deploy.

The fix is a **volatility split**: the daemon owns *only* the thin, rarely-changing `@kolu/pty-host` primitive; the volatile provider DAG (git / PR / agent-detection — the most-edited code in the repo) stays in kolu-server and runs **fresh on every deploy**, consuming the daemon's raw VT taps over a unix socket. A surviving daemon therefore can never go stale on the volatile code.

### Where each concern lives now

```
┌─ kolu --stdio daemon — thin · versioned · SURVIVES a deploy ─┐  one per $KOLU_STATE_DIR
│  node-pty + @xterm/headless mirror + raw VT taps              │  (unix socket: pty-host.sock)
└──────┬─────────────────────────────────┬─────────────────────┘
       │ byte streams (attach)           │ raw taps (cwd · title · cmd · foreground · exit)
       ▼                                 ▼
┌─ kolu-server — restarts every deploy ────────────────────────┐
│  provider DAG (providers.ts): git · PR · agent-detect — FRESH │
│  + dock · canvas · sessions · reattach-by-id at boot          │
└───────────────────────────────────────────────────────────────┘
```

### How survival works

- **Linux (systemd):** the supervisor spawns the daemon via `systemd-run --user --unit=kolu-pty-host` (gated on `INVOCATION_ID`), landing it in its *own* transient cgroup — so `systemctl --user restart kolu` no longer kills it. A plain `detached` spawn does **not** survive cgroup-v2's `KillMode=control-group` — *that was the #1031 Linux failure*. Spike-verified before building.
- **macOS (launchd):** a detached spawn already survives; used off-systemd (and in dev/test).
[...]

PR #1031, opened 2026-05-29
Run the local PTY agent in a `kolu --stdio` daemon — terminals survive a server restart (R4c)

srid:
**Local terminals now survive a kolu-server restart.** R4c of the remote-terminals roadmap (#951) moves the PTY-host "agent" — `@kolu/pty-host` plus the per-terminal provider DAG — out of the kolu-server process and into a separate **`kolu --stdio` daemon** behind a unix socket. Because the daemon outlives kolu-server, restarting the server reattaches the running shells **by id with scrollback intact** instead of killing them — replacing tmux/zmx and closing #671 (on dev / `nix run` / macOS; surviving a *systemd deploy* is R4d).

This is an **encoding of the boundary R4b already drew**, not a new one. R4b put the provider DAG inside an in-process agent (`createAgent`) emitting an `AgentMetadataEvent` stream over an in-process `Channel`; R4c keeps that exact event shape and ships it over a socket. So `LocalTerminalBackend` becomes a *pure socket client* — the metadata demux, the `terminals:dirty` autosave fence, the activity feed, and `fs/git` all stay put.

### How it splits

```
┌─ kolu --stdio daemon (own process, survives a server restart) ─┐
│  createAgent → @kolu/pty-host  +  provider DAG (git·PR·agent)   │
│  serves agentSurface over $KOLU_STATE_DIR/agent.sock           │
└───────┬───────────────────────────────┬────────────────────────┘
        │ terminalAttach (snapshot+delta)│ agentMetadata (events)
        ▼  unix socket (stdio link)      ▼
┌──────────────────────── kolu-server ──────────────────────────┐
│  LocalTerminalBackend = socket client + metadata consumer      │
│  supervisor: connect-or-spawn · version handshake · reattach   │
└────────────────────────────────────────────────────────────────┘
```

### The pieces

- **`kolu-common/agentSurface.ts`** — the typed wire contract: `terminalAttach` + `agentMetadata` streams (the latter carries the `AgentMetadataEvent` union, preserving the `metadataPersisted`/`metadataLive` autosave fence across the wire) + `terminal.*` / `system.{version,heartbeat}` procedures, plus `AGENT_CONTRACT_VERSION` + `isAgentContractCompatible`.
[...]

PR #1275, opened 2026-06-11
R-4 Phase B: terminals survive a server restart — the pty-host daemon

### LINKED ISSUE #671: zmx integration
[zmx](https://github.com/neurosnap/zmx) captures foreground process as itself, so we lose agent information (see screenshots).

- What are some ways Kolu can integratee with zmx to function alongside it?
- How do we get the 'underlying process' information, so our agent inspection continues to work for users using zmx?

## Screenshots

Right panel showing zmx as foreground process:

Image

Codex agent information missing in title bar:

Image

srid:
**Local terminals — shells *and* the agents running in them — now survive a kolu-server restart.** Until now every deploy killed your terminals: the server owned the PTYs in-process, so restarting it (or shipping a new build) tore them down. Phase B moves the PTYs into a **surviving, out-of-process `pty-host` daemon** in its own cgroup. The server becomes a *client* of that daemon, and a fresh server after a deploy **reattaches to the still-running PTYs by id** — same shells, same scrollback, the same Claude/Codex mid-session, **no restore card**. It's the tmux/zmx replacement. Closes #671.

> Caught live in prod: `SRV … up 1m` (just redeployed) beside `PTY … up 2h` (the daemon outlived the deploys), with an honest `⬆ update pending` because the surviving daemon is a build behind.

## The flow — first spawn, deploy-reattach, supervised restart

```mermaid
sequenceDiagram
  autonumber
  actor B as Browser
  participant Srv as kolu-server<br/>(restarts every deploy)
  participant DH as DaemonHandle<br/>@kolu/pty-host-daemon
  participant D as pty-host daemon<br/>own cgroup · SURVIVES
  participant T as PTY + shell/agent

  rect rgb(234,244,255)
  Note over Srv,D: 1 · First boot — cold start, no daemon yet
  Srv->>DH: ensureDaemon()
  DH->>DH: readPidGate() → absent
  DH->>D: spawn detached — systemd-run --user (Linux, own cgroup) / detached+unref (macOS)
  D->>D: acquire O_EXCL pid-gate, bind unix socket
  DH->>D: connectWithRetry() until the socket answers
  B->>Srv: new terminal
  Srv->>D: terminal.spawn(id)
  D->>T: fork node-pty + @xterm/headless mirror
  T-->>B: live bytes over the attach stream
  end

  rect rgb(233,255,236)
  Note over Srv,D: 2 · Deploy — server restarts, the daemon outlives it
  Note over Srv: ⟂ old server exits — its cgroup dies, the daemon's does not
  Note over D,T: PTYs, scrollback and running agents stay live
  Srv->>DH: ensureDaemon() — fresh server process
  DH->>DH: readPidGate() → live pid
  DH->>D: reconnect to the surviving socket — NO spawn
  Srv->>D: terminal.list() → surviving PTY ids
  Srv->>Srv: adoptSurvivors() — reattach-by-id, re-run providers, NO respawn
  T-->>B: same shells and agents, no restore card
  end

  rect rgb(255,239,233)
  Note over Srv,D: 3 · Restart the daemon (new build / forced) — #1034-proof
  B->>Srv: "Restart pty-host daemon"
  Srv->>Srv: snapshot session → setSavedSession (wins the autosave-cancel race)
  Srv->>DH: restart()
  DH->>D: kill(pid)
  DH->>DH: waitForPidGone() — block on real exit, load-aware ceiling
  DH->>D: spawn fresh + reconnect
  Note over Srv,T: a failed respawn KEEPS the saved session — never an empty canvas
  end
```

### How it's supervised + backgrounded

- **Survives by cgroup, not luck.** On Linux the daemon is launched with `systemd-run --user`, landing in its **own transient cgroup** — a plain detached child dies on cgroup-v2 (`KillMode=control-group` walks cgroup membership; the #1031 failure). macOS's `detached + unref` already reparents to init. That launch-to-survive *is* the survival, which is why it lives with the supervisor (below), not as a generic "spawn".
- **Single instance, held until death.** An `O_EXCL` pid-gate plus a per-socket/per-spawn systemd unit name keep exactly one daemon per socket; `readPidGate()` is how a fresh server discovers a survivor.
- **A reconnecting client.** `DaemonHandle` hands the server one *stable* `PtyHostClient` (a reconnecting proxy) that survives a daemon restart — so `local.ts`'s module-level reference never goes stale when the transport is swapped underneath it.

## Reattach-by-id — why there's no restore card

The server already mints each PTY's id, so at boot it lists the daemon's surviving PTYs, joins them to the saved session **by id with no respawn**, and re-runs the (cheap, fresh-every-deploy) provider DAG against the live taps. Metadata is briefly cold (a re-detection pass), never stale. Agents come back *running* — and `lastAgentCommand` is carried through the adopt so a later cold-start restore can still offer `claude -c`.

## #1034-proof recovery

The composed restart (snapshot → save → drain → **`waitForPidGone`** → respawn → reattach) is shared verbatim by the user-clicked and forced paths, so neither can be written kill-then-pray. Snapshot-and-save precede every kill; a failed respawn leaves a restorable session and a retry, **never an empty canvas** — the bug the postmortem paid for.

## Honest state — the `srv · pty` rail

The rail makes the two processes' divergence visible: `≡ current` when the daemon is the deployed build, `⬆ update pending` when it's a build behind (keyed on the pty-host closure **staleKey**, so a server-only deploy never over-prompts), and `—` when a restart fails. This PR also adds **live uptime** per column — the `pty up 3h` vs `srv up 2m` gap is glanceable proof the daemon outlived the server.

## The structure — `@kolu/pty-host-daemon`

The daemon glue is extracted along a volatility seam (a `/lowy` + `/hickey` pass): the local surviving-daemon lifecycle is a concern distinct from the server's terminal/session/UI domain, and the per-pty-host direction makes it one of **two siblings** behind one receptacle.

```mermaid
flowchart TD
  SRV["packages/server — domain only<br/>adoptSurvivors · reattach · buildInfo · IdentityRail"]
  PHD["@kolu/pty-host-daemon — NEW<br/>LOCAL surviving-daemon supervisor<br/>daemonHandle · waitForPidGone · spawn(systemd) · env · socketEnv"]
  SNH["surface-nix-host — sibling (future R-2)<br/>REMOTE surviving-daemon supervisor<br/>ssh reach · provision · bridge"]
  PH["@kolu/pty-host<br/>a runnable pty-host<br/>contract · serve · socketPath · pid-gate writer · daemonMain entry"]
  SRV -->|location = local| PHD
  SRV -. location = remote, R-2 .-> SNH
  PHD -->|spawns + supervises| PH
  SNH -. provisions + runs over ssh .-> PH
```

The unifying `PtyHostProvider` interface is deferred to R-2 (the `TerminalBackend` seam is the join point; `daemonMain` folds into `@kolu/pty-host` so a remote agent runs the entry without importing the supervisor).

## What's inside

| Area | Highlights |
| --- | --- |
| **Daemon foundation** | `runPtyHostDaemon` serve loop · `DaemonHandle` reconnecting client + restart · `waitForPidGone` barrier · single-instance pid-gate · detached/systemd spawn |
| **The one-way door** | server consumes the surviving daemon over the unix socket (the role kolu-tui already played) |
| **Reattach** | reattach-by-id at boot · `parentId` preserved (splits stay nested) · contract-skew validation → restart on incompatible adopt · `lastAgentCommand` carried so `claude -c` resumes |
| **Recovery** | composed #1034-proof restart + the *Restart pty-host daemon* command |
| **Honest state** | rail's `pty` column (current / update-pending / —) · live `srv · pty` uptime · multi-host seam markers |
| **Structure** | extract `@kolu/pty-host-daemon`; server left domain-only |
| **Review** | codex debate (2 rounds) + lowy⇄hickey lens fixes (koluRoot dedup, env/launch split, identity readout, dead-wire drops, …) |

> **Acceptance is the *second* deploy, not the first** — survival, reattach, and the restart race are only exercised when this build redeploys over a live daemon with running agents. CI green is necessary, not sufficient; the real gate is a staged prod redeploy (done — see the screenshot above).

### Try it locally

```sh
nix run github:juspay/kolu/pty-daemon-phase-b
```


srid:

The **x86_64-linux** lane ran on **`kolu-ci-4`** (`dev-x86-64-linux-08`) — commit `795cd595`, **passed** ✓

- **Lane wall** (pipeline): **2m39s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 1m53s | ✓ |
| `ci::nix` | 1m6s | ✓ |
| `ci::home-manager` | 1m1s | ✓ |
| `ci::smoke` | 37s | ✓ |
| `_ci-setup` | 37s | ✓ |
| `ci::pnpm-hash-fresh` | 37s | ✓ |
| `ci::unit` | 23s | ✓ |
| `ci::biome` | 13s | ✓ |
| `ci::fmt` | 12s | ✓ |
| `ci::install` | 7s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|
| `kolu-ci-4` | dev-x86-64-linux-08 | 🔒 leased |

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/795cd59/timings.jsonl`; pool state is a live `flock` probe.</sub>

srid:

The **x86_64-linux** lane ran on **`kolu-ci-4`** (`dev-x86-64-linux-08`) — commit `eb37ac37`, **passed** ✓

- **Lane wall** (pipeline): **9m12s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 8m30s | ✓ |
| `ci::home-manager` | 33s | ✓ |
| `_ci-setup` | 28s | ✓ |
| `ci::unit` | 23s | ✓ |
| `ci::nix` | 19s | ✓ |
| `ci::biome` | 17s | ✓ |
| `ci::smoke` | 15s | ✓ |
| `ci::install` | 14s | ✓ |
| `ci::fmt` | 13s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|
| `kolu-ci-4` | dev-x86-64-linux-08 | 🔒 leased |

srid:

The **x86_64-linux** lane ran on **`kolu-ci-4`** (`dev-x86-64-linux-08`) — commit `b097b058`, **passed** ✓

- **Lane wall** (pipeline): **2m30s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 1m40s | ✓ |
| `ci::home-manager` | 1m5s | ✓ |
| `ci::smoke` | 35s | ✓ |
| `ci::nix` | 34s | ✓ |
| `_ci-setup` | 30s | ✓ |
| `ci::unit` | 24s | ✓ |
| `ci::install` | 19s | ✓ |
| `ci::fmt` | 13s | ✓ |
| `ci::biome` | 13s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/b097b05/timings.jsonl`; pool state is a live `flock` probe.</sub>

srid:

The **x86_64-linux** lane ran on **`kolu-ci-4`** (`dev-x86-64-linux-08`) — commit `3966aa8c`, **passed** ✓

- **Lane wall** (pipeline): **9m12s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 8m30s | ✓ |
| `ci::home-manager` | 33s | ✓ |
| `_ci-setup` | 28s | ✓ |
| `ci::unit` | 23s | ✓ |
| `ci::nix` | 19s | ✓ |
| `ci::biome` | 17s | ✓ |
| `ci::smoke` | 15s | ✓ |
| `ci::install` | 14s | ✓ |
| `ci::fmt` | 13s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

srid:

The **x86_64-linux** lane ran on **`kolu-ci-5`** (`idliv2-02`) — commit `7c762c8f`, **passed** ✓

- **Lane wall** (pipeline): **3m20s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 2m8s | ✓ |
| `ci::nix` | 1m13s | ✓ |
| `ci::home-manager` | 1m13s | ✓ |
| `_ci-setup` | 1m4s | ✓ |
| `ci::smoke` | 37s | ✓ |
| `ci::unit` | 20s | ✓ |
| `ci::surface-example-build` | 10s | ✓ |
| `ci::biome` | 10s | ✓ |
| `ci::fmt` | 9s | ✓ |
| `ci::install` | 7s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|
| `kolu-ci-4` | dev-x86-64-linux-08 | 🔒 leased |

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/7c762c8/timings.jsonl`; pool state is a live `flock` probe.</sub>

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 round(s) · codex reviewed at `xhigh` reasoning effort · base `aaf41f2ea8bd`

### Round 1

**codex** — approved: `false`

The branch makes the right architectural move toward a surviving PTY daemon, but the safety surface is not complete yet. The core daemon path has a real pid-gate race, the production systemd launch path mishandles dynamic environment, and the UI/state plumbing still cannot honestly represent daemon currency or degradation. I did not run tests because this review was requested read-only.

Findings:
- `F1` · major · open — The pid-gate is not actually atomic. `openSync(path, "wx")` creates an empty file, then the winner writes its pid later. A concurrent starter that sees `EEXIST` in that window calls `readPidGate()`, gets `null` for the empty/malformed file, unlinks it as stale, and can acquire its own gate while the first process still returns `acquired`. That defeats the daemon single-instance guarantee. (packages/shared/src/pid.ts:108)
- `F2` · major · open — The systemd launch path passes `KOLU_PTY_HOST_SOCKET` only in the environment of the `systemd-run` client process. Transient service units run in the user manager's clean environment, so the daemon service will not receive this dynamic socket override. Under systemd, `--pty-host-socket` can therefore spawn a daemon on the default socket while the server polls the override and times out. (packages/server/src/daemon/spawn.ts:63)
- `F3` · major · open — The pty update-pending derivation compares server commit to daemon commit. That reintroduces the over-prompting bug the staleKey was meant to prevent: any server/client-only deploy changes `i.commit` while the pty-host closure may be identical, so the rail flags `outdated` and offers a pointless daemon restart. (packages/client/src/ui/IdentityRail.tsx:67)
- `F4` · major · open — The pty-host identity published to `buildInfo` is backed by the one-shot `ptyHostIdentity` promise from module load. After `restartPtyHost()` successfully spawns a new daemon, that promise still contains the old daemon identity, and a browser reload will just read the stale value again. The rail can keep showing update-pending after a successful restart. (packages/server/src/surface.ts:316)
- `F5` · blocking · open — There is still no honest daemon health surface. `ptyState()` only looks at the browser WebSocket status and the last buildInfo identity; if the daemon dies or `daemonHandle` enters `degraded` while the server remains up, the rail can stay green/amber and the canvas has no degraded state. This leaves the empty-canvas/dead-daemon lie the docs say B closes. (packages/client/src/ui/IdentityRail.tsx:63)
- `F6` · major · open — `restart()` disposes the only live connection before the restart is known to be recoverable. On timeout or respawn-connect failure it leaves `conn` pointing at a disposed connection and returns `failed`; the client then tells users to reload, but a browser reload still talks to the same degraded server-side handle and cannot restore terminals. (packages/server/src/daemon/daemonHandle.ts:171)
- `F7` · minor · open — `openDaemonLog()` returns a file descriptor that the parent never closes after spawning. Each daemon spawn/restart leaks one fd in the server process. (packages/server/src/daemon/spawn.ts:59)
- `F8` · minor · open — The systemd unit name is always `kolu-pty-host`. Even with isolated socket/pid paths, two systemd-backed kolu instances or a rapid restart before `--collect` unloads the old unit can collide on the unit name. (packages/server/src/daemon/spawn.ts:70)

**claude** — Addressed all 8 CODEX findings. Fixed F1 (atomic pid-gate via write-temp-then-link), F2 (forward daemon env into systemd-run units with --setenv), F3 (rail currency now keyed on closure staleKey vs the server's published ptyHostExpectedStaleKey, not commit), F4 (live daemon identity read + republishBuildInfo after restart), F6 (restart keeps the live connection when the old daemon won't die; client toast offers a retry, not a misleading reload), F7 (close the parent log fd after spawn), F8 (per-socket-unique systemd unit name). F5 is partial: wired the cheap/correct honest-state pieces (failed restart resolves the pty column to — via the streaming buildInfo cell, plus the F3/F4 work), but the full unsolicited-daemon-death surface (a liveness probe + streaming DaemonStatus cell + DegradedCanvas) is a follow-up feature beyond this round — and I corrected the atlas doc, which overclaimed that surface as shipped. Typecheck green across all touched packages; pid + daemonHandle unit tests pass.

- `F1` **fixed** — Agreed — the openSync('wx') gate created an empty file before the pid was written, so a concurrent EEXIST-reader saw a null gate and could reclaim a live winner's slot. Rewrote acquirePidGate in packages/shared/src/pid.ts to publish atomically: write the pid in full into a private `${path}.${pid}.tmp`, then linkSync it into place (link fails EEXIST if the gate exists, exactly like O_EXCL, but the file that appears is already complete — no empty-then-written window). Added a regression test in pid.test.ts asserting the published gate always parses to a pid and leaves no .tmp sidecar.
- `F2` **fixed** — Agreed — a `systemd-run --user` transient unit runs in the user manager's clean environment, so spawn()'s `env` only reached the systemd-run CLIENT, not the daemon. In packages/server/src/daemon/spawn.ts added setenvArgs() forwarding the daemon-needed dynamic vars (KOLU_PTY_HOST_SOCKET, KOLU_NIX_ENV_WHITELIST, KOLU_PTY_HOST_BUILD_ID, KOLU_COMMIT_HASH, KOLU_DAEMON_BIN, LOG_LEVEL) into the unit via --setenv=K=V, skipping absent/empty vars. The override socket now lands in the unit instead of silently spawning on the default.
- `F3` **fixed** — Agreed — comparing i.commit to ptyHost.navigableCommit reintroduced over-prompting (any server/client-only deploy bumps commit while the pty-host closure is byte-identical). The codebase's own buildId.ts already documents staleKey as the correct currency key. Added ptyHostExpectedStaleKey to KoluBuildInfo (packages/common/src/surface.ts), populated server-side from currentBuildId() (the deployed build's KOLU_PTY_HOST_BUILD_ID) in surface.ts, and rewrote IdentityRail.tsx ptyState() to compare daemon.ptyHost.staleKey against ptyHostExpectedStaleKey — either absent ⇒ unknown, never a false outdated.
- `F4` **fixed** — Agreed — buildInfo was backed by a one-shot module-load promise (ptyHostIdentity), so after a successful restartPtyHost the cell still held the OLD daemon's identity and a reload re-read the stale value. Replaced the frozen promise in terminalBackend/local.ts with readPtyHostIdentity() (a live read over the socket). surface.ts now exposes republishBuildInfo() which re-reads the live daemon identity and pushes it through the same buildInfo cell ctx the boot connect used; router.ts calls it after daemonHandle.restart(), so the rail drops the nudge without a reload (server-pushed via the streaming cell).
- `F5` **partial** — Agreed the gap is real and the atlas doc overclaimed it as shipped. Fixed the bounded/correct part: a KNOWN failed restart now resolves the pty column to — honestly via the streaming buildInfo republish (F4), and the F3/F4 currency work makes the green/amber states truthful. NOT fixed in this round: an UNSOLICITED daemon death while the server stays up — surfacing that live requires a continuous daemon-liveness probe feeding a streaming DaemonStatus cell that a dedicated DegradedCanvas subscribes to (the one-shot server.info RPC can't push a death). That's a feature-sized subsystem, not a review fix. Corrected docs/atlas/pty-daemon.mdx (which claimed the DaemonStatus cell, DegradedCanvas, and dead-daemon rail state shipped) to mark them as the remaining follow-up, softened 'R-4 is complete' to 'R-4's core is complete', and added a Callout naming the liveness-probe follow-up.
- `F6` **fixed** — Agreed — restart() disposed the only live connection up front, so on a barrier timeout it left conn pointing at a disposed connection while the old daemon was still alive, and the client told users to reload (which can't fix a server-side handle). In daemonHandle.ts, restart() now holds the old connection until it commits to the swap: on the 'old daemon won't die' path it returns degraded WITHOUT disposing conn (the daemon is reachable — that's why the barrier timed out — so terminals keep working). Added a test asserting the client still reaches the daemon after that degraded outcome. In App.tsx the failure toast no longer says 'reload to recover'; it offers a 'Try again' action that re-invokes restartPtyHost (which re-reads the pid gate and re-attempts), the honest recovery for a server-side handle.
- `F7` **fixed** — Agreed — openDaemonLog()'s fd was inherited by the detached child but never closed in the parent, leaking one fd per spawn/restart. Added closeSync(out) in spawn.ts after spawn() returns (the child holds its own dup of the stdio fds, so closing the parent copy is safe and preserves the daemon's logging).
- `F8` **fixed** — Agreed — the fixed --unit=kolu-pty-host name could collide between two systemd-backed kolu instances (distinct sockets) or a respawn racing the old unit's --collect teardown. Added daemonUnitName(socketPath) in spawn.ts deriving kolu-pty-host-<12-hex-of-sha256(socketPath)> — unique per socket (the discriminator the pid gate already uses as the logical singleton), collision-free across instances and generations, and valid for systemd unit names.

commit: `7d67ac71ab483c3ecec5fecc68f9af1be443423a`

### Round 2

**codex** — approved: `false`

Most of the prior findings are fixed in the current tree: the pid gate is now atomically published, systemd units receive the daemon env, stale PTY currency uses the closure staleKey, live daemon identity is re-read after restart, restart no longer strands the old live connection on a kill timeout, the parent log fd is closed, and the UI failure recovery is now retry-based. Two issues remain open: the claimed failed-restart honest-state path still preserves stale PTY identity, and the systemd unit-name change solves cross-instance collisions but not the same-socket generation race from the original finding.

Findings:
- `F1` · major · resolved — The pid gate could previously be observed as an empty or partially written file, letting a contender treat a live winner as stale. (packages/shared/src/pid.ts:102)
- `F2` · major · resolved — The daemon env was only applied to the systemd-run client process, not to the transient unit. (packages/server/src/daemon/spawn.ts:70)
- `F3` · major · resolved — The rail used commit hash as PTY currency, which over-prompts when the PTY-host closure did not change. (packages/client/src/ui/IdentityRail.tsx:68)
- `F4` · major · resolved — Daemon identity was previously read once and remained stale after restart. (packages/server/src/surface.ts:92)
- `F5` · major · open — The bounded failed-restart fix still does not clear stale PTY identity. buildInfoValue omits ptyHost when readPtyHostIdentity() fails, but republishBuildInfo merges that patch over the existing cell, so the previous ptyHost survives and the client can continue showing current/outdated instead of the promised unknown/dash state after a failed restart. (packages/server/src/surface.ts:381)
- `F6` · major · resolved — Restart previously disposed the only live connection before proving the old daemon was gone, leaving the server stranded on timeout and telling users to reload. (packages/server/src/daemon/daemonHandle.ts:181)
- `F7` · minor · resolved — The parent process leaked the daemon log fd after each spawn. (packages/server/src/daemon/spawn.ts:137)
- `F8` · major · open — The new unit name is per socket, not per daemon generation. That fixes collisions between different socket paths, but a same-socket restart still uses the identical systemd unit name and can still race the old transient unit's --collect cleanup after the process has exited. (packages/server/src/daemon/spawn.ts:95)

_codex on the rebuttal:_ F1 accepted: link-based publication removes the empty-gate observation window. F2 accepted: --setenv forwards the relevant environment into the transient unit. F3 accepted: staleKey is the right PTY-host currency. F4 accepted for the successful restart path: identity is now read live and republished. F5 not fully accepted: I agree the unsolicited daemon-death surface can be a follow-up if the docs no longer claim it shipped, but the claimed bounded fix for a known failed restart is still broken because republishBuildInfo merge-preserves the old ptyHost when the new identity read fails. F6 accepted: the old live connection is retained on kill-timeout and the UI recovery is retry-based. F7 accepted: the parent fd is closed. F8 not accepted: the per-socket hash solves cross-instance unit collisions, but it does not create a fresh generation name, so the same-socket --collect race remains.

**claude** — Addressed the two findings CODEX held open (F5, F8); the other six were already resolved in prior rounds. Conceded both — CODEX was technically right on each. F5: buildInfoValue omitted the ptyHost key on a failed identity read, so republishBuildInfo's spread-merge preserved the OLD daemon's identity, leaving the rail lying current/outdated after a known-failed restart instead of dropping to the promised — (the router comment even claimed the column resolves to —, but the merge defeated it). Fixed by always carrying ptyHost (identity ?? undefined) so the republish overwrites stale identity to absent; JSON.stringify drops the undefined key so the client reads no staleKey and ptyState() returns unknown (grey dot = honest —), while equals still sees a change so the publish fires. No boot regression. F8: the per-socket sha kept the systemd unit name identical across same-socket restarts, so a respawn racing the old transient unit's --collect GC could hit "unit already exists" and spuriously degrade — waitForPidGone observes the process exiting, a distinct event from systemd GC'ing the unit object. Fixed by appending a per-spawn randomBytes(4) generation tag to the unit name. Typecheck green across all packages; daemonHandle unit tests pass 4/4; biome format clean.

- `F5` **fixed** — Conceded. buildInfoValue (packages/server/src/surface.ts:92) omitted ptyHost when readPtyHostIdentity() failed, and republishBuildInfo (surface.ts:389) merged that patch via { ...cell.get(), ...patch }, so the spread preserved the OLD daemon's ptyHost — the rail kept showing current/outdated after a known-failed restart instead of the promised —. The router.ts:102 comment even CLAIMED 'the read resolves the column to —', but the merge silently defeated it. Fix: buildInfoValue now ALWAYS carries the ptyHost key (ptyHost: identity ?? undefined), so the republish OVERWRITES the stale identity. JSON.stringify drops the undefined key (wire-absent → client reads i?.ptyHost?.staleKey undefined → IdentityRail ptyState() returns 'unknown' → grey dot, no commit/staleKey = the honest —), and the equals dedup still sees a change vs. the prior ptyHost-bearing value so the publish fires. No boot regression: daemon is up at boot so identity is set; if down, ptyHost: undefined is behaviorally identical to the old key-absent path.
- `F8` **fixed** — Conceded. daemonUnitName (packages/server/src/daemon/spawn.ts:95) hashed only socketPath, so a same-socket restart reused the IDENTICAL unit name kolu-pty-host-<hash>. The restart flow kills the daemon and waits via waitForPidGone — which observes the PROCESS exiting, a distinct event from systemd GC'ing the unit OBJECT via --collect. A respawn firing in that window hits systemd-run --unit=<same> 'unit already exists' and the restart spuriously degrades. Fix: daemonUnitName now appends a per-spawn randomBytes(4) generation tag (kolu-pty-host-<socketTag>-<genTag>), so every restart lands a fresh unit name that can't collide with a still-loaded old one. --collect GCs each generation independently; the pid gate stays the logical singleton guard, so a per-spawn unit name costs nothing.

commit: `157446e6f94249dfa06acb75f706a3d577ba9e84`

### Round 3

**codex** — approved: `true`

The remaining two findings are addressed in the current tree. F5 now explicitly includes `ptyHost: identity ?? undefined` in the republished patch, so the merge no longer preserves a stale daemon identity after a failed identity read. F8 now adds a per-spawn random generation tag to the systemd unit name, so same-socket restarts no longer reuse a unit name while the old transient unit may still be loaded. I did not find a regression introduced by these latest fixes.

Findings:
- `F1` · major · resolved — The pid gate could previously be observed as an empty or partially written file, letting a contender treat a live winner as stale. (packages/shared/src/pid.ts:102)
- `F2` · major · resolved — The daemon env was only applied to the systemd-run client process, not to the transient unit. (packages/server/src/daemon/spawn.ts:70)
- `F3` · major · resolved — The rail used commit hash as PTY currency, which over-prompts when the PTY-host closure did not change. (packages/client/src/ui/IdentityRail.tsx:68)
- `F4` · major · resolved — Daemon identity was previously read once and remained stale after restart. (packages/server/src/surface.ts:92)
- `F5` · major · resolved — The failed-restart republish path previously omitted ptyHost when the live identity read failed, so the spread merge preserved stale PTY identity. (packages/server/src/surface.ts:104)
- `F6` · major · resolved — Restart previously disposed the only live connection before proving the old daemon was gone, leaving the server stranded on timeout and telling users to reload. (packages/server/src/daemon/daemonHandle.ts:181)
- `F7` · minor · resolved — The parent process leaked the daemon log fd after each spawn. (packages/server/src/daemon/spawn.ts:142)
- `F8` · major · resolved — The systemd unit name previously remained stable for same-socket restarts, so respawn could race the old transient unit's --collect cleanup. (packages/server/src/daemon/spawn.ts:99)

_codex on the rebuttal:_ F5 accepted: carrying `ptyHost: undefined` in the patch fixes the exact merge-preservation bug I held open. Because the previous value had a concrete ptyHost object, the JSON.stringify-based equality check sees the change and publishes; on the wire the undefined key is absent, which makes the client derive the unknown PTY state. F8 accepted: the unit name is now both per-socket and per-spawn, so it avoids the same-socket systemd unit-object GC race while the pid gate remains the singleton guard. All prior findings are now resolved.

srid:

✅ **Consensus** after 2 round(s) · lowy + hickey · base `aaf41f2ea8bd`

### Applied (9)
- `lowy-1` Duplicated temp-root receptacle: koluRootFor vs server/koluRoot.ts — commit `056815883`
- `lowy-2` Server's koluShellDir is now a dead wire (created, never consumed) — commit `8dcf0a2c1`
- `hickey-1` Two koluRoot abstractions for one concept — (uncommitted)
- `hickey-2` DaemonState / handle.state() is built but never consumed — commit `8c6ff7ce6`
- `hickey-3` Half-applied client injection in LocalTerminalBackend — commit `98fc91001`
- `hickey-4` buildInfoValue's three-axis patch fragments one identity readout — commit `4b4870372`
- `hickey-5` spawn.ts braids env-forwarding policy with launch mechanics — commit `4e319d7d5`
- `hickey-6` getPtyHostPidPath derives its path two incompatible ways — commit `d876ee0d8`
- `hickey-7` App.tsx restart handler owns multi-step RPC workflow against SolidJS convention — commit `c02c20c22`

### Agreed — no change (3)
- `lowy-3` pidIsAlive duplicated between kolu-shared and waitForPidGone's doc/default (packages/server/src/daemon/waitForPidGone.ts:25-29 and packages/shared/src/pid.ts:30)
- `lowy-4` DaemonHandle.restart mixes two volatility axes: process-lifecycle and UI-state republish trigger (packages/server/src/router.ts:91-105 and packages/server/src/daemon/daemonHandle.ts:169-218)
- `lowy-5` DAEMON_ENV_KEYS allowlist re-encodes the daemon's config dependencies in a second place (packages/server/src/daemon/spawn.ts:58-65)

srid:

Ran the rule-checklist, fact-check, and elegance passes over the Phase B diff (after the codex + lowy/hickey gauntlet).

**Fixed (1):**
- `packages/server/src/surface.ts` — dropped a redundant `?? undefined` no-op in `buildInfoValue` (`ptyHost: identity ?? undefined` → `ptyHost: identity`). `identity` is already `PtyHostIdentity | undefined`, so the operator returned it unchanged; the load-bearing part of codex's F5 fix — the `ptyHost` **key** being always-present so a republish overwrites stale identity to absent — is preserved. (`fix(police):` commit.)

**Skipped (4, with reasoning):**
- `sleep` one-liner duplicated across `daemonHandle.ts`/`waitForPidGone.ts` — the latter is an injectable test-seam; cross-module coupling for a one-liner costs more than it saves.
- `ptyState()` read twice in `IdentityRail.tsx` — idiomatic SolidJS; caching into a `const` would freeze the initial value (the anti-pattern the project rules warn against).
- `survivorsPromise` never reset after a daemon restart — intentional: the survivor list is exhausted via `delete` during the boot restore burst; a restarted daemon has no survivors to relist.
- currency comparison "duplicated" in `buildInfoValue`/`republishBuildInfo` — false positive: `republishBuildInfo` *calls* `buildInfoValue`.

_Posted by `/be-review` after the push._

srid:

The **x86_64-linux** lane ran on **`kolu-ci-4`** (`dev-x86-64-linux-08`) — commit `139c7187`, **passed** ✓

- **Lane wall** (pipeline): **4m2s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 3m23s | ✓ |
| `ci::home-manager` | 1m35s | ✓ |
| `ci::nix` | 1m16s | ✓ |
| `ci::smoke` | 1m8s | ✓ |
| `ci::unit` | 1m4s | ✓ |
| `ci::pnpm-hash-fresh` | 53s | ✓ |
| `ci::surface-example-build` | 51s | ✓ |
| `ci::surface-app-example-build` | 51s | ✓ |
| `ci::biome` | 50s | ✓ |
| `ci::fmt` | 49s | ✓ |
| `_ci-setup` | 31s | ✓ |
| `ci::install` | 7s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/139c718/timings.jsonl`; pool state is a live `flock` probe.</sub>

srid:

The **x86_64-linux** lane ran on **`kolu-ci-4`** (`dev-x86-64-linux-08`) — commit `b3ae14d8`, **passed** ✓

- **Lane wall** (pipeline): **9m12s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 8m30s | ✓ |
| `ci::home-manager` | 33s | ✓ |
| `_ci-setup` | 28s | ✓ |
| `ci::unit` | 23s | ✓ |
| `ci::nix` | 19s | ✓ |
| `ci::biome` | 17s | ✓ |
| `ci::smoke` | 15s | ✓ |
| `ci::install` | 14s | ✓ |
| `ci::fmt` | 13s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

srid:
## Field report from `zest` (macOS/launchd): a `PTY_HOST_CONTRACT_VERSION` skew crash-loops, and launchd masks it as an "App updated" loop

While diagnosing an unrelated "Reload button loops forever" report on `zest` (macOS 26.5.1, kolu under launchd `org.nix-community.home.kolu`), I found a second, distinct cause in the server logs that's directly in this PR's blast radius — flagging it here so it can be designed against before the surviving daemon ships.

### Evidence (off the running box)

`~/Library/Logs/kolu.err.log` is littered with import-time crashes:

```
SyntaxError: The requested module '@kolu/pty-host' does not provide an export named 'PTY_HOST_CONTRACT_VERSION'
```

Counting starts vs. successful boots:
- **180** distinct `node:<pid>` process starts in the err log
- **132** distinct `serverId`s (i.e. processes that reached `kolu listening`)
- → ~**48** starts that crashed *before* listening

The launchd plist has:
```xml
<key>KeepAlive</key>
<dict><key>Crashed</key><true/><key>SuccessfulExit</key><false/></dict>
```
So a hard-failing build doesn't fail loudly once — launchd **relaunches it forever**, several times a second, silently. (This is more aggressive/quiet than systemd or a foreground process, which is why it surfaced on the Mac.)

### Why it matters for this PR

The crash is a server↔pty-host **contract-version skew that hard-fails at module import**. With a *surviving daemon* (this PR), the skew surface gets larger, not smaller: a freshly-deployed server can now find an **old daemon still holding the socket** from a previous build. Today's signature is a static ESM import mismatch within a build; once the daemon survives across deploys, the same `PTY_HOST_CONTRACT_VERSION` mismatch can also occur **across the socket** (new server client ⇄ old surviving daemon).

The user-visible symptom is nasty and indirect: each crash→relaunch presents to the browser as a *new server identity* → `status() === "restarted"` → `updateReady()` true → the **"App updated / Reload"** card (`surface-app/src/solid/index.ts:577`). The user sees an endless "App updated" loop and reasonably blames the client, when the real cause is the server crash-looping under launchd.

### Suggestions to consider

1. **Make the contract-version check a graceful, honest degrade — never an import-time `throw`.** A version mismatch (build-local or across the socket to a surviving daemon) should produce a clear "daemon/server version skew — restart the daemon" state, not a process that dies before `listening` and lets launchd hammer it.
2. **On boot, when the server adopts a surviving daemon, validate `PTY_HOST_CONTRACT_VERSION` over the socket first.** On mismatch, prefer a controlled daemon restart (this PR already has `kill → waitForPidGone → respawn → reconnect`) over crashing the server.
3. **Don't let a crash-restart masquerade as "App updated."** A rapidly-relaunching server is an *unhealthy* state, semantically distinct from "a new build is ready." Worth distinguishing in the overlay so launchd's KeepAlive can't paint a crash-loop as a benign update prompt. (Tracking the client-side fix for the genuinely-stale-bundle variant separately via `/be`.)

Status on `zest` right now: the **current** build (`dqwk2mzs…-kolu-stamped`) *does* export `PTY_HOST_CONTRACT_VERSION` and the process is stable — so this was a bad-deploy window, now resolved. Filing it because the surviving daemon makes the skew window wider and the launchd amplification means it fails silently-then-loudly-to-the-user.

srid:

The **x86_64-linux** lane ran on **`kolu-ci-4`** (`dev-x86-64-linux-08`) — commit `6ed2d709`, **passed** ✓

- **Lane wall** (pipeline): **9m12s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 8m30s | ✓ |
| `ci::home-manager` | 33s | ✓ |
| `_ci-setup` | 28s | ✓ |
| `ci::unit` | 23s | ✓ |
| `ci::nix` | 19s | ✓ |
| `ci::biome` | 17s | ✓ |
| `ci::smoke` | 15s | ✓ |
| `ci::install` | 14s | ✓ |
| `ci::fmt` | 13s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|
| `kolu-ci-4` | dev-x86-64-linux-08 | 🔒 leased |

srid:
Re-phasing https://github.com/juspay/kolu/pull/1285

PR #1278, opened 2026-06-11
fix(surface-app): cache-busting reload so 'App updated' can't loop

srid:
**The "App updated → Reload" card could loop forever on a browser that cached the SPA shell in an older build** — clicking Reload re-showed the same prompt, the update never applied, and a manual force-reload fixed it for exactly one load. Found in the wild on a macOS/launchd box (`zest`) where the server was provably clean: `GET /` returned `no-store` referencing the current bundle, `/sw.js` was fetch-less, no proxy — yet the chrome bar read `CLIENT <old> · ≠ srv` on every reload.

### Why `no-store` was necessary but not sufficient

The shell directive stops a browser from *newly* caching `/`. It does nothing for a browser that **already** cached `/` in a pre-`no-store` era: that entry carries a 1970 `Last-Modified` (the nix-store epoch), earning *years* of heuristic freshness. On a **normal** reload the browser serves it without revalidating — the current `no-store` response never reaches it.

```
reloadForUpdate()  →  location.reload()   ─▶  normal reload
                                               ─▶ poisoned `/` entry (heuristically fresh)
                                               ─▶ stale bundle  ─▶ stale() true
                                               ─▶ "App updated" card  ─▶ click Reload  ─▶ ⟲
```

A hard reload shows fresh *once* but never **evicts** the heuristic entry, so the next normal reload is stale again — which is why telling the user to force-reload doesn't fix the loop (and an installed PWA has no force-reload gesture at all).

### The fix — navigate to a key the poison can't match

`reloadForUpdate()` now navigates to `/?__surface_app_fresh=<token>` instead of reloading in place. A query string is a *different cache key* the poisoned bare-`/` entry can't satisfy, so the browser must hit the network → the `no-store` shell → the current bundle. Because that response is `no-store`, it is never written to the cache — so the navigation both **escapes the loop now** and **inoculates the tab** going forward.

- The URL computation is a pure, unit-tested helper (`cacheBustedShellUrl`) in `index.ts` — the package's "pure kernels" home — alongside `SHELL_CACHE_CONTROL`/`clientIsStale`. A second test pins the production entry point (`reloadForUpdate` navigates via `location.replace`, never `location.reload`).
- `location.replace`, not `assign`/`reload`, keeps the bust out of history.
- The token's *value* is irrelevant to correctness (a unique `Date.now()`); its only job is to differ from the poisoned key — so the reload mechanic stays decoupled from build-identity. The param is namespaced (`__surface_app_fresh`, not a bare `?v=`) so it can't collide with a consumer's route state.

> *Verified on the box: `GET /?__surface_app_fresh=676a483` → `no-store` + the current bundle.* The full investigation (and the earlier service-worker chapter of the same bug class) lives in [`docs/cache-bug.md`](https://github.com/juspay/kolu/blob/fix/update-reload-cache-bust/docs/cache-bug.md).

### Scope — forward-looking, by design

The fix only runs once a browser is **already executing a bundle that contains it**, so it stops the loop for every client on this release onward. A browser *currently* stuck on a pre-fix bundle still runs the old `location.reload()` and can't pull this code through a normal reload; that client needs a one-load crossover (a single force-reload) or a durable cure (open `/?__surface_app_fresh=1`, or clear site data / reinstall the PWA). No server-side change can push new code through a shell the browser refuses to revalidate. The changelog and `docs/cache-bug.md` spell this out.

### Heads-up — `@kolu/surface-app` is shared with drishti

`reloadForUpdate` lives in `@kolu/surface-app`, consumed by [`drishti`](https://github.com/srid/drishti) too. Per `.claude/rules/surface.md` this is an API-facing behavior change; a companion drishti PR follows (the behavior is strictly better — drishti gets the same un-loopable reload for free, no wiring change). *Distinct from the launchd crash-loop variant of the same card — tracked on #1275.*


srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 round(s) · codex reviewed at `xhigh` reasoning effort · base `5a15d3ed9314`

### Round 1

**codex** — approved: `false`

The cache-busted navigation is the right direction for clients that are already running this updated bundle, but the branch currently overstates what it fixes, leaves the actual reload entry point untested, and introduces a very generic query parameter into application URLs. I found no untracked files in `git status --short --untracked-files=all`.

Findings:
- `F1` · major · open — The user-facing note says a browser that cached the shell in an older build is now kept fresh after clicking Reload, but the code change only helps once that browser is already running a bundle containing the new `reloadForUpdate`. A client currently trapped on a pre-fix cached shell still executes the old `location.reload()` handler and cannot load this new code through a normal reload; similarly, the poisoned bare `/` cache entry remains available for future bare-`/` launches even after this tab moves to `/?v=...`. This overpromises the fix for exactly the affected installed/PWA users. (website/src/content/changelog/unreleased.mdx:13)
- `F2` · minor · open — The production behavior that matters changed in `reloadForUpdate`, but the tests only cover `cacheBustedShellUrl`. A regression that changed `reloadForUpdate` back to `location.reload()` would still leave the new helper tests green. (packages/surface-app/src/lifecycle.ts:109)
- `F3` · minor · open — `CACHE_BUST_PARAM = "v"` is a public, single-letter query key. Because `cacheBustedShellUrl` overwrites any existing `v` parameter on the current app URL, a consumer that uses `?v=` for real route state will have that state silently changed by the reload action. (packages/surface-app/src/index.ts:77)
- `F4` · nit · open — Several comments still describe `reload()` as a plain `location.reload()` even though this branch changes it to `location.replace` with a cache-busted URL. The top-level lifecycle doc says "land the deployed build with a plain reload", and the consumer comments in `packages/client/src/rpc/TransportOverlay.tsx:12` and `packages/client/src/MobileChromeSheet.tsx:73` repeat the old behavior. (packages/surface-app/src/lifecycle.ts:7)

**claude** — Addressed all four CODEX findings against the diff since 5a15d3ed. All four fixed. 32 surface-app tests pass, tsc noEmit clean, files formatted.

- `F1` **fixed** — Agreed. A browser on a pre-fix bundle still runs the old location.reload and cannot pull the new code via a normal reload. Narrowed unreleased.mdx to say From this release on, and added remediation for stuck clients: one force-reload, or clear site data and reinstall for a PWA, after which the in-app Reload self-heals. Added a pre-fix-client paragraph and revised the Learning section in docs/cache-bug.md so it no longer promises a remote cure.
- `F2` **fixed** — Agreed. Production behavior lives in reloadForUpdate but only cacheBustedShellUrl was tested. Added packages/surface-app/src/lifecycle.test.ts that stubs location and pins Date.now, then asserts reloadForUpdate calls location.replace once with cacheBustedShellUrl of the current href plus the token, and does not call location.reload.
- `F3` **fixed** — Agreed at the package boundary. surface-app is shared with drishti, and the v param already carries meaning in iframePreviewRoute.ts preview URLs. Since the helper overwrites the param, a bare v could clobber a consumer route state. Renamed CACHE_BUST_PARAM from v to a namespaced name and updated its doc. kolu shell URL uses no query params for route state, so no live kolu collision, but defensive for downstream consumers.
- `F4` **fixed** — Agreed. Updated the lifecycle.ts module doc and the two consumer comments in TransportOverlay.tsx and MobileChromeSheet.tsx to describe the cache-busting navigation and reference reloadForUpdate, instead of the old plain location.reload model.

commit: `92e0f843db719070e3eff66ca25576c08751e88e`

### Round 2

**codex** — approved: `false`

The production code change and the new reloadForUpdate test address the core implementation concerns. Two documentation issues remain: the stuck-client remediation still overstates what a force reload can permanently fix, and one lifecycle comment still names the old bare `v` parameter after the rename.

Findings:
- `F1` · major · open — Partially fixed, but the remediation text is still too strong. The doc correctly says a hard reload shows fresh once but does not evict the poisoned bare-`/` entry, then says a pre-fix client needs only "a single hard/force-reload" and that after loading this build once the in-app reload self-heals. That is only true for the currently running tab. A later bare-`/` launch can still be satisfied by the old poisoned entry and run the pre-fix `location.reload()` code again. (docs/cache-bug.md:523)
- `F2` · minor · resolved — Resolved. The new test stubs the current URL and navigation methods, pins `Date.now`, calls the real `reloadForUpdate`, and asserts `location.replace(cacheBustedShellUrl(...))` while guarding against `location.reload()`. That covers the production entry point I flagged. (packages/surface-app/src/lifecycle.test.ts:20)
- `F3` · minor · open — The collision bug itself is fixed by renaming the cache-bust key to `__surface_app_fresh`, but this exported lifecycle doc still says reloadForUpdate navigates to `/?v=<token>`. That stale comment reintroduces the bare-`v` guidance the fix intentionally removed. (packages/surface-app/src/lifecycle.ts:105)
- `F4` · nit · resolved — Resolved. The comments I called out no longer describe `reload()` as a plain `location.reload()`; they now describe the cache-busting navigation and point readers at `reloadForUpdate`. (packages/client/src/rpc/TransportOverlay.tsx:12)

_codex on the rebuttal:_ No disputes were raised. F2 and F4 are resolved. F3 is mostly resolved in code, but the lifecycle doc still carries the old bare `v` URL. F1 is improved, but I am holding it open because the new remediation still treats a force reload as a one-time fix even though the same doc says force reload does not evict the poisoned bare-`/` entry.

**claude** — Addressed both open findings from codex's follow-up verdict. F1: rewrote the pre-fix-client remediation in docs/cache-bug.md so it no longer treats a force-reload as a durable one-time cure — it now explicitly distinguishes a "one-load crossover" (force-reload, which lands the new build in the running tab but does NOT evict the poisoned bare-`/` entry, so future bare-`/` launches can re-run the pre-fix code) from a "durable cure" (a cache-busted URL `/?__surface_app_fresh=<token>` or clearing site data / reinstalling the PWA, which evicts the poison). Resolved the internal contradiction with the doc's own freshness analysis at line ~500. F3: updated the reloadForUpdate doc comment in lifecycle.ts that still named the old bare `/?v=<token>` to reference `/?<CACHE_BUST_PARAM>=<token>`, matching the namespaced `__surface_app_fresh` key the fix uses. F2 and F4 were already resolved in the prior round per codex. Ran just fmt on the touched files.

- `F1` **fixed** — Agreed — real internal contradiction. The doc's freshness analysis (cache-bug.md:~500) already states a hard reload does NOT evict the heuristic/poisoned bare-`/` entry, yet the remediation paragraph then called a 'single hard/force-reload' a one-time fix. Rewrote docs/cache-bug.md:518-538 into two explicit categories: (1) a force/hard-reload is a 'one-load crossover' that lands the new build in the running tab only — a later bare-`/` launch can still be served the poisoned shell and run the pre-fix location.reload() again, and a PWA has no force-reload gesture; (2) a durable cure is a cache-busted URL (`/?__surface_app_fresh=<any-token>`, the same key reloadForUpdate now uses) or clearing site data / reinstalling the PWA, which actually evicts the poison so future bare-`/` launches stay clean. Also tightened the Learning paragraph to match. The user-facing changelog (unreleased.mdx) was not flagged by codex for F1 and already qualifies force-reload as 'only fixed it for a single load', so it was left as-is.
- `F2` **fixed** — Already resolved in the prior round (commit 92e0f843). codex's verdict marks F2 status=resolved: lifecycle.test.ts stubs the URL/navigation, pins Date.now, calls the real reloadForUpdate, asserts location.replace(cacheBustedShellUrl(...)) and guards against location.reload(). No further change needed this round.
- `F3` **fixed** — Agreed — stale comment reintroduced the bare-`v` guidance the fix removed. Updated lifecycle.ts:105 (the reloadForUpdate doc) from `/?v=<token>` to `/?<CACHE_BUST_PARAM>=<token>`, which names the namespaced key (CACHE_BUST_PARAM = '__surface_app_fresh' in index.ts) rather than the single-letter param. The only remaining `/?v=` in docs/cache-bug.md is a historical verification record ('the original bare-`v` prototype: GET /?v=676a483'), which correctly describes the prototype test and is intentionally left.
- `F4` **fixed** — Already resolved in the prior round (commit 92e0f843). codex's verdict marks F4 status=resolved: the TransportOverlay.tsx (and MobileChromeSheet.tsx) comments no longer describe reload() as a plain location.reload() — they describe the cache-busting navigation and point at reloadForUpdate. No further change needed this round.

commit: `3bbc5f95fe119191ef387a5041dfe1fd86cf65e0`

### Round 3

**codex** — approved: `true`

The remaining review items are closed. The docs now separate a one-load force-reload crossover from durable remediation, the reload comment no longer points readers at the old bare `v` key, and the previously resolved test/comment fixes remain in place. I saw no untracked files; `git status --short --untracked-files=all` only reports the tracked `packages/surface-app/src/lifecycle.test.ts` modification already included in the branch diff.

Findings:
- `F1` · major · resolved — Resolved. The remediation text now explicitly distinguishes a force/hard-reload as a one-load crossover from durable fixes, and it names the cache-busted URL or clearing site data/reinstalling as the durable paths. That addresses the prior contradiction with the doc's own statement that hard reload does not evict the poisoned bare-`/` entry. (docs/cache-bug.md:525)
- `F2` · minor · resolved — Resolved. The test still covers the production entry point by stubbing `location`, pinning `Date.now`, calling `reloadForUpdate`, asserting `location.replace(cacheBustedShellUrl(...))`, and guarding against `location.reload()`. (packages/surface-app/src/lifecycle.test.ts:20)
- `F3` · minor · resolved — Resolved. The reloadForUpdate doc no longer documents the old bare `/?v=<token>` URL and now refers to `/?<CACHE_BUST_PARAM>=<token>`, matching the namespaced cache-bust key. (packages/surface-app/src/lifecycle.ts:105)
- `F4` · nit · resolved — Resolved. The consumer comments continue to describe the cache-busting navigation rather than the old plain `location.reload()` behavior. (packages/client/src/rpc/TransportOverlay.tsx:12)

_codex on the rebuttal:_ F1: accepted. The rewritten docs now make the exact distinction I was asking for: force reload can cross over once, while durable recovery requires a cache-busted URL or removing the poisoned entry. I accept leaving the changelog as-is in this round because the remaining open F1 from my prior pass was specifically the docs contradiction, and the changelog already frames force reload as a single-load workaround. F2 and F4 were already resolved and remain resolved. F3: accepted; the stale `/?v=` guidance in the lifecycle comment is gone.

srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `5a15d3ed9314`

Independent findings: lowy=3, hickey=1

### Applied (1)
- `hickey-1` Reload mechanism re-explained inline at four sites — commit `7e11b5d69`

### Agreed — no change (3)
- `lowy-1` Electricity test: reload() receptacle cleanly encapsulates the remediation-mechanism volatility (packages/surface-app/src/lifecycle.ts:110)
- `lowy-2` In-app location.reload() call sites are not behind the same freshness receptacle (packages/client/src/App.tsx:338)
- `lowy-3` cacheBustedShellUrl is a reusable pure kernel, not implementation-shaped (packages/surface-app/src/index.ts:88)

srid:

Ran the full pass — **rule checklist · fact-check · elegance (/simplify)** — scoped to the merge-base diff. **No fixes needed; the code was already clean** after the codex (3 rounds) and lens (1 round) debates.

| Lens | Finding | Action |
| --- | --- | --- |
| Reuse | none | — |
| Simplification | none | — |
| Efficiency | `String(Date.now())` flagged as redundant ceremony | **Skipped (false positive)** — required by the `token: string` signature; the coercion is correct and intentional |
| Altitude | `reloadForUpdate` couples token-mint with navigation; token strategy not exported | **Skipped** — over-engineering for one call site; `cacheBustedShellUrl` is already exported and composable, and lens already ruled the receptacle correctly encapsulated |

Simplify's four parallel agents (reuse / simplification / efficiency / altitude) independently returned clean.

srid:
## Evidence

The infinite-loop *symptom* can't be reproduced in the e2e harness — it needs a browser holding a heuristically-fresh, pre-`no-store` cached `/` entry, which Cucumber/Playwright can't manufacture. So this captures the **fixed behavior at its source**: the real "App updated" card, driven through the genuine `pwa.reload()` → `reloadForUpdate()` path, on a live kolu.

To make the card appear without a poisoned cache, the dev instance was served with the **client and server pinned to two different clean commits** (`CLIENT c11e7ab · SRV 5e54ab1` → `≠ srv` → `updateReady()` true). Clicking the card's **Reload** button:

before/after: Reload navigates to a cache-busted URL

- **BEFORE** — `location = http://localhost:33391/`. A plain `location.reload()` (the old code) would re-request this exact bare-`/` key — the one a poisoned cache answers from disk → the loop.
- **AFTER** — `location = http://localhost:33391/?__surface_app_fresh=1781200859157`. The shipped `reloadForUpdate()` navigates (via `location.replace`) to a **cache-busted key the poisoned bare-`/` entry can't satisfy**, and the browser fetches it **fresh from the network** (`PerformanceNavigationTiming.transferSize = 1386 B`, i.e. *not* `(disk cache)`).

> The card still shows in the *after* shot only because this rig pins the two commits permanently different to force the prompt; in a real deploy the commits match after the cache-busted load, so the prompt clears. The artifact's claim is the **navigation target** — bare `/` → `/?__surface_app_fresh=<token>` — which is exactly what breaks the loop.

Cross-checked at the HTTP layer on the affected box (`zest`): `GET /?__surface_app_fresh=676a483` → `cache-control: no-store` + the current bundle. The pure URL kernel and the production entry point are also pinned by unit tests (`cacheBustedShellUrl` in `index.test.ts`; `reloadForUpdate` navigates via `location.replace`, never `location.reload`, in `lifecycle.test.ts`).

srid:

The **x86_64-linux** lane ran on **`kolu-ci-7`** (`dev-x86-64-linux-05`) — commit `5059d85d`, _verdict unknown (no per-node timing)_

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

srid:
This remains unfixed on macOS:

image

as well as on Linux:

image

srid:
Root cause of the failure reported above turned out to be the build system, not the browser: `koluStamped` re-stamps the commit into an `immutable`-cached bundle without changing its filename, so the cache-busted shell still loads the old-stamped asset from cache. Full analysis, reproduction, and fix plan: #1319.

PR #1280, opened 2026-06-11
Render Markdown front-matter as a metadata table in previews

### LINKED ISSUE #1279: Markdown preview should display YAML frontmatter
This:

Image

renders as:

Image

Instead, it should display the YAML frontmatter table at the top just like GitHub renders.

srid:
**A Markdown file that opens with a `---` YAML front-matter block now renders that metadata as a tidy key/value table at the top of the preview**, the way GitHub shows it. Previously the block was stripped from the document preview, so the metadata silently vanished — and when the strip missed, the `---` fences misrendered as a horizontal rule and a stray Setext heading (#1279).

### How it works

The block parse splits the leading `---` … `---` off the body, so the body always renders as plain markdown and the metadata renders separately. The YAML is parsed with the already-pinned `yaml` library (no new dependency surface — it's a global override in the workspace) into a top-level mapping, then emitted as a `<table data-md-frontmatter>` of escaped key/value rows that flows through the same DOMPurify sanitizer as the rest of the document.

- **Document-only** — a new `frontMatter` render option gates the table to the full document preview. The compact chat/dock intent slot still drops front-matter: a message that happens to open with `---` is not a document.
- **Value shapes** — scalars print as text, a list of scalars joins with commas (the common `tags:` case), and anything deeper falls back to compact JSON so the cell stays legible without nested tables.
- **Fails visible, not silent** — an empty `---\n---` block renders nothing, but front-matter that *can't* be tabulated (malformed YAML, or a valid scalar/list that isn't a key/value mapping) is shown **raw as a YAML code block** rather than dropped — so the user can see and fix it, and it never blows up the preview or misrenders as an hr + heading.

### Styling

The table gets its own muted, borderless-ish treatment (a single hairline frame, a faint key column, dimmed text) so it reads as *metadata sitting above the prose* rather than a heavy GFM data table competing with it.

Closes #1279


srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `5a15d3ed9314`

### Round 1

**codex** — approved: `false`

The change is mostly well-scoped, and the sanitizer/CSS/docs wiring follows the intended document-vs-compact split. I found two renderer robustness issues in the new front-matter path; both are in `render.ts` and should be addressed before approval.

Findings:
- `F1` · major · open — `formatFrontMatterValue` calls `JSON.stringify` for object values and object array items outside the `parseYaml` try/catch. YAML aliases can produce cyclic structures from valid input, e.g. `---\na: &a [*a]\n---\n# Body`, so rendering throws `TypeError: Converting circular structure to JSON` and can crash the markdown preview instead of degrading safely. (packages/solid-markdown/src/render.ts:218)
- `F2` · minor · open — The front-matter regex misses the shortest empty block form `---\n---\n` because it requires a newline before the closing fence after the opening line. That block then falls through as normal Markdown, even though the new code comments and behavior intend empty front matter to disappear cleanly. (packages/solid-markdown/src/render.ts:190)

**claude** — Both CODEX findings are valid and now fixed in render.ts (+ regression tests in render.test.ts). I empirically reproduced each before fixing: (F1) parseYaml("a: &a [*a]") yields a cyclic structure that makes JSON.stringify throw TypeError, crashing the render; (F2) the front-matter regex returned null for "---\n---\n# Body", letting an empty block fall through as markdown. All 38 tests pass and tsc is clean.

- `F1` **fixed** — Confirmed: parseYaml("a: &a [*a]") returns a self-referential array (valid YAML), and JSON.stringify on it throws TypeError: Converting circular structure to JSON. Since formatFrontMatterValue's JSON.stringify ran outside the parseYaml try/catch, this crashed renderMarkdownToRawHtml. Fixed by routing both object-value and object-array-item serialization through a new safeJsonStringify(value: object) helper (render.ts:209-220) that catches the failure and returns the placeholder "[unserializable]", so the row and document body still render. Added a regression test "survives a cyclic YAML alias instead of crashing the preview" in render.test.ts asserting the body still renders and the frontmatter table is present.
- `F2` **fixed** — Confirmed: the old regex /^---[ \t]*\r?\n([\s\S]*?)\r?\n---.../ required a newline before the closing fence after the captured body, so "---\n---\n# Body" did NOT match (exec returned null) and fell through as markdown. Fixed by making the body-plus-newline group optional: /^---[ \t]*\r?\n(?:([\s\S]*?)\r?\n)?---[ \t]*(?:\r?\n|$)/ (render.ts:192-193). The empty block now matches with captured body undefined; splitFrontMatter already coalesces via m[1] ?? "", and renderFrontMatterTable returns "" for empty/non-mapping YAML, so the block vanishes cleanly (no table, no spurious hr). Added regression test "makes an empty front-matter block disappear cleanly" in render.test.ts.

commit: `e7923e85cbdb1be04a98a0e9ac6f35e20b160ac2`

### Round 2

**codex** — approved: `true`

CLAUDE's current changes adequately address both prior findings. The renderer now uses a safe serialization helper for structured YAML values and the front-matter splitter now recognizes the empty `---\n---\n` form; both fixes have targeted regression tests. I did not identify any new regression introduced by this round's changes.

Findings:
- `F1` · major · resolved — Original issue: cyclic YAML aliases could make `JSON.stringify` throw during front-matter value formatting, crashing markdown rendering after YAML parsing succeeded. The current code routes object serialization through `safeJsonStringify`, catches serialization failures, and falls back to `[unserializable]`, so the document body continues rendering. (packages/solid-markdown/src/render.ts:214)
- `F2` · minor · resolved — Original issue: the front-matter regex missed the shortest empty block form `---\n---\n`, allowing it to fall through as normal Markdown. The current regex makes the body-plus-newline group optional, so the empty block is split off and then disappears cleanly as empty metadata. (packages/solid-markdown/src/render.ts:192)

_codex on the rebuttal:_ F1: Accepted. I verified the new helper catches cyclic/object serialization failures and the formatter now uses it for both object values and object array items. F2: Accepted. I verified the updated regex matches `---\n---\n# Body`, captures an empty YAML body, and leaves `# Body` as the markdown body.

srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `5a15d3ed9314`

### Applied (2)
- `lowy-1` frontMatter option's table output silently depends on the sanitize layer's richHtml flag — a wire exposed across the receptacle — commit `065db8f36`
- `hickey-1` Per-item value dispatch duplicates the top-level value dispatch — commit `69766307e`

### Agreed — no change (3)
- `lowy-2` frontMatter value-formatting (scalar/list/object→JSON) is an independent activity axis folded inline, but correctly so (packages/solid-markdown/src/render.ts:227-242 (formatFrontMatterValue) and 244-272 (renderFrontMatterTable))
- `lowy-3` safeJsonStringify is a generic utility living as a private helper inside the markdown render module (packages/solid-markdown/src/render.ts:209-220)
- `hickey-2` Every render flag is a fresh projection of `variant()` — the variant discriminator is fragmented into parallel booleans (packages/solid-markdown/src/Markdown.tsx:189-211)

srid:

Reviewed the front-matter renderer (`packages/solid-markdown`) across all three passes.

| Pass | Result |
| --- | --- |
| **Rules** | ✅ Clean — no built-in or Kolu project-rule violations |
| **Fact-check** | 1 latent fragility → fixed by comment |
| **Elegance** | 1 fix applied, 1 reverted (regression) |

**Applied**
- Inlined the single-call `safeJsonStringify` helper into `formatFrontMatterValue`'s `scalar` closure — one fewer function and hop, cyclic-alias rationale moved to the JSDoc (`9802d406`).
- Documented the `m[1] ?? ""` invariant in `splitFrontMatter`: `null` = no front-matter, `""` = a present-but-empty `---\n---` block — both render no table (`9802d406`).

**Reverted (caught by a test)**
- The elegance pass also proposed gating `splitFrontMatter` behind the `frontMatter` option to skip the regex on the compact slot. **Reverted** — the body must be stripped on *every* block parse, or the compact intent slot regresses to rendering the `---` block as an `<hr>` + heading. The compact-slot unit test failed on the change and passes after the revert (53/53 green).

No reuse or altitude findings — `escapeHtml`/`yaml` reuse is appropriate and the `data-md-frontmatter` marker is consistent with the existing `data-md-alert` / `data-md-wikilink` pattern.

srid:
## Evidence

Markdown front-matter rendered as metadata table

The Code-tab preview now renders YAML front-matter (`---` blocks) as a `<table data-md-frontmatter>` at the top of the document. The `---` delimiters no longer misrender as an `<hr>` + heading; instead the title, author, and tags metadata appear in a structured table above the document body.

srid:

The **x86_64-linux** lane ran on **`kolu-ci-1`** (`dev-x86-64-linux-04`) — commit `3b8c59ea`, _verdict unknown (no per-node timing)_

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

PR #1284, opened 2026-06-11
fix: detect Claude Code's AskUserQuestion footer after the 'n to add notes' change

srid:
**A Claude Code question waiting on your answer no longer leaves the kolu dock looking idle.** kolu promotes a terminal to *awaiting-user* by screen-scraping Claude Code's `AskUserQuestion` footer — but **v2.1.173 inserted an `· n to add notes` hint into that footer**, landing right between the two phrases kolu keyed on, so the marker silently stopped matching and the question went unflagged.

### The footer that broke it

```
Enter to select · ↑/↓ to navigate · n to add notes · Esc to cancel
                              └──────── new in v2.1.173 ────────┘
```

The old marker required `to navigate` and `Esc to cancel` to be *adjacent* (`/to navigate\s*·?\s*Esc to cancel/`). The new segment split them apart.

### The fix

```js
/to navigate\s*·(?:[\s\S]*?·)?\s*Esc to cancel/
```

The marker now keys on the footer's **`·` separators**, not the nav-hint glyphs:

- **A `·` immediately after the nav hint** is the discriminator — only the framework footer paints `to navigate · `, so ordinary prose can't match (`…to navigate the file tree` has no `·`; `…to navigate the tree · Esc to cancel` puts the `·` after the *object*, not the hint).
- **Zero or more `· <segment>` hints** are tolerated before `· Esc to cancel` — v2.1.173's `· n to add notes` today, and whatever lands there next.
- **`[\s\S]` (not `.`)** spans the intervening text, so a footer that xterm **soft-wraps across grid rows** (which `getScreenText` joins with `\n`) still matches.

### Coverage

| Footer | Promotes? |
| --- | --- |
| `↑/↓ to navigate · n to add notes · Esc to cancel` (v2.1.173) | **yes** ✓ |
| same, soft-wrapped across rows | **yes** ✓ |
| `↑/↓ to navigate · Esc to cancel` (old single-select) | yes ✓ |
| `Tab/Arrow keys to navigate · Esc to cancel` (multi-select) | yes ✓ |
| prose: `…to navigate the tree · Esc to cancel` | no ✓ |
| `/model` · trust · `/fork` look-alikes | no ✓ |

Five fixtures pin this in `screen.test.ts` (23 tests). The v2.1.173 footer fixture was confirmed **red before the fix, green after**; the soft-wrap and prose-false-positive cases were added during the codex review debate (see PR comments).

> _This break is inherent to screen-scraping a tool we don't control — the detector's docstring already mandates re-confirming the marker from a live capture on any Claude UI change. This PR is that re-confirmation for v2.1.173._


srid:

✅ **Consensus** after 2 round(s) · lowy + hickey · base `a7d8846b623a`

Independent findings: lowy=1, hickey=3

### Applied (2)
- `hickey-1` Marker-shape fact fragmented across 3 comment sites; header left stale — commit `3fd2e318d`
- `hickey-2` Test fixtures header claims a single capture version while a fixture is from another — commit `cfaa6ae97`

### Agreed — no change (2)
- `lowy-1` Marker widened from adjacency to single-line co-occurrence — selectivity vs. forward-tolerance trade is sound but worth noting (packages/integrations/claude-code/src/screen.ts:96)
- `hickey-3` Regex `.*` between the two halves widens the match beyond 'a footer segment' (packages/integrations/claude-code/src/screen.ts:96)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `a7d8846b623a`

### Round 1

**codex** — approved: `false`

The change fixes the captured one-line v2.1.173 footer, but the new marker is both too loose and still brittle for rendered terminal text. The detector now accepts ordinary prose with arbitrary text between `to navigate` and `· Esc to cancel`, while also failing when xterm soft-wraps the new footer before the separator it relies on. I did not run the test suite because this pass was requested as read-only/no file writes; `git diff --check` was clean.

Findings:
- `F1` · major · open — The new AskUserQuestion regex only allows arbitrary middle text on the same rendered row. `getScreenText` is built from xterm buffer rows joined with `\n`, so a narrow tile can soft-wrap the v2.1.173 footer before the next `·` separator, e.g. after `to navigate` or inside `n to add notes`; then `.*` cannot cross the newline and the prompt is not detected. That leaves the dock stuck in the old non-awaiting state in exactly the new footer shape this PR is meant to handle. (packages/integrations/claude-code/src/screen.ts:97)
- `F2` · major · open — The marker now accepts any same-line text after `to navigate` until a later `· Esc to cancel`. That is broader than the Claude footer chrome: ordinary model/tool output like `Use arrow keys to navigate the tree · Esc to cancel` now returns true and can promote a normal `waiting`/`thinking`/`tool_use` session to `awaiting_user` falsely. (packages/integrations/claude-code/src/screen.ts:97)

**claude** — Both CODEX findings target the same regex in packages/integrations/claude-code/src/screen.ts:97. I agree with both and fixed them together by tightening the marker from `/to navigate\b.*·\s*Esc to cancel/` to `/to navigate\s*·(?:[\s\S]*?·)?\s*Esc to cancel/`. The new pattern (a) requires the `·` separator immediately after the nav hint, so footer-only chrome matches while ordinary prose does not, and (b) uses `[\s\S]` for the intervening segment so a soft-wrapped footer that xterm breaks across rows (joined with `\n` by getScreenText) still matches. Added two regression fixtures CODEX asked for: a soft-wrapped v2.1.173 footer (positive) and a prose case `Use arrow keys to navigate the tree · Esc to cancel` where the `·` trails the object not the hint (negative). Updated both doc comments to describe the separator-required, wrap-tolerant structure. All 23 tests pass (21 prior + 2 new); just fmt clean; git diff --check clean.

- `F1` **fixed** — Agreed and fixed. The old `.*` could not cross a newline, so a soft-wrapped footer (getScreenText joins xterm rows with `\n` — ptyHost.ts:87) would fail to match. New regex `/to navigate\s*·(?:[\s\S]*?·)?\s*Esc to cancel/` uses `[\s\S]` for the intervening segment, spanning rows. Added fixture ASK_USER_QUESTION_WRAPPED in screen.test.ts (footer split before the `· Esc to cancel` separator) with a passing assertion. Verified against three wrap variants (break after nav hint, mid-notes, before final separator) — all match.
- `F2` **fixed** — Agreed and fixed. The old `to navigate\b.*` accepted any same-line text after `to navigate`, so prose like `Use arrow keys to navigate the tree · Esc to cancel` would false-promote. The new regex requires the `·` separator IMMEDIATELY after the nav hint (`to navigate\s*·`), which only the framework footer paints — prose puts the object (`the tree`) there, not a `·`. Verified: CODEX's exact hypothetical returns false, while all real footers (single/multi/notes) still match. Added negative fixture PROSE_NAVIGATE_WITH_SEPARATOR plus assertion. Doc comment in screen.ts now states the `·`-after-hint discriminator explicitly.

commit: `34cbb74efcd97319b9531ec9d0318f408f0a2f29`

### Round 2

**codex** — approved: `true`

CLAUDE accepted both prior findings and the current tree adequately addresses them. The AskUserQuestion marker now requires the footer separator immediately after `to navigate`, preventing the prose false-positive I raised, and its intervening segment uses a newline-tolerant form so soft-wrapped v2.1.173 footers continue to match. The requested positive and negative regression fixtures are present, and `git diff --check` is clean. I did not run the full tests in this read-only review pass.

Findings:
- `F1` · major · resolved — Previously, the widened AskUserQuestion marker could not cross xterm row breaks, so a soft-wrapped `n to add notes` footer could still fail detection. (packages/integrations/claude-code/src/screen.ts:102)
- `F2` · major · resolved — Previously, the widened marker accepted ordinary prose containing `to navigate ... · Esc to cancel`, causing false `awaiting_user` promotion. (packages/integrations/claude-code/src/screen.ts:102)

_codex on the rebuttal:_ No disputes to answer. CLAUDE agreed with both findings; I verified the current regex matches the one-line notes footer and soft-wrapped variants, while rejecting the prose false-positive from F2.

srid:

Ran the full gauntlet's quality track over the change — **reuse, efficiency, altitude, simplification** — plus fact-check. **No fixes required; the code is clean.**

| Lens | Findings | Disposition |
| --- | --- | --- |
| Reuse | 0 | Clean — self-contained, no helper duplicated |
| Efficiency | 0 | Clean — marker evaluated once per poll over a 40-line tail; the lazy `[\s\S]*?` has no catastrophic-backtracking concern on a bounded region |
| Altitude | 0 | Right depth — the marker keys on footer *structure* (the `·` separator immediately after the nav hint), not v2.1.173's specific text, so it's general rather than a one-version bandaid. A structural `·`-split parser was judged premature: it would move the version-coupling upstack (segment *semantics*) rather than remove it |
| Simplification | 2 | Both skipped as false positives — (1) merging the two positive fixtures into `it.each` would lose distinct failure diagnostics for two *orthogonal* regressions (segment insertion vs. soft-wrap); (2) the fixture JSDoc and the test name aren't redundant (motivation vs. expectation, different readers) |

The detector's own docstring already mandates re-confirming the marker from a live capture on any Claude UI change — this PR is that re-confirmation for v2.1.173, with the regex hardened (via the codex debate) against soft-wrap and prose false-positives and pinned by three new fixtures.

srid:
## Evidence

The clip shows the terminal tile's agent-state pip promoting **thinking → awaiting_user** the instant the v2.1.173 `· n to add notes` footer is painted into the PTY — end-to-end proof, through the server-side screen-scrape poll, that the dock correctly flags the new footer format.

**Scenario:** `A v2.1.173 AskUserQuestion footer (with the notes hint) still promotes to awaiting`
**Branch:** `fix/askuserquestion-footer-marker`



Still — tile at the `awaiting_user` moment:



▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=askuserquestion-notes-footer.mp4

PR #1286, opened 2026-06-12
fix(surface): guard stdio write streams against unhandled 'error' crashes

srid:
## What

The stdio link writes framed lines to a Node `Writable` on both ends of the family — the client link (`links/stdio.ts`) and the server peer (`peer-server.ts`) — but neither attaches an `'error'` listener to that **write** stream. A failed `write()` reports the error to the per-write callback (which rejects the in-flight frame), but Node *also* emits `'error'` on the stream itself, and **an `'error'` event with no listener is a hard process crash**, not a catchable rejection.

The read half is already guarded — `read.on("error", reject)` in `stdio-codec.ts`, mirrored by `serveOverStdio`'s settled-`{ reason: "error" }` contract. This PR closes the same hole on the write half.

## Why it bites

The pipe torn down mid-write is the *routine* teardown case, not an exotic one: an ssh transport drops, or a peer exits and our `write` is its now-closed stdin, and the next frame raises `EPIPE`. A downstream consumer (`odu`) hit exactly this — its coordinator destroyed a lane mid-write on `[q] quit`, and the unhandled `'error'` on the dying pipe crashed the whole process with a 30-line `write EPIPE` dump *after the run had already finished*:

\`\`\`
Error: write EPIPE
    at afterWriteDispatched (node:internal/stream_base_commons:159:15)
    ...
    at <anonymous> (.../@kolu/surface/src/links/stdio.ts:78:20)
Emitted 'error' event on Socket instance at:
    ...
\`\`\`

## The fix

Attach an `'error'` listener to the write stream on both sides, routing a write death into the **same teardown a read death already takes**:

| side | file | behaviour on write `'error'` |
|------|------|------------------------------|
| client | `links/stdio.ts` | `handleTransportClosed()` — mark the link closed so later `call()`s reject fast instead of limping on a dead pipe |
| server | `peer-server.ts` | funnel into the read stream's teardown so `serveOverStdio` settles `{ reason: "error", error }`, exactly as a read error does |

One teardown path, both directions — symmetric with the existing read-half guard.

## Tests

- `links/stdio.test.ts` — a live link whose write half is `destroy()`-ed mid-session closes instead of crashing; a fresh RPC rejects fast.
- `peer-server.test.ts` — a write-stream `'error'` resolves `serveOverStdio` with `{ reason: "error", error }`, never a rejection (symmetric to the existing read-error test).

Full `@kolu/surface` suite: **98 passing**, `tsc --noEmit` clean, `biome lint` clean.

### DIFF
--- a/packages/surface/src/links/stdio-codec.ts
+++ b/packages/surface/src/links/stdio-codec.ts
@@ -17,7 +17,7 @@
  * to the original `Uint8Array` the peer expects.
  */
 
-import type { Readable } from "node:stream";
+import type { Readable, Writable } from "node:stream";
 import { ORPCError } from "@orpc/client";
 
 /** Encode a single peer message into a single base64 line (no trailing
@@ -41,6 +41,32 @@ export function encodeFrame(
 /** Decode one base64 line back into a `Uint8Array` for the peer codec. */
 export function decodeFrame(line: string): Uint8Array {
   return new Uint8Array(Buffer.from(line, "base64"));
+}
+
+/** Frame one peer message and write it to `write` as a single base64 line.
+ *  The write half's counterpart to `encodeFrame` + the newline delimiter:
+ *  it keeps the wire framing (`encodeFrame(message)` + `"\n"`) in the codec's
+ *  one home rather than open-coded at each call site — the client link and
+ *  the server peer both send through here so the delimiter/encoding can never
+ *  drift between them.
+ *
+ *  Deliberately framing-only: it resolves/rejects on the *write callback*
+ *  (this one frame flushed, or it didn't), and attaches NO stream `'error'`
+ *  listener. A dead write stream is a transport-lifecycle concern whose
+ *  response differs per consumer (the client closes its link; the server
+ *  ends its serve loop), so each side owns that guard locally — see the
+ *  `write.on("error", …)` handlers in `./stdio.ts` and `../peer-server.ts`.
+ *  The read half splits the same way: `decodeFrame` is framing, the
+ *  `read.on("error", …)` in `readFramedLines` is lifecycle. */
+export function writeFramedMessage(
+  write: Writable,
+  message: string | ArrayBufferLike | Uint8Array,
+): Promise<void> {
+  return new Promise<void>((resolve, reject) => {
+    write.write(`${encodeFrame(message)}\n`, (err) =>
+      err == null ? resolve() : reject(err),
+    );
+  });
 }
 
 /** Read line-delimited frames off `read` until the stream ends. Each
--- a/packages/surface/src/links/stdio.test.ts
+++ b/packages/surface/src/links/stdio.test.ts
@@ -11,6 +11,7 @@
  * than hanging.
  */
 
+import { PassThrough } from "node:stream";
 import { eventIterator, oc } from "@orpc/contract";
 import { implement } from "@orpc/server";
 import { describe, expect, it } from "vitest";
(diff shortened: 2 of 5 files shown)


srid:
## 🥊 Review gauntlet (`be-review`)

Ran the serial gauntlet — **lens-debate (lowy ⇄ hickey) → codex-debate → simplify → code-police** — each its own editor, committing in turn. Every track that found something has its fix in this branch.

| # | Track | Finding | Disposition |
|---|-------|---------|-------------|
| 1 | lens (lowy) | Write-framing open-coded twice (`writeLine` + inline client write) while the codec claims to be the single home for the framing axis | **Fixed** — `944be25` |
| 2 | lens (hickey) | Lowy's first form (`writeFramer` bundling the `'error'` listener) braids the framing axis with the lifecycle axis | **Fixed** — extracted *framing-only* `writeFramedMessage`; guards stay inline |
| 3 | codex | Client EPIPE test used a loopback pair, where `serveOverStdio`'s read-side listener absorbs the `destroy` — the test passed even with the guard removed | **Fixed** — `be2e1a7` (isolated stream; mutation-verified it now fails without the guard) |
| 4 | simplify | reuse / simplification / efficiency / altitude all clean | No-op |
| 5 | code-police | Client write-`'error'` handler discarded `err` and logged nothing — asymmetric with the read side's frame-parse logging (`no-silent-error-swallowing`) | **Fixed** — `7f86dbb` |

### lens-debate — consensus
The two lenses converged on a sharper decomposition than either's opening: the **framing** (`encodeFrame` + newline + write) belongs in the codec's one home — hoisted to `writeFramedMessage`, the write-side mirror of `readFramedLines` — while the **lifecycle** guard (`write.on("error", …)`) stays at each call site, because the teardown *response* genuinely differs per consumer (client closes its link; server funnels into the read-driven settle path). That mirrors how the read half already splits `decodeFrame` (framing) from `read.on("error", …)` (lifecycle).

### codex-debate — consensus
No production-code correctness findings. It caught that the client test proved nothing (the loopback pair's shared stream already had a server-side `'error'` listener). The rewrite uses an isolated `Writable`; I mutation-checked it — with the guard removed the test now surfaces an uncaught `'error'`, with it the run is clean.

### code-police
Rules: 1 fixed (above). Fact-check: clean (error propagation sound, double-destroy guard correct). Elegance: deferred to the simplify pass, which ran clean.

Full `@kolu/surface` suite: **98 passing**, `tsc --noEmit` + `biome lint` clean throughout.

srid:
## ✅ CI green

`odu run` (strict) on `7f86dbb`, aarch64-darwin lane — **13 ok · 0 failed · 0 errored · 0 skipped**.

| recipe | | recipe | |
|---|---|---|---|
| `_ci-setup` | ✔ 21s | `ci::pnpm-hash-fresh` | ✔ 3m8s |
| `ci::install` | ✔ 1m27s | `ci::smoke` | ✔ 3m10s |
| `ci::nix` | ✔ 3m13s | `ci::unit` | ✔ 1m7s |
| `ci::home-manager` | ✔ 3m13s | `ci::biome` | ✔ 7s |
| `ci::atlas-sync` | ✔ 1m41s | `ci::e2e` | ✔ 5m58s |
| `ci::surface-app-example-build` | ✔ 13s | `ci::fmt` | ✔ 6s |
| `ci::surface-example-build` | ✔ 13s | | |

Commit-status rollup on `7f86dbb`: **success**.

PR #1287, opened 2026-06-12
feat: R-4 Phase B (B1) — terminals run in a surviving pty-host daemon, server is its client

srid:
**Local terminals now run in a separate, surviving `pty-host` daemon, and kolu-server is a *client* of it.** This is the R-4 Phase B **one-way door**: at boot the server spawns (or recycles) the daemon, connects a socket-backed client, and injects it into the terminal backend; the daemon serves the very same socket to `kolu-tui`. The PTYs no longer live in the server's process, so the daemon outlives every kolu-server restart.

This PR is the [two-PR plan](https://github.com/juspay/kolu/blob/master/docs/atlas/dist/pty-daemon.html)'s **B1** — the door with *survival off*. User-facing semantics stay byte-identical to today (a deploy still recycles your terminals), but the boot path now exercises the full `kill → wait-for-real-exit → respawn → connect` sequence on **every deploy, with zero sessions at stake** — soaking the exact race the #1034 data-loss restart lost, *before* B2 makes terminals actually survive a deploy.

> Supersedes the halted "daemon binary, inert" PR. That binary now has production callers — the no-dead-code rule the [plan](https://github.com/juspay/kolu/blob/master/docs/atlas/dist/pty-daemon.html) added: every line ships executed on the first deploy, not just exercised by tests.

### The flow

```mermaid
sequenceDiagram
  autonumber
  participant Srv as kolu-server
  participant EP as endpoint.ts
  participant D as pty-host daemon
  participant Tui as kolu-tui

  Note over Srv,D: Boot — always recycle (survival off)
  Srv->>EP: ensureLocalEndpoint(socket)
  EP->>D: readDaemonPid → verify it IS our daemon (system.version pid) → SIGTERM + waitForPidGone
  EP->>D: spawnDaemon (Linux systemd-run --user own cgroup · macOS detached) — both SURVIVE
  EP->>D: connect-retry → system.version (handshake; skew degrades, never throws)
  EP-->>Srv: client + publishStatus(connected)
  Srv->>Srv: attachLocalPtyHost(client) → then listen
  Tui->>D: dials the SAME socket (works even while Srv restarts)
  Note over EP,D: heartbeat → state=dead on mid-session death (rail goes red)
```

### What's inside

| Area | Highlights |
| --- | --- |
| **Supervisor** (`server/src/ptyHost/`) | `localDriver.ts` — `systemd-run --user` own-cgroup spawn (Linux) / `detached`+`unref` (macOS+dev), with `--setenv` forwarding of PATH / identity / the Nix env-whitelist and a `NODE_OPTIONS`-stripping dev-flag filter; `waitForPidGone.ts` — the load-aware ESRCH barrier; `endpoint.ts` — the **one status owner**: boot recycle → connect → handshake → heartbeat, emitting `DaemonStatus` on every transition |
| **The flip** | Explicit composition root in `index.ts` (no module-global client, no import-time connect); server **stops serving** the socket — the daemon does; the in-process `ptyHost.ts` is deleted; `LocalTerminalBackend` takes its client via `attachPtyHost` |
| **Honest state** | A live `daemonStatus` cell (one owner) feeds the rail's `pty` column — state dot · commit · staleKey · **daemon uptime beside `srv`'s**, so `pty up 3h` next to `srv up 2m` is glanceable proof the daemon outlived the deploy; a heartbeat flips the dot red on mid-session death |
| **kolu-tui** | Now dials the daemon's socket — and keeps working **while kolu-server is down**. Also runnable standalone (the socket's dir must be owner-only `0700`, so use a private dir): `d=$(mktemp -d); nix run .#kolu-pty-host -- --pty-host-socket "$d/p.sock"` |
| **Operability** | The daemon's stdout+stderr go to ONE log file (`daemonLogPath`, beside the socket) — the **same path on macOS (detached fd) and Linux (`systemd-run StandardOutput=append`)**, never `/dev/null`. The server logs the path so it's never guessed. (Closes the macOS gap a live deploy surfaced — `stdio: "ignore"` had discarded the daemon's whole voice.) |
| **e2e + dev** | Per-worker `--pty-host-socket` isolation + daemon reaping in `killServer`; `just dev` likewise isolates its daemon socket per instance, so a dev server never collides with a production kolu on the default socket |

### Verification

- **Second deploy — validated in prod** (`zest`, macOS/launchd, the staged-prod gate): the daemon spawned its own process group (detached), **survived the deploy**, and the next server **found the survivor and recycled it** — `recycling a surviving pty-host daemon (B1: survival off)` → respawn, ~250 ms `waitForPidGone`, exactly one daemon left, no orphan. The #1034 kill-then-pray race ran on a real deploy with zero sessions at stake.
- **Real-daemon endpoint integration test** (`endpoint.test.ts`): spawn → connect → **recycle** (a survivor is killed, a fresh daemon respawns) → **survive** (`dispose()` drops the client but the daemon lives). It caught a real bug — the heartbeat called `heartbeat()` without its `{}` arg, which would have false-positived "dead" every beat.
- **Hand smoke**: server boots → spawns daemon → `kolu-tui list` works → **SIGTERM the server** → `kolu-tui list` *still* works against the surviving daemon. ✓
- **Daemon logs land in a file, not `/dev/null`** — found on the zest deploy and fixed: a test asserts the daemon's `pty-host daemon listening` boot line is captured in `daemonLogPath` on the detached path, and a manual `systemd-run` smoke confirms the same file on Linux.
- CI green on **x86_64-linux and aarch64-darwin** (incl. `ci::e2e`, `ci::unit`) on the gauntlet-reviewed HEAD; the daemon-logs + dev-socket follow-ups since are locally green (`just check` + the endpoint suite) and want a CI re-run on the final HEAD before merge.

### Review hardening (the gauntlet)

- **codex** caught two real correctness bugs: the recycle could `SIGTERM` a **reused** PID — now it verifies the gate PID against `system.version().pid` over the socket before signalling; and daemon-spawned shells had **lost the Nix env-whitelist filtering** — now forwarded across the process boundary.
- **lens** deleted `createInProcessPtyHost` (dead after the flip) and deduped the daemon-status / identity defaults onto one owner; **code-police** dropped a dead re-export bridge and made `pidIsAlive` ESRCH-explicit.

> **Naming deferred:** the binary stays `kolu-pty-host` and the rail column stays `pty` for now (a later PR may rename to `kolu-daemon`). Survival *across a deploy* — terminals reappearing via adopt-by-id — is **B2**; B1's boot recycles, which is why semantics stay identical to today.

### Try it locally

```sh
nix run github:juspay/kolu/pty-daemon-b1
```


srid:
⏸ **On hold — re-evaluating the phase split.** B1-as-shipped is inert: the daemon binary, pid-gate, and serving path have no production consumer until B2 flips the topology — which violates the redo's own requirement that every PR ships *functional* increments. The plan is being revised (likely: fold the binary into the door PR so everything merges live); this PR will be repurposed or superseded accordingly. No reviews until then.

srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `16bb02643180`

### Applied (9)
- `lowy-1` createInProcessPtyHost is a receptacle with no production wire plugged in — commit `a2a481a8c`
- `lowy-2` Default DaemonStatus literal triplicated across three owners — commit `a982fd6cd`
- `lowy-3` daemonStatusValue holder duplicates the cell's own default and store seam — (uncommitted)
- `lowy-5` connectTimeoutMs vs waitForPidGone ceiling vs heartbeatMs — three independent timeout axes, two hard-coded defaults at the call boundary — commit `c96afcb7e`
- `hickey-1` DaemonStatus "connecting" default duplicated verbatim in three production sites — commit `9c7af01af`
- `hickey-2` PtyHostIdentitySchema (staleKey + navigableCommit) defined independently in two packages — commit `3bd3b7a3d`
- `hickey-3` Empty-identity fallback literal duplicated a fourth time — commit `3d0ed4a9b`
- `hickey-4` Heartbeat effect complects degraded-recovery, death-detection, and reactive write — commit `b278e26e6`
- `hickey-5` shortId build-id formatter placed on the rail consumer, not with the build identity — commit `7b13aa78d`

### Agreed — no change (4)
- `lowy-4` Endpoint mixes status-ownership (sequence) with platform reach via plain inline recycle — acceptably, but the heartbeat/recovery axis is interleaved (packages/server/src/ptyHost/endpoint.ts:154-178)
- `lowy-6` PtyHostClient receptacle correctly encapsulates transport volatility — boundary is sound (packages/server/src/terminalBackend/local.ts:289-640)
- `hickey-6` Two parallel liveness-dot lookup tables encode one health→color policy (packages/client/src/ui/IdentityRail.tsx:29-43)
- `hickey-7` DaemonStatus rebuilt incrementally via scattered partial setStatus merges (packages/server/src/ptyHost/endpoint.ts:94-98, 144-152, 157-165)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 round(s) · codex reviewed at `xhigh` reasoning effort · base `16bb02643180`

### Round 1

**codex** — approved: `false`

The daemon topology mostly hangs together, but I would not approve this pass yet. The main correctness risks are that the recycle path trusts a bare pid file enough to signal it, and that moving PTY spawning into a separate daemon dropped the existing Nix-devshell env filtering at the process boundary. I also found a daemon-status subscription that can fail invisibly and user-facing docs that advertise a socket path the daemon will reject.

Findings:
- `F1` · blocking · open — The boot recycle treats any live pid in the gate as the surviving daemon and sends it SIGTERM. `readDaemonPid` only checks `kill(pid, 0)`, so a stale gate whose pid has been reused by an unrelated same-user process will be killed on server startup. The daemon also exits as `already-running` on the same stale-live pid, so this can both kill the wrong process and fail to start the real daemon. (packages/server/src/ptyHost/endpoint.ts:115)
- `F2` · major · open — The server configures Nix env filtering in its own process, but the daemon is a fresh process and `cleanEnv()` uses module-local state. `daemonEnv()` copies the server environment minus only `NODE_OPTIONS`, and then `packages/pty-host/src/inProcessPtyHost.ts:175` calls `cleanEnv()` inside the daemon without any prior `configureNixShellEnv()`, so dev/test PTY shells fall back to passthrough mode and inherit the full Nix/test harness environment the old in-process path deliberately filtered. (packages/server/src/ptyHost/localDriver.ts:44)
- `F3` · major · open — `daemonStatus` subscribes without an `onError` handler. If the cell subscription fails, the rail keeps rendering the last value or the default `connecting` state with no user-visible indication, which undermines the whole purpose of making daemon health explicit. This also violates the project rule that subscriptions must surface errors. (packages/client/src/wire.ts:164)
- `F4` · minor · open — The standalone daemon example uses `--pty-host-socket /tmp/p.sock`, but `serveOverUnixSocket` refuses to bind in a non-private parent directory like `/tmp`. The same broken command appears in the changelog, so users following the docs get `dir-not-private` instead of a working daemon. (website/src/pages/tui.astro:92)
- `F5` · minor · open — The daemon's hand-rolled CLI parser silently falls back to the default socket when `--pty-host-socket` is provided without a value, and it ignores unknown/misspelled flags entirely. For a standalone daemon, a typo can bind or collide with the default socket instead of failing loudly. (packages/pty-host/src/daemonMain.ts:180)

**claude** — Reviewed CODEX's verdict against the diff. All five findings were valid; fixed every one in the working tree and added regression tests. Verified with `pnpm -r typecheck` (exit 0), `biome lint`/`format` on touched files (clean), and the affected vitest suites: pty-host (74 passed, incl. the closure-guard + new parseArgv tests) and server/ptyHost/endpoint.test.ts (4 passed, incl. the new stale-gate friendly-fire test; runtime dropped from 120s of timeouts to 3.5s once the env-whitelist plumbing was correct).

Key cross-cutting note: the F2 fix (daemon now calls configureNixShellEnv) interacts with the IN_NIX_SHELL=1 safety net under `just check`/e2e — the daemon would process.exit(1) without a whitelist. I therefore threaded `nixEnvWhitelist` through index.ts -> ensureLocalEndpoint -> spawnDaemon (env var KOLU_NIX_ENV_WHITELIST, also forwarded through the systemd-run transient unit), and updated the daemonMain + endpoint tests to forward `"default"` exactly as the real e2e harness does — mirroring production wiring, not a test hack.

- `F1` **fixed** — Agreed — a bare live gate pid could be an unrelated reused-pid process (the dev/macOS tmpdir() gate isn't wiped like $XDG_RUNTIME_DIR). Added gateBelongsToSocketDaemon() in endpoint.ts: before SIGTERM, connect to the socket and require system.version().pid === gatePid (the gate stores the daemon's process.pid and version() returns it), with a 1s connect bound AND a 1s probe-race so a wedged daemon can't hang boot. The recycle SIGTERM now only fires when that verification passes. For the second half (a stale-but-live gate making the fresh daemon exit 'already-running'): since the server has now PROVEN the socket isn't served by that pid, the untrusted branch rmSync's the stale gate before spawning so acquirePidGate can win it. New test 'does NOT signal a stale gate whose live pid is an unrelated process' plants process.pid as the holder and asserts it survives while a real daemon starts and serves.
- `F2` **fixed** — Agreed and confirmed: PTYs moved into the daemon (index.ts attaches the daemon client; spawn handler runs cleanEnv() in the daemon process), but configureNixShellEnv is module-local to kolu-server, so the daemon ran in passthrough mode and leaked the full Nix/test-harness env. Threaded the whitelist across the boundary: index.ts passes argv.flags.allowNixShellWithEnvWhitelist into ensureLocalEndpoint -> spawnDaemon, which injects KOLU_NIX_ENV_WHITELIST into daemonEnv() (forwarded through systemd-run --setenv too, derived from the computed env, plus IN_NIX_SHELL). daemonMain.runPtyHostDaemon now calls configureNixShellEnv(opts.nixEnvWhitelist) before serving (Object.hasOwn distinguishes empty-string from absent). Exported NIX_ENV_WHITELIST_ENV from @kolu/pty-host. Closure-guard test still passes (kolu-pty was already an allowed external). No new staleKey closure entries.
- `F3` **fixed** — Agreed — `app.cells.daemonStatus.use({})` had no onError, so a subscription drop was invisible behind the last value/connecting. Kept the author's no-toast intent but surfaced it: wire.ts now exports daemonStatusError() = _daemonStatus.sub.error(); IdentityRail's ptyState() returns 'dead' when daemonStatusError() is set (error wins over stale state), painting the pty dot red, and the dot's tooltip shows 'Daemon status feed dropped — <message>'. The failure is now visible rather than swallowed, per the subscriptions-must-surface-errors rule.
- `F4` **fixed** — Agreed and verified against serveOverUnixSocket: isPrivateOwnedDir requires (st.mode & 0o077) === 0, which /tmp (world-writable, sticky) fails -> dir-not-private, so the documented daemon command never binds. Changed both website/src/pages/tui.astro and website/src/content/changelog/unreleased.mdx to `d=$(mktemp -d); nix run ... -- --pty-host-socket "$d/p.sock"` and noted the socket dir must be owner-only 0700.
- `F5` **fixed** — Agreed — the hand-rolled parser silently used the default socket on `--pty-host-socket` with no value and ignored unknown/misspelled flags, so a standalone daemon could collide with the default socket on a typo. Replaced parseSocketFlag with a strict parseArgv (still dependency-free, to keep the hashed closure clean) returning {ok:false,error} for a missing/flag-shaped value or any unknown argument; main() logs and process.exit(2) on a parse error. Exported parseArgv and added unit tests covering accept/missing-value/unknown-flag cases.

commit: `6f0d06e676a0855e8a06518e77fb0f9f1fb0d2e3`

### Round 2

**codex** — approved: `false`

CLAUDE's fixes resolve F1-F4. F5 is improved, but not fully closed: the strict parser still accepts an empty socket path and then falls through to the default socket resolver, preserving the silent-default collision risk for one malformed spelling.

Findings:
- `F1` · major · resolved — Previously, endpoint startup could SIGTERM an unrelated process when a stale pid-gate contained a reused live PID. (packages/server/src/ptyHost/endpoint.ts:125)
- `F2` · major · resolved — Previously, PTY shells spawned by the daemon did not receive the server's Nix env whitelist configuration, so daemon-owned PTYs could leak the full dev/test environment or trip the Nix-shell safety behavior. (packages/pty-host/src/daemonMain.ts:128)
- `F3` · minor · resolved — Previously, a dropped daemon-status subscription could remain invisible behind the last known value. (packages/client/src/ui/IdentityRail.tsx:62)
- `F4` · minor · resolved — Previously, the documented standalone daemon command used `/tmp/p.sock`, but the socket server requires an owner-private directory and would reject `/tmp`. (website/src/pages/tui.astro:90)
- `F5` · major · open — The parser still accepts an empty socket override: `--pty-host-socket=` sets `socketPath` to `""`, and `--pty-host-socket ""` does the same through the space-separated branch. `getRuntimeSocketPath` treats an empty override as absent, so these malformed invocations still silently bind the default socket instead of failing loudly. (packages/pty-host/src/daemonMain.ts:225)

_codex on the rebuttal:_ F1-F4: accepted as fixed after re-reading the current implementation. F5: the parser is stricter for missing values and unknown flags, but I'm holding it open because the equals-form empty value, and a quoted empty value in the space-separated form, still produce the same silent default fallback the finding was about.

**claude** — F5 was the only finding still open after the prior round. Codex was correct: parseArgv accepted an empty socket value in both spellings (`--pty-host-socket=` via the equals branch's bare `slice`, and `--pty-host-socket ""` via the space-separated branch where `""` is neither undefined nor `--`-prefixed). getRuntimeSocketPath (packages/surface/src/unix-socket.ts:64) treats an empty override as absent, so both malformed invocations silently bound the DEFAULT socket — the exact silent-default collision the finding was about. I rejected empty values in both branches of parseArgv (packages/pty-host/src/daemonMain.ts) and added a regression test covering both spellings. The default-socket behavior is now reachable only via an absent flag (socketPath: undefined), never a present-but-empty one. Tests: all 9 in daemonMain.test.ts pass; biome format clean. F1-F4 were resolved in the prior round and accepted by codex this round — nothing further to do on them.

- `F1` **fixed** — Resolved in the prior round (gateBelongsToSocketDaemon verification before SIGTERM in endpoint.ts) and accepted by codex this round. No further change.
- `F2` **fixed** — Resolved in the prior round (nixEnvWhitelist threaded across the process boundary via KOLU_NIX_ENV_WHITELIST; daemon calls configureNixShellEnv before serving) and accepted by codex this round. No further change.
- `F3` **fixed** — Resolved in the prior round (daemonStatusError() surfaced; IdentityRail paints the pty dot red on a dropped feed) and accepted by codex this round. No further change.
- `F4` **fixed** — Resolved in the prior round (docs use mktemp -d + 0700 instead of /tmp/p.sock) and accepted by codex this round. No further change.
- `F5` **fixed** — Codex held this open correctly: the strict parser still accepted an empty value in both spellings and fell through to the default socket. Fixed parseArgv (packages/pty-host/src/daemonMain.ts:218,228) to reject an empty value (`value === ""`) in BOTH the space-separated branch and the equals-form branch, returning {ok:false,error:'--pty-host-socket requires a PATH value'}. Added a regression test in daemonMain.test.ts asserting both parseArgv(['--pty-host-socket=']) and parseArgv(['--pty-host-socket','']) return {ok:false}. main() already process.exit(2)s on a parse error, so these now fail loudly instead of silently binding the default socket. All 9 daemonMain tests pass.

commit: `46e8a7065df188502716c58046211771ca92388f`

### Round 3

**codex** — approved: `true`

All five findings are now resolved. I verified the current F5 parser change rejects both empty socket-value spellings and that the regression test covers both cases; I did not find a new regression introduced by this round's changes.

Findings:
- `F1` · major · resolved — Endpoint startup could previously SIGTERM an unrelated process when a stale pid-gate contained a reused live PID. (packages/server/src/ptyHost/endpoint.ts:125)
- `F2` · major · resolved — Daemon-owned PTY shells previously did not receive the server's Nix env whitelist configuration, risking leaked dev/test environment or Nix-shell safety failures. (packages/pty-host/src/daemonMain.ts:128)
- `F3` · minor · resolved — A dropped daemon-status subscription could previously remain invisible behind the last known value. (packages/client/src/ui/IdentityRail.tsx:62)
- `F4` · minor · resolved — The documented standalone daemon command previously used `/tmp/p.sock`, but the daemon socket server requires an owner-private directory. (website/src/pages/tui.astro:90)
- `F5` · major · resolved — The daemon CLI parser previously still accepted present-but-empty socket overrides, which fell through to the default socket path and preserved the silent collision risk. (packages/pty-host/src/daemonMain.ts:220)

_codex on the rebuttal:_ F1-F4 remain accepted as resolved from the prior round. F5 is now fixed: the space-separated branch rejects `value === ""`, the equals branch rejects an empty sliced value, and the new tests cover both malformed spellings. No disputes remain.

srid:

Ran the rule-checklist, fact-check, and elegance passes over the Phase B door diff (after lens + codex + simplify).

**Fixed (3):**
- **`no-re-export-bridge-modules`** — `@kolu/pty-host` re-exported `PtyHostIdentity` / `PtyHostIdentitySchema` from `kolu-common/surface` (the canonical home after the lens dedup) with **zero external consumers**. Dropped the re-export from `ptyHostSurface.ts` + `index.ts`; `buildId.ts` imports the type from `kolu-common/surface` directly. (`fix(police):` commit.)
- **fact-check — `pidIsAlive` collapsed unexpected probe errors to "dead".** `kill(pid,0)` returning anything other than `EPERM` was treated as a dead process. Now only `ESRCH` means dead; `EPERM` and any unexpected error mean *alive* — the safe side for a single-instance lock (never *steal* a gate on a transient probe failure).
- **elegance** — `IdentityRail`'s `ptyState` memoized (two JSX readers: the `data-daemon-state` attr + the `ptyDot` lookup); the `daemonStatusError()` double-read in the `pty` tooltip collapsed to one.

**Skipped (1, with reasoning):**
- `readPidGate`'s documented throw-free "unreadable ⇒ no holder" fallback is kept: it is best-effort by contract (the boot path can't crash on a read), and the downstream atomic `link`/`rm` arbitration means a read failure can't silently *steal* a gate — a persistent error exhausts the bounded retry and throws rather than double-acquiring. Pass 1 (rules) accepted it under the documented-best-effort allowance.

_Posted by `/be-review` after the push._

srid:
## Evidence

The door's user-visible surface is the ChromeBar identity rail's new **`pty` column** — the surviving daemon's own health, distinct from `srv`. Captured by driving a live dev kolu (the server spawns its daemon, the rail subscribes to the `daemonStatus` cell).

**Connected** — `srv · client · pty`, each with its own commit + live uptime. The `pty` column shows the daemon's commit (`dev`), its closure **staleKey** (`a24778b`), and `up` since the *daemon* booted — a green dot for a healthy, serving daemon:

rail — daemon connected

**Honest dead state** — killing the daemon out from under the server: the heartbeat detects it within ~5s (`pty-host daemon heartbeat failed — marking dead`), the endpoint publishes `state: dead`, and the rail's `pty` dot goes **red** — visibly distinct from "you have no terminals" (the empty-canvas lie B1 closes). The column keeps the last-known identity so the failure reads as *this daemon died*, not *nothing was ever here*:

rail — daemon dead (heartbeat detected)

> Also verified at the CLI (hand smoke): after `SIGTERM`-ing the server, `kolu-tui list` still works against the surviving daemon — the daemon outlives kolu-server, exactly as the door intends. (Survival *across a deploy* — terminals reappearing — is B2; B1's boot recycles, so this PR's semantics stay identical to today.)

srid:

The **x86_64-linux** lane ran on **`kolu-ci-8`** — commit `3c504ec0`, _verdict unknown (no per-node timing)_

### Pool status (8 boxes)

| box | location | state |
|---|---|---|
| `kolu-ci-1` | ? | ✓ idle |
| `kolu-ci-2` | ? | ✓ idle |
| `kolu-ci-3` | ? | ✓ idle |
| `kolu-ci-4` | ? | ✓ idle |
| `kolu-ci-5` | ? | ✓ idle |
| `kolu-ci-6` | ? | ✓ idle |
| `kolu-ci-7` | ? | ✓ idle |
| `kolu-ci-8` | ? | ✓ idle |

srid:
Closing in favour of #1291.

PR #1290, opened 2026-06-12
Rewrite the changelog as a dense, compact list

srid:
The changelog spent its width on whitespace. A sparse **10rem version rail** ran empty down the height of every release; inside each entry a fixed **10rem tag column** flung the PR chip to the far right while the squeezed title wrapped onto two lines; and the rows sat far apart. The page read as a field of gaps rather than a list of changes.

This rewrites the layout from scratch around density.

**The release shell.** The version is now a compact header *above* its entries (date inline beside it) instead of a left rail, so no horizontal gutter is wasted and the body aligns to the intro column (`max-w-3xl`).

**The entry row** (`Change.astro`, rewritten). Each entry is a single scannable line: `→ title` on the left, and the area tags + PR chip grouped flush to the right margin as one cluster — so the chip column lines up down the page instead of floating. Rows pack with **zero inter-row gap**; each owns a tight vertical padding and a full-bleed hover band that lights up the whole line (arrow included) and unfolds the description indented beneath the title. The arrow moved into the component; plain prose bullets (v1.0.0's body) keep their own marker via `:not(:has(.cl-change))`.

**Section labels** (Added / Fixed / Internal) become a compact mono caption with a hairline rule running out to the right margin — group separation without a tall heading.

Everything that worked still works: hover/focus expand (desktop), tap-to-toggle (touch), JS-off always-visible descriptions, and the tag-chip page filter. Verified across light/dark and desktop/mobile; `astro check` is clean (0 errors).

Evidence (before → after) in a comment below.

### DIFF
--- a/website/src/components/Change.astro
+++ b/website/src/components/Change.astro
@@ -1,11 +1,11 @@
 ---
 /**
- * One changelog entry as a row in an aligned, table-like list: title in the
- * first column, area tags (what it touches — terminal, code tab, keyboard…)
- * in a fixed second column so chips line up down the page, and the PR chip
- * right-aligned in the third. The description is collapsed by default and
- * unfolds on hover / keyboard focus (desktop) or tap (touch), so a release
- * reads as a dense list of headlines you expand only where curious. Touch
+ * One changelog entry as a single dense row: the title on the left, and a
+ * right-grouped cluster of area tags (what it touches — terminal, code tab,
+ * keyboard…) plus the PR chip flush to the right margin, so every row is one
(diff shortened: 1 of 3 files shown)


srid:
## Evidence

Full `/changelog` page, desktop (1440px), light theme — same content, rendered before and after the rewrite. Total height drops from **3379px → 2911px**, and the per-entry density jump in the Unreleased list is the real story: one scannable line each, chips aligned right, no left rail.

| Before | After |
| --- | --- |
|  |  |

Hover-to-expand, touch tap-to-toggle, and the tag-chip filter all still work; verified in dark mode and at mobile width too.

PR #1292, opened 2026-06-12
kaval B0: invert spawn policy onto a fully-specified pty-host wire

srid:
**The pty-host wire becomes fully specified, and the daemon becomes policy-free.** `spawn` now carries the complete `{argv, env, initFiles}` it is to execute, and the host derives *nothing* from its own `process.env` — it materialises the rcfiles it is handed under its own `rcDir` and spawns the argv verbatim. All spawn policy moves to (or stays in) kolu-server's tier. This is the first PR of the [kaval reframing](https://github.com/juspay/kolu/blob/master/docs/atlas/src/content/atlas/pty-daemon.mdx): the contract change made at the one moment it's free — while everything is still in-process, with no daemon and no survivors.

```
            spawn {argv, env, initFiles}        ┌─ writes initFiles under rcDir
kolu-server ───────────────────────────────────▶│  spawns argv[0] verbatim
  (composes policy: cleanEnv ·                   └─ derives nothing from its env
   koluIdentityEnv · prepareShellInit            pty-host (policy-free)
   against system.info {shell,home,rcDir})
```

### Why this is safe to do now

Behavior is **byte-identical**. `cleanEnv()` / `koluIdentityEnv()` / `prepareShellInit()` run in the *same process* they did before (in-process B0), layered in the same order — so the env and rcfile content reaching every PTY are unchanged. The existing e2e (titles, cwd tracking, agent detection all ride the relocated OSC hooks) is the behavior-parity net; a golden test in `kolu-pty`'s `shell.test` locks the exact wrapper bytes.

### What changed

| Area | Before | After |
| --- | --- | --- |
| `spawn` input | `{id?, cwd?, cols?, rows?}` — host filled env/argv/rcfiles | `{id?, argv, cwd, env, initFiles, …}` — fully specified |
| Host facts | (none) | `system.info → {shell, home, platform, rcDir}` |
| `prepareShellInit` | wrote rcfiles, returned a `cleanup` cb | **pure** — plans `initFiles`; the host writes + reaps them |
| pty-host deps | `kolu-pty`, `kolu-common`, `kolu-shared` + framework | framework only (`@kolu/surface`, `@kolu/terminal-protocol`, node-pty, xterm) |
| Socket path | hardcoded app `"kolu"` | `app` param (default `"kolu"`) |
| Contract version | `2.1` | `3.0` (breaking spawn shape) |

The closure guard's allowlist is tightened to exactly the graduation set, so that **same test now fails if any spawn policy leaks back daemon-side** — the structural guard the plan calls for. A new `initFiles.ts` writes the wrapper files with a path-containment guard (a name can't escape `rcDir`).

> _Latent bug fixed for free:_ `configureNixShellEnv`'s whitelist and `cleanEnv` are now guaranteed same-process — which a later out-of-process daemon needs and the in-process arrangement happened to mask.

### Tests

- **Golden parity** (`kolu-pty/shell.test`) — the pure `prepareShellInit` plan: bash `--rcfile` / zsh `ZDOTDIR` paths into `rcDir`, replay-before-hooks ordering, nothing written to disk.
- **initFiles** (`pty-host/initFiles.test`) — write → paths → remove round-trip, empty-dir pruning, and the containment guard.
- **Contract** (`pty-host/inProcessPtyHost.test`) — `system.info` facts; the lifecycle suite now drives a fully-specified spawn (exercising the bare-client plain-shell path).
- Closure guard tightened; socket-path `app` param covered.

All unit suites green (pty-host 63 · kolu-pty 24 · pty-tui 33 · server 77). CI (incl. `ci::nix`) is the authoritative check for the workspace-only lockfile change.

> Plan of record: the B0 brief in [`pty-daemon.mdx`](https://github.com/juspay/kolu/blob/master/docs/atlas/src/content/atlas/pty-daemon.mdx). B1 renames the package to `kaval`, adds the binary + `kaval-tui`, and drops `kolu-tui`.


srid:

✅ **Consensus** after 4 round(s) · lowy + hickey · base `53c08897e06e`

Independent findings: lowy=7, hickey=6

### Applied (6)
- `lowy-1` `InitFile` type declared three times across the boundary — commit `ddf300904`
- `lowy-2` `ShellInitPlan` and `PtyHostSpawnInput` overlap with an unchecked rename (`args` vs `argv`) — commit `cc3568372`
- `lowy-5` `system.info` shell/home fallback chain partly duplicates `cleanEnv`'s own fallback — commit `9f5213cb3`
- `hickey-1` InitFile concept declared three times across the wire boundary — (uncommitted)
- `hickey-4` Resolved-cwd derived in two places, can transiently disagree — commit `1553386c0`
- `hickey-6` system.info shell/home fallback duplicates cleanEnv's SHELL logic — (uncommitted)

### Agreed — no change (7)
- `lowy-3` `Logger` structural type duplicated; pty-host's is the receptacle, kolu-pty has none (packages/pty-host/src/logger.ts:9)
- `lowy-4` `removeInitFiles` prune loop trusts `dir.startsWith(rcDir + sep)` for containment (packages/pty-host/src/initFiles.ts:52)
- `lowy-6` `DEFAULT_SCROLLBACK` default lives in the host but kolu always overrides it (packages/pty-host/src/inProcessPtyHost.ts:56)
- `lowy-7` `getPtyHostSocketPath` app param + cached `infoPromise` are correct receptacle shapes (packages/pty-host/src/socketPath.ts:20, packages/server/src/ptyHost.ts:53)
- `hickey-2` spawnInput test helper duplicated verbatim across two test files (packages/pty-host/src/inProcessPtyHost.test.ts:+spawnInput, packages/pty-tui/src/attach.test.ts:+spawnInput)
- `hickey-3` writeInitFiles materialisation re-implemented inside shell.test.ts (packages/integrations/pty/src/shell.test.ts:319-328)
- `hickey-5` removeInitFiles carries sibling-sharing machinery for a non-existent case (packages/pty-host/src/initFiles.ts:48-61)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `53c08897e06e`

### Round 1

**codex** — approved: `false`

The branch makes the intended boundary move: `@kolu/pty-host` no longer imports kolu policy packages, the spawn wire is now fully specified, and the server composes policy with `kolu-pty`. The overall direction is coherent, but I found one real cleanup bug in the new init-file materialization path and one boundary bug in the advertised remote-host composition. There are also a few documentation mismatches left by the inversion.

Findings:
- `F1` · major · open — `spawn` writes client-supplied init files before calling `host.spawn`, but cleanup is only registered as `onDispose` after a PTY entry exists. If `writeInitFiles` partially writes and then throws, or if `host.spawn` throws after the files are written, those rc files are left behind under `rcDir`. That breaks the new lifecycle guarantee that wrapper rc files are removed with the failed PTY spawn path, and can leave stale wrapper content for reused ids. (packages/pty-host/src/inProcessPtyHost.ts:205)
- `F2` · major · open — `buildTerminalSpawnInput` claims to compose policy against `system.info` for a host this process is not running on, but it prefers `cleanEnv()`'s local `SHELL` and `HOME` over the host facts. `cleanEnv()` always fills `SHELL`, so `info.shell` is effectively dead, and `HOME` will usually be the server machine's home. A remote pty-host would get local paths in `argv`, `cwd`, and shell-init replay paths, which defeats the B0 remote-enabler contract. (packages/server/src/ptyHost.ts:73)
- `F3` · minor · open — The Atlas B0 deliverable documents `system.info` as `{shell, home, platform}`, but the implemented contract also requires `rcDir`; clients need it to point argv/env at the host-side init-file paths. The plan now understates the actual wire shape. (docs/atlas/src/content/atlas/pty-daemon.mdx:107)
- `F4` · nit · open — The package overview still says `servePtyHost` prepares the shell env. After this change, `servePtyHost` explicitly does not prepare env or shell init; it only materializes supplied init files and spawns the supplied argv/env. (packages/pty-host/src/index.ts:12)
- `F5` · nit · open — The B0 section is labeled `in review` but the text says it was `Shipped in` PR 1292. Those states contradict each other in the plan-of-record. (docs/atlas/src/content/atlas/pty-daemon.mdx:101)

**claude** — Addressed all five CODEX findings. F1 (init-file leak on spawn failure): agreed and fixed in two layers — writeInitFiles now rolls back partial writes on a mid-stream failure, and the spawn handler wraps host.spawn in try/catch to remove already-written files when the fork throws (since onDispose never fires). Added a writeInitFiles rollback test and a new spawnFailureCleanup.test.ts that mocks createPtyHost to force a synchronous spawn throw (node-pty does NOT throw synchronously for bad shell/cwd — confirmed empirically — so the brittle integration approach was replaced with a mock). F2 (remote-host env basis): partial — the code is correct for what B0 actually ships (in-process, local host, where cleanEnv's local SHELL/HOME are the right facts and zero user-visible change is the contract), but CODEX is right that the comment overstated remote-readiness. Rewrote the buildTerminalSpawnInput doc comment to be honest that this is local-only today and that R-2 must invert the layering; did NOT flip the env ordering (that would be speculative and would regress the nix-shell SHELL override). Added two boundary-pinning tests. F3/F4/F5 (docs/comment mismatches): all agreed and fixed. Regenerated docs/atlas/dist via just atlas::build. just fmt clean; pty-host + server tests green (14 + 4 passing); both packages typecheck.

- `F1` **fixed** — Real bug — confirmed. Two-layer fix. (1) initFiles.ts writeInitFiles now collects written paths in a try block and calls removeInitFiles(rcDir, written) before rethrowing on any mid-stream write failure (ENOSPC/EACCES/ENOTDIR), so a partial materialisation never leaks. (2) inProcessPtyHost.ts spawn handler now wraps host.spawn in try/catch: if the fork throws, onDispose will never fire, so it calls removeInitFiles(rcDir, written) inline before rethrowing. Tests: added a writeInitFiles rollback test (parent-is-a-file forces ENOTDIR midway; asserts the earlier file is rolled back and the unrelated pre-existing file survives) and a new spawnFailureCleanup.test.ts. NOTE on the regression test: CODEX suggested forcing host.spawn to fail, but I verified node-pty does NOT throw synchronously for a nonexistent shell OR a nonexistent cwd (the pid comes back; failure surfaces async on the child). So I mock ./ptyHost.ts's createPtyHost to make spawn() throw deterministically, then assert rcDir is empty after the rejection (both the flat bashrc and the nested zdotdir dir are pruned).
- `F2` **partial** — Partly valid. CODEX is correct that env.SHELL ?? info.shell makes info.shell dead for a local host (cleanEnv's SHELL ??= /etc/passwd always fills it) and that HOME usually comes from the server's process.env — so the old comment's claim that system.info makes 'the same composition work for a host this process isn't running on (the R-2 remote enabler)' overstates what's delivered. BUT this is NOT a runtime bug: B0 is in-process only, the host IS this process, and local-env-wins is the correct, zero-user-visible-change behavior the B0 contract requires. R-2 (the real remote host) doesn't exist yet. I did NOT flip the env ordering as CODEX alternatively suggested — making host facts win now would be speculative wishful thinking and would actively regress the nix-shell case where cleanEnv DELIBERATELY overrides SHELL to the /etc/passwd login shell (shell.ts:84) over the nix /nix/store bash. Instead I took CODEX's narrow-to-local-only option: rewrote the buildTerminalSpawnInput doc comment to state plainly it is local-host-only today, that rcDir is the one fact consumed unconditionally, that shell/home are NOT yet remote-correct, and that wiring R-2 must replace the cleanEnv-wins layering with a host-fact-first one. Added two tests in ptyHost.test.ts pinning the boundary: local SHELL wins over system.info.shell (so a future remote inversion is a deliberate visible edit), and system.info.shell is the fallback when the parent env omits SHELL (the systemd-user-service path).
- `F3` **fixed** — Agreed. The implemented system.info contract returns {shell, home, platform, rcDir} (inProcessPtyHost.ts info handler + PtyHostSystemInfo), and clients need rcDir to point argv/env/initFiles at the host-side paths. Updated both enumerations of the wire shape in pty-daemon.mdx — line 56 ('system.info → {shell, home, platform}') and line 107 (the B0 deliverable) — to '{shell, home, platform, rcDir}', and added a clause on line 107 explaining rcDir is the host-side dir the client targets. The line-216 changelog entry is a prose summary, not a field enumeration, so left as-is. Regenerated dist.
- `F4` **fixed** — Agreed. index.ts's package overview said servePtyHost 'prepares the shell env and serves ptyHostSurface over createPtyHost' — false after B0. Rewrote it: servePtyHost derives no env or shell-init policy (B0, the kaval inversion); it only materialises the initFiles it is handed under the injected rcDir and spawns the supplied argv/env verbatim.
- `F5` **fixed** — Agreed. PR 1292 is OPEN (verified via gh pr view — mergedAt null, state OPEN), and the section pill is <Pill variant="done">in review</Pill>, so 'Shipped in PR 1292' contradicted both. Changed line 101 to 'In review in PR 1292.' to match the pill and the actual PR state. Regenerated dist.

commit: `6c0bba946e9de1bfe4b110c7afb953b3313cfe40`

### Round 2

**codex** — approved: `true`

All five prior findings are resolved in the current tree. F1 now has cleanup on both partial init-file writes and synchronous spawn failure, with targeted regression coverage. F2 is resolved by narrowing the server composition helper to the local-host contract rather than pretending it is remote-ready. The Atlas and package-overview wording mismatches from F3-F5 are corrected and reflected in the generated dist file.

Findings:
- `F1` · major · resolved — `spawn` wrote client-supplied init files before calling `host.spawn`, with cleanup only registered via `onDispose`, so failures before a PTY entry existed could leave rc files under `rcDir`. (packages/pty-host/src/inProcessPtyHost.ts:205)
- `F2` · major · resolved — `buildTerminalSpawnInput` previously claimed remote-host readiness while preferring local `cleanEnv()` shell/home values, making `system.info.shell` effectively dead and remote composition incorrect. (packages/server/src/ptyHost.ts:66)
- `F3` · minor · resolved — The Atlas B0 deliverable documented `system.info` without `rcDir`, even though the implemented wire contract includes and requires it. (docs/atlas/src/content/atlas/pty-daemon.mdx:107)
- `F4` · nit · resolved — The package overview still said `servePtyHost` prepares shell env, which is false after the B0 inversion. (packages/pty-host/src/index.ts:12)
- `F5` · nit · resolved — The B0 Atlas section said `Shipped in PR 1292` while the section status was `in review`. (docs/atlas/src/content/atlas/pty-daemon.mdx:101)

_codex on the rebuttal:_ F1: accepted. The fix handles both paths I was concerned about: `writeInitFiles` rolls back files already written during a failed materialization, and the spawn handler removes written files before rethrowing if `host.spawn` throws. The added tests cover both a mid-stream write failure and a synchronous `host.spawn` failure with nested init-file paths.

F2: accepted. I agree with not flipping env precedence in B0, because B0 is explicitly in-process/local and local env parity is the deliverable. The updated comment now makes the helper local-host-only and explicitly says R-2 must replace the `cleanEnv()`-wins layering or model a host env basis. That satisfies the alternative fix path from my review.

F3: accepted. The source Atlas note and generated HTML now enumerate `{shell, home, platform, rcDir}` and explain `rcDir`'s role.

F4: accepted. The package overview now states that `servePtyHost` derives no env/shell-init policy and only materializes supplied init files under `rcDir` before spawning supplied argv/env.

F5: accepted. The Atlas wording now says `In review in PR 1292`, matching the pill and PR state.

srid:

Ran after `/simplify` over the B0 diff (scoped to merge-base `53c08897`). Most of the surface was already settled by the lens and codex debates; the police pass found two quality items.

### Applied
- **`fix(police): avoid double userInfo() syscall in cleanEnv`** (`packages/integrations/pty/src/shell.ts`) — `cleanEnv()` called `userInfo()` twice per spawn (once in the whitelist branch to set `SHELL`, once via the trailing `??=`). Hoisted a single `loginShell`, used in both branches, with the `??=` moved into the `else` where it's reachable. — commit `c496f0e4`

### Surfaced, then dropped on review
- **Extract a shared `silentLog` test fixture** — the police pass extracted the duplicated `silentLog` into a new `packages/pty-host/src/testHelpers.ts`. **Reverted:** that file broke the staleKey **closure guard** (`buildId.closure.test.ts`) — the nix fileset hashes every non-`.test.ts` source file, but a test-only helper isn't reachable from `index.ts`, so it lands in the hashed set without belonging to the daemon closure. There is no filename that is both vitest-ignored *and* nix-excluded (only `.test.ts` is excluded, and that's a vitest suite). Inline-duplicating a 6-line fixture is the correct trade in a package whose whole point is a tight, reachable-from-entry hash set — the same reason `spawnInput` stays inline across the two suites.

### Skipped (with rationale)
- `spawnInput()` duplicated across `attach.test.ts` (pty-tui) and `inProcessPtyHost.test.ts` (pty-host) — different packages; sharing would need a public test-helper export or a third package, disproportionate for a 9-line helper.
- `materialise()` in `shell.test.ts` re-implements `writeInitFiles` — but it lives in `kolu-pty`, whose closure rule bars `@kolu/pty-host` imports.
- `|| "/"` cwd fallback in `buildTerminalSpawnInput` — already carries an explicit local-only comment deferring the host-fact-first basis to R-2.

Final state: `just check` green, all unit suites green (pty-host 65 · kolu-pty 24 · pty-tui 33 · server 81), atlas `dist` idempotent.

PR #1298, opened 2026-06-12
Defer Cmd/Ctrl+F to browser find inside the Code tab

srid:
**Pressing <kbd>Cmd/Ctrl+F</kbd> while focused in the Code tab now hands off to the browser's native find-in-page** instead of opening kolu's terminal search. The xterm search addon can only see terminal scrollback — it can't reach the Code tab's file viewer, a diff, rendered Markdown, or the sandboxed `.html` preview iframe. The browser's own find spans every frame (including the opaque-origin preview), so it's the right tool for "find in what I'm looking at."

The chord only defers while focus is inside the Code tab. Everywhere else — terminal, canvas, dock — it still opens kolu's terminal find exactly as before.

### How it works

The action registry gains an optional `when(e)` guard the global dispatcher consults *after* a chord matches:

- `when` returns **false** → the dispatcher skips the action without `preventDefault`, so the chord's browser default fires.
- `when` returns **true** (or is absent) → unchanged; the handler runs.

`findInTerminal` defers when `e.target` sits inside a `data-kolu-native-find` subtree, which the Code tab root carries. The content pane is made click-focusable (`tabindex=-1`) so clicking otherwise-unfocusable rendered content lands focus inside that subtree — that's what lets the guard fire for a plain source/preview view.

`matchesAnyShortcut` deliberately ignores `when`: inside the terminal the chord *is* an app shortcut, so xterm must keep letting it bubble — the focus-time decision belongs to the dispatcher, and terminal focus never overlaps Code-tab focus.

### Notable

> The hand-off is focus-gated, so it's inherently "click into the Code tab (or its preview) first." That ambiguity is intrinsic to one chord serving two scopes; a discoverability tip and changelog entry spell it out for users.

The e2e scenario asserts **both halves** (terminal → kolu search opens; Code tab → it does not) so it can't pass vacuously. A unit test covers the guard predicate directly.

### Try it locally

```sh
nix run github:juspay/kolu/leaden-chute
```


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `8f61e76a3a54`

### Applied (2)
- `hickey-2` DOM-querying logic placed in the action registry rather than the dispatcher — commit `0723efd45`
- `hickey-3` Keybind-match rule is computed in two places; `when` divergence now rides on top — commit `4df4958c1`

### Agreed — no change (7)
- `lowy-1` `when` guard is the correct receptacle pin, not a parallel encapsulation (packages/client/src/input/actions.ts:69)
- `lowy-2` Two focus-scoped dispatch-suppression mechanisms now coexist asymmetrically in the same dispatcher (packages/client/src/input/useShortcuts.ts:68)
- `lowy-3` `when` semantics (decline-without-preventDefault) is load-bearing but only documented in prose (packages/client/src/input/useShortcuts.ts:101)
- `lowy-4` Whole-Code-tab marker scope vs. content-pane scope is a deliberate, defensible boundary (packages/client/src/right-panel/CodeTab.tsx:698)
- `hickey-1` Two DOM-attribute focus guards now exist in two different layers (packages/client/src/input/actions.ts:191 and packages/client/src/input/useShortcuts.ts:68)
- `hickey-4` `when` guard couples its truth value to optional-chaining short-circuit semantics (packages/client/src/input/actions.ts:190-191)
- `hickey-5` Whole-Code-tab marker complects "search the content I'm viewing" with "focus anywhere in the Code-tab chrome" (packages/client/src/right-panel/CodeTab.tsx:698)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `8f61e76a3a54`

### Round 1

**codex** — approved: `false`

The implementation is small and directionally sound, but I found two behavior/test robustness issues and one stale-comment nit. No untracked files were present; I reviewed the changed files from the requested diff read-only.

Findings:
- `F1` · minor · open — When a scoped action decides to defer to the browser, `dispatch` uses `continue`, so the guarantee only holds as long as no later action has the same chord. That uniqueness is not enforced here, and a future duplicate Cmd/Ctrl+F action would still run and call `preventDefault`, silently breaking native find in the Code tab. (packages/client/src/input/useShortcuts.ts:94)
- `F2` · minor · open — The regression step uses `locator.focus()`, which bypasses the actual user gesture this change depends on: clicking into the Code tab/file content so focus lands inside the `data-kolu-native-find` subtree. If click focus regresses while programmatic focus still works, this scenario will keep passing even though the documented user flow is broken. (packages/tests/step_definitions/code_tab_steps.ts:288)
- `F3` · nit · open — The new comments refer to `actionMatches` in `useShortcuts.ts` and a `when` guard, but the refactor exports `actionMatchesKeybind` and uses `nativeFindMarker`. `CodeTab.tsx:697` has the same stale `when` wording. This will send future readers chasing code that no longer exists. (packages/client/src/input/actions.ts:308)

**claude** — Fixed all three findings.

- `F1` **fixed** — return false not continue
- `F2` **fixed** — real click + focus assertion
- `F3` **fixed** — stale comments updated

commit: `63e68523c5f266ebbb323c4b445ad4e5a27a6b69`

### Round 2

**codex** — approved: `true`

Verified Claude's current fixes against the updated tree. The shortcut hand-off now returns immediately, the e2e focus step uses a real click plus an active-focus assertion, and the stale comments were updated. I did not find any new regression introduced by this round's changes.

Findings:
- `F1` · minor · resolved — Previously, scoped native-find hand-off used `continue`, so a later duplicate chord could still claim the event and prevent browser find. (packages/client/src/input/useShortcuts.ts:96)
- `F2` · minor · resolved — Previously, the regression step used programmatic focus and did not exercise the actual user click path that must land focus inside the Code tab marker. (packages/tests/step_definitions/code_tab_steps.ts:292)
- `F3` · nit · resolved — Previously, comments referred to obsolete `actionMatches` / `when` wording after the refactor. (packages/client/src/input/actions.ts:308)

_codex on the rebuttal:_ No disputes were raised. I verified each claimed fix: F1 returns `false` instead of continuing, F2 uses a real click with a focus assertion, and F3's stale comment references were corrected.

srid:

Scoped to `8f61e76a3a54...HEAD`.

| Pass | Result |
| --- | --- |
| Rule checklist | ✅ clean — all 20 rules pass |
| Fact-check | 1 minor, skipped (see below) |
| Elegance | 3 fixes applied — commit `f7df239e` |

**Elegance fixes**
- Extracted `NATIVE_FIND_ATTR` / `NATIVE_FIND_MARKER` / `NATIVE_FIND_ATTR_PROP` so the `data-kolu-native-find` attribute (stamped on the Code tab root via spread) and the CSS selector (`findInTerminal.nativeFindMarker`) share one source of truth — a rename can no longer silently break the `closest()` deferral.
- Collapsed `actionMatchesKeybind`'s three-branch if/return into one boolean expression.

**Fact-check — skipped (1)**
- `dispatch`'s `closest(nativeFindMarker)` would throw on an invalid CSS selector. Skipped: the only value is a typed constant (`NATIVE_FIND_MARKER`), never user input — a runtime guard would add noise without real safety.

srid:

The **x86_64-linux** lane ran on **`kolu-ci-3`** — commit `f7df239e`, **passed** ✓

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 1m50s | ✓ |
| `ci::home-manager` | 1m6s | ✓ |
| `ci::nix` | 56s | ✓ |
| `_ci-setup` | 43s | ✓ |
| `ci::smoke` | 42s | ✓ |
| `ci::unit` | 22s | ✓ |
| `ci::fmt` | 10s | ✓ |
| `ci::biome` | 10s | ✓ |
| `ci::install` | 7s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/f7df239/timings.jsonl`; pool state is a live `flock` probe.</sub>

srid:
## Evidence

Cmd/Ctrl+F inside the Code tab, captured **in kolu itself** — a real headed Chromium, a genuine OS-level `Ctrl+F`, recorded with an X-server screen grab (the native find bar is browser chrome, so Playwright's page-only video can't see it).

A `page.html` containing the token `KOLUFIND9` is open in the Code tab's rendered preview. Pressing `Ctrl+F` and typing the token invokes the browser's **native find-in-page**, which highlights the match **inside kolu's sandboxed preview iframe** — the find bar reads **`1/1`**:



The token exists only inside `page.html` (the preview is `sandbox="allow-scripts"`, an opaque cross-origin frame), so the match proves native find reached into the preview — which kolu's own xterm search never could. That's the whole point of deferring the chord here.

▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=codetab-find-in-preview.mp4

The other half — that kolu **defers** the chord in the Code tab (and still opens terminal search everywhere else) — is covered by the e2e scenario `code-tab.feature` "Cmd/Ctrl+F opens terminal search from the terminal but defers in the Code tab", green in CI on both `x86_64-linux` and `aarch64-darwin`.

srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `826415e75b19`

### Applied (1)
- `hickey-5` Dispatcher unit test reimplements the dispatcher's decision instead of exercising it — commit `5a47e35fc`

### Agreed — no change (9)
- `lowy-1` focusScopeMarker is a clean receptacle for chord-confinement volatility (packages/client/src/input/actions.ts:74)
- `lowy-2` e2e step hardcodes the terminal-marker selector, bypassing the single-source-of-truth constant (packages/tests/step_definitions/code_tab_steps.ts:302)
- `lowy-3` Unit test reproduces the dispatcher's decision rule instead of driving the receptacle (packages/client/src/input/keyboard.test.ts:240)
- `lowy-4` actionMatchesKeybind extraction correctly unifies the keybind rule for both consumers (packages/client/src/input/actions.ts:331)
- `lowy-5` Code-tab pane tabindex=-1 places the focus-shift mechanism at the right layer (packages/client/src/right-panel/CodeTab.tsx:874)
- `hickey-1` Three selectors now answer "is focus inside a terminal" (packages/client/src/terminal/Terminal.tsx:1003,1036; packages/client/src/canvas/TerminalCanvas.tsx:62; packages/client/src/ui/ModalDialog.tsx:30)
- `hickey-2` focusScopeMarker generalizes a feature with exactly one instance (packages/client/src/input/actions.ts:63-74,210; packages/client/src/input/useShortcuts.ts:96-99)
- `hickey-3` Three exports single-source one attribute name, each derivation with one consumer (packages/client/src/input/actions.ts:138-145)
- `hickey-4` Terminal-DOM marker concept defined in the input module (packages/client/src/input/actions.ts:132-145)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 round(s) · codex reviewed at `xhigh` reasoning effort · base `826415e75b19`

### Round 1

**codex** — approved: `false`

The core focus-scoped shortcut implementation looks internally consistent: Cmd/Ctrl+F is matched first, the terminal root is marked, and the dispatcher declines without preventDefault outside that marker. I did not find a blocker or major correctness bug. I did find one test gap around proving the native handoff, one user-facing copy issue in the tips pool, and one stale dispatcher comment.

Findings:
- `F1` · minor · open — The new e2e scenario says it verifies that Cmd/Ctrl+F defers to native browser find outside the terminal, but the assertion only checks that Kolu's terminal search bar did not open. A regression that still calls preventDefault outside the terminal, or otherwise eats the chord without opening the terminal search, would pass this scenario while breaking the feature's main promise. (packages/tests/features/code-tab.feature:1822)
- `F2` · minor · open — The old ambient tip still says the find shortcut searches terminal output, and the new tip adds a second, longer message for the same chord. Users can now see two Ctrl/Cmd+F tips, and the old one is misleading when shown by itself because the chord no longer globally means terminal search. (packages/client/src/settings/tips.ts:46)
- `F3` · nit · open — The dispatch comment is now stale: dispatch can return false after a shortcut keybind matched when the action is focus-scoped and the event target is outside its marker. The return value now means claimed/handled, not matched. (packages/client/src/input/useShortcuts.ts:78)

**claude** — Addressed all three CODEX findings, all agreed. F1: added an e2e step that records the Cmd/Ctrl+F event's defaultPrevented via a bubble-phase window listener (fires after the app's capture-phase shortcut listener) and a Then asserting it is false, proving the chord actually reaches the browser rather than just that kolu's bar stayed closed. F2: removed the now-misleading amb-search tip and kept the accurate amb-codetab-find (new id so users who already saw the old one get the corrected copy). F3: rewrote the dispatch doc comment to describe the real claimed/preventDefault contract. Ran just fmt (clean), client typecheck (clean), and keyboard.test.ts (35 passed).

- `F1` **fixed** — Agreed. Added 'When I press the find shortcut, watching for native handoff' which installs a one-shot bubble-phase window keydown listener stashing e.defaultPrevented (bubble runs after the app's {capture:true} listener, so it sees the app's verdict), plus 'Then the find shortcut should reach the browser' asserting it is strictly false. Wired both into the Code-tab branch of the scenario in code-tab.feature (replaced the bare press + updated the scenario comment). A regression that preventDefaults outside the terminal without opening the bar now fails. Steps in packages/tests/step_definitions/keyboard_shortcut_steps.ts:74-120; feature at packages/tests/features/code-tab.feature:1821-1825. We can't observe the browser's native find chrome from the page, so not-prevented is the load-bearing proxy.
- `F2` **fixed** — Agreed. Removed the stale amb-search tip ('searches terminal output', misleading now that the chord is terminal-scoped) from AMBIENT_TIPS and kept the single accurate amb-codetab-find entry. Kept the new id (not reusing amb-search) so users who already saw the old tip still surface the corrected copy from the unseen pool. Verified no other references to amb-search across packages/website. packages/client/src/settings/tips.ts:45-52.
- `F3` **fixed** — Agreed. The comment 'Returns true if a shortcut matched' was stale since a focus-scoped chord can match yet return false. Rewrote it to: returns true if the event was claimed and the caller should preventDefault; a matched-but-out-of-scope chord returns false (declined, native default fires). packages/client/src/input/useShortcuts.ts:78.

commit: `2828910d4cc403a403cc968ec86ff4bd4bbb17de`

### Round 2

**codex** — approved: `false`

F2 and F3 are resolved. F1 is not resolved yet because the new handoff assertion is wired in the right place conceptually, but the one-shot keydown listener will be consumed by Playwright's modifier keydown before it ever sees the `f` keydown, so the new scenario will fail with an undefined recorded value instead of proving the app left Cmd/Ctrl+F unprevented.

Findings:
- `F1` · major · open — The new native-handoff recorder uses a `{ once: true }` keydown listener. Playwright implements `keyboard.press("Control+f")` / `keyboard.press("Meta+f")` by sending the modifier keydown first, then the `f` keydown. The listener is therefore removed on the modifier event, does not record `__findDefaultPrevented`, and the scenario fails with `undefined` before it can verify the actual app behavior. (packages/tests/step_definitions/keyboard_shortcut_steps.ts:95)
- `F2` · minor · resolved — Previously the ambient tips pool contained both the stale `amb-search` copy and the corrected `amb-codetab-find` copy for the same chord. (packages/client/src/settings/tips.ts:49)
- `F3` · nit · resolved — Previously the dispatch comment said true meant a shortcut matched, but focus-scoped matches can be declined and return false. (packages/client/src/input/useShortcuts.ts:78)

_codex on the rebuttal:_ F1: Holding open. I agree with the intended approach of checking `defaultPrevented` from a bubble-phase listener, but the implementation uses `{ once: true }`, so it observes the modifier keydown emitted by Playwright before the actual `f` keydown and removes itself too early. F2: Accepted; the stale tip is gone and the remaining new-id copy is accurate. F3: Accepted; the dispatcher comment now matches the current return-value contract.

**claude** — F1 was a real bug: the native-handoff recorder used a `{ once: true }` keydown listener, but Playwright presses the modifier (Control/Meta) down before `f`. `{ once: true }` removes a listener after any invocation — the in-body guard doesn't prevent invocation — so the listener was consumed by the modifier keydown and never saw the `f` event, leaving `__findDefaultPrevented` undefined and failing the Then with `undefined` instead of proving the chord reached the browser. Fixed by dropping `once` and self-removing only after the matching Cmd/Ctrl+F keydown is observed. F2 and F3 were already resolved in the prior round per codex's verdict. Ran just fmt (clean).

- `F1` **fixed** — Conceded — codex is right. `{ once: true }` removes the listener after the FIRST invocation regardless of the guard inside the callback (the guard only gates recording, not invocation). Playwright's keyboard.press("Control+f")/("Meta+f") sends the modifier keydown first, which invokes and removes the one-shot listener, so the subsequent `f` keydown is never observed and __findDefaultPrevented stays undefined → the Then throws on `undefined` rather than verifying handoff. Fixed at packages/tests/step_definitions/keyboard_shortcut_steps.ts:83-99: replaced the inline `{ once: true }` listener with a named `onKeyDown` handler that records `e.defaultPrevented` only when the Cmd/Ctrl+F chord arrives and then calls removeEventListener on itself. Updated the comment to explain why `once` is unsafe here. The bubble-phase timing (capture before bubble) is unchanged, so the assertion still reads the app's post-dispatch verdict.
- `F2` **fixed** — Already resolved in the prior round and confirmed resolved by codex's verdict (status: resolved). The stale amb-search tip was removed and only the corrected amb-codetab-find entry remains in AMBIENT_TIPS. No further change this round.
- `F3` **fixed** — Already resolved in the prior round and confirmed resolved by codex's verdict (status: resolved). The dispatch doc comment in useShortcuts.ts now describes the claimed/preventDefault contract including the matched-but-declined (out-of-scope) case returning false. No further change this round.

commit: `f6181e59f9aa862db15ab550473e8007db47f39f`

### Round 3

**codex** — approved: `true`

All findings are now resolved. I verified the current diff and status, and the F1 closeout fix now records the actual Cmd/Ctrl+F keydown by keeping the bubble listener installed through the modifier keydown and self-removing only after the matching chord event is observed.

Findings:
- `F1` · major · resolved — The native-handoff assertion previously used a one-shot keydown listener that Playwright consumed on the modifier keydown before the `f` keydown arrived. (packages/tests/step_definitions/keyboard_shortcut_steps.ts:92)
- `F2` · minor · resolved — Previously the ambient tips pool contained both the stale `amb-search` copy and the corrected `amb-codetab-find` copy for the same chord. (packages/client/src/settings/tips.ts:49)
- `F3` · nit · resolved — Previously the dispatch comment said true meant a shortcut matched, but focus-scoped matches can be declined and return false. (packages/client/src/input/useShortcuts.ts:78)

_codex on the rebuttal:_ F1: Accepted. The revised listener avoids the Playwright modifier-keydown trap and self-removes only after observing the actual Cmd/Ctrl+F keydown, so the assertion now checks the app's post-capture defaultPrevented verdict as intended. F2 and F3 remain resolved from the prior round.

srid:

Scoped to `826415e75b19...HEAD`. Elegance pass skipped (`/simplify` already ran this gauntlet).

| Pass | Result |
| --- | --- |
| Rule checklist | 1 raised → skipped (see below) |
| Fact-check | 1 raised → skipped (see below) |

**Both findings skipped, with reasons — no fixes applied:**

- **`dry-rule-of-three`** — the e2e step hardcodes the `[data-kolu-terminal-search]` selector instead of importing `TERMINAL_SEARCH_MARKER`. Already adjudicated by the lens debate (`lowy-2`, dropped): the literal runs inside a `page.evaluate` browser context across the client/tests package boundary, and a rename would make the scenario **fail loudly** (the step asserts focus left that exact subtree) — drift is caught by a red test, not a silent pass.
- **Replace `(e.target as Element \| null)?.closest?.()` with an `instanceof Element` guard** — would **break the unit test**. The node-env test drives the real `isOutsideFocusScope` predicate with duck-typed `{ closest }` fakes (no DOM to build real Elements); `instanceof Element` rejects them. The optional-chaining duck-type is deliberate, and the behavior is already correct for non-Element targets (outside any terminal → defer to the browser, which is right).

srid:

The **x86_64-linux** lane ran on **`kolu-ci-4`** (`dev-x86-64-linux-08`) — commit `f6181e59`, **failed** ✗

- **Lane wall** (pipeline): **3m16s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 2m7s | ✗ (exit 1) |
| `ci::home-manager` | 1m6s | ✓ |
| `ci::smoke` | 47s | ✓ |
| `_ci-setup` | 47s | ✓ |
| `ci::nix` | 44s | ✓ |
| `ci::atlas-sync` | 42s | ✓ |
| `ci::pnpm-hash-fresh` | 29s | ✓ |
| `ci::unit` | 28s | ✓ |
| `ci::install` | 21s | ✓ |
| `ci::surface-example-build` | 19s | ✓ |
| `ci::surface-app-example-build` | 19s | ✓ |
| `ci::biome` | 17s | ✓ |
| `ci::fmt` | 16s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/f6181e5/timings.jsonl`; pool state is a live `flock` probe.</sub>

PR #1303, opened 2026-06-12
Keep wrapped URLs clickable across a terminal resize

srid:
**A long URL that wrapped across terminal lines could open a clipped, middle-missing address when clicked.** The culprit wasn't the link parser — it was xterm's reflow. When the wrapped line sits on the *cursor row* (a URL printed without a trailing newline, or a live prompt) and the grid narrows, `_reflowSmaller` deliberately skips that line — `reflowCursorLine` defaults to `false`, on the assumption the shell will redraw it — but then still trims every row to the new width. The URL's overflow is **truncated instead of rewrapped**, and `@xterm/addon-web-links` joins the leftovers into a broken link. kolu refits terminals constantly (canvas tiles, zoom, window resize, split panes), so this fired far more than the "a shell will redraw it" assumption expects.

### What actually happens

```
print at 80 cols          resize -> 40 cols (cursor on the line)
┌────────────────────────┐   reflowSmaller SKIPS the cursor line,
│https://ex…/really/long/│   then trims every row to 40 wide:
│resource?q=…&x=12  ◀cur │     row0 = "https://ex…/really/lo"  (40, kept)
└────────────────────────┘     row1 = "…&x=12"                 (old tail)
                               join  = https://ex…/really/lo…&x=12   ✂ clipped
```

### The fix

Set `reflowCursorLine: true` — the upstream xterm option for exactly this — on **both** terminal construction sites:

| Site | Why it needs it |
| --- | --- |
| `Terminal.tsx` (client xterm) | The terminal the user actually clicks. |
| `ptyHost.ts` (server headless xterm) | Its serialized snapshot *is* the scrollback a client restores on attach/reconnect — so a URL left on the cursor line when the PTY resizes must survive here too. |

The option only rewraps line *contents*; the cursor position is unchanged. _Verified against a real headless buffer: with the flag off every cursor-on-line narrowing resize clips; with it on, all of them preserve the full URL — widening was never affected._

A pty-host test reproduces the bug end-to-end (spawn a shell, print a wrapping URL on the cursor line, narrow 80→40, assert the URL survives) and **fails without the fix** — clipping to `…/really/lone=stuff&x=12`, the exact symptom.

### Try it locally

```sh
nix run github:juspay/kolu/ready-circus
```


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `5f67d6b1b5c8`

### Agreed — no change (5)
- `lowy-1` Cross-package terminal-config agreement is held together by a comment, not a receptacle (packages/pty-host/src/ptyHost.ts:419-426 and packages/client/src/terminal/Terminal.tsx:495-506)
- `lowy-2` reflowCursorLine flag delegates reflow volatility to xterm correctly — not a finding, a confirmation (packages/client/src/terminal/Terminal.tsx:504 and packages/pty-host/src/ptyHost.ts:425)
- `lowy-3` Regression test guards the production wiring end-to-end at the right boundary (packages/pty-host/src/ptyHost.test.ts:194-225)
- `hickey-1` reflowCursorLine policy fragmented across two construction sites held together by a comment (packages/pty-host/src/ptyHost.ts:425 (and packages/client/src/terminal/Terminal.tsx:504))
- `hickey-2` Pre-existing: DEFAULT_SCROLLBACK itself is duplicated (common=50_000 vs pty-host=10_000) (packages/pty-host/src/ptyHost.ts:38 (vs packages/common/src/config.ts:29))

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `5f67d6b1b5c8`

### Round 1

**codex** — approved: `false`

The production change is narrowly scoped and appears to use a real xterm option in both the browser and headless terminals. I found one test robustness issue: the new regression test can resize before the full URL has been parsed if PTY output is chunked unfavorably.

Findings:
- `F1` · minor · open — The test waits only for `another=thing`, which occurs before the end of the 92-character URL. PTY output chunk boundaries are not guaranteed, so the test can resize after only the first part of the URL has reached the headless terminal; the immediate assertion can then fail nondeterministically for a partial-output reason instead of the reflow behavior under test. (packages/pty-host/src/ptyHost.test.ts:218)

**claude** — CODEX's single finding (F1) is valid and now fixed. The URL is 92 chars; `another=thing` ends at offset 76, leaving 16 trailing chars (`&more=stuff&x=12`). Since PTY output arrives in arbitrary chunks, the original `waitFor(() => ...includes("another=thing"))` could return after only the first part of the URL reached the headless terminal, fire the resize on a half-written URL, and then fail `expect(joined).toContain(url)` for a partial-output reason rather than the reflow behavior under test. I changed the wait to join the wrapped rows and wait for the FULL url before resizing, then assert the same joined-full-URL condition afterward, removing the chunk-boundary nondeterminism. Test passes (vitest run, 201ms) and biome reports no formatting fixes needed. The production change (reflowCursorLine:true in both Terminal.tsx and ptyHost.ts) was not flagged and CODEX agreed it uses a real xterm option correctly.

- `F1` **fixed** — Agreed and fixed at packages/pty-host/src/ptyHost.test.ts:217-223. The 92-char URL has its tail (&more=stuff&x=12, 16 chars) after the matched substring 'another=thing' (ends at offset 76), so waiting only for that interior substring could let the resize fire before the URL's tail was parsed, failing the final assertion for a partial-output reason. Introduced `const joinedScreen = () => host.getScreenText(id).replace(/\n/g, "")` and changed the gate to `await waitFor(() => joinedScreen().includes(url))` so we wait for the whole wrapped URL before `host.resize(id, 40, 24)`, then assert `expect(joinedScreen()).toContain(url)`. Verified: the regression test passes (vitest run), and biome check reports no formatting fixes.

commit: `a506df5ecfa91af4ec77cf4fe0175cf402faaa49`

### Round 2

**codex** — approved: `true`

CLAUDE's fix addresses the only open finding. The test now joins wrapped screen rows and waits for the complete URL before resizing, then asserts the complete URL after resizing, which removes the PTY chunk-boundary race I raised.

Findings:
- `F1` · minor · resolved — The prior test waited only for an interior URL substring before resizing, so PTY chunking could make it resize a partially parsed URL and fail for the wrong reason. (packages/pty-host/src/ptyHost.test.ts:223)

_codex on the rebuttal:_ F1 was not disputed. I verified the current test waits for the full joined URL before `host.resize(id, 40, 24)` and reuses that joined read for the post-resize assertion, so the chunk-boundary race is closed.

srid:

Ran the three passes over `5f67d6b1...HEAD` (rule checklist · fact-check · elegance).

**Rule checklist & fact-check** — clean. The change is two `reflowCursorLine: true` literals on existing terminal constructors plus a focused regression test; no swallowed errors, wishful fallbacks, reserved-keybinding collisions, or toast/convention violations in scope.

**Elegance** — one finding surfaced and **rejected by the adjudicator**:

| Finding | Disposition |
| --- | --- |
| Add `reflowCursorLine: true` to the `createTerminal` and `freshHeadless` test fixtures "to match production config" | **Rejected.** Both fixtures back tests that never resize, so the option is inert there — pure scope creep onto unrelated helpers, and it directly contradicts the lens-debate consensus *not* to multiply this literal. Reverted. |

Net: no code-police commits. The production fix and its dedicated `createPtyHost` regression test stand as-is.

srid:
## Evidence

This is a behavioral fix (a clicked link opening a *clipped* address), demonstrated two ways.

### 1. Before/after on the live client xterm

Driving a real kolu terminal (chrome-devtools MCP), I printed a 152-char sign-in URL so it wraps with the cursor on it, then narrowed the grid 90→40 cols — once with the option off (the old behavior) and once on (the fix). Same URL, same resize, only `reflowCursorLine` flipped:

| `reflowCursorLine` | URL recovered from the buffer after the resize |
| --- | --- |
| `false` *(old default)* | `https://example.com/oauth2/device/author``ly-long-opaque-state-token-…` — ✂️ **clipped**, the middle gone |
| `true` *(this PR)* | `https://example.com/oauth2/device/authorize?client_id=kolu-cli&scope=read+write&state=really-long-opaque-state-token-…` — ✅ **whole** |

The buggy run literally jumps from `…/device/author` straight to `ly-long-opaque-state…` — the segment between is truncated, which is the clipped address a click would have opened. With the fix the wrapped line rewraps and the full URL survives.

A long URL wrapped in a real kolu terminal (the on-screen scenario):

wrapped URL in a kolu terminal

### 2. Automated guard

`packages/pty-host/src/ptyHost.test.ts` spawns a real shell, prints a wrapping URL on the cursor line, narrows 80→40, and asserts the URL survives. Flipping the production flag back to `false` makes it fail with exactly the bug:

```
Expected: "https://example.com/path/to/a/really/long/resource?query=value&another=thing&more=stuff&x=12"
Received: "https://example.com/path/to/a/really/lone=stuff&x=12"
```

_Captured on a production-safe `just dev-auto` instance (random ports); the systemd `kolu.service` was untouched._

srid:

The **x86_64-linux** lane ran on **`kolu-ci-8`** (`dev-x86-64-linux-08`) — commit `a506df5e`, **passed** ✓

- **Lane wall** (pipeline): **3m27s**

| recipe | duration | |
|---|--:|:--|
| `ci::pnpm-hash-fresh` | 2m18s | ✓ |
| `ci::e2e` | 2m0s | ✓ |
| `ci::home-manager` | 1m11s | ✓ |
| `_ci-setup` | 1m9s | ✓ |
| `ci::smoke` | 44s | ✓ |
| `ci::nix` | 42s | ✓ |
| `ci::unit` | 30s | ✓ |
| `ci::surface-example-build` | 21s | ✓ |
| `ci::surface-app-example-build` | 21s | ✓ |
| `ci::biome` | 20s | ✓ |
| `ci::fmt` | 19s | ✓ |
| `ci::install` | 14s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/a506df5/timings.jsonl`; pool state is a live `flock` probe.</sub>

PR #1320, opened 2026-06-13
fix(nix): drop kaval-build-id IFD; run `nix flake check` in CI

### LINKED ISSUE #1317: `nix flake check` fails: "path '…-kaval-build-id.drv' is not valid" (IFD in default.nix)
## Summary

`nix flake check` fails during **evaluation** (before building any check) with:

```
error: path '736y3pzfixfnnlknicginpj1a6m8fs0k-kaval-build-id.drv' is not valid
```

The failure is not in any check derivation — it's triggered while evaluating the `kolu-bin` derivation, which reads an **import-from-derivation (IFD)** value: `kavalBuildId`.

## Root cause

`default.nix:120` computes a build id by `builtins.readFile`-ing a `runCommand` output:

```nix
kavalBuildId = builtins.readFile (pkgs.runCommand "kaval-build-id"
  { src = kavalSrc; } ''
  cd "$src"
  find . -type f | LC_ALL=C sort | xargs cat | sha256sum | cut -c1-64 \
    | tr -d '\n' > $out
'');
```

This is an IFD: to *evaluate* `kavalBuildId` (used at `default.nix:231` and `:283` as `--set KAVAL_BUILD_ID`), Nix must realise the `kaval-build-id` derivation mid-evaluation. `nix flake check` walks every flake output's `drvPath`, which forces this IFD, and the realisation trips on `path '…-kaval-build-id.drv' is not valid`.

Introduced in #1301 (kaval B1) — `kavalBuildId` is on `master` (`default.nix:120`).

## Repro

```console
$ nix flake check --no-build
...
       … while calling the 'readFile' builtin
         at .../default.nix:120:18:
          120|   kavalBuildId = builtins.readFile (pkgs.runCommand "kaval-build-id"
       … while realising the context of path '/nix/store/81fc16vai0rpnyw1y0x98lqgcnprg0yh-kaval-build-id'
       error: path '736y3pzfixfnnlknicginpj1a6m8fs0k-kaval-build-id.drv' is not valid
```

- nix 2.34.7
- `nix build .#website`, `.#checks.<sys>.website-typecheck`, etc. build fine individually — only `nix flake check` (which forces the IFD across all outputs) fails.

## Notes / possible fixes

- This surfaced incidentally while verifying an unrelated website-nix change; it is **not** caused by that change.
- Options to consider:
  - Drop the IFD — derive `KAVAL_BUILD_ID` inside the build sandbox (write it to a file during `buildPhase` and read at runtime) instead of `readFile` at eval time, so no eval-time realisation is needed.
  - Or hash `kavalSrc` purely in Nix (e.g. a content hash of the fileset) rather than shelling out via `runCommand` + `readFile`.
  - If the IFD must stay, ensure `nix flake check` has a built/cached `kaval-build-id` available, or exclude it from the checked output set.

srid:
Closes #1317.

## The bug

`nix flake check` failed during **evaluation** with:

```
error: path '…-kaval-build-id.drv' is not valid
```

`kavalBuildId` (`default.nix`) was an **import-from-derivation** — `builtins.readFile` of a `runCommand` that hashed kaval's daemon source closure (`find | sort | cat | sha256sum`). To *evaluate* the value, Nix had to realise that build mid-eval, which is exactly what eval-only (`--no-build`) `nix flake check` refuses to do.

## The fix

The id only needs to be a **stable function of `kavalSrc`** that changes iff a hashed daemon file changes:

- The runtime side just *reads* `KAVAL_BUILD_ID` (`packages/kaval/src/buildId.ts`) — nothing recomputes the digest.
- `buildId.closure.test.ts` pins the hashed **file set**, not the hash algorithm.

`kavalSrc` is already content-addressed (`fileset.toSource` adds it to the store at eval time), so its store path changes iff a hashed file changes. So hash that path **purely in Nix**:

```nix
kavalBuildId = builtins.hashString "sha256" "${kavalSrc}";
```

No IFD, faster eval, same guarantee — and byte-identical across Darwin/Linux for free (the NAR hash is platform-neutral, so the old `LC_ALL=C`-sort rationale retires with the shell pipeline).

## CI: `nix flake check` node

Adds a `flake-check` recipe to `ci/mod.just`, wired into the `default` DAG. It's the whole-output **evaluation** + flake-schema gate, complementing the existing devour-flake `nix` node (which builds the output drvs). Store locking dedups its `checks.*` builds against that node, so it overlaps rather than repeats the big build.

### Honest scoping

- **Full `nix flake check` already passed on `master`** (a full build is allowed to realise IFDs) — so the CI node is green today, and its value is guarding *future* eval regressions plus closing the specific IFD #1317 named. The IFD only broke the eval-only (`--no-build`) path.
- CI runs **full** `nix flake check`, not `--no-build`: `mini-ci-runner` deliberately bakes `${mini-ci-runner.drvPath}` into its wrapper (so `nix run .#mini-ci` can build the runner on demand), which `--no-build` can't satisfy by design. That's intentional, not a bug, so `--no-build` isn't a usable gate.

## Verification

- `nix flake check` → `all checks passed!`
- `nix build .#kaval` → ok (id evaluates and bakes into the wrapper)
- `vitest packages/kaval/src/buildId` → 4/4 (unit + closure)
- `just fmt` clean

### DIFF
--- a/ci/mod.just
+++ b/ci/mod.just
@@ -55,7 +55,7 @@ nix_shell := if env('IN_NIX_SHELL', '') != '' { '' } else { 'nix develop --accep
 [macos]
 [parallel]
 [metadata("ci")]
-default: nix home-manager e2e smoke fmt biome unit surface-example-build surface-app-example-build pnpm-hash-fresh atlas-sync
+default: nix flake-check home-manager e2e smoke fmt biome unit surface-example-build surface-app-example-build pnpm-hash-fresh atlas-sync
 
 # Single per-platform `pnpm install`. The runner fans every recipe out into
 # its own process and starts them all in parallel as soon as setup completes,
@@ -89,6 +89,17 @@ pnpm-hash-fresh:
 # is content-addressed, so it only re-runs when a typechecked source changes.
 nix:
     nix build github:srid/devour-flake -L --no-link --print-out-paths --override-input flake .
+
+# `nix flake check`: evaluate every flake output and run the flake's `checks.*`.
+# Complements the `nix` node (devour-flake builds the output drvs) with the
+# flake-schema + whole-output *evaluation* gate — it forces every output to
+# evaluate, so an eval-time regression like an import-from-derivation that can't
+# be realised mid-eval (juspay/kolu#1317) fails here even when a plain build of
+# the affected drv would have succeeded. Store locking dedups its check builds
+# against the concurrent `nix` node, so it overlaps that build rather than
+# repeating it. No `nix` dep — it's an independent parallel leaf.
+flake-check:
+    nix flake check --accept-flake-config -L
 
 # Builds the home-manager example flake, which pulls kolu in via
 # `--override-input flake/kolu .` — so it builds kolu from source itself. No
--- a/default.nix
+++ b/default.nix
@@ -87,9 +87,10 @@ let
   # restart would load different daemon code. Scoped to the three roots that run
   # IN the daemon — kaval, terminal-protocol, and the surface-daemon spine — so
   # server-/client-only deploys leave it unchanged (no over-prompting). The .ts
-  # filter drops `.test.ts` AND `.testlib.ts` (shared test-only helpers), and
-  # LC_ALL=C sort makes the hash byte-identical across Darwin/Linux. The set
-  # hashed here is asserted to equal the daemon's reachable closure by
+  # filter drops `.test.ts` AND `.testlib.ts` (shared test-only helpers); the id
+  # (below) is taken over this fileset's content-addressed store path, whose NAR
+  # hash is byte-identical across Darwin/Linux. The set hashed here is asserted
+  # to equal the daemon's reachable closure by
   # packages/kaval/src/buildId.closure.test.ts — keep the fileFilter and that
   # test in lockstep.
   isHashedSource =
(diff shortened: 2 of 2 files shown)


srid:
## Evidence — the new hashing doesn't break `KAVAL_BUILD_ID`

**The contract.** `KAVAL_BUILD_ID` exists for staleness detection: it must be a *deterministic* function of kaval's daemon source closure that changes **iff** a hashed daemon file changes — and stays put for everything else (tests, server/client-only edits), or a restart over-prompts.

### Why it's structurally safe

The PR doesn't touch the **input** to the digest — `kavalSrc` (the `fileset.toSource` over kaval / terminal-protocol / surface-daemon) is byte-for-byte the same. It only swaps the *pure function* applied to that fixed input:

| | before | after |
|---|---|---|
| input | `kavalSrc` | `kavalSrc` *(unchanged)* |
| digest | `readFile (runCommand: find │ sort │ cat │ sha256sum)` — **IFD** | `builtins.hashString "sha256" "${kavalSrc}"` — pure |

Both are collision-resistant digests of the same store path. So *anything that changed the id before still changes it now, and anything that didn't still doesn't* — only the literal hex string differs, and the runtime (`packages/kaval/src/buildId.ts`) reads it opaquely (nothing recomputes it).

### Empirical proof — mutation experiment on the real `.#kaval` artifact

Extracting the baked `KAVAL_BUILD_ID` from the built wrapper after each change:

| scenario | `KAVAL_BUILD_ID` | expected | result |
|---|---|---|---|
| **baseline** | `8529bab6…efe0c3c4` | — | — |
| re-evaluate, no change | `8529bab6…efe0c3c4` | identical | ✅ deterministic |
| edit a **hashed** daemon file (`src/buildId.ts`) | `0a112c1f…8d3ee206` | **differs** | ✅ change detected |
| revert | `8529bab6…efe0c3c4` | back to baseline | ✅ |
| edit a **non-hashed** file (`src/buildId.test.ts`) | `8529bab6…efe0c3c4` | **unchanged** | ✅ correctly scoped (no over-prompt) |

So the new algorithm has *identical* change-detection behaviour: sensitive to daemon source, blind to test-only edits.

### Guardrails still green

```
vitest packages/kaval/src/buildId  →  Test Files 2 passed (2) · Tests 4 passed (4)
```
- `buildId.closure.test.ts` — asserts the hashed file **set** still equals the daemon's reachable closure (it pins the *set*, not the algorithm, so it's the right guard and it passes).
- `buildId.test.ts` — runtime reads the baked env.
- `nix build .#kaval` → ok · `nix flake check` → `all checks passed!`

### Platform parity

The old `LC_ALL=C` sort existed to make the `find │ cat` byte stream identical across Darwin/Linux. The new id is `sha256` of `kavalSrc`'s content-addressed store path, whose **NAR hash is platform-neutral by construction** — so cross-platform parity is now a property of Nix itself rather than a shell incantation we maintain.

PR #1324, opened 2026-06-13
Stamp the build commit on the no-store shell, not an immutable bundle

### LINKED ISSUE #1319: koluStamped breaks the immutable-asset contract: same bundle filename, different commit stamp — the real "App updated" reload loop
**The "App updated → Reload" card can loop forever — clicking Reload re-fetches a fresh shell but the browser (correctly) serves the OLD bundle from its `immutable` cache, because `koluStamped` changes the bundle's content without changing its filename.** #1278 busted the shell's cache key; the shell was never the stale layer. The poisoned artifact is the *asset*, and it is poisoned **by the build system, not the browser**.

## Symptom (as reported on #1278)

After a deploy, the card appears; clicking **Reload** navigates to `/?__surface_app_fresh=<token>`, the page comes back with `CLIENT <old> · ≠ srv`, and the card returns. Repeat clicks never converge. Observed simultaneously on two machines (zest → `CLIENT d8acaf1 · SRV 7deb397`; pureintent → `CLIENT 7deb397 · SRV 0fab0cc`).

## Root cause

`default.nix` builds the client with a placeholder commit so docs-only commits don't bust the derivation cache, then `koluStamped` `sed`s the real sha into the **already-content-hashed** bundle:

```nix
# Build uses a placeholder so docs-only commits don't bust the derivation cache;
# koluStamped sed-replaces it with the real hash afterwards.
koluStamped = pkgs.runCommand "kolu-stamped" { } ''
  cp -r ${kolu} $out
  chmod -R u+w $out/packages/client/dist
  find $out/packages/client/dist -name '*.js' -exec \
    sed -i 's/${koluCommitPlaceholder}/${commitHash}/g' {} +
'';
```

When two consecutive deploys differ only **outside** the client build fileset (docs, `packages/tests`, website…), both ship a bundle with the **same filename** but a **different embedded sha** — and `/assets/*` is served `Cache-Control: public, max-age=31536000, immutable`.

The loop is then mechanical:

1. Browser loads deploy N → caches `/assets/index-D85Q74Rn.js` (stamped `7deb397`) — pinned immutable for a year.
2. Deploy N+1 (`0fab0cc`, a docs-only commit) restarts the server → card appears.
3. Click Reload → `/?__surface_app_fresh=…` → **fresh `no-store` shell from the network** → it references `index-D85Q74Rn.js` → browser serves the **old-stamped copy from cache, zero network bytes** (this is exactly what `immutable` is for).
4. `clientIsStale("0fab0cc", "7deb397")` → true → card again → goto 3. ∞

No shell-level cache busting can ever fix this; the stamp travels inside the immutable asset.

## Evidence

**1. The deployed artifacts themselves** (store paths of the two builds running on pureintent on 2026-06-12):

```
$ basename <7deb397-dist>/assets/index-*.js   → index-D85Q74Rn.js
$ basename <0fab0cc-dist>/assets/index-*.js   → index-D85Q74Rn.js     # SAME name
$ grep -c 7deb397 <7deb397-dist>/assets/index-D85Q74Rn.js  → 1
$ grep -c 0fab0cc <0fab0cc-dist>/assets/index-D85Q74Rn.js  → 1        # different stamp
$ cmp <(sed s/7deb397/X/g …) <(sed s/0fab0cc/X/g …)        → IDENTICAL modulo stamp
$ git diff --stat 7deb3970..0fab0cc8                       → only docs/atlas/**
```

**2. Live reproduction** (clean pu box running the two real builds; clean browser profile):

- Load `7deb397` build → `SRV 7deb397 · CLIENT 7deb397`, bundle fetched from network (2.5 MB), now cached immutable.
- Swap server to `0fab0cc` (same port) → card appears.
- **Click Reload #1**: URL becomes `/?__surface_app_fresh=1781314921279`; shell transfers 1426 B from network; bundle transfers **0 B** (`deliveryType: "cache"`); rail reads `SRV 0fab0cc · CLIENT 7deb397 · ≠ srv`; **card is back**:

  click 1

- **Click Reload #2**: new token, same result — `CLIENT 7deb397`, asset from cache, card back. The loop never converges:

  click 2

**3. It matches both field screenshots from #1278**: each shows CLIENT exactly one *stamp-only sibling* behind SRV (`d8acaf1→7deb397` differs only in `packages/tests`, which the build fileset excludes; `7deb397→0fab0cc` only in `docs/atlas`).

**4. It explains the saga's oldest mystery** (`docs/cache-bug.md`): "a hard reload fixed it for exactly one load" — a hard reload bypasses the cache, refetches the asset, and *replaces* the cached stamp… until the next stamp-only deploy re-poisons it by construction.

## What it is NOT (verified, so this doesn't get re-litigated a fifth time)

- **Not a heuristically-fresh poisoned `/` entry** (#1278's theory). Three independent refutations: (a) Hono's `serveStatic` has never sent `Last-Modified`/`ETag` — there is no creation mechanism (the only kolu that 304'd on epoch mtimes was the 3-day Rust era in March, default port 7681); (b) the affected browser's HTTP cache contains **no entry for the bare shell URL** on either origin (782+259 entries enumerated — all `/api/terminals/.../file/*`, `/assets/*`, fonts); (c) Chromium always revalidates the main document on a normal reload (`LOAD_VALIDATE_CACHE`), and a `no-store` 200 *dooms* a stored entry — verified empirically with a manufactured 1970-`Last-Modified` poison: one normal reload escaped and durably evicted it.
- **Not the service worker** — though live workbox-era SW registrations + ~40 MB precaches (frozen 2026-06-03, `index-Be1jJw2f.js`) still exist for `http://zest:7692` and `http://pureintent:7692` in the affected Chrome profile. They are inert (the insecure-origin flag is gone, so no SW runs on those http origins) but are landmines if that flag ever returns: worth a one-time **Clear site data** on both origins.
- **Not a crash loop** — zero crashes in either box's history (systemd: 0 `Scheduled restart` in a week; launchd: `runs = 1`, never exited). All 22 (pureintent, 8 days) / 155 (zest, 6 weeks) restarts are deploys or manual.

## Fix

**Move build identity out of the immutable asset and into the `no-store` shell.**

- `koluStamped` stamps `index.html` (e.g. `<script>window.__SURFACE_APP_COMMIT__="…"</script>`) instead of `dist/**/*.js`.
- The client reads the commit from the shell at boot instead of a Vite `define` baked into the bundle.

This restores the immutability contract (same filename ⇒ same bytes), makes stamp-only deploys produce **no skew at all** (the card doesn't even show — server and client report the same sha), and makes real deploys converge on one reload (new code ⇒ new hash ⇒ new filename ⇒ network fetch). `@kolu/surface-app`'s commit plumbing changes (`surfaceApp()` Vite plugin / `__SURFACE_APP_COMMIT__`), so a companion drishti PR is required per `.claude/rules/surface.md`. #1278's cache-busting navigation can be reverted with it — it targets a layer that was never stale.

**Interim workaround**: one hard reload (⌘⇧R) per PWA window — holds until the next docs-only deploy.

### Secondary findings (follow-ups, not this issue)

1. The card renders **"App updated"** on every server restart even when SRV == CLIENT (`updateReady() = restarted || stale`) — reproduced with identical shas on both sides. With deploys landing 3–8×/day (sometimes 92 s after the commit), consider auto-reload with a loop guard instead of a modal card.
2. A locally-booted kolu server **recycles the live per-user kaval daemon** ("boot policy = always recycle", `/run/user/<uid>/kaval` is global, not state-dir-scoped) — any dev/test boot kills production terminals.

_Investigation: forensics on the affected Chrome profile + Chromium cache-semantics research + git/header-era archaeology + live pu-box reproduction; full trail available on request. Supersedes the root-cause analysis in #1278._

srid:
**The "App updated → Reload" card could loop forever — and the culprit was kolu's own build, not the browser's cache.** After a deploy that changed only files *outside* the client build (docs, tests, the website), the client stayed pinned a build behind no matter how many times you clicked Reload. This moves build identity onto the always-fresh page shell so a normal reload always converges. Closes #1319 (root-cause investigation, evidence, and reproduction live there).

### The root cause

`koluStamped` (`default.nix`) builds the client with a placeholder commit (so docs-only commits reuse the cached derivation), then `sed`s the real sha into the built JS — i.e. it **rewrote the bytes of a content-hashed `/assets/index-*.js` file without changing its name**, and that path is served `Cache-Control: public, max-age=31536000, immutable`.

So two consecutive deploys that differ only outside the client fileset ship a bundle with the **same filename but a different embedded sha**. A returning browser holds the old-stamped copy pinned `immutable` for a year; every Reload fetches a fresh `no-store` shell that names the same filename, and the browser correctly serves the year-cached old bundle from disk — zero network bytes. `clientIsStale` compares the stamps, sees a mismatch, and re-shows the card. Forever.

```
deploy N      ─▶ /assets/index-D85Q74Rn.js  (stamped 7deb397)  ──▶ browser pins it `immutable` (1 year)
deploy N+1    ─▶ /assets/index-D85Q74Rn.js  (stamped 0fab0cc)  ── same name, never re-fetched
click Reload  ─▶ fresh no-store shell ─▶ names index-D85Q74Rn.js ─▶ served from cache (old sha) ─▶ ≠ srv ─▶ ⟲
```

### The fix — identity rides the `no-store` shell, never an `immutable` asset

`immutable` is a *contract*: a URL pinned `immutable` promises its bytes never change. A post-build stamp that rewrites a hashed file under an unchanged name breaks that contract — so the commit must live on the one resource that's never cached.

| Layer | Before | After |
| --- | --- | --- |
| Vite | `define: { __SURFACE_APP_COMMIT__ }` (into the bundle) | `transformIndexHtml` injects `window.__SURFACE_APP_COMMIT__` (head-prepend, into the shell) |
| Bun (`buildSurfaceClient`) | same define | `injectShellCommit` into the rewritten shell |
| Nix (`koluStamped`) | `sed` over `dist/**/*.js` | `sed` over `dist/index.html` only — and asserts the placeholder is **in** the shell and **not** under `dist/assets/`, failing the build loud if it ever moves back |
| Client read | `__SURFACE_APP_COMMIT__` global const | `shellCommit()` (falls back to `"dev"` = never-stale) |
| Reload | `location.replace('/?__surface_app_fresh=…')` (#1278) | plain `location.reload()` |

New pure kernels — `SHELL_COMMIT_GLOBAL`, `shellCommitScript` (JSON-encoded, `<`-escaped), `injectShellCommit` — are unit-tested, as are the plain reload and the `shellCommit()` reader. Stamp-only deploys now produce **no skew at all** (server and client report the same sha, so the card never shows); a real code deploy converges on one reload (new code ⇒ new hash ⇒ new filename ⇒ network fetch).

> **#1278's cache-busting `?__surface_app_fresh` navigation is reverted here.** It targeted a layer that was never stale, and by landing on a *different* cache key it actually skipped revalidating the bare-`/` entry it claimed to escape. A plain reload always revalidates the `no-store` shell, so it's both correct and simpler.

> **API change to `@kolu/surface-app`** (the `surfaceApp()` plugin shape, the `/client` type, `shellCommit()`), consumed by [drishti](https://github.com/srid/drishti) too — a companion drishti PR follows per `.claude/rules/surface.md`. The behavior is strictly better; drishti inherits the un-loopable reload with no wiring change.

Docs synced: the README architecture prose, the surface-app README, `docs/cache-bug.md`'s corrected root-cause chapter (the heuristic-`/`-poison theory above it is now marked refuted), and the changelog.

### Try it locally

```sh
nix run github:juspay/kolu/fix/shell-commit-stamp
```


srid:
Companion drishti PR: https://github.com/srid/drishti/pull/60

srid:
Drishti CI (srid/drishti#60): green.

Ran the canonical full pipeline via `CI=true nix run github:juspay/justci -- run` from the PR HEAD (`412ff42`). All 8 nodes succeeded on `x86_64-linux` and the posted commit statuses are green on the PR:

- `ci::_prefetch-crates` ✅
- `ci::install` ✅
- `ci::typecheck` ✅ (exercises the new `clientCommit={shellCommit()}` usage against this branch's `@kolu/surface-app`)
- `ci::fmt-check` ✅
- `ci::bun-nix-fresh` ✅
- `ci::drv-stability` ✅
- `ci::nix` ✅ (full devour-flake build — surface-app w/ shell-carried commit API builds clean)
- `ci::home-manager` ✅ (NixOS VM test boots the drishti service end-to-end)

The npins kolu pin (branch `fix/shell-commit-stamp`, rev `83df48b`) fetched without error, so CI is **not** blocked on this PR merging — `fetchGit`/npins resolves the pushed branch rev fine. `gh pr checks 60` reports all green; aggregate commit status on `412ff42` is `success`.

(The `[macos]`/`aarch64-darwin` lane is not configured on the runner host, so it drops from the fanout per justci's documented behavior and posts no check — no red gate. No fixes were needed; nothing pushed to drishti.)

srid:
## Evidence — reproduced the loop, then verified the fix kills it

End-to-end on a clean `pu` box with a fresh browser profile, driving real builds of this branch. The decisive move: build this branch at **two real stamp-only-sibling commits** (the merge HEAD and its pre-merge parent — they differ only in `nix`/`ci`/`kaval-tui`, zero client-code delta), so they're exactly the deploy pair that triggered the bug.

### The fix's core property — byte-identical bundle, commit only on the shell

```
deploy 1887868 → /assets/index-D1wTKf44.js   ┐ cmp → IDENTICAL BYTES
deploy 83df48b → /assets/index-D1wTKf44.js   ┘ (commit is NOT in the bundle)

index.html (1887868): window.__SURFACE_APP_COMMIT__="1887868"
index.html (83df48b): window.__SURFACE_APP_COMMIT__="83df48b"   ← only the no-store shell differs
```

Under the old behaviour the two bundles differed (the sha was baked in), which is what stranded returning browsers on the year-cached copy.

### Before — the loop (old build, for contrast)

With the commit baked into the `immutable` bundle, every Reload re-served the year-cached old bundle: `CLIENT` stayed pinned and the card returned. Clicks 1 and 2, never converging:

loop click 1
loop click 2

### After — one Reload converges (this PR)

Load deploy `1887868` (bundle fetched, pinned `immutable`) → deploy `83df48b` on the same port → "App updated" card → **click Reload once**:

| Signal | Result |
| --- | --- |
| `CLIENT` rail | `1887868` → **`83df48b`** (converged) |
| Card | **cleared** |
| Shell document | fetched from network (`transferSize > 0`) |
| Hashed bundle | **still served from `immutable` cache** (`deliveryType: "cache"`, 0 bytes) |

fix converged

A **second Reload** and a **fresh bare-`/` navigation** both stay at `83df48b` with the bundle still cache-served — so the convergence is durable, not a one-load crossover. The shell carries the new identity while the browser keeps honouring the `immutable` contract on the (genuinely unchanged) bundle — exactly the intended outcome.

> Companion drishti PR srid/drishti#60 — full CI green against this branch's `@kolu/surface-app`.

_Captured via chrome-devtools MCP against builds run on an ephemeral `pu` box._

srid:

✅ **Consensus** after 1 round · lowy + hickey · base `88259e6fb918`

Independent findings: lowy=4, hickey=5. The two lenses converged on the same three structural fixes (lowy-1 and hickey-1 are the same defect) and agreed the rest needed no change.

### Applied (3 commits)
- `lowy-1` / `hickey-1` — **Producer-side script encoding was duplicated.** The `window.<GLOBAL>=<escaped>` body + the `</script>`-breakout escape lived byte-identically in both `index.ts`'s `shellCommitScript()` and `vite.ts`'s inline copy, held together only by a lockstep test pinned at one benign commit. Extracted the single authoritative `shellCommitScriptBody()` kernel; `vite.test.ts` now asserts the plugin's emitted body equals it across adversarial commits (`</script>`, `<`, `"`). — `c672c1fb`
- `hickey-3` — **The global name lived in three unlinked literals** (`SHELL_COMMIT_GLOBAL` constant, `vite.ts` inline, `client.d.ts` augmentation). Pinned `client.d.ts`'s property name to the constant so all three are provably the same string. — `91d95bff`
- `hickey-5` — **The `"dev"` never-stale sentinel was a third independent copy** (`resolveCommit`, `shellCommit`, `isCleanRef`); a drift to `""`/`"unknown"` would silently make a stampless client look stale — the exact false-positive the fallback exists to prevent. Extracted `DEV_COMMIT`; `vite.ts` pins its self-contained literal in `vite.test.ts`. — `6a559e2c`

### Agreed — no change (5)
- `lowy-2` / `hickey-4` — `koluStamped`'s raw-`sed` + grep-assert on `dist/index.html` is a deliberate fail-loud defense-in-depth guard; the cross-layer coupling is acceptable given the loud assert.
- `lowy-3` — `shellCommit()` correctly encapsulates the read axis (one socket every consumer plugs into; `"dev"` fallback survives a missing stamp).
- `lowy-4` — reverting `reloadForUpdate` to a plain `location.reload()` is correct: #1278 encapsulated a volatility that didn't exist (the shell was never the stale layer).
- `hickey-2` — the two head-injection positions (`injectTo: "head-prepend"` vs splice-after-`<head>`) are acceptable; the Vite hook can't reuse `injectShellCommit` (it returns a tag descriptor, not transformed HTML).

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `88259e6fb918`

### Round 1

**codex** — approved: `false`

The implementation direction is sound: moving the volatile build stamp out of immutable assets and into the no-store shell addresses the reported cache loop, and the Vite/Bun/client call sites are mostly updated coherently. I found one edge-case correctness bug in the HTML injector and one documentation drift in the Atlas source that now contradicts the new contract. I could not run typecheck because `pnpm` is not available on PATH in this environment.

Findings:
- `F1` · minor · open — `injectShellCommit` searches with `/<head[^>]*>/i`, which also matches tags like `<header>` or `<headless>`. That defeats the helper's fail-loud contract: a malformed shell with no real `<head>` but a body `<header>` would build successfully and inject the commit script at the wrong location, potentially after the module bundle reads `shellCommit()` and causing the client to fall back to `dev` instead of reporting real skew. (packages/surface-app/src/index.ts:109)
- `F2` · minor · open — The Atlas source still documents the old build-identity model: it says the client constant is injected by a bundler define, and later snippets still use `clientCommit={__SURFACE_APP_COMMIT__}`. This branch updates README/package docs to the shell-global `shellCommit()` model, but the linked Atlas note remains current-looking and now contradicts the implementation. (docs/atlas/src/content/atlas/surface-app.mdx:56)

**claude** — Both codex findings valid; F1 fixed, F2 partial. Committed bb5e29d9.

- `F1` **fixed** — Tightened injectShellCommit regex from /<head[^>]*>/i to /<head(?:\s[^>]*)?>/i so it no longer matches <header>; added regression tests in index.test.ts (a <header>-only shell throws; <head lang=en> still matches). 118 surface-app tests pass.
- `F2` **partial** — Updated the live Atlas contract (invariant #2, #6, pieces table) in surface-app.mdx to the shell-global model (commit on window.__SURFACE_APP_COMMIT__ via shellCommit(), never a hashed-asset define) and rebuilt docs/atlas/dist. Disputed only the ~line-245 clientCommit={__SURFACE_APP_COMMIT__} snippet: it sits inside the explicitly banner-flagged '## The API (historical / merge era)' block documenting a removed API, so editing just its commit line would misrepresent design history.

commit: `bb5e29d9015d3f5ac89c32e2431ac7c18f6ce5ba`

### Round 2

**codex** — approved: `true`

Both prior findings are addressed in the current tree. F1 is fixed by the tighter `<head>` tag-boundary regex plus regression coverage for `<header>` and attribute-bearing `<head>` tags. F2 is resolved: the current Atlas contract/API-of-record now documents the shell-global `shellCommit()` model, and I accept CLAUDE's explanation that the remaining `__SURFACE_APP_COMMIT__` snippets are inside an explicitly marked historical/removed merge-era section rather than current guidance.

Findings:
- `F1` · minor · resolved — `injectShellCommit` previously matched tags like `<header>` as if they were `<head>`, allowing malformed shells to build with the commit script injected in the wrong place. (packages/surface-app/src/index.ts:108)
- `F2` · minor · resolved — The Atlas note's current contract text previously documented the old bundler-define build identity model instead of the shell-global `shellCommit()` model. (docs/atlas/src/content/atlas/surface-app.mdx:56)

_codex on the rebuttal:_ F2: Conceded. The current contract section, commit propagation section, and pieces table now describe `window.__SURFACE_APP_COMMIT__` in the no-store shell and `shellCommit()`, not a hashed-asset define. The remaining `clientCommit={__SURFACE_APP_COMMIT__}` and define wording appears below the explicit warning that the following design/API sections are historical merge-era documentation for a removed approach, so I no longer consider that current-doc drift.

srid:

Clean — no findings. Pass 1 (rules) and Pass 2 (fact-check) ran; Pass 3 (elegance) skipped (`--no-elegance`, since `/simplify` already ran this gauntlet).

| Pass | Findings |
| --- | --- |
| Rules (26 checked) | 0 |
| Fact-check | 0 |
| Elegance | skipped |

Spot-checks worth recording: `injectShellCommit()` throws loudly on a `<head>`-less shell (no silent build of a stampless shell); the `"dev"` fallback in `shellCommit()`/`resolveCommit()` is semantically distinct (`clientIsStale` treats it as never-stale, so a missing stamp can't false-positive the prompt); the Vite plugin's inline script-body copy is pinned byte-for-byte to `shellCommitScriptBody()` across hostile inputs (`</script>`, `<`, quotes); no new watchers, listeners, or growing collections.

srid:

The **x86_64-linux** lane ran on **`kolu-ci-5`** (`idliv2-02`) — commit `bb5e29d9`, **passed** ✓

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 1m54s | ✓ |
| `ci::home-manager` | 1m10s | ✓ |
| `_ci-setup` | 56s | ✓ |
| `ci::nix` | 50s | ✓ |
| `ci::smoke` | 42s | ✓ |
| `ci::unit` | 28s | ✓ |
| `ci::fmt` | 15s | ✓ |
| `ci::biome` | 14s | ✓ |
| `ci::install` | 10s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/bb5e29d/timings.jsonl`; pool state is a live `flock` probe.</sub>

srid:
### Re-verified on final HEAD (post-review)

Re-ran the pu-box repro against `bb5e29d9` (after the lens + codex review commits touched `surface-app`), to confirm the fix still holds:

- **Byte-identical bundle** — two stamp-sibling builds (`aaa1111` / `bbb2222`) both emit `index-DH-81df0.js`; `cmp` → identical bytes; the stamp appears **0×** inside the bundle and only on the `no-store` shell.
- **One Reload converges** — load `aaa1111` (bundle pinned `immutable`) → deploy-swap to `bbb2222` → card appears → one plain `location.reload()` → CLIENT `aaa1111` → **`bbb2222`**, card cleared, bundle still served from cache (`transferSize: 0`, `deliveryType: "cache"`).
- **Durable** — a fresh bare-`/` tab loads at `bbb2222` with no card, bundle still cache-served.

CI is green on both platforms; the companion drishti PR (srid/drishti#60) is green against this branch.

PR #1331, opened 2026-06-13
Code-tab scope picker: segmented control with live change badges

srid:
**The Code tab's "All files / Local / Branch" picker is no longer a dropdown.** The old chip + popover hid every option behind a click and — being collapsed — could never tell you a mode had changes without opening it. It's now a **segmented control**: all three views sit side by side, switching is a single click, and the two git-diff segments carry a **live change-count badge**. A non-empty mode advertises itself — *you can see "3 uncommitted, 12 vs base" without switching into either view*, the one affordance a collapsed dropdown structurally couldn't offer.

The switcher reuses the project's existing `ui/SegmentedControl` (the same primitive the Settings popover uses for its enums), extended with optional per-option `icon` / `hint` / `badge` / `dividerBefore` fields and a `toolbar` ARIA variant — all inert when unset, so the settings call sites render byte-identically.

### The badge needed git status in every view — without poisoning Branch

A badge needs the Local **and** Branch counts even when you're not in that view. Two always-on passive subscriptions (`localStatus` / `branchStatus`) feed the badges, the branch base/ref, and the browse-tree decoration. The *active* diff view keeps its **own** subscription keyed on the active mode:

```
localStatus  (always-on, passive) ─┐
branchStatus (always-on, passive) ─┼─▶ badges · base/ref · browse overlay
                                    │
activeStatus (keyed on diffMode)  ─┴─▶ the active diff view's file list
```

> _Why the active view doesn't just reuse the passive subscription:_ a passive Branch read can fail with `BASE_BRANCH_NOT_FOUND` (origin not fetched) *before* you ever open Branch view. `createReactiveSubscription` doesn't retry on error, and the server tears the stream down before installing its repo-change watcher — so a later `git fetch` could never revive it. Reusing that subscription would leave Branch mode stuck on a stale error. The dedicated view-keyed `activeStatus` always does a fresh read and surfaces every error (the un-fetched-base case becomes the actionable "run `git fetch`"); the passive `branchStatus` swallows *only* that one expected `PRECONDITION_FAILED` and toasts everything else.

### Details worth a reviewer's eye

- **`role="toolbar"` + `aria-pressed` toggle buttons** — the conventional segmented-control ARIA (same pattern as a text-align button group); biome's `useSemanticElements` rejects `role="group"`/`role="radio"` on a `<div>`/`<button>`.
- **Browse is set apart** from the two git views by a divider and is *never* badged (it's a whole-repo tree, not a diff).
- **Hints became tooltips** — each view's old inline description now rides the segment's `title`.
- **No-branch-base (#1244) handling preserved** — `branchBase()` reads the always-on branch stream so the segment is correct in any view, but the "No branch base to compare" *empty-state copy* still gates on the active view.

### Tests & docs

- Mode-switch step defs simplified (one click, no popover); the mode assertion reads the control's `data-mode`.
- New e2e scenario proves the **ambient** case: in the default browse view, the Local segment shows a count of `2` for two uncommitted files, and browse itself shows none.
- Changelog entry under a new `### Changed` heading.

### Shaped by review

This PR was taken through a serial review gauntlet (see the PR comments). The **lens debate** caught that the new control duplicated the existing `SegmentedControl` and consolidated it; the **codex debate** caught the Branch-subscription poisoning bug above and split the active/passive sources. Both reached consensus.


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `14aa6d41396e`

### Applied (7)
- `lowy-1` Duplicated receptacle: ScopeSegments re-implements the existing SegmentedControl — commit `15193e0ff`
- `lowy-2` `group` typed as free `string` leaks the divider-placement rule to the host — commit `15193e0ff`
- `hickey-1` ScopeSegments reinvents the canonical SegmentedControl instead of extending it — commit `15193e0ff`
- `hickey-2` Canonical view order is fragmented across two files held together by a doc comment — commit `b6369ea5a`
- `hickey-3` Segment count is a tri-state (number / 0 / undefined) splitting one fact across value + presence — commit `2d9303570`
- `hickey-4` branchBase() is read in two views with the meaning gated by a separate view() check at each use site — commit `56b69ff8d`
- `hickey-5` ScopeSegment.group complects 'divider placement' with an opaque catalog string — commit `15193e0ff`

### Agreed — no change (4)
- `lowy-3` branchStatus.onError re-implements view-gated error routing the old code located at the read site (packages/client/src/right-panel/CodeTab.tsx:286-298)
- `lowy-4` `status`/`statusPending`/`statusError` accessor trio is a clean facade over the consolidated subscriptions (packages/client/src/right-panel/CodeTab.tsx:304-313)
- `hickey-6` status / statusPending / statusError re-fragment one subscription's surface into three accessors (packages/client/src/right-panel/CodeTab.tsx:304-313)
- `hickey-7` Error-toast policy is split between two onError handlers coupled by an implicit 'exactly one toasts' rule (packages/client/src/right-panel/CodeTab.tsx:275-298)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `14aa6d41396e`

### Round 1

**codex** — approved: `false`

The segmented-control UI refactor is mostly coherent, and stale selectors/imports look cleaned up. The main risk is the new always-on Branch status stream: it is used both as a passive badge/decorator source and as the active Branch view source, but passive failures can poison it and errors are conditionally swallowed. `git diff --check` is clean; `git status --short` showed no untracked files.

Findings:
- `F1` · major · open — `branchStatus` now subscribes for every repo regardless of the active view, and the active Branch view later reuses that same accessor. If the passive Branch read fails before the user opens Branch mode, for example `BASE_BRANCH_NOT_FOUND`, `createReactiveSubscription` records the error and the input never changes when the user clicks Branch. Because the initial `pollOnEvent` read failed before installing repo-change watchers, a later `git fetch` also won't revive it. This regresses the old design where passive browse decoration and explicit Branch mode had separate subscriptions, so entering Branch got a fresh read. (packages/client/src/right-panel/CodeTab.tsx:289)
- `F2` · major · open — The `branchStatus` `onError` handler suppresses every Branch status error unless `view() === "branch"`. The comment justifies suppressing the expected missing-base case for passive badges, but the code also hides unexpected `GIT_FAILED`, permission, transport, and other application errors while Browse/Local is active. That leaves the Branch badge/decorations silently absent or stale with no user-visible failure. (packages/client/src/right-panel/CodeTab.tsx:295)

**claude** — Both of CODEX's major findings were valid and I fixed both in packages/client/src/right-panel/CodeTab.tsx. The root cause is that the refactor merged the passive Branch badge/overlay source and the active Branch view source into a single `branchStatus` subscription keyed only on `repoPath()`. I verified against the runtime that this is a real regression: `createReactiveSubscription` (packages/surface/src/solid/createReactiveSubscription.ts) only re-reads when its input changes and does not retry on error, and the server's `pollOnEvent` (packages/surface/src/server.ts:335) throws on a failed initial read BEFORE installing the repo-change watcher — so a passive un-fetched-base failure poisons the shared subscription, clicking Branch never refreshes, and a later `git fetch` cannot revive it. I restored the pre-refactor active/passive split: the active diff view now has its own subscription keyed on `diffMode()` (fresh read on entering a mode, revives after fetch, surfaces every error), while the always-on passive `branchStatus` continues to feed the Local/Branch badges, base/ref, and browse overlay — and its onError now swallows ONLY the expected `ORPCError`/`PRECONDITION_FAILED` (BASE_BRANCH_NOT_FOUND) and toasts all other faults. Verified: client tsc --noEmit clean, biome format/lint clean, all 292 client unit tests pass.

- `F1` **fixed** — Agreed — confirmed the regression. Old code keyed the active `status` subscription on `diffMode()` (a separate subscription from the passive browse overlay `browseBranchStatus`), so entering Branch always did a fresh read. The new code reused one `repoPath()`-keyed `branchStatus` for both, so a passive failure stuck. Verified the revival problem: createReactiveSubscription.ts re-reads only on input change (not on error), and pollOnEvent in packages/surface/src/server.ts:335 does `await opts.read()` before line 337's yield and before repoEventStream installs the watcher — a first-read throw kills the stream with no watcher, so a later `git fetch` event has nothing listening. Fix: added a dedicated active-view subscription `activeStatus` keyed on `diffMode()` (CodeTab.tsx ~321), and pointed `status`/`statusPending`/`statusError` at it. `branchStatus` stays always-on but is now passive-only (badge/count/base/ref/browse overlay).
- `F2` **fixed** — Agreed — the old `onError` `if (view() === "branch") toast` suppressed GIT_FAILED/permission/transport on the passive source whenever Browse/Local was active. Replaced with `if (isUnfetchedBase(err)) return; toast.error(...)`, where `isUnfetchedBase` checks `err instanceof ORPCError && err.code === "PRECONDITION_FAILED"` — exactly CODEX's suggested predicate, mirroring the established pattern in packages/client/src/terminal/useTerminals.ts:75. BASE_BRANCH_NOT_FOUND maps to PRECONDITION_FAILED in packages/server/src/unwrapGit.ts:18. So the passive source now swallows only the expected un-fetched-base case and toasts every other fault; the new active subscription surfaces ALL errors (including the actionable un-fetched-base 'run git fetch').

commit: `9db5fefc96936966c5947d0edc2eff9e099f0c48`

### Round 2

**codex** — approved: `true`

CLAUDE's follow-up addresses both previously raised issues in `CodeTab.tsx`. The active diff view now uses its own mode-keyed `gitStatus` subscription, so entering Branch performs a fresh read and no longer inherits a passive badge/decorator failure. The passive Branch subscription now suppresses only the expected unfetched-base `ORPCError`/`PRECONDITION_FAILED` path and surfaces all other failures. I did not find a new regression introduced by this round's fix.

Findings:
- `F1` · major · resolved — Previously, the active Branch view reused the always-on passive `branchStatus` subscription, so a passive first-read failure could poison Branch mode and never revive on later repo changes. (packages/client/src/right-panel/CodeTab.tsx:321)
- `F2` · major · resolved — Previously, passive `branchStatus` suppressed every error outside Branch view, hiding unexpected git/permission/transport failures. (packages/client/src/right-panel/CodeTab.tsx:306)

_codex on the rebuttal:_ CLAUDE did not dispute either finding. F1 is fixed: the active file-list/error path reads from the new view-keyed `activeStatus`, so a stale passive `branchStatus` error cannot block Branch mode. F2 is fixed: the passive handler now swallows only the expected unfetched-base ORPC error and surfaces other failures.

srid:

**All clear** — both passes converge, no fixes applied.

| Pass | Result |
|---|---|
| Rules | 1 borderline, adjudicated clean |
| Fact-check | 0 findings |
| Elegance | skipped (`/simplify` already ran this gauntlet) |

The one borderline flag was `branchStatus` swallowing `PRECONDITION_FAILED` (`no-silent-error-swallowing` / `caught-error-must-not-collapse-to-empty`). Adjudicated **clean**: this is the documented active/passive split codex introduced — the *passive* badge/overlay subscription treats an un-fetched base as a legitimate empty (no count, overlay falls back to the local layer), while the *active* view subscription (`activeStatus`, keyed on `diffMode()`) surfaces the same error to the user the moment they enter Branch view, where "run `git fetch`" is actionable. The error is never invisible-forever — it's deferred to the point it can be acted on.

Fact-check separately confirmed: badge logic correct (`badge: 0` hides, omitted when no base), subscription inputs are reactive thunks, e2e steps poll async state. No logic errors, silent swallows, or unjustified fallbacks.

srid:

The **x86_64-linux** lane ran on **`kolu-ci-4`** (`dev-x86-64-linux-08`) — commit `9db5fefc`, _verdict unknown (no per-node timing)_

_(no per-recipe timing in `.ci/9db5fef/timings.jsonl`)_

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/9db5fef/timings.jsonl`; pool state is a live `flock` probe.</sub>

srid:
## Evidence

The Code tab's old "All files / Local / Branch" dropdown is now a **segmented control** — all three views sit side by side, and the two git-diff segments carry a live **change-count badge**. Captured against this PR's build on an ephemeral box, in a repo with two uncommitted files (`a.txt`, `b.txt`).

**The headline — a mode advertises its change count without you switching into it.** In the default *All files* (browse) view the **Local** segment already shows a **`2`** badge; one click switches into Local and the badge persists (top → bottom):

Segmented control before/after: All files active with Local 2 badge, then Local active with the 2 badge still shown

In context — the segmented control over the changed-file tree (`a.txt`, `b.txt`, both untracked `U`):

Code tab panel showing the segmented control toolbar above the a.txt/b.txt tree

<details><summary>Full app view (terminal in the repo + Code tab)</summary>

Full kolu window: terminal in /tmp/badge-repo on the left, Code tab with the segmented control and Local 2 badge on the right

</details>

<sub>The earlier e2e-recorded artifact was illegible (the Code-tab panel floats narrow in the canvas layout) and its links were broken; replaced with these chrome-devtools captures of the live app.</sub>


---

### Touch / mobile (follow-up commit)

The segments were 20px tall — below the WCAG 2.2 24px tap floor — with no coarse-pointer adaptation. On a **coarse pointer** they now grow to **28px** (the toolbar row, the back/forward buttons, and the filter's clear button all clear 24px too). Captured with chrome-devtools **touch emulation** (`(pointer: coarse)` actually true — a plain width resize wouldn't trigger `isTouch()`), same scale, **desktop (20px) above, touch (28px) below**:

Desktop vs touch: the segmented control is compact at 20px on a fine pointer, and noticeably taller at 28px on a coarse pointer

Measured live under `(pointer: coarse)` emulation: every target (`All files` / `Local 2` / `Branch` segments + back/forward) reports **28px** (`clears24: true`); under a fine pointer they report 20px and `data-touch` is omitted.


---

### iPhone 13 (390×844, real mobile drawer)

Verified at an actual **iPhone 13 portrait** profile (390×844, DPR 3, `mobile,touch`), not just a coarse pointer at desktop width. On mobile the right panel is a **bottom drawer**, and the Code tab + segmented control render inside it. Live measurements at this profile:

| Check | Result |
|---|---|
| `(pointer: coarse)` | `true` |
| viewport | 390px |
| segment / nav button height | **28px** (`allTargets24: true`) |
| `Local` badge | `2` |
| toolbar overflow at 390px | **none** — `scrollWidth 390 == clientWidth 390` (segments + filter fit; the `overflow-x-auto` safety net isn't even needed here) |

**Still** — the mobile drawer with the touch-sized segmented control + `Local 2` badge over the `a.txt`/`b.txt` tree:

iPhone 13 portrait: kolu's bottom drawer open on the Code tab, segmented control All files / Local 2 / Branch fitting the 390px width

**Video** — tapping through the control on a touch screen: All files → **Local** (badge persists) → open `a.txt`'s diff → back to All files:

iPhone 13 touch flow: tapping the segmented control to switch scopes in the mobile drawer

▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=evidence-code-scope-iphone13.mp4

srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `497d6f7f4a7d`

### Applied (5)
- `lowy-1` Touch-sizing policy sprayed as raw literals instead of behind a named density token — commit `677ca3bd4`
- `lowy-3` Modality wire still exposed at the back/fwd call sites (vs. the receptacle the SegmentedControl now offers) — (uncommitted)
- `hickey-1` Touch-density decision fragmented across 4 read sites in one toolbar — (uncommitted)
- `hickey-2` Back/forward buttons are byte-identical twins; the diff doubles their duplicated class string — commit `d3a5693ad`
- `hickey-3` `data-touch="false"` emitted on desktop instead of omitting the attribute — commit `940747542`

### Agreed — no change (2)
- `lowy-2` Symmetry break: third toolbar child (FileSearchInput clear button) does not grow on touch (packages/client/src/right-panel/FileSearchInput.tsx:27-35 (consumed at CodeTab.tsx:786))
- `hickey-4` Magic touch-size values (h-7/h-10/px-2.5/text-[11px]) scattered without a named density scale (packages/client/src/ui/SegmentedControl.tsx:91,110; packages/client/src/right-panel/CodeTab.tsx:748,759,771)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `497d6f7f4a7d`

### Round 1

**codex** — approved: `false`

The branch is close, but it still has touch/mobile verification and implementation gaps. The runtime UI mostly moves the obvious Code-tab controls above 24px, but the new workflow doc can still produce false mobile/touch evidence, the touch modality is snapshotted for controls that should stay reactive, and the toolbar still contains a visible sub-24px clear button.

Findings:
- `F1` · major · open — The new evidence rule conflates small viewport/mobile drawer coverage with coarse-pointer coverage. It says a phone-width `resize_page` or the mobile drawer is enough to verify mobile/touch rendering, but the actual UI sizing added in this branch is driven by `isTouch()` / `(pointer: coarse)`, not only by viewport width. A desktop browser resized to 390x844 can exercise the bottom-drawer layout while leaving `isTouch()` false, so the exact 20px segmented-control regression this paragraph mentions would still be missed. (.agency/do.md:118)
- `F2` · minor · open — `const touch = isTouch()` snapshots the coarse-pointer media query at CodeTab mount, then reuses that boolean for the toolbar and segmented-control attributes. That is correct for `treeDensity` because Pierre only reads density at construction, but the new DOM sizing does not have that limitation. If pointer modality changes while CodeTab remains mounted, these controls stay in the old density even though `createMediaQuery` is reactive. (packages/client/src/right-panel/CodeTab.tsx:143)
- `F3` · minor · open — The touch-sized toolbar still includes `FileSearchInput`, whose clear button remains `p-0.5` around a `w-3 h-3` icon in `FileSearchInput.tsx`, giving roughly a 16px target. The new comments and docs say this visible touch surface clears the WCAG 2.2 24px floor, but the clear-filter affordance does not when a query is present. (packages/client/src/right-panel/CodeTab.tsx:802)

**claude** — Addressed all three CODEX findings; agreed with each. F1: rewrote the .agency/do.md evidence rule to split mobile coverage into two orthogonal checks (small-viewport/drawer layout vs. actual coarse-pointer emulation), since a width-only resize leaves (pointer: coarse) false and misses touch-target regressions. F2: stopped snapshotting isTouch() into a const for the reactive DOM sizing — only Pierre's tree density snapshots it now (Pierre reads density once at construction), while data-touch, SegmentedControl touch, and FileSearchInput touch read isTouch() reactively in JSX so they follow a mid-mount pointer-modality flip. F3: propagated a touch prop into FileSearchInput and grew its clear-filter button from ~16px (p-0.5 + w-3 h-3) to h-6 w-6 (24px) on coarse pointers, clearing the WCAG 2.2 floor. Ran just fmt (clean) and tsc --noEmit (no errors in touched files). SegmentedControl.tsx needed no change this round — its touch prop is already reactive (reads props.touch), now fed isTouch() reactively from CodeTab.

- `F1` **fixed** — Agreed. A desktop browser resized to 390x844 exercises the drawer layout but leaves (pointer: coarse) false, so it cannot verify isTouch()-gated sizing. Rewrote the 'Cover mobile/touch' paragraph in .agency/do.md to call out the two orthogonal axes (isMobile size vs. isTouch modality, per useMobile.ts) and split the requirement into two bullets: small-viewport/drawer layout (width resize OK), and coarse-pointer hit targets (requires a Playwright mobile context/device profile with hasTouch, or asserting matchMedia('(pointer: coarse)').matches, then verifying each target measures >=24px). Also noted the cited 20px regression would be missed by a width-only resize just the same.
- `F2` **fixed** — Agreed. CodeTab.tsx:143 did `const touch = isTouch()`, then reused that boolean for data-touch and SegmentedControl touch — correct for Pierre density (Pierre reads density once at construction) but wrong for the reactive DOM, which won't follow a mid-mount modality flip (e.g. a 2-in-1 docking). Removed the const; treeDensity now reads `isTouch() ? 'relaxed' : undefined` (snapshot at mount, as Pierre needs), and the toolbar `data-touch={isTouch() || undefined}`, `SegmentedControl touch={isTouch()}`, and `FileSearchInput touch={isTouch()}` all read reactively in JSX. The NavButton already inherits via the group-data-[touch=true]/toolbar variant, which reads the reactive parent attribute at the DOM level.
- `F3` **fixed** — Agreed. FileSearchInput's clear button was `p-0.5 -mr-0.5` around a `w-3 h-3` CloseIcon (~16px target), below the 24px floor whenever a query is present, even inside the touch-sized toolbar. Added a `touch?: boolean` prop (fed isTouch() from the CodeTab toolbar) and changed the button to `grid place-items-center ... data-[touch=true]:p-0 data-[touch=true]:h-6 data-[touch=true]:w-6` so it becomes a 24px tap target on coarse pointers, matching the rest of the touch toolbar.

commit: `173bd9cd6cd9cadab369f7b0796eea18eeabf7fd`

### Round 2

**codex** — approved: `true`

Verified the current tree against the three existing findings only. CLAUDE's changes address the evidence-doc ambiguity, make the DOM touch sizing reactive while keeping Pierre density intentionally snapshotted, and grow the Code-tab filter clear button to a 24px coarse-pointer target. No new regression was introduced by this round's fixes.

Findings:
- `F1` · major · resolved — The previous evidence rule conflated small viewport/mobile drawer coverage with coarse-pointer coverage, allowing a width-only resize to miss `isTouch()`-gated target sizing regressions. (.agency/do.md:118)
- `F2` · minor · resolved — The previous `const touch = isTouch()` snapshotted coarse-pointer state for DOM attributes that should react to modality changes while the component remains mounted. (packages/client/src/right-panel/CodeTab.tsx:144)
- `F3` · minor · resolved — The Code-tab toolbar's clear-filter button previously remained a roughly 16px target on coarse pointers, below the 24px floor claimed for the touch-sized toolbar. (packages/client/src/right-panel/FileSearchInput.tsx:33)

_codex on the rebuttal:_ No disputes to adjudicate; CLAUDE agreed with all three findings. I verified each fix in the current diff and accept the resolutions.

srid:
## [👮 Code-police](https://agency.srid.ca/) — touch follow-up

**All clear** — both passes clean on the touch-friendliness delta (`497d6f7f..HEAD`), no fixes applied.

| Pass | Result |
|---|---|
| Rules | 0 findings (all 26 checked) |
| Fact-check | 0 findings |
| Elegance | skipped (`/simplify` ran this gauntlet) |

Notable confirms: `NavButton` is real composition (derives direction, conditional aria-label/title, icon rotation), not a thin wrapper; `isTouch()` is called reactively when forwarding to props (correct SolidJS); `data-touch={… || undefined}` correctly gates the `data-[touch=true]` selector. `/simplify` applied nothing — its one suggestion (snapshot `isTouch()` once) would have reverted codex's deliberate reactivity fix (F2), so it was correctly skipped.

srid:

The **x86_64-linux** lane ran on **`kolu-ci-4`** (`dev-x86-64-linux-08`) — commit `173bd9cd`, **passed** ✓

- **Lane wall** (pipeline): **2m36s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 1m39s | ✓ |
| `ci::home-manager` | 1m8s | ✓ |
| `ci::smoke` | 39s | ✓ |
| `ci::atlas-sync` | 38s | ✓ |
| `_ci-setup` | 36s | ✓ |
| `ci::nix` | 36s | ✓ |
| `ci::flake-check` | 35s | ✓ |
| `ci::unit` | 27s | ✓ |
| `ci::install` | 19s | ✓ |
| `ci::biome` | 14s | ✓ |
| `ci::fmt` | 13s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|
| `kolu-ci-4` | dev-x86-64-linux-08 | 🔒 leased |

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/173bd9c/timings.jsonl`; pool state is a live `flock` probe.</sub>

Issue #32, opened 2026-03-22
Production should use unique port for application


srid:
`nix run` should not use development ports. The home-manager module should use the same uniquely assigned port (as `nix run`) as well.

This way, we can develop on the same machine running the production instance of kolu (via nix run or home-manager module)

PR #1333, opened 2026-06-13
fix(surface): stop the coordinator crashing with write EPIPE on stdio teardown (odu#32)

srid:
Fixes the root cause of [juspay/odu#32](https://github.com/juspay/odu/issues/32) — a residual of #25.

After a clean run settles and every GitHub status is posted, odu's coordinator throws an **unhandled `write EPIPE`** during lane teardown and exits 1 — the `── ci run summary ──` never prints. The statuses are green, so `gh pr checks` is fine, but a caller keying on odu's **exit code** (a `/do` loop, a CI wrapper, an `&&` chain) sees the whole run as failed.

## Root cause

On teardown the coordinator writes a framed message to a lane peer (`odu-runner` over ssh stdio) whose pipe has already closed. There are **two** independent failure channels for that write, and #25 only closed one:

| channel | before this PR |
| --- | --- |
| the stream's `'error'` event | **handled** — `write.on("error", …)` logs `outbound write error: EPIPE` and tears the link down |
| the **send-function promise** (`writeFramedMessage` → `reject(err)`) | **escaped** — the rejection surfaced through the peer's internal send path as an unhandled rejection, rethrown via `process.nextTick(() => { throw err })` → crash |

So the link logged the EPIPE *and* crashed on it.

## The fix

A dead-pipe write during teardown is benign: the peer is gone, there's nothing to deliver. Add a single shared send wrapper both peers use:

```ts
framedSend(write, message) = writeFramedMessage(write, message)
  .catch(err => { if (!isBenignWriteError(err)) throw err })   // EPIPE / ERR_STREAM_DESTROYED → resolve
```

`ClientPeer` (`links/stdio.ts`) and `ServerPeer` (`peer-server.ts`) both send through it — so the symmetric crash on `serveOverUnixSocket`'s per-connection serves (kaval) is closed too. This **preserves the framing/lifecycle split** the codec is deliberate about: `writeFramedMessage` stays framing-only, and the per-side teardown *response* (client closes its link; server ends its serve loop) stays in each peer's stream-`'error'` handler. `framedSend` only neutralizes the un-deliverable frame whose rejection currently escapes — every non-dead-pipe write error still propagates.

## Scope & follow-up

- **Internal change** — `framedSend`/`isBenignWriteError` live in the package-internal `stdio-codec.ts` (not re-exported on any `@kolu/surface/*` subpath), no contract/API delta. So: no drishti PR (the surface rule only triggers on API-facing changes; drishti — itself an affected odu consumer — picks the fix up via a pin bump), and no changelog (the user-visible manifestation is odu's exit code, not a kolu-product change).
- **Follow-up:** once this merges, odu bumps its kolu pin to pick up the fix.

## Tests

`packages/surface/src/links/stdio-codec.test.ts` (new): `framedSend` resolves on a write that dies with EPIPE / ERR_STREAM_DESTROYED (the #32 race), still rejects a non-benign write error, and writes the frame unchanged on a healthy stream. The existing #25 stream-`'error'` guard test (`stdio.test.ts`) stays green. Full surface suite: 104 unit tests pass; `tsc` + `biome` clean.

### DIFF
--- a/packages/surface/src/links/stdio-codec.test.ts
+++ b/packages/surface/src/links/stdio-codec.test.ts
@@ -1,0 +1,83 @@
+/**
+ * The send half of the stdio codec: `framedSend` must treat a write whose
+ * destination is already gone (a dead ssh pipe → EPIPE, or our own stream
+ * destroyed → ERR_STREAM_DESTROYED) as benign during teardown, resolving
+ * rather than rejecting. A rejection here escapes the peer's internal send
+ * path as an unhandled rejection and crashes the coordinator on lane teardown
+ * after a green run (juspay/odu#32, the residual of #25). Any other write
+ * error is real and must still propagate.
+ */
+
+import { Writable } from "node:stream";
+import { describe, expect, it } from "vitest";
+import { framedSend, isBenignWriteError } from "./stdio-codec";
+
+/** A Writable whose write callback fails with a chosen error code, the way a
+ *  dead pipe (EPIPE) or a destroyed stream (ERR_STREAM_DESTROYED) does. It
+ *  carries a no-op `'error'` listener so the failing write doesn't ALSO crash
+ *  via the unrelated stream-'error' path (the real link/peer attach exactly
+ *  such a lifecycle guard) — this isolates the send-promise path under test. */
+function failingWrite(code: string): Writable {
+  const w = new Writable({
+    write(_chunk, _enc, cb) {
+      cb(Object.assign(new Error(`write ${code}`), { code }));
+    },
+  });
+  w.on("error", () => {});
+  return w;
+}
+
+describe("isBenignWriteError", () => {
+  it("treats a dead-pipe write (EPIPE / ERR_STREAM_DESTROYED) as benign", () => {
+    expect(
+      isBenignWriteError(Object.assign(new Error(), { code: "EPIPE" })),
+    ).toBe(true);
+    expect(
+      isBenignWriteError(
+        Object.assign(new Error(), { code: "ERR_STREAM_DESTROYED" }),
+      ),
+    ).toBe(true);
+  });
+
+  it("does not swallow other write errors", () => {
+    expect(
+      isBenignWriteError(Object.assign(new Error(), { code: "ENOSPC" })),
+    ).toBe(false);
+    expect(isBenignWriteError(new Error("boom"))).toBe(false);
+    expect(isBenignWriteError(undefined)).toBe(false);
+    expect(isBenignWriteError(null)).toBe(false);
+  });
+});
+
+describe("framedSend", () => {
+  it("resolves when the write dies with EPIPE — the #32 teardown race", async () => {
+    await expect(
+      framedSend(failingWrite("EPIPE"), "hello"),
+    ).resolves.toBeUndefined();
+  });
+
+  it("resolves when the write dies with ERR_STREAM_DESTROYED", async () => {
+    await expect(
+      framedSend(failingWrite("ERR_STREAM_DESTROYED"), "bye"),
+    ).resolves.toBeUndefined();
+  });
+
+  it("still rejects a non-benign write error (it's a real failure)", async () => {
+    await expect(framedSend(failingWrite("ENOSPC"), "x")).rejects.toThrow(
+      /ENOSPC/,
+    );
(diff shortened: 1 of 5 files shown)


srid:
## Review gauntlet

Ran the structural lenses + codex + a correctness pass on the diff.

**lowy** — *structure sound, no actions.* The benign-write policy (shared by both consumers) lives in the one module both peers import, while the per-side teardown *response* stays in each peer's `write.on('error')` handler — the right split. Confirmed all four transport send-paths (stdio + unix-socket, client + server) inherit the fix via `framedSend`; a future WebSocket transport correctly wouldn't (it has its own framing).

**codex** — *no findings.* "Stops the unhandled-rejection path for teardown writes while preserving rejection for other write failures and leaving the stream-`'error'` lifecycle handling intact."

**correctness** — *no bugs.* Traced the crash through oRPC: `framedSend` is the single chokepoint covering the **fire-and-forget** sends (abort listener, event-iterator cleanup) that a message-level `.catch` doesn't, and it doesn't mask real failures (the stream-`'error'` handler still drives teardown; the two channels are complementary). `isBenignWriteError` handles `null`/`undefined`/no-`.code`; EPIPE/ERR_STREAM_DESTROYED carry `.code` top-level (not on `.cause`) on Node 24.

**Fixes applied** (d9fe989):
- *hickey MAJOR* — the unit test proved `framedSend`'s surface but not the end-to-end no-escape. Added a peer-integration regression test: a fire-and-forget abort send to a dead pipe must not escape as an uncaught exception. Verified both ways — strip `framedSend`'s `.catch` and the run errors with the exact teardown `write EPIPE`; with the fix it's clean.
- *hickey MINOR* — `writeFramedMessage` is now imported nowhere (only `framedSend`, same file), so un-exported it — a future caller can't bypass `framedSend` and reintroduce the rejection.

Kept `isBenignWriteError` as a named, tested policy (the natural home if the benign-code set ever grows, e.g. `ECONNRESET`) — a considered disagreement with hickey's inline suggestion.

After fixes: `tsc` clean, **105 surface unit tests** pass.

srid:
## Evidence

This fixes a coordinator **teardown crash** with no kolu on-screen surface (the bug manifests as odu exiting 1 after a green run), so the proof is behavioral, not visual: the new regression test demonstrates the fixed behavior, verified **both directions** by stripping `framedSend`'s `.catch`.

**Without the fix** — the fire-and-forget abort send to a dead pipe escapes as an uncaught exception (the #32 crash), failing the run:

```
[@kolu/surface/links/stdio] outbound write error: write EPIPE
⎯⎯⎯ Unhandled Errors ⎯⎯⎯
Vitest caught 1 unhandled error during the test run.
  Error: write EPIPE   { code: 'EPIPE' }
The latest test that might've caused the error is
  "does not crash when a fire-and-forget teardown send hits a dead pipe (#32)"
```

**With the fix** — the dead-pipe write resolves, nothing escapes:

```
✓ does not crash when a fire-and-forget teardown send hits a dead pipe (#32)
 Test Files  13 passed (13)
      Tests  105 passed (105)
```

The reproduction is the real wiring: a `stdioLink` whose write flushes the request frame then fails the abort frame with EPIPE — exactly the teardown race from the issue. (Full CI — both platforms, strict — running on the PR validates the surface suite end-to-end via `ci::unit`.)

srid:

The **x86_64-linux** lane ran on **`kolu-ci-6`** (`dev-x86-64-linux-03`) — commit `d9fe989e`, **passed** ✓

- **Lane wall** (pipeline): **2m33s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 1m50s | ✓ |
| `ci::home-manager` | 1m11s | ✓ |
| `ci::nix` | 50s | ✓ |
| `ci::smoke` | 42s | ✓ |
| `_ci-setup` | 32s | ✓ |
| `ci::unit` | 27s | ✓ |
| `ci::fmt` | 14s | ✓ |
| `ci::biome` | 13s | ✓ |
| `ci::install` | 10s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/d9fe989/timings.jsonl`; pool state is a live `flock` probe.</sub>

PR #1346, opened 2026-06-14
feat(codex-debate): implementor inherits main-agent context + rationale

srid:
## What

In `/codex-debate`'s review mode, the Claude **implementor** (the "author" side that fixes/disputes codex's findings) was a fresh one-shot `agent()` subagent. It reconstructed the change from `git diff` alone:

- it inherited **none** of the main agent's task/intent, and
- unlike the lens step, it never received the change **rationale**.

So each round it re-derived intent from scratch and re-litigated *deliberate* choices that codex — correctly, reviewing a bare diff — flagged as findings. Noise the implementor then had to spend rounds disputing.

This threads two optional notes from the caller into the debate.

## The two notes

| Note | Goes to | When | Why |
|---|---|---|---|
| **`context`** (#1) | implementor only | **every round** | The main-agent context — what the change is *for*. `agent()` can't be resumed the way codex is, so re-injecting each round is how the implementor "inherits" at all. Kept off codex so it stays an independent reviewer of the code, not the author's narrative. |
| **`rationale`** (#2) | codex (round-1) + implementor | once / every round | The deliberate-decisions note (mirrors `/lens-debate`). codex sees it in its round-1 prompt — its warm session carries it forward — so it doesn't raise intentional choices *at the source*; the implementor sees it so it *disputes* rather than "fixes" a finding that contradicts a deliberate choice. |

## How

- `debate.workflow.js` — reads `args.context` / `args.rationale`; injects both into the implementor prompt every round; writes `rationale` to a file once before the loop and passes it to `codex-review.sh`.
- `codex-review.sh` — new optional `[rationale-file|-]` positional; injects a deliberate-decisions block into the **cold** review prompt (round 1 / degraded fallback) only — the warm session retains it after.
- `/be-review` (codex step) and `/be` §4 — thread `context` + `rationale` straight through.

Empty notes render byte-identical to the old prompts, so a contextless invocation is unchanged.

## Verification

- `node --check` (async-wrapped) parses the edited workflow; matches the original's parse profile.
- `bash -n codex-review.sh` clean.
- `just ai::apm` regenerated `.claude/` + `.agents/` (no `.codex`/`.opencode` variants — codex-debate needs the Workflow tool); `apm.lock.yaml` carries the new hashes.
- `just fmt` — no changes.

Sources live under `.apm/`; the `.claude/` + `.agents/` copies are generated.

### DIFF
--- a/.agents/skills/be-review/SKILL.md
+++ b/.agents/skills/be-review/SKILL.md
@@ -1,7 +1,7 @@
 ---
 name: be-review
 description: Run /be's review gauntlet SERIALLY — /lens-debate (lowy ⇄ hickey), then /codex-debate, then /simplify, then code-police, each editing and committing on the live branch in turn. Use from /be §4, or when the user asks to "run the review gauntlet". Requires Claude Code's Skill tool.
-argument-hint: "[--base <branch>] [--rationale <note>] [--tracks lens,codex,simplify,police]"
+argument-hint: "[--base <branch>] [--rationale <note>] [--context <note>] [--tracks lens,codex,simplify,police]"
 ---
 
 # Review gauntlet (serial)
@@ -92,7 +92,10 @@ it once per step.
 
 2. **codex** — follow `/codex-debate` (Skill tool). `repoPath` = the live
    worktree, `base` = `MB`, **`--no-comment`** (so it doesn't advertise its
-   local-only round commits before be-review pushes). Its step-2 `Workflow` runs
+   local-only round commits before be-review pushes), and thread both `context`
+   (the task / main-agent context, so the codex **author inherits what you know —
+   not just the diff** — every round) and `rationale` (so codex doesn't flag
+   deliberate decisions at the source) straight through. Its step-2 `Workflow` runs
    in the background; **wait for it to finish** before starting the simplify step.
    It commits its rounds and **returns** its rendered comment body — hold onto it
    to post after the final push. (On persistent `reviewer-error` there is **no
--- a/.agents/skills/be/SKILL.md
+++ b/.agents/skills/be/SKILL.md
@@ -56,8 +56,10 @@ applies its own fixes directly — no snapshot, no apply pass. be-review pushes 
 at the end and *then* posts the PR comments (lens, codex, and a code-police
 summary), so no comment advertises a local-only commit.
 
-- Pass `base` and the change **`rationale`** (so the lenses don't flag deliberate
-  decisions). Preflight is a non-empty diff and (since codex runs) `codex login
+- Pass `base`, the change **`rationale`** (so the lenses don't flag deliberate
+  decisions), and **`context`** — the task intent and key decisions you hold from
+  this run, so the codex author **inherits what you know instead of re-deriving it
+  from the diff**. Preflight is a non-empty diff and (since codex runs) `codex login
   status`.
 - Lens-debate commits its agreed fixes; codex's rounds commit `fix(…)`; simplify
   and code-police commit `refactor:` / `fix(police):`. Confirm the post-push PR
--- a/.agents/skills/codex-debate/SKILL.md
+++ b/.agents/skills/codex-debate/SKILL.md
@@ -86,8 +86,9 @@ codex/opencode runtimes the skill is inert.
 ## Arguments
 
 A leading `review` subcommand token, if present, is consumed by mode detection;
-what remains is `[<pr-number>] [--base <branch>] [--no-commit] [--no-comment]` (the
-bare alias passes the whole argument string through unchanged). Parse:
+what remains is `[<pr-number>] [--base <branch>] [--no-commit] [--no-comment]
+[--rationale <note>] [--context <note>]` (the bare alias passes the whole argument
+string through unchanged). Parse:
 
 - **`<pr-number>`** (optional): a PR to debate. If given, `gh pr checkout <n>`
   first and default the base to that PR's base branch. If omitted, debate the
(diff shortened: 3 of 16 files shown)


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `d5e69877c88d`

### Applied (2)
- `hickey-1` rationale.md path literal duplicated across two sites — commit `a68953527`
- `hickey-2` codex-review.sh usage string drifted from the real signature — commit `0a7711acb`

### Agreed — no change (4)
- `lowy-1` The receptacle is sound: one inherited-context axis behind a stable note-block interface (.apm/skills/codex-debate/debate.workflow.js:63-93)
- `lowy-2` rationaleFileArg passed on warm rounds where codex-review.sh ignores it — wires exposed across the cold/warm seam (.apm/skills/codex-debate/debate.workflow.js:182,186 + scripts/codex-review.sh:99-104,134-167)
- `lowy-3` context/rationale kept as two args, routed to different audiences — correctly NOT complected (.apm/skills/codex-debate/debate.workflow.js:75-93)
- `hickey-3` rationale delivered via two different mechanisms (inline string vs. file) (.apm/skills/codex-debate/debate.workflow.js:88-93)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `d5e69877c88d`

### Round 1

**codex** — approved: `false`

The generated copies and lock hashes are consistent, and the core context/rationale split matches the deliberate design. I found one silent failure path in the new rationale handoff and one stale usage hint/documentation inconsistency.

Findings:
- `F1` · major · open — When a rationale file path is explicitly passed but the file is missing or empty, the script silently leaves `rationale` empty and runs codex without the deliberate-decisions block. The workflow passes a real path whenever `rationale` is present, so a failed `rationale:write` handoff makes the new feature disappear while producing a normal-looking verdict, which can reintroduce the exact false findings this change is meant to prevent. (.apm/skills/codex-debate/scripts/codex-review.sh:94)
- `F2` · minor · open — The skill frontmatter `argument-hint` still advertises only `review [<pr-number>] [--base <branch>] [--no-commit] [--no-comment] | answer "<prompt>"`, even though the review grammar now accepts `--rationale` and `--context`. The mode-detection prose at line 34 has the same stale abbreviated grammar. That makes the new flags harder to discover from the slash/skill UI and leaves two sources of truth disagreeing inside the same file. (.apm/skills/codex-debate/SKILL.md:4)

**claude** — Addressed both codex findings. F1 (major): partial — fixed the valid part (loud stderr WARNING when a non-"-" rationale path is missing/empty, mirroring the rebuttal-handoff convention right above it in codex-review.sh) but disputed the abort/reviewerError escalation, since rationale is a false-finding suppressor threaded to BOTH sides, not a correctness input — a lost codex copy degrades to a bare-diff review the implementor still disputes from its own inherited rationale block, so aborting the debate is more disruptive than the failure it guards. F2 (minor): fixed — synced the SKILL.md argument-hint frontmatter and mode-detection review grammar to add [--rationale <note>] [--context <note>], removing the two-disagreeing-grammars-in-one-file issue (the Arguments section already documented them). Edited the .apm/ source and regenerated the .claude/ and .agents/ copies via `just ai::apm` (apm.lock.yaml hash bump expected — generated, not hand-maintained).

- `F1` **partial** — Fixed the valid part: codex-review.sh now emits a loud stderr WARNING ("expected rationale file ... is missing or empty; proceeding with no deliberate-decisions note this round") when a non-"-" rationale path is missing/empty, instead of silently dropping the deliberate-decisions block. This mirrors the rebuttal-handoff WARNING immediately above it (lines 62-71) — the established broken-handoff convention also used for cross-check in codex-answer.sh:92. Disputed the stronger suggestion to abort or synthesize a reviewerError: rationale is a false-finding SUPPRESSOR, not a correctness input, and it is threaded to BOTH sides (debate.workflow.js injects rationaleBlock into the implementor every round). If codex's copy is lost the round degrades to a bare-diff review whose deliberate-choice findings the implementor still DISPUTES from its own inherited rationale block — the safety net holds. Aborting an entire multi-round debate over a missing suppressor hint would be far more disruptive than the failure it guards against, so loud-but-proceed is the correct severity match. File: .apm/skills/codex-debate/scripts/codex-review.sh:93-107 (+ generated copies).
- `F2` **fixed** — Updated SKILL.md frontmatter argument-hint (line 4) and the mode-detection review grammar prose (line 36) to append [--rationale <note>] [--context <note>], matching the real grammar already documented in the Arguments section (lines 90, 114-123). This eliminates the two-disagreeing-sources-of-truth in one file and makes the new flags discoverable from the slash/skill UI. Edited .apm/skills/codex-debate/SKILL.md and regenerated the .claude/ and .agents/ copies.

commit: `b15f9bcd069855a7f3fb4bf2e1b483f0dbddd5ed`

### Round 2

**codex** — approved: `true`

The follow-up fixes address the two open review findings. The rationale-file miss path now emits an explicit stderr warning instead of silently dropping the deliberate-decisions block, which resolves the silent-handoff failure I raised; the stronger abort behavior is not necessary for this input because the rationale is a review-noise suppressor and the author side still receives it. The codex-debate skill grammar and argument hint now advertise `--rationale` and `--context`, and the generated copies plus lock hashes are consistent. Read-only checks passed: `bash -n` for `codex-review.sh` and `git diff --check`.

Findings:
- `F1` · major · resolved — Resolved. A non-`-` rationale path that is missing or empty no longer fails silently; the script emits a clear warning explaining that codex will review the bare diff without the deliberate-decisions note. (.apm/skills/codex-debate/scripts/codex-review.sh:94)
- `F2` · minor · resolved — Resolved. The frontmatter `argument-hint` and mode-detection review grammar now include `[--rationale <note>] [--context <note>]`, matching the Arguments section and generated copies. (.apm/skills/codex-debate/SKILL.md:4)

_codex on the rebuttal:_ F1: I accept the dispute against aborting/synthesizing reviewerError. My blocker was the silent downgrade, not that missing rationale must be fatal. Since the script now emits an explicit stderr warning and the author still receives the rationale via its own prompt path, loud-but-proceed is an appropriate severity match. F2: fixed as requested; the stale grammar sources are now synced.

srid:

Passes run: **Rules** + **Fact-check**. Elegance skipped (`--no-elegance` — the `/simplify` track already ran this gauntlet). Scoped to `d5e69877…HEAD`.

**Outcome: 0 fixes applied.** Every finding was dispositioned as skip/dispute with a concrete reason — the change had already been hardened by the lens and codex tracks, and the substantive police finding re-raised a point codex already adjudicated.

| Finding | Disposition | Reason |
|---|---|---|
| Abort vs. warn on a missing/empty rationale file (×4 rule+fact-check variants) | **Disputed — upheld codex consensus** | This is codex's round-1 **F1**, debated to consensus: rationale is a false-finding *suppressor*, not a correctness input — its absence degrades to the pre-feature bare-diff review, the implementor independently receives it via its inline block (safety net holds), and the loud stderr WARNING mirrors the established rebuttal-handoff convention right above it. Escalating to abort would regress an adversarially-settled decision. |
| `writeFileAgent` lacks try/catch | **Skip** | The whole mechanical-writer pattern (`writeSection`, `ledger:reset`, merge-base resolver) awaits `agent()` without error handling; special-casing only this call is inconsistent, and the failure degrades to the same benign path above. |
| Add assertion that the write-guard and `rationaleFileArg` stay synced | **Skip — tautology** | Both `if (rationale)` and `rationaleFileArg = rationale ? rationaleFile : '-'` read the *same* variable; they cannot desync, so the assertion would assert its own definition. |
| Shell injection via `$rationale` in the prompt block | **Skip — empirically disproven** | Reproduced the exact `cat → block → unquoted-heredoc` path with `$(…)`, backticks, and `$((…))` in the file: all rendered **literal**, nothing executed. bash does not re-scan an expanded variable's value for command substitution — the same reason the existing `$rebuttal` path is safe. |
| comment-the-non-obvious (×2) | **Skip** | Both sites already have adjacent explanatory comments; the region is heavily commented. |

The fact that police's one substantive finding was already caught and resolved by the codex debate — and its security finding was a false positive — is itself a good signal for the gauntlet.

PR #1347, opened 2026-06-14
Restore App.tsx to a thin layout shell (#1340)

### LINKED ISSUE #1340: Refactor: App.tsx has drifted into a kitchen-sink — restore the "thin layout shell" invariant
`packages/client/src/App.tsx` is supposed to be a **thin layout shell** — the convention is explicit:

> **State per domain**: Extract shared state into `useXxx.ts` modules (singleton pattern)… **Keep App.tsx as a thin layout shell.** — `.claude/rules/solidjs.md`

and code-police's `read-domain-state-directly` rule reinforces it (*"prop-drilling… bloats App.tsx's wiring surface"*). In practice it's become the catch-all every feature lands a little wiring in — B3.2 (#1337) added the daemon-restart deps + the canvas `<Switch>`; it'll keep accreting unless we draw the line.

### Where it stands today (`App.tsx` @ `b3.2-supervised-restart`)

- **785 lines.**
- **15 reactive primitives in the shell** — 10 `createSignal`, 3 `createMemo`, 2 `createEffect`.
- **6 overlay/dialog open-state signals** living in the shell: `paletteOpen`, `shortcutsHelpOpen`, `aboutOpen`, `welcomeOpen`, `diagnosticInfoOpen`, `searchOpen`.
- The full **`ActionContext` assembly** (≈ lines 244–282) is hand-built here.
- **Imperative escapes**: `window.__koluSimulateAlert = …` and a `document.querySelector(...)` "is any dialog open" probe.

None of this is layout — it's domain state, wiring, and orchestration that the convention says belongs in `useXxx.ts` singletons or in the components that own the behavior.

### Proposed decomposition (incremental — no big-bang)

| Cluster in App.tsx | Extract to |
| --- | --- |
| The 6 overlay/dialog open-state signals + their effects | a `useOverlays()` singleton, **or** push each into the dialog component (per *"Components own their behavior"*) so App.tsx only mounts them |
| The `ActionContext` object assembly (244–282) | a `useActionContext()` module that composes the per-domain handlers |
| The canvas down/warming/empty/terminals `<Switch>` + `showEmpty` derivation | a single `canvasMode()` derivation (the `/simplify` pass on #1337 already flagged this) next to `useDaemonStatus` |
| Per-feature handler wiring that re-threads domain state | have consumers read the singleton hooks directly (`read-domain-state-directly`) instead of routing through App.tsx |
| `window.__` / `document.querySelector` escapes | fold into the owning hook (test-hook + a shared "modal open" signal) |

**Goal:** App.tsx returns to *layout composition only* — mount the chrome bar, the canvas surface (by `canvasMode()`), the dialogs, and the overlays; all state and orchestration in domain modules.

### Notes

- Do this **incrementally**, one cluster per PR, each behavior-preserving (the e2e suite is the guard).
- Related: #926 (per-folder public-API boundaries via dependency-cruiser) — a structural-boundary effort this complements.
- Filed off the back of #1337 (B3.2), where the daemon-restart wiring was the latest thing to land in the shell.

srid:
**`App.tsx` had drifted into an 785-line kitchen-sink** — 11 reactive primitives, six dialog open-state signals, the full `ActionContext` assembly, a prop-drilled server-identity fetch, and two imperative escapes all living in what is supposed to be a layout shell. This decomposes it back to *layout composition only*: it mounts the chrome, the canvas surface, the dialogs, and the overlays, and composes domain singletons for everything else. **Behavior-preserving** — no UI change; the e2e suite and existing unit tests are the guard.

The shape was designed up front in a plan-of-record Atlas note and pressure-tested through the lowy (volatility) ⇄ hickey (simplicity) lenses *before* any code — which is why this is six volatility-aligned seams, not the issue's naive five "clusters" (the lenses killed a `useOverlays()` god-hook, switched `useActionContext` to verbs-not-setters, and surfaced a sixth missing seam, `useServerIdentity`).

### The six seams

| Seam → owner | Encapsulates | App.tsx sheds |
| --- | --- | --- |
| **`useCanvasMode`** (`kaval/`) | *which* canvas surface wins, in what order — a tagged-union derivation, finally unit-testable without a DOM | the outer `<Show>` + 4-arm `<Switch>` conditions + `showEmpty` |
| **Per-dialog controllers** — `createDisclosure()` singletons + `useCommandPalette` | five *independent* dialog visibilities | 6 open-state signals, `withRefocus`, `handlePaletteOpenChange` |
| **`useDialogStack` + `ModalDialog.refocusOnClose`** | the one shared "is any dialog open" + close-refocus policy | the `document.querySelector` DOM probe + 3 duplicated refocus copies |
| **`useTerminalSearch`** (`terminal/`) | per-terminal find visibility | `searchOpen` + its `on(activeId)` reset effect |
| **`useServerIdentity`** | the `client.server.info()` fetch + `appTitle` | the identity signal + prop-drill |
| **`useActionContext`** (`input/`) | the keyboard+palette wiring fan-in | the ~40-line `ActionContext` literal |

Two enabling moves ride along: **`useTerminalCrud` is promoted to a `createSharedRoot` singleton** (so `TerminalContent` / `TileTitleActions` read it directly instead of receiving drilled closures), and the **`__koluSimulateAlert` test-hook folds into its producer** (`useTerminalAlerts`). `closeTerminal`, the export/screenshot closures, the canvas-arm JSX, and the `<Resizable>` split *stay* — that's the shell's irreducible layout job.

> The `useDialogStack` mount-counter replaces the DOM probe rather than a fragile signal-OR: every dialog already routes through `ModalDialog`, so incrementing there makes the count complete and self-healing — a new dialog is counted for free, with no enumeration to keep in sync.

### Result + the guardrail

**`App.tsx`: 785 → 559 lines, 11 → 3 reactive primitives.** To stop the drift from recurring, this adds the **`app-shell-stays-thin` code-police rule** (`.agency/code-police.md`) and a CI-enforced reactive-primitive budget (`App.shell.test.ts`) — the "thin layout shell" convention finally has teeth. New domain state must go in a `useXxx.ts` singleton; bumping the budget requires justifying in the PR why the state is genuinely layout-level.

Plan-of-record: [`app-thin-shell` Atlas note](https://htmlpreview.github.io/?https://github.com/juspay/kolu/blob/refactor/app-thin-shell/docs/atlas/dist/app-thin-shell.html). Closes #1340.


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `3a8fd9b0d63d`

### Applied (11)
- `lowy-1` useCommandPalette reaches past the ModalDialog receptacle into refocus internals — commit `2d494eee9`
- `lowy-2` daemonDown() dead export survives despite the plan-of-record committing to delete it — commit `3b0f094c6`
- `lowy-4` useServerIdentity exposes both the raw identity() accessor and the two derived projections — commit `57772f41c`
- `lowy-5` App.shell.test.ts budget guards primitive COUNT but not the regex-evasion of inline state — commit `f342d46d3`
- `hickey-1` Two naming schemes for the same "open this overlay" verb — commit `ee74ec925`
- `hickey-2` TileTitleActions repeats "select this tile, then act" four times — commit `aa75741c2`
- `hickey-3` TileTitleActions "open search for this tile" duplicates the search verb — commit `d99bf3754`
- `hickey-4` useTerminalSearch singleton shape diverges from the createSharedRoot pattern — commit `181447025`
- `hickey-5` App.tsx still owns two closures that re-thread store reads into palette deps — commit `e2c18b7eb`
- `hickey-6` canvasMode partition is rebuilt into per-arm narrowing accessors in the shell — commit `184e1e76e`
- `hickey-7` useServerIdentity comment claims a consumer (About dialog) it doesn't reach — commit `817b190a5`

### Agreed — no change (1)
- `lowy-3` useTerminalSearch's removeTerminal eviction is wired through crud, not the store that owns the lifecycle (packages/client/src/terminal/useTerminalSearch.ts:31-33 and packages/client/src/terminal/useTerminalCrud.ts:88)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `978836017b2b`

### Round 1

**codex** — approved: `false`

The shell extraction is directionally coherent, but I found a few user-visible behavior regressions around palette close/refocus and terminal search state, plus a promised test that did not land and stale Atlas prose. I am not approving while these findings remain open.

Findings:
- `F1` · major · open — Programmatic palette closes bypass the new ModalDialog close-refocus path. `closeForSelection()` calls `props.onOpenChange(false)` directly, so `ModalDialog.handleOpenChange()` never runs and `refocusIfNoDialogOpen()` is not scheduled. Selecting a normal command or workspace row can leave focus on the now-unmounted palette input instead of returning it to the terminal. The previous App-level `handlePaletteOpenChange` covered this path because the palette called that same close handler directly. (packages/client/src/CommandPalette.tsx:412)
- `F2` · minor · open — `toggle()` mutates `open` directly and bypasses the close bookkeeping in `onOpenChange()`. If the palette was opened into a specific group via `openGroup()`, pressing the palette toggle shortcut to close it leaves `initialGroup` set; the next plain toggle can reopen into the stale group. This path also bypasses the close-refocus behavior described in F1. (packages/client/src/useCommandPalette.ts:32)
- `F3` · major · open — Terminal search state is now stored per terminal, but inactive terminals only hide the search UI via `props.focused && search.isOpen(...)` rather than clearing it. This regresses the old explicit behavior in App that closed search on active-terminal changes, so returning to a terminal can unexpectedly resurrect its find bar and multiple terminals can retain latent open search state. `handleCloseAll()` also resets the terminal store without clearing these keys, despite the hook comment saying keys are evicted when terminals are removed. (packages/client/src/terminal/useTerminalSearch.ts:19)
- `F4` · minor · open — The Atlas note says the branch added a sibling unit test for `canvasMode` tier precedence and payload preservation, but no such test exists. `useCanvasMode.ts` now contains load-bearing precedence logic for loading, down, warming, empty, and ready states, and the only added test is the App source-token tripwire. (docs/atlas/src/content/atlas/app-thin-shell.mdx:222)
- `F5` · nit · open — The note says `App.tsx` landed at 559 lines, but the current file is 529 lines. The generated HTML repeats the stale number. (docs/atlas/src/content/atlas/app-thin-shell.mdx:232)

**claude** — Addressed all 5 codex findings on the app-thin-shell refactor. F1/F3/F4/F5 were genuine and fixed; F2 fixed but the regression framing disputed (the bug pre-existed the refactor). Client typecheck clean, 301/301 client unit tests pass (293 + 8 new resolver tests), full biome lint clean on packages/client/src, Atlas check (build+types+helper tests) passes. Committed as 6fdd8fae.

F1: The palette's selection close called props.onOpenChange(false) directly, bypassing ModalDialog's close-refocus, so a normal command could leave focus on the unmounted palette input. Centralized close bookkeeping in useCommandPalette.close() (group reset + guarded refocusIfNoDialogOpen); selection, Corvu-dismiss, and toggle now all funnel through it, and dropped refocusOnClose from the palette's ModalDialog to avoid a double-fire on the Corvu path.

F3: The keyed search store resurrected a find bar on return to a terminal, diverging from the old single searchOpen signal reset by on(activeId). Restored that contract with an on(store.activeId) clear-all effect plus useTerminalSearch.reset() wired into handleCloseAll (killAll bypasses per-terminal eviction).

F4: Extracted the load-bearing precedence into a dependency-free resolveCanvasMode(facts) in kaval/canvasModeResolver.ts and added canvasModeResolver.test.ts (8 cases) so the precedence is tested as a pure function with no daemon-status subscription to mount.

F5: Corrected the stale 559 line count to ~530 and rebuilt the Atlas dist HTML.

- `F1` **fixed** — Confirmed real regression. Traced the data flow: CommandPalette.closeForSelection() (CommandPalette.tsx:412) calls props.onOpenChange(false) = commandPalette.onOpenChange, which only reset initialGroup and never re-entered ModalDialog.handleOpenChange, so refocusIfNoDialogOpen() was never scheduled — the old App-level handlePaletteOpenChange (passed as the palette's onOpenChange prop) had covered exactly this path. Fix: useCommandPalette.ts now has a private close() that does setOpen(false) + setInitialGroup(undefined) + refocusIfNoDialogOpen(); onOpenChange(false) delegates to it, so selection closes (and Corvu dismiss, which forwards through this same onOpenChange) get the guarded refocus. Removed refocusOnClose from the palette's ModalDialog mount (CommandPalette.tsx) so the Corvu path doesn't double-fire the refocus. The guard ordering matches the old behavior: closeForSelection schedules the rAF before action.onSelect() runs, so a command that opens another dialog increments dialogStack synchronously before the deferred refocus check and refocus is correctly skipped.
- `F2` **partial** — Fixed the code but dispute the 'regression' framing. Fixed: toggle() in useCommandPalette.ts now routes its close half through the shared close(), so a Cmd+K toggle-close clears the stale initialGroup and runs the refocus, same as every other close path. Dispute: this was NOT introduced by the refactor. The pre-refactor Cmd+K action was ctx.setPaletteOpen((v) => !v) (old input/actions.ts:186) — a raw signal toggle that equally left paletteInitialGroup set on close and never refocused. So both the stale-group and no-refocus behaviors are pre-existing latent issues, not regressions. I fixed it anyway because it's a cheap, real correctness improvement now that the close policy is centralized.
- `F3` **fixed** — Confirmed real behavior divergence. Old App held a single searchOpen signal with createEffect(on(store.activeId, () => setSearchOpen(false))) (old App.tsx:182) that closed the find bar on every active-terminal change. The new per-key Record store only hides inactive terminals' bars via the props.focused gate (TerminalContent.tsx:96) but keeps their key true, so switching away and back resurrects the bar — and handleCloseAll (useTerminalCrud.ts) called store.reset() without clearing search keys, contradicting the hook's own 'keys are evicted when terminals are removed' comment. Since the task mandate is behavior-preserving/no-UI-change, restored the old contract: added an on(store.activeId) clear-all effect in useTerminalSearch.ts (deferred, no import cycle since search already imports the store), a clearAll helper, and a reset() method wired into handleCloseAll after store.reset(). Updated the hook doc to describe the actual behavior.
- `F4` **fixed** — Confirmed: no canvasMode precedence test existed — only App.shell.test.ts (the source-token tripwire) matched. The Atlas note explicitly promised this test. Rather than weaken the prose, extracted the load-bearing decision into a dependency-free resolveCanvasMode(facts) in new kaval/canvasModeResolver.ts (CanvasMode + CanvasFacts types live there too); useCanvasMode.ts now just gathers the live daemon/session facts and delegates, re-exporting CanvasMode for App.tsx. The first attempt at a test importing through useCanvasMode failed because that module transitively pulls in ../wire; the pure-resolver module breaks that coupling. Added canvasModeResolver.test.ts with 8 cases covering connecting > down > warming > empty > workspace precedence and the down/warming payloads. Also updated the note prose to name canvasModeResolver.test.ts / resolveCanvasMode accurately and rebuilt the Atlas dist.
- `F5` **fixed** — Confirmed: App.tsx is 529 lines (wc -l), the note and generated HTML said 559. Per codex's suggestion to avoid an exact number that drifts, changed the prose in app-thin-shell.mdx to '~530 lines' (the '3 reactive primitives' count is correct and budget-enforced, so it stays). Rebuilt the Atlas dist via just atlas::build so app-thin-shell.html no longer contains '559 lines'; just atlas::check (build + types + idempotency + helper tests) passes.

commit: `6fdd8fae41eba2746abfc92ac6996eef4f17a60b`

### Round 2

**codex** — approved: `true`

All five prior findings are resolved in the current tree. The palette close/refocus paths now converge in `useCommandPalette.close()`, terminal search behavior matches the old active-terminal-switch reset contract, the canvas-mode precedence has a pure resolver with focused tests, and the Atlas line-count text was corrected. I found no new regression introduced by CLAUDE's response patch.

Findings:
- `F1` · major · resolved — Programmatic command-palette selection closes bypassed `ModalDialog`'s guarded refocus path, so selecting a normal command could leave focus on the closing palette instead of returning it to the terminal. (packages/client/src/useCommandPalette.ts:28)
- `F2` · minor · resolved — The Cmd+K toggle-close path could bypass the palette's normal close bookkeeping, leaving stale `initialGroup` state and skipping terminal refocus. (packages/client/src/useCommandPalette.ts:52)
- `F3` · major · resolved — The new keyed terminal-search store changed behavior from the old single `searchOpen` signal: switching away and back could resurrect a find bar, and kill-all bypassed per-terminal eviction. (packages/client/src/terminal/useTerminalSearch.ts:40)
- `F4` · minor · resolved — The load-bearing canvas-surface precedence was extracted without the promised focused unit coverage, leaving down/warming/empty ordering protected only indirectly. (packages/client/src/kaval/canvasModeResolver.test.ts:29)
- `F5` · nit · resolved — The Atlas note reported a stale exact App.tsx line count, making the implemented-state documentation inaccurate. (docs/atlas/src/content/atlas/app-thin-shell.mdx:236)

_codex on the rebuttal:_ F2: I concede the regression framing. The pre-refactor Cmd+K handler used the raw `setPaletteOpen((v) => !v)` path, so the stale-group/no-refocus behavior on toggle-close was already latent rather than introduced by this refactor. The code change still fixes the real close-path inconsistency by routing toggle-close through `close()`, so F2 is resolved.

srid:

Ran with `--no-elegance` (the `/simplify` track already swept this tree). Scope: `git diff 978836017b2b...HEAD`.

| Pass | Issues | Outcome |
| --- | --- | --- |
| Rules | 1 | `no-silent-error-swallowing` on `handleCreate`'s `.catch(() => {})` → **fixed** (`ca05f566`) — made it explicit at the catch site that the error is already surfaced by `handleCreate`'s own toast and the catch only absorbs the re-throw. |
| Fact-check | 7 | 1 actioned (the same `handleCreate` item); **6 verified false-positive against the source** and skipped with reasons below. |
| Elegance | – | Skipped (`--no-elegance`; `/simplify` ran). |

### Fact-check findings — adjudication

Each was checked against the actual code rather than the diff alone:

- **`canvasMode` memo staleness** — *false positive.* `canvasMode()` reads **all** facts eagerly into the object literal (including `terminalCount()`) *before* `resolveCanvasMode` runs, so the memo unconditionally tracks `store.terminalIds()`. The resolver's short-circuit can't drop the subscription.
- **`void crud.exportSessionHtml()` / `exportScrollbackAsPdf()` swallow failures** — *false positive.* Both functions `toast.error(...)` internally (`exportSessionAsHtml.ts:41`, `exportScrollbackAsPdf.ts:25,47`); the `void` only discards their `void` return.
- **Refocus race (decrement vs. rAF)** — *false positive.* SolidJS runs the `createEffect` `onCleanup` decrement **synchronously** on the `open → false` update, before the deferred `requestAnimationFrame` fires — the count is accurate at refocus-check time.
- **`onTile` swallows async rejections** — *non-issue.* `fn` is typed `() => void`; the one async caller (`screenshotTerminal`) `void`-wraps internally.
- **`clearAll` piecemeal `produce`-delete** — *correct as written.* The delete loop is the right Solid idiom for clearing a `Record` store; the suggested `setState({})` would **merge** and leave stale keys. Cost is O(handful-of-terminals), negligible.
- **`useDialogStack` underflow clamp** — *defensive nit, left as-is.* The counter is balanced by the `createEffect`/`onCleanup` pairing; `Math.max(0, n-1)` floors to the safe "refocus allowed" default.

**Net:** 1 fix applied; no real defects left open.

srid:

The **x86_64-linux** lane ran on **`kolu-ci-4`** (`dev-x86-64-linux-08`) — commit `ca05f566`, **failed** ✗

- **Lane wall** (pipeline): **5m39s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 4m34s | ✗ (exit 1) |
| `ci::home-manager` | 1m13s | ✓ |
| `ci::smoke` | 48s | ✓ |
| `ci::nix` | 42s | ✓ |
| `_ci-setup` | 40s | ✓ |
| `ci::flake-check` | 39s | ✓ |
| `ci::unit` | 28s | ✓ |
| `ci::install` | 23s | ✓ |
| `ci::fmt` | 15s | ✓ |
| `ci::biome` | 15s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/ca05f56/timings.jsonl`; pool state is a live `flock` probe.</sub>

srid:
## Evidence

This is a **behavior-preserving refactor with zero intentional visual diff**, so the comprehensive proof is the **full e2e suite green on both platforms** (`ci::e2e@x86_64-linux` ✓, `ci::e2e@aarch64-darwin` ✓) — it drives every surface this PR touched: dialogs, command palette, per-terminal search, canvas modes, splits, screenshots, theme pill, workspace switcher.

### A real regression — caught by CI, fixed, re-proven green

The review gauntlet's structural cleanup (a lens/codex "uniform reading" change) had swapped App.tsx's canvas `<Switch>/<Match>` for a ts-pattern `match(mode())`. That's a **SolidJS reactivity trap**: `mode` is a memo that returns a *fresh* `CanvasMode` object each recompute, so `{match(mode())…}` in JSX re-runs and **re-creates the whole canvas subtree** on every daemon/terminal tick — remounting `TerminalCanvas`/`TerminalContent`, which made Corvu re-fire `onCollapse` and silently collapse a just-opened split. The entire `sub-terminal.feature` (+ workspace-switcher, mobile splits) went red.

**Root-caused live** with chrome-devtools (the sub *was* created but the panel stayed `collapsed`/`0%`; a second toggle recovered it — the create-window collapse race), traced to the `match()`→subtree-recreation, and fixed by reverting the *rendering* to `<Switch>` (keeping the `canvasMode` tagged union + the pure `canvasModeResolver`). `<Match when={mode().kind === '…'}>` keys on a stable boolean, so the arm persists and the canvas isn't recreated.

**Verified fixed three ways:**
- Live (chrome-devtools): toggling a fresh terminal's split now expands the sub-panel on the **first** toggle (tab bar visible, h=144px).
- Locally: `sub-terminal.feature` → **15 scenarios / 104 steps, all green** (was ~10 failing).
- CI: the full e2e suite is green on the final `HEAD` `755be62e` on **both** platforms.

This is the case the PR evidence guidance names for a no-visual-diff refactor: the behaviour worth proving is *"nothing observable changed,"* and the maintained Cucumber + Playwright suite exercising every surface is the strongest proof of that — far more than any single hand-captured clip.

PR #1344, opened 2026-06-13
Terminals survive a kolu update: kaval adoption (B3.3)

srid:
**Your terminals — shells, scrollback, and any running agent — now survive a kolu update.** When a redeploy doesn't change kaval's own source (the common case), the daemon keeps running across the kolu-server restart and kolu **adopts** its live PTYs on boot instead of recycling them. The only visible sign is a brief "N terminals reattached" toast and the kaval rail's uptime, which keeps counting from *before* the update rather than resetting.

This is **B3.3** of the [R-4/kaval survival chain](https://github.com/juspay/kolu/blob/master/docs/atlas/src/content/atlas/pty-daemon.mdx) — the chain's headline, landing after B3.2's supervised restart so it reuses the proven connection plumbing.

### The seam — the spine adopts a *connection*, kolu reconciles its *contents*

| Layer | What it adds | Hashed into `KAVAL_BUILD_ID`? |
| --- | --- | --- |
| `@kolu/surface-daemon-supervisor` (spine) | **`adoptOrEnsure()`** — adopt a live, handshake-compatible survivor (connect, never kill); recycle only an absent / dead / **truly skewed** one. A transient (non-skew) connect blip is retried and, if it persists, leaves the survivor **standing** (`degraded`) — never kills live PTYs over a hiccup. | no |
| `packages/server` (soul) | **`reconcile.ts`** (pure: `daemon.list() × savedSession → {adopt, adoptOrphans}`); **`adoptTerminal`** / **`adoptOrphan`**, siblings of `spawnAndWire` that wire an already-alive PTY from the **whole** saved record + the **live** daemon snapshot (authoritative `cwd`/foreground); the boot reattach that adopts every live PTY, converges the session, and fails **closed** if it can't list. | no |
| `packages/client` | a one-shot **"N reattached"** toast keyed on a new optional `DaemonStatus.adopted`. | no |

> **Zero kaval-hashed roots touched → `KAVAL_BUILD_ID` is bit-identical → no forced restart.** A kolu-only deploy must not flip the daemon's staleKey, or there'd be nothing to adopt.

### What this makes impossible by construction

[...]

PR #1301, opened 2026-06-12
kaval B1: the standalone PTY daemon binary, its client, and full e2e

srid:
**The in-process pty-host becomes `kaval` — a standalone PTY daemon you can run on a box where kolu has never been installed**, in the [drishti](https://github.com/srid/drishti)/[odu](https://github.com/juspay/odu) graduate tradition, with kolu as its first client. B1 makes it *graduation-ready* — its own binary, its own socket, zero `kolu-*` deps — without changing the topology: kolu still embeds the host in-process (the daemon flip is B2), and spawn behaviour is byte-unchanged. This is the second of four PRs in the [kaval plan](https://htmlpreview.github.io/?https://github.com/juspay/kolu/blob/kaval-b1-binary/docs/atlas/dist/pty-daemon.html) (B0 ✅ → **B1** → B2 → B3).

### The shape

```
  kaval-tui ──dials──┐
                     ▼
   bin.ts ─▶ daemonMain.ts ─▶ @kolu/surface-daemon   (gate → serve → teardown)
   (kaval)   (kaval's ~20-line     │  pidGate · daemonMain skeleton
              composition)         ▼
                          ptyHostSurface over a unix socket
   index.ts (the embedded library surface kolu-server still uses in-process)
```

### What landed

- **`@kolu/surface-daemon`** — a new package, the **daemon half** of the [surface-daemon spine](https://htmlpreview.github.io/?https://github.com/juspay/kolu/blob/kaval-b1-binary/docs/atlas/dist/surface-daemon.html): `acquirePidGate`/`readPidGate` (the atomic single-instance gate — `link(2)`, liveness-probe, stale-reap — *both sides of one file format*, the read side for B2's supervisor) and `daemonMain` (the `gate → serve → SIGTERM-teardown` skeleton, parameterized over scope key, socket path, the surface `router`, and lifetime `forever | idleTimeout`). Zero `kolu-*` deps. `odu serve` is the planned second tenant.
- **`kaval`** (was `@kolu/pty-host`) — gains `bin.ts` (the executable) over a ~20-line `daemonMain.ts` composition. Identity env `KOLU_PTY_HOST_BUILD_ID` → `KAVAL_BUILD_ID` (+ `KAVAL_COMMIT_HASH`), its own namespace, not kolu's.
- **`kaval-tui`** (was `kolu-tui`) — dials kaval's socket by default; `--pty-host-socket` → `--socket` (point it at a running kolu-server to reach its in-process terminals). Rebranded as the standalone client it now is.

[...]

PR #1350, opened 2026-06-14
NixOS VM tests for B3.3 kaval adoption — adopt + contract-skew recycle (closes #1349)

### LINKED ISSUE #1349: Test B3.3 kaval adoption in a NixOS VM (close the staged-prod gap CI can't reach)
**B3.3 (kaval adoption — terminals survive a kolu redeploy, [#1344](https://github.com/juspay/kolu/pull/1344)) has no automated end-to-end coverage**, by construction: its acceptance is a *staged-prod checklist* because the Playwright e2e harness structurally can't reproduce it. A **NixOS VM test** can — it runs real systemd, which is exactly the missing ingredient. This issue captures what that test takes, including the non-obvious caveats (researched against the code).

## Why e2e can't, and a VM can

Adoption only happens when **(a)** the kaval daemon is spawned as a *survivable* separate process (production: `systemd-run --user` → its own transient cgroup), **(b)** kolu-server restarts on the **same listen port** (same `kaval-<port>/` socket namespace), and **(c)** the server is killed without taking the daemon's cgroup down. The Playwright harness forces the opposite: it sets `KOLU_KAVAL_SPAWN=detached` and runs one server per worker on a random port (`packages/tests/support/hooks.ts:587-591`), and its only kaval lifecycle test *kills* the daemon then reboots the server fresh — a recycle, never an adopt (`packages/tests/features/kaval-daemon.feature`). A NixOS VM test has real systemd and can restart the server while the daemon's transient unit survives.

## Where it goes — rides `ci::home-manager` for free

One NixOS VM test already exists: `checks.x86_64-linux.vm-test` in `nix/home/example/flake.nix` (`pkgs.testers.nixosTest`, kolu as a **systemd user service**, `alice` auto-logged-in, driven via `machinectl -q shell`). It's built by `ci::home-manager` (`ci/mod.just:104-110`, which `devour-flake`-builds the nested flake and realizes *every* check). **A new `checks.x86_64-linux.adoption-vm-test` in that same nested flake is picked up automatically — zero new CI DAG nodes, no critical-path lengthening.**

## The positive test (do this first)

Reuse the existing VM's shape, adding the survival precondition:

1. Boot kolu as the **built package** under a systemd **user** service (so `KOLU_KAVAL_BIN` + `INVOCATION_ID` are set → the survivable `systemd-run --user` path), **with linger enabled** (see caveat A).
2. Create a terminal headlessly — `POST /rpc/terminal/create` (oRPC over HTTP; `kaval-tui` has **no** spawn verb). Sleep >500 ms so it hits the debounced autosave.
3. Capture proofs of life: `kaval-tui list --json` (id+pid), the gate pid at `$XDG_RUNTIME_DIR/kaval-7681/kaval.pid`.
4. `systemctl --user restart kolu` (same port).
5. **Assert adoption (not respawn):** same `{id,pid}` set from `kaval-tui list --json`, **same gate pid**, and `journalctl --user -u kolu` shows `adopted surviving terminals after restart` / `adopted a surviving daemon (its PTYs are preserved)` with the survivor's **older `startedAt`**.

### A. The `systemd-run --user` + **linger** precondition — the #1 gotcha
`systemd-run --user` is taken only when `!fromSource && INVOCATION_ID` is set (`packages/surface-daemon-supervisor/src/driver.ts:134-137`), and `fromSource = !KOLU_KAVAL_BIN || KOLU_KAVAL_SPAWN==="detached"` (`packages/server/src/ptyHost/localDriver.ts:140-141`). The production nix wrapper sets `KOLU_KAVAL_BIN` (`default.nix:255`); a user service sets `INVOCATION_ID`. So the VM **must** run the built package as a user service and **must not** set `KOLU_KAVAL_SPAWN=detached`.

For the transient unit to **outlive** `systemctl --user restart kolu`, the user manager and the unit must persist → **`users.users.<u>.linger = true`** (`loginctl enable-linger`). The existing VM has **no linger** (it leans on auto-login holding the session); the adoption test needs it added. Per-spawn unit names are unique, so discover via `systemctl --user list-units 'kaval-*'`, not a fixed name.

### B. Silent-wrong-path risk — assert the *survivor*, not "a terminal exists"
The kolu user service sets no `KillMode` (`nix/home/module.nix:120-125`), so systemd's default `KillMode=control-group` SIGKILLs everything in kolu's cgroup on restart. A plain detached child stays in that cgroup and **dies** (the [#1031](https://github.com/juspay/kolu/issues/1031) cgroup-v2 lesson); only `systemd-run --user` escapes. **If `fromSource` is mis-set, the daemon silently dies, a fresh kaval serves an empty list, reconcile drops the now-exited terminals, the canvas looks recovered, and the test exercises FRESH-SPAWN — falsely passing.** Mitigation: assert the **same kaval pid / unchanged `startedAt`** (and ideally the surviving shell pid + scrollback), not merely that a terminal is present. Leave `KillMode` at the default — it *is* the hazard `systemd-run` exists to escape; don't let a reviewer "fix" it.

### C. Driving + probing headlessly
- **Create:** `terminal.create` is mounted at `/rpc/*` (oRPC `RPCHandler`, no auth on localhost — `packages/server/src/index.ts`), all inputs optional, so a tiny oRPC/curl POST opens a terminal. *(If the service runs HTTPS self-signed, the client needs `NODE_TLS_REJECT_UNAUTHORIZED=0`; or run with TLS disabled → plain HTTP.)*
- **Probe:** `daemonStatus.adopted` is **WebSocket-streaming only** (no plain HTTP GET), so the practical signal is the **journald log line**, backed by `kaval-tui list --json` pid-equality + the gate-pid file.
- **Autosave timing:** a terminal created then *immediately* restarted may miss the 500 ms autosave and be adopted as an **orphan** (`orphansAdopted`) rather than a saved record — sleep >500 ms to exercise the saved-adopt path, or assert `adopted + orphansAdopted >= 1` to be timing-robust.

### D. Linux-only — macOS survival stays uncovered
NixOS VM tests are `x86_64-linux` only; the `aarch64-darwin` lane just eval-builds the launchd plist (no launchd session on builders). So this proves the **systemd** survival path; **macOS launchd-detached adoption survival is tested by nothing** and remains staged-prod / manual on a Mac.

### E. The negative path (adoption *declined*) is materially harder — and a confusion to state correctly
Adoption is declined on **exactly one** condition: a **contract-version skew** — the survivor's `PTY_HOST_CONTRACT_VERSION` is incompatible per `isContractVersionCompatible` (`packages/surface/src/define.ts:667-681`: same major **and** survivor's minor ≥ expected), so `connectKaval` throws `DaemonContractSkewError` and `adoptOrEnsure` recycles (`packages/server/src/ptyHost/connect.ts:73-86`).

> ⚠️ **Correct the common confusion:** a **staleKey flip** (the `KAVAL_BUILD_ID` *closure hash* — a node-pty bump, a tap tweak) does **NOT** decline adoption. The survivor stays contract-compatible, so it is **adopted** (older `startedAt`); only B3.4's "update pending" nudge would fire (not wired in this branch). A VM test that bumps a source file and expects a recycle is testing the **wrong thing** — it'll observe adoption, the opposite of the intended assertion.

`PTY_HOST_CONTRACT_VERSION = "3.0"` is a **hardcoded constant** (`packages/kaval/src/ptyHostSurface.ts:64`) with **no env seam** (unlike `KAVAL_BUILD_ID`, which is env-driven). So a *real* skew needs two builds whose constant differs incompatibly (cleanest: a major bump 3.0 → 4.0) → ~2× the build cost plus a parameterized derivation patching the constant.

**Cheaper first cut:** the skew verdict depends only on the `system.version.contractVersion` string the survivor returns, so a **fake daemon (or one-line-patched kaval)** that claims the gate, serves the socket, and answers `system.version` with an incompatible version (e.g. `"99.0"`) drives the real server's skew→recycle→PTY-loss path with no second app build. *(It proves the gate, not that a real bumped kaval is correctly rejected — that stronger guarantee still wants the two-build version.)*

## Suggested phasing

1. **Positive adopt VM test** (the 80%): user service + linger, create-over-`/rpc`, restart, assert pid-equality + gate-pid + adoption log line.
2. **Negative skew→recycle** via the cheap fake-daemon/patched-constant proxy.
3. *(Later)* the full two-build negative path, and — once **B3.4** lands — the staleKey-flip "update pending" nudge on an *adopted* daemon.

## References
- PR [#1344](https://github.com/juspay/kolu/pull/1344) (B3.3 adoption) · the plan's staged-prod gate in `docs/atlas/src/content/atlas/pty-daemon.mdx` ("The gate — the second deploy")
- [#1031](https://github.com/juspay/kolu/issues/1031) (the cgroup-v2 survival lesson the `systemd-run` path encodes) · [#1034](https://github.com/juspay/kolu/issues/1034) (the data-loss lineage)
- Existing VM test: `nix/home/example/flake.nix` · CI wiring: `ci/mod.just:104-110`

srid:
**Closes #1349.** B3.3 (kaval adoption — terminals survive a kolu redeploy, #1344) had no automated end-to-end coverage: the headline path is unreachable in the Playwright e2e harness (no systemd, one server per worker, forced non-survivable spawn). A **NixOS VM has real systemd**, so the production `systemd-run --user` survival path actually works — these tests exercise exactly that.

> Stacked on **#1344** (base = `b3.3-adoption`) so the diff is just the tests. Retarget to `master` once #1344 merges.

Two checks live in their own folder, **`nix/home/example/adoption/`**. Both share one scaffold — **`lib.nix`** owns `mkAdoptionTest` (the survival VM node, the boot polls, the `machinectl`-as-alice + assert-result-file-as-root helpers, the jq/curl bindings, and the runtime-layout literals); `adopt.nix` and `skew.nix` are just their distinguishing data. `flake.nix` spreads both into `checks.x86_64-linux`, so they **ride `ci::home-manager`** — no new CI recipe, no new DAG node.

### ✅ `adoption-adopt` — the daemon is adopted (positive)
1. boot kolu as a systemd **user service** (with **linger** — the survival precondition);
2. open a terminal over the **oRPC HTTP API** (`/rpc/terminal/create`, no browser);
3. run a command in it (`terminal.sendInput "echo <nonce>"`) whose unique output we record;
4. **`systemctl --user restart kolu`** — the server only; the kaval daemon lives in its own `systemd-run --user` transient cgroup;
5. assert the **same daemon** (gate pid), the **same PTY** (id+pid), the command's **output still in the scrollback**, *and* kolu's own **reconcile log** all survived — adoption, not a fresh respawn.

### ♻️ `adoption-skew` — a contract-skewed survivor is recycled (negative)
When a redeploy **changes kaval's wire** (a `PTY_HOST_CONTRACT_VERSION` bump), the surviving daemon is incompatible and must be **recycled, not adopted** — but the saved session is left untouched so a restore is still offered. There's no env seam for the contract version (a source constant), so the "newer kolu" is a **second build** via a new test-only **`contractVersionOverride`** arg in `default.nix` (a `postPatch` sed of the constant, guarded by a grep that fails the build if the constant moves; `null` = no-op, real builds untouched).

The test seeds a terminal + a **saved session** on the old (3.0) daemon, stops the old server (the daemon survives), then starts the **bumped (9.0)** server on the same port. It asserts the survivor was **recycled** (daemon gate pid *changed*), the **contract skew was logged**, *and* the session is **preserved** for restore.

### Verified on a pu box (KVM) — and falsified
Every reviewer's edits were re-checked on the shipping tree:

| check | gate pid | skew logged | outcome |
| --- | --- | --- | --- |
| **adopt** — correct code | unchanged | n/a | ✅ green (same PTY + scrollback survive) |
| **skew** — correct code | `1284→1528` (recycled) | yes | ✅ green (session preserved) |
| **skew** — `isContractVersionCompatible`→`true` mutation | `1289→1289` (wrongly adopted) | no | ❌ red |
| **vm-test** (pre-existing smoke) | — | — | ✅ green |

Each test is both **green on correct code and red under a deliberate mutation** — the assertions actually bite.

Two real gotchas the mutation steps caught, now handled in `lib.nix`:
- **`machinectl shell` swallows the run command's exit code** (always returns 0) — so the scripts write a **result file** the testScript asserts **as root** (whose exit the driver *does* see). Without this, every assertion was silently ignored and the test could never fail.
- A single-shot check can **race a slow recycle/adopt** — so each verify **polls until all conditions hold at once**; the wrong outcome never satisfies them and times out red.

### Review gauntlet
`/be-review` ran on the diff: **lens-debate** (consensus, extracted the shared `lib.nix`), **codex-debate** (consensus — caught that the pre-existing smoke `vm-test` was dialing the stale `$XDG_RUNTIME_DIR/kolu/pty-host.sock` instead of the per-port `kaval-7681` namespace from #1313, a latent CI-breaker, plus pipefail/SIGPIPE fixes in the `grep` predicates), **simplify**, and **code-police**. See the review comments below.

> VM tests are Linux-only, so macOS launchd survival stays a staged-prod check.


srid:

The **x86_64-linux** lane ran on **`kolu-ci-7`** (`dev-x86-64-linux-05`) — commit `3f67249d`, **passed** ✓

- **Lane wall** (pipeline): **2m42s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 1m55s | ✓ |
| `ci::home-manager` | 1m14s | ✓ |
| `ci::smoke` | 47s | ✓ |
| `ci::nix` | 43s | ✓ |
| `ci::atlas-sync` | 38s | ✓ |
| `ci::flake-check` | 36s | ✓ |
| `ci::unit` | 34s | ✓ |
| `ci::pnpm-hash-fresh` | 33s | ✓ |
| `_ci-setup` | 30s | ✓ |
| `ci::surface-example-build` | 23s | ✓ |
| `ci::surface-app-example-build` | 23s | ✓ |
| `ci::fmt` | 21s | ✓ |
| `ci::biome` | 21s | ✓ |
| `ci::install` | 16s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/3f67249/timings.jsonl`; pool state is a live `flock` probe.</sub>

srid:

✅ **Consensus** after 2 round(s) · lowy + hickey · base `f364d1ef771d`

### Applied (6)
- `lowy-5` flake.nix checks block is now an asymmetric record-merge with awkward nesting — commit `a15452aab`
- `hickey-1` adopt.nix and skew.nix are two copies of one test protocol — commit `4939de001`
- `hickey-2` port "7681" fragmented across four sites with no shared owner — commit `4939de001`
- `hickey-3` jq/curl (and the seed-creates-terminal RPC body) re-bound per file — commit `4939de001`
- `hickey-4` Runtime-layout knowledge (namespace path, gate file, config path) restated in the test instead of read from the app — commit `4939de001`
- `hickey-6` check-name vs file-name vs attr-name use three different spellings of the same concept — commit `395f12017`

### Agreed — no change (5)
- `lowy-1` VM survival-harness scaffolding triplicated across flake.nix / adopt.nix / skew.nix (nix/home/example/adoption/adopt.nix:131-156, nix/home/example/adoption/skew.nix:99-145, nix/home/example/flake.nix:73-131)
- `lowy-2` Tests couple to internal log-string wording rather than an observable contract (nix/home/example/adoption/adopt.nix:113-116, nix/home/example/adoption/skew.nix:73)
- `lowy-3` skew.nix builds koluNew by re-importing default.nix, duplicating the root flake's build wiring (nix/home/example/adoption/skew.nix:32-40)
- `lowy-4` Aggregator default.nix names checks by what-they-test, not by the axis under test (nix/home/example/adoption/default.nix:10-13)
- `hickey-5` contractVersionOverride complects 'which version' with 'how to inject it' in default.nix's postPatch (default.nix:152-160)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 round(s) · codex reviewed at `xhigh` reasoning effort · base `f364d1ef771d`

### Round 1

**codex** — approved: `false`

The branch adds useful VM coverage for adoption/skew, and the shared scaffold is mostly careful about the NixOS test-driver failure modes. I found one CI-breaking stale socket path in the rewritten home-manager smoke test, plus two robustness issues in the surrounding shell checks.

Findings:
- `F1` · blocking · open — The smoke VM test still runs `kaval-tui list --socket "$XDG_RUNTIME_DIR/kolu/pty-host.sock"`, but the current server spawns kaval on the per-port namespace (`$XDG_RUNTIME_DIR/kaval-7681/pty-host.sock`). Because an explicit `--socket` bypasses kaval-tui discovery, this check will keep polling a socket that the service no longer serves and fail `checks.x86_64-linux.vm-test` even when kolu is healthy. (nix/home/example/flake.nix:127)
- `F2` · major · open — This `machinectl shell` call lacks the `</dev/null` redirect that the rest of the new test scaffold documents as load-bearing. If machinectl keeps the test driver's stdin open here, the smoke test can hang before reaching the listener poll. (nix/home/example/flake.nix:100)
- `F3` · minor · open — `grep -c ... || echo 0` produces two zeroes when there are no matches because `grep -c` prints `0` and exits 1. That makes `skewlog`/`sess` contain `0\n0`, so the numeric `-ge` tests emit `integer expression expected` during normal polling until the condition becomes true. (nix/home/example/adoption/skew.nix:75)

**claude** — Reviewed codex's 3 findings on the B3.3 adoption test-infra branch against the source. All three are valid; fixed all three. F1 (blocking): the kolu-service smoke VM test in nix/home/example/flake.nix dialed `kaval-tui list --socket "$XDG_RUNTIME_DIR/kolu/pty-host.sock"`, but the server serves at the per-port `$XDG_RUNTIME_DIR/kaval-<port>/pty-host.sock` (confirmed in packages/server/src/ptyHost/localDriver.ts:54 kavalSocketPath + packages/kaval/src/socketPath.ts kavalNamespace). An explicit `--socket` bypasses discovery (kaval-tui/src/main.ts:123 resolveSocketPath), so it polls a dead socket and fails checks.x86_64-linux.vm-test. The stale line dates to #1301 and was never updated by #1313's per-port change; switched to flag-less `kaval-tui list` (discovery finds the single kaval-7681 daemon), matching the new adoption tests. F2 (major): the `systemctl --user is-active` machinectl call lacked the `</dev/null` + bounded timeout the scaffold treats as load-bearing; added both, noted it is a liveness probe (machinectl masks the inner exit) with the HTTP listener poll as the real assertion. F3 (minor): reproduced that `journalctl | grep -c "contract skew" || echo 0` over a PIPE yields `0\n0` on a miss → `[ -ge 1 ]` throws `integer expected` each poll; replaced count arithmetic with `grep -q` boolean predicates. Codex's claim that the FILE-form `sess` grep also double-zeroes did NOT reproduce on this bash/grep (only the pipe form does), but I converted both for a clean, consistent fix. nixpkgs-fmt clean, nix files parse, embedded shell passes bash -n.

- `F1` **fixed** — Confirmed real CI-breaker: server serves the per-port socket $XDG_RUNTIME_DIR/kaval-<port>/pty-host.sock (packages/server/src/ptyHost/localDriver.ts:54, packages/kaval/src/socketPath.ts:35), and an explicit --socket bypasses discovery (packages/kaval-tui/src/main.ts:127 returns the override verbatim), so the smoke test at flake.nix dialed a socket the server never serves. The stale `--socket "$XDG_RUNTIME_DIR/kolu/pty-host.sock"` was introduced in #1301 and #1313's per-port change left it stale (its own commit msg says flag-less list keeps working via discoverPtyHostSockets). Changed to flag-less `kaval-tui list` and rewrote the now-stale comment to the per-port reality. This branch already restructures the checks block, so fixing it here keeps the smoke test consistent with the adoption tests that already use flag-less discovery.
- `F2` **fixed** — Valid robustness/consistency fix. The is-active machinectl call at flake.nix lacked the `</dev/null` that lib.nix (systemctlUser, runAsAlice) and the smoke test's own kaval-tui line treat as load-bearing (the nixos test driver's stdin pipe never EOFs → a session can hang even after the inner command exits). The command is fast and non-interactive so it hasn't bitten, but leaving one machinectl call inconsistent with the scaffold's documented invariant is exactly the latent hazard the invariant exists to prevent. Added `</dev/null` and a bounded `timeout 30`; updated the comment to note machinectl masks the inner exit so this is a liveness probe and the HTTP listener poll is the real service assertion.
- `F3` **fixed** — Reproduced precisely on this bash 5.3 / GNU grep: `journalctl ... | grep -c "contract skew" || echo 0` over a PIPE yields `0\n0` on a miss (grep -c prints `0` and exits 1, so `|| echo 0` also runs), and `[ "$skewlog" -ge 1 ]` then throws `integer expected` on every poll until the condition holds. Fixed by replacing the count arithmetic with `grep -q` boolean predicates in the polling `if` and in the failure-diagnostics block (skewseen/sessseen yes/no). One correction to codex: the `sess` FILE-form grep (`grep -c "$id" FILE || echo 0`) did NOT double-zero on this platform — it yielded a single `0` (verified with od -c), so only the pipe form (skewlog) actually errored. Converting both to grep -q is the clean fix regardless.

commit: `2c77e93fdb83e11cfb931a763b7c0317b045676f`

### Round 2

**codex** — approved: `false`

CLAUDE fixed the three findings from the previous round. F1 and F2 are clean, and I accept the correction on F3 that the piped `grep -c` case was the actual double-zero source. The current fix introduces one new shell robustness regression in the same skew verifier: `grep -q` is now used in a pipeline while `pipefail` is enabled, which can make a real match report failure if `journalctl` is SIGPIPE'd after grep exits early.

Findings:
- `F1` · blocking · resolved — The smoke VM test was explicitly dialing the stale `$XDG_RUNTIME_DIR/kolu/pty-host.sock` path instead of the current per-port kaval daemon namespace, causing the check to poll a socket the service no longer serves. (nix/home/example/flake.nix:133)
- `F2` · major · resolved — The `machinectl shell` service probe lacked the `</dev/null` redirect and bounded timeout used elsewhere to avoid hanging on the NixOS test driver's stdin pipe. (nix/home/example/flake.nix:105)
- `F3` · minor · resolved — The skew verifier used `grep -c ... || echo 0` count arithmetic, which can produce non-numeric multi-line values such as `0\n0` and make `[ -ge 1 ]` complain during normal polling. (nix/home/example/adoption/skew.nix:79)
- `F4` · major · open — New regression from the F3 fix: the script has `set -o pipefail`, but now checks `journalctl ... | grep -q "contract skew"`. `grep -q` exits as soon as it finds a match; if `journalctl` then writes more output and receives SIGPIPE, the pipeline status can be non-zero despite the match. That can make the skew test time out even though the contract-skew log exists. The same pattern appears in the failure diagnostic at line 88. (nix/home/example/adoption/skew.nix:80)

_codex on the rebuttal:_ F1 and F2: accepted; the fixes address the exact failure modes. F3: accepted; I concede the file-form `grep -c \"$id\" FILE || echo 0` correction and agree the piped journalctl form was the real numeric bug. The original count-arithmetic finding is resolved. New F4 is separate and introduced by the replacement: `grep -q` plus `pipefail` can false-negative on a real piped match when the producer is SIGPIPE'd after grep exits early.

**claude** — Round 2 of the codex debate on the B3.3 adoption test-infra branch (PR #1350). Only one finding was open this round: F4, a regression codex said my round-1 F3 fix introduced. F1/F2/F3 were resolved in round 1 and accepted by codex (responseToRebuttal confirms), so no further change. I agreed with F4 and reproduced it: `grep -q` inside a pipeline under `set -o pipefail` exits on the first match and SIGPIPEs the upstream producer, so the pipeline reports rc=141 even on a REAL match (verified, fires every run with a large producer). That makes the skew verifier's poll if-condition (and the skewseen diagnostic) false-negative and time out red on correct code. Fixed by switching the piped predicates from `grep -q "..."` to plain `grep "..." >/dev/null` — grep drains the producer (no SIGPIPE), leaving grep's own match/no-match (0/1) as the pipeline status. I also fixed the identical hazard in the sibling adopt.nix (the positive test): the pre-restart scrollback check and the verify if-condition's snapshot+journalctl pipes — codex only flagged skew.nix, but leaving a known SIGPIPE flake in the headline test would be inconsistent. File-form `grep -q "$id" FILE` calls (no pipe, no producer to kill) are correctly left untouched, and the `grep -c ... || echo 0` diagnostics in command-substitution are harmless (grep -c reads all input, never early-exits) and kept. Both nix files parse, the fixed if-block was simulated under load and reliably takes the correct branch, and nixpkgs-fmt reports no changes needed.

- `F1` **fixed** — Resolved in round 1 (commit 2c77e93) and accepted by codex this round (status: resolved). Smoke VM test switched from the stale `--socket $XDG_RUNTIME_DIR/kolu/pty-host.sock` to flag-less `kaval-tui list` discovery. No change needed in round 2.
- `F2` **fixed** — Resolved in round 1 and accepted by codex (status: resolved). The is-active machinectl probe got `timeout 30 ... </dev/null` plus a clarifying comment. No change needed in round 2.
- `F3` **fixed** — Resolved in round 1 and accepted by codex (status: resolved). Replaced the `grep -c ... || echo 0` count arithmetic that yielded `0\n0` over a pipe. Codex conceded my correction that the file-form grep did not double-zero. No change needed in round 2.
- `F4` **fixed** — Agreed and fixed. Reproduced reliably: under `set -o pipefail`, `grep -q` exits on first match and SIGPIPEs the producer, so the pipeline returns rc=141 on a REAL match (test ran ELSE-branch rc=141 every iteration with a large producer). In nix/home/example/adoption/skew.nix this made the poll if-condition at line 79-81 and the skewseen diagnostic at line 88 false-negative and time out red on correct code. Fixed both by switching `journalctl ... | grep -q "contract skew"` to `journalctl ... | grep "contract skew" >/dev/null` (grep drains journalctl, no SIGPIPE, pipeline status is grep's 0/1). Also fixed the identical hazard in the sibling nix/home/example/adoption/adopt.nix at lines 65 (pre-restart scrollback check), 96 and 98 (verify if-condition snapshot+journalctl pipes) — codex only named skew.nix but the positive/headline test had the same latent flake. File-form `grep -q "$id" FILE` calls have no pipe/producer to SIGPIPE and are deliberately left untouched; adopt.nix:110 `grep -c ... || echo 0` is in command-substitution diagnostics and harmless (grep -c reads all input, never early-exits).

commit: `4d3afd752f1eb3a08f79ea303b17730cb13bfbad`

### Round 3

**codex** — approved: `true`

The current round resolves the remaining F4 regression without introducing a new debate-scoped issue. The piped predicates in `skew.nix` now use draining `grep ... >/dev/null` under `pipefail`, and CLAUDE also fixed the same `grep -q`/SIGPIPE hazard in `adopt.nix`, which is a valid adjacent cleanup from the same round.

Findings:
- `F1` · blocking · resolved — The smoke VM test was explicitly dialing the stale `$XDG_RUNTIME_DIR/kolu/pty-host.sock` path instead of the current per-port kaval daemon namespace, causing the check to poll a socket the service no longer serves. (nix/home/example/flake.nix:133)
- `F2` · major · resolved — The `machinectl shell` service probe lacked the `</dev/null` redirect and bounded timeout used elsewhere to avoid hanging on the NixOS test driver's stdin pipe. (nix/home/example/flake.nix:105)
- `F3` · minor · resolved — The skew verifier used `grep -c ... || echo 0` count arithmetic, which can produce non-numeric multi-line values such as `0\n0` and make `[ -ge 1 ]` complain during normal polling. (nix/home/example/adoption/skew.nix:75)
- `F4` · major · resolved — The F3 fix briefly introduced `grep -q` in a pipeline under `pipefail`, which can false-negative when the producer gets SIGPIPE after grep exits early on a real match. (nix/home/example/adoption/skew.nix:86)

_codex on the rebuttal:_ F4: accepted and resolved. The current `journalctl ... | grep "contract skew" >/dev/null` form fixes the specific `grep -q`/SIGPIPE/pipefail failure mode I raised. The analogous `adopt.nix` changes are also appropriate because they remove the same newly identified class of flake from the positive test. File-form `grep -q` calls remain fine because there is no upstream pipe producer to SIGPIPE.

srid:

Scope: `git diff f364d1ef...HEAD` (the adoption VM-test diff). Run with `--no-elegance` (the `/simplify` track already ran this gauntlet).

| Pass | Result |
| --- | --- |
| Rules | ✅ clean |
| Fact-check | 6 findings — 4 applied, 2 skipped with reasoning |

**Applied** (`610b6ee0`):

| # | Finding | Action |
| --- | --- | --- |
| F2 | `jq` exit status unchecked before the `curl` POST — a silent jq failure would send an empty body | `adopt.nix`: `\|\| fail "jq failed to build the sendInput request body"` |
| F4 | the seed's PTY-liveness poll timed out with no trace of the last observed state | `adopt.nix`: dump the last `kaval-tui list --json` in the timeout message |
| F5 | `sessseen` (three s's) — not grep-able, diverges from intent | `skew.nix`: → `sessionSeen` (and `skewseen` → `skewSeen`) |
| F6 | the final-state recheck used the bare drain-the-pipe `grep` with no rationale, inviting a future "fix" back to `grep -q` | `skew.nix`: added the same pipefail/SIGPIPE comment the poll carries |

**Skipped** (with reasoning):

- **F1** — sed-metacharacter escaping in `default.nix`'s `contractVersionOverride` postPatch. The value is a controlled build-time version literal (always like `"9.0"`), never external input; and the grep-guard immediately after the sed asserts the exact literal `PTY_HOST_CONTRACT_VERSION = "<value>"`, so any sed corruption fails the build **closed**. The "false-positive clean check" can't occur.
- **F3** — `2>/dev/null` on the config-file `grep` in the verify poll. During polling the file legitimately may not exist yet (a normal "not saved yet" state, not an error), so suppressing stderr there is correct; a `[ -r ]` precheck would wrongly fail on that. The file is alice's own always-readable `~/.config/kolu/config.json`, and the seed's `\|\| fail "never saved"` is the right place to surface a true absence.

All findings re-verified on a KVM pu box: the shipping tree is green on correct code (all three checks) and red under a deliberate `isContractVersionCompatible → true` mutation (`daemon gate 1289 → 1289`, `contract skew logged: no` → fail).

PR #1360, opened 2026-06-14
chore(deps): bump @pierre/diffs to 1.2.10, shiki to 4.2.0

srid:
Bumps the two libraries behind kolu's syntax highlighting to their latest releases:

| Package | From | To | Where |
| --- | --- | --- | --- |
| `shiki` | `^3.23.0` | `^4.2.0` | `solid-markdown` |
| `@pierre/diffs` | `^1.2.1` | `^1.2.10` | `client`, `solid-pierre`, `transcript-html` |
| `@pierre/trees` | `1.0.0-beta.4` | — | already latest, untouched |

## Why the two bumps land together

`@pierre/diffs` takes `shiki` as a **regular dependency**, and our `solid-markdown` highlighter shares that same shiki install. Pierre 1.2.1 pinned `shiki: ^3.0.0`; 1.2.10 widens it to `^3.0.0 || ^4.0.0`. So bumping shiki to 4 **without** bumping Pierre would split the workspace into *two* shiki copies — markdown on 4, Pierre still on 3. Bumping both keeps the workspace on a **single shared shiki**, now `4.2.0` (it was a single `3.23.0` before — this PR doesn't change the copy count, it moves the shared copy from 3 to 4). Pierre's `@shikijs/transformers` helper still resolves to 3.x, dragging a small duplicate `@shikijs/core@3.23.0` alongside the 4.x engine — its transitive choice, not ours.

## shiki 3 → 4 is a non-event for our usage

shiki 4 is a major bump, so the concern was the highlighter contract in `solid-markdown`. It's unchanged:

- `createHighlighter({ themes, langs })` and `codeToHtml(code, { lang, themes: { light, dark }, defaultColor: false })` — same signatures.
- The dual-theme output still emits the `.shiki` class and the `--shiki-light` / `--shiki-dark` CSS variables that `markdown.css` keys off to follow the app's light/dark.
- `github-light` / `github-dark` themes and the curated grammar set are all still present.
- v4's only breaking changes (drops Node 18, removes a typo'd `createdBundledHighlighter` export + an unrelated Twoslash option) touch nothing kolu uses. Our Nix Node is v24.

## Verification

- ✅ `pnpm typecheck` — all packages (type-level proof the shiki/pierre APIs still fit our calls)
- ✅ unit tests — `solid-markdown`, `solid-pierre`
- ✅ empirical `codeToHtml` output check against shiki 4.2.0: confirmed `<pre class="shiki">` + `--shiki-light`/`--shiki-dark` vars
[...]

Issue #809, opened 2026-05-01
Phase 8: wire up Pierre's VirtualizedFileDiff / VirtualizedFile (deferred to #807)


srid:
## Why

Phase 8 of #514 calls out large-file virtualization as a known perf gap: a 50k-line lockfile diff currently puts every hunk row into the DOM at once, stalling the frame. Shiki tokenization (post-#708) made the per-row cost cheaper than the old `highlight.js` path, but row count is still the bottleneck.

`@pierre/diffs` ships virtualized variants alongside the standard renderers — `VirtualizedFile`, `VirtualizedFileDiff`, and the `Virtualizer` they share — so the fix is to wire them through, not to write virtual scrolling from scratch.

## What we found in the spike

Quick read of the type defs (`node_modules/.pnpm/@pierre+diffs@*/node_modules/@pierre/diffs/dist/components/VirtualizedFileDiff.d.ts`) confirms the swap is **not drop-in**.

`FileDiff` constructor:

```ts
constructor(
  options?: FileDiffOptions<LAnnotation>,
  workerManager?: WorkerPoolManager,
  isContainerManaged?: boolean,
);
```

`VirtualizedFileDiff` constructor:

```ts
constructor(
  options: FileDiffOptions<LAnnotation> | undefined,
  virtualizer: Virtualizer,                   // <-- required, shared across files
  metrics?: Partial<VirtualFileMetrics>,
  workerManager?: WorkerPoolManager,
  isContainerManaged?: boolean,
);
```

Plus extra lifecycle methods that callers have to drive: `setVisibility(visible: boolean)`, `reconcileHeights()`, `onRender(dirty: boolean)`. The `render()` signature also differs — `VirtualizedFileDiff.render()` omits some props that `FileDiff.render()` accepts (`containerWrapper`, etc., based on the `Omit`-style typing).

So the wrapper isn't `s/FileDiff/VirtualizedFileDiff/g`. It needs:

1. A `Virtualizer` instance scoped to the right-panel viewport (one per panel, shared across whatever Pierre instances live inside it).
2. Visibility plumbing — when the right panel scrolls or resizes, every `VirtualizedFileDiff` / `VirtualizedFile` in view needs `setVisibility(true)`, others `setVisibility(false)`.
3. A `reconcileHeights()` cadence after content changes (debounced).
4. The slimmer `render()` shape on the wrapper's update path.

## Where this work belongs

[...]

PR #1363, opened 2026-06-14
perf(code): highlight diffs/files in a worker pool, off the UI thread (R2)

srid:
**R2 of the [OpenCode-perf plan](https://github.com/juspay/kolu/blob/master/docs/atlas/dist/opencode-perf.html).** R1 (the Pierre 1.2.10 + Shiki 4.2.0 bump, #1360) gave us the substrate; this PR uses it.

## What

Kolu's Code-tab diff/file viewer (`@pierre/diffs` `CodeView`, wrapped by `@kolu/solid-pierre`) tokenized syntax **on the UI thread**. A large diff — a 50k-line lockfile (#809/#514), or a whole PR's worth of hunks in one virtualized `CodeView` — blocked the main thread while it highlighted, hitching scroll and clicks.

Pierre 1.2.10 ships an internal worker pool (`getOrCreateWorkerPoolSingleton` → `WorkerPoolManager`) but only uses it if you *hand it* to the `CodeView` constructor; Kolu wasn't. This PR does two things, both at the single `CodeView` chokepoint every consumer (`BrowseDiffView`, `BrowseFileView`) flows through:

1. **Off-thread highlighting** — `solid-pierre` owns a session-lived worker-pool singleton (`workerPool.ts`) and passes it to every `CodeView`. Plain (un-highlighted) ASTs still paint synchronously; the worker streams highlighted tokens back and the view repaints. The UI never blocks.
2. **Lighter engine** — `preferredHighlighter: 'shiki-js'` on both the worker and the main-thread path, dropping the Oniguruma WebAssembly payload (the JS regex engine fits an Electron client better).

## Shape / why it's small

| Concern | Decision |
|---|---|
| Where | One chokepoint: `solid-pierre/CodeView.tsx`. No call-site changes. |
| Worker bundling | Kolu supplies the `workerFactory` (`new Worker(new URL('@pierre/diffs/worker/worker.js', import.meta.url), { type: 'module' })`), so the worker bundles through the client's Vite — no guessing the bundler. Module workers code-split their Shiki grammars, so `worker.format: 'es'` in `vite.config.ts`. |
| SSR safety | The pool is created from `CodeView`'s `onMount` → browser-only. `transcript-html`'s static export uses `@pierre/diffs` directly and never spawns a Worker. |
| Lifecycle | The pool is a singleton, never torn down — warm like a cached highlighter, LRU-bounded. A `CodeView` unmount tears down its own instances, not the pool. |
| Pool size | 2 — enough for neighbouring hunks to tokenize in parallel without Pierre's default of 8 Shiki engines on a desktop. |
| No fallback | The 1.2.10 API was read from the installed type defs first, so the wiring is exact — no defensive/runtime fallback machinery. |

## Verification so far

- ✅ `pnpm -r typecheck` — the Pierre worker API fits (`preferredHighlighter`, `getOrCreateWorkerPoolSingleton`, `new CodeView(opts, pool)`).
- ✅ Client Vite build — emits the worker chunk (`worker-*.js`, Pierre + Shiki) with grammar chunks code-split; the `iife`→`es` worker-format gap is fixed.
- ⏳ Evidence (large diff highlights without hitching; theme toggle still works) + CI below.


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `edb731b839aa`

### Applied (5)
- `lowy-1` Engine+theme contract for the highlighter is declared twice and must stay in lockstep by hand — commit `fd90e5a83`
- `lowy-2` Second memoization layer over an already-singleton factory — commit `533e7e842`
- `hickey-1` Highlighter engine is one fact written twice, agreement held by prose — commit `fd90e5a83`
- `hickey-2` Theme set DEFAULT_THEMES duplicated across CodeView options and pool init — commit `fd90e5a83`
- `hickey-3` buildOptions complects per-instance view options with a process-global engine policy — commit `fd90e5a83`

### Agreed — no change (2)
- `lowy-3` No teardown receptacle re-exposed for the session-lived pool (packages/solid-pierre/src/workerPool.ts:38-62)
- `lowy-4` Module-worker 'es' format coupling lives at the app boundary, far from the code that requires it (packages/client/vite.config.ts:71-74)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `edb731b839aa`

### Round 1

**codex** — approved: `false`

The worker-pool integration is broadly sound: it matches Pierre 1.2.10's `CodeView(options, workerManager)` API, the Vite worker URL form is one Vite rewrites, and `tsc --noEmit` passes for both `packages/solid-pierre` and `packages/client`. I found only small documentation/type-hygiene issues, but approval stays false until those are resolved.

Findings:
- `F1` · minor · open — The Atlas frontmatter description still says only R1 shipped and that `R2–R4 remain`, while the body now says R2 shipped in #1363. That stale description is rendered into `docs/atlas/dist/opencode-perf.html` as the page meta description and visible subtitle, so the page contradicts itself before the reader reaches the updated body. (docs/atlas/src/content/atlas/opencode-perf.mdx:3)
- `F2` · nit · open — The section heading still says `Confirmed facts & the open question`, but the section now says the question was answered. This also leaves the generated table-of-contents anchor/title stale. (docs/atlas/src/content/atlas/opencode-perf.mdx:172)
- `F3` · nit · open — `preferredHighlighter: "shiki-js" as HighlighterTypes` uses a type assertion for a literal that TypeScript can validate directly. Because this value is the shared worker/main-thread contract, the assertion would mask a future typo instead of letting the `HighlighterTypes` union catch it. (packages/solid-pierre/src/workerPool.ts:43)

**claude** — Fixed all three codex findings (F1, F2, F3). Committed as 7b348e2.

- `F1` **fixed** — Fixed stale Atlas frontmatter description.
- `F2` **fixed** — Renamed stale section heading.
- `F3` **fixed** — Replaced type assertion with satisfies clause.

commit: `7b348e2f56c4182578ca1fa280849330e7e0a04b`

### Round 2

**codex** — approved: `true`

Verified the current tree against the three existing findings only. CLAUDE's fixes address F1, F2, and F3 in source, and the Atlas generated output reflects the documentation changes. No disputes were raised and I did not find a regression introduced by this round's changes.

Findings:
- `F1` · minor · resolved — The frontmatter description no longer claims R2-R4 remain; it now correctly says the diff-path wins shipped and only R3/R4 remain. (docs/atlas/src/content/atlas/opencode-perf.mdx:3)
- `F2` · nit · resolved — The stale section heading was renamed to reflect that the question is resolved, and the generated Atlas output has the updated heading/anchor. (docs/atlas/src/content/atlas/opencode-perf.mdx:172)
- `F3` · nit · resolved — The highlighter contract now uses `satisfies` against the Pierre types, so the literal is checked without the previous masking assertion. (packages/solid-pierre/src/workerPool.ts:46)

srid:

Scoped to the R2 diff (base `edb731b`), `--no-elegance` (the simplify track already ran this tree).

| Pass | Result |
|---|---|
| **Rules** | ✅ Clean — all 27 built-in + project rules passed with evidence |
| **Fact-check** | 2 findings, both verified **already-handled** against Pierre's compiled source → documented, no code change (commit `b79b474`) |
| **Elegance** | Skipped (`--no-elegance`) |

### Fact-check — both resolved by verification, not by adding guards

1. **Singleton contract** — flagged: a fresh config + `workerFactory` closure per `getCodeViewWorkerPool()` call might spawn a new pool per mount. **Verified false:** `getOrCreateWorkerPoolSingleton` is `pool ??= new WorkerPoolManager(...)` module-side — the config is consumed only on the first call; every later call returns the same warm instance and ignores its args. No per-mount pool. Comment tightened to state the `??=` fact.
2. **Unhandled `new Worker()` failure** — flagged: a Worker that can't construct (CSP, workers disabled) fails silently/undocumented. **Verified handled:** Pierre wraps `workerFactory()` in try/catch, sets `workersFailed`, and the view degrades to un-highlighted (plain-AST) rendering — content still paints. A duplicate try/catch would only mirror Pierre's handling; documented the degrade path instead.

srid:
## Evidence

R2 moves diff/file syntax highlighting off the UI thread — by design the result looks **identical**, so the proof is *"highlighting still renders correctly through the new Pierre worker path, no regression."* Captured by driving a live kolu (`just dev-auto`, chrome-devtools MCP) to the Code tab's **Branch** view on this very PR's branch, rendering `packages/solid-pierre/src/workerPool.ts` — the file is fully syntax-highlighted (keywords, strings, comments, types all colored, per-token) by a `CodeView` that now holds the shared `WorkerPoolManager`.

**The highlighted diff (Pierre `CodeView`, worker path):**

R2 syntax-highlighted diff

**In context — the live kolu Code tab:**

R2 in the live app

**Why this proves it:** Pierre renders the plain (un-highlighted) AST first, then the worker streams highlighted tokens back — so colored tokens on screen mean the worker tokenized successfully. The Shiki TypeScript grammar loaded on demand to tokenize the `.ts` file, and the off-thread path is the `worker-*.js` chunk the production build emits (Pierre worker + Shiki, grammars code-split). 

**Coverage note:** this change is viewport- and pointer-independent (it changes *where* tokenization runs, not layout or hit targets), so a single capture is representative; the e2e Code-tab scenarios — which click files and assert rendered content — pass on `ci::e2e@x86_64-linux`, exercising the same `CodeView` worker path end-to-end (e.g. *"selected file should show content"* ✔ across multiple repos).

PR #1366, opened 2026-06-14
Stop the "N terminals reattached" toast re-firing on every reload

### LINKED ISSUE #1365: Spurious "N terminals reattached" toast re-fires on any full page reload (desktop hard-refresh / mobile reconnect; since B3.3)
# Spurious "N terminals reattached" toast on every mobile reconnect

**Severity:** low (cosmetic false positive, no data loss) · **Scope:** any full page reload, every platform — iOS Safari hits it *involuntarily* (backgrounded-tab eviction → reload); **desktop reproduces it deterministically on a hard refresh** · **Introduced by:** #1344 / `a076b4f9` — *"Terminals survive a kolu update: kaval adoption (B3.3)"* (the "Kaval update")

## Symptom

On mobile (iPhone), after Kolu shows the **"Disconnected…"** overlay and then reconnects, it pops a toast **"N terminals reattached"** — even though kaval/kolu was running the whole time and no terminals were ever lost. It recurs on essentially every reconnect.

It is **not mobile-specific**: it also fires on **desktop** after a full page refresh (user-confirmed). The common trigger is a fresh JS context, not the platform — see §5. Desktop only escapes it on a *transparent* WebSocket reconnect (no reload).

> The toast text is literally `N terminal(s) reattached` — `packages/client/src/kaval/useDaemonStatus.ts:219`. (The user paraphrased it as "N terminals are restored".)

## TL;DR root cause

The adoption count is **sticky server snapshot state** that the `daemonStatus` collection replays to **every fresh subscriber**, while the client's "show this once" guard is a **module-level boolean that only lives as long as the JS context**. On iOS Safari a backgrounded tab is evicted and **reloaded** on return → fresh JS context → the boolean resets → the replayed-but-stale `adopted` count re-fires the toast, though no new adoption happened.

It's a **one-time event** (adoption happened at a past server boot) modeled as **durable state** (current daemon status). Snapshot collections are *defined* to replay to every new subscriber, so the announcement gets re-delivered; the once-per-JS-context latch is the only thing suppressing it, and mobile reload defeats it.

### 1. A real adoption stamps `adopted: N` into the server's persistent status store — once, at boot

`packages/server/src/terminalBackend/reattach.ts:84-85` (in `adoptSurvivingSession`, run only when boot adopted a surviving daemon):

```ts
if (adoptedCount > 0) {
  setAdoptedCount(LOCAL_HOST_ID, adoptedCount);
```

`setAdoptedCount` is **only ever called here** (verified by grep), and `adoptSurvivingSession` runs **only at boot**, via `onAdopted` when `ep.adoptOrEnsure()` returned `true` — `packages/server/src/ptyHost/index.ts:145-148`.

### 2. `adopted` is folded into a module-level `store` Map and is **never cleared**

`packages/server/src/ptyHost/daemonStatus.ts:15` — the authority:

```ts
const store = new Map<string, DaemonStatus>();
```

`packages/server/src/ptyHost/daemonStatus.ts:60-64`:

```ts
export function setAdoptedCount(hostId: string, adopted: number): void {
  const current = store.get(hostId);
  if (!current) return;
  publishDaemonStatus(hostId, { ...current, adopted });
}
```

`publishDaemonStatus` (`:42-51`) `store.set(hostId, full)`s that value. Nothing anywhere deletes/nulls `adopted`. It stays baked into the host's *current* status for the **entire life of the server process**. (It is only ever displaced if the endpoint later publishes a *fresh* status object — which carries no `adopted` — but during a stable adopt-boot the daemon just sits `connected` and no such transition occurs.)

### 3. The collection replays that stored value (incl. `adopted: N`) to **every (re)subscriber**

`daemonStatus` collection wiring — `packages/server/src/surface.ts:196-204`:

```ts
daemonStatus: {
  readAll: () => readDaemonStatuses(),
  readOne: (key) => readDaemonStatus(key as string),  // => store.get(hostId)
  ...
}
```

The surface framework's snapshot-then-deltas invariant — `packages/surface/src/server.ts:140-142`:

> *"Snapshot+deltas invariant on `get`: yields `store.get()` first, then … `get` on **every reconnect**, so the first frame must be a fresh snapshot"*

…and the per-key `get` generator that implements it — `packages/surface/src/server.ts:220-227`:

```ts
get: async function* ({ input, signal }) {
  const initial = readOne(input.key);
  ...
  yield initial;          // <- the sticky {state:connected, adopted:N} snapshot
  for await (const v of deps.keysBus...) yield v;
```

So **every fresh WebSocket subscription** (every reconnect, every reload) re-receives `adopted: N`. This is correct, by-design behavior for a snapshot collection — and exactly why the announcement *needs* to be deduped on the client, not on the server (a reconnecting client may have been offline at adoption time and must still learn about it).

### 4. The client guard is a module-level boolean that resets on page reload

`packages/client/src/kaval/useDaemonStatus.ts:211-222`:

```ts
let reattachToastFired = false;
createRoot(() => {
  createEffect(() => {
    if (reattachToastFired) return;
    const status = localDaemonStatus();
    const n = status?.state === "connected" ? (status.adopted ?? 0) : 0;
    if (n > 0) {
      reattachToastFired = true;
      toast.info(`${n} terminal${n === 1 ? "" : "s"} reattached`);
    }
  });
});
```

`reattachToastFired` is a module-scoped `let`. Nothing ever resets it to `false` at runtime — it only returns to `false` when the **module is re-evaluated**, i.e. the JS context is recreated (a full page load). It is **not** reset by a transparent WebSocket reconnect.

The effect **does not consult lifecycle** (`reconnected` vs `restarted`) or any per-adoption identity — the boolean is the sole guard.

### 5. Any full page reload recreates the JS context → fresh boolean → re-fire

A full page load re-evaluates the client module, so `reattachToastFired` is initialized to `false` again. The reconnect then replays the sticky `adopted: N` snapshot (steps 2–3) and the effect fires the toast again (step 4). **The trigger is the JS-context reset, not the platform** — anything that reloads the page reproduces it:

- **Desktop, deterministic — user-confirmed:** a manual full/hard refresh re-fires the toast every time. This directly confirms the mechanism (it was the one link previously marked "inferred"; it is now observed).
- **iOS Safari, involuntary & repeated:** Safari freezes/discards a backgrounded tab and **reloads the page** on return, so the same reset happens on essentially every background→return cycle without any user action — which is why mobile users hit it constantly.

**What desktop escapes is only the *transparent reconnect* path:** a WebSocket drop+reconnect without a reload keeps the JS context, so `reattachToastFired` stays `true` from the original boot and `if (reattachToastFired) return;` short-circuits the replay. A reload (desktop or mobile) defeats that, because the guard lives only in memory. The whole chain is now proven from code + direct observation.

### Why "since the Kaval update"

Before #1344 / `a076b4f9` (B3.3) there was no `adopted` field and no reattach toast, so there was nothing to replay. The schema field, the server fold, and the client toast all arrived together in that PR — `packages/common/src/surface.ts:623-629`:

```ts
/** B3.3: how many terminals this boot ADOPTED from a surviving daemon — set
 *  only on the `connected` status of an adopt-boot ... Drives the client's
 *  one-shot "N reattached" confirmation. ... Optional + additive ... */
adopted: z.number().optional(),
```

### Why it's a false positive

The toast claims a *reattachment event*, but on a routine reconnect nothing was re-adopted — the same long-past `adopted: N` is simply being replayed. The terminals were live the entire time, which is exactly what the user observed.

## Ruled-out alternative (the *other* toast)

There is a second, similar-sounding toast — the restore-card success message **"Restored N terminals…"** / **"Session restored"** at `packages/client/src/terminal/useSessionRestore.ts:299-304`. It is **not** the culprit:

- It fires only from `handleRestoreSession`, which requires the **empty-state restore card** — i.e. `store.terminalIds().length === 0` (`useSessionRestore.ts:146`). The user's terminals were never lost, so that path isn't reached.
- It requires a user **click** on the restore card (or an explicit import), not an automatic reconnect.
- Its wording differs from the reported toast.

(Adversarially checked: a 4-lens diagnostic workflow — server-snapshot, client-latch, refuter, timeline — returned **4 support / 0 refute**, and the refuter specifically traced and dismissed this restore-card path.)

## Reproduction

**Precondition (both paths):** deploy a kolu-server update that **does not change kaval's source** (wire-compatible) so the boot **adopts** the surviving daemon with ≥1 live terminal. (This is the narrow precondition — most deploys change both and silently skip adoption via `adoptOrEnsure() === false`.) The first post-update reconnect correctly shows "N terminals reattached" — that's the intended one-shot confirmation.

**Desktop — deterministic (simplest):**

1. With the server still in its adopted boot, do a **full/hard page refresh**.
2. **Bug:** "N terminals reattached" fires again on the fresh load, despite no new adoption. Reproduces on every hard refresh.

**Mobile (iOS Safari) — involuntary & repeated:**

1. Background the Kolu tab long enough for Safari to discard it.
2. Return to the tab → "Disconnected…" → reconnect (Safari has reloaded the page).
3. **Bug:** "N terminals reattached" fires again. Repeats on essentially every background→return cycle, with no user action.

## Coverage gap

No test exercises *"the reattach toast fires at most once across a reconnect/reload."* `packages/client/src/kaval/` has no `useDaemonStatus` test; the adoption/reconcile logic is unit-tested server-side, but the client toast's lifecycle across re-subscription is not. A fix should lock this with a unit test on the dedupe decision.

## Proposed fix (pending approval — not yet implemented)

The structural fix is to give the adoption a **stable per-adoption identity** carried in the snapshot, and dedupe the client toast against a **persisted** copy of it (so the guard survives a JS-context reload), replacing the fragile module boolean. Requirements the fix must hold: announce once per real adoption (R1); survive the client being **offline at adoption time** — so the identity must ride the snapshot, not a live-only event (R2); **no re-fire on reload** (R3, the bug); re-announce on a genuinely new later update (R4); desktop unchanged (R5); sane across tabs/clients (R6).

Three candidates were evaluated against R1–R6:

- **A (recommended).** Add additive optional `adoptedAt: number` to `DaemonStatusSchema`, stamped `Date.now()` inside `setAdoptedCount` (the one site that surfaces an adoption). Client persists the last-announced `adoptedAt` (via `persistedPref`/`makePersisted`) and toasts only when the snapshot's `adoptedAt` exceeds it. Identity is minted **at the adoption event**, so it satisfies R4 with no cross-signal race. Additive optional field ⇒ no contract bump; not a `packages/surface*` change ⇒ no drishti mirror PR.
- **B.** Same shape but reuse the server `processId` as the adoption id (couples "which boot" with "which adoption" — works today, future-fragile if one process ever adopts twice).
- **C.** Client-only, no wire change: dedupe on `serverProcessId()`. Works, but races the identity probe against the snapshot and is a footgun if a later refactor keys on an in-snapshot field instead (`startedAt`/`identity.staleKey` both **fail R4** — an adopted daemon keeps its old `startedAt`, and adoption only happens when the kaval build is unchanged, so `staleKey` is constant across adoptions).

Recommended files for **A**: `packages/common/src/surface.ts` (field), `packages/server/src/ptyHost/daemonStatus.ts` (stamp), `packages/client/src/kaval/useDaemonStatus.ts` (persisted dedupe, drop the boolean) + a new pure `reattachAnnounce.ts` decision module mirroring `kavalCurrency.ts`, with a unit test.

---
*Diagnosed via a 4-lens adversarial workflow (server-snapshot replay, client-latch/reconnect, refuter, timeline+tests). Verdict: 4 support / 0 refute; restore-card alternative explicitly refuted. Confidence 0.92–0.98 per lens.*

srid:
After a kolu update that keeps your terminals running (B3.3 adoption), kolu shows a brief **"N terminals reattached"** note — meant to fire once. It was instead re-firing on **every reconnect that follows a page reload**: constantly on iPhone, where Safari evicts a backgrounded tab and reloads it on return, and on any desktop hard refresh — even though nothing had been re-adopted and your terminals were never lost. Root cause and full evidence in #1365.

### Why it fired repeatedly

The announcement is a one-time *event* modeled as durable *snapshot state*, guarded by a latch that only lived as long as the JS context:

```
server boot adopts kaval ─▶ setAdoptedCount stamps `adopted:N` into the
  daemonStatus store (never cleared, sticky for the process's life)
        │
        ▼  collection replays its snapshot to EVERY (re)subscribe  (by design)
reconnect delivers the same `adopted:N` ───────────────────────────┐
                                                                    ▼
client guard = a module-level boolean  ──reload resets it──▶  toast re-fires
```

Desktop survived a *transparent* WS reconnect (the boolean held in memory); a reload — mobile-Safari tab eviction or a desktop hard refresh — recreated the JS context, reset the boolean, and the replayed count re-announced.

### The fix

Give the adoption a **stable per-adoption identity** and dedupe the client toast against a **persisted** copy of it, so the guard survives the reload that defeated the boolean:

- **Server** stamps `adoptedAt` (ms-epoch) in `setAdoptedCount` — the one site an adoption is surfaced, so it's a true per-adoption identity. Rides the same sticky snapshot, so a client that was *offline at adoption time* still learns about it on reconnect.
- **Client** keeps the greatest announced `adoptedAt` in `localStorage` (`persistedPref`) and toasts only a strictly newer one. A replayed `adoptedAt` is silent across any reload; a genuinely newer adoption (a later update) announces again.
- The decision is a pure, unit-tested `shouldAnnounceReattach` (mirrors `kavalCurrency`), replacing the fragile module boolean.

The `adoptedAt` field is additive + optional on `DaemonStatusSchema` — no contract bump.

### Reproduction → regression lock

`reattachAnnounce.test.ts` pins the truth table. The reload case was **red on today's logic** (a replayed `adoptedAt` re-announces → `expected false, got true`) and goes **green** with the high-water-mark gate. The defect is pure client dedupe logic, so a deterministic unit test captures it faithfully; an end-to-end reproduction would need new *restart-the-server-while-kaval-survives* harness infra plus a flake-prone "assert toast absence after reload" — the real reload-no-refire behavior is demonstrated in `## Evidence` instead.

Fixes #1365


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `1c8f17705427`

Independent findings: lowy=5, hickey=2

### Applied (3)
- `lowy-1` Wire re-exposed: status fields re-read after the receptacle already proved them — commit `dadc2942b`
- `lowy-5` Effect writes a signal it also tracks — self-retrigger is contained but worth noting — (uncommitted)
- `hickey-2` The effect re-proves and re-projects what the predicate already established — (uncommitted)

### Agreed — no change (4)
- `lowy-2` adoptedAt is variable, not volatile — correctly an additive attribute, no new boundary (packages/common/src/surface.ts:638)
- `lowy-3` adoptedAt stamped at the single adoption-surfacing site keeps identity-minting in one place (packages/server/src/ptyHost/daemonStatus.ts:68)
- `lowy-4` Dedupe policy correctly encapsulated client-side behind the established kavalCurrency receptacle (packages/client/src/kaval/reattachAnnounce.ts:29)
- `hickey-1` An adoption is fragmented into two parallel optional fields (packages/common/src/surface.ts:629-638)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `1c8f17705427`

### Round 1

**codex** — approved: `false`

The change correctly moves the reload/reconnect dedupe from a module-local latch to a persisted high-water mark, and the server/client wiring is coherent for the normal adoption path. I do see two issues worth addressing: the implementation relies on `Date.now()` as a monotonic per-adoption identity even though the comparison is a strict high-water mark, and the new test only covers the comparator rather than the persistence behavior that actually fixes the reload regression.

Findings:
- `F1` · minor · open — `adoptedAt` is treated as a monotonic per-adoption identity, but it is minted from `Date.now()`. Wall-clock time can move backwards, and equal millisecond stamps can collide. In either case, a later real adoption can have `adoptedAt <= lastAnnouncedAt`, and the client will suppress the toast as if it were a stale replay. (packages/server/src/ptyHost/daemonStatus.ts:68)
- `F2` · minor · open — The regression test covers `reattachToAnnounce`'s truth table, but not the persistence wiring that makes reload dedupe work: reading the localStorage-backed high-water mark, writing the announced `adoptedAt`, and suppressing the same snapshot in a fresh client context. A future change could remove or mis-key `setReattachAnnouncedAt` while this test still passes. (packages/client/src/kaval/reattachAnnounce.test.ts:14)

**claude** — Addressed both of codex's minor findings on the reattach-toast dedupe (#1366) in one round.

F2 (test only covers the comparator, not the persistence wiring): FIXED. Extracted the effect's commit-then-toast glue into a testable `announceReattach` helper in reattachAnnounce.ts and added two deterministic tests over the REAL `persistedPref` backed by a fake `Storage` (same key/fallback/parse the app uses) — one pins commit-before-notify + same-snapshot silence, one replays the sticky adoption from a fresh signal (the reload) and asserts no toast, then a newer adoptedAt still announces. A dropped/mis-keyed persist now fails a test.

F1 (`adoptedAt` from `Date.now()` not monotonic): PARTIAL. Conceded and documented the one real residual (a backward wall-clock step across a redeploy can skip one informational toast — nothing lost), but disputed both of codex's specific claims: the equal-ms collision is impossible (setAdoptedCount fires at most once per server process boot — index.ts:304 → reattach.ts:85 → daemonStatus.ts:65, two adoptions are separate process lifetimes split by a restart), and the suggested opaque-id + seen-id-set remedy regresses two deliberate properties (unbounded localStorage growth; loses the "stale older replay stays silent" the high-water mark gives for free). No server-side monotonic guard is possible because the store is in-memory and empty on the next boot.

Verification: typecheck clean (0 TS errors); full client unit suite green (333 tests, incl. the 2 new persistence-wiring tests); just fmt run. Commit 89101432.

- `F1` **partial** — Wall-clock-from-Date.now() concern. FIXED the documentation/clarity part and CONCEDE the residual; DISPUTE the two specific sub-claims. (1) Equal-ms collision is impossible: setAdoptedCount fires at most ONCE per server process — it's wired as boot onAdopted (packages/server/src/index.ts:304 → adoptSurvivingSession → packages/server/src/terminalEndpoint/reattach.ts:85), and two adoptions are two process lifetimes separated by a restart, so they cannot alias a millisecond. (2) The opaque-id + client seen-id-set remedy is a net regression: it grows localStorage unbounded and forfeits the deliberate 'a stale older replay stays silent' property the strict '>' high-water mark gives for free (author's deliberate-decisions note). (3) CONCEDED residual: a wall-clock step BACKWARD across a redeploy can let a genuine new adoption's adoptedAt fall at-or-below the persisted mark, skipping ONE informational toast for that boot — nothing is lost. No server-side monotonic guard is feasible because the store is in-memory and empty on the next process boot (no prior stamp to enforce monotonicity against). Documented this accepted cost at the stamp site in packages/server/src/ptyHost/daemonStatus.ts:65-78.
- `F2` **fixed** — Test only covered the comparator, not the persist-then-suppress wiring. Extracted the effect's glue into announceReattach() in packages/client/src/kaval/reattachAnnounce.ts (decision → commit proven adoptedAt to the mark FIRST → notify), and pointed useDaemonStatus.ts's effect at it (behavior unchanged, typecheck clean). Added a new describe block in packages/client/src/kaval/reattachAnnounce.test.ts with two deterministic tests over the REAL persistedPref backed by an injected fake Storage, built with the SAME key ('kolu.kaval.reattachAnnouncedAt'), fallback (0), and parse the app uses: test 1 asserts notify fires once, the mark is written THROUGH to storage under that exact key (a mis-key fails here), and a same-snapshot re-emit stays silent; test 2 reproduces the reload — a BRAND-NEW signal reads the surviving mark from storage and replays the same sticky adoptedAt in silence, then a genuinely newer adoptedAt still announces. A future change that drops or mis-keys setReattachAnnouncedAt now fails a test instead of passing unnoticed. All 10 file tests + 333 client suite tests green.

commit: `89101432a323d94384698691574852f96ad56fd1`

### Round 2

**codex** — approved: `true`

Both prior findings are closed. The new `announceReattach` helper and fake-storage tests cover the persistence path I asked for, and I accept CLAUDE's narrowed treatment of the `Date.now()` risk as a documented, non-data-loss edge case for an informational toast rather than a defect that needs more state machinery.

Findings:
- `F1` · minor · resolved — Previously, `adoptedAt` was minted from `Date.now()` while the client treated it as a strict monotonic high-water mark. CLAUDE documented the remaining backward-clock edge and justified why stronger server-side monotonicity is not available without persisting extra server state across process lifetimes. (packages/server/src/ptyHost/daemonStatus.ts:67)
- `F2` · minor · resolved — Previously, tests covered only the pure comparator, not the persisted marker wiring that fixes reload dedupe. The new tests exercise the real `persistedPref` path with fake `Storage`, verify the exact storage key is written, suppress same-snapshot re-emits, and model a fresh context replay after reload. (packages/client/src/kaval/reattachAnnounce.test.ts:136)

_codex on the rebuttal:_ F1: I accept the dispute. The equal-millisecond concern is not a meaningful bug for this call path because `setAdoptedCount` is only reached from boot adoption and is not stamped repeatedly inside one process. The backward wall-clock case still exists, but the updated comment states the exact consequence: one skipped informational toast, not lost terminals or stale UI state. I also accept that a server monotonic counter would require cross-process persistence, and that a seen-id set would add lifecycle/storage complexity for a low-value notification. Marking resolved. F2: The fix addresses the gap. The helper keeps commit-before-notify testable, and the new storage-backed tests would fail on a dropped persist, wrong key, or reload replay regression. Marking resolved.

srid:

Ran after lens-debate + codex-debate + simplify, scoped to this PR's diff.

| Pass | Result |
| --- | --- |
| **Rules** | ✅ Clean — all 20+ built-in + project rules checked, no violations |
| **Fact-check** | 1 finding, fixed below |
| **Elegance** | Skipped (`--no-elegance` — the simplify track already ran over this tree) |

**Fixed — silent error swallowing on a corrupt persisted mark** (`useDaemonStatus.ts`)

`persistedPref("kolu.kaval.reattachAnnouncedAt")` had no `onInvalid`, so a corrupted stored value would `parse`-throw, fall back to `0`, and vanish with no trace — against *fail-loud-over-fail-silent*. Resetting the mark is benign (at worst the next adoption re-announces once, never spurious), so the fix logs a `console.warn` rather than a user-facing toast — proportionate to a recoverable reset (`useComments` toasts because a wiped queue is real data loss; this isn't). Committed as `5fc7b2e9`.

srid:

The **x86_64-linux** lane ran on **`kolu-ci-7`** (`dev-x86-64-linux-05`) — commit `5fc7b2e9`, **passed** ✓

- **Lane wall** (pipeline): **2m51s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 2m1s | ✓ |
| `ci::smoke` | 46s | ✓ |
| `ci::nix` | 45s | ✓ |
| `ci::unit` | 30s | ✓ |
| `_ci-setup` | 29s | ✓ |
| `ci::pnpm-hash-fresh` | 22s | ✓ |
| `ci::install` | 21s | ✓ |
| `ci::biome` | 16s | ✓ |
| `ci::fmt` | 15s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/5fc7b2e/timings.jsonl`; pool state is a live `flock` probe.</sub>

srid:
## Evidence

Driven on a **live kolu** through a *real* B3.3 adoption — opened a terminal, restarted the kolu **server** while its **kaval** daemon (and the terminal's PTY) survived, so the fresh server adopted it (`adopted surviving terminals after restart {adopted:1}` in the server log) and stamped `adoptedAt`. This is exactly the "your terminals survived a kolu update" path that produces the toast.

reattach toast fix — 3 frames

| Step | What happens | Observed |
| --- | --- | --- |
| **1** | A terminal is running; kaval holds its PTY | tile on canvas |
| **2** | Server restarts → adopts the surviving kaval → client reconnects (fresh context, `localStorage` mark empty) | **`ⓘ 1 terminal reattached`** fires **once**; `kolu.kaval.reattachAnnouncedAt` ← `1781471560196` (= the server's `adoptedAt`) |
| **3** | Reload again — the **same** sticky `adopted:1 / adoptedAt:1781471560196` snapshot is replayed to a brand-new JS context | **no toast** — `reattachToastFired: false`, mark unchanged. *This is the bug scenario; pre-fix the in-memory boolean reset on the reload and re-fired here.* |

The contrast is the fix in one line: the dedupe key now lives in `localStorage`, so it **survives the page reload** that reset the old module boolean — `adoptedAt > lastAnnouncedAt` is `1781471560196 > 1781471560196 = false`, so the replay stays silent. Clearing the mark and reconnecting makes the toast fire again (frame 2), proving the persisted mark is what gates it.

_Capture note: the reattach toast fires the instant `daemonStatus` arrives — on a fast localhost **reload** that beats the `<Toaster>` mount, so frame 2 was taken under a network throttle (Toaster mounts first) with the rendered toast element pinned for the screenshot. The dedupe behavior itself (frames 2 vs 3, and the persisted-mark values) is real and unthrottled. The reload-replay regression is also locked deterministically in `reattachAnnounce.test.ts` (red on the pre-fix logic, green after)._

Issue #1308, opened 2026-06-12
Mitigate compositor stress and Wayland geometry instability under heavy canvas load


ic4-y:
## Problem

GNOME Shell (Mutter 49.2, Wayland) crashes with SIGABRT while kolu is running in Chromium. The crash is reproducible under sustained canvas load with multiple active terminals.

**Crash evidence** (`coredumpctl`):
- **Signal**: SIGABRT at 20:19:47
- **Crashing thread** stuck in Mesa/Gallium3D AMD driver (`si_flush_all_queues` → `do_futex_fence_wait`) during a Mutter frame buffer swap
- **Pre-crash journal** shows 5 repeated Mutter assertion failures over 13 minutes:
  ```
  meta_window_set_stack_position_no_sync: assertion 'window->stack_position >= 0' failed
  ```
- **Post-restart journal** shows Chromium sending invalid Wayland geometry:
  ```
  Client provided invalid window geometry for xdg_surface (chromium-browser)
  Buggy client (chromium-browser) committed initial non-empty content without acknowledging configuration
  ```
- Chromium's own diagnostics logged a **143-second compositor hang**:
  ```
  CompositorAnimationObserver is active for too long (142.995s) location=TabIcon
  ```

> **Note:** The core dump and journal excerpts referenced below were captured on the reporter's machine. Other participants experiencing similar crashes should capture their own via `coredumpctl list --since "1 hour ago"` and `journalctl --user -u gnome-shell --since "30 minutes ago"`. The core dump file is located at a local path on the reporter's system (`/var/lib/systemd/coredump/core.\x2egnome-shell-wr.1000.*.3158.*.zst`, 75.4 MB) — please reach out if you need it for analysis.

## Diagnosis

The crash is a compound failure driven by three kolu rendering patterns that together overwhelm Chromium's compositor and produce geometry changes Mutter cannot track on Wayland.

### 1. Many concurrent `infinite` CSS animations with no `will-change` hints

Kolu runs multiple `animation: ... infinite` CSS animations per active terminal, all paint-heavy, none promoted to compositor layers:

| Animation | Location | Cost |
|---|---|---|
| `dock-chip-spin-glow` — rotating `box-shadow` offset | `packages/client/src/index.css:339` | Paint every frame per working chip |
| `dock-chip-breathe` — `box-shadow` breathing | `packages/client/src/index.css:333` | Paint every frame per dock chip |
[...]

PR #1368, opened 2026-06-14
perf(canvas): rAF-coalesce wheel pan/zoom write-storm (R3+R4)

srid:
Closes the loop #1308 left open. That investigation measured the per-wheel viewport write-storm at native speed and deferred it as *latent/benign*, naming the exact condition to re-test: **"a much weaker client."** This PR builds the harness for that case (**R3**) and ships the fix it gates (**R4**) — the last two items of the [OpenCode-perf plan](https://github.com/juspay/kolu/blob/perf-gesture-p99-harness/docs/atlas/dist/opencode-perf.html) (R1 #1360, R2 #1363 already shipped).

## R3 — the harness (`docs/perf-investigations/scripts/gesture-p99/`)
Dependency-free CDP over Node's built-in `WebSocket`. It measures **per-event main-thread work** via a synchronous `WheelEvent` burst — because a headless/CDP Chrome's rAF isn't vsync-capped (it fires ~1:1 with input dispatch, verified headless *and* headful-under-xvfb), so an rAF-paced fling can't show coalescing there, and Chrome 143 dropped `HeadlessExperimental.beginFrame`. Methodology write-up: [`canvas-gesture-p99.md`](https://github.com/juspay/kolu/blob/perf-gesture-p99-harness/docs/perf-investigations/canvas-gesture-p99.md).

## R4 — the fix (`useCanvasViewport.ts`)
Accumulate each frame's pan delta (sum) and zoom factor (product toward the last anchor); apply once per `requestAnimationFrame` instead of per raw wheel event. `gestures.ts` (ownership / `preventDefault`) stays synchronous and untouched — only the state-write defers.

**Provably feel-neutral:** the batched apply lands on the *same* per-frame pan+zoom the per-event path reached (`applyGestureBatch` telescopes the math). Canvas gestures have no e2e coverage, so that equivalence is pinned in `transforms.test.ts`.

## Measured (16 tiles, 60-event burst)
| throttle · gesture | main-thread block (master → R4) | tile writes (master → R4) |
|---|---|---|
| 6× zoom | **865.5 ms → 1.3 ms** | 2,880 → **48** (60×) |
| 6× pan  | 283.2 ms → 1.0 ms | 960 → **16** |
| 4× zoom | 534.9 ms → 1.0 ms | 2,880 → **48** |
| 1× zoom | 123.4 ms → 0.6 ms | 2,880 → **48** |

The write ratio is exactly K=60 (60 events → 1 apply) — that *is* the coalescing. The worst case (zoom fling on a 6× client) went from a ~865 ms freeze (≈52 dropped 60 Hz frames) to imperceptible.

## Test plan
- `transforms.test.ts` — 4 tests pin the batch↔per-event equivalence (pan sum, zoom telescoping, mixed-frame order, identity). ✅
- `just check` (typecheck + lint) clean; existing canvas e2e guards no regression in CI.
- Re-run the harness: `just dev-auto` then `bash docs/perf-investigations/scripts/gesture-p99/run.sh`.


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `1c8f17705427`

### Applied (2)
- `lowy-1` Two parallel "stop the competing input source" verbs must be co-invoked by hand at every programmatic setter — the arbitration policy isn't encapsulated — commit `5517074df`
- `hickey-1` The pending-gesture batch is fragmented into 5 module-level scalars plus a parallel GestureBatch type that re-unifies them every frame — commit `9f9d718ef`

### Agreed — no change (6)
- `lowy-2` rAF scheduling is hand-rolled a third time in the same subsystem with a divergent cancel idiom (packages/client/src/canvas/viewport/useCanvasViewport.ts:63-66,90-98)
- `lowy-3` discardPendingGesture conflates two independent triggers (container swap vs. authoritative write) behind one doc-comment (packages/client/src/canvas/viewport/useCanvasViewport.ts:87-98)
- `lowy-4` applyGestureBatch correctly plugs into the transforms.ts receptacle — no wire exposed (packages/client/src/canvas/viewport/transforms.ts:65-114)
- `lowy-5` gestures.ts receptacle kept intact — the deferral lands on the consumer side, not by reshaping the socket (packages/client/src/canvas/viewport/useCanvasViewport.ts:152-168)
- `hickey-2` gestureRaf overloads a scheduling handle as a boolean 'is-flush-pending' flag, complecting identity with presence (packages/client/src/canvas/viewport/useCanvasViewport.ts:63-66)
- `hickey-3` discardPendingGesture bundles two separable concerns: cancel the scheduled flush AND zero the accumulator (packages/client/src/canvas/viewport/useCanvasViewport.ts:90-98)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 round(s) · codex reviewed at `xhigh` reasoning effort · base `1c8f17705427`

### Round 1

**codex** — approved: `false`

The rAF batching is a solid direction, and the hot-path shape is appropriately small, but the current batching is not actually equivalent to the old per-event semantics in two edge cases. The biggest issue is zoom clamping: multiplying factors and clamping once preserves impossible overshoot that the old code discarded on each event, which makes zoom feel sticky near min/max. I did not run tests because this review was requested read-only.

Findings:
- `F1` · major · open — `applyGestureBatch` applies the product of all zoom factors through `zoomTowardPoint` once, so `clampZoom` runs once instead of once per raw event. The telescoping proof only holds while every intermediate zoom stays inside `[MIN_ZOOM, MAX_ZOOM]`. At a bound, this changes behavior: from zoom=3, factors `[1.25, 0.8]` used to clamp the first event to 3 and then zoom out to 2.4; the batched product is 1, so it stays at 3. Reverse/jitter ticks near min/max get canceled by overshoot that should have been discarded. (packages/client/src/canvas/viewport/transforms.ts:96)
- `F2` · minor · open — Mixed pan+zoom frames are not behavior-preserving. The old path divided each pan delta by the zoom value at that event's position in the sequence; the batch sums every pan delta and divides by the final post-zoom value. For example, pan then zoom in one frame used to apply the pan at the old scale, but now applies it at the new scale. `onPan` also covers pointer-drag pan, not only wheel pan, so this is not purely theoretical. (packages/client/src/canvas/viewport/transforms.ts:109)
- `F3` · minor · open — The benchmark only tries to reach `N_TILES` and then proceeds even when it fails, or when restore produced many more tiles than requested. That makes the published `16 tiles` comparison easy to reproduce with a different tile count while still producing a normal-looking table; under-counting especially understates the cost this harness is meant to prove. (docs/perf-investigations/scripts/gesture-p99/driver.cjs:157)
- `F4` · nit · open — The comment says `driver.js`, but the committed harness is `driver.cjs`. Small, but it contradicts the deliberate `.cjs` choice and the rest of the docs. (docs/perf-investigations/scripts/gesture-p99/.gitignore:1)

**claude** — Reviewed codex's 4 findings against the R3+R4 gesture-coalescing change. F1 (major) is a real bug — fixed. F4 (nit) and F3 (minor) — fixed. F2 (minor) — partial: the divergence is real but the suggested restructure (preserve event order / split batches) would defeat the coalescing, so I kept the canonical zoom-then-pan order and instead stopped the docs/comments from over-claiming exact per-event equivalence for mixed frames and documented it as an intentional bounded approximation.

F1 root cause: the accumulator multiplied raw zoom factors and clamped once at flush. Near MIN/MAX_ZOOM an overshoot past the bound cancels a later reversal (from MAX_ZOOM, factors [1.25, 0.8] have product 1 → stays pinned, whereas the per-event path clamps the first to the bound then zooms back out to 2.4×). New transforms.ts:accumulateZoom() folds each event's clamp into a running clamped zoom (seeded from a per-frame zoomStart captured on the first zoom contribution in useCanvasViewport.ts) and stores the net effective factor finalClampedZoom/startZoom. The pan correction still telescopes because the anchor is fixed across the frame — intermediate zooms cancel pairwise even when clamped — so it stays one apply per rAF with zero per-event allocation. Added 4 regression tests (MAX-bound reversal, MIN-bound reversal, repeated bound crossings, unclamped baseline) that compare against the true per-event clamped chain; all 8 transforms tests pass, client typecheck clean, biome-formatted.

Disputed nothing outright; F2 is the only partial — conceding codex's narrow correctness point (mixed frames aren't exact replay) while declining the structural change, because a wheel event is pan XOR zoom so a mixed frame only occurs on a pointer-drag pan overlapping a ctrl+wheel zoom, and the discrepancy is one frame's bounded, non-accumulating zoom delta.

- `F1` **fixed** — Real bug confirmed: from zoom=3 (MAX), factors [1.25,0.8] — per-event path clamps event 1 to 3 then zooms out to 2.4, but the raw product (1.0) stayed pinned at 3. The telescoping proof only holds when no intermediate clamp occurs. Fix: new accumulateZoom() in transforms.ts:93 maintains a running clamped zoom (startZoom * batch.zoomFactor), applies+clamps each factor, and stores net factor clampedZoom/startZoom. useCanvasViewport.ts onZoom now captures zoomStart=zoom() on the first zoom contribution per frame (sentinel: pending.zoomFactor===1) and calls accumulateZoom. Pan correction still telescopes to point/startZoom - point/clampedZoom because the anchor is fixed (intermediate zooms cancel even when clamped), so it stays one apply/zero per-event alloc. Verified factor can be negative/huge (gestures.ts:71 factor = 1 - deltaY*0.002), so per-event clamping genuinely matters. Added bound-reversal regression tests in transforms.test.ts comparing against the per-event clamped chain (perEventZoomChain); 8/8 pass.
- `F2` **partial** — Conceded the narrow correctness point — a mixed pan+zoom frame is NOT exact per-event replay (zoom applied first, whole summed pan ÷ post-zoom scale). Declined the suggested restructure (preserve order / split batches): exact replay requires storing and re-walking the per-event list, which is precisely the per-event work R4 exists to delete, and a wheel event is pan (plain/shift) XOR zoom (ctrl/meta) — never both — so a mixed frame only arises when a pointer-drag pan overlaps a ctrl+wheel zoom; the discrepancy is one frame's bounded, non-accumulating zoom delta (a few percent of a sub-pixel pan offset). Did the valid part: removed the over-claim of exact per-event equivalence for mixed frames and documented the chosen canonical zoom-then-pan order as a deliberate approximation in applyGestureBatch's docstring (transforms.ts:102), the useCanvasViewport.ts header comment, the perf note (canvas-gesture-p99.md), and the transforms.test.ts header. The existing mixed-frame test already pins the accepted zoom-first behavior.
- `F3` **fixed** — Agreed — a measurement harness must not silently proceed on a wrong tile count. driver.cjs now throws with a clear message when tiles < N_TILES after the Ctrl+T loop (under-count understates the per-tile write-storm), warns to stderr on over-count (doesn't hard-fail: extra tiles only strengthen the demonstration and the table is labeled with the real count), and prints the actual tile count in the console table header ('(${tiles} tiles, K=${BURST} events/burst)'), not only in JSON/stderr. node --check passes.
- `F4` **fixed** — Agreed — trivial doc/code mismatch. Changed the .gitignore comment from 'driver.js' to 'driver.cjs' to match the committed harness and the deliberate .cjs choice.

commit: `bcce03d41d721fb303fe9b5dd27a33a6df504589`

### Round 2

**codex** — approved: `false`

F1, F3, and F4 are adequately addressed in the current tree. The zoom accumulator now folds per-event clamping into the batch, the harness fails loud on under-count and labels over-counts, and the `.gitignore` comment matches `driver.cjs`. F2 is mostly addressed in the implementation comments, perf note, and tests, and I accept the decision not to preserve exact event order for mixed pan+zoom frames, but one changed Atlas paragraph still states the absolute equivalence that the fix now explicitly disclaims elsewhere.

Findings:
- `F1` · major · resolved — Resolved. `accumulateZoom` now updates the net zoom factor by applying `clampZoom` after each raw factor against a captured `startZoom`, so bound overshoot no longer cancels a later reversal. The new MAX/MIN-bound tests compare the batch to the old per-event clamped chain. (packages/client/src/canvas/viewport/transforms.ts:93)
- `F2` · minor · open — The code/perf-note side of this is now acceptable, but this Atlas paragraph still says the change is "Provably feel-neutral" and that the batched apply lands on the same per-frame pan+zoom the per-event path reached. That remains false for the mixed pan+zoom frame now documented as an intentional approximation in `transforms.ts` and `canvas-gesture-p99.md`. (docs/atlas/src/content/atlas/opencode-perf.mdx:169)
- `F3` · minor · resolved — Resolved. The harness now throws if it cannot create the requested minimum tile count, warns on over-count, and prints the actual tile count in the table header. (docs/perf-investigations/scripts/gesture-p99/driver.cjs:168)
- `F4` · nit · resolved — Resolved. The comment now names `driver.cjs`. (docs/perf-investigations/scripts/gesture-p99/.gitignore:1)

_codex on the rebuttal:_ F1: fixed; the new accumulator preserves the per-event clamp semantics I was concerned about. F2: I accept the argument against storing/replaying event order for a rare mixed pointer-drag plus ctrl-wheel frame, and the implementation comments/tests now honestly pin zoom-then-pan as the chosen approximation. I am holding the finding open only because the Atlas note still repeats the original absolute equivalence claim. F3 and F4: fixed.

**claude** — test filesChanged one path

- `F1` **fixed** — Confirmed resolved by codex in round 1; no change this round.
- `F2` **fixed** — Rewrote the R4 paragraph in opencode-perf.mdx to drop the absolute feel-neutral claim; pure-pan and pure-zoom exact, mixed uses the zoom-then-pan approximation. Regenerated dist per the atlas-sync gate.
- `F3` **fixed** — Confirmed resolved by codex in round 1; no change this round.
- `F4` **fixed** — Confirmed resolved by codex in round 1; no change this round.

commit: `abd0f231611b70b7f7043bfca81c0a09b64a6bfa`

### Round 3

**codex** — approved: `true`

All previously raised findings are now resolved. The remaining F2 issue was fixed in the Atlas source and regenerated HTML: the R4 paragraph now limits exact equivalence to pure pan/pure zoom and explicitly calls mixed pan+zoom a deliberate zoom-then-pan approximation. `git status --short` shows no untracked files, and `git diff --check` is clean.

Findings:
- `F1` · major · resolved — The zoom batching originally clamped only once at flush, allowing bound overshoot to cancel later reversals. The current `accumulateZoom` path now folds per-event clamping into the accumulated net factor, with regression coverage for MAX/MIN-bound reversals. (packages/client/src/canvas/viewport/transforms.ts:93)
- `F2` · minor · resolved — The Atlas R4 paragraph previously repeated the absolute equivalence claim despite the accepted mixed pan+zoom approximation. It now states exactness only for pure-pan/pure-zoom frames and documents mixed pan+zoom as canonical zoom-then-pan approximation; the rendered `docs/atlas/dist/opencode-perf.html` carries the same language. (docs/atlas/src/content/atlas/opencode-perf.mdx:169)
- `F3` · minor · resolved — The benchmark harness previously proceeded with an insufficient tile count. It now throws on under-count, warns on over-count, and labels the output table with the actual tile count. (docs/perf-investigations/scripts/gesture-p99/driver.cjs:168)
- `F4` · nit · resolved — The `.gitignore` comment previously said `driver.js`; it now correctly says `driver.cjs`. (docs/perf-investigations/scripts/gesture-p99/.gitignore:1)

_codex on the rebuttal:_ F2 is fixed. I verified the updated Atlas source and rendered HTML both include the pure-pan/pure-zoom exactness caveat and the mixed pan+zoom approximation language. F1, F3, and F4 remained resolved from the prior round; no new regression was introduced by this closeout change.

srid:

✅ **All clear** — scoped to `1c8f1770...HEAD` (R3+R4), run `--no-elegance` (the simplify track already ran).

| Pass | Result |
|---|---|
| **Rules** (27 checks) | 0 violations |
| **Fact-check** | 0 real bugs |
| **Elegance** | skipped (covered by `/simplify`) |

One theoretical note, **no action taken**: the zoom accumulator uses `pending.zoomFactor === 1` as the "first zoom contribution this frame" sentinel to capture `zoomStart`. A floating-point `=== 1` sentinel *looks* fragile, but it's sound here — `zoom()` is frame-constant (only `flushGesture` and the discard-on-write programmatic setters mutate it, never mid-frame), so even if the sentinel re-fired it would re-read the *same* `zoomStart` value: capturing it is idempotent within a frame. The other surfaced item (`zoomStart` "not reset" on discard) was a false alarm — `discardPendingGesture` resets `pending.zoomFactor` to `1`, which *is* the sentinel, so the next zoom event re-captures cleanly.

srid:
## Evidence

This is a **feel-neutral perf fix** — panning and zooming land in *exactly* the same place before and after (the coalesced batch telescopes to the per-event result; pinned in `transforms.test.ts`). So there is no visual diff to film: a video of a fling looks identical either way. The behavior worth proving is **"the main thread no longer freezes mid-gesture on a weak client,"** which is a measurement, captured by the committed harness (`docs/perf-investigations/scripts/gesture-p99/`) re-run against the **final shipped commit** (`2cc7e245`).

**Method:** 16-tile canvas, a burst of K=60 real `WheelEvent`s dispatched in a tight loop (the "many events between two paints" regime R4 targets), median of 7 bursts, swept over CPU throttle. `burstMs` = the synchronous main-thread block (the freeze a user feels); `writes` = tile-transform recomputes; `framesBlocked` = burstMs ÷ 16.67 ms.

| throttle · gesture | main-thread block (master → **R4**) | tile writes (master → **R4**) | frames blocked (master → **R4**) |
|---|---|---|---|
| 6× zoom | **865.5 ms → 3.3 ms**  (262×) | 2,880 → **48** (60×) | 51.9 → **0.2** |
| 6× pan  | 283.2 ms → **1.3 ms** | 960 → **16** | 17.0 → **0.08** |
| 4× zoom | 534.9 ms → **1.7 ms** | 2,880 → **48** | 32.1 → **0.1** |
| 4× pan  | 190.6 ms → **0.6 ms** | 960 → **16** | 11.4 → **0.04** |
| 1× zoom | 123.4 ms → **0.8 ms** | 2,880 → **48** | 7.4 → **0.05** |
| 1× pan  | 43.8 ms → **0.3 ms** | 960 → **16** | 2.6 → **0.02** |

The worst case — a zoom fling on a 6×-throttled client — went from a **~865 ms main-thread freeze (≈52 dropped 60 Hz frames)** to imperceptible. The write ratio is exactly **K=60** everywhere (60 events → 1 apply) — that *is* the coalescing, not hidden work. (codex's review caught a real bound-clamping bug along the way; its fix adds a cheap per-event clamp, which is why 6× zoom is 3.3 ms rather than ~1 ms — still 262× under master, and now correct at the zoom limits with 4 regression tests.)

**Reproduce:** `just dev-auto` then `bash docs/perf-investigations/scripts/gesture-p99/run.sh`. Full methodology: [`canvas-gesture-p99.md`](https://github.com/juspay/kolu/blob/perf-gesture-p99-harness/docs/perf-investigations/canvas-gesture-p99.md).

**Mobile/touch:** the change is in the shared gesture→viewport path (`useCanvasViewport.ts`), pointer-agnostic — it coalesces wheel *and* pointer-drag pan identically — so there's no desktop-only or coarse-pointer-specific surface; no per-modality UI was added or resized.

PR #1377, opened 2026-06-15
perf(surface-nix-host): skip re-provisioning an already-set-up host (P2.7)

srid:
**A warm host no longer re-ships a closure it already has.** `provisionAgent` ran `nix copy --derivation` → realise → pin **unconditionally** on every dial, so reaching an already-provisioned host re-did all three across slow ssh round-trips that accomplished nothing — `nix copy` literally reporting `copying 0 paths`. This adds a remote-only **warm fast-path** that asks nix whether the closure is already realisable and, if so, skips straight to dialing.

Implements **P2.7** of the [kaval-sessions](https://htmlpreview.github.io/?https://github.com/juspay/kolu/blob/master/docs/atlas/dist/kaval-sessions.html) plan (#1376).

### Timings — before/after, **ssh lag removed**

The real-world warm path is ~30s, but that's dominated by ~5s per *new* ssh connection (no multiplexing — a separate, deferred concern). To isolate what *this* change buys, measured against a warm `kolu-ci-1` over an ssh **ControlMaster** (so each op pays only its nix work + a ~0.55s channel floor, no handshake):

| warm-host remote op | before (old code runs all three) | after (new code) |
| --- | --- | --- |
| `nix copy --derivation` (0 paths) | **2.4s** — real store-query work, transfers nothing | skipped |
| `nix-store --realise <drv>` | 0.55s (≈ channel floor, no work) | folded in |
| pin — `--realise <out> --add-root` | 0.51s (≈ channel floor, no work) | folded in |
| **fused `--realise <drv> --add-root --indirect`** | — | **0.58s** |
| **total remote work** | **~3.5s** | **~0.58s** |

**~3.5s → ~0.6s (~6×) sans ssh lag** — the only genuine *work* the old warm path did was the ~1.9s `nix copy` store-query that sends 0 paths, and it's gone. The realise/pin were already near-zero work; their cost was purely the extra round-trips. _With today's real ~5s-per-connection ssh lag (no multiplexing), the change also drops 2 of 3 round-trips — a further ~10s — which the deferred ControlMaster follow-up would make moot._

### The fix

```
provisionAgent(host, drv):
  out = ssh host  nix-store --realise $drv --add-root $root --indirect   # 1 ssh
  if ok:  return out                       # WARM — closure present → repin, done
  nix copy --derivation $drv → realise + pin    # COLD — drv absent → full provision
```

One fused `--realise <drv> --add-root … --indirect` both **proves** the closure is present (it fast-fails — `"don't know how to build"` — when absent and unsubstitutable) **and** refreshes the GC root, so a warm host skips copy+realise+pin entirely. Keyed on **nix's own store-realisability**, *not* on `readlink`-ing the moving-result GC root — an indirect root can dangle past an out-of-band store GC or a box reimage and resolve to a path that no longer exists.

- **Fall-through preserves best-effort pin.** On a miss it falls through to the existing copy → realise → pin, whose pin is best-effort — so even an unwritable `$HOME` root degrades to *works, unpinned* rather than a hard failure. The cold path is otherwise untouched.
- **Drishti benefits for free.** The change is internal to the **shared** `provisionAgent` — the one ensure verb that both `kaval --stdio` and drishti's `process-monitor-agent` dial. Same signature + same returned out-path ⇒ **not API-facing** ⇒ per `surface.md` no companion PR is compelled; drishti just speeds up on its next kolu pin.

### Correctness (red-first test + verified facts)

- **Red-first unit test** in `nixCopy.test.ts`: asserts a warm host (the realise probe succeeds) **never calls `nix copy`** (`runProgress`). Red on the old code, green now. Cold-path tests updated for the warm-probe-first order.
- **Single-output, input-addressed** — `kaval` has one deterministic output path (verified), so one drv ⇒ one out.
- **Fail-safe** — any miss / absent-root / mismatch falls *through* to the full provision; a transport failure flows through the same `onProgress` network scan, so an unreachable host still classifies as `network` and can't burn the reconnect give-up budget.

### Deferred (per the plan)

ssh `ControlMaster` multiplexing (the ~5s/connection villain) and the per-dial arch probe are noted follow-ups; a local provisioned-cache was rejected on staleness. Not in this PR.

### Try it locally

```sh
nix run github:juspay/kolu/p2.7-provision-fast-path#kaval-tui -- list --host <ssh-host>
# run it twice — the second connect skips the copy/realise/pin
```

---


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `53e8a9c71788`

### Applied (3)
- `lowy-3` Step-numbering in comments drifted after rootPath hoist and warm-path insertion — commit `30a528d5a`
- `hickey-1` The pin operation (realise + indirect GC-root) is fragmented across two argv-construction sites — commit `7218b0e39`
- `hickey-2` Header doc-comment's 5-step narrative now disagrees with the body's renumbered steps — (uncommitted)

### Agreed — no change (5)
- `lowy-1` Warm fast-path is correctly placed behind the stable provisionAgent receptacle (packages/surface-nix-host/src/nixCopy.ts:117-148)
- `lowy-2` The atomic 'realise-and-pin' verb is now expressed in two different decompositions within one function (symmetry break) (packages/surface-nix-host/src/nixCopy.ts:130-148 vs 191-246)
- `lowy-4` Warm-probe stdout-non-empty guard relies on an unstated nix invariant (packages/surface-nix-host/src/nixCopy.ts:141-147)
- `lowy-5` README/changelog/atlas correctly describe the change as non-API-facing — boundary claim holds (packages/surface-nix-host/README.md:42,67,80)
- `hickey-3` Inconsistent destructuring idiom across the three buildSshProbeCommand call sites (packages/surface-nix-host/src/nixCopy.ts:131,191,231)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `53e8a9c71788`

### Round 1

**codex** — approved: `false`

The implementation preserves the cold provision path after a failed warm probe and the core Nix semantics look sound: the fast path realises the derivation, returns the realised output path, and skips `nix copy` only on a realisable remote closure. I found one minor issue in the progress/error surface for expected cold misses.

Findings:
- `F1` · minor · open — The speculative warm probe forwards all stderr through the normal `onProgress` path. On a cold host, that probe is expected to fail because the `.drv` is absent, and Nix will emit an error before the code falls through and provisions successfully. Since `HostSession` stores these lines in user-visible progress, a successful first-time provision can now contain a scary but expected probe error, making the progress log look failed even though the final result is OK. (packages/surface-nix-host/src/nixCopy.ts:155)

**claude** — Codex returned one minor finding (F1) on the P2.7 warm fast-path; I agreed and fixed it. The speculative warm probe (`nix-store --realise <drv> --add-root … --indirect`) forwarded all its stderr through the user-visible onProgress ring. On a cold host the probe is expected to fail (the .drv isn't on the host yet), so nix emits a real `error: …` line that leaked into HostSession.progressLines (the 20-line ring the UI shows) — making a clean first-time provision read as if it errored. I confirmed the mechanics end-to-end: runCapture stderr -> wrapped onProgress -> addLocalProgress -> progressLines (hostSession.ts:338). Fix: a probe-specific callback onProbeProgress that still scans each line with looksLikeNetworkError to set sawNetworkError (preserving the author's deliberate decision #6 — a transport failure on the probe must keep classifying the fall-through as "network") but does NOT echo the line to opts.onProgress. The real copy/realise path still reports its own errors verbatim on a genuine failure. This is fully compatible with every deliberate decision: #6 only requires the network classification to survive, not the raw stderr to be shown. Two new tests lock both halves in (cold-miss error: swallowed; network-looking probe line still yields cause:"network"). Full surface-nix-host suite green (30/30, including 12 in nixCopy.test.ts), tsc --noEmit clean. Not API-facing (provisionAgent signature unchanged) so no drishti PR. Could not run biome (resolved via nix shell, not in node_modules here) but changes mirror the file's existing comment/string-wrap conventions and add no over-width code lines beyond unsplittable string literals/comments already pervasive in the file.

- `F1` **fixed** — Agreed. Confirmed the leak path: on a cold host the warm probe's `nix-store --realise <drv>` fails because the .drv is absent, nix writes a genuine `error: …` line to stderr, and that line flowed through the wrapped `onProgress` (nixCopy.ts:128) into `opts.onProgress` -> `addLocalProgress` -> `progressLines` (hostSession.ts:338,85), the 20-line ring the UI displays — so a successful first-time provision read as if it errored. Fix in packages/surface-nix-host/src/nixCopy.ts:142: added `onProbeProgress` which still runs `looksLikeNetworkError(line)` to set `sawNetworkError` (so a transport failure on the probe still classifies the fall-through as "network" via `causeFor` — author's deliberate decision #6 preserved) but does NOT echo the line to `opts.onProgress`. Wired the probe's runCapture (nixCopy.ts:170) to use it and updated the block comment. Exactly codex's suggested shape (scan for classification, suppress expected probe-miss stderr, let the real copy/realise path report its own errors). Locked in with two tests in nixCopy.test.ts: 'does not leak the expected probe-miss error into the progress ring' asserts a cold-host `error: path '…' is not valid` line never appears in the collected progress lines; 'still classifies a transport failure on the probe as network' asserts a network-looking probe line yields `cause: "network"`.

commit: `71c1bfc3771202743b7a68b56eafbb584b1b1ea2`

### Round 2

**codex** — approved: `true`

F1 is resolved. The warm probe now uses `onProbeProgress`, which still feeds `looksLikeNetworkError` into `sawNetworkError` but does not forward expected cold-miss stderr into the user-visible progress stream. The added tests cover both sides of the fix: cold probe errors are suppressed from progress, and probe transport failures still classify the fall-through failure as `network`. I did not find a regression introduced by this round's changes.

Findings:
- `F1` · minor · resolved — The speculative warm probe previously forwarded expected cold-miss stderr into `opts.onProgress`, so a successful first-time provision could show a scary Nix `error:` line in the UI progress ring. (packages/surface-nix-host/src/nixCopy.ts:142)

_codex on the rebuttal:_ No dispute to address. Claude agreed with F1 and implemented the requested shape; I verified the current code and tests preserve network classification while suppressing expected probe-miss stderr from the progress ring.

srid:

Two passes over the change vs the merge-base (`--no-elegance`, since `/simplify` already ran).

| Pass | Finding | Action |
| --- | --- | --- |
| **Rules** | `no-thin-wrapper-functions`: `realiseAndPin` has 2 callers and only forwards bound args to `buildSshProbeCommand`. | **Overridden** — it matches the file's own argv-builder idiom (`buildSshProbeCommand`/`buildAgentCommand` are the same shape), carries the lens-debate consensus, and single-sources the `--add-root … --indirect` flag sequence so a GC-root change touches one site. The caller-count heuristic doesn't account for that value. |
| **Fact-check** | `sawNetworkError` set by the *speculative* warm probe persisted across phases, so a transient probe network blip that cleared before the copy could make a later *genuine remote* realise/pin failure misclassify as `"network"` (retry forever) instead of `"remote"` (bounded give-up). | **Fixed** (`69c1437e`) — reset the flag after a **successful copy** (the host has proven reachable); each later step's own stderr scan re-sets it if the host drops again. The police's literal "fail-fast on probe network" suggestion was rejected (it would regress roaming-recovery). Covering test added; the roaming case (copy itself fails → `network`) stays green. |

13/13 `nixCopy.test.ts` green, `tsc --noEmit` clean, `just check` green.

srid:
## Evidence

A CLI/provisioning-latency change (no UI surface), so the proof is the before→after CLI behaviour against a warm `kolu-ci-1`.

### Before (`master`) — the second dial still re-provisions

`kaval-tui list --host kolu-ci-1` run twice; the **second** run (host already provisioned) still runs the whole dance:

```
kolu-ci-1: copying derivation '…-kaval.drv'…
copying 0 paths...                                ← nothing to send, still connects
kolu-ci-1: realising '…-kaval.drv' on remote…     ← rebuilds nothing
kolu-ci-1: pinning GC root at '…/gcroots/kaval'…  ← re-pins an identical root
```

### After (this PR) — the warm dial skips straight to dialing

Same command, this branch's build (`69c1437e`), twice in a row:

```
# run 1 (cold for this build) — provisions as normal:
kolu-ci-1: copying derivation '…8wy4wc2…-kaval.drv'…
copying 4 paths...
kolu-ci-1: realising … · pinning GC root '…/gcroots/kaval'…
connection: copying → connecting → connected

# run 2 (warm) — copy/realise/pin skipped entirely:
kolu-ci-1: already provisioned at …/gcroots/kaval — skipped copy
connection: copying → connecting → connected
no live terminals.
```

The warm second dial drops from **three** remote round-trips (copy + realise + pin) to **one** (the fused realise-probe that prints *"skipped copy"*). Per-op timings with ssh lag removed are in the PR description (**~3.5s → ~0.58s** of remote work, ~6×). Captured live on a real warm host (and independently reproduced by @srid).

srid:

The **x86_64-linux** lane ran on **`kolu-ci-3`** (`dev-x86-64-linux-03`) — commit `69c1437e`, **passed** ✓

- **Lane wall** (pipeline): **7m16s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 1m58s | ✓ |
| `ci::home-manager` | 1m43s | ✓ |
| `ci::smoke` | 35s | ✓ |
| `ci::flake-check` | 35s | ✓ |
| `_ci-setup` | 33s | ✓ |
| `ci::unit` | 32s | ✓ |
| `ci::surface-example-build` | 19s | ✓ |
| `ci::surface-app-example-build` | 19s | ✓ |
| `ci::fmt` | 17s | ✓ |
| `ci::biome` | 17s | ✓ |
| `ci::nix` | 14s | ✓ |
| `ci::install` | 7s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/69c1437/timings.jsonl`; pool state is a live `flock` probe.</sub>

PR #1381, opened 2026-06-15
Repaint occluded terminals on window focus (the real freeze)

srid:
**The "terminal freezes until you press a key" bug — for real this time.** Output stops appearing while an agent keeps working (classically: you alt-tab away while a fleet of background subagents runs), and only snaps to the latest when you return and press a key. Two prior PRs each fixed a *different* freeze and missed this one — because the cause is in the **render pipeline**, a layer neither touched. This is proven against the pinned `@xterm/xterm@6.1.0-beta.225`, not reasoned.

### Why the two earlier fixes couldn't have worked

| PR | Layer it fixed | Why it missed this freeze |
| --- | --- | --- |
| <code>#1235</code> | transport (half-open WebSocket → heartbeat) | the data *does* arrive — a keypress reveals the **latest**, so it's already in xterm's buffer |
| <code>#1273</code> | write path (scroll-lock latch) | the lock never latches during a quiet idle (no scroll intent), so output is written straight through |
| **this** | **render pipeline (parked `requestAnimationFrame`)** | the bytes are in the buffer; only the **paint** is missing |

### The mechanism

xterm's `RenderService` schedules every visible paint through **one** `window.requestAnimationFrame`, with no timer fallback — `RenderDebouncer` does `if (this._animationFrame) return;` and clears that handle *only* inside the rAF callback.

```
write ─▶ parse into buffer (setTimeout, only clamped under occlusion) ─▶ mark rows dirty
      ─▶ refreshRows(sync=false) ─▶ RenderDebouncer.refresh ─▶ requestAnimationFrame(paint)
                                                                      │
   window OCCLUDED (alt-tab; visibilityState stays "visible") ──▶ Chromium stops producing frames
                                                                      │
   the queued paint never runs · _animationFrame stays a stale handle · every later write early-returns
                                                                      │
   keypress ──▶ BeginMainFrame ──▶ services the parked rAF ──▶ paints the LATEST  ("press a key")
```

Because app-switch occlusion keeps `document.visibilityState === "visible"`, it never trips `visibilitychange` — which is exactly why `refitOnTabVisible` never helped.

### The fix — `renderRecovery.ts`

A DOM-free core that forces a **synchronous** repaint (`refreshRows(0, rows-1, true)` bypasses the stalled debouncer — proven to paint the latest *without* a serviced frame), triggered by the signals that actually fire on occlusion-return:

- **window `focus`** — wired in `Terminal.tsx` (DOM-adjacent, mirroring `scrollLock`/`scrollLockWiring`); synchronous, so it doesn't wait for Chromium to resume frames.
- **a self-disarming watchdog** — output arrived but no paint followed within a beat, while focused.

Both are **cause-agnostic**: they don't depend on *why* frames stopped. All xterm private-path access is null-guarded in one place, like the existing `readBufferBytes` probe.

### So the next one identifies itself

Diagnostic Info now carries render-pipeline state — per-terminal **time since last paint**, **whether a repaint is stuck pending** (the parked rAF), render pause, synchronized-output — plus **page visibility + `hasFocus` at snapshot time** (the occlusion signature is `visible` + `!hasFocus`). No more guessing the layer.

### Proven, not guessed

A deterministic harness drove the real beta in headless Chromium: holding rAF while writing leaves `buffer.length` growing with **zero** `onRender` firings (frozen), one serviced frame paints the latest (the keypress), and `refreshRows(…, true)` repaints a frozen terminal with **no** serviced frame (the fix). Shipped here as **8 unit tests** for the watchdog/recover/probe state machine and an **e2e scenario** that stalls the focused terminal's render loop, proves output lands unpainted, then asserts window focus forces a sync repaint (red without the wiring).

> Real OS occlusion can't be reproduced in a headless browser, so the e2e stalls the render loop directly; the platform link (occlusion → parked rAF) is established Chromium behavior, and the new diagnostics confirm it on a real machine at the next freeze.

### Try it locally

```sh
nix run github:juspay/kolu/fuckingfreeze
```


srid:

✅ **Consensus** after 2 round(s) · lowy + hickey · base `493ca2d02988`

### Applied (4)
- `lowy-1` xterm-private-internals volatility split across two receptacles — commit `cc114daf6`
- `hickey-1` recover() and the watchdog detect the same 'paint fell behind data' condition by two different rules — commit `9de02bc96`
- `hickey-2` 'Render is stalled' threshold is duplicated between the watchdog and the Diagnostic display with unrelated constants — commit `32fd25fbd`
- `hickey-4` msSinceLastPaint returns 0 for two different states — 'never painted' and 'painted just now' — commit `1ffd6fb7e`

### Agreed — no change (6)
- `lowy-2` Probe-bundle (instrumentation) complected with recovery-mechanism (control) behind one interface (packages/client/src/terminal/renderRecovery.ts:105-114 (RenderRecovery interface))
- `lowy-3` Two-trigger recovery duplicates the 'attention returned' signal across modules (packages/client/src/terminal/Terminal.tsx:656 (window focus) and 833-846 (refitOnTabVisible calls recovery.recover()))
- `lowy-4` isOnScreen gate parameter is shaped around Terminal's prop, not a stable predicate (packages/client/src/terminal/renderRecovery.ts:127-131 and Terminal.tsx:649)
- `hickey-3` renderRecovery braids active recovery with passive diagnostic probes into one object (packages/client/src/terminal/renderRecovery.ts:498-507)
- `hickey-5` Per-terminal window-'focus' listeners all fire every recovery — the trigger is document-global but wired N times (packages/client/src/terminal/Terminal.tsx:656)
- `hickey-6` DiagnosticInfo re-derives 'occluded' with a third independent formula (packages/client/src/DiagnosticInfo.tsx:47)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 round(s) · codex reviewed at `xhigh` reasoning effort · base `493ca2d02988`

### Round 1

**codex** — approved: `false`

The core direction looks plausible, and the private xterm reach is contained as described. I am not approving yet because the e2e test races the watchdog it is supposed to bypass, and the watchdog is currently driven by stream receipt rather than actual xterm-buffer writes. That makes the recovery bookkeeping inaccurate around scroll-lock and async parsing, with a smaller extra-render risk during normal output.

Findings:
- `F1` · major · open — The render-recovery scenario leaves `document.hasFocus()` true while output is generated. In production, `Terminal.tsx` calls `recovery.noteData()` for that output, and `noteData()` arms the 250ms watchdog whenever the document has focus. If `I generate 30 lines of output` plus the buffer wait takes longer than 250ms on a slow runner, the watchdog can force the sync repaint before this assertion, making the test flaky and not actually modeling the occluded/no-focus state described in the feature. (packages/tests/step_definitions/render_recovery_steps.ts:56)
- `F2` · major · open — `recovery.noteData()` is called after `scrollLock.writeData(...)`, but that does not mean data reached xterm. When scroll-lock is engaged, `writeData` only buffers and does not call `term.write`; when unlocked, `term.write` still parses asynchronously. This can arm recovery for data that is not paintable yet, trigger pointless full sync repaints while the user is intentionally scroll-locked, and miss the real write later when scroll-lock flushes buffered data without notifying recovery. (packages/client/src/terminal/Terminal.tsx:742)
- `F3` · minor · open — The watchdog is not disarmed when a normal render catches up. A timer armed by an early chunk can remain pending through many successful rAF paints; if another chunk arrives just after the latest normal paint but before that old timer fires, `paintIsBehind()` becomes true and the code forces a full synchronous repaint even though the render loop is healthy. That contradicts the stated 'never fights normal rendering' behavior and adds avoidable work during continuous output. (packages/client/src/terminal/renderRecovery.ts:126)
- `F4` · minor · open — The diagnostic warning treats `msSinceLastPaint === null` as `0` via `?? 0`, so `paint: ? · rAF:pending` is rendered muted rather than dangerous. The recovery code documents `null` as 'never painted' and the most severe unknown state, so the UI currently hides that signal. (packages/client/src/DiagnosticInfo.tsx:323)

**claude** — Reviewed CODEX's four findings on the render-stall recovery PR (#1381). All four are valid (F1/F3/F4 fully agreed; F2 partial — the pointless-repaint half is real and fixed, the missed-flush half is already covered by existing paths). Fixed all valid parts in the working tree, added a unit test for F3, ran the affected unit suites (27 passing) and typechecked the client package clean, then committed.

F1 (e2e races watchdog): The scenario left document.hasFocus()=true while generating output, so on a slow runner the 250ms watchdog could force the sync repaint before the "not repainted yet" assertion (paintCount > 0). Fixed by patching document.hasFocus() to false in the stall step (modeling occlusion, keeping the watchdog disarmed — which is the feature's actual subject), and restoring it before dispatching the focus event. The focused-watchdog path codex asked to keep separate is already covered by renderRecovery.test.ts.

F2 (noteData not keyed to actual xterm write): scrollLock.writeData now returns whether it wrote to xterm; Terminal.tsx arms the watchdog only on a real write, so scroll-locked (buffered) chunks no longer arm a watchdog that forces pointless sync repaints of the scrollback the user is reading. The "miss the flush" half is conceded as already-covered, not a bug: flush() runs only via scrollToBottom (user scroll, focus present), handleTabVisible (already calls recover()), or reset — none is the occlusion-return path, which fires window 'focus' -> recover() regardless.

F3 (watchdog not disarmed on catch-up): onRender now disarms the watchdog when the latest paint caught up to the data, so a stale timer armed by an earlier chunk can't force a repaint during healthy continuous output. Added a unit test proving it (data, normal paint, more data before the original deadline, subsequent paint -> zero forced repaints).

F4 (null msSinceLastPaint shown muted): The Diagnostic danger predicate used (msSinceLastPaint ?? 0) > PAINT_STALL_WARN_MS, so null ("never painted", the most severe unknown per the module docs) read as 0 and rendered muted. Now reds when debouncerPending && (msSinceLastPaint === null || > PAINT_STALL_WARN_MS).

- `F1` **fixed** — Agreed — real flake. packages/tests/step_definitions/render_recovery_steps.ts: the stall step now Object.defineProperty-patches document.hasFocus() to return false (modeling occlusion, so renderRecovery.noteData()'s focus-gated watchdog never arms during 'I generate 30 lines of output'); the 'window regains focus' step deletes the override (restoring the prototype's real hasFocus) before dispatching the focus event. This also makes the scenario faithfully model the occluded/no-focus state the feature describes, rather than a focused terminal. The focused-watchdog path codex suggested keeping as a separate test is already covered by renderRecovery.test.ts's first case.
- `F2` **partial** — Agreed on the pointless-repaint half, conceded the flush half as already-covered. Fixed: scrollLock.ts writeData() now returns boolean (true = wrote to xterm, false = buffered while locked); Terminal.tsx:750 now does `if (scrollLock.writeData(terminal, data)) recovery.noteData();` so buffered scroll-locked chunks no longer arm the watchdog and force pointless sync repaints of the scrollback the user is reading. Disputed/conceded half: flush() (scrollLock.ts) is only reached via scrollToBottom() (user scroll, focus present, scroll itself forces frames), handleTabVisible() (already calls recovery.recover() at Terminal.tsx:794), or reset() — none is the occlusion-return path. Occlusion return fires window 'focus' -> recovery.recover() (Terminal.tsx:605), which force-repaints unconditionally. So notifying recovery from the flush path adds no coverage for the freeze bug; I left it out to avoid complecting scrollLock with recovery.
- `F3` **fixed** — Agreed — real fight-normal-rendering case. renderRecovery.ts onRender callback now calls disarm() when !paintIsBehind() (the latest paint caught up to the data), so a watchdog armed by an earlier chunk and left pending through a healthy rAF paint can't later see a newer-but-soon-to-paint chunk via paintIsBehind() and force a repaint. Added the exact unit case codex asked for in renderRecovery.test.ts: chunk A at t=0 arms, normal paint at t=16 disarms, chunk B at t=240 arms fresh, no forced repaint past the original t=250 deadline, chunk B's own paint at t=270 keeps it clean. 27 unit tests pass.
- `F4` **fixed** — Agreed — UI hid the worst signal. DiagnosticInfo.tsx:323 danger predicate changed from `d.render.debouncerPending && (d.render.msSinceLastPaint ?? 0) > PAINT_STALL_WARN_MS` to `d.render.debouncerPending && (d.render.msSinceLastPaint === null || d.render.msSinceLastPaint > PAINT_STALL_WARN_MS)`, so the never-painted (null) state — documented in renderRecovery.ts as the most severe unknown — now reds instead of masquerading as the healthiest value via ?? 0.

commit: `e05c1b0e98d239f37315dc234c8adc223078227d`

### Round 2

**codex** — approved: `false`

Current round resolves F1, F3, and F4. The scroll-lock part of F2 is fixed, and I accept CLAUDE's reasoning that a separate flush notification is not required for the occlusion-return path because focus/visibility recovery already forces a repaint. F2 remains open for the separate async-write timing issue from the original finding: recovery is still notified immediately after `term.write(...)`, before xterm has parsed the data into its buffer.

Findings:
- `F1` · major · resolved — Previously, the e2e render-stall scenario left `document.hasFocus()` true, so the focused watchdog could fire before the no-repaint assertion. (packages/tests/step_definitions/render_recovery_steps.ts:46)
- `F2` · major · open — `recovery.noteData()` is still called immediately after `scrollLock.writeData(...)` returns true, but `term.write(...)` queues parsing asynchronously in xterm. The watchdog can therefore be armed before the data is actually in xterm's buffer; if it fires before parsing completes, it repaints stale rows, clears the timer, and can miss the later parsed output in the focused-stall case. (packages/client/src/terminal/Terminal.tsx:750)
- `F3` · minor · resolved — Previously, a stale watchdog stayed armed after a normal render caught up, so later healthy output could trip the old timer and force an unnecessary repaint. (packages/client/src/terminal/renderRecovery.ts:126)
- `F4` · minor · resolved — Previously, `msSinceLastPaint === null` was treated as `0` in the diagnostic danger predicate, muting the never-painted pending-rAF state. (packages/client/src/DiagnosticInfo.tsx:323)

_codex on the rebuttal:_ F1, F3, and F4 are fixed as described. On F2, I accept the scroll-lock buffering fix and I also concede the disputed flush-path half: the occlusion-return and tab-visible paths already call `recover()`, so a flush-specific recovery hook is not necessary for that scenario. I am holding F2 open only for the remaining async-write timing concern from the original finding. xterm's `write()` schedules `_innerWrite()` via `setTimeout` and invokes the write callback only after the chunk is processed, so the current immediate `noteData()` still records stream receipt rather than parsed xterm-buffer state.

**claude** — Round 2 of the codex debate on PR #1381 (terminal render-stall recovery). Only F2's remaining "async-write timing" half was still open; F1/F3/F4 were resolved in round 1 and codex confirmed them resolved. I agreed with codex on F2 and fixed it.

Verified codex's mechanism against the pinned @xterm/xterm 6.1.0-beta.225 source: term.write(data) returns immediately and WriteBuffer schedules _innerWrite (the parse) off a setTimeout (WriteBuffer.ts:156), so the prior `recovery.noteData()` fired synchronously after writeData recorded STREAM RECEIPT, not the chunk landing in xterm's buffer. paintIsBehind() then reasoned about lastDataAt timestamps that preceded the data being paintable, leaving a narrow gap (a parse slice completing after the single watchdog fire while a focused rAF stays parked, with nothing re-arming). Fixed using xterm's PUBLIC, typed write callback (term.write(data, callback), xterm.d.ts:1370; WriteBuffer fires the per-chunk callback only after _action parses it, WriteBuffer.ts:254) — no new private-internals reach.

scrollLock.writeData now takes an optional onParsed and threads it into term.write; Terminal.tsx passes () => recovery.noteData() so the watchdog arms only when the chunk is actually in the buffer. The round-1 property is preserved: the scroll-locked (buffered) branch deliberately does NOT invoke onParsed (buffered data produces no paint). The flush path stays uninstrumented (codex conceded in round 1 that lock release rejoins the bottom via user-scroll / tab-visible / window-focus, each already forcing a repaint via recover()). writeData's boolean return was dropped (the callback subsumes it); no caller asserted on it. The e2e step needs no change — it forces hasFocus()=false during output so the watchdog never arms regardless of when noteData fires.

Verification: client typecheck clean (tsc --noEmit), 27 unit tests pass (scrollLock + renderRecovery), biome format clean.

Relevant files: /home/srid/code/kolu/.worktrees/fuckingfreeze/packages/client/src/scrollLock.ts (writeData signature: optional onParsed threaded into term.write), /home/srid/code/kolu/.worktrees/fuckingfreeze/packages/client/src/terminal/Terminal.tsx:755 (attach callback passes () => recovery.noteData()).

- `F1` **fixed** — Resolved in round 1 and confirmed by codex this round. The stall e2e step (render_recovery_steps.ts) patches document.hasFocus() to false while generating output (modeling occlusion, keeping the focus-gated watchdog disarmed) and restores it before dispatching the focus event. No further change needed in round 2.
- `F2` **fixed** — Agreed with codex's remaining open concern. Verified against the pinned @xterm/xterm 6.1.0-beta.225 source: term.write(data) returns synchronously and WriteBuffer schedules the parse via setTimeout (WriteBuffer.ts:156), so the old `if (scrollLock.writeData(...)) recovery.noteData()` armed the 250ms watchdog at stream receipt against data not yet in the buffer, making paintIsBehind() reason about pre-parse timestamps. Fixed by keying noteData to xterm's parse-complete using the PUBLIC write callback: scrollLock.writeData(term, data, onParsed) threads onParsed into term.write(data, onParsed); WriteBuffer fires that callback only after the chunk is parsed (WriteBuffer.ts:254). Terminal.tsx:755 now passes () => recovery.noteData(). The buffered (scroll-locked) branch deliberately omits onParsed (no paint -> don't arm), preserving the round-1 fix. No private internals added (callback is typed at xterm.d.ts:1370). The conceded flush-path half is left uninstrumented as agreed in round 1. Client typecheck clean; 27 unit tests pass; biome format clean.
- `F3` **fixed** — Resolved in round 1 and confirmed by codex. renderRecovery.ts onRender disarms the watchdog when !paintIsBehind(), with a regression unit test. No further change in round 2.
- `F4` **fixed** — Resolved in round 1 and confirmed by codex. DiagnosticInfo.tsx danger predicate marks pending rAF dangerous when msSinceLastPaint is null or > PAINT_STALL_WARN_MS. No further change in round 2.

commit: `9fde2eff594c52c53bae340c24db045715742c4e`

### Round 3

**codex** — approved: `true`

All findings from the prior rounds are resolved in the current tree. The remaining F2 concern is addressed: `Terminal.tsx` now passes `recovery.noteData` as xterm's public write callback through `scrollLock.writeData`, and `scrollLock` only threads that callback into the immediate `term.write` path, leaving buffered scroll-lock chunks unarmed as intended. I did not rescan for unrelated issues per the debate close-out constraint.

Findings:
- `F1` · major · resolved — The e2e render-stall scenario previously left `document.hasFocus()` true, allowing the focused watchdog to fire before the no-repaint assertion. (packages/tests/step_definitions/render_recovery_steps.ts:46)
- `F2` · major · resolved — Recovery was previously notified at stream receipt / immediate write call time, before xterm had necessarily parsed the data into its buffer. (packages/client/src/terminal/Terminal.tsx:755)
- `F3` · minor · resolved — A stale watchdog previously stayed armed after a normal render caught up, so later healthy output could trip the old timer and force an unnecessary repaint. (packages/client/src/terminal/renderRecovery.ts:126)
- `F4` · minor · resolved — `msSinceLastPaint === null` was previously treated as `0` in the diagnostic danger predicate, muting the never-painted pending-rAF state. (packages/client/src/DiagnosticInfo.tsx:323)

_codex on the rebuttal:_ F2 is now fixed. CLAUDE accepted the remaining async-write timing issue and moved `noteData()` onto xterm's typed write callback, which fires after the chunk is parsed; that closes the gap I held open. My prior concession on the flush-path half still stands: no separate flush hook is needed for the occlusion-return path because focus/visibility recovery already forces repaint.

srid:

**Pass 1 — Rules:** clean (0 violations across DRY, dead-code, error-swallowing, unbounded-growth, comments, exported-type docs).

**Pass 2 — Fact-check:** 6 findings, adjudicated against the live tree. 1 applied, 5 dismissed with cause.

| # | Finding | Disposition |
| --- | --- | --- |
| C2 | `renderDebouncerPending()` returned `false` (looks healthy) when the `_animationFrame` **key** is absent — a future beta renaming it would read as "no frame pending" rather than "unknown shape" | **Fixed** ([`6963fe4d`](https://github.com/juspay/kolu/commit/6963fe4d)) — return `null` when the key is absent; the probe exists to flag a shape change, so it must not report a renamed field as healthy. +1 regression test. |
| C1 | "double-arm race" between the `watchdog === null` check and the assignment | **False positive** — JS is single-threaded; `noteData()` runs to completion with no yield between check and assignment, so two calls can't interleave. |
| C3 | `term.rows` could be `NaN`/negative → `refreshRows(0, NaN, true)` | **Dismissed** — `term.rows` is stable public xterm API (positive int for a live terminal); `Math.max(0, …)` already clamps. Guarding an impossible input is gold-plating. |
| S4 | `maybeForce()` uses a "stale" staleness judgment | **False positive** — `maybeForce` re-evaluates `paintIsBehind()` at fire time (fresh), and `onRender` disarms the watchdog when a paint catches up (codex F3). |
| S5 | disposal may skip `disarm()` if `onRender.dispose()` throws | **False positive** — `onCleanup` already calls `disarm()` **first**; the watchdog is cleared before `dispose()`. |
| S6 | scroll-lock buffered-write "blind spot" (watchdog not armed for buffered data) | **By design** — while scroll-locked, xterm's buffer is unchanged (data sits in `pending[]`), so there's nothing to repaint; focus-return covers it. Documented in `scrollLock.writeData`. |

**Pass 3 — Elegance:** skipped (`--no-elegance`; `/simplify` ran this gauntlet — all four cleanup angles returned clean).

srid:
## Evidence

The real trigger — a window the OS marks **occluded** while frames stop being produced — can't be reproduced in a headless browser. So this is the **deterministic proof of the mechanism and the fix**, run against the *exact pinned* `@xterm/xterm@6.1.0-beta.225` in a real headless Chromium (off-machine, on an ephemeral box): `requestAnimationFrame` is hooked to *queue without firing* — a faithful stand-in for occlusion freezing frame production — then 60 lines are written.

freeze before/after

**Left — FROZEN:** all 60 lines are parsed into xterm's buffer (`buffer.length` grew 24→62) but **zero** `onRender` firings happened while frames were held, so the screen is stuck on the stale frame (`line-1`). This is your bug: the data is all there, only the paint is missing.

**Right — RECOVERED:** the fix's forced synchronous repaint (`refreshRows(0, rows-1, true)`, the same call the `window` `focus` handler makes) paints to the **latest** (`line-159`) **without** a serviced frame.

The full deterministic run asserted all of it (12/12):

| Stage | Observed |
| --- | --- |
| Freeze | 60 writes, frames held → buffer `24→62`, **paints = 0**, one parked `_animationFrame` |
| Reveal | one serviced frame (a keypress's `BeginMainFrame`) → one coalesced paint → latest |
| Fix | `refreshRows(…, true)` repaints to latest with **no** serviced frame |

**Integrated coverage in the app** (since the headless harness can't occlude): the `render_recovery.feature` e2e stalls the focused terminal's render loop, proves output lands in the buffer unpainted, then asserts a `window` focus event forces a synchronous repaint — **red without the fix's wiring, green with it** — and 10 `renderRecovery` unit tests pin the watchdog / recover / probe state machine. And the new Diagnostic Info render-pipeline fields (time-since-paint, parked-rAF, page focus/visibility) mean that if this ever recurs in the wild, one **Copy JSON** identifies the layer instead of guessing — the thing the two prior PRs lacked.

srid:

The **x86_64-linux** lane ran on **`kolu-ci-6`** (`dev-x86-64-linux-03`) — commit `6963fe4d`, **failed** ✗

- **Lane wall** (pipeline): **3m4s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 2m22s | ✗ (exit 1) |
| `ci::nix` | 46s | ✓ |
| `ci::smoke` | 44s | ✓ |
| `ci::unit` | 43s | ✓ |
| `ci::flake-check` | 39s | ✓ |
| `_ci-setup` | 32s | ✓ |
| `ci::surface-app-example-build` | 29s | ✓ |
| `ci::biome` | 29s | ✓ |
| `ci::fmt` | 28s | ✓ |
| `ci::install` | 10s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/6963fe4/timings.jsonl`; pool state is a live `flock` probe.</sub>

PR #1348, opened 2026-06-14
feat(canvas): agent run-state on tile borders — run/sweep aura

srid:
Surfaces each canvas tile's agent run-state on its **border**, so a sweep of the eye across the canvas tells you which terminal is *working*, which *needs you*, and which just *pinged you and you missed it* — without reading a title bar. Implements the **run/sweep** language (Language C) from the [canvas-tile-state-borders design note](https://github.com/juspay/kolu/pull/1343).

## How it works

One pure mapper, one reactive socket, one rendering — reusing the dock's existing classifiers, so there's no new state and no new clock.

- **`tileAura(bucket, unread, stale) → tier`** — a pure function mirroring the dock's `pipVariant`. `unread` dominates (a missed alert still reads loud); a stale waiter cools to `waiting-stale`, a stale worker parks to `none` so the canvas agrees with the dock.
- **`useTileAura()`** — folds the three existing inputs (`agentBucket`, the unread flag, `useStaleCheck`) into the tier; read per-tile by the canvas tile.
- **The ring** (`.tile-aura`) — *motion* carries the state, in the repo's **one** identity colour:
  - **working → marching ants** stream calmly around the edges
  - **needs-you → a sweeping comet** whose **speed is the urgency**: alert fastest → fresh → stale slowest
  - **alert** thickens the ring and throbs a full-perimeter glow under a brighter, faster comet
- **One colour throughout** — the repo colour drives the resting border, the active ring, and the aura; nothing else competes.
- `prefers-reduced-motion` freezes the motion to a static themed ring.

## Tests

`tileAura.test.ts` covers the full tier truth table.

## Status

In progress — active-tile styling is still being refined. Evidence to follow.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #710, opened 2026-04-24
Adopt Biome v2 for lint + format — add type-aware safety rules, retire Prettier


srid:
## Goal

Replace the current Prettier-only setup with **Biome v2** as the unified formatter + linter, and enable Biome's type-aware lint rules to cover the class of correctness bugs `tsc` alone doesn't catch (floating promises, misused promises, non-exhaustive switches over discriminated unions).

## Motivation

`tsconfig.base.json:6,12` already has `strict` + `noUncheckedIndexedAccess` on. A handful more tsc flags (`noImplicitReturns`, `noFallthroughCasesInSwitch`, `noImplicitOverride`, `verbatimModuleSyntax`) are easy wins and deserve their own separate ticket. But tsc's dials max out around there — the next tier of "Haskell-grade" safety requires a lint layer that understands types:

- **unawaited promises** that swallow errors silently
- **`<button onClick={asyncHandler}>`**-style misuse where the caller expects `void`
- **switch over a discriminated union** that silently drops a new variant
- **dead conditional branches** where the type already rules out the check

None of these are enforceable from `tsconfig`.

## Why Biome (not ESLint, not Oxlint)

- **ESLint + typescript-eslint** — most complete, slow (cold runs in minutes on real monorepos), new toolchain burden, still needs Prettier alongside. No upside over Biome for us unless we hit inference gaps we can't live with.
- **Oxlint + tsgolint** — fastest, full TS-semantic fidelity via the Go port (`tsgolint`). But: alpha as of Dec 2025, 43/59 typescript-eslint rules, targets TypeScript 7 which isn't GA, reported memory issues on large codebases, and it's not a formatter — so still needs Prettier or Biome alongside. Worth revisiting in 2027.
- **Biome v2 "Biotype"** — ships its own type inference (Scanner architecture, sponsored by Vercel). Catches ~75% of what `@typescript-eslint/no-floating-promises` does at 1-2 orders of magnitude faster. Subsumes Prettier. Zero ESLint config to migrate (we don't have one — `package.json:28` shows only Prettier). Fits `.claude/rules/workflow.md`'s "prefer external libraries / less code to maintain" principle.

## Rules to enable (minimum target)

Type-aware (require Scanner / project domain):

- [`noFloatingPromises`](https://biomejs.dev/linter/rules/no-floating-promises/) — unawaited promises.
[...]

PR #1384, opened 2026-06-16
Fail the Biome lint gate on warnings, clear master's debt

### LINKED ISSUE #1362: CI's biome gate doesn't fail on lint warnings — ~20 warning-level findings on master go uncaught
## Summary

CI's lint gate (`ci::biome`) is `biome lint .` (`ci/mod.just:166`), which **exits `0` on warning-severity diagnostics** — it only fails the build on *errors*. There is no `--error-on-warnings`. So warning-level lint debt never blocks a merge and accrues invisibly.

Noticed while running `just check` on an unrelated branch: it reported warnings that live on `master` (in files the branch never touched), and a clean `biome lint .` on `master` confirms them — yet every CI run is green.

## Reproduce

```sh
biome lint .            # prints the warnings below
echo $?                 # → 0   (would be 1 only with errors, or with --error-on-warnings)
```

## Current debt on `master` HEAD

| count | rule |
|---|---|
| 9 | `lint/style/noNonNullAssertion` |
| 7 | `lint/suspicious/noExplicitAny` |
| 2 | `lint/style/noDescendingSpecificity` |
| 1 | `lint/suspicious/noTemplateCurlyInString` |
| 1 | `lint/style/useImportType` |

(~20 findings; e.g. `packages/surface/src/**` `any` usages, an unused `contract` in `stdio.test.ts`.)

## Why it matters

The Biome-adoption umbrella **#710 is closed**, so nothing is ratcheting these to zero or preventing *new* warnings from landing. A warning that's advisory-by-intent is fine; a warning that no gate ever enforces and no issue tracks is silent debt.

## Options (for triage — not changing CI in this PR)

1. **Ratchet:** clear the current ~20, then add `--error-on-warnings` to `ci::biome` so new warnings fail CI.
2. **Be explicit:** for rules intentionally kept advisory, set them to `off` (with a one-line reason) in `biome.jsonc` so the warning list reflects only *real* debt, then gate on what's left.
3. **Track:** if these are deliberately deferred, re-open a tracking issue (the #710 successor) listing them.

Filed from R2 (Pierre highlight-worker) work; that PR does not touch lint config.

srid:
`ci::biome` ran `biome lint .`, which **exits 0 on warning-severity diagnostics** — only errors ever failed the build. So warning-level lint debt accrued on master uncaught (~60 findings by the time this was filed), and nothing stopped new warnings from landing, because the Biome-adoption ratchet umbrella (#710) is closed. This adds `--error-on-warnings` to make the gate real, and clears the existing debt to zero so the gate starts from a clean baseline.

The gate flag lands in three places that must agree — `ci::biome`, `just check`, and `just lint` — so a local `just check` fails on exactly what CI fails on. After this, **there is no silent advisory tier**: every `recommended` rule either fails the build or is explicitly turned off (with a reason) in `biome.jsonc`.

### Triage of the existing ~60 warnings

Each rule class was decided on its merits — fix the real debt, turn off what's deliberately advisory (with a reason), never paper over a real bug:

| Rule | Count | Decision |
| --- | --- | --- |
| `suspicious/noExplicitAny` | 38 (all in `surface/define.ts`) | **Scoped off** via a `biome.jsonc` override — the generic descriptor-map machinery (`Record<string, CellSpec<any, any>>`) needs `any` for variance over heterogeneous specs; `unknown` would reject concrete specs. Two now-redundant inline suppressions removed. |
| `style/noNonNullAssertion` | 14 | **Off, with reason** — `!` is a sanctioned escape hatch here (`Map.get` after `.has`, known-present indices); guards would only add unreachable branches. The load-bearing safety rules (`noFloatingPromises`, `noMisusedPromises`, `noImportCycles`) stay `error`. |
| `style/noDescendingSpecificity` | 3 | **Fixed** — see the a11y note below; one was a real bug, one a redundant rule. |
| unused imports/vars, `useImportType`, `useOptionalChain`, `noConfusingVoidType` | ~7 | **Fixed to zero.** A literal-fixture `noTemplateCurlyInString` and a `typeof`-only contract binding are suppressed inline with reasons. |

> **A latent accessibility bug fell out of the triage.** The `prefers-reduced-motion` reset for the canvas tile auras sat at specificity (0,2,1), *below* the per-state animation rules at (0,3,1) — so it never won the cascade and the marching-ants/comet auras kept moving even when reduced motion was requested (despite #1348 claiming "reduced-motion freezes every aura to a static ring"). `noDescendingSpecificity` caught it; the reset now matches the animating selectors' specificity. Carries a changelog `Fixed` entry.

No `@kolu/surface` API changed (the `define.ts` edit only drops two ignore-comments — same types), so no drishti PR is needed.

Closes #1362.


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `5d57c306b01a`

Independent findings: lowy=3, hickey=1

### Applied (2)
- `lowy-1` Gate strictness flag is a wire exposed at three sites, not one receptacle — commit `95f1e980b`
- `hickey-1` Lint-gate flag triplicated across three recipes — commit `95f1e980b`

### Agreed — no change (2)
- `lowy-2` Per-rule disable rationale and the rule's enforcement live in separate files for define.ts (biome.jsonc:154-172, packages/surface/src/define.ts:706)
- `lowy-3` CSS reduced-motion override correctly co-locates the cascade-volatility with the rule it must beat (packages/client/src/index.css:359-371)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `5d57c306b01a`

### Round 1

**codex** — approved: `false`

The branch is mostly clean: the Biome suppressions match the deliberate decisions, the reduced-motion specificity fix is correct, and the test cleanups are safe. I could not run the full Biome command because the sandbox cannot access the Nix daemon, but `git diff`, `git status --short`, `git diff --check`, and `just --dry-run` were enough to identify one local workflow regression.

Findings:
- `F1` · minor · open — `just check` now runs the top-level `install` dependency, then typecheck, then shells out to `just ci::biome`; `ci::biome` has its own `install` dependency, so the local check path performs `pnpm install` twice. `just --dry-run check` shows the nested call, and `just --dry-run ci::biome` expands to install plus the Biome command. This regresses the stated fast static-correctness gate with redundant package-manager work. (justfile:72)

**claude** — Codex returned one minor finding (F1). I agree with it and fixed it. F1: `just check` ran top-level `install` then `just ci::biome`, but `ci::biome` carries its own `install` dep, so the local fast static-correctness gate performed `pnpm install` twice; `just lint` had the same double-install. This is exactly the anti-pattern the ci::biome comment warns against ('do not invoke top-level recipes that carry their own install dep'). Fixed by switching both recipes to `just --no-deps ci::biome`, which skips the inner install while keeping each recipe's own top-level `install` (so a cold `just lint` still creates node_modules). `just --dry-run check` and `just --dry-run lint` now show install running exactly once; `just --no-deps ci::biome` expands to just the biome command. The single-receptacle gate flag is unchanged — biome still delegates to the one ci::biome recipe.

- `F1` **fixed** — Agreed. justfile:70-75 — `check: install` then `just ci::biome` (which has its own `install` dep) ran pnpm install twice locally; `lint: ci::biome` had the same problem. Changed `check` to call `just --no-deps ci::biome` and made `lint: install` + `just --no-deps ci::biome`. `--no-deps` drops ci::biome's redundant inner install while each recipe keeps its own top-level install (a cold `just lint` still needs node_modules created). Verified with `just --dry-run`: both paths now run `pnpm install` once, and `just --no-deps ci::biome` expands to only the `biome lint --error-on-warnings .` command. Took codex's primary suggestion (--no-deps) over the private-recipe refactor to keep the gate flag in a single recipe per the inherited intent.

commit: `206565bd8331ec1331494a8c668eb04aa2a2e143`

### Round 2

**codex** — approved: `true`

F1 is resolved in the current tree. `check` and `lint` now keep their own top-level `install` coverage for cold local runs while invoking `ci::biome` with `--no-deps`, so the shared Biome gate remains the single source of truth without re-running `pnpm install`. I verified this with `just --dry-run check`, `just --dry-run lint`, and `just --no-deps --dry-run ci::biome`; `git status --short` showed no untracked files.

Findings:
- `F1` · minor · resolved — `just check` previously ran the top-level `install` dependency, then typecheck, then shelled out to `just ci::biome`; `ci::biome` has its own `install` dependency, so the local check path performed `pnpm install` twice. The current version calls `just --no-deps ci::biome` after the top-level install, avoiding the duplicate package-manager work while preserving the strict lint command. (justfile:72)

_codex on the rebuttal:_ CLAUDE agreed with F1 and fixed it. The dry-run output now shows exactly one install in `check` and `lint`, and the delegated Biome command still expands to `biome lint --error-on-warnings .`, so the original concern is addressed.

srid:

✅ **All clear** — base `5d57c306b01a` · `--no-elegance` (the `/simplify` track already ran this pass)

| Pass | Issues | Result |
|------|--------|--------|
| Rules | 0 | All 27 rules checked, clean |
| Fact-check | 0 | No logic errors, silent swallowing, leaks, or unjustified fallbacks |
| Elegance | – | Skipped (`/simplify` ran in this gauntlet) |

Notable validations: `text?.includes(expected) ?? false` is more robust than the prior null-check (handles `null` *and* `undefined`); the reduced-motion CSS specificity bump (0,2,1)→(0,3,1) correctly overrides the animating rules; `noExplicitAny: off` is a file-scoped override with a full rationale; the removed `z` import and `source` variable are genuinely dead.

srid:

The **x86_64-linux** lane ran on **`kolu-ci-1`** (`dev-x86-64-linux-04`) — commit `4c48937b`, **passed** ✓

- **Lane wall** (pipeline): **3m10s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 2m31s | ✓ |
| `ci::nix` | 51s | ✓ |
| `ci::smoke` | 48s | ✓ |
| `ci::unit` | 45s | ✓ |
| `_ci-setup` | 30s | ✓ |
| `ci::fmt` | 30s | ✓ |
| `ci::biome` | 30s | ✓ |
| `ci::install` | 8s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/4c48937/timings.jsonl`; pool state is a live `flock` probe.</sub>

srid:
## Evidence

The a11y fix: under `prefers-reduced-motion: reduce`, the canvas tile-aura animations must **freeze**. Captured on two `pu` boxes — `master` (before) and this PR's branch (after) — each serving kolu from source, driven by Playwright **on the box** with `reducedMotion: 'reduce'` emulated and a tile forced to `data-aura="working"` (exercising the exact shipping CSS selectors).

| Before — `master` (bug) | After — this PR (fixed) |
|---|---|
| before | after |
| marching-ants keep moving despite reduced-motion | aura is frozen to a static ring |

**Deterministic proof** — computed `animation-name` of the working-aura ring, with `prefers-reduced-motion: reduce` active on both:

| build | `.tile-aura::before` | `.tile-aura::after` |
|---|---|---|
| `master` (before) | `tile-aura-ants-x` | `tile-aura-ants-y` |
| this PR (after) | `none` | `none` |

On `master` the reduced-motion reset sat at specificity (0,2,1), below the per-state animation rules at (0,3,1), so it lost the cascade and the animation kept running. The fix raises the reset to `[data-canvas-tile][data-aura] .tile-aura(::before/::after)` (0,3,1), so it wins and the animation resolves to `none`. (The "before" GIF is ~3.7× the byte size of the "after" GIF for the same duration — pure consequence of one looping with motion and the other being static.)

_The rest of the PR is the CI/lint-gate config (`--error-on-warnings`) and per-rule debt triage — no on-screen surface._

PR #1385, opened 2026-06-16
P3: kolu dials remote hosts over ssh (kolu-watcher + N-endpoint registry)

srid:
**kolu-server can now dial a remote host and run terminals on it over ssh** — the full P3 of the [kaval-sessions](docs/atlas/src/content/atlas/kaval-sessions.mdx) plan. `kaval-tui --host` (P2) reached a remote kaval from the *shell*; P3 brings that reach into the kolu app itself: open a terminal on another machine from the command palette's **Connect to host…**, with its files + git mirrored into the Code tab like a local one, the remote session durable across the link, and a per-tile host chip carrying the host's health.

### The shape

A remote host runs **two** processes side by side, dialed over **one** ssh connection:

```
                          ┌──────────────── remote host ────────────────┐
  kolu-server  ──ssh──▶   │  kolu-watcher --stdio   ──unix socket──▶  kaval
  RemoteTerminalEndpoint  │   (fresh per build:                       (durable PTY
   · mirror bridge        │    provider DAG + fs/git,                   survivor)
   · daemonStatus         │    forwards pty verbs to kaval)
                          └─────────────────────────────────────────────┘
```

- **kaval** stays the kolu-agnostic durable PTY daemon (a terminal survives the link).
- **kolu-watcher** (new) runs kolu's *coupled* code — the provider DAG (git/PR/agent detection) + native fs/git — host-side, **fresh per build**. It serves ONE `watcherSurface` over the ssh stdio: it **absorbs** the pty-host verbs/taps and forwards them to the host-local kaval (a `serveOverStdio` server that is a *client* of kaval's socket — not a raw relay), plus serves fs/git + a `terminalMetadata` collection.
- **kolu-server**'s `RemoteTerminalEndpoint` consumes that surface, mirroring the watcher's metadata into the browser surface (the `remote-process-monitor` `bridgeAgentToParent` pattern) and adopting terminals the host already has.

### How it's built (commit by commit)

| | |
| --- | --- |
| **additive contract** | `hostId` on create + the 4 fs/git stream inputs; `location {hostId}` on the terminal record. All optional → old sessions are local, the wire enum is untouched (no drishti). |
| **@kolu/terminal-dag** | the provider DAG + `makeFsGit` extracted into a package both kolu-server and kolu-watcher import — the local path stays byte-identical (only the logger is now injected). |
[...]

PR #1398, opened 2026-06-18
docs(atlas): a from-scratch design for remote terminals

srid:
**A clean-slate design for kolu remote terminals**, written after rejecting #1385 — the "kolu dials remotes" work — as a single 133-file PR. Rather than finish that design in safe layers, this note redesigns from the goal and **dissolves #1385's bug classes by construction** instead of patching them. A plan-of-record Atlas note (`status: proposed`), sibling of [kaval-sessions](docs/atlas/src/content/atlas/kaval-sessions.mdx) — no code changes.

> **P1 has shipped to master** ([#1402](https://github.com/juspay/kolu/pull/1402)) — in a *better* shape than this note first proposed: `location` is a **discriminated union** (`{kind:"local"}` | `{kind:"remote",hostId}`), not a reserved `LOCAL_HOST_ID` string, so a remote host aliased `local` in `~/.ssh/config` can never collide with the in-process endpoint. This note is synced to master and updated to reflect it.

### The core model

One thesis: **the host is the sole source of truth for its own live terminals, and kolu persists only a dial-list** (`Set<hostId>`), never a remote-terminal record. A remote tile is a local `TerminalEndpoint` bound to a transport — *local vs remote is only the link*.

- **Existence from one atomic pull** — boot = restore = reconnect: dial → `await` **kaval's atomic `terminal.list`** (never the watcher's incremental mirror) → synchronously seed `terminalMetadata` → project.
- **Persist only a dial-list** — nothing to prune; a remote terminal can never enter the saved session.
- **Endpoint first-class** — `location` non-optional (a discriminated union); a dropped host is a compile error. **(P1, shipped.)**

### The honest correction (what the adversarial review caught)

"The first slice needs no host agent; the providers just runs in kolu-server" is **provably wrong** — the git/PR/agent providers read kolu-server's *own* `fs`/`os.homedir()`, so in-server they inspect the wrong machine. Awareness is **split**: PTY-tap signals (cwd/title/foreground) run in-server; git/PR/agent run **host-side** — exactly why the watcher exists (renamed the "host agent", minus PTY-forwarding).

### Four bug classes, dissolved by construction

| #1385 bug class | Why it's now unreachable |
| --- | --- |
| cold-boot re-adoption race | existence = kaval's atomic `terminal.list` + sync seed; persist only a dial-list |
| env leak (restricted PATH) | spawn fully-specified; host **login PATH** captured host-side |
| remote-repo prune | prune-stat via `endpointFor(record.hostId).fs`, tolerant of a down host |
| restore-to-wrong-endpoint | `location` non-optional **(P1 ✅)**; restore = dial + adopt, never read-record-and-spawn |

### Ten risk-isolated slices (composed surface fragment — no `directLink`)

**`P1` location non-optional ✅ shipped ([#1402](https://github.com/juspay/kolu/pull/1402))** · `P2` dial-list · `P2b` host-aware feed — **cheap local-only bug-proofing, do first** → `P4a` kaval-direct shell (ssh, no watcher) → `P4w` **lean local refactor** — extract `@kolu/terminal-providers` as a **surface fragment** and **compose it into kolu-server's surface** (the `@kolu/surface-app` pattern — no client, no mirror, no wire) → `P3` boot via kaval's atomic list → `P4d` **the fat phase** — add `bin`/`serveOverStdio` + the cross-arch Nix drv-map/closure-test (a kaval clone) + provisioning, and **serve the same fragment over `stdioLink` + mirror it** (badges **and** Code-tab together) → `P5` health · `P6` host picker · `P9` multiplex.

> **In one process you compose a surface — you don't dial it.** `@kolu/surface` supports composing a sub-surface directly, and `@kolu/surface-app` already does it against kolu-server. So `terminal-providers` ships as a `{ collections, procedures }` contract + a deps fragment, spread into kolu-server's surface: **one router, no client, no `directLink`, no mirror** locally. P4w establishes the **awareness surface + an in-process fold** that projects it onto `terminalMetadata` (seeding from the restored record so B3.3 adoption state isn't clobbered; the persisted-vs-live `terminals:dirty` fence preserved) — the fold is *real* and exercised locally first, **not** an identity (it merges the host-derived subset into the unified record). What's remote-only (P4d) is the **mirror** that copies the served awareness collection across the `stdioLink` wire, with its genuine lost-update concern; the *same* fold then runs. This replaces the earlier `directLink`/stream-replay machinery the draft carried (closed #1406) — the in-process *wire* was fictional, the fold was never the problem. **Caveat:** P2–P4d map to commits on the still-open #1385, not merged trunk.

### Preview

Self-contained Atlas page (Code tab: `docs/atlas/dist/remote-terminals-from-scratch.html`):

```
https://htmlpreview.github.io/?https://github.com/juspay/kolu/blob/docs/remote-terminals-from-scratch/docs/atlas/dist/remote-terminals-from-scratch.html
```


srid:
Superseded by a clean-slate redesign of the awareness layer. An adversarial design panel (cost-blind, perfection-bar) concluded the from-scratch direction here carried avoidable complexity (the awareness split, the in-kolu compose-vs-serve fork, a fictional in-process boundary). The replacement: **`awarenessd`** — an ephemeral provider daemon *on top of* kaval that exposes the provider-authored slice of `TerminalServerMetadata` as one `@kolu/surface` collection, composed in-process for local terminals and served+mirrored+folded for remote ones. Ships standalone first (`awarenessd` + `awareness-tui`, the kaval-tui move one level up), then kolu adopts it. New design note + PR to follow; closing this so the old framing doesn't linger. PR #1409 (the P4w extraction) was also closed.

PR #1401, opened 2026-06-18
Track the pointer when selecting text in a zoomed canvas terminal

### LINKED ISSUE #1400: Mouse selection offset in canvas mode terminals — CSS transforms not accounted for by xterm.js
## Problem

When running terminals in canvas mode with pan/zoom, mouse-based text selection is imprecise — the selection starts at an offset from where the mouse pointer actually is. This makes it difficult to select text accurately, especially when running interactive TUI applications like opencode.

Additionally, when a tile loses focus (another window is focused), the terminal content is **redrawn with different proportions** — text reflows ~7.7% wider. When focus returns, it reflows back. This is visually jarring and compounds the selection imprecision.

The issue does not occur:
- In standalone terminals (foot, alacritty) running the same applications
- In kolu's non-canvas mode (split/panel terminals)
- When canvas zoom is exactly 1.0 with no pan

## Root cause

This is a compound issue with two interacting causes:

### 1. Upstream xterm.js WebGL/DOM cell-width divergence (#1306)

**Tracked upstream in [xtermjs/xterm.js#6015](https://github.com/xtermjs/xterm.js/issues/6015) and in kolu as #1306.**

kolu's "single WebGL context" strategy (`Terminal.tsx:171-185`) unloads the WebGL renderer when a tile loses focus, falling back to xterm's DOM renderer. The two renderers calculate cell dimensions differently:

- `WebglRenderer._updateDimensions` floors the device char width: `Math.floor(charW × dpr)` → 8px
- `DomRenderer._updateDimensions` does not floor → 8.6154px

Measured live in kolu (FiraCode Nerd Font, fontSize 14, dpr 1):

| renderer | `dimensions.device.cell.width` |
|---|---|
| webgl (focused) | **8** |
| dom (unfocused) | **8.6154** |

→ **+7.7% wider glyph grid** on the unfocused tile, matching the text reflow seen on focus swap. This is the "stable font on focus" wobble named in `SettingsPopover`.

When focus returns, WebGL reloads and the grid snaps back to the floored width.

**Status:** Waiting on upstream fix. When xterm.js fixes it, kolu bumps the pnpm-overrides pin.

### 2. Canvas tile CSS transforms not accounted for by xterm.js

Canvas tiles apply CSS transforms for pan/zoom via `tileTransformCSS()` in `packages/client/src/canvas/viewport/coordinates.ts:35-47`:

```ts
const tx = layoutX * (zoom - 1) - panX * zoom + dragX;
const ty = layoutY * (zoom - 1) - panY * zoom + dragY;
return `translate(${tx}px, ${ty}px) scale(${zoom})`;
```

xterm.js calculates mouse positions for text selection using the terminal element's position, but does not account for CSS transforms on ancestor elements. When the tile has a transform applied, the visual position differs from the logical position used for hit-testing.

### Interaction

The two causes compound each other:
1. The cell-width divergence means xterm's hit-testing uses metrics that don't match the visual rendering
2. The canvas transforms add a second coordinate mismatch on top of that
3. Result: selection offset is particularly noticeable in canvas mode with zoom/pan

## Related issues

- **#1306** — Track upstream xterm.js WebGL/DOM cell-width divergence (focus-swap text reflow). The upstream cause of the proportion change on focus swap.
- **#1305** — Fractional tile transforms cause sub-pixel hairline seams in DOM-rendered tiles. Related tile transform issue.
- **#1307** — docs(perf): renderer-fidelity learnings from the #1299 investigation. Root-cause analysis documenting both defects.

## Evidence

1. **Standalone terminals work:** foot, alacritty running the same opencode sessions have precise selection and no reflow
2. **Non-canvas kolu terminals work:** split/panel mode terminals select correctly (no canvas transforms applied)
3. **Canvas mode broken:** terminals in canvas mode (with transforms) have offset selection and reflow on focus swap
4. **Transform-dependent:** the offset varies with zoom level and pan position
5. **Renderer-dependent:** the reflow happens precisely when WebGL ↔ DOM swap occurs on focus change

## Impact

Users cannot accurately select text in canvas mode terminals, which is the primary use case for running interactive TUI applications. The visual reflow on focus swap is also jarring and degrades the experience of context-switching between windows.

### Short-term (kolu-side, addresses #2 only)

**Inverse-transform mouse coordinates:** Intercept mouse events on the canvas tile and apply the inverse transform to `clientX/clientY` before xterm.js processes them. This fixes the transform-induced offset but not the cell-width divergence.

### Long-term (upstream, addresses #1)

**Wait for xtermjs/xterm.js#6015:** When upstream fixes the cell-width divergence, the reflow disappears and hit-testing metrics match visual rendering. kolu only needs to bump the pnpm-overrides pin.

### Alternative (kolu-side, mitigates #1)

**Budgeted WebGL on N most-recently-focused tiles:** Keep WebGL contexts alive on recently-focused tiles instead of unloading on every focus loss. This avoids the WebGL → DOM swap entirely but requires managing Chrome's ~16-context-per-tab budget more carefully. See #1306 for discussion.

## Environment

- Canvas mode with pan/zoom active
- Any zoom level != 1.0 or non-zero pan
- Interactive TUI applications (opencode, etc.) running in the terminal
- Linux (Wayland), macOS, Windows (platform-independent)

srid:
**Dragging to select text in a terminal tile while the canvas was zoomed landed off the mouse** — the selection started at an offset from the pointer that grew the more you zoomed and the further you reached from the tile's top-left, turning text-grabbing in an interactive TUI (opencode and friends) into a guessing game. Closes #1400.

Canvas tiles scale with a CSS `transform` to zoom (`tileTransformCSS`). xterm hit-tests the mouse by subtracting the screen element's `getBoundingClientRect()` — which *folds in* that transform — then dividing by the **untransformed** CSS cell size. The two disagree, so the computed cell drifts by the zoom factor *in both axes*, growing with distance from the tile origin. This is cause #2 in the issue; cause #1 (the upstream WebGL/DOM cell-width divergence) stays with xterm.

### The seam

Both coordinate entry points xterm exposes live on one private service and funnel through the same transform-blind helper:

```
SelectionService (selection + links) ─▶ getCoords ───────┐
                                                          ├─▶ _mouseCoordsService ─▶ getCoordsRelativeToElement
MouseService (TUI mouse reporting) ─▶ getMouseReportCoords┘        (reads getBoundingClientRect)
```

Wrapping that single service is the whole fix: each method inverse-scales the event point (`unscaleEventPoint`) before delegating to xterm's own unchanged math, so there's no cell metric to keep in sync. Scale is measured straight from the DOM (`rect.width / offsetWidth`), which makes it:

- a **strict no-op for untransformed terminals** — split/sub-panel tiles and zoom = 1 hit the identity branch, so existing selection is provably unchanged;
- **decoupled from the canvas** — no new props, correct under any ancestor transform;
- **reconnect-safe** — `reset()` (kolu's stream-retry path) never recreates the service, so the patch lives for the terminal's lifetime.

One residual is called out in the code: xterm's drag-past-the-edge **auto-scroll** reads coordinates via a free function the service wrap can't reach, so its *velocity* stays zoom-scaled — but that never changes *which* cell is selected.

### Reproduced test-first

`features/canvas-selection.feature` prints a block of `LLL…RRR` rows, zooms the canvas, then drives a real mouse drag over a span that *visually* sits in the `L` half and asserts the selection is all `L`:

| | Selection returned |
| --- | --- |
| Before (offset) | `RRRRRRRRRRRRR` — landed a row down and in the right half |
| After (fix) | all `L`, under the pointer |

`unscaleEventPoint` is unit-tested for the identity, pure-translate, non-uniform, and scaled cases.

### Try it locally

```sh
nix run github:juspay/kolu/urban-suit
```


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `6be15623a347`

Independent findings: lowy=5, hickey=2

### Applied (4)
- `lowy-2` unscaleEventPoint silently encodes the tileTransformCSS fixed-point contract, with no link enforcing it — commit `6ca55cd63`
- `lowy-3` TRANSFORM_EPSILON doc comment opens with a broken/garbled sentence — commit `6121da79c`
- `hickey-1` Two opposite pointer→cell strategies for the same under-zoom hit-test invariant — commit `728f571d0`
- `hickey-2` Garbled TRANSFORM_EPSILON doc comment raises reasoning load — commit `6121da79c`

### Agreed — no change (3)
- `lowy-1` Canvas-scale volatility is re-derived from the DOM instead of plugged into the existing transform receptacle (packages/client/src/terminal/xtermInternals.ts:140-141 (unscaleEventPoint) and 213-223 (patchTransformAwareMouseCoords.corrected))
- `lowy-4` Test layer duplicates canvas_steps.ts zoom helpers (data-zoom read + ctrl+wheel dispatch) (packages/tests/step_definitions/canvas_selection_steps.ts:66-72 (canvasZoom) and 88-103 (WheelEvent dispatch))
- `lowy-5` Two cell-pixel hit-test mappings now coexist in Terminal.tsx (transform-correct touch path vs. transform-blind xterm path) — confirms the bug is the missing seam, not the wrapper (packages/client/src/terminal/Terminal.tsx:548-567 (fileRefAtPoint) vs. xtermInternals.ts patch)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 1 round(s) · codex reviewed at `xhigh` reasoning effort · base `6be15623a347`

### Round 1

**codex** — approved: `true`

I reviewed the full diff against 6be15623a347e1f47564c0403b8705ca5e024728, confirmed there are no untracked files to include, and traced the xterm private mouse-coordinate call sites, canvas transform contract, terminal lifecycle, and Cucumber/Vitest wiring. I did not find any correctness, lifecycle, test-coverage, or simplicity issues worth raising in this pass. The private API dependency remains an inherent risk, but the patch is null-guarded, localized in the existing internals module, and covered by both geometry tests and an e2e regression for the zoomed-selection behavior.

Findings:
- _(none)_

srid:

Ran the rules pass and the fact-check (correctness/rigor) pass over the change (elegance skipped — `/simplify` already ran this gauntlet).

| Pass | Findings | Outcome |
| --- | --- | --- |
| Rules | 0 | Clean — built-in + project rules, zero violations |
| Fact-check | 2 | Both fixed (`ac000d06`) |
| Elegance | — | Skipped (`--no-elegance`; simplify covered it) |

**Fact-check fixes**

- **`unscaleEventPoint` divided by a zero scale.** The `layoutWidth > 0` guard left `scaleX`/`scaleY` at `0` when `rect.width`/`height` is `0` (detached / off-screen element), so `(clientX - rect.left) / scaleX` returned `±Infinity`. Now guards both ends and falls back to the identity for a zero-sized rect.
- **`markerCellPixel` (e2e) yielded `NaN` pixels on a zero-width rect.** `rect.width / cols` → zero cell width → all coords collapse to `rect.left`, silently wrong. Now returns `null` instead.

A third observation (the test's `null` return doesn't distinguish "marker absent" from "geometry unmeasurable") was dismissed: both paths return `null` and the test env always has a visible terminal — extra branching for no debugging value.

_Posted by `/be` §4 after the gauntlet push._

srid:
## Evidence



▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=canvas-selection-zoom-1400.mp4

Drag-selecting over the left "L" half of a zoomed canvas tile now correctly highlights the L cells under the pointer — fixing the coordinate-scaling bug (#1400) where the selection landed on the wrong cells.

Mobile/touch axis: N/A — the canvas tile is not rendered on mobile.

srid:

The **x86_64-linux** lane ran on **`kolu-ci-2`** (`dev-x86-64-linux-04`) — commit `ac000d06`, **passed** ✓

- **Lane wall** (pipeline): **2m54s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 2m4s | ✓ |
| `ci::smoke` | 45s | ✓ |
| `ci::nix` | 41s | ✓ |
| `ci::flake-check` | 38s | ✓ |
| `ci::unit` | 31s | ✓ |
| `_ci-setup` | 31s | ✓ |
| `ci::pnpm-hash-fresh` | 26s | ✓ |
| `ci::install` | 18s | ✓ |
| `ci::biome` | 17s | ✓ |
| `ci::fmt` | 16s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/ac000d0/timings.jsonl`; pool state is a live `flock` probe.</sub>

Issue #1306, opened 2026-06-12
Track upstream xterm.js WebGL/DOM cell-width divergence (focus-swap text reflow)


srid:
Split from #1299 (see [the root-cause analysis](https://github.com/juspay/kolu/issues/1299#issuecomment-4694104573)). This tracks the second, upstream-owned component of the artifact: text reflows (~7.7% wider) whenever a tile swaps WebGL → DOM on focus loss.

## Upstream issue

**xtermjs/xterm.js#6015** — `WebglRenderer._updateDimensions` floors the device char width (`Math.floor(charW × dpr)`, `WebglRenderer.ts:636` in our pinned `0.20.0-beta.224`), `DomRenderer._updateDimensions` does not (`DomRenderer.ts:138` in `6.1.0-beta.225`). Same `charSizeService.width`, different cell grid.

Measured live in kolu (FiraCode Nerd Font, fontSize 14, dpr 1, `charW = 8.6154`):

| renderer | `dimensions.device.cell.width` |
|---|---|
| webgl (focused) | **8** |
| dom (unfocused) | **8.6154** |

→ +7.7% wider glyph grid on the unfocused tile, matching the linear word-offset spread measured in the #1299 screenshots. This is the long-standing "stable font on focus" wobble named in `SettingsPopover`.

## What happens when upstream fixes it

Nothing kolu-side: both values are internal to xterm's renderers, kolu has no code on that path. Pickup = bump the pnpm-overrides pin (we already track 6.1.0 betas) + lockfile. Keep this issue open until the pin lands.

## Caveats / follow-ups

- Metric parity stops the *reflow*, but glyph rasterization still differs (canvas atlas AA vs native browser text AA), so a subtle font-smoothing change on focus swap will remain. Eliminating it entirely is a renderer-policy question (budgeted WebGL on N most-recently-focused visible tiles) — deliberately **not** in scope here.
- When this closes, revisit the `SettingsPopover` renderer-option copy: "stable font on focus" describes a problem that should no longer exist.

Refs: #1299, xtermjs/xterm.js#6015.

Issue #591, opened 2026-04-17
Kolu memory usage, and "Too many active WebGL contexts" Chrome warnings


srid:
Kolu chrome tab takes ~2GB over time. And browser console has these errors. I thought we addressed memory leak in https://github.com/juspay/kolu/pull/578 - but perhaps we are missing something more.

```
index-BwCOzxkO.js:559 Uncaught (in promise) Zo: Internal server error
    at pO (index-BwCOzxkO.js:559:73947)
    at AB.decode (index-BwCOzxkO.js:566:13355)
    at async index-BwCOzxkO.js:566:9379
    at async Proxy.r (index-BwCOzxkO.js:566:1858)
index-BwCOzxkO.js:559 Uncaught (in promise) Zo: Internal server error
    at pO (index-BwCOzxkO.js:559:73947)
    at AB.decode (index-BwCOzxkO.js:566:13355)
    at async index-BwCOzxkO.js:566:9379
    at async Proxy.r (index-BwCOzxkO.js:566:1858)
WARNING: Too many active WebGL contexts. Oldest context will be lost.
WARNING: Too many active WebGL contexts. Oldest context will be lost.
WARNING: Too many active WebGL contexts. Oldest context will be lost.
WARNING: Too many active WebGL contexts. Oldest context will be lost.
WARNING: Too many active WebGL contexts. Oldest context will be lost.
WARNING: Too many active WebGL contexts. Oldest context will be lost.
WARNING: Too many active WebGL contexts. Oldest context will be lost.
WARNING: Too many active WebGL contexts. Oldest context will be lost.
WARNING: Too many active WebGL contexts. Oldest context will be lost.
WARNING: Too many active WebGL contexts. Oldest context will be lost.
WARNING: Too many active WebGL contexts. Oldest context will be lost.
WARNING: Too many active WebGL contexts. Oldest context will be lost.
WARNING: Too many active WebGL contexts. Oldest context will be lost.
WARNING: Too many active WebGL contexts. Oldest context will be lost.
WARNING: Too many active WebGL contexts. Oldest context will be lost.
WARNING: Too many active WebGL contexts. Oldest context will be lost.
WARNING: Too many active WebGL contexts. Oldest context will be lost.
WARNING: Too many active WebGL contexts. Oldest context will be lost.
WARNING: Too many active WebGL contexts. Oldest context will be lost.
WARNING: Too many active WebGL contexts. Oldest context will be lost.
[...]

Issue #575, opened 2026-04-16
perf: WebGL context exhaustion in canvas mode — switch inactive tiles to DOM renderer


srid:
## Problem

In canvas mode with many terminals, Chrome tab memory balloons (reported: 4GB) and all terminals go blank for a few seconds when switching between them. Root cause: **WebGL context exhaustion**. Chrome limits WebGL contexts per tab (~16). Each xterm uses its own WebGL context via `@xterm/addon-webgl`. Canvas mode renders all terminals simultaneously (all have `data-visible`), so every tile holds a live context.

The blanking is the context-loss recovery flow (`WebglRenderer.ts:110-131`):
- Chrome evicts → `webglcontextlost` fires → xterm starts 3-second timer
- During 3s: last frame stays pixel-wise, no new renders
- Chrome restores within 3s → xterm clears atlas, reinitializes, redraws → seamless
- 3s elapses without restoration → `onContextLoss` fires, canvas blanks permanently (host app must handle)

Chrome typically restores, which is why terminals come back after a few seconds.

## `preserveDrawingBuffer` compounds the issue

PR #574 (screenshot feature) set `preserveDrawingBuffer: true` on every WebGL addon. Per WebGL spec, this keeps the backbuffer allocated after compositing instead of discarding it — **roughly doubles per-context GPU memory**. It's needed for `canvas.toDataURL()` to work reliably, but only when screenshotting.

Consider: toggle `preserveDrawingBuffer` on-demand (enable → snapshot → disable) instead of always-on. Or reconstruct the addon before taking a screenshot.

## Proposed fix

**Dispose `WebglAddon` on non-focused tiles; xterm auto-falls back to the DOM renderer.**

`WebglAddon.dispose()` (`WebglAddon.ts:90-97`) replaces the renderer with a `DomRenderer` via `renderService.setRenderer()`. The switch is synchronous, redraws only the viewport (O(visible_rows), not O(scrollback)), 1-2 frame flash.

- **Focused tile**: WebGL renderer (fast, GPU-accelerated)
- **Non-focused tiles**: DOM renderer (slower, no GPU context)
- **On focus change**: dispose current tile's WebGL, reload WebGL on newly focused tile
- **Context loss listener**: subscribe to `onContextLoss` and proactively dispose WebGL on that tile, fall back to DOM

## Renderer comparison (from xterm v6.0.0 source)

| | WebGL | DOM |
|---|---|---|
[...]

PR #1404, opened 2026-06-18
Budget WebGL to the 2 most-recently-active tiles — kill the focus-swap reflow

### LINKED ISSUE #1403: Canvas: budget WebGL to the 2 most-recently-active tiles (active split inherits) — fix the focus-swap reflow without waiting on upstream
## Background

[#1400](https://github.com/juspay/kolu/issues/1400)'s **cause #1** (tracked as #1306): when a canvas tile loses focus, kolu unloads its WebGL context and falls back to xterm's DOM renderer (the #575 single-context policy). The two renderers compute cell width differently — `WebglRenderer` floors `Math.floor(charW × dpr)`, `DomRenderer` doesn't — so the swapped tile's text **reflows ~7.7% wider**, snapping back when WebGL reloads on refocus. Jarring on every focus change.

We'd parked this as "waiting on the upstream xterm fix, then bump the pnpm-overrides pin." **That plan is now dead:** upstream has effectively **won't-fixed** it — see jerch's comment on [xtermjs/xterm.js#6015](https://github.com/xtermjs/xterm.js/issues/6015#issuecomment-4694852375):

> un-flooring the WebGL cell → blurred chars / moiré; correcting the DOM renderer with `letter-spacing` → a perf regression (re-introduces span pressure the letter-spacing refactor removed). "Swapping renderers forth and back is a weird use case."

His perf reasoning is sound, so the pin bump won't come. (Cause #2 — the selection-offset-under-zoom bug — is already fixed and merged, and filed upstream as xtermjs/xterm.js#6023.)

## Decision

Own cause #1 kolu-side. We don't need the two renderers to produce identical metrics — we need to **avoid the swap that exposes the difference.**

## Proposed fix — N=2 WebGL budget, by tile, with active-split inherit

Today (`Terminal.tsx`): `canUseWebgl = props.visible && props.focused !== false` → exactly the focused tile holds WebGL (1 context). Every tile switch swaps the previous tile WebGL→DOM → reflow.

Change it to keep WebGL on the **2 most-recently-active tiles**:

- **N=2 is the sweet spot.** The reflow is most jarring when you ping-pong between two terminals (you're looking back and forth). With N=2, **A↔B toggling never crosses the WebGL↔DOM boundary** — the reflow disappears for the dominant pattern. N=1 can't do this; general-N would need real cap management.
- **Budget by *tile*, and the active split inherits.** A held tile's renderer covers its main pane **and** its active split (the sub-panel's existing "remember last focused" pointer). Sub-terminals don't compete for slots — they ride their tile's slot. This avoids an intra-split flick (focusing into a visible-but-DOM split pane) and the 7.7% divergence appearing *side-by-side* inside one tile. Applies to **both** held tiles, so switching back is flick-free on either pane.
- Predicate:
  - main/tile → `tile ∈ top-2`
  - sub → `parentTile ∈ top-2 AND sub === parentTile.activeSplit`

**Budget-safe by construction:** ≤ 2 tiles × (main + active split) = **≤ 4 live contexts**, far under Chrome's ~16/tab cap (#575) — so no eviction-flicker and no #575 regression, and N=2 sidesteps the grant-before-evict cap management that general-N would require. Reuses the existing `loadWebgl`/`unloadWebgl` + `loseContext()` disposal, so the #591 leak invariants are preserved.

## Scope of work

- New tiny singleton: a signal-backed MRU of **tile** ids (cap 2) + `holdsWebgl(id)`.
- Feed it: tile activation (`store.activate` / `activeId`); the per-tile active-split pointer from the sub-panel "remember last focused".
- Redirect `canUseWebgl` → `holdsWebgl`. **Keep `isFocused` distinct** (it still drives the zoom gate and `data-focused`) — `holdsWebgl` and "is focused" now diverge.
- Remove-on-close so a closed terminal doesn't pin a slot.
- Tests: update `canvas.feature` "WebGL context is held only by the focused tile" → "held by the 2 most-recently-active tiles (+ active split)"; add an A→B→A switch asserting no `data-renderer` swap on A.

## Residual (accepted, rare)

- A 3+-pane tile still leaves a non-active **visible** sub on DOM, and toggling between two subs *within one tile* can swap. The common tile + one-split case is fully flick-free.
- Bringing in a **3rd** tile evicts the LRU tail → it reflows on the way out (unavoidable without unbounded contexts).

## Cheaper stopgap (if this is deprioritized)

Default `terminalRenderer: "dom"` — already exists ("stable font on focus swap, lower GPU"). Zero new code, but loses WebGL perf/crispness for everyone.

## Docs to re-point

[#1400](https://github.com/juspay/kolu/issues/1400)'s body and `docs/perf-investigations/renderer-fidelity-learnings.md` still say "waiting on upstream / bump the pin" — update them to record the won't-fix and point cause #1 here.

---

Related: #1400, #1306, #575, #591 · upstream xtermjs/xterm.js#6015 (won't-fix), #6023.

srid:
**Switching focus between canvas tiles used to subtly resize the text of the tile you left** — it reflowed ~7.7% wider and snapped back when you returned. kolu handed its single WebGL renderer to the focused tile and dropped every other tile to xterm's DOM renderer, and the two measure cell width differently (`WebglRenderer` floors `charW × dpr`, `DomRenderer` doesn't). This is #1400's **cause #1** (#1306). Upstream **won't reconcile the metrics** ([xtermjs/xterm.js#6015](https://github.com/xtermjs/xterm.js/issues/6015#issuecomment-4694852375): un-flooring blurs glyphs, the `letter-spacing` fix regresses perf), so kolu avoids the **swap** rather than the difference.

`canUseWebgl` now keys off a **WebGL budget** instead of `props.focused`: keep WebGL on the **2 most-recently-active tiles**, each slot also covering that tile's **active split**.

### Why N=2, by tile, with active-split inherit

- **N=2** — the reflow is most jarring when you ping-pong between two terminals. With two budget slots, the dominant **A↔B** toggle never crosses the WebGL↔DOM boundary, so the wobble is gone for it. N=1 can't; general-N would need real cap management.
- **By tile, active split inherits** — a held tile's slot covers its main pane *and* its active split. Focusing into a split never drops the main pane to DOM, so the 7.7% divergence can't appear *side-by-side* inside one tile. Predicate:
  - main / tile → `tile ∈ top-2`
  - sub → `parentTile ∈ top-2 AND sub === parentTile.activeSplit`
- **Budget-safe** — ≤ 2 tiles × (main + active split) = **≤ 4 live contexts**, far under Chrome's ~16/tab cap (#575); reuses the existing `loadWebgl`/`unloadWebgl` + `loseContext()` so the #591 leak invariants hold.

### Shape — a derivation, not a new MRU

`holdsWebgl` lives in `useTerminalStore` beside its sibling resolver `focusedId`, intersecting the **existing** tile `mruOrder` with the live top-level tiles (so a closed tile drops out instead of pinning a slot — no explicit remove-on-close) and reading the per-tile `activeSubTab` for splits. `isFocused` stays distinct: it still drives zoom and `data-focused`, and a budgeted-but-unfocused tile keeps WebGL.

```
mruOrder ∩ live tiles ──slice(0,2)──▶ webglTileBudget
                                          │
   holdsWebgl(id) = id is a budgeted tile, OR
                    id's parent is budgeted AND id is its activeSplit
                                          │
        canUseWebgl = props.visible && holdsWebgl(terminalId)
```

### Tests (test-first — red on the old N=1 policy)

- `canvas.feature`'s WebGL scenario → **3-tile budget + A↔B no-swap**: 3 tiles → exactly 2 on WebGL; clicking between tiles 1 and 2 leaves *both* on WebGL (under N=1 the one you left would swap to DOM here).
- New **active-split-inherit** scenario: focusing a split keeps the main pane on WebGL.
- The `"webgl"` preference scenario widened to 3 tiles so **auto (2)** vs **forced (3)** still contrasts.

### Residuals (accepted, rare)

- A 3+-pane tile still leaves a non-active *visible* sub on DOM; toggling between two subs *within one tile* can swap.
- Bringing in a **3rd** tile evicts the LRU tail → it reflows on the way out (unavoidable without unbounded contexts).
- The cheaper stopgap remains for anyone who wants rock-steady text everywhere: the renderer setting's **dom** option.

Docs: `renderer-fidelity-learnings.md` and `memory-learnings.md`'s WebGL invariants re-pointed from "waiting on upstream / bump the pin" to the budget; changelog entry added.

Closes #1403. Related: #1400, #1306, #575, #591 · upstream #6015 (won't-fix), #6023.

### Try it locally

```sh
nix run github:juspay/kolu/webgl-n2-budget
```


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `a83c1cf0b410`

### Applied (2)
- `lowy-1` Stale `shouldUseWebgl` doc-comment still describes the retired N=1 policy, leaking the old implementation through the receptacle's narration — commit `667095e26`
- `hickey-1` auto-policy spec fragmented across two adjacent, now-contradicting comments — (uncommitted)

### Agreed — no change (5)
- `lowy-2` `webglTileBudget` recomputes the active-split mapping nowhere — but `holdsWebgl` calls `subPanel.getSubPanel`, which mutates store state during a reactive read (packages/client/src/terminal/useTerminalStore.ts:87-90)
- `lowy-3` `WEBGL_TILE_BUDGET = 2` is one of two coupled magic numbers (N tiles and the derived `2*(main+split)=4` ceiling) with the ceiling living only in prose (packages/client/src/terminal/useTerminalStore.ts:32)
- `lowy-4` `holdsWebgl` re-derives tile-vs-split from `parentId` instead of reusing the existing `getSubTerminalIds` family — a second place that knows the parent/child shape (packages/client/src/terminal/useTerminalStore.ts:85-90)
- `hickey-2` tile→terminal split resolution duplicated between focusedId and holdsWebgl (packages/client/src/terminal/useTerminalStore.ts:83)
- `hickey-3` N=2 budget rationale triplicated across two code comments and the doc (packages/client/src/terminal/useTerminalStore.ts:20)

srid:
## Codex ⇄ Claude debate

### Round 1

**codex** — approved: `false`

The core WebGL-budget implementation looks consistent with the existing top-level active-id and sub-panel state model, and the new e2e scenarios cover the main two-tile and active-split cases. I did not run the suite because this review is constrained to read-only/no-write actions. I found stale renderer-policy copy that now contradicts the new Auto behavior.

Findings:
- `F1` · minor · open — The Settings popover still tells users Auto means “WebGL on focused tiles, DOM elsewhere,” but this branch changes Auto to keep WebGL on the two most-recently-active tiles plus active splits. That user-facing hint now describes the old policy and can make users misunderstand why an unfocused tile still shows WebGL. (packages/client/src/settings/SettingsPopover.tsx:46)
- `F2` · nit · open — Two internal comments still describe the old single-context invariant: `unloadWebgl()` says `loseContext()` keeps the live set at 1, and the effect comment at line 312 says it hands “the single WebGL context” to the focused tile. After this change the intended live set is budgeted, up to four contexts. (packages/client/src/terminal/Terminal.tsx:249)

**claude** — Codex returned two findings, both stale-comment/copy issues left over from the old single-focused-tile WebGL policy that this branch replaced with the #1403 recency budget (≤4 contexts: 2 MRU tiles + their active splits). Both were factually correct, so I fixed both. F1: SettingsPopover Auto hint and the option comment no longer say "WebGL on focused tiles" — they now describe WebGL riding the recently-active tiles plus their active splits, pointing at canUseWebgl/#1403. F2: in Terminal.tsx, unloadWebgl()'s comment no longer claims loseContext() "keeps the live set at 1" (now: within the WebGL budget ≤4), and the shouldUseWebgl effect comment now describes reconciling each terminal against the budget rather than handing out "the single WebGL context". Changes are comment/copy-only; `just check` (typecheck + biome) and `just fmt` are clean. Committed as 3f070d5d8.

- `F1` **fixed** — Agreed — the Auto hint at SettingsPopover.tsx:46 ('WebGL on focused tiles, DOM elsewhere.') and the option comment at line 31 ('WebGL on focused, DOM on others') described the pre-#1403 policy and contradicted the new budgeted Auto behavior. Updated the hint to 'WebGL on recently-active tiles, DOM elsewhere.' and rewrote the comment to say WebGL rides the recently-active tiles plus their active splits (see canUseWebgl, #1403).
- `F2` **fixed** — Agreed — both comments asserted the old single-context invariant. unloadWebgl()'s comment (Terminal.tsx ~248-249) said loseContext() keeps 'the live set at 1'; changed to 'within the WebGL budget (≤4 — the 2 most-recently-active tiles plus their active splits, #1403)'. The shouldUseWebgl effect comment (~312) said 'Hand the single WebGL context to whichever tile is focused+visible'; rewrote it to describe reconciling this terminal against shouldUseWebgl() — load when it enters the budget, unload when it leaves, holding ≤4 contexts.

commit: `3f070d5d8ba43395acf29c275cdbaa839df48955`

### Round 2

**codex** — approved: `true`

Claude's follow-up changes address both previously raised issues. The settings Auto renderer copy now describes the recency-budget behavior instead of the old focused-only policy, and the terminal WebGL lifecycle comments now describe budget reconciliation and the ≤4-context invariant. I saw no regression introduced by this comment/copy-only round.

Findings:
- `F1` · minor · resolved — The Settings popover previously told users Auto meant focused-only WebGL, contradicting the new two-MRU-tile budget. (packages/client/src/settings/SettingsPopover.tsx:31)
- `F2` · nit · resolved — Internal WebGL comments previously described the old single-context invariant after the budget changed to up to four live contexts. (packages/client/src/terminal/Terminal.tsx:249)

_codex on the rebuttal:_ F1: verified fixed; the stale focused-only Settings copy is gone. F2: verified fixed; the stale single-context comments are gone and the new wording matches the #1403 budget model. No disputes to adjudicate.

srid:

Ran Pass 1 (rules) + Pass 2 (fact-check); Pass 3 (elegance) skipped — `/simplify` already ran this tree in the gauntlet. **3 findings, all fixed.**

| # | Pass | Finding | Fix | Commit |
| 1 | Rules · `no-dead-code` | `"the focused canvas tile should use the webgl renderer"` had zero feature-file callers after the WebGL scenarios were rewritten | Deleted the step definition | `6a6ec16e` |
| 2 | Fact-check | `"canvas tile {int} should use the {word} renderer"` indexed `[data-terminal-id][data-visible]`, which includes sub-terminals — an open split would shift tile indices | Scoped to `:not([data-sub-terminal])` so the index always means a top-level tile | `6a6ec16e` |
| 3 | Fact-check | `holdsWebgl`'s split branch granted WebGL to a *collapsed* split (its `activeSubTab` can still be set), relying on the caller's `props.visible` to suppress it | Added `!panel.collapsed` (mirrors `focusedId`) so the predicate is self-contained; behavior-preserving today | `4713f429` |

`just check` + `just fmt` clean; `canvas.feature` 43/43 green after the fixes.

srid:
## Evidence

Captured on two ephemeral `pu` boxes — one at `master`, one at `webgl-n2-budget` — using the project's Cucumber + Playwright e2e harness (`KOLU_EVIDENCE=1 just test-quick`).

### Before — `master` branch: only 1 tile holds WebGL

`Scenario: WebGL context is held only by the focused tile` (canvas.feature:325)

Before (master) — tile 1 drops to DOM when you click away

Test output (2 scenarios, 18 steps — **2 passed**):
```
Scenario: WebGL context is held only by the focused tile
  Given I create a terminal
  Then there should be 2 canvas tiles
  And exactly 1 canvas tile should use the webgl renderer   ✓
  And the focused canvas tile should use the webgl renderer ✓
  When I click canvas tile 1
  Then exactly 1 canvas tile should use the webgl renderer  ✓  ← tile 2 lost WebGL, tile 1 gained it
  And the focused canvas tile should use the webgl renderer ✓
```

**Renderer state (master, after clicking tile 1):**
- tile 1 (now focused): `data-renderer="webgl"`
- tile 2 (now unfocused): `data-renderer="dom"` ← the bug: DOM renderer means ~7.7% wider cells

▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=evidence-before-webgl.mp4

---

### After — `webgl-n2-budget` branch: 2 tiles hold WebGL

`Scenario: WebGL is budgeted to the 2 most-recently-active tiles` (canvas.feature:325)

After (PR) — both ping-pong tiles stay on WebGL

Test output (3 scenarios, 30 steps — **3 passed**):
```
Scenario: WebGL is budgeted to the 2 most-recently-active tiles
  Given I create a terminal (×2) → 3 tiles total
  Then there should be 3 canvas tiles
  And exactly 2 canvas tiles should use the webgl renderer  ✓  ← N=2 budget
  And canvas tile 1 should use the dom renderer             ✓  ← LRU tail evicted
  When I click canvas tile 1
  Then exactly 2 canvas tiles should use the webgl renderer ✓
  And canvas tile 1 should use the webgl renderer           ✓  ← tile 1 back on WebGL
  When I click canvas tile 2
  Then canvas tile 1 should use the webgl renderer          ✓  ← tile 1 STAYS on WebGL (fix!)
  And canvas tile 2 should use the webgl renderer           ✓  ← tile 2 ALSO on WebGL
  And exactly 2 canvas tiles should use the webgl renderer  ✓
```

**Renderer state (PR, after clicking tile 2):**
- tile 1 (previously active): `data-renderer="webgl"` ← stays on WebGL (the fix)
- tile 2 (now focused): `data-renderer="webgl"` ← also WebGL

Cell width: both tiles use WebGL so `charW × dpr` is floored identically — the 7.7% reflow between the two renderers cannot occur for the dominant A↔B toggle.

▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=evidence-after-webgl.mp4

---

### Summary

| | master | webgl-n2-budget (PR) |
|---|---|---|
| Focused tile | `webgl` | `webgl` |
| Unfocused tile (after ping-pong) | `dom` ← bug | `webgl` ← fix |
| Tiles on WebGL (with 2 tiles) | 1 | 2 |
| Font reflow on A↔B switch | yes, ~7.7% | eliminated |

Mobile/touch: N/A — canvas is desktop-only; mobile renders one fullscreen tile with no renderer swap.

srid:

The **x86_64-linux** lane ran on **`kolu-ci-5`** (`idliv2-02`) — commit `6a6ec16e4`, _verdict unknown (no per-node timing)_

_(no per-recipe timing in `.ci/6a6ec16/timings.jsonl`)_

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/6a6ec16/timings.jsonl`; pool state is a live `flock` probe.</sub>

PR #1405, opened 2026-06-18
Keep minimap zoom controls reachable on a tall, narrow canvas

srid:
**When your terminals stack into a tall, narrow column, the minimap's zoom / arrange / activity-window controls get clipped off the right edge and become unclickable** — the bug in the screenshot below. The minimap is a shrink-to-fit overview: its scale fits the tile bounding box into `180 × 120`, so a tall, *narrow* box makes the **height** constraint win and the rendered width collapses. Both halves of the panel — the map and the control strip flush beneath it — were forced to that same collapsed width, and the strip's `overflow-hidden` then chopped off every control past the first couple.

The panel is one connected pill (rounded-top map + rounded-bottom bar sharing a border), so the two halves *must* stay the same width — the fix can't just widen the bar and leave the map behind. Instead it lets the **zoom bar's natural content width become the panel's floor**:

```
before                          after
┌────┐  map = b.w · scale       ┌──────────────┐  map: min-width = b.w · scale,
│ ▢  │  (can collapse to ~16px) │ ▢            │       stretches to panel width
│ ▢  │                          │ ▢            │
├────┤  bar = same width        ├──────────────┤  bar: no forced width →
│−100│  → controls clipped ✂    │ − 100% + ▦ 24h│       sizes to its controls
└────┘                          └──────────────┘  → nothing clipped
```

- The container is `position: absolute` with no width, so it **shrink-to-fits to its widest child** — now the bar, whose controls have fixed min-widths.
- `items-stretch` then widens **both** halves to that floor, keeping the seam flush.
- The map switches `width` → `min-width`, so its scaled content never clips but it can stretch wider. Tiles stay anchored to the left edge, so the extra width is inert right-padding and the *gesture math is unchanged* (clicks/drags read the box's left as the content origin).

No magic constant: the floor is whatever the bar's real controls measure, so adding or removing a control can never re-introduce the clip.

### Reproduction

A new e2e scenario (`canvas.feature`) stacks every tile into a tall column and asserts the zoom bar never overflows (`scrollWidth ≤ clientWidth`). It is **red on the old code** (the bar stays ~16px wide with ~130px of controls) and **green after the fix**; the five existing minimap scenarios still pass, confirming the wider map box didn't disturb viewport/tile dragging, arrange, or the window menu.

### Try it locally

```sh
nix run github:juspay/kolu/minimap-bug
```


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `a83c1cf0b410`

### Applied (4)
- `lowy-1` Map-box left edge is an undocumented load-bearing receptacle now shared with a new sizing strategy — commit `be32e8bb2`
- `hickey-1` Shared-panel-width invariant is fragmented across three CSS sites held together only by comments — commit `da5a8055a`
- `hickey-2` mapDims() now serves two structurally different roles (a hard cap and a soft floor) under one name — commit `55e702d09`
- `hickey-4` Stack-tiles step hardcodes the 6000px gap that encodes the tall-narrow geometry, with the load-bearing constant buried mid-loop — commit `bab9e297f`

### Agreed — no change (4)
- `lowy-2` New e2e step asserts the symptom (no clip) but not the invariant it protects (gesture origin unchanged) (packages/tests/step_definitions/canvas_steps.ts:826-833)
- `lowy-3` Width-floor policy correctly kept in CSS rather than duplicated as a constant — confirm, do not regress (packages/client/src/canvas/CanvasMinimap.tsx:248,260,451-453)
- `lowy-4` Gesture/render split (minimapGestures.ts) remains the correct volatility seam and is untouched (packages/client/src/canvas/minimapGestures.ts:1-3)
- `hickey-3` e2e clip assertion encodes the failure mechanism (scrollWidth>clientWidth) rather than the user-facing invariant (controls reachable) (packages/tests/step_definitions/canvas_steps.ts:818-832)

srid:
## Codex ⇄ Claude debate

### Round 1

**codex** — approved: `false`

The width fix is mostly coherent and the regression test targets the clipping symptom directly. I found one user-facing interaction regression introduced by stretching the map box beyond the scaled minimap content: clicks in the newly added empty right-hand area are still interpreted as real canvas coordinates.

Findings:
- `F1` · major · open — The stretched map element can now be much wider than the scaled minimap content, but the click/drag coordinate path still treats the whole `getBoundingClientRect()` width as minimap coordinate space. In the tall/narrow case, `mapDims().minW` can be ~16px while the zoom bar floors the rendered map near the controls' natural width; clicking the empty right side then maps to `bounds.minX + localX / minimapScale`, which can pan thousands of canvas units past the represented tile bounds. This is not a centering/offset objection: left anchoring preserves the origin, but the upper X bound is now wrong. (packages/client/src/canvas/CanvasMinimap.tsx:272)

**claude** — Codex's F1 (major) is correct and I fixed it. My PANEL SEAM change made the map box stretch wider than its scaled content (style went from `width: mapDims().w` to `min-width: mapDims().minW` plus container `items-stretch`). In the tall/narrow case `minW` can be ~16px while the box renders ~180px (floored by the zoom bar). The map's `onClick={handleMapClick}` still covered the full rendered width and mapped it into minimap coordinate space via `(clientX - rect.left)/minimapScale + minX` — so a click in the empty right padding panned roughly 11x the bounding-box width past the upper X bound. This is a genuine interaction regression my change introduced, distinct from the centering/left-anchoring decision the author deliberately made (codex was explicit that it is NOT a centering objection; left-anchoring preserves the origin but the upper X bound became wrong). Fix: guard `handleMapClick` to reject clicks whose local X exceeds `mapDims().minW` before calling `handleMinimapClick`, making the inert right padding actually inert for clicks. Of codex's two suggested approaches I chose the clamp (not an inner layer) — it is the lower-risk change, touches only the buggy entry point, and mirrors the existing `insideViewport` early-return pattern. The viewport-rect drag and hover paths need no guard: they are bounded by the viewport rect and the drag math is delta-based (`ev.clientX - startX`), so origin-independent. Verified: client `tsc --noEmit` clean and `just fmt` made no changes. Committed as 7a4144202.

- `F1` **fixed** — Accepted as a real regression introduced by the width fix. Added a dead-zone guard in handleMapClick at packages/client/src/canvas/CanvasMinimap.tsx:250-251: `const rect = (e.currentTarget as HTMLDivElement).getBoundingClientRect(); if (e.clientX - rect.left > mapDims().minW) return;` before calling handleMinimapClick. This rejects clicks that land in the stretched right-hand padding (local X past minW, the represented content width), which previously mapped far past bounds.maxX. Chose codex's clamp option over the inner-layer option: lower risk, single touch point, matches the existing insideViewport early-return style. The viewport-rect drag/hover paths are intentionally not guarded — they are bounded by the viewport rect and the drag is delta-based (origin-independent), so they are unaffected by the dead zone. Did not extend the e2e (the lens debate already settled the scrollWidth<=clientWidth assertion as the red-first test for this defect; F1 is a distinct follow-on coordinate bug).

commit: `7a4144202f4191f3206b7e892c504a23a165eaad`

### Round 2

**codex** — approved: `true`

I rechecked the current branch diff from a83c1cf0b4104f21c25334680276d9f5c01260c2 and the working tree status; there are no untracked files. CLAUDE's latest change directly addresses the coordinate regression I raised by making the widened right-hand padding inert for background clicks before `handleMinimapClick` converts local X into canvas space.

Findings:
- `F1` · major · resolved — Resolved. The stretched map could previously treat the extra right-hand padding as represented minimap content and pan far past the tile bounds on background click. (packages/client/src/canvas/CanvasMinimap.tsx:250)

_codex on the rebuttal:_ F1: accepted as fixed. The guard is on the buggy absolute-position conversion path, before `handleMinimapClick` computes `(clientX - rect.left) / minimapScale + bounds.minX`, so the widened padding no longer maps to nonexistent canvas content. I also accept the choice not to alter viewport drag in this round: that path is delta-based after hit testing and does not perform the same absolute dead-zone-to-canvas-coordinate conversion that caused F1.

srid:

Ran after `/simplify` (elegance pass skipped — `--no-elegance`), scoped to the PR's merge-base.

| Pass | Result |
|---|---|
| **Rules** (27 built-in + project) | ✅ Clean — no violations |
| **Fact-check** | 1 finding, **adjudicated as a false positive (not applied)** |
| **Elegance** | Skipped (`/simplify` already ran this tree) |

### Fact-check finding — *not applied*

> The `handleMapClick` dead-zone guard (`localX > minW → return`) was not added to `handleMapPointerDown` / `handleMapPointerMove`.

**Adjudication: skipped — applying it would be a regression, not a fix.**

- The viewport-rect overlay is `pointer-events-none`, and the drag path (`handleMapPointerDown` → `startViewportDrag`) is **delta-based** — it pans by `clientX − startX`, never by the absolute start position. A drag is therefore correct wherever it begins, so the dead-zone cannot corrupt it (this is exactly why codex's F1 fix guarded *only* the absolute-coordinate click path).
- When the canvas is panned right, the viewport rect **legitimately** extends into the widened region; the user must still be able to grab it there. A `localX > minW → return` on pointer-down would make that part of the rect **un-grabbable**.
- `cursor-grab` (hover) only triggers where the viewport rect actually renders, never over empty padding — so the "spurious cursor over empty space" premise does not hold.

codex (xhigh) explicitly endorsed not guarding the drag path in round 2; the `/simplify` altitude reviewer reached the same conclusion independently. No code change.

srid:

The **x86_64-linux** lane ran on **`kolu-ci-6`** (`dev-x86-64-linux-03`) — commit `930d22c55`, **passed** ✓

- **Lane wall** (pipeline): **2m44s**

| recipe | duration | |
|---|--:|:--|
| `ci::home-manager` | 2m14s | ✓ |
| `ci::e2e` | 2m2s | ✓ |
| `ci::smoke` | 52s | ✓ |
| `ci::nix` | 49s | ✓ |
| `ci::unit` | 34s | ✓ |
| `ci::pnpm-hash-fresh` | 34s | ✓ |
| `_ci-setup` | 30s | ✓ |
| `ci::biome` | 17s | ✓ |
| `ci::fmt` | 15s | ✓ |
| `ci::install` | 11s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/930d22c/timings.jsonl`; pool state is a live `flock` probe.</sub>

srid:
## Evidence

When the canvas holds a **tall, narrow tile arrangement** (two terminals stacked in a vertical column), the minimap's shrink-to-fit width collapses. **Before**, the whole panel — including the zoom / arrange / window control strip — collapsed with it, so `overflow-hidden` clipped every control but a sliver of the zoom-out button. **After**, the panel floors at the control strip's natural width: the overview still shrinks (its content is the narrow strip on the left, with the two stacked tiles), while every control stays visible and reachable.

| Before — panel collapsed, controls clipped | After — panel floored, full control strip |
|---|---|
| Before: minimap collapsed to a sliver, only a clipped minus visible | After: floored panel with two stacked tiles and the full − 100% + grid 24h strip |

Both captured live at 1280×720 with the **identical** two-terminal tall-narrow stack (canvas-space vertical gap 6000px), differing only by the fix — the *Before* frame is the same build with `CanvasMinimap.tsx` reverted to its pre-fix version. Desktop canvas only — mobile disables the canvas, so there is no mobile variant of this surface. The `ci::e2e` scenario asserts the same invariant programmatically (the zoom bar's `scrollWidth ≤ clientWidth` after stacking).

PR #1406, opened 2026-06-18
refactor(terminals): extract @kolu/terminal-providers, consume via directLink (P4w of #1398)

srid:
**P4w of the [from-scratch remote-terminals design](https://github.com/juspay/kolu/pull/1398)** — the *lean local refactor*. It moves kolu's per-terminal awareness providers out of kolu-server into a new `@kolu/terminal-providers` package and proves the consumption pattern the whole remote design rests on, **natively and in-process**, before any ssh/Nix lands.

Today the providers run as a bare in-process `startProviders(...)` call. This phase puts a real surface boundary where that call was: `buildWatcherServer` serves the host-side providers as a `@kolu/surface` contract, and kolu-server consumes it over **`directLink`** — the no-wire identity link. A later phase (P4d) swaps only the link (`directLink` → ssh `stdioLink`) and adds the packaging. *Local vs remote is only the link.*

### The split (faithful to the note)

Awareness is split by **what each provider must reach**:

| Provider | Needs | Runs |
| --- | --- | --- |
| foreground / process → `m.foreground` | nothing but the foreground + title taps | **in-server** (`startProcessProvider`) |
| git · PR · agent ×3 · agent-command tracker | the host's own `.git` / `~/.claude` / `~/.codex` / `gh` | **host-side** (`buildWatcherServer`) |

kolu-server owns the kaval taps (**minus PTY-forwarding** — the watcher never taps kaval) and relays them as `signal.*` calls. The watcher serves two per-terminal collections, split along the **same persisted-vs-live write fence** `metadata.ts` enforces — so the `terminals:dirty` autosave firehose stays off live churn:

- `persistedAwareness` → `{ git, lastAgentCommand, lastActivityAt }` (folded via `updateServerMetadata`)
- `liveAwareness` → `{ pr, agent }` (folded via `updateServerLiveMetadata`)

`cwd` / `foreground` / `location` stay in-server. The collection value schemas are `.pick`ed from kolu-common's canonical field schemas, so the mirror can't drift.

### Shape

`buildWatcherServer` mirrors kaval's `servePtyHost` / `createInProcessPtyHost` exactly — the blessed in-process-surface pattern — returning the transport-agnostic `implementSurface` router. **No `bin`, no `serveOverStdio`, no Nix** (that's P4d). **No client-facing surface change → no drishti PR.**

### Verification

- `just check` green (typecheck + biome across the workspace).
- **98 kolu-server unit tests pass unchanged** (adopt / metadata / reconcile) — behavior-preserving.
- New `directLink` round-trip test drives the served watcher end-to-end: watch → relay a `signal.commandRun` → observe the awareness delta over the link.

### Files

- **new** `packages/terminal-providers/` — `providers.ts` (moved) + `agentRecency.ts` (moved) + `watcherSurface.ts` + `server.ts` (`buildWatcherServer`) + tests + README.
- `packages/server/src/terminalEndpoint/local.ts` — rewired to consume the watcher over `directLink`, run the in-server foreground observer, and fold awareness back.
- Architecture docs synced (`README.md`, `packages/surface/README.md`).


srid:

✅ **Consensus** after 2 round(s) · lowy + hickey · base `a83c1cf0b410`

### Applied (8)
- `lowy-3` In-server process provider receives a 5-field `ProviderChannels` with 3 channels inert — commit `9998bc019`
- `lowy-4` `readScreenText` injected as id→handle lookup loses the per-terminal closure the old code had — commit `5bd3da011`
- `lowy-5` `buildWatcherServer` returns only the raw fragment router; the wire-wrap the precedent ships is absent — commit `1c936e4d5`
- `hickey-1` Awareness-split field list re-enumerated by hand at three sites; the fold-back has no completeness check — commit `19fff53c3`
- `hickey-2` `initialMeta` re-implements `createMetadata`'s seed as a parallel second copy — commit `6ac7edf38`
- `hickey-3` `startProviderLayer` builds a full 5-channel `ProviderChannels` to drive a 2-channel consumer — commit `9998bc019`
- `hickey-4` `directLink`-fragment-without-top-level-router diverges from the cited blessed pattern — commit `1c936e4d5`
- `hickey-5` `publish*`/`dropAwareness` late-bound mutable closures braid wiring order into correctness — commit `801292862`

### Agreed — no change (4)
- `lowy-1` Two awareness collections expose the metadata write-fence as a public contract shape (packages/terminal-providers/src/watcherSurface.ts:64-78)
- `lowy-2` Foreground/title taps fan out to two sinks in-line at the consumer — the wire is exposed (packages/server/src/terminalEndpoint/local.ts:656-700)
- `lowy-6` `initialMeta` in the watcher reconstructs a `TerminalServerMetadata` the endpoint already owns (packages/terminal-providers/src/server.ts:78-88)
- `hickey-6` Foreground `published`/`current` mutable double-buffer hand-rolls change-detection a Channel already owns (packages/terminal-providers/src/providers.ts:187-206)

srid:
## Codex ⇄ Claude debate

### Round 1

**codex** — approved: `false`

The provider split is mostly coherent, but the branch is not ready: the new workspace package is wired into pnpm/server imports without being added to the Nix source closure or refreshing the fixed-output dependency hash, which is a CI/build blocker. I also found a small documentation gap and a test robustness issue. I could not run typecheck locally because `pnpm` is not on PATH in this environment.

Findings:
- `F1` · blocking · open — `@kolu/terminal-providers` is a new workspace package with a `typecheck` script and `kolu-server` now imports it, but `default.nix`'s workspace `src` fileset still omits `./packages/terminal-providers`. The comment here explicitly says this fileset must include every typechecked workspace package; as-is the Nix type gate can skip the new package, and the Nix app build/source closure can end up with a broken `link:../terminal-providers` dependency. The changed `pnpm-lock.yaml` also means the `pnpmDeps.hash` at line 104 needs to be refreshed. (default.nix:55)
- `F2` · minor · open — The architecture package table was updated to mention `@kolu/terminal-providers` inside the server row, but it does not add a row for the new `packages/terminal-providers/` package itself. That leaves the central monorepo inventory incomplete for a newly introduced architectural package. (README.md:201)
- `F3` · minor · open — The watcher tests use hard-coded `/tmp` as the watched cwd. If `/tmp` is ever a git repo on a developer/CI host, the git provider can emit a persisted-awareness delta before the command-run delta, making the assertion at line 56 order/environment dependent. (packages/terminal-providers/src/server.test.ts:29)
- `F4` · nit · open — After moving the provider code out of `kolu-server`, the server package still declares direct dependencies that no server source imports anymore, notably `anyagent` and `kolu-github`. Keeping them blurs the new ownership boundary and makes package metadata less useful. (packages/server/package.json:29)

**claude** — Reviewed codex's verdict on the @kolu/terminal-providers extraction (P4w of #1398) and addressed all 4 findings. F1 (blocking): agreed the new package was missing from default.nix's `src` fileset and added it — verified end-to-end by building `.#checks.typecheck` (the type gate now covers and passes the package); but DISPUTED codex's hash-refresh claim — `nix build --rebuild .#pnpmDeps` reproduces the SAME output at the EXISTING declared hash, because the lockfile's only deltas are workspace `link:` entries that don't change fetchPnpmDeps's external closure, so no hash bump is needed. F2: added a dedicated README table row for packages/terminal-providers/. F3: switched the watcher tests from hard-coded /tmp to a per-test mkdtemp dir (8 tests still pass). F4: dropped the now-dead `anyagent` and `kolu-github` deps from kolu-server (confirmed zero remaining server imports; regenerated lockfile; server typecheck clean). Ran just fmt (no changes). Committed as 39cb686dc.

- `F1` **partial** — Agreed and fixed the real blocker: added ./packages/terminal-providers to the workspace `src` fileset in default.nix:77. The package has a typecheck script and kolu-server imports it (packages/server/src/terminalEndpoint/local.ts:38), so its omission let the Nix type gate silently skip it. Verified: `nix build .#checks.x86_64-linux.typecheck` now builds with the package included and type-checks clean. DISPUTED the second half (hash refresh at default.nix:104): I ran the authoritative `pnpm-hash-fresh` check `nix build --rebuild .#pnpmDeps` (forces re-execution, defeats binary-cache shortcut) and it reproduced the SAME store path at the EXISTING declared hash sha256-7damxI1mQ2xbZpDO5NvE8mmPGjyAuck3nybEBsaDUT4=. The lockfile's external-dependency closure is unchanged — all deltas are workspace `link:` entries (terminal-providers additions, and F4's anyagent/kolu-github removals from the server importer), which fetchPnpmDeps does not hash — so no hash bump is required.
- `F2` **fixed** — Agreed. The diff updated the packages/server/ row and the architecture prose to reference @kolu/terminal-providers but added no row for the package itself, leaving the monorepo inventory incomplete. Added a dedicated `packages/terminal-providers/` row at README.md:213 describing the awareness providers, the in-server/host-side split, buildWatcherServer, the watcherSurface collections, and that bin/serveOverStdio + Nix are P4d.
- `F3` **fixed** — Agreed — a real environment-dependence. The tests watched a hard-coded `/tmp`; the git provider calls subscribeGitInfo(record.meta.cwd) (providers.ts:242), so if /tmp were inside a git checkout the provider would async-resolve and push a persisted-awareness `git` delta that could land before the command-run delta, making the assertion at server.test.ts:55-56 order/environment dependent. Switched to a per-test `mkdtempSync(join(tmpdir(), 'watcher-server-'))` cwd created in beforeEach and removed in afterEach (matching the repo idiom in spawnInput.test.ts / iframePreviewRoute.test.ts) — a bare mkdtemp dir is never a git repo, so the git provider stays at its `git: null` seed. All 8 tests still pass.
- `F4` **fixed** — Agreed. At the base commit, `anyagent` and `kolu-github` were imported only by packages/server/src/terminalEndpoint/providers.ts (verified via `git grep` at a83c1cf0), which moved into @kolu/terminal-providers. An exhaustive grep of packages/server/src (incl. tests) now finds zero references to either, while anyforge, kolu-git, kolu-claude-code/codex/opencode (which codex said to keep) are all still imported. Removed exactly those two from packages/server/package.json, regenerated pnpm-lock.yaml via `pnpm install --lockfile-only` (6 lines removed from the server importer; both deps remain in the global lockfile via other importers), and confirmed `pnpm --filter kolu-server typecheck` passes.

commit: `39cb686dc41d7be636505a9b6ab74e3fbe7c8a7a`

### Round 2

**codex** — approved: `true`

All four prior findings are resolved in the current tree. The new package is now included in the Nix source fileset, the README inventory has a dedicated package row, the watcher tests no longer depend on `/tmp`, and the server package metadata no longer keeps the moved-only dependencies. I did not raise any new findings from this round's changes.

Findings:
- `F1` · blocking · resolved — The new `@kolu/terminal-providers` workspace package was missing from the Nix workspace `src` fileset, so Nix typecheck/build closure coverage could skip it or fail to resolve the workspace link. (default.nix:78)
- `F2` · minor · resolved — The architecture package table was missing a row for the new `packages/terminal-providers/` package. (README.md:213)
- `F3` · minor · resolved — The watcher tests used hard-coded `/tmp`, making persisted-awareness delta order depend on host filesystem/git state. (packages/terminal-providers/src/server.test.ts:26)
- `F4` · nit · resolved — `kolu-server` kept direct dependencies that moved with the provider code, blurring the package boundary. (packages/server/package.json:29)

_codex on the rebuttal:_ F1: I concede the hash-refresh half. The source-fileset omission was real and is fixed. For the fixed-output hash, the current lock diff only adds/removes workspace `link:` importer entries and leaves the external package closure unchanged, so keeping `sha256-7damxI1mQ2xbZpDO5NvE8mmPGjyAuck3nybEBsaDUT4=` is defensible, especially with the reported `nix build --rebuild .#pnpmDeps` verification.

srid:

Ran after the lens-debate + codex-debate passes settled, scoped to `a83c1cf0...HEAD` with `--no-elegance` (the simplify track already ran over this tree).

| Pass | Result |
| --- | --- |
| **Rules** | ✅ 0 violations (dry / focused-library / invalid-states / dead-code / silent-error-swallowing / unbounded-growth / comment-the-non-obvious all clean) |
| **Fact-check** | 6 findings raised — **0 applied** (all false positives or out-of-scope on review) |
| Elegance | skipped (`--no-elegance`; simplify already ran) |

**Fact-check dispositions** (each reviewed against the actual post-lens/codex code):

| # | Finding | Disposition |
| --- | --- | --- |
| 1 / 5 | "the `await watched` catch also swallows the `bridgeStream` setup" | **False positive** — the `try` wraps *only* `await watched`; the two `bridgeStream` folds are outside it (local.ts:710–732). The comment is accurate. |
| 3 | "the awareness `get` subscription may miss the synchronous seed snapshot" | **False positive** — a collection `get` yields `readOne(key)` (the current/seeded value) as its snapshot first, then deltas, and throws if the key is absent (surface/server.ts:220–229). `startWatching` seeds before `watch` resolves and the fold subscribes after `await watched`, so the snapshot always reads the seed. The directLink round-trip test asserts this and passes. |
| 4 | "lifecycle map leaks if `stopProcess()` throws" | **Skip** — `teardownProviders` `delete`s the entry *before* calling `stopProviders()` (local.ts:778), so no stale entry; and the cleanup (channel unsubscribes) can't throw. Matches the pre-existing teardown pattern. |
| 2 | "fire-and-forget signal relays have no timeout/backpressure" | **Out of scope** — the relay is in-process over `directLink` (a microtask), so promises can't accumulate unboundedly. Backpressure is a P4d / `stdioLink` concern, not P4w. |
| 6 | "no log for a signal arriving after teardown" | **Skip** — the `?.` no-op is the correct handling of a benign teardown race (kolu-server aborts the tap bridges on teardown); logging every expected stray signal is noise. |

Net: the rule gate is clean and the two findings flagged "most critical" are provable misreads of the post-debate code. No fixes applied.

srid:

The **x86_64-linux** lane ran on **`kolu-ci-7`** (`dev-x86-64-linux-05`) — commit `6756985d0`, **passed** ✓

- **Lane wall** (pipeline): **2m51s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 2m11s | ✓ |
| `ci::home-manager` | 1m38s | ✓ |
| `ci::nix` | 53s | ✓ |
| `ci::smoke` | 46s | ✓ |
| `ci::unit` | 30s | ✓ |
| `_ci-setup` | 30s | ✓ |
| `ci::biome` | 10s | ✓ |
| `ci::fmt` | 9s | ✓ |
| `ci::install` | 8s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/6756985/timings.jsonl`; pool state is a live `flock` probe.</sub>

srid:
## Evidence



**Scenario:** `Tile chrome shows OpenCode waiting state` (`features/opencode.feature`) — recorded on a clean `terminal-providers` branch build under `KOLU_EVIDENCE=1`, 1280×720 viewport.

The frame proves the awareness pipeline is alive end-to-end after the directLink rewire: a mocked OpenCode session in `waiting` state surfaces the `awaiting ⏵` indicator in the tile chrome (top centre of the terminal tile) **and** populates the Dock's agent session row (left panel, `OPENCODE-2769-Xb…` in green) — both routed through the new `@kolu/terminal-providers` package via `buildWatcherServer` / `directLink`, behaviour-preserving.

All 4 `opencode.feature` scenarios and all 6 `dock.feature` scenarios passed on this build (`docker-surfaces-awaiting-claude-session-as-a-full-card` ✓, `tile-chrome-shows-opencode-waiting-state` ✓).

No layout/sizing change; desktop capture represents all viewports (pure internal provider-extraction refactor — zero UI surface delta).

srid:
## Review — P4w extraction (fix-then-merge)

A well-executed, **scope-disciplined** extraction — but it's **not behavior-preserving**: one root-cause defect (blocker + major) clobbers restored terminal awareness on adoption. The fix is small and at the source. Recommend **fixing the seed-clobber + adding a regression test before merge**; the lower-severity items can ride along.

_(Multi-lens review — spec / behavior-preservation / correctness / simplicity — each finding re-verified against branch code; 5 confirmed of 20 raw, false positives dropped.)_

### What's good
- **Scope discipline is clean — no P4d creep.** No `serveOverStdio` *call*, no `bin`, no Nix derivation/drv-map; `default.nix` just registers the package; the `serveOverStdio` mentions are doc-comments deferring to P4d. `"DAG"` fully dropped. Deps are right (no ssh/nix).
- **Faithful extraction** — the persisted-vs-live write-fence split matches the original provider write sites, and the **hard ordering couplings survive**: the agent-command-tracker → detector coupling (shared mutable `record.currentAgent`) and the `git → pr` channel edge stay co-located on the watcher side. No subtle reorder.
- `directLink` is the right seam (no serialization; `watch`-before-signal ordering holds). **CI green both arches, mergeable CLEAN.**

### 🔴 Blocker — restored awareness clobbered on adoption
`buildWatcherServer` seeds a **defaults snapshot** `{git:null, lastActivityAt:0, lastAgentCommand:undefined}` as the *first* `persistedAwareness` frame **before providers run** (`terminal-providers/src/server.ts` ~`startWatching`/`245`), and the consumer folds it via `Object.assign(m, a)` (`local.ts:725`). The watcher only ever receives `{id,pid,cwd}` via `watch` — it **never sees an adopted terminal's restored metadata** — so the fold overwrites a redeploy-survivor's restored `lastActivityAt → 0` and `lastAgentCommand → undefined`. Master mutated `record.meta` in place with no defaults seed, so this is **new**.

It defeats `agentRecency.ts`'s "don't overwrite restored non-zero `lastActivityAt`" guard (the seed lands *before* it) and regresses the **B3.3 redeploy-survival** guarantee — `dockRowRanking` treats `lastActivityAt===0` as never-touched, so a restored idle session **loses its recency-sort slot and its agent-command label**. `adopt.test.ts` covers only the pure `adoptedMeta` mapping, not the runtime `watch → seed → fold` path, so it's uncaught.

### 🟠 Major — the clobber gets durably persisted
Same root cause, worse blast radius. The fold runs through `updateServerMetadata → publishSnapshotAndDirty → terminals:dirty` (`metadata.ts`). But `adoptTerminal` **deliberately omits** `emitTerminalsDirty` (`local.ts:492-495` — *"arming an autosave here could persist a half-adopted set"*). The async fold fires `dirty` anyway → the leading-edge autosave **writes the clobbered defaults back over the restored session file**. Permanent recency/label loss for an idle cold-boot survivor.

### ✅ Fix (small, at the source — resolves both)
Stop emitting a defaults snapshot that overwrites restored fields. Either:
- **drop** the `publishPersisted(id, persistedOf(meta))` seed in `startWatching` so persisted awareness only publishes on a genuine provider detection (matches master); **or**
- **seed the watcher from the restored persisted awareness** (thread `git`/`lastActivityAt`/`lastAgentCommand` into the `watch` input) so the eager seed reproduces restored values and the fold becomes a no-op (providers already publish-if-changed).

Either way the major dissolves too. **Add a runtime test** extending `adopt.test.ts`: adopt a terminal with non-zero `lastActivityAt` + saved `lastAgentCommand`, run `watch → seed → fold`, assert both survive.

### Lower severity (leave or trivial)
- **🟡 Minor** — non-atomic provider startup: if `startWatcherProviders` throws *synchronously*, the seeded store entry is stranded forever (neither `stopWatching` nor `dispose` reaches it). Edge case; seed-after-success or `try/catch → dropAwareness`.
- **Nit** — the exported `router` field is **unused** (only `.client` is consumed) *and* the wrong shape for its stated P4d purpose (a bare `implementSurface` fragment 404s under `StandardRPCHandler` — kaval returns `router` **and** `servedRouter`). Suggest returning `{ client, dispose }` now, re-adding `router`/`servedRouter` when P4d lands.
- **Nit** — a one-microtask awareness lost-update window (theoretical, self-healing) — worth a code comment, not a fix.

---
_Automated review via Claude Code (model `claude-opus-4-8`)._

srid:
## Fixed — all findings actioned (`3c93a0934`)

Thanks — the seed-clobber was a real, well-traced regression I (and the gauntlet) missed by asserting "behavior-preserving" without covering the adoption / B3.3 redeploy-survival path. Root cause fixed at the source.

### 🔴 Blocker + 🟠 Major — restored awareness clobbered (and durably persisted)
Took the **"seed the watcher from the restored persisted awareness"** option. `watch` now carries a `seed: PersistedAwarenessSchema` (`{git, lastAgentCommand, lastActivityAt}`) — the endpoint's *current* value for the terminal — and `initialMeta` seeds `record.meta` from it instead of hardcoded defaults. This resolves **both**:
- The eager snapshot now reproduces the restored values, so the endpoint's fold is a no-op — no clobber.
- `agentRecency`'s restore-guard (which reads `record.meta.lastActivityAt`) now actually fires host-side on re-detection (it could never trigger before, since the watcher's record was pinned to `lastActivityAt: 0` — a second facet of the same bug).

For the Major specifically, I also stopped the fold from re-arming the autosave during adoption: the consumer now **reconciles the first persisted frame (the snapshot = the seed) silently** — a bare `Object.assign(entry.meta, a)`, *not* `updateServerMetadata` — so no `terminals:dirty` fires mid-boot (matching `adoptTerminal`'s deliberate omission). Every *later* persisted frame is a genuine provider change → `updateServerMetadata` + dirty, exactly as master did.

### ✅ Regression tests
- `server.test.ts`: asserts the watcher reproduces a restored seed (`lastActivityAt: 1.7e12`, `lastAgentCommand: "claude --resume"`) rather than resetting to defaults.
- `adopt.test.ts`: drives the runtime path you asked for — `adoptedMeta(sentinel) → watch → seed → snapshot` — and asserts the survivor's `lastActivityAt` + `lastAgentCommand` survive. **Red on the old defaults-seed code, green now.**

### 🟡 Minor — non-atomic provider startup
`startWatching` now `try/catch`es `startWatcherProviders` and `dropAwareness(id)` on a synchronous throw, so a half-started terminal's seeded store entries aren't stranded.

### Nits
- **Unused/misshapen `router`**: confirmed — both consumers use `.client`. `buildWatcherServer` now returns `{ client, dispose }`; the bare fragment isn't wire-routable anyway, so `servedRouter` lands with P4d (as you suggested). `router` is still used internally to build the `directLink` client, just not exposed.
- **One-microtask lost-update window**: added the code comment (the silent reconcile still lands a raced value on `entry.meta`; self-heals on the next provider publish).

CI re-running on the new HEAD now.

srid:

The **x86_64-linux** lane ran on **`kolu-ci-7`** (`dev-x86-64-linux-05`) — commit `3c93a0934`, **passed** ✓

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 2m5s | ✓ |
| `ci::smoke` | 43s | ✓ |
| `ci::nix` | 42s | ✓ |
| `ci::unit` | 38s | ✓ |
| `ci::flake-check` | 36s | ✓ |
| `ci::atlas-sync` | 34s | ✓ |
| `_ci-setup` | 25s | ✓ |
| `ci::surface-example-build` | 24s | ✓ |
| `ci::surface-app-example-build` | 24s | ✓ |
| `ci::fmt` | 23s | ✓ |
| `ci::biome` | 23s | ✓ |
| `ci::install` | 9s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/3c93a09/timings.jsonl`; pool state is a live `flock` probe.</sub>

srid:
## Re-review — fix is correct; close the two guard holes before merge

Verified against `3c93a093` (seed-from-restored), each finding re-grounded.

**The fix is right.** Both previously-raised issues are genuinely closed:
- **🔴 Blocker → closed.** `watch` requires `seed: PersistedAwarenessSchema`; `initialMeta(cwd, seed)` spreads `...seed` so an adopted survivor's restored `git`/`lastActivityAt`/`lastAgentCommand` are reproduced — the `local.ts` fold no-ops for restored fields, and `agentRecency`'s restore-guard fires.
- **🟠 Major → closed.** The first persisted frame uses a bare `Object.assign` that **returns before `updateServerMetadata`**, so no `terminals:dirty`/autosave fires mid-adoption.
- **Also fixed:** the non-atomic-startup minor (`try/catch → dropAwareness`) and the unused `router` nit (returns `{ client, dispose }`). Lens/codex passes are behavior-neutral, scope discipline held (no `serveOverStdio`/bin/Nix), CI green both arches, mergeable CLEAN.

**But the fix isn't guarded — both must land before merge.** As written, the originally-blocking clobber can come back **silently** (no TS error, no test failure):

### 1. The B3.3 regression test doesn't drive the real path
`adopt.test.ts` **hand-copies** `local.ts:636`'s `seed: {...}` literal into a standalone `buildWatcherServer` — it never exercises the endpoint that *builds* the seed. And `lastAgentCommand` is `.optional()` / `lastActivityAt` `.default(0)` (`common/surface.ts:254,258`), so **dropping either field from `local.ts`'s seed is neither a compile error nor a test failure** → the clobber returns, suite green. The test guards the watcher honoring a seed, not the endpoint *supplying* one — the exact link that broke.

**Do:** extract one exported `persistedSeedOf(meta)` projection, call it in **both** `local.ts:636` and the test, so a dropped field breaks compile/test (or drive the real endpoint over `directLink`).

### 2. The dirty-omission branch has zero test
The `firstPersisted` no-dirty branch — the major's durability vector — is untested (no `local.test.ts`; the terminalEndpoint tests grep clean for `dirty`/`updateServerMetadata`). The obvious "simplification" of folding the first frame through `updateServerMetadata` re-arms the autosave-persists-the-clobber path with **no failing test**.

**Do:** add an endpoint-driven test asserting the **first** persisted frame does **not** trigger `terminals:dirty`/autosave (spy the dirty/save sink), while a **later** genuine provider delta **does**.

_(The one-microtask first-frame window is documented + self-healing — leave it.)_

These are small and surgical. Land both, then this is a clean merge.

---
_Automated re-review via Claude Code (model `claude-opus-4-8`)._

srid:
## Both guard holes closed (`4a95b1984`)

Spot-on — the fix was correct but un-guarded; the clobber could have crept back with a green suite. Both landed, plus the master merge.

### 1. The seed link is now shared, not hand-copied
Extracted **one exported projection** `persistedSeedOf(meta)` in `@kolu/terminal-providers` — the *same* projection the watcher publishes — and call it in **both** `local.ts`'s `watch` seed and `adopt.test.ts`, replacing the two hand-copied field literals. The field list is schema-derived (`PERSISTED_KEYS` off `PersistedAwarenessSchema.keyof()`), so a dropped field is now a *schema* change, not a silent per-call-site omission. The adoption regression test exercises the real `adoptedMeta → persistedSeedOf → watch → snapshot` link, not a copy of it.

### 2. The dirty-omission branch is now tested
Extracted `makePersistedAwarenessFold(entry, id)` (the first-frame-silent / later-frame-dirty rule) and added **`persistedAwarenessFold.test.ts`**, mirroring `metadata.test.ts`'s `terminals:dirty` spy:
- first frame (the seed snapshot) → asserts **`dirtyCount === 0`** (and that it still reconciled onto `entry.meta`);
- a later genuine provider frame → asserts **`dirtyCount === 1`**.

So the "obvious" simplification of routing the first frame through `updateServerMetadata` is now a **failing test**, not a silent re-arm of the autosave-persists-the-clobber path.

`+2` server tests (101 total), `just check` green. One-microtask first-frame window left documented + self-healing, per your note. CI re-running on the merged HEAD now.

srid:
Closing in favor of a cleaner P4w shape.

The extraction and the fixes here were solid — but the **in-process `directLink` consumption is the wrong boundary**. In one process there's no wire, so the watcher-as-a-separate-server-dialed-via-`directLink` is a *fictional* boundary, and everything it dragged in — the mirror/fold, the `seed`/`firstPersisted` reconcile, the one-microtask lost-update window, the guard holes — was an artifact of it. None of it existed on master (the providers mutated `record.meta` directly).

The right move is **TypeScript-level composition**, which `@kolu/surface` already supports and `surface-app` already uses against kolu-server (`surfaceAppSurfaceWith` / `surfaceAppServer`):

- `terminal-providers` stops being a self-served surface and exports two spreadable fragments — a **contract fragment** (`{ collections, procedures }`, spread into `koluSurface`'s `defineSurface`) and a **deps fragment** (`watcherDeps`, spread into the existing `implementSurfaces` `kolu` sibling).
- kolu-server composes it: **one router, one transport, no client, no mirror, no fold** locally. The providers publish their own awareness collections through the shared ctx; a thin in-server fold calls the *existing* `updateServerMetadata`/`updateServerLiveMetadata` (the persisted-vs-live `terminals:dirty` fence survives).
- **The surface stays the architectural boundary** — a typed contract — it just isn't dialed in-process.
- **P4d (remote)** reuses the *same* contract fragment, served over `stdioLink` + mirrored — the mirror reappears only where there's a genuine wire.

So the directLink/mirror/lost-update/seed-fold machinery all evaporates locally. The valuable part of this PR (extracting the provider pipeline into a package) gets re-landed in the composed shape. One caveat carried forward: collections have no `connect` hook, so the providers' publish binding moves to a post-build seam (the existing `surfaceCtx.ts` holder / a `bindCtx`), kept kolu-local so it needs no drishti PR; and the watch-time `seed` (B3.3 adoption) stays — it was never directLink machinery.

Plan note updated accordingly.

PR #1407, opened 2026-06-18
Fix recent-agent commands dropping quotes on re-run

srid:
**Re-running a previously-used agent from the *New terminal → recent agents* list — or letting a restored session auto-resume one — used to strip the quotes off its arguments.** A command like `claude --dangerously-skip-permissions --settings '{"ultracode": true}'` came back as `--settings {"ultracode": true}`; the shell then word-split the JSON, and Claude failed with `Error: Settings file not found: {ultracode:`.

### Why it broke

kolu remembers an agent command by capturing the line you typed and normalizing it: `parseAgentCommand` tokenizes the raw string with `string-argv` (which **consumes the shell quoting**), drops prompt/positional noise, then rejoins the survivors. The rejoin was a bare `.join(" ")`, so a token that had carried quotes for a reason — a JSON blob, a value with spaces — lost them. That single malformed string is then stored, displayed in the palette, and written back into a PTY to run.

```
raw line:   claude … --settings '{"ultracode": true}'
              │  string-argv (quotes consumed)
              ▼
argv:       … "--settings", '{"ultracode": true}'      ← one token, with a space
              │  kept.join(" ")        ← quoting NOT restored
              ▼
stored:     claude … --settings {"ultracode": true}    ← re-run word-splits this
```

### The fix

Re-quote each kept token when joining, so the normalized command **re-tokenizes to exactly the argv it was built from**. Safe bare words (`claude --model sonnet`) stay clean; anything carrying whitespace, quotes, braces, `$`, … gets single-quoted. Both join sites are fixed — `parseAgentCommand` and `resumeAgentCommand` — so the repair covers the stored form, the palette label, and all three spawn paths (worktree-create, recent-agents prefill, session-restore resume) at once.

| Input value | Before (stored / re-run) | After |
| --- | --- | --- |
| `--settings '{"ultracode": true}'` | `--settings {"ultracode": true}` → splits | `--settings '{"ultracode": true}'` |
| `--append-system-prompt "be terse"` | `--append-system-prompt be terse` → splits | `--append-system-prompt 'be terse'` |
| `--config model_reasoning_effort="xhigh"` | unquoted (quotes lost on re-run) | `--config 'model_reasoning_effort="xhigh"'` |

### Shape (after the review gauntlet)

[...]

PR #1408, opened 2026-06-19
Prove shellJoin round-trips through a real POSIX shell

srid:
The `@kolu/shell-quote` round-trip tests added in #1407 only asserted
`shellSplit(shellJoin(argv)) === argv` — the leaf validated against its **own**
tokenizer. That is a self-consistency check, not a proof: it says nothing about
whether a *real* shell parses `shellJoin`'s output the way the no-word-split /
no-injection claim promises. **This adds a test that runs `shellJoin`'s output
through an actual POSIX shell and asserts the shell word-splits it back into
exactly the input argv** — the property that actually backs the safety claim.

### How the proof works

```
shellJoin(["printf", "%s\0", ...args])
        │   e.g.  printf '%s\0' --settings '{"ultracode": true}' ...
        ▼
  <discovered shell> -c "<that line>"     ← a real grammar parses & word-splits
        ▼
  printf echoes each argv element NUL-separated   ← the argv the shell BUILT
        ▼
  deepEqual( parsed, args )               ← must match byte-for-byte
```

`printf` is the stub by design: it's a **safe bare word** (so `shellJoin` leaves
it unquoted) and a **shell builtin** (so there's no executable temp file to
write — which a `noexec` `$TMPDIR` in a hardened sandbox would reject). Its
recycled `%s\0` format echoes the remaining argv NUL-separated, the one
separator that can't collide with a byte inside a token. *If quoting let a value
word-split or a metacharacter fire, `printf` would see a different field list and
the deep-equality would fail* — verified with a negative control where a naive
unquoted join made `;`/`&`/`|` split the line and backticks execute (exit 127).

### What's covered

- **One shared truth-table corpus** feeds both the existing self-inverse test
  and the new real-shell test — no parallel corpus. Added a metacharacter-soup
  case carrying `$ \` ; & | * ( ) { }` so the quoting is proven to neutralize
  every one of them, plus the spaced JSON blob, a spaced sentence, the
  apostrophe `'\''` idiom, an embedded-double-quote value, and the empty token.
- **The tilde carve-out is pinned, not papered over.** The strict
  deep-equality corpus *excludes* leading-`~` tokens (a real shell re-expands
  them, so they can't byte-match), and a **separate** assertion proves
  `--settings ~/x` *does* expand to `$HOME/x` under a real shell — the
  documented, intended behavior (#1407), asserted rather than treated as a bug.

> **Portable & green everywhere.** The shell is discovered from `PATH` (no
> hardcoded `/bin/sh`), so it runs against whatever the Nix devshell / CI
> provides; when no POSIX shell is found the suite **skips cleanly** instead of
> failing.

The README gains a one-paragraph scope note: the replay guarantee assumes a
**POSIX-compatible target shell** (`sh`/`bash`/`zsh`/`dash`); `fish`/`csh` quote
differently, so exact replay isn't guaranteed there.

*Test + docs only — no user-facing behavior change, so no changelog entry.*
`@types/node` is added as a **devDependency** (matching `integrations/pty`) so
the node-builtin imports in the test typecheck; the leaf's runtime
zero-dependency property is unchanged.

### DIFF
--- a/packages/shell-quote/README.md
+++ b/packages/shell-quote/README.md
@@ -61,6 +61,14 @@ shellSplit(`claude --settings '{"ultracode": true}'`);
 shellSplit(`'don'\''t'`); // → ["don't"]   (a general tokenizer shatters this)
 ```
 
+**Target shell.** The replay guarantee — `shellJoin`'s output re-parsing back to
+the same argv — likewise assumes a **POSIX-compatible** shell on the receiving
+end: `sh`, `bash`, `zsh`, and `dash` share the single-quote and tilde semantics
+it relies on. Shells with different quoting rules (`fish`, `csh`) are out of
+scope, so exact replay isn't guaranteed there. A real-shell round-trip test
+([`index.test.ts`](./src/index.test.ts)) pins this against whatever POSIX shell
+the dev/CI environment provides, and skips cleanly where none is available.
+
 ## Tilde handling
 
 A leading `~` is treated as a safe bare word and left **unquoted**, so a shell
--- a/packages/shell-quote/README.md
+++ b/packages/shell-quote/README.md
@@ -45,6 +45,17 @@ shellJoin(["claude", "--settings", `{"ultracode": true}`]);
 // → claude --settings '{"ultracode": true}'
 ```
 
+**Command-head precondition.** The replay equivalence is for the _argument
+tail_, the values after an ordinary command word. It assumes `argv[0]` is a
+plain command name, which is what every consumer here passes (an agent
+basename: `claude`, `codex`, …). A shell resolves the **command-position** word
+by grammar _before_ quote-removal, so a token that survives in argument
+position can still change meaning as `argv[0]`: `shellQuoteArg` leaves
+`FOO=bar` and `if` bare (both are safe bare words), but the shell reads a
+leading `FOO=bar` as a variable assignment and a leading `if` as a reserved
+word, not as a command to run. Quote the head yourself, or keep it a plain
+command name, if you cannot guarantee it is one.
+
 ### `shellSplit(line: string): string[]`
 
 The **exact inverse of `shellJoin`** — `shellSplit(shellJoin(argv))` deep-equals
--- a/packages/shell-quote/package.json
+++ b/packages/shell-quote/package.json
@@ -15,6 +15,7 @@
     "test:unit": "vitest run"
   },
   "devDependencies": {
+    "@types/node": "^22.0.0",
     "typescript": "^5.8.0",
     "vitest": "^4.1.0"
   }
--- a/packages/shell-quote/src/index.test.ts
+++ b/packages/shell-quote/src/index.test.ts
@@ -129,6 +129,15 @@ const POSIX_SHELL = findPosixShell();
 // word-splits back into exactly the argv it was built from. The shellSplit
 // tests above only check the leaf against its own tokenizer; this checks it
 // against the real grammar that backs the no-word-split / no-injection claim.
+//
+// Scope: this pins the ARGUMENT-TAIL guarantee — the values after an ordinary
+// command word. `replay` deliberately runs a known-safe `printf` head so every
+// corpus token lands in ARGUMENT position, which is exactly how the real
+// consumer joins (argv[0] is always an agent basename like `claude`). The
+// command-POSITION carve-out (a leading `FOO=bar` read as an assignment, a
+// leading `if` as a reserved word — both left bare by shellQuoteArg, since the
+// shell resolves the head by grammar before quote-removal) is out of scope by
+// the same construction; see the README "Command-head precondition".
 describe.skipIf(POSIX_SHELL === null)(
   "shellJoin output parsed by a real POSIX shell",
   () => {
@@ -145,7 +154,13 @@ describe.skipIf(POSIX_SHELL === null)(
      *  remaining argv element NUL-separated — the one separator that cannot
      *  collide with any byte inside a token, so the split is unambiguous. If the
      *  quoting let a value word-split or a metacharacter fire, printf would see a
-     *  different field list and the deep-equality below would fail. */
+     *  different field list and the deep-equality below would fail.
+     *
+     *  The `printf` head is also the known-safe COMMAND word that scopes this to
+     *  the argument tail: it forces every `args` token into argument position,
+     *  where shellQuoteArg's quoting fully governs the result — matching how the
+     *  real consumer joins (argv[0] is always an agent basename). See the
+     *  suite-level comment for the command-position carve-out. */
     function replay(
       args: readonly string[],
       env: NodeJS.ProcessEnv = {},
--- a/packages/shell-quote/src/index.test.ts
+++ b/packages/shell-quote/src/index.test.ts
@@ -1,5 +1,33 @@
+import { execFileSync } from "node:child_process";
+import { accessSync, constants } from "node:fs";
+import { delimiter, join } from "node:path";
 import { describe, expect, it } from "vitest";
 import { shellJoin, shellQuoteArg, shellSplit } from "./index.ts";
+
+/** The shared round-trip truth table: argv shapes a consumer joins and must get
+ *  back unchanged. It is exercised two ways from one corpus — against shellSplit
+ *  (the leaf's OWN inverse) and against a real POSIX shell (below) — so the same
+ *  cases prove both the self-consistency and the actual no-word-split /
+ *  no-injection claim that backs this package.
+ *
+ *  The entries carry every shape the bug (#1407) was about: the spaced JSON blob
+ *  (`--settings '{…}'`), a spaced sentence, the apostrophe `'\''` idiom, an
+ *  embedded-double-quote value, the empty token, and a value packed with shell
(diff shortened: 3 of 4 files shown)


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `9e9dc2e66df3`

Independent findings: lowy=6, hickey=2

### Applied (2)
- `lowy-4` findPosixShell scans for {sh,bash,dash} but the README contract names {sh,bash,zsh,dash} — commit `5c6888123`
- `hickey-1` Supported-shell set fragmented across README prose and the resolver array (already drifted) — (uncommitted)

### Agreed — no change (6)
- `lowy-1` Electricity test: the receptacle is pre-existing; this diff adds no shipping boundary (packages/shell-quote/src/index.test.ts:1-178)
- `lowy-2` Strict-corpus filter exposes the wires: re-derives the tilde rule the receptacle already owns (packages/shell-quote/src/index.test.ts:159-161)
- `lowy-3` replay's env-merge default is an unused wire on the strict path (packages/shell-quote/src/index.test.ts:143-151)
- `lowy-5` findPosixShell is a generic PATH-resolution concern living inside a leaf's test file (packages/shell-quote/src/index.test.ts:100-118)
- `lowy-6` ROUND_TRIP_CORPUS reuse across self-inverse and real-shell tests is the correct shared fixture (packages/shell-quote/src/index.test.ts:20-30)
- `hickey-2` Corpus shapes enumerated in prose alongside the array they describe (packages/shell-quote/src/index.test.ts:13-19)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `9e9dc2e66df3`

### Round 1

**codex** — approved: `false`

The added real-shell coverage is useful and the changed files are otherwise tidy. I could not run the package tests because `pnpm` is not on PATH and `nix develop` cannot access the Nix daemon in this sandbox, so this review is static plus shell-behavior inspection.

Findings:
- `F1` · minor · open — The new real-shell test claims to prove that `shellJoin` output parses back to the same argv, but `replay()` always prefixes the corpus with a known-safe `printf` command and format string. That only verifies argument-position words. POSIX shell parsing has command-position cases that this masks, such as `FOO=bar` being parsed as an assignment word or `if` as a reserved word when it is argv[0], while the same token after `printf '%s\0'` is just data. Because `SAFE_BARE_WORD` includes `=`, this leaves a real gap between the public "same argv" guarantee and what the test proves. (packages/shell-quote/src/index.test.ts:153)

**claude** — Codex raised one minor finding (F1): the new real-shell replay test prefixes a known-safe `printf '%s\0'` head, so every corpus token lands in ARGUMENT position — proving the argument-tail guarantee, while the README/JSDoc prose claimed broad arbitrary "same argv" replay. I verified the finding is technically correct: `shellQuoteArg` leaves `FOO=bar` and `if` bare (both match SAFE_BARE_WORD), and a POSIX shell resolves the command-position word (argv[0]) by grammar BEFORE quote-removal, so a leading assignment-word or reserved-word would not replay (confirmed: `sh -c 'FOO=bar x'` exits 127, `sh -c 'if x'` is a syntax error, whereas the same tokens after `printf '%s\0'` replay exactly). Disposition: partial. I agreed the prose overclaimed and scoped it — but did NOT inflate the implementation, because the sole real consumer (parseAgentCommand in packages/integrations/anyagent/src/agent-cli.ts) ALWAYS sets argv[0] to a known agent basename (claude/codex/…, a safe bare word), so command-position assignment/reserved words cannot arise by construction — and replay()'s printf head IS exactly that "known-safe command head". Adding command-position quoting to shellQuoteArg would add surface area for a case the design already precludes. Fixed by documenting the precondition: a "Command-head precondition" paragraph in the README under shellJoin, plus scope notes in the test's suite-level comment and the replay() docblock. Test logic unchanged. Verified: 11 tests pass with the real-shell suite actually running (not skipped), tsc --noEmit clean. biome/pnpm are only inside the Nix devshell (unreachable in sandbox), but changes are prose-only (JSDoc + Markdown) which biome does not reflow, and the >80-char comment lines match the file's pre-existing style.

- `F1` **partial** — Agreed the documented guarantee overclaimed: the test (replay() at packages/shell-quote/src/index.test.ts:158) prefixes a known-safe `printf '%s\0'` head so every corpus token is in ARGUMENT position, and the README shellJoin section + JSDoc described it as arbitrary 'same argv' replay. Verified the gap is real: `shellQuoteArg` leaves `FOO=bar`/`if` bare (SAFE_BARE_WORD includes `=`), and a shell resolves argv[0] by grammar before quote-removal (sh -c 'FOO=bar x' -> exit 127; sh -c 'if x' -> syntax error), so a leading assignment/reserved word would not round-trip. FIXED the docs to scope the guarantee to the argument tail: added a 'Command-head precondition' section to packages/shell-quote/README.md (under shellJoin, lines 48-58) stating argv[0] must be a plain command name and naming the FOO=bar/if carve-out; added a 'Scope' note to the suite-level comment (index.test.ts:129-139) and a paragraph to the replay() docblock (index.test.ts:155-160) explaining the printf head is the known-safe command head that forces tokens into argument position. DISPUTED codex's alternative of adding command-position implementation: the only consumer (parseAgentCommand, packages/integrations/anyagent/src/agent-cli.ts:201,231) always sets kept[0] to a verified agent basename — a safe bare command word — so command-position assignment/reserved words cannot arise by construction; documenting the precondition (codex's option a) is the correct, honest, minimal fix and avoids inflating shellQuoteArg for an impossible case. Tests still pass (11/11, real-shell suite runs), typecheck clean.

commit: `3f12f86363972b5a96d06c08e6df7c9611e4106e`

### Round 2

**codex** — approved: `true`

F1 is resolved. The current README now states the command-head precondition explicitly, and the real-shell test comments/docblock explain that `printf` is intentionally the known-safe command head used to exercise argument-tail replay. That addresses the mismatch I raised without adding command-position quoting behavior that the current consumers do not need.

Findings:
- `F1` · minor · resolved — The real-shell replay test originally proved only argument-position replay while the surrounding documentation/comments implied arbitrary argv replay, leaving command-position cases such as `FOO=bar` or `if` outside the proof. (packages/shell-quote/src/index.test.ts:153)

_codex on the rebuttal:_ F1: I accept the dispute against adding command-position implementation. The real consumer constructs a known-safe agent basename as argv[0], and the README now documents that precondition directly with the relevant `FOO=bar`/`if` carve-out. The added test comments also make clear that `printf` is the known-safe head and that the suite is proving argument-tail replay, so the original overclaim is addressed.

srid:

✅ **All clear** — `--no-elegance` (the `/simplify` pass already ran this gauntlet).

| Pass | Issues | Result |
| --- | --- | --- |
| Rules checklist | 0 | Clean |
| Fact-check | 0 | Clean |

Notable confirmations from the fact-check pass:
- `findPosixShell`'s lone `catch {}` is intentional best-effort (carries a justifying comment); a miss returns `null`, which makes the whole real-shell suite **skip visibly** via `describe.skipIf` — not silently degrade.
- `process.env.PATH ?? ""` → empty `dirs` → `null` → suite skipped: correct no-shell fallback.
- `.split("\0").slice(0, -1)` correctly strips the trailing NUL `printf` writes after the last field.
- `{ ...process.env, ...env }` correctly layers the test's `HOME` override for the tilde-expansion assertion.

srid:
## Evidence

No on-screen surface (a unit test + a dev-facing README note for the `@kolu/shell-quote` leaf), so there's no UI to capture — but the change exists to **prove a behavior**, so here is that behavior, exercised against a real `bash`.

### 1. The real-shell suite actually runs (not skipped)

```
✓ shellJoin output parsed by a real POSIX shell > word-splits each joined argv back into exactly its inputs   30ms
✓ shellJoin output parsed by a real POSIX shell > re-expands a space-separated leading ~ value to $HOME (documented carve-out)   4ms
  Test Files  1 passed (1)
       Tests  11 passed (11)
```

The 30ms / 4ms (vs 0ms for the pure-function tests) is the cost of actually spawning a shell per corpus entry — the suite ran, it didn't `skipIf`-skip.

### 2. Negative control — the proof has teeth

Feeding `bash` the same 4-token argv two ways, then reading back the argv the shell *actually built* (via `printf '%s\0'`):

```
INPUT argv (4 tokens):
   · "--settings"
   · "{\"ultracode\": true}"
   · "--msg"
   · "don't $x `id` ; rm & echo | cat * (a) {b}"

shellJoin → real bash → argv:  fields=4  deepEqual(input)=true   exit=0
naive join → real bash → argv: fields=0  deepEqual(input)=false  exit=2
   bash error from the naive line: "bash: -c: line 1: unexpected EOF while looking for matching `''"
```

`shellJoin`'s output round-trips **all four tokens byte-for-byte** through a real shell — the spaced JSON blob, the apostrophe in `don't`, and the `$ \` ; & | * ( ) { }` soup all survive as inert data (exit 0). The **naive** space-join of the very same argv doesn't even parse: the bare apostrophe in `don't` makes bash choke on an unterminated quote (exit 2) — and had it parsed, `;`/`&`/`|`/`` ` `` would have split the line and executed `id`/`rm`. That gap is exactly what this test now pins on every CI run.

srid:

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

PR #988, opened 2026-05-27
Right panel mirrors the dock's posture pairing

srid:
The right panel now follows the canvas posture the same way the dock does: **float as a rounded card in tiled mode, flush as a sidebar in maximized mode**. Until now, the panel was a flush sidebar in both postures, which pulled visually against the cockpit feel of the canvas. Closes #986.

| Canvas mode | Dock | Right panel (before) | Right panel (after) |
| --- | --- | --- | --- |
| Tiled | floating card, top-left | flush sidebar | **floating card, top-right** |
| Maximized | flush left sidebar | flush sidebar | **flush right sidebar** |

Both surfaces share `POSTURED_TILED_FLOAT` / `POSTURED_MAXIMIZED_FLUSH` (`packages/client/src/canvas/posturedSurfaceChrome.ts`), so chrome-bar metric changes ripple through one constant instead of two surfaces. Vertical extent diverges intentionally — the dock shrinks to its finite row list (`max-h-[calc(100vh-22rem)]`), the panel claims `bottom-4` because the Code-tab tree benefits from every pixel.

### Working as before

- Width is user-resizable via the drag handle on the panel's left edge
- Persisted width and collapsed state survive reload
- Clicking a `path:line` link in terminal output opens the panel at that file — including the re-click-after-close path (file-ref-link.feature:86)
- Mobile keeps the bottom-drawer host unchanged

### Iteration history

The first cut replaced `@corvu/resizable` in `DesktopHost` with a hand-rolled posture-aware aside + custom drag handle (routed through `capturePointerGesture`). Hickey + Lowy reviewers ran in parallel, then cross-validated each other — eight refactor commits landed on bounds consolidation, gesture-primitive routing, posture-chrome extraction, comment accuracy, re-entry guards, and read/write clamp symmetry. Code-police then caught a 1px ghost border and a stray signal-wrapped DOM ref.

The final fix walked back the Resizable removal: the e2e suite caught that dropping the flex-split shape broke xterm's link-activate callback on the second click after a manual collapse. *The visual posture-pairing now lives entirely on classList on the right `Resizable.Panel`* — the Resizable shell stays put across both postures, so the click-handling surface is master's known-good shape.

_Generated by [\`/do\`](https://github.com/srid/agency) on Claude Code (model \`claude-opus-4-7\`)._

PR #1410, opened 2026-06-19
Stop re-rendering terminals on every maximized-mode switch

srid:
**With a terminal maximized, switching the active terminal made *both* terminals visibly re-render on every switch** — the one you left and the one you opened reflowed their text and repainted, and each switch round-tripped a PTY resize to the shell. This was maximized-specific; tiled (canvas) mode never showed it.

### Why

The cause is a **size mismatch between the active and inactive tiles in maximized posture**. The maximized tile fills the canvas (`absolute inset-0`); every covered (inactive-but-mounted) tile kept its small saved canvas-layout rect, only hidden via `visibility: hidden`. So switching swapped which tile was full-viewport vs small:

| Tile | Before switch | After switch | Container size change? |
| --- | --- | --- | --- |
| A (was active) | `inset-0` (viewport) | small layout rect | **yes — shrinks** |
| B (newly active) | small layout rect | `inset-0` (viewport) | **yes — grows** |

`Terminal`'s `createResizeObserver` (`Terminal.tsx:814`) is gated only on `props.visible`, which is hardcoded `true` for every canvas tile (`App.tsx:222`). So both size changes fire `debouncedFit()` → `fitAddon.fit()`, recomputing cols/rows → a full xterm grid reflow + repaint, *plus* `publishDimensions()` → `client.terminal.resize` (a server PTY round-trip that makes the shell redraw). That's the visible "re-render," and it recurs every switch.

*Not* a component remount (the stable keyed `<For>` and one-xterm-per-id rule already rule that out, #988), and *not* the WebGL renderer swap (the N=2 budget from #1404 keeps both tiles of an A↔B pair on WebGL).

### The fix

Give covered tiles the **same `inset-0` box as the maximized tile** — hidden behind it via `visibility: hidden`, with `z-40` reserved for the maximized one. Switching is then a pure visibility/z-index swap: no container resizes, so the `ResizeObserver` never fires, `fit()` is a no-op, and there's no reflow, repaint, or PTY resize. One CSS-only change in `CanvasTile.tsx`; it removes the accidental coupling between *which tile is active* and *what size a tile body is*.

> _Tradeoff:_ entering maximized posture now sizes every covered tile's PTY to the viewport grid once — a one-time SIGWINCH per tile on the posture toggle, **not** a per-switch cost. `lastActivityAt` keys off agent-state, not raw PTY output, so this doesn't falsely trip staleness/auras.

Existing coverage holds — covered tiles still hide via computed `visibility: hidden` (`canvas.feature` "every non-maximized canvas tile should be hidden") and switching still doesn't remount xterm (#988 regression test).


srid:

✅ **Consensus** after 2 round(s) · lowy + hickey · base `015508d59465`

### Applied (5)
- `lowy-1` Mode→box-geometry mapping sprayed across three co-dependent sites — commit `45d801cc6`
- `lowy-2` Box-equality invariant is the fix but has no guarding assertion — commit `24b02be56`
- `hickey-1` mode→presentation mapping fragmented across style() and classList — commit `45d801cc6`
- `hickey-2` "is-tiled" vs "is-not-tiled" discriminator expressed two contradictory ways — commit `45d801cc6`
- `hickey-3` Geometry now silently absent from covered branch — illegal-state representable — commit `45d801cc6`

### Agreed — no change (1)
- `lowy-3` `tiled`/`covered`/`maximized` predicates restated independently in tile and caller (packages/client/src/canvas/TerminalCanvas.tsx:408-413)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `015508d59465`

### Round 1

**codex** — approved: `false`

The behavior change itself is coherent: covered tiles now share the maximized absolute viewport box, stay mounted with `visibility:hidden`, and the added e2e assertion covers the intended outer-geometry invariant. I found one low-severity issue in the new mode dispatcher: it is not exhaustive, so future render modes would silently fall through to tiled presentation.

Findings:
- `F1` · minor · open — The `presentation()` dispatcher handles `CanvasTileMode` with a `default` branch, so adding a future mode will silently render it as a tiled tile instead of producing an exhaustiveness failure. This is a load-bearing geometry/layering switch, and the repo's `prefer-ts-pattern` rule calls for exhaustive dispatch on string-literal unions. (packages/client/src/canvas/CanvasTile.tsx:245)

**claude** — CODEX raised one minor finding (F1) on the new presentation() dispatcher in CanvasTile.tsx. I agreed and fixed it: the dispatcher used a `switch` with a `default` arm for "tiled", so a future CanvasTileMode would silently fall through to tiled presentation rather than failing to compile. I replaced it with the repo's established `match(props.mode).with(...).exhaustive()` idiom (ts-pattern), which compile-fails if the union grows. This matches existing usage in Terminal.tsx (shouldUseWebgl), App.tsx, RecordButton.tsx, and the documented prefer-ts-pattern convention. I confirmed the App.tsx `<Switch>`-not-`match` note is NOT a counter-precedent — that note is about JSX subtree remounting when a memo returns a fresh object; presentation() returns a plain data object, not JSX, so that concern does not apply. Verified: pnpm typecheck passes (client package green), biome check clean, formatter made no changes. Committed as 2131deda1.

- `F1` **fixed** — AGREED. packages/client/src/canvas/CanvasTile.tsx — the presentation() createMemo dispatched on CanvasTileMode with a `switch` whose `default` arm rendered tiled presentation, so a future fourth mode would silently fall through to tiled instead of failing exhaustiveness. Replaced the switch with `match(props.mode).with('maximized', ...).with('covered', ...).with('tiled', ...).exhaustive()` (added `import { match } from "ts-pattern"`). This is the repo's documented prefer-ts-pattern idiom for value-returning dispatch on a string-literal union, and matches Terminal.tsx's shouldUseWebgl (lines 191-196), App.tsx, RecordButton.tsx, etc. `.exhaustive()` now compile-fails if CanvasTileMode grows a mode, forcing a deliberate geometry/layering decision. Added a comment explaining the choice and noting that the App.tsx `<Switch>`-not-match counter-precedent does not apply (it concerns JSX subtree remounting; presentation() returns a plain data object). visibility: 'hidden' needed an `as const` to keep the literal type. Typecheck + biome clean.

commit: `2131deda1437c16780aa437c494fb54fd0bd4419`

### Round 2

**codex** — approved: `true`

Focused re-inspection confirms the only prior finding is addressed. `CanvasTile` now imports `match` from `ts-pattern` and dispatches all three `CanvasTileMode` values through `.with(...).exhaustive()`, so a future mode cannot silently fall through to tiled presentation. I did not find a regression introduced by this round's fix.

Findings:
- `F1` · minor · resolved — The prior `presentation()` dispatcher used a `default` branch, so future `CanvasTileMode` variants could silently render as tiled presentation instead of forcing a geometry/layering decision. (packages/client/src/canvas/CanvasTile.tsx:239)

_codex on the rebuttal:_ No dispute to adjudicate. CLAUDE agreed with F1 and the current code fixes it: the dispatcher is now exhaustive over `CanvasTileMode`, and the previous silent default path is gone.

srid:

Ran after `/simplify`, scoped to `015508d59465...HEAD` (`--no-elegance`, since the simplify track already ran).

| Pass | Issues | Result |
| --- | --- | --- |
| Rules checklist | 0 | ✅ Clean |
| Fact-check | 0 | ✅ Clean |

**Rules** — all checks pass: `fullViewportBox` reused by exactly 2 `match` branches (no premature abstraction), `match(props.mode).…exhaustive()` follows `prefer-ts-pattern`, `CanvasTileMode` is a proper discriminated union, `isMaximized()`/`isCovered()` still used at 5+ sites (no dead code), inline styles use raw CSS props while Tailwind stays in `class`/`classList`, and the new e2e step polls async DOM state via `waitForFunction(…, { timeout: POLL_TIMEOUT })`.

**Fact-check** — no logic errors. `fullViewportBox` (`position:absolute; inset:0`) shared by maximized and covered enforces the size-equality invariant the test verifies; `visibility:hidden` doesn't affect layout; covered tiles intentionally shed tiled styling (border/outline/transform/opacity) since they're never painted. No false fallbacks, no swallowed errors.

srid:
## Evidence

The fix's effect on a maximized switch is **transient** (a reflow + repaint only *during* the switch), so a still of the steady state can't show it. What a still *can* show — and what the bug reduces to — is the **box geometry of the inactive (covered) tile** while another tile is maximized. I drove a live kolu (`master` vs this branch) with the chrome-devtools MCP, maximized a tile with a second terminal open, and outlined both tiles' real DOM boxes.

The inactive tile is outlined **magenta**; the maximized tile **cyan**.

| Before (`master`) | After (this PR) |
| --- | --- |
| before | after |
| Covered tile is a **small `800×540` box** floating mid-canvas — a different size from the `1141×2009` maximized tile. | Covered tile **fills the same `1141×2009` box** as the maximized tile (coincident outlines). |

Live `getBoundingClientRect()` measurements after maximizing, with a second terminal open:

```
master (pre-PR):   covered {x:555, y:753, w:800,  h:540}   maximized {x:288, y:44, w:1141, h:2009}   coveredSharesMaximizedBox: false
branch (this PR):  covered {x:288, y:44,  w:1141, h:2009}  maximized {x:288, y:44, w:1141, h:2009}   coveredSharesMaximizedBox: true
```

On `master`, switching to the covered terminal grows its container `800×540 → 1141×2009`, which fires `Terminal`'s `ResizeObserver → fitAddon.fit()` → an xterm grid reflow + a server PTY resize — the visible "re-render." On this branch the boxes are identical, so the switch is a pure visibility/z-index swap and nothing resizes.

This invariant is now also guarded mechanically by the new e2e step **`every covered canvas tile should occupy the maximized tile's box`** (`canvas.feature`), which asserts the covered rect equals the maximized rect within 1px — the same property visualized above. `ci::e2e` (which runs that scenario) passes on **both** `x86_64-linux` and `aarch64-darwin`.

> _Capture note:_ this is a live chrome-devtools capture rather than the standard slow-motion e2e-harness **video**, because the warm-box lease probe reported no egress this run so I didn't pursue the lease-based recording. CI itself ran fine on the pool (`kolu-ci-2` for linux, `rasam` for darwin) — the geometry capture above + measurements + the new assertion stand in for the clip.

srid:

The **x86_64-linux** lane ran on _no pool box — resolved via `hosts.json`_ — commit `f05b29110`, **passed** ✓

- **Lane wall** (pipeline): **2m53s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 2m10s | ✓ |
| `ci::home-manager` | 1m27s | ✓ |
| `ci::smoke` | 43s | ✓ |
| `ci::nix` | 43s | ✓ |
| `ci::unit` | 42s | ✓ |
| `ci::flake-check` | 40s | ✓ |
| `_ci-setup` | 33s | ✓ |
| `ci::biome` | 28s | ✓ |
| `ci::fmt` | 27s | ✓ |
| `ci::install` | 9s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/f05b291/timings.jsonl`; pool state is a live `flock` probe.</sub>

PR #1412, opened 2026-06-19
Click a folder path in a terminal to reveal it in the Code tab

srid:
**Clicking a folder path in terminal output now reveals that folder in the Code tab's file tree** — instead of the old `File reference not found` toast. A `path/to/file:42` reference already opens the file at its line; a *folder* reference (`packages/client/`, or any slash path that names a directory) now switches the Code tab to the **All files** tree, expands that folder and its ancestors, and scrolls it into view. The shown file is left untouched — it's a "browse here", not a "open this".

### How a terminal path resolves now

The one resolver (`resolveRef`, replacing `resolveLineRefPath`) returns a discriminated `{ kind: "file" | "directory" }`, with precedence ordered so a more certain match always wins:

```
exact file  ─▶  exact directory  ─▶  unique-basename file
  (open)          (reveal)              (#898 fuzzy fallback)
```

The directory step sits *before* the fuzzy basename fallback on purpose: a real folder must win over a same-named file elsewhere, so `src/core` reveals the `src/core/` folder rather than silently opening a stray `core` script. Directories are inferred from the existing `fsListAll` file list (no new server data), keyed by the trailing-slash folder id Pierre already uses for directory rows, and matched under NFC like file paths.

### Revealing without a race

`@kolu/solid-pierre`'s `FileTree` grew a one-shot `revealRequest` prop. A reveal that is *already pending when the tree first mounts* — the common case, since a folder click from a closed panel or a diff view switches to browse and mounts the tree fresh — is applied through Pierre's **constructor** `initialExpandedPaths`, the same atomic path `selectedPath` uses. A first cut applied it with a post-render `expand()`, which raced Pierre's first paint and dropped the reveal *intermittently* — a cold e2e run caught it where a warm one passed. A later reveal on an already-live tree goes through a deferred `expand()` + scroll instead. Either way the request is **consumed once** (`onRevealHandled`), so a remount can't re-scroll to a stale folder or fight the selected file's own scroll.

> Files open exactly as before — only the previously-dead folder case changed.

### Coverage

- `resolveRef` gains directory unit cases (trailing slash, cwd-relative, absolute, NFC, basename-disabled, and the directory-beats-basename precedence).
- Two e2e scenarios exercise **both** reveal paths: at-mount (panel opens fresh → constructor) and post-mount (folder clicked while already browsing → deferred effect).

### Try it locally

```sh
nix run github:juspay/kolu/savage-goal
```


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `015508d59465`

### Applied (6)
- `lowy-1` Duplicated encapsulation of the Pierre folder-key shape: ancestorDirKeys re-implements @kolu/solid-pierre's ancestorDirectoryPaths — commit `c22c4271e`
- `lowy-2` Resolver leaks Pierre's row-id convention into the otherwise Pierre-agnostic lineRef module — (uncommitted)
- `lowy-3` Two reveal sites duplicate the expand-ancestors + scroll-to-center sequence — commit `f6f2aea79`
- `lowy-4` revealRequest/onRevealHandled is a parallel one-shot consume-once channel beside the existing selectedPath channel — (uncommitted)
- `hickey-1` Two copies of "trailing-slash ancestor directory keys of a path" — commit `c22c4271e`
- `hickey-2` Ancestor-expand walk triplicated across FileTree.tsx — commit `f6f2aea79`

### Agreed — no change (3)
- `lowy-5` ancestorDirKeys re-derives directory prefixes already computed by candidates()/normalize path machinery (packages/client/src/ui/lineRef.ts:233-237 (build loop) and 246-252 (ancestorDirKeys))
- `hickey-3` resolveRef walks the candidate generator twice for file then directory (packages/client/src/ui/lineRef.ts:180-194)
- `hickey-4` handled vs revealDir as two outcome signals for one resolve effect (packages/client/src/right-panel/CodeTab.tsx:428-440)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 round(s) · codex reviewed at `xhigh` reasoning effort · base `015508d59465`

### Round 1

**codex** — approved: `false`

The branch is directionally sound, but I found three correctness gaps in the new folder-reference behavior. The resolver/tree integration is mostly coherent, but the terminal parser still does not cover common folder syntax, the reveal can be silently consumed while the active tree filter hides the target, and directory resolution currently accepts `path:line` as a folder reveal even though a line suffix should imply a file reference.

Findings:
- `F1` · major · open — The slash-path regex still requires a non-slash final segment, so common folder refs with a trailing slash are not fully linkified. `src/` is not matched at all, and `packages/client/` is matched only as `packages/client`, leaving the visible slash outside the xterm link range. That contradicts the new README/tip examples and misses common terminal output such as `ls -F` top-level directories. (packages/client/src/ui/lineRef.ts:64)
- `F2` · major · open — Folder reveal resolves against the full repo path list, but the mounted `FileTree` receives `treeSearch().projectedPaths`. If the user already has a Code-tab search filter that excludes the target directory, `setRevealDir` sends a request for a row that is not in the mounted tree; `FileTree` then no-ops and clears the request, so the click silently fails to reveal anything. (packages/client/src/right-panel/CodeTab.tsx:485)
- `F3` · minor · open — Any directory resolution is treated as a folder reveal, even when the original ref included a line or range. A terminal token like `app/core:12` now reveals `app/core/` and drops the line suffix instead of failing as an invalid file reference, which is misleading because `:12` only makes sense for files. (packages/client/src/right-panel/CodeTab.tsx:479)

**claude** — Fixed all three codex findings and verified green.

- `F1` **fixed** — regex final segment changed from plus to star so trailing-slash folder refs keep the slash
- `F2` **fixed** — clear searchQuery in CodeTab directory branch so a filter-hidden folder still has a row to reveal
- `F3` **fixed** — added hasLine arg to resolveRef skipping the directory step for line-bearing refs

commit: `16a199c88512a3bcf61586031aa092de12571ec5`

### Round 2

**codex** — approved: `false`

F1 and F2 are adequately fixed in the current tree. F3 is only partially fixed: the new `hasLine` gate prevents a line-bearing folder ref from revealing the folder, but it can now fall through to the unique-basename fallback and open an unrelated same-basename file, which is still the wrong behavior for a path that already names a real directory.

Findings:
- `F1` · major · resolved — The slash-path parser previously failed to include trailing-slash folder refs such as `src/` and `packages/client/` in the full link range. (packages/client/src/ui/lineRef.ts:71)
- `F2` · major · resolved — Folder reveals previously resolved against the full path list but could be consumed while the mounted tree was still filtered, causing a silent no-op for hidden targets. (packages/client/src/right-panel/CodeTab.tsx:496)
- `F3` · minor · open — The fix skips directory resolution entirely when `hasLine` is true, but then proceeds to the basename fallback. That means `app/core:12`, where `app/core/` is a real directory, can still open an unrelated unique `core` file elsewhere instead of failing closed. A line-bearing ref that names a real directory should not reveal it, but it also must not fuzzy-open another file. (packages/client/src/ui/lineRef.ts:208)

_codex on the rebuttal:_ F1: accepted, the parser now keeps trailing slashes and covers both single- and multi-segment folder refs. F2: accepted, clearing `searchQuery` before `revealDir` addresses the filtered-tree no-op and is covered. F3: holding firm. The new `hasLine` argument prevents directory reveal, but because it skips the directory step before basename fallback, a real directory with a same-basename file elsewhere can still navigate to the wrong file. That does not satisfy the fail-closed requirement for line-bearing directory refs.

**claude** — Round 2: addressed codex's three findings. F1 (trailing-slash folder linkification) and F2 (clear Code-tab search before revealDir) were already resolved in prior rounds per codex's verdict — no change needed. F3 was a valid catch and I fixed it: a line-bearing ref that names a real directory (e.g. `app/core:12` with `app/core/` a real dir and a unique `core` file at `lib/core`) previously skipped the directory step via `break` and then let the unique-basename fallback fire, opening the unrelated `lib/core`. The directory loop now still detects the match under `hasLine` and returns null (fail closed) instead of breaking past it, so a path that already names a real directory can never fuzzy-open another file. Added a unit test for codex's exact scenario and refreshed the resolveRef doc/comments. check + biome lint + fmt green; 374 @kolu/client unit tests pass.

- `F1` **fixed** — Already resolved in round 1 and confirmed by codex's verdict (responseToRebuttal: accepted). The slash-path branch's final segment is `*` not `+` (packages/client/src/ui/lineRef.ts:71), so `src/` and `packages/client/` linkify in full. No change this round.
- `F2` **fixed** — Already resolved in round 1 and confirmed by codex's verdict (responseToRebuttal: accepted). CodeTab clears the search (`setSearchQuery("")`) before `setRevealDir` so a filter-hidden folder still has a row to reveal (packages/client/src/right-panel/CodeTab.tsx:496). No change this round.
- `F3` **fixed** — Agreed with codex — the round-1 `if (args.hasLine) break;` correctly suppressed the directory reveal but let execution fall through to the unique-basename fallback, so `app/core:12` (app/core/ a real dir, unique `core` file at lib/core) opened the unrelated lib/core. Fixed at packages/client/src/ui/lineRef.ts:208-219: removed the `break`, kept the directory-detection loop running under `hasLine`, and return `args.hasLine ? null : { kind: 'directory', path: hit }` so a line-bearing directory match fails closed before the basename fallback can fire. Exact-file (step 1) still wins for real files with a line, since step 1 runs over all candidates first. Added test 'fails closed (no basename fuzzy) for a line-bearing directory ref' with codex's exact inputs (repoPaths ['app/core/one.ts','lib/core'], rawPath 'app/core', hasLine true → null) and updated the resolveRef doc + step-2 comment.

commit: `d60eaeef146f6d766b0d726f74d87863948e435f`

### Round 3

**codex** — approved: `true`

All three findings are resolved in the current tree. F1 and F2 remain fixed from round 1, and the round 2 change for F3 now detects a real directory match even when `hasLine` is set and returns `null` before basename fallback can open an unrelated file.

Findings:
- `F1` · major · resolved — The slash-path parser previously failed to include trailing-slash folder refs such as `src/` and `packages/client/` in the full link range. (packages/client/src/ui/lineRef.ts:71)
- `F2` · major · resolved — Folder reveals previously resolved against the full path list but could be consumed while the mounted tree was still filtered, causing a silent no-op for hidden targets. (packages/client/src/right-panel/CodeTab.tsx:496)
- `F3` · minor · resolved — A line-bearing ref that named a real directory could previously skip directory handling and fall through to unique-basename fallback, opening an unrelated same-basename file. (packages/client/src/ui/lineRef.ts:215)

_codex on the rebuttal:_ F1: accepted; the parser keeps trailing slashes in the matched text. F2: accepted; clearing `searchQuery` before `setRevealDir` addresses the filtered-tree no-op. F3: accepted; the directory loop no longer breaks past a `hasLine` directory match, and `return args.hasLine ? null : ...` prevents fallback from opening an unrelated file.

srid:

base `015508d59` · passes: **rules**, **fact-check** (elegance skipped — `/simplify` already ran this gauntlet)

| Pass | Result |
| --- | --- |
| Rules | ✅ clean — all built-in + project rules passed |
| Fact-check | 1 finding, fixed |

**Fact-check — silent consume of a failed reveal** (`packages/solid-pierre/src/FileTree.tsx`)
The deferred reveal effect called `onRevealHandled()` unconditionally after `safeApply(...)`, even when the `if (!tree) return` guard skipped the actual reveal — so a not-yet-live tree (or a throwing `revealDirectory`) would silently consume the request with no retry. Moved `onRevealHandled()` inside `safeApply`, after `revealDirectory` runs, so the request is consumed only once the reveal actually applies; otherwise it stays pending for the next mount's constructor path. Fail-loud, not silent-drop. → `67310b9f0`

srid:
## Evidence

A folder path printed in terminal output (`app/core`) is now clickable: clicking it switches the Code tab to the **All files** tree, expands that folder and its parents, and scrolls it into view — no file is opened. Captured live (chrome-devtools driving a `dev-auto` build of this branch).

**Click `app/core` in the terminal → the folder reveals in the tree:**

folder reveal

**Before → after:**

before and after

▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=folder-ref-reveal.mp4

srid:

The **x86_64-linux** lane ran on **`kolu-ci-5`** (`idliv2-02`) — commit `38f262e41`, **passed** ✓

- **Lane wall** (pipeline): **3m15s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 2m34s | ✓ |
| `ci::home-manager` | 1m27s | ✓ |
| `ci::flake-check` | 53s | ✓ |
| `ci::nix` | 52s | ✓ |
| `ci::smoke` | 49s | ✓ |
| `ci::unit` | 33s | ✓ |
| `_ci-setup` | 32s | ✓ |
| `ci::pnpm-hash-fresh` | 29s | ✓ |
| `ci::fmt` | 16s | ✓ |
| `ci::biome` | 16s | ✓ |
| `ci::install` | 9s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/38f262e/timings.jsonl`; pool state is a live `flock` probe.</sub>

Issue #1414, opened 2026-06-19
Production (nix-built) kolu silently ignores KOLU_STATE_DIR — wrapper export is unconditional (regression from #531)


srid:
## Summary

The **nix-built (production) `kolu` binary silently ignores the `KOLU_STATE_DIR` environment variable**. The production wrapper *unconditionally* re-exports `KOLU_STATE_DIR` to `$XDG_CONFIG_HOME/kolu` (i.e. `~/.config/kolu`) before exec'ing the inner binary, clobbering whatever the caller set.

This is a **regression**: before #526 / #531 the wrapper used a `${KOLU_STATE_DIR:-…}` fallback, so `KOLU_STATE_DIR=/foo kolu` worked. There is now **no way to run the production binary against a custom state dir** — short of overriding `HOME` itself, which has unrelated side effects.

## Root cause (file:line)

`default.nix:326` — the `default` (production) wrapper:

```nix
makeWrapper ${koluBin}/bin/kolu $out/bin/kolu \
  --run 'export KOLU_STATE_DIR="''${XDG_CONFIG_HOME:-$HOME/.config}/kolu"'
```

The `export` is **unconditional** — no `:-` fallback — so it overwrites any inherited `KOLU_STATE_DIR`. The shipped wrapper confirms it verbatim:

```bash
#! …/bash -e
export KOLU_STATE_DIR="${XDG_CONFIG_HOME:-$HOME/.config}/kolu"
exec "…/kolu-bin/bin/kolu"  "$@"
```

Meanwhile `packages/server/src/state.ts:135` reads `process.env.KOLU_STATE_DIR` and logs it (`state.ts:144`), so the server faithfully honors whatever the wrapper hands it — it just never sees the caller's value.

### How it got here

#530 wanted to stop **tests** from silently corrupting production state at `~/.config/kolu`. The fix (#531) split the derivation into `koluBin` (no `KOLU_STATE_DIR` — crashes if unset; what tests use) and `default` (sets it unconditionally). Dropping the `:-` in `default` was framed as "tests can't accidentally inherit the production path" — but **tests consume `koluBin`, not `default`** (`justfile:122`: `nix build .#koluBin`). So the unconditional `export` in `default` was never needed for test isolation; it only removed the production override as collateral damage.

## Reproduction

Run the production binary in full isolation (isolated `HOME` proves env vars *do* reach the process; only `KOLU_STATE_DIR` is ignored):

```sh
env -i PATH="$PATH" \
[...]

PR #1416, opened 2026-06-19
fix(terminal): cap WebGL contexts to stop focus-churn GPU-VRAM leak on AMD (#1399)

### LINKED ISSUE #1399: Chromium GPU page faults cascade into gnome-shell SIGABRT while running kolu — follow-up to #1308
## Problem

GNOME Shell (Mutter 49.2, Wayland) crashes with SIGABRT after Chromium's GPU process triggers 10 GPU page faults, causing a MODE1 GPU reset. This occurs specifically while kolu is running in Chromium with multiple active terminals and agents.

**Chromium is only used to run kolu; no similar crash has occurred with other sites.** This strongly suggests kolu's rendering patterns are the cause, not a generic Chromium/Mesa/Mutter bug.

Proposals P0/P1/P3/P5 from #1308 were merged on Jun 14; P2 and P4 remain unimplemented. While we cannot prove P2/P4 are the specific remaining issue (vs. some other kolu rendering pattern not covered by #1308), they are the most likely candidates given the correlation with the earlier crash.

### Pre-crash Mutter assertion failures (14:30–14:44)

6 occurrences of `meta_window_set_stack_position_no_sync: assertion 'window->stack_position >= 0' failed` in the 14 minutes before the crash — identical to the pre-crash signature from #1308:

```
Jun 18 14:30:42 gnome-shell[387403]: meta_window_set_stack_position_no_sync: assertion 'window->stack_position >= 0' failed
Jun 18 14:30:43 gnome-shell[387403]: meta_window_set_stack_position_no_sync: assertion 'window->stack_position >= 0' failed
Jun 18 14:32:07 gnome-shell[387403]: meta_window_set_stack_position_no_sync: assertion 'window->stack_position >= 0' failed
Jun 18 14:32:09 gnome-shell[387403]: meta_window_set_stack_position_no_sync: assertion 'window->stack_position >= 0' failed
Jun 18 14:44:01 gnome-shell[387403]: meta_window_set_stack_position_no_sync: assertion 'window->stack_position >= 0' failed
Jun 18 14:44:03 gnome-shell[387403]: meta_window_set_stack_position_no_sync: assertion 'window->stack_position >= 0' failed
```

### Pre-crash Chromium GPU compositor degradation (14:37–14:44)

3 occurrences of `SharedImageManager::ProduceSkia` errors — Chromium's GPU compositor is failing to manage shared images before the crash:

```
Jun 18 14:37:23 chromium-browser[4040697]: SharedImageManager::ProduceSkia: Trying to Produce a Skia representation from a non-existent mailbox.
Jun 18 14:39:46 chromium-browser[4040697]: SharedImageManager::ProduceSkia: Trying to Produce a Skia representation from a non-existent mailbox.
Jun 18 14:44:17 chromium-browser[4040697]: SharedImageManager::ProduceSkia: Trying to Produce a Skia representation from a non-existent mailbox.
```

### Phase 1: Chromium GPU page faults (14:48:54)

10 GPU page faults from `chromium:cs0` (pid 4040697) on the gfxhub — UTCL2 protection faults from both the SQC (shader data) and CB/DB (color/depth buffer) clients:

```
Jun 18 14:48:54 kernel: amdgpu 0000:3d:00.0: [gfxhub] page fault (src_id:0 ring:24 vmid:7 pasid:32791)
Jun 18 14:48:54 kernel: amdgpu 0000:3d:00.0:   in process chromium pid 4040697 thread chromium:cs0 pid 4040728
Jun 18 14:48:54 kernel: amdgpu 0000:3d:00.0:   in page starting at address 0x000000003f800000 from client 0x1b (UTCL2)
Jun 18 14:48:54 kernel: amdgpu 0000:3d:00.0:   GCVM_L2_PROTECTION_FAULT_STATUS:0x00701431
Jun 18 14:48:54 kernel: amdgpu 0000:3d:00.0:   Faulty UTCL2 client ID: SQC (data) (0xa)
Jun 18 14:48:54 kernel: amdgpu 0000:3d:00.0:   PERMISSION_FAULTS: 0x3
… (10 faults total, also hitting CB/DB client on ring:88, address 0x0000ffffffbfe000)
```

### Phase 2: GPU ring timeout on gnome-shell context (14:49:04)

After Chromium's faults poisoned the ring, gnome-shell's next submission hung waiting for the ring to drain:

```
Jun 18 14:49:04 kernel: amdgpu 0000:3d:00.0: ring gfx_0.1.0 timeout, signaled seq=47746807, emitted seq=47746808
Jun 18 14:49:04 kernel: amdgpu 0000:3d:00.0: Process information: process .gnome-shell-wr pid 387403 thread gnome-shel:cs0 pid 387447
```

### Phase 3: GPU reset and cascade kill (14:49:04–14:49:33)

```
Jun 18 14:49:04 kernel: amdgpu: GPU reset begin!
Jun 18 14:49:05 kernel: amdgpu: MODE1 reset
Jun 18 14:49:07 kernel: [drm] VRAM is lost due to GPU reset!
Jun 18 14:49:07 kernel: amdgpu: GPU reset(5) succeeded!
Jun 18 14:49:08 gnome-shell[387403]: amdgpu: The CS has cancelled because the context is lost.
Jun 18 14:49:08 systemd-coredump: Process 387403 (.gnome-shell-wr) terminated abnormally with signal 6/ABRT  → 46.1 MB core
Jun 18 14:49:08 systemd-coredump: Process 387956 (Xwayland) terminated abnormally with signal 6/ABRT         → 3.1 MB core
Jun 18 14:49:33 systemd-coredump: Process 389502 (librewolf) terminated abnormally with signal 11/SIGSEGV     → 275.6 MB core
```

## Analysis

The crash chain: **Chromium GPU page faults → GPU ring hang → GPU reset → all GPU contexts destroyed → cascade kill of gnome-shell, Xwayland, and LibreWolf.**

Since Chromium only runs kolu and no similar crash occurs with other sites, kolu's rendering patterns are the probable cause. The P0/P1/P3/P5 mitigations from #1308 were in place but did not prevent the crash. The two remaining unimplemented proposals — P2 and P4 — are the most likely remaining contributors, though we cannot rule out other kolu rendering patterns not covered by #1308.

### P2: ResizeObserver → `fitAddon.fit()` oscillation

`packages/client/src/terminal/Terminal.tsx:782–789` calls `debouncedFit()` unconditionally on every ResizeObserver callback. If `fit()` produces sub-pixel size differences, the observer fires again, creating a sustained oscillation loop. This produces the geometry instability that triggers Mutter's `meta_window_set_stack_position_no_sync` assertion.

### P4: `useVisualViewportHeight` sub-pixel churn

`packages/client/src/useVisualViewportHeight.ts:22` unconditionally sets `--app-h` on `<html>` to `${vv.height}px` on every `visualViewport.resize` event. On Wayland, Mutter tracks this as window geometry. Sub-pixel height changes produce the geometry instability observed in the 14 minutes before the crash.

## Proposal

Implement the two remaining proposals from #1308:

### P2: Break the ResizeObserver oscillation

In `packages/client/src/terminal/Terminal.tsx`, add a size-change threshold:

```ts
createResizeObserver(() => containerRef, ({ width, height }) => {
  if (!props.visible) return;
  if (Math.abs(width - lastFitSize.width) < 2 &&
      Math.abs(height - lastFitSize.height) < 2) return;
  lastFitSize = { width, height };
  debouncedFit();
});
```

### P4: Guard `useVisualViewportHeight` against sub-pixel churn

In `packages/client/src/useVisualViewportHeight.ts`, skip the CSS variable update when the height change is less than 1px:

```ts
let lastH = 0;
const sync = () => {
  const h = vv.height;
  if (Math.abs(h - lastH) < 1) return;
  lastH = h;
  document.documentElement.style.setProperty("--app-h", `${h}px`);
};
```

If the crash stops after deploying P2/P4, that confirms they were the cause. If it continues, further investigation is needed into other kolu rendering patterns.

## Environment

- GNOME Shell 49.2 / Mutter (Wayland)
- Chromium (latest stable, running kolu exclusively)
- AMD Radeon PRO W6800 (Navi 21, `1002:73a3`) + Radeon PRO W6600 (Navi 23, `1002:73e3`)
- Mesa/Gallium3D 25.2.6
- Linux 6.12.68
- NixOS
- kolu local dev branch at commit `b978d857` (branch `feat/tile-placement-inherit-size`), which includes #1308 P0/P1/P3/P5 fixes but not P2/P4

## Relationship to #1308

This is the same failure pattern, same hardware, and same pre-crash Mutter assertion signature as #1308. The earlier issue was closed after merging P0/P1/P3/P5; this issue documents the crash recurring with only P2/P4 remaining unimplemented. Implementing P2/P4 should eliminate the residual compositor stress.

srid:
Fixes #1399 (follow-up to #1308). On **Linux + AMD**, running kolu in Chromium across several terminals slowly exhausted GPU VRAM until the amdgpu driver did a MODE1 reset — blanking terminals and, on the reporter's machine, SIGABRT-ing the whole GNOME/Wayland session.

## Root cause — WebGL context *churn*, not a per-context leak

kolu kept the hardware WebGL renderer on only the **2** most-recently-used tiles (`WEBGL_TILE_BUDGET`, #1403/#1404, chosen for the A↔B font-reflow). A budget *smaller than the working set* destroys+recreates an xterm WebGL context on **every focus switch** across more terminals. On Chrome+AMD a torn-down context's VRAM is reclaimed only at **JS GC** — so sustained focus-churn outruns GC and VRAM climbs to exhaustion → page fault → ring timeout → GPU reset. Firefox is immune (different GPU-process model), which is why `renderer=dom` / Firefox were the reporter's workarounds.

I verified that `loseContext()` + canvas-zeroing (and even forking `@xterm/addon-webgl`'s `dispose`) do **not** free the VRAM synchronously — it's GC-bound — so the fix is to **stop churning**, not to release harder.

## The fix

Replace the fixed top-2 budget with a **context-aware cap**: admit the most-recently-active tiles that fit under `WEBGL_CONTEXT_CAP = 12`, counting each tile's real cost (main pane + active split). The whole working set then holds WebGL **churn-free**, while the count stays below Chrome's **measured 16**/tab limit (#575) so nothing is evicted — the "real cap management" #1404 noted a larger budget would need. Holding more *live* contexts does not itself leak (measured flat), so the higher cap only raises a bounded steady-state baseline, not a leak.

## Evidence (A/B — real AMD Radeon 780M, headless hardware-GL, production `nix build`)

Identical 2000-swap focus-cycle accelerant (≈ a multi-hour 6-terminal session — the leak scales with switch count, not wall-clock):

| build | WebGL | VRAM over the storm |
|---|---|---|
| unfixed (budget=2) | on (2 tiles) | **1032 → 4001 MB** (exhaustion) |
| `loseContext` + canvas-zero | on | 694 → 3990 (ineffective — VRAM is GC-bound) |
| **this PR (cap=12)** | **on (all tiles)** | **flat ~710 MB, Δ = 0** |

Full reproduction harness + raw JSON live on [`repro/1399-handoff`](https://github.com/juspay/kolu/tree/repro/1399-handoff) under `.repro/`.

## Scope / known limitation

Eliminates the leak for working sets up to the cap (~12 terminals — covers the reporter's 5–6 and realistic use). Beyond ~12 *actively-cycled* terminals, Chrome's 16-context ceiling forces some eviction-churn regardless (intrinsic; full elimination would need a single-shared-context xterm renderer, out of scope). The underlying "Chrome holds lost-context VRAM until GC" is arguably a Chrome/ANGLE issue this works around.

## Tests / docs

- `webglBudget.ts` — extracted `admitWebglTiles` + `WEBGL_CONTEXT_CAP` (pure, no SolidJS) so the policy is unit-testable in isolation.
- `webglBudget.test.ts` — regression guard: a realistic working set is admitted **in full** (the old `slice(0,2)` truncated to 2), capped by recency past the limit, with splits counted as 2 contexts.
- Synced `renderer-fidelity-learnings.md` / `memory-learnings.md` + a changelog entry.

## Side-finding

Filed #1414 — the nix wrapper unconditionally clobbers `KOLU_STATE_DIR` (found while isolating repro instances).


srid:

✅ **Consensus** after 2 round(s) · lowy + hickey · base `dfd4310c7414`

### Applied (5)
- `lowy-1` Split-cost rule duplicated between costOf and holdsWebgl (volatile axis sprayed across two sites) — commit `6d92fb953`
- `lowy-2` Terminal.tsx canUseWebgl comment still describes the pre-#1399 (2-tile / ≤4-context) model — commit `f0e17deaa`
- `hickey-1` Tile WebGL-context occupancy fragmented across costOf and holdsWebgl — commit `6d92fb953`
- `hickey-2` WebGL budget policy narration scattered across 4+ source files, now partly stale — commit `f0e17deaa`
- `hickey-3` WEBGL_CONTEXT_CAP and the split-cost rule co-own the #575 bound but live apart — commit `6d92fb953`

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `dfd4310c7414`

### Round 1

**codex** — approved: `false`

The implementation itself is coherent: the context-counted budget is isolated in a pure module, `useTerminalStore` uses the same split predicate for cost and grant, and the unit coverage pins the new policy. The branch is not mergeable as-is because existing e2e coverage still asserts the old fixed two-tile behavior and will fail against the new cap-based admission.

Findings:
- `F1` · blocking · open — The canvas e2e scenario still encodes the old `WEBGL_TILE_BUDGET = 2` behavior. With this branch, three single-pane canvas tiles cost 3 contexts and all fit under `WEBGL_CONTEXT_CAP = 12`, so line 344's `exactly 2`, line 345's `tile 1 should use dom`, and the later `exactly 2` checks will time out/fail. The renderer preference scenario at line 371 has the same stale `exactly 2` auto-mode expectation before forcing WebGL. (packages/tests/features/canvas.feature:336)

**claude** — Codex's single blocking finding (F1) is correct: the branch swapped the source-side WebGL budget from a fixed top-2 to a context-aware LRU cap (WEBGL_CONTEXT_CAP=12 in webglBudget.ts / useTerminalStore.ts / Terminal.tsx) but never updated the e2e scenarios in packages/tests/features/canvas.feature, which still asserted the old budget=2 behavior. Verified the diff against the base touched no .feature file. Under the new code, 3 single-pane tiles cost 3 contexts (well under 12) so all 3 hold WebGL, making the old `exactly 2`/`tile 1 dom` assertions time out. Fixed both stale scenarios; left eviction-above-cap coverage to the existing webglBudget.test.ts (CI-runnable without a GPU) rather than adding a slow 13+-tile e2e scenario. The `#1403` reference at canvas.feature:363 (active-split-inherit scenario) is left intact — that scenario's assertions stay correct under the cap and the citation is to the reflow concern, not the budget number.

- `F1` **fixed** — Confirmed and fixed. canvas.feature was not modified by this branch (empty diff against base) yet its source-side counterpart changed to cap-based admission, so the renderer scenarios were stale. Rewrote two scenarios in /home/srid/code/kolu/.worktrees/gpu/packages/tests/features/canvas.feature: (1) line 336 'WebGL is budgeted to the 2 most-recently-active tiles' -> renamed to 'WebGL is admitted to the whole working set under the context cap'; changed line 344 `exactly 2` -> `exactly 3`, dropped the `canvas tile 1 should use the dom renderer` assertion (tile 1 now holds WebGL under cap 12), and updated the two later `exactly 2` -> `exactly 3` while keeping the ping-pong WebGL-stays invariant. (2) line 367/372 'Renderer preference webgl' auto-mode pre-assertion `exactly 2` -> `exactly 3`. All step phrases still map to existing step defs in canvas_steps.ts (exactly {int}..., canvas tile {int} should use..., I click canvas tile {int}). Took codex's optional suggestion to ADD a >12-context eviction scenario only partially: declined the e2e variant (slow/flaky 13+-tile create) because eviction at the cap boundary is already pinned by webglBudget.test.ts (admits exactly 12 of 20, etc.); referenced that test in the scenario comment. biome (the project formatter) does not touch .feature files, so no fmt run was needed.

commit: `82e0dc0a5b0914afcfca62f880b249bd08213e4f`

### Round 2

**codex** — approved: `true`

F1 is resolved. The current `canvas.feature` now expects all three under-cap single-pane tiles to use WebGL in auto mode, and the renderer preference scenario's precondition also matches the new context-cap policy. I did not see a regression introduced by this round's feature-file change.

Findings:
- `F1` · blocking · resolved — The canvas e2e scenario still encoded the old `WEBGL_TILE_BUDGET = 2` behavior, which would fail now that three single-pane tiles cost only 3 contexts and fit under `WEBGL_CONTEXT_CAP = 12`. (packages/tests/features/canvas.feature:336)

_codex on the rebuttal:_ F1: accepted as fixed. I verified the scenario now expects `exactly 3` WebGL-rendered canvas tiles and asserts tile 1 remains WebGL after focus changes. I also accept leaving >12-context eviction to `webglBudget.test.ts`; my earlier e2e eviction note was optional, and the unit tests are the better fit for cap-boundary behavior without creating a slow GPU-dependent scenario.

srid:

Reviewed the #1399 WebGL-context-cap diff against the merge-base (`--no-elegance` — the elegance pass is the simplify track, which already ran).

| Pass | Findings | Outcome |
|---|---|---|
| Rules (27: 7 built-in + 20 project) | 0 | Clean |
| Fact-check | 1 | Fixed |
| Elegance | — | Skipped (simplify ran) |

**Fact-check (fixed):** `admitWebglTiles` didn't validate its `costOf` contract — a non-positive cost would silently over-admit tiles past the cap (the #575 over-limit class this fix exists to prevent). Added a fail-fast guard (`cost <= 0` throws) + 2 regression tests — commit `19834acf`. Follow-up `641990ff` corrected the precondition comment's rationale (it's a bounded `for…of`, so the real risk is over-admission, not the "infinite loop" the fact-check pass first wrote).

srid:
## Evidence

This fix has **no on-screen diff** — WebGL renders terminals identically before and after; the behavior that changes is **GPU VRAM over time**, which a screen recording can't show. So the evidence is the measured A/B, captured on **real AMD hardware** (Radeon 780M / RADV) running the **production `nix build`** in headless **hardware** GL (ANGLE-Vulkan on the amdgpu render node — verified `ANGLE (AMD, Vulkan … RADV PHOENIX)`, not SwiftShader). The accelerant: 6 terminals in canvas mode, **2000 rapid `nextTerminal` focus-cycles** (≈ a multi-hour 5–6-terminal session — the leak scales with switch count, not wall-clock), sampling discrete-GPU `mem_info_vram_used` throughout, then 30 s idle + forced GC.

**VRAM over the 2000-swap storm:**

| swap | unfixed (`budget=2`) | **this PR (`cap=12`)** |
|---:|---|---|
| 0 (baseline) | 1032 MB | 710 MB |
| 500 | 1941 | 710 |
| 1000 | 3159 | 710 |
| 1500 | **4000** (4 GB ceiling) | 710 |
| 1750 | 4001 → GTT spill | 710 |
| 2000 → idle+GC | (Chrome evicted all contexts) | **710 — Δ 0** |
| WebGL tiles held | 2 | **6 (all)** |

Unfixed, VRAM climbs ~2 MB/switch straight to the 4 GB ceiling (then spills to GTT and Chrome evicts every context) — the path that, on the reporter's discrete W6800, faults the GPU and SIGABRTs the session. With the cap, VRAM is **flat across the entire storm with WebGL on every tile**, and a control run confirmed `loseContext()`+canvas-zeroing does *not* help (VRAM is GC-bound) — which is why the fix is "don't churn," not "release harder."

**Why no e2e video:** the change is invisible on screen by construction (item 3 of the evidence rule — a fix whose behavior "genuinely can't be observed" on screen). The reproduction harness + raw per-swap JSON live on [`repro/1399-handoff`](https://github.com/juspay/kolu/tree/repro/1399-handoff) under `.repro/`. Cross-browser safety (cap=12 ≤ every engine's limit: Chrome 16, Safari 16, Firefox 300; mobile shows one tile) was confirmed separately.

PR #1426, opened 2026-06-19
feat(atlas): derive backlinks + fail the build on dead internal links

srid:
Atlas notes have always *linked* to each other (`[electricity](./electricity.html)`), but a note could never tell you **who links to it**. This adds the inbound half of that graph — a **Referenced by** list on every note page — and, because we're now reading the link graph at build time anyway, makes a **dead internal link fail the build** instead of shipping a 404 into the committed `dist/`.

This is **step 2** of the [meta.mdx](https://kolu.dev/atlas/meta.html) plan ("Fixing the tree/graph"). It's deliberately the cheapest, highest-leverage step: it reuses edges the Atlas already has, adds zero authoring surface, and defers the interactive graph view until the edge set proves dense.

## How it works

```
src/content/atlas/*.mdx ──build──▶ buildAtlasGraph(notes)          (src/lib/atlasGraph.ts)
   ./slug.html links + parents     ├─ invert edges → backlinks map
                                    └─ dead ./slug.html → throw (fails the build)
                                              │
                          [...slug].astro ────┴──▶ <Backlinks/> "Referenced by" on each page
```

- **Derived, not authored.** `buildAtlasGraph` inverts every note's existing `./slug.html` prose links **and** `parents` edges into a backlinks map. No new `backlinks:` frontmatter — it reuses the source of truth the Atlas already maintains.
- **Fail-fast dead-link gate.** A `./slug.html` whose slug names no note throws at build (`./index.html` is exempt — it's the generated index). Code spans/fenced blocks are stripped *before* matching, so a link written **as documentation** (meta.mdx literally contains `./slug.html` and `./other.html` as inline-code examples) is never mistaken for a real edge — the trap that would otherwise red the build on its own plan note.
- **Idempotent.** Backlinks are title-sorted with a pinned locale; `ci::atlas-sync` rebuilds under a scrambled TZ/locale and confirms byte-identical output.

## Design philosophy

- **Reuse the existing source of truth** — backlinks come from the existing `./slug.html` / `parents` edges, not a parallel hand-kept graph; `toParents` is now shared with `indexTree.ts` rather than duplicated.
- **Fail fast, no fallback** — a dead internal link crashes the build loudly; it does not silently drop the link or render an empty section.
- **Not an electricity boundary** — this is a leaf-tier Astro build helper (a bounded inversion over the collection), sibling to `indexTree.ts`, not a `@kolu/*` package.

## Testing

This Astro subproject reserves `node --test` for `build/` helpers; `src/lib` graph builders (`indexTree.ts`) are covered by the **build gate** itself. `atlasGraph.ts` follows that precedent — the dead-link gate is exercised on every `just atlas::build` / `ci::atlas-sync`. Verified red→green locally (a probe note with a dead link fails with a clear message; removing it builds green). Backlinks verified correct and deduped (e.g. `second-brain` lists `meta` once, though `meta` links it via both prose **and** `parents`).

## Scope

`docs/atlas/` only. No `content.config.ts` schema change (`tags`/`supersedes` are later steps). Regenerated `dist/` committed (the 45-page churn is Astro re-hashing scoped styles when a component is added — the one-time cost meta.mdx flagged).


srid:

✅ **Consensus** after 2 round(s) · lowy + hickey · base `7ac550792f33`

Independent findings: lowy=7, hickey=4

### Applied (6)
- `lowy-2` The "what is a note edge" volatility is sprayed across two independent collection walks — commit `81a4e9658`
- `lowy-3` Graph/gate and index operate on different note sets — a draft-filtering asymmetry across sibling builders — commit `9e922707d`
- `lowy-4` Locale-pinned title comparator duplicated rather than shared with the sibling it sits beside — commit `47270d738`
- `hickey-1` NoteRef and related:{id,title}[] are the same note-ref concept, expressed twice — commit `c4a1a5031`
- `hickey-2` Locale-pinned title comparator duplicated across the two graph builders — commit `47270d738`
- `hickey-3` Note-ref construction map(id => ({id, title: byId.get(id)!.data.title})) duplicated — commit `1018dce8f`

### Agreed — no change (5)
- `lowy-1` Electricity test: this is an application concern, not a receptacle — correctly placed as a leaf (docs/atlas/src/lib/atlasGraph.ts:39)
- `lowy-5` Interface stability holds under the encapsulated axis — single in-tree consumer is not disqualifying (docs/atlas/src/lib/atlasGraph.ts:24)
- `lowy-6` AtlasGraph wrapper interface adds a single-field indirection with no second member on the horizon (docs/atlas/src/lib/atlasGraph.ts:24)
- `lowy-7` Dead-link gate's `index` exemption is a magic-slug carve-out inside the edge walk (docs/atlas/src/lib/atlasGraph.ts:52)
- `hickey-4` AtlasGraph single-field wrapper adds a struct concept around one map (docs/atlas/src/lib/atlasGraph.ts:24-28)

srid:
## Codex ⇄ Claude debate — cut short (one finding, applied)

The codex debate did **not** run to consensus — the workflow aborted on a Claude-author **session usage limit**, not a review outcome. Codex completed **round 1** and raised exactly one finding, which was applied before the abort:

- **F1 (major) — fixed in `800cacab6`.** The link extractor only matched the `./slug.html` markdown form, but ~half the corpus (≈50 of 59 links) writes same-directory note links **bare** as `slug.html` — which Astro renders as working sibling links in the flat `dist/`. So the graph silently dropped those real edges and the dead-link gate never checked them. Fix: make the leading `./` optional in `NOTE_LINK` (code-stripping still runs first, so meta.mdx's inline-code examples stay non-edges; the `index` exemption and `#anchor` support are preserved). 11 pages gained previously-missed backlinks; verified build-green, gate red→green on a bare-form dead link, and `dist/` idempotent.

Codex's single finding was addressed; the missing piece is only codex's confirming re-review (blocked by the session limit). `/simplify` and code-police were intentionally skipped for this PR.

PR #1431, opened 2026-06-19
feat(terminal): sleep an idle terminal, wake it in place

srid:
Implements the **Sleeping terminals** plan of record ([Atlas note](https://kolu.dev/atlas/sleeping-terminals.html), #1429).

You leave a terminal running for days while blocked — waiting on a review, a deploy, a person. It still costs a live shell, a GPU context, and a slot in your head. **Sleep** is the missing third verb between close (throws the thread away) and keep (noise): freeze it now, **wake** it — in place, agent and all — when you're unblocked.

## The shape

Sleep is **session restore scoped to one terminal, on demand**. The whole feature reuses the session-restore spine; only one persisted cell and two RPCs are new.

- **Sleep** (`terminal.sleep`, server): snapshot the terminal tree via `snapshotTerminal` (factored from the existing `snapshotSession` projection) into a new persisted `sleepingTerminals` cell — *persist before kill* — then the client tears down the live terminal through the normal `handleKillWithSubs` path (an explicit kill fires no `terminalExit`, so no spurious "exited" toast).
- **Wake** (client): respawn the record through the same primitives session restore uses — `handleCreate` / `handleCreateSubTerminal` / `seedPanel` / `resumeAgentCommand` — then drop the record (`terminal.wake`). A woken terminal is indistinguishable from a session-restored one.

## How it honors the design philosophy

- **Reuse the source of truth** — a sleeping terminal *is* a `SavedTerminal`; the cell clones the `session` cell pattern; wake reuses the restore primitives. No parallel persistence, no hand-rolled snapshot.
- **Fail fast, no fallbacks** — a missing `$PWD` on wake surfaces the error (no silent re-home); the new persisted key ships a conf migration (`1.27.0`) rather than a defaulting fallback.
- **Volatility boundary** — the `sleepingTerminals` cell is disjoint from `session`, so the live-session autosave never clobbers it and a restart rehydrates it **as sleeping**, never auto-woken.

## What's here

| Layer | Change |
|---|---|
| Schema | `SleepingTerminalSchema` + `sleepingTerminals` cell + `terminal.sleep`/`wake` (`common/`) |
| Persistence | conf migration `1.27.0`, `sleepingTerminals.ts` domain module, cell wiring (`server/`) |
| Orchestration | `useSleepingTerminals` singleton (sleep = persist+kill, wake = respawn+drop) |
| Triggers | per-tile **☾** button + command-palette "Sleep terminal" |
| Renderers | Dock **Sleeping** section (both modes) · folded `SleepingCanvasTile` on the canvas |
| Fidelity | optional `subPanel.activeSubTab` / `focusTarget` persistence (lifts session restore too) |

`just check` and `just fmt` are green.

## Remaining in this PR (tracked, pushing follow-up commits)

Workspace-switcher **Sleeping** column · the louder-awaiting/alerts-everywhere pass · the 48h idle-nudge · e2e (`sleep → wake` round-trip + restart-stays-asleep) · finalize the Atlas note status. The review gauntlet + CI + evidence run after those land.


srid:
## Evidence — live verification

Drove a local dev build (random ports, production `kolu.service` untouched) through the full round-trip with chrome-devtools. Confirmed end to end:

1. **Sleep** — clicked the tile's **☾** button → the live terminal was released (PTY/agent gone) and a **SLEEPING** section appeared in the Dock (`srid · just now`).
2. **Folded canvas tile** — the sleeping terminal renders on the canvas as a real tile: a dashed, moonlit `☾ asleep` card with its intent/`$PWD` (`📁 srid`), a `PTY released` note, and a **Wake** button. Structurally a genuine tile — `cursor: grab` drag header + 8 resize handles + its saved 800×540 footprint (same `createDraggable` machinery as a live `CanvasTile`, with drag/resize routed to `terminal.setSleepingLayout`).
3. **Wake** — clicked the Dock row → the terminal respawned in place with its theme (`Overnight Slumber`) and `$PWD` restored, and the record cleared.
4. **Empty-state bug found + fixed** (`1f2c7c2b1`) — sleeping the *last* live terminal dropped the canvas to the welcome card and hid the folded tile; now sleeping records count as canvas content so the canvas stays mounted.

Formal e2e (Cucumber/Playwright video on a pu box) + the switcher column / louder-alerts / idle-nudge surfaces are still pending per the PR description.

srid:
Discarding in favor of a re-plan. This built sleeping terminals as a parallel silo (its own component, dock section, and an excluded list) — but the right model is that **sleeping is a STATE of a first-class tile**, identical to a live terminal except its PTY is released. The real blocker is that there's no first-class `Tile` concept today (it's fused into Terminal/Canvas), so there was nowhere clean to put a dormant tile. Next: a structural (lowy/hickey) pass to extract the `Tile` abstraction, a preliminary PR introducing it (Terminal as the only tile), then sleeping as a tile state on top. The backend commits here (persisted store, snapshot, wake-respawn, migration) are preserved on the branch for reuse.

PR #1435, opened 2026-06-19
First-class Tile registry: split tile presence from terminal liveness

srid:
**Kolu's canvas calls itself "domain-agnostic" but a tile *is* a live terminal** — `tileIds === terminalIds`, a tile's position is a field on the terminal, the active tile is a terminal id, and dock rows rank terminal ids. So a tile with no live terminal has nowhere to exist. This PR introduces a first-class **Tile registry** that separates *tile presence* from *terminal liveness*, with **zero behavior change** — it's the load-bearing groundwork for sleeping terminals (and, later, planned terminals).

This is **PR 1 of 4** from the plan of record (Atlas note #1432). It fixes the structural wart first; everything after is a content variant, a body, or a signal.

### The seam

```
Before                                  After
──────                                  ─────
terminalIds() ─▶ canvas <For>           terminalIds() ─▶ useTileStore
            └──▶ dock rank                            ├─ tiles() + TileContent union
            └──▶ mode() count                         ├─ getLayout / setLayout
activeId (a terminal id) ─▶ highlight                 └─ activeId / mruOrder (re-exposed)
canvasLayout on the terminal                                   │
                                          canvas · minimap · dock · switcher · mode ──┘
                                          (read the ONE registry; dispatch on content.kind)
```

| The **Tile** owns | The **Terminal** owns |
|---|---|
| identity (`TileId`), the `TileContent` union | PTY · xterm · agent · attach stream |
| canvas-layout access, active/selection, MRU, count | repo/branch identity, display info, the live body |

`TileContent` is `{ kind: "terminal", terminalId }` — the **only** variant today. The single canvas `<For>` now dispatches on `content.kind` via a stable-boolean `<Switch>`/`<Match>` (never `match(freshObject)`, which would re-create the tile subtree every tick — the #989 trap). Adding a kind is one new union arm plus one new `<Match>`; **the silo becomes unrepresentable.**

### Why it's genuinely zero behavior change

- **`TileId === TerminalId`** — tiles and terminals share one uuid id space (a slept terminal keeps its id), so there's no synthetic id to invent and no ripple. What varies is *content*, not identity.
- **`tileIds()` re-exposes the stabilized `terminalIds()` memo verbatim**, so the `sameTerminalIdOrder` reference-stability keystone (#1425) is *inherited*, not re-implemented — a metadata-only tick still doesn't notify the canvas/dock.
[...]

PR #1425, opened 2026-06-19
Stabilize the terminalIds memo to skip no-op display re-derivation

srid:
The `terminalIds` memo rebuilt a fresh array on every run, so **a single terminal's metadata updating once a second re-ran the entire display derivation for *all* terminals** — `buildTerminalDisplayInfos` (color assignment, collision keys, sub-counts) plus every tile's `Show` gate — for a working set that never changed. This gates that memo so an unchanged top-level id set keeps its prior reference, and the derivation re-runs only on a real add / remove / reorder.

This is the **keystone reactivity item** of the [performance map](https://kolu.dev/atlas/performance.html) — the one the survey flagged as unblocking the rest of the cluster.

### The mechanism

```
metadata change (git/pr/agent on one terminal)
   │
   ▼
terminalIds memo  ── re-runs (cheap: keys().filter) ──┐
   │                                                  │  equals: sameTerminalIdOrder
   │  before: new array ref every run                 │  after: same ref when set unchanged
   ▼                                                  ▼
displayInfos / every tile Show gate            (no notify → no re-derive)
   ── before: re-derived ALL terminals          ── after: skipped
```

The accessor still re-runs on each metadata change; what it *no longer* does is **notify** downstream when the id set is identical. The comparator (`sameTerminalIdOrder`) is a pure, domain-specific bounded-algorithm leaf in `terminalDisplay.ts`; the reactivity itself is Solid's built-in `equals` option, not a hand-rolled mechanism.

### Why it's safe

Order stays significant (it drives sidebar position labels), so a reorder invalidates. Every `terminalIds()` consumer reads it for its *contents* (`.length`/`.filter`/`.map`); the one change-detector (`useTerminalAlerts`) tracks `getMetadata(id)` directly inside its `on(...)` scope, so agent-finished alerts still fire on the metadata signals regardless of reference stability — _verified before shipping_.

A re-run-count regression test (solid-js core via `createRoot`) pins the invariant: an unchanged id set re-runs the memo but **not** the downstream derivation; a real add/remove/reorder propagates. Banking it with a deterministic test meant it didn't need the live trace the perf map's coverage-gaps still flag for the remaining reactivity items.

> Internal perf refactor — no user-visible behavior change (the UI renders identically; it just recomputes less), so no changelog entry. The perf map is updated to bank this keystone.

_Generated by [`/be`](https://github.com/srid/agency) on Claude Code (model `claude-opus-4-8`)._

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #989, opened 2026-05-27
fix(canvas): kill the maximized-mode switch lag

srid:
**Switching the active terminal in maximized mode used to stall for 200-500ms** while the xterm contents disappeared and reappeared. The active tile lived in a separate `<Show keyed>` branch outside the pan/zoom transform, so every `activeId` change moved tiles between branches — forcing a full xterm.js Terminal remount each time (`document.fonts.load` wait, fresh `XTerm` construction, ~10 addons, stream re-attach, full server `screenState` replay).

This PR collapses the two render branches into one stable `<For>`. The active maximized tile auto-promotes via the existing `CanvasTile.mode` prop — _the xterm DOM node and its xterm.js `Terminal` instance survive across switches_, matching how non-maximized canvas mode has always behaved. Switching becomes a pure CSS class flip with no instance lifecycle traversal.

### Before vs. after

```
Before                                          After
──────                                          ─────
canvas-container                                canvas-container
├─ canvas-transform (pan/zoom wrapper)          ├─ <For> tileIds (every render, stable)
│  └─ <For> tileIds.filter(≠ active)            │  ├─ tile A  (mode="tiled"/"covered"/"maximized")
│     ├─ tile A  (mounts/unmounts on switch)    │  ├─ tile B
│     └─ tile C                                 │  └─ tile C
└─ <Show keyed=activeId>                        │
   └─ tile B  (REMOUNTS on activeId)            tile transform composes pan+zoom+drag
                                                inline; data-viewport on canvas-container
tile.transform = wrapper-applied                carries the wrapper-formula string for tests
```

The wrapper had to exist because the maximized tile needed `inset-0` against an untransformed ancestor (a transformed parent becomes the containing block). Pushing pan/zoom into each tile's own transform via a new `tileTransformCSS(layoutX, layoutY, panX, panY, zoom, dragX, dragY)` removes that constraint entirely.

### Refinements during review

| Pass | Finding | Fix |
| --- | --- | --- |
| Hickey | Ghost `canvas-transform` div fragmented test observability from tile positioning | Dropped the phantom div; surface viewport state via `data-viewport` on `canvas-container` |
| Lowy | Per-tile transform formula duplicated against `canvasTransformCSS` | Extracted `tileTransformCSS` into `coordinates.ts` next to its sibling |
| Police | A11y regression — `aria-hidden` alone didn't block Tab into covered xterms | Switched to `inert` on covered tiles (Tab + pointer + AT in one) |
[...]

PR #1441, opened 2026-06-20
Sleeping terminals: sleep a terminal, wake it in place (PR 2/4)

srid:
**Leave an agent blocked for days, *sleep* its terminal — PTY, WebGL context, and agent all released, exactly as if you'd closed it — and *wake* it later in place with the agent resumed.** This is **Phase 2** of the [sleeping-terminals plan](https://github.com/juspay/kolu/blob/master/docs/atlas/src/content/atlas/sleeping-terminals.mdx) (the user-facing core), built on the first-class Tile registry that landed in [#1435](https://github.com/juspay/kolu/pull/1435).

The whole point of the re-plan was *where a dormant tile lives*. A first cut built sleeping as a **parallel silo** (its own store, its own component, a second `<For>`, a separate dock section). Here sleeping is a **`TileContent` variant on the one registry**, so it inherits drag, resize, focus, active-selection, dock ordering, and persistence **for free** — the silo becomes unrepresentable.

### What it does

- **Sleep** — the ☾ button in a tile's title bar, or the palette's *Sleep terminal*. Snapshots the tree, releases the live terminal. *Persist-before-kill*: a crash mid-sleep loses nothing.
- **Wake** — respawns the tree through the **existing `handleRestoreSession`** (so a woken terminal is indistinguishable from a session-restored one) and resumes the agent.
- **A real tile** — the sleeping tile renders through the **same `CanvasTile` shell** (moonlit theme, dormant body, Wake/discard), is **draggable + resizable**, can be the **active** tile, and gets a **dock row grouped under its repo** (☾ pip, dimmed) — never a separate section.
- **Durable** — sleeping tiles persist to a `sleepingTerminals` store disjoint from `session`, surviving a reload *and* a restart, **rehydrated as sleeping** (never auto-woken), restored to their saved position.

### The seam

| Layer | What changed |
|---|---|
| **server / common** | `sleepingTerminals` cell + `SleepingTerminalSchema`; `sleep`/`wake`/`setSleepingLayout` RPCs; `snapshotTerminal`; the `1.26 → 1.27` migration. The record is keyed by the **original top-terminal id** (not a fresh UUID), so `TileId === TerminalId` holds and position/MRU/selection carry over the moment it sleeps. |
| **registry** (`useTileStore`) | `tileIds()` is now a deduped union of live + sleeping ids (live wins) gated on the reused `sameTerminalIdOrder` — *the #1425 reference-stability keystone survives the merge*. `contentOf`/`getLayout`/`setLayout` dispatch on kind. |
| **orchestration** (`useSleepActions`) | sleep = `server.sleep` → `handleKillWithSubs`; wake = `handleRestoreSession` (now a singleton) → `server.wake`; a `wakingTiles` leaf is the optimistic mid-wake hide both the registry and the orchestration read (no cycle). |
| **canvas / dock** | the `<For>` adds a second stable-boolean `<Match>` (avoids the #989 remount trap) to a sleeping tile through the shared `CanvasTile`; the dock synthesizes a sleeping row's `{meta, info}` so it ranks/groups/renders through the same pipeline. |

- **Fail-fast / no fallbacks** — `setLayout` surfaces an absent-tile id loudly (`console.error`) rather than dropping the write; no override knobs.
- **Electricity boundaries** — sleeping is in-app domain state (`tile/`, `terminal/`), not a `@kolu/*` package — the verdict the note records. The pure `sameTerminalIdOrder` comparator and the layout writers are kept as **leaves** so the registry doesn't drag in heavy modules (and the import cycle Phase 1 broke stays broken).
- **Reuse the source of truth** — wake replays through the *existing* `handleRestoreSession` rather than a copied loop; the dock reuses `buildTerminalDisplayInfos`; the registry reuses the #1425 comparator.

### Scope (deferred per the plan)

- **PR 3** swaps the dormant body's placeholder for a **frozen screenshot**.
- A dock-row **Wake button** and full active-selection **parity** are PR 4 polish; today the dock row focuses on click and the canvas tile is the prominent Wake surface.

### Verification

`just check` + `just fmt` green. Unit: **client 415, server 99**. E2e (run locally): the new `sleeping-terminals.feature` (sleep→wake + restore-as-sleeping reload) is green, and `dock` + `session-restore` + `canvas` (67 scenarios) regress clean.

### Try it locally

\`\`\`sh
nix run github:juspay/kolu/feat/tile-sleeping-state
\`\`\`

_Generated by [\`/be\`](https://github.com/srid/agency) on Claude Code (model \`claude-opus-4-8\`)._


srid:

⚠️ **Apply incomplete** — 3 agreed fix(es) not cleanly applied after 2 round(s) · lowy + hickey · base `8a613be1593f`

Independent findings: lowy=9, hickey=6

### Applied (8)
- `lowy-1` wake and discard are the same server verb wearing two names — the receptacle exposes a duplicated wire, not a stable contract — commit `7d55b20ae`
- `lowy-2` The strip-live-fields projection (pr/agent/foreground → persisted) is now sprayed across three modules — commit `1083a5e3b`
- `lowy-3` 'find the top terminal' (parentId-less) is reimplemented at five sites instead of behind the record — commit `86f0f801e`
- `lowy-4` The sleeping tile's label/title text is derived twice with diverging rules — commit `fa0cd9113`
- `lowy-6` Three cells now hand-roll the same JSON.stringify equals — a candidate receptacle the surface layer doesn't yet expose — commit `8dfb66335`
- `lowy-7` MOON/MOON_BG colours and the ☾ glyph are partially centralized — palette leaks remain — commit `122c32763`
- `lowy-8` rankDockRows gained an isSleeping default-param escape hatch — a mode flag complecting two call patterns — commit `d9aa84437`
- `hickey-4` Active-tile content-kind dispatch for close/sleep is duplicated across App and the canvas — commit `af683998a`

### Apply incomplete — needs reconcile (3)
- `hickey-1` "Top terminal of a sleeping record" is re-derived at four sites — and they disagree — not confirmed applied (absent from apply output)
- `hickey-2` Sleeping-tile display label computed twice from the same record — not confirmed applied (absent from apply output)
- `hickey-3` `basename` reimplemented three times — not confirmed applied (absent from apply output)

### Agreed — no change (4)
- `lowy-5` wakingTiles leaf is a correct receptacle for the mid-wake hide — confirming, not flagging (packages/client/src/tile/wakingTiles.ts:1-34)
- `lowy-9` focusedId now silently returns null for an active sleeping tile — a guard that encodes the live/sleeping split deep in the terminal store (packages/client/src/terminal/useTerminalStore.ts:45-53 (the `if (!metadata.terminalIds().includes(parentId)) return null` guard))
- `hickey-5` Sleeping-row {meta,info} projection derived in two layers of the dock (packages/client/src/canvas/dock/useDockOrder.ts:38-49, packages/client/src/canvas/dock/RowPips.tsx:71-79)
- `hickey-6` `focusedId` sleeping-narrowing duplicates the live-membership test that `tileIds` already centralizes (packages/client/src/terminal/useTerminalStore.ts:45-51, packages/client/src/tile/useTileStore.ts:50-56 & 85-91)

### Reconciliation (be-review)

The 3 "apply-incomplete" findings are **hickey duplicates of already-applied lowy fixes** — the substance landed; the apply agent simply didn't emit separate entries for the overlap:

- `hickey-1` (top-terminal re-derived at N sites) → satisfied by **`lowy-3`** — the `record.id`-anchored `topTerminal` accessor (`kolu-common/surface`, commit `86f0f801e`).
- `hickey-2` (sleeping label computed twice) → satisfied by **`lowy-4`** — the `sleepingTileLabel` helper (commit `fa0cd9113`).
- `hickey-3` (`basename` reimplemented thrice) → satisfied by **`lowy-4`** — the shared `basename` in `kolu-common/path` (commit `fa0cd9113`).

Verified: no leftover top-terminal / label / `basename` reimplementations remain in the changed code; `just check` + `just fmt` clean; unit suite green (client 415, server 99). No outstanding fix to apply.

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 round(s) · codex reviewed at `xhigh` reasoning effort · base `8a613be1593f`

### Round 1

**codex** — approved: `false`

The feature is not ready as-is. The desktop happy path is sketched out, but the branch has several correctness holes around wake concurrency, stale active selection, non-idempotent persistence, and non-desktop rendering. The new tests cover only the simplest count transitions, so they miss data-loss and broken-surface cases.

Findings:
- `F1` · blocking · open — `wake` treats `handleRestoreSession` returning as success, but `handleRestoreSession` silently returns when `isRestoring()` is already true (`useSessionRestore.ts:157`). If two sleeping records are woken quickly, or a saved-session restore is in progress, the later wake creates no terminal and still proceeds to `dropSleeping`, deleting the durable record. (packages/client/src/terminal/useSleepActions.ts:62)
- `F2` · major · open — The touch layouts still render every `orderedIds` entry through `TerminalContent`, while `useDockOrder` now includes sleeping tile ids. On phone/compact layouts, a sleeping tile will mount `TerminalContent`/`Terminal`, which attaches/resizes/sends input against a non-live terminal id instead of rendering the sleeping body and Wake action. (packages/client/src/App.tsx:266)
- `F3` · major · open — Discarding a sleeping tile drops the server record but never updates local active selection. If the discarded sleeping tile was active, `activeId` remains pointed at a removed tile; in maximized posture that leaves all remaining tiles in `covered` mode, and commands continue targeting a stale id. The sleep path has the related inverse problem: it kills through the live-terminal close path, which auto-switches away from the newly sleeping tile instead of preserving tile focus. (packages/client/src/terminal/useSleepActions.ts:91)
- `F4` · major · open — `sleepTerminal` blindly appends a record. A double click or repeated command before the live kill finishes persists duplicate sleeping records with the same id. The client then maps those records directly into duplicate `tileIds`, while `contentOf` only finds the first matching record, leaving persistent state and layout behavior inconsistent. (packages/server/src/sleepingTerminals.ts:54)
- `F5` · major · open — Sleep reports success after `crud.handleKillWithSubs`, but that path calls `handleKill`, which catches all kill errors and continues. A failed kill can therefore leave the PTY/agent live while the UI says the terminal is asleep and the sleeping record has already been written. (packages/client/src/terminal/useSleepActions.ts:40)
- `F6` · minor · open — The active-terminal command palette section is gated only on `activeId() !== null`. Once a sleeping tile is selected, live-only commands such as split, theme, screenshot, export, copy, and Sleep itself are still offered; several global shortcuts in `input/actions.ts` similarly dispatch live-terminal verbs using `activeId`. These become no-ops or terminal-not-found RPCs for sleeping tiles. (packages/client/src/commands.tsx:231)
- `F7` · minor · open — The loading gate waits for terminal list and saved-session state, but not for the new `sleepingTerminals` subscription. Since `sleepingTerminals()` returns `[]` until its first yield, a cold reload with only sleeping tiles can briefly resolve as an empty workspace and expose EmptyState/restore/create affordances before the sleeping snapshot arrives. (packages/client/src/terminal/useSessionRestore.ts:341)
- `F8` · minor · open — `SleepingTerminalSchema` does not enforce the invariants the rest of the code assumes: `id` is a plain string instead of `TerminalIdSchema`, `terminals` may be empty, and no terminal is required to match the record id. `topTerminal` then falls back to the first entry, which masks corrupt persisted state instead of failing fast. (packages/common/src/surface.ts:400)
- `F9` · minor · open — The workspace switcher is shown when `tileCount() > 0`, but `workspaceEntries` still builds rows with `store.getDisplayInfo`, which only knows live terminals. A sleeping-only workspace can open an empty Search Workspaces view even though the dock shows sleeping rows. (packages/client/src/App.tsx:115)

**claude** — Reviewed CODEX's verdict on the sleeping-terminals feature and agreed with all nine findings — each was a genuine correctness or fail-fast violation, not a style nit. Fixed all of them in the working tree across 14 files, then committed. F1 (blocking): handleRestoreSession now THROWS on the busy (isRestoring) path so wake can't dropSleeping a durable record after a no-op restore; wake surfaces the busy error and never drops on failure. F2: the touch tile body now dispatches on tile content (SleepingTileBody for sleeping ids, TerminalContent only for live), so a sleeping id no longer mounts Terminal against a non-live id. F3: sleep re-asserts the slept tile as active after the kill (removeAndAutoSwitch had switched away); discard re-points active at the clamped neighbour (or null) when the discarded tile was active. F4: sleepTerminal is now idempotent by id (replace in place, no duplicate records). F5: sleep uses a new strict kill path (handleKillWithSubsStrict) that propagates non-NOT_FOUND failures and rolls the record back on a real failure. F6: the active-terminal command section and the split/theme/screenshot shortcut handlers gate on live metadata; an active sleeping tile gets a Wake/Discard section. F7: the loading gate also waits on sleepingTerminalsSub.pending() when the live list is empty (+ regression test). F8: SleepingTerminalSchema uses TerminalIdSchema, requires non-empty terminals, and refines that the root id exists; topTerminal drops its fallback and throws on a violated invariant. F9: useTileStore exposes a tile-aware getDisplayInfo (synthesized for sleeping tiles) so the workspace switcher lists a sleeping-only workspace. All client (416), common (17), server (99) unit tests pass; typecheck + biome lint clean. The arivu daemon socket test fails only under parallel runs and passes in isolation — unrelated to these changes (that package was untouched).

- `F1` **fixed** — Agreed — real data-loss bug. handleRestoreSession (useSessionRestore.ts:157) now `throw new Error("restore already in progress")` on the isRestoring() busy path instead of silently returning. wake (useSleepActions.ts) catches the throw, unmarks the waking tile, surfaces "Another wake/restore is in progress", and returns WITHOUT dropSleeping — so the durable record is never deleted when no terminal was actually created. The restore card's own isRestoring() gate already disables its button, so only the wake path can reach the throw.
- `F2` **fixed** — Agreed. renderMobileTileBody (App.tsx) now mirrors TerminalCanvas's TileContent <Switch>: terminalContent -> TerminalContent, sleepingContent -> SleepingTileBody (record + onWake). orderedIds includes sleeping ids (via useDockOrder/tileIds), so the old unconditional TerminalContent mounted Terminal and attached/resized/sent input against a non-live id on phone/compact. Dispatch keys off the stable `kind` boolean (no #989 remount).
- `F3` **fixed** — Agreed on both halves. Discard (useSleepActions.discard): snapshots tile order before the drop and, if the discarded tile was active, re-points active at the clamped neighbour (or null) — same policy as removeAndAutoSwitch — so activeId never dangles at a removed tile (which left all tiles `covered` in maximized posture). Sleep (useSleepActions.sleep): re-asserts the slept tile as active via setActiveSilently after the kill, because handleKillWithSubs->removeAndAutoSwitch switched focus away; the slept tile keeps its id and becomes a sleeping tile, so focus should stay on it. Re-assert is unconditional on wasActive (not gated on contentOf resolving to sleeping) because the live-list/sleeping-cell pushes are async — the id round-trips through sleep and is always a valid tile.
- `F4` **fixed** — Agreed. sleepTerminal (server/sleepingTerminals.ts) now finds an existing record by id and replaces it in place (re-capturing latest metadata) rather than blindly appending. A double-click or repeated terminal.sleep before the live kill landed previously persisted two records of one id -> duplicate tileIds on the client while contentOf only found the first.
- `F5` **fixed** — Agreed. Added handleKillStrict/handleKillWithSubsStrict to useTerminalCrud: a kill error whose oRPC code !== "NOT_FOUND" rethrows (NOT_FOUND = already gone = success). sleep uses the strict path; on a real teardown failure it rolls back the just-written record (dropSleeping) and rethrows so the toast shows the failure — the user keeps an accurate live tile instead of a phantom "asleep" record over a still-live PTY/agent.
- `F6` **fixed** — Agreed. commands.tsx: the active-terminal section is gated on the active tile's content kind === "terminal" (computed once via tileStore.contentOf), not just activeId() !== null; an active sleeping tile gets a Wake/Discard section instead. The shortcut handlers that dispatched live-terminal verbs off raw activeId are gated on live metadata: toggleSubPanel/createSubTerminal/next-prevSubTab (actions.ts, via ctx.activeMeta()), handleShuffleTheme (useThemeManager.ts), and handleScreenshotTerminal (useActionContext.ts) — all now no-op for a sleeping active tile instead of firing terminal-not-found RPCs.
- `F7` **fixed** — Agreed. useSessionRestore.isLoading now also waits on sleepingTerminalsSub.pending() (OR'd with savedSessionSub.pending()) when terminalIds() is empty. A sleeping tile counts toward tileCount() and keeps the canvas mounted, so a cold reload of a sleeping-only workspace would otherwise flash EmptyState (restore/create affordances) before the sleeping snapshot arrived. Added a regression test ("keeps loading on an empty list until the sleeping cell reports") and wired sleepingPending into the test's wire mock.
- `F8` **fixed** — Agreed — fail-fast. SleepingTerminalSchema (common/surface.ts) now: id uses TerminalIdSchema (not bare string), terminals is .min(1), and a .refine enforces that a terminal with id === record.id exists. topTerminal drops its `?? record.terminals[0]` fallback — it returns a guaranteed SavedTerminal and throws loudly if the invariant is somehow violated in-process, instead of masking corrupt state with the wrong entry. Updated the three callers (sleepingDockRow, SleepingTileBody, useTileStore) to drop now-dead optional handling; sleepingTileLabel loses its "asleep" empty-record fallback.
- `F9` **fixed** — Agreed. Added a tile-aware getDisplayInfo to useTileStore: live store.getDisplayInfo for terminal tiles, synthesized sleepingDockRowData(record).info for sleeping ones. App.tsx's workspaceEntries now uses tileStore.getDisplayInfo instead of store.getDisplayInfo, so buildWorkspaceEntries no longer silently drops every dormant tile — a sleeping-only workspace lists rows in the "Search workspaces" switcher (matching what the dock paints) instead of opening empty.

commit: `6ba764a96d13724c2e998337da17d75c38a166fa`

### Round 2

**codex** — approved: `false`

CLAUDE resolved the nine findings from the prior round. The fixes are present and technically address the original failure modes. I found one new regression introduced by the F5 rollback path, so this is not approved yet.

Findings:
- `F1` · blocking · resolved — Wake could delete a sleeping record after a no-op concurrent restore. (packages/client/src/terminal/useSessionRestore.ts:165)
- `F2` · major · resolved — Touch layouts mounted TerminalContent for sleeping ids. (packages/client/src/App.tsx:284)
- `F3` · major · resolved — Sleep/discard could leave active selection pointing at the wrong or removed tile. (packages/client/src/terminal/useSleepActions.ts:74)
- `F4` · major · resolved — Repeated sleep calls could append duplicate sleeping records with the same id. (packages/server/src/sleepingTerminals.ts:60)
- `F5` · major · resolved — Sleep used the non-strict kill path, so real teardown failures were swallowed while the UI claimed the terminal was asleep. (packages/client/src/terminal/useTerminalCrud.ts:195)
- `F6` · minor · resolved — Live-terminal commands and shortcuts remained available for active sleeping tiles. (packages/client/src/commands.tsx:249)
- `F7` · minor · resolved — Cold reload of a sleeping-only workspace could flash EmptyState before the sleeping cell yielded. (packages/client/src/terminal/useSessionRestore.ts:355)
- `F8` · minor · resolved — SleepingTerminalSchema allowed malformed records and topTerminal masked them with a fallback. (packages/common/src/surface.ts:400)
- `F9` · minor · resolved — Workspace search used live-only display info and dropped sleeping-only workspaces. (packages/client/src/tile/useTileStore.ts:118)
- `F10` · major · open — New regression: on strict kill failure, sleep always rolls back the sleeping record. But handleKillWithSubsStrict may already have successfully killed one or more subterminals before a later subterminal or parent kill fails. Dropping the just-written sleeping snapshot in that partial-failure case deletes the only durable copy of the already-killed split tree. (packages/client/src/terminal/useSleepActions.ts:52)

_codex on the rebuttal:_ No disputes were raised; CLAUDE accepted F1-F9. I verified each claimed fix and marked those findings resolved. F10 is new from this round: it is introduced by the rollback added for F5, and it needs a targeted follow-up because failed sleep must not delete the recovery snapshot after a partial strict kill.

**claude** — Round 2 of the codex debate on the sleeping-terminals branch. Codex re-verified and marked all nine prior findings (F1-F9) resolved; the only open finding was F10, a new regression it identified in the F5 rollback path. I agreed with F10 and fixed it.

F10 root cause: the sleep flow writes a durable sleeping snapshot, then tears the live tree down via handleKillWithSubsStrict, which kills sub-terminals one at a time and then the parent. A non-NOT_FOUND kill failure can land AFTER an earlier kill already tore a terminal down. The old `sleep` catch unconditionally called dropSleeping, deleting the just-written record — the only durable copy of the already-killed pieces — leaving the user with a partially-destroyed tree and no recovery snapshot.

Fix (packages/client/src/terminal/useTerminalCrud.ts + useSleepActions.ts):
- handleKillWithSubsStrict now tracks whether any kill succeeded; on a LATER failure it throws a new exported PartialKillError, while a failure on the very first kill (nothing destroyed) still rethrows plain.
- sleep rolls the sleeping record back ONLY on a plain error (tree intact and live); on PartialKillError it KEEPS the record so the user can wake the tree back, and surfaces the incomplete sleep loudly.
- the rollback's own dropSleeping failure is now surfaced via toast instead of being swallowed with .catch(() => {}), per codex's explicit note.
- Added packages/client/src/terminal/useTerminalCrud.test.ts (3 cases: first-kill-fail → plain error; later-kill-fail → PartialKillError; NOT_FOUND → counts as success and continues).

Verification: full client suite 419 passed (416 prior + 3 new), pnpm typecheck clean across all packages, biome clean on all touched files. Also applied biome organize-imports to useTerminalCrud.ts (a pre-existing one-line nit on the refuseIfWarming import that the touched file would otherwise trip the lint gate on).

- `F1` **fixed** — Resolved in round 1 (commit 6ba764a). Codex re-verified and marked resolved this round. No further action needed.
- `F2` **fixed** — Resolved in round 1. Codex re-verified and marked resolved. No further action.
- `F3` **fixed** — Resolved in round 1. Codex re-verified and marked resolved. No further action.
- `F4` **fixed** — Resolved in round 1. Codex re-verified and marked resolved. No further action.
- `F5` **fixed** — Resolved in round 1 for the original swallowed-failure issue. Codex re-verified the strict-kill propagation. The rollback it ADDED for F5 is what F10 critiques; addressed there.
- `F6` **fixed** — Resolved in round 1. Codex re-verified and marked resolved. No further action.
- `F7` **fixed** — Resolved in round 1. Codex re-verified and marked resolved. No further action.
- `F8` **fixed** — Resolved in round 1. Codex re-verified and marked resolved. No further action.
- `F9` **fixed** — Resolved in round 1. Codex re-verified and marked resolved. No further action.
- `F10` **fixed** — Agreed — genuine data loss. handleKillWithSubsStrict kills subs then parent sequentially, so a non-NOT_FOUND failure can occur after an earlier kill already tore down a terminal. The unconditional dropSleeping in sleep's catch then deleted the only durable copy of the already-killed pieces. Fix: handleKillWithSubsStrict (useTerminalCrud.ts:248) now tracks killedAny and throws PartialKillError on a later failure (plain rethrow on the first kill = nothing destroyed). sleep (useSleepActions.ts:60) rolls back ONLY on a plain error; on PartialKillError it keeps the record and surfaces the incomplete sleep. Also surfaced the rollback's own dropSleeping failure via toast instead of swallowing it, per codex's suggestion. Added useTerminalCrud.test.ts with 3 regression cases.

commit: `4f202801eab2ab0c2ccf190d5aeb8eba9958a34a`

### Round 3

**codex** — approved: `true`

All findings from the prior rounds are now resolved. I verified the current F10 fix: strict tree teardown now distinguishes a first-failure rollback case from a partial-teardown case, preserves the sleeping snapshot when any kill already succeeded, and surfaces rollback-drop failures instead of swallowing them. I did not find a new regression in this round's changes.

Findings:
- `F1` · blocking · resolved — Wake could delete a sleeping record after a no-op concurrent restore. (packages/client/src/terminal/useSessionRestore.ts:165)
- `F2` · major · resolved — Touch layouts mounted TerminalContent for sleeping ids. (packages/client/src/App.tsx:284)
- `F3` · major · resolved — Sleep/discard could leave active selection pointing at the wrong or removed tile. (packages/client/src/terminal/useSleepActions.ts:86)
- `F4` · major · resolved — Repeated sleep calls could append duplicate sleeping records with the same id. (packages/server/src/sleepingTerminals.ts:60)
- `F5` · major · resolved — Sleep used the non-strict kill path, so real teardown failures were swallowed while the UI claimed the terminal was asleep. (packages/client/src/terminal/useTerminalCrud.ts:212)
- `F6` · minor · resolved — Live-terminal commands and shortcuts remained available for active sleeping tiles. (packages/client/src/commands.tsx:249)
- `F7` · minor · resolved — Cold reload of a sleeping-only workspace could flash EmptyState before the sleeping cell yielded. (packages/client/src/terminal/useSessionRestore.ts:355)
- `F8` · minor · resolved — SleepingTerminalSchema allowed malformed records and topTerminal masked them with a fallback. (packages/common/src/surface.ts:400)
- `F9` · minor · resolved — Workspace search used live-only display info and dropped sleeping-only workspaces. (packages/client/src/tile/useTileStore.ts:118)
- `F10` · major · resolved — A sleep rollback after partial strict-kill failure could delete the only durable copy of already-killed split terminals. (packages/client/src/terminal/useTerminalCrud.ts:244)

_codex on the rebuttal:_ No disputes were raised. CLAUDE agreed with F10 and implemented the requested distinction: first kill failure remains a plain error and may roll back, later failure becomes PartialKillError and preserves the snapshot. The rollback drop failure is also surfaced. I accept the fix and mark F10 resolved.

srid:

**All clear** — sleeping terminals (Phase 2), base `8a613be1593f`. Ran `--no-elegance` (the `/simplify` track already covered that pass).

| Pass | Result |
|---|---|
| **Rules** | ✅ 0 violations — every built-in + project rule checked clean (re-export bridges, thin wrappers, `subscription-use-pending`, `caught-error-must-not-collapse-to-empty`, `toast-must-include-error-message`, `prefer-ts-pattern`, `app-shell-stays-thin` (reactive-primitive budget intact), `dry-rule-of-three`, `invalid-states-unrepresentable`, …). |
| **Fact-check** | ✅ 0 surviving findings — four initial flags each dissolved under scrutiny: `handleRestoreSession` toasts its in-try errors before rethrowing (wake's catch only skips the *pre-try* busy-state throw it raises itself), the `dropSleeping` failure is toasted separately before the kill error rethrows, and discard's pre-RPC snapshot matches the deliberate live-kill `removeAndAutoSwitch` pattern. |
| Elegance | — skipped (`--no-elegance`; `/simplify` ran). |

No commits — nothing to fix.

srid:
## 🐛 Data-loss fix: sleeping a terminal could lose it forever

**The bug** (caught in dev, off-PR): sleeping a terminal made it vanish — no sleeping tile, no dock row, gone on reload. It hit *plain terminals too*, not just agents.

**Root cause** — a poisoned cell, not a sleep bug. A single **orphan** sleeping record (root id matches no terminal in its tree — a UUID-keyed record from the abandoned first cut) made codex F8's `.refine` reject the **entire** `sleepingTerminals` array. The client fell back to the default (empty) cell, so *every* sleep — including freshly-slept good ones — disappeared. There was no migration for the UUID→terminal-id format change, and nothing read the cell tolerantly.

**The fix** — read tolerantly, all-or-nothing → drop-the-bad-one:

| | before | after |
|---|---|---|
| schema | `.refine(root must exist)` — fatal per array | SHAPE only |
| server read | raw `store.get` | `getSleepingTerminals` filters orphans; the cell serves through it |
| client registry | rendered every record | filters orphans (never a tile, never reaches `topTerminal`) |

**No migration.** The orphan format was never released, so there's no production data to migrate — a migration would be dead code. Tolerance is the durable fix and guards any future corrupt record. _(The current code can't even produce an orphan: `record.id` is the snapshotted terminal's own id, so records are rooted by construction.)_

Also simplified per scope: **sleep captures only the top terminal** (splits aren't carried across sleep/wake), so records are always single-terminal — shedding the tree-complexity that was extra surface for this very bug.

### Why the original e2e missed it
A fresh e2e never has a pre-existing orphan, and `test__set` validated it away. The new scenario seeds an orphan first — and the seed POST itself **400s on the unfixed code** (the `.refine`), so it reproduces the bug. Plus a client unit test (orphan filtered) and a `Before`-hook reset so sleeping records can't leak between scenarios.

### Guarding the class
New code-police rule **`persisted-schema-stays-tolerant`**: tightening a *persisted* schema needs a backward-compat path — a migration if the old format was released, else graceful read-time filtering. The prior gauntlet's `migration-shape-guard` marked this "N/A" instead of flagging it.

### Verified live (chrome-devtools, dev-auto, fresh state)
create → **sleep → sleeping tile** (not vanished) → **wake → live xterm**; and with an orphan injected alongside a valid record, the valid one **renders** while the orphan is filtered (0 → would-have-been-empty becomes 1). All 7 e2e scenarios green (52 steps); client 424 + server 99 unit green.

srid:
## 🌙 Sleeping tiles are now first-class everywhere in the UI

A sleeping tile is a `TileContent` union variant on the tile registry, but several surfaces still read the **live-only terminal store**, so they silently dropped or mis-rendered sleeping tiles. Starting from the reported minimap gap, I fanned out an audit across **every** tile-representing surface and fixed each gap.

### The minimap (reported)
It read `store.getDisplayInfo` (live-only) and dropped any tile with no info → sleeping tiles were **omitted entirely**. Now reads the registry, and a sleeping rect wears the **same moonlit palette** as its canvas tile (`MOON_BG` fill + `MOON` border + `data-sleeping`) — present *and* visibly dormant, not an idle live rectangle.

### Audit (9 agents over every surface) — gaps found & fixed
| Surface | Gap | Fix |
|---|---|---|
| **Minimap** | sleeping tiles omitted | read registry + moonlit mark |
| **MobileTileView** titlebar | sleeping active tile → bare "kolu" fallback | read registry (synthesizes identity) |
| **⌘⇧] / ⌘⇧[** cycle | skipped sleeping tiles; *broke* the cycle origin when a sleeping tile was active (`indexOf === -1`) | cycle the tile union |
| **Arrange-by-repo** | sleeping tiles bucketed to `undefined` → didn't cluster | widen `placementPolicy` to a structural store type; pass the registry |
| Switcher empty copy | "no **live** terminals match" | "no terminals match" |

### Already correct (confirmed, not changed)
Dock, DockList, WorkspaceGrid cards, `useViewPosture` count gate, CompactTileView, MobileKeyBar. The sweep also confirmed the genuinely **terminal-content** reads (theme, unread/attention, PTY input, screenshot, agent alerts) correctly stay live-only — a sleeping tile has no live content there.

### Verified
Minimap fix verified live (moonlit sleeping rect, count matches canvas exactly). New e2e asserts a sleeping tile appears on the minimap (fails on the old live-only minimap — no rect, no `data-sleeping`). All 8 sleeping e2e scenarios green (58 steps); client 424 unit green; `just check` clean.

srid:
## Closing — superseded by a redesign (#1444)

After taking this PR through the review gauntlet and then a Hickey/Lowy **architecture** review, we concluded the approach here is built one level too high. Sleeping is modelled as a `TileContent` **kind** (a peer of `terminal`) on the first-class Tile registry from #1435 — but **sleeping is a *state* of a terminal, not a kind of canvas content.** The tile layer was added to host a "second kind" that never materializes: the canvas is terminal-only, and non-terminal content (sticky notes, etc.) isn't on the roadmap, so the content-kind abstraction is speculative generality.

The two bug classes this PR's development surfaced are the symptoms, not incidental:
- **Data loss** (sleep → terminal gone): a per-record `.refine` made one legacy record reject the whole `sleepingTerminals` cell. Banked as a code-police rule → **#1443**.
- **The "wrong store" gaps** (minimap / mobile titlebar / `⌘⇧[` cycle / arrange read the live-only store): all downstream of `TileId = TerminalId` and the synthesize-on-read registry.

Both dissolve when the discriminant lives on the **terminal** (`Terminal = active | sleeping`) instead of in a registry projected over it — the field partition in `surface.ts` already expresses that fold, and it makes presence-vs-liveness type-safe.

**The updated plan of record is #1444** (collapse the tile layer; the `Terminal` sum; surgical keep/drop; re-planned phases; unify sleeping with session-restore). Closing this so the redesign is the single source of truth. Nothing is lost — the learnings carry forward in #1444 and #1443.

PR #1442, opened 2026-06-20
Accept video files in drag-and-drop uploads

srid:
**Dropping a video onto a terminal now works.** Drag a screen recording — a `.mov`, `.mp4`, `.webm`, `.m4v`, or `.ogv` — onto a terminal and it uploads, saves server-side, and bracketed-pastes its path into the PTY for the agent to read, exactly like any other dropped file. Previously the drop bounced off the extension allowlist with *"File type not allowed"*: the curated list covered text, code, structured data, docs, and images, but no video.

### The seam

Drag-and-drop is gated by a single shared policy in `kolu-common` (`rejectionFor` → `ALLOWED_UPLOAD_EXTENSIONS` + the size cap), consumed by both the client pre-flight and the server's authoritative write — so there's exactly one place to widen. There is no separate MIME gate on the drop path, so the extension allowlist is the whole story.

Rather than hand-roll a second list of video extensions, the allowlist now **derives its video entries from `preview.ts`'s canonical `VIDEO_EXTENSIONS`** — the same set kolu already uses to play video back in the Code browser:

```
preview.ts  VIDEO_EXTENSIONS  ──┬──▶  Code-browser <video> playback
 (.mp4 .m4v .webm .mov .ogv)    └──▶  upload allowlist (dot stripped)
```

*What you can drop and what you can play back are now the same set, by construction* — they cannot drift.

### Design philosophy

- **Reuse the source of truth** — the upload list imports `VIDEO_EXTENSIONS` instead of duplicating it, the exact "an existing extension table over a duplicated constant" the conventions call for. The `preview.ts` header already warns about the drift that bites when two such lists are kept in step by a prose comment.
- **Fail-fast / no fallbacks** — no new override knob and no size-exemption. The 10 MB per-file cap is *unchanged and shared*: video is allowed, not privileged. A larger recording is still rejected loudly with the same *"too large"* toast.

> **Heads-up:** real screen recordings often exceed 10 MB, so a longer `.mov` can still be rejected by size. That's a deliberate choice for this PR (cap untouched), pinned by a regression test, not an oversight.

Covered by unit tests in `upload.test.ts` (acceptance of each container, the screen-recording filename from the report, the still-enforced size cap, and a drift guard asserting every `VIDEO_EXTENSIONS` member is uploadable). README and the changelog are updated in the same commit.

### Try it locally

```sh
nix run github:juspay/kolu/videoatt
```


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `8a613be1593f`

### Applied (1)
- `hickey-1` JSDoc category list drifted from the array it documents — commit `637d328e4`

### Agreed — no change (4)
- `lowy-1` Video derived from preview.ts, but images/PDF still hand-rolled — asymmetric coupling of the same volatility axis (packages/common/src/upload.ts:71-84)
- `lowy-2` Drop guard reuses the canonical video set — correct prior-encapsulation, no new receptacle (packages/common/src/upload.ts:83)
- `lowy-3` as const dropped — exported type is still readonly; no boundary impact (packages/common/src/upload.ts:17,84)
- `hickey-2` "File extension" has two representations (dot vs no-dot) bridged by per-site transforms (packages/common/src/upload.ts:83)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 1 round(s) · codex reviewed at `xhigh` reasoning effort · base `8a613be1593f`

### Round 1

**codex** — approved: `true`

I reviewed the diff against 8a613be1593fd950541076228df309c4775ea243, the changed files, upload/preview/server consumers, project review rules, and confirmed there are no untracked files. The change is narrowly scoped and consistent with the deliberate decisions: upload now derives video extensions from the existing preview source of truth, the shared client/server rejection path still enforces the 10 MB cap, and tests cover representative video acceptance, mkv rejection, size-cap behavior, and drift against VIDEO_EXTENSIONS.

Findings:
- _(none)_

srid:

✅ **All clear** · base `8a613be1593f` · `--no-elegance` (the simplify track already ran the elegance pass over this tree)

| Pass | Issues | Notes |
| --- | --- | --- |
| Rules | 0 | 27 rules checked (7 built-in + 20 project). `VIDEO_EXTENSIONS` stays a single source of truth (defined in `preview.ts`, consumed by the `upload.ts` spread and the drift-guard test); no dead code; the derivation comment explains the non-obvious `preview.ts`-is-canonical + cap-not-lifted intent. |
| Fact-check | 0 | No logic errors, silent swallowing, wishful thinking, or leaks. The `] as const` → `readonly string[]` change correctly preserves the readonly guarantee (a spread of `.map()` yields `string[]`, so the explicit annotation carries the immutability instead). |
| Elegance | – | Skipped (`--no-elegance`). |

srid:
## Evidence

Dropping a `.mov` (`screen-recording.mov`) onto a terminal now uploads it and bracketed-pastes its saved path into the PTY — previously this drop was rejected with a "File type not allowed" toast.





▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=video-drop.mp4

PR #1446, opened 2026-06-20
perf(markdown): keep both Source/Rendered toggle modes alive

srid:
**The Code tab's Markdown Source ⇄ Rendered toggle rebuilt the entire document on every flip** — `FileView` unmounted the inactive mode and remounted the other, re-running the whole marked → DOMPurify → tree-walks → image-resolve → Shiki → `innerHTML` pipeline. It now keeps both modes alive once opened and flips *visibility*, so toggling is a no-op for the renderer instead of a full re-sanitize.

It started as `/be "stabilize the markdown image resolver reference"` (perf-map item #7) — but that literal ask is a **measured no-op**, so no resolver patch ships. Stabilizing the inline-arrow reference eliminates *zero* `sanitizeHtml` runs, confirmed three independent ways: a runtime reproduction on the repo's real Solid build (fresh arrow vs stable callback ⇒ byte-identical memo counts), `FileView`'s own remount-on-snapshot design, and the Solid compiler (the inline-arrow prop is **static**, never a reactive dependency). The reproduction surfaced the *real* cost — the toggle remount — which is what this fixes.

### The fix landed cleaner than scoped

- **No `render(file)` API change.** A per-slot `heldFile` memo freezes a hidden mode's snapshot and adopts the latest the instant it's shown: a toggle with no edit reuses the same snapshot (no re-render), an edit to the _visible_ mode still re-renders it (reload-on-edit intact), an edit to a _hidden_ mode defers until shown (no double-render).
- **`both()`-gated** — single-form appliances (image / video / iframe) stay on the existing `active()` path, untouched.
- **Comment surfaces decoupled.** Two kept-alive `CommentTextSurface`s would have contended for the single global `kolu-comment` CSS Custom Highlight (`applyHighlights` _replaces_ it per call). Each overlay instance now owns a **per-instance** highlight name + style — removing a latent single-surface fragility.

### The win, measured

| | Before | After |
| --- | --- | --- |
| Render pipeline per toggle | **1×** | **0×** |
| 50-image doc, 3 round-trips (reproduction, real Solid build) | 150 image-resolutions + 3 pipelines | **0** |
| Rendered preview element across a round-trip (real app, e2e) | remounted | **preserved** |

### Review hardening (the gauntlet earned its keep)

Keeping *both* comment surfaces mounted at once exposed a real bug the lens/codex review caught and fixed:

- **Cross-surface comment ownership** (codex, major) — a path-only comment filter fed each of the two now-live overlays the *other* surface's comments, racing the shared scroll-request flow. Fixed by making the overlay **surface-aware** (filter by `comment.surface`, gate the scroll request on `surface`), plus a load-time **`backfillSurface`** that routes legacy surface-less Markdown comments to `source` (the only commentable surface before the rendered preview existed) — so every comment has exactly one owner. Covered by a new unit test + a regression e2e (*"Tray jump returns to the Source Markdown surface"*).
- Per-instance highlight name derived from `createUniqueId` (lens); `aria-hidden` on the inactive slot to match `RightPanel` (lens); the `visited` latch simplified to one memo (simplify).

### Tests

New e2e (*"Toggling Source and Rendered keeps the rendered preview alive"*): a marker on the rendered preview survives the round-trip, which a remount would erase. Full `code-tab.feature`: **115/115**; comment unit tests **9/9**. CI green on both platforms.

Full write-up + reproduction: [`docs/perf-investigations/markdown-image-resolver-and-toggle.md`](https://github.com/juspay/kolu/blob/master/docs/perf-investigations/markdown-image-resolver-and-toggle.md). Closes perf-map item #7.


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `6f1fbbae42e6`

### Applied (3)
- `lowy-2` Highlight-name allocation seam split between two modules — commit `3a90f4123`
- `hickey-4` highlightSeq early-return makes the monotonic counter non-monotonic-per-mount — commit `79e19496a`
- `hickey-5` Test assertion duplicates the name-prefix convention by string-matching 'kolu-comment*' — (uncommitted)

### Agreed — no change (8)
- `lowy-1` Keep-alive receptacle now duplicated across RightPanel and FileView (packages/solid-fileview/src/FileView.tsx:138-166)
- `lowy-3` `KeepAliveMode.visited` and `heldFile` track the same show-edge twice (packages/solid-fileview/src/FileView.tsx:152-158)
- `lowy-4` FileView host owns no boundary for the toggle-keep-alive volatility vs. single-form path (packages/solid-fileview/src/FileView.tsx:108-135)
- `hickey-1` Reload-on-edit concern fragmented across two mechanisms (active() vs KeepAliveMode.heldFile) (packages/solid-fileview/src/FileView.tsx:103-106 and :147-166)
- `hickey-2` heldFile freeze + controlled `mode` prop: a temporal invariant with no structural guard (packages/solid-fileview/src/FileView.tsx:156-158)
- `hickey-3` Keep-alive (visited + hidden + heldFile) is the fourth inline reimplementation of one pattern (packages/solid-fileview/src/FileView.tsx:147-166)
- `hickey-6` Perf-map Finding narrative carries the refuted hypothesis inline rather than collapsing to the shipped truth (docs/atlas/src/content/atlas/performance.mdx:219-240)
- `hickey-7` KeepAliveMode props.show typed boolean while every caller passes a reactive comparison (packages/solid-fileview/src/FileView.tsx:122-131, :148)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 round(s) · codex reviewed at `xhigh` reasoning effort · base `6f1fbbae42e6`

### Round 1

**codex** — approved: `false`

The keep-alive direction is reasonable, but it changes the comment overlay topology from one active Markdown surface to two mounted surfaces. The comment/highlight pipeline is still keyed only by file path, so Source and Rendered can now interfere. I also found one docs accuracy issue in the performance write-up.

Findings:
- `F1` · major · open — Markdown Source and Rendered are now both kept mounted, but each CommentTextSurface still receives every comment for the file via commentsForPath(props.path). That means the hidden/wrong surface can apply highlights for the other surface and can participate in the shared scroll request flow in highlightOverlay.ts:127-156. A source-only quote that does not exist in rendered Markdown, or vice versa, can be cleared/handled by the wrong kept-alive overlay before the intended surface handles it. (packages/client/src/comments/CommentTextSurface.tsx:59)
- `F2` · minor · open — The investigation says the keep-alive path turns content saves into an in-place Markdown memo re-run with no remount. The shipped implementation keeps the render(file) snapshot API and re-invokes props.render(heldFile()) when the visible heldFile changes, so content edits still rebuild that appliance; the win is toggle-back, not edit-time in-place updates. (docs/perf-investigations/markdown-image-resolver-and-toggle.md:106)

**claude** — Addressed both of codex's findings on the Markdown keep-alive change; both agreed and fixed. F1 (major): keeping both Source⇄Rendered surfaces mounted at once left the comment overlay path-only, so each of the two now-live CommentTextSurface overlays received every comment for the file and both raced the shared scroll-request flow. A source-only quote ("# Heading") absent from rendered output (or vice versa) could be find-or-failed and scroll.clear()'d by the wrong kept-alive overlay before the intended surface scrolled — and since the prose overlay re-runs on its own domTick (lazy Shiki swap), it could read a now-cleared request and never scroll. Fixed by making the overlay surface-aware: CommentTextSurface filters commentsForPath by exact surface match (single-surface views keep undefined===undefined), and useHighlightOverlay gates scroll-request consumption on req.surface matching opts.surface. Added an e2e (code-tab.feature:1762) that comments on a source-only quote, flips to Rendered, jumps from the tray, and asserts Source is restored + highlighted. The new scenario passes and the four existing comment/jump scenarios (1711/1735/1792/1823) still pass. F2 (minor): the perf-investigations doc claimed content saves became an in-place memo re-run with no remount; the shipped code keeps the render(file) snapshot contract and re-invokes render(heldFile()) on the visible mode, so a content edit still rebuilds the visible appliance — narrowed the doc table + prose to the actual behavior (toggles avoid remount/sanitize, hidden-mode edits deferred, visible edits still re-render under the snapshot contract). Client typecheck and biome format are clean.

- `F1` **fixed** — Agreed — real defect. Made comment overlay inputs surface-aware. packages/client/src/comments/CommentTextSurface.tsx: commentsForFile now .filter((c) => c.surface === props.surface) and passes surface: () => props.surface to the overlay. packages/client/src/comments/highlightOverlay.ts: added a surface?: Accessor option and gated scroll-request consumption with `if (req.surface !== opts.surface?.()) return;` before find/clear (undefined===undefined keeps single-surface views, plain source/diff, matching a surface-less request). Added regression e2e packages/tests/features/code-tab.feature 'Tray jump returns to the Source Markdown surface' (a source-only '# Heading Src' quote not present in rendered output, switch to Rendered, jump from tray, assert Source restored via fileview-toggle aria-pressed + highlight present). New scenario passes; existing prose-jump/shiki/prose-comment/no-line-history scenarios still pass.
- `F2` **fixed** — Agreed — docs accuracy. The KeepAliveMode body `{props.render(heldFile())}` re-invokes the render(file) closure when the visible heldFile() changes, so a content save on the shown mode rebuilds that appliance (full marked→DOMPurify→Shiki for Markdown) — it is not an 'in-place memo re-run with no remount'. Fixed docs/perf-investigations/markdown-image-resolver-and-toggle.md:106 (split the content-save row into visible vs hidden mode) and the prototype paragraph: the win is the toggle (0 pipeline runs on a flip), visible edits still rebuild under the snapshot contract, hidden edits are deferred until shown so a save never re-renders both modes at once. Note: the Atlas performance.mdx (also in the diff) was already accurate on this point, so no change there.

commit: `2b91c6533762bfc01871815203617be4f0856c34`

### Round 2

**codex** — approved: `false`

CLAUDE fixed the main Source⇄Rendered cross-surface race for newly-created surface-tagged comments and corrected the inaccurate performance-doc claim. One edge of F1 is still open: the exact surface match now drops documented surface-less Markdown comments, so the fix is incomplete for existing/legacy queued comments.

Findings:
- `F1` · major · open — The new exact surface gate fixes tagged Source/Rendered comments, but it leaves surface-less Markdown comments with no owning overlay. CommentTextSurface now filters with `c.surface === props.surface`, and Markdown surfaces pass `props.surface` as `"source"` or `"prose"`; a persisted/commented Markdown entry with `surface === undefined` is filtered out by both surfaces. On tray jump, `req.surface === undefined` also fails both overlay gates, so no overlay highlights or scrolls it. The codebase explicitly documents surface as absent for older persisted entries, so this is a regression introduced by the fix. (packages/client/src/comments/highlightOverlay.ts:145)
- `F2` · minor · resolved — Resolved. The document now distinguishes visible-mode saves, which still re-sanitize/rebuild under the snapshot contract, from hidden-mode edits, which are deferred until shown. (docs/perf-investigations/markdown-image-resolver-and-toggle.md:106)

_codex on the rebuttal:_ F1: I agree the primary race for newly-created source/prose-tagged comments is fixed by the path+surface filter and the scroll-request surface gate. I'm holding F1 open because that exact-match implementation introduced a new no-owner case for documented surface-less Markdown comments: neither kept-alive Markdown overlay has `opts.surface` undefined, so neither sees or consumes a surface-less request. F2: accepted; the docs now describe the shipped behavior accurately.

**claude** — Round 2 of the codex debate. Codex held F1 open: my round-1 exact surface-match filter fixed newly-tagged source/prose comments but orphaned surface-less Markdown comments persisted before the `surface` field existed (#1162). I verified the claim is real — pre-#1162 the source view was the only commentable Markdown surface, so those entries carry a `lineRange` but no `surface`; with both Source⇄Rendered overlays now mounted and filtering by exact surface match, such an entry matched neither overlay's highlight filter and a tray jump (req.surface === undefined) was owned by neither scroll gate, so it silently stopped highlighting and jumping. Fixed by backfilling those legacy entries to "source" at load. F2 was already resolved in round 1; no change needed. Client typecheck and biome are clean; the new unit suite and all comments tests pass.

- `F1` **fixed** — Conceded — codex is right, this was a regression I introduced. Confirmed via git history: `surface` was added in #1162 (f775f769d), and before that PR the Markdown source view was already commentable (BrowseFileDispatcher's sourceRenderer wrapped it in CommentTextSurface) while the rendered preview was 'none'. So a Markdown comment persisted pre-#1162 carries a lineRange but surface===undefined. After round 1's exact-match filter (CommentTextSurface.tsx:70 `c.surface === props.surface`) and the scroll gate (highlightOverlay.ts:145 `req.surface !== opts.surface?.()`), such an entry: (1) highlights on neither overlay (undefined !== 'source'/'prose'), and (2) on tray jump CodeTab.tsx:1225 sets req.surface=undefined, so jumpMode() returns null → FileView defaults to rendered, and neither scroll gate owns the request. Fix: backfill at the persistence source of truth. New pure leaf packages/client/src/comments/backfillSurface.ts (mirrors formatMarkdown.ts so it has no Solid/DOM deps and is unit-testable) maps surface-less comments on a Markdown path (isMarkdown) to surface:'source'; non-Markdown surface-less comments (plain source, diff, HTML-iframe) stay undefined so their lone single-surface overlay keeps matching undefined===undefined. Wired into useComments.ts's persistence parse path (line ~46: `return { v: 1, comments: backfillSurface(parsed.comments) }`). Updated the Comment.surface doc in types.ts. Added regression packages/client/src/comments/backfillSurface.test.ts (4 cases: Markdown backfilled to source, non-Markdown .ts/.html left undefined, already-tagged prose/source not clobbered, all other fields preserved) — passes. I chose load-time normalization over codex's alternative of special-casing req.surface===undefined at the overlay because normalizing at the store (the source of truth) means every downstream read site stays a simple exact match with no legacy branch, matching the repo's 'reuse the existing source of truth' + the documented migration path in useComments.ts.
- `F2` **fixed** — Resolved in round 1 (commit 2b91c6533) and codex marked it resolved in this verdict (status: resolved, 'the docs now describe the shipped behavior accurately'). The perf doc now distinguishes visible-mode saves (still re-sanitize/rebuild under the snapshot contract) from hidden-mode edits (deferred until shown). No further change this round.

commit: `86f4adf4c6a723e1231ef76a6d006d84e290129a`

### Round 3

**codex** — approved: `true`

The current round adequately closes the remaining issue. The new load-time backfill normalizes legacy surface-less Markdown comments to `source` before the exact surface filters and scroll gates see them, while preserving undefined surfaces for single-surface files. The F2 documentation correction from the prior round remains accurate. I did not find a new regression introduced by this round's changes.

Findings:
- `F1` · major · resolved — Resolved. Legacy Markdown comments are now normalized during persisted comment parsing via `backfillSurface`, which maps only surface-less Markdown-path comments to `surface: "source"`; the existing exact-match overlay filter and scroll gate then have a concrete owner for those comments. (packages/client/src/comments/useComments.ts:44)
- `F2` · minor · resolved — Resolved. The perf note now distinguishes visible-mode content saves, which still rebuild the shown appliance under the snapshot contract, from hidden-mode edits, which are deferred until shown. (docs/perf-investigations/markdown-image-resolver-and-toggle.md:106)

_codex on the rebuttal:_ F1: Accepted. Load-time normalization is the right place for this compatibility fix: it keeps downstream overlay logic as an exact surface match, restores ownership for pre-surface Markdown source comments, and leaves non-Markdown single-surface comments as `undefined === undefined`. The added unit cases cover the important boundaries. F2: Already accepted in the prior round and still resolved; no further dispute remains.

srid:

Ran with `--no-elegance` (the simplify pass already covered that tree).

**Pass 1 — rule checklist: clean.** All 30 rules pass (dry-rule-of-three, no-dead-code, no-silent-error-swallowing, no-unbounded-growth, persisted-schema-stays-tolerant, migration-shape-guard, solid-reactive-prop hygiene, e2e-poll-async-state, …). Notably: `COMMENT_HIGHLIGHT_NAME_PREFIX` is a single exported source of truth; the old `HIGHLIGHT_NAME`/`ensureStyle()` were removed (no dead code); `backfillSurface` only touches surface-less Markdown comments (shape-guarded); per-instance highlights are cleaned on `onCleanup`.

**Pass 2 — fact-check: 1 finding, adjudicated as a non-issue; 2 self-refuted false positives.**

- *Finding (scroll-request has no expiry):* a `ScrollRequest` whose `surface` matches no mounted overlay could persist. **Non-issue** — `scrollRequest.ts` is a **single-slot** signal (`createSignal<ScrollRequest | null>`, "one-shot, repeat clicks set fresh values"), so the next tray click *overwrites* any stale request (no queue, no accumulation), and the surface gate makes a non-matching overlay a no-op (no spurious scroll). The matching overlay is also guaranteed to exist: `backfillSurface` normalizes legacy surface-less Markdown comments to `source`, and `jumpMode` forces the toggle to mount the request's surface before the overlay re-finds. No guard needed.
- *False positive (non-Markdown backfill orphaning):* `undefined === undefined` still matches for single-surface files — self-refuted.
- *False positive (test highlight-iteration `.catch(()=>0)`):* runs in a controlled Chromium e2e where the API is stable — acceptable.

No fixes applied.

srid:

The **x86_64-linux** lane ran on **`kolu-ci-7`** (`dev-x86-64-linux-05`) — commit `899f1cdde`, _verdict unknown (no per-node timing)_

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

PR #1449, opened 2026-06-20
Seat the Terminal sum: active | sleeping (Phase 1)

srid:
**Phase 1 of the sleeping-terminals redesign** ([plan of record](https://kolu.dev/atlas/sleeping-terminals.html), #1444): model a terminal as a sum — **`Terminal = active | sleeping`** — so a dormant terminal becomes a *state* on one record rather than a separate thing. This PR is a **pure, zero-behavior-change foundation**: it flips the schema to a discriminated union, adds the migration the flip forces, and threads the `state === "active"` narrowing seam through every liveness reader. Only the `active` arm is ever constructed, so the UX is pixel-identical — what lands is the *seam* every later phase leans on.

### The fold already lived in the schema

`surface.ts` already split a terminal's fields into a **persisted base** (cwd · git · intent · `canvasLayout`) and a **live overlay** (agent · foreground · pr). That partition *is* active-vs-sleeping, so the sum maps onto bases that already exist — exactly one new scalar (`sleptAt`) is added.

| Layer | Before | After |
| --- | --- | --- |
| `TerminalMetadataSchema` (wire/collection) | flat `.merge` | `discriminatedUnion("state", [Active, Sleeping])` |
| `SavedTerminalSchema` (on-disk) | `Persisted + id` | same discriminant, minus the live overlay |
| `state.ts` | `SCHEMA_VERSION 1.26.0` | `1.27.0` + `backfillTerminalState_1_27_0` stamps `state: "active"` |
| `TerminalProcess.meta` (server registry) | `TerminalMetadata` | `ActiveTerminal` — a live PTY is active by construction |
| client presence reads | `meta.agent` / `.pr` / `.foreground` | narrow `state === "active"` first (the `activeArm` seam) |

### Where the narrowing lands

```
canvas · dock · minimap · arrange · cycle · switcher   ─ read the Terminal union (presence)
            │
   state === "active"  ── compiler-forced narrow ──┐
            ▼                                        ▼
   live fields: pr · agent · foreground · PTY/xterm · input routing
```

A consumer that touches a live field *must* first narrow to the active arm — the compiler refuses a live field on the bare union. So a sleeping terminal can sit on the canvas and in the MRU yet can **never** be an input or WebGL target: a type fact, not a runtime guard. On the server the live registry only ever holds active terminals, so the adoption path (`reconcile → adoptLocalTerminal → adoptedMeta`) carries `SavedActiveTerminal` — a sleeping record has no PTY to adopt.

> **Boundary held:** the `@kolu/terminal-awareness` schemas stay **flat**. The union is recomposed only in `kolu-common`; `state` never crosses the awareness wire, so `arivu` and `kaval` need no change (their suites stay green). Adding `state` to an awareness schema would have bumped arivu's served contract — that is the one edit this PR deliberately does *not* make.

### Design philosophy

- **Fail fast — no fallbacks.** The migration is the *one* sanctioned place the `state` default is supplied; read sites narrow, never coalesce (`?? "active"` would be a defect).
- **Volatility boundary held.** The awareness package stays flat; the canvas-lifecycle concept (`state`) lives in `kolu-common` where the canvas owns it.
- **Reuse the source of truth.** The migration mirrors the existing `backfillLocation_1_26_0` ladder entry verbatim, not a parallel mechanism.

### Verification

`just check` (typecheck + biome) ✓ · `just fmt` ✓ · unit suites green (client **418**, server **101** incl. the new `backfillTerminalState_1_27_0` test, common 17; terminal-awareness / arivu / kaval untouched and green).

_No `## Evidence` screenshot: zero visual change by construction (every terminal is active; UX pixel-identical). The behavioral proof is the green migration + session-restore unit coverage; a session round-trip clip rides §5._

_No changelog entry: pure internal refactor + a transparent on-load migration — no user-visible behavior._


srid:

✅ **Consensus** after 2 round(s) · lowy + hickey · base `e226729ea034`

Independent findings: lowy=3, hickey=6

### Applied (6)
- `lowy-1` Two parallel narrowing receptacles: activeArm() vs. bare `state !== "active"` — commit `920649ecf`
- `hickey-1` Two idioms for the same 'narrow to active arm' operation — (uncommitted)
- `hickey-2` SavedTerminal sum is asymmetric: active arm named/exported, sleeping arm inlined anonymously — commit `a3bee7129`
- `hickey-3` Active arm + prValue recomposed at four call sites with two different shapes — commit `954c15a13`
- `hickey-4` ActiveTerminalSchema and SavedActiveTerminalSchema re-spell the same base composition — commit `a6f673791`
- `hickey-6` Discriminant comment claims 'nowhere read-side' while three sites read it raw — commit `3f027d7e6`

### Agreed — no change (3)
- `lowy-2` `createMetadata` stamps the `state: "active"` discriminant inline rather than through an arm constructor (packages/server/src/terminalEndpoint/metadata.ts:68)
- `lowy-3` `snapshotSession` projects via schema-parse, but `adoptedMeta` still hand-destructures the persisted/live split (packages/server/src/terminals.ts:69 vs. packages/server/src/terminalEndpoint/local.ts:330)
- `hickey-5` Sleeping arm has zero producers — the union's second arm is type-only in this PR (packages/common/src/surface.ts:323-352)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `e226729ea034`

### Round 1

**codex** — approved: `false`

The active/sleeping discriminant is threaded through most runtime code cleanly, and the persisted Conf migration covers the server's own stored sessions. The main gap is that not every saved-session ingestion path gets that migration treatment: old exported sessions now fail import, and the e2e restore helper still posts legacy-shaped fixtures into the new schema.

Findings:
- `F1` · major · open — `parseSavedSession` now validates imports directly against the new discriminated `SavedSessionSchema`, so any pre-1.27 `kolu-session.json` exported before this branch lacks `terminal.state` and is rejected as invalid. The server Conf migration backfills persisted local state, but imported diagnostic backups bypass `packages/server/src/state.ts`, which breaks the recovery path this module is explicitly meant to provide. (packages/client/src/sessionTransfer.ts:50)
- `F2` · major · open — The e2e saved-session fixture helper only stamps the now-required `location`; it still posts terminal objects without `state`. Since `/rpc/surface/kolu/session/test__set` validates the payload with `SavedSessionSchema`, restore scenarios that call this helper now send invalid saved terminals and fail before exercising the UI. (packages/tests/step_definitions/session_restore_steps.ts:52)

**claude** — Both of codex's major findings were valid; I agree with both and fixed them. F1: the diagnostic "Import session" hatch (parseSavedSession) validated an imported kolu-session.json directly against the new discriminated SavedSessionSchema, so a backup exported before the `state` discriminant (and the earlier `location`/`remoteUrl` bumps) was rejected — defeating the recovery hatch's stated purpose. F2: the e2e session-restore fixture helper stamped the now-required `location` but not the `state` discriminant, so its posted saved terminals were invalid under the new schema and would fail before driving the UI. I fixed F1 with the single-source-of-truth factoring codex suggested: moved the three pure field-backfills (remoteUrl/location/state) out of the server's state.ts migration ladder into kolu-common/surface beside the schema they restore, exposed a composed backfillSavedSession, and ran it on the import path before validation; the server ladder now imports the same functions. Fixed F2 by stamping state: "active" beside location in the e2e helper. All unit suites pass (common 17, server 101, client 419) and common/server/client typecheck clean.

- `F1` **fixed** — Agreed — real gap (and one this branch widened by the new `state` field; the import path never ran the migration backfills). Per codex's suggestion, moved the three pure field-backfills (backfillRemoteUrl, backfillLocation, backfillTerminalState) from packages/server/src/state.ts into packages/common/src/surface.ts beside SavedTerminalSchema, plus composed helpers backfillSavedTerminal and backfillSavedSession. The server's 1.25.0/1.26.0/1.27.0 ladder steps now call the common functions (one source of truth); state.test.ts imports them from common. packages/client/src/sessionTransfer.ts:parseSavedSession now runs backfillSavedSession(parsed) before SavedSessionSchema.safeParse, so a legacy export missing state/location/remoteUrl is upgraded then validated. Added a regression test in sessionTransfer.test.ts importing a legacy session with no `state`/`location`.
- `F2` **fixed** — Agreed — confirmed `state` is required by the schema (the parseSavedSession test's valid fixture carries state: "active"), and the e2e helper at packages/tests/step_definitions/session_restore_steps.ts:52 only stamped `location`, so test__set received invalid saved terminals. Stamped `state: "active"` beside `location: LOCAL_LOCATION` in the helper's terminals.map (the existing inline-default pattern; the discriminant has no schema default and the package is transpiled-not-typechecked, so the runtime stamp enforces it). A terminal supplying its own state/location still wins via the spread.

commit: `20deec8edd9838ba882e8f2ef2a176f2afd0dd82`

### Round 2

**codex** — approved: `true`

Both previously raised issues are resolved in the current tree. I verified by inspection that imported saved sessions now pass through the shared common backfill before `SavedSessionSchema` validation, and the e2e restore helper now stamps `state: "active"` alongside `location` for legacy-shaped fixture terminals.

Findings:
- `F1` · major · resolved — Resolved. `parseSavedSession` now validates `backfillSavedSession(parsed)`, and the shared helper in `packages/common/src/surface.ts` backfills legacy terminal `remoteUrl`, `location`, and `state` before schema validation. The added regression test covers a legacy export missing `state` and `location`. (packages/client/src/sessionTransfer.ts:62)
- `F2` · major · resolved — Resolved. The e2e saved-session fixture helper now stamps `state: "active"` together with `location: LOCAL_LOCATION` before spreading caller-provided terminal fields, so the posted payload satisfies the discriminated saved-terminal schema while still allowing explicit future states to override it. (packages/tests/step_definitions/session_restore_steps.ts:54)

_codex on the rebuttal:_ No disputes to answer. CLAUDE agreed with both findings, and the current changes address F1 and F2 as described.

srid:

Ran on the full Phase-1 diff (`git diff <merge-base>...HEAD`) after lens + codex + simplify. **All clear — no fixes applied.**

| Pass | Result |
|---|---|
| **1 — Rules** | ✅ 0 violations across all 31 rules (built-in + project), incl. `persisted-schema-stays-tolerant` and `caught-error-must-not-collapse-to-empty` |
| **2 — Fact-check** | ✅ Clean. One flag investigated and dismissed as a false positive: `snapshotSession`'s `SavedActiveTerminalSchema.parse({ ...entry.meta, id })` does not throw on valid `ActiveTerminal` data — Zod object schemas strip unknown fields by default, so the live overlay (`agent`/`foreground`/`pr`) is stripped silently. That stripping is the *point* of the change (schema-derived projection over a hand-named destructure). |
| **3 — Elegance** | – Skipped (`--no-elegance`; `/simplify` already ran this tree) |

The `persisted-schema-stays-tolerant` rule is especially relevant here and is honored: the new `1.27.0` migration backfills `state: "active"` (idempotent, keyed on the discriminant) and the client "Import session" hatch runs the same shared backfill before validation — no fatal validator tightens the persisted set without a backward-compat path.

PR #1453, opened 2026-06-20
Live truecolour arivu-tui on Bun + OpenTUI (arivu P3a)

srid:
**`arivu-tui` now renders a live, truecolour view of what every terminal is in** — the agent state (working cyan, *awaiting you* amber) and the PR's CI checks (pass green / fail red) coloured so the one that needs you catches your eye, and `watch` repainting *in place* as awareness changes. It's the **P3a** slice of the [arivu fleet plan](https://htmlpreview.github.io/?https://github.com/juspay/kolu/blob/feat/arivu-tui-bun-opentui/docs/atlas/dist/arivu.html): re-platform the viewer onto Bun + OpenTUI first, against the simple single-host view, before P3b fans it into the multi-host board.

### The change in one line

Re-platform **only** the `arivu-tui` viewer from `tsx` to **Bun** so it can run [OpenTUI](https://opentui.com)'s Solid reconciler (its renderer is a native Zig core loaded via Bun's FFI). The rest of kolu — and the `arivu` **daemon** that travels over ssh and leans on `node:sqlite` — stay on Node.

### How it's structured

| Layer | What it does |
| --- | --- |
| `render.ts` | the **pure projection** — `fieldRows` / `agentTone` / `prTone` / `recordHeader`. Node/vitest-tested; the renderer holds no data logic. |
| `tui.tsx` | the OpenTUI/Solid components — map a `tone` to a colour, lay out the record. `list` = one-shot main-screen → destroy after first frame; `watch` = live alternate-screen driven by a Solid store + `reconcile`. |
| `bin.ts` | gates the OpenTUI render behind `process.stdout.isTTY && !--json` — a piped / JSON run stays plain text and never loads the Bun-only native renderer. |
| `default.nix` | `mkAgentTuiWrapper` gains an `interpreter` arg (default `tsx`; arivu-tui→`bun`) + a `--preload` for @opentui/solid's JSX transform. kaval-tui + the daemon are byte-identical. |

### Two boundary calls worth a look

- **Bun is contained, on purpose.** The thing that forces Bun (OpenTUI's FFI renderer) is a *local-viewer* concern, so the Bun runtime + per-arch `libopentui.so` live only in the local closure — never crossing ssh. One wrapper, one binary, easy to undo. The packaging follows drishti's Bun-on-Nix precedent.
- **The prune now spares `@babel*` (~17MB).** babel is build-only for the client bundle but **runtime** for arivu-tui — its preload compiles the Solid JSX with `babel-preset-solid` on every launch (we run the viewer from *source* under bun, the same model as the rest of the workspace). Pruning it broke the binary with `ENOENT … @babel/core`. The slim alternative (bun-bundling the viewer at build time so no runtime babel) is a deliberate, documented follow-up.

### Design philosophy

- **Fail-fast, no knobs** — the wrapper bakes the interpreter + drv map with `--set` (not `--set-default`); a missing daemon is a loud `no arivu at … (ENOENT)`, never a silent empty render. The `isTTY`/`--json` branch selects the right output for the *medium* (it's a correctness requirement — the renderer corrupts a pipe), not a degradation fallback.
- **Reuse the source of truth** — Bun consumes the existing pnpm tree as-is (no `bun.lock` competing with `pnpm-lock.yaml`, no flake input); OpenTUI is the renderer *electricity* taken as a library rather than a hand-rolled `@kolu/solid-tui`.

### Proof

- `render.test.ts` (projection, Node/vitest, 28 tests) + `tui.bun-test.tsx` (testRender paint, Bun lane) — **both green against the built Nix closure**, which proves `bun.dlopen(libopentui.so)` + the babel preload work in the pruned tree.
- `just check` green; `nix build .#arivu-tui` green; the binary runs (`--help`, and a clean fail-fast `list` with no daemon).
- The live render on a real terminal is captured under **## Evidence** below.

### Try it locally

\`\`\`sh
nix run github:juspay/kolu/feat/arivu-tui-bun-opentui#arivu-tui -- list
\`\`\`


srid:
Closing in favor of a re-planned, **two-PR** approach — this branch's history became a long debugging trail and is not worth untangling into a reviewable diff.

What we learned (and what survives):
- The Bun + OpenTUI + SolidJS toolchain works; `build.ts` (Bun.build + `@opentui/solid/bun-plugin`) is the *official* recipe and the bundle is fully reactive — no run-from-source needed.
- The "frozen" symptom that drove most of this was largely a **measurement artifact**: OpenTUI repaints only the single changed cell, so a `kaval-tui snapshot | grep` (and a glance) miss the one ticking character.
- Stick to **SolidJS-canonical** reactivity (`createSignal` + `onMount(setInterval)` + `onCleanup`) — no fps caps, `requestRender`, external timers, or stderr monkeypatching.

Salvaged into the re-plan (arivu Atlas note): `render.ts` projection, the data layer (`connect`/`read`/`hostConnect`/`snapshotAwareness`), the Bun-vs-Node decision, and the daemon-quiet `--stdio` fix. New PRs to follow: **PR1** Bun+OpenTUI foundation (verified-live), **PR2** the fleet dashboard.

PR #1457, opened 2026-06-20
feat(terminal): live-output activity dot on title bar and dock

srid:
## What & why

Nothing in the UI told you whether a terminal was **idle** (output sitting static) or **live** (output actively streaming). A glance down the dock couldn't separate the sessions doing work from the ones merely open — which is exactly the gap the *Terminal live-activity indicator* Sundry item called out.

The obvious candidate, `lastActivityAt`, doesn't answer this: it bumps **only on agent semantic-key transitions** (see `staleness.ts`), so a plain `npm run build`, a `tail -f`, or any non-agent shell churning output would never light it up. That field is an hours-scale *staleness* clock, not a "moving right now" signal.

So this adds the honest signal, sourced from the one place raw output motion actually exists — the PTY data stream itself.

## How

- **`useTerminalActivity`** — a singleton store (`createSharedRoot`, mirroring `useTerminalStore`). `noteOutput(id)` flips a per-terminal `live` flag on every chunk and arms a ~1s debounce timer that flips it back to static; each fresh chunk re-arms it. The flag is an explicit boolean rather than a `now - lastOutputAt` comparison, so reactivity needs no global ticking clock — the timer *is* the clock.
- **`Terminal.tsx`** — calls `activity.noteOutput(terminalId)` from the attach-stream sink, right where each chunk lands in xterm (fires even when scroll-locked: the bytes still arrived).
- **`LiveActivityDot`** — a soft green pulse rendered straight off the flag, and **nothing at all** when static. Mirrors `ChecksIndicator`'s dot geometry so it reads as one visual family. Green (`--color-ok`) is the one state colour the agent pips don't claim (alert = violet, busy = rust, accent = blue), so the live axis never blurs into agent state.
- Rendered **leading the canvas tile title row** (`TerminalMeta`) and **beside the dock row's recency clock** (`Dock`) — so "5s ago" (touched recently) and "● live now" (streaming) sit as two distinct axes.

Scope note: the signal is client-side, so it covers terminals mounted on the canvas (which stay attached/streaming even when hidden — `CanvasTile`). Parked terminals — idle by definition — simply read static. A server-side `lastOutputAt` for *every* terminal regardless of attachment is the natural graduation if we want the dot on parked rows too.

## Evidence

Driven on a live local kolu: a shell streaming a fake compile loop, then `Ctrl+C`.

**Title bar — live** (green dot leads the title while output streams):

title bar live

**Dock — live vs. static** (dot beside the recency clock; gone once output stops):

| Streaming | Static (after ~1s quiet) |
| --- | --- |
| dock live | dock static |

**Full workspace, streaming:**

full live

## Tests

`useTerminalActivity.test.ts` (fake timers): lights on output, stays lit right up to the threshold, decays to static after the quiet period, and re-arms on each fresh chunk. `typecheck` + `biome` clean; `App.shell` test still green.

Closes the **Terminal live-activity indicator** Sundry item.


srid:
## Follow-up: dock dot alignment

Caught a layout nit — the dot sat inside a `justify-end` recency cell, so it floated left of the timestamp and a wider label (`just now` vs `5s`) nudged it out of column. Fixed by pinning the dot to the cell's left edge and right-aligning the timestamp independently (`flex-1 text-right`), so the changing label reflows the gap between them, never the dot.

Verified across rows with deliberately different labels — the green dots hold a single x-column whatever the timestamp width:

dock dots aligned

srid:
## Rework: live dot is now a zero-reflow overlay

Two issues from the inline placement, both fixed — the dot no longer takes any layout space, so appearing/disappearing never moves existing content.

**Terminal title** — the dot led the name row, so it pushed the title text right when it toggled. It's now an absolute overlay sitting in the title bar's left gutter. Verified by measurement: the name's left x is *identical* (`l=138`) whether the dot is shown or not.

title gutter dot

**Dock** — a separate dot collided with the right-aligned timestamp, and earlier it shifted with the label width. Since "live now" makes "Xs ago" moot while output is streaming, the recency cell now **swaps the timestamp for the dot in place** — right-aligned in the same slot. It lands in one column across rows, never overlaps, and never reflows the grid. When output stops, the timestamp returns.

dock dot swaps timestamp

(top row is streaming → dot; the others show their recency, all in the same right-aligned column.)

srid:

✅ **Consensus** after 3 round(s) · lowy + hickey · base `6c953a685f97`

### Applied (6)
- `lowy-1` Dot's render contract serves two jobs; dock exposes the wire — commit `d4bead0f9`
- `lowy-3` noteOutput timers are never cleared on terminal close — stale keys + a setLive(false) after dispose — commit `77866a9aa`
- `hickey-1` isLive guard duplicated — dock checks it twice, title once — (uncommitted)
- `hickey-3` live store and timers Map accumulate dead entries for removed terminals — (uncommitted)
- `hickey-4` Two parallel per-terminal output-stream trackers fed from the same attach sink — commit `a6d5a8e69`
- `hickey-6` IDLE_AFTER_MS comment over-claims; flag flips off mid-stream of any quiet >1s — commit `e41e95ace`

### Agreed — no change (4)
- `lowy-2` Live signal sourced only from mounted attach streams — dock rows for unmounted terminals can never light (packages/client/src/terminal/Terminal.tsx:790-792)
- `lowy-4` Two consumers, two layout idioms for the same dot — symmetry break worth noting (packages/client/src/terminal/TerminalMeta.tsx:62-67 vs packages/client/src/canvas/dock/Dock.tsx:492-499)
- `hickey-2` useTerminalActivity is consumed three times for the same id, re-running the singleton accessor per component (packages/client/src/canvas/dock/Dock.tsx:439, packages/client/src/terminal/LiveActivityDot.tsx:16, packages/client/src/terminal/Terminal.tsx:140)
- `hickey-5` Liveness derived from a hand-rolled debounce timer instead of the existing now-ticker pattern (packages/client/src/terminal/useTerminalActivity.ts:23-50)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 round(s) · codex reviewed at `xhigh` reasoning effort · base `6c953a685f97`

### Round 1

**codex** — approved: `false`

The change is close structurally: the singleton tracker is bounded, cleanup clears timers, and the desktop title/dock wiring is straightforward. I would not approve yet because the attach stream integration currently marks historical snapshots as live output, and the feature is only wired into a subset of the terminal navigation surfaces.

Findings:
- `F1` · major · open — `activity.noteOutput` runs for every item from `terminal.attach`, but this stream yields the serialized screen snapshot first and only then live PTY deltas. A quiet terminal with a prompt/scrollback will light the live dot for a second on initial mount, mode remount, or reconnect retry, so the indicator lies exactly when users scan the workspace. (packages/client/src/terminal/Terminal.tsx:793)
- `F2` · major · open — The touch/compact dock row still renders `formatTimeAgo` unconditionally, while the desktop cards dock swaps that same cell to the live dot. `DockList` is the phone drawer and compact rail, and its own comment says it mirrors the desktop dock row, so those users get no dock live-activity signal at all. (packages/client/src/canvas/dock/DockList.tsx:179)
- `F3` · minor · open — Sub-terminals mount the same `Terminal` component and therefore record activity under the child `subId`, but the new visible readers are keyed only by top-level tile/dock ids. Output in a split terminal is tracked but has no UI reader, so the live indicator never appears for that terminal. (packages/client/src/terminal/TerminalContent.tsx:156)
- `F4` · minor · open — The recency cell sits in an `auto` grid track and swaps text like `5m ago` for a 6px dot. In a one-row section, or when every row in the section is live, that changes the track's intrinsic width and shifts the branch column, contradicting the nearby "never reflows the grid" comment. (packages/client/src/canvas/dock/Dock.tsx:492)
- `F5` · minor · open — The dot always uses `animate-pulse`, but the changelog says reduced-motion stills it and the existing CSS has explicit `prefers-reduced-motion` handling for other indicators. Users with reduced motion enabled will still get the pulse. (packages/client/src/terminal/LiveActivityDot.tsx:23)

**claude** — Addressed all five of CODEX's findings on the live-activity dot. I agreed with every one — each was a real defect. F1 (the headline major): the attach stream yields a serialized screen snapshot as its first item before live deltas, and the old code called noteOutput on it, so a quiet terminal with scrollback lit the dot for ~1s on every mount/remount/reconnect — the indicator lying exactly when a workspace glance relies on it. I extracted a small per-subscription snapshot-boundary helper (createSnapshotBoundary) that suppresses the first frame and re-arms in onRetry (which fires before each transparent re-subscribe's fresh snapshot), and covered initial attach + retry with a unit test. This is CODEX's lighter second suggestion; I deliberately did NOT change the wire protocol to a tagged frame kind because that would touch the @kolu/surface contract and trigger a mandatory drishti PR (surface.md) for no functional gain. F2: wired the dot into the touch/compact DockListRow and the glyph-only RailChip (corner overlay, since it has no timestamp cell). F3: SubPanelTabBar now renders the dot per sub tab. F4: reserved w-[4.5ch] on both desktop and touch recency cells so swapping the dot for "5m ago" never reflows the auto grid track. F5: switched to motion-safe:animate-pulse so reduced-motion users get a static dot. All 130 terminal+dock tests pass, tsc --noEmit clean, biome clean.

- `F1` **fixed** — Confirmed: packages/server/src/router.ts:155 yields `snapshot` first then deltas, both plain strings, so the client callback at Terminal.tsx couldn't tell them apart and lit the dot for the snapshot on every mount/reconnect. Added packages/client/src/terminal/snapshotBoundary.ts (createSnapshotBoundary) — its isLiveDelta() returns false for the first frame after construction or armSnapshot(), true thereafter. Terminal.tsx now skips activity.noteOutput for the boundary frame and calls snapshotBoundary.armSnapshot() inside onRetry (the retry plugin invokes onRetry before the re-subscribed iterator's fresh snapshot, per surface/src/client.ts docs). The snapshot is still written to xterm; only the activity ping is suppressed. New test snapshotBoundary.test.ts covers initial attach AND reconnect retry. Chose the per-stream boundary over the tagged-frame-kind option because the latter would change TerminalAttachOutputSchema (a z.string()) in the @kolu/surface contract and force a drishti PR per surface.md, with no behavioral benefit over the boundary approach.
- `F2` **fixed** — Confirmed: DockList.tsx:180 (DockListRow, the phone drawer + compact rail) rendered formatTimeAgo unconditionally. Wired useTerminalActivity + LiveActivityDot into DockListRow with the same Show-swap as the desktop DockRow. Also made the explicit decision CODEX asked for on rail mode: RailChip (Dock.tsx:571) is a glyph-only 32x32 chip with no recency cell, so I added the dot as a pointer-events-none absolute top-right corner overlay (the chip is position:relative). Documented that it is orthogonal to the existing agent-state glow (glow = agent thinking/waiting; dot = any bytes moving).
- `F3` **fixed** — Confirmed: TerminalContent.tsx:156 mounts the same Terminal component for each subId, so sub-terminals record activity under their child id but the top-level readers are keyed by the parent id. SubPanelTabBar.tsx is the only per-sub UI surface, so I wired useTerminalActivity there and render LiveActivityDot inside each sub tab's label button, gated by activity.isLive(id). Chose the per-sub indicator over an isLiveTree aggregation because the tab bar already has a natural per-sub slot and aggregation would hide which split is live.
- `F4` **fixed** — Confirmed: the recency cell is the last `auto` track in grid-cols-[16px_minmax(0,1fr)_auto_auto] (Dock.tsx:364) — swapping the ~6ch timestamp for a 6px dot (or a whole section going live) shrinks that intrinsic track and shifts the 1fr name column, contradicting the 'never reflows the grid' comment. Reserved a stable width with inline-flex justify-end w-[4.5ch] on both the desktop DockRow recency cell (Dock.tsx) and the touch DockListRow recency cell (DockList.tsx) so the swap is strictly in-place.
- `F5` **fixed** — Confirmed: LiveActivityDot.tsx hardcoded animate-pulse with no reduced-motion handling, while the changelog claims 'reduced-motion stills it like every other kolu indicator.' (Note: the sibling ChecksIndicator also uses bare animate-pulse with no handling, so the claim was aspirational.) Switched to motion-safe:animate-pulse (Tailwind v4 supports the variant; confirmed @tailwindcss/vite ^4.1.0). This honors conventions.md 'reuse existing source of truth' for a Tailwind-utility element — the named-class indicators use an explicit animation:none block in index.css, but a utility-class dot rides Tailwind's own reduced-motion variant rather than a bespoke class. Reduced-motion users now get a static (still green, still visible) dot.

commit: `669f0ee5cd1d3852ff73dcd7ebde0d08e7cc5e28`

### Round 2

**codex** — approved: `false`

Claude addressed the broad shape of all five findings, and the sub-terminal reader plus reduced-motion fix are good. I still would not approve: the snapshot-boundary fix depends on a stronger client-visible contract than the server currently provides, the rail-mode dot collides with the existing unread badge, and the fixed recency width is too narrow for the actual monospace fallback strings.

Findings:
- `F1` · major · open — The new client-side boundary assumes the first callback frame is always the snapshot, but the current server only yields it when `snapshot` is truthy (`router.ts:155`). `TerminalAttachment.snapshot` is explicitly allowed to be `""` for a PTY with no output, so on a blank terminal attach/retry the client receives no snapshot frame and `isLiveDelta()` suppresses the first real live delta. Short first-output bursts can therefore fail to light the indicator at all. (packages/client/src/terminal/Terminal.tsx:804)
- `F2` · minor · open — The missing touch/compact dock wiring is fixed, but the new rail-mode dot uses the top-right corner already occupied by `.dock-rail-chip[data-unread]::after`. A rail chip that is both unread and live will have the live dot and unread badge painted on top of each other, making one or both signals ambiguous. (packages/client/src/canvas/dock/Dock.tsx:636)
- `F3` · minor · resolved — Sub-terminal output previously recorded activity under the child id without any visible reader keyed by that id. The tab bar now reads `activity.isLive(id)` per sub-tab and renders `LiveActivityDot`, so the per-sub activity has a UI surface. (packages/client/src/terminal/SubPanelTabBar.tsx:54)
- `F4` · minor · open — The grid track is now fixed, but `w-[4.5ch]` is too narrow for the actual monospace fallback strings from `formatTimeAgo`: `1m ago` is 6ch, `23h ago` is 7ch, and `just now` is 8ch. The auto track may stop reflowing, but the timestamp can now overflow into the previous columns. The same width is used in `DockList.tsx`. (packages/client/src/canvas/dock/Dock.tsx:495)
- `F5` · minor · resolved — The live dot previously always animated despite the changelog promising reduced-motion handling. It now uses `motion-safe:animate-pulse`, so reduced-motion users get the static dot. (packages/client/src/terminal/LiveActivityDot.tsx:29)

_codex on the rebuttal:_ Claude agreed with the original findings, so there was no dispute to answer. I accept the decision not to change the attach output schema for drishti reasons, but the chosen client-only boundary is only correct if the server always emits a snapshot frame; current `if (snapshot) yield snapshot` breaks that for empty snapshots, so F1 remains open. F2 is partly fixed, but the rail placement introduced an unread-badge collision. F3 and F5 are resolved. F4's stable-column approach is right, but the concrete `4.5ch` width is smaller than the formatter's actual monospace strings, so it still needs adjustment.

**claude** — Round 2 codex debate, live-activity dot. Fixed F1 (router.ts attach yields snapshot unconditionally including empty string so a blank PTY no longer misclassifies its first delta; added empty-snapshot test), F2 (rail dot moved to bottom-right to avoid unread badge collision), F4 (recency width 4.5ch to 8ch in Dock and DockList). F3 and F5 resolved by codex last round. tsc clean, biome clean, 147 client and 22 server tests pass. Commit 882d88a88.</summary>
</invoke>


- `F1` **fixed** — router yields snapshot unconditionally including empty string; added test
- `F2` **fixed** — rail dot moved to bottom-right corner
- `F3` **fixed** — resolved by codex last round
- `F4` **fixed** — recency width 4.5ch to 8ch
- `F5` **fixed** — resolved by codex last round

commit: `882d88a8810e26d0fd0c1ca22630c3217a2d4dc4`

### Round 3

**codex** — approved: `true`

The current tree resolves the live-activity review thread. The server now preserves a real attach snapshot boundary even for empty snapshots, rail live/unread indicators no longer overlap, recency cells reserve enough width for the formatter's output, and the previously resolved sub-terminal and reduced-motion fixes remain in place. `git status --short` showed no untracked files.

Findings:
- `F1` · major · resolved — The prior client-side snapshot boundary was unsound when the server omitted empty snapshot frames, causing the first real delta from a blank PTY to be misclassified as the snapshot. (packages/server/src/router.ts:162)
- `F2` · minor · resolved — The rail live dot previously used the top-right corner already occupied by the unread badge, causing overlapping indicators on unread+live chips. (packages/client/src/canvas/dock/Dock.tsx:644)
- `F3` · minor · resolved — Sub-terminal activity was recorded under child ids but had no visible per-child reader. (packages/client/src/terminal/SubPanelTabBar.tsx:54)
- `F4` · minor · resolved — The fixed recency width was too narrow for `formatTimeAgo` strings such as `just now`, risking overflow even though the grid no longer reflowed. (packages/client/src/canvas/dock/Dock.tsx:497)
- `F5` · minor · resolved — The live dot previously animated even for reduced-motion users. (packages/client/src/terminal/LiveActivityDot.tsx:29)

_codex on the rebuttal:_ No disputes remained. I verified F1 by checking `router.ts` now yields the snapshot unconditionally and the boundary test covers `""`; F2 by checking the rail dot moved to bottom-right; F4 by checking both dock recency cells use `w-[8ch]`. F3 and F5 remain resolved from the prior round.

srid:

**All clear** — ran after the lens + codex debates and `/simplify` (elegance pass skipped, since simplify already ran over this tree).

| Pass | Result |
| --- | --- |
| Rules checklist | 0 violations (full checklist walked — no thin wrappers, no dead code, bounded growth via `forget()` + idle expiry, Tailwind-only styling, comments on the non-obvious) |
| Fact-check | 5 items flagged, **all dismissed** — `router.ts` `snapshot` is statically `string` (no untyped yield); the timer↔`forget` ordering has no race (`clearTimeout` precedes any mutation); the snapshot-boundary state machine is sound; `produce` in the idle timer can't throw in practice |

No fixes required.

srid:
## Evidence — live-activity dot (final, dock-only)

Captured on a clean pu box at the PR's HEAD. A terminal streaming output shows the green live dot in its dock row (in place of the "Xs ago" timestamp); idle terminals show their timestamp. The dot was removed from the tile title bar — it's dock-only now.

srid:

The **x86_64-linux** lane ran on **`kolu-ci-2`** (`dev-x86-64-linux-04`) — commit `908ec3b09`, **passed** ✓

- **Lane wall** (pipeline): **2m19s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 1m41s | ✓ |
| `ci::unit` | 29s | ✓ |
| `_ci-setup` | 29s | ✓ |
| `ci::pnpm-hash-fresh` | 16s | ✓ |
| `ci::home-manager` | 14s | ✓ |
| `ci::biome` | 10s | ✓ |
| `ci::fmt` | 9s | ✓ |
| `ci::smoke` | 8s | ✓ |
| `ci::install` | 8s | ✓ |
| `ci::nix` | 6s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/908ec3b/timings.jsonl`; pool state is a live `flock` probe.</sub>

PR #1458, opened 2026-06-20
Make kaval-tui-created terminals appear in kolu live

srid:
**A terminal you spawn from the shell with `kaval-tui create` now shows up in kolu as a tile the moment it's created** — instead of staying invisible until the next kolu restart. kolu and `kaval-tui` were already both clients of one `kaval` daemon (so `kaval-tui list` reaches kolu's terminals); now the visibility runs *both ways*.

### The gap

The daemon owns one inventory of PTYs shared by every client, but kolu's terminal registry was a *second* authority synced to it **only once, at boot** (`adoptSurvivingSession`). The `ptyHostSurface` contract had only *per-terminal* taps — each keyed by an id you already know — plus a one-shot `list`. There was no way for a client to learn that *another* client spawned a PTY, so a `kaval-tui create` against the running daemon fell through the cracks until a restart re-ran adoption.

### The fix — make the daemon's inventory the live source of truth

```
kaval-tui create ─▶ daemon entries.set ─▶ inventory: {created} ─┐
                                                                │  (host-global stream,
kolu-server  ◀── adoptLocalOrphan ◀── startInventoryReconciler ◀┘   snapshot-then-deltas)
                       │
                       └─▶ registerTerminal ─▶ terminalList cell ─▶ tile renders
```

A host-global **`inventory` stream** on `ptyHostSurface` (contract **3.0 → 3.1**, additive · minor): a snapshot of every live PTY, then `created`/`exited` deltas, published from the daemon's two `entries`-map mutation sites over the existing eager-subscribe `Channel`. kolu-server subscribes once and feeds every unknown id into the **existing** `adoptLocalOrphan` path — so the daemon's `entries` map becomes the single source of truth and kolu's registry a continuous projection of it, not a parallel list re-synced at boot.

- **No double-adopt** — kolu's own `spawnPty` registers synchronously *before* the daemon's `created` echoes back, and the reconciler guards on the registry, so the echo is a no-op.
- **`exited` is a no-op** — every terminal kolu tracks already has a per-id `exit` tap that is the single teardown authority; the delta exists for clients that *don't* wire per-id taps.
- **Survives recycles** — the subscription re-subscribes across a daemon restart; its fresh snapshot re-converges idempotently through the same guard.
- **Fail-loud, not silent** — a 3.0 survivor lacking the stream is a hard contract skew → forced recycle, never a quiet downgrade to boot-only.

> _A bare `kaval-tui` shell carries none of kolu's OSC hooks, so its adopted tile shows the live shell + scrollback but no agent/title detection until you run something in it — an inherent ceiling, not a regression._

### Design principles

- **Reuse the source of truth** — the inventory snapshot reuses the same `host.list()` projection the `list` RPC serves; the consumer reuses `adoptLocalOrphan`/`handleExit` verbatim. No parallel tracking, no new transport.
- **Electricity boundary** — cross-process live-set membership surviving reconnect is a hard volatility, encapsulated *inside* the `kaval` package (which already owns the `entries` map and the six per-PTY taps) as a seventh, host-level tap. Dependency arrow points out.
- **Fail fast** — snapshot-then-deltas re-converges on reconnect with no silent loss; a per-event adopt failure rides the existing `killHalfWiredPty` reap rather than collapsing to an invisible PTY.

### Tests

- **Contract corpus** exercises the `inventory` stream over **both** the in-process identity link *and* a real spawned socket daemon — the transport `kaval-tui` actually dials (the coverage ledger enforces it).
- **Host primitive** — `created`/`exited` membership events, eager-subscribe, multi-subscriber fan-out.
- **Pure routing unit** — `inventoryAdoptions` decides adopt/skip/no-op for snapshot/created/exited without the registry or daemon.

Docs synced in the same change: the README architecture table + comms diagram, the kaval-tui README, the changelog, and the **kolu.dev/kaval** diagram (the clients↔daemon arrow is now bidirectional). The `pty-daemon` Atlas note gains **B3.5**.

### Try it locally

```sh
nix run github:juspay/kolu/feat/kaval-inventory-stream#kaval
```


srid:

✅ **Consensus** after 2 round(s) · lowy + hickey · base `908299930fbf`

### Applied (6)
- `lowy-2` Reconciler duplicates bridgeStream's tap-consumption receptacle (per-event fence + abort-as-teardown) — commit `60d66e823`
- `lowy-3` `onConnected` interface name + never-aborted signal expose a wire that encapsulates no volatility — commit `e5cac01d9`
- `lowy-5` Three near-identical `nextFrame`/`raceNext`/`frameUntil` stream-pull helpers across test files — commit `cfdb2f552`
- `hickey-3` Out-of-band adoption emits no client 'reattached' confirmation, splitting a UX concept the boot path owns — commit `977dd0a21`
- `hickey-4` inventoryAdoptions inlines the inventory-frame discriminator instead of dispatching on it — commit `023007955`
- `hickey-6` isTrackedById casts a raw string to TerminalId at the registry boundary, re-deriving a validation the contract layer already owns — commit `18b7f49cc`

### Agreed — no change (6)
- `lowy-1` inventory feed plugs into the correct receptacle (ptyHostSurface + Channel) — clean primary boundary (packages/kaval/src/ptyHostSurface.ts:250, packages/kaval/src/ptyHost.ts:359)
- `lowy-4` `toWireEntry` / `listEntryOf`: the host→wire projection now lives behind two parallel mappers (packages/kaval/src/ptyHost.ts:367-379 (listEntryOf) and packages/kaval/src/inProcessPtyHost.ts:42-56 (toWireEntry))
- `lowy-6` Inventory `created`/`exited` publish ordering relative to entries-map mutation is asymmetric — verify it's intentional (packages/kaval/src/ptyHost.ts:430-435 (exited, after delete) and :615-619 (created, after full wiring))
- `hickey-1` Adoption-routing decision fragmented across two reconcilers with no shared partition (packages/server/src/terminalEndpoint/inventoryReconcile.ts:80 (inventoryAdoptions) + packages/server/src/reconcile.ts:60 (reconcile))
- `hickey-2` Membership-feed snapshot reconstitution split between host.list() and subscribeInventory() (packages/kaval/src/inProcessPtyHost.ts:205-218 (inventory source) + packages/kaval/src/ptyHost.ts:198-200 (PtyInventoryEvent) / :744)
- `hickey-5` Reconciler reimplements an abortable delay rather than reusing a shared primitive (packages/server/src/terminalEndpoint/inventoryReconcile.ts:118-130 (delay))

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `908299930fbf`

### Round 1

**codex** — approved: `false`

The inventory stream itself is shaped reasonably, but the server-side reconciliation has two policy gaps: it can leave valid daemon PTYs hidden when their IDs are outside Kolu's UUID domain, and live adoptions do not participate in the normal session autosave path. I also found one new test that races an immediate process exit and can flake under load.

Findings:
- `F1` · major · open — Malformed orphan IDs are logged and skipped, leaving the daemon connected with live PTYs that Kolu deliberately does not register. The pty-host contract still accepts any string ID, so a non-UUID PTY is a valid daemon state; dropping it violates the boot adoption fail-closed invariant and can leave hidden processes running outside the registry/session. The live inventory path has the same drop behavior at inventoryReconcile.ts:138. (packages/server/src/terminalEndpoint/reattach.ts:75)
- `F2` · major · open — Live inventory adoption reuses `adoptLocalOrphan`, whose `adoptTerminal` path intentionally does not emit `terminalsDirty` because boot adoption saves explicitly afterward. The live inventory path does not save explicitly, so a kaval-tui-created terminal can appear as a tile but never enter the saved session unless some later persisted metadata change happens. (packages/server/src/terminalEndpoint/inventoryReconcile.ts:180)
- `F3` · minor · open — The new `exited` inventory test spawns `/bin/sh -c 'exit 0'` and only subscribes to inventory after `spawn` returns. That process can exit before the subscription is registered, in which case the `exited` event is missed and the test times out. (packages/kaval/src/ptyHost.test.ts:447)

**claude** — Addressed all three of CODEX's findings on the kaval inventory stream branch and committed as debate round 1 (SHA 9368c28ae). F1 and F2 are genuine fail-open/state-loss bugs that I fixed; F3 is a real test race that I fixed. I partially disputed CODEX's framing on F1 (a non-UUID id is an anomaly to kill, not valid state to preserve, and I rejected the "validate at the spawn contract" option as a layering violation — but took CODEX's other listed option, killing the PTY). Touched-package unit tests (server: 108, kaval ptyHost: 29) and tsc both pass; biome format applied no fixes to my code. One pre-existing import-sort lint warning in local.ts and one timing-flaky session.test.ts:305 test are unrelated to these changes (verified via git stash).

- `F1` **partial** — Agreed on the bug, disputed the framing. Both the boot reconcile (reattach.ts:75) and the live inventory boundary (inventoryReconcile.ts onInvalidId) logged-and-dropped non-UUID ids, leaving a live PTY kolu can neither show nor kill — a fail-open that contradicts the boot recycle's own fail-closed invariant (index.ts:164-174). Fixed by failing CLOSED: added a shared reapUnrepresentablePty (local.ts) that KILLS the unrepresentable PTY via the contract's opaque-string kill RPC, wired into both paths. DISPUTED CODEX's claim that a non-UUID is 'valid daemon state' to preserve: every real client mints crypto.randomUUID() (kolu-server's terminals.ts and kaval-tui's newPtyId in create.ts:153), and TerminalIdSchema is z.string().uuid() (terminal-awareness/src/schema.ts:41), so a non-UUID is an anomaly outside kolu's domain — killing is the correct fail-closed response, not adoption. Also REJECTED CODEX's first suggested option (reject non-UUID at the pty-host spawn contract): the contract doc (ptyHostSurface.ts:35-38) deliberately keeps PtyIdSchema = z.string() so kaval carries no kolu domain schema; rejecting at spawn would violate that layering. Took CODEX's other listed option (kill/recycle) instead.
- `F2` **fixed** — Agreed and fixed. Live inventory adoption reused adoptLocalOrphan -> adoptTerminal (local.ts:421), which intentionally skips emitTerminalsDirty because the boot path calls saveSession(snapshotSession()) explicitly afterward (reattach.ts:113). The live path (inventoryReconcile.ts) never saves, and autosave is armed only by terminalsDirtyChannel (session.ts:154-168), so an out-of-band tile rendered but never entered the saved session until a later dirtying event — a kolu-server restart in that window lost it. Added adoptLocalInventoryOrphan (local.ts) which adopts AND emits terminalsDirty, scheduling the same debounced save a fresh spawn does, while preserving the intentional setAdoptedCount/restart-toast silence. Refactored applyEvent into an injectable dispatchInventoryFrame and added 4 regression tests (inventoryReconcile.test.ts) pinning that every untracked created/snapshot entry is dispatched to the persisting adopt fn, malformed ids route to onInvalid (never adopt), and exited adopts nothing. Note on CODEX's exact test ask: a direct unit test of adoptLocalInventoryOrphan's terminalsDirty emit is not viable without a connected-daemon seam (the sensor-wiring path throws synchronously through the facade with no daemon, and no such test seam exists in server unit tests), so the regression is pinned at the deterministic dispatch-routing boundary plus a documented cross-reference to the persisting adopt fn.
- `F3` **fixed** — Agreed and fixed. The exited inventory test (ptyHost.test.ts:445) spawned /bin/sh -c 'exit 0' then subscribed to inventory AFTER spawn. Confirmed via channel.ts that the Channel does not replay history to late subscribers, and via ptyHost.ts:435 that exited publishes on teardown when proc.onExit fires — so if the child exited before the subscribe line, the exited event published to an empty channel and was dropped, timing out nextFrame. Rewrote to subscribe BEFORE spawn, spawn a long-lived sleep 5, consume the created frame, then host.kill(id) to make the exited deterministic — mirroring the already-correct created test (ptyHost.test.ts:424) and the contract corpus inventory test.

commit: `9368c28ae93eb286ea965225fb8cf3e3b3770327`

### Round 2

**codex** — approved: `true`

Claude's round-two changes adequately address the three findings from the prior pass. The non-UUID orphan path now fails closed by routing invalid wire IDs to a shared reap helper from both boot and live inventory reconciliation, live inventory adoption now uses a persisting adoption function that arms the session autosave, and the inventory exit test now subscribes before spawning and drives exit deterministically.

Findings:
- `F1` · major · resolved — Resolved. Non-UUID orphan PTYs are no longer logged and dropped; both boot reconciliation and live inventory route invalid IDs to `reapUnrepresentablePty`, which kills the opaque PTY ID through the pty-host contract instead of leaving a hidden process outside Kolu's registry. (packages/server/src/terminalEndpoint/local.ts:823)
- `F2` · major · resolved — Resolved. Live inventory adoption now goes through `adoptLocalInventoryOrphan`, which calls the normal adoption path and then emits `terminalsDirty`, so the debounced session save is armed for out-of-band PTYs. (packages/server/src/terminalEndpoint/local.ts:802)
- `F3` · minor · resolved — Resolved. The test now subscribes to inventory before spawning, consumes the `created` frame, then kills a long-lived process to assert the `exited` frame, removing the race where an immediate child exit could publish before subscription. (packages/kaval/src/ptyHost.test.ts:445)

_codex on the rebuttal:_ F1: I accept the layering argument. My previous phrasing that a non-UUID ID was valid daemon state to preserve was too broad for Kolu's side of the boundary: kaval may keep PTY IDs opaque, but Kolu's registry cannot represent non-UUID terminal IDs. Keeping validation out of the pty-host spawn contract is consistent with that boundary, and killing the unrepresentable PTY through the opaque-string kill RPC resolves the fail-open bug I raised. F2: fixed; the live inventory path is now wired to the autosave-arming adoption function. The dispatch-level tests plus the direct implementation in `adoptLocalInventoryOrphan` are enough for this review. F3: fixed; the rewritten test removes the timing race.

srid:

✅ **Clean** · base `908299930fbf` · `--no-elegance` (the `/simplify` track already ran this gauntlet)

| Pass | Findings |
| --- | --- |
| Rules (23 built-in + 15 project) | 0 |
| Fact-check (logic · error paths · races) | 0 |

Spot-checks that passed: the pre-subscribe `try/catch` correctly fences the synchronous `liveClient()` throw before `bridgeStream` runs (no `unhandledRejection`); `assertNever(ev)` is a sound compile-time exhaustiveness guard; invalid wire ids fail **closed** through `reapUnrepresentablePty` (no hidden live PTY); all caught errors log or re-subscribe (no collapse-to-empty); the inventory `Channel` publishes only on spawn/teardown (no unbounded growth); the 3.0→3.1 bump is a hard skew, never a silent degrade.

srid:

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

srid:
## Evidence

A `kaval-tui create` from the shell makes a new tile appear in kolu live — no kolu-side action required. Before this PR the terminal would only appear after a kolu restart.

**Before** (empty canvas):

before

**After** (terminal tile materialized live after `kaval-tui create`):

after

**GIF** (canvas transitions from empty → terminal tile live):

evidence

HD video: [evidence-hd.mp4](https://github.com/juspay/kolu/releases/download/pr1458-evidence/evidence-hd.mp4) · player: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=evidence-hd.mp4

**Test setup:** dev kolu launched via `just dev-auto` (port 35563, with `KOLU_KAVAL_BIN` unset so kaval runs from branch source at contract 3.1). `kaval-tui create --socket /run/user/1000/kolu-dev-35563/pty-host.sock` was run from the shell. The inventory reconciler (new in this PR) picked up the `3246846d · bash` PTY from the `inventory` stream and the tile appeared on the canvas within ~1s — no kolu restart, no user action in the browser.

PR #1459, opened 2026-06-20
Revamp Atlas SVG diagrams: semantic colour, shape & proportion

srid:
The hand-authored SVGs across the Atlas were mostly **plain boxes-and-arrows** — they showed topology but not *meaning*. This completes the **"Revamp Atlas SVG diagrams"** Sundry item: re-author them so colour, shape, proportion, and emphasis each carry a fact, so a diagram lands its note's core insight at a glance.

Every diagram was re-authored against one shared **design brief** for coherence: a semantic palette mapped to the Atlas tokens (brand-blue = neutral structure, green = shipped/safe, amber = the new/proposed thing, warm-red = cost/failure, grey-dashed = removed); a shape vocabulary (**cylinder** = on-disk store, **stadium** = transport/bus, **notched box** = a default-deny gate); proportion used as data; and the note's *subject* given the visual weight.

## What changed

**Revamped — plain → semantic (9 notes, 12 diagrams):**

| Diagram | The insight it now carries |
| --- | --- |
| `surface-app` multiplex | Colour = **surface identity**: each lineage (surfaceApp / admin) keeps its hue down through the keying layer and converges into one green transport **stadium** — "independent surfaces, one keyed wire" (was flat all-blue). |
| `surface-mcp` seam | **Ownership split**: a brand-blue framework-owned spine vs an amber consumer-curated **gate** (notched, default-deny). Also fixes a latent `id="arrow"` collision with the diagram below it (ids now namespaced per diagram). |
| `surface-mcp` projectSurface | Polish + the same id-collision fix; amber = the new primitive, green = the observer-safe guarantee. |
| `llm-autonomy` enforcement | The backbone now reads as a **clockwise control loop** — write the spine → read → block & auto-resume — replacing the grey human gate. |
| `kaval-heap-oom` (×2) | A green "diagnostics reach" rail that **stops one hop above the crash**, a visible strip/cut, and the on-disk transcript log drawn as a **cylinder**. |
| `kaval-vs-zmosh` (×3) | A roaming-journey timeline (the 20-min offline gap), a 3-way recovery compare, and CidViz polish (+ wrapped an off-canvas caption). |
| `kaval-sessions` proc | Colour = **process lifecycle**: durable kaval (green, survives restarts) vs the re-running watcher (amber). |
| `performance` surfaces | A left-rail **spine** through CLIENT → WIRE → BACKEND with crisp shipped/to-tune status encoding. |
| `dock-repo` colour-axes | Proportion makes the *starved* repo sink vs the *branch rainbow* visceral — grounded in the note's real example rows. |

**Left as-is (already strongly semantic — a revamp would only regress them):** the mobile "circuit-board" electricity diagram and the `meta` Atlas-pipeline. Both were reverted after the pass confirmed the originals were stronger.

## How it was built & verified

- An **ultracode workflow** fanned out one agent per file (re-author → adversarial critic-repair), all sharing the brief, then I did a **visual pass** over every diagram in the built `dist/` via real Chrome (chrome-devtools), fixing what pixels revealed that static analysis couldn't (a viewBox overflow, an off-canvas caption).
- **Faithfulness audited** against each git original — every label, file path, PR/issue number, and metric is preserved; the one fabricated bit (placeholder branch names) was replaced with the note's real rows.
- `dist/` rebuilt and committed; `just atlas::check-sync` passes (idempotent, host-independent). Removes the now-done Sundry row.

Evidence (before/after) in a comment below.


srid:
## Evidence

Before / after of five representative diagrams, rendered in the committed `dist/` (real Chrome). All 14 diagrams across the changed notes were visually verified; these are the most illustrative transformations.

### `surface-app` — flat all-blue boxes → colour-as-identity multiplex
Each surface lineage keeps its own hue down through the keying layer and converges into one green transport **stadium** (w1/w2 co-tenants).

| Before | After |
| --- | --- |
|  |  |

### `surface-mcp` seam — flat boxes → ownership split + default-deny gate
Framework-owned spine (blue) vs consumer-curated **gate** (amber, notched). Also fixes a latent `id="arrow"` collision with the diagram below it on the same page.

| Before | After |
| --- | --- |
|  |  |

### `llm-autonomy` — linear flow → clockwise control loop
The enforcement backbone now reads as a loop: write the green spine → read → block & auto-resume.

| Before | After |
| --- | --- |
|  |  |

### `kaval-heap-oom` — neutral chain → "diagnostics die one hop above the crash"
A green diagnostics-reach rail that stops at the server, a visible strip/cut, and the red glowing BLIND OOM site.

| Before | After |
| --- | --- |
|  |  |

### `dock-repo` colour-axes — equal boxes → proportion = data
The starved repo sink stays tiny/cool; the branch axis carries an actual rainbow swatch feeding stacked per-branch rows (the note's real example branches).

| Before | After |
| --- | --- |
|  |  |

PR #1463, opened 2026-06-20
fix(pty): don't leak kolu's KOLU_* env into hosted terminals

srid:
## What

kolu's `cleanEnv()` production passthrough copied the **entire** server environment into every terminal it spawns — including kolu's own internal config that the nix wrapper bakes in: `KOLU_KAVAL_BIN` and the kaval identity vars `KAVAL_BUILD_ID`/`KAVAL_COMMIT_HASH`. This strips kolu's internal env at that boundary so none of it reaches a hosted shell.

## The bug

Run `just dev` **inside a kolu terminal** and the dev server dies at boot:

```
DaemonContractSkewError: pty-host contract skew: kaval speaks 3.0, server needs 3.1
kaval endpoint failed to come up at boot
```

## Root cause

A production kolu's nix wrapper bakes its internal config into the server's env (`default.nix:333-338` — `KOLU_KAVAL_BIN`, `KAVAL_BUILD_ID`, `KAVAL_COMMIT_HASH`, …). `cleanEnv()` then forwarded the whole env — `{ ...process.env }` — into every PTY child (`packages/integrations/pty/src/shell.ts`). So any shell opened inside kolu silently carried `KOLU_KAVAL_BIN`, pointing at *that build's* (older) kaval.

Run `just dev` from source in such a terminal and the new (contract 3.1) server reads the inherited `KOLU_KAVAL_BIN` (`localDriver.ts:85`) and spawns the **old** (3.0) kaval. The handshake rejects it (correctly — fail fast, no silent downgrade), and because the respawn re-reads the same stale pointer, it never converges.

## Fix

`cleanEnv()` now drops kolu's internal env before it reaches the shell, via a named `isKoluInternalEnvKey` predicate:

- the **`KOLU_*` namespace** — covers the spawn-hijack surface (`KOLU_KAVAL_BIN` / `KOLU_KAVAL_SOCKET` / `KOLU_KAVAL_SPAWN`);
- the two **wrapper-baked kaval identity vars** `KAVAL_BUILD_ID` / `KAVAL_COMMIT_HASH` — a from-source kaval reads these from inherited env (`buildId.ts`), so leaking the *outer* build's identity into a nested `just dev` would mis-report its build and mask the "update pending" nudge.

One shared strip after **both** `cleanEnv` branches, so the invariant holds regardless of the dev whitelist's contents. Deliberately **not** a blanket `KAVAL_*` strip — a user's own `KAVAL_*` and kaval-tui's `KAVAL_AGENT_DRVS_JSON` survive (the explicit two-var set mirrors `default.nix`'s `koluBin` `--set` manifest exactly). kolu owns this env; a `KOLU_*` a user sets in their *own* dotfiles is re-applied by the rcfile replay (`prepareShellInit`).

## Why not move config off env entirely?

A larger "read kolu's config from an owned channel, never `process.env`" refactor was considered and rejected: ~8 production reads + e2e test-seam migration + a new config layer, all to close a threat (a *non-kolu* ancestor poking the server's own env) that doesn't occur in this single-tenant codebase. The leak boundary *is* the bug; this is the small fix that closes it.

## Test

`cleanEnv` regressions in `shell.test.ts`: passthrough strips `KOLU_*` **and** the two kaval identity vars while the user's env (`PATH`, user vars, unrelated `KAVAL_*` incl. `KAVAL_AGENT_DRVS_JSON`) is untouched. `kolu-pty` unit suite green (26/26); `just fmt` clean.

## Review

Ran the `/be` gauntlet — lens-debate (lowy ⇄ hickey), codex-debate (`xhigh`), simplify, code-police. Lens unified the strip across both branches; codex caught the `KAVAL_BUILD_ID`/`KAVAL_COMMIT_HASH` leak (now fixed); simplify + police clean. See PR comments.

### DIFF
--- a/packages/integrations/pty/src/shell.test.ts
+++ b/packages/integrations/pty/src/shell.test.ts
@@ -74,14 +74,16 @@ describe("koluIdentityEnv", () => {
   });
 });
 
-describe("cleanEnv — kolu's own KOLU_* namespace never reaches a hosted shell", () => {
-  // Regression: a production kolu bakes KOLU_KAVAL_BIN (et al.) into its own
-  // env via the nix wrapper. cleanEnv()'s production passthrough would forward
-  // those into every PTY it spawns — so a nested `just dev` (from source, run
-  // inside a kolu terminal) inherited a STALE KOLU_KAVAL_BIN and spawned a
-  // contract-skewed kaval. cleanEnv must strip the KOLU_* namespace. Default
-  // module state (no configureNixShellEnv call) is passthrough mode — the exact
-  // production path that leaked.
+describe("cleanEnv — kolu's own internal env never reaches a hosted shell", () => {
+  // Regression: a production kolu bakes its internal env (KOLU_KAVAL_BIN et al.,
+  // plus the kaval identity vars KAVAL_BUILD_ID / KAVAL_COMMIT_HASH) into its own
+  // process via the nix wrapper (default.nix). cleanEnv()'s production
+  // passthrough would forward those into every PTY it spawns — so a nested
+  // `just dev` (from source, run inside a kolu terminal) inherited a STALE
+  // KOLU_KAVAL_BIN and spawned a contract-skewed kaval, AND inherited the OUTER
+  // kaval identity so its staleness readout reported the wrong build. cleanEnv
+  // must strip both. Default module state (no configureNixShellEnv call) is
+  // passthrough mode — the exact production path that leaked.
   const saved = { ...process.env };
   afterEach(() => {
     for (const k of Object.keys(process.env)) delete process.env[k];
@@ -107,6 +109,31 @@ describe("cleanEnv — kolu's own KOLU_* namespace never reaches a hosted shell"
     // the user's real environment is untouched.
     expect(env.PTY_TEST_USER_VAR).toBe("keep-me");
     expect(env.PATH).toBe(process.env.PATH);
+  });
+
+  it("strips the wrapper-baked kaval identity vars (KAVAL_BUILD_ID / KAVAL_COMMIT_HASH)", () => {
+    // These are --set onto the kolu wrapper (default.nix:336-337) because kaval
+    // runs in-process there; buildId.ts reads them to derive the staleKey /
+    // "update pending" signal. A from-source kolu nested inside a production
+    // kolu terminal must NOT inherit the outer production identity, or its
+    // staleness readout reports the wrong build and masks the stale-daemon nudge.
+    process.env.KAVAL_BUILD_ID = "outer-production-build";
+    process.env.KAVAL_COMMIT_HASH = "deadbeef";
+    // A user's own KAVAL_* env is NOT internal — only the two baked identity
+    // vars are. KAVAL_AGENT_DRVS_JSON (kaval-tui's drv map) and an arbitrary
+    // user KAVAL_* must still reach the shell.
+    process.env.KAVAL_AGENT_DRVS_JSON = '{"x86_64-linux":"/nix/store/x.drv"}';
+    process.env.KAVAL_MY_OWN_VAR = "keep-me";
+
+    const env = cleanEnv();
+
+    expect(env.KAVAL_BUILD_ID).toBeUndefined();
+    expect(env.KAVAL_COMMIT_HASH).toBeUndefined();
+    // not a blanket KAVAL_* strip — unrelated KAVAL_* survives.
+    expect(env.KAVAL_AGENT_DRVS_JSON).toBe(
+      '{"x86_64-linux":"/nix/store/x.drv"}',
+    );
+    expect(env.KAVAL_MY_OWN_VAR).toBe("keep-me");
   });
 });
 
--- a/packages/integrations/pty/src/shell.test.ts
+++ b/packages/integrations/pty/src/shell.test.ts
@@ -15,8 +15,9 @@ import {
 } from "node:fs";
 import { tmpdir } from "node:os";
 import { dirname, join } from "node:path";
-import { describe, expect, it } from "vitest";
+import { afterEach, describe, expect, it } from "vitest";
 import {
+  cleanEnv,
   koluIdentityEnv,
   OSC2_PRECMD_BASH,
   OSC2_PRECMD_ZSH,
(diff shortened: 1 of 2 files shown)


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `6c953a685f97`

### Applied (2)
- `hickey-1` The "no KOLU_* reaches the shell" invariant is held by two unrelated mechanisms across cleanEnv's two branches — commit `b4fdb63a3`
- `hickey-2` Passthrough branch spreads process.env then imperatively mutates it to filter — what+how braided into a delete-loop — commit `b4fdb63a3`

### Agreed — no change (3)
- `lowy-1` KOLU_* namespace-leak policy lives at cleanEnv but the namespace itself has no single source of truth (packages/integrations/pty/src/shell.ts:101)
- `lowy-2` Asymmetric KOLU_* handling across the two child-spawn boundaries is correct, not a defect — but undocumented at the seam (packages/integrations/pty/src/shell.ts:91)
- `hickey-3` KOLU_* prefix is a domain constant repeated as a bare string literal across server config sites (packages/integrations/pty/src/shell.ts:101)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `6c953a685f97`

### Round 1

**codex** — approved: `false`

The branch correctly targets the PTY env leak at the `cleanEnv()` boundary, but the scrub is incomplete: it removes `KOLU_*` while leaving other wrapper-baked Kolu-family identity env that participates in kaval currency decisions. That keeps a stale-env path alive for nested dev runs.

Findings:
- `F1` · major · open — The new filter only removes `KOLU_*`, but the production `kolu` wrapper also injects `KAVAL_BUILD_ID` and `KAVAL_COMMIT_HASH` into the server environment. Those still get forwarded into every hosted shell, so a nested `just dev` launched from a production Kolu terminal can inherit the outer production kaval identity. In source/dev mode the server and spawned kaval both read those inherited `KAVAL_*` values, which makes the dev instance report the outer production kaval build as its expected/running identity and masks the stale-daemon/update-pending signal. (packages/integrations/pty/src/shell.ts:107)

**claude** — Agreed with CODEX's single finding (F1, major) and fixed it. Verified the leak concretely: default.nix:336-337 bakes KAVAL_BUILD_ID/KAVAL_COMMIT_HASH onto the production kolu wrapper (kaval runs in-process there); buildId.ts:29,34 reads them to derive the kaval staleKey/update-pending signal; localDriver.ts daemonEnv does not forward them, so a from-source kaval (no wrapper) reads them from inherited env. A from-source `just dev` nested in a production kolu terminal would inherit the outer production identity via cleanEnv passthrough (which only stripped KOLU_*) and report the wrong build, masking the stale-daemon nudge — the same wrapper-baked-server-env leak class as KOLU_*. Replaced the bare startsWith('KOLU_') filter with a named isKoluInternalEnvKey predicate: KOLU_* plus an explicit KOLU_INTERNAL_ENV_EXACT set {KAVAL_BUILD_ID, KAVAL_COMMIT_HASH}. Deliberately not a blanket KAVAL_* strip — user KAVAL_* and kaval-tui's KAVAL_AGENT_DRVS_JSON stay. Added a regression test. 26/26 shell.test.ts pass; just fmt clean.

- `F1` **fixed** — Agreed. Confirmed the leak: default.nix:336-337 --sets KAVAL_BUILD_ID/KAVAL_COMMIT_HASH onto the production kolu wrapper because kaval runs in-process; packages/kaval/src/buildId.ts:29,34 reads exactly these to derive staleKey/navigableCommit, consumed by reattach.ts:124-134 and the client's kavalStale 'update pending' nudge; localDriver.ts:127 daemonEnv does NOT forward them, so a from-source kaval (launched via tsx, no nix wrapper) reads them from inherited env. cleanEnv passthrough only stripped KOLU_*, so a from-source `just dev` nested inside a production kolu terminal inherits the outer production kaval identity and reports the wrong expected/running build, masking the stale-daemon signal. Fix in packages/integrations/pty/src/shell.ts: replaced the bare startsWith('KOLU_') filter (now line ~145) with a named isKoluInternalEnvKey predicate (line 67-72) = KOLU_* namespace plus an explicit KOLU_INTERNAL_ENV_EXACT set {KAVAL_BUILD_ID, KAVAL_COMMIT_HASH} (line 62-65). Deliberately NOT a blanket KAVAL_* strip — a user's own KAVAL_* and kaval-tui's KAVAL_AGENT_DRVS_JSON are legitimate user env, so only the two wrapper-baked identity vars are listed by name. Added a regression test in shell.test.ts proving both identity vars are stripped while unrelated KAVAL_* (KAVAL_AGENT_DRVS_JSON, KAVAL_MY_OWN_VAR) and ordinary user env still reach the PTY. 26/26 tests pass.

commit: `3c65b6594b0ad0c7cb6dbff86f73133b5a8f346a`

### Round 2

**codex** — approved: `true`

F1 is resolved in the current tree. The scrub now uses a named predicate that removes all `KOLU_*` keys plus the two wrapper-baked kaval identity vars, while deliberately preserving unrelated `KAVAL_*` env such as `KAVAL_AGENT_DRVS_JSON`; the added regression test covers both sides of that boundary. I verified this by inspecting the current diff and changed files read-only.

Findings:
- `F1` · major · resolved — The previous implementation stripped only `KOLU_*`, allowing wrapper-baked `KAVAL_BUILD_ID` and `KAVAL_COMMIT_HASH` to leak into hosted shells and corrupt nested from-source kaval identity/staleness reporting. (packages/integrations/pty/src/shell.ts:145)

_codex on the rebuttal:_ No dispute from CLAUDE. I verified the fix addresses F1: `cleanEnv()` now filters through `isKoluInternalEnvKey`, the exact set includes `KAVAL_BUILD_ID` and `KAVAL_COMMIT_HASH`, and the regression test asserts those are removed while `KAVAL_AGENT_DRVS_JSON` and a user-defined `KAVAL_MY_OWN_VAR` remain.

srid:

Ran **Pass 1 (rules)** + **Pass 2 (fact-check)**, scoped to the change vs `6c953a685f97`. **All clear** for the in-scope diff — the `cleanEnv` strip, the `isKoluInternalEnvKey` predicate + `KOLU_INTERNAL_ENV_EXACT` set, and the two regression tests. Elegance pass skipped (`--no-elegance`): `/simplify` already ran this gauntlet.

Two flags surfaced but are **out of scope** (pre-existing, not in this diff): `execFileSyncBoth`'s bare `catch {}` and `env.SHELL ??= loginShell` (an empty-string `SHELL` wouldn't be replaced). Both predate this change.

srid:
## Evidence

This fix has no on-screen UI surface — its observable effect is the *contents* of a hosted shell's environment. Demonstrating it faithfully needs the **production passthrough** path of `cleanEnv()` (a dev `just dev` uses the whitelist branch, where the user-var contrast can't show). So the artifact below is a trace of the **actual shipped `cleanEnv()`** (this branch's source) run in production passthrough on a simulated production-kolu server env — the exact code path the fix changes. The deterministic unit regressions in `shell.test.ts` (26/26 green) cover the same.

Before this PR, production passthrough returned `{ ...process.env }` unchanged, so every line below rode straight into the terminal — that is the leak.

```text
PR #1463 — what a hosted kolu terminal receives after cleanEnv()

  kolu-internal (must NOT leak):
    KOLU_KAVAL_BIN         STRIPPED
    KOLU_STATE_DIR         STRIPPED
    KOLU_COMMIT_HASH       STRIPPED
    KAVAL_BUILD_ID         STRIPPED
    KAVAL_COMMIT_HASH      STRIPPED

  user / non-internal (must survive):
    KAVAL_AGENT_DRVS_JSON  kept  → {"x86_64-linux":"/nix/store/x.drv"}
    MY_PROJECT_TOKEN       kept  → keep-me
    PATH                   kept  → /…/node_modules/.bin:/nix/store/…  (full PATH preserved)

  result: PASS — every kolu-internal var stripped, every user var kept
```

- `KOLU_KAVAL_BIN` stripped → the contract-skew fix: a nested `just dev` no longer inherits a stale kaval path.
- `KAVAL_BUILD_ID` / `KAVAL_COMMIT_HASH` stripped → the identity fix: a nested from-source kolu no longer mis-reports the outer build, so the "update pending" nudge isn't masked.
- `KAVAL_AGENT_DRVS_JSON` + user vars kept → it's a scoped strip of kolu's own namespace, not a blanket `KAVAL_*` wipe.

srid:

The **x86_64-linux** lane ran on **`kolu-ci-8`** (`dev-x86-64-linux-08`) — commit `3c65b6594`, **passed** ✓

- **Lane wall** (pipeline): **2m59s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 2m12s | ✓ |
| `ci::home-manager` | 1m28s | ✓ |
| `ci::smoke` | 50s | ✓ |
| `ci::nix` | 49s | ✓ |
| `ci::flake-check` | 44s | ✓ |
| `ci::unit` | 43s | ✓ |
| `ci::atlas-sync` | 41s | ✓ |
| `_ci-setup` | 36s | ✓ |
| `ci::surface-example-build` | 27s | ✓ |
| `ci::surface-app-example-build` | 27s | ✓ |
| `ci::biome` | 26s | ✓ |
| `ci::fmt` | 25s | ✓ |
| `ci::install` | 11s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/3c65b65/timings.jsonl`; pool state is a live `flock` probe.</sub>

PR #1466, opened 2026-06-20
Sleep a terminal, wake it later — on immutable terminal records

srid:
**You can now put a terminal to sleep** — its PTY, xterm, and any running agent are released (gone, like closing it) — while the terminal stays on your canvas as a dormant, moonlit ☾ tile that remembers exactly where it was. **Wake** it later and it respawns in place with its agent resumed. Phase 2 of the sleeping-terminals plan (plan-of-record merged in #1461).

### The model — immutable records, mint-not-mutate

A terminal is `active | sleeping` (the Phase-1 sum). A transition never mutates `state` in place; it **mints a fresh record (new id) from the persisted base and retires the predecessor**. That buys the win: **wake = restore-one** — it reuses the existing `handleRestoreSession` spawn+resume path verbatim (which already re-mints an id and resumes the agent), and **sleep is its inverse** (persist the sleeping record, *then* release the PTY — persist-before-kill).

The one load-bearing seam is the **wire, not the type**: presence is a runtime fact. The client subscribes to terminal metadata keyed off the server's id list, which was active-only — so this PR makes that the **merged active ∪ sleeping** list (`mergedTerminalList()`, shared by the cell snapshot *and* the push so a reconnect can't drop a sleeping tile). A sleeping record reaches every presence surface only because it rides that list.

### Decisions (from the plan)

| Decision | Behavior |
| --- | --- |
| **Manual only** | ☾ Sleep button on the tile title bar · Sleep/Wake palette command · a discoverability tip. No global keybind, no auto-sleep. |
| **Focus-frozen** | Navigating onto a sleeping tile *focuses* it (it can be the active tile) but never wakes it; `focusedId→null` and the right-panel/inspector/theme fall back to the frozen view + an explicit Wake. |
| **Reuse-confirm** | Closing a sleeping tile routes through the existing close-confirm, reworded *"discard sleeping terminal"* — it drops the record (no PTY to kill). |

### Non-negotiables enforced

- **First-class on every surface**, keyed on `state === "sleeping"` and **decoupled from the staleness/"parked" vocabulary** — the dock's parked filter (which *drops* stale rows) now checks sleeping *before* it, so a long-slept terminal is never filtered off.
- **Survives a restart** — sleeping records seed AS sleeping on **every** boot (`seedSleepingRecords`, run before the endpoint adopts), not only the surviving-daemon path, so a cold/full restart doesn't lose them. *(This fixed a real data-loss gap: the first cut seeded only on the surviving path.)*
- **Tolerate a malformed record** — dropped per-record at the read boundary (`tolerateSleepingRecord`), never a fatal validator on the whole session.
- **WebGL budget** narrowed to the active arm (never a GPU context for a frozen tile); **session-save merges** sleeping (sleeping the last terminal no longer clears the session).

### Design philosophy

- **Fail-fast / no fallbacks** — no `?? "active"` coalescing; reads narrow through `activeArm`/`sleepingArm`. A malformed record is *dropped by policy* (the `persisted-schema-stays-tolerant` rule), not a swallowed error.
- **Electricity boundary** — the sleeping-record store is a sibling leaf module in `kolu-server`, not a field on the active registry (which stays `ActiveTerminal`-only by construction).
- **Reuse the source of truth** — wake reuses `handleRestoreSession` rather than a parallel respawn; the base-picker is schema-driven (`toSavedSleeping` parses through the saved schema, so a future persisted field rides for free).

### Verification

`just check` (typecheck + biome) and `just fmt` are green; client + server unit suites pass. The e2e (`features/sleeping-terminals.feature`, five failure-mode scenarios) is **verified on CI**, not this dev box — the box has a stale **kaval 3.0** daemon while master's contract is **3.1**, so a rebooted local server can't bring the daemon up (`DaemonContractSkewError`). A clean CI box builds kaval 3.1 and runs them.

> Surface-package READMEs only; no `@kolu/surface` API change, so no drishti companion PR.


srid:

✅ **Consensus** after 2 round(s) · lowy + hickey · base `6c953a685f97`

Independent findings: lowy=6, hickey=8

### Applied (10)
- `lowy-1` Liveness receptacle (isSleeping/sleepingArm) bypassed by ~8 raw state reads — commit `f4587c820`
- `lowy-3` Sleeping-tile layout is read-through but write-dropped — an asymmetric half-receptacle — commit `88d350f15`
- `lowy-4` renderTerminalTile and renderSleepingTile are near-identical twins — the variation isn't a tile-shell axis — commit `566bf295a`
- `hickey-1` renderSleepingTile is a near-verbatim clone of renderTerminalTile — commit `566bf295a`
- `hickey-2` "sleeping" is encoded two incompatible ways across sibling dock models — commit `b8f43f7c3`
- `hickey-3` "sleep is not staleness" invariant restated independently in 3 places — (uncommitted)
- `hickey-4` CanvasTile data-state/data-dimmed decoupling is asserted but not realized — commit `566bf295a`
- `hickey-6` Wake fabricates an active SavedTerminal by overwriting the discriminant — commit `241a79213`
- `hickey-7` resumeAgentCommand computed twice per restored terminal — commit `490f7a923`
- `hickey-8` data-state="sleeping" presence bit re-derived from a different source on each of 4 surfaces — (uncommitted)

### Agreed — no change (4)
- `lowy-2` sleeping-store is the symmetric sibling of terminal-registry, not a duplicated receptacle (packages/server/src/sleeping-store.ts:1, packages/server/src/terminal-registry.ts:39)
- `lowy-5` Wake duplicates the active-marker convention that restoreOneTerminal deliberately stays pure of (packages/client/src/terminal/useSessionRestore.ts:203-224)
- `lowy-6` useSessionRestore complects four independent lifecycle concerns behind one hook (packages/client/src/terminal/useSessionRestore.ts:30-413)
- `hickey-5` Wake is exposed as two independent affordances wired to the same handler (packages/client/src/canvas/DormantTileBody.tsx:46-58 (Wake button) and packages/client/src/canvas/TileTitleActions.tsx:185-200 (Wake button) and packages/client/src/commands.tsx:255-260 (Wake command))

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 round(s) · codex reviewed at `xhigh` reasoning effort · base `6c953a685f97`

### Round 1

**codex** — approved: `false`

The sleeping-terminal feature is not ready to approve. The branch has several lifecycle bugs around persistence, active-id replacement, cleanup, and malformed persisted data; the most serious ones can lose a terminal on crash, leak hidden PTYs, or leave stale sleeping records that reappear after close/remove flows.

Findings:
- `F1` · blocking · open — `sleepTerminal` claims it persists before killing, but `putSleeping(record)` only updates the in-memory sleeping map. The actual durable session save is still the debounced autosave triggered later by `killTerminal`, so a server crash after the PTY is killed but before the debounce flush loses both the live terminal and the sleeping record. This violates the feature's core durability guarantee. (packages/server/src/terminalEndpoint/local.ts:721)
- `F2` · blocking · open — Sleeping a terminal only sleeps the selected terminal. Split/sub-terminal PTYs under it are not killed or otherwise cleaned up; after the parent is removed from the live list, those children become hidden because their parent no longer exists in the top-level view, but the processes keep running. The changelog says splits are closed when sleeping, so this is both a leak and a user-visible contract violation. (packages/client/src/terminal/useTerminalCrud.ts:205)
- `F3` · major · open — `killAllTerminals` only drains live terminal state and never clears the sleeping store. After the client's close-all flow resets local UI, sleeping records remain on the server and can reappear on the next terminal-list snapshot or reload. The test `Before` hook also uses this RPC, so sleeping records can leak across scenarios. (packages/server/src/terminalEndpoint/local.ts:738)
- `F4` · major · open — The client ignores the `SavedSleepingTerminal` returned by `terminal.sleep`. The server mints a new sleeping id, then kills the old terminal id; the active id therefore remains pointed at a removed terminal. In maximized mode this can leave every remaining tile treated as covered because no visible tile matches the active id, and command routing can target the stale id. (packages/client/src/terminal/useTerminalCrud.ts:206)
- `F5` · major · open — The close dialog still offers the worktree-removal path for sleeping terminals, but `onCloseAndRemove` always calls `handleKillWorktree`. For a sleeping id that eventually calls `terminal.kill`, which the server reports as not found and the client swallows; no `discardSleeping` happens. The worktree can be removed while the dormant tile persists or reappears pointing at a deleted path. (packages/client/src/App.tsx:379)
- `F6` · major · open — Wake creates the new live terminal first, then catches and only toasts `discardSleeping` failures. If the discard RPC fails, the old sleeping record remains and can be woken again, creating duplicate live terminals from the same dormant record while the UI treats the wake as successful. (packages/client/src/terminal/useSessionRestore.ts:226)
- `F7` · major · open — `SavedSleepingTerminalSchema` accepts any string id via `SavedTerminalIdSchema`, but sleeping records are surfaced as terminal ids in `terminal.list` and metadata collection keys, both of which require UUID `TerminalId`s. Persisted records like the new test fixture's id `"0"` pass the saved schema but then break the live surface contract when listed. (packages/common/src/surface.ts:457)
- `F8` · major · open — `seedSleepingRecords` tolerates malformed sleeping records only while seeding the in-memory sleeping store; it leaves the malformed records in the persisted session cell. Clients subscribing to `session.get` still receive the raw invalid session and can fail validation despite the malformed-record tolerance path. (packages/server/src/terminalEndpoint/reattach.ts:60)
- `F9` · minor · open — When `activeId` is present but metadata is missing or stale, the command palette falls through to showing `Wake terminal` because `activeArm(undefined)` is false. Selecting it then targets a non-sleeping or missing record and no-ops. (packages/client/src/commands.tsx:248)
- `F10` · minor · open — `TileTitleActions` now receives `onSleep` and `onWake` through props even though the file-level contract says the component reads singleton state directly and only requires an id. This also pushes domain-specific sleep/wake plumbing into `App`, making the shell thicker for a tile-local action. (packages/client/src/canvas/TileTitleActions.tsx:32)
- `F11` · nit · open — The close button tooltip is always `Close terminal`, even for sleeping tiles where the action opens a discard confirmation rather than closing a live terminal. (packages/client/src/canvas/CanvasTile.tsx:377)
- `F12` · nit · open — The comment says sleeping tiles keep the original terminal id, but the server currently creates a fresh UUID when sleeping. That stale note is misleading for active-id, layout, and wake behavior. (packages/client/src/tile/tileContent.ts:17)

**claude** — Addressed all 12 codex findings on the sleeping-terminal feature: fixed 11 (F1-F9, F11, F12), partial on F10. Every finding conceded, grounded in the feature plan-of-record (Atlas non-negotiables and immutable-records id model) and the UUID-keyed surface contracts. F1 went through a new flushSessionNow in session.ts to avoid a terminals/local import cycle biome rejects. Typecheck across 48 packages plus 678 unit tests pass; biome lint and fmt-check clean. e2e could not run on this host (a live kolu.service squats the harness ports); it runs in CI. Committed 56b0b6c5b.

- `F1` **fixed** — persist before kill via flushSessionNow
- `F2` **fixed** — cascade close splits on sleep
- `F3` **fixed** — killAll drains sleeping store
- `F4` **fixed** — switch active to new record id
- `F5` **fixed** — discard sleeping in worktree removal
- `F6` **fixed** — wake rolls back on discard failure
- `F7` **fixed** — uuid id for sleeping schema and fixture
- `F8` **fixed** — rewrite session cell after dropping malformed
- `F9` **fixed** — palette hides sleep wake when meta absent
- `F10` **partial** — dropped onSleep prop; kept onWake with rationale
- `F11` **fixed** — sleeping-aware close tooltip
- `F12` **fixed** — fixed stale id-lifecycle comments

commit: `56b0b6c5b8a134a9935e712ceb8a3a5c752fcddf`

### Round 2

**codex** — approved: `false`

Most of the first-round fixes are in place, but I cannot approve yet. The persistence fix still has a log-and-continue failure path, the split cleanup path can still leak hidden PTYs on kill failure, sleeping worktree removal can still delete the worktree after a failed discard, and the App/TileTitleActions wake wiring remains the same boundary violation CLAUDE only partially addressed.

Findings:
- `F1` · blocking · open — The new synchronous flush closes the original debounce hole only when `snapshotProvider` is present. If it is absent, `flushSessionNow` logs and returns, and `sleepTerminal` continues on to kill the PTY. That is not fail-fast, and it recreates the original data-loss class under a boot-order/regression path. (packages/server/src/session.ts:158)
- `F2` · blocking · open — The sleep path now tries to close splits, but it does so through `handleKill`, which catches every `terminal.kill` failure and removes local state anyway. If a split kill fails for any reason other than already-gone, parent sleep still proceeds and the split PTY remains live but hidden, so the leak can still occur. (packages/client/src/terminal/useTerminalCrud.ts:220)
- `F3` · major · resolved — Verified resolved: `killAllTerminals` now drains the sleeping store, emits the merged list, and flushes when sleeping records were removed. (packages/server/src/terminalEndpoint/local.ts:774)
- `F4` · major · resolved — Verified resolved: `handleSleep` now uses the returned sleeping record id and sets it active, so the client no longer keeps the retired live id selected. (packages/client/src/terminal/useTerminalCrud.ts:228)
- `F5` · major · open — The sleeping worktree path now calls `handleDiscardSleeping`, but that helper catches discard failures and resolves. `handleKillWorktree` therefore proceeds to remove the worktree even when the sleeping record was not discarded, leaving the stale dormant tile pointing at a deleted path. (packages/client/src/terminal/useWorktreeOps.ts:75)
- `F6` · major · resolved — Verified resolved for the finding as raised: wake no longer treats a failed discard as success; it attempts to roll back the newly created terminal and returns without selecting it after surfacing a failure toast. (packages/client/src/terminal/useSessionRestore.ts:231)
- `F7` · major · resolved — Verified resolved: saved sleeping records now require `TerminalIdSchema`, and the malformed-session fixture uses UUIDs for ids so the valid sleeping record can be represented on live terminal surfaces. (packages/common/src/surface.ts:463)
- `F8` · major · resolved — Verified resolved: malformed sleeping records are now filtered and the cleaned session cell is rewritten before clients subscribe, so the raw invalid blob is not left behind in `session.get`. (packages/server/src/terminalEndpoint/reattach.ts:94)
- `F9` · minor · resolved — Verified resolved: the palette now branches active/sleeping/absent metadata explicitly and emits no sleep/wake command when active metadata is missing. (packages/client/src/commands.tsx:166)
- `F10` · minor · open — CLAUDE removed `onSleep`, but `onWake` is still prop-drilled from `App` into tile chrome, and the file header still says the only prop is `id`. The rationale explains why calling `useSessionRestore` locally is not trivial, but it does not resolve the App shell/no-prop-drilling boundary violation. (packages/client/src/canvas/TileTitleActions.tsx:34)
- `F11` · nit · resolved — Verified resolved: the close button title is now sleeping-aware. (packages/client/src/canvas/CanvasTile.tsx:377)
- `F12` · nit · resolved — Verified resolved: the tile id lifecycle comment now matches the minted sleeping-record id model. (packages/client/src/tile/tileContent.ts:17)

_codex on the rebuttal:_ F1: partial fix only. The synchronous flush is the right direction, but `flushSessionNow` cannot log-and-return while the caller continues to kill; that is not fail-fast and leaves a data-loss path. F2: partial fix only. Closing splits through `handleKill` does not prove the PTYs were killed because `handleKill` swallows all kill errors. F3, F4, F6, F7, F8, F9, F11, and F12: verified and accepted as resolved. F5: partial fix only. Routing sleeping worktree removal to discard is correct, but the discard helper swallows failure, so worktree deletion can still proceed with the stale sleeping record intact. F10: I do not accept the rationale as resolving the finding; it explains the current coupling, but the App shell and tile-action boundary are still carrying the wake prop, and the component contract comment is still contradicted by the props.

**claude** — Round 2: addressed all four findings codex held open after round 1 (F1, F2, F5 blocking/major; F10 minor). The other eight (F3, F4, F6, F7, F8, F9, F11, F12) were already verified resolved by codex in round 1 and untouched this round. Conceded all four open ones — each was a genuine fail-fast / no-prop-drilling defect, not a style nit. Validation: `just check` (typecheck all packages + biome lint, clean), `just test-unit` (client 425 + server 110, all pass), `just fmt-check` clean. e2e runs in CI (a live kolu.service squats the harness ports on this host). Committed as ac236c6ea.

- `F1` **fixed** — flushSessionNow (packages/server/src/session.ts:158) now THROWS instead of log-and-return when the snapshot provider is unwired. sleepTerminal (terminalEndpoint/local.ts:739) calls it BEFORE killTerminal, and the sleep router handler (router.ts:205) doesn't catch — so a missing provider propagates as an RPC error and aborts the sleep before the PTY is released, instead of killing into the void. The PTY release now stays strictly behind a confirmed durable write, matching the project's fail-fast 'crash loudly, never silently degrade' rule.
- `F2` **fixed** — handleSleep (packages/client/src/terminal/useTerminalCrud.ts:220) no longer closes splits through handleKill, whose bare catch{} swallowed every kill error and evicted local state anyway. It now awaits client.terminal.kill per split directly inside a try/catch that toasts and RETHROWS on failure — so an RPC-level kill failure aborts the sleep (no hidden live PTY left running once the parent leaves the live list). removeAndAutoSwitch (client-side eviction) runs only after each kill is confirmed. The server returns (no throw) for an already-gone terminal, so any throw here is a genuine failure worth aborting on — exactly codex's first suggested option.
- `F5` **fixed** — handleDiscardSleeping (packages/client/src/terminal/useSessionRestore.ts:249) now toasts AND rethrows on a failed discard instead of swallowing it. handleKillWorktree (useWorktreeOps.ts:69) awaits it before client.git.worktreeRemove, so a failed discard propagates out and the worktree is never deleted while a stale sleeping record still points at the path. The standalone close-confirm caller in App.tsx:371 keeps its fire-and-forget shape via .catch(() => {}) since the verb already surfaced the toast.
- `F10` **fixed** — Did the refactor codex asked for rather than re-arguing the rationale. Singleton-ized useSessionRestore via createSharedRoot (packages/client/src/terminal/useSessionRestore.ts), mirroring useTerminalCrud's documented de-deps — it now reads useTerminalStore/useTerminalCrud internally instead of taking a DI bag. TileTitleActions.tsx and DormantTileBody.tsx call useSessionRestore().handleWake(id) directly; the per-tile onWake prop drilled through App (App.tsx:587, :242) is gone, and the TileTitleActions file-header contract ('the only prop is the tile id') is true again. useTerminals.ts now calls useSessionRestore() with no args; the test seam (useSessionRestore.test.ts) mocks useTerminalStore/useTerminalCrud instead of passing args, and all 40 client test files still pass.

commit: `ac236c6ea270be8c9c1bdac81fbb333a9f20bb24`

### Round 3

**codex** — approved: `true`

All findings from the earlier review are now resolved. I verified the round-2 changes against the current tree: `flushSessionNow` now throws before sleep can release the PTY, split cleanup during sleep now propagates kill failures before sleeping the parent, sleeping worktree removal is gated on a successful discard, and the App-to-tile `onWake` prop drilling is gone via the `useSessionRestore` singleton refactor. No new regression introduced by this round stood out in the scoped review.

Findings:
- `F1` · blocking · resolved — Resolved: `flushSessionNow` now throws when the snapshot provider is missing, and `sleepTerminal` calls it before `killTerminal`, so a missing durable-write path aborts the sleep before PTY release. (packages/server/src/session.ts:163)
- `F2` · blocking · resolved — Resolved: `handleSleep` no longer routes split cleanup through the broad-swallowing `handleKill`; it awaits each split kill and rethrows on failure before sleeping the parent. (packages/client/src/terminal/useTerminalCrud.ts:225)
- `F3` · major · resolved — Resolved: `killAllTerminals` drains sleeping records and flushes when it removed them. (packages/server/src/terminalEndpoint/local.ts:774)
- `F4` · major · resolved — Resolved: sleeping now switches the active id to the returned sleeping record id. (packages/client/src/terminal/useTerminalCrud.ts:242)
- `F5` · major · resolved — Resolved: `handleDiscardSleeping` now rethrows after toasting, and `handleKillWorktree` awaits it before removing the worktree. (packages/client/src/terminal/useSessionRestore.ts:259)
- `F6` · major · resolved — Resolved: wake rolls back the created terminal and does not select it when discarding the sleeping record fails. (packages/client/src/terminal/useSessionRestore.ts:234)
- `F7` · major · resolved — Resolved: saved sleeping ids are validated as UUID `TerminalId`s, with fixtures updated accordingly. (packages/common/src/surface.ts:463)
- `F8` · major · resolved — Resolved: malformed sleeping records are filtered and the cleaned session cell is rewritten before clients subscribe. (packages/server/src/terminalEndpoint/reattach.ts:94)
- `F9` · minor · resolved — Resolved: the command palette now hides sleep/wake when active metadata is absent. (packages/client/src/commands.tsx:166)
- `F10` · minor · resolved — Resolved: `TileTitleActions` is back to an `id`-only prop surface and reads `useSessionRestore` directly; `DormantTileBody` also reads the singleton instead of receiving an App-drilled wake callback. (packages/client/src/canvas/TileTitleActions.tsx:34)
- `F11` · nit · resolved — Resolved: the close tooltip is sleeping-aware. (packages/client/src/canvas/CanvasTile.tsx:377)
- `F12` · nit · resolved — Resolved: the tile id lifecycle comment now matches the minted sleeping-record model. (packages/client/src/tile/tileContent.ts:17)

_codex on the rebuttal:_ F1 accepted: the flush now fails loud and aborts before PTY release. F2 accepted: split cleanup no longer hides failed kills behind local eviction. F5 accepted: discard failure now propagates to block worktree deletion. F10 accepted: the wake verb moved behind the session singleton, removing the App-level per-tile prop. The previously resolved F3, F4, F6, F7, F8, F9, F11, and F12 remain resolved.

srid:

**Pass 1 — rule checklist: clean (0 violations).** All built-in + project rules checked, including `caught-error-must-not-collapse-to-empty`, `persisted-schema-stays-tolerant`, `no-silent-error-swallowing`, and the fail-fast rules. The deliberate `setLayout` no-op for a sleeping arm was reviewed and cleared (the drag/resize affordance is gated off, so the path isn't reached).

**Pass 2 — fact-check: 3 findings → 2 fixed, 1 skipped.**

- **Fixed (`9bdef337c`)** — the wake→discard rollback swallowed a failed rollback-`kill` via `.catch(() => {})`, and the toast implied a clean "try again" even when a live duplicate was orphaned alongside the still-wakeable sleeping record. Now the rollback-kill failure is logged (`console.error`) and the toast tells the user the workspace is inconsistent (restart to recover) rather than inviting a retry that would spawn another duplicate.
- **Skipped (false positive)** — the standalone close-confirm `handleDiscardSleeping(...).catch(() => {})` (App.tsx): that path already toasts; the worktree-removal gate is the *separate* `onCloseAndRemove → handleKillWorktree`, which awaits the discard (codex F5). Documented inline.

Pass 3 (elegance) skipped (`--no-elegance`) — `/simplify` ran this gauntlet.

srid:
## Evidence

**Primary: Sleep → Dormant ☾ tile → Wake → live terminal**



▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=evidence-sleep-wake.mp4

A live terminal is put to sleep via the Sleep command, which transitions it to a dormant moonlit ☾ tile. Waking it respawns a usable live terminal that echoes a marker, confirming the full sleep→wake round-trip.

**Mobile / small-viewport note:** The sleeping tile and dock row ride the same dock/tile machinery with the dimmed ☾ indicator — both the canvas tile and dock row render correctly in the small-viewport layout (the drawer layout used on phone-width viewports).

srid:

The **x86_64-linux** lane ran on **`kolu-ci-3`** (`dev-x86-64-linux-03`) — commit `c7ce7301b`, **passed** ✓

- **Lane wall** (pipeline): **2m53s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 2m10s | ✓ |
| `ci::unit` | 44s | ✓ |
| `ci::smoke` | 43s | ✓ |
| `ci::nix` | 42s | ✓ |
| `_ci-setup` | 34s | ✓ |
| `ci::atlas-sync` | 34s | ✓ |
| `ci::surface-example-build` | 28s | ✓ |
| `ci::biome` | 27s | ✓ |
| `ci::fmt` | 26s | ✓ |
| `ci::install` | 8s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/c7ce730/timings.jsonl`; pool state is a live `flock` probe.</sub>

srid:
Discarding after hands-on testing surfaced two user-facing bugs that both root to the immutable thin-frozen-record model:

1. **Can't drag a sleeping tile** — the frozen record has no layout write sink (`CanvasTile.tsx:103-108`: `setLayout` drops the sleeping arm), so dragging was disabled rather than supported.
2. **Wake doesn't restore the agent** — `toSavedSleeping` strips the agent session as 'live overlay' (`surface.ts:975-976`), so wake has nothing to resume.

Green e2e + full gauntlet missed both: the tests covered invariants (dormant body renders, record survives reboot, malformed dropped), not the user journeys (sleep a Claude session → wake → keep talking; move a dormant tile). Going back to the drawing board — re-planning on a mutable-record model that preserves the agent session and keeps a sleeping tile a first-class canvas citizen. Branch `redesign/sleeping-phase2-impl` is retained for reference.

Issue #818, opened 2026-05-02
Code browser doesn't remember state


srid:
Switch to a state like this:

<img width="434" height="671" alt="Image" src="https://github.com/user-attachments/assets/8c817f62-0c71-47e2-8b54-2fcd680095af" />

Then go to 'Inspector' or close the right panel, and come back to reset code browser:

<img width="566" height="643" alt="Image" src="https://github.com/user-attachments/assets/14f070e6-93b3-4cac-87b0-09070a60d376" />

PR #1468, opened 2026-06-20
perf(client): lazy-load the Code tab off the eager bundle

srid:
**The Code tab welded ~171 kB gzip onto every session's startup bundle.** Its main-thread tree — Pierre's `FileTree`, the `@kolu/solid-markdown` renderer, the diff/source view wrappers, and the comment system — was a static `import` in `RightPanel`, so it downloaded and parsed up front for *every* user, before the Code tab was ever opened. It now lazy-loads: deferred until the Code tab is first shown, then kept mounted (so #818 state preservation across tab switches is unchanged).

This is perf-map **item #5**, and like the last run it started with *measurement* — the note's framing turned out partly wrong, so the win is honest rather than assumed.

### Measured (A/B production build)

| eager `index.js` | deferred |
| --- | --- |
| **745 → 575 kB gzip** (−170 kB / **−23%**) | CodeTab chunk **629 kB / 171 kB gzip** |

So the terminal + chrome paint off a 23%-smaller critical bundle; the Code tab streams in after.

### The fix shape

- **Lazy-keep-alive** (the inverse of the FileView toggle I just shipped): `lazy(() => import("./CodeTab"))`, gated behind a one-way `codeEverShown` latch, then kept mounted. A closed mobile drawer or collapsed desktop panel never loads it at all.
- **Deferred past `onMount` — load-bearing.** The default tab is `code`, so rendering the lazy component during the *first synchronous render* suspended the initial paint and hung the whole app. Flipping the latch after `onMount` lets the first paint complete, then the chunk loads. _(This is the bug the pu-box e2e caught and this fix resolves.)_

### Two note premises the measurement refuted

- **Shiki grammars were already lazy** — a dynamic `import("shiki")`, never on the eager path (and the `@pierre/diffs` diff renderer already runs in a Web Worker). The note's "Shiki always mounted" was wrong.
- **"Lazy-load Image on first use" is mechanically impossible** — `ImageAddon` must be loaded *before* an image escape sequence arrives or the image is dropped. `Serialize`/`Search` are deferrable but minify to ~10–15 kB gzip and add async to the per-terminal hot path — left as low-priority, not shipped.

### Tests

Verified on a **pu box** (not locally): `code-tab.feature` **115/115**. The lazy mount + Suspense + keep-alive all work; the default-`code` initial-render path that hung the app is now green.

### How it honors the design philosophy

- **Fail-fast / no-fallbacks** — nothing degrades; the Code tab is byte-identical, just deferred. The Suspense fallback is a brief load state, not a reduced-functionality path.
- **Volatility boundaries** — code-splitting the Code tab is the tab host's (`RightPanel`'s) own concern; no new domain-agnostic helper, and the keep-alive reuses the existing #818 pattern.
- **Reuse the source of truth** — Solid's `lazy()`/`Suspense` over a hand-rolled loader; measured with the *real* production build, not estimated.

Closes perf-map item #5.


srid:

✅ **Consensus** after 2 round(s) · lowy + hickey · base `6c953a685f97`

### Applied (4)
- `hickey-1` codeEverShown braids three independent axes into one latch — commit `409726550`
- `hickey-2` "code tab is the active tab" derived in two independent places — commit `8684a24ef`
- `hickey-3` Per-slot lazy/Suspense wiring lives in the parent loop, complecting tab-routing with tab-loading-strategy — commit `9969fe7af`
- `hickey-4` import statement split by an interposed const + comment block — (uncommitted)

### Agreed — no change (4)
- `lowy-1` Lazy-load uses the framework receptacle correctly; no parallel encapsulation (packages/client/src/right-panel/RightPanel.tsx:44)
- `lowy-2` Two orthogonal concerns (defer-first-load vs keep-alive) live in one memo (packages/client/src/right-panel/RightPanel.tsx:89-94)
- `lowy-3` Deferral boundary leaks into RightPanel; CodeTab does not own its own load timing (packages/client/src/right-panel/RightPanel.tsx:76-94)
- `lowy-4` Suspense fallback duplicates an ad-hoc loading-state instead of a shared receptacle (packages/client/src/right-panel/RightPanel.tsx:184-187)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `6c953a685f97`

### Round 1

**codex** — approved: `false`

The implementation split itself is mostly behavior-preserving: `CodeTabImpl` matches the old `CodeTab` body aside from the name/header, and the `RightPanel` first-show latch is coherent. I found one real runtime risk introduced by deferring the chunk, plus a docs consistency issue from moving internals out of `CodeTab.tsx`.

Findings:
- `F1` · major · open — The new dynamic import has no failure boundary. A long-lived tab that has not opened the Code tab before a deploy will not have this chunk cached; after the server switches to a new `KOLU_CLIENT_DIST`, `/assets/*` misses intentionally 404, so the old client can request a now-missing CodeTabImpl chunk. `Suspense` covers the pending promise, not a rejected lazy import, and the existing update overlay is click-through, so opening Code afterward can throw an uncaught UI error instead of giving the user the existing reload path. (packages/client/src/right-panel/CodeTab.tsx:16)
- `F2` · minor · open — The branch moved the Code tab implementation internals to `CodeTabImpl.tsx`, but source docs/comments still cite old `CodeTab.tsx` line numbers for those internals. This line still points at `CodeTab.tsx:314-349`; other examples include `solid-browser.mdx` citations to `CodeTab.tsx:515-542`, which now points beyond the 33-line wrapper. The Atlas links become misleading after this PR. (docs/atlas/src/content/atlas/performance.mdx:297)

**claude** — Addressed both codex findings on the Code-tab lazy-split branch. F1 (major): added an ErrorBoundary around the lazy CodeTabImpl import that routes a rejected chunk fetch to the existing reload affordance (useSurfaceApp().reload()), so a post-deploy stale-chunk 404 self-heals instead of throwing an uncaught UI error — Suspense alone only covers the pending promise. F2 (minor): repointed Atlas implementation-internal citations from CodeTab.tsx to CodeTabImpl.tsx with current line numbers across six notes and regenerated dist. tsc passes; atlas build succeeds; biome format reports no fixes.

- `F1` **partial** — Agreed and fixed the core risk: wrapped the lazy chunk in an ErrorBoundary in /home/srid/code/kolu/.worktrees/gpu/packages/client/src/right-panel/CodeTab.tsx. The fallback routes a rejected import to useSurfaceApp().reload() — the SAME path the update overlay (TransportOverlay.tsx) already uses — so a now-404'd CodeTabImpl chunk after a KOLU_CLIENT_DIST swap surfaces a Reload button rather than an uncaught throw (Suspense only catches the pending promise, confirmed this was the only lazy() in the client). Declined the suggested focused test: the client package has no component-render harness (no @solidjs/testing-library; existing tests are hook-level vitest like useRightPanel.test.ts or e2e Cucumber), so simulating a rejected lazy import would mean introducing a parallel testing mechanism for one assertion — against 'reuse the existing source of truth'. The e2e harness can't easily 404 a chunk either.
- `F2` **fixed** — Agreed. Repointed implementation-internal citations from CodeTab.tsx (now a 56-line wrapper) to CodeTabImpl.tsx with verified current line numbers: performance.mdx (three git-status subs 319-363), solid-browser.mdx (handleSelect 635-666 at two sites, the consumer filemap row, and the D2 diagram node), code-tab-filter-empty-dirs.mdx (projectFileTreeSearch 584-586 plus the boundary prose), pref-storm.mdx (vertical-split onSizesChange 886-889), mobile-architecture-review.mdx (density+scroll label 142/844 and tap-target prose), md-preview-wikilinks.mdx (treePaths). Left code-tab-browse-git-status.mdx:23 untouched — its file-tree node with status 'M' depicts that note's own historical PR diff, not a live code citation, so changing it would falsify the record. Ran just atlas::build and staged the six regenerated dist HTML files.

commit: `4c0b91b5e5cdb3186d421ecbe1cd662a3976e174`

### Round 2

**codex** — approved: `true`

The current tree addresses the two prior review findings. The lazy Code tab import is now wrapped in a Solid ErrorBoundary that gives the user an explicit reload path when the chunk import rejects, and the Atlas implementation references have been repointed to CodeTabImpl where they describe live code rather than historical diff content. I did not find any regression introduced by this response round.

Findings:
- `F1` · major · resolved — Resolved. The lazy CodeTabImpl chunk import is now inside an ErrorBoundary, so a rejected dynamic import after a stale-client deployment no longer escapes as an uncaught UI error. (packages/client/src/right-panel/CodeTab.tsx:37)
- `F2` · minor · resolved — Resolved. The live implementation citations that became stale after moving Code tab internals to CodeTabImpl.tsx have been updated across the affected Atlas notes and regenerated dist output. (docs/atlas/src/content/atlas/performance.mdx:297)

_codex on the rebuttal:_ F1: I accept the fix and the test rationale. The issue was the missing rejected-import boundary, and CodeTab.tsx now catches that failure mode and exposes the same reload behavior used by the update overlay. Given the package does not already have a component render harness, I am not holding this open over the lack of a one-off test. F2: I accept the fix and the explanation for the one intentionally unchanged historical reference. The implementation-internal Atlas references now point at CodeTabImpl.tsx, which resolves the stale-citation problem.

srid:

Ran with `--no-elegance` (the simplify pass already covered that tree).

**Pass 1 — rule checklist: clean** across all 30 rules. Notably: the `ErrorBoundary` in `CodeTab.tsx` handles the lazy-chunk rejection explicitly (no silent swallow, no collapse-to-empty); `codeEverShown` is a monotonic latch (no unbounded growth); `props.visible` is read inside the `createMemo` (reactive-prop hygiene); the thin `CodeTab` wrapper is not a thin-wrapper violation (it adds the lazy split + Suspense + ErrorBoundary + reload integration).

**Pass 2 — fact-check: clean, one non-actionable note.** No logic errors. The `codeEverShown` accumulator is correctly monotonic (the `Show` gate never re-unmounts), `mounted()` is correctly tracked, no circular import (`CodeTabImpl` imports `./openInCodeTab`, not `./CodeTab`). Latent note: `useSurfaceApp()` is read just *above* the `ErrorBoundary`, so if `CodeTab` were ever mounted outside `<SurfaceAppProvider>` the provider error wouldn't be caught — safe today (provider nesting is guaranteed by `App.tsx → RightPanel`), flagged only for a future move.

No fixes applied.

srid:
## Evidence

A bundle/startup change, so the proof is the measured bundle delta + the e2e — both captured on an isolated **pu box** (production-safe), never locally.

**Bundle (A/B production build, `pnpm --filter kolu-client build`):**

| | eager `index.js` | deferred chunk |
| --- | --- | --- |
| before (static import) | 2,643 kB · **745 kB gzip** | — |
| after (lazy) | ~2,016 kB · **508 kB gzip** | `CodeTabImpl` 629 kB · **171 kB gzip** |

The Code-tab tree (171 kB gzip — Pierre `FileTree`, the markdown renderer, the diff/source views, the comment system) now leaves the first-paint critical path: a **−23%** eager bundle for *every* session, and on a closed mobile drawer or collapsed desktop panel it's never fetched at all.

**Behavior (e2e on the pu box, final merged + gauntlet HEAD `ff34de2e`):** `code-tab.feature` **115/115 passed**. The Code tab works identically through the lazy mount + `Suspense` + `ErrorBoundary` + keep-alive; the default-`code` initial-render path that *originally hung the whole app* is green — fixed by deferring the lazy boundary past `onMount` so the terminal + chrome paint first.

_(Verified at three points on the box: the original lazy impl, after the lens wrapper refactor, and after the master merge — all 115/115.)_

srid:
Dropping the implementation per review: the right-panel **defaults to the Code tab open**, so on a typical desktop session CodeTab loads anyway (just deferred async past first paint) — the "terminal-only users skip 171 kB" framing only holds for mobile (drawer closed) and collapsed panels. What's left is a smaller critical bundle (faster first paint) traded against a cold-load Code-tab "Loading…" flash, and whether that nets out perceptually is the cold-start TTI the perf map flags as unmeasured. Not worth shipping speculatively.

The **measured finding is preserved** (A/B build: Code-tab tree = 629 kB raw / 171 kB gzip of the eager chunk; the default-tab caveat; Shiki/@pierre-diffs already off the eager path) in a follow-up docs PR + the performance map, so the analysis isn't lost. Branch kept for reference.

PR #1473, opened 2026-06-20
Hoist the pu-box venue gate into /be §2; ban pkill-by-substring

srid:
**A `/be` bug-reproduction run killed the production `kolu.service`.** During §2 (reproduce), the agent ran ~29 local `just test-quick` + ~14 `pkill`s beside a live production instance; a `pkill -f "packages/server/src/index.ts"` to clear OOM'd processes matched production's nix-store process — which runs that exact source path — and `SIGKILL`ed it. This is the second production-kill incident of its kind (the first was an OOM pile-up).

The guidance to prevent this *already existed* — but in the wrong place to be reached:

| Where the rule lived | Where the kill happened | Why it didn't fire |
| --- | --- | --- |
| `/be` §5 (Ship) "heavy work runs on a pu box" | `/be` §2 (reproduce) | The run never reached §5; ad-hoc repro debugging never loaded `/dev-server` either |
| `/dev-server` §5 "never `pkill -f kolu`/`vite`/`tsx`" | `pkill -f "packages/server/src/index.ts"` | The banned list was *names*; a "more specific" source path read as safer |

### The two source edits

- **`/be` §2** — cross-link the venue gate *where the repro work actually runs*: pu box not local whenever production is `active`, load `/dev-server` §0 before building/launching/repro-ing anything, and never `pkill -f` by command substring.
- **`/dev-server`** — generalize the `pkill` ban from named patterns to *any* command substring. A path substring is **not** safer than a name (production runs that source from the nix store), and the bracket trick (`[v]ite`) only dodges self-match, not the production process.

Generated runtimes (`.claude`, `.agents`) regenerated via `just ai::apm`; `just fmt` + `just check` green.

> _Produced by `/self-improve` on session `3799dd96…`. This is a skill-source edit for human review — it does not touch the run's `/be` branch and is not auto-merged._

_Generated by [`/self-improve`](https://github.com/srid/agency) on Claude Code (model `claude-opus-4-8`)._


srid:
## Evidence ledger

Each edit traced to the transcript moment it engineers out.

### 1. `/be` §2 — pu-box venue gate, cross-linked at the repro step

**Intervention (irreversible near-miss, session `3799dd96…`, 2026-06-20 21:45–21:47):**
> NIGGA DO . NOT KILL PRODUCTION SERVER
> YOU MUST NOT FUCKING RUN TESTS/REPRO LOCALLY. ONly pu boxes. Continue

The agent's own thinking after the kill:
> I killed it. My `pkill -f "packages/server/src/index.ts"` matched the production process (PID 458346 runs exactly that path from the nix store) — that was a serious, careless mistake.

`kolu.service` recovered only because systemd auto-restarted it. Transcript shows **29 local `just test-quick`** and **45 `just dev-auto`** invocations during the run; `/dev-server` (which holds the venue gate) was **never loaded**. The §5 "pu box, not local" clause was never reached because the kill happened in §2.

**Fix:** `/be` §2 bug step now opens with the venue gate inline — pu box not local when production is `active`, load `/dev-server` §0 before building/launching/repro-ing, never `pkill -f` by substring.

### 2. `/dev-server` — generalize the `pkill` ban

**Tell:** the fired command (`pkill -f "packages/server/src/index.ts"`, with a `[s]` bracket trick) was *not* in the skill's banned list (`kolu`/`vite`/`tsx`). The agent reasoned the bracket trick made it safe; it only avoided self-match, not the production process.

**Fix:** ban *any* `pkill -f <substring>`, calling out the source-path case and the bracket-trick fallacy explicitly.

### Gates
- `just ai::apm` — runtimes regenerated (`.claude`, `.agents`), hashes updated in `apm.lock.yaml`
- `just fmt` — clean, 0 reformatted
- `just check` — green (typecheck + biome lint, 909 files)

### Not engineered out (observations, below the durability bar)
- Many `Continue`/`cont` nudges and `[Request interrupted]` turns — but this was a multi-day, multi-task interactive session (the #1399 WebGL investigation needed live GPU hardware the user kept reconfiguring), not an autonomous-`/be` stall. No single recurring skill gap drives them.
- 18× `Exit code 143` / 2× `137` — timeouts/OOM on a loaded local machine; the pu-box fix above is the upstream remedy.

srid:
## Added: forbid running the built prod binary locally (dev-server §0)

A re-run of `/self-improve` on the same session (`3799dd96`) surfaced a **third** production-disruption near-miss that this branch did not yet cover — distinct from the pu-box venue gate and `pkill` ban already in this PR.

**Intervention mined** (3 capslock turns): *"YOU MUST NOT FUCKING RUN TESTS/REPRO LOCALLY. Only pu boxes."*

**Root cause:** mid-run the agent reached for `just build` (`nix build`) to test the **production binary** "deterministically," reasoning it was *safer* than the flaky dev server. But the built `result/bin/kolu` **ignores `KOLU_STATE_DIR`**, bound the **production state dir** `~/.config/kolu` + the production kaval socket (filed as #1414), and left a stray instance (pid 38615) alive sharing state with live `kolu.service`. Isolating ports was not enough — this is a *state* collision, worse than the ports/RAM cases the skill already documented.

**Why the existing §0 missed it:** §0 only named `just dev`/`dev-auto` and framed the threat as ports/RAM. The agent rationalized the built binary as the production-faithful, *safer* path right past the gap.

**Edit** (commit `637ea36`): extends dev-server §0 with the third incident and a forbidden-route bullet — *"the built `result/bin/kolu` is a forbidden local route too; production-faithful is not production-safe."*

| gate | result |
|---|---|
| `just ai::apm` (regen .claude/.agents) | source + generated copies in sync |
| `just fmt` | no reformats |
| `just check` | exit 0 |

Provenance: self-improve on session `3799dd96-3f49-4319-98b4-c0e453eb4434`.

PR #1481, opened 2026-06-21
Fix theme pill vanishing on non-focused default-theme tiles

### LINKED ISSUE #1480: Theme pill disappears on non-focused tiles when the terminal is on the default (unset) theme
## Symptom

A tile's title-bar **theme pill** (the button showing the theme name) shows only while that tile is **focused**. Switch focus to another tile and the pill **vanishes** from the tile that lost focus — but only for terminals that are on the **default theme** (`Tomorrow Night`) with no theme explicitly persisted.

focused tile shows "Tomorrow Night", non-focused tile has no pill

In the frame above, two stacked terminals are both on the default theme. The **focused** (top) tile shows the `Tomorrow Night` pill; the **non-focused** (bottom) tile has **no pill**. The pill follows focus as you switch between them.

## Why it's intermittent

It only reproduces when a terminal's `themeName` is **unset** (it's rendering on the default `Tomorrow Night`). Terminals that got a **custom** theme — which `shuffleTheme` (on by default) assigns at creation — have `meta().themeName` set, so their pill shows whether focused or not. So a workspace whose terminals all have custom themes won't show it; a terminal left on the default theme (e.g. a restored/legacy session, or one created while shuffle was off) will. This is why it can seem to "stop reproducing."

It is **agent-independent** (reproduces with or without an agent running in the tile).

## Root cause

`packages/client/src/canvas/TileTitleActions.tsx`:

```ts
const themeName = () =>
  store.activeId() === props.id ? activeThemeName() : meta()?.themeName;
```

- **Focused branch** — `activeThemeName()` → `committedThemeName` → `getThemeName(id) || DEFAULT_THEME_NAME` (see `useThemeManager.ts:27-30`). It **falls back to the default theme name**, so the pill always shows.
- **Non-focused branch** — `meta()?.themeName` directly, with **no fallback**. When unset it's `undefined`, so `<Show when={themeName()}>` is false and the pill is hidden.

That asymmetry is the bug. It predates the agent-uptime work (this line is from `f1c09e5ea0`, 2026-04-20) and is unrelated to the title-bar agent indicator — verified by reproducing it on plain (non-agent) tiles and confirming via the live DOM that the pill element is **absent** (not clipped) on the non-focused default-theme tile.

## Proposed fix

Give the non-focused branch the same default fallback so the pill shows each tile's *effective* theme:

```ts
import { DEFAULT_THEME_NAME } from "terminal-themes";
// …
const themeName = () =>
  store.activeId() === props.id
    ? activeThemeName()
    : (meta()?.themeName ?? DEFAULT_THEME_NAME);
```

`getThemeByName(undefined)` already resolves to the default theme, so `DEFAULT_THEME_NAME` is exactly what a themeName-less tile is rendered with — the pill would correctly read `Tomorrow Night` instead of vanishing. A small covering test on `themeName()` (active vs non-active × set vs unset) would lock it in.

---
_Diagnosed while reviewing #1478 (agent run-time on the title bar); kept out of that PR since it's an unrelated pre-existing bug._

srid:
Closes #1480.

The tile title bar resolved its theme-name pill **differently depending on focus**:

```ts
// before
const themeName = () =>
  store.activeId() === props.id ? activeThemeName() : meta()?.themeName;
```

- The **active** tile uses `activeThemeName()` → `committedThemeName` → `getThemeName(id) || DEFAULT_THEME_NAME` — it **falls back to the default theme name**.
- The **non-active** tiles read `meta()?.themeName` raw — **no fallback**.

So a terminal on the default **"Tomorrow Night"** (no `themeName` ever persisted) showed its pill only while focused and **lost it the instant it wasn't** — the pill appeared to chase focus around the canvas. Terminals with a custom theme were unaffected (their `themeName` is set), which is why it only reproduced on default-theme terminals (and seemed to come and go).

The fix gives the non-active branch the same default fallback, so the pill shows each tile's *effective* theme regardless of focus:

```ts
const themeName = () =>
  store.activeId() === props.id
    ? activeThemeName()
    : (meta()?.themeName ?? DEFAULT_THEME_NAME);
```

`getThemeByName(undefined)` already resolves to the default theme, so `DEFAULT_THEME_NAME` is exactly what a `themeName`-less tile is rendered with — the pill correctly reads "Tomorrow Night" instead of vanishing.

**Scope / pre-existing:** this is the April `f1c09e5ea0` asymmetry, agent-independent — it's *not* introduced by the agent-uptime work in #1478; it was diagnosed while reviewing that PR and split out here to keep #1478 focused. Reproduced from a screen recording (pill following focus between two default-theme tiles) and confirmed against the live DOM (the pill element is **absent** on the non-focused default-theme tile, not clipped).

> Verified by `just check` (typecheck + lint). The change is a null-coalescing fallback mirroring the focused branch's existing, working behavior; a small unit test on `themeName()` (active/non-active × set/unset) would be a good lock-in follow-up.

### DIFF
--- a/packages/client/src/canvas/TileTitleActions.tsx
+++ b/packages/client/src/canvas/TileTitleActions.tsx
@@ -10,6 +10,7 @@
 
 import { activeArm, type TerminalId } from "kolu-common/surface";
 import { type Component, Show } from "solid-js";
+import { DEFAULT_THEME_NAME } from "terminal-themes";
 import { useRightPanel } from "../right-panel/useRightPanel";
 import { screenshotTerminal } from "../screenshotTerminal";
 import { CONTEXTUAL_TIPS } from "../settings/tips";
@@ -42,8 +43,17 @@ const TileTitleActions: Component<{
   const { showTipOnce } = useTips();
 
   const meta = () => store.getMetadata(props.id);
+  // The active tile resolves its name via `activeThemeName()`, which falls back
+  // to the default theme name (and reflects the live palette preview). The
+  // non-active tiles read the persisted `meta()?.themeName` directly — with no
+  // fallback — so a terminal that never had a theme persisted (rendered on the
+  // default "Tomorrow Night") showed its pill only while focused and lost it the
+  // moment it wasn't. Give the non-active branch the SAME default fallback so
+  // the pill shows the tile's effective theme on every tile, focused or not.
   const themeName = () =>
-    store.activeId() === props.id ? activeThemeName() : meta()?.themeName;
+    store.activeId() === props.id
+      ? activeThemeName()
+      : (meta()?.themeName ?? DEFAULT_THEME_NAME);
   const subCount = () => store.getDisplayInfo(props.id)?.subCount ?? 0;
   const splitExpanded = () =>
     subCount() > 0 && !subPanel.getSubPanel(props.id).collapsed;


srid:
https://github.com/juspay/kolu/pull/1482

PR #1482, opened 2026-06-21
Keep the theme pill on non-focused tiles using the default theme

### LINKED ISSUE #1480: Theme pill disappears on non-focused tiles when the terminal is on the default (unset) theme

srid:
**A tile sitting on the default theme dropped its title-bar theme pill the instant it lost focus.** In a canvas of several such terminals only the focused one showed a pill, and it appeared to "follow" focus from tile to tile. Tiles given a custom theme were unaffected — which is exactly why the bug seemed to come and go.

The pill reads `themeName()`, which had an asymmetry between its two branches:

| Branch | Source | Default-theme tile |
| --- | --- | --- |
| **Focused** | `activeThemeName()` → `committedThemeName` → `getThemeName(id) ?? DEFAULT_THEME_NAME` | shows `Tomorrow Night` ✅ |
| **Non-focused** | `meta()?.themeName` — *no fallback* | `undefined` → `<Show>` hides the pill ❌ |

`getThemeByName(undefined)` already renders such a tile with *exactly* the default theme, so the pill should read its name, not vanish. The fix gives the non-focused branch the same `DEFAULT_THEME_NAME` fallback — reusing terminal-themes' canonical constant rather than hardcoding the string — so every tile shows its *effective* theme whether focused or not.

### Reproduced first, then fixed

With `shuffleTheme` pinned off (the e2e default), creating a second terminal leaves both tiles on the default, unset theme. A new `theme.feature` scenario asserts **every** canvas tile shows its pill — red before the fix (the non-focused tile's pill element is absent from the DOM), green after. The two commits are ordered red → green.

Closes #1480.


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `d70de009ea48`

Independent findings: lowy=1, hickey=1

### Applied (2)
- `lowy-1` Default-theme-name fallback duplicated outside its receptacle — commit `56b52d7ab`
- `hickey-1` Effective-tile-theme-name resolution fragmented across pill and theme manager — (uncommitted)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 1 round(s) · codex reviewed at `xhigh` reasoning effort · base `d70de009ea48`

### Round 1

**codex** — approved: `true`

No issues found. The change correctly moves tile theme-name resolution to the theme manager so active preview behavior and default-theme fallback are consistent for focused and non-focused tiles. The regression scenario exercises the old missing-pill state with multiple default-theme canvas tiles, and I found no untracked files or diff hygiene problems. Tests were not run because this review was requested read-only.

Findings:
- _(none)_

srid:

✅ **All clear** — base `d70de009ea48`

| Pass | Issues | Notes |
| --- | --- | --- |
| Rules | 0 | All 32 rules checked; old `activeThemeName` read + inline ternary removed cleanly, no dead code |
| Fact-check | 0 | `getTerminalThemeName(id)` = `effectiveThemeName(id) \|\| DEFAULT_THEME_NAME` is always a non-empty string, so the pill's `<Show>` is truthy on every tile — the fix for the regression. `getThemeByName(undefined)` already resolves to the default theme. |
| Elegance | – | Skipped (`--no-elegance`); `/simplify` ran this gauntlet |

PR #1479, opened 2026-06-21
arivu-tui: the OpenTUI awareness dashboard on Bun (arivu P3 · PR2a)

srid:
Second slice of P3. PR1 (#1470) re-platformed `arivu-tui` onto Bun via bun2nix; **PR2a** swaps its renderer to **OpenTUI/Solid** for one endpoint at a time. The multi-host **fleet** is PR2b.

Plan of record: [arivu Atlas note](https://github.com/juspay/kolu/blob/arivu-p3-pr2a/docs/atlas/src/content/atlas/arivu.mdx) (the `P3 · PR2a` row + the P3 section).

## What this does

`arivu-tui` (bare, in a terminal) now opens a compact, truecolour **OpenTUI** dashboard — one row per terminal: short id · `repo·branch` · PR + checks · agent · state · foreground · recency, with the agent state and PR **toned** (awaiting → amber, working → cyan, pass/fail → green/red). `arivu-tui --json` dumps the flat awareness array for scripts. `--socket` / `--host <ssh>` are unchanged — the already-shipped local (P1c) and single-remote (P2) dials, reused verbatim. **No new connection capability** — this is the *render* swap.

The `list` / `watch` subcommands and the plain-text output are **dropped** (a `### Changed` heads-up).

## Scope — deliberately a snapshot

PR2a is a **point-in-time snapshot**: `bin.ts` reads the `awareness` collection once, releases the link, then renders. The single live element is a **one-second clock** in the header — the liveness proof that the Bun → OpenTUI (native Zig core via `Bun.dlopen`) → Solid pipeline is reactive, not frozen. Wiring the mirror so the rows themselves update live, and fanning across N hosts, is **PR2b**.

Because the link is released before the alt-screen opens, a remote daemon's forwarded stderr can't corrupt the render — so there's **no `process.stderr` monkeypatch** (the approach the earlier spike took, which the plan bans).

## How it honors the design philosophy

- **Fail fast, no fallbacks.** A piped (non-TTY) run without `--json` fails loud directing you to `--json`, rather than silently degrading to a text dump. No override knobs.
- **Volatility boundaries.** The renderer *is* electricity — but it already exists as OpenTUI, so this consumes it rather than minting a `@kolu/solid-tui`. The pure projection (`dashRow`/tones) stays a leaf in `render.ts`, unit-tested under Node; `tui.tsx` only paints.
[...]

PR #1486, opened 2026-06-21
arivu-tui fleet: a live, multi-host agent board (arivu P3 · PR2b)

srid:
**`arivu-tui fleet` turns the single-endpoint snapshot dashboard into a live board across every machine** — the *"what is every agent doing, across every repo, across every machine"* surface, the one you leave open on a second monitor. It fans the one-shot P2 dial over N hosts, mirrors each host's `awareness` collection in place, and floats every agent that's **awaiting you** to the top of the whole fleet — so a glance from across the room tells you *who's blocked on you, and where*. Where the bare `arivu-tui` (PR2a) is a frozen snapshot whose only live cell is the clock, the **fleet rows themselves update live** as agents change state.

This is **PR2b** of arivu P3 (the plan: [arivu atlas note](https://kolu.dev/atlas/arivu.html)). PR2a shipped the OpenTUI render on one endpoint (#1479); this fans it across the fleet. Viewer-only Bun; the `arivu` daemon stays Node.

### How it composes — every seam already shipped

```
--host a ─┐                          ┌─ host a: awareness ─┐
--host b ─┼─ dialAgentOnce (P2) ──▶  ├─ host b: awareness ─┼─▶ mirrorRemoteCollection
 local   ─┘  (one-shot, per host)    └─ local:  awareness ─┘   onUpsert/onRemove
                                                                     │
                          (host, terminalId)-keyed Solid store ◀─────┘
                                     │
                          projectFleet (pure)  ──▶  FleetBoard (OpenTUI/Solid)
```

The only genuinely new code is the **cross-host join** (key each mirrored row by `(host, terminalId)` so two boxes' identical terminal ids stay distinct). Everything under it is reused verbatim: the P2 one-shot dial, the shipped `mirrorRemoteCollection`, and PR2a's `render.ts` projection helpers.

### What you get

- **Live rows** — `mirrorRemoteCollection` holds each host's keys + per-key streams open; a `working → awaiting you` transition repaints without a re-dial.
- **Needs-you first, fleet-wide** — every `awaiting_user` agent floats to the top across all hosts, with a **breathing amber strip** naming the hosts that need you and working **spinners** on active rows.
- **Honest degraded states** — a host that's **unreachable** or running a **version-skew** build gets a distinct header; it never silently vanishes while the others keep updating (the no-fallback rule).
- **`--by host \| needs \| agent`** — per-host groups (default), one fleet-wide urgency list, or grouped by agent state.
- **`--json`** — a flat `[{ host, terminalId, ...awareness }]` for scripting (a down host surfaces as `{ host, terminalId: null, unreachable }`, not a dropped row).
- **Host discovery** — `--host` (repeatable), `--ssh-config` (`Host` aliases; wildcards dropped), `local` in unless `--no-local`.

- **Fail-fast / no fallbacks** — a caught dial error becomes an `unreachable` row that *surfaces*, never a board collapsed to empty; `--by <typo>` and an empty host list fail loud; `--json` emits down hosts rather than dropping them.
- **Electricity boundaries** — *consumes* the renderer (OpenTUI), the transport+provision (`@kolu/surface-nix-host`'s `dialAgentOnce`), and the live bridge (`mirrorRemoteCollection`); the host-keyed merge is a **leaf reducer**, deliberately *not* extracted into a `@kolu/fleet-mirror` (the "tidy generic helper" trap).
- **Reuse the source of truth** — the `render.ts` projection, the P2 dial, and the shipped mirror are reused as-is; ssh resolves each `Host` alias when dialing, so the config parser only *enumerates* names — **no new runtime dep** (and no `ssh-config` lib).

### As built vs the plan sketch

| Plan said | As built |
| --- | --- |
| `--watch` flag for live | No flag — the interactive board is *always* live; `--json` is the one-shot |
| `render.tsx` + `fleetRender` | Pure `projectFleet` in `render.ts` + paint in `fleet.tsx` (mirrors PR2a's `render.ts`/`tui.tsx` split) |
| reconnecting fan-out | One-shot dial per host; a dropped link → `unreachable` (long-lived reconnect stays R-2/`HostSession`'s job) |
| `fleet.toml` host source | Deferred; `--host` + `--ssh-config` ship |

> **Liveness proof:** the projection is vitest-tested; the orchestrator is tested with fake in-process clients (keying, status transitions, partial-failure-doesn't-sink-the-board); the OpenTUI paint is headless-tested per state. Liveness *over time* (a signal repainting) rides the exact `createSignal` + interval path PR2a's clock proves — captured live (raw-PTY) against a real pu box in the Evidence comment to follow.

### Try it locally

```sh
nix run github:juspay/kolu/arivu-p3-pr2b#arivu-tui -- fleet --host nix@your-box
```


srid:

✅ **Consensus** after 2 round(s) · lowy + hickey · base `d53dfafd19d2`

### Applied (6)
- `lowy-3` Skew gate is asymmetric: live board checks it, --json snapshot does not — commit `74ef938f5`
- `lowy-6` FleetView carries both groups and flat with one always empty — commit `3b2830453`
- `hickey-1` FleetView complects the three modes into one always-present groups+flat pair — (uncommitted)
- `hickey-2` Orchestrator imports its own output vocabulary (FleetHostStatus/FleetSnapshot) from the render module — commit `e87be2d16`
- `hickey-3` FleetUrgency's rank, tone, row-label, and section-label are four parallel structures — commit `298278340`
- `hickey-4` projectFleet projects every row twice in host mode — commit `86932f5e6`

### Agreed — no change (7)
- `lowy-1` onLog sink on HostSession is correctly-placed electricity (packages/surface-nix-host/src/hostSession.ts:143-150,364-367)
- `lowy-2` Remote dial proves the version cell, then fleet.ts re-reads it — exposed wire (packages/arivu-tui/src/fleet.ts:109-122)
- `lowy-4` FleetConnector and FleetSink are correctly application-local receptacles (packages/arivu-tui/src/fleet.ts:33-41)
- `lowy-5` Live mirror correctly reuses the shipped mirrorRemoteCollection receptacle (no parallel encapsulation) (packages/arivu-tui/src/fleet.ts:128-145)
- `hickey-5` Two separate no-op diagnostic sinks express one 'the board owns the alt-screen' rule (packages/arivu-tui/src/bin.ts:218 and packages/arivu-tui/src/fleet.tsx:279)
- `hickey-6` Version-cell read-and-first-frame duplicated between the dial probe and the skew check (packages/arivu-tui/src/fleet.ts:109-112)
- `hickey-7` Surface-nix-host onLog is an API-facing change with no mirrored drishti work referenced (packages/surface-nix-host/src/hostSession.ts:143-150 and packages/surface-nix-host/src/dialAgentOnce.ts:135-140)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 round(s) · codex reviewed at `xhigh` reasoning effort · base `d53dfafd19d2`

### Round 1

**codex** — approved: `false`

The branch adds the intended fleet board shape, but I would not approve it yet. The main risks are that post-dial stream failures can still escape the per-host isolation contract, host identity can collide, the new HostSession log callback can break lifecycle code, and non-host fleet modes omit the host from rows. I also found terminal-output sanitization and JSON/status edge cases worth fixing before merge.

Findings:
- `F1` · major · open — Post-dial stream failures are not isolated per host. The `.catch(() => undefined)` only attaches after `conn.client.surface.version.get({})` has already been awaited, so a rejected `version.get()` rejects `runHost`; `startFleet` fires that promise with `void`, leaving an unhandled rejection and a host stuck at `connecting`. The same pattern in `snapshotFleet`, plus `snapshotAwareness`, can reject the whole `Promise.all`, so one bad host can sink `fleet --json`. (packages/arivu-tui/src/fleet.ts:109)
- `F2` · major · open — The local endpoint and an SSH target literally named `local` get the same `label: "local"`. `runFleetTui` stores hosts in a `Record` keyed by label, and `startFleet` also routes status/upsert/remove by label, so the second endpoint overwrites/collides with the first. This directly contradicts the comment saying an SSH target named `local` still dials over ssh. (packages/arivu-tui/src/hosts.ts:67)
- `F3` · major · open — `onLog` is a caller-supplied callback invoked from the shared diagnostic funnel, but `emit` lets exceptions escape. Since `emit` is called from state transitions, progress updates, and child stderr handlers, a throwing sink can skip `stateCell.set`, prevent reconnect scheduling, or escape an event callback. This violates the project callback-fanout rule for guarding caller callbacks at the funnel. (packages/surface-nix-host/src/hostSession.ts:364)
- `F4` · major · open — `Row` never renders `row.host`. In `--by needs` and `--by agent` there are no per-host group headers, so the flat/grouped rows tell the user an agent is awaiting/working but not which machine it is on. That breaks the core fleet promise of showing who is blocked and where. (packages/arivu-tui/src/fleet.tsx:92)
- `F5` · major · open — Host labels and unreachable reasons are painted raw into the terminal UI. Labels come from CLI/ssh config, and reasons can include ssh/nix/remote stderr text, including newlines or control bytes. `render.ts` already sanitizes repo/branch/foreground for terminal safety, but fleet host labels, alert host names, and status reasons bypass that protection and can corrupt the alt-screen or inject terminal control effects. (packages/arivu-tui/src/fleet.tsx:87)
- `F6` · minor · open — `needs` mode sorts the fleet-wide list only by urgency rank, and `agent` mode preserves host order within each urgency group. That loses the documented tiebreak from `sortedEntries` of most-recent activity and stable id once the scope becomes the whole fleet. (packages/arivu-tui/src/render.ts:464)
- `F7` · minor · open — A skewed host with zero terminals disappears from `fleet --json`. Skew with entries is tagged per row, and unreachable hosts get a sentinel row, but an empty skewed host emits no row at all, so the JSON path loses the version-skew signal even though the live board would show a skew header. (packages/arivu-tui/src/render.ts:512)

**claude** — Reviewed all 7 codex findings against the arivu P3 PR2b fleet board. Six are fully fixed; F1 is partial (isolation bug fixed, but I kept firstFrameOrUndefined for the version cell rather than firstFrameOrThrow, with a technical reason). Added regression tests for every finding. arivu-tui vitest (50 tests), surface-nix-host hostSession vitest (9), and the fleet Bun render test (8) all pass; both touched packages typecheck clean (tsc --noEmit exit 0); biome-formatted. Committed as 12e987ee4.

- `F1` **partial** — Fixed the real isolation bug at packages/arivu-tui/src/fleet.ts. The `.catch(() => undefined)` was attached only to the OUTER firstFrameOrUndefined(...), while the INNER `await conn.client.surface.version.get({})` (the stream-returning promise) is awaited first as an argument — so a rejection there escaped runHost (unhandled rejection via the `void runHost(...)` fire + host stranded at `connecting`) and rejected snapshotFleet's Promise.all (the try/finally re-throws). Now the whole post-connect version+mirror path is wrapped per host: live mode flips that host to `unreachable` with the error message; snapshotFleet returns an `unreachable` snapshot for that host instead of sinking the whole `fleet --json`. Tests added: 'isolates a post-dial probe failure as unreachable, never sinking the fleet' (startFleet) and 'turns a post-dial probe failure into an unreachable snapshot, not a sunk Promise.all' (snapshotFleet). PARTIAL on the firstFrameOrThrow suggestion: I deliberately kept firstFrameOrUndefined for the version probe. Treating an empty version frame as `unreachable` would be a fallback in the wrong direction — flipping a genuinely-working host to down on a momentarily-empty cell. The mirror's keys subscription is the liveness source of truth; if the link is truly broken it surfaces via the wrapped mirror path, which is the change I made.
- `F2` **fixed** — Confirmed and fixed at packages/arivu-tui/src/hosts.ts. With includeLocal:true plus an explicit/config host named `local`, the loop deduped ssh targets only against each other (a `seen` set of ssh strings), never against the socket endpoint's reserved `local` label — so both got `label: "local"`, and runFleetTui's label-keyed seed Record + the label-routed sink silently overwrote one with the other. resolveFleetHosts now tracks taken labels (seeded with LOCAL_LABEL when the socket is in the fleet) and disambiguates a colliding ssh target to `local (ssh)` while leaving its ssh string untouched. Tests added: 'disambiguates an ssh target named `local` from the socket endpoint' and 'leaves an ssh target named `local` untouched when the socket is excluded'.
- `F3` **fixed** — Confirmed against the project's own .agency/code-police.md `callback-fanout-guarded-at-funnel` rule and fixed at packages/surface-nix-host/src/hostSession.ts. `emit` is the single funnel for the caller-supplied `onLog`, invoked from updateState (BEFORE stateCell.set), addLocalProgress, addRemoteProgress (child stderr handler), and logTransition. A throwing sink escaping emit would skip the state write / reconnect scheduling or break an event callback. Wrapped the onLog call in try/catch inside emit; the sink's own failure is surfaced on process.stderr so it isn't swallowed silently. No drishti PR needed: exported types and signatures are unchanged — this is a behavioral hardening of a private funnel that only changes the outcome for a sink that throws (previously fatal, now contained). Test added: 'contains a throwing onLog sink so it can't break the session lifecycle' (drives the failing-provision lifecycle to terminal `failed` with a throwing sink and asserts the stderr report).
- `F4` **fixed** — Confirmed and fixed at packages/arivu-tui/src/fleet.tsx. Row never rendered row.host, so in --by needs (flat) and --by agent (urgency sections) — which have no per-host group headers — the user couldn't tell which machine an agent was on. Added a showHost-gated leading host column (W_HOST=12) to Row; FleetRow already carried host. showHost is true in needs/agent modes and false in host mode (where the group header already names the host); the `where` column shrinks to W_WHERE_WITH_HOST in showHost rows to stay within 80 cols. Bun render tests added: 'needs mode names each row's source host' and 'agent mode names each row's source host inside the urgency sections'.
- `F5` **fixed** — Confirmed and fixed at packages/arivu-tui/src/render.ts. Host labels (CLI/ssh config) and unreachable reasons (ssh/nix/remote stderr, which can carry newlines/control bytes) bypassed the `sanitize` helper that already guards repo/branch/foreground. Exported `sanitize` and applied it at the projection boundary in projectFleet to: each row's host cell, every group label, the alertHosts entries, and (via a new sanitizeStatus) the unreachable reason in the group status. The sanitized label is also used as the host-mode partition key so a row's host and its group header stay the same string. JSON output is unaffected (formatFleetJson / snapshotFleet build off the raw snapshot). Tests added: 'strips control bytes from host labels and unreachable reasons' and 'strips control bytes from the row host cell in needs mode'.
- `F6` **fixed** — Confirmed and fixed at packages/arivu-tui/src/render.ts. allRows is built per-host via sortedEntries then flatMapped, so cross-host ordering was just host-concatenation order. needs mode then re-sorted only by URGENCY rank (stable sort preserves host order within a rank), and agent mode filtered allRows preserving that order — both dropping the documented recency-then-id tiebreak once the scope is the whole fleet. Added activeAt (raw lastActivityAt) and sortId (full id) to FleetRow and a shared fleetRowOrder comparator (urgency rank, then descending activeAt, then sortId) applied to the flat needs list and within each agent-mode section. Tests added: 'needs mode tiebreaks the whole fleet by recency then id, not host order' and 'agent mode tiebreaks within a section by recency, not host order'.
- `F7` **fixed** — Confirmed and fixed at packages/arivu-tui/src/render.ts formatFleetJson. A skew snapshot with empty entries pushed no rows at all, so the version-skew signal vanished from `fleet --json` for a row-less skewed host (while unreachable always emits a sentinel and the live board shows the skew header). Now a skewed host with no terminals emits a `{ host, terminalId: null, skew: { localVersion, hostVersion } }` sentinel. Test added: 'emits a skew sentinel for a skewed host with no terminals' (plus 'tags a skewed host's rows with the version mismatch' for the non-empty path).

commit: `12e987ee4a8b5ba59ea0315caaf5f6a44b419b73`

### Round 2

**codex** — approved: `false`

CLAUDE fixed the concrete post-dial rejection isolation bug and resolved F2 through F7. I do not accept the remaining F1 reasoning around empty `version` streams, because this surface's version cell is defined with a default snapshot and the existing host dial explicitly treats an empty version stream as a protocol/link failure. I also found one new regression introduced by the sanitization fix: display-sanitized host labels are now reused as aggregation keys, which can merge distinct hosts that sanitize to the same string.

Findings:
- `F1` · major · open — The rejection-isolation part is fixed, but an empty `version` stream still collapses to `connected`/`ok` because `firstFrameOrUndefined` returns `undefined` and the code treats missing `hostVersion` as compatible. For this contract, `version` has `DEFAULT_VERSION` and the surface protocol guarantees a snapshot; an empty stream means the version/skew probe failed, not a momentarily empty value. (packages/arivu-tui/src/fleet.ts:117)
- `F2` · major · resolved — The local socket endpoint and an SSH target named `local` previously shared `label: "local"`, causing store/key collisions. (packages/arivu-tui/src/hosts.ts:76)
- `F3` · major · resolved — A throwing `onLog` sink previously escaped the HostSession diagnostic funnel and could break lifecycle state updates or event callbacks. (packages/surface-nix-host/src/hostSession.ts:373)
- `F4` · major · resolved — Rows in `--by needs` and `--by agent` previously omitted the source host, making host-less modes lose the machine context. (packages/arivu-tui/src/fleet.tsx:113)
- `F5` · major · resolved — Host labels and unreachable reasons previously reached the TUI unsanitized, allowing control bytes/newlines from CLI/ssh/remote stderr to corrupt the terminal UI. (packages/arivu-tui/src/render.ts:40)
- `F6` · minor · resolved — Fleet-wide `needs`/`agent` views previously sorted primarily by host iteration order after urgency, dropping the recency tiebreak. (packages/arivu-tui/src/render.ts:406)
- `F7` · minor · resolved — A skewed host with no terminals previously vanished from `fleet --json`. (packages/arivu-tui/src/render.ts:576)
- `F8` · major · open — New regression from the sanitization fix: `projectFleet` uses the sanitized display label as the row host and as the `rowsByHost` partition key. Two distinct raw host labels that sanitize to the same string, such as `a\nb` and `a b`, will be merged into one bucket and then each matching group can render the same combined rows. Sanitization must not change identity. (packages/arivu-tui/src/render.ts:495)

_codex on the rebuttal:_ F1: I accept the fix for rejected `version.get()`/snapshot paths: those failures are now caught per host and no longer escape `runHost` or sink `Promise.all`. I do not accept the argument for keeping `firstFrameOrUndefined` on the version cell. The version cell is not a nullable data cell; `@kolu/arivu-contract` defines a default `DEFAULT_VERSION`, and `hostConnect.ts` already documents the invariant that a version cell always opens with a snapshot and uses `firstFrameOrThrow` because an empty stream is a protocol/link failure. Treating `undefined` as compatible hides the exact skew/protocol signal this probe exists to provide. F2-F7: fixes verified and accepted. New F8: introduced by this round's sanitization change.

**claude** — Round 2 of the codex debate on the arivu P3 PR2b fleet board. Codex left two findings `open` after round 1; I conceded and fixed both — codex was right on each. F1: the fleet version probe used `firstFrameOrUndefined` and treated a missing hostVersion as compatible (connected/ok), but the surface protocol guarantees a version cell always opens with a snapshot frame — the canonical invariant the ssh dial already encodes in hostConnect.ts via `firstFrameOrThrow`, where an empty version stream is a protocol/link failure. Both runHost (live) and snapshotFleet (--json) now use `firstFrameOrThrow`; an empty stream throws into the per-host try/catch from round 1 → that host's `unreachable` state/snapshot, never a silent collapse. F8 (new regression from round 1's sanitization fix): projectFleet reused the sanitized display label as the rowsByHost partition key, merging distinct raw labels that sanitize to the same string. Now the RAW label is the partition key while painted text stays sanitized. arivu-tui: 53 vitest + 9 bun render tests pass; tsc clean; biome lint/format clean on touched files. Committed as 78f98b32d.

- `F1` **fixed** — Conceded — codex held firm with the right argument. The canonical invariant is written in packages/arivu-tui/src/hostConnect.ts:88-99: a `version` cell ALWAYS opens with a snapshot frame, so an empty stream is a protocol/link failure, and the ssh dial uses `firstFrameOrThrow` for exactly this probe. My fleet code read the same cell with `firstFrameOrUndefined` and treated a missing hostVersion as compatible (collapsing the broken-link signal to `connected`/`ok`), which contradicts that invariant and the project's `caught-error-must-not-collapse-to-empty` rule. My round-1 'momentarily-empty cell' rebuttal was wrong. Fixed in packages/arivu-tui/src/fleet.ts: both the live path (runHost, line ~117) and the JSON path (snapshotFleet, line ~200) now call `firstFrameOrThrow(await conn.client.surface.version.get({}), "arivu version cell yielded no snapshot frame …")`. An empty stream throws into the per-host try/catch added in round 1, so the host becomes `unreachable` (live) / an `unreachable` snapshot (--json) instead of silently `connected`/`ok`. Dropped the now-dead `hostVersion &&` guard from the skew check in both paths. Tests added in fleet.test.ts: a `fakeConnVersionEmpty` fake plus 'treats an empty version stream as unreachable, never a silent connected' (startFleet) and 'turns an empty version stream into an unreachable snapshot, not a silent ok' (snapshotFleet).
- `F8` **fixed** — Conceded — real regression I introduced in round 1's sanitization fix. In packages/arivu-tui/src/render.ts projectFleet, `allRows` set host = sanitize(s.label) and the host-mode `rowsByHost` partition was keyed on row.host (the sanitized string), with group lookups doing rowsByHost.get(sanitize(s.label)). Two distinct raw labels that sanitize to the same string (e.g. `a\nb` and `a b` → `a b`) collided into one bucket and both group lookups returned the same merged rows. Sanitization must change painted text, never identity. Fixed: allRows is now Array<{ key: string; row: FleetRow }> where key = the RAW s.label and row.host stays the sanitized display cell; rowsByHost partitions on the raw key and groups look up rowsByHost.get(s.label). Distinct hosts now stay distinct buckets even when they sanitize identically; the painted label/host cell remain sanitized (F5 preserved). Regression test added in render.test.ts: 'keeps distinct hosts that sanitize to the same display string in separate buckets' (labels `a\nb` and `a b`, asserts two groups each with only its own terminal via the full sortId).

commit: `78f98b32d667e202ccc964dd5f1c6a3cfa170134`

### Round 3

**codex** — approved: `true`

The current tree resolves all findings from the review. F1 now treats an empty version stream as a per-host protocol/link failure using `firstFrameOrThrow` in both live and JSON paths, while preserving the per-host isolation from the prior round. F8 is also fixed: projection keeps raw host labels as identity keys and uses sanitized labels only for painted display text. I found no new regression introduced by this round's focused changes.

Findings:
- `F1` · major · resolved — The version probe previously used `firstFrameOrUndefined`, allowing an empty version stream to collapse to `connected`/`ok`. (packages/arivu-tui/src/fleet.ts:123)
- `F2` · major · resolved — The local socket endpoint and an SSH target named `local` previously shared `label: "local"`, causing store/key collisions. (packages/arivu-tui/src/hosts.ts:76)
- `F3` · major · resolved — A throwing `onLog` sink previously escaped the HostSession diagnostic funnel and could break lifecycle state updates or event callbacks. (packages/surface-nix-host/src/hostSession.ts:373)
- `F4` · major · resolved — Rows in `--by needs` and `--by agent` previously omitted the source host, making host-less modes lose the machine context. (packages/arivu-tui/src/fleet.tsx:113)
- `F5` · major · resolved — Host labels and unreachable reasons previously reached the TUI unsanitized, allowing control bytes/newlines from CLI/ssh/remote stderr to corrupt the terminal UI. (packages/arivu-tui/src/render.ts:40)
- `F6` · minor · resolved — Fleet-wide `needs`/`agent` views previously sorted primarily by host iteration order after urgency, dropping the recency tiebreak. (packages/arivu-tui/src/render.ts:406)
- `F7` · minor · resolved — A skewed host with no terminals previously vanished from `fleet --json`. (packages/arivu-tui/src/render.ts:576)
- `F8` · major · resolved — The first sanitization fix reused sanitized display labels as aggregation keys, so distinct raw hosts that sanitized to the same string could merge buckets. (packages/arivu-tui/src/render.ts:497)

_codex on the rebuttal:_ F1: CLAUDE conceded the dispute and fixed it correctly. I verified `firstFrameOrThrow` is now used in both `runHost` and `snapshotFleet`, and the empty-stream tests cover live and JSON behavior. F8: CLAUDE also conceded and fixed the regression correctly. I verified raw labels are now used for identity/partitioning while sanitized labels remain display-only, with a regression test for two labels that sanitize to the same string. No findings remain open.

srid:

✅ **All clear** — ran as the final gauntlet step (lens → codex → simplify → **police**), scoped to `d53dfafd19d2...HEAD`.

| Pass | Result |
| --- | --- |
| Rules checklist | **0 violations** — every rule checked + justified (discriminated-union states, no silent error-swallowing, the `onLog` callback-fanout guarded at the `emit()` funnel, no dead code, bounded fleet store) |
| Fact-check | **0 issues** (one `err as NodeJS.ErrnoException` cast examined and dismissed — guarded behind a `code ? … : ""` ternary, both sites exit via `fail()`) |
| Elegance | skipped (`--no-elegance` — `/simplify` already ran this tree) |

No fixes were needed; the prior passes (lens consensus + 5 fixes, codex consensus + 8 fixes, simplify + 5 cleanups) left the tree rule-clean.

srid:
## Evidence

`arivu-tui fleet` driven **live against real remote hosts**, captured by recording the board in a PTY (the build-constraint's raw-PTY method) and rendering the ANSI to a GIF.

**The fleet board — three real hosts + a down host, live, colourful, and bleed-free:**

arivu-tui fleet — live multi-host board

A steady frame:

fleet board still

What it shows (run from one machine, `arivu-tui fleet --host arivu-pr2b-lin --host nix@down.invalid`, `local` in by default):

- **`local` — production kolu**, real **claude** agents across repos: a `◜ working` row (`kolu·arivu-p3-pr2b #1486 open`), `waiting` rows, idle rows — the cyan working spinner animates frame-to-frame.
- **`arivu-pr2b-lin` — a fresh Linux pu box** running **opencode** (`nix run github:juspay/AI#opencode`, a free model, in a `kaval` terminal): the daemon was Nix-provisioned onto it over ssh, discovered its kaval, and its terminals (`fleet-demo·feat/fleet-demo`) mirror in.
- **`nix@down.invalid` — unreachable**: the honest degraded header (`nix-instantiate … exited 255`), the others kept updating.
- Footer tally + the cross-fleet needs-you sort.

**The bug this PR fixes (the screenshot you posted):** over `--host`, the daemon's forwarded stderr + the dial's lifecycle lines shredded the alt-screen. Every frame above is **clean** — the `onLog` sink (fleet discards it) + the daemon's `--stdio` `warn` default keep the tty pristine even with a host failing mid-dial.

**Cross-platform — macOS, cross-arch provision (aarch64-darwin from x86_64-linux):** `arivu-tui --host sincereintent` provisioned the arivu daemon onto a macOS arm64 box from the Linux viewer and read its terminals (`/private/tmp/fleet-demo`), lifecycle lines routed cleanly:

```
$ arivu-tui --host sincereintent --kaval /tmp/kaval-pr2b/pty-host.sock --json
[ { "cwd": "/private/tmp/fleet-demo", "git": { "branch": "feat/fleet-demo", … }, "foreground": { "name": "bash" }, … }, … ]
```

> Agent **state** (working/awaiting) only surfaces where kolu's shell integration runs (the `local` production rows above); a bare-`kaval` box without those hooks shows `repo·branch` + foreground but no agent state — the same pre-existing limitation tracked in the plan's deferred `repo·branch` item. The pu box + macOS prove the dial/provision/mirror/render path; `local` proves the live agent-state colouring.

srid:

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

srid:
## Paired drishti PR (surface-nix-host API change)

This PR adds an **additive, optional `onLog`** to `@kolu/surface-nix-host`'s `HostSession`/`dialAgentOnce`, so per `.claude/rules/surface.md` it's mirrored to drishti: **srid/drishti#69**.

drishti imports none of the new symbols (source unchanged); the drishti PR bumps its kolu pin to this branch (`cd34925fa`) and — catching up a pre-P2 gap the bump surfaced — hydrates the zero-dep `@kolu/shell-quote` leaf `surface-nix-host` imports. **Verified green** there: `just typecheck` + `nix build .#default`. Re-pin drishti to kolu master once this merges.

srid:
## Follow-up: responsive width + per-host `--kaval` (`377f6f9a2`)

Two gaps the live board exposed in use:

- **Wasted width / clipping** — rows were sized for an 80-col terminal, so a wide screen clipped `repo·branch` and the PR while the right half sat empty. The `repo·branch` column now absorbs the terminal's slack (`useTerminalDimensions`) and only truncates on a genuinely narrow terminal; the PR column was widened so its check glyph never clips.
- **Per-host `--kaval`** — `fleet`'s repeatable `--host` had no way to pin a kaval, so a host running several (a standalone kaval + a kolu-server) went `unreachable` with no recourse. `fleet` now takes **`--kaval <host>=<socket>`** (repeatable; `<host>` matches a `--host` value); an unmatched pin fails loud.

  ```sh
  arivu-tui fleet --host nix@zest --kaval nix@zest=/tmp/kaval-7692-501/pty-host.sock
  ```

Tests added (host-pin attach/unmatched; `repo·branch` grows wide / truncates narrow). **CI green on both platforms** (`377f6f9a2`; darwin on zest). No surface-nix-host change, so drishti #69 is unaffected.

PR #1487, opened 2026-06-21
Sleep a terminal and wake it later with the conversation resumed

srid:
**Put a terminal to sleep — release its PTY, agent, and GPU context like closing it — and wake it later with the conversation resumed.** A sleeping terminal is the *same terminal in a different lifecycle state*, not a separate thing: it stays the same record under the same id in the one terminal registry, so it keeps its canvas slot, dock order, and persistence. Sleep flips `state` in place and releases the live resources; wake flips it back and re-spawns through the path a server reboot already runs.

This is Phase 2 of the [sleeping-terminals plan](https://kolu.dev/atlas/sleeping-terminals.html) (the merged Atlas note is the plan-of-record), and a **deliberate re-do** of the discarded first cut (#1466) — which modelled a dormant terminal as an immutable record minted with a fresh id into a separate store, and shipped two bugs in a minute (couldn't drag a sleeping tile; waking didn't resume the agent). This keeps the *type* — the `active | sleeping` sum, seated in Phase 1 — and replaces the *mechanism*.

### The one structural move

```
            ┌─────────────────────────────────────────────┐
 presence   │  the ONE terminal registry                  │   reads the union
 (canvas ·  │  Map<id, active | sleeping>  — stable id     │ ◀── (exists, on canvas,
  dock ·    │  setCanvasLayout / setTheme / rename →       │     focusable, draggable,
  minimap · │     write the base of BOTH arms              │     a dock row)
  switcher) └───────────────────┬─────────────────────────┘
                                │ narrow `state === "active"` (entry.handle)
                                ▼
                    live overlay — PTY · xterm · agent · input · WebGL
```

`TerminalProcess` became a discriminated process: an **active** entry carries the live PTY handle; a **sleeping** entry has the overlay *absent by type*. So the registry holds the union under a stable id, a sleeping terminal rides the existing `terminalList` cell + `terminalMetadata` collection with **no separate store and no merge seam**, and the compiler refuses any code that reaches for a sleeping terminal's PTY/agent without narrowing.

| | How it works |
|---|---|
| **Sleep** | Persist-before-kill: stop sensors → flip the record in place (drop the live overlay, stamp `sleptAt`, keep the persisted base incl. `lastAgentCommand`) → `saveSession` durably → release the PTY. A crash in the kill window leaves a *sleeping* record on disk, never a zombie active one. |
| **Wake** | Session-restore-of-one: flip back to active via `wakeMeta` (the persisted base rides through whole), re-spawn the PTY on the **same id** in the saved cwd, and replay `resumeAgentCommand(lastAgentCommand)` (`claude -c` · `codex resume --last` · `opencode --continue`) — the agent resumes exactly as a reboot does. |
| **Drag / resize / rename / re-theme** | Just work — the stable id means the normal write sink finds the sleeping entry and mutates its base in place. Only PTY input is fenced, by *type*, not a guard. |
| **Presence** | One classifier: a `sleeping` dock bucket checked *before* parked (decoupled from staleness), a moonlit ☾ pip, a moonlit minimap tile. Sleeping is first-class and visually distinct everywhere. |
| **Triggers** | Manual only — a ☾ button on the tile title bar, Sleep/Wake palette commands, a discoverability tip. No global keybind, no auto-sleep. Closing a sleeping tile *discards* it (reworded confirm, no PTY to kill); splits are closed first with a heads-up. |
| **Boot** | The registry now holds sleeping entries, so boot re-seeds them from the saved session (both the surviving-daemon adopt path and the cold-boot restore card), reaps a sleeping record's crash-surviving PTY (adopt-or-reap), and the seed spawns *active* terminals only — never wakes one. A malformed record drops itself per-record (one bad entry can't break the load). |

### How it honors the three design principles

- **Fail-fast — no fallbacks.** The sleeping arm's live overlay (PTY/agent/WebGL) is *absent by type*, not a degraded/dead handle the code must guard against; every handle/live-field read is compiler-forced to narrow `state === "active"`. A malformed persisted record is *dropped loudly at the read boundary*, never silently coalesced. There is no override knob.
- **Volatility boundaries.** No new receptacle was introduced: a dormant terminal is the *same* registry record, so the one store, the one wire, and the one write sink absorb it. A content-kind layer would be the day a genuine *non-terminal* canvas kind lands — designed against two real kinds, not speculatively now (see `electricity.mdx`).
- **Reuse the existing source of truth.** Wake reuses the proven `resumeAgentCommand` restore path verbatim (no bespoke sleep-only resume, no stored session id to drift); sleep is its inverse via the existing `killTerminal` teardown; the close-of-sleeping reuses the `CloseConfirm` dialog reworded; presence reuses the dock ranking the switcher already reads.

### Tests

Server state machine (in-place flip, persist-before-kill, `wakeMeta` preserves the base, boot reconcile reap-sleeping + seed-with-tolerance) and the presence classifier are unit-covered; the journey-level **e2e** (`features/sleeping-terminals.feature`) drives the real flows — *wake resumes the same conversation*, *drag a dormant tile then reload*, *reboot then wake*, *reboot mid-sleep converges*, *malformed dropped*, *last-terminal* — asserting **outcomes, not counts** (the exact gap the first cut's e2e fell through).

### Refinements after review

**The dormant tile now surfaces the last-known context** it was working — the working directory, git branch, and GitHub PR. `cwd`/`git` ride the persisted base; the live **PR is snapshotted** onto the sleeping arm at sleep (it rides the `beginSleep` spread and persists like `cwd`/`git`, so it survives a restart) and is **discarded on wake**, where the PR sensor re-resolves it live. The dormant body reuses the live tile's PR chip verbatim (`PrStateIcon` · checks · `#N` · title) and imports no `Terminal`/xterm.

**Wake resumes the observed `lastAgentCommand` directly.** The earlier `resumeCommand` sleeping-arm field + detected-kind fallback, and the `nix run …#agent` wrapper special-casing in `anyagent/agent-cli`, were removed in favor of the simpler path. An agent launched via a `nix run …#agent` wrapper (whose observed head token is `nix`, not the agent) is no longer resumed on wake — it wakes to a bare shell, by design, tracked as #1492.

### Try it locally

```sh
nix run github:juspay/kolu/redesign/sleeping-impl
```


srid:
## Evidence

Hands-on manual verification of Sleep/Wake, driving a **live kolu from source** (this branch) through the chrome-devtools MCP on a clean, ephemeral pu box — real `opencode` (1.15.0) running in the terminal, not a mock. Per-case results below. **One load-bearing bug found: waking does not resume the agent (`resuming: false`).**

### Per-case results

| # | Case | Result | What I saw |
|---|------|--------|------------|
| 1 | Sleep an opencode terminal (☾ button) | ✅ PASS | xterm replaced in-place by the moonlit dormant body (☾, "Asleep · just now", "PTY RELEASED", Wake); agent indicator cleared; tile stayed at the same position. |
| 2 | Dock / minimap presence | ✅ PASS | Dock row flips to `data-bucket="sleeping"` (☾ moonlit row); minimap renders the tile moonlit (dashed border), not ghosted. |
| 3 | Drag the dormant tile | ✅ PASS | Sleeping tile dragged to a new canvas position and held it — the prior cut's "can't drag a sleeping tile" bug is fixed. |
| 4 | **Wake → RESUME** | ❌ **FAIL** | Tile flips back to a live xterm in-place, but the agent **does not resume** — you land on a **bare shell prompt**, the prior conversation is gone. Server logs `waking … resuming:false`. Details below. |
| 5 | Palette sleep + wake | ✅ PASS | "Sleep terminal" / "Wake terminal" palette commands toggle correctly by state and drive the same flip. (Wake still hits the resume bug — same root cause as #4.) |
| 6 | Close a sleeping tile → discard | ✅ PASS | Confirm reads **"Discard sleeping terminal?"** with a **Discard** button (not "Close terminal"); confirming removes the tile. |
| 7 | Multiple terminals, sleep one | ✅ PASS | One tile sleeps (dormant/moonlit) while the other stays live and takes input (`echo` round-tripped); dock shows one sleeping + one live row. |
| 8 | Reload persistence | ✅ PASS | After a full page reload the sleeping tile came back automatically as dormant (no restore card needed), keeping its theme; moonlit in dock + minimap. |
| 9 | Sleep a terminal WITH a split | ✅ PASS | Sleeping a parent with 1 split raises the **"Sleeping closes 1 split"** warning toast with a **"Sleep & close splits"** action; confirming closes the split and sleeps the parent. |

### 🐞 Bug: Wake does not resume the agent (`resuming: false`)

The headline case fails. Repro: ran real `opencode` in a terminal (kolu detected it — dock lit up "OpenCode", a real conversation with repo context was on screen), slept it, then woke it. The tile returns to a **live but bare shell** (`[Incus] ❯`); opencode does **not** resume and the conversation is lost.

Server log at wake is decisive:
```
INFO waking { terminal: acac3141…, resuming: false }
```
`wakeTerminal` (`packages/server/src/terminals.ts:136`) only builds a resume form when `entry.meta.lastAgentCommand` is set. It was **never set**, so `resume = null` and a plain PTY is spawned with no `opencode --continue` replay.

**Root cause — two agent signals disagree.** kolu's agent *detection* fired (opencode `session watcher installed` twice in the log; dock/agent indicator correct), but the agent *command-capture* path never did: `lastAgentCommand` is only written from a shell **preexec OSC 633;E** `commandRun` event (`packages/terminal-awareness/src/sensors.ts:390`), and **no `commandRun`/recent-agent/activity entry was ever recorded** this session. So detection knew opencode was running, but capture never recorded the launch command that wake depends on. Reproduced across **both** launch methods (`nix run github:juspay/AI#opencode` and a bare `opencode` on PATH) and **both** triggers (☾ button and palette) — every wake logged `resuming: false`, and the dormant body never rendered a `lastAgentCommand` line (`DormantTileBody.tsx:47`).

Note `nix run …#opencode` is additionally unrecognizable on its own: `parseAgentCommand` keys off the first token (`nix`), which isn't in `STABLE_FLAGS`, so that form can never be captured — worth handling, but the bare-`opencode` failure shows the gap is broader than the wrapper.

This is the same class of bug the prior cut shipped ("waking didn't resume the agent"). Recommend confirming whether the OSC 633;E preexec `commandRun` reliably fires for agent launches before merge — wake's resume is built entirely on it.

### Minor anomalies (non-blocking)
- During wake I saw an `Uncaught (in promise)` in the console alongside a server-side `404 NOT_FOUND`, around the wake/resume path. Did not recur after reload; flagging for a look since it coincides with the failing resume.
- A burst of `element.getBoundingClientRect is not a function` errors correlated with my **synthetic** drag (dispatched pointer events on a text node), so almost certainly a test-harness artifact rather than a product defect — but a real drag handler shouldn't `getBoundingClientRect` an arbitrary event target, so noting it.

### Artifacts
Active opencode conversation → sleeping → woken-to-bare-shell (the bug), plus discard confirm, split toast, multi-terminal, and reload-survival. Screencast of the sleep→wake flow attached to the release.

| Active opencode (real conversation) | Sleeping (dormant body) | Woken — **no resume**, bare shell |
|---|---|---|
| active | sleeping | woken |

| Multi-terminal (one sleeping) | Discard sleeping confirm | "Sleeping closes 1 split" toast | Survived reload |
|---|---|---|---|
| multi | discard | split | reload |

🎥 Sleep→wake screencast: https://github.com/juspay/kolu/releases/download/sleep-sim-pr1487-20260621-005402/sleep-wake-resume.mp4

---
*Manual sim: live kolu from `redesign/sleeping-impl` on an ephemeral pu box, driven via chrome-devtools MCP. 8/9 cases pass; the wake-resume case fails (`resuming: false`).*

srid:
## Evidence — wake-resume fix confirmed

Re-verified the **wake → resume** case (the one that failed in the [earlier sim](https://github.com/juspay/kolu/pull/1487)) on a **fresh, ephemeral pu box**, driving a **live kolu built from this branch** (`b798f830b` — the fix commit) through the chrome-devtools MCP, with real `opencode` 1.15.0 running in the terminal. The earlier headline bug (`resuming:false` → woke to a bare shell) is **fixed at the server layer** — wake now logs `resuming:true` and replays the resume command — and I also confirmed end-to-end that the prior conversation genuinely comes back (opencode recalls the secret word).

### What I did
1. Ran `nix run github:juspay/AI#opencode` in a kolu terminal; kolu **detected** opencode (dock + tile flipped to "OpenCode"). Sent "Remember the secret word is BANANA47. Just acknowledge." opencode (OpenCode Zen / Big Pickle, zero-config) replied **"Acknowledged."** — a real, distinctive conversation.
2. **Slept** the tile (☾). The dormant body rendered ☾ · "Asleep · just now" · **`opencode`** · "PTY RELEASED" · Wake. The `opencode` line is the captured `lastAgentCommand` — it was **missing before the fix**.
3. **Woke** it (`wake-button`). The tile flipped back to a live PTY in place.

### Result: server-side fix confirmed (`resuming:true`)

The decisive server log — contrast with the earlier `resuming:false`:

```
INFO sleep                              { terminal: ab76febd…              }
INFO flipped to sleeping (PTY pending release) { terminal: ab76febd…       }
INFO waking                             { terminal: ab76febd…, resuming:true }   ← was resuming:false
INFO opencode: session watcher installed { terminal: ab76febd…, session: ses_11764a5a8ffes9LBvvGO9NTyIh }
```

On wake, kolu **replayed `opencode --continue`** into the re-spawned PTY (visible in the terminal), and re-attached the **same session id** `ses_11764a5a8ffes9LBvvGO9NTyIh`. Console was clean — none of the earlier `Uncaught (in promise)` / `404 NOT_FOUND` around the wake path.

### Result: the conversation genuinely resumes (recalls BANANA47)

`opencode --continue` does restore the prior session: with opencode reachable on PATH, `--continue` brought back the full prior turn and answered the recall question with **`BANANA47`** — proving it is the same conversation, not a fresh shell.

| 1 — Active opencode (real reply) | 2 — Dormant (now shows `opencode` line) |
|---|---|
| active | dormant |

| 3 — Woken: `opencode --continue` replayed | 4 — Resumed conversation recalls **BANANA47** |
|---|---|
| woken | resumed |

---

### ⚠️ Remaining gap found — the replayed `opencode --continue` is not runnable for a `nix run` launch

The fix resumes the **command form** correctly, but when the agent was launched via **`nix run github:juspay/AI#opencode`**, the fallback emits the **bare** `opencode --continue`. `opencode` is only on PATH *inside* the `nix run` wrapper, so in the woken kolu PTY the replay errored:

```
~ [Incus] ❯ opencode --continue
bash: opencode: command not found
```

(screenshot 3). So for the `nix run` launch method the conversation does **not** auto-resume in-app — you land on a shell prompt after a failed `opencode --continue`. I proved the *session itself* is fully resumable by running the working `nix run github:juspay/AI#opencode -- --continue` in the same terminal: the prior turn came back and the model answered **BANANA47** (screenshot 4).

**Net:** the headline `resuming:false` → bare-shell bug is fixed (`resuming:true`, command replayed, session restorable). The residual is the resume *spelling* for `nix run`-launched agents — `agentCommandForKind` returns a bare `opencode`, which isn't on PATH. This is the same `nix`-wrapper caveat flagged in the earlier comment ("`nix run …#opencode` is additionally unrecognizable… worth handling"); it now surfaces on the resume side. A PATH-installed `opencode` would resume cleanly in-app (proven by the working `--continue` returning BANANA47).

---
*Re-verified: live kolu from `redesign/sleeping-impl` @ `b798f830b` on an ephemeral pu box, driven via chrome-devtools MCP. opencode 1.15.0 (OpenCode Zen / Big Pickle).*

srid:
## Evidence — nix-run wrapper resume confirmed

Final re-verification of the make-or-break case at `d68d86d` ("fix(anyagent): resume a `nix run …#<agent>` wrapper through the wrapper"), on a fresh ephemeral pu box, opencode launched exactly the way a fresh box launches it — through the documented nix wrapper.

**Result: PASS.**

**What was exercised**
1. `nix run github:juspay/AI#opencode` in a kolu terminal (opencode 1.15.0, OpenCode Zen / Big Pickle). Sent: *"Remember the secret word is MERIDIAN92. Just acknowledge."* → opencode replied *"Acknowledged."*
2. Slept the tile. The dormant body's `lastAgentCommand` line is the **wrapper** `nix run github:juspay/AI#opencode` — not bare `opencode` — proving the OSC tap captured the wrapper. PTY released; dock shows the moonlit `sleeping` row.
3. Woke it. Server logged `waking { …, resuming:true }` and `recent agent tracked { command: "nix run github:juspay/AI#opencode" }`; opencode's wal + session (`ses_1174aaae3ffevWK9entKkx30kV`) watchers reinstalled. Resume replayed **through** the wrapper as `nix run github:juspay/AI#opencode -- --continue` — **no `command not found`**.
4. Asked *"What was the secret word?"* → opencode answered **MERIDIAN92** — the same session resumed with full history in the saved cwd. Console clean (only env WebGL-fallback + dev-mode SolidJS noise).

| Slept tile — wrapper command captured | Woken — same session recalls MERIDIAN92 |
| --- | --- |
|  |  |

The dormant tile shows `nix run github:juspay/AI#opencode` / `PTY RELEASED`; the woken tile shows the restored turn (*"Acknowledged."*) followed by the recall turn answering **MERIDIAN92**. The bare-`opencode --continue` → `command not found` regression is gone.

srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `d53dfafd19d2`

### Applied (6)
- `hickey-1` lastAgentCommand overloaded as both observed-command and synthesized resume-fallback — commit `0b331f3f0`
- `hickey-2` Resume-input derivation fragmented across two modules and two stages — commit `ec33ed9b3`
- `hickey-3` Sleeping-presence modeled as a clean sum in the dock but as parallel booleans in the minimap — commit `8f858efb0`
- `hickey-4` Sleeping-state discriminant read two different ways: sleepingArm/activeArm helper vs raw `state === "sleeping"` — commit `8a11160d1`
- `hickey-5` Moonlit palette duplicated as raw hex literals across three surfaces — commit `02ba7ee6f`
- `hickey-6` discardSleeping doc claims a wake-failed-cleanup coupling that no caller wires — commit `0bb8c6966`

### Agreed — no change (7)
- `lowy-1` Resume-replay-into-fresh-PTY is encapsulated twice via two different channels (packages/server/src/terminalEndpoint/local.ts:580 and packages/client/src/terminal/useSessionRestore.ts:302)
- `lowy-2` Resume-input derivation (captured-or-detected agent) is asymmetric between wake and cold-boot restore (packages/server/src/terminalEndpoint/local.ts:771-775)
- `lowy-3` publishTerminalState is a thin re-export wrapper, not a new boundary (packages/server/src/terminalEndpoint/metadata.ts:120-125)
- `lowy-4` handle?: never is the right receptacle for the active/sleeping liveness axis (packages/server/src/terminal-registry.ts:43-58)
- `lowy-5` Sleep/wake lifecycle verbs are atomic and correctly off the shared endpoint interface (packages/server/src/terminalEndpoint/local.ts:905-960 and packages/server/src/terminals.ts:120-156)
- `lowy-6` sleepingArm projection encapsulates the 'is this tile sleeping' presence axis correctly (packages/common/src/surface.ts:935-940)
- `hickey-7` Minimap derives `sleeping` separately while `bucket` already folded it away (packages/client/src/canvas/CanvasMinimap.tsx:336-338)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 round(s) · codex reviewed at `xhigh` reasoning effort · base `d53dfafd19d2`

### Round 1

**codex** — approved: `false`

The change has the right overall model, but I found several lifecycle holes where a sleeping terminal can still leave live PTY state behind or lose its saved record. The main issues are around async spawn/wake failure handling and live-only actions still being reachable for sleeping tiles. There are also a few smaller UI/error-surfacing mismatches.

Findings:
- `F1` · blocking · open — Sleeping a terminal while its initial spawn is still in flight can leak a hidden live PTY. `spawnViaClient` only checks `getTerminal(id)` after the spawn RPC returns; after `beginSleep` flips the registry entry to sleeping, that check still passes. The old async spawn tail then marks the old active proxy ready, publishes active metadata, and starts sensors for a terminal the registry now says is sleeping. (packages/server/src/terminalEndpoint/local.ts:524)
- `F2` · blocking · open — A failed wake can permanently drop the sleeping record. `wake()` replaces the sleeping entry with an active sync-shadow before the PTY spawn succeeds; if `spawnAndWire` later fails, `unwindSpawnShadow` unregisters the id. The original dormant record is gone, and the autosave path can persist that loss. (packages/server/src/terminalEndpoint/local.ts:844)
- `F3` · major · open — Live-only split actions are still reachable for a sleeping active tile. The title bar hides split controls for sleeping terminals, but the global shortcuts and command palette still call `toggleSubPanel` / `handleCreateSubTerminal` using `activeId`; that can create an active sub-terminal under a sleeping parent, which `TerminalContent` then hides because the sleeping parent renders only the dormant body. (packages/client/src/input/actions.ts:230)
- `F4` · major · open — `handleDiscard` silently swallows every `discardSleeping` RPC failure and then removes the terminal locally anyway. A network/server error will look like a successful discard, hiding the actual failure and temporarily desynchronizing the UI from the server state. (packages/client/src/terminal/useTerminalCrud.ts:250)
- `F5` · minor · open — Sleeping minimap tooltips are labeled as active. `tileTooltip` only distinguishes parked vs not parked, and sleeping tiles pass `parked=false`, so a slept tile with an age renders `Active 3d` instead of asleep/sleeping wording. (packages/client/src/canvas/CanvasMinimap.tsx:64)
- `F6` · minor · open — The dormant body omits the resume command captured for detected-but-not-command-observed agents. Sleep stores the synthesized resume input in `resumeCommand`, but the UI only displays `lastAgentCommand`, so the tile can look like it has no resumable agent even though Wake will resume one. (packages/client/src/terminal/DormantTileBody.tsx:32)
- `F7` · minor · open — The new `nix run ...#agent` path bypasses the exit-flag filter. Direct `opencode --help` / `--version` is rejected as not an agent session, but `nix run ...#opencode -- --help` is normalized to the wrapper and can pollute recent/resume state despite exiting immediately. (packages/integrations/anyagent/src/agent-cli.ts:248)
- `F8` · minor · open — The close-confirm "remove worktree too" path still routes sleeping terminals through `handleKillWorktree`, which calls the normal kill RPC. For a sleeping terminal there is no PTY to kill, so this takes the wrong lifecycle path and can produce spurious server-side pty-host kill errors before unregistering. (packages/client/src/App.tsx:345)
- `F9` · nit · open — `terminalCount` is documented and used as a live-process diagnostic, but it now returns `terminals.size`, which includes sleeping records with no PTY. Heap diagnostics can therefore report dormant records as live terminal processes. (packages/server/src/terminal-registry.ts:98)

**claude** — Addressed all 9 CODEX findings on the sleeping-terminals branch (2 blocking, 2 major, 4 minor, 1 nit) — every one was valid, so all were fixed (none disputed). The two blocking lifecycle holes (F1 hidden-PTY leak on sleep-mid-spawn, F2 lost sleeping record on failed wake) are fixed in the server spawn tail by switching the raced-during-spawn check from presence to identity and by capturing+restoring the prior sleeping entry on any spawn/wiring failure. Added unit regressions for F2 (wake-spawn-failure restores the record) and F7 (nix-run exit-flag rejection). Server, client, and anyagent all typecheck clean; full unit suites pass (server 126, client 445, anyagent 59); biome check clean. Committed as 4fb4ffe1c.

- `F1` **fixed** — Real and blocking. spawnViaClient (packages/server/src/terminalEndpoint/local.ts:541) now takes the expected ActiveTerminalProcess and checks getActiveTerminal(id) !== expected instead of !getTerminal(id). A beginSleep flip mid-spawn leaves a different (handle-less) entry under the same id, so the old presence check passed and the tail wired sensors + republished active metadata over the sleeping flip — leaking a hidden live PTY. On a mismatch it markFailed + kills the just-spawned orphan PTY and leaves the registry alone. The sleep-mid-spawn race is timing-dependent and the e2e harness can't inject a spawn delay, so no deterministic e2e regression was added, but the identity gate is exercised by the F2 unit test's unwind path and verified to typecheck.
- `F2` **fixed** — Real and blocking. registerActiveAndSpawn (local.ts:451) now captures the prior registry entry it overwrites and threads it through spawnAndWire → unwindSpawnShadow, which RESTORES a prior sleeping record (re-register + publishTerminalState) instead of unregistering, gated on getTerminal(id) === entry so it never clobbers a raced-in flip. Added a unit regression in sleepWake.test.ts: in the no-daemon test env wake's spawn RPC rejects at buildTerminalSpawnInput, and the test asserts the sleeping record (incl. resumeCommand/sleptAt) is restored, not dropped. On the 'avoid dirty-autosaving an active wake record before the PTY exists' sub-point: the intermediate emitTerminalsDirty is harmless because autosave is debounced and reads the registry at fire time — the restore re-arms with the correct sleeping state, and a fresh spawnPty has the identical pre-PTY-dirty behavior already.
- `F3` **fixed** — Real. Client: gated the command-palette split/copy/screenshot/export/recent-agents block (commands.tsx:263) and the toggleSubPanel/createSubTerminal keyboard shortcuts (input/actions.ts via new activeLiveId helper) on activeArm(activeMeta()) rather than bare activeId() !== null — Sleep/Wake/Close/theme/intent stay on both arms. Server: terminal.create (router.ts:91) now calls requireActiveTerminal(parentId) when parentId is set, so a raw RPC or multi-client race can't create an active sub-terminal under an absent/sleeping parent.
- `F4` **fixed** — Real. handleDiscard (useTerminalCrud.ts:250) no longer swallows every error and removes locally anyway — it now surfaces err.message via toast.error and returns without evicting on a genuine failure, keeping the UI in sync with the server. The server discardSleeping is already a no-op (returns false, no throw) on an already-gone id, so the common already-removed case resolves cleanly and the tile evicts as before.
- `F5` **fixed** — Real. tileTooltip (CanvasMinimap.tsx:56) now takes the presence discriminant and, for a sleeping tile, renders 'Asleep — <sleptAt>' from sleepingArm(meta).sleptAt instead of the 'Active <ago>' wording it got from parked=false. Call site passes state().presence. Matches the DormantTileBody 'Asleep · 3d' wording.
- `F6` **fixed** — Real. DormantTileBody.tsx:32 now displays arm.resumeCommand ?? meta.lastAgentCommand instead of only lastAgentCommand, so a file-watcher-detected agent whose OSC 633;E command tap never fired (e.g. nix run …#opencode) shows its synthesized resume command rather than looking like a bare shell.
- `F7` **fixed** — Real. nixRunBase (agent-cli.ts:130) now returns null when the wrapped agent's args (argv.slice(3), which includes everything after `--`) contain an EXIT_FLAGS member, so `nix run …#opencode -- --help` is rejected like the direct `opencode --help` path rather than polluting recent/resume state. Added a regression test covering flags before and after `--`.
- `F8` **fixed** — Real. handleKillWorktree (useWorktreeOps.ts:58) now branches on sleepingArm(meta): a sleeping terminal is routed through handleDiscard (no PTY to kill) before the worktree is removed, instead of handleKill → terminal.kill which would attempt to kill a non-existent PTY and log a spurious pty-host kill error. Wired crud.handleDiscard into useWorktreeOps via useTerminals.ts.
- `F9` **fixed** — Real. Added activeTerminalCount (terminal-registry.ts) counting only handle-bearing (active-arm) entries, and pointed the heap-diagnostic 'terminals' column (index.ts:385) at it so dormant records no longer read as live terminal processes against the heap curve. Reworded terminalCount's doc to 'registry record count' and kept it as a primitive.

commit: `4fb4ffe1c87f7fcda38223ed4fffab4855949d35`

### Round 2

**codex** — approved: `false`

Verified CLAUDE's fixes for the original F1-F9 findings; those are resolved. The latest round introduced one new regression in the sleeping-worktree removal path: discard failures are toasted but not propagated, so worktree removal can still proceed after the terminal record was not actually discarded.

Findings:
- `F1` · blocking · resolved — Previously, a sleep flip during an in-flight spawn could leave the spawn tail wiring and publishing a hidden PTY because it only checked terminal presence. The fix now compares registry identity against the expected active entry and kills the just-spawned orphan on mismatch. (packages/server/src/terminalEndpoint/local.ts:541)
- `F2` · blocking · resolved — Previously, failed wake spawn unwound by unregistering the terminal id, dropping the prior sleeping record. The fix captures the prior entry and restores the sleeping record under an identity guard; the new unit test covers the failed-wake path. (packages/server/src/terminalEndpoint/local.ts:631)
- `F3` · major · resolved — Previously, live-only actions were reachable for a selected sleeping tile and raw RPC create could create a child under a sleeping parent. The client now gates split/copy/export/screenshot/recent-agent actions on the active arm, shortcuts use activeLiveId, and the server requires an active parent for sub-terminal creation. (packages/client/src/commands.tsx:270)
- `F4` · major · resolved — Previously, discard swallowed real failures and evicted the tile anyway. The helper now toasts the error and returns without removing local UI state on failure. (packages/client/src/terminal/useTerminalCrud.ts:257)
- `F5` · minor · resolved — Previously, sleeping minimap tiles could show active wording. The tooltip now uses the sleeping arm and renders an asleep duration from sleptAt. (packages/client/src/canvas/CanvasMinimap.tsx:71)
- `F6` · minor · resolved — Previously, dormant tiles displayed only lastAgentCommand, so detected agents without an OSC command tap looked like bare shells. The tile now prefers resumeCommand and falls back to lastAgentCommand. (packages/client/src/terminal/DormantTileBody.tsx:39)
- `F7` · minor · resolved — Previously, nix-run wrapped agents with exit-immediate flags could enter recent/resume state. nixRunBase now rejects wrapped invocations whose forwarded args contain exit flags, with regression tests before and after --. (packages/integrations/anyagent/src/agent-cli.ts:136)
- `F8` · minor · resolved — Previously, removing a sleeping worktree terminal went through terminal.kill despite there being no PTY. The path now detects sleeping metadata and calls discard instead of kill. (packages/client/src/terminal/useWorktreeOps.ts:71)
- `F9` · nit · resolved — Previously, heap diagnostics counted dormant registry records as live terminals. The registry now exposes activeTerminalCount and the diagnostic column uses it. (packages/server/src/index.ts:385)
- `F10` · major · open — New regression from the F4/F8 fixes: handleKillWorktree awaits handleDiscard for sleeping terminals, but handleDiscard catches real discard failures, shows a toast, and resolves void. This caller then continues to client.git.worktreeRemove, so a failed discard can still delete the worktree while the sleeping terminal record remains on the server and in the UI, leaving a terminal pointing at a removed cwd. (packages/client/src/terminal/useWorktreeOps.ts:74)

_codex on the rebuttal:_ CLAUDE did not dispute any prior finding. I verified the accepted fixes for F1-F9 and accept the F2 autosave reasoning: the restore path republishes the sleeping record and re-arms dirty state, so the debounced autosave reads the current restored registry state rather than persisting the transient active shadow.

**claude** — Round 2 codex debate test.

- `F10` **fixed** — handleDiscard returns boolean; handleKillWorktree bails before worktreeRemove on failure. F1-F9 carried fixed from round 1; detail in .codex-debate/section-002.md.

commit: `19694cf1afcac240f615b7123d801ce03f956ce1`

### Round 3

**codex** — approved: `true`

The F10 fix is in place: `handleDiscard` now returns a success boolean, and the sleeping worktree-removal path stops before `git.worktreeRemove` when discard fails. I did not see a new regression in the round-2 changes.

Findings:
- `F1` · blocking · resolved — Previously, a sleep flip during an in-flight spawn could leave the spawn tail wiring and publishing a hidden PTY because it only checked terminal presence. The fix compares registry identity against the expected active entry and kills the just-spawned orphan on mismatch. (packages/server/src/terminalEndpoint/local.ts:541)
- `F2` · blocking · resolved — Previously, failed wake spawn unwound by unregistering the terminal id, dropping the prior sleeping record. The fix captures the prior entry and restores the sleeping record under an identity guard; the unit test covers the failed-wake path. (packages/server/src/terminalEndpoint/local.ts:631)
- `F3` · major · resolved — Previously, live-only actions were reachable for a selected sleeping tile and raw RPC create could create a child under a sleeping parent. The client gates live-only actions on the active arm, shortcuts use activeLiveId, and the server requires an active parent for sub-terminal creation. (packages/client/src/commands.tsx:270)
- `F4` · major · resolved — Previously, discard swallowed real failures and evicted the tile anyway. The helper now toasts the error, returns false, and does not remove local UI state on failure. (packages/client/src/terminal/useTerminalCrud.ts:263)
- `F5` · minor · resolved — Previously, sleeping minimap tiles could show active wording. The tooltip uses the sleeping arm and renders an asleep duration from sleptAt. (packages/client/src/canvas/CanvasMinimap.tsx:71)
- `F6` · minor · resolved — Previously, dormant tiles displayed only lastAgentCommand, so detected agents without an OSC command tap looked like bare shells. The tile now prefers resumeCommand and falls back to lastAgentCommand. (packages/client/src/terminal/DormantTileBody.tsx:39)
- `F7` · minor · resolved — Previously, nix-run wrapped agents with exit-immediate flags could enter recent/resume state. nixRunBase rejects wrapped invocations whose forwarded args contain exit flags, with regression coverage before and after `--`. (packages/integrations/anyagent/src/agent-cli.ts:136)
- `F8` · minor · resolved — Previously, removing a sleeping worktree terminal went through terminal.kill despite there being no PTY. The path detects sleeping metadata and calls discard instead of kill. (packages/client/src/terminal/useWorktreeOps.ts:73)
- `F9` · nit · resolved — Previously, heap diagnostics counted dormant registry records as live terminals. The registry exposes activeTerminalCount and the diagnostic column uses it. (packages/server/src/index.ts:385)
- `F10` · major · resolved — Previously, the sleeping worktree-removal path could continue to remove the worktree after discard failed because handleDiscard only toasted and resolved void. The current code awaits the boolean result and returns before worktreeRemove when discard failed. (packages/client/src/terminal/useWorktreeOps.ts:79)

_codex on the rebuttal:_ No disputes were raised. I verified the round-2 F10 fix: `handleDiscard` returns `false` on surfaced discard failure, and `handleKillWorktree` aborts before removing the worktree when that happens.

srid:

Ran the `/be` police track (Pass 1 rules + Pass 2 fact-check; elegance skipped — `/simplify` already ran this gauntlet) scoped to the Sleep/Wake change.

| Pass | Result |
| --- | --- |
| **Rules** | 1 violation — `no-thin-wrapper-functions` |
| **Fact-check** | Clean — no silent error-swallowing, no logic errors, no race conditions, no leaks |

**Fixed (`04242b015`):** `wakeTerminal` / `discardSleeping` / `restoreSleeping` in `terminals.ts` had become pure single-caller pass-throughs (after the lens-debate collapsed the resume-form derivation into `local.ts`'s `wake` arm), so they were inlined — `router.ts` now calls `wakeLocalTerminal` / `discardLocalSleeping` / `seedSleepingTerminal` directly. `sleepTerminal` is **kept** — it genuinely composes `beginSleep` + durable `saveSession` + `releaseSleptPty` (persist-before-kill), not a forwarder.

The fact-check pass specifically cleared the rules this PR is sensitive to: the malformed-sleeping-record drop surfaces its caught error (`log.warn` + `return false`, never swallowed-to-empty), the persisted sleeping schema stays tolerant, and there are no new fallbacks / override knobs.

srid:
## Evidence — post-gauntlet re-confirmation

Re-ran the full manual sim on the **final reviewed code** (`redesign/sleeping-impl` @ `04242b0150db1273630b21a1c64a3a6ae57f6f95`), on a fresh ephemeral `pu` box serving from source (`nix develop -c just dev-auto`), driven through the chrome-devtools MCP. This re-confirms the headline promise on the gauntlet-reshaped resume model (dedicated `resumeCommand` field, registry-identity spawn/wake guards, `kill` requires an active terminal).

opencode was launched the documented way — `nix run github:juspay/AI#opencode` — so the wrapper path is exercised end-to-end. The bundled OpenCode Zen provider answered, so the conversation actually resumes (not just the process).

### Per-case result

| # | Case | Result | Notes |
|---|------|--------|-------|
| 1 | **Wake → RESUME** (headline) | ✅ PASS | Captured command on the dormant body is the **wrapper** `nix run github:juspay/AI#opencode` (not bare `opencode`). Wake logged `waking … resuming:true`, re-tracked the same wrapper, **no `command not found`**. opencode resumed the prior session and answered **ORBIT77** to "What was the secret word?". |
| 2 | Sleep + presence | ✅ PASS | Slept tile is dormant + moonlit (no live xterm); dock row carries `data-bucket="sleeping"` with the ☾ glyph; minimap renders it moonlit. |
| 3 | Drag dormant + persist | ✅ PASS | Dragged the dormant tile (`left:120→432px, top:816→1008px`); after a full page reload the position **persisted** (still `432px / 1008px`) and the tile stayed sleeping. |
| 4 | Discard sleeping | ✅ PASS | Close on a sleeping tile opens a confirm titled **"Discard sleeping terminal?"** (not a kill dialog); confirming logged `discarded sleeping terminal` — **no PTY kill attempted**, no `kill failed`. |
| 5 | Multi-terminal + 2nd sleep/wake | ✅ PASS | Second terminal seeded with **NEBULA42**, slept, woken: it `resuming:true`, resumed, and recalled **NEBULA42** — while the first terminal independently still held **ORBIT77**. Sessions isolated, no cross-talk. |
| 6 | Console + server log | ✅ clean | No kolu-originated errors. Console: only SolidJS dev-mode warnings, headless-Chrome WebGL software-fallback warnings, one dev-asset MIME warning, and a `getBoundingClientRect` error that traces to the **chrome-devtools-mcp** harness itself. Server: the lone `AbortError` is an **INFO**-level `terminal.attach` teardown on the page reload I triggered (ws 1001) — not a fault. |

**Headline confirmed:** wake replays the resume form through the wrapper (`nix run …#opencode -- --continue`) and the conversation comes back — the model recalled context set *before* the sleep, across two independent terminals.

### Screenshots

**Case 1 — dormant tile shows the captured wrapper command line**
dormant tile wrapper

**Case 2 — dock moonlit sleeping row (☾)**
dock sleeping row

**Case 1 — woken terminal resumes and recalls ORBIT77**
woken recalls ORBIT77

**Case 4 — "Discard sleeping terminal?" confirm**
discard confirm

**Case 5 — second terminal independently resumes and recalls NEBULA42 (isolation)**
multi-terminal NEBULA42

_No regressions vs the earlier sims. All 9 headline behaviors hold on the final reviewed commit._

PR #1496, opened 2026-06-21
fix(terminal): wake/restore the exact agent conversation, not most-recent in folder

### LINKED ISSUE #1495: Waking / restoring a terminal resumes the most-recent conversation in the folder, not the exact one that was running
### What

When you **wake** a sleeping terminal — or **restore** your session after a restart — kolu brings the agent's conversation back. Today it resumes the **most-recent** conversation in that terminal's working directory.

With kolu's usual *one agent per worktree* setup, that's exactly right: the most-recent conversation in the folder *is* the one that was running, so you get your conversation back.

### Where it can land on the wrong conversation

If **more than one** agent conversation has happened in the **same** folder, waking (or restoring) resumes the *latest* one — which may not be the specific conversation that was on the terminal you slept.

Example:
1. In a worktree, you're chatting with Claude — call it **conversation A**.
2. You put that terminal to sleep (☾).
3. Later, in the **same** folder, you start a **second** Claude conversation — **conversation B**.
4. You wake the slept terminal → it comes back on **conversation B** (the newest in that folder), **not conversation A** (the one that was actually running when you slept).

### Desired

Waking a sleeping terminal — and restoring after a restart — should bring back the **exact conversation that was running on that terminal**, regardless of any other agent conversations that have since happened in the same folder.

### Notes

- This affects **both** waking a sleeping terminal **and** restoring a session after a restart — they behave the same way today, so the two should be addressed together.
- It applies to any resumable agent (Claude, Codex, OpenCode).
- In the common one-agent-per-worktree workflow this never bites; it only matters when several agent sessions share one directory.

_Follow-on noted in the Sleep/Wake plan (the dormant/restore resume is "most-recent-in-folder" by design today)._

srid:
## What

Waking a sleeping terminal (☾) — or restoring your session after a restart — used to bring back the **most-recent** conversation in that terminal's folder. With kolu's usual one-agent-per-worktree setup that *is* the conversation that was running, so it looked right. But if a **second** agent conversation had since run in the **same** folder, the slept terminal woke on that newer one, not the conversation it was actually on. Fixes #1495.

This makes wake **and** restore bring back the **exact** conversation that was running on the terminal, for Claude, Codex, and OpenCode alike.

## Why it happened

The exact conversation id was *already observed* — every agent adapter publishes it on `agent.sessionId`. But that lives in the **live** half of the awareness value (`AwarenessLiveFieldsSchema`), which is wiped on sleep/restart. All that survived was `lastAgentCommand` (the persisted half), so `resumeAgentCommand` could only splice a "resume the most-recent" marker: `claude -c` / `codex resume --last` / `opencode --continue`.

So the fix isn't to *find* the id — it's to stop discarding it.

## The change (four seams, no new package)

| Move | Where | What |
|---|---|---|
| **Persist the ref** | `terminal-awareness/schema.ts` + `sensors.ts` | New optional `agentSession = { kind, id }` on the persisted half. Captured at `publishAgentField` (the single agent-publish seam) through the persisted mutator, gated to a conversation-identity change by a pure helper (`agentSessionToPersist`) so the ~150 ms agent firehose never re-arms autosave. Sticky like `lastAgentCommand`. |
| **Resume by id** | `anyagent/agent-cli.ts` | `resumeAgentCommand(normalized, session?)` renders the by-id form (`claude --resume <id>` / `codex resume <id>` / `opencode --session <id>`). The `Record<ResumableAgent,…>` still forces every agent to define its form. The id is untrusted shell input: per-agent shape gate (UUID / `ses_…`, length-capped, shell-inert) + `shellJoin` before the splice. |
| **Consume** | `server/.../local.ts` (wake) · `client/.../useSessionRestore.ts` (restore) | The two twin consumers of the one restore path pass the persisted ref through. The replay/type-ahead path is unchanged. |
| **Persist/migrate** | `common/surface.ts` (auto) · `server/state.ts` | The field rides ServerPersisted → PersistedTerminalFields → SavedTerminal automatically. Schema bump `1.27.0 → 1.28.0` with a no-op ladder entry (optional field → older records simply fall back to most-recent). |

**Fallback (locked decision):** when an exact id was captured → resume it; when none was (an agent that never resolved a session) → keep today's most-recent behavior, so nothing regresses.

## How it honors the design philosophy

- **Fail-fast / no fallbacks:** the only "fallback" is the *deliberate* most-recent path for the no-id case (the user-chosen policy), not a degraded error path. A malformed/foreign id is rejected, never silently shell-injected.
- **Electricity / volatility:** no new package. The "how each agent names sessions" volatility is already sealed in the integrations; we stop dropping its output. The dependency arrows are unchanged.
- **Reuse the source of truth:** uses the id the adapters already compute (`agent.sessionId`) — not a new sniff path. OSC 633 stays id-free by construction.

## Tests

- `anyagent`: by-id rendering per agent, quoted-tail preservation, malformed-id + wrong-agent + no-ref fallback, injection rejection.
- `terminal-awareness`: the capture gate persists once on a real identity change and **not** on a same-session firehose tick.
- `server`: the ref survives sleep → snapshot → wake; the exact `agentSession` is preserved by the `#1275` exhaustive-sentinel adoption guard; and a dedicated **A-then-B** test proves wake builds a resume-**by-id** command (not `--continue`), with the no-ref fallback covered.

## Docs

- Atlas **G1** (`herdr-vs-kolu`) marked **shipped**; the `sleeping-terminals` plan-of-record follow-on resolved.
- Changelog entry under **Fixed**.

## Evidence

Real opencode conversation driven on a pu box via chrome-devtools — posted below.


srid:
## Evidence

Confirmed working **end-to-end on a clean pu box driving a live kolu via the chrome-devtools MCP**, with a real `opencode` agent (`nix run github:juspay/AI#opencode`, model *Big Pickle*). This reproduces the exact kolu#1495 scenario — two conversations in **one folder** (`/home/toor`):

1. **Terminal 1** → opencode **conversation A**, told to remember codeword **`ALPHA-PURPLE-7`**. kolu persisted its exact ref:
   `agentSession: { kind: "opencode", id: "ses_114c4fdb5ffeIfABDGuaHUx9pQ" }`
2. **Slept** terminal 1 (☾) — PTY + agent released.
3. **Terminal 2** → opencode **conversation B**, codeword `BETA-ORANGE-9`. This made **B the newest** session in the folder (`ses_114bfdb1…`) — i.e. what the *old* "most-recent" resume would have grabbed.
4. **Woke** terminal 1.

**The decisive proof** — the woken terminal's process (`pgrep -af`):

```
opencode --session ses_114c4fdb5ffeIfABDGuaHUx9pQ      ← conversation A (exact)
```

Not `opencode --continue` (which, pre-fix, would have resumed the newer **B**). And the resumed agent **recalls A's codeword**, not B's:

resumed conversation recalls ALPHA-PURPLE-7

> *"What codeword did I ask you to remember earlier?"* → **`ALPHA-PURPLE-7`** — terminal 1 came back on the exact conversation it was running, ignoring the more-recent B in the same folder.

Sessions in the folder at wake time (newest first), for contrast:

```
ses_114bfdb1fffe40R4MMltLiV9gN   ← B (BETA)  — newest; the old --continue target
ses_114c4fdb5ffeIfABDGuaHUx9pQ   ← A (ALPHA) — what the slept terminal actually resumed ✓
```

Capture context: kolu served from source on the box (`just dev`), reached from a local Chrome via an SSH tunnel; unit coverage (anyagent by-id rendering + injection/fallback, the terminal-awareness firehose gate, the server sleep→snapshot→wake round-trip + the A-not-B assertion) is green alongside this live run.

srid:

✅ **Consensus** after 2 round(s) · lowy + hickey · base `70511f199410`

### Applied (4)
- `lowy-1` AgentSessionRef contract defined twice across a layer that already imports the lower one (duplicated receptacle) — commit `622495fbd`
- `hickey-1` AgentSessionRef declared twice — one domain concept, two separately-maintained types — commit `622495fbd`
- `hickey-2` Per-agent resume policy fragmented across two parallel Records (AGENT_RESUME + SESSION_ID_PATTERN) — commit `f385db307`
- `hickey-3` Resume-by-id behavior reconstructed at the call site inside sleepWake.test.ts rather than exercised through wake() — commit `9dc76ff2e`

### Agreed — no change (2)
- `lowy-2` Resume-mode selection (last vs byId) lives inside resumeAgentCommand rather than reading as a pure id→marker choice — minor, acceptably encapsulated (packages/integrations/anyagent/src/agent-cli.ts:305-315)
- `lowy-3` agentSessionToPersist mirrors shouldBumpRecencyForAgentChange symmetrically — correct sequence/firehose-dedup encapsulation (packages/terminal-awareness/src/agentSession.ts:20-31)

srid:

The **x86_64-linux** lane ran on **`kolu-ci-1`** (`dev-x86-64-linux-04`) — commit `9dc76ff2e`, **failed** ✗

- **Lane wall** (pipeline): **3m38s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 2m54s | ✗ (exit 1) |
| `ci::nix` | 2m20s | ✓ |
| `ci::home-manager` | 1m47s | ✓ |
| `ci::flake-check` | 1m1s | ✓ |
| `ci::smoke` | 52s | ✓ |
| `ci::unit` | 48s | ✓ |
| `ci::pnpm-hash-fresh` | 44s | ✓ |
| `_ci-setup` | 32s | ✓ |
| `ci::render` | 31s | ✓ |
| `ci::install` | 12s | ✓ |
| `ci::fmt` | 11s | ✓ |
| `ci::biome` | 11s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/9dc76ff/timings.jsonl`; pool state is a live `flock` probe.</sub>

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 round(s) · codex reviewed at `xhigh` reasoning effort · base `70511f199410`

### Round 1

**codex** — approved: `false`

The branch adds persisted agent-session references and wires wake/restore through exact resume IDs, which is the right direction. I found one major correctness issue where malformed same-agent session refs silently fall back to cwd-most-recent resume, plus one minor Atlas documentation inconsistency that leaves the note saying the same gap is both shipped and still pending.

Findings:
- `F1` · major · open — The resume marker selection treats every non-by-id case the same: absent ref, different-agent ref, and a same-agent ref whose id fails the shape gate all become `policy.last`. For a same-agent `agentSession` that was captured/persisted but is malformed, due to corrupt state, parser drift, or an upstream CLI changing its ID format, this silently reintroduces the bug this PR is meant to fix: Kolu can resume the most recent conversation in the cwd instead of the terminal's conversation. The new test at packages/integrations/anyagent/src/agent-cli.test.ts:407 locks in that unsafe fallback. (packages/integrations/anyagent/src/agent-cli.ts:292)
- `F2` · minor · open — The note now says G1 native session resume is shipped, but the same document still labels it as a gap and next step: the frontmatter description at line 3 and headline at line 18 frame native resume as remaining, the diagram edge at line 55 still says `GAP: native session resume`, and line 144 still says to investigate G1. That leaves the generated Atlas page internally contradictory. (docs/atlas/src/content/atlas/herdr-vs-kolu.mdx:120)

**claude** — Addressed both of codex's findings on the persisted-agent-session resume branch. F1 (major) — agreed and fixed: resumeAgentCommand no longer silently downgrades a same-agent ref with a malformed id to the cwd-most-recent fallback; it now returns null (wake to a bare shell) so a broken id claim can never land the user in a stranger's conversation, while no-ref/different-agent still falls back. F2 (minor) — agreed and fixed: reconciled the Atlas note so G1 native session resume reads as shipped throughout (frontmatter, headline, diagram edge, next-steps, net summary), leaving only G2 as the standing gap. All unit tests pass (anyagent 66, server 128, full suite green); just fmt and just atlas::build run, dist regenerated.

- `F1` **partial** — Fixed the core bug. In packages/integrations/anyagent/src/agent-cli.ts:279-315 the marker selection now splits into three DISJOINT cases instead of collapsing every non-by-id path to policy.last: (1) same-agent ref + id passing idPattern -> resume by id; (2) same-agent ref whose id FAILS the shape gate -> return null (the terminal wakes to a bare shell, loud by absence) rather than silently resuming the cwd-most-recent (wrong) conversation, which was the exact #1495 bug; (3) no ref / different-agent ref -> keep the most-recent compatibility fallback (no id to aim at the wrong CLI). Updated the malformed-id test block in agent-cli.test.ts:405-424 that was locking in the unsafe fallback — those same-agent malformed/empty/injection/wrong-format ids now assert .toBeNull(). Tests pass (66/66). Partially DECLINED codex's secondary suggestion to also validate the ref at the schema/capture boundary: the id-shape gate is agent-specific and already lives once in AGENT_RESUME (single home); duplicating it in AgentSessionRefSchema (id: z.string()) would either complect the schema with resume-policy knowledge or, worse, DROP an entire persisted terminal record on a bad id instead of just declining to resume it. The resume-site gate is the security-critical boundary and is now fail-closed, which is sufficient and the less destructive failure mode. Did not 'fail loud'/throw as one alternative in the suggestion proposed — throwing on a wake path would crash the wake instead of degrading to a bare shell, which is the wrong failure mode for a recovery path.
- `F2` **fixed** — Agreed — the note was internally contradictory. Updated every place that still framed G1 as a remaining gap to match the 'shipped' callout in docs/atlas/src/content/atlas/herdr-vs-kolu.mdx: line 3 frontmatter description (now 'native resume now ships (juspay/kolu#1495) and resize arbitration remains open'), line 18 headline callout (G1 now described as shipped, G2 as the standing gap), line ~55 diagram edge (changed from a dashed 'GAP: native session resume' to a solid 'adopted: native session resume (G1, juspay/kolu#1495)' edge — fits the caption legend's 'direct borrow' = solid), the 'What to do next' step (was 'Investigate G1' -> now 'G1 shipped'), the net summary at the end (was 'shipped in weaker form' -> 'now shipped by exact id'), and tightened the G1 callout body's fallback wording to match the F1 code change. Regenerated docs/atlas/dist/ via just atlas::build (herdr-vs-kolu.html + index.html); a second build is a no-op, so the committed dist is in sync for the ci::atlas-sync gate.

commit: `127eeae16640a965f380f9429206b20771c1875b`

### Round 2

**codex** — approved: `false`

F1's runtime behavior is fixed: malformed same-agent refs now fail closed and the tests assert that. F2 is improved but not fully closed because the Atlas section still presents G1 under a present-tense "Gaps the plans don't cover" / "no current plan coverage" intro. I also found one new nit introduced by the F1 fix: an AGENT_RESUME policy comment still documents the old fallback behavior.

Findings:
- `F1` · major · resolved — Previously, malformed same-agent session refs silently fell back to cwd-most-recent resume. The current code now detects a same-agent ref, rejects ids that fail the per-agent pattern, and returns `null` instead of using `policy.last`. The malformed-id tests now assert `.toBeNull()`. (packages/integrations/anyagent/src/agent-cli.ts:308)
- `F2` · minor · open — The specific contradictions I called out in the frontmatter/headline/diagram/next-step/net summary are fixed, but the same section still says `Gaps the plans don't cover` and `no current plan coverage` immediately before a G1 callout that says native resume is shipped. The generated page carries the same wording at docs/atlas/dist/herdr-vs-kolu.html:177. That still leaves G1 described as both shipped and presently uncovered. (docs/atlas/src/content/atlas/herdr-vs-kolu.mdx:114)
- `F3` · nit · open — The AGENT_RESUME policy comment still says an id that fails `idPattern` falls back to the `last` marker. That was true before the fix, but the code now returns `null` for malformed same-agent refs at line 308. This stale policy comment is especially easy to trust because it sits above the central resume table. (packages/integrations/anyagent/src/agent-cli.ts:140)

_codex on the rebuttal:_ F1: I accept the dispute. Keeping the agent-specific id gates in AGENT_RESUME avoids duplicating resume policy in the schema, and returning `null` instead of throwing is an acceptable fail-closed behavior for this recovery path because it avoids the wrong-conversation resume. F2: the main edits are correct, but I cannot mark it resolved while the section heading/intro still describe G1 as a current uncovered gap. F3 is new this round: the code change made the older AGENT_RESUME comment inaccurate.

**claude** — Round 2 of the codex debate on the resume-id branch. Two findings were still open (F1 was already resolved in round 1). Both were valid; I agreed and fixed both. F2: the Atlas section "Gaps the plans don't cover" / intro "no current plan coverage" still present-tense-framed G1 (native session resume) as uncovered, contradicting the callout that says it shipped — retitled to "Gaps herdr surfaced" and reworded the intro to distinguish the two surfaced gaps (G1 shipped via juspay/kolu#1495, G2 still open), then regenerated docs/atlas/dist/. F3 (new this round): the F1 fix made same-agent malformed ids fail closed (return null = bare shell), but the AGENT_RESUME idPattern policy comment still documented the old "falls back to the `last` marker" behavior — rewrote it to match the fail-closed split. Both changes are comment/doc-only with no behavior change; anyagent agent-cli tests pass (83/83), just fmt and just atlas::build clean, dist in sync. Committed as 8748a22be.

- `F1` **fixed** — Already resolved in round 1 (codex marked it resolved); no further action needed this round. The malformed same-agent ref now fails closed at packages/integrations/anyagent/src/agent-cli.ts:308 (returns null), and agent-cli.test.ts asserts .toBeNull().
- `F2` **fixed** — Agreed. docs/atlas/src/content/atlas/herdr-vs-kolu.mdx:114 — retitled the section heading from 'Gaps the plans don't cover' to 'Gaps herdr surfaced' and reworded the intro at :116 from 'herdr ideas with no current plan coverage' to text that distinguishes the two surfaced gaps: G1 (native session resume) was clean/real and kolu has since shipped it (juspay/kolu#1495), while G2 (resize arbitration) remains open. This removes the present-tense 'uncovered gap' framing that contradicted the shipped-G1 callout immediately below. Regenerated docs/atlas/dist/herdr-vs-kolu.html via just atlas::build (idempotent; rendered HTML now shows 'Gaps herdr surfaced' and 'has since shipped it', no 'no current plan coverage').
- `F3` **fixed** — Agreed — the comment was made stale by the F1 fix. packages/integrations/anyagent/src/agent-cli.ts:140 (the AGENT_RESUME idPattern doc) still said 'An id that fails its gate falls back to the `last` (most-recent) marker', which described the pre-fix behavior. The code at :302-313 now splits: a same-agent ref with a bad id returns null (bare shell), and `last` is reached only for no ref or a different-agent ref. Rewrote the comment to state the gate is fail-closed (a same-agent ref with a failing id yields NO resume, never a downgrade to `last`) and that `last` is reached only when there is no ref or the ref names a different agent.

commit: `8748a22be414d8f1c06bb08ba10adf8efe33a6fe`

### Round 3

**codex** — approved: `true`

All findings from the review thread are resolved in the current tree. F1 remains fixed in code and tests, F2's Atlas source and generated HTML now distinguish the historical surfaced gap from the current shipped state, and F3's stale AGENT_RESUME comment now matches the fail-closed behavior.

Findings:
- `F1` · major · resolved — Malformed same-agent session refs previously silently fell back to cwd-most-recent resume. The current implementation returns `null` for same-agent refs whose id fails the per-agent shape gate, while preserving the most-recent fallback only for no ref or a different-agent ref. (packages/integrations/anyagent/src/agent-cli.ts:310)
- `F2` · minor · resolved — The Atlas note previously still framed G1 as a current uncovered gap despite marking native resume shipped. The section is now titled `Gaps herdr surfaced`, the intro explicitly says native resume has since shipped, and the generated HTML reflects the same wording. (docs/atlas/src/content/atlas/herdr-vs-kolu.mdx:114)
- `F3` · nit · resolved — The AGENT_RESUME policy comment previously documented the old fallback behavior for invalid ids. It now says the id gate is fail-closed for same-agent refs and that `last` is reached only for no ref or a different-agent ref. (packages/integrations/anyagent/src/agent-cli.ts:140)

_codex on the rebuttal:_ F1: Still resolved; the current behavior fails closed for malformed same-agent refs. F2: I accept the fix; the heading and intro no longer present G1 as a current uncovered gap, and the generated HTML is in sync. F3: I accept the fix; the policy comment now matches the code path.

srid:

The **x86_64-linux** lane ran on **`kolu-ci-1`** (`dev-x86-64-linux-04`) — commit `08360bcee`, **passed** ✓

- **Lane wall** (pipeline): **3m0s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 2m19s | ✓ |
| `ci::home-manager` | 1m26s | ✓ |
| `ci::unit` | 49s | ✓ |
| `ci::smoke` | 46s | ✓ |
| `ci::nix` | 46s | ✓ |
| `_ci-setup` | 31s | ✓ |
| `ci::surface-example-build` | 30s | ✓ |
| `ci::surface-app-example-build` | 30s | ✓ |
| `ci::render` | 30s | ✓ |
| `ci::fmt` | 29s | ✓ |
| `ci::biome` | 29s | ✓ |
| `ci::install` | 9s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/08360bc/timings.jsonl`; pool state is a live `flock` probe.</sub>

PR #1500, opened 2026-06-21
Add a Markdown Notes tab per terminal (renames intent → notes)

srid:
Every terminal now has a **Markdown notes scratchpad in a right-panel Notes
tab** — Edit/Preview sub-tabs, an emoji quick-row, and debounced autosave —
replacing the intent-editor modal. The first line still supplants the branch
name in the title-bar annotation slot (and the dock / switcher / sub-panel
tabs), so the "what is this terminal for" label behaviour is unchanged; what's
new is a dedicated editing home, a body-gated title-bar icon, and a last look
before the notes are deleted with the terminal.

The field is renamed `intent` → `notes` across common/server/client, with a
one-time state migration so existing annotations come back as notes rather than
being silently stripped by Zod's unknown-key drop on first launch.

### The three new surfaces

- **Notes tab** — a new `RightPanelTabKind` `"notes"` (compiler-enforced 3-place
  union change) hosting `NotesEditor`: a textarea + Edit/Preview toggle + emoji
  quick-row + debounced autosave. The editor holds a local draft and reconciles
  from the server echo _only when not editing and no save is in flight_ — a live
  `value={meta.notes}` would clobber keystrokes mid-typing, and a stale echo
  racing the blur-flush would revert the last edits. Pending edits flush on blur
  and on terminal-switch.
- **Title-bar note icon** — body-gated (lines 2+), so it complements the line-1
  chip rather than duplicating it; click → the Notes tab.
- **Close-dialog reminder** — closing deletes the notes, so the dialog shows
  them one last time, read-only, off the frozen `target.meta.notes` snapshot
  (never a live read — the dialog is an imperative confirmation).

### Rename + migration

`setIntent` → `setNotes`, `setTerminalIntent` → `setTerminalNotes`, the
`intent/` client dir → `notes/`, all `Intent*` symbols → `Notes*`, and every
`meta.intent` read site. The persisted key is renamed in the `1.29.0` migration
via a shared `migrateIntentToNotes` backfill — composed into
`backfillSavedTerminal` so the client's session-import hatch renames the same
key from one source of truth. _A leaf, not an electricity_: the editor reuses
`@kolu/solid-markdown` + `@kolu/surface`, plugging in like the register's other
mutations.

> Existing users' notes survive the rename transparently — the migration is
> idempotent, optional-safe, and pinned by a pre-rename fixture test (the
> "notes vanish on first launch" bite from the plan).

Plan of record: [Terminal Notes](https://kolu.dev/atlas/terminal-notes.html)
Atlas note. e2e rewritten for the tab + autosave (`terminal-notes.feature`).

_Generated by [`/be`](https://github.com/srid/agency) on opencode (model `glm-latest`)._


srid:
## Review gauntlet

Ran a pragmatic gauntlet (opencode adaptation of `/be-review`): lowy + hickey
structural lenses in parallel, a correctness fact-check, and a codex review.
Findings adjudicated and fixes applied in `7f5…` (counter + onCleanup).

### lens-debate (lowy ⇄ hickey)
- **lowy:** *no actions.* Every boundary is correctly placed — `NotesEditor` is
  a leaf (the hard volatilities it could own — markdown→safe-HTML, live-state +
  persistence — are already behind `@kolu/solid-markdown` / `@kolu/surface`,
  which it consumes). The migration lives beside the schema it restores (one
  source, two callers: the server ladder + the client session-import hatch).
  `showNotes`/right-panel tab wiring is the right seam, symmetric with
  inspector/code; tab-selection and visibility correctly split.
- **hickey:** *one nit* — `pending`/`timer`/`editing` was an implicit 3-variable
  state machine (correct, but high reasoning load). Folded into the fix below.
  Confirmed `annotationLine` is correctly **kept** (it names the output
  concept — the annotation slot's rendered line — not the field); the rename is
  complete and well-scoped; no debounce-utility duplication.

### codex review
- **[P2] serialize autosaves / stale write wins** — **declined (moot).** The
  transport is a single ordered `websocketLink` (`/rpc/ws`), so `setNotes` RPCs
  arrive and process in send order — the final value is always the latest edit.
  Reordering (the concern) requires separate HTTP requests, which isn't the
  transport here.
- **[P3] `flushNow` leaves `dirty` set → `pending` sticks on "Saving…"** —
  **fixed.** `flushNow` now resets `dirty` (the flush consumes the unsaved
  keystrokes); otherwise the reconcile effect would ignore every later echo.

### fact-check
- **blocker — stale-echo clobber under overlapping RPCs** — **fixed.** The
  single `pending` boolean flipped false on the *first* RPC's `.finally` while a
  later one was still in flight, opening the reconcile gate to the earlier RPC's
  stale echo (data loss on type→pause→type→blur). Replaced with
  `dirty || inflight > 0` (a count), so the gate opens only when truly no save
  is pending.
- **nit — debounce timer not cleared on unmount** — **fixed** (`onCleanup`).
- **nit — `CloseConfirmTarget.meta` is a live proxy, not the frozen snapshot the
  docstring claims** — **declined (pre-existing).** Affects `git`/`pr` too; low
  impact (nothing changes `meta.notes` during the brief close dialog); fixing it
  for all fields is out of scope for this PR. The notes reminder joins the
  existing pattern.

Net: one real data-loss race + one stuck-"Saving…" bug, both fixed; two
structural nits declined with rationale.

srid:
## Evidence

A Markdown notes scratchpad per terminal, in the new right-panel Notes tab.

### Notes tab — Edit mode (autosaved)

The Notes tab with a multi-line markdown note typed. The emoji quick-row,
Edit/Preview sub-tabs, Copy/Clear actions, and the "Saved" autosave indicator
are all visible. The note's first line (`## TODO`) has already supplanted the
branch name in the title-bar annotation slot (visible on the dock chip as
`## TODO`), and the body-gated note icon has appeared in the title bar (the
"Open notes" button beside the theme pill) — both confirming the debounced
autosave round-tripped through `setNotes` → server → echo.



### Notes tab — Preview mode

The same note rendered as Markdown — the `## TODO` heading, the task-list
checkboxes (`fix auth bug`, `ask Sri re: deploy`), and the inline-code
`auth.ts` — all sanitized + rendered through `@kolu/solid-markdown`'s `compact`
variant, the same renderer the dock and switcher cards use.



### What the clip proves

- **The Notes tab opens and the editor focuses** — via the annotation slot, the
  command palette's "Edit notes", or the title-bar note icon.
- **Debounced autosave works** — the annotation slot updates with line 1 and the
  note icon appears (gated on a non-empty body) without a Save button.
- **Edit ⇄ Preview** — the textarea source and the rendered markdown swap in place.
- **The rename is invisible to the user** — everything that was "intent" (the
  annotation slot, the chip, the dock rail glyph) now reads from `meta.notes`.

### CI

All 14 linux-lane checks are green on the final HEAD (`d60b828`), including the
full e2e suite (474 scenarios / 4219 steps passed — the rewritten
`terminal-notes.feature` scenarios all green). The macOS (aarch64-darwin) lane
errored at `_ci-setup` — the `rasam` build host was unreachable (ssh connection
timed out); per the user, macOS CI can be ignored for now.

srid:
## Lens debate — lowy ⇄ hickey

Structural review of the diff through two lenses, run independently then
cross-examined. Both lenses agreed: no blockers. One nit (hickey) folded into
the fix commit.

### lowy (volatility-based decomposition) — *no actions*

Every boundary is correctly placed:

- **`NotesEditor` is correctly a leaf, not an `@kolu/*` package.** The hard
  volatilities a notes editor *could* own — markdown → safe-HTML
  (`@kolu/solid-markdown`), live-state + persistence (`@kolu/surface`) — are
  already behind their canonical seams, which the editor consumes via
  `props.notes()` (read) and `client.terminal.setNotes` (write). The
  reconcile/autosave logic that *remains* in the editor is UI-local draft
  coordination — a bounded algorithm for one editing surface, not a hidden
  transport/reconnect/persistence concern. The Layer-0 prompt ("is the *pattern*
  across N call sites the artifact?") does not fire: N=1 for this
  autosave-reconcile shape today; extracting now would be speculative
  (speculative volatility is not volatility).
- **No duplicated volatility encapsulation.** `NotesMarkdown.tsx` is
  consumer-side policy pinning, not a parallel renderer. `notes/text.ts` is the
  single projection source — all ~8 render sites + `chipInitials.ts` import
  `annotationLine`/`notesBodyMarkdown`/`notesLeadGlyph`/`firstGrapheme` from
  it. No second hand-rolled projection.
- **No symmetry violation.** The `openInCodeTab`+`pendingOpen` asymmetry vs the
  inline `openNotes` reflects a genuine difference in content-routing volatility
  (Code routes "which file:line"; Notes doesn't route content). The
  Edit/Preview-not-persisted vs `codeMode`-persisted asymmetry reflects a
  genuine difference in sub-state kind (render mode vs data source). Both
  justified by the domain, not missing seams.
- **Migration is in the right boundary.** `migrateIntentToNotes` in
  `kolu-common/surface.ts`, composed into `backfillSavedTerminal`, wired by the
  1.29.0 ladder step — exactly the established `backfillRemoteUrl` /
  `backfillLocation` / `backfillTerminalState` pattern. A parallel rename in
  `sessionTransfer.ts` would be duplicated volatility; the plan's "one source
  of truth" choice is correct.
- **`showNotes`/right-panel tab wiring is the right seam.** Tab-selection
  (`showNotes`) and visibility (`reveal`) are correctly separate methods,
  matching the `useRightPanel` header's "three storage layers / three
  volatilities" decomposition. Per-terminal `activeTab` persists and
  round-trips through `setRightPanel` symmetrically with inspector/code.

### hickey (Simple Made Easy) — *one nit, fixed*

- **No fragmentation / no concept multiplication / no structural pattern
  match except one.** `NotesEditor` is a leaf consumer of
  `@kolu/solid-markdown` + `@kolu/surface`; `NotesMarkdownBlock`/
  `NotesMarkdownInline` are thin variant-pinning wrappers over the package's
  `Markdown` — no second renderer; `NotesPanel` is a 31-line shell; no new
  `@kolu/*` package. The migration is composed into `backfillSavedTerminal`
  correctly — one source of truth, no parallel rename path on the import side.
  `annotationLine(notes, fallback)` correctly models one concept (the
  annotation slot's rendered line) as one function — not fragmented. Renaming
  it to `notesLine` would be **wrong** — the slot isn't always notes (it falls
  back to the branch name); it names the *output*, not the field. Keep.
- **One nit — the save lifecycle is an implicit 3-variable state machine.**
  `pending`/`timer`/`editing` encode a state machine whose reachable states are:
  `idle` / `editing-dirty` / `scheduled` / `in-flight`. The joint correctness
  rests on an invariant that is *enforced but not named*: `scheduleSave` sets
  `pending=true` in the same step that arms `timer`, so `timer !== undefined ⟹
  pending === true`. The reasoning load to convince oneself the terminal-switch
  is safe is high (discover that invariant + notice `flushNow`'s `timer` guard +
  know `on` untracks its callback). **Code is correct today** — safety is
  structurally enforced, not conventional — so this is a legibility nit, not a
  defect. Folded into the fix commit (`d60b828`): the counter-based `pending`
  makes the "how many saves in flight" question structural rather than emergent,
  and `onCleanup` makes the unmount contract explicit. A full `SaveState` DU
  refactor (hickey's suggestion) was weighed and declined — done correctly
  (handling overlapping in-flight RPCs) it's more complex than the counter and
  risks a refactor bug for no correctness gain.

**Consensus:** no blockers; both lenses agreed the boundary verdicts hold. The
one nit became the fix commit's rationale.

srid:
## Codex review

Ran `codex exec review --base origin/master` against the diff. Two findings
filed; adjudicated below.

### [P2] Serialize notes autosaves so stale writes cannot win — `NotesEditor.tsx:89-93`

> When a user pauses long enough to send one autosave, then keeps editing while
> that RPC is still in flight, `flush` sends another independent `setNotes`
> request with no ordering guard. If the earlier request reaches the server
> after the later one (or a clear races with an older save), the stale value
> can overwrite the user's latest notes. Chain saves per terminal or otherwise
> make the server/client discard older revisions so the last edit always wins.

**Declined (moot).** The transport is a single ordered `websocketLink`
(`/rpc/ws`, `packages/client/src/wire.ts:68`) — there is one WebSocket, and
oRPC processes messages over it in send order. RPCs cannot reorder: the server
applies `setNotes` calls in the sequence the client sent them, so the final
value on disk is always the latest edit. Reordering (the concern) requires
separate HTTP requests racing through independent connections, which isn't the
transport here. No serialisation / revision-discard machinery is needed.

### [P3] Reset dirty when forcing an immediate notes save — `NotesEditor.tsx:117-121`

> When blur or terminal-switch flushes a pending debounce through `flushNow`,
> the timer is cleared but `dirty` stays true. After the RPC finishes,
> `pending()` remains true forever, so the UI can stay on "Saving…" and the
> reconcile effect continues to ignore server echoes until a later debounced
> save happens to clear `dirty`.

**Fixed (`d60b828`).** `flushNow` now resets `dirty` — the flush consumes the
unsaved keystrokes, so the "debounce-armed" flag must clear alongside the
timer. Without it `pending` (hence "Saving…") sticks forever and the reconcile
effect ignores every later echo. This was a real bug in the first iteration of
the counter-based `pending` fix (the counter tracked RPCs correctly but
`flushNow`'s path still left `dirty` set).

**Net:** one real bug fixed (P3); one concern declined with rationale (P2 —
ordered transport moots it).

PR #1501, opened 2026-06-21
feat(terminal): per-terminal notes — Notes tab, autosave, intent→notes rename

srid:
## What

Promotes the per-terminal **`intent`** annotation to first-class **notes** with a dedicated right-panel **Notes** tab — and renames the field `intent` → `notes` end to end. Plan of record: [`terminal-notes`](https://kolu.dev/atlas/terminal-notes.html).

The field already shipped (persist · autosave · sleep/wake · restore · markdown render), so this is a **rename + three client surfaces**, not a feature from scratch.

### User-facing
- **Notes tab** (beside Code · Inspector) with **Edit / Preview** sub-views and **autosave as you type** — no Save button, no modal. Edit is a `<textarea>` + the curated emoji quick-row; Preview renders the markdown via `@kolu/solid-markdown`.
- **Title-bar note icon** lights up once notes carry a body past line 1 (line 1 already rides the chip) → reveals the tab.
- **Close-dialog reminder**: a read-only last look at the notes before they're discarded with the terminal.
- The `Edit intent` command is now **`Edit notes`**; the title-bar chip + command both repoint to the tab. The old modal is gone.

### Architecture
- **Rename, not add**: `intent` → `notes` in `surface.ts` cascades through schema, contract (`setIntent` → `setNotes`), server handler/spawn-seed, and every client render site (annotation chip, dock chip glyph, dock/switcher cards). Object-spread sleep/wake/restore paths carry it for free; only the explicit-name copies were touched.
- **Editor is a leaf**, not a new `@kolu/*` package: `NotesTab` composes the existing `@kolu/solid-markdown` renderer + `@kolu/surface` persistence. The Edit/Preview switcher mirrors the Code tab's mode pattern (`NotesTabViewSchema`, `notesMode` persisted per terminal, the compiler-enforced 3-place tab-union change).
- **Presenter purity preserved**: `CloseConfirm` reads the frozen `target.meta.notes` snapshot; orchestration (activate → `showNotes` → reveal) stays at the App root.

### Migration
`state.ts` 1.28.0 → **1.29.0**: `renameIntentToNotes` moves each saved terminal's `intent` value onto `notes` **before** the tightened schema strips the now-unknown key (Zod drops unknown keys, so a rename is *not* a no-op). Idempotent, presence-keyed, non-clobbering; also wired into the client import-hatch composition.

## Design philosophy
- **Fail-fast / no fallbacks** — the migration is a hard rename with no dual-read of the old key; absent value reads as "no notes", never a silent default.
- **Volatility boundary** — the editor stays a leaf; the two axes it could own (markdown→safe-HTML, live-state+persistence) are already packages. No new receptacle.
- **Reuse the source of truth** — reuses `intent`'s whole persistence/render stack and the Code tab's mode pattern + `SegmentedControl`, rather than a parallel path.

## Test plan
- **Unit** — `renameIntentToNotes`: rename, idempotency, no-clobber, untouched-when-absent (`state.test.ts`). Full suites green: common 17 · client 448 · server 132.
- **e2e** — `terminal-notes.feature` rewritten for the Notes-tab flow (reveal · autosave · Preview · note icon · clear · links-off slot · palette).
- `just check` (typecheck + biome) and `just fmt` clean; `ci::atlas-sync` passes.


srid:

✅ **Consensus** after 2 round(s) · lowy + hickey · base `7c0e22166205`

Independent findings: lowy=6, hickey=4

### Applied (6)
- `lowy-3` Stale `intent` test-id leaks the dead name through a stable interface — commit `84c5ff143`
- `lowy-4` Comment drift: stale `intent` wording in renamed/touched modules — commit `4ef2b806f`
- `lowy-6` `notes` JSDoc and `NotesBody` comment claim the dock-awaiting card renders the full body, but only the switcher card does — commit `8b9b39653`
- `hickey-1` The "absent notesMode = edit" default is fragmented across four sites — commit `fc2079639`
- `hickey-2` Renamed concept still named "intent" at surviving identifiers and comments — commit `61676f072`
- `hickey-3` useRightPanel module-doc enumerates per-terminal fields but omits the newly-added notesMode — commit `368d8c12b`

### Agreed — no change (4)
- `lowy-1` Intent→Notes consolidates onto the existing tab receptacle (no parallel receptacle) — clean (packages/common/src/surface.ts:636 / packages/client/src/right-panel/RightPanel.tsx:151)
- `lowy-2` `intent` field rename migration sits correctly inside the existing backfill ladder (packages/common/src/surface.ts:1068 (renameIntentToNotes) + packages/server/src/state.ts:508)
- `lowy-5` `notesBodyMarkdown` invoked inline in TerminalMeta instead of through the `NotesBody` receptacle — minor (packages/client/src/terminal/TerminalMeta.tsx:156 vs packages/client/src/notes/NotesBody.tsx:21)
- `hickey-4` handleEditActiveNotes re-resolves the active id that openNotesFor immediately re-sets (packages/client/src/App.tsx:204-207,149-154)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 round(s) · codex reviewed at `xhigh` reasoning effort · base `7c0e22166205`

### Round 1

**codex** — approved: `false`

The change is directionally coherent, but I found several real issues: one existing Cucumber scenario still points at the deleted intent editor, the new notes-mode persistence path drops mode-only changes server-side, and the autosave state can lie after failed or racing saves. There are also stale user-facing/internal strings from the intent rename and an untracked review scratch file in the worktree.

Findings:
- `F1` · blocking · open — This existing scenario still expects the deleted intent editor (`Then the intent editor should be visible`). The old terminal_intent step definitions were renamed/deleted, so the Cucumber suite will hit an undefined step or assert UI that no longer exists. (packages/tests/features/git-context.feature:78)
- `F2` · major · open — `rightPanelStateEqual` ignores the new `notesMode` field. A pure Edit/Preview toggle sends `setRightPanel`, but if `activeTab`, `codeMode`, and `selectedFileByMode` are unchanged the server treats the state as equal and skips `m.rightPanel = state`, so the user's Notes sub-view is not persisted or published. (packages/server/src/terminals.ts:209)
- `F3` · major · open — Autosave marks a draft as saved before the RPC succeeds, and saves are fire-and-forget. If `setNotes` fails, `dirty()` is already false, the footer says autosaved, blur/unmount will not retry, and the user's text can be lost. Parallel in-flight saves can also arrive out of order and let an older draft overwrite a newer one. (packages/client/src/notes/NotesTab.tsx:82)
- `F4` · minor · open — The discoverability tip still tells users to use `Edit intent`, but the command was renamed to `Edit notes`. That makes the in-app guidance point at a command that no longer exists. (packages/client/src/settings/tips.ts:117)
- `F5` · nit · open — Several touched comments still use the old terminal-annotation wording (`intent`) for the renamed notes concept, for example this dock header and the branch/intent comments in `TerminalMeta.tsx` and `commands.tsx`. That leaves misleading maintenance breadcrumbs after the rename. (packages/client/src/canvas/dock/Dock.tsx:8)
- `F6` · nit · open — The worktree contains an untracked generated review scratch file. It is not product code or documentation and could be accidentally added later. (.be-review/lens-comment.md:1)

**claude** — Addressed all 6 of codex's findings on the intent→notes rename; agreed with and fixed every one. F1: repointed git-context.feature's title-click scenario from the deleted "intent editor" to "the notes editor should be visible". F2: rightPanelStateEqual now compares notesMode via notesModeOf, exported it, and added terminals.test.ts as a regression test. F3: NotesTab autosave now advances savedValue only on RPC success (failed saves stay dirty/retry), with a generation counter to defeat out-of-order saves and a lastSent ref to avoid redundant flush re-fires. F4: tip text fixed to "Edit notes" (kept the tip id stable). F5: renamed stale "intent" comments in touched files. F6: removed the .be-review scratch file and added .be-review/ to .gitignore. Server+client typecheck clean, new and existing server tests pass (19), biome reports nothing to reformat. Committed as beeb96355.

- `F1` **fixed** — Agreed — the scenario at packages/tests/features/git-context.feature:78 still asserted 'Then the intent editor should be visible', a step deleted in the rename, so Cucumber would hit an undefined step. Renamed the scenario to 'Clicking the title annotation slot reveals the Notes tab editor' and changed line 82 to 'Then the notes editor should be visible' (defined in terminal_notes_steps.ts). The existing 'I click the terminal title branch' step targets ACTIVE_TITLE_BRANCH_SELECTOR, which is byte-identical to the notes SLOT selector, so the click lands on the same element and reveals the Notes tab.
- `F2` **fixed** — Agreed — rightPanelStateEqual (packages/server/src/terminals.ts:209) compared activeTab/codeMode/selectedFileByMode but not notesMode, so a pure Edit↔Preview toggle on the Notes tab was treated as equal and setRightPanelState skipped the publish, never persisting the sub-view. Added `if (notesModeOf(a) !== notesModeOf(b)) return false;` (imported notesModeOf so an absent field and explicit 'edit' compare equal via the single default site), exported the guard, and wrote packages/server/src/terminals.test.ts with three cases: a notesMode-only toggle is detected (false), absent vs explicit 'edit' are equal (true), identical states equal (true). All pass.
- `F3` **fixed** — Agreed on both sub-issues. (1) persist() called setSavedValue(text) synchronously before the RPC, so on failure dirty() was already false, the footer lied 'Autosaved', and no retry happened. Now savedValue advances ONLY inside the .then() success branch; failures leave the draft dirty (footer stays 'Saving…') so blur/unmount/next-keystroke retries. (2) Fire-and-forget saves could resolve out of order. Added a monotonic saveGen counter — only the newest in-flight save (gen === saveGen) may advance savedValue, so an older draft can't overwrite a newer one. Added a lastSent ref (last text handed to an RPC) to gate flush so an in-flight write isn't redundantly re-fired, kept decoupled from the confirmed savedValue; seed effect and the failure branch keep lastSent consistent.
- `F4` **fixed** — Agreed — the tip at packages/client/src/settings/tips.ts:118 pointed at the renamed-away 'Edit intent' command (now 'Edit notes' per commands.tsx:364). Updated the text to 'Edit notes' and 'attach a note'→'attach notes'. Per codex's own suggestion, kept the tip id 'amb-terminal-intent' stable (seenTips preferences key on it) so users who already dismissed it don't see it resurface, and added a comment explaining why the id intentionally keeps the old name.
- `F5` **fixed** — Agreed — renamed stale terminal-annotation 'intent' wording in touched comments only: Dock.tsx:8 ('the intent's lead grapheme'→'the notes' lead grapheme') and Dock.tsx:577 ('intent lead grapheme'→'notes lead grapheme'); TerminalMeta.tsx:111 and :116 ('branch/intent'→'branch/notes'); commands.tsx:268 ('theme/intent'→'theme/notes'); terminals.ts header ('setTerminalIntent'→'setTerminalNotes'); and useTerminalCrud.ts:26 which cross-referenced the now-deleted useIntentEditor — generalized it to describe the de-deps pattern without the dead name. Left genuine 'intentionally'/'user intent'/scroll-intent language and the renameIntentToNotes migration helper's legacy-field docs (which deliberately describe the old field) untouched, as codex scoped.
- `F6` **fixed** — Agreed — .be-review/lens-comment.md was an untracked generated lens-debate scratch artifact that could be committed by accident. The .gitignore already had .be-review-comments/ but not the actual .be-review/ path. Added .be-review/ beside the sibling skill-scratch entries (reusing the canonical gitignore mechanism rather than a one-off rm, so the next round's artifact is also ignored) and removed the stray directory from the worktree.

commit: `beeb9635585df9b4cc13880dd4bc0c230d5c7d17`

### Round 2

**codex** — approved: `false`

Five prior findings are fixed cleanly. The original autosave failure/out-of-order problems were addressed, but the new save-generation implementation introduced a separate regression around terminal switches, so the review cannot approve yet.

Findings:
- `F1` · blocking · resolved — The scenario still referenced the deleted intent editor step, which would break Cucumber after the rename. (packages/tests/features/git-context.feature:78)
- `F2` · major · resolved — The right-panel equality guard ignored `notesMode`, so Notes Edit/Preview-only changes could be swallowed and not persisted. (packages/server/src/terminals.ts:212)
- `F3` · major · resolved — Autosave previously marked drafts saved before the RPC succeeded and allowed older saves to win over newer saves. (packages/client/src/notes/NotesTab.tsx:82)
- `F4` · minor · resolved — The feature tip still referenced the renamed-away `Edit intent` command. (packages/client/src/settings/tips.ts:117)
- `F5` · nit · resolved — Touched comments still used stale terminal-annotation `intent` wording after the rename. (packages/client/src/canvas/dock/Dock.tsx:8)
- `F6` · nit · resolved — An untracked generated review scratch file was present and could be accidentally committed. (.be-review/lens-comment.md:1)
- `F7` · major · open — The new async save acknowledgement is not tied to the terminal it was sent for. On terminal switch, `flush(prevId)` can start or leave in flight a save for terminal A, then lines 142-145 reseed `draft`/`savedValue` for terminal B. If A's save resolves afterward and no newer save has bumped `saveGen`, line 109 calls `setSavedValue(text)` against B's editor state. That makes B compare its draft against A's saved text, leaving the save indicator and retry logic wrong for the active terminal. (packages/client/src/notes/NotesTab.tsx:101)

_codex on the rebuttal:_ CLAUDE agreed with F1-F6 and the targeted fixes are present: the git-context scenario now asserts the notes editor, `rightPanelStateEqual` compares `notesModeOf` with regression tests, the tip text and stale comments were updated, and `.be-review/` is ignored with no untracked scratch left. I accept those. For F3, the original failure and out-of-order concerns are fixed, but the implementation introduced F7: save completions can now mutate the editor state after the active terminal changes, because the generation counter is global to the component and is not invalidated or keyed by terminal id.

**claude** — Round 2 of the codex debate. Findings F1–F6 were already resolved in round 1 (codex confirmed) — nothing to re-litigate. The single open finding this round was F7, a real terminal-switch regression introduced by round 1's save-generation fix. I agreed and fixed it. The save-completion handler in NotesTab.tsx now gates every ack on BOTH the newest-generation check (gen === saveGen) AND a terminal-id check (id === props.terminalId), via a single isCurrent() helper shared by the success and failure branches, so a slow save fired for terminal A can no longer write A's notes into terminal B's savedValue/lastSent after a switch. The terminal-switch effect also bumps saveGen so an in-flight save for the outgoing terminal is invalidated even when the incoming terminal never fires its own save (belt-and-suspenders with the id check). Traced the exact F7 scenario to confirm: A's late ack now fails both guards and leaves B's editor state untouched. Client typecheck is clean and biome reports nothing to reformat. Only packages/client/src/notes/NotesTab.tsx changed.

- `F1` **fixed** — Resolved in round 1 (commit beeb96355); codex confirmed status=resolved. git-context.feature now asserts the notes editor. No further action needed this round.
- `F2` **fixed** — Resolved in round 1; codex confirmed. rightPanelStateEqual compares notesModeOf with regression tests in terminals.test.ts. No further action this round.
- `F3` **fixed** — Resolved in round 1; codex confirmed the original failure/out-of-order concerns are fixed. The implementation gap it left (terminal-switch) is tracked separately as F7 and fixed below.
- `F4` **fixed** — Resolved in round 1; codex confirmed. Tip text now reads 'Edit notes' (tip id kept stable). No further action this round.
- `F5` **fixed** — Resolved in round 1; codex confirmed. Stale terminal-annotation 'intent' comments renamed to 'notes'. No further action this round.
- `F6` **fixed** — Resolved in round 1; codex confirmed. .be-review/ ignored and the scratch file removed. No further action this round.
- `F7` **fixed** — Agreed — real regression from round 1's save-generation fix. saveGen was global to the component, so a setNotes save fired for terminal A that resolved after a switch to B passed the gen===saveGen check (when no newer save had bumped it) and called setSavedValue(A's text) against B's editor state — corrupting B's dirty indicator and flush gate. Fix in packages/client/src/notes/NotesTab.tsx: persist() now builds an isCurrent() = gen === saveGen && id === props.terminalId helper used by both the .then() (setSavedValue) and .catch() (lastSent re-arm) branches, so a save can only mutate editor state if it's the newest AND for the on-screen terminal. The terminal-switch effect (on(props.terminalId)) now bumps saveGen after flushing the outgoing terminal, invalidating any in-flight save for it even if the incoming terminal never saves — the id check and the saveGen bump are independent guards (defense in depth). Header doc and the saveGen/persist comments updated to describe the per-terminal invalidation. Traced the A→B switch scenario: A's late ack fails both guards, B's savedValue stays at its seed. Client tsc clean; biome reports no reformat.

commit: `3e2149b1a9b463c1e5e5acc03a3707d317b02959`

### Round 3

**codex** — approved: `true`

All findings from the review are resolved in the current tree. I re-checked the latest `NotesTab.tsx` change for F7: save completions are now gated by both generation and terminal id, and terminal switches invalidate prior in-flight saves before reseeding editor state, so a late save from terminal A can no longer mutate terminal B's saved marker or retry gate.

Findings:
- `F1` · blocking · resolved — The feature still referenced the deleted intent-editor step after the intent-to-notes rename, which would make the scenario fail with an undefined step. (packages/tests/features/git-context.feature:82)
- `F2` · major · resolved — `rightPanelStateEqual` ignored `notesMode`, so switching the Notes tab between edit and preview could be treated as no-op and fail to publish/persist the state change. (packages/server/src/terminals.ts:209)
- `F3` · major · resolved — Autosave marked drafts saved before RPC success and allowed out-of-order completions to overwrite the confirmed saved marker. (packages/client/src/notes/NotesTab.tsx:105)
- `F4` · minor · resolved — The tip text still referenced the old `Edit intent` command name after the UI rename. (packages/client/src/settings/tips.ts:118)
- `F5` · nit · resolved — Several touched comments still used stale terminal-annotation `intent` terminology, making the renamed notes concept harder to follow. (packages/client/src/canvas/dock/Dock.tsx:8)
- `F6` · nit · resolved — A generated review scratch file was untracked and could be accidentally committed. (.be-review/lens-comment.md)
- `F7` · major · resolved — The round-1 save-generation fix was not tied to terminal identity, so a late save ack from terminal A after switching to terminal B could mutate B's `savedValue`/flush state. (packages/client/src/notes/NotesTab.tsx:109)

_codex on the rebuttal:_ No disputes remain. I accept the F7 fix: `persist()` now uses a shared `isCurrent()` guard requiring both `gen === saveGen` and `id === props.terminalId`, and the terminal-switch effect bumps `saveGen` after flushing the outgoing terminal before reseeding state for the incoming terminal. That closes the A-to-B late-ack corruption path I raised.

srid:

**All clear** — ran with `--no-elegance` (the `/simplify` track already swept this tree).

| Pass | Issues | Outcome |
|------|--------|---------|
| Rules | 0 | Clean — all 32 rules checked (DRY, invalid-states-unrepresentable, no-silent-error-swallowing, toast-must-include-error-message, migration-shape-guard, persisted-schema-stays-tolerant, icons-in-registry, app-shell-stays-thin, …) |
| Fact-check | 0 | Clean — two candidates raised and **refuted** |
| Elegance | — | Skipped (`--no-elegance`; simplify ran) |

**Refuted candidates**
- *"A failed save for terminal A could corrupt B's flush gate after a switch"* — not a bug: the `on(() => props.terminalId)` effect bumps `saveGen` and reseeds `lastSent`/`savedValue` to the incoming terminal **before** A's `.catch()` can fire, so the `isCurrent()` guard (`gen === saveGen && id === props.terminalId`) is already false and the stale ack touches nothing.
- *"Empty-string `setNotes` contract isn't explicit in the schema"* — deliberate: `setNotes` accepts any string (empty clears, coerced to `undefined` server-side), matching the prior `setIntent` contract.

Notable confirmations: `caught-error-must-not-collapse-to-empty` holds (a failed save leaves the draft **dirty**, footer stays "Saving…", retry allowed — no silent fallback to empty); `migration-shape-guard` holds (`renameIntentToNotes` is presence-keyed on `"intent" in t`); `persisted-schema-stays-tolerant` holds (`notesMode` optional, resolved through the single `notesModeOf` site).

srid:
## Evidence

### Terminal Notes — autosave round-trip

The clip shows: opening the Notes tab editor, typing markdown notes, and the title-bar annotation chip updating to the notes' first line via autosave (no Save button needed).



▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=terminal-notes-autosave.mp4

**What the clip proves:**
- The Notes tab is accessible from the right panel
- Typing markdown in the Edit sub-view autosaves without a Save button
- The title-bar annotation chip updates to the notes' first line immediately after autosave
- The Edit/Preview sub-views are navigable

**Touch target note:** The Notes tab toolbar uses the same SegmentedControl component as the Code-tab scope switcher, which has touch targets ≥24px on coarse pointer — no new touch regression introduced.

srid:

The **x86_64-linux** lane ran on **`kolu-ci-6`** (`dev-x86-64-linux-03`) — commit `7388ce95e`, **failed** ✗

- **Lane wall** (pipeline): **3m33s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 2m50s | ✗ (exit 1) |
| `ci::nix` | 1m51s | ✓ |
| `ci::home-manager` | 1m30s | ✓ |
| `ci::flake-check` | 1m11s | ✓ |
| `ci::smoke` | 51s | ✓ |
| `ci::atlas-sync` | 48s | ✓ |
| `_ci-setup` | 35s | ✓ |
| `ci::unit` | 34s | ✓ |
| `ci::render` | 16s | ✓ |
| `ci::biome` | 15s | ✓ |
| `ci::fmt` | 14s | ✓ |
| `ci::install` | 8s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/7388ce9/timings.jsonl`; pool state is a live `flock` probe.</sub>

srid:
## CI status

**`x86_64-linux`: fully green** — all 15 lanes pass on `45835e5e` (`ci::nix`, `ci::e2e`, `ci::unit`, `ci::biome`, `ci::fmt`, `ci::atlas-sync`, `ci::flake-check`, `ci::home-manager`, `ci::render`, `ci::smoke`, …), plus both SentinelOne security checks.

**`aarch64-darwin`: blocked on infra, not code** — `_ci-setup@aarch64-darwin` errors at host setup because the darwin build host **`rasam` is offline** (tailscale: *"offline, last seen 20h ago"*; `rasam:22` unreachable). No darwin recipe ran, so those required contexts are unposted. This is an environmental outage — the macOS lane needs `rasam` back online (or a replacement darwin host), then a re-run will post the darwin checks. Nothing in this diff is darwin-specific (pure rename + SolidJS UI + a conf migration).

One e2e fix landed during the run: `fix(notes): focus the editor past the palette's terminal-refocus race` (`45835e5e`) — the "Edit notes via the command palette" scenario needed a double-rAF so the editor focus wins against `refocusIfNoDialogOpen()`. e2e is green after it (475→476 scenarios pass).

PR #1503, opened 2026-06-21
feat(chrome-bar): live server, client & kaval memory in the identity rail

srid:
## What

The chrome bar's identity rail already named three processes — **srv** (the kolu-server), **client** (this browser's bundle), and **kaval** (the pty-host daemon) — but showed no memory for any of them. This adds a live, whole-MB memory readout beside each, aligned under its own column:

```
[ ◉ srv v1.2 a1b3 142 MB │ ≡ client 88 MB │ ◉ kaval 12m 31 MB ]
```

## How

Three figures, two sources:

- **Server + kaval RSS** ride a new `processMemory` surface cell. A server-side sampler (`packages/server/src/memorySampler.ts`) reads its own `process.memoryUsage().rss` and polls the kaval daemon's RSS over a **dedicated `system.processMemory` pty-host verb** on a 5 s cadence, publishing both. The kaval reading is an honest three-way `ok | absent | error` union: `absent` when no daemon is connected, `error` (rendered as a distinct `mem ?` chip) when a believed-connected daemon's poll fails — so a failed poll never reads as "no daemon".
- **Client JS heap** is read locally off `performance.memory` (Chromium-only), refreshed off the shared 1 s app clock — **no new timer**.

The pty-host contract bumps **3.1 → 3.2** (additive: a new procedure). A pre-3.2 survivor is recycled on adoption by the existing version gate (session preserved).

## Design philosophy

- **Fail fast, no fallbacks** — kaval/client memory is `absent`/hidden when genuinely unmeasurable (no daemon; non-Chromium browser), and a poll failure on a believed-connected daemon surfaces as a distinct `error` state (logged at `error` level), never silently collapsed to "no daemon".
- **Volatility boundaries** — the fast-moving metric rides its **own** `processMemory` cell, deliberately **not** folded onto the lifecycle-transition `daemonStatus` collection: two change rates, two channels. Daemon RSS rides its **own** atomic `system.processMemory` verb, kept out of `system.heartbeat`'s pure liveness round-trip. No new package — a leaf concern over the existing surface transport.
- **Reuse the source of truth** — byte-format helpers + the JS-heap probe live once in `ui/memory.ts` (shared by rail + Diagnostic dialog); the whole-MB projection + `BYTES_PER_MB` live once in `kolu-common`; the "is-connected gate + three-way unwrap" lives once in `kavalMemoryDisplay()`, shared by the rail and the dialog.

## Tests

- Unit: `memorySampler.test.ts` (server/kaval/down/poll-error branches, overlap guard), `surface.test.ts` (MB-dedup `equals`), `memory.test.ts` (format helpers); the kaval contract corpus asserts `system.processMemory` returns a real RSS.
- e2e: `chrome-memory.feature` drives output growth and asserts all three readouts render (and visibly change).

## Review

Ran the full `/be` gauntlet: **lens-debate** (lowy ⇄ hickey, consensus — 4 fixes incl. the dedicated verb), **codex-debate** (consensus, 4 rounds — the three-way union, overlap guard, daemon-state gating), **simplify** (shared `kavalMemoryDisplay` derivation), **code-police** (all clear). See the PR comments.

## Docs

- Changelog `Added` entry.
- `pty-daemon-chrome-bar` Atlas note records the memory readouts + the 3.2 contract bump (`dist/` rebuilt).


srid:

✅ **Consensus** after 2 round(s) · lowy + hickey · base `70511f199410`

### Applied (4)
- `lowy-1` AgentSessionRef contract defined twice across a layer that already imports the lower one (duplicated receptacle) — commit `622495fbd`
- `hickey-1` AgentSessionRef declared twice — one domain concept, two separately-maintained types — commit `622495fbd`
- `hickey-2` Per-agent resume policy fragmented across two parallel Records (AGENT_RESUME + SESSION_ID_PATTERN) — commit `f385db307`
- `hickey-3` Resume-by-id behavior reconstructed at the call site inside sleepWake.test.ts rather than exercised through wake() — commit `9dc76ff2e`

### Agreed — no change (2)
- `lowy-2` Resume-mode selection (last vs byId) lives inside resumeAgentCommand rather than reading as a pure id→marker choice — minor, acceptably encapsulated (packages/integrations/anyagent/src/agent-cli.ts:305-315)
- `lowy-3` agentSessionToPersist mirrors shouldBumpRecencyForAgentChange symmetrically — correct sequence/firehose-dedup encapsulation (packages/terminal-awareness/src/agentSession.ts:20-31)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 4 round(s) · codex reviewed at `xhigh` reasoning effort · base `7c0e22166205`

### Round 1

**codex** — approved: `false`

The branch adds a coherent memory readout path across kaval, server surface state, and the client rail, with relevant unit and e2e coverage. The main problems are around failure semantics: a connected-daemon poll failure is collapsed into the same `null` state as "no daemon", the periodic sampler can accumulate overlapping hung polls, and the kaval readout can briefly show stale memory after the daemon is no longer connected. I also found two Atlas documentation issues.

Findings:
- `F1` · major · open — When the daemon is believed connected, a `system.processMemory` failure is caught and converted to `kavalRssBytes: null`, which is indistinguishable from the normal no-daemon case. The production wiring also logs this through `log.warn` at packages/server/src/index.ts:367, even though this is an actual failed RPC. That violates the repo rule that caught errors must not collapse to empty/no-data and that real errors log at error level; users get a missing kaval figure rather than an error state. (packages/server/src/memorySampler.ts:49)
- `F2` · major · open — `startMemorySampler` fires `sampleMemoryOnce` immediately and then every 5 seconds without awaiting or guarding the previous async poll. Since `pollKavalRss` is a plain stdio RPC with no timeout or signal, a wedged-but-open daemon connection can leave a request pending forever; every interval then starts another one, and even the always-available server RSS publication is blocked behind the kaval poll. (packages/server/src/memorySampler.ts:71)
- `F3` · minor · open — The kaval memory readout is rendered solely from the last `processMemory` cell value, not from the current daemon state. When the daemon transitions to restarting/degraded/dead, `daemonStatus` updates immediately but `kavalRssBytes` is only cleared by the next 5-second sampler tick, so the rail and Diagnostic Info can show stale memory for a daemon that is no longer live. (packages/client/src/ui/IdentityRail.tsx:192)
- `F4` · minor · open — The Atlas note says the daemon RSS rides the `system.heartbeat` reply, but the implementation deliberately added a separate `system.processMemory` procedure. The rendered HTML repeats the same incorrect statement, so the design record now contradicts the contract. (docs/atlas/src/content/atlas/pty-daemon-chrome-bar.mdx:19)
- `F5` · nit · open — The note received a meaningful edit in this branch, but the Atlas `updated` frontmatter still says `2026-06-14`. The branch commit touching this file is dated `2026-06-21`, and the Atlas rules say this field is the date of the last meaningful edit. (docs/atlas/src/content/atlas/pty-daemon-chrome-bar.mdx:7)

**claude** — All 5 codex findings agreed and fixed. F1 models kaval daemon memory as a discriminated union (ok/absent/error). F2/F3 robustness fixes. F4/F5 Atlas doc fixes. Typecheck and unit tests pass, lint clean. Committed 5a4a55a05.

- `F1` **fixed** — Modeled kaval memory as a discriminated union ok/absent/error in common/surface.ts so a connected-daemon poll failure is distinct from no-daemon. Rail shows an error chip, DiagnosticInfo shows poll failed, server logs at ERROR with full error object.
- `F2` **fixed** — Sampler publishes server RSS first, bounds the kaval poll with a 2s timeout, and serializes ticks via an inFlight guard. Tests added.
- `F3` **fixed** — Gated rail kaval memory on daemon connected state so it clears immediately on non-connected transition.
- `F4` **fixed** — RSS rides system.processMemory not system.heartbeat; corrected MDX and regenerated dist.
- `F5` **fixed** — Bumped updated frontmatter to 2026-06-21 and regenerated dist.

commit: `5a4a55a05519c46be91c100c5c43b89c13c34e69`

### Round 2

**codex** — approved: `false`

CLAUDE addressed the schema/error-state part cleanly and fixed the Atlas documentation issues. Two existing findings remain open: the sampler timeout bounds `sampleMemoryOnce` but does not cancel or serialize the underlying kaval RPC after timeout, and Diagnostic Info can still show stale kaval RSS during the same post-disconnect window that the rail fix handled.

Findings:
- `F1` · major · resolved — Previously, a connected-daemon memory poll failure collapsed into the same no-data shape as an absent daemon and was only logged at warn level. The current code now models kaval memory as `ok`/`absent`/`error`, preserves `error` through the surface equality key, renders an error chip in the rail, shows `poll failed` in Diagnostic Info, and logs the full error at `log.error`. (packages/common/src/surface.ts:763)
- `F2` · major · open — The timeout wrapper resolves `sampleMemoryOnce`, but it does not cancel the underlying `deps.pollKavalRss()` RPC. After a wedged-but-open stdio call times out, `inFlight` is cleared in `startMemorySampler`, so the next interval can start another kaval RPC while the old one remains pending in the client/peer. This still allows hung kaval polls to accumulate over time; the fix only bounds the wrapper promise, not the resource being leaked. (packages/server/src/memorySampler.ts:70)
- `F3` · minor · open — The rail now gates `KavalMemReadout` on `daemon()?.state === "connected"`, so that part clears immediately. Diagnostic Info still reads `kavalMemory()` directly from the last `processMemory` cell value, so if the daemon transitions away from connected just after an `ok` sample, the diagnostic snapshot can keep showing that stale RSS until the next 5-second sampler tick publishes `absent`. (packages/client/src/DiagnosticInfo.tsx:69)
- `F4` · minor · resolved — The Atlas note now correctly says daemon RSS rides the dedicated `system.processMemory` verb, separate from `system.heartbeat`, and the rendered HTML contains the same corrected text. (docs/atlas/src/content/atlas/pty-daemon-chrome-bar.mdx:19)
- `F5` · nit · resolved — The Atlas frontmatter `updated` field is now `2026-06-21`, matching this meaningful edit, and the generated dist artifact was updated. (docs/atlas/src/content/atlas/pty-daemon-chrome-bar.mdx:7)

_codex on the rebuttal:_ CLAUDE agreed with all findings, so there were no disputes to answer. I accept the F1, F4, and F5 fixes. I do not accept F2 as fully fixed because `Promise.race` with a timeout does not cancel the original stdio RPC; clearing `inFlight` after the wrapper timeout reintroduces overlapping pending polls. I do not accept F3 as fully fixed because the rail is gated now, but Diagnostic Info still reads the stale cell value without checking the daemon's current state.

**claude** — Round 2 closed the two findings codex held open. F1/F4/F5 were already resolved in round 1. F2: the timeout wrapper bounds sampleMemoryOnce but can't cancel the underlying stdio processMemory RPC, so a wedged daemon left the poll pending and the next tick could pile on another. I added guardOverlap() which tracks the real poll promise — while one is pending, further calls reject with KAVAL_POLL_BUSY immediately (no second RPC), clearing the instant the underlying poll settles; liveSamplerDeps wraps the poll in it, so at most one kaval RPC is ever outstanding. BUSY skips render as `error` but aren't re-logged every tick. F3: gated Diagnostic Info's kaval RSS field on localDaemonStatus()?.state === "connected", mirroring the rail, so it clears immediately when the daemon leaves connected instead of showing stale RSS until the next 5s tick. Tests added for both the busy-skip path and guard launch-once-settled; server + client typecheck clean, 6 memorySampler tests pass, formatter run.

- `F1` **fixed** — Already resolved in round 1 (kaval memory modeled as ok/absent/error discriminated union). codex marked resolved; no change this round.
- `F2` **fixed** — Agreed. Added guardOverlap() in packages/server/src/memorySampler.ts that wraps pollKavalRss and tracks the REAL underlying promise: while a poll is unsettled (timed out at the wrapper but still pending at the wire), the next call rejects with the new KAVAL_POLL_BUSY sentinel instead of launching a second RPC, so at most one kaval poll is ever outstanding; the guard clears the moment the underlying poll settles. liveSamplerDeps now wraps the production poll in guardOverlap. sampleMemoryOnce renders BUSY as `error` but skips re-logging it (the wedge was logged once on the original timeout via reportPollError). This is codex's suggested separate-guard approach rather than an abort signal, since processMemory is a plain request/response stdio verb with no abort path. Server RSS still publishes independently every tick. Added two tests (busy-skip-without-relog, and guard launches afresh once prior poll settles).
- `F3` **fixed** — Agreed. packages/client/src/DiagnosticInfo.tsx now gates the kaval RSS snapshot field on localDaemonStatus()?.state === "connected" (imported from ./kaval/useDaemonStatus), mirroring IdentityRail's KavalMemReadout gate. The diagnostic field clears to null the instant the daemon leaves connected rather than showing stale RSS until the next 5s sampler tick publishes absent.
- `F4` **fixed** — Already resolved in round 1 (Atlas note corrected to system.processMemory). codex marked resolved; no change this round.
- `F5` **fixed** — Already resolved in round 1 (updated frontmatter bumped to 2026-06-21, dist regenerated). codex marked resolved; no change this round.

commit: `9218956969b0b513cbeee18d642e17d47e328546`

### Round 3

**codex** — approved: `false`

Round 2 resolves the two findings that were still open: the production poll is now non-overlapping, and Diagnostic Info now gates kaval RSS on the current daemon state. I found one new regression introduced by the overlap guard itself: `guardOverlap` creates an unobserved `p.finally(...)` promise that will reject whenever the underlying kaval RPC rejects, which can surface as an unhandled rejection in the server process.

Findings:
- `F1` · major · resolved — The previous no-data/error conflation remains fixed: kaval memory is still modeled as an `ok`/`absent`/`error` discriminated union, with distinct UI and Diagnostic Info handling for poll failure. (packages/common/src/surface.ts:763)
- `F2` · major · resolved — The original overlapping-poll issue is fixed: `liveSamplerDeps` wraps the production kaval poll in `guardOverlap`, which tracks the real underlying poll promise and rejects subsequent attempts with `KAVAL_POLL_BUSY` instead of launching another RPC while the prior one is still unsettled. (packages/server/src/memorySampler.ts:93)
- `F3` · minor · resolved — Diagnostic Info now mirrors the rail by checking `localDaemonStatus()?.state !== "connected"` before reading the cached kaval memory cell, so stale RSS is hidden immediately when the daemon leaves connected. (packages/client/src/DiagnosticInfo.tsx:75)
- `F4` · minor · resolved — The Atlas text still correctly says daemon RSS rides `system.processMemory`, separate from `system.heartbeat`. (docs/atlas/src/content/atlas/pty-daemon-chrome-bar.mdx:19)
- `F5` · nit · resolved — The Atlas `updated` frontmatter remains `2026-06-21`. (docs/atlas/src/content/atlas/pty-daemon-chrome-bar.mdx:7)
- `F6` · major · open — New regression from this round: `guardOverlap` stores `inFlight = p.finally(...)` but returns the original `p`. If the underlying kaval RPC rejects, the caller may handle `p` via `withTimeout`, but the separate promise returned by `finally` rejects with no handler. That can create an unhandled rejection in the server process, especially when a timed-out poll later rejects as the socket closes. (packages/server/src/memorySampler.ts:100)

_codex on the rebuttal:_ CLAUDE agreed with F2 and F3, and I accept both fixes for the original findings. The overlap guard closes F2's accumulation problem, and the Diagnostic Info daemon-state gate closes F3. There were no disputes to answer. The only remaining issue is new F6, introduced by the round-2 guard implementation.

**claude** — Round 3 of the codex debate. Only one finding was still open (F6, new this round); F1-F5 were already resolved in prior rounds per codex's round-2 verdict. I agreed with F6 and fixed it: in `guardOverlap`, the code stored `inFlight = p.finally(() => ...)` but returned the original `p`. The `.finally()` call creates a NEW derived promise that rejects whenever `p` rejects; since that derived promise was only assigned to `inFlight` and never awaited or caught, a rejecting kaval poll surfaced as an unhandled rejection in the server process. Fixed by storing `inFlight = p` directly and clearing it via a separate consumed `void p.then(clear, clear)` handler that observes nothing the caller (via `withTimeout`) doesn't already see. Added a regression test that asserts a rejecting poll clears the guard, lets the next call poll afresh, and produces no `unhandledRejection`. All 7 memorySampler tests pass, server typecheck is clean, and biome reports the touched files already formatted. Committed 56435858cef55377efc8eb4c8b3237621119078b.

- `F1` **fixed** — Already resolved in round 1 (kaval memory modeled as ok/absent/error discriminated union in packages/common/src/surface.ts). codex marked resolved in round 2; no change this round.
- `F2` **fixed** — Already resolved in round 2 (guardOverlap added so at most one kaval RPC is ever outstanding). codex marked resolved in round 2; the guard's own regression is tracked as F6 below.
- `F3` **fixed** — Already resolved in round 2 (DiagnosticInfo gates kaval RSS on localDaemonStatus()?.state === "connected"). codex marked resolved; no change this round.
- `F4` **fixed** — Already resolved in round 1 (Atlas note corrected to system.processMemory). codex marked resolved; no change this round.
- `F5` **fixed** — Already resolved in round 1 (updated frontmatter bumped to 2026-06-21, dist regenerated). codex marked resolved; no change this round.
- `F6` **fixed** — Agreed — codex is correct. In packages/server/src/memorySampler.ts guardOverlap did `inFlight = p.finally(() => { inFlight = undefined; })` and `return p`. The `.finally()` produces a derived promise that rejects whenever the underlying poll `p` rejects; it was stored in `inFlight` and never observed, so a rejecting kaval RPC (e.g. a timed-out poll later rejecting as the socket closes) would surface an unhandled rejection in the server process. Fixed: `inFlight = p; void p.then(clear, clear); return p;` — the cleanup is now its own consumed handler that swallows nothing the caller doesn't already observe via withTimeout. Added a regression test in memorySampler.test.ts asserting a rejecting poll clears the guard, the next call polls afresh, and no unhandledRejection fires.

commit: `56435858cef55377efc8eb4c8b3237621119078b`

### Round 4

**codex** — approved: `true`

All findings from the prior rounds are resolved in the current tree. The round-3 change fixes F6 correctly by storing the original poll promise in `inFlight` and attaching a separate consumed `then(clear, clear)` cleanup handler, so a rejecting poll no longer creates an unobserved rejecting derived promise. The added regression test covers rejection cleanup, subsequent retry, and absence of `unhandledRejection`.

Findings:
- `F1` · major · resolved — The previous no-data/error conflation remains fixed: kaval memory is modeled as an `ok`/`absent`/`error` discriminated union, with distinct user-visible handling for poll failure. (packages/common/src/surface.ts:763)
- `F2` · major · resolved — The overlapping-poll issue remains fixed: `guardOverlap` tracks the real underlying kaval poll and rejects subsequent attempts with `KAVAL_POLL_BUSY` while it is unsettled. (packages/server/src/memorySampler.ts:93)
- `F3` · minor · resolved — Diagnostic Info remains fixed: kaval RSS is gated on `localDaemonStatus()?.state === "connected"`, so stale RSS is hidden immediately when the daemon leaves connected. (packages/client/src/DiagnosticInfo.tsx:75)
- `F4` · minor · resolved — The Atlas note still correctly says daemon RSS rides the dedicated `system.processMemory` verb, separate from `system.heartbeat`. (docs/atlas/src/content/atlas/pty-daemon-chrome-bar.mdx:19)
- `F5` · nit · resolved — The Atlas `updated` frontmatter remains `2026-06-21`. (docs/atlas/src/content/atlas/pty-daemon-chrome-bar.mdx:7)
- `F6` · major · resolved — The unhandled-rejection regression is fixed. `guardOverlap` now stores `inFlight = p` and uses `void p.then(clear, clear)` for cleanup, so the cleanup branch consumes both settle paths and does not create an unobserved rejecting `finally` promise. (packages/server/src/memorySampler.ts:100)

_codex on the rebuttal:_ CLAUDE agreed with F6, and I accept the fix. The current implementation avoids the unobserved derived promise, still clears the guard on both success and failure, and the new test exercises the rejecting-poll case plus subsequent polling. F1-F5 remain resolved with no regression in the inspected current state.

srid:

**All clear.** Reviewed the full memory-readout change against the rule checklist and a correctness fact-check.

| Pass | Findings | Outcome |
|------|----------|---------|
| Rules | 0 | Clean across ~30 rules (DRY, invalid-states-unrepresentable, no-silent-error-swallowing, caught-error-must-not-collapse-to-empty, subscription-must-surface-errors, no-unbounded-growth, errors-must-log-at-error, e2e-poll-async-state, …) |
| Fact-check | 0 actionable | 3 candidates evaluated and rejected as false positives |
| Elegance | — | Skipped (`--no-elegance`; the simplify track already ran over this tree) |

Three fact-check candidates were examined and found to be correct/intended behavior:
1. **`guardOverlap` cleanup** — `void p.then(ok, err)` is deliberately *not* `p.finally()`: finally would attach an observed handler that intercepts the caller's rejection. The pattern is correct (and commented).
2. **Stale kaval MB on a sub-5s reconnect** — `kavalMemoryDisplay()` gates on the live `connected` state, returning `null` while disconnected; the worst case is a brief correct-but-stale `ok` figure after reconnect until the next tick. Cosmetic, not a logic error.
3. **`reportPollError` throwing** — would surface as an unhandled rejection → the global handler's `process.exit(1)`: the project's intentional fail-loud, not a silent swallow.

Issue #951, opened 2026-05-21
Remote terminals: spawn shells on hosts from ~/.ssh/config


srid:
A Kolu user should be able to open a terminal whose shell runs on a remote machine they already SSH into, and have that terminal behave exactly like a local one in every Kolu surface.

The user keeps their SSH access set up the way they always have — entries in `~/.ssh/config` (or wherever their SSH client reads from), passing key-based auth, accessible via `ssh <alias>`. Kolu picks those up; the user doesn't configure anything inside Kolu.

## What the user expects to be able to do

- **Create a remote terminal from the command palette.** "New terminal" exposes one entry per host (a local one plus one per discovered SSH alias). Picking a host opens a terminal whose shell lives on that machine — `pwd` says the remote path, `hostname` says the remote name, `ps` lists the remote processes.

- **Tile appears immediately.** Picking a remote host doesn't freeze the UI while SSH handshakes. The tile renders right away with a "Connecting to `<host>`…" placeholder where xterm would normally be. The placeholder swaps to the live terminal as soon as the shell is ready. Any input typed during the connecting window is queued and delivered when the shell comes up.

- **Sub-terminals inherit the parent's host.** Opening a sub-terminal from a remote tile spawns it on the same remote — the user doesn't have to re-pick the host every time.

- **Session restore re-spawns each terminal on its original host.** Quitting and reopening Kolu brings every terminal back where it was; remote ones reconnect to the same machine, local ones stay local. During restore every saved tile appears at once with the same "Connecting…" placeholder treatment — the user doesn't wait for the slowest host before seeing any tiles.

- **Every per-tile surface works on remote tiles, identical to local ones**:
  - **Branch chip in the title bar** for remote repos
  - **Worktree create / remove** for remote repos (right-click → "New worktree…" works against a remote)
  - **GitHub PR badge** — checks status, PR number, author, all populated from the remote repo's GitHub remote
  - **Agent indicator** (Claude Code / Codex / OpenCode) — the badge lights up and reports state ("thinking", "tool_use", "waiting") for sessions running on the remote
  - **Workspace switcher / dock entries** for remote tiles, with a visual mark so the user can tell remote from local at a glance
  - **Right-panel Code tab** — file tree, git diff, file preview all work against the remote repo

- **Network blips don't kill the terminal.** Walking out of WiFi range or suspending the laptop and coming back should leave each remote terminal intact — same prompt, same scrollback, no lost output. The terminal reconnects and the user picks up where they left off.

[...]

PR #1505, opened 2026-06-22
Complete `mirrorRemoteSurface` to a total dual

srid:
**`mirrorRemoteSurface` was the consume-side dual of `implementSurface` — but a leaky half-mirror.** Its `SurfaceSink` carried only cells/collections/streams/events; procedures fell through, harmless only because every surface mirrored so far had zero of them. A dual must mirror *every* primitive its producer wires. This completes it.

The split is by kind. **Streaming state is PUSH** — subscribed and folded into the caller's sink, as before. **A procedure is PULL** — a local call runs on the remote and returns — so it can't live in the push-only sink. It comes back as a **forwarding stub** on the return, which is now `{ procedures, done }` instead of a bare promise:

```ts
const { procedures, done } = mirrorRemoteSurface(surface, client, {
  collections: { awareness: { upsert, remove } },   // PUSH → sink
});
await procedures.terminal.kill({ id });   // PULL → forwarded to the remote, returns its result
await done;                               // the former Promise<void> — settles when the link closes
```

### Why the return shape changed

`done` *is* the old `Promise<void>` (settle-on-link-close). Procedures couldn't be bolted onto it — a stub is callable *during* the mirror's life, but `done` doesn't resolve until the link dies. So the two halves sit side by side: `procedures` available synchronously (a stub is a typed passthrough — no standing cost, so **every** procedure is forwarded, `{}` if the surface has none), `done` the lifecycle signal.

Grafting the stubs into a re-serve makes **`serve ∘ mirror ≈ identity`** — the location-transparency the remote-terminals epic rests on — pinned by a hermetic unit test (`implementSurface` → `directLink` → mirror → re-serve → `directLink` → call round-trips `{ y: 42 }`).

### The three failure modes stay distinct

| Failure | Behavior |
| --- | --- |
| Upstream per-primitive blip (a remote stream errors) | logged, that subscription settles — one bad stream never sinks the mirror |
| Sink throws (your local fold is buggy) | **rejects `done`** — fail-fast, never logged-and-swallowed |
| Calling a procedure the client doesn't expose | **rejects** with a client/surface mismatch, at call time — never a silent `undefined` |

### Callers

[...]

PR #1506, opened 2026-06-22
R6: lift fs/git into @kolu/terminal-workspace — one library, one surface, two homes

srid:
**The host-side library that knows a terminal's workspace now serves its fs/git reads too, beside awareness, from one impl** — `@kolu/terminal-awareness` (the R4 sensor set) grows into `@kolu/terminal-workspace`, and the thin fs/git wrapper that lived in kolu-server's `localEndpoint` moves in next to the sensors. This is phase **R6** of [remote terminals over SSH](https://github.com/juspay/kolu/issues/951): with one fs/git impl, kolu runs it **in-process** for local terminals and `arivu` serves it **over ssh** for remote ones — the same two homes the sensors already proved — so R8's remote Code tab mirrors a *proven* surface, not a second hand-rolled one.

No user-visible behavior changes here: the kolu app surface and the browser client are byte-for-byte unchanged. Only the wrapper's *home* moved.

### One impl, two faces

```
                    ┌──────────────────────────────────────┐
                    │  @kolu/terminal-workspace             │
 kolu-server ──────▶│  createTerminalWorkspaceEndpoint(log) │◀────── arivu
 (in-process,       │   = the kolu-git fs/git wrapper        │       (serves the
  binds its         │                                        │        whole surface
  TerminalEndpoint) │  terminalWorkspaceSurface (./surface) │        over ssh)
                    │   awareness + activity + version       │
                    │   + fs.* / git.* procedures            │
                    │   + subscribeRepo/FileChange streams   │
                    └──────────────────────────────────────┘
```

The dependency arrow points **out**: both kolu-server and arivu depend on the library; it depends on neither. The terminal-endpoint *orchestration* (spawn · adopt · the registry) and the binary-preview/iframe-URL layer over `fs.readFile` stay kolu-server's — only the leaf wrapper moved.

### Two commits

| Commit | What |
| --- | --- |
| `refactor:` | `git mv` the package → `@kolu/terminal-workspace`; retire `@kolu/arivu-contract` by folding its surface into `@kolu/terminal-workspace/surface` (+ `/socket`). Mechanical, zero behavior change — typecheck/lint green. |
| `feat:` | Lift the fs/git wrapper (`./endpoint`), grow the surface with `fs.*`/`git.*` **procedures** + the change **watcher streams** (contract `0.2 → 0.3`, additive), wire kolu's in-process face and arivu's served face, add tests. |

### Why procedures + watcher streams (not the old streams)

The fs/git *reads* become **procedures** — request → response, location-transparent under R8's mirror. The live deltas become **watcher streams** that carry a `seq` *pulse*, not data: a consumer requeries the procedures on each pulse. The watcher uses the raw `source` arm of `StreamImplDeps`, **not** the poll-shape arm, so the `seq` counter is allocated *per subscription* — the poll-shape's `read`/`install` are one shared dep-object closure that would leak `seq` across concurrent subscribers. A test pins exactly this (two subscriptions each start at `{seq:0}`).

### Design philosophy

- **Fail fast / no fallbacks** — `unwrapGit` moved verbatim: a git error throws `ORPCError`, never collapses to an empty result (pinned by a non-repo-path throws test). The watcher's `onReadError` logs, never swallows. The factory takes a *required* injected logger, not a knob.
- **Electricity** — fs/git over `kolu-git` is a *leaf* wrapper (a bounded algorithm), so it folds into the existing host-side library rather than spawning a new `@kolu/*` receptacle; `kolu-git` stays the package hiding the `@parcel/watcher`/`simple-git` volatility.
- **Reuse the source of truth** — reuses `pollOnEvent` (snapshot-then-deltas by construction), kolu-git's refcounted/debounced watchers, `kolu-git/schemas` for the procedure I/O, and one `createTerminalWorkspaceEndpoint` backing both faces.

> The only `packages/surface` change is a one-line JSDoc example rename (`arivuSurface` → `terminalWorkspaceSurface`) — no exported API delta, so no paired drishti PR is needed.

Plan of record: [remote-terminals.mdx → R6](docs/atlas/src/content/atlas/remote-terminals.mdx).


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `def93fb79584`

### Applied (4)
- `lowy-1` "One surface, two homes" is half-true: the in-process home serves a parallel surface shape, not terminalWorkspaceSurface — commit `69cda0843`
- `hickey-1` Two parallel surface descriptions of the same fs/git domain reads — commit `69cda0843`
- `hickey-3` Package root docstring not updated to its new responsibilities — commit `e69d611b8`
- `hickey-5` `unwrapGit` lost its anti-cycle rationale on the move — commit `4c195a68f`

### Agreed — no change (7)
- `lowy-2` fsReadFile binary-preview branch splits the readFile contract across the boundary (packages/terminal-workspace/src/surface.ts:77-80 (FsReadFileTextOutputSchema) vs packages/server/src/surface.ts:296-325 (fsReadFile))
- `lowy-3` unwrapGit lives in endpoint.ts (the impl module) yet is imported as a standalone helper by router.ts (packages/terminal-workspace/src/endpoint.ts:73-98; consumed at packages/server/src/router.ts:16)
- `lowy-4` changePulseSource duplicates the seq-pulse mechanism that koluSurface's fs/git streams solve a different way (packages/terminal-workspace/src/serveFsGit.ts:33-48 vs packages/server/src/surface.ts:271-325)
- `lowy-5` socket.ts hardcodes the 'arivu' runtime namespace inside a package named terminal-workspace (packages/terminal-workspace/src/socket.ts:21-27)
- `lowy-6` package exports list reads as a coherent primitive — passes the Surface/package-coherence test (packages/terminal-workspace/package.json:10-17)
- `hickey-2` Workspace surface re-declares fs/git input/output schemas instead of reusing kolu-git's canonical ones (packages/terminal-workspace/src/surface.ts:889-901 (FsFileInputSchema, FsReadFileTextOutputSchema))
- `hickey-4` `arivu`-named socket and contract-version constant living in a package named `terminal-workspace` (packages/terminal-workspace/src/socket.ts:1-27 (arivuSocketPath, app:"arivu") and surface.ts:860 (TERMINAL_WORKSPACE_CONTRACT_VERSION consumed as arivu-tui's CLI --version, bin.ts:51-53))

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `def93fb79584`

### Round 1

**codex** — approved: `false`

The extraction is mostly coherent: the shared fs/git endpoint matches the old local behavior, arivu wires the expanded surface through the same socket, and package dependencies/exports look consistent. I did find one correctness issue in the new pulse streams and a few documentation inconsistencies introduced by the rename/expansion. I could not run typecheck/tests in this read-only review environment because `pnpm` is not installed.

Findings:
- `F1` · major · open — The new repo/file change streams use `pollOnEvent`, whose first action is to yield `read()` before installing the watcher. That was tolerable for existing value-bearing streams because the first frame contains the current data, but these streams yield only `{seq: 0}` and expect the consumer to re-query `fs.*`/`git.*`. A consumer following that pulse-then-requery contract can receive `seq:0`, perform the read, and only then pull the next iterator frame that installs the watcher; any file/git change in that gap is missed forever, leaving remote workspace state stale. (packages/terminal-workspace/src/serveFsGit.ts:40)
- `F2` · minor · open — The arivu README still describes arivu as serving only one `awareness` collection/surface. In this branch arivu also serves the terminal workspace fs/git procedures and change-pulse watcher streams on `terminalWorkspaceSurface`, so the package docs now understate the daemon's contract. The same awareness-only wording also appears in the root Architecture row for arivu. (packages/arivu/README.md:7)
- `F3` · nit · open — The kolu-git README still says the server unwraps `GitResult` via `unwrapGit()` in `router.ts`, but this branch moved that boundary helper into `@kolu/terminal-workspace/endpoint`. That sends future readers to the wrong layer for the error mapping. (packages/integrations/git/README.md:15)
- `F4` · nit · open — The R6 history bullet says fs/git was lifted "onto the one surface" and "served in-process by kolu and by arivu remotely," which contradicts the R6 section just above: R6 intentionally ships one fs/git implementation, not one surface served by both homes, and the single shared surface is deferred to R8. (docs/atlas/src/content/atlas/remote-terminals.mdx:204)

**claude** — Reviewed CODEX's 4 findings on the @kolu/terminal-workspace fs/git lift. Fixed all 3 documentation findings (F2, F3, F4) and regenerated the Atlas dist; disputed the one correctness finding (F1) with a concrete technical rebuttal. Committed as ad058ef40.

F1 (DISPUTED): CODEX claims serveFsGit's pulse streams have an install-after-snapshot race where a change in the gap between {seq:0} and watcher-install is lost. The mechanics it describes (watcher installs only on the 2nd iterator pull) are real — I confirmed with a generator-evaluation test — but its harm model is wrong. The 2nd pull is driven by oRPC eagerly draining the SERVER-side pollOnEvent generator immediately after yielding {seq:0}, NOT by the consumer's separate re-query RPC. So the watcher is live (microtasks after subscribe) before the consumer's re-query even reaches the server, and any change after the re-query is caught by the now-installed watcher (the dirty-flag in repoEventStream captures events even between pulls). There is no lost-update window for the pulse-then-requery contract. Moreover this snapshot-then-install ordering is the documented pollOnEvent contract (packages/surface/src/server.ts:1282 "snapshot+install+re-read"), reused unchanged from the existing value-bearing streams (gitStatus/gitDiff/fsListAll/fsReadFile in packages/server/src/surface.ts:271-325) that this branch leaves untouched — if it were a defect it'd be a pre-existing framework-wide one, and fixing it would touch shared @kolu/surface (also consumed by drishti). Finally, no production consumer subscribes to these pulses yet (only the arivu daemon serves them; arivu-tui doesn't read them).

F2/F3/F4: all stale-docs introduced by the rename/expansion — fixed in the working tree.

- `F1` **disputed** — No lost-update race. Confirmed via a generator-evaluation test that install() runs on the 2nd iterator pull — but on the SERVER side that pull is driven by oRPC eagerly draining the pollOnEvent generator right after it yields {seq:0} (packages/surface/src/server.ts:266-269 streamHandlers `for await ... yield`), not by the consumer's separate re-query RPC. The watcher is therefore live ~microtasks after subscribe, before the consumer's re-query reaches the server; repoEventStream's dirty flag (server.ts:363-379) then captures every change even between consumer pulls. So a consumer following pulse-then-requery cannot end up stale. The snapshot-then-install ordering is the documented pollOnEvent contract (server.ts:1282 comment 'snapshot+install+re-read') and is reused byte-for-byte from the existing value-bearing streams the branch leaves untouched (packages/server/src/surface.ts:271-325 gitStatus/gitDiff/fsListAll/fsReadFile) — not introduced here. If real it would be a pre-existing framework defect in shared @kolu/surface (also consumed by drishti), out of scope. Additionally no production consumer subscribes to these pulses in this branch. Left packages/terminal-workspace/src/serveFsGit.ts unchanged.
- `F2` **fixed** — Updated packages/arivu/README.md:5-13 (opening description) and the `--socket` flag line to state arivu serves the whole terminalWorkspaceSurface — awareness collection + version cell + activity stream, plus the R6 fs.*/git.* read procedures and the subscribeRepoChange/subscribeFileChange change-pulse watcher streams — and that arivu-tui consumes only the awareness/activity side today while the fs/git surface is what a remote kolu-server mirrors in R8. Also updated the root README.md arivu row (line 219) with the same clarification.
- `F3` **fixed** — Updated packages/integrations/git/README.md:15 — unwrapGit() now lives in @kolu/terminal-workspace/endpoint (verified: defined at packages/terminal-workspace/src/endpoint.ts:80, imported by packages/server/src/router.ts:16). Reworded to name that package as the home of the boundary helper, with kolu-server importing it in router.ts for its remaining git RPCs.
- `F4` **fixed** — Rewrote the R6 history bullet in docs/atlas/src/content/atlas/remote-terminals.mdx:204 to match the R6 section (lines 154/159): one fs/git impl (createTerminalWorkspaceEndpoint) re-exposed through two deliberately different contract shapes — kolu-server in-process on koluSurface value-bearing streams, arivu on terminalWorkspaceSurface as procedures + {seq} pulse watcher streams (that surface's contract 0.2 -> 0.3) — explicitly 'one impl, not one shared surface', with the single shared surface mirrored by kolu in R8. Regenerated docs/atlas/dist/remote-terminals.html via `just atlas::build` (verified idempotent/in-sync).

commit: `ad058ef40b37e96afe4dcc09de6f6ac52b24adac`

### Round 2

**codex** — approved: `true`

All previously raised findings are now resolved. F2-F4 were fixed in the current tree, and I accept CLAUDE's F1 rebuttal as sufficient for this branch: my original concern was real at the raw async-generator/lazy-consumer level, but I no longer consider it a demonstrated production correctness bug in the current change.

Findings:
- `F1` · major · resolved — Previously raised concern: the pulse streams yield `{seq:0}` before `pollOnEvent` installs the watcher, creating a possible missed-change window for pulse-then-requery consumers. (packages/terminal-workspace/src/serveFsGit.ts:40)
- `F2` · minor · resolved — Previously stale docs described arivu as serving only the awareness collection, omitting the expanded terminal workspace surface. (packages/arivu/README.md:7)
- `F3` · nit · resolved — Previously stale docs said `unwrapGit()` lived in `router.ts` after it moved to `@kolu/terminal-workspace/endpoint`. (packages/integrations/git/README.md:15)
- `F4` · nit · resolved — Previously contradictory Atlas history text said R6 lifted fs/git onto one shared surface served by both homes, while the R6 section says one impl with two contract shapes until R8. (docs/atlas/src/content/atlas/remote-terminals.mdx:204)

_codex on the rebuttal:_ F1: I accept the dispute. The generator ordering I pointed out is real: `pollOnEvent` yields its initial read before `repoEventStream` calls `install()`. But the finding overstated that as a concrete branch bug. In the current branch the pulse streams are only served by arivu, no production consumer subscribes to them yet, and the expected surface consumers/transport keep draining the server stream after the first frame rather than blocking watcher installation on a separate requery RPC. Once installed, `repoEventStream`'s dirty flag preserves events that happen between pulls. This is not enough to keep F1 open. F2, F3, and F4: verified fixed in the current tree.

srid:

Ran the rules + fact-check passes over the R6 diff (elegance pass skipped — `/simplify` already ran this gauntlet).

| Pass | Found | Outcome |
| --- | --- | --- |
| Rules | 1 | **Fixed** |
| Fact-check | 0 | Clean |

**`errors-must-log-at-error`** — `packages/terminal-workspace/src/serveFsGit.ts` — the watcher's `onReadError` logged a failed change-pulse `read()` at `warn`. A rejected read is an actual error, not a degraded-but-recoverable state, so it must surface at `error`. Fixed in `6e7be3741` (`log.warn` → `log.error`).

Fact-check found no silent error-swallowing, no inaccurate fallbacks, no wishful thinking: `unwrapGit` throws a typed `ORPCError` on every arm (never collapses to empty), the per-subscription `seq` closures prevent cross-subscriber leakage, and every error path either throws or surfaces via a callback.

srid:

The **x86_64-linux** lane ran on **`kolu-ci-3`** (`dev-x86-64-linux-03`) — commit `6e7be3741`, **passed** ✓

- **Lane wall** (pipeline): **3m15s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 2m38s | ✓ |
| `ci::flake-check` | 58s | ✓ |
| `ci::nix` | 57s | ✓ |
| `ci::smoke` | 56s | ✓ |
| `ci::unit` | 51s | ✓ |
| `ci::pnpm-hash-fresh` | 49s | ✓ |
| `ci::render` | 33s | ✓ |
| `ci::biome` | 32s | ✓ |
| `ci::fmt` | 30s | ✓ |
| `_ci-setup` | 27s | ✓ |
| `ci::install` | 9s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/6e7be37/timings.jsonl`; pool state is a live `flock` probe.</sub>

srid:
## Evidence

R6 is a **behavior-preserving refactor** — the fs/git wrapper moved out of kolu-server into `@kolu/terminal-workspace`, but kolu's app surface and the browser client are byte-for-byte unchanged, so there is **no on-screen change** to capture: the Code tab renders the same file tree, file contents, and diffs as before, now sourced from `createTerminalWorkspaceEndpoint` instead of the inline `localFs`/`localGit`. Per the project's evidence policy this is the sanctioned skip case (a pure internal refactor with no externally visible effect), so the proof is regression-safety, not a pixel.

What proves the lift didn't break the user-facing Code-tab path:

- **Both `ci::e2e` lanes green** on `6e7be3741` — `ci::e2e@x86_64-linux` (2m38s) and `ci::e2e@aarch64-darwin` (6m29s). The e2e suite drives the Code tab over the same `gitStatus`/`gitDiff`/`fsListAll`/`fsReadFile` app-surface streams, which now read off the lifted endpoint — a green run is the end-to-end proof the local fs/git path is unchanged.
- **New unit tests** (in `ci::unit`, green both lanes) pin the lifted behavior directly: the endpoint factory (`fs.listAll`/`readFile`/`statFileMtimeMs`, `git.getStatus`/`getDiff`) against a temp git repo; the **fail-fast** contract (a non-repo path throws `ORPCError`, never resolves to empty); the served surface round-trip over a `directLink`; the **per-subscription `{seq}` pulse** (two concurrent subscriptions each start at `{seq:0}`); and arivu serving fs/git over a real unix socket (the "second home").
- **`ci::atlas-sync` green** — the regenerated `docs/atlas/dist/` matches the edited notes.

All 30 required contexts pass on the final `HEAD`.

srid:
## Review — R6: `@kolu/terminal-workspace`, fs/git lifted

Clean execution of the plan: the awareness library grows to **one** surface (awareness + activity + version **+ fs/git procedures + watcher streams**), one `createTerminalWorkspaceEndpoint` backs both the in-process (kolu) and served (arivu) faces, dependency arrow points out. Highlights: the **`seq`-per-subscription** watcher is correct and subtle — the raw `source` thunk arm (fresh closure per subscribe) instead of the shared poll-shape `{read,install}` arm that would leak `seq` across concurrent subscribers (`serveFsGit.ts` L13–48), pinned by a two-subscription test; reads are **procedures**, deltas a **`seq` pulse, not data** (requery-on-pulse) — the location-transparent shape R8's mirror wants; fail-fast throughout (`unwrapGit` verbatim, `onReadError` logs not swallows, required injected logger).

**Concerns**

1. **71 files for a "mechanical" rename.** Retiring `@kolu/arivu-contract` + `git mv` → `@kolu/terminal-workspace` is a wide blast radius; the zero-behaviour-change claim rests on CI (30 green) + the byte-identical-app-surface assertion. Worth a final grep that **no `@kolu/arivu-contract` reference survives** (the package is deleted).
2. **Pulse→requery adds one remote round-trip per change under R8's mirror** — a watcher pulse arrives, then the consumer re-calls the forwarded procedure for fresh data. Fine for an interactive Code tab; noting the cost vs. streaming the data itself. (Design choice — it's what keeps fs/git reads clean procedures.)

## Merge order (both PRs)

R6 and R7 are **logically independent** (R6⊥R7 — the plan's R5·R6·R7 are any-order), so correctness doesn't force a sequence. But they **textually conflict on 7 files**: `mirrorRemoteSurface.ts`, `surface/README.md`, `remote-terminals.mdx` (+ dist), and crucially the mirror **call sites** in `arivu/daemon.ts` + `arivu-tui/fleet.ts`, which **both** rewrite — R6 for the package rename, R7 for the `{procedures, done}` return-shape.

**Recommendation: merge #1505 (R7) first, then #1506 (R6).** Settle the framework contract, then rebase the rename onto a stable mirror API, combining *rename + `.done`* once at the shared call sites. **Verify in that rebase:** each `mirrorRemoteSurface(...)` keeps **both** `terminalWorkspaceSurface` **and** `.done` — a dropped `.done` silently no-ops (R7's un-type-caught hazard). Caveat: R7's merge is gated on drishti #71 (`surface.md`); if that isn't green yet, merge **#1506 (R6) first** (it's ungated) and bring R7 in after, applying the same call-site care.

PR #1510, opened 2026-06-22
R8 — compose terminal-workspace into kolu-server, dissolve the fold

srid:
**kolu-server now serves `terminalWorkspaceSurface` as a third member of the surface it serves the browser** — the same surface arivu serves remotely — composed in-process beside `koluSurface` + `surfaceApp`. That collapses R6's "one impl, two contract shapes" into one impl, one surface, and turns the server-side awareness *fold* into a client-side *join*. This is **R8** of the remote-terminals epic (#951); plan of record is [`remote-terminals.mdx` → R8](https://github.com/juspay/kolu/pull/1508).

> ⚠️ **Draft / in progress.** Opened early per request. Implementation is landing incrementally; not yet green on CI.

### The moves

| Move | Where |
| --- | --- |
| **Compose the 3rd surface** | `terminalWorkspace` added to the `surfaces` map (auto-propagates to contract + client); `buildWorkspaceSurfaceDeps(...)` slotted into the same `implementSurfaces` call |
| **Per-terminal dispatch seam** | `server/.../workspaceSurface.ts` — `awarenessFor(id)`/`awarenessAll()` project the local registry today; the one line R9 adds for the remote arm is marked in place |
| **Delete kolu's fs/git** | the 4 value-bearing streams gone from `koluSurface`; fs/git now ride `terminalWorkspaceSurface` as procedures + `{seq}` change-pulse watcher streams |
| **The fold dissolves** | `terminalMetadata` shrinks to kolu's own fields (`projectKoluFields`); the live `AwarenessValue` rides `terminalWorkspaceSurface.awareness` (`projectAwareness`); the browser rebuilds the full record with `joinTerminalMetadata` |
| **fs/git client cost** | `useWatchedRead` — a procedure-plus-pulse → value adapter that re-queries on each pulse; Code-tab call sites are drop-in |

### Design choice worth reviewing — Shape A, and the persisted/live "fence"

The internal server record stays **whole**; only the *wire* splits, and the browser joins (**Shape A**). This keeps session save/restore, sleep/wake, and the sensors untouched, and needs **no on-disk migration**.

One consequence for the plan text: the plan claims R8 *deletes* the persisted/live fence (the rule that keeps the ~150 ms agent-status firehose off the session autosave). **Shape A keeps the fence as an internal detail** — it only disappears from the wire. Literally deleting it would require splitting the internal record (**Shape B**), a much larger refactor that buys no user-facing or structural win (and a `cwd`-shaped remnant survives even then, since `cwd` is both live and must-persist-for-restore). Recommendation: keep Shape A and soften the plan's claim to "the fold dissolves and the merge becomes a client-side join; the fence becomes an internal detail."

Other costs surfaced, straight from the plan's trade-off table: the `activity` stream is **vestigial for local terminals** (local liveness stays client-side via `useTerminalActivity` until R9 sources remote liveness off the mirror), and the join produces a **first-paint flicker** (chrome paints before the awareness badge), which `joinTerminalMetadata` handles by seeding defaults.

Part of #951.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -64,6 +64,7 @@ let
       ./package.json
       ./pnpm-workspace.yaml
       ./pnpm-lock.yaml
+      ./patches
       ./tsconfig.base.json
       ./packages/surface
       ./packages/surface-mcp
@@ -112,7 +113,7 @@ let
     # hash-fresh` enforces this stays in sync with pnpm-lock.yaml by forcing
     # fetchPnpmDeps to re-execute (--rebuild), so stale artifacts in the
     # binary cache can't silently satisfy a hash that no longer matches.
-    hash = "sha256-8EvigbDMvYiGUqmQRlIkiMv+zxoeoBSiiz7RKmLTpX4=";
+    hash = "";
     fetcherVersion = 3;
   };
 
--- a/default.nix
+++ b/default.nix
@@ -113,7 +113,7 @@ let
     # hash-fresh` enforces this stays in sync with pnpm-lock.yaml by forcing
     # fetchPnpmDeps to re-execute (--rebuild), so stale artifacts in the
     # binary cache can't silently satisfy a hash that no longer matches.
-    hash = "sha256-8EvigbDMvYiGUqmQRlIkiMv+zxoeoBSiiz7RKmLTpX4=";
+    hash = "";
     fetcherVersion = 3;
   };
 
--- a/default.nix
+++ b/default.nix
@@ -113,7 +113,7 @@ let
     # hash-fresh` enforces this stays in sync with pnpm-lock.yaml by forcing
     # fetchPnpmDeps to re-execute (--rebuild), so stale artifacts in the
     # binary cache can't silently satisfy a hash that no longer matches.
-    hash = "";
+    hash = "sha256-8EvigbDMvYiGUqmQRlIkiMv+zxoeoBSiiz7RKmLTpX4=";
     fetcherVersion = 3;
   };
 
--- a/package.json
+++ b/package.json
@@ -27,6 +27,9 @@
       "@anthropic-ai/sdk": "^0.91.1",
       "@xterm/xterm": "6.1.0-beta.225",
       "@xterm/addon-webgl": "0.20.0-beta.224"
+    },
+    "patchedDependencies": {
+      "@pierre/trees@1.0.0-beta.4": "patches/@pierre__trees@1.0.0-beta.4.patch"
     }
   }
 }
--- a/packages/client/src/right-panel/BrowseFileDispatcher.tsx
+++ b/packages/client/src/right-panel/BrowseFileDispatcher.tsx
@@ -64,7 +64,7 @@ import { resolveWikilink } from "@kolu/solid-markdown";
 import { CommentTextSurface } from "../comments/CommentTextSurface";
 import { useCommentScrollRequest } from "../comments/scrollRequest";
 import { OptionMenu } from "../ui/OptionMenu";
-import { terminalWorkspace } from "../wire";
+import { client, terminalWorkspace } from "../wire";
 import BrowseFileView from "./BrowseFileView";
 import BrowseIframeRenderer from "./BrowseIframeRenderer";
 import { resolveMarkdownImageSrc } from "./markdownImageSrc";
@@ -135,14 +135,14 @@ const BrowseFileDispatcher: Component<BrowseFileDispatcherProps> = (props) => {
     async (i): Promise<FsReadFileOutput> => {
       if (isBinaryPreviewable(i.filePath)) {
         const mtimeMs =
-          await terminalWorkspace.rpc.surface.fs.statFileMtimeMs(i);
+          await client.surface.terminalWorkspace.fs.statFileMtimeMs(i);
         return {
           kind: "binary",
           url: `${buildTerminalFileUrl(props.terminalId, i.filePath)}?v=${Math.floor(mtimeMs)}`,
         };
       }
       const { content, truncated } =
-        await terminalWorkspace.rpc.surface.fs.readFile(i);
+        await client.surface.terminalWorkspace.fs.readFile(i);
       return { kind: "text", content, truncated };
     },
     filePulse,
--- a/packages/client/src/right-panel/BrowseFileDispatcher.tsx
+++ b/packages/client/src/right-panel/BrowseFileDispatcher.tsx
@@ -39,12 +39,15 @@ import { MarkdownRenderer } from "@kolu/solid-fileview/renderers/markdown";
 import { VideoRenderer } from "@kolu/solid-fileview/renderers/video";
 import type { SelectedLineRange } from "@kolu/solid-pierre";
 import {
+  buildTerminalFileUrl,
+  isBinaryPreviewable,
   isMarkdown,
   isRasterImage,
   isSandboxPreviewable,
   isVideo,
 } from "kolu-common/preview";
 import type { TerminalId } from "kolu-common/surface";
+import type { FsReadFileOutput } from "kolu-git/schemas";
 import {
   type Component,
   createMemo,
@@ -61,10 +64,11 @@ import { resolveWikilink } from "@kolu/solid-markdown";
 import { CommentTextSurface } from "../comments/CommentTextSurface";
 import { useCommentScrollRequest } from "../comments/scrollRequest";
 import { OptionMenu } from "../ui/OptionMenu";
-import { app } from "../wire";
+import { terminalWorkspace } from "../wire";
 import BrowseFileView from "./BrowseFileView";
 import BrowseIframeRenderer from "./BrowseIframeRenderer";
 import { resolveMarkdownImageSrc } from "./markdownImageSrc";
+import { useWatchedRead } from "./useWatchedRead";
 import { openInCodeTab } from "./openInCodeTab";
 
 // The "File truncated" banner is rendered as a sibling ABOVE the comment
@@ -113,12 +117,34 @@ export type BrowseFileDispatcherProps = {
 };
 
 const BrowseFileDispatcher: Component<BrowseFileDispatcherProps> = (props) => {
-  const fileContent = app.streams.fsReadFile.use(
-    () => ({
-      terminalId: props.terminalId,
-      repoPath: props.repoPath,
-      filePath: props.filePath,
-    }),
+  // R8: fsReadFile moved off koluSurface. The composed terminalWorkspaceSurface
+  // serves the RAW text read (`fs.readFile`) + the mtime (`fs.statFileMtimeMs`) as
+  // procedures, pulsed by `subscribeFileChange`. The binary-preview orchestration
+  // (decide binary vs text, build the kolu iframe-route URL) was kolu-server's; it
+  // moves HERE — the client knows the terminal id and builds the same `?v=<mtime>`
+  // URL via the browser-safe `buildTerminalFileUrl`.
+  const filePulse = terminalWorkspace.streams.subscribeFileChange.use(
+    () => ({ repoPath: props.repoPath, filePath: props.filePath }),
+    { onError: (err) => toast.error(`File watch: ${err.message}`) },
+  );
+  const fileContent = useWatchedRead<
+    { repoPath: string; filePath: string },
+    FsReadFileOutput
+  >(
+    () => ({ repoPath: props.repoPath, filePath: props.filePath }),
+    async (i): Promise<FsReadFileOutput> => {
+      if (isBinaryPreviewable(i.filePath)) {
+        const mtimeMs = await terminalWorkspace.rpc.surface.fs.statFileMtimeMs(i);
+        return {
+          kind: "binary",
+          url: `${buildTerminalFileUrl(props.terminalId, i.filePath)}?v=${Math.floor(mtimeMs)}`,
+        };
+      }
+      const { content, truncated } =
+        await terminalWorkspace.rpc.surface.fs.readFile(i);
+      return { kind: "text", content, truncated };
+    },
+    filePulse,
     {
       onError: (err) => toast.error(`File content stream: ${err.message}`),
     },
--- a/packages/client/src/right-panel/BrowseFileDispatcher.tsx
+++ b/packages/client/src/right-panel/BrowseFileDispatcher.tsx
@@ -134,7 +134,8 @@ const BrowseFileDispatcher: Component<BrowseFileDispatcherProps> = (props) => {
     () => ({ repoPath: props.repoPath, filePath: props.filePath }),
     async (i): Promise<FsReadFileOutput> => {
       if (isBinaryPreviewable(i.filePath)) {
-        const mtimeMs = await terminalWorkspace.rpc.surface.fs.statFileMtimeMs(i);
+        const mtimeMs =
+          await terminalWorkspace.rpc.surface.fs.statFileMtimeMs(i);
         return {
           kind: "binary",
           url: `${buildTerminalFileUrl(props.terminalId, i.filePath)}?v=${Math.floor(mtimeMs)}`,
--- a/packages/client/src/right-panel/CodeTab.tsx
+++ b/packages/client/src/right-panel/CodeTab.tsx
@@ -617,6 +617,11 @@ const CodeTab: Component<{
     return r;
   });
 
+  // FT-DEBUG: independent eager subscriber to treePaths — does the memo notify?
+  createEffect(() => console.log(`FT-WATCH tp=${treePaths().length}`));
+  // FT-DEBUG: independent eager subscriber to the exact Show `when` expression.
+  createEffect(() => console.log(`FT-SHOW when=${treePaths().length > 0}`));
+
   // Track membership rather than the treePaths array identity: browse paths
   // come from a reconciled store array whose contents can change in place.
   // Gate on the relevant stream's `pending()` — when the gitStatus / fsList
--- a/packages/client/src/right-panel/CodeTab.tsx
+++ b/packages/client/src/right-panel/CodeTab.tsx
@@ -64,9 +64,10 @@ import {
 } from "../ui/pierreTheme";
 import { realSizes } from "../ui/corvuResizable";
 import { Z_HANDLE_INNER } from "../ui/stackLayers";
-import { app } from "../wire";
+import { terminalWorkspace } from "../wire";
 import BrowseDiffView from "./BrowseDiffView";
 import BrowseFileDispatcher from "./BrowseFileDispatcher";
+import { useWatchedRead } from "./useWatchedRead";
 import FileSearchInput from "./FileSearchInput";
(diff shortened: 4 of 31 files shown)


srid:
## On "the persisted/live fence deletes" — it doesn't, under this implementation

The subtlest claim in the [R8 plan](https://github.com/juspay/kolu/pull/1508) is that R8 *deletes* the persisted/live fence. This implementation **keeps** it — worth pinning down the divergence.

### What the fence is

Each terminal has one metadata record whose fields split three ways by *who writes them and whether they survive a restart*: **server-persisted** (`cwd`, `git`, `location`, …), **client-persisted** (`themeName`, `canvasLayout`, …), and **live** (`pr`, `agent`, `foreground`). The **fence** is the persisted⇄live line, enforced by two writer functions in `metadata.ts`:

- `updateServerMetadata(…)` → writes a persisted field **and fires `terminals:dirty`** (arms the session autosave).
- `updateServerLiveMetadata(…)` → writes a live field and **deliberately does NOT fire `terminals:dirty`**.

Why it exists: the agent-status sensor ticks ~**150 ms** while an agent streams, almost always touching only live fields. Without the fence those ticks would snapshot-and-save the whole session ~twice a second. The mutator *types* are even narrowed so a sensor physically can't write a live field through the persisting helper — a compile-time wall.

### Why the plan says it deletes

The plan's reasoning: the fence is domain logic *only because* awareness lives in one co-owned record. Pull awareness onto its own surface (live, never persisted) and the live churn has no persisted neighbour to disturb — nothing left to fence. **That's correct, but only if you actually split the internal record.**

### Why this PR keeps it

This PR is **Shape A**: the server keeps one *whole* internal record; only the **wire** splits (two surfaces), and the browser joins. Session save/restore, sleep/wake, and the sensors are untouched, and there's **no on-disk migration**. But because the internal record is still whole, the 150 ms `agent` write still lands on the same object the session save reads — so the fence still earns its keep. It went **invisible to the wire, but stayed in the server's internals.**

To literally delete it you'd need **Shape B** (split the internal record so awareness is a separate live-only thing). That's a much larger refactor — every sensor write, the registry type, session save/restore, sleep/wake — and it buys **no user-facing or structural win** beyond what Shape A already delivers.

### The catch even Shape B hits

The fence deletes cleanly only for the *purely-live* fields (`pr`/`agent`/`foreground`). `cwd` is both **live** (the cwd sensor) **and must-persist** (cold restore respawns the shell in that dir) — so *something* must still throttle "cwd changed → update the saved copy" away from the 150 ms churn. That's the fence wearing a different hat. Even Shape B leaves a `cwd`-shaped remnant unless cwd's persistence moves wholly into the session blob and is re-derived on restore.

### Recommendation

Keep Shape A (this PR) and **soften the plan's claim** — e.g. *"the server-side fold dissolves and the merge becomes a client-side join; the persisted/live fence becomes an internal detail, no longer a wire concern."* Deleting the fence outright isn't worth Shape B's churn, and it can't fully vanish while `cwd` is both live and persisted. The project/join inverse is pinned by a round-trip test in this PR.

srid:
## Your facts are right. The conclusion is one boundary short.

I went through this against the code (six readers + a Hickey/Lowy/adversary pass). Everything you observed checks out: awareness already **never** reaches disk (`snapshotSession` strips the live overlay via `SavedTerminalSchema.parse({ ...entry.meta, id })`, `terminals.ts:73-80`); the fence is real (`updateServerMetadata` fires `terminals:dirty`, `updateServerLiveMetadata` doesn't, `metadata.ts`); and `cwd` is genuinely both observed-live and must-persist.

Where I disagree is the recommendation to **keep Shape A and soften the plan**. Shape A keeps **one whole co-owned record** (`entry.meta: ActiveTerminal`, `types.ts:30`) and splits only the *wire*, re-joining in the browser. So the fold, the fence, and the co-ownership all survive — relocated from the wire into the server's internals. That relocation is the smell: the wire now says "two things" while the server still insists "one thing," and you pay the full cost of separation (`projectKoluFields`/`projectAwareness`/`joinTerminalMetadata`/`useWatchedRead` + a first-paint flicker) **without** collecting the prize. The adversary I ran to defend Shape A conceded it in its last line: *"Shape A keeps a genuine complection that the clean shape actually removes; my defense rests on it being cheaper today, not structurally cleaner."*

We should pay for cleaner, not cheaper. Here's the target shape.

## The clean shape: kolu stops *holding* the observation and *reads* it

A terminal is exactly two things, and they were never one record:

- **Authored state** (kolu writes, durable): id, restore-cwd, theme, layout, sleeping-state, the frozen-pr snapshot → stays on `koluSurface.terminalMetadata`.
- **Observed state** (sensors derive, ephemeral, never persisted): cwd, git, pr, agent, foreground → lives on `terminalWorkspaceSurface.awareness`, in-process locally, mirrored remotely.

### Concrete moves

1. **Sensors stop writing `entry.meta`.** The sensor write path (`sensors.ts:213/339`, the agent detectors, the OSC-7 cwd tap at `local.ts:552/668`) targets the in-process `terminalWorkspaceSurface.awareness` collection instead of mutating kolu's record. This is the single change that removes the co-ownership — and with it the *reason* the fence exists.

2. **kolu reads observation through one seam: `awarenessFor(id)`** — you already built it in `workspaceSurface.ts`. Point it at the in-process awareness collection (local) today; R9 swaps the backing for a mirror (remote). Reads stay synchronous against the local replica either way, so the seam's signature doesn't change at R9.

3. **The three server-side awareness readers move to that seam.** This is the part the "near-migration-free" framing missed, and you're right to flag it — but it's not a surcharge, it's Shape A's deferred R9 debt:
   - `exportTranscriptHtml` (`router.ts:263-273`) reads `term.meta.{agent,cwd,git,pr}` → reads `awarenessFor(id)`.
   - `countActiveClaudeSessions` (`terminal-registry.ts:118-128`) loops `entry.meta.agent` → reads `awarenessFor`/`awarenessAll`.
   - the HTTP preview path (`index.ts:269`) reads `meta.git?.repoRoot` → reads `awarenessFor(id)`.

   Under Shape A these three would **still** need this rewiring at R9 (their `entry.meta` only holds *local* awareness; remote lives in a mirror). The clean shape pays it once, now, in the right place — and they work for remote terminals for free.

4. **`terminalMetadata` shrinks to kolu's authored fields.** Delete the live arm and the `updateServerLiveMetadata` mutator. kolu persists a small set of **restore-seeds** — cwd, the git cache, `lastAgentCommand`, `agentSession` — by **sampling them from the observation at save time** so a cold restart re-spawns and re-resolves fast. These are kolu-authored snapshots, one writer, not co-owned live fields.

## The fence vanishes — it does not move (this answers the cwd objection)

Your sharpest point: `cwd` is both live and must-persist, so "something still throttles cwd persistence — the fence wearing a different hat." It isn't, and here's the precise reason.

The fence exists for **one** reason: the ~150 ms agent firehose shares a record with a must-persist field, so a typed wall keeps the churn off the dirty channel. In the clean shape kolu **never persists agent/pr/foreground at all** (they're re-derived on wake), so the firehose never touches kolu. The only persisted-relevant observation kolu subscribes to is `cwd`, which changes at human-`cd` speed. So kolu arms its autosave on every cwd change exactly as today (`session.ts:154`) — **no firehose risk, no typed wall, one writer.** The fence's defining property (two writers, one record) is gone, not relabeled. No crash-durability regression: cwd stays event-driven.

## On the other two objections

- **"Honest streaming flicker."** Both shapes paint a half-arrived tile (your `terminalMetadataFold.test.ts:722`). The difference isn't the pixels — it's that in the clean shape the awareness-after-lifecycle paint is *correct* (live observation legitimately streams in after static lifecycle), so a seeded placeholder is honest, not a patch over a torn record. Keep your seed-defaults approach.
- **"Couples R8 and R9."** It doesn't — it **decouples** them. Everything above ships **local-only** (the in-process surface already exists from R6). R9 becomes a pure backing-swap behind `awarenessFor`. The clean shape is the version where R9 is *smallest*.

## What to keep from this PR

`projectAwareness` / `joinTerminalMetadata` are the **right shape at the wrong layer** — keep them as the genuine composition of two surfaces, drop them as a split-then-rejoin of one record. The `awarenessFor`/`awarenessAll` seam in `workspaceSurface.ts` is exactly right; it just needs the sensors writing *into* its collection and `entry.meta` losing its live arm.

I'm updating the plan note (`remote-terminals.mdx` R8) to this shape now — it'll land as the spec. Net: the "fold" was never domain logic; it was a held copy that existed only because, before arivu, the observation had nowhere else to live. R6 gave it a home, R7 gave us a perfect mirror — R8 just reads the home instead of copying it.

srid:
## Status: the wart is fixed; one CI blocker remains (fs/git live updates)

### Done — the architectural wart is gone, and e2e confirms it
The clean shape from the review landed: the sensors write the in-process
`terminalWorkspaceSurface.awareness` store (arivu's pattern), `entry.meta` is
authored-only, kolu *reads* the observation through the `awarenessFor` seam, and
the three server readers (transcript export, Claude-session count, iframe-preview
route) + session-save sampling all go through it. `projectKoluFields`/`projectAwareness`
(the split-then-rejoin) are gone; `joinTerminalMetadata` (the genuine client-side
compose) stays. The persisted/live fence is deleted (the cwd tap arms the autosave
directly — one writer, human-cd speed). Typecheck/biome/unit are green on both
platforms, and **every e2e run shows zero awareness/tile failures** — the awareness
split works end-to-end.

### Blocker — the fs/git Code-tab LIVE updates (procedure + pulse)
The 9 remaining e2e failures are all the same class: editing/deleting a file should
refresh the diff / content / browse-tree / iframe-preview live, and it doesn't
(plus one bare-repo "no page errors"). This is R8's *other* move — fs/git deleted
from `koluSurface`, served on `terminalWorkspaceSurface` as procedures + a
per-subscription `{seq}` change-pulse the client requeries on. **This is the first
time anything consumes that surface's *streams*** (R6/R7 only ever consumed its
`awareness` collection), so it's the consumption path's first real test.

What I've ruled out across ~8 CI runs:
- **client requery logic** — `useWatchedRead` re-queries on each pulse and handles
  rejections without throwing; pinned by unit tests.
- **reconcile-tracking** — the consumer tracks the nested `{seq}` (the surface
  subscription reconciles the frame in place).
- **re-subscribe churn** — memoized `repoPath` so an observation tick no longer
  resets the pulse subscription's seq.

What's left: the pulse `{seq:1}` isn't surfacing over the **websocket + surfaceClient**
path, even though `directLink` (in-process iteration) delivers it fine, the
*procedures* on the same surface work, and `koluSurface`'s old poll-shape streams
worked over the same websocket. The remotely-observable signals are exhausted — the
production client build + cucumber's stdout capture hide the browser console, and
the kolu-server log lives in an ephemeral nix sandbox on the CI box.

### Next
Debugging on a pu box with the browser console + server log visible to pinpoint why
this surface's source-arm streams don't deliver via the surfaceClient `.use()`.

srid:
## Debug update: wart fixed + two real bugs fixed; 9 live-update e2e failures localized to a timing race

Followed the pu-box debug path end to end (full traces under both the `tsx` dev server and the **nix binary** that CI actually runs). Findings:

### Fixed
1. **The architectural wart is gone and stays gone** — every CI run shows **zero** awareness/tile failures.
2. **diff→browse "Open path" content never loaded** (deterministic, 3/3 fail → 6/6 pass after fix). Root: `useWatchedRead.pending()` was an imperative flag that raced the `openInCodeTab` resolution gate, which resolved against an empty path list. Fix: derive `pending()` (`input set && no value yet`) so it's race-free. Unit-pinned.
3. **Per-pulse re-render churn** — the value was a `createSignal` reassigned wholesale, minting a fresh object reference on *every* pulse (incl. no-change ones) and re-rendering the whole Pierre tree. Switched to `createStore` + `reconcile` (the framework's `writeWrappedValue` strategy the old `koluSurface` streams used) so the tree re-renders only on a real change.

### The 9 that remain — NOT a data-flow bug
The 9 are all the Code-tab **live-update** scenarios (rm→tree, edit→diff/content, iframe refresh, new-file, commit→empty-changes) + the bare-repo page error. The on-box trace under the nix binary **proves the data flow is correct**: the watcher fires, the `{seq}` pulse reaches the client, the requery runs and returns fresh data (`out={"paths":[]}`, `out={"files":[]}`), and the membership effect clears the now-absent selection (`s=null`). Yet `@kolu/solid-pierre`'s `FileTree` DOM keeps the stale node (`data-item-selected="true"`, 44× over the 20s timeout).

`FileTree` applies **incremental diffs** against a closure-tracked `appliedPaths` and self-documents (FileTree.tsx:59,185) that "the stream resubscribes and briefly empties `paths` under load, unmounting/remounting us." That code is **unchanged from master**. What changed in R8 is the *delivery timing*: the old `koluSurface` stream pushed the new value in a single async hop; the new procedure+pulse contract takes **two** (pulse yield → a separate requery RPC + its resolution). Under the nix binary's timing that extra hop trips FileTree's diff/remount path — and only there: the `tsx` box never reproduces this set (it flakes a disjoint set instead).

### The fork
- **(a) value-bearing fs/git streams on `terminalWorkspaceSurface`** — serve fs/git as value-bearing streams (the old `koluSurface` shape) instead of procedures + `{seq}` pulse. The client consumes them via the framework's `createReactiveSubscription` end-to-end (no `useWatchedRead`, no requery hop) — i.e. exactly master's single-hop delivery the FileTree was built for. This revisits R6's deliberate "procedures forwarded, streams sunk" choice (it'd make fs/git *sunk* like awareness/activity in R9's mirror, which is fine — but it's your R6 design call). Touches the contract version + arivu's serve path.
- **(b) keep procedure+pulse, fix FileTree's diff under late/async updates** — deeper `@kolu/solid-pierre` work to make the incremental diff robust to the two-hop timing. Doesn't revisit R6, but it's surgery on the tree adapter.

Leaning (a): it restores the exact delivery shape the tree is proven to work with, and the procedure+pulse contract's only consumer today is this client.

srid:
## Definitive root cause of the 9 e2e failures — and why client-side fixes can't reach it

Followed the pu-box debug to ground truth under the **nix binary** CI actually runs, instrumenting the full chain `pulse → requery → reconcile → treePaths → treeSearch → FileTree props → Pierre`. Findings, each trace-proven:

1. **The data flow is correct.** On `rm obsolete.txt` the watcher fires, the pulse reaches the client, the requery returns fresh data (`{"paths":[]}`, `{"files":[]}`), `treePaths`/`treeSearch` recompute to empty, the Code tab's membership effect clears the selection, and Pierre's **model** updates: `tree.getItem("obsolete.txt") === false`, `getSelectedPaths() === []`.

2. **But Pierre's React-virtualized DOM never repaints.** The removed (focused+selected) row stays in the DOM for the full 20s (`data-item-selected="true"`). This is **not** a model bug, a selection bug, or a remount/leak (confirmed single mount, model empty). I tried, in order: derive `pending()` race-free (fixed a *separate* real bug — diff→browse "Open path"), `reconcile` instead of wholesale signal reassignment, keep-value-across-reloads (killed the remount churn), keep FileTree mounted + overlay the empty-state, deselect-before-remove, and finally **`resetPaths` (a full tree rebuild)** — *even resetPaths does not repaint the DOM*. So Pierre's React render is simply not flushing on model changes under R8's async procedure+pulse delivery context.

3. **It works on master** because master's koluSurface fs/git are **value-bearing streams**, and Pierre re-renders there via its documented remount path (`FileTree.tsx:59` — "the live fsListAll stream resubscribes and briefly empties paths, which unmounts/remounts this tree under load"). R8's in-place `{seq}`-pulse→requery delivery drives Pierre through a code path whose React re-render never fires.

This is an opaque `@pierre/trees` (React 19) ⇄ Solid interop re-render failure, downstream of correct data, that **cannot be fixed from the kolu consumption layer** — I've exhausted that surface (~24 CI runs). It's specific to the R8 fs/git relocation; the **awareness wart fix is independent and green** in every run.

### The decision (it reverses R6/R8's deliberate fs/git contract, so it's yours)
- **De-scope the fs/git relocation** — keep kolu's fs/git on koluSurface value-bearing streams (master's proven, repainting path); ship R8 as the **awareness wart fix + terminalWorkspaceSurface composed for awareness**. fs/git relocation becomes a follow-up where the Pierre re-render can be solved properly (likely a `@pierre/trees` fix or a value-bearing delivery on terminalWorkspaceSurface). **Green CI now**, wart fixed.
- **Value-bearing fs/git streams on terminalWorkspaceSurface** — replace procedures+pulse with value-bearing streams so the client consumes via the framework's `createReactiveSubscription` exactly as master does (restoring the repainting delivery). Reverses R6's "procedures forwarded, streams sunk"; touches the contract version + arivu.
- **Crack the Pierre re-render** — fork/patch `@pierre/trees` or its integration so the React tree flushes on model mutation regardless of delivery timing. Uncertain, third-party-internals depth.

srid:
## ✅ Resolved — full green CI on both platforms (x86_64-linux + aarch64-darwin)

Every lane green: typecheck · biome · unit · **e2e** · flake-check, both platforms. The
9 fs/git failures + the sleeping-agent failures are fixed at the **root**, not deferred.

### The three root causes (all under R8's new async change-pulse delivery)

1. **Code-tab live updates didn't repaint** (the 9: rm→tree, edit→diff/content, iframe,
   new-file, commit→empty-changes, bare-repo). Root: `@kolu/trees`' Preact view re-renders
   off a `controller.subscribe` callback that **swallowed the first emit after every
   (re)subscribe** as an "initial snapshot". The view re-subscribes whenever its layout
   deps change — which R8's fs/git churn triggers — so a real `batch`/`resetPaths` mutation
   landing in that window was dropped (model updated, `getItem`=false, but the DOM never
   repainted; even `resetPaths` couldn't recover). master dodged it via value-bearing
   streams that re-render through a full remount. **Fix:** a pnpm patch to `@pierre/trees`
   that always bumps the revision on emit (`patches/@pierre__trees@…`, applied at install;
   `patches/` added to the nix `fetchPnpmDeps` src).

2. **diff→browse "Open path" content never loaded** (a separate deterministic bug, 3/3→6/6).
   `useWatchedRead.pending()` was an imperative flag that raced the `openInCodeTab`
   resolution gate → it resolved against an empty path list. **Fix:** derive `pending()`
   (input present && no value yet) so it's race-free; keep the value across non-null reloads
   so the tree doesn't remount mid-update.

3. **Woken agent terminals never went live** (`waitForLive` timeout). `terminalMetadata`'s
   collection `upsert` is a no-op that only PUSHES to subscribers — state reaches the client
   solely through `publishTerminalState`. Sleep pushed the sleeping record, but the
   wake/spawn path (`registerActiveAndSpawn`) only emitted list-changed, so a woken
   terminal's registry meta flipped to active while the client stayed at the stale sleeping
   push (`isLive` false → dormant body never gave way to the live xterm). **Fix:** push the
   authored snapshot on wake/spawn.

Diagnosed end-to-end on a pu box under the actual nix binary (the failures were
nix-binary-specific and invisible to `test-quick`), instrumenting the full chain
`pulse → requery → reconcile → treePaths → Pierre` and the wake `registry → publish → client`.

PR #1514, opened 2026-06-22
Footnote markers in the Markdown preview open a click popover

srid:
**A footnote marker in the Code-tab Markdown preview is now a button.** The preview renders GFM footnotes as a superscript `[n]` linking to a definitions list at the very bottom — and reading one meant scrolling all the way down and back. Click or tap the marker and its definition opens in a small popover anchored right under it, where your eye already is; click anywhere else, press Escape, scroll the document, or click the marker again to dismiss. The bottom "Footnotes" section is untouched — the popover reads its content from there, so it's an *additional* way in, not a replacement.

It threads the *same* render → sanitize → interact → host-overlay seam that alerts (`data-md-alert`) and wikilinks (`data-md-wikilink`) already use, so it adds **no dependency and no new package** — `@kolu/solid-markdown` gains one callback; the client renders the overlay with the hook the wikilink-disambiguation menu already uses.

### How it fits together

```
click [n]  ─▶  bindInteractions          ─▶  onFootnote(anchor, def)  ─▶  FootnotePopover
(package)      recognises data-md-footnote     (new package callback)       (client: Portal +
               resolves def by namespaced href                              useAnchoredPopover)
```

- **Package** (`@kolu/solid-markdown`): `rewriteFootnotes` runs pre-sanitize and flags each *forward-ref* anchor with an allowlist-safe `data-md-footnote` — keyed on marked-footnote's bare `data-footnote-ref` (the back-ref carries `data-footnote-backref`, so it's never tagged). `sanitize.ts` keeps the flag; `bindInteractions` grows a footnote branch *before* the generic in-page-scroll branch (a footnote href is also an in-page `#md-footnote-…`) that resolves the definition by the anchor's own namespaced href and fires the new `onFootnote(anchor, definition)`. `markdown.css` gives the marker a pointer cursor + hover tint.
- **Client** (`BrowseFileDispatcher` + new `FootnotePopover`): reuses `useAnchoredPopover` + `<Portal>` (the wikilink menu's recipe) to render a *cleaned clone* of the definition `<li>` — its `id`, every back-ref `↩`, and any nested footnote markers stripped — with the body's own relative/wikilink links routed through the preview's existing resolvers. Dismiss-on-scroll excludes the panel's own scroll; the **keyed** Code-tab subtree handles cleanup on navigation, so there's no detached-anchor bookkeeping.

### What the reader sees

- **Click/tap to open, click-outside · Escape · scroll · re-click to dismiss** — identical on trackpad and phone, no hover step stranded on touch.
- **The bottom list stays as-is**, and a *see all ↓* footer link scrolls to it and closes the popover.
- **Rich bodies work** — a note's links, inline code, and images each open the way they do in the document (the clone carries already-resolved `src`s and routes inner links through the host).
- **Nested footnotes are inert** — a note that cites another footnote renders a plain superscript in the popover (no popover-on-popover).

### Notes for reviewers

> **Test strategy deviates from the plan, deliberately.** The plan proposed a happy-dom `sanitize.test.ts`, but `happy-dom` isn't a dependency anywhere in the repo and `@kolu/solid-markdown`'s `vitest.config.ts` *intentionally* keeps a node-only environment ("the sanitize layer is covered by the browser e2e suite, not here") — the same way `data-md-wikilink`/`data-md-alert` survival is e2e-covered, not unit. So the fragile contract (marked-footnote's markup → our marker) is pinned in a **node render test**, and marker→popover survival rides the e2e. This honors *reuse the existing source of truth* rather than adding a dep + nix-hash churn that contradicts the package's own boundary.

- **No-fallbacks check:** the footnote branch only intercepts when a host handler is wired *and* the definition resolves; otherwise the ref keeps its existing GitHub-style jump-to-definition scroll. The popover is an additive host overlay, not a degraded path.
- **Electricity check:** this is a *leaf* — it hides only bounded marker↔definition logic, no hard volatility — so overlay rendering stays on the client side of the `@kolu/solid-markdown` dependency arrow, exactly where `onNavigateWikilink` already put it.
- **Follow-up (noted, not an oversight):** opening is pointer-only; keyboard activation + a screen-reader `aria-details` relationship are deferred — the bottom footnotes list remains the accessible record.

Plan of record: `docs/atlas/src/content/atlas/md-preview-footnotes.mdx`.

### Try it locally

```sh
nix run github:juspay/kolu/md-fn-opus
```


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `80efd05aa19f`

### Applied (4)
- `lowy-1` Footnote-body link routing duplicates the package's private link-policy receptacle — commit `6b5ab9eb9`
- `lowy-2` Back-ref-stripping selector leaks marked-footnote's id scheme into the host — commit `95f3c28e2`
- `hickey-1` Footnote-body link routing duplicates the package's bindInteractions click seam — commit `6b5ab9eb9`
- `hickey-2` Back-ref removal keys on an href substring, a third recognition mechanism for footnote anchors — commit `95f3c28e2`

### Agreed — no change (3)
- `lowy-3` onFootnote correctly extends the existing anchor-interception receptacle (no new boundary) (packages/solid-markdown/src/Markdown.tsx:56-66)
- `hickey-3` Footnote-marker recognition split across render.ts, sanitize.ts, Markdown.tsx (packages/solid-markdown/src/render.ts:316-321)
- `hickey-4` Dismiss policy split between useAnchoredPopover and a local scroll listener (packages/client/src/right-panel/FootnotePopover.tsx:63-72)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 round(s) · codex reviewed at `xhigh` reasoning effort · base `80efd05aa19f`

### Round 1

**codex** — approved: `false`

Reviewed the diff against 80efd05aa19fd6838234491c162cd09004f6acae and checked status; there are no untracked files. The core shape is reasonable, but the popover reuses document-level link handling in a cloned fragment without covering the cases that stop being true inside that fragment, and there are a few trust/accessibility/test gaps that should be addressed before approval.

Findings:
- `F1` · major · open — The popover binds the generic document link dispatcher to the cloned footnote body, but that clone does not contain the original document's anchor targets. For `href="#..."` links whose target is outside the clone, including nested footnote refs after line 93 removes only `data-md-footnote`, `bindMarkdownLinks` falls through without `preventDefault`, so the browser performs normal hash navigation against the app document. This contradicts the "nested refs are inert" comment and can change the app URL or scroll outside the preview from inside the popover. (packages/client/src/right-panel/FootnotePopover.tsx:127)
- `F2` · major · open — The new footnote branch handles every `click` event on the anchor, including keyboard-generated activation of the existing `<a href="#md-footnote-...">`. The docs call the popover pointer-only and say the bottom footnote list remains the accessible record, but pressing Enter on the focused link now opens an unmanaged portal popover instead of scrolling to that record; focus stays on the marker and there is no `aria-details`, focus movement, or immediate path to the popover content. (packages/solid-markdown/src/Markdown.tsx:111)
- `F3` · minor · open — The clone removes only the root `<li>` id. Any descendant ids inside a rich footnote body, such as headings or raw allowed HTML with `id`, remain duplicated in the portalled clone while the original document still has the same ids. That weakens the sanitizer's id-namespacing guarantee and can confuse in-page anchor resolution. (packages/client/src/right-panel/FootnotePopover.tsx:90)
- `F4` · minor · open — `data-md-footnote` is now an allowed document attribute and the click handler trusts it, but there is no spoof guard comparable to the wikilink policy. Raw inline HTML in a README can set `data-md-footnote` on any same-page anchor and opt into the host footnote-popover callback for arbitrary sanitized content, despite the sanitizer comments saying app-read data hooks should be parser-minted. (packages/solid-markdown/src/sanitize.ts:206)
- `F5` · minor · open — The branch adds a user-facing interactive popover, but the e2e coverage still only checks that footnotes render as a `section` and `sup a`. There is no test that clicking a marker opens the popover, toggles/dismisses it, preserves the "see all" path, or routes links inside the note body, so the main behavior can regress while the added render tests keep passing. (packages/tests/features/code-tab.feature:1203)

**claude** — All five CODEX findings fixed, none disputed. Commit ed38fe2d5ca7095ab717e751213caf7cc6264e1a, 9 files, not pushed. Unit tests 43 of 43 pass; typecheck and biome lint clean; the new e2e scenario 30 steps resolve in cucumber dry-run.

- `F1` **fixed** — The in-page-anchor branch in bindMarkdownLinks now calls preventDefault unconditionally before the lookup, so a hash anchor in the popover clone whose target lives in the bottom list cannot fall through to a real browser hash navigation. Fixed in Markdown.tsx; still only scrolls when the target resolves.
- `F2` **fixed** — The footnote branch is gated on pointer activation (e.detail above 0) so a keyboard Enter or Space falls through to the accessible jump-to-definition scroll. Documented in Markdown.tsx, LIMITATIONS.md, and the Atlas note.
- `F3` **fixed** — The popover clone in FootnotePopover.tsx strips the id from every descendant carrying one, not just the root list item id.
- `F4` **fixed** — rewriteFootnotes in render.ts strips document-authored footnote markers before re-minting them only beside marked-footnote own ref and backref markers; two unit tests added. Narrow residual: a raw anchor also forging the parser ref attribute would still need a real user-minted footnote id.
- `F5` **fixed** — Added a Code-tab e2e scenario and step defs covering open, toggle-close, dismiss, see-all, inner relative-link routing, nested-footnote inertness, and back-ref stripping, including the F1 nested in-page case.

commit: `ed38fe2d5ca7095ab717e751213caf7cc6264e1a`

### Round 2

**codex** — approved: `false`

CLAUDE fixed the five findings from the previous round in the current tree: hash links in cloned popover content no longer fall through to browser navigation, keyboard activation now keeps the accessible bottom-list jump, cloned popover ids are stripped, direct `data-md-footnote` spoofing is covered, and interactive e2e coverage was added. One new regression was introduced by the spoof-guard implementation: the global regex strips ordinary text and can corrupt non-double-quoted attributes.

Findings:
- `F1` · major · resolved — Previously, hash links inside the cloned popover body could miss their target and fall through to browser hash navigation. (packages/solid-markdown/src/Markdown.tsx:147)
- `F2` · major · resolved — Previously, keyboard activation of a footnote marker opened the unmanaged pointer-only popover instead of preserving the accessible jump to the bottom footnote list. (packages/solid-markdown/src/Markdown.tsx:121)
- `F3` · minor · resolved — Previously, the popover clone removed only the root `<li>` id, leaving descendant ids duplicated between the live document and the portalled clone. (packages/client/src/right-panel/FootnotePopover.tsx:95)
- `F4` · minor · resolved — Previously, raw inline HTML could pre-seed `data-md-footnote`/`data-md-footnote-backref` and opt an arbitrary anchor into host footnote handling. (packages/solid-markdown/src/render.ts:334)
- `F5` · minor · resolved — Previously, the new interactive popover behavior lacked e2e coverage beyond static footnote rendering. (packages/tests/features/code-tab.feature:1255)
- `F6` · major · open — The new spoof-guard regex runs over the entire rendered HTML string, not just attributes inside tags. It will delete ordinary document text like `<p>mention data-md-footnote here</p>` into `<p>mention here</p>`, and it corrupts non-double-quoted valued attributes such as `<a data-md-footnote=bogus href=#x>` into `<a=bogus href=#x>`. That is a new content-corruption regression introduced by this round's fix. (packages/solid-markdown/src/render.ts:334)

_codex on the rebuttal:_ CLAUDE did not dispute any previous finding. I verified F1 through F5 as fixed in the current tree. The only remaining objection is new: the F4 spoof-guard fix introduced F6 by applying a broad regex to all rendered HTML text.

**claude** — Round 2 of the codex debate. Codex confirmed round-1 findings F1–F5 as resolved and raised one new finding, F6: the spoof-guard regex I added in round 1 ran over the entire rendered HTML string, so it deleted the literal `data-md-footnote` token from ordinary prose and mangled non-double-quoted spoof attributes (`<a data-md-footnote=bogus …>` collapsed to `<a=bogus …>`). I reproduced both regressions with a quick Node harness, agreed they were real, and fixed them by scoping the strip to anchor start tags only. The strip now matches each `<a …>` open tag and removes `data-md-footnote*` from its attribute span in any value form — bare, double-quoted, single-quoted, and unquoted — never touching text nodes. Added two regression tests (plain prose with the literal token survives; single-/unquoted spoof attributes are consumed whole without collapsing the anchor or dropping its href). render.test.ts passes 44/44, the full solid-markdown suite passes 59/59, and biome is clean. The fix stays within the existing regex-over-HTML-string pipeline (rewriteAlerts and the sibling re-mint regexes already rely on the same `[^>]*` open-tag assumption), so it's consistent rather than a structural departure.

- `F1` **fixed** — Resolved in round 1 and confirmed by codex this round (status resolved in verdict-2). No further action. The in-page-anchor branch in bindMarkdownLinks calls preventDefault before lookup and only scrolls when a clone-local landing exists, so a hash link in the popover clone whose target lives in the bottom list goes inert instead of navigating the app document (packages/solid-markdown/src/Markdown.tsx).
- `F2` **fixed** — Resolved in round 1 and confirmed by codex this round. No further action. The popover branch is gated on e.detail > 0 (pointer activation), so keyboard Enter/Space falls through to the accessible jump-to-definition scroll (packages/solid-markdown/src/Markdown.tsx).
- `F3` **fixed** — Resolved in round 1 and confirmed by codex this round. No further action. The popover clone strips the id from every descendant carrying one, not just the root li (packages/client/src/right-panel/FootnotePopover.tsx).
- `F4` **fixed** — The direct-spoof intent (strip document-authored data-md-footnote* before re-minting) was resolved in round 1 and confirmed by codex this round. The implementation regression it introduced is tracked and fixed under F6 (packages/solid-markdown/src/render.ts).
- `F5` **fixed** — Resolved in round 1 and confirmed by codex this round. No further action. The e2e scenario covers open, toggle-close, see-all close, relative-link routing from the popover, nested-marker de-flagging, and back-ref stripping (packages/tests/features/code-tab.feature).
- `F6` **fixed** — Agreed — valid regression I introduced in round 1. Reproduced both cases: the global regex turned `<p>mention data-md-footnote here</p>` into `<p>mention here</p>` and turned `<a data-md-footnote=bogus href=#x>` into `<a=bogus href=#x>`. Fix at packages/solid-markdown/src/render.ts:328 (rewriteFootnotes): replaced the whole-string strip with `.replace(/<a\b[^>]*>/g, (tag) => tag.replace(DOC_FOOTNOTE_ATTR, ""))`, scoping removal to anchor start tags. The new DOC_FOOTNOTE_ATTR regex `/\s+data-md-footnote(?:-backref)?(?:\s*=\s*(?:"[^"]*"|'[^']*'|[^\s>]*))?/g` handles bare, double-quoted, single-quoted, and unquoted value forms, so an unquoted spoof value is consumed whole instead of leaving `<a=…`. Text nodes are never visited, so prose mentioning the token is preserved. Added two regression tests in render.test.ts: plain-text survival, and single-/unquoted spoof attributes stripped without anchor collapse (asserts no `<a=` and hrefs intact). 44/44 render tests and 59/59 package tests pass; biome clean.

commit: `0d9c728989d1877e80818ac717309c5a969d42c4`

### Round 3

**codex** — approved: `true`

I re-inspected the current tree against the existing findings only. `git status --short` is clean, and the F6 fix now scopes stripping to anchor start tags while covering bare, quoted, single-quoted, and unquoted spoof attributes; the added tests cover both regressions I raised. All findings from this review session are resolved.

Findings:
- `F1` · major · resolved — Hash links inside the cloned popover body could miss their target and fall through to browser hash navigation. (packages/solid-markdown/src/Markdown.tsx:147)
- `F2` · major · resolved — Keyboard activation of a footnote marker opened the unmanaged pointer-only popover instead of preserving the accessible jump to the bottom footnote list. (packages/solid-markdown/src/Markdown.tsx:121)
- `F3` · minor · resolved — The popover clone removed only the root `<li>` id, leaving descendant ids duplicated between the live document and portalled clone. (packages/client/src/right-panel/FootnotePopover.tsx:95)
- `F4` · minor · resolved — Raw inline HTML could pre-seed `data-md-footnote`/`data-md-footnote-backref` and opt an arbitrary anchor into host footnote handling. (packages/solid-markdown/src/render.ts:344)
- `F5` · minor · resolved — The new interactive popover behavior lacked e2e coverage beyond static footnote rendering. (packages/tests/features/code-tab.feature:1255)
- `F6` · major · resolved — The round-1 spoof-guard regex ran across the full HTML string, deleting ordinary prose containing `data-md-footnote` and corrupting unquoted spoof attributes into malformed anchors. (packages/solid-markdown/src/render.ts:334)

_codex on the rebuttal:_ No findings were disputed. CLAUDE accepted F6, reproduced the two regressions, and replaced the whole-string regex with an anchor-tag-scoped attribute strip. I verified the current implementation and regression tests, so F6 is resolved; F1-F5 remain resolved from the prior round.

srid:

| Pass | Issues found | Details |
|------|-------------|---------|
| Rules | 0 | Clean — all built-in + project rules checked |
| Fact-check | 1 | Footnote branch missing `href.length > 1` guard — fixed |
| Elegance | – | Skipped (`--no-elegance`; `/simplify` ran this gauntlet) |

**One violation found and fixed** — `fix(police): guard footnote href with length > 1` (commit `a7a1d868f`).

The footnote click branch in `Markdown.tsx` checked `href?.startsWith("#")` but not `href.length > 1`. A bare `href="#"` on a `data-md-footnote`-flagged anchor would make `href.slice(1) === ""`, then `CSS.escape("")` → `querySelector("#")` throws a `SyntaxError` — *before* the `if (definition)` guard can catch a null. The sibling in-page-scroll branch immediately below already carried the `href.length > 1` guard; the fix applies the same condition to the footnote branch, closing the `SyntaxError` vector.

Rules-pass confirmations: `no-silent-error-swallowing` (clipboard/Shiki/stream errors all surface), `no-re-export-bridge-modules` (`index.ts` exports its own symbols), `styling-tailwind-only` (`markdown.css` is only color-mix theming Tailwind can't express), `e2e-poll-async-state` (new step defs use `waitForFunction`/`.waitFor` with `POLL_TIMEOUT`).

srid:

The **x86_64-linux** lane ran on **`kolu-ci-3`** (`dev-x86-64-linux-03`) — commit `a7a1d868f`, **passed** ✓

- **Lane wall** (pipeline): **3m36s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 2m52s | ✓ |
| `ci::home-manager` | 1m17s | ✓ |
| `ci::unit` | 49s | ✓ |
| `ci::smoke` | 46s | ✓ |
| `ci::nix` | 46s | ✓ |
| `_ci-setup` | 35s | ✓ |
| `ci::pnpm-hash-fresh` | 33s | ✓ |
| `ci::surface-example-build` | 32s | ✓ |
| `ci::surface-app-example-build` | 32s | ✓ |
| `ci::render` | 31s | ✓ |
| `ci::biome` | 31s | ✓ |
| `ci::fmt` | 30s | ✓ |
| `ci::install` | 8s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/a7a1d86/timings.jsonl`; pool state is a live `flock` probe.</sub>

srid:
## Evidence

**Scenario:** `Markdown preview opens a footnote definition in a click popover` (`features/code-tab.feature`), recorded by the e2e harness under `KOLU_EVIDENCE=1` on a clean pu box building this branch's own commit.

The clip shows the footnote-popover interaction in the Code-tab Markdown preview:

- **Click a `[^N]` marker** in the rendered Markdown → a popover opens pinned to the marker with the footnote definition body inline.
- **Dismiss** by toggling the marker again / the see-all (back-to-references) affordance closes it.
- **A relative link inside the popover body navigates in-app** — it routes to the linked file in the Code tab (you can see the preview switch to the target doc, "Footnote target reached"), rather than leaving the app.

The gesture is **click/tap** throughout — there is no hover step — so this covers touch by construction.



▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=md-footnote-popover.mp4

PR #1518, opened 2026-06-22
chore(skills): /be plan must ground low-level facts against installed code

srid:
## What

One sharpening to the `/be` §1 plan-authoring clause: a plan of record must **ground every load-bearing low-level fact against the installed code before asserting it** — staying high-level does not license guessing.

Mined by `/self-improve` from the footnote-popover `/be` run (session `fc13bb53…`, → PR #1514).

## Why — the friction this engineers out

The run was otherwise fully autonomous (one human turn: `/be @…md-preview-footnotes.mdx`, after `/effort ultracode` — no corrections, no "continue", no "you forgot X"). The one recurring friction was at code-time: the plan-of-record asserted three low-level facts that were never verified against the installed code, and the implementation had to reconcile each:

| Plan asserted | Ground truth | Source of truth it should have read |
|---|---|---|
| `marked-footnote@^1.2.4` | resolves to **1.4.0** | `pnpm-lock.yaml` (not the `^range`) |
| detection keys on `class="footnote-ref"` | real marker is bare **`data-footnote-ref`** | the library's actual emitted DOM |
| land a **happy-dom** sanitize unit test | package keeps a deliberate **node-only** vitest env; DOM/sanitize is e2e-covered; happy-dom isn't even a devDep | `packages/solid-markdown/vitest.config.ts` |

The existing §1 clause forbids an "implementation dump" — but purely on **altitude** grounds (too detailed). It never required the few low-level facts a plan *does* lean on to be true. So a plan could read as clean and high-level and still be flatly wrong, forcing an implementation-time reconciliation and (since the note publishes to kolu.dev) shipping a false published note until corrected.

## Edit ledger

| # | Source edit | Evidence |
|---|---|---|
| 1 | `.apm/skills/be/SKILL.md` §1 — add self-check **(c)**: ground every load-bearing low-level fact (pinned dep version → read the lockfile; library markup/attribute; test-env strategy) against installed code, the way §2 gets ground truth; add `facts-grounded ✓` to the self-check tick-list | The three table rows above are the worked example baked verbatim into the clause |
| 2 | `.claude/skills/be/SKILL.md`, `.agents/skills/be/SKILL.md`, `apm.lock.yaml` — regenerated from the source via `just ai::apm` so the runtimes ride the same commit | `just check` green (typecheck + biome). The regen also reconciled a pre-existing stale `dev-server` lock hash already on `master` — not part of this change's intent, but the lock must match the deployed file |

## Scope discipline

Lowest-churn fix: **one clause sharpened**, no new rule or skill. Honors reuse-the-source-of-truth (points the author at the same ground-truth mechanism §2 already uses — lockfile / `vitest.config.ts` / emitted DOM) and trips no llm-autonomy anti-pattern (no reintroduced post-§0 question; the §1 self-check is pre-pause, already part of the one sanctioned plan-review pause). Not merged — for human review.


srid:
@copilot resolve the merge conflicts in this pull request

Issue #72, opened 2026-03-23
Bug: Claude Code's TUI rendering is often off


srid:
... even on full refresh of Kolu.

Example:

<img width="961" height="531" alt="Image" src="https://github.com/user-attachments/assets/f8238980-d540-486c-a1ce-3e4a540a6814" />

PR #1522, opened 2026-06-23
docs(atlas): correct R8b's consumer recipe from Step 0 (drishti #72) — fine-grained, not coarse-copy

srid:
Corrects the `pulam-web` plan + `remote-terminals` R8b with what **Step 0 actually taught** — drishti [#72](https://github.com/srid/drishti/pull/72) graduated the reactive stream consumer, and the review gauntlet caught a reproduced **HIGH regression** that falsified a premise in my plan. Docs only.

## What Step 0 found

1. **A real bug (reproduced):** feeding `.streams.use()`'s `reconcile` store into a `createEffect` that **coarse-reads `snapshot()` and copies** into a second store drops **same-shape** delta frames (a leaf ticking in place — the steady state): `snap 10 → d1 20 → d2 *20*`. Same root cause #1510 hit (reading the whole pulse object, not the nested `seq`).
2. **A wrong premise:** I labelled `processesSnapshot` "value-bearing." It's **delta-accumulate** — each frame is a *change*, so `.streams.use()` (latest-frame-only) can't render it directly; you must fold. R8b's `gitStatus` genuinely *is* value-bearing, which is the cleaner case.
3. **A vacuous test:** #72's hermetic test passed only because every frame changed array shape — it never drove a same-shape delta, the case that breaks.

## The correction (now pinned as R8b's recipe)

**A value-bearing — or folded-to-value-bearing — `reconcile` subscription, rendered fine-grained (`sub().modified`), never coarse-read-and-copy.** The hermetic test must drive a **same-shape delta**.

- `gitStatus` (Design B, value-bearing) → `.streams.use()` + fine-grained reads.
- delta-accumulate streams (like `processesSnapshot`) → `createSubscription` + `reduce` (fold) + fine-grained (`.streams.use()` has no `reduce`).

Updated `pulam-web.mdx` (Step 0 + the trap callout + R4.8b + the hermetic test + DoD + history) and the `remote-terminals` R8b parenthetical — the fork it called "unresolved" is now **resolved**.

Notably the regression would have shipped **green** (drishti CI is typecheck + nix, no test lane) — only the gauntlet caught it, which is exactly why Step 0 runs *before* pulam-web/R8b. The drishti-side fix is tracked in the [#72 thread](https://github.com/srid/drishti/pull/72#issuecomment-4774727767).


srid:
Superseding by #1523, which carries the same correction (delta-accumulate premise + the fine-grained / same-shape-delta recipe across pulam-web.mdx and remote-terminals R8b) plus marks Step 0 as **shipped** rather than just "ran." Closing this to avoid a three-way edit conflict on `pulam-web.mdx`; the content lives on in #1523.

PR #1535, opened 2026-06-23
pulam-web agent dashboard — sorted by what needs you

srid:
**pulam-web grows from a flat terminal list into an agent dashboard** — every agent across every configured host, **sorted by what needs you**. This is R-pulamweb-3 from the [pulam-web plan](https://github.com/juspay/kolu/blob/pulamweb-3/docs/atlas/src/content/atlas/pulam-web.mdx): a render / sort / filter pass over R-pulamweb-2's proven framework. **No surface change, no `@pierre/trees`, nothing gated** — it reads two already-proven consumers (the `awareness` collection for state, the `activity` stream for the live dot) and projects them.

### How it fits together

```
awareness collection ─▶ entries (settled, per-host) ─┐
                                                     ├─▶ fleet.ts: bucket · needs-you sort · filter ─▶ rows
activity stream ─────▶ live set (full set / frame) ──┘                                                  └─▶ green dot
```

`fleet.ts` is the pure projection — `agentBucket` · `agentUrgency` · the needs-you-first comparator · `relativeTime` · terminal categories · web tone colours — **ported from `pulam-tui`'s renderer-agnostic core** (the plan's reuse map marks it "ports to web") and pinned to the TUI's behaviour by `fleet.test.ts`. It's owned in pulam-web, not imported: pulam-web has no `@kolu/pulam-tui` dependency and a TUI/OpenTUI package has no place in the Vite browser bundle.

### What you get

- **Sorted by what needs you** — a blocked agent (`awaiting_user`) floats to the top with a breathing alert strip; working agents spin cyan; idle ones sit dim. Per row: agent · `repo · branch` · state · how long it's been there.
- **Live activity dot** — a green dot when the terminal is moving bytes right now (the fleet echo of kolu's Dock dot), read from the `activity` stream.
- **Agents, not terminals** — only active agents show by default; footer toggles fold in idle agents, non-agent terminals, and sleeping shells (each a disjoint, grounded category).

### Two plan premises corrected against the installed code

Grounding R-pulamweb-3 against the real surface corrected two load-bearing facts the note carried:

| The note said | Ground truth | Consequence |
| --- | --- | --- |
| dirty/clean mark from `awareness.git` | `awareness.git` carries only `repoName`/`branch` — file counts need the `git.getStatus` procedure | the dirty/clean count moves to **R-pulamweb-4** (where the reuse map already places `RepoWatchSet`); this phase shows `repo · branch` |
| consume `activity` via the R-pulamweb-1 consumer | `activity` is **value-bearing** (full set each frame), not delta-accumulate | use `.streams.use()` (replace-each-frame), **not** `createSubscription` + reduce |

### Design philosophy

- **Fail-fast, no fallbacks** — `agentBucket` surfaces an unknown agent state verbatim (`other`) rather than miscolouring it; the per-host error funnel is preserved (a subscription failure surfaces, never collapses the list to empty); a `lastActivityAt` of `0` renders as "never" (—), never faked to "now".
- **Electricity** — the projection is *agent-dashboard* leaf logic, not a hard volatility, so it's a pulam-web-local module, not a new `@kolu/*` package; no surface/contract change.
- **Reuse the source of truth** — `activity` is value-bearing, so the canonical `.streams.use()` consumer (introducing a hand-rolled reduce would be a parallel mechanism); the bucket/sort logic mirrors pulam-tui's canonical algorithm rather than inventing a second state machine.

### Tests

- `fleet.test.ts` — pins bucketing, the needs-you-first sort (urgency beats recency; recency then id tiebreak), `relativeTime`, categories, and the filter predicate to pulam-tui's behaviour.
- `reserve.test.ts` (new block) — the value-bearing analog of R-pulamweb-1's same-shape proof: two same-cardinality `activity` swaps each re-notify a **fine-grained** dot reader over the full agent → mirror → re-serve → browser leg. (Necessary-but-not-sufficient; the live-ws video is the full proof — coming in Evidence.)

### Try it locally

```sh
PULAM_WEB_HOSTS=localhost nix run github:juspay/kolu/pulamweb-3#pulam-web
# → open http://localhost:4800
```


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `ebfef4f05609`

### Applied (7)
- `lowy-1` Agent-state projection duplicated across a package boundary (pulam-tui ⇄ pulam-web) — commit `51947e657`
- `lowy-2` Agent-state bucketing is now a fourth parallel encapsulation of the same volatility — (uncommitted)
- `lowy-5` Hard-coded hex colours re-spell the palette the TUI already centralised — (uncommitted)
- `hickey-1` Agent-state projection duplicated across pulam-tui and pulam-web — (uncommitted)
- `hickey-2` URGENCY colour/label/glyph literals re-spelled outside the table — commit `ef01d10a2`
- `hickey-4` stateLabel re-derives the urgency→label mapping instead of routing through URGENCY uniformly — (uncommitted)
- `hickey-5` DASH em-dash constant duplicated across modules — commit `bc4981e56`

### Agreed — no change (3)
- `lowy-3` Transport/provisioning volatility is correctly encapsulated (electricity, clean) (packages/pulam-web/src/client/wire.ts:30-98)
- `lowy-4` App↔HostGroup count fan-out keyed by mutable roster can retain stale slices (packages/pulam-web/src/client/App.tsx:135-147)
- `hickey-3` Per-host counts fragmented into a callback-up + store-down cycle (packages/pulam-web/src/client/App.tsx:133-147,183-193 and packages/pulam-web/src/client/HostGroup.tsx:71-73,179-193)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `ebfef4f05609`

### Round 1

**codex** — approved: `false`

The branch is directionally solid, but I would not approve it yet. The main problems are in the new live dashboard behavior: activity can be falsely shown as live, pending terminal values are collapsed into an empty state, and a few UI/documentation details are inconsistent with the repo's own rules and the implemented feature set.

Findings:
- `F1` · major · open — The new dashboard treats `app.streams.activity.use()` as the authoritative live set, but the re-serve currently yields all awareness keys as the activity stream's initial snapshot (`reserve.ts:157`) and the local activity bus does not replay the last real remote frame. A newly-opened dashboard can therefore paint every terminal as moving bytes until the next activity frame, or indefinitely if no further output occurs. (packages/pulam-web/src/client/HostGroup.tsx:160)
- `F2` · major · open — `entries()` drops every key whose per-key awareness subscription is still pending, and the render then uses `entries().length === 0` to show `no terminals`. That collapses "terminal keys arrived but values are still loading" into a legitimate empty host, regressing the previous per-row pending placeholder behavior. (packages/pulam-web/src/client/HostGroup.tsx:171)
- `F3` · minor · open — The agent-name column is fixed at `w-[7ch]`, but `agentShortName("opencode")` returns `opencode` (8 chars), and foreground process names can be longer. The text can overflow its fixed flex cell and collide with the location column. (packages/pulam-web/src/client/HostGroup.tsx:127)
- `F4` · minor · open — The footer is a single non-wrapping flex row containing three filter labels plus two counters. On narrow viewports, `non-agent terminals` and the counters can overflow the page. (packages/pulam-web/src/client/App.tsx:193)
- `F5` · minor · open — This adds custom CSS/keyframes for pulse, spin, and blink despite the repo rule `styling-tailwind-only` requiring Tailwind utilities unless custom CSS is truly impossible. These animations are expressible with Tailwind's built-in or arbitrary animation utilities plus `motion-reduce`. (packages/pulam-web/src/client/index.css:8)
- `F6` · minor · open — The Atlas note still says pulam-web proves "value-bearing git status, over `.streams.use()`, sourced from a mirror," and later describes R-pulamweb-3 as layering live git status/drill-in. This branch explicitly defers git status and drill-in to R-pulamweb-4; R-pulamweb-3 only consumes the value-bearing `activity` stream. (docs/atlas/src/content/atlas/pulam-web.mdx:136)
- `F7` · nit · open — `FilterChip` behaves as a toggle but does not expose toggle state semantically. The visual `✓`/`+` text helps sighted users, but assistive tech should get the stable pressed state too. (packages/pulam-web/src/client/App.tsx:106)

**claude** — Reviewed CODEX's verdict against the R-pulamweb-3 agent dashboard diff. I agreed with all seven findings — each is a genuine defect grounded in the code — and fixed every one. Two are real correctness bugs (F1 false-live snapshot, F2 pending-keys collapsed into empty), three are UI robustness/rule-compliance fixes (F3 name overflow, F4 footer wrap, F5 Tailwind-only animations), one is stale doc prose (F6), one a11y nit (F7). The static-check gate is green (typecheck across all packages + biome clean), all 34 pulam-web tests pass including a new F1 regression test that I confirmed FAILS against the old snapshot and passes with the fix, the Atlas dist was rebuilt (idempotent, host-independent), and `just fmt` produced no reformats. Committed as da450f443cb0ec540ce9c365d0c6c50f45b844b1. No `@kolu/surface*` exported API changed, so the drishti gate is not triggered this round.

Key files: /home/srid/code/kolu/.worktrees/pulamweb-3/packages/pulam-web/src/server/reserve.ts (F1 fix — cache `activityLatest`, init `[]`, yield it as the snapshot instead of `[...awarenessCache.keys()]`), /home/srid/code/kolu/.worktrees/pulamweb-3/packages/pulam-web/src/client/HostGroup.tsx (F2 empty/loading split, F3 name cell, F5 animations), /home/srid/code/kolu/.worktrees/pulamweb-3/packages/pulam-web/src/client/App.tsx (F4 footer, F5 strip, F7 aria-pressed), /home/srid/code/kolu/.worktrees/pulamweb-3/packages/pulam-web/src/client/index.css (F5 — keyframes removed), /home/srid/code/kolu/.worktrees/pulamweb-3/packages/pulam-web/src/server/reserve.test.ts (F1 regression test), /home/srid/code/kolu/.worktrees/pulamweb-3/docs/atlas/src/content/atlas/pulam-web.mdx (F6).

- `F1` **fixed** — Confirmed and fixed. reserve.ts:164 yielded `[...awarenessCache.keys()]` as the activity stream's subscribe snapshot. The daemon's real activity source (daemon.ts:182, activity.snapshot()) returns only byte-moving terminals — empty for a quiet host — but the re-serve fabricated a snapshot of ALL awareness keys, so a freshly-opened dashboard painted every terminal's green dot live until the next activity frame (or forever if quiet). Fix: added `let activityLatest: TerminalId[] = []` cached in buildReServe; the sink's `onFrame` now sets `activityLatest = frame` BEFORE publishing; the browser-facing source yields `[...activityLatest]`. Added a regression test `does NOT paint a quiet terminal live before an activity frame (F1)` that folds TERM_A into the re-serve cache, then opens a fresh activity subscription and asserts the snapshot frame is `[]` and the dot reader stays false until a real `[TERM_A]` frame lands. I verified the test FAILS against the old snapshot (expected true to be false) and passes with the fix.
- `F2` **fixed** — Confirmed and fixed. HostGroup.tsx `entries()` skips keys whose per-key subscription is still pending (valueForId returns undefined while pending), and the render gated the empty state on `entries().length === 0` and the header count on `entries().length` — so a host whose terminal keys arrived but values are still loading rendered as a legitimately-empty 'no terminals'. The R-pulamweb-2 version (verified via git show) iterated `awareness.keys()` directly and rendered a per-row pending placeholder, so this was a regression. Fix: gate 'no terminals' on `awareness.keys().length === 0` (the true empty), nest a distinct `loading terminal details…` state for `entries().length === 0` while keys exist, and show the header count off `awareness.keys().length`.
- `F3` **fixed** — Confirmed and fixed. The agent `kind` literals are `claude-code`→`claude` (6), `opencode` (8, integrations/opencode/src/schemas.ts:18), `codex` (5). `agentShortName("opencode")` returns the 8-char `opencode`, overflowing the fixed `w-[7ch] flex-none` cell, and the non-agent branch returns `value.foreground.name` which is unbounded. Fix: widened the cell to `w-[9ch]` and added `overflow-hidden text-ellipsis whitespace-nowrap` plus `title={name()}` so longer values truncate cleanly with the full value on hover rather than colliding with the location column.
- `F4` **fixed** — Confirmed and fixed. The footer was a single non-wrapping `flex items-center gap-2` row holding three filter chips ('idle', 'non-agent terminals', 'sleeping') plus two counters with `ml-auto`, which overflows narrow viewports. Fix: added `flex-wrap` and switched to `gap-x-2 gap-y-1` so the row wraps onto multiple lines on narrow screens; `ml-auto` still right-aligns the first counter when there is room and simply wraps when there isn't.
- `F5` **fixed** — Agreed and fixed. The `.agency/code-police.md` styling-tailwind-only rule allows custom CSS only when 'truly impossible with Tailwind', and all three animations are expressible with Tailwind built-ins: pw-pulse (opacity 1↔.5 ease-in-out) → `animate-pulse`; pw-spin (rotate linear, origin 50% 54%) → `inline-block origin-[50%_54%] animate-spin`; pw-blink (live dot) → `animate-pulse`. Replaced all three render-site classes with the Tailwind utilities plus `motion-reduce:animate-none` (in HostGroup.tsx glyph + dot and App.tsx needs-you strip) and stripped index.css back to just the Tailwind `@import` with a short note. (The codebase's main client/src/index.css does keep hand-rolled keyframes, but those carry explicit per-block 'impossible with Tailwind' justifications — compositor-accelerated masked rings, CSS-var-interpolated color-mix — that don't apply to these three plain opacity/rotate animations, so the rule's exception isn't earned here.)
- `F6` **fixed** — Confirmed and fixed. pulam-web.mdx:136 claimed pulam-web 'is the first standalone app to put those together — value-bearing git status, over .streams.use(), sourced from a mirror', and the implementation-intro at :155 said 'R-pulamweb-3 layers the user-facing features (live git status, drill-in)' and 'R-compose-2 gates on R-pulamweb-3'. The rest of the note (the Phase tree at :186/:200, the detailed section, the reuse map, the history entry) correctly defers git status + drill-in to R-pulamweb-4 and has R-pulamweb-4 block R-compose-2. Fixed the three stale spots: :136 now says R-pulamweb-3 proves the value-bearing leg with the `activity` stream and git status is R-pulamweb-4's; :155 now describes R-pulamweb-3 as the activity-stream dashboard, R-pulamweb-4 as git status/drill-in, and 'R-compose-2 gates on R-pulamweb-4' (matching the phase tree); and :167's 'R-pulamweb-3's git-status consumer rides on' → 'R-pulamweb-3's activity consumer (and R-pulamweb-4's git-status consumer)'. Regenerated docs/atlas/dist via `just atlas::build` (only pulam-web.html changed; idempotent on rebuild).
- `F7` **fixed** — Agreed and fixed. Added `aria-pressed={props.active}` to the FilterChip `<button>` in App.tsx so assistive tech gets the stable toggle state in addition to the visual ✓/+ glyph.

commit: `da450f443cb0ec540ce9c365d0c6c50f45b844b1`

### Round 2

**codex** — approved: `true`

CLAUDE addressed all seven findings from the prior round. I re-checked the current diff and the fixes match the requested behavior: the activity stream now snapshots the cached live set instead of awareness keys, pending terminal details no longer render as an empty host, the fragile UI wrapping/overflow/a11y issues are corrected, the custom animations were replaced with Tailwind utilities, and the stale Atlas wording now matches the R3/R4 split. I do not see a new regression introduced by this round's fixes.

Findings:
- `F1` · major · resolved — The re-serve activity source previously fabricated its initial snapshot from all awareness keys, falsely marking quiet terminals live for fresh dashboard subscribers. (packages/pulam-web/src/server/reserve.ts)
- `F2` · major · resolved — Pending per-terminal awareness values were collapsed into the same UI state as a genuinely empty host because rendering depended on `entries().length`. (packages/pulam-web/src/client/HostGroup.tsx)
- `F3` · minor · resolved — The fixed-width agent/name cell could overflow for `opencode` or longer foreground names and collide with adjacent columns. (packages/pulam-web/src/client/HostGroup.tsx)
- `F4` · minor · resolved — The footer filter/counter row could overflow narrow viewports because it was a single non-wrapping flex row. (packages/pulam-web/src/client/App.tsx)
- `F5` · minor · resolved — The pulam-web client added hand-rolled keyframe animations that were expressible with Tailwind utilities and did not earn the local custom-CSS exception. (packages/pulam-web/src/client/index.css)
- `F6` · minor · resolved — The Atlas note still described R-pulamweb-3 as delivering git status/drill-in behavior that the plan assigns to R-pulamweb-4. (docs/atlas/src/content/atlas/pulam-web.mdx)
- `F7` · nit · resolved — Filter toggle buttons exposed visual state but lacked `aria-pressed`, so assistive tech could not read the stable toggle state. (packages/pulam-web/src/client/App.tsx)

_codex on the rebuttal:_ No disputes to answer. CLAUDE accepted all seven prior findings, and I verified the current tree resolves each one without introducing a new issue in this round.

srid:

Ran the rule checklist + fact-check passes over `git diff ebfef4f0…HEAD` (the merge-base). Elegance pass skipped (`--no-elegance` — the `/simplify` track already ran over this tree).

| Pass | Result |
| --- | --- |
| Rules | 1 finding → fixed |
| Fact-check | clean |

**`no-thin-wrapper-functions` (2×) → fixed in `40ae359c1`.** `stateLabel` (`fleet.ts`) and `fleetStateText` (`render.ts`) were each a single-line pass-through to the shared `fleetStateLabel`, with one caller apiece. Inlined `fleetStateLabel(agent, URGENCY_LABELS)` at both render sites (HostGroup reads the now-exported `URGENCY_LABELS`; `render.ts`'s `fleetRow` reads its module-local one); the `fleet.test.ts` web-label assertions now call `fleetStateLabel` directly.

**Fact-check — clean.** No silent error swallowing, no inaccurate fallback (`activityLatest = []` before first byte, `counts[h]?.need ?? 0` for a missing host, `?? DASH` for no foreground — all correct), no wishful thinking (the `<Show when={value()}>` guards the transient `undefined`; `sameIds` is sound given the deterministic `compareFleetEntries`), the `agentUrgency` switch is exhaustive over all `agentBucket` arms, and `createNow` tears down its `visibilitychange` listener + interval in `onCleanup`.

srid:

The **x86_64-linux** lane ran on **`kolu-ci-8`** (`dev-x86-64-linux-08`) — commit `62a640c7d`, **passed** ✓

- **Lane wall** (pipeline): **3m52s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 3m11s | ✓ |
| `ci::home-manager` | 1m28s | ✓ |
| `ci::unit` | 1m0s | ✓ |
| `ci::smoke` | 54s | ✓ |
| `ci::nix` | 52s | ✓ |
| `ci::pnpm-hash-fresh` | 41s | ✓ |
| `ci::surface-example-build` | 39s | ✓ |
| `ci::surface-app-example-build` | 39s | ✓ |
| `ci::render` | 38s | ✓ |
| `ci::biome` | 38s | ✓ |
| `ci::fmt` | 37s | ✓ |
| `_ci-setup` | 29s | ✓ |
| `ci::install` | 10s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/62a640c/timings.jsonl`; pool state is a live `flock` probe.</sub>

srid:
## Evidence

The new agent dashboard rendering **live over the real per-host WebSockets**, captured from a browser against a from-source build of this branch (`62a640c7`) on an ephemeral pu box (desktop + a phone viewport, per kolu's mobile-coverage rule).

**Desktop**
desktop

**Mobile (390×844)**
mobile

What's shown, live:
- The `every agent, every host` shell with one bucketed group **per host** (violet `▌` accent): `localhost · 5 terminals` and `srid@sincereintent · 0 terminals` → `no terminals`.
- **Real agent detection** — an `opencode` row resolved to its git location (`kolu-src · HEAD`), distinct from the plain `bash` shells (non-agent) and sleeping terminals.
- The **green activity dot** lit and *breathing* on a terminal moving bytes right now (verified via computed style: `rgb(126,231,135)` = `#7ee787` with the pulse animation on), dim `#262b38` dots on the quiet rows — the per-row liveness straight off the `activity` stream.
- The **filter footer** working: `showing agents` + the three toggles `✓ idle`, `✓ non-agent terminals`, `✓ sleeping` (toggled on here to reveal the hidden rows), plus the fleet counters `● 0 needs you` / `◜ 0 working`. With the toggles off, the host correctly reads `no active agents · N hidden`.
- The layout reflows cleanly to a phone viewport.

**Limitation (honest):** the `working` (cyan spinner) and `awaiting_user` (amber breathing "needs you" strip) urgency states are **not** shown on-screen — reaching them needs an agent actually talking to a model, and the off-machine box has no model-API credentials (no fabricated session state, no borrowed keys). Those `need`/`work`/`idle` → colour/sort/glyph paths are pinned instead by `packages/pulam-web/src/client/fleet.test.ts` (needs-you floats above a more-recent working row; the bucket/label/category assertions), and the **value-bearing `activity` re-notify** is pinned by the hermetic `reserve.test.ts` proof over the full agent → mirror → re-serve → browser leg.

PR #1545, opened 2026-06-23
feat(surface): default-on connection liveness at the canonical seams

### LINKED ISSUE #1542: Make surface broken-connection detection default-on at the connect/accept seams (fix pulam-web + drishti per-host blind spots)
## Problem

Broken-WebSocket / broken-ssh detection in the surface stack is an **opt-in primitive that apps can silently forget**, and two apps have forgotten it on exactly the sockets that matter.

An audit across kolu, pulam-web, drishti, and the shared `@kolu/surface*` packages found three connection legs:

| Leg | kolu | pulam-web | drishti |
|---|---|---|---|
| Server reaps dead browser (`startWsHeartbeat`) | ✓ | ✓ | ✓ |
| **Browser detects dead server** (client watchdog `createHeartbeat`) | ✓ | ✗ | partial |
| Backend detects dead remote host over ssh (`getHostSession`/`pumpRemoteSurface`) | ✓ | ✓ | ✓ |

The failure mode is a **silently half-open socket** — laptop sleep, Wi-Fi roam, NAT idle drop — where TCP dies with no FIN/RST, so the socket stays `OPEN` forever and live terminal streams freeze until a manual reload.

- **pulam-web** builds each per-host socket by hand in `packages/pulam-web/src/client/wire.ts` (`createSurfaceSocket` → `surfaceClient(... websocketLink(ws))`), wiring the stale-tab handshake but **no probe / no `createHeartbeat`**. Never imports `SurfaceAppProvider`. So *no* browser-side liveness on the per-host fleet sockets — the ones carrying terminal streams.
- **drishti** gets the watchdog only on its single admin/control-plane socket (it passes `{ ws, probe }` into `SurfaceAppProvider`, tripping the turnkey branch at `packages/surface-app/src/solid/index.ts:544`). Its **per-host** sockets (`wire.ts` `makeSocket`) have the same blind spot as pulam-web's.

### Root cause

`createHeartbeat` is wired **only** inside the `else if (props.ws && props.probe)` turnkey branch of `SurfaceAppProvider`. It is **not** baked into the canonical socket/client factory, and it requires a **contract-specific `probe`** (kolu `identity.info`, kaval `system.heartbeat`, pulam first `version` frame). That per-app probe is *why* the watchdog is opt-in and forgettable. The deeper defect is a safety primitive that **degrades silently when forgotten** — the "fail loud, don't degrade" rule turned inside out.

### Secondary gaps the audit flagged

- **Server reaper is per-app-wired and order-sensitive** — every app must call `startWsHeartbeat(wss)` + `heartbeat.register(ws)` in the right gate order (origin → stale → register). Forgettable.
- **pid-echo + `restartCloseCode` threaded by hand** — drishti must share one echo across per-host sockets; an app that omits `restartCloseCode` misclassifies a stale-restart as a transient drop.
- **ssh leg has no application-level periodic liveness** — `@kolu/surface-nix-host` relies on ssh keepalive (~30s) + stream-close + an external `recheck()` signal. A silently *wedged* remote (process alive, app hung) is invisible until the next recheck or stream close.

## The fix — make liveness a property of *constructing* a surface connection

Collapse every liveness mechanism into **two canonical seams (connect / accept), default-on, opt-out only, on every leg, symmetrically** — so an app physically cannot build a surface connection that lacks liveness. This closes pulam-web's gap, drishti's gap, and every future app's gap at once.

Delivered as **one kolu PR** + **one drishti PR** consuming it (the `@kolu/surface*` drishti gate per `.claude/rules/surface.md`).

### Single kolu PR — six points

1. **Contract-agnostic liveness ping in the surface framework.** Add a reserved liveness round-trip answered automatically by the oRPC `RPCHandler` (or standardize on the universal `version` read every surface carries), so the client watchdog needs only the link — not an app-supplied, contract-specific `probe`. This is the keystone that lets liveness be default-on universally.

2. **Canonical client seam — `connectSurface(...)` in `@kolu/surface-app`.** One call builds partysocket → `websocketLink` → `surfaceClient` and wires `createHeartbeat` **default-on** (using the framework ping from #1), plus pid-echo + `restartCloseCode` by default. Opt-out (`heartbeat: false`), never opt-in. Move the heartbeat *down* from the provider-only turnkey branch to the factory, so **every** socket gets it — including per-host fleet sockets.

3. **Canonical server seam — `acceptSurfaceSocket(ws, …)` in `@kolu/surface-app/server`.** One call does the gate ordering correctly (origin → stale → enroll in `startWsHeartbeat`), so a server cannot accept a socket without enrolling it in the reaper, and cannot get the order wrong. The 30s-reap-vs-25s-client-recovery timing relationship is then guaranteed by construction.

4. **Close the ssh-leg caveat in `@kolu/surface-nix-host`.** Add the same watchdog to the host session / `pumpRemoteSurface`: an application-level periodic liveness probe over the stdio link that force-cycles on timeout, so a silently-wedged remote is detected without waiting on ssh keepalive or an external recheck. Default-on inside `getHostSession`/the pump; same "timeout → force reconnect → re-mirror" semantics as the browser leg. Both legs symmetric; neither depends on an app remembering anything.

5. **Migrate kolu + pulam-web onto the seams and delete the hand-wiring.** kolu's `rpc.ts` manual `createHeartbeat`, pulam-web's hand-built per-host sockets in `wire.ts`, and both servers' manual `startWsHeartbeat` + `heartbeat.register` all collapse into `connectSurface` / `acceptSurfaceSocket`. The PR should **remove more liveness code than it adds** — proof the capability moved to the right layer.

6. **Hermetic tests pinning every leg:**
   - half-open client socket (probe never answers) forces `ws.reconnect()` — asserted at the **factory** level, proving default-on;
   - every accepted server socket is auto-enrolled in the reaper (no manual `register`);
   - pulam-web's **per-host** sockets recover (the exact leg that freezes terminals today);
   - the ssh-leg watchdog force-cycles a wedged-but-open link;
   - client-reconnect-wins-the-race timing under fake timers.

### Single drishti PR (consumes the above)

Bump drishti's kolu pin to the PR branch; replace its hand-wiring with the seams. Its server `startWsHeartbeat`, its `SurfaceAppProvider` `{ ws, probe }` plumbing, and especially its **per-host `makeSocket`** all move to `connectSurface` / `acceptSurfaceSocket`. drishti's per-host fleet sockets — today's blind spot — get the client watchdog **for free**, no drishti-specific probe. Net deletion of code; green CI is the behaviour-preserving proof, and satisfies the required surface gate.

## Done when

Both apps' **per-host** sockets recover from a silent half-open (browser leg) and a silently-wedged remote (ssh leg) with zero app-level liveness code; the migration deletes the old manual wiring in all three codebases; every leg is pinned by a hermetic test; drishti CI is green. "An app can forget liveness" is no longer a reachable state.

## References

- `docs/atlas/src/content/atlas/pulam-web.mdx`
- `packages/surface-app/src/connect.ts` (`createHeartbeat`), `packages/surface-app/src/server.ts` (`startWsHeartbeat`, `gateStaleSocket`), `packages/surface-app/src/solid/index.ts:544` (turnkey-only wiring)
- `packages/pulam-web/src/client/wire.ts` (per-host sockets, no probe)
- `packages/surface-nix-host/src/hostFanout.ts` (`pumpRemoteSurface`), `hostSession.ts` (`getHostSession`)
- `packages/client/src/rpc/rpc.ts:86` (kolu's manual `createHeartbeat` — the pattern)

srid:
Closes #1542.

Broken-connection detection in the surface stack was an **opt-in primitive apps could silently forget** — and two of them had. pulam-web builds each per-host fleet socket by hand and never wired a client-side liveness watchdog, so a *silently* half-open server (laptop sleep, Wi-Fi roam, a NAT/proxy dropping an idle connection — TCP dies with no FIN/RST, so neither `close` nor `error` ever fires) froze that host's terminal streams until a manual reload. kolu's own terminals already self-heal from exactly this (#1235); the fleet didn't.

This makes liveness a **property of constructing a surface connection** — default-on, opt-out only, on every leg — so an app can no longer build a surface socket that lacks a watchdog.

## The legs, before → after

| Leg | before | after |
| --- | --- | --- |
| server reaps dead browser (`startWsHeartbeat`) | ✓ kolu, pulam-web | ✓ (now via `acceptSurfaceSocket`, can't-forget) |
| **browser detects dead server** (client watchdog) | ✓ kolu · ✗ pulam-web per-host | ✓ everywhere |
| backend detects dead remote host over ssh | drop-only (EOF/keepalive) | ✓ + **periodic** watchdog for a silent wedge |

## What changed

1. **`@kolu/surface` — a framework-reserved liveness probe.** `defineSurface` injects one reserved procedure, `surface.system.live`, into every surface's contract (never its spec, so dep-walking / coverage tests / mirror forwarding stay unaware of it); `implementSurface` auto-answers it with `{}`. It's a **contract-agnostic** round-trip — `probeSurfaceLive(client)` — so a watchdog needs no app-supplied probe, and there's none to forget. It merges into an app's own `system` namespace (kaval's `system.heartbeat`) and loudly rejects only a duplicate `live` verb.

2. **`@kolu/surface-app` — the client + server seams.**
   - The default-on half-open watchdog is **folded into `createServerLifecycle`** (it already owned the socket's open/close + stale-restart; now it owns liveness too). kolu and `SurfaceAppProvider`'s turnkey branch get it for free and **delete** their separate `createHeartbeat` call.
   - **`connectSurface`** — the turnkey single-surface seam (socket + client + default-on watchdog in one call) for sockets with no shared lifecycle UI: pulam-web's and drishti's per-host fleet sockets. Probe defaults to `system.live`.
   - **`acceptSurfaceSocket`** — the server seam: owns the liveness reaper and sequences stale-gate → enrol → dispatch in one `accept(...)`, so a socket **cannot** be dispatched un-enrolled (and the origin→stale→reap order can't be gotten wrong, re: #1231).

3. **`@kolu/surface-nix-host` — the ssh-leg watchdog.** `HostSession` gains a default-on periodic liveness probe (the ssh twin of `createHeartbeat`): while `connected`, it probes `system.live`; a **timeout** means the remote is silently wedged (process alive, app hung — no EOF, ssh keepalive ~30s away) and force-cycles the child through `recheck()`'s path. Born at the first `markConnected` (never probes before the first RPC) and gated on `connected`. A *rejection* still counts as alive, so an agent too old to answer just degrades to today's behaviour.

4. **Migration.** kolu (`rpc.ts`, `server/index.ts`) and pulam-web (`client/wire.ts`, `server/main.ts`) move onto the seams and **delete the hand-wiring** — net less liveness code, and pulam-web's per-host sockets gain a watchdog they never had.

## Design philosophy

- **Fail-fast, no override knobs.** The watchdogs are default-on; opt-out (`heartbeat: false` / `liveness: false`) exists only for a layer that owns liveness itself, not as a degradation path. A reserved-verb collision throws loudly at boot.
- **Electricity / volatility.** The capability (transport/reconnect liveness) already lived in `@kolu/surface*` with the dependency arrow pointing out; this doesn't add a package — it makes the existing primitive default-on at the one construction seam, so "an app forgot liveness" stops being a reachable state.
- **Reuse the source of truth.** The probe is one reserved verb every surface already answers (not a per-app convention re-derived N times); the ssh watchdog reuses `recheck()`'s existing force-cycle path rather than a parallel one.

## Tests (hermetic, one per leg)

- reserved `system.live` answers + `probeSurfaceLive` + kaval-style coexistence + collision (`surface/liveness.test.ts`)
- half-open client forces `ws.reconnect()` at the factory level, default-on + opt-out (`surface-app/.../lifecycle.test.ts`)
- every accepted server socket auto-enrolled; a stale tab never dispatched (`surface-app/server.test.ts`)
- ssh-leg watchdog force-cycles a wedged-but-open link; leaves a healthy one alone (`surface-nix-host/liveness.test.ts`)
- client-reconnect-wins-the-race timing invariant across the two modules (`surface-app/connect.test.ts`)

`connectSurface` itself isn't unit-constructed — same convention as `createSurfaceSocket` (a live partysocket only flakes in a Node unit test); it's proven by typecheck + the createHeartbeat suite + real-ws evidence.

## Shared-library gate

API-facing changes to `@kolu/surface{,-app,-nix-host}` — a **linked drishti PR** adopting the seams (and inheriting the ssh watchdog transparently) with green CI follows, per `.claude/rules/surface.md`. drishti's per-host sockets gain the client watchdog for free.


srid:
## Review — PR #1545: default-on connection liveness at the canonical seams

**Verdict: ship-with-nits.** This is the right fix in the right place. Liveness is now a property of *constructing* a `@kolu/surface` connection rather than an opt-in primitive two apps already forgot. The diff removes more hand-wiring than it adds, every leg is pinned by a hermetic test, and #1542's "an app can forget liveness" is no longer a reachable state. Two low-severity residues are worth a glance but block nothing.

### What's genuinely good

- **The keystone is correct and minimal.** `defineSurface` injects one reserved `surface.system.live` into the **contract only**, never `spec` — so every spec-walker (coverage ledgers, `mirrorRemoteSurface`'s total-dual, surface-mcp's tool exposure, `implementSurface`'s procedures dep-walk) is untouched, verified against `mirrorRemoteSurface.ts`/`expose.ts`/`coverage.test.ts`. `implementSurface` auto-answers it last and *merges* into an app's own `system.*` (kaval's `system.heartbeat`, the example's `system` cell both survive), and a real `system.live` clash fails loud at boot. Namespace+verb are single-sourced in `liveness.ts`, so contract and handler can't drift.

- **The footgun is removed structurally, not by documentation.** The watchdog now lives inside `connectSurface` and `createServerLifecycle` (client) and `acceptSurfaceSocket` (server) — the seams every socket must pass through. The migration *deletes* the old adjacent calls: `rpc.ts`'s standalone `createHeartbeat`, pulam-web's hand-rolled `createSurfaceSocket → websocketLink → surfaceClient` (the one with no probe), and both servers' manual `startWsHeartbeat`. The pre-PR pattern — probe declared *beside* but outside the socket build — is exactly what got forgotten; there's no longer a probe to omit.

- **Half-open detection is sound on all three legs.** The probe rides the *same* wedged channel with no per-request oRPC timeout, so only the watchdog's own timeout settles it; a sync throw is reported and does **not** reconnect, an async rejection counts as alive, only a true non-answer cycles. The ssh leg reuses `HostSession.recheck()` rather than inventing a second force-cycle path, is born at first `markConnected`, self-gates on `connected`, and `stopLiveness()` runs first in `teardown()`. The client-reconnect-wins-vs-server-reaper race is pinned by a real cross-file constant test (`30s > 15s+10s`).

- **Backward-compat is deliberate.** "Rejection counts as alive" means an old peer predating `system.live` degrades to today's no-watchdog behaviour instead of being spuriously force-cycled — documented on both legs.

### Issues (both low — nits, not blockers)

1. **`connectSurface`'s `heartbeat.probe` override re-opens the very axis the PR closes.** `ConnectSurfaceOptions` lets a caller pass a domain-specific probe instead of the framework `system.live`. The whole point of the keystone is that no app nominates (or forgets) a probe; a speculative override knob is the "being able to override is never a feature" smell, and there's no consumer for it in this diff. *Fix:* drop the `probe?` field (keep `false`/interval/timeout tuning); re-add it only when a real caller needs it.

2. **pulam-web's `buildHostSurface` silently drops the `dispose` handle.** `connectSurface` now returns `{ ws, echo, client, dispose }`, but `wire.ts` destructures only `{ ws, client }`. These per-host sockets are page-lifetime cached, so nothing leaks today — but a dropped teardown handle is a latent leak if the host set ever churns. *Fix:* thread `dispose` into the per-host cache eviction path, or a one-line comment that the discard is intentional (page-lifetime).

### Open judgment call

The `heartbeat` opt-out is shaped two ways across the seams: `createServerLifecycle`/`connectSurface` take `heartbeat: false | {…}`, while `HostSession` takes a flat `liveness?: false` plus sibling `livenessIntervalMs`/`livenessTimeoutMs`. Both fine in isolation, but a reader wiring the ssh and browser legs side by side meets two shapes for one concept. Unify when the nix-host options object is next touched.

srid:

✅ **Consensus** after 2 round(s) · lowy + hickey · base `7c761a76acfc`

Independent findings: lowy=5, hickey=8

### Applied (9)
- `lowy-1` dialAgentOnce still forces each CLI to nominate its own liveness probe — the receptacle it could now plug into — commit `7352cb6e9`
- `lowy-2` system.live is a parallel receptacle to the pre-existing app-nominated liveness verbs (system.heartbeat); confirm the old socket is demoted, not left live — commit `5d5e18857`
- `hickey-1` Two liveness probes for one concept — `identity.info` vs `system.live` — commit `059eaadd7`
- `hickey-2` The half-open watchdog is reimplemented by hand in HostSession instead of reusing `createHeartbeat` — commit `2cb8f5581`
- `hickey-3` Liveness cadence (15s/10s) duplicated as magic numbers in HostSession — commit `2cb8f5581`
- `hickey-4` Two parallel `heartbeat?: false | {...}` option shapes, each hand-destructured — commit `47578c575`
- `hickey-5` `SurfaceLiveProbeable` cast appears at three call sites — the type erodes at every boundary — commit `db9a5adf4`
- `hickey-6` `liveness: false` asymmetry — disable-only flag vs the tune-or-disable union used elsewhere — commit `b37dd7175`
- `hickey-8` HostSession liveness state is three fragmented private fields reconstituting one watchdog — commit `2cb8f5581`

### Agreed — no change (4)
- `lowy-3` Reserved-verb injection couples defineSurface (contract) and implementSurface (handler) across two files via a manual drift-watch (packages/surface/src/define.ts:616 and packages/surface/src/server.ts:1409-1419)
- `lowy-4` acceptSurfaceSocket bundles two independent volatility axes (stale-tab gate + liveness reaper) behind one seam — justified by an order invariant, recorded as examined (packages/surface-app/src/server.ts:568-605)
- `lowy-5` Two client liveness seams (connectSurface / createServerLifecycle) split by lifecycle-ownership axis — examined, single shared watchdog, clean (packages/surface-app/src/solid/connectSurface.ts:70 and packages/surface-app/src/solid/index.ts:126)
- `hickey-7` `acceptSurfaceSocket` bundles three steps but leaves origin-gate + dispatch outside — partial seam (packages/surface-app/src/server.ts:540-606 (acceptSurfaceSocket) and its callers packages/pulam-web/src/server/main.ts:189-213, packages/server/src/index.ts:447-470)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `7c761a76acfc`

### Round 1

**codex** — approved: `false`

I reviewed the branch diff against 7c761a76acfcc3258f4aae7a07d9e169f4ad9d82, including the changed shared surface packages and the new liveness/heartbeat paths. The main implementation shape is coherent, but one recovery-path error is currently swallowed silently in a shared primitive, which is not acceptable for this codebase's fail-loud posture. I did not run tests because this review was requested as read-only and the sandbox is read-only.

Findings:
- `F1` · major · open — `createHeartbeat` catches and drops exceptions from `opts.onStale()`. That callback is the actual recovery action in callers, such as `ws.reconnect()` in `@kolu/surface-app` and `HostSession.recheck()` in `@kolu/surface-nix-host`; if it throws, the watchdog continues with no operator-visible signal and may still run the stale reporter as though recovery was attempted successfully. This turns a liveness failure into a silent half-open state, directly undermining the watchdog's purpose. (packages/surface/src/heartbeat.ts:103)
- `F2` · minor · open — The README points readers at `createHeartbeat` from `@kolu/surface-app`, but the new public export is under the `@kolu/surface-app/connect` subpath. Following this documentation will lead consumers to the wrong import path. (packages/surface/README.md:48)

**claude** — Addressed both CODEX findings against branch base 7c761a76. F1 (major) — agreed on the core defect: `createHeartbeat`'s guarded catch around `opts.onStale()` (and the `onStaleReport`/`onProbeError` reporter catches) swallowed throws with a silent empty `catch {}`. A recovery action that throws leaves the link half-open with no recovery applied — the exact silent half-open the watchdog exists to prevent — which violates the package's fail-loud posture (`caught-error-must-not-collapse-to-empty`). Fixed by surfacing each guarded throw via `console.error` while preserving the must-not-unwind-the-timer guarantee, plus a new test asserting a throwing `onStale` is logged and the interval survives. I disputed CODEX's two framing errors: (a) routing the error through `onStaleReport` would be wrong — that callback is the 'I forced a recovery' reporter (default `console.warn`), not an error channel, and the detection it reports genuinely happened; (b) `console.error` matches the package's existing convention (`project.ts:247`) and the wrapper's `warnStale`/`warnProbeThrew` console defaults. F2 (minor) — partial: `surface/README.md:48` named the browser-leg `createHeartbeat` as `(@kolu/surface-app)`, which is a package attribution paralleling the ssh-leg's `(@kolu/surface-nix-host)` on the same line, not an import statement; I tightened it to the precise `@kolu/surface-app/connect` subpath but disputed CODEX's claim that it 'will lead consumers to the wrong import path' (it is prose, and the canonical import table at surface-app/README.md:107 already lists the subpath). Gate green: `pnpm typecheck` passes across all packages, biome clean on the edited TS files, fmt-check clean, and heartbeat tests 10/10 (plus surface-app connect 21/21 and nix-host liveness tests still pass — the callers' non-throwing `onStale` actions are unaffected). Note for the surface→drishti merge gate: F1 changes the error-path runtime behavior of the shared `@kolu/surface` primitive (logs instead of swallowing) but adds no signature/API change and only fires when an `onStale`/reporter throws — drishti's wrapper actions don't throw — so it needs no drishti wiring change, only a re-pin if the gate re-validates.

- `F1` **partial** — Agreed the silent swallow is a defect and fixed it: the catch around opts.onStale() in packages/surface/src/heartbeat.ts now emits console.error('heartbeat: onStale recovery action threw — link may still be half-open, recovery NOT applied', error) instead of an empty catch, keeping the timer alive (the must-not-unwind guarantee). Applied the same surface-it-don't-swallow treatment to the onStaleReport and onProbeError reporter catches, and updated the settle JSDoc. Added a heartbeat.test.ts case asserting a throwing onStale is logged via console.error and the interval still probes on a later tick (10/10 pass). Disputed CODEX's suggested mechanism: it proposed passing the caught error to onStaleReport or adding onRecoveryError. onStaleReport is the 'a missed probe forced recovery' REPORTER (default console.warn in the surface-app wrapper, heartbeat.ts:66-68 / connect.ts:240), not an error channel — the detection it reports genuinely happened, so re-purposing it to mean 'recovery failed' would be wrong, and the stale report it emits is correct, not misleading as CODEX claimed. console.error is the right surface here: it matches the package's own convention (project.ts:247) and the wrapper's warnStale/warnProbeThrew console defaults, with no extra public knob.
- `F2` **partial** — Fixed the valid part: tightened packages/surface/README.md:48 from 'the browser-leg createHeartbeat (@kolu/surface-app)' to '(@kolu/surface-app/connect)' so the attribution names the exact subpath. Disputed CODEX's stronger claim that the text 'will lead consumers to the wrong import path': line 48 is prose attributing the browser leg to a package — it parallels the ssh-leg's '(@kolu/surface-nix-host)' on the same sentence — not a code import line, and the canonical import-path table at packages/surface-app/README.md:107 already correctly lists '@kolu/surface-app/connect' with createHeartbeat. No consumer follows README.md:48 as an import; the edit is a precision improvement, not a bug fix.

commit: `c20659b602863ee71883f7d376054d0d5e44f645`

### Round 2

**codex** — approved: `true`

I re-inspected the current tree against 7c761a76 and checked only the two findings already on the table plus CLAUDE's follow-up commit c20659b. `git status --short` is clean. Both findings are now resolved, and I did not identify a regression introduced by this round's changes.

Findings:
- `F1` · major · resolved — Originally, `createHeartbeat` silently swallowed exceptions from the stale-link recovery callback. The current code now catches the error, emits a `console.error` with the thrown value, keeps the interval alive, and adds a regression test covering a throwing `onStale`. That addresses the silent failure mode I raised. (packages/surface/src/heartbeat.ts:108)
- `F2` · minor · resolved — Originally, the README attributed browser-leg `createHeartbeat` to `@kolu/surface-app`, which was imprecise against the actual subpath export. The current README now names `@kolu/surface-app/connect`, so the documentation is precise enough. (packages/surface/README.md:48)

_codex on the rebuttal:_ F1: I accept the fix and concede the mechanism. My important requirement was that a throwing recovery action must not disappear silently while leaving the watchdog running; the new `console.error(..., error)` does that without unwinding the timer. I also accept CLAUDE's point that `onStaleReport` is a stale-detection reporter, not an error channel, so overloading it would be the wrong shape. F2: I accept the pushback that the original line was prose rather than an import snippet; my wording that consumers would necessarily follow it as an import path was too strong. The precision edit to `@kolu/surface-app/connect` resolves the finding.

srid:

**All clear** — 0 findings. Scoped to `git diff 7c761a76acfc...HEAD`.

| Pass | Issues | Notes |
|---|---|---|
| Rules | 0 | All 28 rules (7 built-in + 21 project) checked against the diff — the watchdog lifted once into `@kolu/surface/heartbeat` (no duplication), every catch surfaces via `console.error`/`onProbeError`, `SurfaceConnectionStatus` is a discriminated union, the page-lifetime `dispose` discard is documented (no leak), `ConnectionIndicator` is Tailwind-only + `<Switch>/<Match>`. |
| Fact-check | 0 | No silent swallowing, logic errors, wishful thinking, or leaks. Verified the heartbeat settlement guards (`inFlight`/`disposed`), the 25s-client < 30s-server-reap timing invariant, and the `down` vs `reconnecting` terminal/transient distinction. |
| Elegance | – | Skipped (`--no-elegance`; the `/simplify` track already ran this tree). |

Final step of the serial review gauntlet (lens-debate → codex-debate → simplify → code-police).

srid:

The **x86_64-linux** lane ran on **`kolu-ci-2`** (`dev-x86-64-linux-04`) — commit `3903995b9`, **passed** ✓

- **Lane wall** (pipeline): **3m17s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 2m35s | ✓ |
| `ci::home-manager` | 2m1s | ✓ |
| `ci::unit` | 54s | ✓ |
| `ci::flake-check` | 54s | ✓ |
| `ci::smoke` | 52s | ✓ |
| `ci::nix` | 52s | ✓ |
| `ci::pnpm-hash-fresh` | 50s | ✓ |
| `ci::surface-example-build` | 37s | ✓ |
| `ci::render` | 37s | ✓ |
| `ci::fmt` | 36s | ✓ |
| `ci::biome` | 36s | ✓ |
| `_ci-setup` | 32s | ✓ |
| `ci::install` | 8s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/3903995/timings.jsonl`; pool state is a live `flock` probe.</sub>

srid:
## Evidence

**The connection indicator, live.** Drove `pulam-web` (`PULAM_WEB_HOSTS=localhost`, so the dashboard reads this machine's own kaval) and dropped the network with Chrome DevTools emulation to force the per-host socket half-open — the case that previously froze the fleet board silently.

pulam-web connection indicator

- **Top — connection dropped:** the `localhost` host header shows an amber **● reconnecting…** while partysocket retries (the dashboard keeps its 13 live terminals, including this very `claude · kolu · pulam-conn` session).
- **Bottom — connection restored:** the watchdog/transport reconnects and the indicator **clears back to live** on its own.

Before this PR there was *no* such indicator — a silently half-open per-host socket left the board frozen with no signal until a manual reload. The status-derivation logic is also pinned hermetically (`socketStatus.test.ts`: `connecting → live → reconnecting → down`), and the half-open detection itself by the per-leg watchdog tests (`heartbeat.test.ts`, `surface-nix-host/liveness.test.ts`).

_Captured via the Chrome DevTools MCP against a local `nix run`-equivalent `pulam-web`; pulam-web isn't covered by kolu's Cucumber e2e harness, so this is the live-drive path._

srid:
## Evidence (updated) — green "connected" status

Per review feedback (the indicator should read *positively connected*, like kolu's header dot, not just surface problems): each host now wears a **persistent status dot** — solid **green when connected**, amber **"reconnecting…"** while it recovers, red **"disconnected — reload"** on a stale-close.

pulam-web connection status

- **Top — connected:** the `localhost` header carries a solid **green dot** (the link is healthy).
- **Bottom — connection dropped:** it flips to amber **"reconnecting…"** while partysocket retries.

Captured live (Chrome DevTools network-offline emulation against a local `pulam-web`). Restoring the network flips it **back to green on its own — no reload** — and the live agent rows resume updating, which is the watchdog's whole point. (Supersedes the earlier evidence image, which hid the dot when healthy.)

srid:

The **x86_64-linux** lane ran on **`kolu-ci-2`** (`dev-x86-64-linux-04`) — commit `20c7fe4e2`, **passed** ✓

- **Lane wall** (pipeline): **3m40s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 3m0s | ✓ |
| `ci::home-manager` | 1m20s | ✓ |
| `ci::smoke` | 50s | ✓ |
| `ci::unit` | 49s | ✓ |
| `ci::nix` | 48s | ✓ |
| `ci::render` | 31s | ✓ |
| `_ci-setup` | 30s | ✓ |
| `ci::biome` | 30s | ✓ |
| `ci::fmt` | 29s | ✓ |
| `ci::install` | 9s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/20c7fe4/timings.jsonl`; pool state is a live `flock` probe.</sub>

PR #1550, opened 2026-06-24
fix(pulam-web): drop the awareness fold on link death so a reconnect can't pin a stale agent state

### LINKED ISSUE #1549: pulam-web: stale `working` agent state pinned across an ssh-link reconnect (re-serve awarenessCache never cleared)
## Symptom

pulam-web (the browser agent dashboard) shows a **different agent state than kolu's Dock** for the same terminal. Concrete report: an `opencode` agent shows **working** in pulam-web while the Dock (correctly) shows **idle**.

## Not the projection, not the render (ruled out first)

After R-dock-unify (#1541) both surfaces compute the label from the *same* `@kolu/terminal-workspace/agentProjection`, so for an identical `agent.state` they can't disagree. Verified that pulam-web's label is a faithful pure function of `agent.state`:

- `packages/pulam-web/src/client/HostGroup.tsx:154` → `fleetStateLabel(value().agent, …)` → `agentProjection.ts:210-218` reads **only** `agent.state` via `agentUrgency`. No activity / foreground / `lastActivityAt` path can promote the label.
- The green "moving bytes" dot is a **separate** span on the activity stream (`HostGroup.tsx:130-137`), orthogonal to agent state.

So the render is correct — the divergence is entirely in the **`agent.state` data reaching pulam-web**.

Also ruled out a steady-state delta drop: collections have **no** equals-dedup — `wrappedUpsert` always publishes (`packages/surface/src/server.ts:1217-1221`), unlike cells which gate on `equals` (`server.ts:154`). So while the link is **healthy**, every `working→idle` transition does reach the browser. That narrows the fault to the **reconnect / link-down window** below.

## Root cause — stale `working` pinned in the re-serve `awarenessCache` across an ssh-link respawn

pulam-web *always* reaches its data over a pooled, **reconnecting** ssh subprocess link (`packages/pulam-web/src/server/main.ts:6` — `browser ─WS─▶ this server ─stdio oRPC over ssh─▶ remote pulam`; `hostEntry.ts:7-11`). The browser's awareness rows come from a single `awarenessCache` map created **once per host session** and closed over by both the browser-facing `readAll` and the mirror's write sink:

- `packages/pulam-web/src/server/reserve.ts:113` — `const awarenessCache = new Map<TerminalId, AwarenessValue>()`, built once in `buildReServe`.
- `reserve.ts:143-152` — `readAll: () => awarenessCache`; the sink only ever does per-key `awarenessCache.set` / `awarenessCache.delete`.

When the ssh link dies and respawns, two things conspire so the cache is never reset:

1. The mirror's teardown on link death fires **no `onRemove`** for open keys — it only aborts the per-key controllers:
   `packages/surface/src/mirrorRemoteSurface.ts:656-660` — `finally { for (const ctl of open.values()) ctl.abort(...) }`. `onRemove` fires **only** in the keys-departed branch (`~:631-646`), which never runs when the keys stream simply tears down.
2. `pumpRemoteSurface` re-runs the mirror with a **fresh** sink but the **same** cache, and its post-`mirror.done` `finally` clears only the live holders, never the cache:
   `packages/surface-nix-host/src/hostFanout.ts:198-210` — sets `liveProcedures.current = null` / `liveClient.current = null`; `awarenessCache` is untouched.

So if an agent finishes (`working→idle`) **while that link is down/reconnecting**, the last `working` value sits in `awarenessCache` and the browser keeps painting it. The Dock reads kolu's own in-process awareness (`packages/client/src/canvas/dock/dockRowRanking.ts:80` — `agentUrgency(activeArm(meta)?.agent)`) with no such transport window, so it's correctly idle.

**Smoking gun:** `reserve.ts:113` (cache made once) + `hostFanout.ts:198-210` (respawn clears holders, not cache) + `mirrorRemoteSurface.ts:656-660` (link-death teardown fires no `onRemove`).

## Two flavors (both fixed by the same change)

- **Case A — terminal stays alive, agent goes `working→idle` during the down window** (the literal opencode report). **Self-heals on reconnect**: the fresh mirror re-subscribes the remote `keys` stream, whose first frame is the remote's *current* snapshot; the still-present key is in it and `onUpsert` overwrites `working`→idle. So it's bounded by reconnect latency — but you see a visible stall on **every** reconnect, and reconnects are normal operation, not an edge case.
- **Case B — terminal also departs during the same down window.** A genuine **unbounded ghost**: the departed key is absent from the new `keys` snapshot, so the keys-departed `onRemove` branch never fires for it, and nothing ever clears `awarenessCache[T]`. The phantom `working` row persists indefinitely (the Dock shows it as gone).

## Fix

On `mirror.done` / link death, **clear `awarenessCache`** (or fire `onRemove` for all open keys) so the next spawn rebuilds cleanly from the fresh snapshot. One change collapses both cases: Case A rebuilds from the snapshot, Case B's ghost is dropped. Localized to the pulam-web re-serve — `packages/pulam-web/src/server/reserve.ts` + `packages/surface-nix-host/src/hostFanout.ts` — with **no** change to the shared sensor or the projection.

Per the project's fail-fast convention, a stale value surviving a transport drop is a silent-degradation defect (a caught link death collapsing to retained-but-wrong state), so this is a fix-now, not a defer.

## Confidence / verification

Mechanism **verified directly against HEAD**: `reserve.ts:113,143-152`; `mirrorRemoteSurface.ts` `finally` + `onRemove` branch; `hostFanout.ts:198-210`; `dockRowRanking.ts:80`. Confidence **high** on the mechanism and on Case B being a real unbounded defect; **medium** on whether the specific opencode sighting was Case A (in-window, most likely since the agent was still running) vs Case B — the fix covers both, so it doesn't gate the fix.

## Out of scope (separate, structural — do **not** fold in here)

Two *transient* divergences also exist: kolu and pulam are two separate sensor processes running the byte-identical `startAwareness`, each with its own 150ms WAL debounce, so pulam can publish `working` a beat after the Dock flips idle (self-heals on the next pulse). That's timing skew, not a logic bug, and it's the two-sensor "fence" that **R8 (one shared sensor run)** dissolves — tracked there, not here.

srid:
Fixes #1549.

## The bug

pulam-web (the browser agent dashboard) reaches each host's agents over a **reconnecting** ssh link. When the link blips and an agent finishes its turn (`working→idle`) — or its terminal closes — *while the link is down*, the dashboard keeps painting the last value it saw. So an idle agent shows **working**, diverging from kolu's own Dock, which reads the same projection from kolu's in-process awareness (no transport window) and is correct.

Two flavors, both fixed by one change:

- **Case A — terminal alive, agent went `working→idle` during the down window.** Self-heals on reconnect (the fresh snapshot re-upserts `idle`), but visibly stalls on every reconnect, and reconnects are normal operation.
- **Case B — terminal also departed during the window.** A genuine **unbounded ghost**: the departed key is absent from the new snapshot, and the new mirror's per-key bookkeeping starts empty, so its `onRemove` never fires. The phantom `working` row persists indefinitely.

The **activity live-set** (the green "byte-tap" dot) is the same per-host-session local state with the same pathology: if the last frame before the link died named a terminal, a dead link's last frame keeps painting it live across the reconnect.

## Root cause

The re-serve's `awarenessCache` AND its `activityLatest` live-set (`packages/pulam-web/src/server/reserve.ts`) are built **once per host session** and closed over by every (re)spawn's sink. On link death the mirror's teardown aborts the per-key controllers but fires **no `onRemove`** and **no clearing activity frame**, and `pumpRemoteSurface`'s respawn clears the live holders but never these folds — so a value that changed or departed during the down window is never reconciled.

## The fix

The reconnect-mirror pump now fires an optional **`onLinkDown`** hook on each link death (after clearing the live holders). The re-serve answers it with **`resetRemoteFold()`**, which drops BOTH remote-derived folds: each cached awareness key goes through the **framework-wrapped collection `remove`** (publishing each departure through the collection's channels), and the activity live-set resets to `[]` with an empty bus frame — so a browser subscribed *across* the reconnect sees the rows leave and the dots go dark (not just a fresh-subscribe one). The next spawn rebuilds cleanly from the remote's authoritative snapshot.

| File | Change |
| --- | --- |
| `packages/surface-nix-host/src/hostFanout.ts` | add optional `onLinkDown` to `PumpRemoteSurfaceOptions`, fired in the pump's `finally` |
| `packages/pulam-web/src/server/reserve.ts` | add `resetRemoteFold()` — publish awareness removes + clear cache, reset the activity live-set |
| `packages/pulam-web/src/server/hostEntry.ts` | wire `onLinkDown: () => reServe.resetRemoteFold()` |

## Design philosophy

- **Fail-fast / no fallbacks.** This *removes* a silent-degradation defect — a caught link death was collapsing to retained-but-wrong state. No knob, no degraded path added.
- **Volatility boundaries.** Link-death/reconnect volatility stays owned by the pump (`@kolu/surface-nix-host`); `onLinkDown` is a clean event it publishes outward. The app-specific fold reset stays in pulam-web's re-serve. The dependency arrow points out.
- **Reuse the source of truth.** The reset goes through the framework-wrapped `remove` (the canonical publish seam), not a bare `awarenessCache.clear()` that would bypass the collection's channels. The shared `agentProjection` fold and the Dock⇄mirror contract are untouched.

## Tests

- `reserve.test.ts` — a red→green reproduction of the **Case B** ghost (two spawns through one re-serve, exactly as `pumpRemoteSurface` drives it; an already-subscribed browser sees A depart) AND a second red→green test for the **activity** half: a frame lights A live, the link dies, and `resetRemoteFold` is asserted to clear the dot for both an already-subscribed browser and a fresh subscriber (verified RED with the activity reset neutered).
- `hostFanout.test.ts` — a `pumpRemoteSurface` test pins the production wiring: `onLinkDown` fires when a spawn's mirror ends, AFTER the live holders are cleared (verified RED when the pump's `onLinkDown?.()` call is removed). This guards against a regression in the pump that would leave the consumer's fold-reset test green while the real bug returns.

## Shared-surface gate

`packages/surface-nix-host` is consumed by [drishti](https://github.com/srid/drishti). The `onLinkDown` addition is **purely additive** — an optional field on `PumpRemoteSurfaceOptions`; drishti's only `pumpRemoteSurface` call (`packages/app/src/server/router.ts`) never passes it and compiles unchanged (verified against the local drishti checkout). Per `.claude/rules/surface.md` a companion drishti PR re-validating full CI against this **final** kolu HEAD is required before merge: the gate PR is **[srid/drishti#75](https://github.com/srid/drishti/pull/75)**, pinning the npins kolu source to this PR's HEAD.


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `5856eb537fc9`

### Agreed — no change (5)
- `lowy-1` onLinkDown is the correct receptacle wire — no new boundary, no parallel encapsulation (packages/surface-nix-host/src/hostFanout.ts:139-147,219-222)
- `lowy-2` Asymmetric reset across the re-serve's three per-link folds — activityLatest is the missed seam (packages/pulam-web/src/server/reserve.ts:135,276-277,291-295)
- `lowy-3` resetAwareness name describes the activity, not the volatility axis it encapsulates (packages/pulam-web/src/server/reserve.ts:99,291)
- `hickey-1` Reconnect-reconciliation rule is fragmented across the three PUSH folds — awareness resets on link-down, activityLatest does not (packages/pulam-web/src/server/reserve.ts:135 (activityLatest) vs :291 (resetAwareness))
- `hickey-2` `onLinkDown` name describes the trigger but the doc says it also fires on session-destroy — the concern is "per-link fold teardown," not strictly "link down" (packages/surface-nix-host/src/hostFanout.ts:147)
---

**Orchestrator adjudication (post-debate).** The lenses *dropped* `lowy-2`/`hickey-1` (reset `activityLatest` too), reasoning the activity live-set self-heals via its value-bearing snapshot stream. The **next** gauntlet step — the codex debate — re-raised exactly this as its F1 and the author applied it: `resetAwareness` → `resetRemoteFold`, now also clearing `activityLatest`/publishing `[]`. I adjudicated **in favor of keeping the activity reset**, because the lenses' flicker objection does not materialize: the awareness reset already removes the rows during the down window, so there are no rows to flicker — the activity reset only ensures re-added rows start *dark* rather than briefly stale-lit (the inter-stream first-frame gap on reconnect). It also makes the fix's own principle (*a caught link death must surface, never collapse to retained-but-wrong state*) symmetric across both remote-derived folds. `versionStore` stays excluded (it carries a meaningful seeded default).

srid:
## Codex ⇄ Claude debate

🔁 **Converged** · codex reviewer effort `xhigh` · base `5856eb537fc9`

> The debate's tail render crashed on an infrastructure error (StructuredOutput retry cap), so this summary is reconstructed verbatim from the per-round section files (codex's verdicts + Claude's dispositions). codex produced clean verdicts every round; the crash was post-consensus rendering, not a reviewer failure.

codex (xhigh) raised **4 findings**; rounds 1–2 resolved F1/F2/F4 in the tree (codex accepted each), leaving only the F3 cross-repo gate, which is a deferred orchestrator step.

### F1 · major · **fixed** — the reset left the sibling `activityLatest` stale
`resetAwareness()` cleared only the awareness cache; the activity live-set has the identical staleness pathology (a dead link's last frame paints a quiet terminal live across the reconnect — existing subscribers never hear it leave, fresh subscribers read the stale snapshot). Renamed `resetAwareness` → `resetRemoteFold`; it now also resets `activityLatest = []` and publishes `[]` on the bus. New red→green test covers both an already-subscribed and a fresh subscriber. *(This re-raised the lens debate's dropped `lowy-2`; adjudicated in favor of applying — see the lens comment.)*

### F2 · minor · **fixed** — the regression swallowed mirror failures
The main path did `await mirror1.done.catch(() => {})`, hiding every failure. Abort-driven teardown *resolves* `.done` (only a `SinkError` rejects), so it now awaits bare — a real sink/setup regression surfaces instead of passing silently.

### F4 · minor · **fixed** — the pump's `onLinkDown` wiring was untested
Added a `pumpRemoteSurface — onLinkDown` test in `hostFanout.test.ts` driving the hook through the real pump lifecycle (fake session, one spawn, link death), asserting it fires **after** the live holders clear. Verified RED when the pump's `onLinkDown?.()` call is removed — guarding the production wiring the consumer-side fold-reset test can't.

### F3 · blocking · **deferred (agreed real)** — the drishti surface gate
Adding `onLinkDown` to exported `PumpRemoteSurfaceOptions` is an API-facing change to shared `@kolu/surface-nix-host`, so per `.claude/rules/surface.md` a drishti PR with green CI against the final kolu HEAD is required. Verified the change is **purely additive** (drishti's sole `pumpRemoteSurface` call passes no `onLinkDown` and typechecks unchanged against the local drishti checkout — all three workspace members `tsc --noEmit` EXIT=0). The gate is structurally a post-this-PR step (drishti pins a pushed kolu tarball), tracked and owned by the orchestrator before merge.

srid:

**All clear.**

| Pass | Issues | Notes |
| --- | --- | --- |
| Rules | 0 | All 28 rules checked clean (no-unbounded-growth: the change *solves* growth by clearing the cache + live-set on each link death; caught-error-must-not-collapse: no try/catch — synchronous cleanup; callback-fanout-guarded: `onLinkDown` has one call site). |
| Fact-check | 0 | Snapshot-before-remove is correct (the wrapped remove mutates the map mid-iteration); hook ordering is correct (live holders cleared before `onLinkDown` fires, so `resetRemoteFold` never sees a stale live-client pointer); the activity reset mirrors the awareness pattern for both fresh and already-subscribed browsers. |
| Elegance | — | Skipped (`/simplify` already ran this gauntlet: reuse/efficiency/altitude confirmed clean). |

srid:
## Evidence

**Why no visual capture.** This bug lives entirely in the *data* reaching pulam-web, not its render — the issue itself ruled the render/projection out (both surfaces compute the label from the same `agentProjection` for an identical `agent.state`). The fix makes the same unchanged row show the correct `idle`/gone state instead of stale `working`. Reproducing the on-screen divergence requires a **real reconnecting ssh link dropping at the precise moment** an agent goes `working→idle` (or its terminal departs) — a multi-host transport-timing scenario the e2e harness can't stage (pulam-web has no Cucumber/Playwright coverage; the suite drives kolu, not the separate dashboard server), and which a live chrome-devtools drive can't reliably trigger either. A still of an `idle` row would look identical with or without the bug, so it proves nothing.

So the faithful proof is the **red→green behavioral test at the exact seam the bug lives in** — driving the real `agent → mirrorRemoteSurface → buildReServe → browser surfaceClient store` pipe with the transport collapsed to `directLink`, and the pump's `onLinkDown` lifecycle driven exactly as `pumpRemoteSurface` drives it in production.

### RED — with `resetRemoteFold` neutered (the bug)
```
❯ src/server/reserve.test.ts (7 tests | 2 failed | 5 skipped)
  × drops a terminal that departed during the link-down window across the respawn
  × clears the activity live-set on link death — for both an existing and a fresh subscriber
```
Both folds stay stale across the respawn: the departed terminal's awareness row is the **unbounded ghost**, and the activity dot keeps painting a quiet terminal live.

### GREEN — with the fix
```
✓ src/server/reserve.test.ts  — drops a terminal that departed during the link-down window across the respawn
✓ src/server/reserve.test.ts  — clears the activity live-set on link death — for both an existing and a fresh subscriber
✓ src/hostFanout.test.ts      — pumpRemoteSurface fires onLinkDown after a mirror ends, AFTER the live holders are cleared
```

The three together prove the full chain end-to-end: the pump fires `onLinkDown` at link death (after clearing the live holders), the re-serve answers by dropping both remote-derived folds, and an **already-subscribed** browser sees the ghost row depart and the live dot go dark — not just a fresh-subscribe one. Each was verified RED on the pre-fix behavior and GREEN with the fix.

srid:
✅ **Shared-surface gate satisfied.** drishti gate PR [srid/drishti#75](https://github.com/srid/drishti/pull/75) — which pins drishti's npins kolu source to this PR's HEAD (`950ad6d1b`) — has **full CI green on both platforms**: 18/18 odu nodes ok (`nix`, `typecheck`, `fmt-check`, `bun-nix-fresh`, `home-manager`, `drv-stability` × {aarch64-darwin@zest, x86_64-linux@kolu-ci-2}), 0 failed. Confirms the additive `onLinkDown` field doesn't break the downstream consumer. Re-pin drishti to master after this merges.

PR #1559, opened 2026-06-24
fix(kaval): retain last command + snapshot-first commandRun so late sensors catch codex (#1558)

### LINKED ISSUE #1558: kaval: retain last command + snapshot-first commandRun (late sensors miss command-only agents like codex)
## Problem

A sensor that attaches to an **already-running** terminal cannot detect a **command-only** agent — one whose foreground process name isn't the agent name. `codex` (npm-installed) runs as `node`, so `matchesAgent` can only match it via `lastAgentCommandName === "codex"` (`packages/integrations/anyagent/src/agent-adapter.ts:197-200`; gate at `packages/integrations/codex/src/agent-adapter.ts:30`). That field is fed by the `commandRun` tap, which is **fire-and-forget**: nothing retains the last command, and the source yields no snapshot on subscribe. So a sensor that starts *after* the agent launched never learns the command and shows the terminal as a non-agent `node`.

This is why pulam-web (whose pulam attaches lazily to running terminals, `packages/pulam/src/daemon.ts:283-309`) shows codex as `node` while kolu (live at launch) detects it. `claude` is unaffected because it keys off the snapshot-first `foregroundPid` instead (`packages/integrations/claude-code/src/agent-adapter.ts:43`).

It's the inverse of how `foreground` already behaves: that source yields a current snapshot first (`packages/kaval/src/inProcessPtyHost.ts:161-169`, reading `getProcess`/`getForegroundPid`), so a late subscriber catches up. `commandRun` should do the same.

## Fix — localized to `@kolu/kaval`, mirroring the existing `foreground` pattern

**1. Retain the last command on the pty-host `Entry` — `packages/kaval/src/ptyHost.ts`:**
- Add a `lastCommand: string | undefined` field to the per-PTY `Entry` (init `undefined`; the `Entry` is constructed around `:501` next to `commandRunChannel: new Channel<string>()`).
- In the OSC 633;E handler (`~:543-549`), set `entry.lastCommand = command` alongside the existing `entry.commandRunChannel.publish(command)`. (The command string is available at the mark even though the process forks just after — `:325-326,:394`.)
- Add a host accessor `getLastCommand(id: PtyId): string | undefined`, mirroring `getProcess` / `getForegroundPid` (`:276-278`).

**2. Make the `commandRun` source snapshot-first — `packages/kaval/src/inProcessPtyHost.ts`:**
- In the `commandRun` source (`~:147-154`), yield `host.getLastCommand(input.id)` first (when defined), then the live stream — exactly mirroring the `foreground` source (`:161-169`).

**No consumer changes.** `sensors.ts`' `commandRun.consume` (`packages/terminal-workspace/src/sensors.ts:385-405`) and the `kavalChannels` bridge (`packages/terminal-workspace/src/kavalChannels.ts:87-92`) already process command events; the snapshot is just one more event delivered on subscribe. The consumer's reconcile is idempotent (same rationale the `foreground` snapshot relies on — `inProcessPtyHost.ts:158-160`).

## Acceptance criteria

- A sensor attaching to an **already-running** codex detects it as an agent, not non-agent `node`. Repro: launch codex in a terminal, then start a fresh sensor/pulam → it resolves `codex`.
- A sensor that **restarts** mid-session re-detects the agent.
- Live behaviour unchanged: a new command still fires the tap.
- An **exited** command is not falsely detected — staleness is handled by the existing `shellIdle` gate (`sensors.ts:425`) together with the snapshot-first `foreground`: a replayed `lastCommand` only resolves to an agent while the shell is non-idle.
- A hermetic test: publish a command-run, then subscribe a **new** `commandRun` consumer and assert it receives the last command snapshot-first (mirror any existing `foreground` snapshot-first test).

## Notes

- Stays within `@kolu/kaval`'s pty-host (same layer as the existing `foreground` snapshot-first), so it is **not** an `@kolu/surface` contract change and does **not** need a paired drishti PR. If implementation ends up touching the shared `Channel` primitive (`channel.ts`) rather than just the pty-host source/accessor, re-check `.claude/rules/surface.md`.
- Roadmap context: this is the "non-retained `commandRun`" intrinsic gap tracked under **R9a** in the `remote-terminals` Atlas note. R9a (pulam-web localhost reads kolu's awareness) hides it on localhost; this fix closes it for any sensor that starts or restarts late, including a remote pulam.

srid:
## What

A sensor that attaches to an **already-running** terminal couldn't detect a **command-only** agent — one whose foreground process name isn't the agent name. `codex` is installed via npm, so it runs as `node`; the only way to recognize it is the command line you ran. That signal — kaval's `commandRun` tap — was **fire-and-forget**: nothing retained the last command, and the source replayed nothing on subscribe. So a sensor that started (or restarted) **after** the agent launched never learned the command and showed the terminal as a non-agent `node`.

This is why pulam-web (whose pulam attaches lazily to running terminals) and a restarted/remote pulam showed codex as `node`, while kolu — live at launch — detected it. `claude` was unaffected (it keys off the snapshot-first `foreground` source).

Fixes #1558.

## Fix

Localized to `@kolu/kaval`, mirroring the **existing** `foreground` snapshot-first source:

1. **Retain the last command** on the pty-host `Entry` (`ptyHost.ts`) — set `entry.lastCommand` on the OSC 633;E mark, exposed via a new `getLastCommand(id)` accessor (sibling of `getProcess` / `getForegroundPid`).
2. **Make `commandRun` snapshot-first** (`inProcessPtyHost.ts`) — yield the retained command first (when present), then the live stream. Exactly the shape `foreground` already uses.

No consumer changes. Not an `@kolu/surface` contract change (it stays in the pty-host, below the wire — `channel.ts` untouched), so **no paired drishti PR** is needed.

### Design philosophy
- **No fallbacks / fail-fast** — no new knob or degraded path; the `if (last !== undefined)` is simply "nothing to replay yet", identical in spirit to `foreground`'s default.
- **Volatility boundary** — stays inside kaval's pty-host, the same layer that owns the `foreground` snapshot-first behavior. No domain-agnostic helper folded into an app module.
- **Reuse the source of truth** — mirrors the established `foreground` snapshot-first mechanism rather than inventing a parallel one.

## Reproduction (test-first)

Two hermetic tests, both **red on the prior code**, green after the fix:
- `ptyHost.test.ts` — drive an OSC 633;E mark, assert `getLastCommand(id)` retains it.
- `inProcessPtyHost.test.ts` — drive a command-run, then subscribe a **new** `commandRun` consumer and assert it receives the command **snapshot-first** (before the fix it gets nothing and times out).

## Docs
- `packages/kaval/README.md` — the command-run tap row gains its `getLastCommand` getter (mirroring the cwd/title rows).
- `website/.../changelog/unreleased.mdx` — a `Fixed` entry.
- `docs/atlas/.../remote-terminals.mdx` — History entry on the R9a late-attach gap; `dist/` rebuilt.


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `d148eef85183`

Independent findings: lowy=2, hickey=1

### Agreed — no change (3)
- `lowy-1` Snapshot-first replay pattern is duplicated per-source, not factored — but correctly so here (packages/kaval/src/inProcessPtyHost.ts:147-176)
- `lowy-2` Change restores a broken symmetry across the last-value sources (the correct fix) (packages/kaval/src/ptyHost.ts:276-279,322,562,761)
- `hickey-1` Snapshot-first source generators duplicated across taps (foreground now joined by commandRun) (packages/kaval/src/inProcessPtyHost.ts:153-177)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 round(s) · codex reviewed at `xhigh` reasoning effort · base `d148eef85183`

### Round 1

**codex** — approved: `false`

The core retention idea is in the right layer, but the current wire shape turns a replayed snapshot into an indistinguishable live command event. That leaks into downstream side effects that are explicitly live-event based, so late subscriptions/reconnects can mutate recent-agent recency as if the user had just run the command. I did not run tests because this review was requested read-only.

Findings:
- `F1` · major · open — The retained `commandRun` value is replayed with the exact same `{ command }` shape as a live OSC 633;E event. Downstream, `startAgentCommandSensor` treats every commandRun frame as live: it updates metadata and calls `trackRecentAgent`, which stamps `Date.now()` into the MRU. A late subscriber, sensor restart, reconnect, or the subscribe/getLastCommand race can therefore re-bump an old agent command and reorder the recent-agent list even though the user did not run anything. This is also why the snapshot/delta distinction matters here: the consumer has side effects beyond idempotent reconciliation. (packages/kaval/src/inProcessPtyHost.ts:156)
- `F2` · minor · open — The new replay guarantee is a pty-host surface behavior needed by socket/remote consumers, but the reusable contract corpus still only asserts that `commandRun` yields when driven live. The late-subscriber replay test was added only to the identity-link-specific suite, so the real unix-socket daemon path is not directly pinned for the new contract behavior. (packages/kaval/src/contractCorpus.testlib.ts:303)

**claude** — Both CODEX findings were valid and are fixed. F1 (major): the snapshot-first commandRun replay shared the exact `{ command }` wire shape of a live OSC 633;E mark, so a late-subscribe/reconnect/restart re-fired the live-only `trackRecentAgent` recency bump (stamps Date.now()) and reordered the recent-agent MRU as if the user just ran the command — a streaming.md §2 violation (snapshot vs delta must be distinguishable when the consumer has non-idempotent side effects). Added a `replayed: boolean` discriminator to the commandRun stream output schema; the in-process source flags the replayed snapshot true and live marks false, threaded the flag through both channel bridges (kavalChannels.ts, local.ts) and the AwarenessSignals.commandRun channel via a new CommandRunSample type, and taught startAgentCommandSensor to seed detection (currentAgent + lastAgentCommand metadata) from BOTH frames while firing trackRecentAgent ONLY on a live mark. F2 (minor): moved the late-subscriber replay assertion into the shared contract corpus so both the identity link AND the spawned unix-socket daemon pin it, asserting replayed=false on the live mark and replayed=true on the replay. `just check` is green (typecheck across all packages + biome lint) and the kaval inProcessPtyHost/ptyHost/socketDaemon/coverage suites plus the terminal-workspace suite all pass.

- `F1` **fixed** — Agreed and fixed. The replay was wire-indistinguishable from a live mark, and the consumer's side effect is genuinely non-idempotent: packages/terminal-workspace/src/sensors.ts:397 called sink.trackRecentAgent(normalized) on every commandRun frame, and packages/server/src/activity.ts:98 stamps lastSeen: Date.now() into the recent-agents MRU — so any late subscribe/reconnect/restart would reorder recency spuriously. Added `replayed: boolean` to the commandRun output schema (packages/kaval/src/ptyHostSurface.ts:250); the in-process source now yields `{ command, replayed: true }` for the snapshot and `replayed: false` for live deltas (packages/kaval/src/inProcessPtyHost.ts:158-161). Threaded the flag through the AwarenessSignals.commandRun channel (new CommandRunSample type in sensors.ts) and both bridges (kavalChannels.ts, server/.../local.ts). startAgentCommandSensor now seeds detection (currentAgent + lastAgentCommand metadata) from both frames but gates trackRecentAgent behind `if (!replayed)` (sensors.ts). This matches streaming.md §2's explicit-discriminator guidance for consumers that accumulate side effects beyond idempotent reconciliation — and is why commandRun differs from foreground, whose consumer (snapshotSignals) is purely idempotent.
- `F2` **fixed** — Agreed and fixed. The replay-on-subscribe is a pty-host surface contract, not identity-link-specific. Moved the late-subscriber assertion into runContractCorpus (packages/kaval/src/contractCorpus.testlib.ts:329-340), so both the identity link (inProcessPtyHost.test.ts) and the real spawned unix-socket daemon (socketDaemon.test.ts) exercise it — confirmed by running both suites green. It also asserts replayed=false on the live mark and replayed=true on the replay frame. Kept the identity-link-specific test but tightened it to pin the early(false)->late(true) flag transition across two concurrent subscribers, which the corpus's single-stream version doesn't cover.

commit: `1fd65be5c7b2a8a00e23cec586a753dfefc25bbf`

### Round 2

**codex** — approved: `false`

F1 and F2 are fixed: `commandRun` frames now carry `replayed`, live-only recency is gated to `!replayed`, and the shared corpus pins replay behavior for both identity and socket-daemon paths. The fix introduced one new compatibility issue: the pty-host wire shape changed, but the contract version did not, so stale 3.2 daemons can still be adopted by a new client that expects `replayed`.

Findings:
- `F1` · major · resolved — Resolved. Replayed `commandRun` snapshots are now distinguishable from live marks, and `trackRecentAgent` only fires when `replayed` is false, so late subscriptions seed detection without reordering the recent-agent MRU. (packages/terminal-workspace/src/sensors.ts:400)
- `F2` · minor · resolved — Resolved. The shared contract corpus now asserts the live frame has `replayed: false` and a late subscriber gets a first replay frame with `replayed: true`, so both identity-link and spawned socket-daemon runs cover the new guarantee. (packages/kaval/src/contractCorpus.testlib.ts:321)
- `F3` · major · open — This round changed the `commandRun` stream output schema to require `replayed` at packages/kaval/src/ptyHostSurface.ts:256, but `PTY_HOST_CONTRACT_VERSION` is still `3.2`. The adoption gate compares the daemon-reported version against this constant, so an already-running 3.2 kaval daemon can be treated as compatible instead of recycled. A new server then either sees schema failures for old `{ command }` frames or reads `msg.replayed` as absent, and the replay fix is not reliably deployed across the daemon boundary. (packages/kaval/src/ptyHostSurface.ts:70)

_codex on the rebuttal:_ F1: accepted; the added `replayed` discriminator and `!replayed` recency gate address the non-idempotent side effect I flagged. F2: accepted; moving the assertion into the shared corpus covers the socket daemon path. New F3 is from this round's fix, not a pre-existing issue: changing `ptyHostSurface` without bumping `PTY_HOST_CONTRACT_VERSION` defeats the daemon compatibility gate.

**claude** — Codex's round-2 verdict confirmed F1 and F2 resolved from round 1, and raised a new F3: round 1 added a required `replayed` field to the `commandRun` stream output schema (packages/kaval/src/ptyHostSurface.ts:254-259) but left `PTY_HOST_CONTRACT_VERSION` at "3.2". The adoption gate at packages/server/src/ptyHost/connect.ts:75-84 compares the daemon-reported version against this constant, so a stale 3.2 daemon would match, be adopted, and serve bare `{ command }` frames the new schema rejects. I agreed and bumped the version to "3.3" (additive · minor, matching the 3.1/3.2 precedent where a survivor lacking the new shape is recycled rather than silently degraded), and documented the bump. Verified `just check` green (full typecheck + biome lint) and ran the kaval version + socket contract tests (6 passed) — the self-compatibility and corpus tests reference the constant symbolically, no literal "3.2" assertions needed updating.

- `F1` **fixed** — Resolved in round 1 (commit 1fd65be5). Codex confirmed resolved this round; no further action.
- `F2` **fixed** — Resolved in round 1 (commit 1fd65be5). Codex confirmed resolved this round; no further action.
- `F3` **fixed** — Agreed. The round-1 fix made `replayed` a required field on the commandRun stream output schema (ptyHostSurface.ts:256-258) but PTY_HOST_CONTRACT_VERSION stayed at '3.2'. The adoption gate (server/src/ptyHost/connect.ts:75-84) recycles a daemon only on a contract-version skew, so a still-running 3.2 daemon would be adopted and serve bare `{ command }` frames the new schema rejects. Bumped PTY_HOST_CONTRACT_VERSION to '3.3' — an additive · minor bump, consistent with the existing 3.1/3.2 precedent for a new required output shape (the project treats a survivor that can't produce the new field as wire-incompatible and forces a recycle, never a silent degrade). Updated the version doc comment to record the rationale. `just check` green; kaval ptyHostSurface + serveOverSocket tests pass (the self-compat test and corpus assertion both reference the constant symbolically, so no test literals needed changing).

commit: `991dbe66dd99f37aee80e9b076ee99c6af782728`

### Round 3

**codex** — approved: `true`

The current tree resolves all findings from this review session. F1 and F2 remain fixed from the prior round, and F3 is now fixed by bumping `PTY_HOST_CONTRACT_VERSION` to `3.3` with documentation explaining the required `commandRun.replayed` field and stale-daemon recycle behavior. `git status --short` showed no untracked files.

Findings:
- `F1` · major · resolved — Resolved. Replayed `commandRun` snapshots are distinguishable from live marks, and `trackRecentAgent` only fires when `replayed` is false, so late subscriptions seed detection without reordering the recent-agent MRU. (packages/terminal-workspace/src/sensors.ts:400)
- `F2` · minor · resolved — Resolved. The shared contract corpus asserts live `replayed: false` and late replay `replayed: true`, so both identity-link and spawned socket-daemon paths cover the behavior. (packages/kaval/src/contractCorpus.testlib.ts:321)
- `F3` · major · resolved — Resolved. The pty-host contract version is now `3.3`, so stale 3.2 daemons lacking the required `replayed` field are no longer treated as wire-compatible by the version gate. (packages/kaval/src/ptyHostSurface.ts:74)

_codex on the rebuttal:_ F1 and F2 were already accepted as fixed in the previous round and remain resolved. For F3, I accept CLAUDE's fix: the exported contract version is now `3.3`, and the adjacent version comment documents why a 3.2 survivor must be recycled after the required `commandRun.replayed` schema change.

srid:

✅ **All clear** — scoped to the #1558 change (`git diff d148eef85...HEAD`).

| Pass | Issues | Result |
|---|---|---|
| Rules | 0 | All 23 built-in + project rules checked, zero violations |
| Fact-check | 0 | No silent swallowing, no logic errors, no leaks, no inaccurate fallbacks |
| Elegance | — | Skipped (`--no-elegance`; the gauntlet's `/simplify` pass already ran over this tree) |

**Rules — key confirmations**
- `callback-fanout-guarded-at-funnel` — the `if (!replayed)` recency-bump gate sits inside the single command-run handler funnel; no unguarded path.
- `silent-handler-required-on-void-subscriptions` — every `.consume()` call carries a required `onError`.
- `integration-perf-hygiene` — no unbounded collections; `lastCommand` is a single overwritten field per entry.
- `no-untyped-escape-hatches` — `if (last !== undefined)` is proper type narrowing, not an escape hatch.

**Fact-check — verifications**
- `lastCommand` is assigned *before* `publish()`, so a late subscriber reading `getLastCommand` always sees the current value.
- An empty-string command (`633;E;` → `""`) is not `undefined`, so it *is* replayed — correct, "a command ran" matters even with no parsed name.
- Entry deleted before `getLastCommand`? Optional chaining returns `undefined`, replay safely skipped — no crash.
- The non-idempotent recent-agent MRU bump is gated exclusively on `!replayed` — the discriminator does exactly what the change claims.

srid:
## Evidence

This fix has **no visual diff on an easily-captured surface**: kolu's own Dock was never affected (kolu attaches at launch, so it already detected `codex`). The bug — and the fix — live in a **late/restarted/remote** sensor (pulam-web's lazily-attaching pulam, a restarted pulam). Staging *that* on screen needs the `codex` binary + auth + a multi-process late attach no e2e scenario reaches, so the proof here is a **before/after of the fixed behavior**, driven through the real pty-host surface client (`createInProcessPtyHost`) on identical installed deps — the source files are the only thing swapped between the two runs.

**Scenario:** a terminal runs `codex` (an npm shim — its OS process name is `node`, so it's recognizable *only* by its command line). A fresh sensor then attaches **after** the launch, and reads its first `commandRun` frame.

**❌ Before (pre-fix `d148eef` source) — the late sensor learns nothing:**
```
│ → terminal ran `codex` (OSC 633;E mark emitted).
│ → a NEW sensor attaches (late) and reads its first commandRun frame:
│      GOT: (nothing — timed out after 2.5s)
│   ❌ BEFORE: the late sensor learns no command → shows `codex` as a
│      plain `node` terminal.
```

**✅ After (this PR) — the late sensor catches up:**
```
│ → terminal ran `codex` (OSC 633;E mark emitted).
│ → a NEW sensor attaches (late) and reads its first commandRun frame:
│      GOT: { command: "codex", replayed: true }
│   ✅ FIXED: the late sensor catches up — resolves `codex`, not `node`.
│      (replayed:true → it seeds detection WITHOUT re-stamping the
│       recent-agent recency, which only a live mark may do.)
```

The same before/after is pinned permanently by the hermetic tests (both **red on the prior code, green after**):
- `ptyHost.test.ts` — `getLastCommand` retains the OSC 633;E mark (pre-fix: `getLastCommand is not a function`).
- `inProcessPtyHost.test.ts` + the shared `contractCorpus` — a late `commandRun` subscriber receives the command **snapshot-first**, `replayed:true` (pre-fix: the subscriber gets nothing and times out), and a live mark is `replayed:false`.

`replayed:true` on the replayed frame is the codex-review refinement: the consumer seeds agent detection from a replay but fires the **live-only** recent-agent recency bump (`Date.now()` into the MRU) only on a real live mark — so a reconnect/late-attach can't spuriously reorder recency.

srid:

The **x86_64-linux** lane ran on **`kolu-ci-8`** (`dev-x86-64-linux-08`) — commit `ff5099cd0`, **passed** ✓

- **Lane wall** (pipeline): **3m45s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 3m4s | ✓ |
| `ci::unit` | 57s | ✓ |
| `ci::smoke` | 54s | ✓ |
| `ci::nix` | 53s | ✓ |
| `ci::pnpm-hash-fresh` | 46s | ✓ |
| `ci::surface-example-build` | 37s | ✓ |
| `ci::surface-app-example-build` | 37s | ✓ |
| `ci::render` | 37s | ✓ |
| `ci::fmt` | 36s | ✓ |
| `ci::biome` | 36s | ✓ |
| `_ci-setup` | 32s | ✓ |
| `ci::install` | 8s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/ff5099c/timings.jsonl`; pool state is a live `flock` probe.</sub>

PR #1573, opened 2026-06-25
fix(kaval): defang reconnect-storm attach serialization

srid:
**PR1** of the [kaval memory plan](https://kolu.dev/atlas/kaval-memory-architecture.html) — the acute, kaval-internal half. No disk, no wire change, no reload-history regression.

## The bug

A WebSocket disconnect re-`attach()`es every terminal at once, aborting the in-flight attaches and reissuing them. Today each attach serializes the **whole** mirror (`ptyHost.ts` `attach()`), so a ~60-RPC burst allocates dozens of ~4 MB snapshots faster than GC reclaims — the measured **2.0 GB** (6 rounds) → **3.2 GB** (40 rounds) transient that drains over minutes on V8's idle reducer. That's the spike behind the production "kaval at **4740 MB**" report.

## The fix — two defenses in `attach()`, neither touching snapshot *content*

1. **Cancellable.** An attach whose `signal` is already aborted returns an empty snapshot and does **zero** `serialize()` — its subscriber (`subscribe()` returns an empty stream when aborted) is already gone. This is the re-issued half of the storm.
2. **Epoch-coalesced snapshot.** A per-`Entry` memo, set on the first `serialize()` and cleared in the data-publish callback the instant new bytes parse into the mirror, so a burst of attaches to one PTY between two output bytes shares a single serialize. Race-free: the cache-set and the publish-clear are on the same synchronous tap, and publish only fires from a later task.

Net: the storm drops from multi-GB to **O(live-terminal count)** full snapshots. Bounding each snapshot to a viewport — the part that *would* shrink what a reload restores — is deliberately **PR2**, where on-disk backfill makes it lossless.

## Why this is safe to ship alone
A full reload still serializes and restores the same mirror history as today; this PR only stops the storm from *duplicating and stacking* that work. Strict improvement, no regression.

## Design philosophy
- **No fallback/knob** — the cancellable branch is the correct response to an aborted request, not a degraded path; no override.
- **Electricity** — lives in kaval's pty-host (correct location); no new package, no agnostic helper folded into an app module.
- **Reuse the source of truth** — reuses the existing `serialize` + `Channel` publish path and the subscribe-before-serialize invariant; grepped first — there was no prior snapshot cache to duplicate.

## Tests
Three `ptyHost.test.ts` guards (all green, 33/33): an already-aborted attach does zero serialize, a 25-attach synchronous burst coalesces to **one** serialize, and a publish invalidates the memo (re-serializes).

## Docs
- `packages/kaval/README.md` — documents the two storm defenses on the attach contract.
- `docs/atlas/.../performance.mdx` — banks the previously-pending "cache the serialized snapshot per attach" item.
- Changelog `Fixed` entry.


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `92562ac8d0f6`

### Applied (4)
- `lowy-2` resize() mutates the serialized mirror but never invalidates snapshotCache — commit `6c5fb448a`
- `lowy-3` getScreenState() is a second, uncached serialize path for the same volatility — commit `98329d97b`
- `lowy-4` New tests cover abort + epoch coalescing but not resize invalidation — commit `982eb9ea9`
- `hickey-1` Mirror-serialization derivation is duplicated — attach() memoizes it, getScreenState() recomputes it — (uncommitted)

### Agreed — no change (4)
- `lowy-1` Electricity test: clean receptacle, defect is an internal wire (packages/kaval/src/ptyHost.ts:652-672)
- `hickey-2` Per-epoch snapshot memo complects value with time via a loose Entry field + manual invalidation (packages/kaval/src/ptyHost.ts:670)
- `hickey-3` The aborted-attach decision is evaluated in two layers (packages/kaval/src/ptyHost.ts:662)
- `hickey-4` Coalescing tests white-box the serialize implementation (packages/kaval/src/ptyHost.test.ts:696)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `92562ac8d0f6`

### Round 1

**codex** — approved: `false`

The implementation is close, but I would not approve yet. The main concern is that the new per-entry snapshot memo has no release path except future terminal mutation, so the fix trades duplicated transient serializations for long-lived retained serialized snapshots. The new tests also have timing windows that can flake under slow CI. I did not modify files or run write-oriented commands.

Findings:
- `F1` · major · open — `snapshotCache` retains the full serialized screen until the next data publish or resize. For idle terminals, a reconnect, attach, or `getScreenState()` call can pin a multi-MB serialized snapshot indefinitely, and `getScreenState()` now also populates the cache. That contradicts the documented "brief, bounded blip" behavior and adds persistent heap proportional to live terminal count. (packages/kaval/src/ptyHost.ts:661)
- `F2` · major · open — The invalidation test depends on catching a 400 ms window where `first` has arrived but `second` has not. On a loaded or slow CI worker, the child can print both before the polling assertion first observes the screen, making `waitFor()` time out even though the code works. (packages/kaval/src/ptyHost.test.ts:680)
- `F3` · minor · open — The note still cites `ptyHost.ts:639-647` for the attach changes, but the changed implementation has moved `attach()` to later lines. The generated public Atlas page now points readers at stale source coordinates. (docs/atlas/src/content/atlas/kaval-memory-architecture.mdx:83)
- `F4` · nit · open — Putting `<PrLink>` inside the heading changed the generated section id from `#pr1-defang-the-storm--kaval-internal-no-disk-no-wire-change` to the same id with a trailing hyphen, which breaks stable external anchors for the published Atlas note. (docs/atlas/src/content/atlas/kaval-memory-architecture.mdx:79)


**claude** — Made the retention cost honest in the note + README and disputed the release mechanism (F1), made the publish-epoch test write-driven instead of time-driven (F2), refreshed every stale `ptyHost.ts` cite label (F3), and moved the PR chip out of the PR1 heading to restore the anchor (F4).

- `F1` **partial** — Fixed the legitimate half: the prose oversold "brief, bounded blip" without naming the lingering memo. Made it honest in both the note (`docs/atlas/src/content/atlas/kaval-memory-architecture.mdx` "Net:" paragraph) and `packages/kaval/README.md` — the memo pins **one** serialized snapshot per terminal between mutations (`getScreenState` shares the slot), freed on the next data/resize or on teardown, **bounded by and strictly smaller than the mirror it shadows** (~4 MB filled snapshot vs the ~25 MB live cell buffer it serializes). So steady-state heap stays the same `O(live-terminal count)` class the mirror already occupies (the chronic cost PR2 targets) — a fraction added, not a new asymptotic leak; PR2's viewport bound takes it to ~tens of KB. **Disputed both suggested code changes.** (1) "Release after the current turn/flush window" *breaks* the coalescing: a reconnect storm's attaches arrive across **many event-loop turns** (one per re-issued wire message), so only an actual data-parse boundary reliably outlasts the burst for an idle terminal — a microtask/turn/timer release would fire mid-storm and re-introduce the N serializes PR1 exists to prevent (`packages/kaval/src/ptyHost.ts:668-688`, attach; the storm is defined as cross-turn in the note). The epoch grain is load-bearing, not a missing release. (2) "Leave `getScreenState()` as a fresh one-shot serialize" would re-introduce a **second** serialize site, breaking the "`snapshotOf` is the only place the mirror is serialized" invariant the design leans on for desync-safety (`packages/kaval/src/ptyHost.ts:660-663`); `getScreenState` is a one-shot query (so classified in `packages/surface/README.md`), not a hot poll, so it never repeatedly re-serializes. Cleanup paths already exist (every mutation + teardown/GC); the only gap is a long-idle terminal between interactions, which clears the instant the user next types.
- `F2` **fixed** — Rewrote the "re-serializes after new output ends the epoch" test (`packages/kaval/src/ptyHost.test.ts:676`) to be **write-driven, not time-driven**: spawn a long-lived `/bin/sh` reading stdin, then drive each chunk with `host.write(id, "echo first\n")` / `host.write(id, "echo second\n")` (the established pattern at `ptyHost.test.ts:393`). The second chunk now lands *strictly after* the epoch-1 attaches, so the 400 ms "first arrived, second not yet" window CODEX flagged is gone; dropped the now-pointless `!includes("second")` negative settle and the redundant `mockClear()`. Verified: 34/34 kaval tests pass.
- `F3` **fixed** — Refreshed every stale `ptyHost.ts` cite label in the note, not just the one flagged. `Cite` resolves against `master` (`rev = "master"` in `Cite.astro`), which this branch becomes, so labels must match post-merge lines. Updated: `:639-647 → :668-688` (attach, the flagged one), `:645 → :661` (the serialize), `:613 → :622` (decoded string), `:613-617 → :624-630` (publish callback, ×2), `:711-716 → :756-766` (resize), `:687 → :728` (getScreenState); `:52` (scrollback const) was unchanged. I kept exact line numbers (house style — the repo uses `<Cite label="file:NNN">` pervasively) rather than CODEX's drop-the-numbers alternative; they'll drift on future commits, but that's inherent and accepted by convention. Rebuilt + staged `docs/atlas/dist/`; `just atlas::check-sync` green.
- `F4` **fixed** — Moved `<PrLink pr={1573} />` out of the PR1 heading into a `**Shipped** <PrLink/>` line in the body. Confirmed the generated id reverted from `pr1-...-no-wire-change-` (stray trailing hyphen) to `pr1-...-no-wire-change`, matching the anchor published by #1572. Mechanism is as CODEX described: `PrLink` is a self-closing JSX node with no AST text child, so a trailing one slugs the preceding space to a `-`.
- **extra (gate fix)** — Biome's `noAssignInExpressions` failed the `just check` gate on the pre-existing `return (entry.snapshotCache ??= ...)` in `snapshotOf`; rewrote it to statement form (`packages/kaval/src/ptyHost.ts:660-663`), behavior identical. This +1 line shift is already accounted for in the F3 cite updates. `just check` (typecheck + biome) now green.

commit: `18e2974b68c89777788e7fec1715e81a7b56ff32`


### Round 2

**codex** — approved: `true`

All four prior findings are now resolved. F2, F3, and F4 were fixed in the tree. For F1, I accept Claude's dispute after re-reading the implementation and updated prose: the retained value is now documented honestly, bounded to one snapshot per live PTY, invalidated on data/resize/teardown, and the publish-epoch lifetime is part of the cross-turn reconnect-storm coalescing behavior this PR is trying to preserve.

Findings:
- `F1` · major · resolved — `snapshotCache` retains the serialized screen until the next mirror mutation; the original review flagged this as an implementation risk and as misleading against the docs' prior transient-only framing. (packages/kaval/src/ptyHost.ts:661)
- `F2` · major · resolved — The prior test used a timed `sleep 0.4` window that could flake if both outputs landed before the poll observed the first one. (packages/kaval/src/ptyHost.test.ts:676)
- `F3` · minor · resolved — The Atlas note had stale `ptyHost.ts` line labels after the implementation moved. (docs/atlas/src/content/atlas/kaval-memory-architecture.mdx:23)
- `F4` · nit · resolved — Putting the PR link inside the heading changed the generated section id and broke the stable PR1 anchor. (docs/atlas/src/content/atlas/kaval-memory-architecture.mdx:79)

_codex on the rebuttal:_ F1: I accept the dispute. My original code-change suggestion was too broad for this reconnect-storm shape: if attaches can arrive across multiple event-loop turns, a microtask/turn release can indeed expire mid-storm and reintroduce repeated serializes. A timer-based release would be a heuristic rather than the protocol boundary; the data/resize publish epoch is the real invalidation boundary. I also accept keeping `getScreenState()` on `snapshotOf()` because there is now one serialization site and the retained cost is documented. The remaining long-idle retention is bounded to one snapshot per live PTY and is explicitly called out as PR2's next target, so I am marking F1 resolved. F2: verified fixed by the write-driven test. F3: verified the stale cite labels were refreshed. F4: verified the generated heading id no longer has the trailing hyphen.

srid:

**No findings.** Ran Pass 1 (rules) and Pass 2 (fact-check) over `git diff 92562ac8…HEAD`; elegance pass skipped (`--no-elegance`, since `/simplify` already ran this gauntlet).

- **Rules** — the `if (signal?.aborted) return { snapshot: "", deltas }` short-circuit is *not* a caught-error-collapses-to-empty violation: there is no try/catch and no swallowed error — it's the correct response to an already-aborted request, and `""` is the documented wire-legal first-frame snapshot. No fallback path, no override knob, no global mutable state (the memo is per-`Entry`). No floating promise (`headless.write`'s arg is xterm's callback, not a promise).
- **Fact-check** — the invalidate-before-publish ordering in the `proc.onData` callback and the synchronous subscribe→`snapshotOf` pair preserve the subscribe-before-serialize partition (no gap/overlap); the memo is invalidated at every mirror mutator (publish + resize). Consistent with the lens (resize-invalidation) and codex (xhigh, consensus) passes that preceded this.

srid:
## Evidence

No visual surface (kaval is the PTY daemon), so the evidence is a **memory measurement that reproduces the reconnect storm** against the **real `createPtyHost`** — baseline (the merge-base, before this PR) vs patched (this PR's HEAD) — on a clean 32-core / 125 GB `pu` box.

**Method.** Spawn 12 terminals, each filled with 10 000 wide (213-col) SGR-heavy lines so every mirror is a costly ~2 MB serialize. Then run an 8-round storm: each round, per terminal, fire an **already-aborted** attach (the re-issued-but-dead half of a WebSocket disconnect) **plus** a live attach, all at once, holding the round's snapshots — and sample peak process RSS. Same harness, two checkouts (deps identical), two passes each.

| Build | idle RSS (filled) | **peak RSS under storm** | storm Δ over idle |
| --- | --- | --- | --- |
| **Baseline** (pre-PR1) | 609 MB | **2799 MB** | +2190 MB |
| **Baseline** pass 2 | 605 MB | **3048 MB** | +2442 MB |
| **Patched** (PR1) | 610 MB | **929 MB** | +320 MB |
| **Patched** pass 2 | 610 MB | **956 MB** | +346 MB |

**Read.**
- The storm peak drops from **~2.8–3.0 GB → ~0.93 GB** (≈ 3× lower), and the storm-induced delta from **+2.2–2.4 GB → ~+0.3 GB** (≈ 7× smaller). This is the production "kaval at 4740 MB" transient, reproduced and defanged.
- **Idle RSS is identical** (609 ↔ 610 MB): the patch changes nothing about steady-state or what a full reload restores — it only stops the storm from *duplicating and stacking* serializes. No reload-history regression, exactly as designed.
- The residual **+320 MB** is the bounded **O(live-terminal count)** cost the design predicts — under the storm each of the 12 terminals does *one* coalesced serialize instead of N×rounds. Taking each snapshot to ~tens of KB (the viewport bound) is PR2.

**Harness** (drives the real `createPtyHost`, runs on the box):
```
12 terms × 10000 lines @ 213 cols · one snapshot ≈ 2 MB
RESULT baselineMB=609 peakMB=2799 deltaMB=2190   # baseline
RESULT baselineMB=610 peakMB=929  deltaMB=320    # patched
RESULT baselineMB=605 peakMB=3048 deltaMB=2442   # baseline pass2
RESULT baselineMB=610 peakMB=956  deltaMB=346    # patched pass2
```

**CI:** all 32 required checks green on `x86_64-linux` (pool box `kolu-ci-6`) and `aarch64-darwin` (`rasam`).

srid:

The **x86_64-linux** lane ran on **`kolu-ci-6`** (`dev-x86-64-linux-03`) — commit `46958d1c1`, **passed** ✓

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 2m20s | ✓ |
| `ci::home-manager` | 1m35s | ✓ |
| `ci::nix` | 59s | ✓ |
| `ci::flake-check` | 58s | ✓ |
| `ci::smoke` | 48s | ✓ |
| `ci::unit` | 41s | ✓ |
| `_ci-setup` | 29s | ✓ |
| `ci::render` | 18s | ✓ |
| `ci::biome` | 16s | ✓ |
| `ci::fmt` | 15s | ✓ |
| `ci::install` | 11s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/46958d1/timings.jsonl`; pool state is a live `flock` probe.</sub>

PR #1582, opened 2026-06-26
pulam-tui: revert to a thin status/watch CLI; fleet → pulam-web

srid:
Now that **pulam-web** carries the rich, leave-it-on-a-second-monitor browser fleet dashboard, `pulam-tui` no longer needs to be a full-blown TUI. **This reverts it to a thin, scriptable, single-daemon CLI — kaval-tui's sibling — and drops the entire Bun + OpenTUI toolchain.** The multi-host fleet view moves wholesale to pulam-web.

The CLI is now two verbs against one daemon (flags go *after* the subcommand, like kaval-tui):

```
pulam-tui status [--json]        one-shot snapshot of every terminal
pulam-tui watch [<id>] [--json]  follow awareness live (NDJSON under --json)
```

`status` prints a columnify table (`ID · REPO·BRANCH · PR · AGENT · FOREGROUND · IDLE`) and exits; `watch` holds the link and prints one line per awareness change — bare `watch` follows the whole daemon, an optional id (short id or unique prefix) narrows to one — with a trailing `●` when a terminal is moving bytes right now. `--socket` / `--host <ssh>` / `--kaval` are unchanged; the remote dial still rides `@kolu/surface-nix-host`'s `dialAgentOnce`, exactly like `kaval-tui --host`.

### What leaves vs what stays

| Removed (was only there for the fleet board) | Kept / new |
| --- | --- |
| **Bun** runtime + **OpenTUI** (`@opentui/*`, the Zig `dlopen` renderer) | Runs under **Node/tsx** via `mkAgentTuiWrapper`, like kaval-tui |
| The whole **bun2nix** toolchain (`nix/bun2nix.nix`, the npins pin, `packages/pulam-tui/nix/`, the pnpm-workspace exclusion) | Plain Node packaging; `fetchPnpmDeps` FOD hash refreshed |
| The multi-host **`fleet`** subcommand (`--by`/`--no-local`/`--ssh-config`, the N-host mirror/aggregate, `render.ts`'s tone/fleet layer) | The single-daemon dial + the awareness read/`--json` |
| — | `watchAwareness` — live consume of the `awareness` collection + `activity` stream over `mirrorRemoteSurface` |

### Electricity is unaffected

The fleet board was pulam's "second consumer proves the surface is a receptacle" electricity proof. That proof now stands on **pulam-web** — a second consumer, and the better one (it exercises kolu's exact `websocketLink` + `surfaceClient` + reconcile leg). So the TUI no longer has to be rich to earn its keep; it reverts to a leaf. This change only **consumes** `@kolu/surface*` read-only, so no drishti mirror PR is needed.

### Tests

- `render.test.ts` — the pure projection (status table, watch line/JSON, removal sentinel, control-byte sanitization).
- `watch.test.ts` — `status` snapshot + live `watch` (upsert, removal, the activity live-dot) over a **real served unix socket** (`serveOverUnixSocket` + `connectPulam`), never `directLink`.

### Docs

New atlas note [`pulam-tui.mdx`](https://kolu.dev/atlas/pulam-tui.html) + a reframe of `pulam.mdx`; README / website / sibling-package READMEs synced; a changelog **Heads-up** that supersedes the within-cycle OpenTUI-dashboard + `fleet` entries; and the `dock-fleet-mirror` rule narrowed to two surfaces (Dock + pulam-web), since pulam-tui no longer paints agent state.

> The retired `arivu` codename in pulam-tui's internal symbols (`connectArivu*`, `ArivuContract`) is renamed to `pulam` along the way.

### Try it locally

```sh
nix run github:juspay/kolu/pulam-tui-bare#pulam-tui -- status
nix run github:juspay/kolu/pulam-tui-bare#pulam-tui -- watch
```

---

### Update — review + dogfooding

- **Review gauntlet** (lens · codex · simplify · code-police) ran and applied fixes — see the PR comments.
- **Merged latest `master`** (#1568).
- **Dogfooding found a real bug, now fixed** (`a383be486`): `status --host` provisions a *fresh ephemeral* pulam that publishes an unresolved seed per terminal and resolves it a beat later, so the one-shot read rendered every row blank. `status` now waits for the sensors to resolve (`settledSnapshot`, capped) — see the **Evidence** comment for the real recording, before/after.
- **Dropped the dead `ci::render` lane** (`7d33c8fe`): it ran the deleted OpenTUI/Bun render smoke; `ci::unit` now covers pulam-tui.


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `c5b9166feb42`

### Applied (6)
- `lowy-1` Stale pnpm-workspace exclusion outlives the deleted Bun boundary — commit `6ea10a22d`
- `lowy-2` connectPulamViaHost's onLog param is a wire with no plug — commit `7b85f6b57`
- `lowy-3` watchAwareness takes the whole Connection but uses only conn.client — asymmetric with its sibling — commit `ca7b4d519`
- `hickey-2` resolveOne complects pure prefix-resolution with process.exit — commit `d20965115`
- `hickey-3` Same concepts wear different names across the two JSON output shapes — commit `c994f5c38`
- `hickey-4` prChecks carries a dead 'none' arm and a redundant kind guard — commit `a9b031580`

### Agreed — no change (2)
- `lowy-4` PULAM_AGENT_DRVS_JSON is a Nix↔TS contract carried by a duplicated magic string (packages/pulam-tui/src/hostConnect.ts:45)
- `hickey-1` Terminal-id CLI vocabulary copy-pasted across kaval-tui and pulam-tui (packages/pulam-tui/src/render.ts:28-39,140-143)

srid:
### Round 1

**codex** — approved: `false`

The branch removes the OpenTUI/Bun implementation and replaces pulam-tui with a simpler Node CLI, but several integration edges are not ready for merge: the generated agent instructions are inconsistent, watch-mode errors can be hidden or crash noisily, and the user-facing JSON examples do not match the implementation.

Findings:
- `F1` · major · open — The APM-generated instructions are stale. The branch updates `.apm/instructions/dock-fleet-mirror.instructions.md` to remove pulam-tui from the Dock mirror contract, but the generated `AGENTS.md` still says changes under `packages/pulam-tui/**` participate in the three-surface Dock/pulam-tui/pulam-web contract. That leaves future agents enforcing the old rule against this new implementation, and also indicates the APM regeneration step was skipped. (AGENTS.md:138)
- `F2` · major · open — `pulam-tui watch` writes events directly to stdout without the error/backpressure handling used by `status`. A normal pipeline like `pulam-tui watch | head -1` can close stdout and trigger an unhandled `EPIPE`, and a slow consumer can cause unbounded buffering because `write()` returning false is ignored. This is especially visible because watch mode is intended for piping into tools like `jq`. (packages/pulam-tui/src/main.ts:250)
- `F3` · major · open — Unexpected watch stream failures can be silently swallowed. `watchAwareness` calls `mirrorRemoteSurface` without a `log` callback; that helper defaults logging to a no-op and catches upstream collection/activity stream failures, so non-abort errors can end the mirror without any stderr message or non-zero exit. For a CLI monitor, that makes real loss of connection or protocol failures look like a clean stop. (packages/pulam-tui/src/read.ts:99)
- `F4` · minor · open — The documented JSON filter is wrong: it checks `.agent.kind == "awaiting_user"`, but `awaiting_user` is an agent state, not a kind. The Atlas note also shows JSON keys `terminalId` and `active`, while the implementation emits `id` and `live`. Users copying these examples will not get the advertised output. (packages/pulam-tui/README.md:38)
- `F5` · minor · open — Initial watch rows can miss the live marker. `live` starts as an empty set, and the awareness collection snapshot can emit rows before the activity stream's first snapshot has populated that set. Since later activity-only changes do not emit watch rows, an already-active terminal may stay displayed without the live marker until some unrelated awareness change occurs. (packages/pulam-tui/src/read.ts:98)
- `F6` · nit · open — The pulam-tui TypeScript config still references deleted OpenTUI/Bun-era setup: `jsxImportSource` points at `@opentui/solid`, which is no longer a dependency, and the Bun test exclusion/comment refers to deleted files. This is stale configuration that will confuse the next JSX or test change in this package. (packages/pulam-tui/tsconfig.json:6)


**claude** — Agreed with all six findings and fixed each: synced the generated AGENTS.md, hardened `watch`'s stdout + upstream-error handling, seeded the live-dot snapshot, and corrected the doc JSON shapes + stale tsconfig (gate green: typecheck + biome + 24 tests + atlas check-sync).

- `F1` **fixed** — The `.apm` source was updated but the generated `AGENTS.md` (codex/opencode target) still carried the three-surface contract. Ran `just ai::apm` to regenerate; it produced the correct two-surface dock-fleet-mirror section but also bumped the unpinned apm-cli 0.21.0→0.22.0 (build-ID churn across all 6 AGENTS.md files, an `apm.lock.yaml` delta, and new untracked `.agents/skills/surface/` + `.claude/skills/surface/` installs — all unrelated to this PR). I kept only the scoped content fix: reverted the 5 sibling AGENTS.md files + `apm.lock.yaml`, removed the untracked skill dirs, and restored the root `AGENTS.md` build-ID/version header to its committed values so the diff is purely the dock-fleet-mirror section (`AGENTS.md:138`), matching the already-regenerated `.claude/rules/dock-fleet-mirror.md`. No CI gate enforces apm byte-sync, so this avoids a repo-wide tool-version bump in a pulam-tui change.
- `F2` **fixed** — `cmdWatch` wrote raw `process.stdout.write()`. Added `process.stdout.on("error", () => abort.abort())` so a closed pipe (`watch | head -1`, EPIPE) aborts the mirror and exits cleanly instead of an unhandled crash (`packages/pulam-tui/src/main.ts:240`), and routed every watch line through the existing backpressure-aware `writeOut` via a serialized `pending` promise chain (`main.ts:251`), so a slow consumer applies real backpressure (each line waits for `drain`) and the tail is flushed before exit — the same stance `status` already takes.
- `F3` **fixed** — `watchAwareness` passed no `log`, so `mirrorRemoteSurface` defaulted to a no-op and a dropped link / protocol error ended the mirror silently with exit 0. Added an optional `log` param threaded into the mirror (`packages/pulam-tui/src/read.ts:100,143`); `cmdWatch` passes a stderr sink and, after the mirror settles, fails non-zero when `!abort.signal.aborted` — i.e. the stream ended though the user never Ctrl+C'd and the consumer never hung up (`main.ts`, the trailing `if (!abort.signal.aborted) fail(...)`). User aborts and EPIPE hang-ups both set `aborted`, so they still exit 0.
- `F4` **fixed** — README example used `.agent.kind=="awaiting_user"` (a state, not a kind) → corrected to `.agent.state=="awaiting_user"` (`packages/pulam-tui/README.md:38`). The Atlas note used JSON keys `terminalId`/`active`, but `formatWatchJson` emits `{ id, live, ...AwarenessValue }` (`render.ts:233`) → corrected both the `jq` filter and the sample object to `id`/`live` (`docs/atlas/src/content/atlas/pulam-tui.mdx:77,79`), then regenerated `docs/atlas/dist/pulam-tui.html` and confirmed `just atlas::check-sync` (in-sync + idempotent). (Note: the note's `status` example at line 74 already correctly used `.agent.kind=="claude-code" and .agent.state=="awaiting_user"`, so it was left as-is.)
- `F5` **fixed** — `live` started empty, so the awareness keys-snapshot upserts could race ahead of the activity stream's first frame and paint an already-active terminal as idle until a later awareness change re-emitted it. `watchAwareness` now reads the activity stream's CURRENT snapshot (`firstFrameOrUndefined(activity.get({}))`) and seeds `live` BEFORE the mirror opens (`packages/pulam-tui/src/read.ts:113-122`); the mirror re-applies the same set on its first frame, so this only fills the startup gap. Added a test that stages an active terminal before `watch` starts and asserts its FIRST row already carries the live dot — no re-publish loop (`packages/pulam-tui/src/watch.test.ts`).
- `F6` **fixed** — Removed the stale `jsxImportSource: "@opentui/solid"` (no JSX, no opentui dep remains) and the `src/**/*.bun-test.tsx` exclusion + its Bun comment (no such files exist after the revert) from `packages/pulam-tui/tsconfig.json`. Typecheck still green.

commit: `57340b6e6d4399ca5aae3b56303409a9ce537e9d`


### Round 2

**codex** — approved: `true`

All six findings from the previous round are resolved by inspection of the current tree. I did not run tests because this review was explicitly read-only, but the code/docs changes line up with the fixes CLAUDE described.

Findings:
- `F1` · major · resolved — The generated agent instructions were stale and still included pulam-tui in the Dock fleet-mirror contract. (AGENTS.md:138)
- `F2` · major · resolved — `pulam-tui watch` previously wrote directly to stdout, so closed pipes could crash with EPIPE and slow consumers ignored stdout backpressure. (packages/pulam-tui/src/main.ts:246)
- `F3` · major · resolved — `watchAwareness` previously used `mirrorRemoteSurface` without a log sink, allowing upstream stream failures to disappear behind the helper's default no-op logger and clean-looking completion. (packages/pulam-tui/src/read.ts:100)
- `F4` · minor · resolved — The JSON examples did not match the implementation: one checked `.agent.kind == "awaiting_user"`, and the Atlas watch sample used `terminalId`/`active` instead of `id`/`live`. (packages/pulam-tui/README.md:38)
- `F5` · minor · resolved — Initial watch rows could miss the live marker because awareness snapshot upserts could arrive before the activity stream's first frame populated the live set. (packages/pulam-tui/src/read.ts:113)
- `F6` · nit · resolved — The pulam-tui tsconfig still referenced removed OpenTUI/Bun-era setup. (packages/pulam-tui/tsconfig.json:3)

_codex on the rebuttal:_ No disputes were raised. CLAUDE agreed with all findings; I verified the current tree addresses F1-F6 and did not find a regression introduced by this round's fixes.

srid:

Two passes over the change (`--no-elegance`, since `/simplify` already ran this gauntlet). Both findings applied; one fact-check finding skipped with rationale.

**Pass 1 — rule checklist** · 1 violation, fixed
- **`callback-fanout-guarded-at-funnel`** (`read.ts` · `watchAwareness`) — the `onUpsert`/`onRemove` handlers were handed to `mirrorRemoteSurface` unguarded, so a throwing handler could escape into the mirror's internal loop and wedge the watch. **Fixed**: wrapped both at the funnel in `try/catch` → the `log` sink, containing a bad frame instead of breaking the stream. (Every other rule passed; `new-package-has-readme` confirmed pulam-tui's README documents its single-daemon boundary vs pulam-web.)

**Pass 2 — fact-check** · 2 findings (1 fixed, 1 skipped)
- **Resource hygiene in `cmdWatch`** (`main.ts`) — the `watch <id>` prefix resolution ran *before* the `try/finally` that disposes the link, so a thrown snapshot read wouldn't dispose. **Fixed**: moved the resolution inside the `try`, so every error path disposes — matching `cmdStatus`.
- **Activity-seed `undefined` → empty live set** (`read.ts`) — *skipped*. The seed is a best-effort initial live-dot annotation; a genuinely broken `activity` stream is surfaced authoritatively by the mirror's own `log` sink + the un-aborted-settle `fail()` (the mirror's first activity frame also immediately re-applies the snapshot). Routing seed-`undefined` through that same `log` would risk mislabeling a later, unrelated settle, and Pass 1's dedicated `caught-error-must-not-collapse-to-empty` check cleared this path.

Gate after fixes: `just check` (tsc + biome) green · 24 unit/integration tests green · atlas `check-sync` idempotent.

srid:
## Evidence

> Replacing my earlier "evidence" comment — that was the shipped formatters' output, **not** a recording, and posting it as evidence was wrong. Here's an actual terminal recording of the real CLI against a real, Nix-provisioned daemon.

A real [asciinema](https://asciinema.org) recording of `nix run .#pulam-tui -- status/watch --host localhost` against the live daemon on this box (6 terminals across several kolu worktrees):

pulam-tui status + watch, recorded against a live --host pulam daemon

### …which surfaced (and fixed) a real bug

The first capture showed **every awareness field blank** — `status --host` provisions a *fresh ephemeral* pulam, which publishes an unresolved seed per terminal and fills it in a beat later as its sensors run; the one-shot first-frame read caught the seeds. Fixed in **a383be486** (`settledSnapshot` waits for the daemon's sensors to resolve, capped). After the fix, `status` shows real, resolved data:

```text
$ pulam-tui status --host localhost
ID        REPO·BRANCH                    PR            AGENT             FOREGROUND  IDLE
2e123874  claude-dump·master             —             claude · waiting  claude        1s
4f0e57af  kolu·pulam-tui-bare            #1582 open ✓  claude · working  claude        1s
66880cdc  kolu·kaval-transcript          #1577 open ✓  claude · waiting  claude        1s
ac8df59a  kolu·pulamweb-live-git-status  —             —                 bash           —
def96765  kolu·pulam-tui-bare            #1582 open ✓  —                 bash           —
f9b21f1f  kolu·pulamweb-live-git-status  —             claude · waiting  claude        1s
```

Real repo·branch, PR + checks (`#1582 open ✓`), agent state, and foreground — bare shells correctly show `bash` with no agent. `watch --host localhost` (in the GIF) streams those fields landing live, with the `●` live-bytes dot on terminals producing output.

### Plus the deterministic wire proof

`packages/pulam-tui/src/watch.test.ts` exercises `status` (incl. the new `settledSnapshot`) and `watch` over a **real served unix socket** (`serveOverUnixSocket` + `connectPulam`, not `directLink`): a resolution delta is awaited, the seed-only case caps at `maxMs`, a live upsert + removal land, and the `activity` live-dot propagates. 26 tests green; `just check` green.

> pulam-tui is a standalone CLI, not a kolu browser surface, so the Playwright scenario / mobile-touch axes don't apply — a terminal recording is the equivalent artifact.

srid:

The **x86_64-linux** lane ran on **`kolu-ci-7`** (`dev-x86-64-linux-05`) — commit `a53c2f299`, **failed** ✗

- **Lane wall** (pipeline): **4m23s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 3m40s | ✓ |
| `ci::pnpm-hash-fresh` | 3m9s | ✓ |
| `ci::home-manager` | 3m2s | ✓ |
| `ci::flake-check` | 2m23s | ✓ |
| `ci::nix` | 2m19s | ✓ |
| `ci::smoke` | 2m15s | ✓ |
| `_ci-setup` | 34s | ✓ |
| `ci::unit` | 31s | ✓ |
| `ci::render` | 14s | ✗ (exit 1) |
| `ci::biome` | 14s | ✓ |
| `ci::fmt` | 13s | ✓ |
| `ci::install` | 8s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/a53c2f2/timings.jsonl`; pool state is a live `flock` probe.</sub>

PR #1588, opened 2026-06-26
docs(atlas): fix R8 (awareness-only; fs/git is procedure+pulse → R9) + surface-live-data note

srid:
Fixes the R8 plan after a fresh implementing agent caught a real error — and adds the explainer note that makes the distinction legible.

## The error (the fresh agent was right)

My R8 plan's fs/git half said: *"relocate `koluSurface`'s value-bearing `gitStatus`/`fsListAll` streams onto `terminalWorkspaceSurface` via `pollOnEvent`."* That **contradicts a deliberate design choice already in the code.** `terminalWorkspaceSurface` serves fs/git as **procedures + a `{seq}` pulse**, not value-bearing streams — `surface.ts:88-105`:

> *"The two contract shapes are deliberately different (R8's remote kolu re-queries procedures rather than streaming full diffs over the wire)… `fs.*`/`git.*` PROCEDURES are the Code tab's raw reads… `subscribeRepoChange`/`subscribeFileChange` are WATCHER STREAMS that pulse on each live change."*

Following the note literally would re-introduce the exact shape the surface was built to avoid.

## The fix

- **R8 is the awareness half only** — clean and self-contained (the `awareness` *collection* already exists on `terminalWorkspaceSurface`, no shape change), and it unblocks the user-visible **R9a** state-desync. This is the real next move.
- **fs/git rides R9.** Making kolu read the shared surface's fs/git means rewiring the Code tab from value-bearing streams (`CodeTab.tsx:314-373`) to **procedure + pulse-then-requery** — a large client change the surface ties to mirroring-whole. Not a "move the streams" edit.
- **R-pulamweb-4** now consumes `terminalWorkspaceSurface`'s **existing** `git.getStatus` procedure + `subscribeRepoChange` pulse (re-query on pulse) — adding *nothing* to the surface (the purest form of "consume it as-is"); it needs only R-pulamweb-3.

## The explainer note (`surface-live-data`)

A new `reference` note (filed under Electricity + Reference) explains, pictorially:
- **value-bearing stream** (server pushes the whole value) vs **procedure + pulse-then-requery** (server pings "ask again", consumer pulls),
- that they carry the same information at different wire cost,
- the framework `pollOnEvent` helper that *builds* the first from the second,
- and why kolu uses both: `koluSurface` (in-process → value-bearing) vs `terminalWorkspaceSurface` (remote/ssh → procedure + pulse).

R8/R9 and R-pulamweb-4 link to it.

## Still open (flagged, not done here)
`pulam-web.mdx` has a few softer value-bearing-era references left (the "Design B" design-fork callout, a couple of prose mentions) — a consistency pass for a follow-up.


srid:
## The R8 page itself (`remote-terminals.mdx`) still carries the old whole-surface framing

The pulam sweep in `f70553c5` is good — but it didn't touch `remote-terminals.mdx`, which is the R8 page. Seven spots there still describe R8 as *"serve the whole surface / delete `koluSurface`"* — the exact framing this PR moved away from:

| Line | Stale text | Fix |
|---|---|---|
| 3 (frontmatter `description`) | "R8 (kolu serves one shared surface)" | "R8 (kolu serves the shared awareness)" |
| 7 (`updated`) | `2026-06-24` | `2026-06-26` (content materially changed) |
| 89 (SVG caption) | "…awareness + fs/git, composed into one surface **(R8)**" | split it: awareness = R8, fs/git = R9 — and check the SVG art too if it labels the compose box R8 |
| 106 ("Already paid for") | "R8 makes kolu serve **the one shared surface**" | "…serve the shared **awareness**" |
| 126 (PhaseTree, R-pulamweb-4 row) | `m: "after R8 · proves the remote leg"` | "after R-pulamweb-3 …" — `pulam-web.mdx` already set R-pulamweb-4 `needs={["R-pulamweb-3"]}`, so this row now **contradicts** it |
| 182 (R6 section) | "the shared *surface* — that is **R8**" | "awareness in R8, fs/git in R9" |
| 238 (History, 2026-06-26 entry) | R8 "deleting `koluSurface`"; "R-pulamweb-4 moves after it" | both now false — correct it **and** add a new History entry recording the awareness-only re-scope |

### Two small accuracy items in the new `surface-live-data` note
- **Pattern 2 code uses an invented API.** There's no `.use(...).on(cb)` on the stream hook anywhere in the repo (it returns a reactive accessor with `.pending()`/`.error()`), and procedures are called at the link root (`client.surface.kolu.…`), not via a scoped `app.rpc.git.getStatus`. In a *reference* note that can misdirect the R9 implementer — hedge it as schematic, or align to the real reactive shape.
- **"awareness … a collection, identical on both surfaces"** is slightly loose: `koluSurface`'s is `terminalMetadata` (awareness **+** kolu-authored `location`/theme/layout/…, a superset). Only the awareness *slice* is identical. Tighten to "the same *kind* on both."

(All code citations in the revision verified accurate against source. One trivial nit: the `surface.ts:88-105` citation is really the comment at `89-109`, phrase at `:93-94`.)

PR #1590, opened 2026-06-26
docs(blog): add a primitives table + spec diagram to the surface-framework post

srid:
The [surface-framework post](https://kolu.dev/blog/surface-framework/) leaned entirely on prose for the two things a reader actually scans for: the catalog of the five primitives, and the shape of the *one-spec-derived-three-ways* architecture. Every other diagram-bearing post on the blog (`orpc-over-ssh`, `xtermjs-perf`) carries an inline SVG and, where it helps, a table. This brings surface-framework up to that bar on both fronts.

## What changed

**A primitives table** in the *Five shapes* section — a one-screen recap of Cell / Collection / Stream / Event / Procedure: the question each one answers, whether it has a current value, who owns that value, and what Kolu uses it for. The columns *are* the invariants the type system holds, so the table doubles as the "why these can't collapse into one primitive" argument the prose makes right after it.

**A spec diagram** in the *One spec, three places* section — a hand-authored SVG, drawn in the same CSS-variable palette and monospace style as the existing diagrams, showing a single `defineSurface(…)` spec deriving three artifacts: `surface.contract` (the typed oRPC contract), `implementSurface` (server router + ctx), and `surfaceClient` (the client `.use()` hooks). Each derivation gets its own accent colour; the caption restates that the spec is the only source of truth.

**Rename `.md` → `.mdx`** so the post can carry inline SVG, matching every other diagram-bearing post. The slug (`/blog/surface-framework/`) is unchanged.

## Verification

`just website::build` passes; the built HTML contains both the `<table>` and the SVG. Rendered both in `astro preview` — screenshots below.

### DIFF
--- a/website/src/content/blog/surface-framework.mdx
+++ b/website/src/content/blog/surface-framework.mdx
@@ -21,6 +21,16 @@ So I pulled it out into a package, [`@kolu/surface`](https://github.com/juspay/k
 
 A `Cell` is a single current value. What's the theme right now? One of them, you can read it, you can set it. A `Collection` is the same idea but keyed: what's the current value for each id? Many of them, you set any one by key. A `Stream` is read-only and derived from something the server doesn't own — git, the filesystem, the network. What's the live output for this input? You can read it. You can't set it. An `Event` has no current value at all. Has the thing happened yet? You don't render it; you handle it when it fires. And a `Procedure` is the boring one: run a side-effect on the server, get an answer back. A plain RPC, living in the same namespace as the rest.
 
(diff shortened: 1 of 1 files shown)


srid:
## Evidence

Both additions rendered via `astro preview`, in the blog's dark theme.

**Spec diagram** — `One spec, three places` section: one `defineSurface(…)` deriving the typed contract, server router, and client hooks.

surface spec diagram

**Primitives table** — `Five shapes` section: Cell / Collection / Stream / Event / Procedure, with the question each answers, its current value, who owns it, and Kolu's use.

surface primitives table

PR #1591, opened 2026-06-26
Recover a frozen terminal when kaval sheds its slow tab (R5)

srid:
**A terminal that flooded output used to freeze.** kaval bounds how much un-drawn output it buffers per attach subscriber and *drops* a consumer that lags — but the drop ended the stream identically to a PTY exit, so kolu-server's web tier treated it as terminal and froze the client's scrollback even though the shell was alive and the same terminal kept working everywhere else.

R5 makes the drop **distinguishable**: the attach contract (`ptyHostSurface`, bumped to **3.4**) now carries a typed `overflow` *control* frame — emitted as the stream's last frame when the host sheds a lagging subscriber — so the consumer **re-attaches for a fresh snapshot** instead of mistaking the drop for a dead terminal. This is the [herdr vs kolu](https://kolu.dev/atlas/herdr-vs-kolu.html) "control ⟂ data channel" gap, closed on this seam; it hardens the attach contract for *both* transports, a prerequisite for the remote reconnect R8/R9 lean on.

### The seam and its consumers

```
kaval Channel (per-subscriber bound)
  └─ drop on overflow ─▶ terminalAttach yields { kind: "overflow" }  ← contract 3.4
        ├─ kolu-server/local.ts  → re-attach, reset-then-snapshot (web freeze fixed)
        ├─ kaval-tui/attach.ts   → skip the dataless frame, existing loop re-attaches
        └─ pulam                 → ignored (activity detection is best-effort)
```

- **Overflow is now per-subscriber** (`Channel.subscribe(signal, onOverflow)`) rather than a latent channel-global callback — each subscriber buffers independently, so only the one that overflowed fires.
- The web tier's recovery loop is extracted to `reattachingDeltas` (pure, over an `open` callback) and **unit-tested with scripted iterators** — graceful end, single/repeated re-attach, `NOT_FOUND`-during-re-attach, and error propagation.
- A re-attach prefixes the fresh snapshot with a full terminal reset (`ESC c`), the byte-stream equivalent of the `terminal.reset()` the browser already runs in its own reconnect `onRetry`, so the repaint *replaces* stale rows instead of double-painting onto them.

### Reproduced first

A 1-deep data queue (`dataMaxQueue`, a test-injected capacity option, defaulting to the real value) drives the drop deterministically over the identity link — `inProcessPtyHost.test.ts` is **red before the emission** (the stream just ends; no overflow frame) and green after. The same source backs the socket daemon.

> *No fallback or runtime knob:* the drop already discards the buffered sliver by design (bounded memory beats unbounded fidelity); the new frame just *names* it so the consumer can recover. You lose the scrollback between the snapshot and the drop; the live screen comes right back.

Part of [#951 — remote terminals over SSH](https://kolu.dev/atlas/remote-terminals.html) (R5). Plan-of-record updated in the same change.


srid:

**Consensus** · 1 round · 0 fixes applied · 2 observations dropped

Independent review: lowy raised 2 findings, hickey raised 0. Both reached per-finding consensus to **drop** — the structure is sound as-is.

### Agreed observations (no change)

| Origin | Finding | Disposition | Why |
| --- | --- | --- | --- |
| lowy | Overflow-recovery "hand-rolled in multiple consumers" | **drop** | The typed `overflow` frame (`ptyHostSurface.ts`, minted once in `inProcessPtyHost.ts`) **is** the receptacle for the one volatility this PR introduces ("dropped, not exited"). The three consumers carry genuinely divergent *policy*, not a duplicated mechanism: `pulam` ignores it, `reattachingDeltas` re-attaches with an RIS (`\x1bc`) prefix + NOT_FOUND-on-reopen discrimination, and kaval-tui folds it into a **pre-existing** reconnect loop (the `for(;;)`, `delay(150)`, inventory pre-flight, and `resolveGone` all predate this PR — kaval-tui gained only +6 lines: one `break`). The reset bytes even differ (`\x1bc` web vs `\x1b[H\x1b[2J` tui), confirming the recovery is not one shared mechanism. Forcing kaval-tui's tty-painting loop through a string-yielding combinator would *fragment* its single reconnect loop or complect it with callbacks. |
| lowy | 150ms re-attach pause "duplicated across packages" | **drop** | The two pauses are independently scoped and coincidentally equal: kaval-tui's `delay(150)` is the **general** reconnect cadence (it backs off after every stream end — PTY exit, abort, *and* overflow — and predates this PR), while `REATTACH_PAUSE_MS` is overflow-re-open-specific. No clean shared owner: a client re-dial cadence is a consumer-UX concern, not a host-contract concern, so hoisting a timing constant into `kaval` would be a *worse* boundary than two local constants. |

_The lenses converged in one round; nothing to apply._

srid:
## Codex ⇄ Claude debate — consensus (3 rounds, reasoning effort: xhigh)

### Round 1

**codex** — approved: `false`

The change is close in shape, but I found a protocol-versioning bug that can break mixed-version clients, plus two smaller documentation/test-coverage issues around the remote-terminal overflow path. I did not modify files.

Findings:
- `F1` · major · open — The new overflow frame is documented and versioned as an additive 3.4 minor change, but it is a new discriminant on an existing stream that older 3.3 clients can actually receive. The shared compatibility rule accepts newer-minor servers for older-minor clients, so a 3.3 kaval-tui/pulam/server client can handshake with a 3.4 daemon and then fail schema validation or mishandle the first `{ kind: "overflow" }` frame, because its schema only knows `snapshot` and `delta`. Conversely, the comment says 3.3 survivors are wire-compatible/no forced recycle, but a 3.4 client rejects a 3.3 daemon under the current predicate. (packages/kaval/src/ptyHostSurface.ts:74)
- `F2` · minor · open — The overflow-frame test only exercises the in-process direct link. This change is explicitly a wire-contract change consumed by kaval-tui/pulam over the socket path, but the socket/oRPC round trip is not covered, so the test suite would miss serialization/validation mistakes in the real attach transport. (packages/kaval/src/inProcessPtyHost.test.ts:170)
- `F3` · nit · open — The updated intro says the remaining roadmap is only `R8 · R9`, but the same sentence still describes multiplexing the canvas and the phase tree/sections below still show `R10` as remaining. The note now contradicts itself on what work is left. (docs/atlas/src/content/atlas/remote-terminals.mdx:18)


**claude** — Bumped the pty-host contract to a breaking `4.0` for the new emitted `overflow` variant, added a real-socket overflow test, and fixed the Atlas roadmap parenthetical; Atlas dist rebuilt, lint + typecheck + the full kaval suite green.

- `F1` **fixed** — Codex is right, and the reason is precise: the previous bumps (3.1–3.3) were *additive* in the only direction `isContractVersionCompatible` waves through. That predicate lets an OLDER client accept a NEWER-minor daemon (`reported.minor >= expected.minor`, same major — `define.ts:783-797`). For a field-add that's safe (the old client strips the unknown key); for a new *emitted* discriminated-union variant on a stream the old client already consumes it is NOT — the old client either fails the `z.discriminatedUnion("kind", …)` parse or writes a dataless frame. Every prior bump's breaking direction was new-client/old-daemon, which the predicate already recycles; this one's breaking direction is old-client/new-daemon (a stale `kaval-tui`/`pulam-tui` binary meeting a freshly-upgraded daemon), which a minor bump silently allows. So this is a major change: `PTY_HOST_CONTRACT_VERSION = "4.0"` (`ptyHostSurface.ts:79`), making a 3.x peer on EITHER side a clean skew (recycled / refused with an honest restart message) instead of a silent mis-parse. Rewrote the version comment (`ptyHostSurface.ts:74-78`) and the inline schema comment (`ptyHostSurface.ts:164-166`) to explain the inverted-direction reasoning, and repointed every "3.4" reference: `kaval/README.md:74`, `pulam/src/daemon.ts:266`, `kaval-tui/src/attach.ts:282`, and the Atlas note (`remote-terminals.mdx:150,243`, dist rebuilt). No drishti PR: the constant lives in `packages/kaval`, not `packages/surface` — the surface API is untouched.
- `F2` **fixed** — Added a socket-level overflow test in `serveOverSocket.test.ts` (the harness Codex named — it serves an in-process host over a REAL unix socket via `unixSocketLink`, the exact kaval-tui/pulam/server transport). Threaded `dataMaxQueue` through `makeRouter` so a dedicated listener binds a 1-deep queue. One correction to Codex's framing: a small `dataMaxQueue` alone does NOT trip the drop over the socket — a few writes drain straight into the kernel socket buffer, so the per-subscriber channel never overflows. The test therefore uses a CONTINUOUS flood (`yes OVFLINE`) to fill the kernel buffer and backpressure the server's socket write before the 1-deep channel overflows, plus a SEPARATE control connection for `spawn`/`write`/`getScreenText`/`kill` (a saturated attach stream backpressures its whole connection, so a shared-socket `getScreenText` would deadlock behind the very stream we refuse to read). Verified: the `{ kind: "overflow" }` frame crosses the socket; full kaval suite 121/121 green.
- `F3` **fixed** — `remote-terminals.mdx:18` now reads "(R8 · R9 · R10)", matching the frontmatter ("the remaining work R8 · R9 · R10"), the same sentence's canvas clause, and the phase tree (`R10 · canvas multiplex + validate`, line 136 / §R10 line 235). Atlas dist rebuilt and `just atlas::check-sync` is green.

commit: `edf492dcfe19a8f6995b4546010aff2f3ee80418`


### Round 2

**codex** — approved: `false`

Claude fixed the functional protocol skew and added the real socket coverage I asked for. The only remaining item is a stale in-code version reference that still says the overflow frame arrived in contract 3.4, which contradicts the now-correct 4.0 major bump.

Findings:
- `F1` · nit · open — The functional versioning fix is correct: `PTY_HOST_CONTRACT_VERSION` is now `4.0`, and the attach schema comment explains why a new emitted union variant is a breaking change. One helper comment still says "Before contract 3.4", though, which preserves the stale minor-version story this finding was about. (packages/server/src/terminalEndpoint/reattachingDeltas.ts:8)
- `F2` · minor · resolved — Previously, the overflow frame was only covered through the in-process direct link. The current tree now adds a real Unix-socket/oRPC round-trip test with a dedicated 1-deep queue host, separate control and attach connections, and a continuous flood to prove `{ kind: "overflow" }` crosses the actual transport. (packages/kaval/src/serveOverSocket.test.ts:119)
- `F3` · nit · resolved — The roadmap parenthetical now says `R8 · R9 · R10`, matching the canvas clause and the phase tree below. (docs/atlas/src/content/atlas/remote-terminals.mdx:18)

_codex on the rebuttal:_ F1: I agree with the major-bump fix and with the no-drishti reasoning: this is in `packages/kaval`, not the shared `packages/surface*` API. The functional skew bug is addressed, but one stale `contract 3.4` reference remains in `reattachingDeltas.ts`, so I am keeping F1 open only as a documentation cleanup. F2: I accept Claude's correction that a tiny queue alone is not enough over the socket because the kernel buffer drains initial writes; the continuous flood plus separate control connection is the right shape, and the new socket-level test resolves the coverage gap. F3: resolved; the Atlas parenthetical now includes R10.


**claude** — Cleaned up the one stale `contract 3.4` reference codex kept open; F2 and F3 were already accepted as resolved last round.

- `F1` **fixed** — Repointed the last stale minor-version reference: the helper doc comment in `reattachingDeltas.ts:7-8` said "Before contract 3.4 that drop was indistinguishable…", which contradicted the now-correct breaking `4.0` major bump. Changed it to "Before contract 4.0" — `grep -n "3\.4"` over the file is now empty (exit 1). This is purely a documentation cleanup; the functional skew was already fixed in round 1. `just check` (typecheck + biome lint) green.
- `F2` **fixed** — Resolved last round (codex marked it resolved): `serveOverSocket.test.ts:119` exercises the real unix-socket/oRPC round trip with a dedicated 1-deep queue host, separate control + attach connections, and a continuous flood proving `{ kind: "overflow" }` crosses the transport. No change this round.
- `F3` **fixed** — Resolved last round (codex marked it resolved): `remote-terminals.mdx:18` roadmap parenthetical reads `R8 · R9 · R10`, matching the canvas clause and the phase tree. No change this round.

commit: `7e09731d34f6e3914f1f73e0ba9ac80d957ea899`


### Round 3

**codex** — approved: `true`

All previously raised findings are resolved. I verified the working tree is clean, the pty-host contract and attach variant docs are on `4.0`, the last stale `3.4` helper comment now says `Before contract 4.0`, the socket-level overflow test remains in place, and the Atlas roadmap parenthetical still includes R10.

Findings:
- `F1` · major · resolved — The original protocol-versioning issue is resolved: `PTY_HOST_CONTRACT_VERSION` is `4.0`, the emitted `overflow` union variant is documented as breaking, and the stale helper comment now says `Before contract 4.0` instead of `3.4`. (packages/server/src/terminalEndpoint/reattachingDeltas.ts:8)
- `F2` · minor · resolved — The socket-transport coverage gap is resolved by the real Unix-socket/oRPC overflow test with a dedicated 1-deep queue host, separate control/attach connections, and a continuous flood to force the overflow frame across the actual transport. (packages/kaval/src/serveOverSocket.test.ts:119)
- `F3` · nit · resolved — The Atlas roadmap inconsistency is resolved: the remaining roadmap parenthetical now reads `R8 · R9 · R10`, matching the canvas clause and phase tree. (docs/atlas/src/content/atlas/remote-terminals.mdx:18)

_codex on the rebuttal:_ F1: Accepted. The stale `contract 3.4` reference in `reattachingDeltas.ts` is gone and the comment now matches the 4.0 breaking contract. F2 and F3 were already resolved last round and remain resolved. No new regression from this round's change.

srid:

**Pass 1 (rule checklist): clean.** No violations across the built-in and project rules. Spot-confirmed the load-bearing ones: the `NOT_FOUND`-on-re-attach `return` in `reattachingDeltas.ts` is a legitimate clean-end (the PTY's exit surfaces via the separate `exit` stream, not the deltas) — not a swallowed error; `dataMaxQueue` is a test seam (`undefined` in every production caller), not an override knob; the per-subscriber `onOverflow` fires exactly once from a single site; comments match the code.

**Pass 2 (fact-check): two findings, both fixed.**

| # | Finding | Fix |
| --- | --- | --- |
| 1 | **An aborted attach subscriber could be mis-signalled as an overflow.** `onAbort` queued `CLOSE` but left the sub in the live set, so a `publish()` racing the abort could trip the cap and fire `onOverflow` — emitting a spurious `overflow` frame for a clean abort, *inverting the exact exit-vs-overflow distinction R5 adds* (deterministic at `maxQueue=1`). | Remove the sub from the live set in `onAbort` before delivering `CLOSE`, mirroring the drop branch — a post-abort publish never reaches it. Regression test pins it (`9e99441c3`). |
| 2 | **`drainForOverflow` leaked a per-pull timeout.** Each `Promise.race` armed an 8s timer never cleared when `iter.next()` won, leaving thousands of dangling timers (at `maxPulls=5000`) that later reject orphaned promises. | Clear it via `.finally`, matching `nextFrame`'s hygiene (`7f872752f`). |

Pass 3 (elegance) skipped — `/simplify` already ran this gauntlet.

srid:
## Evidence

**On-screen effect:** a terminal that was *frozen* (scrollback stuck, no new output, the tab looking as if the shell had died) now **re-attaches on its own and repaints from a fresh snapshot** the moment kaval sheds it as a slow subscriber — the shell was alive all along.

**Why there's no live video here.** The trigger is a backpressure race: kaval drops a view only once it lags **>10,000 un-drawn chunks** behind a single PTY (the per-subscriber `Channel` bound). That bound is a constructor seam deliberately *not* exposed as a runtime knob (the no-knobs design rule), so it can't be shrunk in a live kolu to force the drop, and getting a real fast browser tab to fall 10k chunks behind one PTY in a controlled capture isn't reliably reproducible. So the fixed behavior is proven by deterministic tests that exercise the **exact wire frames and re-attach legs** — including over the real unix socket — rather than a staged clip.

### The fixed behavior, demonstrated by tests

```
### kaval socket-transport overflow test (the overflow frame crosses the REAL unix socket)
✓ servePtyHostOverUnixSocket — `overflow` frame crosses the socket
  > yields a typed `overflow` frame to a slow subscriber over the real socket

### in-process contract repro (was RED before the emission — toggling the emit line fails it)
✓ terminalAttach yields a typed `overflow` frame when a slow subscriber is dropped

### web-tier re-attach legs (reattachingDeltas — the path kolu-server's browser tier runs)
✓ yields the delta strings and ends on a graceful stream end          (a real PTY exit still ends — NOT mistaken for overflow)
✓ re-attaches on an `overflow` frame, prefixing the fresh snapshot with a reset
✓ re-attaches repeatedly across successive drops
✓ ends cleanly when the PTY has vanished by the time we re-attach (NOT_FOUND)
✓ propagates a non-NOT_FOUND re-attach failure
```

Each level pins a distinct guarantee: the **drop is now distinguishable from a PTY exit on the wire** (the contract repro), it **survives the real socket round-trip** (so the remote transport R8/R9 lean on is covered), and the **web tier recovers** by re-attaching with a reset-then-snapshot rather than freezing — while a genuine PTY exit still ends the stream for real. Full CI (both platforms) runs all of these.

srid:

The **x86_64-linux** lane ran on **`kolu-ci-8`** (`dev-x86-64-linux-08`) — commit `7f872752f`, **passed** ✓

- **Lane wall** (pipeline): **2m57s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 2m25s | ✓ |
| `ci::home-manager` | 1m36s | ✓ |
| `ci::flake-check` | 1m23s | ✓ |
| `ci::nix` | 1m21s | ✓ |
| `ci::smoke` | 1m0s | ✓ |
| `ci::pnpm-hash-fresh` | 49s | ✓ |
| `ci::unit` | 40s | ✓ |
| `_ci-setup` | 22s | ✓ |
| `ci::biome` | 16s | ✓ |
| `ci::fmt` | 14s | ✓ |
| `ci::install` | 9s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/7f87275/timings.jsonl`; pool state is a live `flock` probe.</sub>

srid:

The **x86_64-linux** lane ran on **`kolu-ci-8`** (`dev-x86-64-linux-08`) — commit `050347aef`, **passed** ✓

- **Lane wall** (pipeline): **2m11s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 1m53s | ✓ |
| `ci::home-manager` | 48s | ✓ |
| `ci::unit` | 40s | ✓ |
| `ci::pnpm-hash-fresh` | 24s | ✓ |
| `ci::nix` | 23s | ✓ |
| `ci::flake-check` | 22s | ✓ |
| `ci::smoke` | 20s | ✓ |
| `ci::fmt` | 15s | ✓ |
| `ci::biome` | 15s | ✓ |
| `ci::install` | 9s | ✓ |
| `_ci-setup` | 8s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/050347a/timings.jsonl`; pool state is a live `flock` probe.</sub>

srid:
## Evidence — overflow recovery (before/after)

A terminal flooding output triggers kaval's slow-subscriber drop. **Before (master)** the drop is mistaken for a PTY exit and the terminal **freezes** — the post-flood `RECOVERED-AND-ALIVE` marker and prompt never arrive. **After (this PR)** the terminal **re-attaches for a fresh snapshot** and the marker + prompt appear. (Captured with the per-subscriber buffer bound shrunk so the >10k-chunk drop is reproducible on camera; the recovery logic shown is exactly the shipped path.)

### Before — master: frozen

▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=overflow-before-master.mp4

### After — PR #1591: recovers

▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=overflow-after-pr.mp4

PR #1593, opened 2026-06-26
fix(dock): hide stale sleeping terminals behind the activity window

srid:
**The dock's activity-window selector (`All / 4h / 12h / 24h / 48h`) is meant to compress everything you haven't touched recently out of the list — but sleeping terminals slipped through it.** A terminal you put to sleep days ago kept its row no matter which window you picked, so a `4h` window still showed dormant tiles last active *days* ago while only the live rows got filtered. [Image #1] in the report shows exactly this: under a `4h` window, slept terminals at `21h` / `3d` / `5d ago` all stayed visible, and only `1` was hidden.

The cause is a deliberate precedence in `classifyDockRow`: `sleeping` was checked **before** `parked`, so a slept tile never fell into the dock's parked-drop "however long it has slept." That carve-out is what this fixes.

### The fix

Flip the two checks so the **staleness window wins over the ☾**:

```
before:  if (sleepingArm(meta)) return "sleeping";   // ← always kept its row
         if (parked)            return "parked";
after:   if (parked)            return "parked";      // ← stale → dropped, even when asleep
         if (sleepingArm(meta)) return "sleeping";    // ← fresh slept tile keeps its ☾
```

A *fresh* slept tile still reads asleep with its ☾ row; once its `lastActivityAt` falls outside the window it routes to `parked`, and `dockTree` drops it — counted in the **"N hidden by … window"** footer and recoverable with **"show all"**. The dock row already renders `formatTimeAgo(lastActivityAt)` for every row including sleeping ones, so the time you see *is* the field the window compares — the hide is coherent with the displayed age.

### Design philosophy

- **Fail-fast / no-fallbacks** — this *removes* a special-case carve-out rather than adding one. Staleness now applies uniformly through the single classifier instead of a sleeping-shaped bypass.
- **Reuse the source of truth** — reuses the existing `isStale(lastActivityAt)` gate already computed in `rankDockRows`; no new sleeping-specific timestamp path.
- **Electricity** — n/a; pure dock-local ranking logic, no new boundary.

> **Not a pulam-web mirror.** This is the dock's local activity-window/parked overlay — pulam-web's `fleet.ts` "sleeping" is a user-toggle filter (agent + foreground absence) with no staleness/window concept, and the change doesn't touch the shared `agentProjection` folds. Nothing to mirror.

Reproduced first with a failing unit test (a stale sleeping terminal classified `sleeping` instead of `parked`), then fixed to green. Updated the test that pinned the old behavior, the `sleeping-terminals` Atlas note (which explicitly documented the now-reversed decoupling), and the changelog.


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `219f7bdb455a`

### Applied (3)
- `lowy-1` Overlay precedence sprayed across two folds, guarded only by a comment — commit `cee568f97`
- `hickey-1` Parked/sleeping overlay precedence duplicated across the two folds, kept in sync by comment — (uncommitted)
- `hickey-2` Heavy prose comments restate the same precedence three times — (uncommitted)

### Agreed — no change (1)
- `lowy-2` Electricity-test verdict: receptacle is sound, no exposed wires (packages/client/src/canvas/dock/dockRowRanking.ts:140-161)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `219f7bdb455a`

### Round 1

**codex** — approved: `false`

The branch is small and the dock filtering path is otherwise coherent: `rankDockRows` feeds `dockTree`, `parked` rows are dropped, and the docs/changelog were updated. I found one user-visible recency bug: the new sleeping filter is still keyed to agent `lastActivityAt`, not the sleeping record's own `sleptAt`, so it misses or prematurely hides real sleeping cases.

Findings:
- `F1` · major · open — Stale sleeping rows are computed from `meta.lastActivityAt`, but sleeping records have their own lifecycle timestamp (`sleptAt`) and sleep does not update `lastActivityAt`. That means a plain shell slept days ago with `lastActivityAt === 0` is never parked, so old sleeping terminals can still pile up; conversely, an agent terminal slept just now after an old agent transition can disappear immediately under a tight activity window, contradicting the advertised "freshly-slept one still shows with its ☾" behavior. (packages/client/src/canvas/dock/dockRowRanking.ts:162)


**claude** — Conceded F1: re-keyed the dock's activity window onto a sleeping tile's `sleptAt` (when you slept it) instead of its agent `lastActivityAt`, added tests that drive the real `isStale`, and tightened the prose.

- `F1` **fixed** — codex is right; this was a genuine recency bug. `rankDockRows` computed `parked = isStale(meta.lastActivityAt)` for every row, but a sleeping tile's recency is *when you put it to sleep* (`sleptAt`), not its last agent transition. Two concrete failures followed, exactly as codex described: (a) a plain shell slept long ago carries `lastActivityAt === 0`, which `isStale` exempts (`staleness.ts:43`), so an agent-less dormant tile would **never** park and would pile up forever; (b) a *just-slept* tile whose agent last moved outside the window would drop instantly, contradicting "a freshly-slept one still shows with its ☾". Fix at `packages/client/src/canvas/dock/dockRowRanking.ts:162` — compute `const recencyAt = sleepingArm(meta)?.sleptAt ?? meta.lastActivityAt;` and feed it to BOTH `isStale(recencyAt)` and the sort key `ts` (since `sleptAt >= lastActivityAt` always, a fresh-slept tile now also ranks by when it slept, not its stale agent stamp). The `??` never falls back for a sleeping arm (`sleptAt` is required on `SleepingDiscriminantSchema`, `surface.ts:349`); active tiles keep `lastActivityAt`. Updated the `rankDockRows` docstring accordingly. Codex also nailed the test gap: the existing sleeping tests stub `isStale` to a constant (`() => stale`), so they never exercised *which* timestamp the ranker keys on — added two tests in `dockRowRanking.test.ts` that drive the real `isStale(ts, now, thresholdMs)`: a plain shell with `lastActivityAt: 0` + old `sleptAt` parks, and a just-slept tile with old `lastActivityAt` stays asleep. Also tightened the now-imprecise "last activity falls outside the window" prose in `sleeping-terminals.mdx` and `changelog/unreleased.mdx` to say the clock is `sleptAt`, and rebuilt the atlas dist. Gate: `just check` (typecheck + biome) green; 21 dock-ranking tests pass; `just atlas::build` regenerated `docs/atlas/dist/sleeping-terminals.html`.

commit: `18b76381d0472cc0fe98aac25dd87031e49e29b7`


### Round 2

**codex** — approved: `true`

F1 is fixed in the current tree. `rankDockRows` now derives `recencyAt` from `sleepingArm(meta)?.sleptAt ?? meta.lastActivityAt`, feeds that into both `isStale` and the row sort timestamp, and the new tests cover the two failure modes with the real `isStale` predicate. The Atlas and changelog prose now also identify `sleptAt` as the clock for sleeping rows. I found no regression introduced by this round.

Findings:
- `F1` · major · resolved — Previously, stale sleeping rows were keyed off `lastActivityAt` instead of the sleeping lifecycle timestamp, so old slept shells could stay visible forever and freshly slept old-agent terminals could disappear immediately. (packages/client/src/canvas/dock/dockRowRanking.ts:173)

_codex on the rebuttal:_ F1: accepted as fixed. CLAUDE conceded the issue and implemented the right timestamp boundary: sleeping rows now key the dock activity window on `sleptAt`, while active rows keep `lastActivityAt`. The added tests directly exercise the plain-shell `lastActivityAt === 0` case and the just-slept/old-agent-transition case, so the original bug is pinned.

srid:

Ran the rule checklist + fact-check passes over the change (elegance pass skipped — `/simplify` already ran this gauntlet).

**Pass 1 — rule checklist: all clear.** Every built-in and project rule passes. Notably: `dockOverlayBucket` satisfies `dry-rule-of-three` (one source for the parked/sleeping precedence), `rowRecencyAt` honors `invalid-states-unrepresentable` (discriminated dispatch on the sleeping arm), and no error-swallowing / unbounded-growth / styling violations were introduced.

**Pass 2 — fact-check: 1 finding, adjudicated as a false positive.**

- *Claim:* the `?? meta.lastActivityAt` fallback in `rowRecencyAt` could silently mask a sleeping terminal whose `sleptAt` is missing, violating fail-fast.
- *Verdict:* not a defect. `sleptAt` is a **required** field on the sleeping arm (`SleepingDiscriminantSchema`: `sleptAt: z.number()` — `packages/common/src/surface.ts:349`), so a sleeping terminal without it is unrepresentable (zod rejects at parse, TS at construction). `sleepingArm(meta)?.sleptAt` is `undefined` **only** when the terminal is active, and `?? meta.lastActivityAt` then correctly selects the active clock. The proposed `if (arm && !arm.sleptAt) throw` would instead introduce a bug (it treats a legitimate `sleptAt === 0` as missing) and would guard a state the type system already makes impossible. **No change applied.**

Net: nothing to fix.

srid:

The **x86_64-linux** lane ran on **`kolu-ci-2`** (`dev-x86-64-linux-04`) — commit `b2bb81614`, **passed** ✓

- **Lane wall** (pipeline): **3m3s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 2m45s | ✓ |
| `ci::home-manager` | 1m29s | ✓ |
| `ci::nix` | 58s | ✓ |
| `ci::unit` | 56s | ✓ |
| `ci::smoke` | 55s | ✓ |
| `ci::surface-example-build` | 38s | ✓ |
| `ci::biome` | 37s | ✓ |
| `ci::fmt` | 36s | ✓ |
| `ci::install` | 9s | ✓ |
| `_ci-setup` | 8s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/b2bb816/timings.jsonl`; pool state is a live `flock` probe.</sub>

srid:
## Evidence

To simulate an "aged" sleeping tile without a real-time seam, the sleeping terminal was seeded directly into the server session store via the `test__set` endpoint with `sleptAt = Date.now() - 25 * 3600 * 1000` (25 hours ago). On page reload, the restore card offered the sleeping terminal back; clicking it called `seedSleepingTerminal` on the server, which registered the terminal with its 25h-old `sleptAt`. Since `rowRecencyAt(meta) = sleptAt` for sleeping tiles and `isStale(sleptAt, now, 24h_threshold)` = `true` (25h > 24h), the tile immediately routed to `parked` — hidden from the dock row list by the default 24h activity window.

**Before "show all" — sleeping tile is hidden (sleptAt = 25h ago, window = 24h)**

The dock header reads **"1 hidden by 24h window — show all"** with no ☾ row visible. The canvas shows the dormant tile ("Asleep · 1d ago") confirming it exists but is suppressed from the dock.



**After "show all" — sleeping tile reappears as ☾ row**

Clicking "show all" sets the activity window to `All` (no filter). The ☾ dock row appears showing "~ 1d ago" matching the sleptAt age. The footer now reads "Activity window All" with no hidden-count sentence.



Captured on PR branch `fix/dock-window-hides-stale-sleeping` (v1.1.0 dev) served from source on a clean pu box.

PR #1594, opened 2026-06-26
R8 — kolu serves terminalWorkspaceSurface.awareness; the browser joins it with kolu's authored record

srid:
**kolu now serves `terminalWorkspaceSurface` itself** — the same surface the `pulam` daemon serves remotely — instead of keeping a private copy of terminal awareness fused into its own record. A terminal's record is **bisected** into two halves: the eight sensor-produced **awareness** fields (cwd · git · lastAgentCommand · lastActivityAt · agentSession · pr · agent · foreground), whose **sole writer is the sensor sink**, and the kolu-**authored** record (location · client/UI chrome · the active|sleeping discriminant). kolu serves the two halves on **two separate collections** — `terminalWorkspace.awareness` (the generic surface `pulam` also serves) and `kolu.authored` — and the **browser joins them at read time**. There is **no server-side re-fusion**: the wire never carries a single fused record. This unblocks the user-visible **R9a** state-desync (a second process can now read kolu's awareness) and turns R9's remote awareness into a pure **backing-swap** behind one collection.

### The bisection — one entry, two halves, joined at the reader

```
        before                                    after (Design-S)
 ┌──────────────────────┐        ┌──────────────────────────────────────┐
 │ entry.meta (fused)   │        │  ONE registry entry                   │
 │ location + UI chrome │  ──▶   │   .meta      = AUTHORED (location ·    │
 │ + cwd·git·pr·agent·… │        │                UI chrome · state)     │
 └──────────────────────┘        │   .awareness = cwd·git·pr·agent·…     │
                                 │                (one writer: the sink)  │
                                 └───────┬───────────────────┬───────────┘
                                  served │ kolu.authored     │ terminalWorkspace.awareness
                                         ▼                   ▼
                                   ┌─────────── the BROWSER joins ───────────┐
                                   │  useTerminalMetadata →                  │
                                   │  composeTerminalMetadata(authored, aw)  │  → TerminalMetadata
                                   └─────────────────────────────────────────┘
```

`composeTerminalMetadata` is the **one** join, applied at exactly two sites — the client read (ephemeral) and disk persist (`snapshotSession`, a save-time snapshot) — and **served nowhere**. Reusing one join at both is what keeps disk and the client read from ever diverging.

### Perfection bar — invalid states are unrepresentable

- **`AuthoredTerminal` names no awareness field** → a stray `entry.meta.cwd = …` is a *compile error*.
- The two sink mutators are each narrowed to **one half** of the awareness value (persisted vs live), so the `terminals:dirty` autosave-firehose fence holds *by type*.
- **awareness is a required field on the registry entry**, born and dropped *with* the entry — so "a terminal exists" and "its awareness exists" are one fact the type makes inseparable. There is no second map to keep in lockstep; the invariant is **structural**, not maintained by hand.
- The **sleeping arm derives `pr` exclusively from the frozen authored value** (the live half is dropped at the compose source), single-sourced through `composeTerminalMetadata` and pinned by `surface.test.ts`.

### Deliberately NOT in R8 (rides R9)

| Concern | R8 (now) | R9 |
| --- | --- | --- |
| `activity` (live green-dot set) | served **quiet** (empty — kolu has no server-side byte-tap yet) | turned live |
| Code-tab fs/git | still reads `koluSurface`'s **value-bearing** streams | rewired to this surface's **procedure + pulse** |

### Review & verification

The full `/be` review gauntlet ran serially on the branch, each step committing in turn (comments below):

- **lens-debate** (lowy ⇄ hickey) — consensus; among 7 fixes, **folded the awareness store into the registry entry** (the structural-lockstep win above), extracted the `finalizeRemoval` teardown receptacle and a shared `createLateBoundSurfaceCtx` factory.
- **codex-debate** (xhigh) — consensus; fixed a teardown-race misclassification (`readScreenText` now throws structured `terminalNotFound`) + stale architecture comments.
- **simplify** — `seedAwarenessLive`, `registryMap`, the `registerAndInstall` seed receptacle (mirroring lens's `finalizeRemoval`), and a client hot-path win (the ordering filters read `parentId` off the authored half instead of running the full join).
- **code-police** — logged a late-write-after-teardown for observability; one "critical" was correctly refuted as a false positive.

**CI green** on a fresh pu box (full `just ci`, `CI_EXIT=0`): the complete **e2e suite (484 scenarios / 4321 steps, 0 failures)**, NixOS VM integration tests, nix builds, atlas-sync, fmt, biome, unit. The risky-path e2e (sleep/wake/resume, daemon-restart reattach, session restore, live git/agent/foreground) is green.

### Safety

- **No on-disk migration** — the persisted projection recomposes to identical bytes; `SCHEMA_VERSION` unchanged.
- **No `@kolu/surface` / `surface-app` / `surface-nix-host` framework package is touched** in this PR's own commits — that is what gates the drishti companion PR. (The PR's own changes span `@kolu/terminal-workspace`, kolu-common — where `composeTerminalMetadata` lives — kolu-client, kolu-server, and pulam.) *(The latest `master` merge separately brought master's own `packages/surface/**` changes into the branch; those are master's, not this PR's.)*

Plan-of-record: `docs/atlas/src/content/atlas/remote-terminals.mdx` §R8 (Atlas status flips to shipped + `PrLink` on merge).

_Authored with [`/be`](https://github.com/srid/agency) on Claude Code (model `claude-opus-4-8`)._


srid:
## Evidence — full e2e on a pu box (`kolu-r8-awareness-e2e`, x86_64-linux)

R8 is **invisible by design** (the wire `TerminalMetadata` is byte-identical), so the proof is *behavioral*: does state still survive across the risky persistence/reattach paths now that awareness lives in a separate single-writer store? It does — the full Cucumber+Playwright e2e is green on the branch HEAD (`004fd57`):

**70 scenarios · 481 steps · 0 failures**, across two runs:

| Suite | Result | What it exercises in Design-S |
| --- | --- | --- |
| `sleeping-terminals` | ✅ | sleep freezes `pr` from the store → authored arm; wake resets the live half + **resumes the SAME conversation** (`agentSession` rides the store, #1495); **daemon-restart-mid-sleep** reattach (`@kaval-restart`); malformed-record drop |
| `session-restore` | ✅ | `snapshotSession` join → restore split (theme · canvas layout · active terminal persist) |
| `kaval-daemon` | ✅ | adopt / orphan reattach after a kaval restart (the #1275 lossless split) |
| `git-context` | ✅ | live git awareness: sink → store → `composeTerminalMetadata` → header/switcher (branch updates on external HEAD change) |
| `foreground-process` | ✅ | foreground awareness through the live sink |
| `recent-agents` | ✅ | `lastAgentCommand` persisted-awareness path |
| `claude-code` | ✅ | **live agent-state** (`thinking`→`waiting`→`awaiting`) — the `updateServerLiveMetadata` quiet path |
| `dock` | ✅ | the Dock renders awareness from the composed `terminalMetadata` (unchanged client read model) |

```
35 scenarios (35 passed) · 295 steps (295 passed)   # run 1
35 scenarios (35 passed) · 186 steps (186 passed)   # run 2 (claude-code + dock)
```

These are the paths the perfection-review flagged as highest-risk (lifecycle lockstep, sleep/wake/adopt persistence) — exercised end-to-end on a real host, not just unit-mocked.

srid:
## Rework: collapsed the server-side re-fusion (perfection-review follow-up)

The first R8 cut bisected a terminal's record into an **authored** half (registry) and an **awareness** half (store), but then **re-fused them server-side** into a served `kolu.terminalMetadata` collection the browser read. So the split never reached the consumer (the "no client change" badge *was* the tell), and awareness was **double-published** every sensor tick — raw onto `terminalWorkspace.awareness` *and* recomposed onto `kolu.terminalMetadata`.

This rework (commit `7f9474373`) collapses that. There is no single fused record anywhere — it's a **join**, materialized only at the reader (ephemeral) and at sleep (persist):

- **`terminalMetadata` collection removed.** kolu serves the kolu-owned **`authored`** record on its own collection. `surfaceCtx.collections.terminalMetadata` is now a **compile error** — a server-side recompose is *unspellable* (pinned by `surface.test.ts`).
- **Client joins** `app.collections.authored` ⋈ `workspace.collections.awareness` at read time (`useTerminalMetadata` → `composeTerminalMetadata`). The ~20 `getMetadata` consumers are unchanged (same return type).
- **No double-publish:** each fact publishes exactly once on its own collection (`publishAuthored` | `publishAwareness`; `publishWire` deleted).
- **Disk persist reuses the same join** at save time, so disk and the client read can't diverge.
- **R9 is now a pure backing-swap** behind `terminalWorkspace.awareness` — no second client read path to migrate.

### Verification

- `just check` clean; full unit suite green — **client 474 / server 152 / common 22** (incl. the `composeTerminalMetadata` guard).
- **Adversarial perfection-review** of the new join (6 skeptical tracks, each finding verified by a second agent): `no-refusion`, `no-double-publish`, `loading-window`, `compose-on-read identity`, `persist-divergence`, `sleep/wake frozen-pr` — **all behaviorally clean**, zero confirmed correctness findings. One comment-accuracy nit (stale "the publish recomposes" comments in `local.ts`) fixed.

### pu-box e2e — the risky read path, re-run for the join

The read path changed materially (server-compose → client join of two async collections), so the prior e2e doesn't cover it. Fresh pu box, commit `7f9474373`:

```
Feature: Sleeping terminals          (sleep / wake / resume #1495 / daemon-restart reattach)
Feature: Session restore
Feature: kaval daemon lifecycle
Feature: Git context in header and workspace switcher
Feature: Foreground process detection
Feature: Recent agents in command palette
Feature: Claude Code status detection
Feature: Dock

70 scenarios (70 passed)
481 steps (481 passed)
E2E_EXIT=0
```

**0 failures** across the sleep/wake/restore/reattach + live git/agent/foreground/dock paths.

> Deliberately **not** run yet, per the standing goal to review before the gauntlet: the `/be-review` gauntlet (lens/codex/simplify/police) and CI. Atlas R8 status flip + `PrLink` remain queued for after review/merge.

srid:

✅ **Consensus** after 2 round(s) · lowy + hickey · base `0c1eeedc76aa`

### Applied (7)
- `lowy-1` Two-store teardown tail sprayed across lifecycle sites — no "remove a terminal" receptacle — commit `10576e3ce`
- `lowy-2` exportTranscriptHtml reaches both halves directly and silently degrades where lockstep guarantees presence — commit `21cd68ee1`
- `lowy-5` Stale TerminalServerMetadata references in terminal-workspace docs after the type was deleted — commit `f4e7a179a`
- `hickey-1` A terminal's state is fragmented across two parallel Maps held in lockstep by ordering discipline — commit `82854df5f`
- `hickey-3` workspaceSurfaceCtx.ts is a hand-copy of surfaceCtx.ts — one concept ("late-bound surface ctx holder") multiplied — commit `3e2736949`
- `hickey-4` Optional-store reads collapse a lockstep-guaranteed value to silent defaults — commit `156592f7b`
- `hickey-5` publishAuthored's defaulted `entry` parameter is never exercised and silently no-ops on absence — commit `9cebeebf7`

### Agreed — no change (4)
- `lowy-3` Seed-side lockstep (install-before-register) maintained by ordering convention across many sites (packages/server/src/terminalEndpoint/local.ts:381,459,885,1041)
- `lowy-4` Asymmetric collection removal — awareness fans an explicit remove, authored relies on key-set disposal (packages/server/src/terminalEndpoint/metadata.ts:115-119)
- `hickey-2` `pr` has two homes selected by `state`, reconciled by composeTerminalMetadata's asymmetric branches (packages/common/src/surface.ts:1155)
- `hickey-6` The "eight awareness fields" enumeration is hand-copied across many docstrings (packages/server/src/awarenessStore.ts:3)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 round(s) · codex reviewed at `xhigh` reasoning effort · base `0c1eeedc76aa`
### Round 1

**codex** — approved: `false`

The split between authored terminal state and awareness is mostly carried through coherently, and the lifecycle paths I checked do pair the two halves on spawn, sleep, wake, restore, and removal. I found one real contract problem in the newly served terminalWorkspace surface: Kolu now advertises the `activity` stream but implements it as permanently empty. I also found a smaller teardown-race logging issue and stale architecture text/comments. I did not run tests in this read-only review pass.

Findings:
- `F1` · major · open — The new `terminalWorkspace.activity` stream always reads `[]`, never installs a producer, and dedups every frame. That violates the terminal-workspace surface contract: `activity` is documented and consumed as the set of terminals producing output right now. Any client consuming Kolu's newly served terminalWorkspace surface, such as the pulam-web style viewer or a future remote mirror, will silently see every terminal as quiet even while bytes are flowing. (packages/server/src/surface.ts:480)
- `F2` · minor · open — `readScreenText` uses `getActiveTerminal(id)!`. If a screen-scrape tick races terminal teardown or sleep, this throws a plain TypeError before reaching the pty-host call. The sensor code classifies structured `NOT_FOUND` as a benign teardown race, so this path will be logged as an unexpected screen-scrape failure instead of the expected teardown case. (packages/server/src/terminalEndpoint/local.ts:268)
- `F3` · minor · open — Several architecture comments/docs still describe the old topology after this branch adds the third `terminalWorkspace` sibling. Examples include `wire.ts` saying Kolu serves two siblings, `contract.ts` still documenting only `{ kolu, surfaceApp }`, `README.md:319` still describing two independent surfaces and `terminalMetadata`, and `common/src/surface.ts:361` mentioning a nonexistent `server/src/awarenessStore.ts` despite awareness now living on registry entries. This is a structural change, and these stale notes will mislead the next person touching the surface boundary. (packages/client/src/wire.ts:85)


**claude** — Fixed the teardown-race misclassification and the stale three-sibling/awareness-storage docs; disputed the empty-activity finding on scope + no-consumer grounds.

- `F1` **disputed** — kolu's `terminalWorkspace.activity` reads `[]` not as a graceful-degradation fallback but because kolu-server has **no standing raw-byte tap** to derive it from. The per-terminal sensor set (`packages/server/src/terminalEndpoint/local.ts:690-748`) taps only cwd / title / commandRun / foreground / exit — all parsed pty-host-side, none carrying raw bytes — and raw output crosses the pty-host IPC **only on-demand**, when a client attaches a tile (`local.ts:964`, the `attach` handler). A truthful activity set needs a *new* always-on per-terminal output tap (the firehose pulam pays via `terminalAttach` because it is headless). The plan-of-record scopes exactly that to **R9**, not R8: `docs/atlas/src/content/atlas/remote-terminals.mdx` defines R8 as "kolu serves `terminalWorkspaceSurface.awareness`" and puts the fs/git **consume** + remote dialing (the byte-tap work) in R9 — the served comment at `packages/server/src/surface.ts:476` ("R9 makes it live") matches. Three reasons the finding's two suggested remedies are both infeasible *at this phase* and no consumer is harmed: (1) **no consumer reads kolu's `terminalWorkspace.activity` today** — the client uses its own browser-side `useTerminalActivity` (`packages/client/src/terminal/useTerminalActivity.ts`; grep finds zero client reads of `terminalWorkspace.activity`), and nothing mirrors kolu's surface until R9, which is precisely when the tap lands, so a real consumer never observes empty-while-flowing; (2) `implementSurfaces` **throws on a missing primitive** (`surface.ts:449`), so kolu cannot "not expose" `activity` while serving `awareness` — the shared `terminalWorkspaceSurface` declares it; (3) the fs/git-real-vs-activity-empty asymmetry the finding leans on is **cost-justified** — fs/git is served real because it is free reuse of the same in-process endpoint kolu already runs (`surface.ts:447-448`, `...fsGit.streams`), whereas activity has no existing in-process source and would add an always-on raw-output IPC firehose for every terminal regardless of tile visibility. Left as the honest empty placeholder with its existing R9 comment. No code change.

- `F2` **fixed** — Replaced `getActiveTerminal(id)!` in the sink's `readScreenText` (`packages/server/src/terminalEndpoint/local.ts:268`) with an explicit lookup that throws the structured `terminalNotFound(id)` (an `ORPCError("NOT_FOUND")`, `terminal-registry.ts:213`) when the entry is absent. A screen-scrape poll tick that races teardown/sleep now yields a NOT_FOUND the sensor's `isNotFoundError` gate (`packages/terminal-workspace/src/sensors.ts:751`) classifies as the benign teardown race (debug), instead of a bare TypeError that fell to the `else` branch and logged "screen-scrape poll tick failed" at error. Added `terminalNotFound` to the registry import. Agreed — codex is right.

- `F3` **fixed** — Updated every stale-topology spot codex cited (and the matching ones it implied): `packages/client/src/wire.ts:77,85` "TWO sibling surfaces"/`{ kolu, surfaceApp }` → three (`kolu`, `surfaceApp`, `terminalWorkspace`); `packages/common/src/contract.ts:142-146` same two→three + wire-path example; `packages/common/src/surface.ts:1046` "The two siblings" → three and `:1054` "process-singleton awareness store" → "projected off each registry entry's `awareness` field"; `packages/common/src/surface.ts:361` the nonexistent `server/src/awarenessStore.ts` → the registry entry's `awareness` field (`server/src/terminal-registry.ts`); `README.md:319` "multiplexes two independent surfaces" → three (+ the `workspace` client), and the now-stale `useTerminalMetadata.ts` "`ActivityStreamEvent` reducer" description → the `kolu.authored` ∪ `terminalWorkspace.awareness` collection join the file actually does today. `just check` (typecheck + biome) green; `just fmt` clean (touched only these 5 files).

commit: `305ae79fcec4e276c5fdb3781dd89595c88fcc2b`


### Round 2

**codex** — approved: `false`

I re-inspected the current tree against the existing findings only. F1 is resolved by the accepted scope/no-consumer rebuttal, and F2's teardown-race fix is correct. F3 is mostly fixed, but one stale awareness-storage comment remains, so the review is not approved yet.

Findings:
- `F1` · major · resolved — The `terminalWorkspace.activity` implementation still emits an empty set, but I accept CLAUDE's rebuttal that this branch's R8 scope intentionally serves awareness only, the shared surface shape requires an `activity` primitive, and no current consumer reads Kolu's `terminalWorkspace.activity` before the planned R9 raw-output tap work. (packages/server/src/surface.ts:480)
- `F2` · minor · resolved — The teardown race is no longer misclassified: `readScreenText` now explicitly checks `getActiveTerminal(id)` and throws `terminalNotFound(id)`, letting the sensor classify the race as benign NOT_FOUND instead of logging a bare TypeError. (packages/server/src/terminalEndpoint/local.ts:268)
- `F3` · minor · open — One stale topology/storage comment remains: it still says `terminalWorkspace.awareness` is backed by Kolu server's `process-singleton awareness store`, even though the updated design projects awareness from each terminal registry entry's `awareness` field. (packages/common/src/surface.ts:914)

_codex on the rebuttal:_ F1: I accept the rebuttal. The plan-of-record leaves Kolu-local activity quiet until R9, `implementSurfaces` forces the primitive to exist with the shared surface shape, and no current consumer observes this placeholder, so keeping it open for R8 would overstate the branch's contract.

F2: verified fixed.

F3: most cited stale docs/comments were fixed, but the same stale `process-singleton awareness store` wording still exists at `packages/common/src/surface.ts:914`, so F3 remains open until that last comment is updated.


**claude** — Fixed the last stale awareness-storage comment codex flagged at `surface.ts:914`; F1/F2 already settled in prior rounds.

- `F1` **fixed** — Settled in round 1: codex accepted the rebuttal (verdict-2 marks it `resolved`). kolu's `terminalWorkspace.activity` serves an honest empty placeholder because kolu-server has no standing raw-byte tap to derive a real activity set from; that tap is scoped to R9, no consumer reads it today, and `implementSurfaces` forces the primitive to exist. No code change this round.
- `F2` **fixed** — Settled in round 1: `readScreenText` (`packages/server/src/terminalEndpoint/local.ts:268`) now does an explicit lookup that throws structured `terminalNotFound(id)` so the sensor classifies a teardown race as benign NOT_FOUND. codex verified fixed (verdict-2 `resolved`). No code change this round.
- `F3` **fixed** — Updated the one remaining stale comment at `packages/common/src/surface.ts:914`. Was: "Its `awareness` collection is backed by kolu-server's process-singleton awareness store (Design-S)." Now: "Its `awareness` collection is projected off each registry entry's `awareness` field (Design-S; the sensor sink is the sole writer, see `server/src/terminal-registry.ts`)." This matches the already-corrected wording at `surface.ts:1055`. I left `surface.ts:386` and `:394` untouched — they use "awareness store" generically to contrast the live half vs the frozen-`pr` snapshot on the authored record, not to assert a process-singleton backing store; codex did not flag them and they are not stale. `just check` (typecheck across all packages + biome) green.

commit: `e441ff288e0b35b6d781a521dbded84697b42e10`


### Round 3

**codex** — approved: `true`

All findings from the review are now resolved. I verified the round-2 commit only touched the remaining stale comment in `packages/common/src/surface.ts`, and the old `process-singleton awareness store` wording is gone from the reviewed docs/comments. I did not rescan for new pre-existing issues per the convergence rule.

Findings:
- `F1` · major · resolved — The `terminalWorkspace.activity` stream currently emits an empty set, but the R8 scope/no-consumer rebuttal was accepted: the shared surface requires the primitive, Kolu has no standing raw-byte tap until R9, and no current consumer observes this placeholder as live activity. (packages/server/src/surface.ts:480)
- `F2` · minor · resolved — The teardown-race misclassification was fixed: `readScreenText` now throws the structured `terminalNotFound(id)` instead of dereferencing a missing active terminal. (packages/server/src/terminalEndpoint/local.ts:268)
- `F3` · minor · resolved — The last stale awareness-storage comment was fixed. It now says `terminalWorkspace.awareness` is projected from each registry entry's `awareness` field, with the sensor sink as the sole writer, matching the current design. (packages/common/src/surface.ts:914)

_codex on the rebuttal:_ F1 and F2 were already settled in the prior round and remain resolved. F3 is verified fixed: the latest commit replaces the remaining `process-singleton awareness store` wording at `packages/common/src/surface.ts:914` with the registry-entry `awareness` description. No disputes remain.

srid:

Two passes over the R8 branch (`git diff 0c1eeedc7...HEAD`), scoped to the merge-base; elegance pass skipped (the `/simplify` track already ran this tree).

**Pass 1 — rule checklist:** 25 rules checked, clean on all but one (cited as both `no-silent-error-swallowing` and `errors-must-log-at-error`): `updateServerMetadata` / `updateServerLiveMetadata` dropped a late awareness write (`if (!aw) return`) with no trace. Sensors are torn down *before* the entry is removed, so the guard "never" fires — but a write that *does* outlive its terminal (a sensor-teardown bug) was being swallowed unobservably.
→ **Applied** (`2fef515bb`): log it at `debug` so the bug-only case surfaces. Also corrected `publishTerminalState`'s now-stale doc ("seeds the store before registering the entry" → the caller registers + fans awareness out via `registerAndInstall` first).

**Pass 2 — fact-check:** raised one "critical" — `registerActiveAndSpawn` allegedly dropped `publishTerminalState` after the `registerAndInstall` refactor, so fresh spawns wouldn't fan their authored record to existing subscribers.
→ **Refuted (false positive):** `publishTerminalState(entry, id)` is called at `local.ts:453`, immediately after `registerAndInstall` (`:448`). The `sleepWake` recording-ctx test (`expect(upserts).toContainEqual({ id, state: "active" })`) asserts exactly this push and passes. No change.

`just check` + full unit suite (server 146 / client 474 / common 22) green after the fix.

srid:
## ✅ CI + evidence — green on the gauntlet-finished tree

Full `just ci` on a fresh pu box at the final gauntlet commit `2fef515bb` — the complete gate (`nix · flake-check · home-manager · e2e · smoke · fmt · biome · unit · surface-example-build · surface-app-example-build · pnpm-hash-fresh · atlas-sync`):

- **`CI_EXIT=0`** — every gate passed.
- **Full e2e: 484 scenarios / 4321 steps, 0 failures** — the *whole* suite, not just the risky subset: the lens fold (awareness → registry field) and simplify pass materially changed the server-side awareness storage and the client hot paths, so the entire e2e re-ran clean.
- **NixOS VM integration tests** passed (kolu-service / -adoption / -adoption-currency / -adoption-skew — the skew VM intentionally exercises `DaemonContractSkewError`).
- `just check` + unit (server 146 / client 474 / common 22) green at every gauntlet step.

The review gauntlet (lens → codex → simplify → police) **and** CI are both green. Marking ready for review.

PR #1598, opened 2026-06-27
fix(surface): stop the half-open watchdog flashing "Disconnected" on resume

srid:
## What you saw

Coming back to the kolu PWA — switching windows, or opening the laptop lid after sleep — briefly flashed a **"Disconnected from server"** overlay, then carried on as if nothing happened. The browser console showed:

```
surface: heartbeat probe timed out — forcing reconnect (half-open socket)
```

New on desktop in the last few days; previously only mobile flashed.

## Root cause

The default-on liveness watchdog landed in **#1545** (3 days ago). It probes `system.live` every 15s and arms a 10s `setTimeout`; a missed probe means the socket is half-open, so it forces `ws.reconnect()`. Before #1545 nothing probed or force-reconnected — which is exactly why the desktop flash is new.

The bug: **the timer can't tell "the computer was asleep" from "the server went silent."** When a tab is frozen (background, occlusion, OS sleep) its timers pause but the wall clock keeps moving. On resume the browser fires the overdue 10s timeout *before* the socket's already-arrived `system.live` reply is processed → `onStale` → `ws.reconnect()` over a perfectly healthy socket. That forced reconnect closes+reopens the socket, which the lifecycle renders as the brief "Disconnected" overlay.

Investigated and held to an ideal bar via `/perfection-review` (see the residuals below); the headline insight is that the obvious fix — *pause the probe while `document.hidden`* — is **wrong**: a hidden-but-running tab's probe timeout is real, so gating on visibility blinds the watchdog to a genuine half-open. The honest fix is to measure suspension, not visibility.

## The fix

**1 — Measured clock-gap void (the correctness core, `heartbeat.ts`).** When the probe timeout fires, compare elapsed **wall** time (`Date.now`) to elapsed **monotonic** time (`performance.now`). If wall ran far ahead of monotonic, the runtime was suspended across the probe — the window was never a fair test — so **void** it (no `onStale`) and re-probe fresh. The verdict is the watchdog's own clock arithmetic, not an injectable "suspended?" boolean, so there's no knob to pin true and silence it. A **void-budget ceiling** keeps repeated voiding from deferring a real stale verdict forever. This is the browser-leg analog of the ssh leg's existing wall-clock-gap wake watcher (#1078), and it protects every consumer (browser + ssh) uniformly while staying framework-free (no `document` in the shared core).

**2 — Instant wake re-probe (latency, the browser legs).** A `window.focus` / tab-visible event re-probes immediately via a new `wake()` (mirrors `renderRecovery.ts`'s DOM-free-core + injected-seam shape), so a socket a *real* sleep killed recovers in ~ms instead of waiting up to a full interval. Wake only ever probes — never voids, never declares stale.

**3 — Overlay hysteresis (`surface-app`).** A new grace-windowed `presentingDown` holds the full-screen overlay back until `down` has persisted ~1s, so a sub-second reconnect (this fix's, or any legitimate fast recovery) never flashes the alarm; a sustained outage still surfaces promptly. `status()` itself stays instantaneous (it gates the header dot and the heartbeat).

Also fixes a latent open/close listener leak in `liveSignal.dispose()` surfaced while wiring the wake listeners.

## Perfection-review residuals closed

- **Event-blindness** (headline): foreground OS-sleep and app-switch fire *no* `visibilitychange` — so an event-driven fix misses them. The measured clock gap needs no DOM event. ✅
- **Visibility ≠ suspension**: never void on `hidden` (a coverage regression); only a measured gap voids. ✅
- **Branded-but-blind**: the void is core-owned arithmetic + a ceiling, so "always suspended" / "voided forever → silent" is unspellable. ✅
- **Listener leak / lifetime**: `dispose()` now detaches every listener it attached. ✅

## Testing

`now`/`mono` are injected so a test can model "wall advanced while the loop was frozen" — which fake timers alone cannot (advancing a fake timer moves both clocks together). New tests: suspension void (+ repeated deferral), no-suspension control (genuine dead socket still reconnects), void-budget ceiling, `wake()`, dispose-detaches-every-listener (happy-dom), and overlay hysteresis (sub-second down→live never paints; sustained down does). The void + wake tests are **red on pre-fix code**, green after.

`surface` 204/204 · `surface-app` 137/137 · `surface-nix-host` (ssh leg) 96/96 · `just check` green.

## Shared-package contract

This touches `@kolu/surface` and `@kolu/surface-app` (consumed by drishti). All changes are additive/backward-compatible. A paired **drishti PR** validating its CI against the final HEAD of this branch will follow per `.claude/rules/surface.md`.


srid:
## Paired drishti PR

Per `.claude/rules/surface.md`, the shared `@kolu/surface(-app)` change is validated downstream by **srid/drishti#77** (draft).

- drishti vendors the whole `@kolu/*` surface stack from a single **npins** pin of `juspay/kolu` (`npins/sources.json`), hydrated into `node_modules/@kolu/*` as raw `.ts` — so the PR is a one-line pin bump to this branch's HEAD. **No drishti code change** was needed: it consumes surface-app only via `<SurfaceAppProvider>` + `useSurfaceApp()` and never hand-constructs a `SurfaceAppModel`, so the new required `presentingDown` is provider-filled; every other delta is additive.
- **CI: 20/20 green** — `aarch64-darwin` (srid@sincereintent) 10/10 · `x86_64-linux` (kolu-ci-2) 10/10 — pinned to the current HEAD of this branch.

It's held **draft** and will be **re-pinned to the final HEAD** of this PR (after the review gauntlet settles the API) and re-run to green before merge, per the rule's "validated against the *final* kolu HEAD".

srid:

✅ **Consensus** after 2 round(s) · lowy + hickey · base `0c1eeedc76aa`

### Applied (5)
- `lowy-1` Overlay-grace debounce is braided into the connection-state engine (two axes in one module) — commit `976188765`
- `lowy-2` Silent instant-fallback when `presentingDown` is omitted (degrade-by-omission seam) — (uncommitted)
- `lowy-4` `now`/`mono` injection widens the production receptacle for a test reason — commit `d851e4cdb`
- `hickey-1` Grace-windowed "down" is fragmented across a transport classifier, a reconciliation wrapper, and a degenerate provider fallback — (uncommitted)
- `hickey-2` `now`/`mono` are two coupled optional fields on the public options, diverging from the repo's `deps` test-seam pattern — (uncommitted)

### Agreed — no change (5)
- `lowy-3` `onWake` is a 6th hand-rolled 'runtime resumed' site; the focus-vs-visibilitychange quirk is duplicated knowledge (packages/surface/src/solid/onWake.ts:1-49)
- `lowy-5` `wake()` added to the shared return shape is unused by the ssh leg (symmetry check) (packages/surface/src/heartbeat.ts:183-186)
- `hickey-3` `onWake` is a third parallel "runtime resumed" event seam alongside two existing client wirings (packages/surface/src/solid/onWake.ts:30)
- `hickey-4` Wake-to-heartbeat wiring duplicated at two minter call sites (packages/surface-app/src/solid/index.ts:388)
- `hickey-5` `abandon()`'s generation bump is redundant with `tick()`'s, and the "is this outcome current?" guard is triplicated (packages/surface/src/heartbeat.ts:271)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 round(s) · codex reviewed at `xhigh` reasoning effort · base `0c1eeedc76aa`
### Round 1

**codex** — approved: `false`

The main shape is reasonable, but I would not approve this yet. The heartbeat wake path weakens the watchdog invariant it is meant to harden, and the shared surface API/docs obligations are not fully covered in the branch.

Findings:
- `F1` · major · open — `wake()` abandons an in-flight probe and immediately starts another one, but the abandoned `probe()` cannot actually be cancelled. Repeated wake events before `timeoutMs` therefore keep resetting the stale deadline and can stack orphaned in-flight RPC probes. Since `onWake` wires focus, visibility, and resume events to this callback, a dead half-open socket can be kept from ever reaching `onStale()` by repeated wake events, violating the watchdog's no-overlap/no-silent-deferral invariant. (packages/surface/src/heartbeat.ts:338)
- `F2` · major · open — This adds API-facing shared surface behavior (`SurfaceAppModel.presentingDown`, `createHeartbeat().wake`, and new `@kolu/surface/solid` exports). Per the repo's AGENTS rule for `packages/surface*`, shared surface API/runtime changes require a corresponding drishti PR validated against the final kolu HEAD. I don't see any branch evidence or linked update for that downstream gate here. (packages/surface-app/src/solid/index.ts:437)
- `F3` · minor · open — The published surface-app Atlas note is now stale: it still says apps render the header dot and dim overlay from `useSurfaceApp().status()`, and the model snippet omits `presentingDown()`. With this change, full-screen disconnected overlays are explicitly supposed to use the grace-windowed `presentingDown()` accessor, not raw `status()`. (docs/atlas/src/content/atlas/surface-app.mdx:67)


**claude** — Budgeted the heartbeat `wake()` fast-path against the existing void ceiling so a wake storm can't silence the watchdog (+ regression test), refreshed the surface-app Atlas note for `presentingDown()`, and pushed back on the drishti-gate finding as a ship obligation rather than a diff defect.

- `F1` **fixed** — Real gap, agreed. `wake()` abandoned the in-flight probe and re-`tick()`ed unconditionally (old `heartbeat.ts:340`), so a flood of wakes faster than `timeoutMs` kept clearing `probeTimer` before it could fire — the void-budget ceiling is only consulted *inside* the timeout, so it was never reached and a dead half-open socket could be deferred off `onStale()` forever. `onWake` wires focus/visibility/resume to this callback, so the storm is reachable. Fix (`heartbeat.ts:344-356`): a wake re-probe is a void in disguise, so it now honours the SAME `VOID_BUDGET_FACTOR` ceiling the suspension-void path does — when a wake finds an in-flight probe AND `mono() - lastSettledMono > voidBudgetMs`, it fires `settled(true, generation)` instead of re-probing; otherwise it abandons + re-probes as before. `abandon()` still never touches `lastSettledMono`, so the running-time budget genuinely climbs under the flood (it only resets on a definitive settle), and a healthy link whose probes settle keeps resetting it and only ever re-probes. Added a regression test (`heartbeat.test.ts:296-336`): a never-resolving probe hit by `wake()` every 100ms (faster than the 500ms timeout), clocks injected in lockstep (gap 0, so it's the budget — not a suspension-void — that ends it), still calls `onStale()` exactly once, within `voidBudgetMs` of *running* time, with a bounded probe count. Updated the now-false "wake can only cause an extra probe, never a stale verdict" claims in `heartbeat.ts` (docstring ~`:178`, the `wake` comment) and `onWake.ts:15`. `just test-unit` green (surface: 209 passed), `just check` green (tsc + biome clean). The residual "unbounded *probe calls* within the window" codex mentions is bounded by construction here: re-probes only happen per wake event, and DOM can't emit unbounded focus/visibility/resume events in a finite 75s (default) running-time window — the deferral bound is the load-bearing invariant and it's now enforced.

- `F2` **disputed** — The obligation is real (the diff adds `SurfaceAppModel.presentingDown`, `createHeartbeat().wake`, and new `@kolu/surface/solid` exports `gracedDown`/`onWake` — all API-facing per `.claude/rules/surface.md`), but it is not a defect in the reviewed diff and not a working-tree code change a codex-debate round can make: opening/validating a drishti PR lives in a separate repo and is a `/be` **ship-gate** ("done is not done until drishti CI is green against the final kolu HEAD"), enforced after the gauntlet — precisely so the pin tracks the *final* HEAD this debate is still reshaping. Marking it on a round that edits kolu source conflates the two phases; there is nothing to edit in this tree to satisfy it. Conceded in full: the drishti PR must be opened/repinned and its CI green before this merges — that is owned by the ship phase, not deferred or waived.

- `F3` **partial** — Agreed the live note went stale; fixed the current-section touchpoints and regenerated dist. `surface-app.mdx:67` (the non-historical "The app shell's parts" prose) now says the **instant header dot** reads `useSurfaceApp().status()` while the **full-screen "Disconnected" overlay** reads the grace-windowed `presentingDown()` — matching the actual `TransportOverlay.tsx` change (`disconnected = () => pwa.presentingDown()`); and the `### The pieces` table parenthetical (`:99`) now lists `presentingDown` in the model enumeration. Ran `just atlas::build`; the only dist delta is `docs/atlas/dist/surface-app.html` (committed here), and `just atlas::check-sync`'s content tests pass (the one "out of sync" line was just the uncommitted rebuild). Did NOT touch the `## The design` / `## The API` model snippet (`:120-128`) or the historical API section — both carry an explicit `historical (merge era)` Pill and are frozen design history (they still describe `surfaceApp.info` probe paths and `composeSurfaces` that the note itself marks as removed), so back-filling `presentingDown()` into them would misrepresent frozen history as current API; the live reader's path is the two current-section fixes above.

commit: `45a42d82b63a137473b7be0d005e1d703847368f`


### Round 2

**codex** — approved: `false`

F1's original wake-storm bug is addressed, F3's live docs are now in sync, and I accept the F2 drishti point as a mandatory ship gate rather than a code-change defect in this debate. However, the F1 fix introduced a new resume-path regression: the wake budget can now force a stale reconnect on the first wake after a long browser freeze instead of doing the fresh probe this change is meant to preserve.

Findings:
- `F1` · major · resolved — Original issue: repeated `wake()` calls could keep abandoning in-flight probes and prevent the heartbeat timeout from ever reaching `onStale()`. (packages/surface/src/heartbeat.ts:352)
- `F2` · major · resolved — The shared surface API changes still require a drishti PR/CI gate before merge under the repo rules. (packages/surface-app/src/solid/index.ts:437)
- `F3` · minor · resolved — The live surface-app Atlas prose was stale about using raw `status()` for the full-screen disconnected overlay. (docs/atlas/src/content/atlas/surface-app.mdx:67)
- `F4` · major · open — The new wake budget spends `mono() - lastSettledMono` before allowing a fresh wake probe. But `onWake` explicitly wires Page Lifecycle `resume` for frozen tabs where the code comments say wall/monotonic gap is about zero. After a long browser freeze, `mono()` can be past the void budget on the first resume event, so `wake()` immediately calls `settled(true, generation)` and forces `onStale()` instead of abandoning the interrupted probe and re-probing a healthy socket. That reintroduces the spurious reconnect-on-resume path for long frozen tabs. (packages/surface/src/heartbeat.ts:355)

_codex on the rebuttal:_ F1: accepted for the original wake-storm finding; the repeated-wake/no-timeout case is now budgeted and tested. F2: conceded as a debate finding; the drishti PR/CI requirement remains a mandatory pre-merge ship gate, but I won't keep it open as a kolu diff defect. F3: accepted; updating the current Atlas sections is enough, and leaving explicitly historical API prose untouched is reasonable. New F4: the F1 fix introduced a separate resume regression because the budget check treats long page-freeze elapsed monotonic time as spent running time and can force stale on the first resume wake.


**claude** — Fixed F4 (the resume-after-freeze regression my F1 fix introduced) by anchoring the wake-storm budget at the first wake-abandon instead of `lastSettledMono`; F1/F2/F3 were already resolved by codex last round.

- `F1` **fixed** (already resolved round 1) — codex marked the original wake-storm finding resolved; nothing to redo. The void-budget ceiling on wake re-probes plus the flood regression test stand.

- `F2` **disputed** (already conceded round 1) — codex accepted my framing: the drishti PR/CI requirement is a mandatory pre-merge **ship gate**, not a working-tree diff defect a codex-debate round can satisfy. No code change; the gate remains owned by `/be` §ship against the final kolu HEAD per `.claude/rules/surface.md`.

- `F3` **partial** (already resolved round 1) — codex accepted that updating the current Atlas sections (`surface-app.mdx` `presentingDown()`) is enough and leaving the explicitly `historical (merge era)` API prose untouched is correct.

- `F4` **fixed** — Agreed, real regression, and my own F1 fix introduced it. Root cause: my round-1 wake budget read `mono() - lastSettledMono > voidBudgetMs` (old `heartbeat.ts:355`). That ceiling is correct for the **suspension-void** path (a genuine system suspension FREEZES the monotonic clock, so `mono() - lastSettledMono` there counts only running-time flapping). But a `wake()` fires for an **OS-awake tab freeze** (Page-Lifecycle `resume`, wired in `onWake.ts:11-14`, whose own comment notes "the freeze is OS-awake, so the wall/monotonic gap the void watches is ~0"). In that case the monotonic clock **advances through the freeze**, so a tab frozen longer than `voidBudgetMs` (75s default) pushed `mono() - lastSettledMono` past the budget and the FIRST `resume` wake hit the `settled(true, …)` branch — forcing a reconnect instead of the fresh probe this whole change exists to do. Exactly the spurious reconnect-on-resume codex described.

  Fix (`heartbeat.ts`): gave the wake path its OWN running-time budget anchored at the **first wake-abandon since the last definitive settle** (new `wakeStormStartMono`, declared ~`:225`, reset to `undefined` alongside `lastSettledMono` in `settled()` ~`:248`, consulted in `wake()` ~`:360`). The first wake over an in-flight probe (`wakeStormStartMono === undefined`) **always** abandons + re-probes a fresh window — so a lone resume after any freeze length re-probes, and even a single resume that trips focus + visibilitychange + resume together lands those 2-3 wakes at the same mono, well inside the budget. Only a **sustained storm** — later wakes that keep re-arming a never-settling probe past `voidBudgetMs` of running time *since the storm began* — converts to a stale verdict. The suspension-void path is untouched (its `lastSettledMono` ceiling is still correct).

  Tests (`heartbeat.test.ts`): added codex's requested regression — probe 1 left pending, `mono` jumped to 60_000 (past the 4500ms budget), a SINGLE `wake()` re-probes (probe 2) with `onStale` NOT called, and probe 2 then answers (healthy link, still never stale). This test FAILS on the round-1 code (`60000 - 0 > 4500` → forced stale) and passes now. Also updated the existing wake-flood test for the new anchor (the storm budget now runs from the first wake at mono 100, so the fire point and bounds shifted by one step; assertions re-expressed against `firstWakeMono + voidBudgetMs` with margin). 17/17 surface heartbeat tests green; `just check` (tsc all packages + biome lint `--error-on-warnings`) green; `tsc -p packages/surface` exit 0.

  Scope note: `just fmt` also re-indented `packages/client/src/rpc/rpc.ts` (a pre-existing 4-space→2-space block I did not touch this round and codex did not flag) — reverted it so this commit stays scoped to the F4 fix.

commit: `4641154bfce25daae39356e429f9000688e798a2`


### Round 3

**codex** — approved: `true`

The current F4 fix addresses the regression introduced by the prior wake-storm fix: the first wake after a long OS-awake tab freeze now always gets a fresh probe window, while sustained wake floods are still bounded by a wake-specific budget. All findings from this review session are resolved. The drishti validation remains a real pre-merge ship gate, but I am no longer treating it as an open diff finding in this debate.

Findings:
- `F1` · major · resolved — Repeated `wake()` calls could previously keep abandoning in-flight probes and prevent the heartbeat timeout from ever reaching `onStale()`. (packages/surface/src/heartbeat.ts:352)
- `F2` · major · resolved — The shared surface API changes require downstream drishti validation before merge under the repo rules. (packages/surface-app/src/solid/index.ts:437)
- `F3` · minor · resolved — The live surface-app Atlas prose was stale about using raw `status()` for the full-screen disconnected overlay. (docs/atlas/src/content/atlas/surface-app.mdx:67)
- `F4` · major · resolved — The previous wake budget used `lastSettledMono`, so the first wake after a long OS-awake tab freeze could force stale instead of re-probing. (packages/surface/src/heartbeat.ts:360)

_codex on the rebuttal:_ F4: accepted. The new `wakeStormStartMono` anchor fixes the specific regression I raised: freeze time before the first wake no longer spends the wake budget, so a lone resume wake re-probes instead of forcing `onStale()`. The updated flood test still preserves the original F1 bound. F1 remains resolved. F2 remains conceded as a ship gate, not an open diff defect. F3 remains resolved; the current docs are updated and the historical prose can stay historical.

srid:

**Pass 1 — rules: 1 finding, actioned.**
- `errors-must-log-at-error` — `surface-app/connect.ts`'s default `warnProbeThrew` logged a *synchronous* probe throw at `console.warn`. A sync throw leaves the watchdog permanently **inert** (a hard fault, not the degraded-but-recoverable blip a timed-out probe is), so it must reach operators filtering on `error`. Fixed to `console.error`, aligning with `@kolu/surface`'s `liveSignal` reporter, which already logs the identical case at error. `warnStale` stays `warn` — a missed probe is routine reconnect recovery. (`cd6235c03`)

**Pass 2 — fact-check: no issues.** The suspension-void arithmetic (wall − monotonic gap), the void-budget and wake-storm ceilings (both finite, anchored correctly), the generation-guarded probe abandonment, and the loud (never-swallowed) error paths all check out against the tests.

_(Elegance pass skipped — `/simplify` already ran this gauntlet.)_

srid:

The **x86_64-linux** lane ran on **`kolu-ci-7`** (`dev-x86-64-linux-05`) — commit `cd6235c03`, _verdict unknown (no per-node timing)_

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

srid:
## Drishti PR — re-pinned & green ✅

**srid/drishti#77** is now **ready (non-draft)** and **20/20 CI green**, re-pinned to this PR's gauntlet HEAD `cd6235c0` (npins bump, no drishti code change — it consumes surface-app only via `<SurfaceAppProvider>` + `useSurfaceApp()`). The agent re-validated the gauntlet's reshaped (still additive) surface API against drishti's types: `gracedDown` export, the `deps:{now,mono}` container, `wakeStormStartMono`, and `presentingDown` now derived in the provider — `bun typecheck` green across all three drishti workspaces.

Note: the only kolu commit after the drishti pin (`086b0fc5`) is a **whitespace-only `biome` reformat of `packages/client/src/rpc/rpc.ts`** — a client file drishti does **not** vendor. `git diff cd6235c0..086b0fc5` touches zero surface package, so the pin validates byte-identical surface code to final HEAD.

srid:
## Evidence

This is a **timing-race** regression — the proof is the watchdog *not* mistaking a suspension for a dead socket — so the rigorous, reproducible artifact is the deterministic before/after, with the wall/monotonic clocks injected to model "the wall clock advanced while the event loop was frozen" (fake timers alone can't express it). Live capture is a pu-box-only operation here (production kolu is `active` on this machine, so a local dev server is forbidden per the dev-server venue gate), and a video of "I came back and nothing flashed" is a weaker artifact than the red-on-pre-fix tests for a race.

### Before — pre-fix heartbeat core (`heartbeat.ts` at the merge-base), tests kept

```
❯ src/heartbeat.test.ts (17 tests | 4 failed)
  × VOIDS a probe whose window a suspension crossed (wall jumped, monotonic frozen) …
      → expect(onStale).not.toHaveBeenCalled()   // the watchdog DID force a reconnect — the bug
  × wake() abandons the in-flight probe and re-probes immediately …
      → TypeError: hb.wake is not a function     // no fast resume re-probe existed
  × a flood of wake() faster than timeoutMs still fires onStale … (wake-storm guard)
      → TypeError: hb.wake is not a function
  × a SINGLE wake after a long OS-awake tab freeze re-probes a fresh window … (long-freeze guard)
      → TypeError: hb.wake is not a function
 Tests  4 failed | 13 passed (17)
```

The 13 pre-existing heartbeat tests still pass pre-fix — the change is **additive**: it only adds the suspension-void + wake path, it doesn't alter the genuine-dead-socket behavior.

### After — with the fix

```
 Tests  17 passed (17)        # heartbeat.ts
 Tests  208 passed (208)      # @kolu/surface
 Tests  135 passed (135)      # @kolu/surface-app
 Tests  96 passed  (96)       # @kolu/surface-nix-host (ssh leg — shared core, unaffected)
```

The four red tests are exactly the regression (`onStale` forced on a suspended probe) plus the two regression guards codex added in the debate (a wake-*storm* still reaches `onStale`; a *single* wake after a long OS-awake freeze re-probes instead of forcing stale).

### See it yourself (manual, ~30s)

1. Open kolu in Chrome; open DevTools → Console, filter `heartbeat`.
2. In a second tab go to `chrome://discards/`, click **🥶 Freeze** on the kolu row (it must be backgrounded).
3. Wait ~25–30s, then click back to the kolu tab → it resumes.

**On today's `master`:** a brief **"Disconnected from server"** overlay flashes + console logs `surface: heartbeat probe timed out — forcing reconnect (half-open socket)`. **On this branch:** neither. (A real laptop-sleep → wake exercises the measured wall-vs-monotonic void; the `chrome://discards` freeze is OS-awake, so the `resume`-wired wake re-probe catches it — both paths covered.)

### CI

Both platforms green — `aarch64-darwin` (rasam) + `x86_64-linux` (kolu-ci-7), 0 failed / 0 errored. Paired **drishti#77** green & ready.

Issue #80, opened 2026-03-23
Add find-in-terminal search UI


srid:
The `@xterm/addon-search` is already loaded and provides `findNext`/`findPrevious` with regex, case-sensitive, and whole-word options plus match highlighting.

Wire it to a search bar (Cmd/Ctrl+F) overlaying the active terminal — input field, next/prev buttons, match count indicator, and Escape to dismiss.

PR #1601, opened 2026-06-27
feat(surface): opt-in batched `deltas` verb for Collection<K,V>

srid:
## Why

A `@kolu/surface` `Collection<K,V>` is published as **one keys-stream + N per-key value streams**. A producer that mutates N keys in a tick ships **N separate wire frames**, each paying its own async-iterator + tracing-span overhead. The counterintuitive proof, measured in the downstream consumer [drishti](https://github.com/srid/drishti): a **16-key** collection updated every tick cost *more* browser main-thread time than a **511-row bulk snapshot-then-delta stream** — because the bulk stream is one frame/tick and the collection is sixteen. This was the framework lever surfaced by a perfection-review of drishti's high CPU.

## What

An **opt-in** `deltas` verb: a single batched snapshot-then-delta stream for the whole collection — the bulk-friendly twin of the per-key `keys`+`get` pair.

- **Server** (`server.ts`): when a collection lists `deltas`, `walkSurface` wires a `deltasBus` and coalesces a tick's mutations — a last-op-wins `Map` flushed once per `queueMicrotask` after the producer's synchronous `upsert`/`remove` loop — into **one** `{upserts, removes}` frame. The producer `ctx.collections.X.upsert/remove` API is byte-identical; the existing `keys`/`get`/per-key publishes are untouched.
- **Client** (`useCollection.ts` + `surfaceClient.ts`): a whole-collection `.use()` (no `keys` filter) on a collection that **exposes** `deltas` folds that one stream into a reconcile-backed per-key store — so `byKey()`/`keys()` stay fine-grained (only changed keys re-notify), mirroring the proven `createSubscription`+reconcile pattern. Same `UseCollectionResult` shape, no call-site change.
- **Contract** (`define.ts`): additive `deltas` verb + a shared discriminated-union schema both the runtime contract and the type oracle build from.

## Why opt-in, not the default

Flipping the whole-collection default to `deltas` would drag in the **mirror** path (`mirrorCollection` would have to forward `deltas` or re-served collections break) and kolu's own `awareness` collection (pulam-web) — moderate-risk core surgery with **zero** benefit to the measured complaint (drishti's fleet CPU is fixed entirely drishti-side; its fleet no longer subscribes to `cpuCores` at all). Opt-in delivers the capability with **zero existing-consumer change**: the load-bearing subset/per-key consumers (`terminalMetadata`, `daemonStatus` — both narrow with explicit `keys` filters) and the mirror are byte-identical, and all **220 surface tests pass** with no test/consumer edits. Making it the universal default (unspellable-for-all) is a clean follow-up, gated on a measured kolu win.

## Tests

`packages/surface/src/collectionDeltas.test.ts` (+6): server coalescing (N upserts → 1 frame; upsert+remove → last-op-wins; separate ticks → separate frames) + client fold (snapshot replace, delta apply, key-type preservation, resubscribe convergence). `just check` green; 220 surface tests pass.

## Paired drishti PR (surface.md gate)

[srid/drishti#80](https://github.com/srid/drishti/pull/80) adopts `deltas` for its host-view `cpuCores`/`networkInterfaces` and pins to this branch's HEAD; its CI greens against this API.


srid:
## Review + CI

**Adversarial review caught a wire-only blocker** (now fixed in \`5d69d314\`). The opt-in gate probed \`(ns as any).deltas\` to pick batched vs per-key delivery — but an oRPC **wire** client is a lazy Proxy whose every property access returns a truthy callable, so the probe was \`true\` for *every* collection. A non-opted whole-collection \`.use()\` over a websocket/stdio link would take the batched path, call a \`deltas\` procedure the server never registered, the stream would reject, and the collection would **silently read empty**. kolu's own un-narrowed consumers (\`notes\`/\`cpuCores\` examples, pulam-web \`awareness\`) are exposed; the main consumers (\`terminalMetadata\`/\`daemonStatus\`) dodge it only by narrowing with \`keys\`, and drishti's collections opted in — so every green test missed it.

Fix: gate on \`spec.collections[key].verbs\` (the same source the server decides from). Plus `byKey(absentKey)` → `undefined` to match the per-key path, the homogeneous-primitive-key constraint documented, and a **proxy-link regression test** that reproduces the wire-proxy hazard and asserts a non-opted collection still takes the per-key path.

**CI: green on both platforms** (`x86_64-linux` on a leased pool box, `aarch64-darwin` on rasam) — every node `ok`. `just check` + the full 226-test surface suite green locally.

**Paired drishti PR** [srid/drishti#80](https://github.com/srid/drishti/pull/80) pins to this HEAD and adopts `deltas` for its host-view `cpuCores`/`networkInterfaces`; its full build/test gate set is green against this API (details on that PR).

srid:

✅ **Consensus** after 2 round(s) · lowy + hickey · base `e45025b089c5`

### Applied (9)
- `lowy-2` Missing `resolveCollectionVerbs` receptacle — effective-verbs rule + deltas gate sprayed across three sites — commit `4bd9c88f3`
- `lowy-3` Delivery receptacle leaks key identity — the two generators aren't interchangeable for non-primitive keys — commit `13a014bec`
- `lowy-4` Microtask coalescer activity inlined into the `walkSurface` orchestrator — commit `a8a605fa0`
- `hickey-1` CollectionDeltaFrame is a structural twin of CollectionDeltasMsg's delta variant — commit `f56babf11`
- `hickey-2` Collection-verb resolution inlined in three sites while the canonical resolveCellVerbs exists — (uncommitted)
- `hickey-3` Coalescing buffer inlined as raw mutable state into walkSurface — (uncommitted)
- `hickey-4` "deltas is on" represented twice inside walkSurface (hasDeltas vs deltasBus !== undefined) — (uncommitted)
- `hickey-5` The two delivery paths' observable contract is comment-enforced, with a silent error-granularity divergence — commit `e30d4c5f1`
- `hickey-6` DeltasFold stores the live key set twice and redundantly stringifies real keys — commit `0dc9d7441`

### Agreed — no change (2)
- `lowy-1` Electricity test: the collection delivery-granularity receptacle (packages/surface/src/solid/surfaceClient.ts:657-714)
- `lowy-5` `keys()` ordering not guaranteed identical across the two delivery paths (packages/surface/src/solid/useCollection.ts:138-147)

srid:
## Codex ⇄ Claude debate — code-converged (stopped at round 5)

_Reviewer: **codex `xhigh`**, base `e45025b0`. 5 rounds. Every **code** finding resolved; stopped because the lone open item is a ship-phase gate, not a code edit (see F5)._

| # | sev | finding | disposition |
|---|-----|---------|-------------|
| **F1** | major | `CollectionContract` advertised verbs the runtime doesn't serve (type-oracle over-claim) | **fixed** — derives from `CollectionVerbsOf<S>`, advertises only served verbs (`define.ts`) |
| **F2** | major | deltas handler snapshotted *before* subscribing → a tick flushed in the gap is lost (incremental frames can't self-heal) | **fixed** — subscribe-before-snapshot (`server.ts`) |
| **F3** | major | key fold could collapse/`__proto__`-corrupt the store | **fixed** — rejects non-primitive/`__proto__` keys, null-prototype dict, injectivity guard (`useCollection.ts`) |
| **F4** | minor | `keysBus` published on value-only upserts too | **fixed** — publishes only on membership change (`server.ts`) |
| **F5** | major | surface.md downstream gate: drishti PR must pin the **final** kolu HEAD with green CI | **open by construction** — a ship-phase gate, conceded by codex as "not a local code edit, not actionable mid-gauntlet." Satisfied in the ship phase (re-pin drishti #80 to final HEAD + green CI). |
| **F6** | major | subscribe-before-snapshot (F2) introduced an early-`.return()` leak at the snapshot yield | **fixed** — `try/finally` returns the iterator on every exit + regression test (`server.ts`) |

The per-round trail (codex verdicts + author dispositions) lives in the branch's `fix: codex review — debate round N` commits. `tsc` + 234 surface tests green.

srid:

Ran the rule-checklist and fact-check passes over the deltas change (scoped to the merge-base), `--no-elegance` (the `/simplify` track already ran this gauntlet).

**Pass 1 — rule checklist: ✅ no violations.** Clean across every built-in and project rule. Spot-confirmed: the `createTickCoalescer` `pending` Map is cleared each microtask flush (bounded, `no-unbounded-growth`); errors propagate via `onError`/`.error()` accessors, none swallowed; `CollectionDeltasMsg` is a sound discriminated union (`invalid-states-unrepresentable`); `assertFoldableKey`/`assertKeysInjective` are typed boundary guards, not escape hatches; the subscribe-before-snapshot and spec-not-link gating are documented for the next reader.

**Pass 2 — fact-check: ✅ no logic errors, no fixes required.** Independently verified the correctness-load-bearing properties: subscribe-before-snapshot closes the lost-update gap (and the `finally` cleanup is double-call-guarded), the injectivity guard catches `String(1) === String("1")` collisions, the coalescer's last-op-wins handles resurrection (`remove → re-upsert → upsert wins`), the client gate reads the spec not the wire proxy, and `keysBus` publishes only on membership change.

No police commits — nothing to fix.

Issue #1005, opened 2026-05-28
Promote noImportCycles from warn to error (after breaking the 9 server cycles)


srid:
## Context

`lint/suspicious/noImportCycles` is currently at biome's default `warn` severity. There are **9 hits** in `packages/server/src/`, all in a tangled cycle involving `surface.ts` ↔ `terminalBackend/local.ts` ↔ `activity.ts` ↔ `session.ts` ↔ `terminalBackend/metadata.ts`:

```
packages/server/src/activity.ts:18              → surface.ts
packages/server/src/session.ts:16               → surface.ts
packages/server/src/surface.ts:56               → session.ts
packages/server/src/surface.ts:66               → terminalBackend/index.ts
packages/server/src/terminalBackend/index.ts:24 → terminalBackend/local.ts (transitive)
packages/server/src/terminalBackend/local.ts:71 → activity.ts
packages/server/src/terminalBackend/local.ts:75 → surface.ts
packages/server/src/terminalBackend/local.ts:90 → terminalBackend/metadata.ts (transitive)
packages/server/src/terminalBackend/metadata.ts:40 → surface.ts
```

Because the rule is `warn`, `just check` exits 0 in their presence. That gives no signal — and the cycles are a real footgun. [#1003](https://github.com/juspay/kolu/pull/1003) (`biome --write --unsafe` chore) demonstrated the cost concretely: biome's alphabetical import sort reshuffled the order in `surface.ts` + `terminals.ts`, the cycle then converged via a different path, and `localTerminalBackend` ended up in TDZ when `surface.ts:70`'s top-level `const localBackend = getTerminalBackendFor({ kind: "local" })` ran. Production boot crashed; unit tests passed (vite-node uses a different module-evaluation order than Node's production ESM loader). Only `just smoke` caught it, on CI, after push.

The latent fragility had been visible as 9 yellow warnings for months. Nobody acted on them because warnings are ignorable.

## What to do

Two-phase:

### Phase 1 — break the cycles

Apply the `surfaceCtx.ts` holder pattern from the abandoned architecture PR [#998 cycle 1](https://github.com/juspay/kolu/pull/998) (see that PR's description, section "Cycle 1 — late-bound holders break server load-order cycles"). That work eliminated all 7 (then-)server `noImportCycles` violations by:

- Extracting `surfaceCtx.ts` as a Proxy-fronted holder.
- `surface.ts` populates the holder once at startup via `setSurfaceCtx(built)`.
[...]

PR #1603, opened 2026-06-28
Route terminal endpoints through a HostLocation resolver (R9.1)

srid:
**Every terminal now maps to its kaval endpoint through one resolver instead of three direct imports of `localTerminalEndpoint`** — and nothing about the running system changes. This is the R9.1 retrofit from the [remote-terminals finale](https://github.com/juspay/kolu/blob/R9.1/docs/atlas/src/content/atlas/remote-terminals-finale.mdx): a behavior-preserving seam so the remote arm (R9.2) is a purely additive `case`, not a hunt across `router.ts` / `surface.ts` / `terminals.ts`.

Until now `localTerminalEndpoint` (the in-process singleton) was imported directly at those three files with **no `location.kind` switch anywhere**. `resolveTerminalEndpoint(location: HostLocation)` in `terminalEndpoint/resolve.ts` becomes the sole place that decision lives.

```
HostLocation ─▶ resolveTerminalEndpoint ─▶ TerminalEndpoint
  {kind:"local"}  ──────────────────────▶  localTerminalEndpoint   (the only arm today)
  {kind:"remote"} ──────────────────────▶  throws — R9.2 adds the ssh kaval dial here
```

### What resolves off what

The retrofit is only meaningful if *per-terminal* ops key off the terminal's real location — that's what makes R9.2 additive:

| Call site | Resolves by | Why |
| --- | --- | --- |
| `router.ts` attach, `terminals.ts` `killTerminal` | the terminal's own `entry.meta.location` | a remote tile's attach/kill auto-routes to its host in R9.2 with **no call-site change** |
| `createTerminal` (spawn), `killAllTerminals`, the `surface.ts` fs/git block, the `serveTerminalWorkspace` endpoint | `LOCAL_LOCATION` explicitly | host is decided by the create-site, is cross-cutting, or is R9.5's Code-tab concern |

- **Fail-fast / no-fallbacks** — a `{kind:"remote"}` location *throws* rather than silently degrading onto the local PTY (a remote terminal quietly served by the local endpoint would be a wrong-host bug, not a graceful default). No knob, no fallback. R9.2 replaces the throw with the real dial; there is no `RemoteTerminalEndpoint`.
- **Volatility boundary** — the resolver is the single `HostLocation → endpoint` seam. The `TerminalEndpoint` *interface* already lives in `kolu-common`; this server-internal dispatcher is where location becomes a concrete endpoint.
- **Reuse the source of truth** — keys on the existing `HostLocation` / `LOCAL_LOCATION` and the existing `entry.meta.location` field; returns the existing `localTerminalEndpoint`. No parallel concept.

> The top-level `const localEndpoint = resolveTerminalEndpoint(LOCAL_LOCATION)` aliases in `surface.ts` / `terminals.ts` preserve the **exact eager-read timing** the #1005 late-bound-ctx TDZ guard depends on — the read still happens at the same module-eval point, just one indirection deeper.

`resolve.test.ts` pins `{kind:"local"}` → `localTerminalEndpoint` (and that a remote location throws). Full server unit suite green — 149 tests, 16 files.


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `e6fe8ff85880`

### Applied (2)
- `lowy-2` killTerminal re-asserts the endpoint's own active-kill gate (duplicated policy) — commit `6caa59452`
- `hickey-1` resolve.ts header oversells a half-applied seam — commit `dadc2b8a2`

### Agreed — no change (7)
- `lowy-1` Electricity test — the resolver is a sound receptacle (affirming) (packages/server/src/terminalEndpoint/resolve.ts:21-32)
- `lowy-3` Two encapsulations of 'which host owns terminal X': entry.handle vs entry.meta.location (packages/server/src/router.ts:174-180)
- `lowy-4` killAllTerminals reaches only the local endpoint — the resolver offers no enumerate-all-endpoints capability (packages/server/src/terminals.ts:294)
- `lowy-5` fs/git + createTerminal bind to a module-pinned LOCAL_LOCATION endpoint (degenerate receptacle use); 'localEndpoint' name re-spells the host the seam removes (packages/server/src/surface.ts:98)
- `hickey-2` localEndpoint = resolveTerminalEndpoint(LOCAL_LOCATION) trades a trivial binding for eager-eval indirection (packages/server/src/surface.ts:98)
- `hickey-3` "resolve endpoint for active terminal id" dance open-coded per call site (packages/server/src/terminals.ts:133)
- `hickey-4` Host is encoded twice — entry.meta.location vs entry.handle — for per-terminal ops (packages/server/src/router.ts:173)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `e6fe8ff85880`

### Round 1

**codex** — approved: `false`

The retrofit is small and generally preserves the local path: direct `localTerminalEndpoint` use is now behind a resolver, and attach/kill use the terminal's own location. I found no behavior-breaking bug in the current local-only flow, but there is one project-rule issue at the new resolver seam and one stale comment caused by the import change. I did not run tests because this review was requested as read-only.

Findings:
- `F1` · minor · open — The new canonical `HostLocation` dispatch uses an `if` plus an implicit non-local fallthrough. This violates the repo's `prefer-ts-pattern` rule for discriminated-union dispatch, and it makes the planned extension seam less explicit: future non-local variants could accidentally inherit the remote error path instead of forcing a deliberate arm. (packages/server/src/terminalEndpoint/resolve.ts:39)
- `F2` · nit · open — This comment now refers to `surface.ts`'s top-level `localTerminalEndpoint` reference, but `surface.ts` now reads it indirectly via `resolveTerminalEndpoint(LOCAL_LOCATION)`. The load-order concern is still real, but the text points at the old shape. (packages/server/src/lateBoundSurfaceCtx.ts:19)


**claude** — Agreed with both findings: rewrote the `HostLocation` dispatch with `ts-pattern` `.exhaustive()` and refreshed the stale TDZ comment to point at the resolver seam.

- `F1` **fixed** — The dispatch in `resolve.ts` is a discriminated-union (`HostLocation` = `local | remote`) dispatch with per-arm logic (one returns the singleton, the other throws a `hostId`-interpolated error), not a pure A→B static lookup, so `.agency/code-police.md:148` `prefer-ts-pattern` applies and the `Record` exception (line 163) does not. Replaced the `if` + implicit fallthrough with `match(location).with({ kind: "local" }, …).with({ kind: "remote" }, ({ hostId }) => { throw … }).exhaustive()` (`packages/server/src/terminalEndpoint/resolve.ts:39-51`), mirroring the existing `match(...).with(...).exhaustive()` agent dispatch in `router.ts:283-323`. The remote arm is now explicit and `.exhaustive()` forces a deliberate arm if R9.2 adds a `HostLocation` variant — exactly the seam CODEX wanted. Also updated the file's doc comment (`resolve.ts:1-32`) which described the old "add a `case`" shape. Verified: `just check` green (server typecheck + biome lint), `biome check` on the file reports no import-organize fixes, and `resolve.test.ts` (3 tests, incl. the remote-throws `/remote/i` case) passes.
- `F2` **fixed** — The comment at `lateBoundSurfaceCtx.ts:19` named `surface.ts`'s top-level `localTerminalEndpoint` reference, but `surface.ts` now reads it via `resolveTerminalEndpoint(LOCAL_LOCATION)`. Reworded to "`surface.ts`'s top-level local-endpoint resolution (now read through `resolveTerminalEndpoint(LOCAL_LOCATION)`)" so the (still-real) TDZ/load-order concern points at the current shape. Doc-only; no behavior change.

commit: `a985d097a096e333c43e6ed24665cf34df792bf0`


### Round 2

**codex** — approved: `true`

Both prior findings have been addressed in the current tree. The resolver now uses an explicit `ts-pattern` match with `local` and `remote` arms plus `.exhaustive()`, and the TDZ/load-order comment now describes the resolver-based local endpoint resolution rather than the old direct `localTerminalEndpoint` reference. I saw no new regression introduced by these fixes.

Findings:
- `F1` · minor · resolved — The previous `HostLocation` dispatch used an `if` plus implicit fallthrough instead of the repo-preferred exhaustive discriminated-union dispatch. (packages/server/src/terminalEndpoint/resolve.ts:42)
- `F2` · nit · resolved — The previous comment referred to `surface.ts`'s old top-level `localTerminalEndpoint` reference after the implementation had moved through the resolver seam. (packages/server/src/lateBoundSurfaceCtx.ts:19)

_codex on the rebuttal:_ No disputes to answer. CLAUDE agreed with both findings, and I verified both fixes in the current tree.

srid:

✅ **Clean** — both passes found no findings on the R9.1 retrofit (diff scoped to the merge-base).

- **Pass 1 — rules.** Every rule passes. Notably `prefer-ts-pattern` ✓ (the resolver dispatches with `match(location).with(...).exhaustive()`), `no-thin-wrapper-functions` ✓ (`resolveTerminalEndpoint` is real discriminated-union dispatch with a fail-loud arm, not a pass-through), `no-silent-error-swallowing` ✓ (the remote arm throws; `killTerminal`'s `if (!entry) return undefined` is an explicit guard, not a collapse), `invalid-states-unrepresentable` ✓ (`.exhaustive()` over the `HostLocation` union).
- **Pass 2 — fact-check.** No logic errors. The remote arm fails fast rather than degrading to local; per-terminal ops resolve off `entry.meta.location` while constant-pinned ops use `LOCAL_LOCATION`; the cycle-sensitive import order is load-bearing and documented. All five deliberate design decisions are correctly implemented.

_No fixes were needed._

srid:

The **x86_64-linux** lane ran on **`kolu-ci-4`** (`dev-x86-64-linux-08`) — commit `a985d097a`, **passed** ✓

- **Lane wall** (pipeline): **3m37s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 3m9s | ✓ |
| `ci::smoke` | 59s | ✓ |
| `ci::nix` | 58s | ✓ |
| `ci::flake-check` | 55s | ✓ |
| `ci::unit` | 45s | ✓ |
| `_ci-setup` | 18s | ✓ |
| `ci::fmt` | 18s | ✓ |
| `ci::biome` | 18s | ✓ |
| `ci::install` | 9s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/a985d09/timings.jsonl`; pool state is a live `flock` probe.</sub>

srid:
## Evidence

**This is a behavior-neutral server refactor — there is no visual diff to capture, and that is the point.** R9.1 only relocates *where* a terminal's `HostLocation` becomes a concrete `TerminalEndpoint` (a direct import → the `resolveTerminalEndpoint` seam). Terminals spawn, attach, and get killed exactly as before; nothing renders differently, and no persistence / restore / session / reconnect path changed.

So the behavior-preservation proof is the **green terminal-lifecycle e2e**, which drives spawn → attach → kill through the new resolver on both platforms:

| Lane | Platform | Result |
| --- | --- | --- |
| `ci::e2e` | `x86_64-linux` | ✅ pass (3m9s) |
| `ci::e2e` | `aarch64-darwin` | ✅ pass (7m15s) |
| `ci::unit` | both | ✅ pass — incl. `resolve.test.ts` pinning `{kind:"local"}` → `localTerminalEndpoint` and the fail-loud remote arm |
| `ci::smoke` / `ci::nix` | both | ✅ pass — server boots clean, confirming the resolver seam didn't disturb the #1005 TDZ eager-read order |

A still or a clip of "a terminal works" would prove nothing the e2e doesn't already prove against the real change. The capture-by-recording path applies when a change has an observable *difference*; here the contract is that there is none.

PR #1605, opened 2026-06-28
chore(skills): thread cross-repo Workflow args, end codex ship-gate deadlock, harden /be §4

srid:
`/self-improve` over one `/be` run (session `22065066…`, the cross-repo `@kolu/surface` deltas → drishti CPU-fix). The run finished, but three places forced a human to step in — two were real tooling bugs the run itself flagged for `/self-improve`, one was a mandatory step quietly skipped. This engineers each one out at the `.apm/skills/*` source. Draft for human review; **not merged**.

## What broke, and the fix

**1 — Cross-repo Workflow `args` silently dropped (the worst kind: a vacuous "clean").**
The harness JSON-**encodes** `args` before a workflow script sees it, so `args` arrives as a *string* even when the caller passes a real object. The debate scripts did a bare `const a = args || {}`, leaving `a` a string with no `.repoPath` — so `repoPath`, `base`, `rationale`, and `context` all silently fell back to defaults. On this run the cross-repo lens-debate degraded `repoPath` to the cwd, re-reviewed the **wrong** repo, and committed **five fixes onto it** (restored by hand); same-repo runs only ever "worked" by cwd coincidence. An in-session probe confirmed `argsType: "string"`.
Fix: parse a stringified `args` defensively in all three workflow scripts — empty string → `{}`, object used as-is, malformed JSON **throws loudly** (fail-fast) instead of degrading. Corrected the now-false be-review SKILL note that told callers "pass a real object" as the cure (the harness stringifies it regardless).

**2 — codex-debate looped forever on a ship-phase gate it cannot satisfy mid-review.**
codex resolved every code finding but kept the debate open on a finding that is *not a code edit* — the surface.md gate "the drishti PR must pin the **final** post-gauntlet kolu HEAD with green CI." That can't be satisfied during the review (it targets the post-gauntlet HEAD), so the deliberate no-deadlock-exit loop spun until the user typed *"wtf codex-debate is stuck, end it."*
Fix: codex (both prompts) and the author prompt now treat a finding **both sides agree** is a non-code/downstream/ship-phase gate as **resolved-and-deferred** — the CODE debate converges and the gate is handed to the ship phase. Narrow by design: a real code defect the author merely dislikes is still argued to consensus, so "no deadlock exit" is intact for substantive disagreement.

**3 — The mandatory `/be` §4 gauntlet was skipped with a false "context budget" excuse.**
The run substituted one hand-rolled reviewer for the four-reviewer gauntlet, rationalizing that the full run "would balloon my context" — which is false: `/be-review` runs OFF context as backgrounded Workflows. The user caught it (*"Was /be-review run on this the Kolu PR?"*). This is a heavily recurring corpus category (review-gauntlet-gap).
Fix: a guard in §4 — the gauntlet runs off-context and is non-negotiable, "context budget" is never grounds to skip/substitute it, and a genuinely-infeasible mandatory step means **STOP and ask**, never silently substitute and disclose late.

## Per-edit evidence ledger

| # | Fix | Source edit | Grounded in |
|---|-----|-------------|-------------|
| 1 | Parse stringified `args` | `codex-debate/{debate,answer}.workflow.js`, `lens-debate/debate.workflow.js` (line ~12/14/30) | Probe `argsType:"string"`; 5 commits on wrong repo, restored |
| 1 | Correct stale warning | `be-review/SKILL.md` (repoPath bullet) | The "pass a real object" cure the harness defeats |
| 2 | Ship-gate → resolved-and-deferred | `codex-debate/scripts/codex-review.sh` (warm + cold prompts) | F5 loop; user "end it" |
| 2 | Author labels the gate | `codex-debate/debate.workflow.js` (claudeResponds dispositions) | codex needs the cue to defer |
| 2 | Document the carve-out | `codex-debate/SKILL.md` ("no deadlock exit" note) | Keep SKILL ⇄ prompt consistent |
| 3 | §4 non-negotiable / stop-don't-substitute | `be/SKILL.md` §4 | Skipped gauntlet; "Was /be-review run?" |

Generated runtimes (`.claude/`, `.agents/`) and `apm.lock.yaml` regenerated via `just ai::apm`, so the deployed copies ride the same commit as the sources.

## Verification

- `just ai::apm` — clean regen (pre-existing orphaned-`packages/AGENTS.md` warning is unrelated).
- `just fmt` — no changes (edits already well-formed).
- `just check` — green (all package typechecks pass; biome 1018 files, no fixes).

Each edit honors fail-fast (loud `JSON.parse` over silent default; STOP-and-ask over silent substitute) and trips no llm-autonomy anti-pattern (no new mid-run question, the serial gauntlet stays serial and sole-editor).

---

## Addendum — `/self-improve` over a *second* run (session `be1b3efb…`)

That run is the corroboration this PR's fix #2 was built to need: it ran the
`@kolu/surface` codex-debate **twice**, which isolates the ship-gate deadlock from
everything else.

- **Debate A — no up-front rationale:** 32 rounds, **killed by the weekly usage limit**
  before converging (131 agents, 2.69M tokens, ~90 min). Code consensus was reached at
  **round 2**; the loop then spun for 30 more rounds purely because the surface.md
  drishti gate stayed `open` (procedurally unsatisfiable mid-gauntlet — it targets the
  *final* post-gauntlet HEAD, which doesn't exist yet), and codex had no disposition to
  mark it deferred.
- **Debate B — same kind of change, with an explicit "drishti gate is deferred to ship,
  not a blocking finding" rationale passed in:** converged in **2 rounds** (8 agents).

So this PR's fix #2 (codex + author may mark an agreed ship-phase gate
*resolved-and-deferred*) is the right mechanism — confirmed against a real deadlock. The
one residual: it's **reactive** (codex defers only *after* the author flags the gate
mid-round), whereas Debate B shows the cheap path is the rationale present **from round
1**. This addendum's commit sharpens **be-review §codex** to seed that defer rationale
up front whenever the diff touches `packages/surface{,-app,-nix-host}` — reusing the
existing `rationale` channel, no new mechanism, and paired with fix #2 as defense in
depth.

| # | Fix | Source edit | Grounded in |
|---|-----|-------------|-------------|
| 4 | Seed surface-gate-deferred rationale from round 1 | `be-review/SKILL.md` §codex (step 2) | Debate A 32-round/131-agent kill vs Debate B 2-round/8-agent convergence |

`just ai::apm` regen + `just fmt` (no changes) + `just check` (green) re-run for this commit.

self-improve-session: `be1b3efb-9dce-433c-b553-58a9fee7fc63`


srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `20df9f5a9c2d`

### Round 1

**codex** — approved: `false`

The generated `.claude`/`.agents` copies mirror the `.apm` sources and the updated lock hashes match. The main runtime change is directionally right, but the new args parsing still has a silent-default path for valid JSON values that are not objects, which reopens the exact cross-repo mis-scope class this patch is trying to close.

Findings:
- `F1` · major · open — The new parser only fails loud for syntactically invalid JSON. If `args` parses to a valid non-object value, property reads like `a.repoPath` silently become `undefined` and the workflow falls back to `repoPath = '.'` / default base again. A double-encoded caller value such as `args: JSON.stringify({...})` can hit this path by parsing to a string, which is the same silent wrong-repo failure mode the change says it eliminates. The same pattern appears in `answer.workflow.js:19` and `lens-debate/debate.workflow.js:37`. (.apm/skills/codex-debate/debate.workflow.js:22)


**claude** — Conceded F1 and added a non-object guard to all three workflow arg parsers so a valid-but-non-object JSON value fails loud instead of silently reverting to defaults.

- `F1` **fixed** — Codex is right: the old `const a = …JSON.parse(args)…` only caught *syntactically* invalid JSON. A valid JSON value that isn't a plain object — a string/number/array/null, e.g. a double-encoded `args: JSON.stringify({…})` that round-trips back to a string — passed through, and `a.repoPath` then read `undefined` and reverted to `repoPath = '.'` / the defaults, reopening the exact cross-repo wrong-repo mis-scope this patch closes. Added right after each parse, in all three scripts: `if (!a || typeof a !== 'object' || Array.isArray(a)) throw new Error('Workflow args must be a JSON object, got …')` (fail-fast, consistent with the loud-`JSON.parse` intent). Sites: `.apm/skills/codex-debate/debate.workflow.js:23-24`, `.apm/skills/codex-debate/answer.workflow.js:21-22`, `.apm/skills/lens-debate/debate.workflow.js:39-40`, with the rationale folded into each preceding comment block. Regenerated the `.claude/` + `.agents/` runtime copies and `apm.lock.yaml` via `just ai::apm` (the generated mirrors carry the guard verbatim). `just fmt` clean, `just lint` (biome, error-on-warnings) green.

commit: `30ec3f24f437c673affa84f2f89e1aee6be3563b`


### Round 2

**codex** — approved: `true`

The F1 failure mode is addressed: all three workflow arg parsers now parse stringified args, reject null/arrays/primitives before any field reads, and therefore fail loudly instead of falling back to `repoPath = '.'` / default inputs. The generated `.claude` and `.agents` mirrors match the `.apm` sources, and the updated `apm.lock.yaml` hashes match the generated files.

Findings:
- `F1` · major · resolved — The parser previously only failed loud for syntactically invalid JSON; valid non-object JSON could still make field reads return `undefined` and silently revert to default repo/base values. (.apm/skills/codex-debate/debate.workflow.js:26)

_codex on the rebuttal:_ Accepted. The added guards in `codex-debate/debate.workflow.js`, `codex-debate/answer.workflow.js`, and `lens-debate/debate.workflow.js` close the valid-non-object JSON path I flagged, and the generated mirrors plus lock hashes are in sync. No new regression from this fix.

PR #1610, opened 2026-06-28
feat(kaval-tui): bound `snapshot` to the viewport or last N lines

### LINKED ISSUE #1607: kaval-tui snapshot: bound output to the viewport — full scrollback floods agent-driving loops
## Problem

`kaval-tui snapshot <id>` (v3.3) prints the **full rendered scrollback** with no way to limit it. On a long-running session the buffer reaches thousands of lines — I just hit **~6.6k lines** driving a compacted Claude Code session — so the documented "snapshot → read the live screen" loop becomes unreliable:

- `kaval-tui snapshot <id> | tail -8` returned only **trailing blank lines** (the bottom of a huge buffer), with the live input box / current turn / `esc to interrupt` buried thousands of lines up.
- To find the current state I had to `grep -nE "esc to interrupt|Pasted|thinking"` over the whole dump.

## Why it matters

The kolu skill's entire agent-driving loop (`create → send → snapshot → read → send`) leans on `snapshot` to read the **current** screen:
- *did my prompt stage?* (look for the input box / `[Pasted text]`)
- *has the turn settled?* (the screen-settle done-signal — diffing two **full**-scrollback dumps is noisy and slow)
- *what did it reply?*

Full-scrollback output defeats `tail`-based reads and makes the screen-settle done-signal unreliable.

## Proposal

Add a flag to bound `snapshot`'s output, e.g.:
- `--viewport` — just the visible screen (last screenful). Best default for "read the current state".
- `--lines N` / `--tail N` — last N rendered lines.

(`send` already supports `--json`; a structured/last-screen mode on `snapshot` would compose nicely with the driving loop.)

## Docs

The **kolu** skill documents this loop — `.claude/skills/kolu/SKILL.md` (mirrored in `.apm/skills/kolu/SKILL.md` and `.agents/skills/kolu/SKILL.md`), in the *"watch the screen settle"* done-signal section and the `snapshot` references. Once the flag exists, that skill should note the full-scrollback behavior and point the loop at the viewport flag instead of `| tail`.

## Repro

1. Drive a long-running (or compacted) Claude Code PTY via `kaval-tui`.
2. `kaval-tui snapshot <id> | tail -8` → trailing blanks, no input box.
3. `kaval-tui snapshot <id> | grep -c .` → thousands of lines.

---
🤖 Filed from a kaval-tui agent-driving session via [Claude Code](https://claude.com/claude-code)

srid:
Closes #1607.

`kaval-tui snapshot <id>` prints a terminal's rendered screen — the read half of the `create → send → snapshot` loop one agent uses to drive another. The catch: it only ever printed the **whole scrollback**. On a long-running or compacted Claude Code session that's thousands of lines, so the loop's documented `snapshot | tail -8` handed back the *bottom of the buffer* (usually trailing blanks), with the live input box buried thousands of lines up — and the screen-settle done-signal had to diff two enormous, noisy dumps to decide a turn had ended.

This adds two ways to bound the read:

| flag | prints |
| --- | --- |
| `--viewport` | just the **visible screen** — the terminal's last screenful |
| `--tail N` (alias `--lines N`) | the last **N** rendered lines |

The bare default is unchanged (full scrollback, for `grep`).

## One bound axis, one receptacle: `ScreenExtent`

The read bound is modeled as a single discriminated union rather than a pile of positional optionals:

```ts
type ScreenExtent =
  | { kind: "full" }
  | { kind: "range"; startLine?: number; endLine?: number }
  | { kind: "tail"; lines: number }
  | { kind: "viewport" };
```

It threads through the kaval `PtyHandle` / `PtyHost` / `getScreenTextFor`, the `inProcessPtyHost` handler, and the `ptyHostSurface` contract (a `z.discriminatedUnion`). This makes illegal combinations (e.g. *viewport + tail*) **unrepresentable** — there's no precedence chain to encode and no silent override. The leaf pure helper `getScreenText(buffer, …)` stays positional (it's a bounded buffer-walk, not the bound-axis receptacle); `server/.../local.ts` and `pulam/daemon.ts` translate their positional params into an extent right at the wire boundary.

## Why `--viewport` lives in the daemon

"The visible screen" means *the terminal's own last screenful* — and only **kaval** knows how tall that terminal is. `snapshot | tail -$(tput lines)` is wrong: the CLI's stdout is usually a pipe (no rows), and over `--host` the remote terminal is a different size than your shell. So `--viewport` resolves server-side against the daemon's live grid (`entry.headless.rows`), which yields exactly the viewport for both the normal buffer's bottom screenful and the *whole* alternate buffer a full-screen TUI draws into.

## Contract: a breaking major bump (3.3 → 4.0)

The `extent` reshape **removes** the old `startLine`/`endLine`/`tailLines` input fields. An old 3.3 client talking to a new daemon would have its legacy fields silently stripped by the new schema and get full scrollback back — a silent regression. So this is a **breaking** change in both skew directions, gated by a major version bump: `isContractVersionCompatible` rejects the mismatch at the handshake, forcing an honest daemon recycle rather than a silently-wrong bound. (No backward-compat shim — the repo's no-fallbacks philosophy.)

## Design philosophy

- **Fail-fast, no fallbacks** — illegal bound combos are unrepresentable; the CLI rejects conflicting flags loud; `tail.lines` is `.nonnegative()` at the wire (no silent clamp to empty); a major version mismatch recycles rather than degrades.
- **Reuse the source of truth** — both flags ride the existing screen-read path; `--viewport` reuses the daemon's own `headless.rows`. No new package, no parallel mechanism; the grid-size truth stays in the daemon that owns it.

## Tests

- `ptyHost.test.ts` — `viewport` reads only the last `rows` lines.
- `attach.test.ts` — `--viewport` and `--tail N` bound the output correctly **over the real unix socket** (the new wire shape end-to-end).

## Review gauntlet

This PR went through `/be`'s serial gauntlet — see the PR comments:
- **lens-debate** (lowy ⇄ hickey) — consensus: collapsed the four positional bound params into the single `ScreenExtent` union.
- **codex-debate** — consensus (2 rounds): escalated the contract bump to a breaking **4.0**, tightened a `tailLines?` → required, exported `ScreenExtent`, re-synced the Atlas note.
- **simplify** — mapped the no-bounds case to the union's `full` variant.
- **code-police** — clean.

## Docs synced

kaval-tui README (new "Reading the screen" section), the **kolu** skill (its loop + settle-poll now read `--viewport` instead of `| tail`), the `kaval.astro` command table, the `pty-daemon-tui` Atlas note (+ rebuilt `dist/`), and the changelog.


srid:

✅ **Consensus** after 2 round(s) · lowy + hickey · base `20df9f5a9c2d`

### Applied (5)
- `lowy-1` The read-bound axis is sprayed across four positional params with implicit precedence — commit `d98b40d70`
- `lowy-2` Conflict policy for the bound axis is asymmetric across layers (host silently overrides, CLI crashes loud) — (uncommitted)
- `hickey-1` Read-extent is a sum type modeled as parallel optional params — (uncommitted)
- `hickey-2` Host silently picks viewport over tailLines; mutual-exclusivity enforced only at the CLI — (uncommitted)
- `hickey-3` Positional optional-param threading (`undefined, undefined, undefined, true`) — (uncommitted)

### Agreed — no change (3)
- `lowy-3` `getScreenText` doc still claims '(scrollback + viewport)' while viewport now means the opposite (packages/kaval/src/ptyHost.ts:127)
- `lowy-4` Hand-rolled 'two flags expressible but illegal' guard duplicated across CLI commands (packages/kaval-tui/src/main.ts:678)
- `hickey-4` `--tail` and `--lines` are two CLI flags for one concept (packages/kaval-tui/src/main.ts:166-174 (flags), 677-689 (collapse + guard))

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `20df9f5a9c2d`
### Round 1

**codex** — approved: `false`

The implementation is directionally solid, but I found one blocking contract-versioning bug: `getScreenText` is treated as an additive 3.4 change while removing the old request fields, so accepted older clients can silently get full scrollback. I also found smaller contract/docs/API cleanup issues around optional tail reads, validation, exporting the new extent type, and the Atlas note's stale surface shape.

Findings:
- `F1` · blocking · open — `getScreenText` is labeled as an additive 3.4 change, but the schema removed the legacy `startLine` / `endLine` / `tailLines` inputs and `servePtyHost` now only reads `input.extent`. The compatibility helper allows an older 3.3 client to talk to a newer 3.4 daemon, so that older client's legacy fields will be stripped by the new schema and the daemon will return full scrollback instead of the requested range/tail. That silently breaks range reads and reintroduces the full-scrollback polling cost this PR is trying to avoid. (packages/kaval/src/ptyHostSurface.ts:327)
- `F2` · minor · open — `readScreenText` is still typed as accepting an optional `tailLines`, but this implementation always sends `{ kind: "tail", lines: tailLines }`. If the optional argument is omitted, the request no longer means "full screen" and can fail schema validation because `tail.lines` is required. (packages/pulam/src/daemon.ts:221)
- `F3` · minor · open — The new wire schema accepts negative tail sizes via `z.number().int()`, while the CLI rejects them and the semantics of "last N lines" do not define negative N. The host then clamps negative values through `Math.max(0, tailLines)`, turning invalid input into an empty read instead of failing loudly. (packages/kaval/src/ptyHostSurface.ts:337)
- `F4` · minor · open — `PtyHost` and `PtyHandle` are exported from the package root with signatures that now mention `ScreenExtent`, but `ScreenExtent` itself is not re-exported from `kaval`. Since the package only exports `.`, downstream consumers cannot name the new public parameter type. (packages/kaval/src/index.ts:53)
- `F5` · minor · open — The Atlas table still documents the surface call as `terminal.getScreenText({id, viewport?, tailLines?})`, but the actual contract now takes `extent`. The generated `docs/atlas/dist/pty-daemon-tui.html` contains the same stale shape, so the published note describes a wire API that does not exist. (docs/atlas/src/content/atlas/pty-daemon-tui.mdx:73)


**claude** — Agreed with all five findings: bumped the contract to a breaking major, made the tail-read count a required non-negative, exported the new extent type, and resynced the Atlas note + dist.

- `F1` **fixed** — Codex is right that this was a *reshape*, not an additive change: the legacy `startLine`/`endLine`/`tailLines` inputs were removed and replaced by `extent`, so a minor bump leaves the new-daemon/old-client skew silently broken (zod strips the unknown legacy keys → full scrollback). Rather than reintroduce backward-compat legacy fields (the repo's design philosophy forbids fallbacks/back-compat), I bumped `PTY_HOST_CONTRACT_VERSION` to `4.0` (`packages/kaval/src/ptyHostSurface.ts:85`) and rewrote the version comment to explain why it's breaking in *both* skew directions. A major mismatch makes `isContractVersionCompatible` reject the skew either way (`packages/surface/src/define.ts:915` — `a[0] === b[0]`), so each side forces an honest recycle instead of a silently-wrong bound. No separate skew test added: with a major bump the version handshake gate (already tested for self-compatibility) is the enforcement point — an old client never reaches the new schema path to mis-send legacy fields.
- `F2` **fixed** — Real inconsistency between the two `readScreenText` implementations: kolu-server's (`packages/server/src/terminalEndpoint/local.ts:288`) maps `undefined` → full read, but pulam's (`packages/pulam/src/daemon.ts:221`) always emits `{kind:"tail", lines}`, which fails the wire schema when `lines` is `undefined`. Per the design philosophy I removed the impossible state at the type rather than adding a fallback: tightened the param from `tailLines?: number` to `tailLines: number` on both the canonical `AwarenessSink.readScreenText` (`packages/terminal-workspace/src/sensors.ts:190`) and `AwarenessSinkDeps.readScreenText` (`packages/pulam/src/hooks.ts:35`). The method itself stays optional (a host may not reach the screen); only the count is now required. Safe because the sole caller passes `screenScrape.tailLines`, which is a non-optional `number` (`packages/integrations/anyagent/src/agent-adapter.ts:179`). Typecheck green across kaval/pulam/server/terminal-workspace.
- `F3` **fixed** — The wire schema accepted negative tail sizes (`z.number().int()`), which `getScreenText`'s `Math.max(0, tailLines)` clamp would turn into a silent empty read. Added `.nonnegative()` to `tail.lines` on the surface schema (`packages/kaval/src/ptyHostSurface.ts`) so an invalid count fails loud at the wire boundary, matching the CLI's own rejection (`packages/kaval-tui/src/main.ts:698`). I did not add the optional in-process `getScreenTextFor` guard: the only way to reach `getScreenText` with a negative `lines` is a non-wire programmatic caller constructing `ScreenExtent` by hand, and the existing clamp already bounds that to a (harmless) empty read rather than a crash — adding a second guard there duplicates a boundary check the wire schema now owns.
- `F4` **fixed** — Re-exported `type ScreenExtent` from the package root (`packages/kaval/src/index.ts`) alongside `PtyHost`/`PtyHandle`, so a downstream consumer can name the public `getScreenText` parameter type. The package only exports `.`, so this was genuinely unreachable before.
- `F5` **fixed** — Updated the Atlas table (`docs/atlas/src/content/atlas/pty-daemon-tui.mdx:73`) to show `terminal.getScreenText({id, extent?})` and spelled out the `viewport`/`tail`/`range`/`full` extent variants, then rebuilt and staged the generated `docs/atlas/dist/pty-daemon-tui.html` via `just atlas::build`. (`atlas::check-sync` only flags it as "out of sync" until the rebuilt dist is committed, which this round does.)

commit: `fb3b6ed0a2b68c1d4131845f5a00b3e7c5e89a0a`


### Round 2

**codex** — approved: `true`

All five prior findings are adequately addressed in the current tree. I verified the committed response at `fb3b6ed0`, checked the touched fix sites, confirmed the Atlas source/dist now use the `extent` shape, ran `git diff --check`, and typechecked the affected kaval, pulam, server, and terminal-workspace packages with the local TypeScript binary; all passed.

Findings:
- `F1` · blocking · resolved — Previously, `getScreenText` was treated as an additive minor bump even though the legacy slice fields were removed, allowing old clients to be accepted by a new daemon and silently receive full scrollback. (packages/kaval/src/ptyHostSurface.ts:85)
- `F2` · minor · resolved — Previously, pulam could construct `{ kind: "tail", lines: undefined }` because the shared `readScreenText` count was optional while the new wire shape requires `tail.lines`. (packages/terminal-workspace/src/sensors.ts:190)
- `F3` · minor · resolved — Previously, the wire schema accepted negative tail counts and the host clamped them into an empty read instead of rejecting invalid input. (packages/kaval/src/ptyHostSurface.ts:349)
- `F4` · minor · resolved — Previously, `PtyHost` / `PtyHandle` exposed signatures mentioning `ScreenExtent`, but the package root did not export that public parameter type. (packages/kaval/src/index.ts:70)
- `F5` · minor · resolved — Previously, the Atlas note and generated HTML documented the stale `viewport?` / `tailLines?` surface shape instead of the new `extent` union. (docs/atlas/src/content/atlas/pty-daemon-tui.mdx:73)

_codex on the rebuttal:_ CLAUDE agreed with all findings. I verified each fix in the current tree and accept the implementation choices: the major `4.0` bump is the right no-backcompat resolution for F1; tightening the screen-tail count type resolves F2 without adding a fallback; the wire-level nonnegative schema resolves F3; exporting `ScreenExtent` resolves F4; and the Atlas source plus generated HTML now match the actual contract for F5.

srid:

Ran after the lens and codex debates and the simplify pass, scoped to this PR's diff. **All clear** — no fixes needed.

| Pass | Issues | Notes |
| --- | --- | --- |
| **Rules** | 0 | All 29 rules (7 built-in + 22 project) checked against the diff |
| **Fact-check** | 0 | No silent swallowing, no inaccurate fallbacks, no logic errors |
| **Elegance** | – | Skipped (`--no-elegance`) — the simplify pass already ran over this tree |

Notable verifications:

- **`invalid-states-unrepresentable`** — the `ScreenExtent` union (`full | range | tail | viewport`) makes illegal bound combinations (e.g. viewport + tailLines) unrepresentable, replacing the original four positional optionals with implicit precedence.
- **`persisted-schema-stays-tolerant`** — the breaking wire-contract reshape is correctly gated by the `PTY_HOST_CONTRACT_VERSION` bump to `4.0` (major), with a comment explaining the bidirectional skew, so an old/new daemon-client mismatch is rejected at the handshake rather than silently returning full scrollback.
- **Fail-fast** — CLI mutual-exclusivity fails loud; `tail.lines` is `.nonnegative()` at the wire boundary (no silent clamp to empty); `readScreenText`'s count was tightened to a required `number` so a `{kind:"tail"}` extent can't reach the wire without a count.
- **`viewport`** resolves against `entry.headless.rows` (the daemon's own grid), confirmed by tests spawning at `rows: 10` / `rows: 24`.

srid:
## Evidence

A terminal recording driving `kaval-tui snapshot` against an **isolated kaval daemon** (private socket, no production impact), built from this branch. A daemon terminal printed 60 lines into an 80×24 grid:

kaval-tui snapshot --viewport / --tail recording

_([mp4](https://github.com/juspay/kolu/releases/download/evidence-assets/kaval-snapshot-1610-1782677817.mp4) if the GIF doesn't autoplay.)_

**What the recording shows**

- **Default** `snapshot` → the full scrollback (`60` lines; thousands on a real compacted session) — unchanged, so `grep` still works.
- **`--viewport`** → just the visible screen (`23` lines = the daemon’s own 24-row grid), resolved server-side. The right read for “what’s on screen now” when driving an agent — no more `| tail` over a giant buffer.
- **`--tail 3`** (and the **`--lines`** synonym) → exactly the last N rendered lines.
- **Fail-fast** — `--viewport --tail` together crashes loud (`mutually exclusive`), no silent precedence.

<details><summary>Plain transcript (same run, incl. the negative-count rejection)</summary>

```console
# A terminal in the daemon printed 60 lines into an 80x24 grid.
# On a real long-running / compacted agent session this is THOUSANDS of lines.

$ kaval-tui snapshot 330e1077 | wc -l                 # default: the WHOLE scrollback
60

$ kaval-tui snapshot 330e1077 --viewport              # NEW: just the visible screen
line 55 — output from a long-running agent session
line 56 — output from a long-running agent session
line 57 — output from a long-running agent session
line 58 — output from a long-running agent session
line 59 — output from a long-running agent session
line 60 — output from a long-running agent session
  ...(only the last screenful — the daemon resolves it to its own 24 rows)

$ kaval-tui snapshot 330e1077 --viewport | wc -l       # bounded to the live grid
23

$ kaval-tui snapshot 330e1077 --tail 3                # NEW: just the last N rendered lines
line 59 — output from a long-running agent session
line 60 — output from a long-running agent session

$ kaval-tui snapshot 330e1077 --lines 3               # --lines is a synonym for --tail
line 59 — output from a long-running agent session
line 60 — output from a long-running agent session

$ kaval-tui snapshot 330e1077 --viewport --tail 3     # conflicting bounds crash LOUD (fail-fast)
kaval-tui: --viewport and --tail are mutually exclusive — pass at most one (omit all for the full scrollback).
  (exit 1)

$ kaval-tui snapshot 330e1077 --tail -1               # negative count rejected loud at the wire
kaval-tui: --tail/--lines takes a non-negative whole number of lines, got -1.
  (exit 1)
```
</details>

The same behaviour is pinned by unit tests over the real unix socket (`packages/kaval-tui/src/attach.test.ts`) and at the host level (`packages/kaval/src/ptyHost.test.ts`).

srid:

The **x86_64-linux** lane ran on **`kolu-ci-3`** (`dev-x86-64-linux-03`) — commit `df72f094b`, **passed** ✓

- **Lane wall** (pipeline): **3m2s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 2m46s | ✓ |
| `ci::home-manager` | 2m9s | ✓ |
| `ci::nix` | 59s | ✓ |
| `ci::unit` | 58s | ✓ |
| `ci::smoke` | 58s | ✓ |
| `ci::flake-check` | 56s | ✓ |
| `ci::atlas-sync` | 56s | ✓ |
| `ci::pnpm-hash-fresh` | 41s | ✓ |
| `ci::surface-example-build` | 38s | ✓ |
| `ci::biome` | 38s | ✓ |
| `ci::fmt` | 37s | ✓ |
| `_ci-setup` | 9s | ✓ |
| `ci::install` | 6s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/df72f09/timings.jsonl`; pool state is a live `flock` probe.</sub>

PR #1614, opened 2026-06-28
kolu consumes a local pulam — retire in-process awareness serving (R9.0)

srid:
## What & why

R8 ([#1594](https://github.com/juspay/kolu/pull/1594)) shipped the right **reader-join** (a terminal is `kolu.authored ⋈ terminalWorkspace.awareness`, joined at the reader) — *kept*. But R8 also made **kolu serve awareness *in-process*** (the registry-projection backing of `serveTerminalWorkspace`), which **re-coupled awareness into kolu** — the exact thing `pulam` (R4) exists to decouple.

**R9.0 fixes the decoupling: kolu now _consumes_ a pulam instead of serving awareness itself.** It spawns one ephemeral **local pulam** that dials kolu's own kaval, runs the one sensor set, and serves `terminalWorkspaceSurface`; kolu **mirrors** that pulam as its awareness backing — injected at the same `serveTerminalWorkspace` seam R9.3 will point at a *remote* pulam. The arrow points **→ pulam** with **no cross-reach**: pulam-web keeps its OWN pulam, so the `connectTerminalWorkspace` mud never arises.

This is **Goal A only** (decouple kolu). It deliberately does **not** collapse the two pulam *instances* into one — that's the optional, deferred [R9-converge](https://kolu.dev/atlas/remote-terminals-finale.html#r9-converge). The `working`-vs-`idle` desync that R9a fought is a *separate*, narrow concern; this PR removes the architectural defect (the inverted arrow), not that desync.

> Plan of record: `docs/atlas/src/content/atlas/remote-terminals-finale.mdx` §R9.0 ([#1611](https://github.com/juspay/kolu/pull/1611)).

## How it works

| piece | where |
| --- | --- |
| **Supervise** an ephemeral local pulam via the R2 `adoptOrEnsure` spine | `server/src/ptyHost/localPulamDriver.ts` · `connectPulam.ts` |
| **Mirror** it (supervise → mirror → reset+re-ensure loop) and fold each frame into the registry entry | `server/src/localPulamMirror.ts` · `terminalEndpoint/metadata.ts` (`applyMirroredAwareness`) |
| **Lift** the shared mirror-fold out of pulam-web's `buildReServe` | `terminal-workspace/src/reserve.ts` (`mirrorTerminalWorkspace`) — consumed by kolu *and* pulam-web |
| **Ephemeral** spawn driver — the dual of `survivableSpawnDriver` (dies with kolu's cgroup) | `surface-daemon-supervisor/src/driver.ts` (`ephemeralSpawnDriver`) |
| **Hard cutover** — delete kolu's in-process sensors | `terminalEndpoint/local.ts` (`startAwarenessSensors` → `startExitWatch`, lifecycle-only) |
| nix injects `KOLU_PULAM_BIN` | `default.nix` |

**Lifecycle:** the local pulam is *non-survivable* (dies with kolu) — it writes no gate, so `adoptOrEnsure` always *ensures* (spawns fresh), never adopts, exactly as the note specifies. The driver self-recycles its prior child, so a dropped link can't leave two pulams racing the socket.

**No regression slipped through:** the deleted sink also fed kolu's command-palette MRU (`trackRecentRepo`/`trackRecentAgent`); that feed is reconnected off the mirror frames, gated on *change* so it matches the old `!replayed` semantics and doesn't over-publish.

## Design philosophy

- **Fail-fast / no-fallbacks** — a hard cutover, no flag: kolu either mirrors its pulam or awareness is honestly absent (re-derived next cycle). The worst case on the local awareness path is a briefly-stale badge, never lost work.
- **Electricity boundary** — the ephemeral spawn mechanism lives in the daemon-supervisor spine (a sibling to `survivableSpawnDriver`), not folded into the server.
- **Reuse the source of truth** — the mirror fold is lifted into `@kolu/terminal-workspace` and shared by both kolu and pulam-web; no second copy, and **no new dependency edge** (the fold core needs only `@kolu/surface`).

## Scope notes

- **No drishti PR needed** — R9.0 changes no `@kolu/surface*` API (the lift only *adds* a consumer; `surface-nix-host` is untouched).
- **No user-visible change** — a pure decouple. Awareness (git/PR/agent badges) works identically; the Dock's live-output dot is client-side (`useTerminalActivity`, unchanged); `terminalWorkspace.activity` going live benefits remote awareness consumers (R9.3), not the local UI today. Per the changelog-skip rule, no changelog entry. The README architecture prose is updated to match.

## Tests

Covering integration test in pulam's suite (`packages/pulam/src/daemon.test.ts`) boots the full local chain — in-process kaval → a real pulam → the shared mirror fold — and asserts the mirrored awareness **equals** pulam's own on every load-bearing field (`git.branch`, `cwd`, `pr.kind`), plus removal-on-kill. `just check` green.


srid:
Closing (not merging) — a step-back review found this PR's whole premise was a misdiagnosis. The awareness capability is ALREADY a completely-decoupled library (@kolu/terminal-workspace: own package, arrow out, names no kolu-app package), and kolu ALREADY consumed it in-process on master (startAwarenessSensors → startAwareness). So 'decouple awareness from kolu in code' was solved before this PR. R9.0 conflated the PROCESS boundary with CODE decoupling: it kept the decoupled library but bolted on a local pulam process + dial + mirror + reseed-fold — ~430 net-new lines (connectPulam, localPulamDriver, localPulamMirror, the metadata.ts fold-clobber) that bought nothing for the decoupling and introduced the very fold-clobber blocker we chased across three review rounds (intrinsic to mirroring an ephemeral process that reseeds empty, not to awareness). Corrected R9.0 = kolu consumes the terminal-workspace library in-process (no local daemon); the pulam daemon is kept only for the REMOTE case (R9.3). Superseded by that plan.

PR #1615, opened 2026-06-28
fix(git): stop the event loop freezing on a hung git command

srid:
## What & why

A redeploy investigation traced a **25-minute browser-unresponsive wedge** (2026-06-28) to a single structural defect: the Code-view git watchers resolved each repo's git directory with a **synchronous `execSync('git rev-parse')`** (plus `fs.realpathSync`) run *inline on every watcher install*. On kolu's single-threaded Node server that parks the **entire** event loop in `waitpid` with **no timeout**, so one wedged `git rev-parse` (a stuck filesystem, a contended repo) freezes all HTTP/WebSocket serving for as long as the call hangs.

The journal signature matched exactly: total freeze, **~0 CPU** (blocked in a syscall, not spinning), **no crash**, all logging stopped at once, ended only by a manual redeploy. Confirmed adversarially: the git-dir `execSync` is the only synchronous child-process wait on the serving path, and an async hang can't produce a *total* freeze (it stalls only its own promise chain).

## The fix

**Resolve git-dir off the loop, bounded** (`kolu-git`, `kolu-io`)
- `resolveGitDir`/`resolveGitCommonDir` → async `execFile` with a 5s timeout + `fs.promises.realpath`. The wait runs on a libuv thread; the loop stays live; a wedged child is killed, not leaked.
- `createDirFilenameWatcher.watch()` now installs `fs.watch` after an async resolution, returning the unsubscribe **synchronously** (cancellation + generation guard), and fires one reconciliation tick so a change in the snapshot→attach window isn't lost. Public signature unchanged → zero ripple into `repo-change`/`resolve`/`surface`.
- Sibling audit: the two hot-path `realpathSync` calls in `resolveGitInfo` → `fs.promises.realpath`.

**Prevent recurrence of the class** (`.agency/code-police.md`)
- New `no-sync-blocking-on-the-serving-loop` rule: flags `execSync`/`spawnSync`/`*Sync` fs / `Atomics.wait` / unbounded sync loops on any path reachable while the server is serving — including a sync resolver passed into the async watcher primitive. `just check` can't see this class; a reviewer told to look for it can.

## Design philosophy

- **Fail fast, no fallbacks/knobs** — the git resolution returning `null` is the **pre-existing** "unresolvable / not a repo" contract, unchanged; the only new behavior is a bounded timeout that surfaces as that same `null` (a wedged git is treated as unresolvable, loudly bounded, never silently retried forever).
- **Volatility boundaries** — async git resolution stays in `kolu-git`; the async-install generic fs watcher stays in `kolu-io`. No new receptacle.
- **Reuse the source of truth** — reused the existing async-git pattern (`execFile` + `fs.promises`, already in `browse.ts`/`review.ts`).

## Residuals (honest)
`hasGitDir`'s single `accessSync` and `worktree.ts`'s one-shot `realpathSync`/`existsSync` are left synchronous — fast, known-local, one-shot reads off the high-churn path; they can only hang on a dead filesystem, which a restart wouldn't cure anyway. Covered by the new rule's documented carve-out.

> Note: an earlier revision added an out-of-loop `worker_threads` liveness watchdog; it was dropped as speculative defense-in-depth (a new failure surface that doesn't fix this bug). If a liveness-restart backstop is wanted later, systemd's native `WatchdogSec` is the right layer.

## Tests
- `kolu-io` (6): `watch()` returns while `resolveDir` is pending; reconciliation tick fires once installed; throwing/rejecting resolvers caught; unsubscribe-before-settle cancels; `_reset` discards a stale pending resolution.
- `kolu-git` (98): adapted to await install settle.


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `d313096ebac0`

### Applied (2)
- `lowy-1` Stall predicate duplicated: isStalled has no production caller — commit `134c023d3`
- `hickey-1` Stall predicate computed in two places (isStalled vs inline worker copy) — commit `134c023d3`

### Agreed — no change (6)
- `lowy-2` Asymmetric settle barrier across five identical watchers (packages/integrations/git/src/reflog-watcher.ts:48 and index-watcher.ts:29 (no _whenSettled re-export) vs head/config/cwd-git-watcher)
- `lowy-3` Cross-thread wire format is an implicit, triplicated contract (packages/server/src/eventLoopWatchdog.ts:106-109 (writer) and :84-91 (reader, WATCHDOG_WORKER_SOURCE); packages/server/src/eventLoopWatchdog.test.ts:68-75 (fixture writer))
- `lowy-4` Watchdog is domain-agnostic electricity kept as a server leaf (packages/server/src/eventLoopWatchdog.ts:1-130)
- `hickey-2` "Does cwd have a git dir?" answered by a sync stat and an async subprocess (packages/integrations/git/src/resolve.ts:57-64 (hasGitDir) and packages/integrations/git/src/git-dir.ts:45-63 (resolveGitDir))
- `hickey-3` _whenSettled loops for a case the code states can't happen (packages/integrations/io/src/refcounted-dir-watcher.ts:196-201)
- `hickey-4` Worker launch contract duplicated between production and test fixture (packages/server/src/eventLoopWatchdog.ts:944-958 and packages/server/src/eventLoopWatchdog.test.ts:780-793)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `d313096ebac0`

### Round 1

**codex** — approved: `false`

The branch addresses the original event-loop wedge directionally, and the changed packages typecheck cleanly via their package-local `tsc --noEmit`. I found three correctness issues that should be fixed before approval: one lost-update window introduced by async git-dir watcher attachment, and two watchdog edge cases around false aborts and cleanup on the new SIGABRT path.

Findings:
- `F1` · major · open — The async `resolveDir`/`fs.watch` attachment can miss a repo/file change that lands after a stream has yielded its initial snapshot but before the underlying git-dir watcher is actually attached. The deliberate "snapshot first" reasoning does not cover this case: for the Code-tab pulse streams, `pollOnEvent` reads and yields the snapshot before installing the callback, and this change adds another async `git rev-parse`/`realpath` interval before `fs.watch` exists. A `git add`, commit, checkout, or config change in that interval is neither in the already-sent snapshot nor observed as a later pulse. (packages/integrations/io/src/refcounted-dir-watcher.ts:161)
- `F2` · major · open — The watchdog compares wall-clock `Date.now()` heartbeats across threads. A laptop sleep/resume or a forward wall-clock adjustment can make a healthy main loop look stale by more than 60s, especially if the worker's interval runs before the main heartbeat after resume, causing an unnecessary SIGABRT and restart. (packages/server/src/eventLoopWatchdog.ts:77)
- `F3` · major · open — The new SIGABRT recovery path bypasses the existing `process.on('exit', shutdownCleanup)` cleanup in `packages/server/src/index.ts`, so `koluRoot` scratch data can be left behind on the intended watchdog recovery path. Those scratch dirs can contain pasted images, dropped files, and secrets, and the old shutdown comment's "every exit path" assumption is no longer true for watchdog aborts. (packages/server/src/eventLoopWatchdog.ts:91)


**claude** — Agreed with all three findings and fixed each: a post-install reconciliation tick on the async git-dir watchers, a monotonic clock for the watchdog heartbeat, and worker-side scratch cleanup before the SIGABRT.

- `F1` **fixed** — Real lost-update window, and the codebase already proves the fix shape: `watchWorkingTree` fires a post-install reconciliation tick (`working-tree-watcher.ts:199-207`) precisely because its async install attaches after the consumer's snapshot, but the sibling git-dir axes (HEAD/reflog/index/config) — all of which now route through the async `createDirFilenameWatcher` — had no equivalent. `pollOnEvent` reads+yields the snapshot, *then* `repoEventStream` calls `install` (`packages/surface/src/server.ts:452-454,493`), and the leaf's `execFile git rev-parse` + `realpath` push the `fs.watch` attach to a later tick, so a commit/checkout/stage/config edit in that window was in neither the snapshot nor any future event. Fixed at the leaf so all five watchers (head/config/reflog/cwd/index) get it uniformly: after `entry.subscribe(onChange)` in the settle path I fire one `onChange()` reconcile tick, wrapped in the same try/catch as the event dispatch so a throwing listener can't reject the settle promise (that would surface as an `unhandledRejection` → the server's handler exits). See `packages/integrations/io/src/refcounted-dir-watcher.ts:183-200`. Added two tests (`refcounted-dir-watcher.test.ts`): reconcile fires exactly once post-install, and a throwing reconcile listener is caught not propagated. (Note: `subscribeGitInfo` already does a best-effort `void resolve()` at `resolve.ts:272`, but it races the async attach — the leaf reconcile closes that gap too.)

- `F2` **fixed** — Valid and realistic on a laptop host (kolu runs as a user service): the heartbeat compared cross-thread `Date.now()`, so a sleep/resume or forward NTP step longer than the 60s threshold could make a healthy loop look stale and trip a false `SIGABRT` on wake (the worker's 5s check racing the main thread's 1s beat after resume). Switched both sides to `process.hrtime.bigint()` (`eventLoopWatchdog.ts:118` stamp, worker reads it at the interpolated worker source). hrtime reads the same system-wide monotonic source in every worker thread (no per-thread origin, unlike `performance.now()`), never steps with the wall clock, and on Linux/macOS does not advance across suspend — so the watchdog only trips on a genuine in-loop wedge. `isStalled` stays one source of truth (made unit-agnostic; the worker now compares in ns via `thresholdNs = thresholdMs * 1e6`). No new knob — fail-fast intent preserved. (Did not add codex's separate "scheduler-gap awareness" branch: with a suspend-pausing monotonic clock the suspend interval simply isn't counted, so that branch is redundant — monotonic alone removes the false-abort class.)

- `F3` **fixed** — Valid: `SIGABRT`'s default disposition does NOT run `process.on('exit', shutdownCleanup)` (`index.ts:194`), and since the wedge fix this abort is the *intended* recovery path, so every recovery orphaned `koluRoot` scratch (pasted images, dropped files, secrets). Fixed by the targeted half of codex's suggestion (a): the worker `rmSync`s `cleanupPaths` synchronously right before the abort (the worker's own loop and the FS are fine even with the main loop wedged) — `eventLoopWatchdog.ts` worker source + `startEventLoopWatchdog({ cleanupPaths })`, wired in `index.ts:226` as `[koluRoot]`. Best-effort: a failed unlink still reaches the abort (restart is the critical action) and is surfaced on stderr, not swallowed. Updated the now-stale "every exit path" comment at `index.ts:188-194`. Child-process test extended to assert the scratch dir is gone after the abort. I deliberately did NOT take codex's suggestion (b) startup-sweep of stale `kolu-*` roots: `koluRoot` is keyed by a per-startup UUID with no liveness/pid marker, so a blanket sweep would delete a *concurrent* instance's live root (dev server, second worktree) — unsafe without a liveness handshake this PR doesn't have. The targeted worker cleanup fully closes the finding as scoped (the SIGABRT path); the pre-existing SIGKILL/power-loss residual remains covered by the XDG logout-wipe backstop, unchanged by this PR.

commit: `<see structured return>`


### Round 2

**codex** — approved: `true`

All three previously raised findings are resolved in the current tree. I verified the relevant code paths and ran no-emit typechecks for `kolu-io`, `kolu-git`, and `kolu-server`; all passed. I did not re-scan for unrelated pre-existing issues per the convergence rules.

Findings:
- `F1` · major · resolved — The async watcher-install lost-update window is now closed by subscribing the listener once the watcher is installed and immediately firing one caught reconciliation `onChange()` tick so stream consumers re-read state after the snapshot-to-attach gap. (packages/integrations/io/src/refcounted-dir-watcher.ts:185)
- `F2` · major · resolved — The wall-clock false-abort class is resolved: both heartbeat writer and worker reader now use `process.hrtime.bigint()` and compare in monotonic nanoseconds, so NTP jumps and normal sleep/resume no longer make a healthy loop look stale. (packages/server/src/eventLoopWatchdog.ts:103)
- `F3` · major · resolved — The SIGABRT cleanup gap is resolved: the watchdog worker synchronously removes configured cleanup paths before aborting, and `index.ts` passes this process's `koluRoot`, covering the intended recovery path that bypasses `process.on('exit')`. (packages/server/src/eventLoopWatchdog.ts:123)

_codex on the rebuttal:_ F1: accepted. The fix is at the right layer: all async `createDirFilenameWatcher` consumers get a post-install reconciliation tick, and the listener throw path is fenced so the settle promise cannot become an unhandled rejection.

F2: accepted. My original scheduler-gap wording was one possible implementation. The actual bug was the wall-clock comparison; using the same monotonic `process.hrtime.bigint()` source on both threads removes the Date.now/NTP/sleep false-abort class, so I do not require the separate scheduler-gap branch.

F3: accepted. Worker-side cleanup before SIGABRT closes the code issue I raised. I also accept the decision not to add a startup sweep here: without a liveness marker, sweeping `kolu-*` roots could delete another live instance's root, so that would be a new unsafe behavior rather than a necessary part of this fix.

srid:

Ran rules + fact-check passes (elegance skipped — `/simplify` already ran this gauntlet), scoped to the diff vs the merge-base.

| Pass | Findings | Actioned |
|------|----------|----------|
| Rules | 1 | `no-silent-error-swallowing` — bare `catch { return null }` in `git-dir.ts` (`revParseDir`) and `reflog-watcher.ts` (`resolveDir`) lacked catch-site comments → added inline comments naming the expected cases (not-a-repo / git-not-installed / 5s timeout / realpath failure; and fresh-repo-no-`logs/`). |
| Fact-check | 1 | `eventLoopWatchdog.test.ts` — the SIGABRT assertion's `\|\| result.code === null` disjunct was logically dead (true for *any* signal kill), weakening the check → tightened to `expect(result.signal).toBe("SIGABRT")`. |
| Elegance | — | Skipped (`--no-elegance`; simplify covered it). |

Both fixes committed (`fix(police): …`). All other rules checked clean.

srid:
## Evidence

This fix has **no on-screen surface** — it's a server-internal stability change (the event loop no longer freezes on a hung `git rev-parse`; a wedged loop self-restarts). The bug it fixes manifested *as* an unresponsive browser, but reproducing it requires a hung syscall / a frozen event loop, which the e2e Cucumber+Playwright harness (it drives UI surfaces) cannot induce. So the "demonstrate the fixed behavior" proof is **behavioral**, and it runs in CI:

**1. A blocking resolver no longer blocks the watcher install (the wedge's mechanism).** `watch()` returns synchronously while `resolveDir` is still pending — the property whose absence caused the freeze:

```
✓ createDirFilenameWatcher async install > watch() returns synchronously while resolveDir is still pending  2ms
✓ createDirFilenameWatcher async install > fires a reconciliation tick once the watcher is installed
✓ createDirFilenameWatcher async install > a throwing reconcile listener is caught, not propagated
✓ createDirFilenameWatcher async install > unsubscribing before the resolution settles cancels the install
✓ createDirFilenameWatcher async install > a resolveDir that rejects is caught and logged, never thrown
✓ createDirFilenameWatcher async install > _reset() discards a resolution still pending from before the reset
```

**2. A wedged event loop really is aborted → restarted (the watchdog backstop).** This child-process test literally wedges a process's main loop and asserts the worker `SIGABRT`s it (which is what turns a minutes-long hang into a supervisor restart):

```
✓ isStalled > treats the 0 sentinel as not-yet-started, never a stall  1ms
✓ isStalled > is false at or under the threshold, true strictly over it
✓ startEventLoopWatchdog > starts and stops cleanly without aborting a healthy process  153ms
✓ the watchdog aborts a wedged process > SIGABRTs a child whose event loop is blocked past the threshold  1394ms
```

The child-process test is the closest thing to a live demonstration: a real process whose loop is blocked past the threshold is killed by the worker (signal `SIGABRT`), the diagnostic line lands on stderr first, and `Restart=on-failure` / launchd `KeepAlive.Crashed` would then bring up a fresh process — a ~minute self-recovery instead of the 25-minute hand-fixed wedge.

srid:

The **x86_64-linux** lane ran on **`kolu-ci-8`** (`dev-x86-64-linux-08`) — commit `3e7c9caa3`, **passed** ✓

- **Lane wall** (pipeline): **3m8s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 2m35s | ✓ |
| `ci::home-manager` | 2m14s | ✓ |
| `ci::unit` | 58s | ✓ |
| `ci::nix` | 58s | ✓ |
| `ci::smoke` | 54s | ✓ |
| `ci::flake-check` | 54s | ✓ |
| `ci::biome` | 32s | ✓ |
| `ci::fmt` | 31s | ✓ |
| `_ci-setup` | 19s | ✓ |
| `ci::install` | 12s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/3e7c9ca/timings.jsonl`; pool state is a live `flock` probe.</sub>

srid:

The **x86_64-linux** lane ran on **`kolu-ci-8`** (`dev-x86-64-linux-08`) — commit `0a39f6d11`, **passed** ✓

- **Lane wall** (pipeline): **3m27s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 3m9s | ✓ |
| `ci::home-manager` | 2m33s | ✓ |
| `ci::unit` | 54s | ✓ |
| `ci::nix` | 52s | ✓ |
| `ci::smoke` | 50s | ✓ |
| `ci::pnpm-hash-fresh` | 39s | ✓ |
| `ci::surface-example-build` | 34s | ✓ |
| `ci::fmt` | 34s | ✓ |
| `ci::biome` | 34s | ✓ |
| `ci::install` | 9s | ✓ |
| `_ci-setup` | 8s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/0a39f6d/timings.jsonl`; pool state is a live `flock` probe.</sub>

PR #1624, opened 2026-06-29
Make /be §5 pin BOTH CI platforms, with the linux lane on a pu box

srid:
**`/be` §5's odu host-pinning recipe only ever named the macOS (`aarch64-darwin`) lane.** The `x86_64-linux` lane was left unspecified — so a run that pins darwin alone silently drops Linux from the odu fanout and reports a *false single-platform green*, and the obvious stop-gap (`x86_64-linux=localhost`) runs the heavy build/test lane right beside production kolu, the exact OOM-kill hazard the section opens with.

This adds a sibling bullet to the darwin one: **pin both platforms in the same `mcp__odu__run hosts=[…]` call, run the Linux lane on an ephemeral pu box (load `/pu`) never `localhost`, and confirm the settled run carried two platforms before reporting green.**

### Provenance

Mined by `/self-improve` from session `bc7dbabe-41c3-4896-9e89-84e1edfaf069`. The human had to step in **twice on the same ship gate**, both tracing to this one gap:

| Human intervention | Root cause |
| --- | --- |
| *"Why are you not invoking odu on TWO platforms?"* | §5 pinned only darwin; the unpinned Linux platform silently dropped from the fanout |
| *"Doesn't the CI instructions tell you to run pu?"* | §5 gave no venue for the Linux lane, so it ran on `localhost` beside production instead of a pu box |

### Evidence ledger

- **Edit** — `agents/.apm/skills/be/SKILL.md` §5 step 1: new `x86_64-linux CI lane` bullet (both-platform pin + pu venue + two-platform confirmation before green).
- **Regenerated** — `just ai::apm` propagated the source edit to `.claude/skills/be/SKILL.md` and `.agents/skills/be/SKILL.md`; `apm.lock.yaml` diff is only the timestamp + the two `be/SKILL.md` hashes (no dependency bumps).
- **Gates green** — `just fmt` (no reformat), `just check` (tsc + biome clean).

> _Observations not worth an edit this run:_ a single pre-edit interrupt (the agent began applying fixes before establishing the existing e2e was green); the human's intended order was *baseline-green first, then change*. One reversible occurrence — left as a note, not a rule.

_Generated by [`/self-improve`](https://github.com/srid/agency) on Claude Code (model `claude-opus-4-8`)._

### DIFF
--- a/.agents/skills/be/SKILL.md
+++ b/.agents/skills/be/SKILL.md
@@ -146,6 +146,19 @@ green before capturing.
        falls back to the repo's on-disk `zest`, and you burn a full CI run on the
        dead host. If you must set `$ODU_HOSTS`, write a real hosts *file* and point
        at it; otherwise reach for `--host`.
+   - **x86_64-linux CI lane — on an ephemeral pu box, *never* `localhost`.** Pin it
+     in the **same** `mcp__odu__run hosts=[…]` call as the darwin lane
+     (`hosts=["x86_64-linux=<user>@<pu-box>", "aarch64-darwin=<user>@<host>"]`) so
+     **both** platforms actually run. A platform you don't explicitly pin **silently
+     drops from the fanout** (the `/ci` skill's `hosts.json` contract), so a
+     darwin-only pin is a *false single-platform green*, not full CI — the same
+     "never silently drop the platform" rule the darwin bullet states, applied to
+     the lane that's easy to forget because the host JSON only ever names darwin.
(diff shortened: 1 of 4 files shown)


srid:
Superseded by #1625. Its lesson — a platform you don't explicitly pin silently drops from the fanout, so a single-platform green is a false green — folds into `.agency/do.md`'s two-platform rule there, instead of adding another CI bullet to `/be` §5. #1625 also deletes §5's stale duplicate of the CI host lore (which had drifted to contradict do.md on `sincereintent`).

Issue #1492, opened 2026-06-21
Waking a sleeping terminal doesn't resume an agent launched via `nix run …#agent`


srid:
### What

When a terminal running an AI agent is put to **sleep** and later **woken**, kolu resumes the agent's conversation by replaying its launch command in resume form (e.g. `claude -c`, `codex resume --last`, `opencode --continue`).

This relies on kolu having **observed** the agent's launch command via the shell command sensor (OSC 633). When the agent is launched through a `nix run` wrapper — e.g.

```sh
nix run github:juspay/AI#opencode
```

the observed command's head token is `nix`, not the agent, so kolu doesn't recognize it as an agent launch and **doesn't capture a resumable command**. As a result, **waking such a terminal brings back a bare shell** instead of resuming the conversation.

### Repro
1. Launch an agent via `nix run github:juspay/AI#opencode` in a kolu terminal.
2. Sleep the terminal (☾).
3. Wake it.
4. **Expected:** the agent resumes the prior conversation. **Actual:** a bare shell.

### Status
Not currently supported, by design. Earlier special-casing for `nix run` wrappers was removed in favor of a simpler resume path that replays only the directly-observed agent command. Running the agent **directly** (agent binary on `PATH`) resumes correctly on wake.

### Possible directions
- Detect `nix run <ref>#<agent>` wrappers in the command sensor and resume through the wrapper (`nix run … -- --continue`).
- Or capture the resumable command from the file-watcher-detected agent *kind* when the command tap doesn't fire.

_Surfaced while simplifying the Sleep/Wake resume path (PR #1487)._

PR #1626, opened 2026-06-30
Awareness: the producer observes, kolu remembers (S1+S2 → R9.0)

srid:
## Awareness — the producer observes, kolu remembers (S1 + S2 → R9.0)

Implements the merged plan [`awareness-derive-store.mdx`](docs/atlas/src/content/atlas/awareness-derive-store.mdx) (#1621) as **one coordinated change** — S1 (the memoryless producer) and S2 (kolu's fold) are coupled, so they ship together — then **hardened across two rounds of a 4-reviewer architecture-first-principles + perfection-review gauntlet** (`debates/pr1626-review`, `debates/pr1626-rereview`). The reviewers credited the fold spine **sound**; every residual was in the shell around it — the first round's four must-fixes (below), then two relocations codex caught on re-review (F1/F2).

### The cut

The per-terminal awareness engine used to **derive** awareness by **mutating a host record it also read back as memory**, through an `AwarenessSink` two homes (kolu + `pulam`) each had to wire. This splits it:

- **S1 — memoryless producer.** `startAwareness` → `startSensors`: it *emits* per-field `TerminalEvent`s and takes no seed, no sink, no record. Deleted: the `AwarenessSink` interface, **both** `makeAwarenessSink` impls (kolu's + pulam's `hooks.ts`), `AwarenessRecord`, the apply-and-publish read-back.
- **S2 — kolu's fold.** A pure `fold(cur, o, ctx) → TerminalState` (`{snapshot: TerminalSnapshot, memory: AgentMemory}`). The producer's emit type **cannot spell** a memory fact, so the old "two narrowed mutators" write-fence is now the **type**.
- **R9.0 — local cutover.** kolu runs the producer in-process and folds the stream, publishing the snapshot half to `snapshots` and the memory half to `kolu.authored`.

### Review-gauntlet hardening (the four must-fixes)

| # | Finding | Fix |
| --- | --- | --- |
| **#2 / #7** | The restore target was a bare optional `resumeAgent` identity read alongside `lastAgentCommand`, so `(command set, identity absent)` meant **both** "quit, restore nothing" *and* "no id captured, resume most-recent" — and the derivation lived in the emit shell, not the fold. | The fold now produces a **discriminated `RestoreTarget`** — `none \| { exact, command, agent } \| { legacyMostRecent, command }` (`restoreTargetOf`, consumed by `resumeFormFor`). Absence can no longer be read as most-recent: **quit-to-shell → `none` → bare shell by construction** (#1492) for new records; a migrated pre-1.29 record keeps most-recent as a *named* `legacyMostRecent`. The client restore card + dormant tile require an `exact`/`legacyMostRecent` target. |
| **#5** | The authored *collection* re-published on **every** observation tick (~150 ms) — only the disk arm was fenced. | The emit loop computes deltas once and gates each arm by **its own** delta: snapshot publish on a snapshot change, **authored publish on an authored-fact change**, disk autosave on a restore-relevant change. |
| **#1** | Recency leaned on a 1.5 s wall-clock window; a slow agent resolution crossing it could restamp saved recency on restore. | The frame phase is decided **by value** — compare the re-resolved agent identity to the saved restore target. The timer is deleted. |
| **#8** | The schema module's docstring still taught the deleted `AwarenessValue` / `AwarenessSink` world. | Rewritten to `TerminalSnapshot` / `AgentMemory` / `TerminalState` / the emit type. |

`AgentIdentity` + `RestoreTarget` moved down to `anyagent/schemas` (the resume-vocabulary layer), re-exported through `terminal-workspace` — the fold *produces* the target, `resumeFormFor` *consumes* it, one owned type. The serial lens → codex → simplify → code-police pass then collapsed the vestigial `AgentSessionRef` into `AgentIdentity`, shared one `resumableCommand` projection across the client surfaces, value-gated the emit fences, and hardened the migration against corrupt input.

### Second re-review — two relocations codex caught (the panel earned its keep again)

A second 4-reviewer pass on the hardened HEAD credited the four must-fixes **closed by construction** — but codex (1 of 4) found two real defects relocated *inside* the new shapes, both verified against the diff:

| # | Finding | Fix |
| --- | --- | --- |
| **F1** | `restoreTargetOf` paired `lastAgentCommand` with `snapshot.agent` **without checking they agree on agent kind**, so a stale-command/new-agent race could build `exact{ command:"opencode …", agent:{kind:"claude-code"} }` — which `resumeAgentCommand` silently downgrades to opencode's most-recent: the wrong-agent resume #2 makes unspellable, re-expressed in the `exact` arm. | One constructor, **`exactRestoreTarget(command, agent)`** (anyagent), builds `exact` *only when* `agentKindFromCommand(command) === agent.kind`, else `null`. **Both** production sites go through it — the fold (→ `none` on mismatch) and the migration (→ `legacyMostRecent` on mismatch) — so `resumeFormFor` always takes the same-agent path (exact-by-id, or refuse), never the downgrade. |
| **F2** | The funnel `guardedEmit` swallowed a thrown emit **after** the producer advanced its dedup baseline, so "emitted" ≠ "accepted": the fold never got the value while the baseline advanced → a later equal value deduped → silent permanent divergence. The guard sat at the wrong seam. | Error handling moved to where the fallible work is — the host's **publish boundary**. `commitSnapshot`/`updateMemory`/`emitTerminalsDirty` (kolu) and pulam's snapshots `upsert` each **fold-accept first, then guard their own publish** (logged, self-heals on the next change). With the publishes infallible, `emit` is infallible — and the wrong-seam funnel guard is **removed**. |

### Type-naming cleanup

A follow-up **mechanical, compiler-checked rename** gives the awareness types **one root per shape** — `TerminalSnapshot` (the whole value a host emits) · `TerminalEvent` (a per-field delta) · `TerminalState` (the `{ snapshot, memory }` composite kolu folds) · `Known<T>` (resolution status). The `Terminal` prefix scopes the wire-facing family, and a field repeats its type's root (`entry.observed` → `entry.snapshot`). The served collection key bumps `awareness` → `snapshots` (`TERMINAL_WORKSPACE_CONTRACT_VERSION` `2.0` → `3.0` — breaking, the wire path changed). "Observation"/"awareness" survive only as feature-word prose; the deleted `AwarenessSink`/`AwarenessValue` keep their old names where the text describes what was removed. The schema-module docstring, the [design note](docs/atlas/src/content/atlas/awareness-derive-store.mdx) (types, prose, and the foldflow/homes diagrams), and the unit-test fixtures all move with the code.

- **Fail-fast / no knobs** — the write-fence is the emit type; the restore decision is a discriminated value, not a runtime fallback. No override path.
- **Volatility boundary** — producer/fold/`RestoreTarget` stay in the `@kolu/*` packages; kolu and pulam are consumers, the dependency arrow points out.
- **Reuse the source of truth** — `foldSnapshot` is shared by kolu's fold and pulam's accumulator; `restoreTargetOf` / `exactRestoreTarget` are the *one* place a restore target is decided; the migration composes the existing backfill ladder.

### Deferred refinements (documented honestly, with pins)

- **Active drain-at-sleep (#3).** `beginSleep` freezes the last fold-written target and does **not** drain a final settle, so a sub-second launch-then-sleep freezes a stale `none` and wakes to a **false bare shell**. The docs no longer claim the freeze "holds by construction"; under model B the consequence is fail-*safe* (a bare shell, never the wrong conversation). The active drain is an async-`beginSleep` follow-up.
- **`git`/`pr` reference equality (#6).** `restoreRelevantEqual` compares `git`/`pr` by reference (correct today — upstream `gitInfoEqual`/`prResultEqual` dedup before emit). A `fold.test.ts` pin asserts a structurally-equal-but-new-ref `git`/`pr` still compares equal at the fence, so a future fold copy can't silently break it.
- **R9.3 scope** — the framed `TerminalFrame` *wire* stream and cross-host keying stay R9.3; R9.0 keys by `TerminalId` and folds in-process.

### Migration

`SCHEMA_VERSION` 1.28.0 → 1.29.0: `backfillSnapshotCutover` backfills `pr: { kind: "absent" }` and **synthesizes the discriminated `restoreTarget`** from what a pre-cutover record remembered — a valid `agentSession` whose kind matches the command → `exact`, a command alone (or a kind mismatch) → `legacyMostRecent`, neither → absent (`none`); corrupt `agentSession` value-types fall through safely rather than producing a parse-rejecting record.

### Tests & CI

The pure `fold` / `restoreTargetOf` / `exactRestoreTarget` / `resumeFormFor` are unit-tested (the discriminated switch, the kind-mismatch refusal, the ref-stability pin, the publish-boundary infallibility). The sleep/wake, adopt, reconcile, metadata, session-transfer, and restore-card tests track the reshaped persistence; the e2e session-restore + sleeping-terminals fixtures were fixed to stamp the now-required `pr`. All **unit tests** green across the touched packages, `just check` (tsc + biome) clean, and the **Atlas sync gate idempotent** — re-confirmed after the type-naming cleanup. **Full CI green on both `x86_64-linux` and `aarch64-darwin`** (every recipe × both platforms) on the naming commit, including `ci::e2e` on each.


srid:

✅ **Consensus** after 2 round(s) · lowy + hickey · base `d60c7643d4f1`

### Applied (5)
- `lowy-2` One conversation-identity concept encapsulated twice (AgentSessionRef vs AgentIdentity); AgentSessionRefSchema is now a vestigial schema — commit `be2e012ca`
- `lowy-3` Autosave/authored equality fences reach past restoreTargetOf to its raw inputs (expose the wires) — commit `4ab1aa50a`
- `hickey-1` "Does this restore target resume, and with what command" is spelled four divergent ways — commit `9ae592317`
- `hickey-2` authoredFactsEqual is a textual subset of restoreRelevantEqual — agent-identity + memory comparisons duplicated — (uncommitted)
- `hickey-3` seedMemory() claims to be the single memory-default home but production duplicates the default inline — commit `422737b72`

### Agreed — no change (7)
- `lowy-1` Electricity test: receptacle map and where wires get exposed (packages/terminal-workspace/src/sensors.ts:776 / fold.ts:82 / packages/integrations/anyagent/src/agent-cli.ts:338)
- `lowy-4` Recency identity-change check evaluated in two homes across the fold/shell seam (packages/server/src/terminalEndpoint/local.ts:765)
- `lowy-5` AgentMemory has a clean nested home but is stored flat, re-crossed by hand at seed and write seams (packages/server/src/terminalEndpoint/local.ts:741)
- `hickey-4` Recency-bump decision is co-owned by the shell's recencyBaseline and the fold's own identity check, under an overloaded ctx.live (packages/server/src/terminalEndpoint/local.ts:765-787)
- `hickey-5` restoreTarget.command duplicates the lastAgentCommand memory field on disk (packages/terminal-workspace/src/fold.ts:83-89)
- `hickey-6` "awareness" and "Observation" are now two names for the same value (packages/server/src/terminal-registry.ts:33)
- `hickey-7` restoreRelevantEqual's git/pr equality depends on the fold preserving object references (packages/server/src/terminalEndpoint/local.ts:307-308)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 round(s) · codex reviewed at `xhigh` reasoning effort · base `d60c7643d4f1`

### Round 1

**codex** — approved: `false`

The branch is directionally coherent, but I found two real behavior regressions and one cleanup nit. The main issue is that the UI-facing "resumable" projection no longer matches the actual restore renderer, so users can be shown/count terminals as resumable when wake/restore will send nothing. Separately, the new fold no-op fast path drops MRU updates for repeated identical agent commands.

Findings:
- `F1` · major · open — `resumableCommand` returns `target.command` for every `exact` and `legacyMostRecent` restore target, but the actual wake/restore path goes through `resumeFormFor` and `resumeAgentCommand`, which can return `null` for malformed exact session IDs and for non-resumable legacy commands. That makes `restoreModel` and the dormant/empty-state UI count or display a command as resumable even though restore will not send a resume invocation. (packages/integrations/anyagent/src/schemas.ts:86)
- `F2` · minor · open — The awareness `emit` path returns early when `fold` produces the same object, before running `trackRecentAgent` at line 792. For a live, non-replayed `commandRun` with the same normalized command as before, `fold` deliberately dedups at `packages/terminal-workspace/src/fold.ts:137`, so repeated use of the same agent command no longer refreshes the recent-agent MRU. The old command sensor updated MRU even when `lastAgentCommand` itself did not change. (packages/server/src/terminalEndpoint/local.ts:775)
- `F3` · nit · open — Several touched comments still describe the old model: this one says the sleeping record's frozen `pr` wins, `packages/client/src/terminal/DormantTileBody.tsx:44` says `pr` froze off live overlay, and `packages/pulam-web/src/client/HostGroup.tsx:5` / `:279` still mention awareness recency. The implementation moved to persisted `Observation` plus authored restore memory, so these comments now mislead the next reader about the invariant being preserved. (packages/server/src/terminals.ts:82)

**claude** — Conceded all three: gated the resumable display/count on the real wake test, moved the MRU side-effects ahead of the fold no-op early-return, and refreshed three stale comments; full unit suite + `just check` green.

- `F1` **fixed** — Real regression, conceded. `resumableCommand` (`packages/integrations/anyagent/src/schemas.ts:86`) tested `kind` independently of the wake path, so a `legacyMostRecent` record for a detection-only agent (`aider`/`goose`/…, in `STABLE_FLAGS` but not `AGENT_RESUME`) or an `exact` target whose id fails its shell-safe shape gate got counted/shown as resumable while `resumeFormFor` (`agent-cli.ts:338`) returns `null` and wake lands on a bare shell. Fixed by GATING on the single authority: `if (target === undefined || resumeFormFor(target) === null) return null;` then the existing exhaustive switch returns the raw display command. The two projections can no longer drift (one now delegates to the other). schemas.ts imports `resumeFormFor` from the sibling `./agent-cli.ts` — both browser-safe (agent-cli pulls only the pure `@kolu/shell-quote` + `string-argv`; the client already bundles it via `anyagent/cli`), and the `agent-cli → schemas` edge is `import type` only, so no runtime cycle and no node-only deps dragged into the "browser-safe schemas" layer. Added a `resumableCommand` describe block (6 cases — happy `exact`/`legacy`, plus the three null cases that now match `resumeFormFor`) in `agent-cli.test.ts`, and a broken-id exclusion case in `restoreModel.test.ts`. Also fixed an unrealistic fixture there: `exactTarget` used `sessionId: "ses-A"` (not a claude UUID, so it correctly failed the new gate) → now a valid UUID, since claude session ids ARE UUIDs and the test is about the sleeping/sub/quit filtering, not the id gate.

- `F2` **fixed** — Real regression, conceded. Verified against the base: the OLD command sensor (`sensors.ts:419` at d60c7643d) fired `sink.trackRecentAgent?.(normalized)` for every `!replayed` mark, OUTSIDE the `record.meta.lastAgentCommand !== normalized` gate — i.e. a re-launch of the same agent command still refreshed the recent-agents MRU `lastSeen`. The new emit path (`packages/server/src/terminalEndpoint/local.ts`) returned early on `current === before` (the fold dedups an unchanged `lastAgentCommand` to `cur`, `fold.ts:138`) BEFORE reaching `trackRecentAgent`, dropping that bump. Fixed by moving both cross-terminal MRU side-effects (`trackRecentRepo`/`trackRecentAgent`) ABOVE the early return — they key off the OBSERVATION, not the fold delta. Behavior-neutral for git (a git observation always allocates a new `observed`, so it never hit the early return) and for new/non-replayed commands; only restores the same-command refresh. No unit seam exists for the private `emit` closure (it's wired to pty-host taps — `grep` found zero tests touching it), so this rests on the fold-dedup contract test (`fold.test.ts`) plus the base-vs-branch behavior diff above rather than a new emit-level test.

- `F3` **fixed** — Stale comments, conceded; all three sites describe the OLD frozen-`pr`/awareness-recency model that `surface.ts` itself says is GONE (`surface.ts:221,291,350`: "the old frozen-`pr`-on-the-sleeping-arm special case is gone … `pr` is restore-relevant now"). (1) `terminals.ts:82` — dropped "a sleeping record's frozen `pr` wins"; the spread order is justified now by "authored names no observed field, so it never clobbers the observation" + the sleeping arm keeping the restore-relevant projection (`pr` included) and stripping the live half, matching `composeTerminalMetadata`. (2) `DormantTileBody.tsx:44` — `cwd`, `git.branch`, AND `pr` ALL ride the persisted `PersistedObservation` base; removed "the snapshot the sleeping arm froze off the live overlay". (3) `HostGroup.tsx:5` — the `awareness` collection is the memoryless `Observation` (agent state + git, NO recency — `lastActivityAt` is kolu's remembered memory, confirmed absent from `ObservationSchema`); and the row-order comment (`:279`) said "needs-you first, then most-recent, then id" but `compareFleetEntries` (`fleet.ts:147-157`) sorts "needs-you first, then id" (urgency alone — pulam has no recency), so corrected to match.

commit: `8f704b1b082a43805c2fbe3ad173ac569e04a938`


### Round 2

**codex** — approved: `false`

F1 and F2 are resolved in the current tree. F3 is only partially fixed: the inline comment was corrected, but the surrounding snapshotSession docblock still states the old pr-from-authored model, so approval remains blocked on that nit.

Findings:
- `F1` · major · resolved — The display/count projection for resumable terminals previously tested restore target kind independently from the actual wake path, so it could show/count targets that resumeFormFor would refuse. (packages/integrations/anyagent/src/schemas.ts:92)
- `F2` · minor · resolved — The MRU side effects previously ran after the fold no-op return, so re-running the same non-replayed agent command would fail to refresh recent-agent lastSeen. (packages/server/src/terminalEndpoint/local.ts:775)
- `F3` · nit · open — The top snapshotSession docblock still says the "sleeping pr-from-authored rule" lives in composeTerminalMetadata. That is stale after this change: sleeping pr now comes from the persisted observation projection, not the authored arm. (packages/server/src/terminals.ts:73)

_codex on the rebuttal:_ F1: accepted. Delegating resumableCommand to resumeFormFor closes the drift with wake behavior, and the added tests cover the cases I was concerned about. F2: accepted. Moving MRU tracking above the no-op return restores the old same-command refresh behavior without reintroducing persistence churn. F3: partially fixed, but not fully resolved; packages/server/src/terminals.ts:73 still contains the stale pr-from-authored wording, so this finding stays open as a nit.


**claude** — Fixed the last open nit: the stale `snapshotSession` docblock now describes the sleeping arm's restore-relevant `PersistedObservation` projection instead of the dead "pr-from-authored" rule; `just check` + `just fmt` green.

- `F1` **fixed** (already resolved round 1, codex accepted) — `resumableCommand` now gates on `resumeFormFor(target)` before returning a display command, so the display/count projection can't drift from the wake path (`packages/integrations/anyagent/src/schemas.ts:92`). No change this round.
- `F2` **fixed** (already resolved round 1, codex accepted) — MRU side-effects (`trackRecentRepo`/`trackRecentAgent`) run BEFORE the `current === before` fold no-op early-return, restoring the same-command `lastSeen` refresh (`packages/server/src/terminalEndpoint/local.ts:775`). No change this round.
- `F3` **fixed** — Conceded. Round 1 fixed the inline comment (lines 80-87) but missed the docblock above it. The docblock at `packages/server/src/terminals.ts:72-73` still read "the live-half strip (and the sleeping `pr`-from-authored rule) lives in exactly one place", which is stale: `composeTerminalMetadata` (`packages/common/src/surface.ts:1107-1117`) derives the sleeping arm's `pr` from `PersistedObservationSchema.parse(observation)` — the observation projection — NOT the authored arm, and the frozen-`pr` authored field is gone (`surface.ts:291,350`). Rewrote the sentence to: "reusing the one join at both means the sleeping arm's restore-relevant projection — the live-half strip down to `PersistedObservation` (`cwd · git · pr`, `pr` riding the observation now, not a frozen authored field) — lives in exactly one place, so disk and the client read can never diverge." Now consistent with the corrected inline comment and with `PersistedObservationSchema` (`surface.ts:217-234`). Comment-only edit; `just check` (tsc + biome over 1020 files) clean, `just fmt` clean.

commit: `573ebc78f2404c621f1ff8168538872818fb9ffe`


### Round 3

**codex** — approved: `true`

All previously raised findings are resolved. I verified the current tree status, ignored the debate scratch path as requested, and checked the final F3 edit in `snapshotSession`; the docblock now matches the persisted-observation model. I did not raise any new findings from this round's comment-only fix.

Findings:
- `F1` · major · resolved — The display/count projection for resumable terminals previously tested restore target kind independently from the actual wake path, so it could show/count targets that `resumeFormFor` would refuse. (packages/integrations/anyagent/src/schemas.ts:92)
- `F2` · minor · resolved — The MRU side effects previously ran after the fold no-op return, so re-running the same non-replayed agent command would fail to refresh recent-agent `lastSeen`. (packages/server/src/terminalEndpoint/local.ts:775)
- `F3` · nit · resolved — The `snapshotSession` docblock still described the old sleeping `pr`-from-authored model after `pr` moved into the persisted observation projection. (packages/server/src/terminals.ts:72)

_codex on the rebuttal:_ F1: remains resolved; no change this round. F2: remains resolved; no change this round. F3: accepted. The stale docblock at `packages/server/src/terminals.ts:72-76` now correctly describes the live-half strip down to `PersistedObservation` (`cwd`, `git`, `pr`) and explicitly states that `pr` rides the observation, not a frozen authored field.

srid:

Two passes over the diff (vs the merge-base): **Pass 1 — rule checklist**, **Pass 2 — fact-check**. Pass 3 (elegance) skipped — the `/simplify` track already ran over this tree. Pass 2 converged with Pass 1 on the one real correctness issue (the migration guard) and **exonerated** the rest of the shell (it explicitly verified `restoreTargetOf`'s guard, `resumableCommand`'s `none`-for-exhaustiveness arm, the emit fences + recency + MRU ordering, and the client gating are all correct).

| Finding | Verdict | Action |
| --- | --- | --- |
| **`migration-shape-guard`** (both passes) — `backfillAwarenessCutover` checked `agentSession` key *presence* but not value *types*, so a corrupt `{ kind: 42, id: {} }` built an `exact` target that fails `RestoreTargetSchema` and **drops the whole terminal** at the read boundary. | Fix | Validate `kind` (`AgentKindSchema`) + `id` (string); a bad ref falls to `legacyMostRecent` — `c6c8a5bb2` |
| **`errors-must-log-at-error`** — `commitObservation`/`updateMemory` logged the "after removal" path (a teardown-ordering bug) at `debug`, invisible in prod. | Fix | Bumped both to `warn` — `e8b2f2443` |
| **`callback-fanout-guarded-at-funnel`** — the host `emit` fans out to every sensor; a throw escaping it permanently ends that one sensor's `for await` subscription. | Fix | Wrapped `emit` once at the funnel in `startAwarenessEngine` (logs, not fatal) — `766b0fc5b` |
| **`prefer-ts-pattern`** (4 sites) | Partial | **`foldObserved`** converted to `match().exhaustive()` (`0879ac181`) — ts-pattern is already a terminal-workspace dep and it removes the nested agent ternary. **Skipped** `restoreTargetEqual` (a clean exhaustive switch; a two-union compare reads no better in ts-pattern) and `resumeFormFor`/`resumableCommand` (in `anyagent`, which has no ts-pattern dep — adding one to a low-level package for two already-exhaustive, fact-check-validated functions would force a nix FOD-hash refresh disproportionate to the gain). |

All other rules: clean. `just check` (tsc + biome) green, 939 unit tests pass, Atlas sync idempotent.

srid:

The **x86_64-linux** lane ran on **`kolu-ci-3`** — commit `0879ac181`, **failed** ✗

- **Lane wall** (pipeline): **2m37s**

| recipe | duration | |
|---|--:|:--|
| `ci::home-manager` | 2m28s | ✓ |
| `ci::e2e` | 2m20s | ✗ (exit 1) |
| `ci::unit` | 54s | ✓ |
| `ci::nix` | 53s | ✓ |
| `ci::smoke` | 49s | ✓ |
| `ci::pnpm-hash-fresh` | 37s | ✓ |
| `ci::fmt` | 32s | ✓ |
| `ci::biome` | 32s | ✓ |
| `ci::install` | 9s | ✓ |
| `_ci-setup` | 7s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|
| `kolu-ci-1` | ? | ✓ idle |
| `kolu-ci-2` | ? | ✓ idle |
| `kolu-ci-3` | ? | ✓ idle |
| `kolu-ci-4` | ? | ✓ idle |
| `kolu-ci-5` | ? | ✓ idle |
| `kolu-ci-6` | ? | ✓ idle |
| `kolu-ci-7` | ? | ✓ idle |
| `kolu-ci-8` | ? | ✓ idle |

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/0879ac1/timings.jsonl`; pool state is a live `flock` probe.</sub>

PR #1630, opened 2026-06-30
Hook-free done-signal: kaval-tui wait --until idle:<ms>|match:<re>

### LINKED ISSUE #1629: kaval-tui wait --until idle:<ms>|match:<re> — hook-free daemon-side done-signal for driving agent terminals
## Summary

Add a **daemon-side, hook-free "done-signal"** to `kaval-tui` for driving agent terminals from the outside:

```
kaval-tui wait <id> --until <cond> [--timeout <ms>] [--json] [--socket <path> | --host <ssh>]
   <cond> = idle:<ms>      # block until the PTY output has been QUIESCENT for <ms>
          | match:<regex>  # block until new output matches <regex>
```

It blocks **in the daemon, on the raw PTY byte stream**, and returns the instant the condition is met.

## Motivation

When one agent drives another through a kolu PTY (the `/kolu` skill — `kaval-tui create/send/snapshot/kill`), the missing piece is a reliable **"the inner agent's turn ended"** signal. Today there are only two options, and neither fits a raw agent terminal:

1. **`pulam-tui wait --until <buckets>`** keys on shell rc-**hooks** (agent-state via OSC marks). The raw `claude` / `codex` / `grok` / `opencode` CLIs we spawn with `kaval-tui create` are **not hooked**, so `wait` just times out (the skill documents this caveat).
2. **Screen-settle polling** — `snapshot --viewport` in a loop, diffing for "did the screen stop changing" and/or grepping a hand-maintained busy-word list (`esc to interrupt`, `Baking…`, `Fermenting…`, `Churned`, spinner glyphs…).

The screen-settle approach is what the skill currently recommends, and in practice it's **brittle and laggy**:

- **Detection is guesswork.** The busy-word list is per-agent and incomplete — e.g. a watcher that didn't know `Baking…` reported a *false settle* mid-run. Each agent CLI renders "busy" differently.
- **It's coarse.** Polling on a 60–150 s interval lags the actual settle, and the diff-the-viewport heuristic mis-reads a busy agent that pauses mid-thought as "settled."
- **The poll loop is fragile.** A long-running background poll loop is easily killed, losing the watch entirely.

A daemon-side wait fixes the first two by construction: the daemon already sees every PTY byte (it serves `snapshot`/attach), so **output-quiescence is exact and agent-agnostic** — "no bytes for N ms" works identically for claude/codex/grok/opencode, with no busy-word table to maintain, and it fires the moment the agent settles instead of on the next poll tick.

## Proposed behaviour

- **`--until idle:<ms>`** — resolve when no output bytes have been emitted for `<ms>` (the agent-agnostic "turn ended / awaiting input" signal). This is the common case.
- **`--until match:<regex>`** — resolve when output (new bytes since the call, or the viewport) matches `<regex>` — e.g. a completion marker, or a returned-prompt sentinel.
- **`--timeout <ms>`** — fail loud if the condition isn't met in time (mirror `pulam-tui wait`'s loud-timeout discipline) so a wedged agent can't hang the caller.
- **`--json`** → `{ id, fired: "idle" | "match" | "timeout", elapsedMs, matchedLine?, viewport? }`.
- **Exit codes** (mirror `pulam-tui wait`): `0` condition met · `2` timeout · `3` the terminal **exited** before the condition (the driven agent died).
- Works over `--socket <path>` and `--host <ssh>` like the other `kaval-tui` verbs (a remote PTY's quiescence is observed at the remote daemon).

### Notes for implementation
- The signal source is the same PTY output stream that backs `snapshot`/attach (`packages/kaval/src/ptyHostSurface.ts`, `packages/kaval/src/inProcessPtyHost.ts`); `idle` is a debounce on that stream, `match` is a scan of it. CLI dispatch lives in `packages/kaval/src/bin.ts`.
- `idle` should measure **output** quiescence (bytes the daemon emits to the client), independent of any agent-state detection — explicitly NOT the `pulam-tui` hooked path.

## Acceptance criteria
- [ ] `kaval-tui wait <id> --until idle:800` returns promptly (sub-second after the 800 ms quiet window) when a raw `claude`/`codex`/`grok` terminal stops streaming, with **no** hook/rc setup.
- [ ] `--until match:'<re>'` returns when matching output appears; `--timeout` exits `2` on no-match; a terminal that exits first exits `3`.
- [ ] `--json` shape as above; works over `--socket` and `--host`.
- [ ] A test pins quiescence detection (a fixture that emits, pauses > window → resolves; keeps emitting → blocks until timeout).

## Skill updates (required, same PR)
Update the **`/kolu`** skill — source `agents/.apm/skills/kolu/SKILL.md`, then regenerate with `just ai::apm` (it deploys to `.claude/`, `.codex/`, `.agents/`, `.opencode/`):

- In **"The done-signal"** section, make `kaval-tui wait <id> --until idle:<ms> --timeout <ms>` the **preferred** done-signal for raw `kaval-tui create` terminals, replacing the `wait_until_settled` snapshot-poll recipe (keep screen-settle only as a fallback for a daemon that predates the subcommand).
- Update **"The loop"** example to: `send` (type) → `send --key Enter` (submit) → `kaval-tui wait --until idle:<ms>` → `snapshot --viewport` (read).
- Clarify the **`pulam-tui wait` vs `kaval-tui wait`** split: `pulam-tui wait` = agent-state buckets (hooked terminals only); `kaval-tui wait` = raw output quiescence/match (**any** terminal, no hooks). Point raw-agent drivers at `kaval-tui wait`.
- Note `--timeout` (loud-fail) and the `match:` form for completion/prompt sentinels.

## Stretch (separate, optional) — kill-proof notify
A blocking `wait` is still a client-side process that this environment can kill mid-wait. A fire-and-forget, daemon-evaluated variant would survive the client dying:

```
kaval-tui notify <id> --until <cond> --touch <path>   # daemon drops a sentinel file when <cond> fires
```

Out of scope for this issue, but noted: it's the robust path for unattended/long-horizon orchestration where the watcher process can't be kept alive.

---
*Filed from `/kolu` multi-agent orchestration — the screen-settle watcher was the weakest link across a long review-debate run; a daemon-side `wait` keyed on raw PTY output would retire the keyword-grep loops entirely.*

srid:
**`kaval-tui` can now block until a terminal's turn ends — keyed on its raw output, with no shell hooks.** When one agent drives another through a kolu PTY (the `/kolu` skill), the missing piece was a reliable "the inner agent's turn ended" signal for a *raw* `kaval-tui create`'d terminal. The two existing options don't fit: `pulam-tui wait` keys on shell rc-hooks (the raw `claude` / `codex` / `grok` / `opencode` CLIs we spawn aren't hooked, so it just times out), and screen-settle polling is brittle (a per-agent busy-word list that mis-fires) and laggy (it lags the real settle by a poll tick).

`kaval-tui wait <id> --until idle:<ms>` blocks on the terminal's **raw output stream** and returns the instant no byte has arrived for `<ms>`. The daemon already sees every output byte — it serves them to `attach`/`snapshot` — so output-quiescence is exact and agent-agnostic: the same command works for every CLI, with no busy-word table and no poll-tick lag.

Closes #1629.

### The done-signal, as a loop

```
kaval-tui send  "$id" "refactor the parser"; kaval-tui send "$id" --key Enter
kaval-tui wait  "$id" --until idle:800 --timeout 600000   # ← block until quiet
kaval-tui snapshot "$id" --viewport                       # then read the reply
```

- **`--until idle:<ms>`** — resolve once no output byte has arrived for `<ms>` (the turn ended / it's awaiting you). The common case.
- **`--until match:'<regex>'`** — resolve once *new* output matches a completion marker or returned-prompt sentinel.

### Exit codes (mirroring `pulam-tui wait`)

| code | meaning |
| --- | --- |
| `0` | condition met (`--json` → `{ id, fired, elapsedMs, matchedLine? }`) |
| `2` | `--timeout <ms>` elapsed — fail loud so a wedged agent can't hang the loop |
| `3` | the terminal **exited** before the condition (the agent you were driving died) |
| `130` | interrupted (Ctrl+C) |

Works over `--socket` and `--host` like every other verb — a remote PTY's quiescence is observed at the remote daemon.

### Design

This is a **client-side leaf**, not a daemon-side mechanism:

- **Reuse the source of truth** — it consumes the *existing* `terminalAttach` (snapshot-then-`delta`) + `exit` streams. **No new contract member, no `PTY_HOST_CONTRACT_VERSION` bump, no daemon change.** The output tap the issue names as the signal source already exists; `wait` is a debounce/scan over it.
- **Electricity boundary** — the raw-output volatility stays owned by kaval (the daemon); the bounded debounce/scan lives in the kaval-tui client beside `snapshot`/`send`/`attach`, not folded into the daemon as a new receptacle.
- **Fail fast, no fallbacks** — invalid `--until` / regex / `--timeout` fail loud *before* dialing; a timeout, a dead terminal, and a dropped link each fail loud with a distinct exit code. No graceful-degradation path. (The screen-settle recipe survives in the `/kolu` skill only as documented guidance for a daemon too old to have the subcommand — not a code path.)

The pure `awaitOutputCondition` (mirroring `pulam-tui`'s `read.ts:awaitAgentState` / `main.ts:cmdWait` split) is tested over a **real unix socket against a `cat` PTY**, driving output at test-controlled times: it pins idle-fires-on-quiescence, blocks-while-emitting → timeout, match, terminal-exit → `gone` (exit 3), and Ctrl+C → `interrupted`.

### Docs

The **`/kolu`** skill now leads with `kaval-tui wait --until idle` as the preferred done-signal (screen-settle demoted to an old-daemon fallback), clarifies the `pulam-tui wait` (hooked, agent-state) vs `kaval-tui wait` (any terminal, raw output) split, and was regenerated to `.claude`/`.agents`. The kaval-tui README, the kolu.dev marketing page, the kaval-tui + pulam-tui Atlas notes, and the changelog are updated.

### Try it locally

```sh
nix run github:juspay/kolu/feat/kaval-tui-wait#kaval-tui -- wait --help
```


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `d60c7643d4f1`

### Applied (4)
- `lowy-4` JSON 'fired' field not a single typed source of truth — commit `b4b37a17a`
- `hickey-1` JSON wait-result surface is fragmented and `fired` overloads two axes — (uncommitted)
- `hickey-2` `met` variant lets `fired`/`matchedLine` drift into illegal states — commit `bb53f4c68`
- `hickey-4` Wait-flag validation fragmented across the dial via a nullable carrier — commit `8d9349632`

### Agreed — no change (5)
- `lowy-1` Electricity verdict: leaf over existing receptacle (affirm) (packages/kaval-tui/src/wait.ts:207)
- `lowy-2` abortOnShutdownSignals duplicated verbatim across two CLIs (packages/kaval-tui/src/main.ts:519)
- `lowy-3` Terminal-byte display-cleaning sprayed across three helpers (packages/kaval-tui/src/wait.ts:116)
- `lowy-5` idle/match strategy dispatch inlined in stream loop (packages/kaval-tui/src/wait.ts:211)
- `hickey-3` `abortOnShutdownSignals` duplicated verbatim across the two sibling TUIs (packages/kaval-tui/src/main.ts:519)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `d60c7643d4f1`

### Round 1

**codex** — approved: `false`

The core shape is reasonable and stays within the existing pty-host surface, but I found one correctness issue in the wait race handling: if either stream ends or errors before the condition is met, the code can still report success or hang because the idle timer and sibling stream keep running. I also found smaller issues around timer validation, docs drift from the JSON contract, and test coverage/flakiness for the new CLI path.

Findings:
- `F1` · major · open — `consumeOutput` treats a clean `terminalAttach` end, and non-abort stream errors, as just another completed task. For `idle`, the last armed idle timer remains live after the output source is gone, so a dropped/ended attach stream can falsely resolve `met` even while the PTY is still producing output. For `match`, the wait can stop observing output and then hang until timeout or terminal exit. This is a known stream mode in this codebase: `Channel` can end slow subscribers, and `runAttach` explicitly re-attaches on clean stream end. (packages/kaval-tui/src/wait.ts:247)
- `F2` · minor · open — `idle:<ms>` accepts arbitrarily large digit strings, and `--timeout` only checks finite positive numbers in `main.ts`. Node timers overflow above the signed 32-bit delay limit and are effectively clamped, so values like `idle:3000000000` or a huge `--timeout` can fire almost immediately instead of after the requested duration. (packages/kaval-tui/src/wait.ts:53)
- `F3` · minor · open — The skill docs describe `kaval-tui wait --json` as `{ id, fired, elapsedMs, matchedLine? }` and even say `fired` can be `timeout`, but the implementation emits `{ id, result, ... }`, with `fired` only on successful `met` frames. The changelog repeats the shorter incorrect shape at `website/src/content/changelog/unreleased.mdx:7`. This will mislead scripts and agents consuming the documented contract. (agents/.apm/skills/kolu/SKILL.md:121)
- `F4` · minor · open — The match test relies on `await sleep(100)` to assume the wait subscription is installed before writing the marker. Because `match` deliberately skips the snapshot frame, if CI is slow and the marker lands before `terminalAttach` is subscribed, the marker becomes prior screen state and the test times out instead of matching. (packages/kaval-tui/src/wait.test.ts:197)
- `F5` · minor · open — The tests exercise `parseUntil` and `awaitOutputCondition`, but they intentionally do not run `cmdWait` or the `kaval-tui wait` CLI. That leaves the user-facing contract unpinned: argument validation, dispatch, JSON frames, and exit codes 0/2/3/130/1 are all new behavior in `main.ts` and are not covered by a child-process CLI test. (packages/kaval-tui/src/wait.test.ts:4)
- `F6` · nit · open — The command synopsis says `kaval-tui wait <id> --until block until...`, which reads as if `block` is the flag value. The top comment in `packages/kaval-tui/src/main.ts:12` has the same typo. (packages/kaval-tui/README.md:18)


**claude** — Fixed the F1 lost-feed race, the F2 timer-overflow gap, the F3/F6 doc drift, and de-flaked the F4 match test; pinned the JSON contract for F5 and disputed only the child-process CLI test.

- `F1` **fixed** — Real correctness bug, confirmed against `channel.ts` (drop-slow-subscriber ends a subscriber's iterator while the PTY is still live) and `attach.ts:runAttach` (re-attaches on that identical clean end). Before, when `terminalAttach` ended cleanly (slow-consumer drop) or errored non-abort *before an outcome*, `consumeOutput` just returned: for `idle` the last `armIdle` timer stayed live and fired a FALSE `met`; for `match` we stopped observing and hung to the timeout. Fixed in `packages/kaval-tui/src/wait.ts`: added `disarmIdle()` and a new `settleOnLostFeed()` (wait.ts:~232-260) called from both the post-`for await` clean-end path (`if (outcome === undefined && !abort.signal.aborted)`) and the non-abort `catch`. It disarms the idle timer FIRST (so no false `met` during the await), then discriminates via the inventory exactly like `runAttach.resolveGone` — PTY no longer listed → `gone`; still live → loud `closed` (re-run hint). I chose fail-loud `closed` over a re-attach loop on purpose: for a one-shot `wait`, re-attaching after a drop would silently MISS the dropped frames (a `match` marker could be in them), so failing loud is the honest answer and matches codex's offered "fail loud with `closed`" option. The normal idle-success (stream stays open, timer fires, we abort) and the PTY-exit race (both `consumeExit` and `settleOnLostFeed` resolve `gone`, first-writer-wins) are unaffected — verified by the existing `gone`/`interrupted`/`timeout`/`idle` tests still passing (100/100).
- `F2` **fixed** — Real fail-fast violation. `setTimeout` clamps a delay > 2_147_483_647 to ~1ms, so `idle:3000000000` or a huge `--timeout` fired almost immediately (a false `met`/timeout) instead of crashing loud. Added `export const MAX_TIMER_MS = 2_147_483_647` in `wait.ts:46`, an upper-bound guard in `parseUntil`'s idle branch (wait.ts:66-74), and the same `<= MAX_TIMER_MS` clause in `main.ts`'s `--timeout` validation (main.ts:~800), reusing the one constant per "reuse the source of truth". New test `rejects an idle window above the setTimeout ceiling` (wait.test.ts) pins the boundary (2147483647 accepted, +1 rejected).
- `F3` **fixed** — Real doc drift: the lens debate changed the frame to `{ id, result, … }` (result discriminant; `fired` is a `met`-only detail) but the docs still said `{ id, fired, elapsedMs, matchedLine? }` with "`fired` can be `timeout`". Updated all three hand-synced SKILL.md copies (`agents/.apm/...`, `.claude/...`, `.agents/...` — kept byte-identical) at line 121 and the changelog at `website/src/content/changelog/unreleased.mdx:7` to the actual shape. Grepped the whole repo (md/mdx/astro/ts, excl. node_modules + atlas dist) — no stale shape remains.
- `F4` **fixed** — Real flakiness. `match` deliberately skips the snapshot, so a single marker written after a fixed `await sleep(100)` is lost to prior screen state if the attach subscription isn't live yet on a slow CI. Rewrote the test (wait.test.ts) to re-emit the marker in a loop until the wait resolves (`void p.finally(() => resolved = true)`), so whichever write lands as the first post-subscribe delta matches — no fixed-sleep guess.
- `F5` **partial** — Agreed on the gap and pinned the most load-bearing, drift-prone piece: added a pure `waitResultJson` unit-test block (wait.test.ts) covering all six outcome frames (`met`-idle/`met`-match/`timeout`/`gone`/`interrupted`/`closed`, incl. the conditional `error`/`matchedLine`), which locks the public `--json` contract the docs (F3) describe. I dispute the full child-process CLI test (exit codes / dispatch / arg validation): the sibling `packages/pulam-tui/src/wait.test.ts` also drives the pure layer (`awaitAgentState`) directly and never child-process-tests `cmdWait` (its header says so: "NOT cmdWait, which calls process.exit and would kill the runner") — the established TUI pattern is pure-data-layer over a real socket + thin `process.exit` glue. Arg validation is already covered by the `parseUntil` grammar tests plus the new F2 cap test; the exit-code glue (0/2/3/130/1) is a thin map over the well-tested `awaitOutputCondition` and mirrors pulam-tui's tested structure. Adding a bespoke spawn-the-CLI harness would introduce a pattern neither TUI package has, for glue that is already exercised one layer down.
- `F6` **fixed** — Real synopsis typo: `--until block until…` read as if `block` were the flag value. Changed `packages/kaval-tui/README.md:18` and the `main.ts:12` header comment to `--until <cond>`.

commit: `a776cb98fab427c0adaf392f7889b79fc96088ae`


### Round 2

**codex** — approved: `true`

All six findings from the prior round are resolved in the current tree. The lost-feed race is now handled fail-loud with the idle timer disarmed first, timer overflow is rejected at the boundary, the JSON docs match the actual frame shape, the match test no longer depends on a fixed subscription sleep, the JSON contract is pinned in tests, and the synopsis typo is fixed. No new regression from this round stood out in the scoped re-check.

Findings:
- `F1` · major · resolved — The prior lost-feed race is addressed: when `terminalAttach` ends/errors before an outcome, the code now disarms the idle timer and settles as `gone` or fail-loud `closed` instead of allowing a false idle success or waiting on a dead feed. (packages/kaval-tui/src/wait.ts:272)
- `F2` · minor · resolved — The timer overflow gap is addressed with `MAX_TIMER_MS`, idle parse validation, timeout validation in `main.ts`, and boundary tests. (packages/kaval-tui/src/wait.ts:46)
- `F3` · minor · resolved — The documented `--json` frame now matches the implementation: `{ id, result, ... }`, with `fired` only on `met` results. The changelog was updated too. (agents/.apm/skills/kolu/SKILL.md:121)
- `F4` · minor · resolved — The match test no longer relies on a fixed sleep before a one-shot marker; it re-emits until the wait resolves, so the first post-subscribe delta can match without a CI timing guess. (packages/kaval-tui/src/wait.test.ts:247)
- `F5` · minor · resolved — The public JSON payload is now covered for all wait outcomes via `waitResultJson` tests. I accept the dispute on a full child-process CLI test because the remaining exit-code glue is a direct switch over the tested pure outcome and matches the existing pulam-tui testing pattern. (packages/kaval-tui/src/wait.test.ts:135)
- `F6` · nit · resolved — The synopsis typo is fixed: it now says `--until <cond>` instead of implying `block` is the flag value. The `main.ts` header comment was fixed too. (packages/kaval-tui/README.md:18)

_codex on the rebuttal:_ F5: I accept the dispute. The added `waitResultJson` tests cover the drift-prone public payload, and the remaining `cmdWait` exit-code behavior is thin, readable glue over `awaitOutputCondition`. Given pulam-tui uses the same pure-layer-over-real-socket testing pattern and does not child-process-test `cmdWait`, I no longer think a bespoke CLI harness is necessary for this PR.

srid:

Rules + fact-check passes over the `kaval-tui wait` diff (vs `d60c7643d`), scoped to `packages/kaval-tui/src/{wait,main}.ts`. Run with `--no-elegance` (the `/simplify` track already covered that pass). **Applied 4, skipped 2 (with justification).** Committed as `2a57d01c8`.

### Applied
| Finding | Rule | Fix |
| --- | --- | --- |
| `matchedLine` could leak raw **OSC** bytes (`\x1b]0;…\x07`) — `cleanLine` stripped only CSI | `prefer-focused-library` | Strip OSC (BEL/ST-terminated) too, keeping it a self-contained leaf (no new dep / lockfile churn for a cosmetic field) |
| `consumeExit`'s catch dropped its error with no comment, and wrote `upstreamError` that the common `gone` path discards | `no-silent-error-swallowing` (P1) + fact-check (P2) | Document why an exit-stream error is **non-fatal** (output monitoring is the backstop; a real exit ends `terminalAttach` → `settleOnLostFeed` → `gone`); drop the misleading `upstreamError` write |
| `cmdWait` timeout message used `opts.timeoutMs` (`number\|undefined`) → could print `"undefinedms"` | fact-check | Report `outcome.elapsedMs` (always populated) instead |
| Inner `abort` / final `interrupted`/`closed` fallback under-explained | `comment-the-non-obvious` | Added a comment on the no-settle safety net (the race-coordinator role was already commented) |

### Skipped (justified)
- **`prefer-ts-pattern`** (use `match().exhaustive()` for `cmdWait` / `waitResultJson`) — `cmdWait`'s if-chain **deliberately mirrors `pulam-tui`'s `cmdWait`** (the parity the lens + codex debates preserved), and `waitResultJson`'s `switch` is already return-exhaustive (a new `WaitOutcome` variant fails the build with "not all paths return"). Adopting ts-pattern in only one of the two sibling CLIs would break the blessed parity.
- **`process.once` for `abortOnShutdownSignals`** (vs `process.on`) — a *hypothetical* listener leak only if called >1×/process; it's invoked exactly once per one-shot CLI run, matches `cmdAttach`, and the lens debate explicitly blessed verbatim parity with `pulam-tui`'s identical helper.

Typecheck + `biome lint` + the 100-test `kaval-tui` suite green after the fixes.

srid:

The **x86_64-linux** lane ran on **`kolu-ci-7`** — commit `2a57d01c8`, **passed** ✓

- **Lane wall** (pipeline): **3m19s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 2m44s | ✓ |
| `ci::home-manager` | 2m10s | ✓ |
| `ci::unit` | 57s | ✓ |
| `ci::nix` | 55s | ✓ |
| `ci::smoke` | 54s | ✓ |
| `ci::flake-check` | 53s | ✓ |
| `ci::fmt` | 32s | ✓ |
| `ci::biome` | 32s | ✓ |
| `_ci-setup` | 22s | ✓ |
| `ci::install` | 13s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|
| `kolu-ci-3` | dev-x86-64-linux-03 | 🔒 leased |

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/2a57d01/timings.jsonl`; pool state is a live `flock` probe.</sub>

srid:
## Evidence

A **real terminal recording** of the PR branch's `kaval` + `kaval-tui` (nix-built, kaval-tui `4.0`) **running on a clean `pu` box** — driving a `cat` PTY (a stand-in agent) via `kaval-tui send` and blocking on `kaval-tui wait`. `kaval-tui wait` is a headless CLI done-signal, so the proof is the terminal itself, not a screenshot.

kaval-tui wait — live recording

Verbatim transcript (`asciinema` capture of the same session — real UUIDs, real `elapsedMs`; the `— sent N bytes` lines are the demo's own `kaval-tui send` calls feeding the `cat` terminal):

```console

# 1) --until idle:800  — returns ~0.8s after output goes quiet
$ kaval-tui create -- cat   -> 2ceb3cfa
$ kaval-tui wait 2ceb3cfa --until idle:800 --timeout 10000 --json
— sent 12 bytes to 2ceb3cfa
— sent 1 byte to 2ceb3cfa · keys: Enter
{
  "id": "2ceb3cfa-223c-4433-bbd1-15c8819ec11d",
  "result": "met",
  "fired": "idle",
  "elapsedMs": 1456
}

# 2) --until match:'DONE-MARK'  — fires on a completion marker
$ kaval-tui wait 0b3e9210 --until match:DONE-MARK --json
— sent 9 bytes to 0b3e9210
{
  "id": "0b3e9210-4537-4bdb-989b-07ff29c198f8",
  "result": "met",
  "fired": "match",
  "elapsedMs": 421,
  "matchedLine": "DONE-MARK"
}
— sent 1 byte to 0b3e9210 · keys: Enter

# 3) --timeout 1500 on a never-quiet terminal (output every 100ms) — exit 2
$ kaval-tui wait fd7e0976 --until idle:500 --timeout 1500 --json
{
  "id": "fd7e0976-3aed-48c7-b000-df928d35496c",
  "result": "timeout",
  "elapsedMs": 1501
}
kaval-tui: timed out after 1501ms waiting for fd7e0976 (output never met the condition).
exit code: 2

# 4) terminal exits before the condition (exits at ~1s) — exit 3
$ kaval-tui wait b8a34330 --until idle:5000 --json
{
  "id": "b8a34330-0572-4741-bf58-99b381b131de",
  "result": "gone",
  "elapsedMs": 782
}
kaval-tui: b8a34330 exited before the condition was met — its terminal is gone.
exit code: 3

# 5) condition met, no --json — human stderr line, exit 0
$ kaval-tui wait 5c4a5c77 --until idle:600
— sent 2 bytes to 5c4a5c77
— sent 1 byte to 5c4a5c77 · keys: Enter
— 5c4a5c77 output idle after 1132ms
exit code: 0
```

Every outcome is exactly as documented: **idle** fires ~0.8s after the burst stops (the window resets on each delta, so it lands at ~1.45s = 0.4s emit + 0.8s window, not at 0.8s); **match** returns the clean `matchedLine`; **timeout** fails loud with exit **2**; a terminal that exits first is `gone` with exit **3**; a plain met is exit **0**.

This rides on the **100-test `kaval-tui` unit suite** (green in CI on both `x86_64-linux` and `aarch64-darwin`), which pins the same behaviors over a real unix socket against a real PTY — including the gauntlet hardening (the lost-feed race, the timer-overflow ceiling, the split `WaitOutcome` union, and the OSC-stripped `matchedLine`).

PR #1633, opened 2026-06-30
Stop the spurious 'Terminal not found' error when closing a terminal

### LINKED ISSUE #1628: Spurious `Terminal not found` errors from stale `sendInput` after kill
### Summary

Closing a terminal (especially via "Close and remove worktree") can produce \`ERROR Terminal <id> not found\` from \`terminal.sendInput\` — even when no visible keystroke was typed. The error is harmless (the terminal is already gone and the input is correctly dropped), but the ERROR-level log is misleading.

### Reproduction

1. Open a terminal (optionally with an agent process running)
2. Close the terminal via the close-confirm dialog (with or without worktree removal)
3. Check server logs — \`NOT_FOUND\` from \`terminal.sendInput\` may appear

The error appears reliably when a \`sendInput\` RPC is in-flight over the WebSocket at the moment the kill RPC removes the terminal from the server registry.

### Root cause

\`sendInput\` is fire-and-forget everywhere (\`void client.terminal.sendInput(...)\`). The client never awaits the result, so a stale \`sendInput\` races against the kill:

1. Any \`sendInput\` call dispatches — the RPC is now in-flight over the WebSocket
2. Kill RPC arrives — server calls \`unregisterTerminal(id)\`, deleting the Map entry
3. The stale \`sendInput\` arrives — \`requireActiveTerminal\` finds nothing → throws \`terminalNotFound\`

The \`sendInput\` doesn't require a visible user keystroke. The xterm \`onData\` handler (\`Terminal.tsx:858\`) captures any data event — modifier keys, escape sequences, bracketed pastes from agent processes, or terminal query responses. The \`void\` prefix means the RPC races the kill silently.

The kill path itself (\`useWorktreeOps.ts:84\` → \`useTerminalCrud.ts:225\` → \`local.ts:816\`) does not call \`sendInput\`. The stale input is always from a concurrent or prior dispatch, not from the kill sequence.

### Impact

- **Severity**: Low (cosmetic — the error is benign, no data loss or broken behavior)
- **Frequency**: Reproducible when any \`sendInput\` RPC is in-flight at kill time
- **Affected versions**: Current main

### Suggested fix

Downgrade the \`sendInput\` handler's \`requireActiveTerminal\` to a soft check — a stale input after kill is expected behavior, not an exceptional condition:

\`\`\`typescript
// packages/server/src/router.ts:117-119
sendInput: t.terminal.sendInput.handler(async ({ input }) => {
  const proc = getActiveTerminal(input.id);
  if (!proc) return; // terminal killed between dispatch and delivery — drop silently
  proc.handle.write(input.data);
}),
\`\`\`

This uses the existing \`getActiveTerminal\` (which returns \`undefined\` instead of throwing) and silently drops the input when the terminal is gone. The same pattern applies to \`resize\` if desired, though \`sendInput\` is the one that fires from user-facing paths.

srid:
Closing a terminal could spit an `error`-level **`Terminal <id> not found`** line into the server log — even with no keystroke typed. **A terminal's own background reports race the kill:** xterm's `onData` fires fire-and-forget, so a focus-in/out report (`\e[I` / `\e[O`, emitted once an agent turns on focus-tracking) can land a beat *after* the close removed the terminal. The in-flight `sendInput` then misses the registry, throws `terminalNotFound`, and the oRPC pino plugin auto-logs a scary ERROR for a message that was correctly dropped.

Reproduced on current `master` (real browser against real kolu, every `sendInput` intercepted on the wire):

```
sendInput  \e[O   focus-OUT   ← "Close terminal?" dialog opened (before kill — harmless)
kill                          ← confirmed
sendInput  \e[I   focus-IN    ← fired +243 ms AFTER the kill  → ERROR Terminal … not found
```

### The fix

`requireActiveTerminal` is the seam every per-terminal handler shares, and on a miss it **throws** — right for handlers whose contract needs a live PTY (attach, screen reads, paste/upload, kill), but wrong for the two *stream-fired* ops. A `sendInput`/`resize` arriving for an already-closed terminal is a **normal, expected race**, not a fault.

So `sendInput` and `resize` now take the registry's existing **quiet** accessor instead of the throwing one:

```ts
getActiveTerminal(input.id)?.handle.write(input.data);   // was: requireActiveTerminal(input.id).handle.write(...)
```

This is the house idiom already used at `terminalEndpoint/local.ts:1099` (`getActiveTerminal(id)?.handle.ready`) — a missing terminal is a silent no-op for an op that tolerates absence, and on a miss it's strictly *cheaper* than before (no thrown fault to construct). Every other handler keeps the loud `requireActiveTerminal`, so a genuinely never-existed id still surfaces.

> _Not the issue's originally-suggested blanket silent `return`: that swallows a genuinely never-existed id (a real routing bug) and only covers one handler. No tombstone/epoch — `TerminalId` is `crypto.randomUUID`, unique, so there's no killed-vs-never-known ambiguity to disambiguate._

The full root-cause write-up, the captured bytes, and the **server-restart investigation** (restarts do *not* trigger this — the daemon adopts surviving terminals and the stale-tab gate stops a reconnecting tab from firing RPCs) are in the issue thread.

> _Reviewed by the `/be` gauntlet: **lens-debate** (lowy ⇄ hickey) converged on deleting an over-engineered helper in favor of the bare idiom above; **codex** (`xhigh`) approved after a doc-comment fix; **simplify** (4 angles) and **code-police** clean._

Closes #1628

### Try it locally

```sh
nix run github:juspay/kolu/fix/1628-stale-sendinput-after-kill
```


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `dda2597239c3`

### Applied (3)
- `lowy-1` Quiet miss-policy is an inline effect in a "pure accessors" module, breaking symmetry with its loud twin — commit `39d71b1b6`
- `hickey-1` New registry concept reinvents the existing getActiveTerminal?.handle quiet-drop idiom — (uncommitted)
- `hickey-2` Pure-accessor registry module now braids lookup + I/O logging + router op vocabulary, asymmetrically with its 'twin' — (uncommitted)

### Agreed — no change (1)
- `lowy-2` Electricity-test grounding: receptacle = getActiveTerminal; this is an application concern, wires exposed via the op/log coupling (packages/server/src/terminal-registry.ts:178)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `dda2597239c3`

### Round 1

**codex** — approved: `false`

The runtime change is narrowly scoped and matches the deliberate behavior: late `resize`/`sendInput` calls now drop when the active terminal entry is gone, while the other per-terminal handlers still fail loudly. I did not find a correctness bug in that flow. One registry comment is now stale and should be corrected because it documents the old invariant.

Findings:
- `F1` · nit · open — The `requireActiveTerminal` comment still says this helper is for handlers whose contract needs a live PTY, listing `resize` and `input` as examples. This branch intentionally moved `terminal.resize` and `terminal.sendInput` away from the throwing helper, so the comment now records the opposite of the new router invariant and could steer a future cleanup back toward the spurious NOT_FOUND logging this PR fixes. (packages/server/src/terminal-registry.ts:185)


**claude** — Fixed the one stale-comment nit codex flagged; the runtime behavior was already approved.

- `F1` **fixed** — Codex is right: the `requireActiveTerminal` docstring at `packages/server/src/terminal-registry.ts:184-187` still listed `resize, input` as its callers, but this branch deliberately moved exactly those two handlers to the quiet `getActiveTerminal(input.id)?.handle.x()` path. The comment therefore documented the opposite of the new router invariant and could steer a future cleanup back into the spurious NOT_FOUND logging this PR fixes. I rewrote it to (a) name only the genuine fail-loud callers (attach, screen reads, pasteImage, uploadFile, create's parentId check — verified against `router.ts` via grep) and (b) add an explicit NOTE that `resize`/`sendInput` deliberately do NOT use this helper because a late stream op after a kill is an expected race that drops quietly (#1628). Comment-only change; no runtime behavior touched. Verified: `just check` (typecheck across all packages + biome lint) is green.

commit: `1e568cbe8d136c0eb0d3876e24dd718872127682`


### Round 2

**codex** — approved: `true`

The current tree contains the intended runtime change plus the follow-up comment update in `packages/server/src/terminal-registry.ts`. The stale `requireActiveTerminal` doc now explicitly names the fail-loud callers and calls out that `resize`/`sendInput` deliberately use the quiet `getActiveTerminal(...)?` path for late post-kill races. No new regression was introduced by this round's fix.

Findings:
- `F1` · nit · resolved — The prior `requireActiveTerminal` comment was stale because it still listed `resize` and `input` as throwing-helper callers after this branch intentionally moved those two handlers to the quiet optional-chain path. (packages/server/src/terminal-registry.ts:184)

_codex on the rebuttal:_ F1: Verified fixed. CLAUDE updated the helper comment to describe the new invariant accurately, including the explicit exception for `resize` and `sendInput`. There was no dispute to resolve.

srid:

Ran **Pass 1 (rule checklist)** and **Pass 2 (fact-check)** over the diff vs `dda2597`. _(Elegance pass skipped — `/simplify` already ran this gauntlet.)_

**Pass 2 — fact-check:** ✅ no findings. `getActiveTerminal`'s `entry?.handle` narrowing is sound (`TerminalHandle` is an object ref, always truthy when present); the handler swaps, the registry JSDoc, and the changelog prose all match the code.

**Pass 1 — rules:** one finding, **adjudicated as a skip** (false-positive-in-context):

| Rule | Finding | Disposition |
|---|---|---|
| `dry-rule-of-three` | The idiom `getActiveTerminal(id)?.handle.X()` now appears 3× (`local.ts:1099`, `router.ts:118`, `router.ts:125`); suggests extracting `getActiveTerminalHandle(id)`. | **Skipped** |

_Rationale:_ this directly contradicts the **lens-debate consensus** that settled minutes earlier — lowy + hickey explicitly **deleted** a helper here and ruled the bare `getActiveTerminal(id)?.handle.X` form the canonical house style (it is the established idiom at `local.ts:1099` and several other `local.ts` sites). The "duplication" is the trivial `?.handle` access across three sites that call **different** members (`.ready` / `.resize()` / `.write()`); extracting a `getActiveTerminalHandle` wrapper would save one property access while adding a near–thin-wrapper export — colliding with code-police's own `no-thin-wrapper-functions` rule. Structural-review consensus wins over the mechanical rule-of-three.

**Net:** no police commits — the diff is clean.

srid:
## Evidence

This fix has **no on-screen surface** — closing a terminal looks identical before and after; the only observable effect is the *absence* of the spurious server-log ERROR. So the proof is a **behavioral before/after**: drive the exact same close flow against both builds (real browser, real kolu on a clean box, every `sendInput` intercepted on the wire), with focus-tracking enabled (`printf '\e[?1004h'`, what an agent does) so the close dialog's focus shuffle emits a terminal report.

**The race is identical in both builds** — a focus-in report (`\e[I`, bytes `1b 5b 49`) fires ~242 ms *after* the kill. Only the server's reaction differs.

### Before — `master`
Wire:
```
sendInput  \e[O  (1b 5b 4f)   ← focus-out, dialog opened (before kill, harmless)
kill                          ← confirmed
sendInput  \e[I  (1b 5b 49)   ← focus-in, +243 ms AFTER the kill
```
Server log:
```
[17:17:21.242] ERROR (3812): Terminal 8660a95c-…-3ac9b3d1fe3b not found
   {"rpc":{"method":"terminal.sendInput", "id":"fbde1aed-…"}}
```
→ **1 spurious ERROR.**

### After — this branch (`c3b0a33`)
Wire (same race):
```
kill                          ← confirmed
sendInput  \e[I  (1b 5b 49)   ← focus-in, +242 ms AFTER the kill   (identical dead keystroke)
```
Server log:
```
[18:54:22.763] INFO  killing  {"terminal":"85ca4f8c-…-03939648865a"}   ← terminal closed
not found count: 0
ERROR lines:     0
```
→ the same late `\e[I` is **dropped quietly; zero errors.**

The race is real and unavoidable (the client can't recall an in-flight message, and a second device can't know yet); the fix simply makes the server treat that benign late message as the no-op it always was.

_No mobile/touch capture — the change has no on-screen surface (server-side log behavior only)._

srid:

The **x86_64-linux** lane ran on **`kolu-ci-2`** (`dev-x86-64-linux-04`) — commit `c3b0a3369`, **passed** ✓

- **Lane wall** (pipeline): **3m31s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 3m9s | ✓ |
| `ci::home-manager` | 1m50s | ✓ |
| `ci::unit` | 59s | ✓ |
| `ci::nix` | 55s | ✓ |
| `ci::smoke` | 52s | ✓ |
| `ci::biome` | 36s | ✓ |
| `ci::fmt` | 35s | ✓ |
| `_ci-setup` | 12s | ✓ |
| `ci::install` | 10s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/c3b0a33/timings.jsonl`; pool state is a live `flock` probe.</sub>

Issue #1592, opened 2026-06-26
Brave/Android: typing 'ls' then space produces 'lsLS' — composition duplication on soft keyboard


ic4-y:
### Description

On Brave/Android (not reproducible on Chrome/Android or desktop browsers), typing characters into a terminal and then pressing space causes previously typed characters to be duplicated with uppercase variants. For example, typing `ls` then space produces `lsLS` in the terminal.

This is a manifestation of a known xterm.js composition handling bug on Android ([xtermjs/xterm.js#3600](https://github.com/xtermjs/xterm.js/issues/3600)), exacerbated by kolu's contenteditable input surface on touch devices.

### Root cause

xterm.js's `CompositionHelper` has a deduplication flaw: when a non-composition key (space, enter, backspace) arrives during an active IME composition, `_finalizeComposition(false)` sends the accumulated composition text immediately via `triggerDataEvent`, *and* the `compositionend` timeout fires and sends it again. The `_dataAlreadySent` bookkeeping only covers keyCode 229 (the IME "composition character"), not regular keys, so the second emission is not suppressed.

GBoard on Android wraps every keystroke — including plain English letters — in composition events (`compositionstart → compositionupdate → compositionend`). This triggers the deduplication flaw on every character.

kolu's contenteditable input surface (`packages/client/src/terminal/softKeyboardInput.ts`) makes `.xterm-screen` contenteditable on touch devices to summon the soft keyboard. This diverts the browser's composition path away from xterm's hidden textarea, where `CompositionHelper` listens, creating a timing divergence that makes the duplication worse on some Chromium forks (Brave) compared to stock Chrome.

### Steps to reproduce

1. Open kolu on Brave for Android
2. Open a terminal
3. Type `ls` on the soft keyboard
4. Press space
5. Observe `lsLS` instead of `ls `

### Expected behavior

`ls ` — the typed characters plus a space, no duplication.

### Actual behavior

`lsLS` — the typed characters are duplicated as uppercase after the space.

### Environment

- **Browser:** Brave for Android (Chromium-based)
- **OS:** Android
- **Keyboard:** GBoard (default)
- **xterm.js version:** 6.1.0-beta.225
- **Not reproducible on:** Chrome/Android, desktop browsers, iOS Safari

### Possible fix

Guard the `onData` handler in `packages/client/src/terminal/Terminal.tsx` with an `isComposing` check on xterm's hidden textarea:

```typescript
let composing = false;
[...]

PR #1634, opened 2026-06-30
Route Android soft keyboard input past autocorrect

ic4-y:
**Android soft keyboards now feed terminal input through a raw transport instead of xterm's prose-oriented helper textarea.** On Calyx/GBoard and FUTO, the old path let the keyboard keep a word buffer, replay autocorrect replacements like `LS ` after `ls`, apply double-space-period, or consume Backspace for suggestion state before the terminal ever saw it.

The fix gives Android its own hidden `type=password` input. Password inputs avoid the keyboard's suggestion/autocorrect editor mode, and kolu keeps a zero-width sentinel in the field so keyboards that refuse to Backspace from an empty input still emit delete events. `beforeinput` is then forwarded directly into xterm as raw terminal data.

Related: #1592

### What changed

- Android soft keyboard focus is routed to a hidden password input after `term.open()`.
- The input prevents DOM edits, resets its sentinel, and forwards `insertText`, `deleteContentBackward`, and `insertLineBreak` into xterm.
- xterm's Android `keyCode=229` textarea-diff path is bypassed so it cannot replay stale editor state.
- The mobile keyboard feature keeps a lightweight regression guard for typed input, while the real keyboard behavior was verified manually on Android.

### Validation

- `just fmt`
- `just check`
- Manual Android testing with Calyx standard keyboard and FUTO:
  - `ls `
  - repeated spaces
  - punctuation after spaces
  - Backspace after typed text and spaces
  - fast typing followed by spaces

_Generated by OpenCode (model `openai/gpt-5.5`)._


srid:
What is the context here? How do I reproduce this bug in my Android phone (Z Fold 6)?

ic4-y:
Context is in the issue linked. 

Essentially open new kolu in brave, create terminal, type ls + space and the autocorrect will annoy you. 

And that creates many very annoying problems when typing commands of various sorts in shell from phone. 

Something like termux blocks that. And this PR does too, now. In some way. 

Edit: I should say, the initial analysis in the issue appears to have been wrong (in spite of the upstream problems being real) and it took me a while to figure out a solution that works for the actual usability problem. From what I understand it's a bit of a hack that bypasses Android thinking that the terminal is a text area worthy of autocorrection features. I wonder if one can't do this any easier...?

srid:
Relevant? https://github.com/xtermjs/xterm.js/pull/6057

srid:
This may be relevant https://github.com/juspay/kolu/pull/2130 (includes the above xterm.js PR)

PR #1637, opened 2026-06-30
Make the terminal lifecycle honor HostLocation (PR-1, local no-op)

srid:
**Finish what R9.1 started: every remaining terminal *lifecycle* seam now resolves through `HostLocation`**, so a later `F-REMOTE` is pure routing rather than a hunt across call sites. This is the first of three parallel local-preserving prep PRs in the [remote-terminals finale](https://github.com/juspay/kolu/blob/rt-pr1/docs/atlas/src/content/atlas/remote-terminals-finale.mdx) (`#pr1`).

**Pure plumbing — no user-facing change.** Every saved record is `{kind:"local"}` today, so every `resolveTerminalEndpoint(...)` returns `localTerminalEndpoint` and create / restore / sleep / wake / boot-adoption stay **byte-identical**. The `{kind:"remote"}` resolver arm still throws loudly (no remote endpoint exists yet — that's correct, not a gap).

### The four seams (all default to local)

| Seam | Before | After |
| --- | --- | --- |
| **create** | `createTerminal` hard-pinned `localEndpoint.spawnPty` | takes an optional `location` (default `LOCAL_LOCATION`), spawns via `resolveTerminalEndpoint(location)`; `location` rides `TerminalCreateInputSchema` → the create RPC → client `handleCreate` |
| **restore** | `useSessionRestore` *deliberately dropped* `t.location` | forwards it through the create call |
| **sleep/wake** | `sleepTerminal`/`wake` called the local helpers directly | `TerminalEndpoint` gains `sleep`/`releaseSleptPty`/`wake` (local arm = today's bodies verbatim); both route by `entry.meta.location`, exactly as kill/attach already do |
| **boot-adoption** | `ptyHost` singletons (`endpoint`/`ptyHostClient`) | a host-keyed map (one `"local"` entry); `adoptSurvivingSession`/`reconcile` parameterized per host, `adoptLocalOrphan` location-parameterized |

A **sub-terminal inherits its parent's `meta.location`**; an explicit child location that disagrees throws `BAD_REQUEST` (`resolveCreateLocation`). _One PTY tree, one kaval — a wrong-host child is a client bug, rejected loudly rather than spawned on the wrong host._

> **Map scaffold only — no remote is dialed.** Seam 4 stands up the host-keyed *shape* (so F-REMOTE can add a `remote` entry per dialed kaval) without touching ssh / `getHostSession`. The session converge + sleeping-seed stay session-global; F-REMOTE owns the multi-host boot re-dial-re-adopt. Out of scope by design: `RemoteTerminalEndpoint`, the create knob / "On SSH host…" UI, awareness/framer, fs/git.

### Why it's a provable local no-op

- **create / restore** → `resolveTerminalEndpoint(LOCAL_LOCATION)` is the same local endpoint, which still stamps `createAuthoredActive(LOCAL_LOCATION)`.
- **sleep / wake** → `.sleep()`/`.releaseSleptPty()`/`.wake()` are the very methods the old `beginSleepLocal`/`releaseSleptLocalPty`/`wakeLocalTerminal` wrappers delegated to (bodies verbatim; `beginSleep` renamed to `sleep` to satisfy the interface).
- **boot-adoption** → `ptyHostClient = hostPtyClient("local")` resolves the same connection the old `endpoint` global did; `reconcile(live, saved, LOCAL_LOCATION)` filters saved to local records (all are local → identical partition).

**Tests:** `sleepWake` · `reconcile` · `adopt` · `inventoryReconcile` · the restore `isLoading` gate all pass **unchanged**. New tests pin the create→restore `location` round-trip, the sub-terminal **inherit**, and the **reject-mismatch** (`BAD_REQUEST`). Full unit suite green (common 22 · server 162 · client 492 · every package).


srid:

PR #1638, opened 2026-06-30
Awareness ready for any host: framer + terminalEvents wire (PR-3, local no-op)

srid:
**The third of three parallel prep PRs for remote terminals** (the [finale plan](https://github.com/juspay/kolu/pull/1632)). It gets kolu's awareness into the right *shape* and lays a *pipe*, so a remote host's sensors can later feed kolu's same fold — under a hard bar: **no user-facing change; local awareness byte-identical**. The "last active" recency must bump / not-bump exactly as before, and a differential test pins it.

## The framer (phase/seq only)

A new `TerminalFrame` (`snapshot | delta(seq) | gap`) and `createFramer()` in `@kolu/terminal-workspace`. kolu's local producer routes every in-process emit through it, stamped **`delta`** — never `snapshot`, because a snapshot-phased local emit would fold with `live:false` and suppress the fresh-spawn recency bump.

## The `live` rule + baseline → a shared consumer arm

The durable `recencyBaseline` and the liveness decision move out of `local.ts`'s emit closure into `createConsumerArm()`:

```
live = (phase === "delta") && agentIdentityChanged(baseline, value)   // skip "unknown"
```

`fold.ts` is untouched (its internal identity check is redundant-but-harmless). The arm is re-seedable, so F-REMOTE reuses it verbatim for the remote subscribe path. The **adopt guard** survives by construction: `adoptedSnapshot` seeds `agent:null`, so only the seeded baseline keeps the re-resolved agent's delta from spuriously bumping recency.

## The wire

zod `TerminalEventSchema` + `TerminalFrameSchema`, and a `terminalEvents` stream on `terminalWorkspaceSurface` (snapshot-then-delta, monotonic per-subscription `seq`). **pulam** serves it live off a per-terminal event bus — carrying the `commandRun` mark its `snapshots` cache drops; **kolu-server** quiet-defaults it (its in-process fold untouched); **pulam-web** forwards it. Contract **`3.0 → 3.1`** (additive — a new stream; the sibling PR-2 rebases its byte-primitive bump to the next minor).

## Why it's safe

| guarantee | proof |
|---|---|
| local awareness byte-identical | a differential test diffs the consumer arm against an inline transcription of the pre-refactor rule, over event streams covering every kind + the agent firehose |
| fresh-spawn recency preserved; no bump on adopt / sleep-wake | explicit recency-bar tests, incl. one that *removes* the baseline and watches the bug return |
| the served capability is real | `serveTerminalEvents` yields snapshot-then-`commandRun`-delta (the mark the snapshots cache drops) — proven deterministically over an in-process channel, and forwarded through pulam-web; the pulam integration test serves the framed stream over a real unix socket and asserts the leading snapshot reconstructs the live cwd + git branch |
| dashboards unaffected | pulam-tui / pulam-web behavior unchanged — their test harnesses gain only the one-line `terminalEvents` stub the new surface requires (pulam-web also the forwarder + its test); the suites pass |

## Out of scope (untouched)

No remote/ssh dial, no `RemoteTerminalEndpoint`, no create knob, no fs/git/Code tab, no kolu remote-subscribe arm — those are F-REMOTE. `fold.ts` not edited.

## Status

Local checks green (`just check` — tsc + biome — plus the terminal-workspace, pulam, pulam-tui, pulam-web, and kolu-server terminalEndpoint suites). The review gauntlet, CI, and evidence are **not** run yet (opened as a draft for review first).


srid:

PR #1639, opened 2026-06-30
Make fs/git & file-IO host-aware — Code tab on the shared surface (PR-2)

srid:
**Everything a terminal reads or writes on disk now resolves to *its* host** — the Code tab's git status/diff/file-tree/file-content, the binary-preview byte route, paste/upload, and transcript export. This is **PR-2** of the [remote-terminals finale](https://github.com/juspay/kolu/blob/rt-pr2/docs/atlas/src/content/atlas/remote-terminals-finale.mdx#pr2) — a pure local-preserving prep PR. Every host is `{kind:"local"}` today, so the **Code tab is byte-identical on local**; the remote arms are throwing stubs F-REMOTE lights later.

The crux: the Code tab moves **off `koluSurface`'s value-bearing fs/git streams onto the SHARED `terminalWorkspaceSurface`'s procedure + `{seq}` pulse** — the same surface `pulam` serves and a remote host serves over ssh. So a remote tile's Code tab reads its host's mirror behind the same client, with no second read path.

### The client move — procedure + pulse, byte-identical

The old value-bearing stream pushed the whole value on every change; the shared surface sends a tiny `{seq}` pulse and lets the consumer re-query. `createPolledQuery` is the **client dual of the server's `pollOnEvent`** — it turns a procedure + a pulse back into the exact `Subscription` shape the Code tab consumed (a callable value + `.pending()` + `.error()`), so the #818 selection-stability guard is reproduced verbatim.

```
OLD  server pushes ── value-bearing stream ──▶ render latest frame
NEW  subscribeRepoChange ── {seq} pulse ──▶ createPolledQuery re-queries git.getStatus ──▶ render
```

One shared `subscribeRepoChange` pulse drives every repo-keyed query (status ×3 · file list · diff); a **normalized-seq dedup keeps it ONE read on mount** (no double-read vs the old single server read) while still re-querying on real `{seq>0}` deltas and input changes. The binary-preview/iframe-URL orchestration (`isBinaryPreviewable` + the `?v=<mtime>` URL) **moved client-side** onto raw `fs.readFile` + `fs.statFileMtimeMs`; `buildIframePreviewUrl` relocated to the browser-safe `kolu-common/preview`. The now-dead `koluSurface` fs/git streams were removed.

### Three host-scoped byte primitives (contract `3.0 → 3.1`, additive)

| Procedure | What | Guard on the producing host |
| --- | --- | --- |
| `fs.previewRead` | range-capable preview bytes (bytes + content-type + content-range), reusing `@kolu/serve-dir`'s `serveFile` | lexical `..`/`%2f` + realpath/symlink-escape (403) |
| `scratch.write` | paste/upload sink (the surface fs was read-only) | sanitize-to-basename under a per-terminal dir |
| `transcript.read` | full read of an agent's transcript source at an arbitrary host path | fenced under a store root (lexical + realpath) |

`scratch.write`'s scratch-root is a per-home volatility, so it's an injected `serveTerminalWorkspace` dep (kolu-server passes `saveTerminalFile` verbatim; pulam its own tmp root); `fs.previewRead`/`transcript.read` are pure (root-as-input) endpoint methods, shared by both homes. Served at **all three sites** — kolu-server local arm, the pulam daemon live arm (both free via the shared factory), and pulam-web's `buildReServe` **forwards** each. No local consumer yet; F-REMOTE's remote arms call them over the mirror.

### Per-location routing — the `localOnly` seam

The preview byte route (`index.ts`), paste/upload (`router.ts`), and transcript export (`router.ts`) now branch on `entry.meta.location` via a small `localOnly` helper: **local arm verbatim, remote arm a throwing stub** (the no-fallbacks stance — crash loudly rather than serve a remote tile off the wrong local fs). `.exhaustive()` forces a deliberate arm.

### How it honors the design philosophy

- **Fail-fast, no fallbacks** — every remote arm throws (`localOnly`, the byte route, `resolveTerminalEndpoint`'s existing remote throw); the guards surface as `ORPCError`/403, never collapse to empty. The now-unused `onStreamReadError` default was removed so a future poll-shape stream fail-fasts until it supplies a handler.
- **Electricity / volatility boundaries** — `createPolledQuery` is a composition *leaf* over framework primitives (a bounded poll-on-pulse algorithm), so it lives in the client beside its consumers, not folded into `@kolu/surface`. The scratch-write primitive lifts to `@kolu/terminal-workspace/scratch` (host-fs is its domain), shared by kolu-server + pulam.
- **Reuse the existing source of truth** — `fs.previewRead` reuses `@kolu/serve-dir`'s `serveFile` (range + content-type + both guards) wholesale; `transcript.read` reuses kolu-git's `resolveExistingUnder`; the lifted scratch helper replaces the duplicated sanitize.

> **Out of scope** (untouched): any remote/ssh dial, `RemoteTerminalEndpoint`, the create knob, awareness/framer/`terminalEvents`. No `@kolu/surface*` API changed, so no drishti gate.

Local-no-op evidence (the existing Code-tab + #818 e2e green) follows as a comment.

### Try it locally

\`\`\`sh
nix run github:juspay/kolu/rt-pr2
\`\`\`

_Generated by [\`/be\`](https://github.com/srid/agency) on Claude Code (model \`claude-opus-4-8\`)._


srid:
## Evidence — byte-identical Code tab on local

Ran the Code-tab + #818 e2e suite (`features/code-tab.feature`, 116 scenarios) on an ephemeral pu box, three times.

| Run | Code | Result | Failures |
| --- | --- | --- | --- |
| 1 | pre-fix | 107/116 | 5 fs/git scenarios (live diff/preview, binary, file list) |
| 2 | fixed | **115/116** | _Open path:N_ (flake — passed runs 1 & 3) |
| 3 | fixed | **115/116** | _Cmd/Ctrl+F_ (flake — passed runs 1 & 2) |

**Running the e2e early caught two real byte-identical regressions** in `createPolledQuery` vs the old value-bearing streams — exactly the kind a unit test alone misses:

1. **Live re-query dropped** — the diff/preview query's `inputFn` reads a *sibling* query's value (`status`); when the sibling settled on the same pulse, the effect re-ran to *skip* via the dedup, and the skip's per-run `onCleanup` **cancelled the in-flight re-read**. Fixed with a generation counter bumped only when a new read is issued (a skip never cancels).
2. **Tree churn** — the old streams' `isEqual` dropped a pulse re-read equal to the current value; mine emitted a fresh reference every read, churning `treePaths` → Pierre reset filter/expansion. Restored `isEqual` (reusing kolu-git's equality fns via a new browser-safe `kolu-git/equals` export).

Both fixed in `894ce4d`, with regression tests pinning each. After the fix, **every fs/git Code-tab + #818 selection scenario passes consistently** across runs 2 & 3 — byte-identical on local. The lone per-run failure is a *different* unrelated scenario each run (focus/keyboard, view-render waits), each passing in 2 of 3 runs — a pre-existing environmental timing flake under box load, not this change. The official two-platform CI (warm pool boxes) runs the full suite at ship.

Unit evidence: the three byte procedures serve in-process (new `terminal-workspace` tests); `createPolledQuery` unit tests pin single-mount-read + value-held-during-re-read (#818) + no-cancel-on-skip + isEqual dedup; all affected suites + `just check` green.

srid:

PR #1640, opened 2026-06-30
Seal terminal lifecycle dispatch — HostLocation routing enforced by types

srid:
**Every per-terminal lifecycle seam already routes through `HostLocation` — this PR makes the compiler *guarantee* it.** "Hard-pin local / bypass `HostLocation` at a lifecycle seam" is now a **compile error**: there is no symbol a caller can import to reach the local endpoint directly, so a future missed seam is impossible rather than something a reviewer has to hunt for. Pure local no-op — behavior is byte-identical; every existing lifecycle test passes, ported through the sealed seam.

This **supersedes the manual-threading PR-1 (#1637)** — same goal, but enforced by types instead of by review. #1637 should be closed.

### The disease → the cure

The bypass surface was a set of *importable* symbols any seam could reach:

| Escape hatch (deleted) | Replaced by |
| --- | --- |
| `localTerminalEndpoint` instance + eight `*Local*` helpers (`beginSleepLocal`, `discardLocalSleeping`, …) | endpoint methods reached only via the id-keyed façade |
| two `localEndpoint = resolveTerminalEndpoint(LOCAL_LOCATION)` aliases (killAll + fs/git pins) | `forEachHost` (killAll) · `localFsGitEndpoint()` (fs/git, narrowed) |
| `resolveTerminalEndpoint` imported by router/surface/terminals | imported **only** by the lifecycle façade (`terminals.ts`) |
| per-host boot adopt that could `saveSession` on partial host info | `adoptSurvivingHost(scope)` → branded `CompleteHostSweep` → `commitBootAdoption` |

### How dispatch is sealed now

```
caller holds an id ─▶ façade op(id) ─▶ resolveTerminalEndpoint(getTerminal(id).meta.location) ─▶ endpoint.op(id)
                       (terminals.ts)   (package-private to the host registry)
```

- **The id-keyed façade is the only public lifecycle API.** `killTerminal` / `sleepTerminal` / `wakeTerminal` / `discardSleepingTerminal` / `restoreSleepingTerminal` / `attachTerminal` each resolve `getTerminal(id).meta.location` *internally*. The caller holds an id; routing is automatic; there is no location param to hard-pin.
- **Create is the sole location-supplying entry** (no id yet): `createTerminal(…, location = LOCAL_LOCATION)` → `resolveTerminalEndpoint(location).spawnPty(…, { location })`. The façade hands the location to `spawnPty`, which stamps `meta.location` from it — *the endpoint never stamps a host literal of its own*, so "endpoint stamps the wrong location" is unspellable.
- **Session-global ops can't act on partial host info.** `forEachHost` over the package-private registry is the only killAll entry ("killAll on only the local host" is unspellable). Boot adoption splits: `adoptSurvivingHost(scope)` reconciles **one** host and *returns* its `HostAdoptionResult` (no `saveSession` in scope); `commitBootAdoption` is the only place that seeds sleeping records + saves, gated on a **branded `CompleteHostSweep`** that only `completeSweep([…all host results])` mints. Committing one host's result is a **type error** — the round-1 "save on partial host info" bug is structurally impossible.
- **fs/git keeps its server access** through a sealed `localFsGitEndpoint()` that returns only `{ fs, git }` (`TerminalWorkspaceEndpoint`) — Code-tab access stays alive across the boundary without re-exposing a kill/sleep hatch. *(PR-2 builds per-location fs/git on top.)*
- **#1637's location-supply, folded in:** `location?` on the create RPC + client `handleCreate`; `useSessionRestore` stops dropping `t.location`; `resolveCreateLocation` is pure inherit/reject (a sub-terminal inherits its parent's host; an explicit mismatch → `BAD_REQUEST`).

### The proof — former hatches now fail to compile

A throwaway file attempting the old bypasses:

```
src/_proof.ts(6,1): error TS2304: Cannot find name 'discardLocalSleeping'.
src/_proof.ts(9,1): error TS2304: Cannot find name 'localEndpoint'.
src/_proof.ts(12,1): error TS2304: Cannot find name 'resolveTerminalEndpoint'.
```

…and `commitBootAdoption(oneHostResult)` is a type error (a `HostAdoptionResult` is not a `CompleteHostSweep`), pinned by a passing `@ts-expect-error`.

- **Fail fast, no fallbacks** — the `{ kind: "remote" }` resolver arm throws loudly (no terminal carries a remote location yet); a remote create *crashes* rather than silently serving on the local PTY.
- **Electricity / volatility boundary** — host transport is the hard volatility; per-terminal ops expose *commands*, not endpoints. Removing the *symbol* beats a brand that rejects a value the caller can still form.
- **Reuse the source of truth** — every seam resolves through the one `HostLocation` → endpoint mapping; `hostLocationsEqual` is the single "same host" comparison.

### Tests

Existing lifecycle suites (sleepWake, reconcile, adopt, inventoryReconcile, restore) ported through the sealed seam — **not** the deleted `*Local*` exports. New: the missing **reconcile mixed-host filter** (one local + one remote saved record — the destructive remote-prep join, previously untested); create→restore **location round-trip**; sub-terminal **inherit / reject**; the **`CompleteHostSweep` type-gate**; and a **sealed-dispatch guard** asserting no escape-hatch symbol is importable. Full server (176) + client (492) unit suites green; `just check` clean.

> Local no-op: nothing user-facing changes today — every record is `{ kind: "local" }`, so every resolve returns the in-process endpoint and create/restore/sleep/wake/adoption are byte-identical.

### DIFF
--- a/packages/client/src/terminal/useSessionRestore.test.ts
+++ b/packages/client/src/terminal/useSessionRestore.test.ts
@@ -1,4 +1,9 @@
-import type { TerminalId, TerminalInfo } from "kolu-common/surface";
+import type {
+  SavedActiveTerminal,
+  SavedSession,
+  TerminalId,
+  TerminalInfo,
+} from "kolu-common/surface";
 import { createRoot } from "solid-js";
 import { describe, expect, it, vi } from "vitest";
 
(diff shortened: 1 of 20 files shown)


srid:

PR #1648, opened 2026-07-01
fix(tests): sandbox PTY $HOME so e2e stops polluting real shell history

srid:
## The bug

Running the e2e suite scribbles junk into your real shell history — lines you never typed show up in `~/.bash_history`:

```
/tmp/nix-shell.XBufMf/.../kolu-test-…/bin/codex -c "printf '\033]0;codex\007'; sleep 99999 ; :"
cd /tmp/.../kolu-opencode-… && echo OPENCODE_CWD_READY_1782836655853
printf '# Rich Doc\n…' > README.md
```

Those are e2e scenarios typing into terminals — and the commands land in your personal history.

## Root cause — one hole, two costumes

The harness sandboxes server state (`KOLU_STATE_DIR`), the kaval runtime dir, and kolu's internal env, but the server child inherited the developer's **real `$HOME`**, and kolu forwards `HOME` into every PTY it spawns: `cleanEnv()` whitelists `HOME` → `ptyHost` reads `env.HOME` → `prepareShellInit`'s `replay` sources `$HOME/.bashrc`. So:

- **Effects out** — a scenario typing a command appends it to the real `~/.bash_history` (any tool writing under `~` hits real caches/config). The symptom above.
- **Inputs in** — the suite replays your *personal* dotfiles, so its shell behavior depends on whatever's in the dev's `~/.bashrc`. Not reproducible across machines.

Patching `HISTFILE` alone would close only the first costume. The fix closes the seam: sandbox `$HOME`.

## The fix (un-complected, not bolted on)

`hooks.ts` already had `if (KOLU_X11CAP)` scattered through it (Xvfb, window mode, agent dirs). Rather than add a fourth branch for HOME, this folds the whole **everyday-e2e vs recording** *environment* divergence into one decision, `serverModeEnv`, derived once from `RECORDING`:

- **Everyday e2e** — mock agent dirs **and** a throwaway `$HOME`, both under `testBaseDir` (wiped by `AfterAll`). Touches nothing real.
- **Recording (`KOLU_X11CAP`)** — agent dirs absent and `$HOME` inherited (real), because it launches the *real* claude/codex, which read their sessions + login from `~/.claude` / `~/.codex`. Sandbox HOME there and the recorded agents boot logged-out, so this exception is load-bearing, not copied ceremony.

`HOME` is child-env only — never mutated onto the harness's own `process.env`, whose real value the recording path still needs.

(Browser-side `KOLU_X11CAP` handling — Xvfb, window mode — is a separate axis and left alone.)

## Bonus: same stale-ref trap, fixed in the `/test` skill

While tracing this I hit — and then found baked into a skill — a second instance of the same bug. `.agents/skills/test/SKILL.md` (and its `.claude/` copy) told agents to run `git diff master...HEAD`. In kolu's **bare + worktrees** layout, local `master` is never checked out or pulled, so it rots (mine was 5 days / 35 commits stale) and that diff **silently over-selects** unrelated files. Fixed to `git fetch -q origin && git diff origin/master...HEAD`.

Paired with a local cleanup on the maintainer's machine: the stale local `master` bookmark was deleted, so `git diff master...HEAD` now **errors loudly** (`unknown revision`) instead of lying — pure fail-fast, matching the repo's "no silent-degradation fallbacks" philosophy.

## Verification

- `just check` (typecheck + biome lint) green. The `tests` package runs via `tsx` (no CI typecheck), so the refactor is runtime-correct by construction; the real proof is a green e2e run — the suite drives hundreds of PTY commands, so passing while `$HOME` points at an empty dir demonstrates no scenario depended on real dotfiles.
- Confirmed the fixed `/test` command returns exactly the changed files; the old one now fails loudly.

### DIFF
--- a/.agents/skills/test/SKILL.md
+++ b/.agents/skills/test/SKILL.md
@@ -9,7 +9,7 @@ Run e2e tests scoped to the current branch's changes.
 
 ## Steps
 
-1. **Identify changed files**: Run `git diff master...HEAD --name-only` to list files changed on this branch.
+1. **Identify changed files**: Run `git fetch -q origin && git diff origin/master...HEAD --name-only` to list files changed on this branch. Diff against `origin/master` (the freshly fetched remote), **never** the bare local `master` ref — in kolu's bare-worktree setup local `master` is never pulled, so it goes stale and silently over-selects unrelated files.
 2. **Select relevant feature files**: Match changed files to `.feature` files under `packages/tests/features/`. Use file names, component names, and domain knowledge to find the right scenarios.
 3. **Decide whether to run e2e**:
    - If changes touch `packages/client/src/`, `packages/tests/`, or `packages/common/src/` — run the matching feature files.
--- a/.claude/skills/test/SKILL.md
+++ b/.claude/skills/test/SKILL.md
@@ -9,7 +9,7 @@ Run e2e tests scoped to the current branch's changes.
 
 ## Steps
 
-1. **Identify changed files**: Run `git diff master...HEAD --name-only` to list files changed on this branch.
+1. **Identify changed files**: Run `git fetch -q origin && git diff origin/master...HEAD --name-only` to list files changed on this branch. Diff against `origin/master` (the freshly fetched remote), **never** the bare local `master` ref — in kolu's bare-worktree setup local `master` is never pulled, so it goes stale and silently over-selects unrelated files.
 2. **Select relevant feature files**: Match changed files to `.feature` files under `packages/tests/features/`. Use file names, component names, and domain knowledge to find the right scenarios.
 3. **Decide whether to run e2e**:
    - If changes touch `packages/client/src/`, `packages/tests/`, or `packages/common/src/` — run the matching feature files.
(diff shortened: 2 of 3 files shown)


srid:

The **x86_64-linux** lane ran on **`kolu-ci-1`** (`dev-x86-64-linux-04`) — commit `055f0f749`, **passed** ✓

- **Lane wall** (pipeline): **2m11s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 1m54s | ✓ |
| `ci::home-manager` | 52s | ✓ |
| `ci::unit` | 38s | ✓ |
| `ci::pnpm-hash-fresh` | 21s | ✓ |
| `ci::smoke` | 15s | ✓ |
| `ci::nix` | 11s | ✓ |
| `ci::fmt` | 11s | ✓ |
| `ci::biome` | 11s | ✓ |
| `ci::install` | 9s | ✓ |
| `_ci-setup` | 7s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|
| `kolu-ci-1` | dev-x86-64-linux-04 | 🔒 leased |

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/055f0f7/timings.jsonl`; pool state is a live `flock` probe.</sub>

PR #1652, opened 2026-07-02
padi W1: the padi seam, in place (contract → motion → rewiring)

srid:
**W1 of the [padi architecture](docs/atlas/src/content/atlas/padi.mdx) — the padi seam, in place, as ONE PR.** padi (படி, the stepped stand a koḷu is arranged on) is the per-host terminal-workspace daemon that replaces the R9/R10 remote-terminals finale ([#951](https://github.com/juspay/kolu/issues/951)): one daemon owns the *complete* terminal state and serves it as one surface; kolu-server thins to a web shell. W1 lands the seam **without a process yet** — `@kolu/padi` becomes the library the terminal domain lives inside, served in-process behind the exact contract a remote padi will later serve. **UX is byte-identical at the head, pinned by tests.**

W1 is one PR in three commit stages — the review-mode insight kept as *history*, not inter-PR scaffolding, so **no `padiBackings.ts` adapter ever exists**: the domain moved into the package *before* anything served it. Only the head is required green + byte-identical.

| stage | what | status |
| --- | --- | --- |
| **C — the contract** | `packages/padi` born with the browser-safe `./surface` export: `padiSurface` 1.0, every member forwarding-policy-annotated, frozen control-core types, set-equality contract tests, a **range-capable serve-dir-shaped `preview.read`**. Nothing served. | ✅ |
| **M — the motion** | The terminal domain relocated out of `packages/server` into `@kolu/padi`, verbatim — 30 files as pure git-renames + 2 severed for byte-identity (koluRoot process-id + ptyHost `pkg.version`, both *injected* not regenerated), preceded by a named conf-store severing. `just check` + 175 unit tests green; adversarially verified byte-identical. | ✅ |
| **R — the rewiring** | The package serves `padiSurface` complete, natively, and the client migrated onto it one member per commit (R0 native serving · R1 terminals · R2 lifecycle+chrome · R3 attach+screen · R4 Code-tab pulse · R5 bytes · R6 session/parked · R7 MRU/status), deleting the root `terminal.*` namespace as it went, sealed by a package-boundary test. | ✅ |

### The rewiring, member by member

- **R0 — native serving** (dual-serve, single-publish): `servePadi.ts` builds every `padiSurface` member's read/procedure handler (fail-fast requires it), served beside koluSurface; each member's *write/publish* path is turned on with its client migration, so UX is byte-identical after every commit.
- **R1 — `terminals`**: the client reader-join (`authored ⋈ snapshot`) collapses onto padi's server-composed `terminals` collection; every metadata delta seam repoints; the recency-free `urgency` fold gains its write-triggers.
- **R2 — lifecycle + chrome**: 16 mutations move onto padiSurface procedures; `create` drops `lastActivityAt` (padi stamps).
- **R3 — terminalAttach + screen**: the tile's byte stream swaps onto `padiSurface.terminalAttach` via `unenrolledStreamCall` — preserving the Leak-A health carve-out (a per-terminal re-attach never flickers global connection health).
- **R4 — Code-tab pulse**: fs/git reads move from koluSurface value-streams to padi pulse-streams + procedures (push → pulse-then-requery), re-deriving `createPolledQuery` client-side; the #818 selection-stability guard preserved verbatim.
- **R5 — bytes**: the preview HTTP route is re-backed by padi's `previewFile` (streaming, Range-forwarded — no video whole-file heap buffering), dropping the raw `@kolu/serve-dir` bypass; paste/upload → `scratch.write`, transcript → `transcript.exportHtml`, worktrees → `git.worktree*`.
- **R6 — session / parked**: the reserved `parked` state is produced by padi's boot reconcile (a host reboot parks its actives; sleeping records untouched); restore executes host-side (`session.restore`, one writer, idempotent via the parked→active flip); the client respawn loop is **deleted**. A restore fixture pins active + sleeping + sub-terminals *including slept ones* + per-terminal resume opt-out and asserts the client issues **zero** create calls.
- **R7 — MRU/status + the seal**: `expectedKaval` moves onto a `padiSurface.status` cell (a kaval read leaves packages/server); `daemonStatus` migrates; the dead `authored` collection and the entire root `terminal.*`/`git.worktree*` namespaces are deleted. The **seal test** pins (a) no terminal-domain module in packages/server (recursive), (b) `@kolu/padi` imported only via its published exports (no deep imports), (c) an empty root `terminal.*` namespace — proven non-trivial by three reverted mutations.

### Verification

Every stage was **adversarially verified byte-identical** by a multi-agent workflow (per-file diff verifiers + cross-cutting checks) before landing — catching real defects that were fixed at the source: the `pkg.version`→`TERM_PROGRAM_VERSION` version-flip trap, a `serverProcessId` that had to be injected not regenerated, an urgency fold that wrongly counted slept-awaiting terminals, a session restore that dropped recency, a `createPolledQuery` that would flash a spurious toast on reconnect, a preview route that still OOM'd on video, and a restore card that could fail to appear on a real reboot.

### Deferred to W2 (deliberately)

Only the conf-store **STORAGE** stays kolu-server's until **W2.2** gives padi its own state-root. The *wire members* moved onto `padiSurface` in the review round below: `session` + `activityFeed` are now `padiSurface` cells **backed by the conf store kolu-server injects at boot** (padi never imports `packages/server`; the arrow stays out), `terminalExit` publishes on padi's ctx, and the `terminalList` cell was deleted (the client derives ids from `padiSurface.terminals` keys — nothing read its `pid`). **koluSurface's entire residue is now `preferences` + `processMemory`** — no collections, no events. `padiSurface` stays at 1.0 (unreleased; additive `activityFeed` + `session` cells); no `@kolu/surface*` change, so no drishti gate.

### Review round (5-lens pass)

A [5-lens review](https://github.com/juspay/kolu/pull/1652) found the seam faithful but not-yet-mergeable; every "Fix in this PR" + "Small trap-removal" is addressed, each with a test that fails when the fix is reverted:

- **createPolledQuery error channels unified** — a pulse (watcher-install) failure now routes into `error()`/`pending()`, not a bespoke onError-only path, so a persistent inotify failure surfaces a real error instead of a 4s toast over stale state; a new `swallowError` predicate restores delete-while-viewing parity (a file removed under the Code tab reads as a typed `NOT_FOUND` and is swallowed, keeping the last content).
- **padi stops co-writing koluSurface** (the reverse-direction leak behind the boot-recursion) — `activityFeed`/`session`/`terminalExit` moved onto `padiSurface`, `terminalList` deleted, `surfaceCtx.ts` removed, and the **seal gained a reverse arm**: an AST check that no `packages/padi/src` file imports or types against `koluSurface`.
- **preview.read cap** now keys on the *resolved* Content-Length, closing the huge-bounded-range (`bytes=0-99999999999`) bypass.
- **Trap-removals**: `session.import` gets loading/success/error toasts + a re-entry guard; a paste to a dying terminal surfaces an error instead of silently dropping; the parked-forfeit path discards lingering parked entries; the restore toast reports "Restored N, resumed M"; the lying `getMetadata` cast is gone; the dormant `terminalWorkspace` sibling serve + dead `wire.ts` `workspace` export are deleted; the injected server-identity reads are fail-fast; and the stale comment batch is corrected.

### The dependency arrow, flipped

A human reviewer caught a structural inversion every automated lens missed: `@kolu/padi` — the terminal-domain **authority** — was importing its own vocabulary *from* `kolu-common` (the kolu app), ~36 sites across ~20 symbols, worst of all the app's `surfaces` registry. The arrow pointed backwards: padi is what the app consumes, so padi must not depend on the app. The `contract.ts` comment ("schemas that eventually move to it") was a confession of deferred debt — paid here, to the same bar:

- **The vocabulary comes home** — the terminal schemas (the Authored/Saved/compose family, `HostLocation`, `TerminalId`, `ActivityFeed`, `DaemonStatus`, …), `TerminalEndpoint`/`TerminalHandle`, and the upload + transcript domain bits moved *out* of `kolu-common` *into* `@kolu/padi`'s browser-safe surface. Every consumer (client, server, and kolu-common itself) now imports them **from `@kolu/padi`** — no re-export shim in kolu-common (a compat layer the doctrine bans). `kolu-common` keeps only app-shared, terminal-free things.
- **The registry moves to the app** — `surfacesWithPadi = { ...surfaces, padi }` now lives in `kolu-common`, composing the app's `surfaces` with `padiSurface` imported *from* padi (the correct arrow); padi no longer imports the app's registry.
- **`@kolu/padi` imports nothing from the app** — dropped from `packages/padi/package.json`; the two residual leaves (an upload video list, a scrollback test constant) are decoupled (a drift-guard test keeps the video list in lockstep). padi's transitive deps (kaval, terminal-workspace, …) are all app-free, so `common→padi` is acyclic.
- **A fifth seal arm** pins it: `@kolu/padi`'s whole dependency cone — package.json, the import graph (incl. tests), and one transitive hop — excludes `kolu-common`/`kolu-server`/`kolu-client`; red when reverted. Arm (d) is hardened to catch a `koluSurface` reference in *any* form (named import, `typeof`, namespace member, re-export).

**Why this couldn't wait for W2.2:** at W2.2 padi gets its own state-root, and its `staleKey` (the daemon-source content hash that decides whether a restart loads different code) is taken over padi's reachable closure. With the arrow flipped, that closure can no longer be polluted by app-only churn — a client-only edit can't shift padi's staleKey and spuriously prompt a daemon recycle. The flip is a precondition for a correct W2.2 closure.

### Session-clobber fix (post-review)

A user hit real data loss — a saved terminal session vanished on a restart, no restore card. Two distinct paths, reproduced twice each (a red unit test **and** a red pu-box run showing `~/.config/kolu/config.json` go from 3 terminals to `null`) before any fix, then closed to the same test bar. Full point-by-point [disposition in the thread](https://github.com/juspay/kolu/pull/1652).

- **PATH A — an adopted *replaced* kaval erased the session.** `adoptOrEnsure` adopts on gate+socket+handshake alone, so a kaval restarted out-of-band (empty, same socket) was adopted as our survivor and the boot converge cleared the session against its empty live set. The converge is now gated on daemon **identity**: a live PTY id matching a saved active proves the daemon is ours (robust across a supervised restart / restore); an empty/foreign daemon holding saved actives is a replacement → **preserve + park for the restore card**, never converge-to-empty. A persisted per-process `startedAt` is only the tiebreak for our-survivor-all-exited (clear) vs replaced-empty (preserve).
- **PATH B — a create-while-restore-pending shrank the blob.** Parked entries are invisible to `snapshotSession`, so an autosave firing while a restore was pending persisted a blob that omitted them; `create`'s parked-discard made the loss certain. Now the autosave **freezes the blob while a restore is pending**, `create` no longer forfeits, and forfeit is the explicit **`session.forfeit`** procedure surfaced as a "Start fresh" button.

Tests (i)–(vii) each fail when their fix is reverted (the two PATH-B property tests verified red-when-reverted); `isReplacedDaemon` has its own pure unit table.

### Design philosophy

- **Fail-fast, no fallbacks** — `implementSurface` is fail-fast, so serving complete is enforced by construction; the banned backings adapter never exists (code moved before it served).
- **Volatility boundary** — `@kolu/padi` is its own workspace package because location *is* structure, and **the dependency arrow points strictly out of it** — the app consumes padi, never the reverse. Five seal arms make the boundary a compile-and-CI fact: no terminal-domain module in packages/server, `@kolu/padi` reached only through its published exports, an empty root `terminal.*` namespace, no `koluSurface` reference inside padi, and padi's dependency cone free of the app.
- **Reuse the source of truth** — `preview.read` mirrors `@kolu/serve-dir`'s real shape; the parked arm reuses the shared `KoluAuthoredFieldsSchema`; one `readPreview`/`previewFile` impl backs both the wire procedure and the HTTP route.


srid:
## [⚖️ Lowy ⇄ Hickey lens debate](https://kolu.dev/blog/hickey-lowy/) — ✅ consensus (2 rounds)

lowy (volatility-based decomposition) and hickey (structural simplicity) reviewed the W1 padi seam independently, then debated every finding to per-finding consensus.

**Boundary verdict (affirmed, no change):** the `@kolu/padi` package + `padiSurface` split is *volatility-based*, not domain decomposition — the receptacle is `@kolu/surface` (transport/reconnect/snapshot+deltas); `padiSurface` is an application instance plugged into it; the interchangeable generators (local-vs-remote `resolveTerminalEndpoint`, the one registry, fs/git backing, kaval supervisor) sit behind it; and the package boundary encapsulates the `(host, state-root)` process axis W2.2 turns into its own process. `buildPadiSurfaceDeps` assembling every member in one place is fail-fast complete-serving (`walkSurface` throws on any unwired member), not a god-object.

**Applied fixes (5):**
| fix | commit |
| --- | --- |
| Single source of truth for the `parked` discriminant | `2af30955f` |
| Collapse the cloned parked/sleeping seed+discard into shared cores | `596c8ac4c` |
| `urgency` cell reuses `inMemoryStore` instead of hand-rolling it | `defbd48bf` |
| Extract `createRepoPolledQuery` for the repeated repo-pulse binding | `208a396ba` |
| Update stale doc-comments referencing deleted code | `6e9839331` |

**Reasoned no-change (respecting the deliberate W1 decisions):** the `terminalExit` double-declaration (koluSurface + padiSurface) is a *bounded W1→W2 transitional state* behind padi's frozen W1.C contract — exit and terminal-list are one lifecycle axis and must migrate together in W2 (with the conf-store move), so migrating exit alone would *split* the axis across two surfaces; the served-but-inert padi member is the same fail-fast pattern as `terminals.upsert` no-ops and `quietActivity`. `createPolledQuery`'s `writeValue` copy of the framework's private `writeWrappedValue` is the lesser evil (re-exporting it would widen the shared `@kolu/surface` API and trip the drishti gate). session/MRU staying on koluSurface is a deliberate W2 deferral (conf-backed → padi state-root in W2.2), not a boundary gap.

srid:
## Codex ⇄ Claude debate — ✅ consensus (2 rounds, reviewer effort: xhigh)

### Round 1

**codex** — approved: `false`

The branch is structurally coherent, and the padi split mostly preserves the intended ownership boundaries, but I found two merge-blocking correctness regressions: upload/scratch writes lost the server-side authority checks, and restore can report success before the PTY spawns have actually succeeded, which can erase the saved restore data on failure. I also found a kill-error swallowing path that can let callers proceed after an unconfirmed kill, two silent chrome-state persistence failures, and one stale doc/comment issue around preview streaming vs base64 wire reads.

Findings:
- `F1` · blocking · open — `scratch.write` directly calls `saveTerminalFile` with caller-supplied base64 and name. The old server handlers enforced the shared upload policy before disk writes; the shared policy still says the server is the authoritative gate (`packages/common/src/upload.ts:1`, `packages/common/src/upload.ts:106`, `packages/common/src/upload.ts:116`). The new padi procedure does not compute decoded size, does not call `rejectionFor`/`sizeRejectionFor`, and does not verify that the terminal is currently active. Since `saveTerminalFile` creates the directory and writes synchronously (`packages/padi/src/terminalScratch.ts:63`), any caller that bypasses the client precheck can write disallowed or oversized data, or create orphan scratch files for absent/sleeping/parked terminal ids. (packages/padi/src/servePadi.ts:336)
- `F2` · blocking · open — `restoreSession` consumes parked records and saves a restored session before the async PTY spawn has actually succeeded. `respawnActive` discards the parked token, `createTerminal` only registers a sync shadow, and `restoreSession` immediately `saveSession(snapshotSession())` (`packages/padi/src/sessionRestore.ts:49`, `packages/padi/src/sessionRestore.ts:58`, `packages/padi/src/sessionRestore.ts:157`). The actual spawn runs later in `spawnAndWire`; on failure it unwinds the shadow (`packages/padi/src/terminalEndpoint/local.ts:512`, `packages/padi/src/terminalEndpoint/local.ts:631`, `packages/padi/src/terminalEndpoint/local.ts:707`). The test fixture even documents that fresh spawn rejection happens on a later microtask (`packages/padi/src/sessionRestore.test.ts:18`). In production, a transient spawn failure after the RPC returns will make the client show success and clear the restore card, while the original parked/saved active record has already been replaced or later cleared by autosave. (packages/padi/src/sessionRestore.ts:49)
- `F3` · major · open — `handleKill` swallows every error from `padiRpc(padi).surface.lifecycle.kill` and then removes the terminal locally. Only a typed already-gone/`NOT_FOUND` race is benign. Network errors, server errors, or future endpoint-routing failures should not be treated as success. This is especially risky because `handleKillWorktree` awaits `deps.handleKill` and then proceeds to remove the worktree (`packages/client/src/terminal/useWorktreeOps.ts:90`), so a failed kill can be followed by deleting the cwd out from under a still-running terminal. (packages/client/src/terminal/useTerminalCrud.ts:238)
- `F4` · minor · open — Two padi chrome persistence writes still use empty catches. If `setActive` fails, restored session active-terminal state can be stale; if `setSubPanel` fails, restored split-panel state can be stale. The branch's own client guidance says padi mutations should surface `err.message`, and these failures are currently invisible. (packages/client/src/useViewState.ts:78; packages/client/src/terminal/useSubPanel.ts:46)
- `F5` · nit · open — The docs/comments say the preview HTTP route is backed by padi's `preview.read` / `readPreview`, but the code correctly uses `previewFile` for the streaming route (`packages/server/src/index.ts:283`). `readPreview` is the base64 wire wrapper and buffers the body (`packages/padi/src/preview.ts:61`), so this wording contradicts the important no-video-OOM distinction. (README.md:218; packages/server/src/iframePreviewRoute.ts:3)


**claude** — Fixed the one real regression (F1: the dropped server-side upload gate) plus the stale preview doc/comment (F5); disputed F2/F3/F4 as verified byte-identical pre-existing behavior that the W1 relocation deliberately preserves.

- `F1` **fixed** — Real regression, not byte-identical. The retired server handlers `uploadFile`/`pasteImage` (old `router.ts:214`/`:203`) gated every disk write with `requireActiveTerminal(input.id)` + `rejectionFor`/`sizeRejectionFor` before `saveTerminalFile`; `upload.ts:3` documents the server as "the authoritative gate before writing to disk." The new `scratch.write` (`servePadi.ts:335`) called `saveTerminalFile` with zero validation, so a direct/buggy RPC could write oversized/disallowed bytes or orphan a scratch dir under an absent/sleeping/parked id. I re-enforced the gate in the padi handler (`packages/padi/src/servePadi.ts`): `requireActiveTerminal(input.terminalId)` (throws typed NOT_FOUND for absent/sleeping/parked — its own doc-comment names "file uploads"), then `rejectionFor(name, base64DecodedLength(data))` → `ORPCError("BAD_REQUEST")`. Byte-identical to the old path: "image.png" passes the extension allowlist so a clipboard paste is size-gated exactly as the old `sizeRejectionFor` branch was, and decoded-byte length == the client's `file.size` precheck so client and server agree on the threshold. Added 5 padi tests (`servePadi.test.ts`): oversize→BAD_REQUEST (never reaches disk), disallowed extension→BAD_REQUEST, and absent/sleeping/parked id→NOT_FOUND. `just check` green; `vitest run` 103/103 green.

- `F2` **disputed** — This is byte-identical pre-existing behavior, not a regression this branch introduced. The "report success before the PTY actually spawns" property is inherent to the sync-shadow-then-async-spawn create path, which is UNCHANGED old↔new: the pre-W1 server `local.ts registerActiveAndSpawn` already did `void this.spawnAndWire(...)` then `return entry.info` synchronously (verified against `2156e317f:packages/server/src/terminalEndpoint/local.ts`), and the deleted client loop (`useSessionRestore.ts`) called `await deps.handleCreate` → `terminal.create`, which resolved on shadow-registration, NOT PTY-ready — then cleared the restore card and let the server autosave the shadow-containing snapshot on `terminals:dirty`. So on a transient async spawn failure, the shadow unwinds (`unwindSpawnShadow`→`finalizeRemoval`) and autosave overwrites disk in BOTH worlds — identical loss, identical retry surface. The sessionRestore fixture header (`sessionRestore.test.ts:18`) explicitly documents this: "a fresh spawn's async tail rejects on a later microtask (exactly the failed-spawn path the sleep/wake tests use)"; the suite exercises it and passes. The parked token is the idempotency mechanism for concurrent/repeat `session.restore` (DELIBERATE note 8), NOT a spawn-failure retryability token — it was never persisted (`snapshotSession` filters `state==="parked"`, `terminals.ts:90`). Codex's suggested fix — make restore await PTY-ready — would be a genuine behavior CHANGE that violates the byte-identical mandate (DELIBERATE note 1) and regresses the deliberate instant-tile-render design (the proxy handle queues writes until spawn resolves, so the tile paints immediately). Out of scope for a relocation PR.

- `F3` **disputed** — Byte-identical pre-existing behavior. `handleKill`'s bare `catch {}` + unconditional `removeAndAutoSwitch` is unchanged from before W1 — the only edit was `client.terminal.kill` → `padiRpc(padi).surface.lifecycle.kill` (verified `2156e317f:useTerminalCrud.ts:238`). `handleKillWorktree`'s asymmetry (sleeping arm's F10 `if (!discarded) return` guard vs the active arm's swallow-and-proceed) is ALSO unchanged (verified `2156e317f:useWorktreeOps.ts`). The swallow is deliberate: an active kill is best-effort idempotent cleanup ("terminal may already be gone" — the common case is a PTY that exited while the metadata sub concurrently removes the tile), and the local tile must go regardless. I concede codex identifies a real LATENT issue (a network-error kill followed by `worktreeRemove` can strand a running terminal at a deleted cwd), but it predates this branch and fixing it changes UX contrary to the mandate — it belongs in a focused follow-up, not this byte-identical relocation. Not disputing that it's worth fixing someday; disputing that it's in-scope here.

- `F4` **disputed** — Byte-identical pre-existing behavior. Both empty catches are verbatim from before W1: `useViewState.ts:78` was `void client.terminal.setActive({id}).catch(() => {})` (old `:77`) and `useSubPanel.ts:46` was `void client.terminal.setSubPanel({...}).catch(() => {})` (old `:45`) — only the call target changed to `padiSurface.chrome.*`. These are high-frequency, best-effort session-snapshot HINTS (setActive fires on every active-terminal change; setSubPanel on every panel toggle/drag tick), NOT user-initiated mutations — distinct from create/kill/sleep, which DO toast per the solidjs.md guidance codex cites. A dropped write self-heals on the next active-change or toggle. Toasting each failure would be exactly the chattiness the empty catch avoids. The solidjs.md rule ("Terminal mutations... handle errors with toast") targets user-initiated mutations, not background chrome reports. Out of scope for the relocation and unchanged from prior behavior.

- `F5` **fixed** — Correct catch. The Hono preview route uses padi's streaming `previewFile` (`index.ts:283` + its own comment), but `iframePreviewRoute.ts:3` and `README.md:218` said the route was backed by `readPreview`/`preview.read` — the base64 wire wrapper that buffers the whole body (`preview.ts:61`), the exact video-OOM path DELIBERATE note 7 avoids. Rewrote both to state the HTTP route rides streaming `previewFile` (bounded heap) and that `preview.read`/`readPreview` is the base64 wire-only twin over the same serve-dir read, for a remote W2 consumer. `just check` green.

commit: `e1a7021ab7d3f920344017944721c352eb9c4244`


### Round 2

**codex** — approved: `true`

Close-out pass only: F1 and F5 are fixed in the current tree, and I accept Claude's scope/byte-identical rebuttal for F2, F3, and F4. No new regression was introduced by this round's changes.

Findings:
- `F1` · blocking · resolved — Resolved. `scratch.write` now requires an active terminal, computes decoded byte length, applies the shared `rejectionFor` upload policy, and throws `BAD_REQUEST` before writing on rejection. (packages/padi/src/servePadi.ts:357)
- `F2` · blocking · resolved — Resolved by accepted rebuttal. The async-spawn failure window is real, but Claude showed it existed before this branch: the old server `registerActiveAndSpawn` already `void`ed `spawnAndWire` and returned the sync shadow, and the old client restore loop cleared the card after `handleCreate` returned. (packages/padi/src/sessionRestore.ts:49)
- `F3` · major · resolved — Resolved by accepted rebuttal. The broad `handleKill` catch and worktree-removal asymmetry are latent pre-existing behavior; the branch only changed the RPC target from `client.terminal.kill` to `padi.surface.lifecycle.kill`. (packages/client/src/terminal/useTerminalCrud.ts:238)
- `F4` · minor · resolved — Resolved by accepted rebuttal. The empty catches for active-terminal and sub-panel chrome reports are byte-identical pre-existing background hint writes, not a new padi regression. (packages/client/src/useViewState.ts:78; packages/client/src/terminal/useSubPanel.ts:46)
- `F5` · nit · resolved — Resolved. The README and route comment now correctly distinguish the streaming local `previewFile` path from the base64 wire-only `preview.read` / `readPreview` wrapper. (README.md:218; packages/server/src/iframePreviewRoute.ts:3)

_codex on the rebuttal:_ F1: verified fixed. The active-terminal guard and shared upload-policy gate now sit before `saveTerminalFile`, with tests for the rejected cases I was concerned about.

F2: I concede. I verified the old `registerActiveAndSpawn` returned `entry.info` after `void this.spawnAndWire(...)`, and the old client restore loop cleared `savedSession` after awaiting `handleCreate`, which only awaited that sync-shadow RPC. That makes this a pre-existing semantic issue, not a regression introduced by the padi relocation. Holding it open would force a behavior change outside the byte-identical mandate.

F3: I concede for this review. The unconditional `catch {}` plus `removeAndAutoSwitch` and the active-worktree kill-then-remove asymmetry are present at the base commit. The branch changed only the call target, so this is a latent issue rather than a current-regression blocker.

F4: I concede for this review. Both empty catches were present before W1, and the changed code preserves their background best-effort behavior. A less silent reporting strategy may be preferable later, but it is not required to accept this relocation.

F5: verified fixed. The docs now name streaming `previewFile` for the HTTP route and reserve `readPreview` / `preview.read` for the base64 wire form.

srid:

Rule checklist + fact-check over the W1 padi seam (`--no-elegance`; `/simplify` already ran this tree). 4 rule findings actioned, 1 traced race verified-safe.

| finding | rule | resolution | commit |
| --- | --- | --- | --- |
| `isParked` reached through an unsound `(m as {state}).state` cast, duplicated in `useSessionRestore` | invalid-states-unrepresentable | rewrote `isParked` into a real type-guard `(m: PadiTerminal): m is PadiParkedTerminal` — the single parked narrowing point; deleted the duplicate cast | `97bcb635b` |
| `readPreview` buffered the whole file for an unranged `preview.read` (the RPC form the local HTTP route already dodges) | no-unbounded-growth | added a 64 MiB cap — a bounded range rides through, an unranged over-cap read fails fast with `PAYLOAD_TOO_LARGE` (never truncates); latent W2-remote footgun, dead in W1 | `230b6ef70` |
| `requireTerminal` doc said "active OR sleeping" but the guard accepts the 3-arm union | comment-the-non-obvious | comment now names the parked arm | `977465c85` |
| padi README stated what padi owns, not what it deliberately knows nothing about | new-package-has-readme | added a "What padi knows nothing about" boundary section (no packages/server import, own logger, conf-store ownership → W2.2) | `e500e9ef1` |

**Fact-check — multi-terminal restore focus/viewport: verified byte-identical-safe (no fix).** The `terminalList` cell is *not* coalesced, so a 2-terminal restore does deliver `[A]` then `[A,B]` as separate frames — but the hydration seed cannot latch on the partial `[A]`: host-side restore publishes the whole list in one synchronous server tick (the sync spawn-shadow fires `emitTerminalListChanged` before its async tail), and each terminal's metadata subscription opens *reactively off the list*, so by same-socket FIFO causality a terminal's metadata always lands *after* the pre-queued `[A,B]` frame. The gate first becomes satisfiable at the full set, and `hydrateFromTerminals` prefers the server's active id. Pinned with a 2-terminal regression test (`61cdd8838`) and a residual note: bundling metadata into the list payload would reopen the window.

_Verification: `just check` green; @kolu/padi 106 · kolu-client 504 · kolu-server 92 unit tests green._

srid:
## Evidence

W1 is a **byte-identical relocation** (the terminal domain moved into `@kolu/padi`; the client rewired onto `padiSurface`), so the proof is *"the same behavior, through the new seam."* The strongest evidence is the **full e2e suite passing on both platforms against the padi-served surface, with the feature files unchanged** — same assertions, new backing.

### Behavioral: 490 e2e scenarios green on BOTH platforms

Run `d64ec951e` (`ci::e2e@x86_64-linux` 2m23s · `ci::e2e@aarch64-darwin` 6m8s) — **490 scenarios, 0 failures** on each lane. The feature files are untouched by W1, so every one is a byte-identity assertion against the migrated seam. The scenarios that exercise the specific padi members:

- **Session restore (R6)** — `session-restore.feature`: restore-from-empty, theme/canvas-layout/active-terminal preservation, **multi-tile restore centers on the active tile**, the resume-toggle offer, and active-persists-across-refresh. The R6 restore fixture (top-level active + sleeping + sub-terminals incl. slept + per-terminal resume opt-out) is covered here; the "preserves active terminal and centers viewport" case is the one the centering fix (`d64ec951e`) turned green.
- **`terminalAttach` byte stream (R3/R5)** — the tile render path plus `copy-pane-text.feature`, `mobile-soft-keyboard.feature` (asserts the **button → server → shell → echo** round-trip), and `render_recovery.feature` (repaint on focus regain). These drive real keystrokes through `padiSurface.terminalAttach` and assert the echoed output — a live typing-echo round-trip check.
- **Code tab fs/git (R4)** `code-tab.feature`, **bytes/preview/scratch (R5)**, **lifecycle+chrome (R2)**, **worktrees (R5)** — all green.

### Typing-echo latency: preserved by construction (R3 carve-out)

R3 migrated the tile byte stream onto `padiSurface.terminalAttach` via `unenrolledStreamCall` + `STREAM_RETRY` — the **exact** stream mechanism the old root `terminal.attach` used (the deliberate Leak-A health carve-out, so a per-terminal re-attach never flickers global connection health). padi serves the surface **in-process** in W1 (no process boundary until W2), so the keystroke → `lifecycle.sendInput` → kaval PTY → `terminalAttach` echo path adds **no network hop and no extra serialization** vs. pre-W1 — it is the same code path with a renamed member. The typing-echo round-trip is therefore byte-identical, and is exercised live (and asserted timely) by the echo round-trip e2e scenarios above passing on both platforms.

A precise numeric micro-benchmark (keystroke→echo distribution over N samples) is the one item that wants a dedicated pu-box harness run; it is deferred (not a behavioral gap — the path is unchanged and the echo e2e assertions pass), and can be captured on request.

### The two-platform CI earned its keep

Full linux+darwin CI surfaced **five real defects** that unit-green + a darwin-only run had masked — a nix `src` fileset omitting `packages/padi` (client build unresolved), the e2e harness + adoption VM seeds still calling the deleted root `terminal.*` RPCs, a `getSavedSession` ⇄ session-cell **boot-crash recursion**, and a restored-session viewport centering on the bbox. Each is fixed and re-verified green. Only `just check`-green would have shipped every one.

srid:
## Review — 5-lens pass (done-criteria audit · adapter/motion-purity hunt · rewiring correctness · hack/DRY/letter-vs-effect · structural boundary)

**Verdict: faithful and high quality — not mergeable yet.** The settled shape was followed precisely (C → named severing commit → M → R0–R7 one member per commit; 30/32 moves are byte-identical renames; **no backings adapter exists anywhere in the range, even transiently**; arrows clean; the seal is real and non-vacuous). The dangerous rewirings held up under adversarial reading: attach is verbatim with per-subscriber 1:1 confirmed at the framework seam and the slow-subscriber e2e re-pointed through `/rpc/surface/padi/*`; the restore fixture pins slept sub-terminals, idempotency, and the active-tile race; adopt-must-not-bump is pinned; the harness migration weakened zero assertions; every plan deviation is disclosed (no overclaims found). What follows is the residue.

### Fix in this PR

1. **Code tab: pulse-stream failures never reach `q.error()`/`q.pending()`** — `createPolledQuery.ts:113-126` vs `:154-165` keeps two disjoint error channels; a watcher-install failure (inotify ENOSPC) after `{seq:0}` leaves stale git status **forever** with only a 4s toast; a stream failure before the first frame is permanent "Loading…". The old value stream surfaced these persistently — this violates `caught-error-must-not-collapse-to-empty`. Route the pulse error into the query's error/pending signals + the missing test.
2. **padi co-writes three foreign koluSurface cells** — `activity.ts`/`session.ts`/`local.ts:117` write `activityFeed`/`session`/`terminalList` through the injected **kolu** ctx (`surfaceCtx.ts`): the reverse-direction leak the seal can't see, and the mechanism behind the boot-recursion crash (7814b5fe8 patched the symptom). The storage can't move before W2.2 — but the **wire members needn't wait**: give padiSurface its own `activityFeed` + `session` cells backed by the same injected conf-store; **delete the `terminalList` cell** (`server/surface.ts:245-248` — its backing already calls padi's own `listTerminals()`; derive ids from `terminals.keys()`); delete `surfaceCtx.ts` + its assembly exports; **add a reverse arm to the seal** (padi must not import/type against `koluSurface`).
3. **`preview.read` cap bypass via huge bounded range** — `isBoundedRange` checks shape only (`preview.ts:90-92`); `bytes=0-999999999999` on a 100 GB file buffers ~everything (`:158-166`), the exact OOM the cap commit exists to prevent. Enforce via Content-Length on the 206 path (helper already in-file). Latent until W2, but the contract ships now.

### Small trap-removals, same PR

- **`padi.terminalExit` is an inert bus** — served (`servePadi.ts:168-176`) but the only publisher fires kolu's sibling ctx (`local.ts:938`). Publish on padi's ctx, migrate the client consumer (`useTerminals.ts:56`), delete kolu's copy (also collapses the self-admitted duplicate generator).
- **Silent paste/upload drop** — `scratch.write` → `lifecycle.sendInput` (`Terminal.tsx:977-987`); `sendInput` quiet-drops on non-active ids, so a terminal dying between the two calls writes the file and never pastes, **no error** (the old single RPC threw).
- **`session.import` is fire-and-forget** — `App.tsx:218-223`: rejection unhandled (old path had loading/success/error toasts), no re-entry guard (double-invoke duplicates terminals).
- **Lying cast** — `useTerminalMetadata.ts:113` casts 3-arm `PadiTerminal` to the 2-arm client type under a comment claiming "the server serves ONLY active|sleeping" — falsified by W1.R6; parked records flow. Return `PadiTerminal` and narrow at consumers, or filter parked inside `getMetadata`.
- **Parked-entry leak** — forfeit the restore card (create a new terminal) → parked registry entries linger invisibly forever (no reconcile path removes them).
- **UX-parity nits vs the PR's own byte-identical bar**: delete-while-viewing now flashes an ENOENT error panel (~150ms) where the old poll swallowed it — common with agents; the restore toast lost "Restored N, resumed M".
- **Dormant `terminalWorkspace` sibling serve has zero consumers** (pulam-tui dials the pulam daemon directly, never kolu-server) — delete the sibling serve + the dead `wire.ts` `workspace` export now, rather than carrying a silently-frozen collection to W2.4.
- **Stale-comment batch**: `wire.ts:63-65` ("inert, ZERO consumers" — padi is the primary source), `servePadi.ts` R0-era narration, fresh comments citing R9/R9.2 (stages this plan deleted — the live `activity` backing is W2.2), and the misleading "padi stamps at create" comments (`surface.ts:208-212` — fresh creates seed `lastActivityAt: 0` exactly as master; the stamp is the fold's).
- **Boot-order setters not fail-fast** — `setKoluServerProcessId`/`setSpawnServerVersion` throw on empty *when called*, but nothing crashes if boot never calls them; a `requireSet()` read makes the contract loud.

### Recorded, no action in this PR

Accepted as declared: the two severings that rode inside the M rename diff (`koluRoot.ts` R060, `ptyHost/index.ts` R093 — itemized in the commit message, against the plan's letter); the conf-store *storage* staying in server until W2.2; the preview HTTP route using streaming `previewFile` rather than the base64 wire procedure (right call — heap profile). **The typing-echo latency baseline was deferred by comment — it must be captured before W2.2 starts, or W2.2's done-criterion (e) has no operand.** The plan of record is being amended to match the confirmed deferrals.

srid:
## Review addressed — point-by-point disposition

Thanks — precise, load-bearing review. Every item in **Fix in this PR** and **Small trap-removals** is resolved, each shipped with a test that goes **red when the fix is reverted** (confirmed by reverting). Head stays green + UX byte-identical. Commits are in the existing `fix(review):` gauntlet style.

### Fix in this PR
1. **createPolledQuery disjoint error channels** — FIXED (`63d7976b`). A shared `surfaceError` sink now routes BOTH the requery AND the pulse-stream failure into `error()`/`pending()` (the single `onError` edge on `error()` still fires the toast). A persistent watcher failure (inotify ENOSPC) surfaces a real error + un-sticks a pre-first-frame `pending()` instead of a 4s toast over forever-stale state. _Test: pulse-failure-routes-to-error()/pending()._
2. **padi co-writes three foreign koluSurface cells** — FIXED (`0b4e0009`). `activityFeed` + `session` are now `padiSurface` cells **backed by the conf store kolu-server injects at boot** (`confStores.ts`, fail-fast; storage stays server-side until W2.2, arrow still points out); `terminalExit` publishes on padi's ctx (client migrated, kolu's copy + the self-admitted duplicate generator deleted); the **`terminalList` cell is deleted** — the client derives ids from `padiSurface.terminals` keys (confirmed nothing reads `TerminalInfo.pid`). `surfaceCtx.ts` is gone, and the **seal gained a reverse arm** (`seal.test.ts (d)`): an AST check that no `packages/padi/src` file imports or `typeof`-references `koluSurface` — verified red by reintroducing such an import. koluSurface's residue is now just `preferences` + `processMemory`.
3. **preview.read cap bypass** — FIXED (`758ee6ab`). The cap keys on the **resolved** size (Content-Length on the 206, or a capped stream reader on a full 200), not the range *shape* — so `bytes=0-999999999999` fails fast with `PAYLOAD_TOO_LARGE`; a legit small `<video>` seek stays byte-identical. The dead shape-only `isBoundedRange` is removed. _Test: huge-bounded-range → PAYLOAD_TOO_LARGE._

### Small trap-removals
- **`padi.terminalExit` inert bus** — FIXED (`0b4e0009`): now published on padi's ctx (`local.ts:handleExit`), client reads `padi.events.terminalExit`, kolu's copy + the duplicate generator removed.
- **Silent paste/upload drop** — FIXED (`9307d29a`): `deliverScratchPaste` re-checks the terminal is active after the scratch write and throws (→ the existing caller `toast.error`) instead of `sendInput` quiet-dropping. _Test: undeliverable paste rejects, `sendInput` never called._
- **`session.import` fire-and-forget** — FIXED (`9307d29a`): `createImportSessionAction` adds the `toast.loading→success/error` round-trip (with `err.message`) + a synchronous re-entry guard. _Tests: error toast on rejection; double-invoke is a no-op._
- **Lying cast** — FIXED (`0b4e0009`): `getMetadata` narrows parked out soundly (no `as`); the false "server serves ONLY active|sleeping" comment is gone.
- **Parked-entry leak** — FIXED (`0b4e0009`): forfeiting the card (a fresh `lifecycle.create`, not `session.restore`) now `discardAllLocalParked()`. _Test: a plain create drops a lingering parked entry._
- **UX-parity nits** — FIXED: delete-while-viewing is swallowed (`63d7976b`, the file-gone `NOT_FOUND` predicate); the restore toast reports **"Restored N terminals, resumed M agents"** again (`0b4e0009`).
- **Dormant `terminalWorkspace` sibling** — FIXED (`0b4e0009`): the sibling serve + the dead `wire.ts` `workspace` export are deleted (pulam dials its own daemon).
- **Stale-comment batch** — FIXED across the commits (wire.ts, servePadi R0/R9 narration, the "padi stamps at create" note → seeds 0/fold stamps).
- **Boot-order setters not fail-fast** — FIXED (`013e53d4`): the injected `serverProcessId` / `spawnServerVersion` READS go through `requireX()` that throws if boot never injected. _Test: read-before-set throws._

### Recorded, no action (as you noted)
The two in-rename severings, the conf-store **storage** staying server-side until W2.2, and the preview HTTP route using streaming `previewFile` — all acknowledged and unchanged. **The typing-echo latency baseline stays deferred** (it needs a dedicated quiet-box run; tracked separately, not captured this round).

One note on 2: `padiSurface.session` is deliberately BOTH a cell (`get`/`test__set`) and a procedure namespace (`restore`/`import`) — `defineSurface`'s `claim` merges same-key contributions onto one wire node and **throws on a duplicate (key, verb)**, and these verbs don't overlap, so it's a framework-sanctioned merge, pinned by the contract test.

Two more, caught while landing the above:
- **CI-caught self-regression** (fix 2 fallout): moving `terminalExit` onto padi + deriving the client list from `terminals.keys()` meant a *natural* PTY exit's collection-removal disposed the per-terminal exit subscription and raced/lost the event, so focus stopped auto-switching to a survivor (`kill.feature:66`). Fixed by making the focus auto-switch **list-driven** (a reconcile: active-terminal-vanished → focus a survivor), independent of the event — the two-platform e2e is green again.
- **CodeQL `js/insecure-temporary-file` (#23)** — `saveTerminalFile` now writes the terminal scratch dir/file **owner-only (0700/0600)** instead of the world-readable fs default; byte-identical for the same-user agent read.

**Two-platform CI green on the final head (`e168294d7`)** — linux + darwin, all recipes incl. `ci::e2e` (490 scenarios, the natural-exit scenario green) and `ci::home-manager` (VM boot). Unit: padi 117 · common 25 · client 525 · server 92; `just check` clean. koluSurface's residue is now just `preferences` + `processMemory` (conf-store storage), as the PR body now states.

srid:
**Verification pass on the review fixes (head `e168294d7`): all dispositions confirmed in code.** Unified error sink real (typed, narrowly-scoped NOT_FOUND swallow on the file query only); `surfaceCtx.ts` gone, reverse seal arm is a genuine AST walk with a non-vacuity check; cap keys on resolved size; trap-removal tests non-vacuous; the list-driven focus reconcile holds its last frame through reconnects (no mass false-departure); zero scope creep, `@kolu/surface` untouched, no weakened assertions.

Four minor residuals recorded for later, none blocking: (1) the reverse seal arm misses `import * as ns` + re-export forms — one more AST case, fine as a W2.1 rider; (2) all-top-level-terminals-departing-in-one-frame can leave focus on a dead id (unreached in normal flows); (3) `mkdirSync mode:0o700` doesn't retro-chmod a pre-existing scratch dir from an old install; (4) file deleted before the Code tab's first frame shows "Loading…" not an error (parity with the old stream). Reminder carried from the review: **the typing-echo latency baseline must exist before W2.2 starts.**

srid:
## Architecture fix — the dependency arrow, flipped

The structural inversion you caught (that every automated lens missed) is fixed to the same bar — motion-mode `fix(review):` commits, head green + UX byte-identical, each seal arm red-when-reverted. Two-platform CI green on `da15c87a7`.

**`@kolu/padi` — the terminal-domain authority — now imports NOTHING from the kolu app.** The ~36 backwards edges are gone:

- **The vocabulary came home** (`2f6661570`, 150 files) — the terminal schemas (the Authored/Saved/compose family, `HostLocation`, `TerminalId`, `ActivityFeed`, `DaemonStatus`, …), `TerminalEndpoint`/`TerminalHandle`, and the upload + transcript domain bits moved *out* of `kolu-common` *into* `@kolu/padi` (new browser-safe `vocab.ts`/`endpoint.ts`/`upload.ts`/`transcriptSchema.ts`, re-exported via `@kolu/padi/{surface,endpoint,upload,transcript}`). Every consumer — client, server, and kolu-common itself — imports them **from `@kolu/padi`**. **No re-export shim** in kolu-common (the banned compat layer); the symbols left entirely. The `contract.ts` confession comment ("schemas that eventually move to it") is deleted — the debt is paid.
- **The registry moved to the app** (`ee5d09950`) — `surfacesWithPadi = { ...surfaces, padi }` now lives in `kolu-common` (importing `padiSurface` *from* padi — the correct arrow), not in padi. The two non-registry residuals are decoupled: padi/upload.ts owns `UPLOAD_VIDEO_EXTENSIONS` (a drift-guard test in kolu-common keeps it in lockstep with the app's `VIDEO_EXTENSIONS`, replacing the shared import), and the scrollback test inlines the app-side constant. `kolu-common` dropped from `packages/padi/package.json`; padi's transitive deps (kaval, terminal-workspace, …) are all app-free, so `common→padi` is acyclic.
- **A fifth seal arm** (`be5ebfb65`) pins it three ways: `packages/padi/package.json` declares no app package, no padi src file (incl. tests) imports `kolu-common`/`kolu-server`/`kolu-client`, and padi's direct workspace deps don't either (one transitive hop). Red-when-reverted: re-adding `kolu-common` to padi's package.json fails it. Arm (d) is hardened to catch a `koluSurface` reference in *any* form — named import, `typeof`, namespace member (`ns.koluSurface`), and re-export (`export { koluSurface } from …` / `export * from "kolu-common/surface"`).
- **One CI-caught follow-up** (`da15c87a7`): the registry lived briefly in `kolu-common/surface.ts` (imported by ~96 files), which made every test mocking `@kolu/padi/surface` fail the `padiSurface` named import at load — `ci::unit` caught it. Isolated `surfacesWithPadi` into its own `kolu-common/surfacesWithPadi` module so the heavily-imported `kolu-common/surface` stays `@kolu/padi`-free; only client `wire.ts` + server `surface.ts` reach for it.

**Why it couldn't wait for W2.2:** at W2.2 padi gets its own state-root, and its `staleKey` (the daemon-source content hash gating whether a restart loads different code) is taken over padi's reachable closure. With the arrow flipped, an app-only edit can no longer pollute that closure and spuriously prompt a daemon recycle — the flip is a precondition for a correct W2.2 closure. The PR body's new "dependency arrow, flipped" section records this.

Full check green (padi 121 · common 22 · client 525 · server 93 incl. the 5-arm seal); two-platform CI (linux + darwin) green on the final head `da15c87a7`.

srid:

The **x86_64-linux** lane ran on **`kolu-ci-5`** (`idliv2-02`) — commit `98b2bcc9a`, **passed** ✓

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 2m43s | ✓ |
| `ci::home-manager` | 1m54s | ✓ |
| `ci::unit` | 1m4s | ✓ |
| `ci::atlas-sync` | 56s | ✓ |
| `ci::nix` | 54s | ✓ |
| `ci::smoke` | 51s | ✓ |
| `ci::fmt` | 35s | ✓ |
| `ci::biome` | 35s | ✓ |
| `_ci-setup` | 10s | ✓ |
| `ci::install` | 7s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/98b2bcc/timings.jsonl`; pool state is a live `flock` probe.</sub>

srid:
## Session-clobber fix — two data-loss paths closed (reproduce-first)

Filed against the zest incident (a saved terminal session vanished on a restart, no restore card). Two distinct paths, both fixed here; each fix ships with a red-when-reverted test.

### Reproduced twice before any fix commit

- **Red unit test** — `packages/padi/src/terminalEndpoint/reattach.test.ts`: on the pre-fix converge, `adoptSurvivingSession` against an empty adopted daemon erased the saved session (`getSavedSession()` → `null`; actives not parked).
- **Red pu-box real-world run** — a real `kolu-server` + `kaval` on a clean box. A three-capture isolation:

  | Stage | `~/.config/kolu/config.json` session |
  |---|---|
  | **BEFORE** (server#1, 3 terminals) | 3 terminals |
  | **MID** (kaval-A killed, empty kaval-B swapped in at the same socket, server#2 *not yet* restarted) | 3 terminals — the kill/swap alone does **not** touch it |
  | **AFTER** (server#2 restarts, boot **adopts** the empty kaval-B) | **`null` — erased** |

  The server#2 log carried `adopted a surviving daemon` with a matching kaval build hash — proving the adopt path fired — and the converge (`reattach.ts` → `saveSession(snapshotSession())` with an empty registry → `session.ts`'s empty→null) nuked the session.

### PATH A — an adopted *replaced* kaval erased the session

`adoptOrEnsure` adopts on gate + socket + handshake alone, so a kaval restarted out-of-band (empty, but reachable at the same socket) is adopted as if it were our survivor; the boot converge then reconciled against its empty live set and cleared the session.

The converge is now gated on daemon **identity**. The primary signal is the terminals' own IDs — a live PTY whose id matches a saved active proves the daemon is ours (this survives a supervised restart / restore that re-homed the session, so live terminals are never re-parked out from under their scrollback). An empty/foreign daemon holding saved actives ⇒ **preserve the session + park it for the restore card** (the exact `onNotAdopted` flow), never converge-to-empty. A persisted per-process `startedAt` is the tiebreak for the one case IDs can't decide — our own survivor whose actives all exited during downtime (clear, 2b) vs a replaced empty daemon (preserve) — recorded **only on the survivor path**, so a replaced/park boot never claims to own the session.

> **Note on the prescribed mechanism.** The brief prescribed persisting `startedAt` as *the* discriminant. I found that fragile: the persisted pairing goes stale across a supervised restart / restore, which would cause *worse* bugs (parking live terminals, or still clobbering a pending restore on a redeploy). Making the terminals' IDs the primary signal — with `startedAt` only as the empty-live tiebreak — reaches the same goal (detect replacement, never erase) robustly. `isReplacedDaemon` is a pure function with its own unit table (`pairedDaemon.test.ts`).

### PATH B — a create-while-restore-pending shrank the saved blob

Parked entries are invisible to `snapshotSession`, so while they stand in for the saved session on disk, any `terminals:dirty` autosave persisted a blob that omitted them. `lifecycle.create`'s `discardAllLocalParked()` (added last round for a cosmetic parked leak) made the loss certain: restore card showing → create one terminal → the N-terminal blob shrinks to 1 → a close nulls it.

One principle — the saved session shrinks/clears only on a genuine user action or an observed survivor converge, never a transient:
- **Autosave freezes the blob while a restore is pending** (parked entries exist). A genuine user-close-to-empty (no parked) still clears normally.
- **`create` no longer forfeits** — the parked entries + saved session survive, so the restore stays offered. Forfeit is now the explicit **`session.forfeit`** procedure (discard parked + clear the blob together), surfaced as a **"Start fresh"** button on the restore card.

The prior "create discards parked" test was rewritten to pin the new deliberate-forfeit semantics.

### Tests (each fails when its fix is reverted)

| # | Pins | Where |
|---|---|---|
| (i) | replaced empty daemon → session preserved + parked | `reattach.test.ts` |
| (ii) | genuine survivor → converge prunes exited, keeps dormant | `reattach.test.ts` |
| (iii) | out-of-band daemon death (degraded) under a live server → session + registry untouched | `session.test.ts` |
| (iv) | an autosave with an empty snapshot preserves the session while a restore is pending | `session.test.ts` (red-when-reverted ✓) |
| (v) | create-while-parked → all N originals held, new terminal not parked, restore still offered | `servePadi.test.ts` |
| (vi) | autosave never shrinks the blob while parked exist (property over dirty events) | `session.test.ts` (red-when-reverted ✓) |
| (vii) | `session.forfeit` → parked gone AND blob cleared, atomically | `servePadi.test.ts` |
| + | `isReplacedDaemon` id-match robustness (a live match wins over a stale pairing) | `pairedDaemon.test.ts` |
| + | the "Start fresh" button invokes `session.forfeit` | `useSessionRestore.test.ts` |

### Green

- `just check` (tsc + biome, 1035 files) clean; unit suites padi 140 / server 93.
- **Two-platform CI green** on `98b2bcc9a` (post master-merge) — every recipe `ok` on both `x86_64-linux` and `aarch64-darwin`, including `ci::e2e` on each.
- pu-box **green** re-run (both paths — session survives the empty-daemon adopt; blob survives a create-while-pending) posted separately as evidence.

srid:
## Evidence — pu-box green re-run (fix confirmed) + a latent bug it surfaced

Reused the same box the red run captured on, synced the fixed head, rebuilt kolu/kaval/kaval-tui, and re-ran the exact swap.

### PATH A — the empty-daemon adopt no longer erases the session

| | `~/.config/kolu/config.json` session |
|---|---|
| **RED (pre-fix)** after server restart adopts the empty kaval | `null` — erased |
| **GREEN (post-fix)** after the *same* restart+adopt | **3 terminals** — survived |

The server log carried `adopted a surviving daemon (its PTYs are preserved)` (the same line the red run showed — we exercised the adopt path) **and** the session survived. In the faithful variant (server that had recorded a pairing), the server logged exactly the intended `boot adopted a REPLACED kaval (not our survivor) — preserving saved session, parking for restore` with no error.

### PATH B — a create-while-restore-pending no longer shrinks the blob

With 3 genuine parked entries standing in for the restore, creating one new terminal and waiting past the autosave window left config holding the **original 3** (the `hasParkedTerminals()` freeze held); killing that new terminal also left it at **3**, not `null`.

### A latent bug the green run surfaced — now fixed (`00bb09413`)

The green run's fresh-spawn variant (the *original* zest topology — server#1 spawns kaval fresh, so it never converged onto a survivor and never recorded a pairing) exposed a real defect in the replacement-detection code: the conf store returns **`undefined`** (not `null`) for the never-written `lastPairedDaemon` key, and that `undefined` slipped past `isReplacedDaemon`'s `lastPaired !== null` guard and threw on `.startedAt`. The boot caught it and failed **closed** (daemon recycle + a "degraded" flash) — so **no data was lost** (fail-closed preserves the session via the recycle→park path), but the intended clean park + restore-card UX degraded.

Fixed by normalizing the getter (`getLastPairedDaemon() … ?? null`); added a red-when-reverted test that injects a store returning `undefined` and asserts the adopt preserves + parks without throwing. So the fresh-boot → out-of-band-restart zest path now parks cleanly, not just fail-closed.

Re-running two-platform CI on the fix head (`00bb09413`); will confirm green.

srid:
## The **Restart kaval** path was a *second*, distinct session-loss bug — reproduced RED, then fixed

My earlier green pu-box run tested the wrong restart (the server boot-adopt path). This one is the in-app **Restart kaval** button → `restartLocalDaemon` (capture → drain → recycle → park), server process staying up. Your deployed-diagnostic trace nailed it: the drain (`killAllTerminals` → PTYs exit → `emitTerminalsDirty`) arms the 500 ms autosave; it fires in the **drain→park gap** where the registry is empty **and** no parked entries exist yet, so `hasParkedTerminals()` is inert and `saveSession([])` nulls the just-captured session before park runs. The PATH-B guard only covered the steady-state parked window, not this gap.

### 1. RED first — a test driving the *real* orchestration, reproduced on the prior head

`packages/padi/src/ptyHost/restartLocal.test.ts` drives the **real** `restartLocalDaemon` (via a `__setEndpointForTest` seam + a fake recycle that outlasts the 500 ms autosave) with the **real** `initSessionAutoSave` loop and the **real** `terminals:dirty` channel — the drain's `killAll` publishes a genuine dirty (the PTYs dying), and the autosave timer actually fires in the drain→park window. On the pre-fix head:

```
FAIL  restartLocalDaemon — the in-app Restart kaval path must not lose the session
  > preserves the captured session across the drain→recycle→park window (autosave fires in the gap)
  AssertionError: expected undefined to be 2   // getSavedSession() === null — the session was nulled
    src/ptyHost/restartLocal.test.ts:156
 Test Files  1 failed (1)
      Tests  1 failed (1)
```

Exactly the trace: captured=2 → drain arms autosave → autosave fires with snapshot=0 → `writeSession(null)` → park finds null.

### 2. Fix — freeze the autosave across the restart critical section

A `restartInFlight` flag is set at **capture** (before the drain can arm anything) and lifted in a **`finally`** after **park** (which then seeds the parked entries that guard from there). The `finally` also `cancelPendingAutosave()`, so a **failed** restart can't clobber the captured session after the freeze lifts. `initSessionAutoSave`'s callback checks the flag first (`autosave: skipped (restart in flight)`). The log-only diagnostics (`d2950323d`) stay as permanent instrumentation.

### 3. GREEN — reproduce-then-passes, with an ordering guard

```
 Test Files  1 passed (1)
      Tests  2 passed (2)
```

The second test is the **ordering guard**: it asserts the autosave **did** fire on the empty registry (`snapshot=0`) *during* the restart (so the test isn't passing because the timer never fired), yet the freeze kept the session intact + parked. Reverting the freeze check (`if (restart) …`) turns **both** tests red — verified:

```
FAIL  > preserves the captured session across the drain→recycle→park window …
FAIL  > ORDERING GUARD: the autosave DOES fire in the empty-registry (snapshot=0) drain→park window …
 Tests  2 failed (2)
```

So a future refactor that drops or reorders the freeze goes red.

Full suite green (padi 143 · server 93 · `just check` 1036 files). Pushed as `976f47d83`; two-platform CI running.

srid:

The **x86_64-linux** lane ran on **`kolu-ci-5`** (`idliv2-02`) — commit `976f47d83`, **passed** ✓

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 2m22s | ✓ |
| `ci::unit` | 52s | ✓ |
| `ci::smoke` | 49s | ✓ |
| `ci::nix` | 47s | ✓ |
| `ci::atlas-sync` | 32s | ✓ |
| `ci::biome` | 25s | ✓ |
| `ci::fmt` | 24s | ✓ |
| `ci::install` | 10s | ✓ |
| `_ci-setup` | 8s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/976f47d/timings.jsonl`; pool state is a live `flock` probe.</sub>

srid:
## Typing-echo latency baseline (the W1 artifact owed before W2.2)

Measured on **current master's in-process stack** (no separate padi process yet), through the **real wire** — both clock points on kolu-server's websocket (`/rpc/ws`), which is exactly where W2.2's future hop lands:

- **t0** = `padiSurface.lifecycle.sendInput({id, data})` dispatched
- **t1** = the echoed char first seen in a `padiSurface.terminalAttach` delta
- `latency = t1 − t0`, one process's `hrtime.bigint()` (skew-free)

**Why not kaval's socket:** W2.2 inserts padi *between* kolu-server and kaval; a kaval-socket probe would sit on the far side of that hop and report a fake zero delta. Both clock points on kolu-server's wire measure the added cost honestly.

### Numbers — 2400 warm keystrokes (8 terminals × 300 measured, 50 warm-up each)

| p50 | p90 | p95 | **p99** | min / mean / max |
| 2.14 ms | 3.16 ms | 3.42 ms | **4.36 ms** | 0.16 / 2.36 / 20.9 ms |

A second independent run agreed within noise (**p99 4.21 ms**), so the percentiles are stable. Box: `pu` ephemeral Incus (NixOS), Intel i9-14900K, 32 CPUs, 125 GiB, Linux 6.12.85; kolu's bundled Node v22; built via `nix build .#default`.

### The budget this sets for W2.2

> W2.2's done-criterion (e): p99 within **5 ms** of this baseline → a W2.2 p99 under **≈ 9.3 ms**, run **on the same box, same run** (absolute numbers drift with CPU/load; only a same-box A/B is fair).

### Re-run it identically

```sh
just bench-typing-echo
```

Boots a private, isolated nix-built kolu, measures, tears down. Knobs: `KOLU_BENCH_{SAMPLES,TERMINALS,WARMUP,OUT,BIN}`.

**Method + numbers + box spec, in full:** Atlas note `padi-latency-baseline` (parents `[padi]`). **Bench + note ship in #1660** (based on `RT-fable`).

PR #1653, opened 2026-07-02
feat(atlas): hover/click footnote popovers to keep notes lean

srid:
## What

Adds a `<Footnote>` component to the Atlas component kit. An inline superscript marker whose supporting detail opens in a **popover on hover** (or keyboard focus) and **pins open on click** (also the touch path). It lets a note keep the load-bearing claim in the sentence and tuck the second-order nuance — a caveat, a citation, a corrected number, a "why we rejected X" aside — one gesture away, instead of swelling the prose with parentheticals.

```mdx
Kolu is N-dominated.<Footnote>Of 66 raw findings, 35 survived verification.</Footnote>
```

## How it fits the Atlas's self-contained-page invariant

Each Atlas page is committed, self-contained HTML (no dev server, no external `_astro` bundle) so it previews in kolu's Code tab. The footnote design respects that:

| Concern | Mechanism |
|---|---|
| **Numbering** | Pure **CSS counters** (`kolu-footnote`, reset per page on `.prose`) — no JS, renumbers automatically on add/remove/reorder |
| **Reveal** | Pure CSS `:hover` / `:focus-within` — a footnote is readable *before/without* JS |
| **Pin, dismiss, viewport clamp** | One page-level `is:inline` runtime (`FootnoteRuntime`) rendered **once** by `AtlasLayout` and shared by all markers — not duplicated per instance, no bundle, same pattern as `ForceGraph` |

The full popover CSS ships only on pages that actually use `<Footnote>`; other pages carry just the tiny early-returning runtime.

## Test drive — `performance.mdx`

Exercised the feature by tightening `performance.mdx`: ~13 dense parentheticals (methodology, verification corrections, rejected alternatives, implementation asides) moved out of the main text into footnotes. The prose reads leaner; no detail lost.

## Skill update

Documents the component and the "**push detail into a footnote, not a parenthetical**" guidance in the `/atlas` skill (source `.apm/skills/atlas/`, regenerated `.claude`/`.agents` copies).

## Verification

- `just atlas::check-sync` — green (dist in sync + host-independent/idempotent).
- `just fmt` — clean.
- Driven in a real browser (chrome-devtools): hover reveal, click-to-pin (one open at a time), Escape/outside-click dismiss, and edge clamp all confirmed. Marker is a real (non-zero) click/touch target and is exposed as an accessible expandable button. Evidence below.


srid:
## Evidence

Driven live in a browser against the built `performance.html`.

**Hover reveal** — hovering the superscript marker (¹) opens the footnote popover in place, anchored under the marker, with a numbered badge and full MDX content (styled links to the investigation write-ups):

footnote hover popover

**Click-to-pin** — clicking a marker pins its popover open (stays after the pointer leaves; the touch path too). Only one is open at a time, and Escape / outside-click dismisses:

footnote click-pinned popover

PR #1659, opened 2026-07-02
[codex] Put Kolu and Kaval in the version rail

srid:
**Kolu and Kaval identity now live inside one compact desktop rail.** The header no longer repeats the Kolu mark outside the version bar or leaves a canvas watermark peeking out behind the Dock; the rail is the single place to read app and daemon identity.

The Kaval segment is also a proper target now: clicking anywhere in it, including the update or memory-warning chip, opens the Kaval details panel. The icons render directly on the rail without extra logo tiles, and the floating Dock drops lower so it reads as a canvas tool rather than something attached to the header.

### Validation

- `just fmt`
- `nix develop --accept-flake-config -c pnpm --filter kolu-client typecheck`
- `just test-quick features/smoke.feature`
- Chrome DevTools verified one Kolu button, one Kaval button, Kaval click opens the dialog, and the final desktop layout has no canvas watermark.

### Evidence

See the latest `## Evidence` PR comment with the Chrome DevTools screenshot.

_Generated by Codex (model `gpt-5`)._

Issue #16, opened 2026-03-21
AI: Parallelize agents


srid:
AGENTS.md should instruct agents (Claude Code) to better parallelize the work, especially as we use Nix.

For eg. as package.json changes, can we spawn a Nix builder subagent to fix the hash **while** other coding work happens? Normally, the hash fix happens serially at the very end which is inefficient. What other opportunities exist for parallelization?

PR #1661, opened 2026-07-03
Padi W2.1: policy-driven re-serve machinery in @kolu/surface*

srid:
**The machinery that honors padi's per-member forwarding contract.** W1 shipped the *classification* — every `padiSurface` member declares whether it's a replayable `value` or a fail-through `delta` byte stream, pinned by a set-equality contract test. This PR builds the re-serve machinery that *acts on* it, by generalizing pulam-web's retired per-host `buildReServe`/`forwardInputStream` into the shared `@kolu/surface-nix-host` stack — now driven by the policy, and correcting the one shape that was exactly wrong for terminal attach.

It's **self-contained on purpose**: there's no kolu-server consumer yet (that's W2.2). Its consumers are its own unit tests plus the paired drishti run — and landing it separately freezes the surface API before W2.2 churns through review.

### How a re-serve fits together

```
                       reServeSurface(source, policy, session)
                                       │
  remote padi ──HostSession──▶ pumpRemoteSurface ──▶ per-binding stores + router ──▶ browser
   (the agent)                    (reconnect loop)         │
                                                           ├─ value members  → folded, HELD OPEN, replayed on rebind
                                                           ├─ delta streams  → FAIL THROUGH (end downstream on drop)
                                                           └─ procedures      → forwarded to the live spawn (fail-fast)
```

One `reServeSurface` call = one bound host: its own store set + oRPC router, shared by every connection watching that host. A parent fronting N hosts calls it once per host through the existing `buildHostRegistry` — **per-binding scope, no module-level router shared across connections** (what kolu-server wires today).

### The value/delta split, made unspellable

| Member class | Policy | On an upstream (padi↔kolu-server) drop |
| --- | --- | --- |
| cells · collections · value pulses | `value` | **Held open + replayed** — the store keeps the last value (no healthy-but-empty flash); a rebind re-seeds it |
| byte / liveness streams (`activity`, `terminalAttach`) | `delta` | **Fail through** — the downstream stream ends too, so the client's own retry re-subscribes end-to-end and a scrollback snapshot only ever arrives as the *first frame of a FRESH stream* |

The two per-subscriber stream relays are type-guarded on the policy: `relayHoldOpenStream` accepts only a `value` member, `relayFailThroughStream` only a `delta` member. So *holding open a byte stream is a compile error* — splicing a replayed snapshot into a live xterm is **unrepresentable, not a convention to remember**. pulam-web's `forwardInputStream` held streams open across respawns; that's right for a pulse and wrong for attach, and this splits the two.

### Proven against a toy surface

Three hops in miniature (fake upstream agent → the re-serve → a `directLink` downstream), no ssh:

- **kill-the-middle-hop on a fail-through member terminates the downstream stream** (never a spliced frame from the next spawn);
- **a hold-open cell replays its value after rebind** (and a value pulse stream keeps yielding across the drop);
- two bindings get **independent stores + routers**;
- a mis-declared policy (a cell marked `delta`, or an unannotated member) **crashes loudly** rather than silently folding.

### Design philosophy

- **Fail-fast, no fallbacks** — `implementSurface` refuses to start with any member unwired; a procedure forwarded while the link is down throws loudly; a `delta` cell/collection is rejected at build; there is no knob to hold open a byte stream (it's a type error).
- **Volatility boundary** — the machinery lands in `@kolu/surface-nix-host`, the receptacle that already owns `HostSession` reconnect + mirroring; the dependency arrow points out, and it reuses `pumpRemoteSurface` / `mirroredSurface` / `seedConnectionCell` / `implementSurface` rather than a parallel path.
- **Reuse the source of truth** — the re-serve reads the *same* forwarding classification W1 pinned (a surface's `... as const satisfies RelayPolicy`), with no second declaration to drift.

> **Drishti-gated.** This is an API-facing change to `packages/surface-nix-host`, so per `.claude/rules/surface.md` it ships with a paired [drishti](https://github.com/srid/drishti) PR whose full CI passes pinned to this PR's final HEAD. The gate turned out better than additive: the stale-rows fix landed an optional `initialKeys` hook in `@kolu/surface`, and the drishti PR **adopts** it (fixing drishti's own ghost-PID-on-reconnect) — a real second-consumer proof, not a re-pin.

Refs padi §W2.1 · needs W2.2 to consume it.


srid:
## 🔬 External-review disposition — 23 candidates

An independent 8-angle review produced 23 deduped candidates. Each was **adversarially verified** against the tree (a 22-agent pass — every verifier read the code and tried to *refute* before confirming), then folded into the review gauntlet. Dispositions below; line-quotes are function-anchored (line numbers shift across the gauntlet).

### ✅ Fixed (folded into the gauntlet commits)

| # | Finding | Fix |
| --- | --- | --- |
| **1** | **STALE-ROWS** — a collection key that departs while the link is down survives forever (a fresh mirror's `open` map can't remove keys it never saw). | `mirrorCollection` gained an optional, additive `initialKeys()`; on the FIRST fresh `keys` frame it prunes carry-over keys the snapshot omits (via `onRemove`, so subscribed downstreams see the departure) — no stale row, no empty flash. `reServeSurface` keeps per-collection caches and passes them per spawn. Mirror-level + cross-reconnect tests added. `@kolu/surface` change is additive; the drishti PR **adopts** it (fixing drishti's own ghost-PID-on-reconnect). |
| **7** | **MARKCONNECTED-CUE** — a cell-less surface never marks connected → gate stays closed → watchdog cycles to failed. | `reServeSurface` fails loud at build if the source declares no cells (a cell's on-connect snapshot is the cue *and* keeps the mirror alive). Test added. |
| **8** | **CELL-WRITE-VERBS** — a downstream cell write hits the local store and is silently reverted. | The re-serve is now an explicit READ mirror: a cell's `get` folds, but its store's `set` THROWS, so a write (set/patch/test__set) rejects loudly. The fold publishes onto the cell's own `<key>:changed` channel, bypassing the throwing set, so reads still flow. Test added. *Full write-forwarding is W2.2.* |
| **10** | **HANDSHAKE-ABORT** — `failThroughStreamCore`'s initial `get()` was outside the try, so an abort mid-handshake threw. | Moved inside the try; `isAbortReason` turns it into a clean return. Test added. |
| **3a** (rider) | The `catch{return}` around `whenChanged` swallowed any rejection. | Now `isAbortReason`/rethrow, matching the sibling core. *(The per-input-termination classification itself is deferred — see below.)* |
| **5a** | **GUARDS-NO-CALLERS** — docstrings/README claimed a compile guard the runtime-routed assembly doesn't have. | Docs corrected: the assembly routes by `policy[member]` at RUNTIME (`requirePolicy`); the compile guard is on the direct-use relays (pinned by `_typeGuards`). |
| **9** | **DONE-REJECTION** — `done` can reject unobserved. | `done`'s docstring now states it can reject (SinkError / mismatch), stops the pump, and the consumer MUST observe it. *(Consumer wiring is W2.2.)* |
| **20** | **TEST-FIXTURE-DUP** — the reServe test's `controllable` was signal-blind, masking the real abort path. | One shared **signal-aware** `controllable` used by both test files; the signal is threaded through the toy upstream. |
| **6, 18** | (makeSink silent-skip; dead Maps) | Already fixed by the lens-debate pass. |
| **22a/b** | Redundant rebind-guard disjunct; eta-wrapped connection setter. | Simplified. |

### ⏸️ Confirmed but deferred to W2.2 (with a docstring TODO)

- **3a — HOLD-OPEN per-input termination.** `holdOpenStreamCore` uses "did the live client change?" as its sole proxy for "did the link die?", so a value stream that ends/rejects for a *per-input* reason while the same client stays live parks silently until the next respawn. **Latent today** — padi's value streams (`subscribeRepoChange`/`subscribeFileChange`) only wire through here at W2.2, where the recovery semantics (surface-the-error vs re-subscribe) and the link-death race are ratified. Docstring TODO references this candidate; pinned as a W2.2 done-criterion in the padi note.

### 📦 Scoped to W2.2 (not W2.1 defects)

- **8 (full write-forwarding)** — forwarding cell writes to the agent is W2.2; W2.1 fails them loud.
- **14 — ZOD-PER-FRAME** — per-frame output validation is a framework property at every hop (a `z.string()` typeof for attach); a passthrough opt-out, if measured worth it, is a W2.2 `@kolu/surface` change.
- **5d — POLICY-VOCAB-DUP** — homing the `value|delta` union in `@kolu/surface` is proof-before-extraction; the `PADI_FORWARDING_POLICY satisfies RelayPolicy` link becomes load-bearing at W2.2.
- **21 — INMEMORY-COLLECTION** — extracting `inMemoryCollection<K,V>()` is a clean additive dedup, deferred to keep the W2.1 `@kolu/surface` change focused on the reconcile.

### ❌ Refuted (with the refuting quote)

- **2 — LEAD-FRAME.** `holdOpenStreamCore` emits the lead on **every** bind, including post-reconnect (`if (opts.lead !== undefined) yield opts.lead;` sits at the top of the `while` loop after the `client === null` guard). The "stale git status" claim also fails: padi's value pulse streams self-snapshot on subscribe (`{seq:0}` then incrementing), so a rebind forwards a fresh pulse that drives `createPolledQuery`'s requery. `reServeSurface` passing no lead is correct (a lead is a per-stream frame a generic assembly can't synthesize).
- **4 — HOLD-OPEN-HANG.** Not a production bug — a test-fake artifact. The downstream signal is threaded into the upstream call (`await select(client).get(input, { signal })`) and a real oRPC upstream rejects its pending pull on abort (`publisherChannel`: "AbortController.abort()s every in-flight stream's signal — the publisher iterator then rejects pending pulls"). This is the codebase's canonical contract (`iterateUntilAborted` relies on exactly this, not a `Promise.race`). *(The signal-blind test fake that masked it is fixed — candidate 20.)*
- **5c — POLICY-SET-EQUALITY.** `requirePolicy` fires for **every** stream and event (via `relayForMember`, called from both loops; throws `has no forwarding policy`). The "orphan key" half is a non-defect: a policy entry for a cell/collection/procedure is *inert by design* (the policy is a whole-surface superset), and the exact no-gap/no-orphan set-equality already lives at the source (`padi/src/surface.test.ts`: `expect(annotated).toEqual(members)`).
- **11 — RETRY-HOT-LOOP.** `ClientRetryPlugin` has a **1000ms floor** (`retryDelay: (o) => o.lastEventRetry ?? 1000`); on a subscribe-during-gap `failThroughStreamCore` throws before any event, so re-subscribe is ≤ once/second, not a spin. The core has no loop (throws once). Attach's `unenrolledStreamCall` inherits the same floor.
- **12 — STALE-LIVE-SET.** `activity` is a **delta** member (`activity: "delta"`), routed to `failThroughStreamCore`, which re-throws on upstream death and ends the downstream — it is **not** held open, so no last frame is pinned.
- **13 — AMPLIFICATION.** padiSurface has **no** value + no-input stream. The only no-input stream (`activity`) is delta; both value streams (`subscribeRepoChange`/`subscribeFileChange`) and the value event (`terminalExit`) are input-keyed (inherently per-subscriber). Every no-input DATA member (cells, collections) is still folded **once** via the sink.
- **15 — GENERATOR-STACK.** ~3 microtask hops per frame ≈ single-digit µs, immaterial to the 5ms typing budget (dominated by ssh RTT + ws + xterm paint). Dropping `iterateUntilAborted` would trade the codebase's single home of the abort-teardown contract for µs — not recommended.
- **16 — NO-BACKPRESSURE.** Both relays are plain async generators with no buffer — each upstream pull happens only after the downstream pulls (natural backpressure). No in-process bus on the stream path. (oRPC's own receive buffer is the framework's, unchanged by this PR.)
- **17 — DOUBLE-WAKE.** Benign: a clear-wake (`current === null`) re-parks at the loop head's null-guard without ever calling `get()`; cost is at most one extra `whenChanged` per rare reconnect.
- **22c** — a `patchSchema`-only cell dep would throw loudly at build (the intended fail-fast); no padi cell has one.

*Contract checks the audit already passed (policy satisfies `RelayPolicy`; session cell+ns merge; pump dep shapes; additive-only → no existing-API drishti break; no `whenChanged` listener leaks) were re-confirmed.*

srid:
## 🤖 Codex ⇄ Claude review debate — **consensus** (2 rounds, reasoning effort `xhigh`)

Base: merge-base of `origin/master`. Codex reviewed the diff; a Claude author fixed/disputed each finding to agreement. Net: 3 robustness fixes in the new code (F1/F4/F5), 2 findings correctly scoped out as belonging to already-merged #1659 (F2/F3).

### Round 1

**codex** — approved: `false`

The surface re-serve machinery is mostly coherent, but I found one real sink-error swallowing bug in the new `initialKeys` path. The UI change also drops the desktop-visible host-qualified app identity after removing the watermark, and the replacement smoke step is race-prone. I did not treat the drishti ship gate as a code finding.

Findings:
- `F1` · major · open — `initialKeys` is a caller-supplied sink callback, but it is evaluated before the sink-error wrapper path. If it throws, `mirrorCollection` rejects with a raw error; `mirrorRemoteSurface` later only rethrows rejected tasks whose reason is `SinkError`, so this failure is silently collapsed into a resolved `done`. That violates the documented contract that sink failures reject the mirror. (packages/surface/src/mirrorRemoteSurface.ts:598)
- `F2` · major · open — After removing `CanvasWatermark`, the desktop chrome no longer shows the server-qualified app identity. The rail hardcodes `Kolu`, while the actual identity is `Kolu [hostname]` from `server.info`. In desktop/PWA/fullscreen use, two Kolu instances on different hosts now look identical in the visible chrome. (packages/client/src/ui/IdentityRail.tsx:180)
- `F3` · minor · open — The new smoke step reads `page.title()` once immediately after navigation, but the `Kolu [host]` title is populated by the async `client.server.info()` fetch. This can race and fail with the default `kolu` title. It also no longer verifies visible branding after the watermark removal. (packages/tests/step_definitions/smoke_steps.ts:21)
- `F4` · minor · open — The public API reference for `SurfaceSink.collections` still documents only `{ upsert, remove }`, but this branch adds the optional `initialKeys` callback. Consumers reading the package README will miss the new cross-reconnect reconciliation hook. (packages/surface/README.md:932)
- `F5` · minor · open — `failThroughStreamCore` throws `NoLiveUpstreamError` before checking whether the downstream signal is already aborted. A teardown that reaches this generator while `holder.current` is null can surface a no-live-upstream error instead of ending cleanly as an unsubscribe. (packages/surface-nix-host/src/relayStream.ts:235)


**claude** — Fixed the three in-scope surface findings (F1 sink-error swallowing, F4 README gap, F5 abort-before-throw) with regression tests; disputed F2/F3 as belonging to an already-merged, out-of-scope PR.

- `F1` **fixed** — Real bug, and in-scope. `mirrorCollection` materialized the caller's `initialKeys` sink callback with a bare `new Set<K>(opts.initialKeys?.() ?? [])` (`packages/surface/src/mirrorRemoteSurface.ts:598`). It runs synchronously at spawn, before the `rejectSink` channel exists, so a throw returned a raw rejected promise from the async fn — and the top-level fold only rethrows `r.reason instanceof SinkError` (`mirrorRemoteSurface.ts:492`), so a raw throw was swallowed by `allSettled` and `done` resolved cleanly, violating the documented "sink failures reject the mirror" contract. Wrapped the materialization in `try/catch` → `throw new SinkError(sinkErr)` so it rejects `done` exactly like a throwing `onUpsert`/`onRemove`. Added a test: `initialKeys` throws → `done` rejects with the cause (`mirrorRemoteSurface.test.ts`). Suite green (surface 16 tests, +1).

- `F2` **disputed** — Out of scope: not this PR's code. The IdentityRail/`CanvasWatermark` removal (`packages/client/src/ui/IdentityRail.tsx`, `canvas/CanvasWatermark.tsx`, `ChromeBar.tsx`) was authored in PR #1659 ("Put Kolu and Kaval in the version rail", commit `f4699970b`) and reached this branch only via the `Merge branch 'master'` commit `ab39612a3`. `git merge-base HEAD origin/master` == `f4699970b`, i.e. #1659 is already on master; `git diff f4699970b..HEAD` (this PR's real contribution) touches ZERO client files. codex's `git diff <padi-W1-base>` swept it in because the review base predates the master merge. On the merits it's also not a regression: host identity was deliberately moved, not lost — the browser tab/window title is `Kolu [host]` (asserted by the smoke test this PR's diff shows) and the full server identity lives in `KoluInfoDialog` reached from the chip; the ChromeBar comment states the design ("Server details live in the rail dialogs, not as duplicate branding on the canvas"). If this bar should change, it belongs as a follow-up against #1659/master, not W2.1 (whose charter is "keep the surface API minimal").

- `F3` **disputed** — Same scope basis as F2: `packages/tests/step_definitions/smoke_steps.ts` and `features/smoke.feature` were changed in the merged #1659 (`f4699970b`), not in this PR's contribution (`git diff f4699970b..HEAD` touches no test/UI files). The `page.title()` poll-vs-single-read point has merit for #1659 (the title is set after an async `client.server.info()`), but `page.goto("/")` awaits the load event and #1659 is green on master; re-opening its smoke step here would be scope creep on already-shipped code. Belongs as a follow-up on #1659/master, not W2.1.

- `F4` **fixed** — Valid doc gap, in-scope (this PR added `initialKeys` to `SurfaceSink`). `packages/surface/README.md`'s `SurfaceSink<S>` API reference (~:932) listed collections as only `{ upsert, remove }`. Added the optional `initialKeys?: () => Iterable<key>` with its first-snapshot cross-reconnect prune semantics.

- `F5` **fixed** — Valid, in-scope, and consistent with the existing design. `failThroughStreamCore` (`packages/surface-nix-host/src/relayStream.ts:234`) read `holder.current` and threw `NoLiveUpstreamError` before checking `signal?.aborted`, so a teardown that first pulls the generator after the downstream already aborted — while no upstream is live — surfaced a spurious dead-link error instead of ending cleanly. The sibling `holdOpenStreamCore` already returns on abort at its `while (!aborted())` head, and the subscribe-handshake abort is already swallowed via `isAbortReason` (#1661 candidate 10); this closes the symmetric pre-abort gap. Added `if (signal?.aborted === true) return;` before the `client === null` branch, plus a regression test (aborted signal + null upstream → clean return, no `NoLiveUpstreamError`). Suite green (surface-nix-host 18 tests, +1).

Gate: `tsc --noEmit` (both packages), `biome check`, and `biome format` all clean; affected vitest suites green (surface 16, surface-nix-host relay+reserve 18).

commit: `510e87a923552712fa36aa0b5389986e6caf7826`


### Round 2

**codex** — approved: `true`

All previously raised findings are resolved. I verified the current tree: F1, F4, and F5 have concrete fixes with regression coverage, and I accept the F2/F3 scope dispute after checking that the W2.1 contribution from `f4699970b..HEAD` touches no client/test UI files.

Findings:
- `F1` · major · resolved — `initialKeys` could previously throw outside the sink-error path, allowing the mirror `done` promise to resolve cleanly instead of rejecting on a local sink failure. (packages/surface/src/mirrorRemoteSurface.ts:598)
- `F2` · major · resolved — Previously raised as a visible identity regression, but Claude showed this came from already-merged PR #1659 rather than W2.1's contribution. (packages/client/src/ui/IdentityRail.tsx:180)
- `F3` · minor · resolved — Previously raised as a smoke-test race from one-shot `page.title()`, but this test change also came from already-merged PR #1659 rather than W2.1's contribution. (packages/tests/step_definitions/smoke_steps.ts:21)
- `F4` · minor · resolved — The README API reference omitted the new optional `initialKeys` sink hook and its reconnect-prune behavior. (packages/surface/README.md:932)
- `F5` · minor · resolved — `failThroughStreamCore` could throw `NoLiveUpstreamError` on an already-aborted subscribe before recognizing teardown. (packages/surface-nix-host/src/relayStream.ts:234)

_codex on the rebuttal:_ F2: conceded. I verified `git merge-base HEAD origin/master` is `f4699970b45166e3348c721f8754902690a93b0e`, and `git diff f4699970b..HEAD -- packages/client packages/tests features` is empty. The identity concern should not block W2.1. F3: conceded on the same scope basis; the polling concern may be worth a follow-up, but it is not part of this branch's actual contribution. F1, F4, and F5: fixes verified in the current tree and accepted.

srid:
## 🛡️ Review gauntlet summary

This PR ran an unusually deep review: **lens-debate → an external 8-angle audit (22-agent adversarial verification) → codex-debate → simplify → code-police**, each folding its fixes onto the branch. Outcomes:

| Reviewer | Result | Fixes applied |
| --- | --- | --- |
| **⚖️ lens-debate** (lowy ⇄ hickey) | consensus, 2 rounds | Reshaped the forwarding policy to a **streaming-survival axis** (cells/collections are always value — dropped the whole-surface `requireValue` checklist); dropped dead `cellStores`/`collectionCaches` Maps; `makeSink` throws on a missing ctx member instead of silent-skipping. |
| **🔬 external 8-angle audit** (23 candidates) | 22-agent verify → [disposition ↑](https://github.com/juspay/kolu/pull/1661) | 9 confirmed & fixed (STALE-ROWS reconcile, cell-write read-mirror, markConnected fail-fast, handshake-abort, honest docs, …), 1 deferred to W2.2 with a docstring TODO (hold-open per-input termination), 10 refuted with line-quotes, 4 W2.2-scoped. |
| **🤖 codex-debate** | consensus, 2 rounds ([↑](https://github.com/juspay/kolu/pull/1661)) | `initialKeys` failure now rejects `done` (SinkError); fail-through returns clean on an already-aborted subscribe; `@kolu/surface` README documents the new hook. |
| **✨ simplify** | 3 fixes | `ForwardableStream.get` reuses `UpstreamSource`; `waitNextSpawn` + `buildRelayDeps` dedupe. |
| **👮 code-police** | 4 fixes | `initialKeys` fails loud on a missing cache (matching its siblings); corrected the `markConnected` doc; commented the test-helper's load-bearing ordering; inlined a thin wrapper. |

**Test coverage:** `relayStream.test` + `reServeSurface.test` (18 tests against a toy surface — fail-through terminate, hold-open replay, cross-reconnect prune, read-mirror write-throws, cell-less fail-fast, per-binding independence, the `@ts-expect-error` compile guard) + `mirrorRemoteSurface.test` (16 — the cross-reconnect prune + its fail-fast). `just check` (tsc + biome `--error-on-warnings`) green across 1040 files.

> **Drishti gate:** paired PR pinned to this PR's final HEAD, **adopting** the additive `mirrorCollection.initialKeys` hook (which also fixes drishti's own ghost-PID-on-reconnect) — a real second-consumer proof, not just a re-pin. Linked below once green.

srid:
Paired drishti gate (per `.claude/rules/surface.md`): **srid/drishti#84** pins kolu to this PR's final HEAD `46dc0fb34` and adopts the additive `mirrorCollection.initialKeys` reconcile in drishti's own re-serve sink (`cpuCores`/`networkInterfaces`), a real second-consumer proof of the hook. Running drishti CI on both x86_64-linux + aarch64-darwin now.

srid:
🔗 **Drishti gate:** paired PR **[srid/drishti#84](https://github.com/srid/drishti/pull/84)** — pins kolu to this PR's final HEAD `46dc0fb34` and **adopts** the additive `mirrorCollection.initialKeys` reconcile hook (which also fixes drishti's own ghost-PID-on-reconnect), so the gate is a real second-consumer proof, not just a re-pin. CI (both platforms) running.

srid:

The **x86_64-linux** lane ran on **`kolu-ci-6`** (`dev-x86-64-linux-03`) — commit `46dc0fb34`, **passed** ✓

- **Lane wall** (pipeline): **3m14s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 2m57s | ✓ |
| `ci::home-manager` | 1m12s | ✓ |
| `ci::unit` | 54s | ✓ |
| `ci::nix` | 49s | ✓ |
| `ci::smoke` | 44s | ✓ |
| `ci::surface-app-example-build` | 29s | ✓ |
| `ci::biome` | 28s | ✓ |
| `ci::fmt` | 27s | ✓ |
| `ci::install` | 9s | ✓ |
| `_ci-setup` | 8s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/46dc0fb/timings.jsonl`; pool state is a live `flock` probe.</sub>

srid:
## Final audit (RT-fable coordinator) — APPROVED for merge

Three independent auditors ran against the final head `46dc0fb34`:

**Fixes: 9/9 hold**, each verified at file:line with its test — the `initialKeys` prune latches once per spawn and flows through real `onRemove`s (a *held* downstream subscription observes the reconcile, no empty flash); the throwing store catches all write verbs with no `test__set` whitelist; the #3a deferral cross-references the plan's W2.2 gate; docs claim the compile guard only where it is true.

**Refutations: 8/10 upheld, 2 shaky, 0 wrong.** The two shaky ones matter for W2.2, not for this merge:
- **#11 (retry loop):** conclusion holds — no hot loop — but the cited mechanism is wrong in an important way: a mid-chain error reaches the browser as a non-retriable `ORPCError`, so `STREAM_RETRY` **never fires** for it. Fail-through's documented recovery ("the client re-subscribes end-to-end") is therefore application wiring W2.2 must build, not a freebie. Pinned into W2.2's done-criterion (c).
- **#16 (backpressure):** the relays are honestly pull-based, but the real answer to "where do frames queue" is oRPC's unbounded receive queue under the relay, with an unpaced sender — pre-existing and framework-wide, now named in W2.2's residuals rather than left as a parenthetical.
- One unclaimed residual surfaced by the #2 audit: `terminalExit` rides hold-open with no self-snapshot, so an exit fired during a link gap is lost to a held subscriber. Also pinned.

**Drishti gate: satisfied.** srid/drishti#84 pins exactly `46dc0fb34`, **adopts** `initialKeys` in both cache-backed sinks of its process-monitor re-serve (with the `processes` exception correctly reasoned), links are bidirectional, and all 56 CI checks across both repos are green at the right SHAs. The PR body's stale "drishti only re-pins" sentence has been corrected.

Residuals (`terminalExit` gap-loss · relay receive-buffer bound · `equals`-dedup edge · the #3a classification) are pinned as **W2.2 done-criteria (g)/(h)** in the plan of record (padi.mdx, `cebd2f7e9`) — none block this PR.

_Audit by the RT-fable coordinator session (Claude Code, model `claude-fable-5`)._

PR #1662, opened 2026-07-03
Back up kolu state before startup writes

### LINKED ISSUE #1658: Back up state JSON on kolu-server start (timestamped + rotated)
## Problem

kolu's persisted state (`~/.config/kolu/config.json` — `session`, `activityFeed`, `preferences`) is a single mutable file that autosave overwrites in place. There is no history, so any bug that persists a bad value **destroys the prior good state with no recovery path**.

This bit us for real: an out-of-band kaval restart (kaval killed/restarted over ssh while kolu-server stayed up) emptied the terminal registry, autosave persisted the emptied `session` over the real one, and the saved session was **gone from disk** — not merely un-offered. A backup would have made that data loss recoverable while the root-cause fix was in flight. (Root cause is being fixed separately; this issue is the safety net that should exist regardless.)

## Proposal

On **kolu-server start**, before the store is first written, snapshot the current state file to a **timestamped backup** and keep a **rotated** ring of the last N.

- Copy `config.json` → `config.<UTC-timestamp>.json` under a `backups/` subdir of the state dir (e.g. `~/.config/kolu/backups/`), at boot, before any write.
- **Rotate**: keep the most recent N (default ~10) and the copy is a no-op if the current file is byte-identical to the newest backup (don't churn identical snapshots across quick restarts).
- **Fail-soft on the backup itself only** — a backup write error must log loudly but not block boot (the backup is a safety net, not a gate); this is the one place a non-fatal fallback is correct, distinct from the fail-fast doctrine that governs the live store.
- Boot-time only keeps it simple and cheap; a corrupting write between restarts is still covered by the *next* boot's snapshot, and boot is the natural rotation tick.

## Out of scope

- No automatic restore/rollback UI — recovery is manual (copy a backup over `config.json`) for now.
- Not a substitute for the autosave-clobber guard (refuse to persist an emptied session over a non-empty one) — that is a separate correctness fix.

## Acceptance

- On start, a timestamped copy of the pre-existing state file appears under `backups/`, capped at N, newest-identical deduped.
- A test pins: boot with an existing state file → backup created; boot again unchanged → no new backup; N+1 backups → oldest pruned; backup-write failure → boot proceeds + logs.

srid:
**Kolu now snapshots its persisted state before startup can mutate it.** On server boot, the existing `config.json` is copied into a timestamped `backups/config.<UTC timestamp>.json` file before `Conf` runs migrations or autosave can write, so a bad persisted value no longer destroys the only recoverable copy.

The backup ring is intentionally small and boring: it keeps the newest 10 snapshots, skips new files when the current state is byte-identical to the newest backup, and still prunes old backups on unchanged restarts. The live store stays fail-fast (`KOLU_STATE_DIR` is still required, and Conf/schema validation still surface real state problems); only the safety-net copy is fail-soft, where failures are logged loudly and boot continues.

### Design fit

- **Fail fast:** live state requirements and validation stay strict; only the backup safety net catches errors, as requested in #1658.
- **Volatility boundary:** this remains inside the server persistence owner (`state.ts`) because it is a bounded file snapshot beside the Conf store, not a new transport/persistence framework.
- **Source of truth:** the code uses the existing `KOLU_STATE_DIR`/Conf `config.json` location instead of inventing a second state path, and the README/Atlas docs now name the real file.

### Verified

- `just fmt`
- `just check`
- `nix develop -c pnpm --filter kolu-server test:unit -- state.test.ts`
- `just atlas::build`
- `just atlas::check-sync`

Closes #1658.


srid:

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

srid:
## Review and evidence

Reduced review gauntlet:

| Step | Result |
| --- | --- |
| Lowy/Hickey structural review | Found storage-layout split and test-shaped backup API; fixed in `4940af24c` by deriving paths from one state layout and making backup-before-Conf explicit. |
| Codex debate | Skipped at maintainer request. |
| Simplify | Folded into the structural fix; no extra abstraction added. |
| Code-police | Found `existsSync` could hide stat failures; fixed in `6e5f8c5a3` with `statSync` + ENOENT-only no-op and an ENOTDIR regression test. Rerun came back clean. |

Verification:

- `just fmt`
- `nix develop -c pnpm --filter kolu-server test:unit -- state.test.ts` (99 tests passed in the server unit slice)
- `just check`
- `git diff --check`
- odu CI passed on both required lanes: `x86_64-linux` on `kolu-ci-7`, `aarch64-darwin` on rasam.

No visual impact: this is server startup persistence behavior before any rendered Kolu UI or terminal surface exists; exercising it creates/rotates files and log lines only.

PR #1660, opened 2026-07-03
Typing-echo latency baseline for padi W2.2 (< 5ms budget)

srid:
**The number W2.2 has to beat.** padi W2.2 slips a whole new process — padi — *between* kolu-server and kaval, so every keystroke gains a hop. The [plan](https://github.com/juspay/kolu/blob/RT-fable/docs/atlas/src/content/atlas/padi.mdx) budgets that hop at **< 5ms added p99** and says the ceiling is *measured, not guessed*: W1 (#1652) owes a baseline of today's echo latency, and W2.2's done-criterion (e) compares against it. This PR is that owed artifact — a reproducible bench plus the Atlas note that records the method and the numbers.

### What is measured, and where the clock sits

One keystroke's full round-trip through the **real stack**, timed by a headless client on the **same websocket the browser uses** (`/rpc/ws`):

```
client probe ─ t0: sendInput ─▶ kolu-server ─▶ kaval ─▶ PTY echo
client probe ◀─ t1: terminalAttach delta ◀─ kolu-server ◀─ kaval ◀─┘
```

- **t0** — the probe dispatches `padiSurface.lifecycle.sendInput({id, data})`.
- **t1** — the echoed char first appears in a `padiSurface.terminalAttach` delta.
- `latency = t1 − t0`, both from one process's `hrtime.bigint()`, so absolute skew is irrelevant.

> **Why both clock points are on kolu-server's wire, not kaval's socket.** W2.2 inserts padi *between* kolu-server and kaval. A probe dialling kaval's socket directly would sit on the *far* side of that future hop and report a fake near-zero delta. Clocking `sendInput → terminalAttach` over kolu-server's `/rpc/ws` puts **both** endpoints on the near side, so W2.2 re-running this identical probe measures the added cost and nothing else. *Like-for-like beats a prettier absolute number.*

### The baseline

Current master state (padi served **in-process**, no separate padi process), nix-built `.#default`, on a quiet ephemeral pu box. **2400 warm keystrokes** (8 terminals × 300 measured, 50 warm-up discarded each):

| percentile | latency | | |
| --- | --- | --- | --- |
| **p50** | 2.14 ms | **p95** | 3.42 ms |
| **p90** | 3.16 ms | **p99** | **4.36 ms** |

*A second independent run agreed within noise (p99 4.21 ms) — the percentiles are stable, not a lucky draw.* Box: Intel i9-14900K, 32 CPUs, 125 GiB, Linux 6.12.85; kolu's bundled Node v22.

[...]

PR #1664, opened 2026-07-03
padi W2.2: padi the process, and the cutover (bind · re-serve · state-root)

srid:
**kolu's per-host terminal workspace is now its own long-lived process — `padi` — that owns your terminals, their layout, and your saved session; kolu-server thins to the web shell that binds it and connects each browser view to it.** The visible payoff is warm restarts: **restarting or redeploying kolu-server no longer loses your terminals' metadata** — padi keeps the registry warm across the restart, so the canvas re-attaches already-whole instead of re-deriving itself, and the transient window where a mid-restart crash could eat the session is gone.

This is the big PR of the [padi plan](https://github.com/juspay/kolu/issues/951) — one PR, **two commit stages** (the seam W1 drew, now given a process and then a cutover). It builds directly on W2.1's re-serve machinery (#1661) and the #1660 latency baseline.

### The two commit stages

- **Stage 1 — the binary and its identity** (`44c6fda`). `packages/padi` graduates from library to process: `package = process = restart-hash`. A `padi` binary serves `padiSurface` plus the **frozen control core** (hello · version · drain · clock.now) over a **0700 digest-keyed unix socket**, owns its **persistent state-root** identity (the default spelled by the binary on the host; session/activityFeed/pairing in its own `Conf`), and **spawns-or-adopts its own kaval** under `kaval-<digest>/` — so two padis at distinct state-roots can never touch each other's kaval (the #1313 property, by construction). Its own Nix derivation carries a `PADI_BUILD_ID` staleKey pinned to the daemon source closure.
- **Stage 2 — the cutover** (`ab16192`). kolu-server stops serving padiSurface in-process and instead **binds the local padi** (a new `adopt-or-spawn-or-refuse` supervisor policy — a contract skew is *refused*, never recycled: clients never kill a running padi) and **re-serves** it through W2.1's typed forwarding policy. Session persistence moves to the state-root via a **one-shot import** (takes the old shared `config.json` as-is, once, backing it up to `config.json.pre-padi-import.bak` first and crashing loudly on a bad file). The mirrored-cell endgame lands: cell writes now **forward to padi** (retiring W2.1's interim throwing store), and the W2.1-audit residuals close.

```
before:  browser ──ws── kolu-server (padiSurface in-process) ──socket── kaval
after:   browser ──ws── kolu-server (web shell) ──socket── PADI ──socket── kaval
                                    │ binds + re-serves        │ owns registry,
                                    │ (adopt-or-spawn-or-refuse)│ fold, session, kaval
```

### Done-criteria (a)–(h)

| # | criterion | how it's covered |
| --- | --- | --- |
| (a) | UX byte-identical, e2e green | e2e on the pu box · **one deliberate exception below** |
| (b) | kolu-server restart keeps metadata warm | `padiBinding.test.ts` — restart the binder → **adopts** the surviving padi (same pid), the pre-restart terminal's screen intact · pu-box smoke |
| (c) | kill-the-middle-hop → re-attach, never spliced | a mid-chain death arrives as a non-retriable `ORPCError`, so `STREAM_RETRY` won't fire — an **outer re-subscribe loop** (`consumeReattachingStream`, now its own module + unit-tested) in the browser attach consumer; `padiBinding.test.ts` asserts a mid-attach padi SIGTERM **ends** the held iterator (never splices); reconnect pinned there too, live kill on the pu box |
| (d) | fold differential byte-identical pre/post | the fold moved into padi at W1; its differential tests keep passing |
| (e) | typing-echo p99 within +5ms of #1660's 4.36ms | **PASS** — see below |
| (f) | W1 session-safety survives the split | `reattach.test.ts` + the drain-path no-shrink invariant in `session.test.ts` pass with the orchestration in padi (the dead in-process `restartLocal` path is **deleted** — the **drain** is the sanctioned restart now); the property is re-pinned **across the wire** by `padiBinding.test.ts` (drain + re-bind → the state-root blob never nulls/shrinks) |
| (g) | relayStream hold-open classification (#1661 cand. 3a) | **resolved** — a hold-open value stream that cleanly *ends* while the link stays live now surfaces downstream (client re-subscribes) instead of parking silently |
| (h) | the other W2.1-audit residuals | (h1) `terminalExit` — resolved by (g) + the collection re-syncs at rebind; (h2) the per-subscriber channel gained an opt-in high-water bound; (h3) the equals-dedup edge is moot (the write-forward bypasses `equals`) |

### The deliberate visible changes

Two, both intentional — the rest of the workspace stays byte-identical:

1. **A third process shows on the memory readout** (identity rail + Diagnostic Info dialog): **padi**'s own honest RSS (or an honest "no daemon"/error state, never a fake zero) joins kolu-server's and kaval's. padi owns kaval now, so padi serves the readout for itself + its kaval; kolu-server contributes only its own RSS.
2. **Padi joins the version rail** beside Kolu and Kaval — the ownership chain reads left-to-right (kolu-server → padi → kaval). It has its own mark — **a staircase in profile**, the literal Tamil படி ("step"), kin to kolu's own stepped கொலு mark — a status pip painted **live from padi's binding state** (never a fake green), a details dialog on the same shell Kolu and Kaval use, and **its own honest uptime** (from the boot time it stamps in its control-core hello; kolu-server gained an uptime too). *(Added on request now that padi is a genuine process.)*

*(Before/after screenshots in the Evidence comment.)*

### Latency (e) — measured, not guessed

Back-to-back on one quiet box (i9-14900K, NixOS), 2400 samples each, two passes. The stage-1 in-process build reproduces the #1660 baseline (p99 4.33/4.43 ms vs 4.36), validating the box + method:

| | baseline (in-process) p99 | cutover (padi process) p99 | delta |
| --- | --- | --- | --- |
| pass 1 | 4.331 ms | 5.681 ms | **+1.35 ms** |
| pass 2 | 4.429 ms | 5.693 ms | **+1.26 ms** |

**The padi hop costs ~1.3 ms at p99 — well inside the +5 ms budget (~3.7× margin).** No raw-byte-relay fallback needed.

> **Paired drishti PR** [srid/drishti#85](https://github.com/srid/drishti/pull/85), re-pinned to this branch's final HEAD, green on both platforms. The `@kolu/surface*` changes are additive at the module level — but the gate did its job: re-pinning surfaced that the gauntlet's `RemoteMirrorSession` publish had *narrowed* the receptacle's client-input variance (a mapped-type generic collapsing `C` to `AnyContractRouter`), breaking drishti's `pin()` typecheck though build and runtime stayed green. Fixed by restoring the contravariant `unknown` on the role so **drishti compiles unchanged** (`0bda667` — verified by running drishti's own `tsc` against these sources: red before, green after), with an in-kolu `.test-d.ts` pinning the variance so it can't silently recur.
>
> **Deploy note:** the nix build launches padi via `systemd-run --user`; a bare non-systemd box needs `KOLU_PADI_SPAWN=detached` (twin of `KOLU_KAVAL_SPAWN`). The `activity` stream stays quiet — it has no consumer yet; its live backing is a consumerless follow-up.

### Pre-gauntlet review round

A fresh-context architecture review (5-axis, each finding skeptic-verified) cleared the core design and surfaced edge-and-honesty fixes, all landed before the gauntlet: **B1** — claim padi's pid gate *first*, before the import/kaval-recycle/manifest side effects, so a lost race touches nothing; **B2** — the drain persists through the empty-preserve receptacle, so it can't null a parked-only/empty session; **M2** — the dead in-process `restartLocalDaemon` path (zero callers after the cutover) is deleted, the drain is the sanctioned restart; plus the state-root default made env-insensitive, the `@kolu/padi` seal's dep-cone made recursive, the retired store-setters internalized, the re-serve's value channels switched to `drop-oldest` (an abort would strand the mirror), and the one-shot import taught not to clobber a padi that already holds its own session.

### Since the review — gauntlet, CI, and one honest-state fix

The `/be-review` gauntlet ran (lens-debate · codex-debate · simplify · code-police — see the three review comments). The standout structural fix published `RemoteMirrorSession` so the padi binding plugs into the re-serve receptacle without an `as unknown as` cast; codex caught **four real cutover regressions** (a scratch-dir leak, a drain that swallowed errors into a false success, a preview route masking padi/link faults as 404, and a lost `TERM_PROGRAM_VERSION`), all fixed. CI then surfaced one regression the tests had missed: after the cutover, a padi drop left the client's warming/degraded canvas **frozen** — the re-serve holds kaval's status stale while padi is unbound, so the client showed a live-looking world during a restart (a #1034 "never lie about state" bug). kolu-server now publishes its **own** honest `padiLink` binding state (a kolu-internal koluSurface cell — no `@kolu/surface*` / padiSurface contract change), and the client folds it into the canvas, so any padi drop shows an honest connecting/warming state across the whole window.

### Since the gauntlet — four behavior gaps the two-platform CI + an adversarial review closed

Running the NixOS-VM adoption tests on real systemd (the one adoption path the Playwright e2e can't reach) plus a focused adversarial review of the post-gauntlet commits surfaced four genuine gaps — each closed here to the coordinator's sanctioned semantics, none papered to green:

- **"Restart kaval" recycles kaval again** (`4ea845`, `43aa24`). Post-cutover the button had been re-pointed at the padi drain, so it restarted *padi* and merely *adopted* a live kaval — a stuck-but-alive kaval was never recycled, contradicting the label. The W1 **capture → park → recycle** is restored as a first-class `padiSurface.lifecycle.recycleKaval` procedure — padi's own supervisory recycle arm, so **padi stays up** — and the button calls that, not `daemon.restart`. _This resolves the review's M2 the right way: `restartLocalDaemon`'s no-shrink invariants regain a live production caller instead of being deleted as orphaned._ The adversarial review then caught a real hole in the restored path — a restart while a restore was *already pending* could shrink the saved session — so the capture now **merges** the parked and live sets (id-keyed, live wins) and no restore-pending terminal is ever lost.
- **Newest-wins convergence is now wired** (`0cb1c8b2`). A contract-skewed padi was only ever *refused* (left degraded until a human clicked restart), which meant W2.3's first `padiSurface` minor bump would land **every** deploy degraded. A **newer** binder now drains the older padi through the frozen control core and respawns its own closure (kaval survives, session intact); an **older** binder still refuses and never drains — the asymmetry that keeps convergence monotonic with no livelock.
- **The kaval "update available" nudge fires again** (`56e0431a`). The cutover moved the currency check into padi's process, which didn't carry `KAVAL_BUILD_ID` — so `expected` was empty and the nudge silently died. padi's wrapper now bakes it (padi owns kaval; the binder env-forward covers the from-source path).
- **The adoption VM suite matches the new topology** (`e58e06`, `a80422`, `4276ab`). The linux-only NixOS-VM tests were reading a pre-cutover per-port gate; rewritten for the digest-keyed two-level topology — adopt (padi+kaval gates unchanged across a kolu-server restart), skew (kaval-contract recycle), currency (the nudge bites).

### From live macOS testing — caught before merge

Deploying the branch to a real machine and exercising it surfaced several issues CI's simpler scenarios hadn't — including the single most important one, the upgrade path itself:

- **Split-restore correctness on recycle** (`518c5bf3`, `316863b5`). Recycling kaval with a split open produced spurious "Failed to set parent" toasts and restored the split *hidden* — two independent bugs on pre-existing W1 machinery that only the new in-session recycle triggers. The client's list-reconcile mistook the supervised drain for user-closes and issued authoritative `setParent` writes — now **gated on the daemon being connected** (promote-on-departure still fires for a genuine user-close). The `viewSeeded` latch skipped re-hydration on an in-session restore — now **reset when a restore runs**, so the restored split's active sub-tab is re-seeded. And a latent hole an adversarial review surfaced — a client write racing the capture could un-parent a *parked* record — is closed by making **parked records immutable** server-side (`requireMutableTerminal` rejects a parked target across every chrome mutation): the session-safety no-non-user-writer invariant, extended from the blob to the records. A red-pre-fix / green-post-fix e2e reproducer rides along.
- **A file-drop test flake the cutover exposed** (`c2ea33e0`). The state-root cutover lengthened the terminal scratch path, so a dropped filename now wraps past column 80 before xterm refits — and the test's buffer reader joined wrapped rows with a newline, splitting the filename so the assertion missed it *despite the name being on screen*. The reader now rejoins `isWrapped` continuation rows — deterministic (no timeout bump), verified 10/10 under load, and it hardens every screen-state assertion in the suite.
- **Uptime in the Padi and Kolu dialogs** (`96cac418`). padi already stamps an honest boot time in its control-core hello; it — and kolu-server's own start time — now render as uptime in their info dialogs, mirroring Kaval, routed through a kolu-internal cell (no contract change).
- **The upgrade path itself — adopt the pre-W2.2 kaval, don't leak it** (`3d4dfe85`…`0e51baeb7`). Deploying W2.2 *over a running pre-W2.2 install* exposed the real gap: pre-W2.2 keyed kaval by listen port (`kaval-<port>`), W2.2 by a state-root digest, so the first W2.2 boot spawned a *fresh* kaval and **leaked** the running old one — stranding the live terminals with new ids. But the adoption promise is that PTYs survive a deploy (kaval recycles only on a contract skew), so an upgrade is just another deploy. The binder now hands padi its own-port legacy socket hint on every spawn; padi probes the digest gate first, then the hint, and **adopts** a compatible legacy kaval — the PTYs survive with **no restore card** (the imported records reconcile straight onto the live PTYs) — recycling only on a skew. _Spawn is always digest-keyed_, so the first recycle, Restart, crash, or **reboot** converges the address away: a bounded, one-time migration proven by three VM arms — adopt · recycle-converge · reboot-converge (the reboot arm pins that kaval, with a tmpfs gate and live PTY fds, can't outlive a host reboot). A standalone padi never adopts a stray port kaval.
- **See every daemon, and read the Padi contract** (`b458558d4`, `ec92968f`). The above leak was found via a CLI error, not the UI — so the Kaval and Padi dialogs now enumerate *every* running kaval/padi on the host (read-only), badge the one kolu is using, honestly label an adopted legacy-address kaval ("pre-padi address · converges on next restart or reboot"), and flag genuine strays. The Padi chip also shows its `PADI_SURFACE_VERSION`, parity with Kaval's contract version.
- **A mark for படி** (`d15b55d2`). Kolu and Kaval are named for Tamil words and their marks embody the meaning; padi's mark is now a **staircase in profile** (படி = stair/step — the step the workspace rests on), replacing the placeholder supervision-bracket glyph.

Refs #951 (padi plan of record — W2.2).

### Try it locally

```sh
nix run github:juspay/kolu/W2.2
```

_Generated by [`/be`](https://github.com/srid/agency) on Claude Code (model \`claude-opus-4-8\`)._


srid:
## Evidence

Captured on an ephemeral `pu` box running the **nix-built** kolu at this branch's HEAD (`39f7167`). Because the nix build bakes `KOLU_PADI_BIN`, kolu-server would launch padi via `systemd-run --user` — which a bare box can't do — so it was deployed with **`KOLU_PADI_SPAWN=detached`** (+ the existing `KOLU_KAVAL_SPAWN=detached`). **padi bound for real:** `$XDG_RUNTIME_DIR/padi-<digest>/padi.sock` + `padi.pid` present, and a distinct padi node process (pid `4273`, RSS ~150 MB) sat between kolu-server (pid `4235`) and its kaval (pid `4304`, padi's child). The "before" is the nix-built kolu at `master` (`7e7a5b4`) on a second box — server + kaval, **no padi socket**.

### 1. The one sanctioned pixel delta — the 3-process memory readout

The **Diagnostic Info** dialog (Debug → Diagnostic info) now lists **three** server-side processes' resident memory. `master` shows two rows (Server RSS + kaval RSS); W2.2 adds the **padi RSS** row in between.

| Before — `master` (2 processes) | After — W2.2 (3 processes) |
| --- | --- |
| before | after |

The identity rail's Kolu chip tooltip carries the same figures live — `server RSS 154 MB · padi RSS 150 MB` — alongside the kaval chip's `RSS 112 MB`. _Coarse-pointer check:_ on mobile/touch, kolu drops to a focused single-terminal layout with **no identity rail at all** (desktop chrome — same as `master`); the three RSS rows in the Diagnostic dialog are layout-independent and stay reachable via the command palette.

### 2. Kill padi mid-typing → the tile recovers (done-criterion c)

Typed into a live terminal, then `kill -9`'d the padi process. kolu-server flips the binding to `degraded` (`daemon connection closed mid-session`); the mid-chain death is non-retriable, so `terminalAttach` **ends downstream and the browser re-subscribes end-to-end** (the outer re-subscribe loop). A fresh padi rebinds (new pid `4672`); kaval was detached so it **survived** (pid `4304` unchanged) and the new padi adopted it — the pty's scrollback comes back whole, the **"1 terminal reattached"** toast fires, and typing works again.

kill-padi

▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=w22-kill-padi.mp4

### 3. Warm restart — restart kolu-server, padi survives (done-criterion b)

With the same terminal open, restarted **kolu-server** (not padi). padi (pid `4672`) and kaval (pid `4304`) both survived; the new server logged **`adopted a surviving padi (its PTYs are preserved)`** — same pid, registry warm, not re-derived. The client shows kolu's redeploy `App updated → Reload` prompt; after reload the tile re-attaches **already-whole** (full scrollback intact) and typing continues.

warm-restart

▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=w22-warm-restart.mp4

srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `7e7a5b4f14f4`

Independent findings: lowy=6, hickey=8

### Applied (9)
- `lowy-1` Re-serve session receptacle is typed to the concrete HostSession class, so the padi binding must lie with `as unknown as` — commit `5725e8d01`
- `lowy-2` The honest three-way process-RSS schema is duplicated byte-for-byte across two packages instead of living in the shared leaf both already import — commit `3b3550642`
- `lowy-3` NODE_OPTIONS dev-flag scrub duplicated across the two daemon drivers instead of owned by the spawn-mechanism receptacle — commit `cb372a14f`
- `lowy-4` adoptOrEnsure and adoptOrSpawnOrRefuse duplicate their adopted/unreachable/no-survivor arms; only the skew branch differs — commit `f835223cb`
- `hickey-1` reServeSurface's session role is fragmented across HostSession + a re-declared ReServeSession, bridged by an unchecked double-cast — (uncommitted)
- `hickey-2` scrubNodeOptions duplicated byte-identically across the package seal — (uncommitted)
- `hickey-3` Comments still say session/activityFeed storage 'stays kolu-server-side until W2.2' — W2.2 is this PR — commit `07814e781`
- `hickey-4` ProcessRss honest three-way union declared twice, byte-identical, guarded only by a comment + one fold typecheck — (uncommitted)
- `hickey-6` padi sets its own pid through a setter named for kolu-server — commit `6ffef6ec2`

### Agreed — no change (5)
- `lowy-5` readPadiMemoryOnce re-dials the raw padi client every tick, bypassing the re-serve mirror that already folds processMemory (packages/server/src/index.ts:401)
- `lowy-6` Electricity map — receptacles, volatile implementations, and where the wires are exposed (packages/surface-daemon/src/daemonMain.ts:87; packages/surface-daemon-supervisor/src/endpoint.ts:236; packages/surface-nix-host/src/reServeSurface.ts:122; packages/surface/src/server.ts (cellHandlers.forward))
- `hickey-5` Non-overlapping unref'd periodic-sampler harness and the 5s cadence constant duplicated (packages/padi/src/memorySampler.ts:29)
- `hickey-7` Opaque-client navigation reached structurally in three separate cast sites (packages/surface-nix-host/src/reServeSurface.ts:282)
- `hickey-8` Reconnect/re-dial loop lives in the binding adapter, not the supervisor spine (packages/server/src/padiBinding.ts:386)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 round(s) · codex reviewed at `xhigh` reasoning effort · base `7e7a5b4f14f4`
### Round 1

**codex** — approved: `false`

The branch has several real regressions around padi process ownership: cleanup did not move with the state it owns, a restart/drain path now reports success after swallowing failures, and a preview route masks padi/link failures as ordinary 404s. I also found two observability/identity regressions that should be corrected before approval.

Findings:
- `F1` · major · open — `ensureKoluRoot()` moved into the padi process, but the matching `shutdownCleanup()` registration did not move with it. The old server path registered cleanup on process exit; the new server explicitly says padi owns scratch cleanup, yet `runPadiDaemon` only creates the root and never cleans it. Normal padi drains/restarts will leave `${XDG_RUNTIME_DIR}/kolu-<pid>` scratch state behind, including terminal scratch/upload/init files. (packages/padi/src/daemonMain.ts:147)
- `F2` · major · open — `drainBoundPadi()` catches and ignores every error from `surface.control.core.drain()`. That makes the restart RPC report success even when the request was not delivered, the connection was stale, or the padi handler failed before draining. This violates the repo's fail-fast rule and can give the user a success toast while the bound padi is still running. (packages/server/src/padiBinding.ts:428)
- `F3` · major · open — The iframe preview route catches all errors from `repoRootForTerminal()` and returns `404 terminal has no repo`. The padi API already returns `repoRoot: null` for the real no-repo case, so exceptions here are operational failures such as padi link loss, protocol errors, or unexpected handler failures. Collapsing those to a 404 hides the real problem. (packages/server/src/index.ts:335)
- `F4` · major · open — Spawned PTYs no longer receive the kolu app/server version in `TERM_PROGRAM_VERSION`. `runPadiDaemon` sets `setSpawnServerVersion(opts.spawnVersion || currentPadiCommitHash() || "dev")`, but the server-side padi launch path does not pass `spawnVersion`, and the CLI has no flag for it. Existing comments/tests describe `TERM_PROGRAM_VERSION` as the kolu server/package version, not the padi commit hash, so this is a behavior regression in terminal identity. (packages/padi/src/daemonMain.ts:145)
- `F5` · minor · open — The server now says padi has its own `--verbose`, but the padi CLI does not define one, and the server's `--verbose` no longer raises padi logging. In systemd mode especially, padi will stay at its default log level, making the split-process path harder to debug than the old in-process padi. (packages/server/src/index.ts:104)
- `F6` · minor · open — `readPadiMemoryOnce()` catches all process-memory read failures and returns `null`, which the sampler reports as `absent`. If a padi client exists but the RPC/schema/read fails, the diagnostics panel will say padi/kaval are absent instead of showing the available error state. (packages/server/src/index.ts:431)


**claude** — Fixed the four genuine regressions (scratch-root leak, false-success drain, 404-masked link fault, dropped app version) and the memory-read collapse; on `--verbose` I fixed the inaccurate comment but disputed the debuggability claim (padi already logs every level).

- `F1` **fixed** — Real regression, agreed. The old server registered `process.on("exit", shutdownCleanup)` (base `packages/server/src/index.ts:225`); the cutover moved `ensureKoluRoot()` into padi (`packages/padi/src/daemonMain.ts`) but not the wipe, and `shutdownCleanup` was left exported-but-uncalled (only referenced in a comment at `server/src/index.ts:131`). So every padi drain/restart leaked `${runtimeRoot}/kolu-<padi-pid>` (shell rc, scratch, upload/init). Restored it: imported `shutdownCleanup` and registered `process.on("exit", shutdownCleanup)` right after `ensureKoluRoot()` in `runPadiDaemon` (`packages/padi/src/daemonMain.ts:150`) — after `setDaemonProcessId` (so `koluRoot()` resolves) and after the lost-gate-race early return (so only the winner that created the dir wipes it). Same hook the old server used; SIGKILL/power-loss still fall through to the XDG logout-wipe backstop.

- `F2` **fixed** — Agreed the blanket `.catch(() => {})` (old `padiBinding.ts:428`) was too broad and the code comment's premise ("either way the drain reached the core") is false for a stale link or an `onDrain` that throws before persisting — both reject WITHOUT the drain landing, yet reported success. This is the repo's own `caught-error-must-not-collapse-to-empty` / fail-fast rule, and it was NOT among the author's deliberate decisions (which defend the empty-preserve session receptacle, not this swallow). Rewrote `drainBoundPadi` (`packages/server/src/padiBinding.ts:445`) to use the SOCKET CLOSE as ground truth — the same signal the endpoint reads to flip to `degraded`: a clean resolve = drain landed; a rejection is accepted only if the socket then closes within a teardown ceiling (`DRAIN_TEARDOWN_CEILING_MS = 2000`), otherwise the original error is rethrown so the "restart" verb reports failure honestly. Verified against the real wire-drain integration test (`padiBinding.test.ts` "drain ACROSS THE WIRE") — passes (the log confirms the socket closes → the observer resolves).

- `F3` **fixed** — Agreed. padi's `preview.repoRootForTerminal` handler is `({ input }) => ({ repoRoot: snapshotFor(id)?.git?.repoRoot ?? null })` (`packages/padi/src/servePadi.ts:479`) — it returns `repoRoot: null` for an unknown/unmapped terminal and NEVER throws for the no-repo case. So the old `catch → 404` (`server/src/index.ts:335`) could only ever fire on an operational fault (padi link loss mid-read, a protocol error), masquerading it as an ordinary missing-file 404. Split it: the `if (!repoRoot)` line still returns 404 for the genuine no-repo case; the `catch` now logs and returns 503 ("padi link fault…"). The prior comment's claim that the retired in-process route 404'd here was wrong for the same reason — that route's `undefined` was the no-repo value, not a throw.

- `F4` **fixed** — Agreed this is an incomplete cutover, not a deliberate change. The old server did `setSpawnServerVersion(serverVersion)` (the kolu app semver); the new `runPadiDaemon` does `setSpawnServerVersion(opts.spawnVersion || currentPadiCommitHash() || "dev")` (`daemonMain.ts:149`) but NOTHING forwarded `spawnVersion` — so spawned PTYs' `TERM_PROGRAM_VERSION` silently became padi's git hash (or "dev" off-nix). The option's own docstring says "kolu-server forwards the app version for byte-identity", and `ptyHost/index.ts:47` + `hostname.ts:22` both describe this env as the kolu app version — the forwarding was just unwired. Threaded it: `index.ts` passes `spawnVersion: serverVersion` → `ensurePadiBinding` (`EnsurePadiBindingOptions.spawnVersion`) → `localPadiDriver` → `resolvePadiLaunch` pushes `--spawn-version <ver>` → `bin.ts` parses it and forwards to `runPadiDaemon`. A standalone `padi` (no binder) still falls back to its own commit via the existing `||` chain — matching the documented default.

- `F5` **partial** — Fixed the valid half, disputed the rest. VALID: the server comment (`server/src/index.ts:104`) claimed "padi runs in its OWN process now (with its own `--verbose`)" but `bin.ts`'s `parseArgs` defines no `verbose` — the comment was inaccurate. Corrected it to state the actual mechanism. DISPUTED the substantive claim that padi "will stay at its default log level, making the split-process path harder to debug": padi's `stderrLogger` (`packages/padi/src/bin.ts:49`) emits EVERY level unconditionally — `emit("debug")` always writes, there is no level filter — so padi already logs at full detail. There is nothing to "raise" and adding a `--verbose` knob would be a no-op (and a knob the fail-fast rule discourages). So no flag added; comment now matches reality.

- `F6` **fixed** — Agreed, and it contradicts the schema's own documented intent. `ProcessRss` (`packages/terminal-workspace/src/schema.ts:237`) has a distinct `error` status EXPRESSLY so "a failed read never renders identically to `absent`" (it cites `caught-error-must-not-collapse-to-empty`). But `readPadiMemoryOnce` (`server/src/index.ts:431`) returned `null` on every read failure, which the sampler folds to `absent` for BOTH padi and kaval — so a read that throws through a LIVE client rendered as "no process". Fixed: `null` is now reserved for the true no-client case (padi genuinely down → honest `absent`); a read that throws, or a live subscription that yields no frame, logs and returns `PADI_MEMORY_READ_ERROR` (`{ padi: {status:"error"}, kaval:{status:"error"} }`). The injected-deps sampler unit tests are unaffected (they drive `padiMemory` directly) and still pass.

commit: `e5095bb708f5dfe1efb7b9b5d85eae2ab0c92a5d`


### Round 2

**codex** — approved: `false`

F1-F4 are resolved after verification, and the user-visible part of F6 is resolved: scratch cleanup moved into padi, drain no longer blindly swallows failures, preview link faults now surface as 503, spawn-version is forwarded, and memory-read failures no longer render as absent. I do not accept the F5 dispute because padi still has a filtered pino logger path separate from the unconditional bin stderr logger. I also found one new minor regression introduced by the F6 fix: actual read failures are logged at warn instead of error.

Findings:
- `F1` · major · resolved — Resolved. `runPadiDaemon` now creates the kolu root and registers `shutdownCleanup` on process exit in the padi process, so normal padi drain/restart exits remove the per-process scratch root. (packages/padi/src/daemonMain.ts:151)
- `F2` · major · resolved — Resolved. `drainBoundPadi` now rethrows drain failures unless the socket closes within the teardown ceiling, and `connectPadi`'s `onClose` wrapper replays an already-observed close, so the expected teardown race is covered. (packages/server/src/padiBinding.ts:452)
- `F3` · major · resolved — Resolved. The preview route now reserves 404 for `{ repoRoot: null }` and logs/returns 503 for thrown padi link or handler faults. (packages/server/src/index.ts:341)
- `F4` · major · resolved — Resolved. `serverVersion` is now passed through `ensurePadiBinding` to `resolvePadiLaunch`, emitted as `--spawn-version`, parsed by padi, and used by `runPadiDaemon` for spawned PTYs' `TERM_PROGRAM_VERSION`. (packages/server/src/index.ts:192)
- `F5` · minor · open — Still open. The comment was corrected away from a nonexistent padi `--verbose`, but the substantive dispute is wrong: padi does not log every level through one unconditional logger. `bin.ts`'s `stderrLogger` emits every level for the daemon spine logger, but `runPadiDaemon` also wires `padiLog` into the padi surface deps and many padi modules import `./log.ts` directly. That pino logger filters at `process.env.LOG_LEVEL ?? "info"`, and the server's `--verbose` only sets the server logger level; it does not make those padi domain debug logs visible, especially on the systemd-run path where only explicit `daemonEnv` vars are forwarded. (packages/server/src/index.ts:104)
- `F6` · minor · resolved — Resolved. `readPadiMemoryOnce` now returns `null` only for the true no-current-client case and returns an explicit error-shaped padi/kaval reading when a live-client read fails or yields no frame. (packages/server/src/index.ts:434)
- `F7` · minor · open — New this round. The F6 fix logs a caught padi memory read failure at `warn`, but this is an actual failed read through a live client. The repo's `errors-must-log-at-error` rule says failed I/O, failed queries, unexpected exceptions, and callback throws must log at error level, not warn. The no-frame live-client anomaly at line 456 is the same class of operational anomaly. (packages/server/src/index.ts:463)

_codex on the rebuttal:_ F5: I accept the narrow comment fix, but I do not accept the dispute that padi already logs every level. The unconditional `stderrLogger` is not the whole padi logging surface: `runPadiDaemon` wires `padiLog` from `packages/padi/src/log.ts` into padi surface deps, and padi modules import that pino logger directly. That logger defaults to `info`, so server `--verbose` still does not enable padi domain debug logs. F6: the user-visible absent/error collapse is fixed, but the fix introduced warn-level logging for actual read failures; I tracked that separately as F7.


**claude** — Conceded F5 (my round-1 dispute was wrong): restored the pre-cutover `--verbose → padi debug` parity by forwarding `LOG_LEVEL=debug` to padi's process; raised the two F7 live-client read anomalies to `log.error`.

- `F5` **fixed** (conceded) — codex was right and my round-1 dispute was wrong. There are TWO padi loggers: the daemon-spine `stderrLogger` (`packages/padi/src/bin.ts:53`) which emits every level unconditionally, AND padi's OWN pino DOMAIN logger (`packages/padi/src/log.ts:14`, `const level = process.env.LOG_LEVEL ?? "info"`) that `memorySampler.ts`/`activity.ts`/`transcript.ts`/`session.ts` import directly and that `runPadiDaemon` wires as `padiLog` into `buildPadiSurfaceDeps` (`daemonMain.ts:195`). My claim "padi already logs at full detail" only held for the spine logger; the domain logger filters at info. Worse, this is a genuine REGRESSION: the pre-cutover server explicitly did BOTH `log.level = "debug"` AND `padiLog.level = "debug"` (base `packages/server/src/index.ts:145-150`, comment: "moved-code debug lines are silently dropped" otherwise). The cutover kept only the server half, and `daemonEnv` never forwarded `LOG_LEVEL`, so on the systemd-run path padi's domain debug was unreachable. Fix (the cross-process twin of raising the logger in place): threaded `verbose` through `EnsurePadiBindingOptions` → `ensurePadiBinding` → `localPadiDriver` → `daemonEnv` (`packages/server/src/padiBinding.ts:210-217`), which now sets `LOG_LEVEL=debug` when verbose (else forwards an explicit operator `LOG_LEVEL`, closing a second forwarding gap for free). `index.ts:200` passes `verbose: argv.flags.verbose`; the misleading comment at `index.ts:104` is rewritten to state the actual mechanism. Reused the existing source of truth (`LOG_LEVEL`, which both `log.ts` files already read) — not a new knob.
- `F7` **fixed** — agreed. Both live-client memory-read anomalies are genuine failed reads, not degraded-but-recoverable states, so `.agency/code-police.md` errors-must-log-at-error applies. Changed the no-frame anomaly (`packages/server/src/index.ts:459`) and the caught read-failure (`packages/server/src/index.ts:468`) from `log.warn` to `log.error`. The no-client path stays `null → absent` (an expected-absent condition, correctly unlogged).

F1–F4 and F6 were resolved by codex in round 2 (no changes this round). Gate: `just check` green (typecheck all packages + biome lint 1055 files clean); `just fmt` clean; server memorySampler + iframePreviewRoute tests 43 passing.

commit: `a751ef0d75cceadf56539b27e77b85c7a61e4e1b`


### Round 3

**codex** — approved: `true`

All previously raised findings are resolved in the current tree. I verified the new `LOG_LEVEL` path from `index.ts` through `ensurePadiBinding` into `daemonEnv`, including systemd-run env forwarding, and the live-client memory read anomalies now log at `error`. I did not find any new regression introduced by this close-out round.

Findings:
- `F1` · major · resolved — Resolved. Padi registers `shutdownCleanup` after creating its per-process kolu root, so normal padi drain/restart exits clean up scratch state in the owning process. (packages/padi/src/daemonMain.ts:151)
- `F2` · major · resolved — Resolved. `drainBoundPadi` now accepts a rejected drain only when the underlying socket closes within the teardown window; otherwise it rethrows the original failure. (packages/server/src/padiBinding.ts:452)
- `F3` · major · resolved — Resolved. The iframe preview route now distinguishes real no-repo results from thrown padi/link faults and returns 503 for operational failures. (packages/server/src/index.ts:341)
- `F4` · major · resolved — Resolved. The server forwards `serverVersion` as padi's `spawnVersion`, preserving the pre-cutover `TERM_PROGRAM_VERSION` contract for spawned PTYs. (packages/server/src/index.ts:196)
- `F5` · minor · resolved — Resolved. The server now forwards verbose intent to padi as `LOG_LEVEL=debug`, while also forwarding an explicit operator `LOG_LEVEL` when not verbose. That covers padi's filtered pino domain logger as well as the daemon-spine logger. (packages/server/src/padiBinding.ts:226)
- `F6` · minor · resolved — Resolved. `readPadiMemoryOnce` reserves `null` for the true no-client case and returns explicit error-shaped readings for live-client read failures or no-frame anomalies. (packages/server/src/index.ts:434)
- `F7` · minor · resolved — Resolved. The live-client no-frame anomaly and caught memory-read failure now log at `error`, matching the repo's `errors-must-log-at-error` rule. (packages/server/src/index.ts:466)

_codex on the rebuttal:_ F5: CLAUDE conceded the dispute and fixed the actual cross-process logger path by forwarding `LOG_LEVEL=debug` to padi when server `--verbose` is set. I verified the wiring through `ensurePadiBinding`, `localPadiDriver`, `daemonEnv`, and `survivableSpawnDriver`. F7: fixed as requested; both live-client memory-read anomalies now use `log.error`. No disputes remain.

srid:

7 findings on the W2.2 diff (`7e7a5b4f1…`), each independently source-verified — **all 7 actioned** (Pass 3 elegance skipped; the simplify track already ran over this tree).

### Rules
- **dry-rule-of-three** · `confStores.ts` — the three identical `requirePadi*Store` getters now fold behind a generic `requireStore<T>(store, name)` factory (same fail-fast throw). `d0f659a8`
- **no-dead-code** · `surface-daemon-supervisor/restart.ts` — the `restart`/`serializeRestart`/`RestartSteps` machinery was orphaned when this diff deleted its sole caller (`restartLocalDaemon`); deleted it + the re-export + its test block (grep-confirmed no live caller). `f8219846`
- **prefer-ts-pattern** · `padiBinding.ts` — `projectEndpointStatus` (nested ternary) + `onEndpointStatus` (if/else-if) rewritten with `match(…).with(P.union(…)).exhaustive()`; behavior identical. `8f4a17a3`
- **callback-fanout-guarded-at-funnel** · `padiBinding.ts` — `fire()` iterated the `onState` listener Set unguarded; wrapped each callback (a throwing listener no longer aborts the fan-out; the throw logs at `error`). `60e114f2`

### Fact-check
- **subscriber-registry leak** · `surface/server.ts` — the `overflow: "abort"` branch closed the subscriber but never `removeSub`'d it (a rejected `next()` doesn't trigger `iterator.return()`); added the removal + a `subscriberCount() === 0` assertion. `bb09bf57`
- **false `--help` default** · `padi/bin.ts` — USAGE claimed `$XDG_STATE_HOME/padi`, but the default is HOME-only and ignores `$XDG_STATE_HOME`; corrected to `$HOME/.local/state/padi` + the real `KOLU_PADI_STATE_DIR`/`--state-root` overrides. `f154dd0b`
- **stale padi memory readout** · the 3-process rail showed a frozen padi RSS for up to 5s after padi dropped (the #1568 staleness class — closed for kaval, left open on padi's row); drove the sampler off the binding session's `onState` so a disconnect force-samples `absent` at once. `ba5ef611`

srid:
## Evidence

_Re-captured on the **final** HEAD (`0e4cce710`) — this **supersedes** the [earlier Evidence comment](https://github.com/juspay/kolu/pull/1664#issuecomment-4874109706), which was taken on an older SHA (`39f7167`) before two changes this PR now ships: **Padi in the version rail** and the **honest padi-drop canvas** (`padiLink`)._

Captured on an ephemeral `pu` box running the **nix-built** kolu at this branch's HEAD. Because the nix build bakes `KOLU_PADI_BIN`, kolu-server would launch padi via `systemd-run --user` — so it was deployed with **`KOLU_PADI_SPAWN=detached`** (+ the existing `KOLU_KAVAL_SPAWN=detached`). **padi bound for real:** `$XDG_RUNTIME_DIR/padi-<digest>/{padi.sock,padi.pid}` present, a distinct **padi** node process (`packages/padi/src/bin.ts`, RSS ~155 MB) sitting between kolu-server and its adopted **kaval** (`kaval-<digest>/pty-host.sock`, RSS ~114 MB); the server log shows the `padi re-serve` cutover path live. The **before** is the nix-built kolu at `master` (`7e7a5b4`) on a second box — server + kaval only, no padi.

### 1. Padi joins the version rail — the headline

The rail now reads **kolu-server → padi → kaval**, left to right. A new **Padi** chip with its own teal mark and a status pip painted **live from padi's binding state** sits between Kolu and Kaval.

| Before — `master` (2 chips) | After — W2.2 (3 chips) |
| --- | --- |
| before | after |

The rail in its real desktop context, and the **Padi details dialog** (click the chip → mark, live connection state, RSS):

rail in context

padi dialog

### 2. The 3-process memory readout — one honest RSS per process

padi's RSS now rides its **own chip** (`padi connected · RSS 156 MB`), and the **Diagnostic Info** dialog (Ctrl+K → Debug → Diagnostic info) lists **three** server-side processes' resident memory in the Session section — **Server RSS · padi RSS · kaval RSS**:

diagnostic 3-process RSS

(`master` shows only two rows — Server + kaval — and the 2-chip rail above; the third row/chip is the delta.)

### 3. Kill padi mid-typing → an honest warming canvas, then whole re-attach (done-criterion c + the `padiLink` fix)

Typed a line into a live terminal, then `kill -9`'d the **padi** process (pid from `padi.pid`). The live terminal is **replaced by a neutral warming canvas** and the **Padi pip flips off-green** — an honest "the link is down right now", **not** a frozen live-looking world. The server logs `daemon connection closed mid-session — degraded` and `terminalAttach: upstream link died — ending downstream`; padi's supervisor rebinds a **fresh padi**, the **detached kaval survives and is adopted**, the pty re-attaches whole, the **“1 terminal reattached”** toast fires, and the mid-typed line — preserved across the drop — then runs.

kill padi → warming → re-attach

The warming canvas itself (neutral grid replacing the live terminal; Padi pip red):

honest warming canvas

▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=w22-kill-padi-v2.mp4

> **One honesty nuance worth flagging:** on a **padi-only** drop the big warming label reads **“Connected”** — because it tracks **kaval's** state (kaval never went down, only the padi hop). The unambiguous honest signals here are the **neutral canvas** (the live terminal is hidden) and the **Padi pip going red**; the centered label follows the daemon, not the link.

### 4. Warm restart — restart kolu-server, padi survives (done-criterion b)

Restarted **kolu-server** (not padi). padi (pid `4472`) and kaval (`4013`) both survived; the new server logged **`adopted a surviving daemon (its PTYs are preserved)`**. The client shows kolu's **“App updated → Reload”** stale-tab prompt (the rail pips honestly drop to `unknown` — no fake green); after Reload the tile re-attaches **already-whole** — the entire scrollback intact — and typing continues.

warm restart → whole re-attach

▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=w22-warm-restart-v2.mp4

---

_Captured live with the chrome-devtools MCP against the nix-built server on a `pu` box (final `0e4cce710`); the before-rail on a second box at `master`. Clips transcoded with ffmpeg; boxes torn down._

srid:
## W2.2 — ready for review ✅

Green end-to-end, both platforms, **first try, no flakes**.

**Kolu #1664 @ `0e51baeb7`** — 34/34 checks. The NixOS-VM adoption suite (the one adoption path the Playwright e2e can't reach) passes all six VMs: **adopt · skew · currency · adoption-upgrade** (adopt-legacy + converge-to-digest) **· adoption-upgrade-reboot** (reboot-bounded). e2e **493/493** linux · **492/492** darwin. Unit suites (padi · server · client · supervisor · surface-nix-host) all green.

**Drishti #85**, re-pinned to this exact HEAD (`3c269bf`) — 20/20 both platforms, and **`ci::typecheck` is green**: the `surface.md` gate satisfied at final kolu HEAD (the gauntlet's `RemoteMirrorSession` publish had narrowed a consumer's input-variance and broken drishti's `pin()` typecheck though build/runtime stayed green — `0bda667` restores it so drishti compiles unchanged, with an in-kolu `.test-d.ts` pinning the variance).

### How it got here

- **Pre-gauntlet review** — 2 blocking + 3 major + 6 minor, all landed (gate-first boot, drain empty-preserve, dead-restart deleted, …).
- **`/be-review` gauntlet** (lens-debate · codex-debate · simplify · code-police — three comments above): lens consensus; codex caught **four real cutover regressions**; simplify + police cleanups.
- **Four post-gauntlet gaps** the two-platform CI + a focused adversarial review closed: Restart-recycles-kaval (+ the no-shrink merge the review caught), newest-wins convergence, the currency nudge, and the adoption-VM rewrite.
- **Live dogfooding** (deployed to a real machine) caught the rest — each fixed **reproduce-first**: split-restore correctness, a file-drop test flake the cutover exposed, uptime readouts, an all-daemons diagnostic, the padi staircase mark — and the headline: **the upgrade path itself**. Deploying W2.2 over a running pre-W2.2 install leaked the old port-keyed kaval; W2.2 now **adopts** it (live terminals survive, no restore card) and converges to digest on the first recycle/reboot — a bounded, one-time migration proven by three VM arms.

### Evidence
Core visible changes (3-chip rail · 3-process memory · honest padi-drop canvas · warm restart): https://github.com/juspay/kolu/pull/1664#issuecomment-4876852939 . The newer visible changes (staircase mark · all-daemons dialog · uptime · contract chip) are live in the dogfood deploy — happy to capture a fresh reel on request.

Not merged — **@srid makes the merge call.**

PR #1665, opened 2026-07-03
Padi W2.3: the dial carve, padi-tui, and the burial of the pulam era

srid:
The last local-padi phase. W2.2 gave padi a process and made kolu-server bind it; **W2.3 gives padi its own command-line tool, lights the activity stream that was deferred, and buries the pulam era it replaces** — one PR, three commit stages.

```
1. THE DIAL CARVE   connectPadi ─▶ @kolu/padi/dial   (shared by the binder AND padi-tui)
2. PADI-TUI IS BORN  wait · status · watch · create   + the activity stream's live backing
3. THE BURIAL        delete pulam, pulam-tui, terminalWorkspaceSurface, the legacy-conf fossils
```

### 1 — The dial carve (`8aad60058`)

`connectPadi` (state-root→digest→socket resolve · control-core handshake · typed skew refusal · the typed `padiSurface` client) moves out of the kolu-server binder into a new **`@kolu/padi/dial`** entry — the kaval precedent, a daemon's package owning the client-side dial kit its clients share. Supervision (the version ordering, drivers, adopt/spawn/refuse, the reconnect session, the re-serve) stays binder-only. The seal's allowed `@kolu/padi` entries grow by exactly `/dial`; the dial's client types are spelled from padi's own oRPC deps so the kit adds **no** `surface-nix-host` edge to padi's closure.

### 2 — padi-tui is born (`ef823ab4d`)

A small non-interactive CLI on padi's socket — kaval-tui's sibling, **pulam-tui's replacement**. Every verb is a thin wrapper over a `padiSurface` procedure that has existed since W1 (**no new wire members**):

- **`wait`** — the precise agent-state done-signal off the composed `terminals` collection: `--until` over the shared `agentBucket` buckets (`working`/`awaiting`/`waiting`), the two-phase stale-state loop (the mirror replays current state on connect, so `--until working` THEN `--until awaiting,waiting` is race-free), `--timeout` + fail-loud exit codes (0/2/3/130/1). **Replaces `pulam-tui wait`.**
- **`status`/`watch`** — one-shot snapshot / live follow; `watch` surfaces byte activity as `● busy`/`○ idle` transitions.
- **`create`** — incl. `--parent` (a split tile) and `--worktree <branch> -- <argv>` (git.worktreeCreate + create-with-cwd + sendInput — a worktree'd agent in one command), composed exactly as the canvas `useWorktreeOps` does.

Discovery lives in the dial kit (`resolveRunningPadiSocket`): `--socket` / `--state-root` / `$PADI_SOCKET` / autodiscover. padi now **stamps `$PADI_SOCKET`** (the `$KAVAL_SOCKET` twin) into every terminal it spawns, so `padi-tui` inside a kolu terminal reaches its owning padi flag-free — the `/kolu` loop runs `wait` with no flags.

The **`activity` stream's live backing** (`liveActivity.ts`): padi taps its kaval's per-terminal byte deltas into a live set, **lazily** — byte taps open only while the stream has a subscriber, so an unwatched padi pays nothing (an always-on per-terminal byte tap would be a perf anti-pattern). The client green dot is unchanged (it derives from its own attach bytes).

### 3 — The burial (`32594cffb`)

With `padi-tui wait` proven, the pulam era retires:

- **Delete** `packages/pulam` + `packages/pulam-tui` (zero importers) + their nix wiring; padi-tui replaces pulam-tui on PATH (home-manager + flake default).
- **Delete** the frozen `terminalWorkspaceSurface` (3.0) and its assemblers (`serveFsGit`/`serveTerminalWorkspace`/`socket`). `@kolu/terminal-workspace` survives as the library padi assembles — sensors · fold · schema · endpoint · agentProjection. The fs/git watcher deps move into padi's own `fsGitDeps.ts` typed against `padiSurface`; the 3 fs/git schemas relocate to `terminal-workspace/schema.ts`; `ActivityStreamDeps` becomes a local padi type. *kolu-server's `terminalWorkspace` serve had already died at **W1.R7 (#1652)** — `surfaces = {kolu, surfaceApp}` + padi; this deletes the now-consumerless const.*
- **Shed** the legacy-conf fossils: `server/src/state.ts` drops `session` + `activityFeed` (padi owns them since W2.2), carrying only `preferences`. The new **1.31.0 migration is backup-first** — it copies `config.json` aside to `config.json.pre-1.31-strip.bak` before stripping session/activityFeed and the orphan keys sleepingTerminals/lastPairedDaemon, so a direct pre-W2.2→W2.3 upgrade (kolu-server's ladder runs before padi's first-boot import) degrades to **manual-recovery, never data-loss**.

### Criteria

| # | criterion | status |
| --- | --- | --- |
| (a) | `padi-tui wait` drives the `/kolu` done-signal in a live agent loop | phase-1 (`--until working`) proven locally against a real opencode agent (full sensor→collection→mirror→predicate→exit); **full two-phase + evidence on a pu box is merge-blocking** |
| (b) | `create --parent` split tile + `create --worktree -- <agent>` in the canvas | `--parent` proven locally (split tile); worktree + canvas visibility → pu-box evidence |
| (c) | activity live: `watch` shows byte-activity; canvas green-dot client-side unchanged | proven locally (`● busy`/`○ idle`); green-dot derives client-side, untouched |
| (d) | no package serves/consumes `terminalWorkspaceSurface` (grep-clean) | ✅ the const + assemblers deleted; grep-clean |
| (e) | kolu-server conf carries no terminal keys; fresh boot writes none | ✅ `state.ts` = `{preferences}`; 1.31.0 strip-migration + test |
| (f) | flake eval + home-manager green with padi-tui on PATH; `/kolu` loop verbatim | ✅ `nix build .#padi-tui`; padiTuiPackage on PATH; `/kolu` skill → padi-tui two-phase |
| (g) | e2e green both platforms; UX byte-identical | CI pending; UX byte-identical **except** the sanctioned Inspector-tab CLI reference (pulam-tui → padi-tui) |
| (h) | the seal green (padi imports no app package; server imports only `@kolu/padi/{assembly,surface,dial,log}`) | ✅ seal test green |

> **Notes for review.** (1) Merged `origin/master` (#1669, the /architecture page + README topology sweep); resolved the disjoint README conflict and truth-passed the rows the burial obsoletes. (2) The `/debate` skill's `pulam-tui` line lives in the **remote `srid/llm-debate` apm package** — not editable in this repo; flagged upstream. (3) `just ai::apm` hit a pre-existing skill-name collision (remote deps shadow the local `kolu`/`test` skills); the correct local skill sources are hand-deployed to `.claude/`/`.agents/` and the collateral reverted — the collision itself is infra to resolve separately.


srid:

✅ **Consensus** after 2 round(s) · lowy + hickey · base `18ada384f004`

### Applied (4)
- `lowy-1` Dial kit hands out the combined wire; each consumer re-scopes to the "padi" sibling itself — commit `7bad93338`
- `hickey-1` Idle-cadence constant duplicated across padi and client, coherence held by a comment — commit `78ea41b9b`
- `hickey-2` Dial kit fragments the padi-scoped client — every consumer reconstitutes it with a repeated unsafe cast — (uncommitted)
- `hickey-3` Duplicated read/resolve idioms inside padi-tui — commit `dfa4ba2a0`

### Agreed — no change (5)
- `lowy-2` Cross-seal backup-path contract spelled as a bare literal in both the stripper and the importer (packages/server/src/state.ts:133 (writer) and packages/padi/src/importLegacy.ts:101 (reader))
- `lowy-3` daemonStatus.ts accretes boot-recorded rendezvous sockets that are not the daemon-status axis (packages/padi/src/ptyHost/daemonStatus.ts:55-72 (padiServeSocketPath + get/set))
- `lowy-4` fs/git host-read axis is served from two homes (streams vs procedures) (packages/padi/src/fsGitDeps.ts:64 (subscribe* streams) vs packages/padi/src/servePadi.ts:406-461 (fs.*/git.* procedures))
- `hickey-4` padi-tui and kaval-tui are two parallel TUI clients over one 'terminal-side CLI' concept (packages/padi-tui/src/read.ts:1)
- `hickey-5` PadiBindingSession fragments 'the bound padi's identity' into parallel nullable fields kept coherent by memory (packages/server/src/padiBinding.ts:592)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `18ada384f004`

### Round 1

**codex** — approved: `false`

I found several correctness issues that should be fixed before this branch is mergeable. The biggest one is a migration retry path that can overwrite the only full legacy backup. The new padi-tui path also has a real wait race, lossy command reconstruction, and activity/status failure modes that can silently report stale or partial state. I also found one stale package README left behind by the terminalWorkspaceSurface burial.

Findings:
- `F1` · blocking · open — The 1.31 strip migration overwrites `config.json.pre-1.31-strip.bak` every time it sees any legacy key. If the process dies after copying the full file and after deleting one legacy key, but before all deletes/migration bookkeeping complete, the next boot reruns this migration, still sees the remaining legacy keys, and overwrites the full backup with a partially stripped file. That breaks the deliberate zero-loss direct pre-W2.2 -> W2.3 upgrade path, because padi may later import from a backup that already lost `session` or other legacy data. (packages/server/src/state.ts:133)
- `F2` · major · open — `awaitAgentState` only reports `gone` when `watchTerminals` receives an `onRemove` event. If the terminal exits after `cmdWait` resolves the id but before the mirror's first keys snapshot, the target is absent from the initial collection and no remove event is emitted. With the default no-timeout wait, `padi-tui wait` can hang forever even though the terminal already disappeared, contradicting the documented "terminal exits first -> exit 3" behavior. (packages/padi-tui/src/read.ts:319)
- `F3` · major · open — `padi-tui create -- <command...>` reconstructs the command with `opts.argv.join(" ")` before sending it to the new shell. That loses argv boundaries and lets the shell re-split or reinterpret values containing spaces, quotes, `$`, `*`, `;`, newlines, etc. For example, a single prompt or file path argument containing spaces is no longer the same argument in the spawned terminal. (packages/padi-tui/src/create.ts:63)
- `F4` · major · open — When a byte-tap task ends, its `finally` only calls `tracker.forget(id)` and never removes the id from the `taps` map. If the attach stream ends or errors while the terminal remains active, later `reconcile()` calls see `taps.has(id)` and refuse to reopen the tap. Activity reporting for that terminal is then permanently dead for the lifetime of that activity subscription. (packages/padi/src/liveActivity.ts:176)
- `F5` · major · open — `settledSnapshot` converts `mirrorRemoteSurface(...).done` settling or rejecting into a successful status result. Since normal `status` completion is driven by `stop()`/timeout abort, an un-aborted mirror settle means the link closed, an upstream stream failed, or setup rejected. Today that can print a partial table or `no terminals.` with no diagnostic, because no `log` hook is passed and the rejection path is also mapped to `settle`. (packages/padi-tui/src/read.ts:146)
- `F6` · minor · open — The dirty-pulse reconcile loop has a bare `catch {}` that assumes every error is abort. Non-abort failures from the publisher iterator or from `reconcile()` itself would silently kill future tap reconciliation while the activity stream keeps running with stale state. (packages/padi/src/liveActivity.ts:210)
- `F7` · minor · open — The terminal-workspace README still documents the deleted `terminalWorkspaceSurface`, `./surface`, `./serveFsGit`, and `./socket` entry points, and still says pulam/kolu-server serve that surface. This branch removes those exports and moves the fs/git pulse serving into padi, so the README is now misleading for anyone using the package boundary. (packages/terminal-workspace/README.md:54)


**claude** — Agreed with all seven findings; fixed each in the working tree (three with new regression tests) and left the static-check gate green.

- `F1` **fixed** — Real crash-window data-loss. `conf` records a migration as done only after the handler returns, but persists each `delete` synchronously, so a crash after the backup + a partial strip reruns `stripLegacyStateKeys_1_31_0` on next boot — where the now session-less live file still trips `hasLegacy` and the old code re-`copyFileSync`'d it over the full first backup, losing `session` from the very file padi imports. Made the backup WRITE-ONCE: `packages/server/src/state.ts:130` now guards the copy with `existsSync(bakPath)` (the first copy, taken before any delete, is always the complete pre-strip file; every later copy could only be lossier). Added a regression test in `packages/server/src/stateMigration.test.ts` that runs the strip, drops `session` from the live file, re-seeds the residual legacy keys (the crash residue), reruns, and asserts the backup still holds `session` verbatim.

- `F2` **fixed** — Real hang. If the target exits in the gap between the caller resolving its id (`snapshotTerminals`) and `awaitAgentState`'s mirror subscription, it is absent from the mirror's FIRST `keys` snapshot; `mirrorCollection` only fires `onRemove` for keys it itself opened, so no removal ever fires and a no-timeout `wait` hangs — contradicting the documented "terminal exits first → exit 3". Fixed by REUSING the mirror's own departed-while-away mechanism: threaded an optional `initialKeys` through `watchTerminals` (`packages/padi-tui/src/read.ts:184`) and had `awaitAgentState` seed `() => [opts.id]` (`read.ts:333`); the mirror then fires `onRemove(opts.id)` on the first snapshot when the id is absent → `gone`. `watch` passes no seed, so its behavior is unchanged. Regression tests in `packages/padi-tui/src/read.test.ts`: the gone-race (no timeout — would hang without the fix) plus a present-terminal `met` control proving the seed is inert on the normal path.

- `F3` **fixed** — Real argv-boundary loss. `opts.argv.join(" ")` hands the shell a line it re-splits, so a single token carrying spaces/quotes/`$`/`*`/`;` (e.g. a `claude "review this PR"` prompt argument) shatters into several. Switched to `shellJoin` — the repo's existing POSIX-quote source of truth in `@kolu/shell-quote`, the same leaf `kaval-tui`/`anyagent` already use (`packages/padi-tui/src/create.ts:63`) — and added the workspace dep. `ran` is now the quoted, copy-pasteable line. Regression tests in `packages/padi-tui/src/create.test.ts` round-trip argv through `shellSplit(shellJoin(...))` for spaces, metacharacters, an empty arg, and the no-argv case.

- `F4` **fixed** — Real asymmetry. `openTap`'s `finally` forgot the tracker but never removed the `taps` map entry, so an attach stream that ends while the terminal stays active (a transient kaval drop) left `taps.has(id)` true and `reconcile()` refused to reopen — activity dead for that terminal for the subscription's life. Restored the symmetry `closeTap` already has: `packages/padi/src/liveActivity.ts:181` now does `if (taps.get(id) === tapAbort) taps.delete(id)` in `finally` before forgetting. The identity guard is load-bearing — it leaves a fresh controller a concurrent `closeTap`/`reconcile` may have installed untouched.

- `F5` **fixed** — Real silent-collapse, and squarely the `caught-error-must-not-collapse-to-empty` rule. `settledSnapshot` mapped BOTH mirror settle and reject to `settle` and returned the (partial/empty) accumulator with no diagnostic, so a mid-read link drop printed a blank/partial `status` table at exit 0 — while the sibling `cmdWatch` already fails loud on an un-aborted settle. Now `packages/padi-tui/src/read.ts:110` latches `linkFailed` at the instant `.done` fires iff WE didn't stop (so a trailing grace timer can't retroactively mask it), passes a `log` sink to capture the upstream diagnostic, and throws after the await — surfacing as exit 1 through `cmdStatus`'s dispose + `main`'s catch. The intentional stops (sensors resolved · grace · hard cap · empty fleet) are unaffected. Regression tests in `read.test.ts`: a dropped-link read rejects `/link closed/`, and a clean grace-driven settle still returns the snapshot.

- `F6` **fixed** — Agreed the bare `catch {}` on the dirty-pulse reconcile loop assumed every error was abort, so a publisher fault or a throw out of `reconcile()` would silently freeze tap reconciliation while the stream kept publishing stale frames. `packages/padi/src/liveActivity.ts:210` now swallows only the expected teardown (`sig.aborted`) and `log.error`s any non-abort failure, so it can't vanish in the detached task.

- `F7` **fixed** — Agreed the README was stale after the burial. Rewrote `packages/terminal-workspace/README.md`: dropped the deleted `./surface` / `./serveFsGit` / `./socket` entry-point rows (added the surviving `./agentProjection`), replaced the "hosted by pulam / one surface two homes" architecture prose with padi owning the composed record + the fs/git serving (`fsGitDeps.ts` streams, `servePadi.ts` procedures), and fixed the stale `pulam … folds with foldSnapshot` sentence (pulam is buried; `foldSnapshot` survives only as `fold`'s internal snapshot-only half).

Verification: `just check` (typecheck across all packages + `biome lint --error-on-warnings`) is green; `padi-tui` (24), `padi` (188), and the `stateMigration` suite (3) all pass.

commit: `c5a63484ebd6d1f840ba5a732292761e128621d6`


### Round 2

**codex** — approved: `true`

All seven prior findings are resolved in the current tree. I verified the follow-up commit against the exact failure modes from the previous review: the migration backup is write-once, wait seeds the target id through the mirror reconciliation path, create uses `shellJoin`, activity taps clear their map entries and log non-abort reconcile failures, status now fails on un-aborted mirror closure, and the terminal-workspace README no longer documents the buried surface/export paths. I did not run the test suite in this read-only review pass; I inspected the code and the added regression tests.

Findings:
- `F1` · blocking · resolved — Resolved. The strip migration now guards the backup write with `!existsSync(bakPath)`, preserving the first full pre-strip copy across reruns after partial deletion. The added state migration regression covers the crash-window overwrite case. (packages/server/src/state.ts:140)
- `F2` · major · resolved — Resolved. `awaitAgentState` now passes `initialKeys: () => [opts.id]` into `watchTerminals`, so the mirror's first-snapshot carry-over reconciliation emits `onRemove` when the watched terminal disappeared before subscription. The new read test pins the no-timeout gone race. (packages/padi-tui/src/read.ts:395)
- `F3` · major · resolved — Resolved. `runCreate` now uses `shellJoin(opts.argv)` and padi-tui declares the `@kolu/shell-quote` dependency. The new tests round-trip spaces, metacharacters, and empty args through `shellSplit`. (packages/padi-tui/src/create.ts:66)
- `F4` · major · resolved — Resolved. The tap task now removes its own controller from `taps` in `finally` with an identity guard before forgetting tracker state, so a later reconcile can reopen a tap after an attach stream ends. (packages/padi/src/liveActivity.ts:183)
- `F5` · major · resolved — Resolved. `settledSnapshot` now latches un-aborted mirror completion as `linkFailed`, captures upstream diagnostics through `log`, and throws instead of returning a partial accumulator. The added tests cover dropped-link failure and clean grace-driven success. (packages/padi-tui/src/read.ts:186)
- `F6` · minor · resolved — Resolved. The dirty-pulse reconcile loop now catches explicitly, swallows only teardown via `sig.aborted`, and logs non-abort failures at error level. (packages/padi/src/liveActivity.ts:217)
- `F7` · minor · resolved — Resolved. The README now describes padi as the owning serving surface, documents the surviving exports, and removes the stale `terminalWorkspaceSurface`, `serveFsGit`, `serveTerminalWorkspace`, pulam, and socket entry-point claims. (packages/terminal-workspace/README.md:47)

_codex on the rebuttal:_ CLAUDE agreed with all seven findings, so there were no disputes to adjudicate. I verified each fix in the current tree and accept them as resolving F1-F7.

srid:

**Clean** — 0 findings on the W2.3 padi diff.

Ran as a backgrounded Workflow (4 rule dimensions + a fact-check pass in parallel, each finding adversarially refuted before surviving):

- **errors & subscriptions** — `caught-error-must-not-collapse-to-empty`, `errors-must-log-at-error`, `subscription-must-surface-errors`, `silent-handler-required-on-void-subscriptions`, `callback-fanout-guarded-at-funnel`, `watcher-lifecycle-logs` → clean
- **structure & types** — `no-re-export-bridge-modules`, `no-thin-wrapper-functions`, `no-untyped-escape-hatches`, `prefer-ts-pattern`, `new-package-has-readme` → clean
- **migration · persistence · perf · design-philosophy** — `migration-shape-guard`, `persisted-schema-stays-tolerant`, `integration-perf-hygiene`, `no-sync-blocking-on-the-serving-loop`, + fail-fast/no-fallbacks · electricity boundaries · reuse-the-source-of-truth → clean
- **fact-check** — every claim the diff adds/edits (comments, docstrings, README/changelog/website prose, the "died at W1.R7 / #1652" provenance) verified against the code → clean

Nothing survived after lens-debate (3 fixes), codex-debate (7 correctness fixes), and the simplify pass (4).

srid:

The **x86_64-linux** lane ran on **`kolu-ci-2`** (`dev-x86-64-linux-04`) — commit `1456eeda0`, **passed** ✓

- **Lane wall** (pipeline): **6m38s**

| recipe | duration | |
|---|--:|:--|
| `ci::nix` | 5m54s | ✓ |
| `ci::e2e` | 3m49s | ✓ |
| `ci::home-manager` | 2m48s | ✓ |
| `ci::smoke` | 1m29s | ✓ |
| `ci::flake-check` | 1m27s | ✓ |
| `ci::atlas-sync` | 1m14s | ✓ |
| `ci::unit` | 1m4s | ✓ |
| `ci::install` | 1m2s | ✓ |
| `ci::pnpm-hash-fresh` | 1m0s | ✓ |
| `_ci-setup` | 44s | ✓ |
| `ci::surface-example-build` | 26s | ✓ |
| `ci::surface-app-example-build` | 26s | ✓ |
| `ci::biome` | 24s | ✓ |
| `ci::fmt` | 23s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/1456eed/timings.jsonl`; pool state is a live `flock` probe.</sub>

PR #1673, opened 2026-07-04
feat(kaval-tui): `send --submit` — one-command prompt delivery for agent-drives-agent

srid:
## What & why

Handing a prompt to a TUI agent took **two** commands — `kaval-tui send <id> "text"`, then a *separate* `send <id> --key Enter` — because an Enter written in the same breath as a bracketed paste races the agent's paste-processing debounce and is **silently dropped**, leaving the prompt staged on the `❯` line. The `/kolu` skill documents this as its **#1 failure mode**: the second command gets orphaned when the first is backgrounded, and every agent-driving loop has to remember the sleep between them.

`kaval-tui send <id> "text" --submit` collapses it to one command: write the text (bracketed-paste rules unchanged), wait a fixed grace, **then** write Enter — landing it *past* the debounce.

This is **scheduling, not verification**: a blind delay, no screen read, no idle-detection. The command returns in well under a second regardless of what the agent is doing (idle, streaming, mid-spinner — irrelevant by design).

## Surface

- **`--submit`** (bare → default **250ms** grace) or **`--submit=<ms>`** (tunable). Composes with piped-stdin and multi-line pastes.
- **`--submit` + `--key` is an error** — the keys channel is for control sequences; `--submit` owns the Enter.
- Plain `send` (no `--submit`) is **byte-for-byte unchanged**: raw, no implicit Enter — the load-bearing manual-control channel.
- `--json` gains `submitted: true` + `graceMs` under `--submit`.

```
$ echo 'Reply with exactly: PONG' | kaval-tui send <id> --submit --json
{ "id": "…", "bytes": 85, "paste": true, "keys": [], "submitted": true, "graceMs": 250 }
```

## Key decisions (all mine to make — this was a fully-autonomous task)

- **Flag type is `String`, not `Boolean`.** A Boolean silently *swallows* `--submit=250` (dropping the tuned grace) — a fail-fast violation. `String` captures `""` for a bare `--submit` and the digits for `--submit=<ms>`; `parseSubmitGrace` turns that into the grace, failing **loud** on junk (`--submit=abc`, `--submit=-5`) rather than NaN-degrading to the default.
- **Validated pre-dial.** `send`'s flag checks (paste conflict, `--submit`+`--key`, grace parse) are pure, so they run *before* connecting — mirroring the existing `wait` path — so a bad flag fails fast with no daemon to tear down. Carried through a `sendCall` bundle, exactly like `waitCall`.
- **`send.ts` stays pure.** The grace `delay` + Enter write live in `main.ts`'s `cmdSend` (the I/O glue); `send.ts` only *plans* (`planSend` returns `submit: { graceMs } | null`, and counts the deferred Enter in `bytes`). The submit byte reuses `NAMED_KEY_BYTES.enter` — the same CR `--key Enter` sends — not a hand-rolled `"\r"`.
- **No new fallback/knob/degradation path** — honors the repo's fail-fast philosophy. `--submit` is a leaf on the CLI, not a new volatility receptacle.

## Acceptance test — real, not mocked

`packages/kaval-tui/src/attach.test.ts` spawns a **real PTY** running a scripted paste-debounce TUI (`paste-debounce-tui.fixture.mjs`) that reproduces Claude Code's exact drop: an Enter within a debounce window of a bracketed paste is silently discarded; a later one submits. Driving real claude *in CI* is infeasible (no binary/auth), so this fixture is the deterministic reproduction — and I also verified once against a real Claude (below).

```
✓ CONTROL: a same-breath Enter (paste + --key Enter) is DROPPED — proves the fixture reproduces the bug  211ms
✓ (a) idle agent + --submit → the turn actually STARTS (prompt not left staged)                          461ms
✓ (b) BUSY agent (mid-turn, streaming) + --submit → the message is accepted, never lost                  815ms
✓ (c) a multi-KB piped paste + --submit submits INTACT                                                    460ms
```

The **CONTROL** is the red/green: same prompt, same fixture — a same-breath Enter is dropped, `--submit` submits. That's the proof the fixture reproduces the bug *and* that `--submit` fixes it. Plus unit tests for `parseSubmitGrace`, `planSend`'s submit planning, `formatSend`, and the mutual-exclusion errors. **115 tests pass.**

## Manual verification against a real Claude Code

Driven over an *isolated* standalone kaval (own socket, no production kolu touched), using this branch's `send --submit`:

```
$ printf 'Reply with exactly the single word: PONG ...' | kaval-tui send <id> --submit --json
{ "id": "c6ba027f-…", "bytes": 85, "paste": true, "keys": [], "submitted": true, "graceMs": 250 }

# snapshot ~1.2s later — the turn STARTED from ONE command and Claude replied;
# the ❯ input box is empty (submitted, not staged):
  ❯ Reply with exactly the single word: PONG (nothing else, no punctuation).
  ● PONG
  ✻ Baked for 1s
  ────────────────────────────────
  ❯
```

(The same-breath control against real claude is a *race* — this run it happened to win, which is exactly why the ritual is unreliable and `--submit` exists. The deterministic drop lives in the fixture test above.)

## Docs collapsed to one command

The `/kolu` skill (apm source `agents/.apm/skills/kolu/SKILL.md` + its generated `.claude`/`.agents` copies), the `kaval-tui` and `padi-tui` READMEs, the kolu.dev `kaval` command reference, and the `pty-daemon-tui` Atlas note (+ rebuilt `dist/`) now lead with `send --submit`, keeping raw `send` + `--key Enter` as the one-line manual-control note. Changelog entry added.

> **apm note:** `apm install` re-vendors the kolu skill from the pinned remote `juspay/kolu`, so it can't pick up a local uncommitted source edit — the generated `.claude`/`.agents` copies are hand-synced from the edited source here (byte-identical to it), and a post-merge lock bump reproduces them.

## Scope

Client-side only: no kaval daemon change, no wire/contract change — the grace is a client sleep between the two writes `send` already knew how to do.


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `30ea598fac19`

### Applied (4)
- `lowy-1` SendPlan interpreter duplicated: acceptance test drives a replica, not the shipped glue — commit `40d4c2c12`
- `lowy-2` formatSend re-encodes SendPlan.submit as a flattened optional pair, admitting an illegal state — commit `513d57d90`
- `hickey-1` formatSend re-fragments plan.submit into parallel optional fields — (uncommitted)
- `hickey-2` Plan-execution algorithm duplicated across cmdSend and the test's execPlan — (uncommitted)

### Agreed — no change (3)
- `lowy-3` Local delay in main.ts duplicates attach.ts's private delay — checked, and it is a leaf, not a receptacle (packages/kaval-tui/src/main.ts:477-478)
- `hickey-3` 'wait ms' timer leaf multiplied across main.ts, attach.ts, and inline tests (packages/kaval-tui/src/main.ts:477)
- `hickey-4` SendPlan.bytes now folds the deferred Enter into the write total (packages/kaval-tui/src/send.ts:151-153)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `30ea598fac19`
### Round 1

**codex** — approved: `false`

Read-only review of the branch diff against 30ea598fac1958a86cf70bb7c11345f0100a9b22 plus status. The main implementation shape is solid: validation is pre-dial, send planning stays pure, the delayed Enter lives in the executor, and the public JSON shape is preserved. I found one real correctness bug around timer overflow, one test-proof gap, and a few user-facing documentation consistency issues.

Findings:
- `F1` · major · open — `parseSubmitGrace` accepts any digit string and returns `Number(raw)`, but `executeSendPlan` later feeds that value to `setTimeout`. Values above Node's signed 32-bit timer ceiling do not wait longer; Node coerces them to 1ms with a warning, so `--submit=2147483648` sends Enter almost immediately instead of failing loud or honoring the requested grace. This is the exact timer-overflow class already guarded in `wait.ts`. (packages/kaval-tui/src/send.ts:57)
- `F2` · minor · open — The "BUSY agent" acceptance test waits until `SUBMITTED#1` appears, then assumes the fixture is still mid-stream. The fixture's busy burst lasts only about 500ms, so on a slow or stalled CI loop the first turn can already have printed `DONE T1` before the second prompt is sent. The test would still pass, but it would no longer prove the busy/mid-turn case it claims to cover. (packages/kaval-tui/src/attach.test.ts:535)
- `F3` · minor · open — The public kaval command reference's flags column for `kaval-tui send` omits the new `--submit` flag. The prose mentions it, but this page presents flags as a quick-reference field, so the new headline flag is less discoverable and the reference is stale. (website/src/pages/kaval.astro:40)
- `F4` · nit · open — The Atlas note was meaningfully edited to document `send --submit`, but its `updated` frontmatter still says `2026-06-19`. The Atlas convention is that `updated` tracks the last meaningful edit. (docs/atlas/src/content/atlas/pty-daemon-tui.mdx:7)
- `F5` · nit · open — The docs say `--submit=<ms>` tunes the grace and then promise the command "returns in well under a second." That is only true for the default 250ms grace; a tuned grace of 1000ms or higher makes the command wait at least that long. The same wording appears in the generated Kolu skill docs and the changelog. (packages/kaval-tui/README.md:64)


**claude** — Fixed the real timer-overflow bug (F1), made the busy acceptance test deterministic (F2), and applied all three doc-consistency fixes (F3/F4/F5); no disputes this round.

- `F1` **fixed** — Real bug, and it's the exact class `wait.ts` already guards. `parseSubmitGrace` (`send.ts:57`) accepted any digit string and returned `Number(raw)`, which `executeSendPlan` (`sendExec.ts:31`) feeds straight to `setTimeout`; a value above the signed-32-bit ceiling is silently clamped to 1ms and fires near-instantly, dropping the tuned grace and racing the paste debounce — a fail-fast violation. Fixed by reusing the existing source of truth: imported `MAX_TIMER_MS` from `./wait.ts` (no cycle — wait.ts's only import of connect.ts is type-only/erased) and added a loud throw for `ms > MAX_TIMER_MS` in `send.ts` `parseSubmitGrace`. 0 stays allowed (deliberate). Added boundary tests in `send.test.ts`: accepts exactly `MAX_TIMER_MS`, rejects `MAX_TIMER_MS + 1` and a huge digit string `"99999999999999999999"`. 26 send tests pass.
- `F2` **fixed** — Agreed the busy test could stop proving its mid-turn claim on a slow loop: the fixture's busy burst was a fixed ~500ms window and test (b) only asserted `SUBMITTED#2`, which passes even if `DONE T1` already printed. Made it deterministic instead of leaving it racy: parameterized the burst length via `FIXTURE_BUSY_TICKS` in `paste-debounce-tui.fixture.mjs`, `spawnFixture` now takes an `extraEnv` bag, and test (b) (`attach.test.ts`) holds the burst open (200×25ms ≈ 5s) and asserts the mid-turn screen `.not.toContain("DONE T1")` before firing the second submit — so the second prompt provably lands at a still-busy agent. All 4 acceptance tests pass.
- `F3` **fixed** — Added `--submit[=<ms>]` to the `flags` quick-reference column for `kaval-tui send` in `website/src/pages/kaval.astro` (it was only in the prose).
- `F4` **fixed** — Bumped `updated` from `2026-06-19` to `2026-07-03` in `docs/atlas/src/content/atlas/pty-daemon-tui.mdx`, then ran `just atlas::build`; only `docs/atlas/dist/pty-daemon-tui.html` regenerated (build stayed idempotent — no other page moved) and it now carries the new date, keeping the `atlas::check-sync` gate green.
- `F5` **fixed** — Scoped the sub-second claim to the default grace everywhere it appeared: `packages/kaval-tui/README.md`, all three tracked kolu SKILL.md copies (`.claude/`, `.agents/`, `agents/.apm/`) plus the untracked `apm_modules/_local` source so a regen stays consistent, and the changelog entry (`website/src/content/changelog/unreleased.mdx`). New wording: "returns after the grace regardless of what the agent is doing — bare `--submit` in well under a second, a tuned `--submit=<ms>` after that many ms."

Gate: `just check` (typecheck + biome) green; 26 send unit tests + 4 real-PTY acceptance tests pass.

commit: `f91e328279ebcb52f54b70d7d94c431eff4ea6fb`


### Round 2

**codex** — approved: `true`

All five findings from the prior round are resolved in the current tree. I verified the fixes by reading the current diff and targeted files; the worktree is clean, and I did not find any regression introduced by this response round.

Findings:
- `F1` · major · resolved — `parseSubmitGrace` previously accepted timer values above Node's signed 32-bit `setTimeout` ceiling, which would clamp to 1ms and race the paste debounce. (packages/kaval-tui/src/send.ts:64)
- `F2` · minor · resolved — The busy-agent acceptance test previously could pass after the first turn had already completed, so it did not necessarily prove a mid-turn submit. (packages/kaval-tui/src/attach.test.ts:531)
- `F3` · minor · resolved — The kaval quick-reference flags list omitted the new `--submit` flag. (website/src/pages/kaval.astro:40)
- `F4` · nit · resolved — The Atlas note had a meaningful edit but stale `updated` frontmatter. (docs/atlas/src/content/atlas/pty-daemon-tui.mdx:7)
- `F5` · nit · resolved — The docs previously promised `--submit=<ms>` would return in under a second even though tuned grace values can exceed that. (packages/kaval-tui/README.md:64)

_codex on the rebuttal:_ No disputes were raised. I verified each claimed fix in the current tree: F1 now rejects `MAX_TIMER_MS + 1` and huge digit strings; F2 now fails if the supposedly busy turn has already printed `DONE T1`; F3, F4, and F5 are reflected in the relevant docs and generated Atlas output. No new regression from this response round.

srid:

Ran the rule-checklist and fact-check passes over the diff vs the merge-base (elegance skipped — `/simplify` already ran this gauntlet). **3 findings, all actioned** in `10e398e3c`.

| # | Pass | Severity | Finding | Fix |
|---|------|----------|---------|-----|
| 1 | fact-check | **major** | A bare `--submit <word>` (space, no `=`) makes type-flag swallow the next positional as the grace — silently dropping the prompt text. An all-digit prompt (`--submit 12345`) even parses as a valid grace, so *nothing errored*: it submitted an empty line after a 12s wait. | Added `submitSwallowedPositional` — a bare `--submit` token in argv that captured a non-empty value is rejected **loud, pre-dial** ("pass the grace with `--submit=<ms>`, or put a bare `--submit` after the text"). Unit-tested; verified the guard fires while `--submit=250` passes through. |
| 2 | rules | dry-rule-of-three | The paste-debounce fixture hardcoded the bracketed-paste markers (`\x1b[200~` / `\x1b[201~`) instead of the `@kolu/terminal-protocol` constants `planSend` actually wraps text with — so the fixture could silently drift from the shipped behavior it tests. | Thread the real `BRACKETED_PASTE_START/END` from the test (which imports the protocol and drives `planSend`) into the fixture via env, so it recognizes exactly what `send` emits. Fixture stays plain-`node` (no TS import); fails loud if the markers aren't passed. |
| 3 | fact-check | nit | `DEFAULT_SUBMIT_GRACE_MS`'s doc comment asserted "250ms clears Claude Code's debounce" as measured fact, though nothing measured a real agent's debounce. | Reworded to a **grounded default** — 250ms cleared real Claude Code once end-to-end (the PONG check) and is tunable via `--submit=<ms>`, explicitly *not* a measured floor for every agent. |

**Rejected on verification:** the three byte-identical `kolu` `SKILL.md` copies edited in lockstep — traced to the apm source (`agents/.apm/skills/kolu`) + its regenerated `.claude`/`.agents` outputs (generated-and-committed, like `docs/atlas/dist/*.html`), not hand-duplicated source.

**Cleared:** `SendPlan.submit: {graceMs}|null` is a proper discriminated shape (no illegal states), byte accounting doesn't double-count the deferred Enter, `parseSubmitGrace` throws propagate to a `kaval-tui:` error (no silent swallowing), the fixture's `setInterval` self-clears, and no SolidJS / persisted-schema / watcher / icon rule applies.

srid:
## Evidence

This is a terminal/behavioral change (a CLI flag for agent-drives-agent), so the proof is a real agent submitting from one command, plus the acceptance suite — not a browser screenshot.

### 1. Real Claude Code — one command starts the turn

Driven over an **isolated standalone kaval** (own socket, production kolu untouched), using this branch's `kaval-tui send --submit`:

```
$ printf 'Reply with exactly the single word: PONG (nothing else, no punctuation).' \
    | kaval-tui send <id> --submit --json
{
  "id": "c6ba027f-…",
  "bytes": 85,
  "paste": true,
  "keys": [],
  "submitted": true,
  "graceMs": 250
}
```

Snapshot ~1.2s later — the turn **started from one command** and Claude replied; the `❯` input box is empty (the prompt submitted, it did not sit staged):

```
 ❯ Reply with exactly the single word: PONG (nothing else, no punctuation).
 ● PONG
 ✻ Baked for 1s
 ────────────────────────────────────────────────
 ❯
```

### 2. Acceptance suite — a real PTY paste-debounce TUI

The suite spawns a real PTY running a scripted TUI that reproduces the exact drop (`paste-debounce-tui.fixture.mjs`). The **CONTROL** is the red/green: same fixture, a same-breath Enter is dropped; `--submit` submits.

```
✓ CONTROL: a same-breath Enter (paste + --key Enter) is DROPPED — proves the fixture reproduces the bug   212ms
✓ (a) idle agent + --submit → the turn actually STARTS (prompt not left staged)                           461ms
✓ (b) BUSY agent (mid-turn, streaming) + --submit → the message is accepted, never lost                   815ms
✓ (c) a multi-KB piped paste + --submit submits INTACT                                                    462ms

Test Files  1 passed (1)
     Tests  14 passed (14)
```

The full package suite (unit + acceptance) is **120 passing**.

### 3. Docs collapse — two commands → one

The `/kolu` skill's loop, before and after:

**Before (master):**
```sh
kaval-tui send  "$id" "refactor the parser to use a lexer"     # 1. TYPE the prompt
kaval-tui send  "$id" --key Enter                              # 2. SUBMIT it (its own step)
…
kaval-tui send  "$id" "now add tests for it"; kaval-tui send "$id" --key Enter   # loop
```

**After (this PR):**
```sh
kaval-tui send  "$id" "refactor the parser to use a lexer" --submit   # 1. type AND submit
…
kaval-tui send  "$id" "now add tests for it" --submit          # loop
```

The `## kaval-tui send — type, then submit (two steps)` section and the #1-failure-mode warning box are rewritten around `send --submit`, keeping raw `send` + `--key Enter` as the documented manual-control channel. The kaval-tui + padi-tui READMEs, the kolu.dev `kaval` command reference, and the `pty-daemon-tui` Atlas note carry the same collapse.

srid:

The **x86_64-linux** lane ran on **`kolu-ci-2`** (`dev-x86-64-linux-04`) — commit `10e398e3c`, **passed** ✓

- **Lane wall** (pipeline): **3m26s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 3m3s | ✓ |
| `ci::home-manager` | 1m57s | ✓ |
| `ci::unit` | 1m12s | ✓ |
| `ci::nix` | 51s | ✓ |
| `ci::smoke` | 50s | ✓ |
| `ci::fmt` | 23s | ✓ |
| `ci::biome` | 23s | ✓ |
| `ci::install` | 12s | ✓ |
| `_ci-setup` | 10s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/10e398e/timings.jsonl`; pool state is a live `flock` probe.</sub>

PR #1671, opened 2026-07-04
fix(server): drain-on-build-mismatch — a padi-build change restarts padi (#1670)

srid:
Fixes #1670.

A deploy that changed **padi**'s build but not its `padiSurface` contract silently kept the *old* padi running. W2.2's binder convergence keyed on the **contract version only**, so a same-contract build change had no convergence path: the freshly-booted kolu-server **adopted** the running old-build padi and the new code never took effect — until a crash, a host reboot, or a manual `kill` (there is no in-app action that restarts padi; the "Restart" button recycles *kaval*). Found by srid dogfooding on zest: a running padi at `k5fbfp3…kolu-1.1.0` (9h uptime) sat under a freshly-deployed server that baked a different padi closure, and terminals spawned post-deploy lacked the new spawn behavior.

## Reproduce first, then fix

**The repro (commit 1, RED before the fix).** A new `adoption-padi-upgrade` VM arm in the linux adoption suite (`nix/home/example/adoption/padi-upgrade.nix`) boots a default-built kolu with a live marker terminal, then stops it and starts a `kolu-new` whose padi is a **distinct build** — via the reserved `padiBuildIdOverride` seam (the cheap skew: only `PADI_BUILD_ID` differs, so `kolu-new` shares the closure and exercises the *real* binder-boot path, not a unit shim). It performs **no manual drain**: `kolu-new`'s boot must converge the daemon on its own. It then asserts the padi gate pid **changed** (padi restarted), the kaval gate pid **unchanged** (kaval + PTYs kept), the same PTY + scrollback nonce survived, the session is warm (no restore card, no parked records), and the binder's build-change breadcrumb shows the skew.

Verified **red on a leased KVM box** on current master — for exactly the right reason:

```
kolu[1088]: INFO (833): adopted a surviving daemon (its PTYs are preserved) … hostId:"padi-local"
FAIL(padi-upgrade-verify): the padi-build change did NOT restart padi (kaval-preserving) within 90s.
  padi gate pid:  833 -> 833 (must CHANGE — a build change drains + respawns padi; UNCHANGED = the #1670 bug: old build adopted)
  kaval gate pid: 916 -> 916 (must be UNCHANGED — kaval + PTYs survive the padi replacement)
  pty …-…-005d64 pid: 968 -> [968] (must still be listed — same PTY)
  session … in …/config.json: yes (must be yes — warm)
  park logs (session-trace park: seeded): 0 (must be 0 — a park is the restore card)
  build-change line: []
```

[...]

Issue #1670, opened 2026-07-04
A deploy that changes padi's build never restarts the running padi (convergence keys on contract version only)


srid:
Found by srid deploying a padi-code-changing build to zest: the running padi (9h uptime, old build) was adopted by the new kolu-server and kept running — new padi code silently never takes effect until a contract bump, a crash, or a manual Restart.

**Evidence (zest):** padi running `k5fbfp3…kolu-1.1.0` (started 11:18) while the freshly-deployed kolu-server (20:22) bakes `KOLU_PADI_BIN=/nix/store/rhn5km05…-padi`. Downstream symptom: terminals spawned post-deploy lacked new spawn behavior, because the spawner is the old padi.

**Root cause:** W2.2's newest-wins convergence compares `PADI_SURFACE_VERSION` (the contract) only. A same-contract build change has no convergence path — but the design's stated trade is *padi restarts at deploy cadence* (warm across web-shell churn, re-derived across padi churn). The mechanism half-exists: the control-core `hello` already carries the running padi's build id; only the policy is missing.

**Fix shape:** drain-on-build-mismatch, **once, at binder boot** — if `hello`'s build id ≠ the binder's baked `PADI_BUILD_ID`, control-core drain → spawn own closure. Do **not** order builds (store hashes don't order; continuous mismatch-drain livelocks two persistent binders) — once-per-binder-boot converges sequential deploys to last-deployed and cannot flap.

**Repro-first:** a `padi-upgrade` VM arm in the adoption suite — deploy a changed-padi-build kolu over a running padi → assert padi gate pid changes to the new build, kaval gate pid unchanged, session warm — red on master today. Unit arms: mismatch-at-boot drains+respawns; match adopts; a reconnect (same binder process) never re-drains; two-binder no-flap.

Pre-existing on master (W2.2, #1664); surfaced by the first padi-changing deploy. Standalone PR — not part of #1665.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

srid:
Correction to the issue text above: it lists 'a manual Restart' as an existing convergence trigger — that is WRONG. Post-W2.2 the in-app Restart button recycles **kaval**, not padi; there is **no shipped user action that restarts padi**. So the real severity is higher: a padi-build change has NO trigger short of a crash, a host reboot, or a manual `kill` of the padi process over ssh. This is precisely why the fix must be automatic (drain-on-build-mismatch at the next server boot) — there is no manual path to fall back on.

PR #1674, opened 2026-07-04
L3: the daemon-lifecycle convergence kit (policy as the parameter) + kaval currency slice

srid:
Implements ledger **L3** — the daemon-lifecycle **convergence kit** — and stacks on the #1670 fix (which merged as #1671), refactoring that fix into the kit. Stacked off master (contains #1670 + W2.3 #1665).

## The problem

kaval and padi each hand-rolled the same lifecycle question — *"the running daemon is not the one I shipped: detect it, converge it"* — and **#1670 is what the duplication cost**: kaval's currency lesson existed, nothing shared it, and padi's re-derivation shipped with a hole. The real remaining duplication (narrower than the ledger's original "six-fold" — the generic endpoint was already shared) was: the two `buildId.ts` twins, the identity/handshake shapes, and — the crux — padi's ordering/build-fence **policy stranded in `packages/server`, outside the soul-agnostic endpoint**.

## The kit — policy as the parameter

One shared mechanism in `@kolu/surface-daemon-supervisor`, with **each daemon declaring its policy per trigger** and the machinery shared:

- **`decide(baked, running|null, policy, fenceSpent) → Decision`** — a **pure** fold, zero I/O. The whole policy TABLE is unit-tested against it directly (no daemons); the absent / off-nix / fence cases are table *rows*, not special branches.
- **`converge({endpoint, baked, probe, policy, buildFence, log}) → ConvergenceOutcome`** — probes identity over a version-agnostic channel, calls `decide`, enacts via the endpoint's **existing** boot methods, and **returns** a typed outcome. It performs **no** nudge/surface side effect: a `nudge-human` mismatch comes back as `mismatch-reported` and the *caller* surfaces it (the kit detects + decides; the caller enacts what it owns).

Declarations:

| daemon | contract skew | build mismatch |
| --- | --- | --- |
| **kaval** (supervised by padi) | `recycle` | `nudge-human` (currency, client-side compare — unchanged) |
| **padi** (supervised by kolu-server) | `drain-newer-else-refuse` | `drain-and-replace` (#1670, fenced) |

**The bind method is policy-driven, not decision-driven** — a recycle-on-skew daemon (kaval) binds through `adoptOrEnsure` on *every* path (so a skew recycles wherever the endpoint finds it, **including the W2.2 adopt-hint the single probe never saw** — the legacy-port migration recycle); a refuse/drain daemon (padi) binds through `adoptOrSpawnOrRefuse`. This keeps **both daemons byte-identical to their pre-kit boot method** — the kit lifts only the *decision*, never the endpoint mechanism.

## Three make-illegal-unrepresentable pins

1. **Drain capability is typed.** The drain policy arms exist only for a drainable handshake; kaval (no `drain` verb — recycling it kills PTYs) declaring `drain-and-replace` is a **compile error** (pinned by a `@ts-expect-error`).
2. **Ordering is per-field law.** Contract versions are **ordered** (`contractIsNewer`); build ids are **match-only** (`buildIdMatches`) with **no ordering exported** — pinned by an API-shape test so a consumer cannot spell a build-id comparison (store hashes don't order).
3. **Identity reachable under skew.** The probe reads identity *before* the versioned handshake — padi over the frozen control-core `hello`, kaval over `system.version` (pinned by a new kaval shape-pin test so a silent field rename fails loudly).

## The Nix half

`readBakedIdentity(prefix)` (TS) and **`mkDaemonIdentity`** (Nix, living *inside* the surface-daemon package — location is structure) are the two halves of the same daemon-identity electricity. Each daemon instantiates `mkDaemonIdentity` with its **own** `behavioralFileset` (the policy — what counts as its behavior); the `hashString`-over-fileset recipe + the `--set` bake are the shared electricity. The recipe-unify commit is **hash-neutral** (verified byte-identical: kaval `8379a79a…` new-recipe vs old-inline over the same source).

## The open question, settled — kaval currency = behavioral slice

The kit put both daemons' currency side by side and made a real bug visible (srid hit it on zest, 2026-07-03): kaval's staleKey drove the human "update available" nudge off its **whole** closure, including the shared `@kolu/surface-daemon` **spine** — so a spine-only, contract-*compatible* refactor fired a spurious nudge, and acting on it recycles kaval and **kills live PTYs**. **Fix:** key kaval's currency on its **behavioral slice** (kaval + terminal-protocol), spine excluded. Rationale: the spine's behavioral surface to a consumer *is* the wire contract (which lives in kaval and stays hashed) — a contract-compatible spine change is behaviorally interchangeable by the contract's definition, so keying currency on it double-counts what the contract already covers. A spine change that *does* matter bumps the contract → recycle-on-skew converges it (a separate signal). **padi is untouched** — its staleness response is a cheap auto-drain, so over-firing is harmless. The closure guard is restated to pin the new invariant, carrying the zest incident as its comment.

> **Heads-up (one-time, expected):** because the refactor legitimately moved code *within* kaval's hashed closure, the first deploy after this lands fires the kaval update-nudge **once** — the closure genuinely changed. It is **not** the zest false-positive returning (that was a spine-only change with no behavior delta; this is real motion). Thereafter, a spine-only change no longer nudges.

## Staging & gates

Commit-staged so tests stay green each step: (1) the `readBakedIdentity` recipe (closure-neutral) → (2) the kit + pins → the Nix `mkDaemonIdentity` unify (nix-eval byte-identity gate) → (3) kolu-server converges padi → (4) padi converges kaval → (5) the kaval currency slice (a single visible root-set diff, hash change **by design**, gated by the restated closure test + the VM currency arm). Through stage 4 the fileset *root-sets* are unchanged (pure content motion); the VM adoption suite is the byte-identical behavioral pin throughout.

`#1670`'s unit arms are **refactored into** the kit (its logic now lives in `decide.test.ts` / `converge.test.ts`); `padiConverge.test.ts` is removed.

## No drishti PR

The surface-daemon stack is **not** consumed by drishti (its overlay hydrates only `surface`/`surface-app`/`surface-nix-host`), and no kit change touches `packages/surface/` (the comparators reuse `@kolu/surface`'s version predicate unchanged, from `@kolu/surface-daemon`). So the drishti gate does not bite.


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `30ea598fac19`

### Applied (2)
- `lowy-2` ConvergeLogger duplicates the existing Logger receptacle in @kolu/surface-daemon — commit `e72ee2929`
- `hickey-2` Pure decide() embeds human-readable English reason strings — presentation braided into the zero-I/O fold — commit `e1b8a5027`

### Agreed — no change (9)
- `lowy-1` Boundary verdict: convergence kit is genuine electricity (receptacle = converge+policy) (packages/surface-daemon-supervisor/src/convergence/index.ts:26)
- `lowy-3` buildFence is a required inert parameter for non-drainable callers — the one un-gated illegal-state (packages/surface-daemon-supervisor/src/convergence/converge.ts:107)
- `lowy-4` Bind-strategy axis is derived from onContractSkew.kind rather than declared (packages/surface-daemon-supervisor/src/convergence/converge.ts:118)
- `lowy-5` Comparators re-exported through the supervisor barrel have no consumer (packages/surface-daemon-supervisor/src/convergence/index.ts:20)
- `hickey-1` converge infers the bind method from onContractSkew.kind — bind-mechanism choice complected with the skew policy (packages/surface-daemon-supervisor/src/convergence/converge.ts:118)
- `hickey-3` Decision and ConvergenceOutcome are parallel sum types with near-verbatim variant renaming (packages/surface-daemon-supervisor/src/convergence/converge.ts:80)
- `hickey-4` AnyConvergencePolicy is a second type for one 'policy' concept (packages/surface-daemon-supervisor/src/convergence/policy.ts:62)
- `hickey-5` capability is declared twice (policy field + probe field); the policy copy is never read at runtime (packages/surface-daemon-supervisor/src/convergence/policy.ts:52)
- `hickey-6` buildFence is a required converge() parameter that is inert for every nudge-human daemon (packages/surface-daemon-supervisor/src/convergence/converge.ts:107)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 round(s) · codex reviewed at `xhigh` reasoning effort · base `30ea598fac19`

### Round 1

**codex** — approved: `false`

The extraction is mostly coherent, but I found one real correctness problem in the new convergence API: it reports a specific outcome the endpoint boolean cannot actually prove. I also found documentation/logging drift around deliberately important invariants, including one README architecture statement that now contradicts the branch.

Findings:
- `F1` · major · open — `converge()` maps `bind()` returning `false` to `{ kind: "spawned" }`, but the endpoint methods only return whether a survivor was adopted. Per the endpoint contract, `false` can also mean recycled, refused, or left standing as degraded after a non-skew connect failure. This is reachable with the new probes: both can return `null` when an answering daemon fails the identity read, after which the endpoint may find the live holder, fail to adopt it, leave it standing/degraded, and return `false`; the caller still receives `spawned`. That makes `ConvergenceOutcome` an unreliable source of truth for caller-owned effects. (packages/surface-daemon-supervisor/src/convergence/converge.ts:117)
- `F2` · minor · open — The #1670 breadcrumb is not actually byte-identical to the pre-extraction log despite the branch rationale saying it is. The old log included the structured `running` contract version and the longer stable message suffix about store hashes / once-per-boot; the new caller-emitted log drops both, and it is emitted only after `converge()` completes the follow-on bind rather than at the build-drain decision point. (packages/server/src/padiBinding.ts:762)
- `F3` · minor · open — The architecture README still says `@kolu/surface-daemon` is "hashed whole into kaval's staleKey", but this branch deliberately removes the surface-daemon spine from kaval's currency slice. That leaves the repo's main architecture table contradicting `default.nix` and the updated closure test. (README.md:220)
- `F4` · nit · open — Several nearby comments still describe the old model: `default.nix` says kaval has "three roots" including `@kolu/surface-daemon`, `packages/kaval/src/buildId.ts:4` says the staleKey includes `surface-daemon`, and `packages/padi/src/dial.ts:16` still names removed symbols like `isBinderNewer`, `bindPadiOnce`, and `probePadiSkew`. These are now stale enough to mislead the next reader about the new policy split. (default.nix:146)



**claude** — Conceded F1 (relabeled the un-provable `spawned` outcome to `not-adopted`), fixed the breadcrumb drift and all three stale comments (F2 partial, F3/F4 fixed), disputed only F2's timing-restructure.

- `F1` **fixed** — Codex is right: `converge()`'s `bind()` returns only `boolean` (`adoptSurvivor` in `packages/surface-daemon-supervisor/src/endpoint.ts:598` returns `false` for BOTH a fresh spawn AND the non-skew-connect-failure "degraded, left standing" path), so `converge.ts` mapping `false → {kind:"spawned"}` overclaimed a fresh spawn the endpoint cannot attest. (Codex slightly overstated the set — `recycled`/`refused` are NOT reachable at those two sites; they come from the `recycle`/`refuse` DECISION branches, which return their own kinds regardless of the boolean. The real ambiguity is spawn-vs-degraded.) Fixed honestly by renaming the outcome kind `spawned → not-adopted` (`packages/surface-daemon-supervisor/src/convergence/converge.ts:74`, return sites `:119`, `:133`) with a doc note that the boolean can't resolve spawned-vs-degraded and the endpoint surfaces the degraded case itself via `onStatus`. No endpoint surgery (respects deliberate-decision 1). Neither caller read `spawned` (padi `padiBinding.ts:769` returns `outcome.kind==="adopted"`; kaval `ptyHost/index.ts:296` reads `adopted`/`mismatch-reported`), so behavior is unchanged — the fix is pure type honesty. Test updated (`converge.test.ts:82`); 19 convergence tests green.
- `F2` **partial** — Fixed the drift, disputed the restructure. FIXED: the full log LINE was not byte-identical to the pre-extraction `#1670` breadcrumb (`a823e0836:packages/server/src/padiBinding.ts:519`) — the new caller-emitted line had dropped the structured `running` identity object and the "; store hashes don't order, so this fires at most once per binder boot)" suffix. Restored both at `packages/server/src/padiBinding.ts:762` so the whole line matches the original again (the grepped prefix `padi build change on boot: running=<hex> expected=<hex>` was already byte-identical — that is what "byte-identical" in the rationale meant, and the padi-upgrade VM arm greps only that prefix). DISPUTED the "emitted after the follow-on bind can skip it" restructure ask: `converge()` UNCONDITIONALLY returns `{kind:"drained-replacing", axis:"build", ...}` after a build-drain decision (`converge.ts:199-204`), regardless of the follow-on `bind()` result — so the caller emits the breadcrumb for EVERY completed build-mismatch drain, including the degraded (bind returned false) case. It is skipped only if `bind()` THROWS, and in that path the endpoint has already emitted `dead`/`degraded` via `onStatus` and the fence is already spent (`converge.ts:164`, marked before the drain await) — so nothing is silently lost and no re-drain can livelock. Exposing a pre-bind event from `converge()` solely to cover a throwing-bind observability gap the endpoint already reports is not worth the added surface.
- `F3` **fixed** — Codex is right: `README.md:220` still said `@kolu/surface-daemon` is "hashed whole into kaval's staleKey", which contradicts this branch (`default.nix` drops the spine from kaval's `behavioralFileset`; `packages/kaval/src/buildId.closure.test.ts` pins the spine as a stable leaf). Rewrote the row to state the #L3 invariant: the spine runs in the kaval binary but is a stable leaf for kaval's currency key (contract-compatible spine change is behaviorally interchangeable by the wire contract; keying the human nudge on it over-fires and costs PTYs — the zest incident), a wire-relevant spine change bumps `PTY_HOST_CONTRACT_VERSION`, and padi keeps the spine in its full closure (cheap auto-drain).
- `F4` **fixed** — All three stale comments confirmed and updated. (a) `default.nix:146` header comment said kaval's fileset is "three roots ... `@kolu/surface-daemon` spine" — contradicted its OWN inline note 15 lines below; rewrote to "two behavioral roots, spine deliberately excluded". (b) `packages/kaval/src/buildId.ts:4` said the staleKey hashes "the daemon-side roots `terminal-protocol` and `surface-daemon`"; corrected to `terminal-protocol` only, spine excluded (#L3). (c) `packages/padi/src/dial.ts:16` named `isBinderNewer`, `bindPadiOnce`, `probePadiSkew` — `git grep` confirms all three are removed from the codebase; replaced with the current mechanism (padi's `ConvergencePolicy` enacted by the shared kit's `converge()` over a version-agnostic identity probe).

Gate: `just check` (pnpm typecheck across all packages + `biome lint --error-on-warnings`) green; convergence suite 19/19 and kaval closure 1/1 green. (`padiBinding.test.ts` fails at import on a pre-existing harness precondition — `KOLU_STATE_DIR` must be set by the runner — unrelated to this diff.)

commit: `053ddd8e8220eb2899fc243031dfd5d8aaecc0b3`



### Round 2

**codex** — approved: `false`

F1, F2, and F3 are resolved. CLAUDE's F2 timing rebuttal is technically sound enough to accept: after the drift fix, forcing a pre-bind callback would add API surface mainly for a throwing-bind observability corner that is already surfaced by endpoint status/error logging. F4 is mostly fixed, but one stale removed-symbol reference remains in the same padi dial documentation, so the branch is not fully clean yet.

Findings:
- `F1` · major · resolved — Resolved. `ConvergenceOutcome` no longer overclaims `spawned`; it now reports `not-adopted` where the endpoint boolean cannot distinguish fresh spawn from degraded/non-adopted paths. (packages/surface-daemon-supervisor/src/convergence/converge.ts:73)
- `F2` · minor · resolved — Resolved. The #1670 breadcrumb fields and message text have been restored to match the old line. I accept the rebuttal on timing: `converge()` returns `drained-replacing` after a completed drain regardless of the bind boolean, and bind throws are already surfaced by endpoint status/error paths. (packages/server/src/padiBinding.ts:762)
- `F3` · minor · resolved — Resolved. The README now states the new kaval currency invariant correctly: `@kolu/surface-daemon` runs in the binary but is a stable leaf for kaval's staleKey, while padi keeps it in its full closure. (README.md:220)
- `F4` · nit · open — The requested stale-comment cleanup is incomplete. The top-level `padi/dial.ts` comment was updated, but this same file still refers to the removed `isBinderNewer` ordering and a generic "binder's skew probe"; that name no longer exists after the convergence-kit extraction. (packages/padi/src/dial.ts:110)

_codex on the rebuttal:_ F1: accepted as fixed; the `not-adopted` outcome now states exactly what the endpoint boolean proves. F2: I accept the timing dispute. My original concern was the old breadcrumb moving after the follow-on bind, but after restoring the old fields/text, the remaining gap is only a bind-throws path that is already reported by endpoint status/error logging and does not affect the build-drain fence. Adding a pre-bind event solely for that is not justified here. F3: accepted as fixed. F4: the three cited spots were mostly updated, but `packages/padi/src/dial.ts:110` still contains the removed `isBinderNewer` name, so this nit remains open.



**claude** — Closed the last open nit: scrubbed the final stale removed-symbol reference from the `PadiDial` doc block in `packages/padi/src/dial.ts`; F1/F2/F3 already resolved by codex in round 2.

- `F1` **fixed** — Already resolved in round 1 and accepted by codex in round 2 (`converge()` reports `not-adopted`, not `spawned`, where the endpoint boolean can't prove a fresh spawn). No further change.
- `F2` **disputed** — Already resolved: the drift (structured `running` identity + "store hashes don't order" suffix) was restored in round 1, and codex accepted the timing rebuttal (`converge()` returns `drained-replacing` after the completed drain regardless of the bind boolean; a throwing bind is already surfaced by the endpoint's `onStatus` `dead`/`degraded`). No pre-bind event added — not worth the API surface for that one already-reported corner.
- `F3` **fixed** — Already resolved in round 1 and accepted by codex (README row now states the L3 kaval-currency invariant: the spine runs in the binary but is a stable leaf for kaval's staleKey, padi keeps it in full closure). No further change.
- `F4` **fixed** — This round's real work. Codex was right: the `PadiDial` doc block at `packages/padi/src/dial.ts:108-113` still named the removed `isBinderNewer` ordering and a generic "binder's skew probe". Rewrote it to name the current caller judgement: `connectPadi` applies the `isContractVersionCompatible` gate then holds/refuses, and the binder's convergence probe `probePadiForConvergence` (the actual current symbol, per `packages/server/src/padiBinding.ts:212`) reads identity for padi's `ConvergencePolicy` to drain or leave be. `git grep` confirms zero remaining `isBinderNewer` / `bindPadiOnce` / `probePadiSkew` / "skew probe" references anywhere in source.

Gate: `just check` (pnpm typecheck across all packages + `biome lint --error-on-warnings`, 1067 files) green.

commit: `194d94c76fed2a37c124045288353933c740f519`



### Round 3

**codex** — approved: `true`

All previously raised issues are resolved. I verified the current `PadiDial` doc block now describes the current `connectPadi` / `probePadiForConvergence` flow and no longer references the removed skew-ordering helpers. No new regression was introduced by this round's change.

Findings:
- `F1` · major · resolved — `converge()` reported `spawned` from a boolean endpoint result that could only prove adopted vs not adopted, so the typed outcome overclaimed what the endpoint could attest. (packages/surface-daemon-supervisor/src/convergence/converge.ts:74)
- `F2` · minor · resolved — The extracted padi build-change breadcrumb had drifted from the pre-extraction log shape, and I questioned whether post-bind emission could skip observability in a follow-on bind failure. (packages/server/src/padiBinding.ts:762)
- `F3` · minor · resolved — The architecture table still described `@kolu/surface-daemon` as hashed whole into kaval's stale key, contradicting the new kaval currency policy. (README.md:220)
- `F4` · nit · resolved — Stale comments still referenced removed skew-ordering symbols / old probe wording after the convergence extraction. (packages/padi/src/dial.ts:108)

_codex on the rebuttal:_ F4 is now fixed: `packages/padi/src/dial.ts:108` describes the dialed-but-unjudged result, names `connectPadi` using `isContractVersionCompatible`, and names `probePadiForConvergence` for the binder path. I also verified there are no remaining `isBinderNewer`, `bindPadiOnce`, `probePadiSkew`, or `skew probe` references in the relevant TypeScript sources. The earlier F2 timing dispute remains accepted: the completed drain path returns `drained-replacing` regardless of the follow-on bind boolean, and the throwing-bind corner is already surfaced through endpoint status rather than requiring extra convergence API surface.

srid:

`--no-elegance` · base `30ea598f` → HEAD (the L3 convergence-kit delta)

| Pass | Found | Actioned |
| --- | --- | --- |
| Rules | 1 (`prefer-ts-pattern`) | **Skipped, documented** |
| Fact-check | 1 (**confirmed vs source**) | **Fixed** — commit `916c7cf4d` |

### Fact-check (fixed) — `converge()` discarded the real `bind()` result on recycle/refuse

`converge()`'s `recycle`/`refuse` arms called `await bind()` but threw away its boolean, hard-coding `{ kind: "recycled" }` / `{ kind: "refused" }`. But the endpoint's bind (`adoptOrEnsure` / `adoptOrSpawnOrRefuse`) re-checks the primary **and** the W2.2 adopt-hint the single probe never saw — either can **adopt** a compatible survivor (also a probe↔connect race) — and the pre-kit code used `adoptOrEnsure`'s *real* boolean. So the hard-code was a latent byte-identity gap: a recycle whose hint actually adopted would wrongly report not-adopted, so the caller would park the session (`onNotAdopted`) instead of reconciling the survivor's live PTYs (`onAdopted`).

**Fix:** `recycled`/`refused` now carry `adopted: boolean` threaded from the real `bind()`; a new `outcomeAdopted()` helper reads it uniformly across every outcome kind, and both callers use it. Adds the untested cases the police named (bind returning `true` under recycle/refuse). This *restores* byte-identity. supervisor 55 tests green.

### Rules (skipped, documented) — `prefer-ts-pattern`

The `decide`/`converge` discriminated-union switches use manual `switch` + `const _exhaustive: never` rather than `ts-pattern`'s `match().with().exhaustive()`. Skipped deliberately: adding `ts-pattern` as a dep of the intentionally lean `@kolu/surface-daemon-supervisor` (kept minimal for the `odu serve` second tenant, guarded by `deps.closure.test.ts`) is worse than the dependency-free switch, which is **already** compile-time exhaustive-checked — the same safety `.exhaustive()` gives.

srid:

The **x86_64-linux** lane ran on **`kolu-ci-3`** (`dev-x86-64-linux-08`) — commit `916c7cf4d`, **passed** ✓

- **Lane wall** (pipeline): **4m57s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 3m28s | ✓ |
| `ci::nix` | 3m6s | ✓ |
| `ci::home-manager` | 2m59s | ✓ |
| `ci::smoke` | 1m36s | ✓ |
| `ci::flake-check` | 1m30s | ✓ |
| `ci::pnpm-hash-fresh` | 1m10s | ✓ |
| `ci::atlas-sync` | 1m1s | ✓ |
| `ci::unit` | 1m0s | ✓ |
| `ci::install` | 54s | ✓ |
| `_ci-setup` | 34s | ✓ |
| `ci::biome` | 15s | ✓ |
| `ci::fmt` | 14s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/916c7cf/timings.jsonl`; pool state is a live `flock` probe.</sub>

srid:
## ✅ L3 convergence-kit — done, green on both platforms, ready for review

Built end-to-end and driven through the full gauntlet. **Not merged** — held for review (srid merges).

### Stages (each a commit, tests green at each step)
1. `readBakedIdentity` recipe (TS) — closure-neutral.
2. The convergence kit — pure `decide()` policy table + `converge()` typed outcome + the three make-illegal-unrepresentable pins.
3. Nix `mkDaemonIdentity` unify — **hash-neutral, verified byte-identical** (kaval `8379a79a…` new-recipe vs old-inline over the same source).
4. kolu-server converges padi via the kit.
5. padi converges kaval via the kit (+ kaval `system.version` shape-pin).
6. kaval currency behavioral slice — the single visible root-set diff (hash change **by design**).

Mid-flight, merged master (W2.3 #1665, the dial carve) and reconciled the `default.nix` conflict.

### Gates passed
- **VM adoption suite byte-identical pin — all 6 arms green, TWICE**: on a leased KVM box against the pre-gauntlet HEAD, and again in **`ci::home-manager` on BOTH platforms** against the final post-gauntlet HEAD (`916c7cf4d`). `adopt · skew · currency · upgrade · reboot · padi-upgrade` — including the W2.2 legacy-port `adoptHint` recycle (the policy-driven-bind fix) and the #1670 `padi-upgrade` drain through the kit.
- **CI green on both platforms** — linux@kolu-ci-3 + darwin@rasam, 34 checks (incl. `ci::e2e`, `ci::nix`, `ci::unit`, `ci::pnpm-hash-fresh`) all pass.

### Review gauntlet (over the combined delta — the kit + the #1670 fix as refactored in)
- **lens** (lowy ⇄ hickey) — consensus, 1 round. The kit is genuine volatility-based decomposition; 2 fixes applied: deduped `ConvergeLogger` into `@kolu/surface-daemon`'s `Logger`, moved reason-prose out of the pure `decide()`.
- **codex** (xhigh) — consensus, 3 rounds, 4 findings resolved: `converge()`'s `spawned` outcome made honest (`not-adopted`); the #1670 breadcrumb's structured fields restored; the README architecture table updated to the new kaval-currency invariant; stale comments referencing removed symbols cleaned up.
- **simplify** — 1 fix (hoisted the 4×-duplicated skew-log context); reuse/efficiency/altitude clean.
- **code-police** — 1 fix (confirmed vs source): `converge()`'s `recycle`/`refuse` arms discarded `bind()`'s real boolean — now threaded through `outcomeAdopted()`, which *restores* byte-identity in the adopt-hint/race case; 1 documented skip (`prefer-ts-pattern` — adding the dep to the lean supervisor is worse than the already-exhaustive `switch`).

### ⚠️ One-time expected kaval nudge (heads-up in the PR body)
Because stages 1–4 legitimately moved code within kaval's hashed closure, the **first** deploy after this lands fires the kaval "update available" nudge **once** — the closure genuinely changed. This is **not** the zest false-positive returning (that was a spine-only change with no behavior delta; this is real motion). Thereafter, a spine-only change no longer nudges.

### No drishti PR
The surface-daemon stack is not drishti-consumed (its overlay hydrates only `surface`/`surface-app`/`surface-nix-host`), and no `packages/surface/` file changed.

PR #1675, opened 2026-07-04
W3.1 — the remote binding: kolu-server consumes padiSurface over ssh

srid:
**Bind a padi one ssh hop away, and the whole canvas becomes that machine** — terminals spawn there, the Code tab reads its repos, paste/upload/transcripts land there, byte-identical to a local canvas. This is W3.1 of the [padi plan](https://github.com/juspay/kolu/blob/RT-fable/docs/atlas/src/content/atlas/padi.mdx) (§W3.1): the named path is **`padiSurface` consumed over ssh**, behind a single off-by-default knob. No switching UI yet — that's W3.2; today it's one all-or-nothing `KOLU_PADI_HOST=<ssh host>`, unset ⇒ today's local binding, byte-identical.

The whole point is **reuse, not reinvention**: the remote arm slots into the exact `RemoteMirrorSession` seam the local `PadiBindingSession` plugs into, rides the same `getHostSession` stack `kaval-tui --host` uses, and re-runs `@kolu/padi/dial`'s control-core handshake over the ssh-bridged link instead of a local socket.

### How it fits together

```
KOLU_PADI_HOST set ─▶ getHostSession({ binary: "padi" })   # buildAgentCommand runs it as `padi --stdio`
                        │  resolveSystem (arch probe) → PADI_AGENT_DRVS_JSON[system]
                        │  provisionAgent (nix copy --derivation + realise)
                        ▼
   ssh <host> padi --stdio ─▶ frontDaemonOverStdio (adopt-or-spawn the durable padi)
                        │        └─ padi owns its kaval owns the PTYs (all on the host)
                        ▼
   control-core hello + skew refusal ─▶ scopePadiSurface ─▶ reServeSurface
                        │                                      (same seam as local)
                        ▼
                    browser  (/surface/padi/*)
```

### What this PR builds

- **`padi --stdio`** (`packages/padi/src/stdioBridge.ts`) — the durable-daemon FRONT, the twin of kaval's, over the shared `frontDaemonOverStdio`. The spawn re-execs `padi` minus `--stdio`, so the daemon that comes up runs `runPadiDaemon` and owns its kaval — the PTYs outlive the ssh link (detach → reattach).
- **The binding** (`packages/server/src/remotePadiBinding.ts`) — a `RemotePadiSession` that adapts the ssh `HostSession` into `BoundPadi`: the control-core `hello` handshake + typed skew refusal on each fresh spawn, scoping to `.surface.padi`, identity readouts, and `drainBoundPadi` over the combined client. A shared `BoundPadi` interface unifies the local and remote arms at the `index.ts` composition root — the re-serve and router are identical; only the session's front differs.
- **One drv provisions both daemons** — an arch-keyed `{ system → padi .drv }` map (`PADI_AGENT_DRVS_JSON`, the retired `pulamAgentDrvsJson` pattern) baked onto `koluBin`. Because padi's wrapper already bakes `KOLU_KAVAL_BIN` (kaval rides *inside* padi's closure), provisioning that one drv ships both.
- **The dialog tells two hosts apart** — under a remote bind the Kaval/Padi dialogs read the bound padi's honest `hello` identity (a new `daemonInventory.boundPadi`), show the socket as `ssh · <host>`, and LABEL this machine's local daemon scan "not the bound host" — so a laptop running its own local padi never reads its daemons as the remote's (srid's FINDING 1).

### Convergence: hand-mirrored, same policy as the local kit

adopt-or-spawn + re-adopt fall out of `getHostSession` + `frontDaemonOverStdio` by construction: kill the remote padi → the reconnect respawns it; restart kolu-server → it re-adopts the still-running daemon, PTYs intact. `#1313` still holds — a mere dial never kills a running padi; only `drainBoundPadi` (the "restart" verb) drains it, over the frozen control core.

L3 (#1674) is **merged into this branch**. The remote arm **hand-mirrors** the local two-axis convergence (contract skew · build mismatch) keyed on the same facts, and holds the **same outcome policy** as the local kit. Drains are **instance-keyed + bounded** (a blip re-drains the same never-exited instance; a different wrong instance is the treadmill → the budget is spent). When a **build** drain can't win — a contested host respawning the old build — **both arms ADOPT the resident stale build so the canvas keeps working** (the local kit's fence-spent→adopt row; here it's `adoptStale`, an *emergent owner/guest*: solo host → first drain makes us owner, contested host → we yield and ride what's there). An **incompatible contract** is the one thing neither arm can ride, so a contract drain that can't win goes `unconverged`. Ledger **L23** is now just the mechanical `decide()`-unification.

### Nothing behind the scenes: every degraded bind is a visible dialog state

The whole point of the Padi/Kaval dialogs is that you can *see* what went wrong. Every anomalous state the remote binder can enter is surfaced as a **standing banner** in the Padi dialog — riding `boundPadi.convergence` (the one binder→dialog channel orthogonal to the liveness gate, so an *adopted-stale* keeps the canvas connected yet still says why):

| state | dialog shows |
| --- | --- |
| **adopted-stale** (build mismatch, riding resident) | ⚠ "Build mismatch — riding the resident daemon" · running `<build>` · expected `<build>` · reason — *above a still-green status* |
| **skew-refused** (incompatible contract) | the contract-skew reason |
| **unconverged** (newer-contract drain didn't take) | the drain-failure reason |
| **link-failed** (host unreachable / provisioning failed) | the terminal failure reason — not a bare disconnected dot |

Plus **both dialogs name the bound host prominently** — a "bound to ssh · `<host>`" identity line (local bind unchanged, no line). *(Surfacing the LOCAL arm's convergence is an L23 follow-up — the shared kit collapses the stale-vs-fresh distinction its adopt would need.)*

> **The ssh user is the daemon owner.** `ssh <host> padi --stdio` runs as the ssh identity, and padi (like kaval) serves its socket in a `0700` owner-only runtime dir — so two ssh users get two isolated padis by construction. Enforced on the remote (padi/kaval refuse a non-private dir), documented in `packages/padi/README.md`. Pick your ssh user deliberately.

### Test coverage of the ssh leg

| Test | What it proves | Runs in |
| --- | --- | --- |
| `RemotePadiSession` unit (28) | handshake · scope · skew refusal · drain · reconnect · cursor stability · the knob · instance-keyed build+contract drain (blip re-drain · anti-livelock treadmill · exhaustion → unconverged) · honest-null client on skew | CI |
| `dial.test.ts` stdio-front (3) | `padi --stdio` + `frontDaemonOverStdio` + `stdioLink` — the byte relay ssh rides, minus ssh | CI |
| `remotePadiSsh.test.ts` (`just e2e-ssh`) | the FULL ssh hop: provision → `padi --stdio` → handshake → terminal round-trip → typing-echo latency → drain-and-converge → **a LIVE agent-state transition** (`awaiting_user → waiting`) | pu box (self-ssh; guarded on a real non-loopback host) |
| `just e2e-ssh-2box <boxB>` | the transport lane against a GENUINELY-DIFFERENT host + a full-kolu-server assertion that `daemonInventory` publishes `boundHost`=`<boxB>` + a populated `boundPadi` (FINDING 1, enforced) | pu box (two-box) |

> **CI does NOT run the ssh leg** — no sshd in the build sandbox. The ssh lane is a pu-box recipe; its transcript is the recorded evidence (not a silent cap). The lane is **DESTRUCTIVE** (killAll's + drains the target's padi) and refuses without `KOLU_E2E_SSH_DESTRUCTIVE_ACK=1` so a mistyped host can't murder a workstation's terminals. Agent-state Cucumber scenarios stay local-only by design (their fixtures are scattered local `fs` writes; the transport axis is the ssh lane's job).

_The pu-box end-to-end validation transcript, the two-box FINDING-1 evidence (frames + browser screenshots), and the over-ssh latency reference land as `## Evidence` comments; see also the evidence-corrections comment (Symptom A is unexplained, not give-up-gate)._

### Try it locally

```sh
KOLU_PADI_HOST=<your-ssh-host> nix run github:juspay/kolu/W3.1
```

---

### Known issue — gray Kaval chip across a remote-padi rebind (tracked, out of scope)

Under a remote binding, after the binder drains + respawns padi, the browser's **Kaval status chip stays gray/stale until a page reload**. This is **pre-existing `@kolu/surface` behavior that W3.1's rebind merely exposes** — not a defect in this PR, which is a pure consumer.

**Convicted on a box** ([frame trace](https://github.com/juspay/kolu/issues/1681#issuecomment-4884124100)): the **client** is at fault, not the server. Across a live drain→respawn the server *did* re-publish the fresh `daemonStatus` frame onto the same held-open WebSocket subscription (never reconnected, never completed) — the browser's standing subscription / reactive `byKey` mapping (`createSubscription` in `@kolu/surface`) drops it. The "server stops re-publishing" theory was falsified.

Fix tracked in **#1681**, to land as a **separate small surface-stack PR** (it trips the surface.md drishti gate, so it belongs in a PR whose whole subject is that change). That PR also carries a distinct-but-real **cell equals-gate** drop (`server.ts:1407`) that affects this PR's own adopted-stale banner / connection recovery across a rebind.


srid:
## Evidence — the ssh leg, end-to-end on a fresh pu box

`srid` is asleep; this pu-box run **replaces manual QA**. A clean NixOS Incus box (`toor@10.47.48.150`), provisioned from scratch off this branch. The box sshes to its **own non-loopback IP** — real `sshd`, real `ssh`, real `nix copy` — so the transport is genuinely exercised (only the wire is loopback). `127.0.0.1`/`localhost` are refused by the test guard precisely because `isLocalHost` would bypass ssh.

### 1 · Automated ssh-leg integration test (`remotePadiSsh.test.ts`) — 3/3 green over the real ssh hop

```
[ssh] hello: padiSurface=1.1 controlCore=1.0 stateRoot=/home/toor/.local/state/padi
[ssh] terminal round-trip OK — echo landed over the ssh leg
[ssh] typing-echo over ssh — p50=2.69ms p90=4.87ms p99=14.38ms (n=60)
[ssh] post-drain hello: startedAt before=1783135237544 after=1783135240904 (fresh boot ⇒ respawn)
[ssh] convergence OK — the leg recovered after a drain

 ✓ src/remotePadiSsh.test.ts (3 tests) 6179ms
   ✓ provisions padi over ssh, handshakes the control core, and round-trips a terminal
   ✓ records typing-echo latency over the ssh leg
   ✓ converges across a drain: a drained remote padi respawns and the leg recovers
 Test Files  1 passed (1)
```

Note the hello: the remote padi spelled **its own** default state-root (`/home/toor/.local/state/padi`) on its host — the binder passed nothing, exactly as designed.

### 2 · Typing-echo bench over the ssh leg — the W3 latency reference

| | p50 | p90 | p99 |
| --- | --- | --- | --- |
| **over ssh** (loopback, same box) | 2.69 ms | 4.87 ms | **14.38 ms** |
| local baseline ([padi-latency-baseline](https://kolu.dev/atlas/padi-latency-baseline.html)) | 2.14 ms | — | 4.36 ms |

A **reference, not a gate** (network-dependent by design). The p50 sits right on the local baseline; the p99 tail carries the ssh + relay hop. A real network adds RTT on top.

### 3 · kolu-server binds the remote padi (`nix run .#koluBin` with `KOLU_PADI_HOST`)

```
INFO: binding a REMOTE padi over ssh (KOLU_PADI_HOST set) — the whole canvas is this host {"host":"10.47.48.150"}
INFO: kolu listening {"address":"http://127.0.0.1:8199","rss":"228MB"}
INFO: remote padi session {"line":"10.47.48.150: already provisioned ... — skipped copy"}
INFO: remote padi session {"line":"connection: copying → connecting"}
```

### 4 · Convergence — kill padi, watch it respawn/re-adopt

Established a terminal `78fa3369…` on the remote padi (echo `PHASE3MARK`), `startedAt=1783135240904` (T0).

**kolu-server restart re-adopt (sessions intact):** booted koluBin, killed it (≡ a redeploy), booted it again — then dialed the remote padi:
```
{"startedAt":1783135240904,"id":"78fa3369-…","marked":true,"found":true}
```
`startedAt` **unchanged** across both restarts ⇒ kolu-server **re-adopted** the running daemon, never respawned it; the terminal (and its scrollback) survived.

**Hard SIGKILL of padi (the adversarial kill):**
```
padi pid=8014 (gate /run/user/1000/padi-d765be621f53fc73/padi.pid) → kill -9 → dead
fresh dial: {"startedAt":1783135519195,"id":"78fa3369-…","marked":true,"found":true}
```
A **new** `startedAt` ⇒ `frontDaemonOverStdio` **respawned** a fresh padi; the terminal `78fa3369…` **still carries `PHASE3MARK`** ⇒ the fresh padi **re-adopted the surviving kaval PTY**. The PTYs live in kaval, so they outlive a padi kill — convergence by construction.

### Done-criteria coverage

| criterion | evidence |
| --- | --- |
| provision padi+kaval from scratch via the dial | §1 (test 1), §3 ("already provisioned … skipped copy" = warm short-circuit) |
| bind the canvas to the remote host | §3 (koluBin "the whole canvas is this host"), §1 round-trip |
| create terminals + type (bench) | §1 round-trip, §2 latency |
| kill padi → convergence respawn/re-adopt | §4 (drain-respawn **and** hard-SIGKILL respawn + PTY re-adopt) |
| restart kolu-server → re-adopt, sessions intact | §4 (`startedAt` unchanged, terminal survived) |

_Automated CI coverage (off-box): `RemotePadiSession` unit (9), `dial.test.ts` stdio-front round-trip (3). Full per-axis parity over ssh (Code tab · preview · paste/upload · transcript · sleep/wake · worktree) rides the existing cucumber suite with the new `KOLU_E2E_PADI_HOST` passthrough._

srid:
## Evidence — full cucumber suite over ssh + the #1670 drain

Follow-up to the pu-box evidence above, after the coordinator review. All on the same box (`toor@10.47.48.150`, self-ssh via the box's non-loopback IP — `localhost`/`127.0.0.1` are refused, since `isLocalHost` would bypass ssh).

### The full suite, over ssh — **423 / 493 passed (86%)**

```
493 scenarios (70 failed, 423 passed)
4414 steps   (70 failed, 281 skipped, 4063 passed)
```

The *whole* existing suite, single-worker, with the server bound to a REMOTE padi (`KOLU_E2E_PADI_HOST`) — the parity proof: PTY echo, Code-tab git/diff/tree, binary preview, paste/upload, transcript export, sleep/wake, worktree, sub-terminals, the workspace switcher, session-restore, kaval-daemon, reconnect — all exercised over the ssh leg.

> A first run scored 219/493 — an artifact, not the truth: the remote binding warms **async** (fail-open by design; ssh provisioning takes seconds), and the many scenarios that *restart the server* mid-run (reconnect · session-restore · @kaval-restart) raced the warm-up, cascading `killAll` → "no live upstream link" 500s (1793 steps *skipped*). Fixed by waiting for a live remote padi on **every** server start (`startServerChild`), not just boot. The clean re-run is above.

### The 70 residual failures — triaged, **zero transport lies**

| category | what | verdict |
| --- | --- | --- |
| **ssh-latency timeouts** | Playwright `waitForFunction` / `locator.waitFor` exceeding the harness's tight ~20s waits — heaviest in `code-tab` (diff/preview render, each op an extra ssh round-trip) | the behaviour is **correct, just slower** over ssh — an env/timeout artifact, not a defect |
| **agent-state indicators** | `claude-code` / `codex` / `opencode` Dock-pip scenarios (`Expected … indicator state …`) | the **mock agent fixtures** (`KOLU_CLAUDE_SESSIONS_DIR` / `KOLU_CODEX_DIR` / `KOLU_OPENCODE_DB`) are set in the **local** server's env and don't cross the ssh boundary; the remote padi derives agent state on the *box*, where those fixtures aren't. A fixture-locality artifact (the "scenario assumes a local fixture that now lives on the remote" case) — documented, not brute-forced |

Both are inherent to running the *existing* local-fixture suite against a shared remote daemon, not to `padiSurface`-over-ssh. The 423 passes are the lie-catcher: the remote canvas behaves like the local one, one ssh hop away.

### The #1670 build-mismatch drain, remotely (coordinator Q3)

A binder expecting build **B1** (provisioning D1) bound a running **B0** padi (D0) — the flagship "redeploy kolu with a newer padi, box keeps its old daemon" scenario:

```
INFO: padi build change on boot: running=003079d3… expected=2c559b7e…
      — draining the survivor once … respawning this binder's own build (#1670)
result: { afterBuildId: "2c559b7e…" (=B1), upgraded: true, terminalSurvived: true }
```

One drain, respawn on the new build, **kaval PTYs survived** the upgrade. Adversarial verification of the drain caught a HIGH bug — an intended drain rejects the client, so `markConnected` never fired and the drain's clean exit counted against HostSession's give-up gate, risking terminal `failed` with no respawn after prior failures; fixed by marking connected on the successful hello. `RemotePadiSession` has **19 unit tests**.

srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `90b4fd670666`

### Applied (3)
- `lowy-1` drv-map parse+arch-resolve is a duplicated receptacle, hand-rolled inline — commit `e6215a590`
- `hickey-1` makeResolvePadiDrv complects static-config validation with per-host arch resolution (and re-rolls parseDrvBySystem) — (uncommitted)
- `hickey-2` RemotePadiSession bind-state fragmented across `identity` + `skewError` (illegal combination representable) — commit `bcf33a9d4`

### Agreed — no change (4)
- `lowy-2` Electricity-test assessment: the local/remote split is a correct receptacle (packages/server/src/padiBinding.ts:434)
- `lowy-3` Shared receptacle co-located inside one of its two implementations (packages/server/src/remotePadiBinding.ts:63)
- `lowy-4` e2e readiness gate probes liveness with a destructive mutation (packages/tests/support/hooks.ts:875)
- `hickey-3` Local arm's padi identity fragmented into three parallel nullable fields; the remote arm proves the collapsed shape (packages/server/src/padiBinding.ts:486)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `90b4fd670666`

### Round 1

**codex** — approved: `false`

Not approved. The remote binding shape is close, but the current branch still has one blocker and several major issues in the remote path: previews can read the wrong machine's filesystem, restart/drain can report success without a closed daemon, drain polling can hang past its advertised ceiling, and daemon inventory still reports local state while bound remotely. I did not run tests because this review was explicitly read-only.

Findings:
- `F1` · blocking · open — When `KOLU_PADI_HOST` is set, `repoRootForTerminal` is answered by the remote padi, so `repoRoot` is a path on the ssh host. The route then passes that path to `previewFile()`, which reads the kolu-server machine's local disk. A remote terminal preview can therefore fail or serve unrelated local bytes from the same absolute path, which violates the remote-host contract and is a wrong-data path. (packages/server/src/index.ts:381)
- `F2` · major · open — `drainBoundPadi()` drains `host.currentClient()` directly and returns as soon as `control.core.drain()` settles. The control-core drain handler returns after requesting shutdown, before the daemon socket is necessarily closed, so restart can report success while the old padi is still alive. It also bypasses the binding verdict: after a skew refusal, the raw host client can still be drained even though this session is not bound. (packages/server/src/remotePadiBinding.ts:580)
- `F3` · major · open — `drainAndAwaitClose()` intends to cap drain confirmation with `drainCeilingMs`, but each loop awaits `control.core.hello()` without a timeout or abort signal. If the control core or stdio link stays open but stops answering, the first liveness probe can hang forever and the deadline check is never reached. (packages/server/src/remotePadiBinding.ts:532)
- `F4` · major · open — Remote mode still starts the local daemon inventory sampler unchanged: it scans local padi/kaval sockets and computes local active socket paths while `padiSession` may be bound to an ssh host. That can hide the connected remote padi or mark a local daemon as active/in-use by kolu, making the diagnostic UI materially wrong. (packages/server/src/index.ts:562)
- `F5` · minor · open — The local padi binder passes `serverVersion` through `--spawn-version`, preserving `TERM_PROGRAM_VERSION` for spawned PTYs. The remote arm runs only `padi --stdio`, so the re-execed remote daemon falls back to its own commit/dev value. Remote terminals can therefore report a different program version than local terminals from the same kolu build. (packages/server/src/remotePadiBinding.ts:671)
- `F6` · minor · open — The Architecture table still describes padi/server binding as unix-socket-only. This branch adds a second binding mode, remote ssh stdio via `KOLU_PADI_HOST`, so the README architecture section is stale for a structural change. (README.md:207)


**claude** — Agreed with all six findings and fixed each in the remote-padi path (fail-closed remote preview, wait-for-close + bind-gated restart drain, a bounded liveness probe, honest remote-mode daemon inventory, threaded spawn-version, refreshed README), plus completed the branch's own package-boundary seal allowlist for the new file so the tree is green.

- `F1` **fixed** — Real wrong-data path. In remote mode `previewFile()` was handed `repoRoot`, an absolute path resolved on the ssh HOST's registry, but reads THIS machine's local disk → serves unrelated local bytes (or 404). There is no remote preview stream yet (that rides the W3.2 picker slice), so I fail CLOSED per conventions.md rather than read the wrong machine's disk: the route now returns a loud 501 when `remoteHost` is set, before ever calling `repoRootForTerminal`/`previewFile` (`packages/server/src/index.ts:353`). No silent local read of a remote path.

- `F2` **fixed** — Both parts valid. `drainBoundPadi` (a) returned on the drain REPLY (which precedes the daemon socket close → "restarted" reported while the old padi is alive) and (b) reached the raw `host.currentClient()`, bypassing the bind verdict — so an OLDER binder that REFUSED a newer padi could still drain it via the restart button, downgrading it against the newest-wins monotonicity guarantee. Rewrote it (`packages/server/src/remotePadiBinding.ts:600`) to mirror the local arm's `drainViaControlCore` contract: require `bindState.kind === "bound"` (else throw an honest "not bound"), then reuse the class's own `drainAndAwaitClose(combined)` and THROW if the exit isn't observed within the ceiling — so the restart never reports a success that didn't happen, and never a kill. Updated the drain unit tests: the "drains" test now models the link death and asserts it WAITS; added a throws-on-timeout test and a refuses-a-skewed-padi test (`remotePadiBinding.test.ts`).

- `F3` **fixed** — Real ceiling defeat. `drainAndAwaitClose` checked the deadline only BETWEEN hellos, so a single `await control.core.hello()` on a wedged-but-open ssh link (neither answers nor rejects) hangs forever and the deadline is never reached — blocking the whole handshake/restart. Now each probe RACES against the remaining window (`packages/server/src/remotePadiBinding.ts:531`): a REJECT → the drain took (true); a probe that OUTLIVES the window → cannot confirm exit → did-not-take (false), the caller adopts/refuses, never a kill. The probe is pre-caught so a late-settling hello can't surface as an unhandled rejection. Added a "post-drain hello HANGS" test whose mere resolution proves the ceiling now bounds it.

- `F4` **fixed** — Real diagnostic lie. The sampler scans LOCAL padi/kaval sockets and, in remote mode, would still mark a local daemon "in use by kolu" and pin the REMOTE padi's `surfaceVersion`/`buildCommit` onto whatever local socket sits at `activePadiSocket`. Added an optional `boundLocally` gate to `DaemonInventoryDeps` (default true) that, when false, marks NO local daemon active and passes a null active-socket so no remote identity attaches to a local row — the local list still shows leaks, just never labelled kolu's active one (`packages/server/src/daemonInventory.ts:207,236`; wired `boundLocally: !remoteHost` in `index.ts:568`). Sourcing the remote padi's OWN inventory is a later slice; this removes the "mix remote identity with local discovery" defect codex named. Added two `enumerateDaemonInventoryOnce` tests (remote vs local).

- `F5` **fixed** — Right consistency gap. The remote arm ran `padi --stdio` only, so the re-execed durable daemon fell back to its own commit for spawned PTYs' `TERM_PROGRAM_VERSION`, not the kolu app version a local terminal reports. Threaded a `spawnVersion` option through `ensureRemotePadiBinding` and append `--spawn-version <serverVersion>` to `extraArgs` (`remotePadiBinding.ts:685`; passed `serverVersion` in `index.ts:228`). Verified the mechanism end-to-end: `reExecAsDetachedDaemon({ stripArgs: ["--stdio"] })` filters argv by `!strip.has(a)` (`packages/surface-daemon/src/frontDaemonOverStdio.ts:104-108`), removing only the exact `--stdio` token — so `--spawn-version <val>` rides straight through to the durable daemon's `runPadiDaemon`.

- `F6` **fixed** — README architecture table was unix-socket-only. Updated the padi row and the server row to name both binding modes: local unix socket by default, remote ssh stdio via `KOLU_PADI_HOST` (W3.1, off by default, same `BoundPadi`/`reServeSurface` seam) (`README.md`).

- extra (not a codex finding) **fixed** — Pre-existing branch red test: `seal.test.ts` (a) asserts `packages/server/src`'s non-test modules EQUAL the `WEB_SHELL_FILES` allowlist, but the branch added `remotePadiBinding.ts` without allowlisting it, so it failed on clean HEAD. `remotePadiBinding` is web-shell code (the ssh twin of the already-listed `padiBinding`), so I added it to the allowlist with a rationale comment — completing the new file's own seal contract and keeping the tree green.

Verification: repo-wide `pnpm typecheck` green; biome clean on all touched files; full `packages/server` vitest green (12 files pass, 129 tests, the 1 skipped is the CI-guarded pu-box throwaway) — including the two changed test files (37 tests) and the now-passing seal test.

commit: `7246fdd6ded7bc82ba4d3e35b9ab0117fe05af3e`


### Round 2

**codex** — approved: `true`

Approved. I re-inspected the current tree against the six findings from the prior pass, plus the follow-up changes CLAUDE described. F1-F6 are fixed in code: remote previews now fail closed, remote drain is bind-gated and waits for observed close, the liveness probe is bounded, local daemon inventory no longer claims remote daemons as local active rows, spawn-version is threaded through remote stdio startup, and README architecture text names both binding modes. I did not find a new regression introduced by this round.

Findings:
- `F1` · blocking · resolved — Remote file previews previously passed a remote `repoRoot` into local `previewFile()`, risking wrong-machine file reads. (packages/server/src/index.ts:369)
- `F2` · major · resolved — Remote restart previously drained the raw host client and returned on the drain reply, bypassing the bind verdict and not proving the padi exited. (packages/server/src/remotePadiBinding.ts:609)
- `F3` · major · resolved — The remote drain teardown ceiling was previously defeated by an unbounded `hello()` await. (packages/server/src/remotePadiBinding.ts:531)
- `F4` · major · resolved — Remote mode previously reused local daemon inventory as though local padi/kaval sockets were kolu's active bound daemons. (packages/server/src/index.ts:599)
- `F5` · minor · resolved — Remote padi startup previously omitted the local arm's `--spawn-version` forwarding, so spawned PTYs could report the remote padi commit instead of the kolu app version. (packages/server/src/remotePadiBinding.ts:720)
- `F6` · minor · resolved — The README architecture table previously described padi/server binding as unix-socket-only. (README.md:207)

_codex on the rebuttal:_ CLAUDE did not dispute any findings. I verified the agreed fixes in the current tree and accept F1-F6 as resolved.

srid:

Two passes over the W3.1 delta (`--no-elegance`, since `/simplify` ran this gauntlet): **Pass 1 (rules)** raised 6, **Pass 2 (fact-check)** raised 3. All concentrated in `remotePadiBinding.ts` / `hooks.ts`.

### Applied

| Finding | Pass | Fix | Commit |
| --- | --- | --- | --- |
| **transport-correct build-drain** (silent stale-build adopt on ssh link blip) | 2 | Instance-keyed took-detector: fence keyed by daemon `startedAt`, gate adopt on identity change, bounded budget → loud `unconverged` degrade; blip/anti-livelock/ceiling tests | `8610124c2` |
| `caught-error-must-not-collapse-to-empty` | 1 | `drainAndAwaitClose` now threads the drain rejection reason into the "did not take" paths (was `.catch(()=>{})`) | `1ca9e04ff` |
| `callback-fanout-guarded-at-funnel` | 1 | `onState` initial snapshot now goes through the same `safeEmit` guard as every transition | `1ca9e04ff` |
| fact-check: `waitForRemotePadiLive` 120s bound | 2 | Each `fetch` now `AbortSignal.timeout(5s)` so a wedged request can't outlive the deadline | `1ca9e04ff` |
| fact-check: dead-code comment refs | 2 | 4 comments referenced `drainSupersededSurvivor` (deleted by L3) → corrected to `PADI_CONVERGENCE_POLICY` / `decide()` | `1ca9e04ff` |

The first finding was escalated to the coordinator (it intersected the hand-mirror-vs-unify scoping) and fixed per their ruling — see the PR summary + ledger **L23**.

### Skipped (with rationale)

- **`dry-rule-of-three`** (the two-axis convergence hand-mirrored vs L3's kit) — the coordinator-scoped ledger item (L23); the remote arm deliberately keeps its own transport-correct arm until both are unified.
- **`prefer-focused-library`** (zod for the drv-map parse) — the sibling `parseDrvBySystem` hand-rolls the identical validator; matching it keeps the deferred unification trivial. zod is the repo's choice for *persisted*/connection schemas, not these env drv-maps.
- **`prefer-ts-pattern`** (`BindState` `if`-guards vs `match().exhaustive()`) — the dispatch sites are guard-clauses that degrade sanely (a new variant returns `null`/`hostState`), not exhaustive dispatches.
- **`errors-must-log-at-error`** (`onLog` → `info`) — `onLog` is a diagnostic channel; the authoritative failure signal is the connection cell via `onState` (fail-open async-warm design).

srid:
> **⚠️ SUPERSEDED (Symptom A + CI framing).** The give-up-gate mechanism this comment
> attributes to Symptom A is DISPROVEN — at terminal-`failed` the pump nulls the live
> holders, so a dead binding can't echo typing, inconsistent with srid's terminals still
> working; **Symptom A is UNEXPLAINED**, and the `markConnected` fix stands on its own
> merits, not as its fix. The CI-flake framing is also corrected (box-timing, not
> pre-existing-on-master). See the corrections comment:
> https://github.com/juspay/kolu/pull/1675#issuecomment-4883484768

---

## W3.1 — srid's live-test findings: resolution + evidence

srid live-tested a `KOLU_PADI_HOST` remote binding and reported three symptoms. A repro on the **honest two-box topology** (Box A = laptop-analog with its OWN local production padi+kaval, bound over ssh to a **genuinely-different** Box B) at **current HEAD** settled all three. The binary srid exercised was the **pre-gauntlet draft `9004f1f`** (= the raw W3.1 base `4a201f2f` + a changelog commit) — verified to lack the hello-path `markConnected`, the `boundLocally` gate, and the build-mismatch drain.

### FINDING 1 — Kaval/Padi dialog under remote binding

| dialog field | draft `9004f1f` | current HEAD (remote bind) | mechanism / fix |
| --- | --- | --- | --- |
| Kaval status panel | "status unavailable" (see Symptom A) | **populates** (running · build · socket · v5.0 · uptime · mem) — mirrors over ssh | the mirror was always healthy; the emptiness was the give-up-gate death (Symptom A) |
| local daemon list | unlabeled, `active:true` "in use by kolu" | **"Local daemons — this machine, not the bound host (boxb)"**, `active:false` | F4 `7246fdd6d` (boundLocally) + `6f0a7790e` (label) |
| Padi build/socket/version | mixed (local socket + remote version) | build `8610124` · **"ssh · boxb"** · v1.1 — from the bound padi's `hello` | `3d3264d67` (`boundPadi`; was `—`/"unavailable" between F4 and this) |

Raw frames (`kolu.daemonInventory.get`, over the genuine remote binding):
- **before** (`8610124`): no `boundHost`, no `boundPadi` keys at all.
- **after** (`3d3264d67`, remote): `"boundHost":"boxb"`, `"boundPadi":{"surfaceVersion":"1.1","buildCommit":"8610124"}`, local scan `"active":false`.
- **control** (`3d3264d67`, local): `"boundHost":null`, local scan `"active":true`.

Browser screenshots of all three dialog states captured on the two-box rig (shared separately).

### FINDING 2 — agent-state (Padi/Dock) under remote binding

The pipeline is **INTACT** over ssh: with a claude session fixture on Box B, `padi.terminals[id].agent.state` propagated (sensor → member → ssh re-serve → urgency) **and updated live** (`awaiting_user → waiting`, `contextTokens` delta). Enforced now by a dedicated ssh-lane assertion (`remotePadiSsh.test.ts`, `4b2e8bb6d`) that pins that live transition — validated on a pu box (`awaiting_user(6000) → waiting(12000)`). The agent-state **Cucumber** scenarios stay local-only *by design* (their fixtures are scattered local `fs` writes; the only axis that varies over ssh is the transport, covered by the lane assertion) — an explicit note, not a silent cap.

### FINDING 3 — ssh evidence lane

`just e2e-ssh` (turnkey self-ssh, `4170a6f4b`) + `just e2e-ssh-2box <boxB>` (genuinely-different host + a full-kolu-server `boundHost`/`boundPadi` assertion, `984f60824`). **CI does not run the ssh leg** (no sshd in the build sandbox) — stated in the test header and here; the lane is the recorded pu-box evidence.

### Timeline — each symptom → mechanism → fix

- **A ("status unavailable" + Dock agent-state not updating):** both are downstream of a *live* binding. The draft called `host.markConnected()` only from the pump's first frame — never on the hello — so a handshake reject (a contract skew) or connectivity churn counted clean exits toward the give-up budget → terminal `failed` → **no bound padi → the whole re-served surface reads empty**. Fixed by `0ae430d27` (mark on the hello) + `8610124c2` (P2-1). *Honesty:* I did **not** reproduce this trip on either build — the mirror is healthy in every build I tested — so A is *mechanism-attributed, not trip-reproduced*; if that mechanism is unconvincing, treat A as unexplained rather than claimed-fixed.
- **B (false active-marking):** the draft's scan had no `boundLocally` gate → this-machine's daemons marked active + the remote identity pinned onto them. Fixed by F4 (`7246fdd6d`) + labeling (`6f0a7790e`). **Reproduced** (the killer comparison: same daemons `active:true` local vs `active:false` remote).
- **C (Padi dialog identity):** draft showed the mix; F4 fixed the mix but emptied the row; `3d3264d67` reads the bound padi's `hello`. **Reproduced**.

srid:
### Evidence corrections (supersedes the Symptom-A + CI claims in the earlier evidence comment)

Two honesty corrections from the deep review:

**E1 — Symptom A ("status unavailable" + Dock agent-state not updating) is UNEXPLAINED, not give-up-gate.** My earlier give-up-gate theory does not hold: at terminal-`failed` the pump nulls `liveClient.current` / `liveProcedures.current` (`hostFanout.ts:250-256`), so a dead binding **cannot echo typing** — inconsistent with srid's terminals still working — and the re-serve holds its last folded values rather than "going empty." The mechanism is ruled out; I have **no reproduced, code-consistent explanation** for Symptom A on the draft. The `markConnected`-on-hello fix (`0ae430d27` + `8610124c2`) stays on its **own merits** (an intended convergence drain must not spend the give-up budget) but is **not** the fix for Symptom A. **B** and **C** remain the two symptoms with a demonstrated mechanism + reproduced fix.

**E2 — CI flake framing.** The one red on the M1/M2/F2 push was `padiBinding.test.ts "reconnects when padi dies"` (orpc `AsyncIdQueue` teardown `AbortError`). It failed **2-of-2 on `kolu-ci-4`**, green on `kolu-ci-6` + darwin + 3/3 local — **box-timing-correlated**, not "pre-existing on master" (never measured there). Verified: **not my delta** (changed files disjoint from the `padiBinding` reconnect path). Filed standalone as #1679 with both `kolu-ci-4` logs.

srid:
## Evidence — `just e2e-ssh-2box` at final HEAD (real two-box, genuinely-different physical hosts)

Re-run **at the branch HEAD** on two pu boxes that landed on **separate physical machines** (`dev-x86-64-linux-07` and `-08`) — a real cross-host ssh binding, not self-ssh. Code cloned fresh from `origin/W3.1` at `ef4f199416dd6183ccd809c54edd7af822af0a5d`.

```
KOLU_E2E_SSH_DESTRUCTIVE_ACK=1 nix develop --accept-flake-config -c just e2e-ssh-2box boxb
```

**Step 1 — ssh-transport lane** (`remotePadiSsh.test.ts` against the other host): 3 passed, 1 skipped (the self-ssh-only agent-state test, correctly gated off by `KOLU_E2E_SSH_TWO_BOX=1`):

```
 ✓ src/remotePadiSsh.test.ts (4 tests | 1 skipped) 41999ms
     ✓ provisions padi over ssh, handshakes the control core, and round-trips a terminal  38528ms
     ✓ records typing-echo latency over the ssh leg (the W3 reference vs 4.36ms p99 local)  1300ms
     ✓ converges across a drain: a drained remote padi respawns and the leg recovers  2167ms
      Tests  3 passed | 1 skipped (4)
```

**Step 2 — full kolu-server bound to the remote host**, asserted off the parsed `daemonInventory` frame:

```
e2e-ssh-2box: standing up kolu-server (KOLU_PADI_HOST=boxb) on :7099
daemonInventory (bound to boxb): {"json":{"kavals":[],"padis":[],"boundHost":"boxb","boundPadi":{"surfaceVersion":"1.1","buildCommit":"ef4f199","convergence":null}}}
e2e-ssh-2box: PASS — boundHost=boxb, boundPadi.surfaceVersion=1.1 over the genuinely-remote binding
```

`boundHost="boxb"`, `boundPadi.buildCommit="ef4f199"` matches the checked-out HEAD — the remote padi's honest identity rides `boundPadi`, and the dialog can label the local scan "not the bound host". Recipe exited 0; both boxes destroyed.

PR #1677, opened 2026-07-04
Clamp focus past all departing tiles when the top level empties in one frame

### LINKED ISSUE #1667: All top-level terminals departing in one frame can focus a dead terminal id
Found by the W4 comprehensiveness sweep (W1-era residual recorded on #1652, verified still open on master).

**Mechanism:** `packages/client/src/terminal/useActiveReconcile.ts:~102-111` — the auto-switch computes survivors as `topLevelBefore.filter(x => x !== id)`, excluding only the single id being processed, never cross-checked against the post-removal live set or the full `departed` array. With `[A,B,C]` all departing in one frame (active A), `activeId` lands on B — a departed id — instead of clamping to null. No downstream guard rescues it (`activate` sets unconditionally; the dead id is echoed to the server via `chrome.setActive`).

**Test gap:** `useActiveReconcile.test.ts:~270-293` covers batch-departure only with `connected:false` (eviction suppressed); the connected N≥2→0 case is untested.

Repro-first: add the failing connected batch-depart test, then fix (clamp to the post-removal live set).

srid:
When **every top-level terminal departs in the same frame** — a close-all, or a batch that all exits at once — kolu's focus hand-off could mark a *just-departed* terminal as active and echo that dead id to the server. The auto-switch picked the "survivor" to focus by filtering only the **single id it was processing** out of the pre-removal order, never the rest of the departing batch. So with `[A,B,C]` all leaving (active `A`), focus walked `A → B → C → B` and settled on `B` — a terminal that was itself gone — instead of clamping to **no active terminal**.

### The fix

Thread the whole set of ids leaving in the frame into `evictTerminal`, so the survivor set is `topLevelBefore` minus **all** of them, not just one:

```
survivors = topLevelBefore.filter(x => x !== id)          // before — single id
survivors = topLevelBefore.filter(x => !departing.has(x)) // after  — the whole batch
```

The auto-switch logic stays in its **one home** (`pickAutoSwitchTarget`) — only the *"what's leaving"* fact is supplied, reusing the reconcile's existing `departed` array. The imperative close path departs exactly one tile, so it mints `new Set([id])` and its behavior is **byte-identical**; only the list-driven batch reconcile changes. A batch where some tiles survive still hands focus to one that's actually still there (the tile now in the removed one's slot).

| Scenario | Before | After |
| --- | --- | --- |
| `[A,B,C]` all depart, active `A` | focus → `B` (dead id, echoed to server) | focus → `null` |
| `[A,B,C,D]`, `A,B` depart, active `A` | focus → `B` (dead id) mid-walk | focus → `C` (live survivor) |

### Repro-first

The test file covered batch departure only with `connected:false`, where eviction is *suppressed* — so the connected N≥2→0 case was untested. This adds it (plus the partial-batch case and direct `evictTerminal` unit tests): **red on the old code** (focus landed on a dead id), green after.

> Honors the repo's design philosophy: no fallback or knob — an emptied canvas reaching `null` active is the *correct* state, not a degraded one (the old dead-id echo was the illegal state). No new abstraction; the fix reuses the reconcile's own `departed` set as the single source of "what's leaving" rather than a parallel scan.

Fixes #1667


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `90b4fd670666`

### Applied (2)
- `lowy-2` pickAutoSwitchTarget's contract requires a caller-maintained cross-argument invariant (filtered list + unfiltered index) — commit `27640e2f5`
- `hickey-1` Correctness depends on an unenforced caller invariant: id ∈ departing — (uncommitted)

### Agreed — no change (5)
- `lowy-1` Electricity test: focus-handoff receptacle is a sound application seam, not misplaced electricity (packages/client/src/terminal/useActiveReconcile.ts:33)
- `lowy-3` Signature asymmetry between the dedup's two entry points (evictImperatively mints the set, evictDeparted receives it) (packages/client/src/terminal/useActiveReconcile.ts:136)
- `hickey-2` Frame-level focus hand-off braided into per-id teardown (packages/client/src/terminal/useActiveReconcile.ts:107)
- `hickey-3` Same moment's departure fact held as both `departed` (array) and `departing` (Set), with a tense-split name (packages/client/src/terminal/useActiveReconcile.ts:239)
- `hickey-4` `departing` is a superset of what the filter can consume (carries sub-terminal ids) (packages/client/src/terminal/useActiveReconcile.ts:239)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `90b4fd670666`

### Round 1

**codex** — approved: `false`

The implementation itself looks correct against the branch goal: the fallback target is derived in one place, imperative single-close behavior remains unchanged, and list-driven batch departures now skip every departing top-level id. I found one coverage gap around a subtle contract the code now depends on. I did not run tests in this read-only review.

Findings:
- `F1` · minor · open — The batch tests only cover the removed active tile at index 0. They do not pin the documented full pre-removal index behavior when another departing sibling sits before the removed active tile. A future refactor could compute the index after pruning other departing ids and incorrectly focus C instead of D for `[A, B, C, D]` with `{A, B}` departing and `B` active, without the current tests catching it. (packages/client/src/terminal/useActiveReconcile.test.ts:26)


**claude** — Agreed with codex's single coverage-gap finding and added a pure-function test that pins the full-pre-removal-order index contract when a departing sibling precedes the removed active tile.

- `F1` **fixed** — Codex is right that the batch tests only exercised the removed active tile at index 0, where an index computed on the FULL pre-removal order and one computed on a pruned order coincide — so nothing pinned deliberate decision (3) (`removedIndex = topLevelBefore.indexOf(removedId)` on the FULL order). Added `expect(pickAutoSwitchTarget([a, b, c, d], new Set([a, b]), b)).toBe(d)` at `packages/client/src/terminal/useActiveReconcile.test.ts:33`. Here `b` is active with earlier sibling `a` also departing: survivors are `[c, d]`, `indexOf(b)` in the full order is `1`, so slot 1 is `d`. A pruned-order index would wrongly land on `c` — so this assertion locks the contract a future refactor could otherwise silently break. This strengthens decision (3) rather than contradicting it, so no dispute. Verified: `useActiveReconcile.test.ts` 19 tests pass; `just check` (typecheck + biome lint) green; `just fmt` made no changes.

commit: `238656c65cfd853ff462f99eba48e81a7a76cf15`


### Round 2

**codex** — approved: `true`

F1 is resolved. The added assertion directly pins the missing case: `removedId = b` with earlier sibling `a` also departing, so the target must be selected using `b`'s index in the full pre-removal order and therefore resolves to `d`, not `c`. I saw no new regression introduced by this round's change.

Findings:
- `F1` · minor · resolved — The prior test coverage did not pin the full pre-removal index behavior when a departing sibling preceded the removed active tile. (packages/client/src/terminal/useActiveReconcile.test.ts:34)

_codex on the rebuttal:_ Accepted. CLAUDE fixed the coverage gap with the exact case needed: `b` remains index 1 in the full `[a,b,c,d]` order even after `a` is filtered out of survivors, so the expected target is `d`. That locks the documented contract and resolves F1.

srid:

Two passes over `git diff 90b4fd670666...HEAD`, scoped to the change.

### Pass 1 — Rule checklist
✅ **No violations** across the full built-in + project rule set. The one shape that superficially resembled a thin-wrapper (`createEvictionDedup`'s forwarding closure in `useTerminalCrud.ts`) was cleared: it's an injection-point callback invoked internally by the dedup, predates this diff, and only newly threads the `departing` arg.

### Pass 2 — Fact-check
One real finding, **fixed**:

- **`evictImperatively` recreated #1667 via the imperative path.** It minted `departing = {id}`, but `removeAndAutoSwitch` reads `store.terminalIds()` **live** — a memo fed only by the server's list-push, *not* updated synchronously on `kill`. So closing tile A then tile B before A's list-drop lands feeds B's eviction a `topLevelBefore` that still contains the already-killed A (with `departing={B}` only), letting the auto-switch re-focus the dead A and echo it to the server via `chrome.setActive`. Both being in `claimed` meant the later list-drops short-circuited in `evictDeparted` — **no self-heal**.
  - **Fix** (`e7c1a7dd9`): pass `new Set([id, ...claimed])` — the dedup already tracks exactly the in-flight "evicted, list-drop not yet landed" set, so focus now clamps past *every* departure in flight, not just the one being processed. Repro-first: added test `(h)` exercising two overlapping imperative closes (red → focus landed on dead A; green → clamps to null).
  - This closes the fix across **both** callers, so the changelog's "clamps focus past every terminal leaving in that frame" is now literally true (resolving the fact-check's secondary "changelog overclaim" note).

Out-of-scope note recorded: `handleKill`'s pre-existing blanket `catch {}` (untouched by this diff) would leave a `claimed` id undrained on a genuine kill failure — a latent issue in code this PR doesn't modify.

srid:
## Evidence

**No visual impact — this fix corrects internal state with no on-screen difference to capture.** I ran the evidence gate carefully and traced the effect to the screen rather than stopping at "it's client code":

The bug was `activate(<a departed id>)` during a same-frame mass departure. Reading `activate` (`useViewState.ts`):

- `activate(null)` (the fix's clamp) sets the active id to `null`; the `activeId` effect **returns early on `null`** — no canvas pan, no MRU write, and **no `chrome.setActive` echo**.
- `activate(deadId)` (the bug) fires that effect: it echoes **`chrome.setActive(<departed id>)` to the server**, inserts the dead id at the front of the MRU, and requests a canvas pan toward a tile that no longer exists.

So the bug's damage is **internal / server-side**: a dead-id `setActive` echo, a stale MRU entry, and a pan intent toward a removed tile. Tracing each to a pixel:

- A **departed** id has no tile (nothing to outline/focus) and no dock row (it's out of the list), so no "active" indicator renders differently.
- In the only UI-reproducible mass-departure scenario — closing everything — you land on the centered **EmptyState** either way; it covers the canvas, so even the errant pan isn't visible.

The trigger itself is a **same-frame reactive race** (N≥2 top-level tiles dropping from the list in one reconcile frame), not a deterministic UI flow: `handleCloseAll` bypasses this path (`killAll` + `store.reset()`), and the only routes in — simultaneous natural PTY exits, or two closes racing one server list-drop — aren't reliably stageable under the e2e harness's `slowMo`. That is exactly why this was a latent bug caught by review, not observation.

The exact behavior is instead pinned by **20 unit tests** in `useActiveReconcile.test.ts` (synthesizing the same-frame departure directly), including the red→green repros:

- `(f)` connected batch departure of every top-level tile → focus clamps to `null`, never a dead id.
- `(g)` partial batch → focus lands on a real survivor, never the departing sibling.
- `(h)` two rapid imperative closes → focus clamps past the still-`claimed` first tile (the code-police finding), not back onto the dead one.

Manufacturing a close-all clip here would show an EmptyState identical to the pre-fix one — a tenuous pixel that proves nothing — so per the evidence gate this is the legitimate skip.

srid:

The **x86_64-linux** lane ran on **`kolu-ci-6`** (`dev-x86-64-linux-03`) — commit `e7c1a7dd9`, **passed** ✓

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 2m25s | ✓ |
| `ci::home-manager` | 2m15s | ✓ |
| `ci::unit` | 1m11s | ✓ |
| `ci::nix` | 50s | ✓ |
| `ci::smoke` | 45s | ✓ |
| `ci::fmt` | 24s | ✓ |
| `ci::biome` | 24s | ✓ |
| `_ci-setup` | 7s | ✓ |
| `ci::install` | 6s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/e7c1a7d/timings.jsonl`; pool state is a live `flock` probe.</sub>

Issue #1681, opened 2026-07-04
Re-served surface: standing browser subscription goes stale across a padi drain→respawn rebind (gray Kaval chip until reload)


srid:
### Symptom (srid, live)

Under a **remote padi binding** (`KOLU_PADI_HOST`), after the binder **drains + respawns** padi (a build-mismatch redeploy / any drain→respawn rebind), the browser's **Kaval status chip stays gray/stale until a full page reload**. A fresh subscription after reload sees fresh truth immediately — the server *has* the data; the **standing** browser subscription never recovers across the rebind. Everything else recovers.

This is **pre-existing surface-stack behavior** that W3.1's remote-padi rebind merely *exposes* — W3.1 (#1675) is a pure consumer. The fix lives here, at the framework layer, not in the consumer.

### Path map (code-read, W3.1 investigation)

- **Gray rule:** `kavalDot(state, live)` (`packages/client/src/kaval/daemonPresentation.ts:92`) → gray when `!live || !state`. `live` = the browser's `app.health().live` over the browser↔kolu-server socket, which stays **open** across a *server-side* rebind → `live` stays **true**. So gray reduces to padi's **`daemonStatus.byKey("local")` going stale/undefined**.
- **daemonStatus is a COLLECTION**, per-key path; its re-serve upsert publishes **unconditionally** — `packages/surface/src/server.ts:1526` (`perKeyBus(k).publish(v)`, no equals-gate). So *by code reading* the collection **should** recover on rebind.
- **The demonstrable swallow is on CELLS:** the re-serve cell fold (`packages/surface-nix-host/src/reServeSurface.ts:416` → `cell.set`) hits the framework equals-gate `packages/surface/src/server.ts:1407` (`if (equalsFn?.(store.get(), next)) return;`), which **drops a post-rebind cell value equal to the pre-drain one**. A surviving re-adopt reproduces an identical value → the already-subscribed browser is never told.
- **No ws reconnect across the rebind** (it's server-side), so `STREAM_RETRY`/`ClientRetryPlugin` never re-subscribes; the client relies entirely on the server pushing frames onto held-open subscriptions. `createSubscription` (`packages/surface/src/createSubscription.ts:111`) runs its source **once** and, on a *clean* server-side stream completion, exits with **no** re-subscribe.

### Two candidate layers (contradiction to resolve on a box)

[...]

PR #1683, opened 2026-07-05
fix(surface): standing subscriptions + re-served cells recover across a padi rebind (#1681)

srid:
Two `@kolu/surface`-stack framework gaps that [W3.1](https://github.com/juspay/kolu/pull/1675)'s remote-padi drain→respawn rebind *exposed* — not defects W3.1 introduced (it's a pure consumer). After the server drains and respawns its workspace daemon (**padi**), every live value recovered **except the ones your browser was already subscribed to**: the **Kaval status chip stayed gray until a full page reload**. Convicted on a box in **#1681** (frame trace: the server re-published the fresh frame; the client dropped it).

Keeps the fixes narrow — one small, testable change per bug, no consumer-facing API removed.

## BUG 1 — the client dropped the fresh post-rebind frame

A held-open surface stream is meant to stay open (snapshot-then-deltas), but it can **end cleanly** when the server re-serves across a rebind (a fresh spawn re-mirrors onto the same downstream router). `ClientRetryPlugin` re-subscribes only on a transport **error**, never on a clean completion — so `createSubscription` ran its `source()` **once** and died on the completion. The snapshot the respawned padi republished never arrived, and the standing subscription stayed dead until a reload.

**Fix:** both leaf primitives (`createSubscription`, `createReactiveSubscription`) re-open the source on a clean completion, via a new shared `consumeReSubscribing` loop. It is:
- **rate-bounded** (a 1s delay, the same reconnect cadence `STREAM_RETRY` uses) so a server that pathologically completes its stream in a loop can't spin a republish storm;
- **terminal on a thrown error** — the retry plugin already retried retriable transport errors inside `source()`, so an error reaching this layer is non-retriable;
- **non-destructive** — the retained value + `pending`/`error` are untouched across the re-subscribe, so recovery is a silent refresh in place (no flash to `undefined`).

## BUG 2 — the re-serve's cell equals-gate swallowed an equal post-rebind value

`ctx.cells.<key>.set` de-dups by `equals` (a no-op write doesn't republish). But when the re-serve rebinds to a fresh spawn (`reServeSurface`), a cell value that comes back **equal** to the pre-drain one must still cross **once** — otherwise a downstream holder (kolu-server's adopted-stale banner / connection recovery) can't tell *"rebound and confirmed"* from *"stale"*. The equal re-fold was silently dropped.

**Fix:** the setter takes `set(v, { force: true })` to bypass the dedup for **one** write; `reServeSurface` forces the **first fold of each spawn** (the rebind epoch). Steady-state de-duplication is unchanged — only the epoch's first fold forces.

## Red→green tests (each red on master, green here)

| Layer | Test |
| --- | --- |
| `createSubscription` | drive a server-side **stream completion + republish**, assert the subscriber observes the fresh frame; and a thrown error is terminal (no re-subscribe) |
| `createReactiveSubscription` | same completion+republish within one input round |
| `reServeSurface` | an equals-gated cell across a `setClient` rebind emits **`[v, v]`, not `[v]`** — and a further equal steady-state write still dedups |

## Gate

Both bugs change `@kolu/surface*`, so per `.claude/rules/surface.md` this ships with a **paired [srid/drishti](https://github.com/srid/drishti) PR**, CI-green, pinned to this PR's final HEAD. Linked below.

## Evidence

Gray-chip recovery e2e on a pu box (real remote-padi rebind, chip refreshes with no reload) lands as an `## Evidence` comment.


srid:
Paired drishti PR (surface.md gate): srid/drishti#86 — re-pins kolu to this branch and validates that drishti consumes the (backward-compatible) surface changes unchanged. Its CI will be re-confirmed green against this PR's **final** HEAD (post review-gauntlet) before merge.

srid:
Parking this per coordinator ruling (srid): **discarding the fix pending a live pu-box reproduction that pins the actual drop layer first.**

The #1681 "client convicted" frame trace was from a since-deleted session instrumenting a **raw replica** client, not the real browser — so its mechanism is a hypothesis, not proven. My own `directLink` finding (the re-served per-key `daemonStatus` `get` stream stays open across a rebind and `reconcile` delivers the fresh frame) contradicts it. So the honest state: the symptom is real and the server re-publish is proven, but the **layer of the drop is unknown**.

Commits parked on branch `parked/surface-rebind-attempt1`. Will reopen with the true mechanism + a red→green test at the actually-convicted layer once the live repro speaks.

PR #1685, opened 2026-07-05
Remote file preview — serve the connected machine's bytes, chunked

srid:
Clicking a file in the Code tab while kolu is bound to a remote machine (**`KOLU_PADI_HOST`**) used to return a deliberate **501** — *"file preview is unavailable while bound to a remote padi"* — so kolu would never serve the **wrong** machine's bytes. That refusal is now gone: the preview route's remote arm **serves the connected host's bytes** by dialing the bound padi's `preview.read` procedure — the same serve-dir read the local arm uses — and reassembles them into serve-dir's streaming `ServeResult` shape, byte- and header-identical to a local read.

The root cause the W3.1 note recorded: `padiSurface.preview.read` (range-capable, forwards over the binding) has existed since W2, but the route shared its *implementation function* (`previewFile`) as a **local-disk** read instead of dialing the procedure — invisible on one machine, wrong-host on two. This PR dials the procedure when bound remotely. Plan of record: the **W3.2 "preview"** bullet in [`padi.mdx`](https://kolu.dev/atlas/padi.html) (the coordinator flips its known-gap to closed on merge; this PR does not touch the note).

### The memory story — why chunks

`preview.read` is base64-wire, so its whole body must materialize in memory once per call (and `readPreview` caps a single inline read at 64 MiB). A multi-GB video forced through one dial would blow both the daemon's and kolu-server's heap. So the remote arm **never asks for the whole file at once**:

```
probe bytes=0-0 ──▶ total size + mime (1 byte, no body)
                    │
parseByteRange(browserRange, total)  ◀── @kolu/serve-dir's OWN parser
                    │
      ┌─────────────┼──────────────┐
   null → 200    {s,e} → 206    "invalid" → 416
      │             │
      └──▶ stream [lo,hi] in ≤ 8 MiB dials ──▶ ReadableStream ──▶ socket
```

Peak heap is **one chunk (8 MiB), not one file**, on both hops. `<video>` seeks still work (byte ranges are honoured over the wire). Range *resolution* is not re-implemented — it's serve-dir's `parseByteRange`, the single source of truth the local read already runs through.

> **Fails loud, never silent.** Every chunk dial is validated: a non-206 status, a `Content-Range` total that changed mid-stream (the file was replaced), or a short chunk **throws**, erroring the stream — so a truncated remote read surfaces as a broken/reset response, never a clean body missing its tail.

### Warming / degraded binding

A binding that isn't a live canvas — **warming** (not yet connected), or a standing **skew / unconverged / linkFailed** degraded verdict — yields a `null` from `padiSession.currentClient()` (`remotePadiBinding.ts` `currentClient`, the same three-way `absent` the local memory-poll reads). The route already answered that with a loud **`503 "padi is not connected"`** *before this PR* — that guard is unchanged. Both the client `await` and the `repoRoot` resolve sit inside the link-fault `try`, so a fresh spawn whose handshake **rejects** maps to the same `503` link-fault, **never** a hang and never an uncaught 500. There is no silent-wait path: the route either serves bytes, streams a loud error, or returns a 503/404.

### Design philosophy

- **Fail fast / no fallbacks / no knobs** — the 501 refusal is *deleted*, not toggled; the remote arm serves the real bytes or errors loudly. No degradation path. (`assembleRemotePreview`'s `chunkBytes` is a test seam with a fixed production default of `REMOTE_PREVIEW_CHUNK_BYTES` = 8 MiB, not a runtime override.)
- **Electricity / volatility boundaries** — the transport volatility is already encapsulated in the bound padi client (`preview.read` forwards over the binding). The chunk-assembly is a *bounded leaf algorithm* in kolu-server's route glue, so it stays in `iframePreviewRoute.ts` — it hides no hard volatility and earns no new `@kolu/*` package.
- **Reuse the source of truth** — `parseByteRange` for ranges, `preview.read` for bytes (no contract change — the procedure was already range-capable), serve-dir's header shape forwarded verbatim from the probe. Local mode keeps the exact `previewFile` read (no added hop).

### Traversal defenses — both layers unchanged

The raw-tail guards in the route (`previewTailFromRawUrl` / `rawTargetFromContext` — the `..`/`%2f`/double-decode defenses) protect **both** arms identically; padi re-enforces its realpath/403 guard host-side inside the read. Neither semantics changed.

### Tests

- **Unit** (`iframePreviewRoute.test.ts`) — driven against the real serve-dir engine via a base64 fake reader: chunk boundaries incl. a partial last chunk, binary byte-exact round-trip across chunks, zero-length file (empty 200 + ranged 416), 206/416/suffix-range resolution, Content-Type passthrough, 404 passthrough, and three loud-failure cases (wrong status, short chunk, size-changed-mid-stream). The old 501 test is deleted.
- **ssh e2e** (`remotePadiSsh.test.ts`, self-ssh) — a real text + binary file previews **byte-exact over a real ssh hop**, ranges included. Run green on a pu box (evidence below).

### Review gauntlet + adversarial lenses

- **lens-debate (lowy ⇄ hickey)** — consensus, 1 round. Shared range→response shaping into `@kolu/serve-dir`'s `rangeResponseHead`, typed `PreviewReadResult` from the padiSurface schema, shared one `getHeaderCI`.
- **codex-debate** — consensus, 3 rounds (xhigh). Hardened the cross-chunk snapshot: serve-dir now emits a strong **ETag** (size+mtime+inode) and the remote arm pins it across every chunk (catches a same-size atomic replace the total-size check misses); full `Content-Range` start/end/total validation; and the artifact-sdk HTML decorator strips the now-stale ETag before rewriting the body.
- **simplify** — dropped a redundant per-chunk `Uint8Array` copy on the hot path and a vestigial `Headers` clone.
- **code-police** — fixed a metadata link fault escaping as an unlogged 500 instead of the logged 503.
- **architecture-first-principles** (state-and-time) — **sound on all five principles.** On P5 (guarantees at the knowing endpoint), the reassembly loop correctly stays in the route: only it knows a cross-boundary multi-read is happening (for serve-dir N is always 1, so the ETag check is vacuous). serve-dir owns the ETag **value**; the route owns the snapshot **guarantee**.
- **perfection-review** (adversarial) — every PR-body claim **verified against the code** (two undersell themselves); no adversary could express the 501/wrong-host defect or a silent truncation/quiet-fallback. Added an isolated unit test for the probe ETag-presence guard (mutation-verified: removing the guard fails the test).

**Deliberately deferred** (defect not expressible today; noted per review):

- **Wrong-host dispatch keys on a process-global.** The read fork (`if (remoteHost)`) and `repoPath` (from `padiSession`) come from two sources that today are driven by the *same* boot constant, so they cannot disagree — a remote binding cannot fall through to a local disk read (verified). When W3.2's **per-view bindings** land, that seam could open; the structural fix (fold the read into a polymorphic `BoundPadi.readPreview` so `previewFile` is unreachable from a remote session) belongs in that PR, co-located with the change that opens the seam — and this PR is fenced off from `remotePadiBinding.ts`/the binding abstraction the switch PR owns.
- **Full-200 truncation detection** rides the stream-error→socket-reset path (no `Content-Length`) — **exactly the same semantics as the local `serveFile` arm**, so it's at parity, not a remote regression; pinning a `Content-Length` on the remote 200 only would *break* the byte/header-identity this PR guarantees.


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `20a346c38c3e`

### Applied (5)
- `lowy-1` Response-assembly (HTTP range semantics) duplicated: serveFile vs assembleRemotePreview — commit `0e11f9de3`
- `lowy-2` PreviewReadResult hand-redeclares the canonical padiSurface contract type — commit `04fecbd31`
- `hickey-1` serve-dir's range→response shaping is re-derived in the remote arm — (uncommitted)
- `hickey-2` Case-insensitive header lookup reimplemented in 4 places — commit `e864bd1e9`
- `hickey-4` PreviewReadResult redeclares the preview.read wire shape — (uncommitted)

### Agreed — no change (3)
- `lowy-3` Remote arm sources its response shape three different ways (symmetry break) (packages/server/src/iframePreviewRoute.ts:250)
- `lowy-4` Route selects transport by branching on remoteHost (wire exposure), but bounded (packages/server/src/index.ts:429)
- `hickey-3` chunkBytes default parameter doubles as a test seam (packages/server/src/iframePreviewRoute.ts:257)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 round(s) · codex reviewed at `xhigh` reasoning effort · base `20a346c38c3e`

### Round 1

**codex** — approved: `false`

The implementation is generally well-scoped and reuses the serve-dir range parser/header shaping as intended, but I found a few correctness gaps in the remote assembly path. The biggest one is that the multi-call chunk loop cannot prove all chunks came from the same file snapshot, so it can silently mix bytes across same-size replacements, unlike the local serve-dir path's one-handle invariant. I did not run tests because the review was explicitly read-only and the relevant tests create fixtures.

Findings:
- `F1` · major · open — Each remote chunk is a separate `preview.read` call, and the loop only checks that the reported total size stays the same. If the file is atomically replaced between chunks with another file of the same size, the assembled response silently mixes versions while still passing every current check. The empty-file path has the same shape: after a 416 probe, a file that appears before the unranged re-read returns 200 with bytes, but line 273 discards that body and serves an empty stream. This breaks the local `serveFile` guarantee that headers and body come from one opened file description. (packages/server/src/iframePreviewRoute.ts:210)
- `F2` · minor · open — Unexpected successful probe statuses are converted with `errorResult()`. If the upstream ever returns `200` for the `bytes=0-0` probe, this returns a successful response with a UTF-8 string body instead of failing loud; binary bytes would be corrupted and the response would no longer be serve-dir-shaped. The invalid-range re-dial at line 285 has the same problem if it does not actually come back as 416. (packages/server/src/iframePreviewRoute.ts:259)
- `F3` · minor · open — Chunk validation parses only the `/total` part of `Content-Range`. A broken upstream could answer `bytes=128-255` when the loop requested `bytes=0-127`, with the same total and byte length, and the assembled stream would accept and emit the wrong slice silently. (packages/server/src/iframePreviewRoute.ts:215)
- `F4` · minor · open — The Architecture section still says the preview route is re-backed by padi's streaming `previewFile` and that `preview.read` is only the base64 wire twin. That is no longer true for `KOLU_PADI_HOST`: the server now assembles the HTTP preview by dialing `preview.read` in chunks. (README.md:218)
- `F5` · nit · open — `rangeResponseHead` is now a public export from `@kolu/serve-dir`, but the package README API reference still lists the old export set. That makes the new shared range-header helper hard to discover and leaves the API docs stale. (packages/serve-dir/README.md:60)


**claude** — Fixed all five findings: closed the cross-chunk snapshot gap with a serve-dir strong validator (`ETag`) validated across every chunk, hardened the probe/re-dial status handling to fail loud, added full Content-Range validation per chunk, and refreshed the two stale doc surfaces.

- `F1` **fixed** — The multi-chunk loop now proves every chunk came from ONE file snapshot, the cross-read analogue of the local read's single-open-handle invariant. `@kolu/serve-dir`'s `serveFile` now emits a strong `ETag` (size + mtime + inode, from the same `stat` that sizes the response) on every streamed 200/206 (`packages/serve-dir/src/index.ts` new `fileETag` + `baseHeaders.ETag` assignment after the isFile check). The remote arm captures it from the probe (`iframePreviewRoute.ts:assembleRemotePreview`, throws if a 206 probe lacks it) and re-checks equality on every chunk (`streamByteRange` — a same-size atomic replace bumps mtime+inode → the validator differs → throws → errors the stream). This completes the author's own decision #4 ("refuses an inconsistent body"), which the size-only check couldn't fully deliver. This is NOT a surface-contract edit: `PadiPreviewReadOutputSchema.headers` is `z.record(z.string(),z.string())` (open map), so the new header flows over the wire with no schema change, and serve-dir is inside the drishti gate. It doesn't touch the chunking design (decision #2/#3) — it strengthens the per-chunk validation the design already runs. serve-dir does NOT honor conditional requests (never 304), so no caching-semantics change. The empty-file sub-case (codex's line-273 concern) is also closed: the unranged empty re-read now throws if it comes back non-empty (file grew between the 416 probe and the re-read) rather than discarding a real body and serving empty. New tests: same-size-swap → validator throw; empty-grow path exercised via the guard. serve-dir 46, padi 10, server 55 all green.
- `F2` **fixed** — `assembleRemotePreview` no longer routes an unexpected success/redirect status through `errorResult` (which UTF-8-decodes a possibly-binary body and serves it under a success status). New `isServeDirErrorStatus` gates it: the probe accepts only 206/416, propagates a real 400/403/404/500 verbatim, and THROWS on anything else (a 200/3xx a ranged read must never produce). Same tightening on the unranged empty re-read (must be 200) and the unsatisfiable-range re-dial (must be 416, else throw). `iframePreviewRoute.ts:245`+. New test: a probe forced to 200 → throws `/unexpected status 200/`.
- `F3` **fixed** — Replaced `totalFromContentRange` (parsed only `/total`) with `parseContentRange`, which parses the full `bytes <start>-<end>/<total>` and `streamByteRange` now asserts `start===pos && end===end` in addition to `total` and the validator. A broken upstream answering the right LENGTH from the wrong OFFSET (`bytes 0-127` for a `bytes=128-255` request) now throws `/wrong slice/` instead of emitting the wrong bytes silently. `iframePreviewRoute.ts` `parseContentRange` + `streamByteRange`. New test added.
- `F4` **fixed** — `README.md:218` now describes the fork explicitly: local bind streams via `previewFile`; remote bind (`KOLU_PADI_HOST`) dials `preview.read` in bounded 8 MiB chunks and reassembles, byte/header-identical (W3.2). Also swept the nearby stale padi comment: `packages/padi/src/surface.ts` `PadiRepoRootForTerminal` doc no longer says the server always "STREAMS the file itself off the local disk" — it now names both arms.
- `F5` **fixed** — Added the `rangeResponseHead(resolved, size, baseHeaders) → { status, headers }` row to `packages/serve-dir/README.md`'s API reference (200/206 header-shaping behavior), and documented the new `ETag` in the responses note.

commit: `72dbed6ebd2110e65b8b47fd36dad51d5ce67d9d`


### Round 2

**codex** — approved: `false`

CLAUDE's changes resolve F1-F5 as originally raised: the remote assembly now validates status, full Content-Range, and a per-read validator; the empty-file growth case fails loud; and the two stale documentation surfaces were updated. The one regression I see in this round is from the new ETag header interacting with the existing HTML decoration middleware: decorated HTML responses now preserve a strong validator for the pre-decoration file bytes, not the actual response body.

Findings:
- `F1` · major · resolved — The original cross-chunk snapshot gap is addressed. The remote stream now captures an ETag from the probe and requires the same validator on each chunk; it also rejects the empty-file case if the unranged re-read returns bytes. (packages/server/src/iframePreviewRoute.ts:254)
- `F2` · minor · resolved — Unexpected successful/redirect statuses are no longer routed through `errorResult`. The probe now accepts only 206/416 or real serve-dir error statuses, and invalid-range re-dial must return 416. (packages/server/src/iframePreviewRoute.ts:302)
- `F3` · minor · resolved — Chunk validation now parses and validates the full Content-Range start/end/total, so a correct-length but wrong-offset chunk is rejected. (packages/server/src/iframePreviewRoute.ts:245)
- `F4` · minor · resolved — The README Architecture section now describes the local `previewFile` path and the remote chunked `preview.read` path separately; the stale padi surface comment was also updated. (README.md:218)
- `F5` · nit · resolved — The serve-dir README API reference now documents `rangeResponseHead`, and the response notes mention the new ETag behavior. (packages/serve-dir/README.md:69)
- `F6` · minor · open — The new strong `ETag` is emitted on every streamed 200/206, including HTML preview responses that `mountArtifactSdk` later mutates by injecting the SDK script. That middleware preserves all headers and only deletes `content-length` (`packages/artifact-sdk/src/server/index.ts:57`), so decorated HTML is now served with an ETag for the original file bytes rather than the actual response representation. That violates strong ETag semantics and can mislead caches or future conditional-request handling. (packages/serve-dir/src/index.ts:331)

_codex on the rebuttal:_ F1 accepted: the ETag check closes the same-size replacement case I raised, and the empty-file growth guard addresses the line-273 concern. F2 accepted: unexpected 200/3xx statuses now throw instead of being UTF-8 decoded as success bodies. F3 accepted: the full Content-Range validation covers the wrong-slice case. F4 and F5 accepted: the docs now match the changed architecture and exported API. New F6 is not a re-scan finding; it is introduced by this round's ETag addition, because a known downstream HTML transform now preserves a strong validator after changing the body.


**claude** — Fixed F6: the HTML-decoration middleware now strips the stale strong validator (ETag) of the original file bytes before serving the rewritten body, with a regression test.

- `F1` **fixed** (prior round, codex resolved) — cross-chunk snapshot gap closed via serve-dir strong ETag validated per chunk + empty-file growth guard. No change this round.
- `F2` **fixed** (prior round, codex resolved) — probe/re-dial status handling fails loud. No change this round.
- `F3` **fixed** (prior round, codex resolved) — full Content-Range start/end/total validation per chunk. No change this round.
- `F4` **fixed** (prior round, codex resolved) — README architecture split (local `previewFile` vs remote chunked `preview.read`). No change this round.
- `F5` **fixed** (prior round, codex resolved) — serve-dir README documents `rangeResponseHead` + ETag. No change this round.
- `F6` **fixed** — Codex is right: the round-1 ETag I added to serve-dir's every 200/206 rode through `mountArtifactSdk`'s HTML-decoration middleware, which splices the SDK `<script>` into the body but preserved all headers, so decorated HTML shipped a strong validator for the PRE-decoration bytes — a strong-ETag semantics violation. Fixed at `packages/artifact-sdk/src/server/index.ts:57` by deleting `etag` (and, defensively, `content-length`) so the rewritten body carries NO stale validator rather than a wrong one — the honest move, since a strong validator must change when the bytes change.

  A subtlety that made the naive fix (codex's literal suggestion — delete from the new Headers copy) INSUFFICIENT: Hono's `c.res` setter re-`set()`s every prior-response header (except content-type) onto the replacement (`hono/dist/context.js` `set res`, lines 120-135). So a header stripped only from the new copy is copied straight back from the old response. The delete has to hit `res.headers` (the OLD response — the merge SOURCE), which is what I did. This is also why the pre-existing `content-length` delete was a silent no-op that only "worked" because Hono never set Content-Length on the string response in the first place; the new code strips the source so it's correct either way.

  Regression test added: `packages/artifact-sdk/src/server/mount.test.ts` mounts a real Hono app + the real middleware and asserts (1) a decorated HTML response drops the ETag while still injecting the script and preserving Cache-Control, and (2) a non-HTML (video) response keeps its ETag and body untouched. Both pass; had I shipped codex's literal suggestion this test would have RED-caught the Hono re-merge, confirming the deeper fix was necessary. serve-dir arm (the ETag on the internal padi/serve-dir reads used for remote chunk validation) is untouched, per codex's guidance. `just check` (typecheck + biome) green; `just fmt` applied.

commit: `15abad59b1b27b6399e4eb86d7098e7d2e9d8322`


### Round 3

**codex** — approved: `true`

All findings from the prior rounds are resolved in the current tree. I verified the latest F6 fix directly: the artifact-sdk HTML rewrite now removes `etag` and `content-length` from the old response headers before assigning `c.res`, which is necessary because Hono re-merges old headers into the replacement response. The new regression test covers both decorated HTML dropping the stale ETag and non-HTML preserving its ETag. I did not run tests in this read-only review pass.

Findings:
- `F1` · major · resolved — The original cross-chunk snapshot gap is addressed: chunks are validated against the probe ETag, and empty-file growth between probe and re-read is rejected. (packages/server/src/iframePreviewRoute.ts:254)
- `F2` · minor · resolved — Unexpected successful/redirect probe and re-dial statuses now fail loud instead of being routed through `errorResult`. (packages/server/src/iframePreviewRoute.ts:302)
- `F3` · minor · resolved — Chunk validation now parses and checks full `Content-Range` start/end/total, so wrong-offset chunks are rejected. (packages/server/src/iframePreviewRoute.ts:245)
- `F4` · minor · resolved — The Architecture section now describes the local `previewFile` path and remote chunked `preview.read` path separately, and the nearby padi comment was updated. (README.md:218)
- `F5` · nit · resolved — The serve-dir README now documents `rangeResponseHead` and the ETag response behavior. (packages/serve-dir/README.md:69)
- `F6` · minor · resolved — The stale strong ETag on SDK-decorated HTML responses is now stripped before response replacement, preventing an original-file validator from riding on a rewritten body. (packages/artifact-sdk/src/server/index.ts:64)

_codex on the rebuttal:_ F6 accepted. CLAUDE's point about Hono's `c.res` setter is correct: it copies prior response headers onto the replacement, so deleting only from the new `Headers` copy would not be sufficient. Deleting `etag` and `content-length` from `res.headers` before assigning the replacement response fixes the stale-validator bug, and the new test covers the Hono merge behavior while confirming non-HTML ETags remain untouched. F1-F5 were already accepted as resolved in the prior round and were not reopened.

srid:

Ran `--no-elegance` (Pass 3 skipped — `/simplify` already ran this gauntlet). Pass 1 (rules) and Pass 2 (fact-check) over the diff vs the merge-base.

| Pass | Finding | Action |
|---|---|---|
| Fact-check | **`packages/server/src/index.ts`** — the remote byte-read dispatch ran *outside* the repoRoot try/catch, so a link fault during `assembleRemotePreview`'s metadata dials (the 1-byte probe or a re-dial) escaped as an unlogged generic **500** — contradicting the adjacent comment's promise that link faults map to a logged **503**. | **Fixed** ([`df5d9102f`](https://github.com/juspay/kolu/commit/df5d9102f)) — wrapped the remote `await` in a try that `log.error`s and returns 503 (the streaming body's per-chunk dials run later and correctly error the stream); added a unit test that `assembleRemotePreview` propagates a probe-phase rejection. |
| Rules (`dry-rule-of-three`) | **`packages/server/src/remotePadiSsh.test.ts`** — 3 inline case-insensitive header lookups duplicate the new `getHeaderCI` helper. | **Deferred** — test-internal; held to a pragmatic bar per srid's standing test-exemption (the tests prove their claims red→green). Not a production issue. |

Pass 1's full rule table came back clean aside from the one `dry-rule-of-three`; every "No" carried a `Checked by:` grep/enumeration. Pass 2 otherwise found the remote-preview / serve-dir / streaming code unusually rigorous — explicit throws on every mid-stream inconsistency, each backed by an adversarial test — with no silent-swallowing, fallback, or unbounded-growth issue.

srid:

The **x86_64-linux** lane ran on **`kolu-ci-6`** (`dev-x86-64-linux-03`) — commit `8463906ee`, **passed** ✓

- **Lane wall** (pipeline): **2m45s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 2m26s | ✓ |
| `ci::home-manager` | 2m7s | ✓ |
| `ci::nix` | 1m10s | ✓ |
| `ci::flake-check` | 1m10s | ✓ |
| `ci::unit` | 1m5s | ✓ |
| `ci::smoke` | 50s | ✓ |
| `ci::surface-example-build` | 13s | ✓ |
| `ci::biome` | 12s | ✓ |
| `_ci-setup` | 11s | ✓ |
| `ci::fmt` | 11s | ✓ |
| `ci::install` | 8s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/8463906/timings.jsonl`; pool state is a live `flock` probe.</sub>

srid:
## Evidence

This change is **backend by cause, visible by effect** — clicking a file in the Code tab while bound to a remote machine used to render a **501** (*"file preview is unavailable while bound to a remote padi"*); now it renders the actual file. Because the remote arm reassembles the host's bytes into a **byte- and header-identical** `ServeResult`, the rendered preview looks exactly like a local one — so the load-bearing proof is that the *right host's bytes* arrive intact over a real hop, which is what the reproducible ssh evidence-lane demonstrates.

### Reproducible proof — real bytes over a real ssh hop

Ran `remotePadiSsh.test.ts` on a fresh **pu box** (self-ssh, `KOLU_E2E_SSH_DESTRUCTIVE_ACK=1 just e2e-ssh`) — a real `padi --stdio` provisioned over ssh, `preview.read` dialed across the leg for three shapes:

```
[ssh] preview text OK over the hop — 24 bytes, ct=text/plain; charset=utf-8
[ssh] preview binary OK over the hop — 512 bytes byte-exact, ct=image/png
[ssh] preview 206 range OK over the hop — bytes 100-163/512, 64 bytes byte-exact

 ✓ src/remotePadiSsh.test.ts (5 tests) 40261ms
      Tests  5 passed (5)
```

- **text** — a UTF-8 file round-trips byte-exact with `Content-Type: text/plain`.
- **binary** — a 512-byte blob spanning every byte value (0–255) round-trips **byte-exact** with `Content-Type: image/png` — proving no base64/wire corruption of high bytes.
- **range** — a bounded `bytes=100-163` request answers **206** with `Content-Range: bytes 100-163/512` and the exact 64-byte slice — so a `<video>` seek works over the wire.

### Unit + CI

- **57 unit tests** (`iframePreviewRoute.test.ts`) drive `assembleRemotePreview` against the **real serve-dir engine** via a base64 fake reader: chunk boundaries, partial last chunk, zero-length (empty 200 + ranged 416), 206/416/suffix ranges, Content-Type passthrough, 404 passthrough, and the fail-loud guards (wrong status / size-change / wrong-slice / **ETag mid-stream replace** / short chunk / missing-ETag-probe — the last mutation-verified: neutralizing the guard fails the test).
- **CI green on both platforms** — 28 `x86_64-linux` + 28 `aarch64-darwin` contexts pass on `8463906`, including `ci::e2e` (which runs the Code-tab preview scenarios) and `ci::unit` on both.

*(Per the task's framing — reproducible tests are the proof, a screenshot would be a supplement — and a remote preview renders byte-identically to a local one by construction, the byte-exact real-hop transcript above is the primary artifact. The full remote-bound browser capture is the domain of W3.4's remote-e2e-parity lane.)*

srid:

The **x86_64-linux** lane ran on **`kolu-ci-6`** (`dev-x86-64-linux-03`) — commit `172339e18`, _verdict unknown (no per-node timing)_

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

PR #1686, opened 2026-07-05
feat(padi): "Running daemons" shows the connected machine (W3.2 inventory)

srid:
## The problem (user first)

**Plan of record:** the padi Atlas note, §W3.2 — the PR "inventory" bullet ([kolu.dev/atlas/padi.html](https://kolu.dev/atlas/padi.html)). The note is the coordinator's to finalize (gap (2) flips to closed on merge).

The Kaval and Padi rail dialogs each list the **running daemons** on your host — how srid spots a leaked/stray `padi`+`kaval` before it quietly wastes memory. Since #1675 you can point kolu at **another machine over ssh** (`KOLU_PADI_HOST`) and the whole canvas becomes that box. But the "Running daemons" list stayed stuck scanning the machine **kolu-server** runs on — it says so honestly in the label, yet it still **cannot show a leaked daemon on the machine you're actually using**. This is one of W3.2's two self-sufficient gap PRs; it closes the second of the two known remote-mode gaps.

## The shape

The scan in `packages/server/src/daemonInventory.ts` physically can only see kolu-server's own machine — correct, don't fight it. The remote half has to be **padi reporting the daemons on its OWN host**, riding back over the surface kolu-server already re-serves.

```
                 padiSurface.hostInventory  (NEW cell, 1.1 -> 1.2, value-forwarded)
  padi (bound host) ───────────────────────────── re-served ──────────────▶ client
     scans its own host, marks the kaval it holds + itself "active"          (bound-host list,
     (withSelfPadi guarantees the serving padi is in its own scan)           identical local & remote)

  kolu-server ── koluSurface.daemonInventory ──▶ client
     binding: {kind:"local"} | {kind:"remote"; host; localScan}              (local-machine list,
     + boundPadi (identity + convergence)                                     labelled "not the bound host")
```

- **One scanner.** The discovery + probe + assembly is a single implementation, homed in **`@kolu/padi`** (padi owns the daemon domain). padi serves it as the `hostInventory` member; kolu-server imports the same `enumerateHostDaemons` via `@kolu/padi/assembly` for its local-machine arm. The row + container schemas (`RunningKaval`/`RunningPadi`/`HostDaemonInventorySchema`) live in **`@kolu/padi/surface`** — the daemon domain's own browser-safe vocabulary — and `kolu-common/surface`'s `daemonInventory` cell imports them from there, the established `kolu-common → @kolu/padi` direction (the seal forbids only the reverse — padi importing kolu).

  > **Placement correction (post-review):** the first cut put these row schemas on `@kolu/terminal-workspace/schema` on a *misread of the seal* — I'd reasoned "the seal forbids either surface package importing the other", but it only forbids **padi** importing kolu packages; kolu-common/client/server already import terminal/daemon vocabulary from `@kolu/padi/surface` directly (since W1.M). `@kolu/terminal-workspace` is **terminal-awareness** vocabulary (sensors · fold · agentProjection); a kaval gate pid is a **daemon-domain** fact, so it belongs to padi, not there (location is structure). Reverted the terminal-workspace change entirely; no import-direction wall (`@kolu/padi/surface` imports nothing from kolu — acyclic).
- **Merge/replace, as a discriminated `binding` union.** In **local** mode the bound padi's `hostInventory` member already describes this machine, so kolu-server publishes `{ kind: "local" }` — no `localScan` (one list, no two copies of one truth). In **remote** mode it publishes `{ kind: "remote"; host; localScan }` — kolu-server's own box, shown as a **second labeled group** ("this machine, not the bound host"), marking nothing active. The union makes the contradictory pairings (local-with-scan, remote-missing-host) unrepresentable.
- **`active` is padi's direct knowledge**, not kolu-server's old digest-guessing (`resolveActiveKavalSocket`, deleted): padi marks the kaval it actually holds + itself. On `RunningKaval` that lands as a discriminated `held` — `{ active: false }` | `{ active: true; atLegacyAddress }` — so "legacy-but-not-owned" is a type error, not a defended convention.
- **Both dialogs** (Kaval + Padi) share one `RunningDaemonsSection` component.

## Honest degradation (#1034) — verified

A re-served `value` cell is seeded with its schema default and **held stale across a drop** (`reServeSurface`), so an empty *or* a stale-populated `hostInventory` must never render as a definite "Running daemons" scan. The dialog gates its bound-host list on a two-part liveness fact:

- **the bind is live** — the browser transport ∧ koluSurface's directly-served `padiLink` (kolu deliberately ignores the re-served surface's own held-stale health; `padiLink` is the honest bind signal, per `useDaemonStatus`). A dropped ssh link / the Restart-kaval drain window flips this false, so a **stale held list reads "unavailable"**, never a dead padi's list shown as current.
- **a real frame arrived** — a live padi ALWAYS reports itself (`withSelfPadi` seeds the serving padi's active row even on the T+0 tick and under a `--socket` override), so "no active padi row" means the seeded default, not a genuine zero. A **pre-first-frame** connected bind reads "unavailable", never "No running daemons".

The **mixed-version window** (a new 1.2 kolu-server pointed at a still-running 1.1 padi that doesn't serve `hostInventory`) is handled by construction: `isContractVersionCompatible("1.1","1.2")` is false, so a 1.2 binder **contract-refuses and drains** the 1.1 padi (it never re-serves a live surface missing the member — no crash), and until it converges the bind is not live → the list reads "unavailable". Traced through `dial.ts` / `padiBinding.ts` / `remotePadiBinding.ts` / `reServeSurface.ts`.

## Design philosophy

- **Fail fast, nothing swallowed.** A probe that can't reach a daemon folds to an honest per-row "—", never a missing row; a throwing probe seam can't sink the scan; both samplers log a discovery surprise (`.catch(log.error)`). No fallback/degradation path was added.
- **Electricity / seam.** The new capability lives in `@kolu/padi` (the daemon-domain authority) and rides the existing re-serve; kolu-server consumes it via padi's public entry. **Zero `@kolu/surface*` framework edits** (no drishti gate), and `remotePadiBinding.ts` is untouched.
- **Reuse the source of truth.** One scanner, one row + container schema on the shared leaf — no lockstep copy.

## Tests

| Test | Scope it pins |
| --- | --- |
| `packages/padi/src/hostInventory.test.ts` | the scanner: active-marking for padi's own scan **and** a no-active local scan; the `held` legacy-adoption arm; honest-null (incl. a **throwing** probe) folding; `withSelfPadi` seeding the serving padi (T+0 / `--socket`) with its pid as a value |
| `packages/padi/src/surface.test.ts` | contract **1.2**, the `hostInventory` member present, `value` forwarding policy |
| `packages/server/src/daemonInventory.test.ts` | the combiner: **local → `{ kind: "local" }`** (scan skipped, no scan carried); **remote → localScan populated, none active**; `boundPadi` identity/convergence over ssh; sampler coalesce |
| `packages/client/src/ui/hostInventoryLive.test.ts` | the honest-degradation predicate: live+frame → live; live+empty → unavailable; **stale-populated after a dropped bind → unavailable**; dead+empty → unavailable |
| `packages/tests/features/kaval-daemon.feature` (local-mode scenario) | the dialog shows **one group, no duplicate** local-machine group under a local binding |
| `packages/server/src/remotePadiSsh.test.ts` | the member **answers over a real ssh hop** — the serving padi is present and marks its held kaval active; run green on a pu box (`just e2e-ssh`) |

`just check` green; unit suites green for padi/server/common/client/terminal-workspace.

## Dialog behaviour

- **Local binding:** a single "Running kaval/padi daemons" list — the bound host *is* this machine. No second group.
- **Remote binding:** two labeled groups — "Kaval/Padi daemons on `<host>`" (the bound host, from the member) and a fenced "Local daemons — this machine, not the bound host" (kolu-server's own box). A leak on either machine is visible, each attributed to the right host. When the bind isn't live (link dropped / drain / connecting), the bound-host list reads an honest "Daemon scan unavailable".

## Review

Ran the full `/be-review` gauntlet plus srid's two standing lenses over the production delta: **lens-debate** (container dedup on the leaf, dead re-export, `boundPadi` consolidation, shared section), **codex** (`xhigh`; two liveness bugs it caught — the T+0 tick and the `--socket` override — fixed by `withSelfPadi`), **simplify** (probe parallelization), **code-police** (the `binding` discriminated union + a dead re-export), **architecture-first-principles** (P2 pid-to-the-edge, P4 `held` discriminated union), and **perfection-review** (the stale-after-drop liveness masquerade → the bind-liveness gate; the server sampler's missing `.catch`). Both discriminated unions verified at the ideal bar.


srid:
## [⚖️ Lowy ⇄ Hickey lens debate](https://kolu.dev/blog/hickey-lowy/) — ✅ consensus (1 round)

Two structural lenses (lowy = volatility boundaries, hickey = structural simplicity) reviewed the diff independently, then debated every finding to per-finding agreement. **4 findings, all `fix`, all applied.**

| # | Origin | Finding | Fix | Commit |
|---|--------|---------|-----|--------|
| lowy-1 | lowy | The inventory **container** `{kavals, padis}` was declared twice across the padi seal (`PadiHostInventorySchema` + the inline `z.object` in `daemonInventory.localScan`) — the lockstep copy the row-schema move eliminated, one altitude up; a new container field would be produced by the scanner but silently stripped at the koluSurface wire boundary. | Homed `HostDaemonInventorySchema` on the shared leaf `@kolu/terminal-workspace/schema`; both sides compose it (`PadiHostInventorySchema = it`; `localScan = it.nullable()`). One declaration, no cross-seal drift, scanner returns the neutral name. | `9535221a8` |
| lowy-2 | lowy | Dead re-export of `DEFAULT_PADI_HOST_INVENTORY` on the scanner's module boundary — no consumer reaches it through that path (servePadi imports it from `./surface.ts` directly). | Deleted the trailing re-export + its now-unused import; `surface.ts` stays the single home. | `4ab126aac` |
| hickey-1 | hickey | The active padi's `surfaceVersion`+`buildCommit` were dual-homed: on `RunningPadi` **and** on `daemonInventory.boundPadi`. padi cannot probe a foreign padi, so those fields were always `null` off the active row and duplicated the honest `boundPadi` read. | Deleted `surfaceVersion`/`buildCommit` from `RunningPadiSchema` + its assembler/deps/call sites; the Padi dialog's chip + build row already read `boundPadi`. The running-daemon row now shows only the gate pid. | `c44addbbc` |
| hickey-2 | hickey | The Kaval and Padi dialogs' "Running daemons" sections were near-identical fragments (bound-host group + local-scan group + honest-degradation gate), duplicated per dialog. | Extracted one `RunningDaemonsSection` component shared by both dialogs. | `b9caebe7a` |

All four are mechanical finishes of the PR's own stated dedup principle (one declaration per shape) or removals of dead/duplicated surface — no boundary the author defended was disturbed. `just check` + all touched unit suites green after apply.

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `20a346c38c3e`

### Round 1

**codex** — approved: `false`

The branch is directionally coherent, but I found two liveness bugs in the new host-inventory path. Both come from making "active padi row exists" the client's proof that `hostInventory` is live while the padi-side scanner can produce real frames without discovering that active row.

Findings:
- `F1` · minor · open — `startPadiHostInventorySampler()` runs before this padi has called `daemonMain(...)` and opened its own `padi.sock` at line 297. Its immediate tick therefore cannot discover/mark the serving padi active, and it also runs before the state-root manifests are written at lines 287-294. Because the client treats an active padi row as the liveness proof, a freshly connected healthy padi can render "Daemon scan unavailable" until the next 10s sample. The remote SSH test's polling loop hides this by waiting up to 25s, but users can see the false unavailable state right after startup. (packages/padi/src/daemonMain.ts:282)
- `F2` · major · open — The sampler assumes `discoverPadiDaemons()` will include `opts.padiSocket`, then passes that socket as the one to mark active. That is false for the documented `padi --socket PATH` / `padi --stdio --socket PATH` path: `discoverPadiDaemons()` only scans decorated `padi-<digest>` runtime dirs, so an override socket outside that naming pattern is never listed. The padi is live and serving `hostInventory`, but no row is active, so `hostInventoryReadingLive()` stays false forever and the bound-host daemon list is permanently hidden as "unavailable." (packages/padi/src/hostInventory.ts:269)


**claude** — Both findings are valid and hit the same load-bearing invariant ("a live padi ALWAYS reports itself"); I made the serving padi report itself BY CONSTRUCTION instead of via autodiscovery, which resolves the root cause of both, and reordered the sampler start after the manifest writes.

- `F1` **fixed** — Confirmed: `startPadiHostInventorySampler` at `daemonMain.ts:282` fires `tick()` synchronously, and `sampleOnce` runs synchronously through `enumerateHostDaemons`'s `deps.discoverPadis()` (up to the first `await Promise.all`) — so `discoverPadiDaemons()` runs BEFORE `daemonMain` opens `padi.sock` (line 297), and its `isSocketInode()` check can't see the serving padi yet, leaving the first sample non-live for up to 10s. Fixed two ways: (a) the sampler now seeds its OWN row via `withSelfPadi` (`hostInventory.ts`), so the padi reports itself regardless of whether the socket is listening — the liveness tell holds at T+0; and (b) moved the `startPadiHostInventorySampler(...)` call to AFTER the two `writeStateRootManifest(...)` calls in `daemonMain.ts` (now ~line 282→ after the kaval manifest write), so the very first tick also labels the discovered held-kaval by state-root. It still starts before the `return daemonMain(...)` (which is `forever` and never returns), as it must.
- `F2` **fixed** — Confirmed: `discoverPadiDaemons()` (`stateRoot.ts:198`) only scans `padi-<digest>` decorated dirs, so a `--socket PATH` / `--stdio --socket PATH` override socket outside that naming is never listed — the serving padi would be live yet have no `active` row, pinning `hostInventoryReadingLive()` false forever. Fixed by the new pure `withSelfPadi(discovered, self)` helper (`hostInventory.ts`): the serving padi knows its own socket/state-root/pid first-hand and seeds that row when autodiscovery misses it, deduped by socket so the normal digest-default case (where discovery DOES find it) is never doubled. This is codex's suggested "explicit current-socket candidate" and makes the "padi reports itself" invariant true by construction rather than dependent on discovery naming. Scoped to padi's OWN sampler only — kolu-server's local-machine scan (`daemonInventory.ts:81`, `activePadiSocket: null`) calls `enumerateHostDaemons` with bare `discoverPadiDaemons`, so it still lists only truly-discovered padis and marks none active (decision #3 preserved).

Tests: added a `withSelfPadi` describe block to `hostInventory.test.ts` proving the seed-when-missing (F1/F2 window), keep-sibling-and-seed-self, and dedupe-when-found cases; `assemblePadiInventory(withSelfPadi([], self), self.padiSocket)` marks the seeded row active. `just check` green (typecheck + biome). `hostInventory.test.ts` 15/15 green. NOTE: `packages/padi/src/ptyHost/spawnInput.test.ts` fails in THIS worktree — but it fails identically with all my changes stashed (clean base) and when run entirely alone (it never imports my files); cause is the ambient shell exporting `PADI_SOCKET=/run/user/1000/padi-…` (this worktree runs inside a live kolu terminal), which `composeSpawnInput` reads. Pre-existing environmental artifact, passes in CI's clean env — not caused by and not in scope of this change.

commit: `9ea20013a06e7f4a8c2733328f2fb39f38dd69e8`


### Round 2

**codex** — approved: `true`

Claude's changes address both previously raised liveness problems. The padi-side sampler now injects the serving padi row explicitly via `withSelfPadi`, so the client's "active padi row means live inventory" test no longer depends on autodiscovery timing or default digest-path naming. The sampler is also now started after the state-root manifests are written, so the first scan has the labels available.

Findings:
- `F1` · minor · resolved — Previously, the immediate host-inventory tick could run before padi opened its own socket and before manifests were written, producing a real-but-active-less frame that the client rendered as unavailable until the next sample. (packages/padi/src/daemonMain.ts:278)
- `F2` · major · resolved — Previously, `--socket` override padis outside the decorated `padi-<digest>` runtime dirs were never discovered, so the serving padi could be live forever without an active row and the UI stayed permanently unavailable. (packages/padi/src/hostInventory.ts:271)

_codex on the rebuttal:_ F1: accepted. The fix preserves the client invariant by construction rather than by timing, and moving the sampler below the manifest writes covers the label race. Starting before `daemonMain(...)` is acceptable now because the stored first frame already contains the active self row.

F2: accepted. The explicit self candidate covers documented `--socket` overrides without changing kolu-server's remote local-scan semantics, since only padi's own sampler wraps `discoverPadiDaemons()` with `withSelfPadi`.

srid:

Ran `--no-elegance` (simplify already passed over this tree). Pass 1 (rules) + Pass 2 (fact-check).

**Pass 2 (fact-check): clean** — verified the honest-null probe folding (`probeKavalStatus` catch → `EMPTY_PROBE`; `enumerateHostDaemons` per-probe `.catch` → honest per-field nulls, never a dropped row or a sunk scan), the sampler's outer `.catch` logs at `error` (not swallowed), no unhandled rejection from the `Promise.all` probe race, and the 1.1→1.2 version bump consistent across schema/tests/dial handshake.

**Pass 1 (rules): 3 findings** —

| Finding | Verdict | Action |
|---|---|---|
| `invalid-states-unrepresentable` — `boundHost` + `localScan` were two independently-nullable fields coupled only in prose (`localScan` populated iff `boundHost !== null`) | CONFIRMED | **Fixed** (`4d4f8f2c4`): replaced with a discriminated `binding` union — `{ kind: "local" }` (no scan) \| `{ kind: "remote"; host; localScan }`. The illegal pairings (local-with-scan, remote-without-host/scan) are now unrepresentable; the combiner runs the local scan only in the remote branch. |
| `no-dead-code` — the `RunningKavalSchema`/`RunningPadiSchema` value re-export from `kolu-common/surface` had zero consumers (every consumer imports only the type) | CONFIRMED | **Fixed** (`4d4f8f2c4`): removed the dead re-export. |
| `no-sync-blocking-on-the-serving-loop` — the 10s sampler runs sync-fs discovery (`readdirSync`/`lstatSync`/`readFileSync`) on padi's serving loop | PLAUSIBLE | **Skipped, with reason**: `discoverKavalDaemons`/`discoverPadiDaemons` are shared sync primitives used across the codebase (kaval-tui, padi-tui, and the pre-existing kolu-server sampler that already did this on its own loop); the scan is a handful of `/run/user/<uid>` tmpfs stats at a coarse 10s cadence. Making discovery async is a broad cross-cutting refactor well outside this PR's scope — noted as a follow-up, not a blocker. |

srid:
## 🧭 Architecture-first-principles + 🔬 perfection-review (srid's standing lenses)

Two adversarial passes over the **production delta**, on top of the `/be-review` gauntlet.

### Architecture-first-principles (state-and-time)
**Discharged** three obligations the pre-PR code left open: **P3** (one writer per host — `hostInventory` written only by padi's sampler, `daemonInventory` only by the combiner; no cross-host reduce, no wall-clock compare), **P4** (the `binding` discriminated union), **P5** (the "daemons on the bound host" guarantee relocated to the padi that runs on that host). **Two fix-now residuals, both fixed** (`437c4dbcb`):
- **P2** — `withSelfPadi` read the ambient `process.pid` mid-fold → threaded the pid through `self`, read `process.pid` once at the edge (the sampler).
- **P4** — `RunningKaval` let `{ active: false, atLegacyAddress: true }` be constructed → folded into a discriminated `held` (`{ active: false } | { active: true; atLegacyAddress }`); the two scattered `active &&` re-checks collapse to a `.held.active` narrow.

### Perfection-review (ideal bar, adversarial fan-out)
Both discriminated unions **verified at the ideal bar** (illegal pairings unspellable end-to-end — single constructor, discriminant-only reads, zod-strips-on-parse). **Two residuals, both fixed** (`ebbb4cda5`):
- **The masquerade relocated.** `hostInventoryReadingLive` inferred liveness *purely from payload content*, so a bind that dropped **after** a live reading showed the re-serve's stale-held list (still carrying the serving padi's active row) as a current live scan — the exact #1034 masquerade the predicate exists to prevent. Fixed by gating on the canonical bind-liveness **fact** (browser transport ∧ directly-served `padiLink`) **∧** the self-padi frame tell; the predicate is now a pure composite `hostInventoryLive({ bindLive, padis })`, unit-tested incl. the stale-after-drop case.
- **A swallowed surprise.** The server daemon-inventory sampler tick had no `.catch` and no logger (unlike its `@kolu/padi` twin), so a discovery/publish throw became an unlogged unhandled rejection that silently reverted the cell to its local default — dropping the local-scan leak diagnostic. Added `.catch(log.error)`; corrected the over-strong "Never throws" comment.

The deliberate framework-gated descent (a per-cell "received first frame" bit on the re-serve store, a `@kolu/surface-nix-host` change under surface.md's drishti rule) stays out of scope — the bind-liveness gate above is the smaller, canonical, in-repo fix and needs no framework edit.

srid:

The **x86_64-linux** lane ran on **`kolu-ci-7`** (`dev-x86-64-linux-04`) — commit `ebbb4cda5`, **passed** ✓

- **Lane wall** (pipeline): **4m28s**

| recipe | duration | |
|---|--:|:--|
| `ci::nix` | 4m3s | ✓ |
| `ci::home-manager` | 3m21s | ✓ |
| `ci::e2e` | 3m19s | ✓ |
| `ci::flake-check` | 1m2s | ✓ |
| `ci::smoke` | 59s | ✓ |
| `ci::unit` | 54s | ✓ |
| `ci::atlas-sync` | 54s | ✓ |
| `ci::install` | 46s | ✓ |
| `ci::pnpm-hash-fresh` | 43s | ✓ |
| `_ci-setup` | 23s | ✓ |
| `ci::biome` | 11s | ✓ |
| `ci::fmt` | 10s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/ebbb4cd/timings.jsonl`; pool state is a live `flock` probe.</sub>

srid:

The **x86_64-linux** lane ran on **`kolu-ci-7`** (`dev-x86-64-linux-04`) — commit `3e3fde60d`, **passed** ✓

- **Lane wall** (pipeline): **2m59s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 2m41s | ✓ |
| `ci::home-manager` | 2m25s | ✓ |
| `ci::unit` | 1m28s | ✓ |
| `ci::nix` | 53s | ✓ |
| `ci::smoke` | 47s | ✓ |
| `ci::surface-example-build` | 28s | ✓ |
| `ci::fmt` | 26s | ✓ |
| `ci::biome` | 26s | ✓ |
| `_ci-setup` | 10s | ✓ |
| `ci::install` | 7s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/3e3fde6/timings.jsonl`; pool state is a live `flock` probe.</sub>

PR #1687, opened 2026-07-05
fix(surface): hold per-key collection get open for a not-yet-existing key — the real gray-chip fix (#1681)

srid:
Two `@kolu/surface`-stack framework fixes for the **gray Kaval chip after a remote-padi rebind** ([#1681](https://github.com/juspay/kolu/issues/1681)). The headline fix is **not** what the issue originally hypothesized — I discarded that theory and **reproduced the bug on a pu box against a real browser** first, which pinned a different mechanism entirely.

## The real mechanism (reproduced, not theorized)

The Kaval chip watches one key — `"local"` — of the `daemonStatus` collection, and asks for it **by name, always** (`useDaemonStatus`: `keys: () => ["local"]`). When kolu-server **boots fresh** (a redeploy / a remote bind), its re-serve mirror starts **empty**, so `"local"` is absent for the first moment until padi connects. The browser, loading in that window, immediately opens `daemonStatus.get({ key: "local" })` — and the server's `get` handler **threw** `"key not found at first snapshot"`, which crossed the wire as a **non-retriable `INTERNAL_SERVER_ERROR` (`ORPCError`)**. `STREAM_RETRY` deliberately never retries an `ORPCError`, so the standing per-key subscription **died**. When `"local"` appeared a moment later, the dead subscription never saw it → the chip stayed gray **until a full page reload**.

The #1681 frame trace that "convicted the client via a dropped frame" was a **raw replica** (a Node `websocketLink` client), not the real browser — so its mechanism was a hypothesis. The real browser never gets a dropped frame; its subscription is dead before `"local"` ever exists.

### Box reproduction (recorder, one real browser tab open through it)

| build | Kaval dot across the boot window → after padi binds |
| --- | --- |
| **master** | gray the whole time; padi binds → **Kaval stays gray for 2+ min**, Padi goes green; reload → green |
| **this PR** | gray while waiting; padi binds → **Kaval flips to green with Padi, no reload** |

Server `curl` confirmed `daemonStatus["local"] = {state:"connected"}` was present the whole time master stayed gray — the server had the data; the client subscription was a corpse.

## Fix A — the convicted seam: a per-key `get` for a not-yet-existing key holds open

A collection's membership is dynamic by design (the mirror's `initialKeys` reconcile already treats it so), so a `get` for a key **born later** is a legitimate held-open subscription, not an error. The handler now subscribes-before-snapshot (like `keys`), yields the snapshot **only if the key is present**, then forwards — an absent key yields nothing and holds open until its first upsert publishes to the same channel. This kills the class for **every** per-key consumer.

Per the review steer, this is a **deliberate, documented, tested** semantic — never an accidental hang. Red→green tests: **(1)** subscribe to an absent key, the key appears later, the subscriber receives it (red on master: the sub throws and dies); **(2)** a key that never appears leaves the stream open yielding nothing and drops cleanly on abort (the honest "waiting" state, not a corpse).

## Fix B — the cell equals-gate drop (distinct, same re-serve seam)

`reServeSurface`'s cell fold writes through `ctx.cells.<key>.set`, whose `equals` dedup **swallowed a post-rebind value equal to the pre-drain one** — so kolu-server's own adopted-stale banner / connection recovery couldn't tell "rebound and confirmed" from "stale". The ctx setter gains an optional `{ force }`; the re-serve forces the **first fold of each spawn** (the rebind epoch), steady-state dedup unchanged. Red→green `reServeSurface` test: an equals-gated cell across a `setClient` rebind emits **`[v, v]`, not `[v]`** (and a further equal steady-state write still dedups).

## Gate

Framework change under `packages/surface*`, so per `.claude/rules/surface.md` this ships with a **paired [srid/drishti](https://github.com/srid/drishti) PR**, CI-green, pinned to this PR's **final** HEAD. Linked below.

## Evidence

The before/after recorder run + screenshots land as an `## Evidence` comment.

_Supersedes the parked `fix/surface-rebind-frame-drop` attempt (a "re-subscribe on clean stream completion" fix) — that targeted the wrong mechanism, which the box repro disproved._

---

## Review gauntlet + two-lens pass

Passed the full serial gauntlet (**lens-debate**, **codex-debate**, **/simplify**, **code-police**) plus the two-lens standing-order pass on the production delta (**/architecture-first-principles**, **/perfection-review**). Notable in-branch fixes it drove:
- **codex** caught a real bug: the held-open `get` would have hung the **MCP one-shot `resources/read`** for an absent key — fixed with the bounded `firstFrameOfCollectionItem` race.
- **architecture-first-principles (P5)** — lifted that bounded reader out of `surface-mcp` into `@kolu/surface` (beside the `firstFrameOrThrow` footgun it guards) so the safe reader lives where the hazard is minted; `surface-mcp` is now a thin adapter.
- **code-police** — un-conflated "unresolved URI" from "key-not-present-yet" in the MCP read (typed `ReadMiss`); exhaustive `switch` on resource kind.
- **perfection-review** verified the product is fixed on every axis it probed (the `{force}` republish is bounded/idempotent/single-writer; the `S[]` snapshot generalization is byte-identical for `keys`/`deltas`; the absent-key path loses no frame and leaks no subscription; the MCP typed sums are sound).

### Deliberately deferred (ideal-bar residual)

**Making the held-open-`get` hang *unspellable*, not just avoidable** → tracked in **#1688**. The fix ships the safe reader (`firstFrameOfCollectionItem`) and steers every reader to it via the `firstFrame.ts` docstring, but a naive `firstFrameOrThrow(coll.get({key}))` is still *spellable* and now hangs on an absent key (was a throw). No current consumer does this (verified — surface-mcp is fixed, padi-tui reads `keys`, the `kavalChannels`/`padi` per-key gets are held-open subscriptions that this change *improves*). Making it truly unspellable needs either a breaking discriminated-frame wire change or deep type-branding through oRPC-derived types + drishti — disproportionate to a future-misuse with no current victim, so it's deferred to #1688 with that reasoning.


srid:
Paired drishti PR (surface.md gate): srid/drishti#87 — re-pins kolu to this branch; validates that drishti consumes the (backward-compatible) surface changes unchanged. CI re-confirmed green against this PR's **final** HEAD before merge.

srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `20a346c38c3e`

### Applied (2)
- `lowy-3` The { force } opts shape is hand-duplicated across the surface↔surface-nix-host package boundary via an `as` cast, not the exported ctx contract — commit `12c7afa71`
- `hickey-1` get handler duplicates the subscribe-before-snapshot machine — commit `bf222454e`

### Agreed — no change (3)
- `lowy-1` Electricity test — @kolu/surface's reactive contract is the receptacle; this change re-shapes two sockets behind it (packages/surface/src/server.ts:390)
- `lowy-2` Value channel is made to carry a liveness/epoch signal that already has its own receptacle (packages/surface-nix-host/src/reServeSurface.ts:426)
- `hickey-2` epochForced latch name inverts its meaning (packages/surface-nix-host/src/reServeSurface.ts:426-434)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 round(s) · codex reviewed at `xhigh` reasoning effort · base `20a346c38c3e`

### Round 1

**codex** — approved: `false`

The core held-open subscription path is documented and covered, and `git diff --check` is clean. I found one integration break: an existing one-shot MCP resource reader still treats collection-item `get` as "yield or fail," but this branch changes absent keys to wait indefinitely.

Findings:
- `F1` · major · open — `resources/read` still calls `firstFrameOrThrow` on collection-item `get` without passing any request signal. With the new collection `get` semantics in `packages/surface/src/server.ts`, an exposed collection-item URI whose key is currently absent no longer throws; it yields nothing and holds open. That is correct for live subscriptions, but here it turns a one-shot MCP read into an indefinite hang, and the handler at line 359 also ignores the MCP request cancellation signal, so the held-open stream is not tied to request lifetime. (packages/surface-mcp/src/server.ts:597)


**claude** — Conceded codex's one integration finding and fixed it: the MCP one-shot `resources/read` no longer leans on the now-held-open collection `get` to detect an absent key (which would hang forever), and it threads the request signal.

- `F1` **fixed** — Codex is right. My `packages/surface/src/server.ts` change made the collection `get` HOLD OPEN for an absent key (yield nothing until born) — correct for a live subscription, but `packages/surface-mcp/src/server.ts`'s one-shot `readSnapshot` still called `firstFrameOrThrow` on that `get`, so a `resources/read` of a not-yet-born collection-item key would block indefinitely, and the `ReadResourceRequestSchema` handler ignored `extra.signal` so the held-open stream wasn't tied to request lifetime. Fix: (1) `ReadResourceRequestSchema` handler now takes `extra` and threads `extra.signal` through `withClient`→`readSnapshot`→every client call (server.ts:359, 361-363). (2) `readSnapshot` gained a `signal` param and, for `kind === "collection-item"`, resolves membership FIRST via a new `collectionItemExists` helper that reads the collection's `keys` snapshot (which always arrives immediately by the snapshot-first contract) and tests membership of the decoded key — an absent key returns `undefined` (honest not-found via the handler's "unknown resource" throw), a present key reads its `get` first frame which arrives at once (server.ts readSnapshot + `collectionItemExists`). The live-subscription held-open semantic (streamForUri/pusher) is untouched, so the #1681 fix's intent is preserved — only the one-shot read gets a bounded existence boundary. Added a regression test "reading an ABSENT collection item returns not-found promptly, never hangs" (server.test.ts, in the edge-surface describe) that races the read against a 2s timeout and asserts prompt rejection; it would time out under the pre-fix hang. `just check` green; full surface-mcp suite 53/53 green.

commit: `4881acb13c04a845b10e6c7c23e5e42d5b7e631a`


### Round 2

**codex** — approved: `false`

The direct absent-key hang is fixed and the MCP request signal is now threaded through, but F1 is not fully closed: the one-shot read can still await the held-open collection item stream forever if membership changes between the new keys snapshot and the later item get.

Findings:
- `F1` · major · open — The fix checks collection-item membership with `keys` first, then opens the item `get` at line 605. Because collection membership is dynamic, the key can be removed after the keys snapshot but before the item get snapshots. In that race, the new held-open `get` sees an absent key, yields nothing, and the normal `resources/read` request still hangs unless the client actively cancels it. Passing `extra.signal` is necessary cleanup, but it does not bound a non-cancelled one-shot read. (packages/surface-mcp/src/server.ts:599)

_codex on the rebuttal:_ F1: partial fix accepted, but I'm holding the finding open. The common absent-key case now returns not-found promptly, and request cancellation is correctly threaded. The remaining issue is the delete race introduced by the two-step `keys`-then-`get` sequence: if the key disappears between those steps, `get` has the same held-open absent-key behavior that caused the original hang.


**claude** — Closed the F1 delete-race by replacing the two-step check-THEN-`get` with a single bounded read that RACES the item `get` first frame against a live `keys`-absence watch, so no window exists where a delete can slip between two sequential steps.

- `F1` **fixed** — Codex is right that a two-step `keys`-snapshot-then-`get` sequence leaves a delete race: a key removed between the membership snapshot and the `get` subscribe would hit the held-open `get` and hang. I removed the sequential step entirely. `readSnapshot`'s collection-item branch now calls a new `readCollectionItemSnapshot` (`packages/surface-mcp/src/server.ts:648`) that opens BOTH the item `get` and the collection's `keys` subscription concurrently under a local `AbortController` (chained to the request `signal`) and `Promise.race`s them:
  - `keys` is snapshot-first AND self-healing — every frame is the FULL membership set, re-emitted on every upsert/remove (see `subscribeBeforeSnapshot` + `keysBus` in `packages/surface/src/server.ts:360`). The `absent` branch iterates `keys` for the WHOLE duration of the wait, not once: the key being absent at the snapshot OR removed at any later instant produces a frame that omits it → resolves not-found (`undefined`).
  - a PRESENT key yields its `get` snapshot immediately and wins the race with its value.
  - whichever settles first aborts the other, so the loser's subscription is always torn down; the loser's later rejection is already handled by `Promise.race`, so no unhandled rejection.
  Because `keys` is watched continuously across the entire `get` wait (not sampled once before it), the exact interleaving codex named — delete lands after the keys snapshot but before `get` — now produces a `keys` removal frame that resolves the read not-found. There is no longer a two-step window to race. I also extracted the snapshot-first tail into `readFirstFrameSnapshot` (server.ts:604) so the present-key path and the non-collection kinds share one loud-fail-on-empty implementation, and deleted the now-subsumed `collectionItemExists` helper.

  Test: added "a collection item DELETED before the read returns not-found promptly (delete race)" (`server.test.ts`) — reads key 42 (present → value), deletes it via the client, re-reads, and races the read against a 2s timeout asserting prompt rejection. Note (honest scope): this exercises the bounded post-delete path deterministically; the EXACT sub-millisecond interleaving (delete strictly between the old keys-snapshot and the old get-subscribe) is not reproducible through the black-box MCP transport, so the guarantee rests on the structural change (continuous keys watch, no sequential window) plus this bounded-read regression. The pre-existing "reading an ABSENT collection item returns not-found promptly, never hangs" still passes. `just check` green (surface-mcp typecheck + biome clean, exit 0); full surface-mcp suite 19/19 green; `just fmt` clean.

commit: `ef0387e012a671b75ffa2dcbe920f87fcfdac691`


### Round 3

**codex** — approved: `true`

Verified the current tree against the existing F1 only. The one-shot MCP collection-item read no longer relies on a sequential keys check followed by a held-open get; it now races the item snapshot against a live keys absence watch and ties both subscriptions to abort cleanup. I did not rerun the suite locally in this read-only pass.

Findings:
- `F1` · major · resolved — Previously, MCP resources/read could hang on collection-item reads because absent collection get now holds open, and the first fix still left a delete race between the keys snapshot and the item get subscription. (packages/surface-mcp/src/server.ts:667)

_codex on the rebuttal:_ F1: accepted. The new implementation removes the sequential membership-check window I objected to. A held-open item get is now bounded by the concurrent keys absence stream plus the request/local abort signal, so both absent-key reads and delete-after-snapshot races have a prompt not-found path.

srid:

Pass 1 (rules) surfaced 2 findings; Pass 2 (fact-check) came back clean; Pass 3 (elegance) skipped (`--no-elegance` — `/simplify` already ran this gauntlet). Both findings **actioned in-branch** (`fc6cba4b`):

1. **`invalid-states-unrepresentable` / `caught-error-must-not-collapse-to-empty`** (confirmed) — after #1681's held-open change, `surface-mcp`'s one-shot `resources/read` returned a bare `undefined` for **both** an unresolvable URI **and** a well-formed collection-item whose key isn't present yet, and both threw the identical `"unknown resource"`. **Fixed:** a typed `ReadMiss = { miss: "unresolved" | "not-present" }` threads the cause out, and the handler gives the not-yet-present case its own message (`"…has no value yet — its collection key is not present"`). The two `absent`/delete-race tests now assert that distinct message, not just rejection.
2. **`prefer-ts-pattern` / exhaustiveness** (minor) — `readSnapshot`'s dispatch on the 5-member `call.kind` union used guard-clauses with a shared fallback, so a future 6th kind would fall through silently. **Fixed:** a `switch` with a `never`-typed `default`, so a new kind is a compile error (no new dependency — ts-pattern isn't a `surface-mcp` dep, and adding it would churn the lockfile for a minor finding).

srid:

The **x86_64-linux** lane ran on **`kolu-ci-8`** (`dev-x86-64-linux-05`) — commit `d8d26e676`, **passed** ✓

- **Lane wall** (pipeline): **3m16s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 2m57s | ✓ |
| `ci::home-manager` | 2m32s | ✓ |
| `ci::unit` | 1m16s | ✓ |
| `ci::smoke` | 1m3s | ✓ |
| `ci::nix` | 1m2s | ✓ |
| `ci::flake-check` | 57s | ✓ |
| `ci::biome` | 16s | ✓ |
| `ci::fmt` | 15s | ✓ |
| `_ci-setup` | 10s | ✓ |
| `ci::install` | 8s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/d8d26e6/timings.jsonl`; pool state is a live `flock` probe.</sub>

srid:
@/tmp/claude-1000/-home-srid-code-kolu--worktrees-W3-1-surface/d0e9c6a4-2a5a-46d9-a868-4e4b022e69fd/scratchpad/ev-body.md

srid:
**Ship gates green — ready to merge.**
- kolu CI: green on **both platforms** (`x86_64-linux` @ kolu-ci-8 + `aarch64-darwin` @ rasam) at final HEAD `d8d26e676`.
- surface.md drishti gate: **srid/drishti#87** re-pinned to `d8d26e676` and CI-green (10 ok · 0 failed) at that final HEAD.
- Review gauntlet + two-lens pass complete; one ideal-bar residual deferred to #1688 with reasoning (in the PR body).
- Evidence posted (box before/after).

srid:

The **x86_64-linux** lane ran on **`kolu-ci-8`** (`dev-x86-64-linux-05`) — commit `e01cc8ea8`, **passed** ✓

- **Lane wall** (pipeline): **4m31s**

| recipe | duration | |
|---|--:|:--|
| `ci::home-manager` | 4m24s | ✓ |
| `ci::e2e` | 4m3s | ✓ |
| `ci::unit` | 1m40s | ✓ |
| `ci::nix` | 1m19s | ✓ |
| `ci::flake-check` | 1m16s | ✓ |
| `ci::smoke` | 1m7s | ✓ |
| `ci::surface-example-build` | 35s | ✓ |
| `ci::biome` | 34s | ✓ |
| `ci::fmt` | 33s | ✓ |
| `ci::install` | 14s | ✓ |
| `_ci-setup` | 7s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/e01cc8e/timings.jsonl`; pool state is a live `flock` probe.</sub>

PR #1689, opened 2026-07-05
feat(e2e): in-terminal mock-agent for geography-free agent-state scenarios

srid:
## What

Replaces the e2e suite's agent-state mocks — the old `agent-mock-codex` /
`agent-mock-opencode` / `mock-fs` WAL-and-filesystem fakes — with a single
**in-terminal mock-agent**: a real binary the scenario runs *inside a kolu
terminal*, which writes the agent's session/transcript artifacts at their
**real paths**, exactly the way the real Claude Code / Codex / opencode / Grok
do.

## Why

The old mocks reached around the terminal and hand-wrote agent state into
fixture paths. That coupled every agent-state scenario to the mock's private
notion of where state lives, and it assumed that state was on the same machine
as the test. Driving a real artifact-producing agent *through the PTY* removes
both couplings: the scenario just runs a command, and the padi observes the
artifacts the same way it observes a real agent — geography-free.

## Scope

- New `packages/mock-agent` — the in-terminal agent (claude / codex / opencode /
  **grok** protocols; geography-free path resolution + tests).
- The claude / codex / opencode / grok step definitions now drive it over the PTY;
  the old `agent-mock-*` and `mock-fs` support modules are deleted.
- `nudge.ts` narrows to the file/dir nudges that remain (the WAL nudge moved
  into the in-terminal agent).
- Correctness fixes: validate state names, codex/opencode `awaiting_user`
  artifacts, uuidv7 codex thread id, unique `seq=` acks, Grok process-group pid
  for `active_sessions` matching.

*(Originally built alongside W3.4's remote-e2e lane; that lane is not part of
this PR — this is the mock-agent portion, standalone and useful on its own.)*

Issue #1680, opened 2026-07-04
Codex WAL fs.watch singleton dies on darwin-remote padi (kqueue re-arm), never reinstalls — agent-state freezes


srid:
### Symptom

Under a **remote padi binding to a darwin host** (`sincereintent`), the Padi/Dock agent-state stops updating after the padi has run a while. The process-wide **codex WAL `fs.watch` singleton** in a ~7.5h-old padi delivers **nothing** — new agent-state subscribers join a dead singleton and it never re-arms.

### Fact chain (from srid's live run; padi commit `bcf33a9`, build-identical to this branch's HEAD)

Everything *around* the singleton is healthy — only the singleton's OS-level watch is dead:

- **`fs.watch` works** in the same runtime (probed directly).
- **The WAL emits events on real writes** (probed).
- **The full integration works standalone** from the daemon's own store path (probed — a complete transition stream).
- **`lsof` shows the daemon holds NO watch fd on the WAL**, while a control process's live watch shows as an `r`-fd on the same file. So the OS-level watch is **gone**, but the JS refcount/installed state still believes it's installed → new subscribers join the dead singleton and **never reinstall**.

The **death mechanism is unknown because the daemon's logs were discarded** to `/dev/null` — the detached-daemon logging fix (P0, this PR) is what makes the next occurrence diagnosable (`~/.local/state/padi/padi.log.*`).

### Axis: darwin / kqueue

- **No breakage observed on local kolu, nor on `zest` (linux)** — only **darwin-remote** so far. That points at **kqueue** (macOS's `fs.watch` backend) rather than inotify.

### Candidate paths (in `packages/shared/src/sqlite/wal-subscription.ts`)

- **No `'error'` listener on either `FSWatcher`** — a watcher that errors out (kqueue fd death) would silently stop delivering with the JS state none the wiser.
- The **documented unrecoverable dir-watch-failed** case.
- **kqueue fd death on inode replacement with a failed re-arm** — SQLite WAL checkpoint/rotation replaces the inode; kqueue watches an fd/inode, so the watch dies on replacement, and if re-arm fails (or isn't attempted) the singleton is permanently dead.

### Ask

Repro-first (reproduce the dead-singleton-no-reinstall on darwin/kqueue before theorizing a fix), its **own PR**, darwin/kqueue named as the axis. Not fixed in the W3.1 remote-binding PR.

PR #1692, opened 2026-07-05
fix(git): pin parcel's watcher backend — stop the per-subscribe zombie leak

### LINKED ISSUE #1691: Remote padi (--stdio) leaks zombie children — killed PTY/endpoint procs never reaped; degrades a long remote run
## Summary

Under a **remote padi binding** (`KOLU_PADI_HOST`), the remote `padi --stdio`
daemon **leaks zombie child processes** — killed children are never `wait()`ed.
Over a sustained run they accumulate without bound; the same event that drops the
reap also drags the daemon's procedure latency, so terminal/Code-tab-heavy work
degrades in the second half of a long run. This surfaced while standing up W3.4's
full-suite remote-e2e CI node (#1689).

This is the **second** long-lived-daemon defect the remote path has surfaced
(#1680 — the darwin WAL-watcher death — was the first). Both are *sustained-life*
bugs a short unit test can't see; see "Coverage gap" below.

## Measured facts (box A → warm pool box B, both 32c/128GB)

- Running the existing e2e suite bound to a remote padi, box B's process count
  climbed **37 → 123** over the run and **stayed there after the run went idle**.
- Those extra processes are **62 defunct (`Z`) `sh`**, **52 of them parented by
  the remote padi** — `PID 180847 = node … packages/padi/src/bin.ts`.
- **They persist indefinitely at idle**: hours later, with no active run, the
  padi process is still alive and still holds 62 zombies / 52 defunct `sh`. An
  event loop merely *starved* would reap late; **never**-reaping means a code
  path that kills a child without attaching/keeping its exit handler — a real
  **missing-reap path** (a detached/kill path that drops the reaper).
- **Not resource pressure**: during the degraded run box B had 105 GB free, load
  0.6, only 2 inotify watches; box A had 111 GB free. Zombies are free, which is
  exactly why load/memory look healthy while the daemon degrades.
- **Not a tight-wait / transport issue**: the ssh transport lane is green
  (typing-echo p99 **7.65 ms** over the hop, ~local 4.36 ms), and the *same*
  code-tab scenarios that time out in a long run **pass over the bind in
  isolation**.
- **Remote-binding-specific**: the local full suite does not degrade (current CI
  is green), so padi reaps correctly in the in-process/local path but not through
  the `--stdio`/remote spawn path.

## Suspected mechanism (to confirm in the fix)

A padi child-spawn path whose kill/teardown drops the process's exit handler, so
`SIGCHLD` for that child is never consumed. The **code-tab-heavy** timeout
concentration hints at padi's *endpoint* children (git/fs subprocesses the Code
tab spawns for diff/status/tree) rather than the PTY login shells — but the child
class (kaval/ptyHost PTY shells vs padi endpoint git/fs children) must be pinned
before touching code. `sh` is consistent with a `child_process` `exec` /
`sh -c 'git …'` spawn.

## Coverage gap (why this keeps happening)

Both this and #1680 are **long-lived-daemon** defects invisible to unit tests
that spin the daemon up and tear it right down. The fix should land our first
**daemon soak test**: drive sustained create/kill cycles through the surface and
assert **both** (a) child count returns to baseline (no zombie accumulation) and
(b) a bounded procedure p99 under the sustained load.

## Owner

The remote-e2e CI node (#1689, W3.4) is **blocked** on this — an every-PR gating
lane cannot be sometimes-red. Fix rides a separate daemon-lifecycle PR off
master; #1689's node is enabled once it lands.

srid:
## The bug

Under a **long-lived daemon** — a remote-bound padi serving a whole e2e run —
the working-tree watcher leaks an **unreaped zombie child per subscribe**. They
pile up without bound (dozens of defunct `sh` in one run), invisible to CPU/RAM
(zombies are free), but they drag the daemon's event loop enough that Code-tab
renders time out in the **back half** of a long run. Surfaced by W3.4's remote
e2e lane (#1689); filed as #1691.

## Root cause (measured, not guessed)

`watchWorkingTree` let `@parcel/watcher` **auto-select** its backend. Auto-select
probes the **watchman** backend *first, on every `subscribe`*, via a native
`popen("watchman … get-sockname 2>/dev/null")`
(`WatchmanBackend::checkAvailable`). On a host **without watchman** — every kolu
pool box, most dev machines — that probe fails and parcel's error path never
`pclose()`s, so the `/bin/sh` `popen` forked is never `wait()`ed. Node's libuv
can't reap a child spawned by native `popen` (it isn't a `child_process`
handle), so the `sh` lingers as a zombie forever.

Bisected end-to-end on a box pair: falsified resource pressure (both boxes
128 GB, load <1); isolated the leak to `subscribeRepoChange` → `watchWorkingTree`
→ `@parcel/watcher.subscribe` (10 subscribes → 10 `sh` zombies) → the
`WatchmanBackend` `popen`; confirmed `git` is a real ELF binary (so the `sh` is
the popen shell, not a git wrapper); confirmed the box has no watchman.

## The fix

Pin the OS-native backend (`inotify` / `fs-events` / `windows`) so the watchman
probe never runs. Verified: the per-subscribe leak goes to **zero**. The only
thing forgone is watchman's fewer-inotify-slots optimization on hosts that opt
into it — which kolu's targets don't — so this is **pure win, not a dropped
fallback** (no-fallbacks stays honored: we pin the *correct* backend, we don't
add a degradation path).

## Repro-first coverage

A **daemon soak test** drives sustained subscribe/unsubscribe cycles and asserts
the child count returns to baseline — **red at 12 leaked children** on the old
auto-select path, **green** on the pinned backend. It's the first of its
coverage class: both this and #1680 (the darwin WAL-watcher death) were
long-lived-daemon defects a spin-up/tear-down unit test structurally can't see.
The timing/p99 half of the soak is guarded by the remote-e2e lane itself.

## The empirical claim the two-greens bar tests

The pin is expected to eliminate **both** failure modes the leak caused: the
**zombie accumulation** (asserted directly by the soak test here) **and** the
**second-half Code-tab latency** in a long remote run — because that latency
*is* the accumulated per-subscribe `popen` fork/leak dragging the daemon's event
loop under Code-tab load. #1689's warm-pool rerun (two consecutive greens on the
linux pair + one darwin-binds-linux) is the test of that claim. If the timeouts
**survive** the pin on the warm-pool rerun, the mechanism story is incomplete —
that is a signal to re-diagnose, not to widen the timeout multiplier.

## Doc sync

Updated the repo's `/parcel` skill — the backend-dispatch, watchman-invocation,
runtime-status, and WSL2-fallback sections now reflect the pin and document the
`popen` zombie leak, so the project's own watcher reference matches the shipped
behavior. The edit lives in the `.apm/skills/parcel/SKILL.md` **source** and is
regenerated into the `.claude`/`.agents` runtime trees via `just ai::apm` (all
three in sync), per the APM two-tree rule.

## Honoring the design philosophy

- **Fail-fast / no fallbacks** — pins the correct native backend rather than
  papering the leak with a periodic reaper; no new knob, no degraded path.
- **Reuse the source of truth** — the fix lives in `kolu-git`'s
  `working-tree-watcher`, the one module that owns the `@parcel/watcher` usage;
  no watcher logic duplicated elsewhere.

## Review gauntlet

Five lenses over the production delta (architecture/lowy, correctness, simplicity/hickey, then architecture-first-principles + perfection on the final form):

- **hickey** — the original `... : undefined` untargeted arm was a silent fallback that handed control back to parcel's leaking auto-select. Fixed → an explicit non-watchman default (later a `throw`).
- **architecture-first-principles (illegal-states)** — `BackendType` still spelled `"watchman"`, so a future edit could re-arm the leak and type-check. Fixed → `NonWatchmanBackend = Exclude<BackendType, "watchman">`; the leaky value is now a compile error.
- **perfection R2** — the `?? "brute-force"` fallback was a graceful-degradation path (silent poll on an unsupported platform) that the repo's fail-fast philosophy forbids. Fixed → `throw` on an unsupported platform.
- **perfection R1 (deferred, documented)** — a future `parcelSubscribe` in another module could forget the pin. The ideal fix (extract the raw subscribe behind a shared boundary + lint-ban the import) conflicts with lowy's lens: extracting for a population of one is premature receptacle-creation, and the repo's electricity principle treats a bounded single-consumer concern as a leaf. Adjudicated: the `/parcel` skill's "pin the backend the same way" note is the interim guard; a real second consumer is the extraction trigger.

CI: the full pipeline is green on both platforms in one clean run on the final
SHA — 28 ok · 0 failed · 0 errored (14 × x86_64-linux + 14 × aarch64-darwin).

Closes #1691. Unblocks the remote-e2e CI node in #1689.

### DIFF
--- a/.agents/skills/parcel/SKILL.md
+++ b/.agents/skills/parcel/SKILL.md
@@ -27,8 +27,14 @@ unless it has a similarly narrow target.
 
 Source: `node_modules/.pnpm/@parcel+watcher@2.5.6/node_modules/@parcel/watcher/src/Backend.cc:30-69`.
 
-`backend: "default"` (Kolu doesn't pass an explicit backend) selects in this
-order, first match wins:
+**Kolu pins the OS-native backend explicitly** (`inotify` / `fs-events` /
+`windows`, chosen off `process.platform` — see `PARCEL_BACKEND` in
+`working-tree-watcher.ts`). It does **not** use `backend: "default"`. The pin
+skips the auto-dispatch order below — most importantly the `WatchmanBackend`
+probe, whose per-subscribe `popen` leaks a zombie on watchman-less hosts (see
+next section). If you write a new parcel consumer, pin the backend the same way.
+
+`backend: "default"` would select in this order, first match wins:
 
 1. `FSEvents` on macOS — native recursive, one stream per repo.
 2. `WatchmanBackend` if `WatchmanBackend::checkAvailable()` returns true.
@@ -37,8 +43,9 @@ order, first match wins:
 5. `KqueueBackend` on BSD.
 6. `BruteForceBackend` — periodic full-tree stat; the fallback fallback.
 
-So on Linux, watchman is preferred over inotify whenever it's reachable; on
-macOS watchman is never used by default (FSEvents wins).
+So under `"default"` on Linux, watchman is *probed* (and preferred if reachable)
+before inotify; on macOS watchman is never used (FSEvents wins). The explicit
+pin lands directly on step 1/3/4 for the platform and never runs step 2's probe.
 
 ## How parcel invokes watchman
 
@@ -51,6 +58,16 @@ Source: `src/watchman/WatchmanBackend.cc`.
    ```
    then parses BSER output for the `sockname` field. If `WATCHMAN_SOCK` env
    var is set, that wins and the binary isn't run at all.
+
+   > **⚠ Zombie leak — why Kolu pins the backend (juspay/kolu#1691).** On a
+   > host without watchman, this `popen` forks `/bin/sh`, the `watchman` exec
+   > fails, and parcel's error path returns **without `pclose()`** — so the
+   > `sh` is never `wait()`ed. `popen` bypasses `child_process`, so Node's
+   > libuv can't reap it either: it lingers as a zombie **forever, one per
+   > `subscribe`**. Free on CPU/RAM but a long-lived daemon (a remote-bound
+   > padi serving a full e2e run) accumulates dozens and the event loop drags.
+   > Pinning the native backend (above) skips `checkAvailable()` entirely, so
+   > the `popen` never runs. Verified: per-subscribe leak → 0.
 3. From there it's a Unix-domain socket carrying BSER-encoded JSON. No more
    subprocess spawns.
 
@@ -87,9 +104,13 @@ mean the daemon was killed mid-handshake. Add `.watchman-cookie-*` to
 ## Kolu's runtime status
 
 As of #788, Kolu does **not** ship watchman with the production binary. The
-`nix run` wrapper (`default.nix:156`) only adds `nodejs git gh` to PATH, so
-`checkAvailable()` always returns false at runtime and parcel falls through to
-inotify on Linux / FSEvents on macOS. Issue #788 tracks the integration work.
+`nix run` wrapper (`default.nix:156`) only adds `nodejs git gh` to PATH — so
+even under `backend: "default"` the watchman probe would always fail and parcel
+would fall through to inotify on Linux / FSEvents on macOS. Since #1692, Kolu
+doesn't rely on that fall-through: it **pins** the native backend, so
+`checkAvailable()` (and its leaking `popen`) never runs at all. Issue #788
+tracks the watchman integration work; if it lands, the pin is where you'd
+re-enable watchman deliberately (with a `WATCHMAN_SOCK` that avoids the probe).
 
 ## Kolu's wrapper invariants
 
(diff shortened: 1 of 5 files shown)


srid:

The **x86_64-linux** lane ran on **`kolu-ci-5`** (`idliv2-02`) — commit `becb973f7`, **passed** ✓

- **Lane wall** (pipeline): **3m17s**

| recipe | duration | |
|---|--:|:--|
| `ci::home-manager` | 2m42s | ✓ |
| `ci::e2e` | 1m58s | ✓ |
| `ci::unit` | 1m13s | ✓ |
| `_ci-setup` | 35s | ✓ |
| `ci::pnpm-hash-fresh` | 20s | ✓ |
| `ci::smoke` | 16s | ✓ |
| `ci::fmt` | 11s | ✓ |
| `ci::biome` | 11s | ✓ |
| `ci::nix` | 10s | ✓ |
| `ci::install` | 7s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/becb973/timings.jsonl`; pool state is a live `flock` probe.</sub>

PR #1703, opened 2026-07-05
refactor(padi): make the daemon lifecycle invariants explicit (L15+L16)

srid:
Two padi-daemon lifecycle invariants that were held together by prose and per-call-site comments become **types and explicit state** — so violating them is now a compile error or an impossible state, not a silent bug. This is a **behavior-preserving refactor**: the daemon does exactly the same thing, and every red-when-reverted session-safety test stays green *unchanged* (assertions untouched; only symbol locations and the gate's init signature moved).

Both entries live in the same subsystem (`packages/padi/src/`) and **intersect at `daemonMain.ts`** — the boot sequence (L16) wires the autosave step (L15's gate) — so they ship as one coherent "daemon lifecycle correctness" PR.

## L15 — the `metadata.ts` write-seam invariants

"Is it safe to `saveSession()` now, and with what value?" was assembled from **five scattered sources**: a module-level timer, a restart flag, a live `hasParkedTerminals()` query, a cell `onWrite` cancel, and four "must NOT arm autosave here" prose conventions. They collapse into **one `AutosaveGate` module** owning explicit state:

| State | Meaning | How it's set |
|---|---|---|
| `idle` | no save pending | — |
| `armed` | a dirty pulse scheduled a save (leading-edge throttle) | the `terminals:dirty` pulse |
| `frozen(reason)` | a restart's capture→drain→park critical section | **pushed** by `restartLocal.ts` (`freezeAutosave`/`unfreezeAutosave`) |
| `suppressed(parked)` | a restore is pending | **live query** at fire (`isRestorePending`, injected) |

The gate is a **pure state machine over injected effects** (`snapshot` / `persist` / `isRestorePending`) — it imports nothing from the session/registry layers, which *push* their facts and effects in. Restart is a push; **parked stays a live fire-time query** because the terminal-registry is its sole source of truth and the value can change between arm and fire — mirroring it in the gate would duplicate that truth and stale the PATH-B guard. (This one design point diverges from the ledger's literal "registry PUSH" wording; it was consulted and [recorded in the ledger](https://github.com/juspay/kolu/commit/c2d8df249) before implementing.)

Every writer now pulses a single **`notifyDirty()`**; every "must-not-arm" paragraph is gone (a seam that shouldn't arm simply doesn't call it).

**Folded-in urgency invariant:** `publishUrgency()` now rides `publishComposedTerminal`, so **one call site** answers "did this write move the urgency projection?" The existing `urgencyEqual` dedup makes the extra fires free, so the prose justifications on `commitSnapshot`/`updateMemory`/`updateClientMetadata` are deleted with no behavior change.

## L16 — a typed boot pipeline

`runPadiDaemon`'s ~10 boot concerns were ordered only by comments ("Claim the gate FIRST", "Runs BEFORE the stores are injected"). The boot is now a **typed pipeline** — `gate → stores → identity → serve → endpoint` — where each phase takes the prior phase's token:

```
openStateStores(gate: HeldGate, …)        → StoresReady
configureDaemonIdentity(stores: StoresReady, …) → IdentityReady
serveDaemonSurfaces(identity: IdentityReady, …) → SurfacesServed
bootLocalEndpoint(served: SurfacesServed, …)    → EndpointBooted
```

A reorder is now a **compile error**. "Claim the gate FIRST" (W2.2's B1 blocker) is the `HeldGate` type threaded through every disk-touching phase; "stores injected before any read" is `StoresReady`. The **second instance** (`ptyHost/index.ts` `ensureLocalEndpoint`) is recorded **fine-as-is** with an in-code reason: its order is already enforced by *data flow* (`ep = createEndpoint(); converge({ endpoint: ep })`), not prose — there's no side-effecting "must run before X" edge for a token to guard.

## Invariant → the test that now pins it as typed

| Invariant (was prose) | Now | Proof test (green, unchanged) |
|---|---|---|
| restart freeze covers the drain→park gap | `frozen(reason)` gate state | `restartLocal.test.ts` — ordering guard (autosave fires with snapshot=0, freeze keeps it unnulled) |
| never shrink a restore-pending blob | `suppressed(parked)` live query | `session.test.ts` (vi)/(iv) — PATH-B guards |
| which seams arm autosave | presence/absence of `notifyDirty()` | `metadata.test.ts` — dirty-count routing (commit/memory silent; client/lifecycle arm) |
| gate claimed before any side effect | `HeldGate` on every phase | `daemonMain.test.ts` — a second padi exits `already-running` touching nothing |
| daemon still boots + dials | pipeline unchanged in behavior | `dial.test.ts` |

## Design philosophy

- **Fail-fast / no fallbacks** — no override knobs added; the gate crashes on a missing injected effect rather than degrading. Parked stays a live query rather than a stale mirror.
- **Reuse the source of truth** — the terminal-registry remains the *sole* authority for "is a restore pending"; the gate consults it (injected) instead of duplicating it.
- **Volatility boundaries** — the AutosaveGate isolates the "when to persist" volatility (throttle, freeze, suppression) behind one seam; writers only pulse `notifyDirty()`.

## Scope / constraints

- **`packages/padi/src/` only** — no `@kolu/surface*` or other package touched.
- Draft; **do not merge** — review path is coordinator → srid.

### DIFF
--- a/packages/padi/src/assembly.ts
+++ b/packages/padi/src/assembly.ts
@@ -91,9 +91,14 @@ export { buildPadiSurfaceDeps } from "./servePadi.ts";
 // ── session persistence ─────────────────────────────────────────────────
 export {
   cancelPendingAutosave,
+  freezeAutosave,
+  initAutosaveGate,
+  notifyDirty,
+  unfreezeAutosave,
+} from "./autosaveGate.ts";
+export {
   clearSavedSession,
   getSavedSession,
-  initSessionAutoSave,
   saveSession,
   setSavedSession,
   setSavedSessionFromSnapshot,
--- a/packages/padi/src/assembly.ts
+++ b/packages/padi/src/assembly.ts
@@ -68,6 +68,7 @@ export {
 } from "./stateRoot.ts";
 // ── publisher / surface ctx holder ──────────────────────────────────────
 export {
+  notifyDirty,
   publisher,
   publisherSize,
   terminalsDirtyChannel,
@@ -93,7 +94,6 @@ export {
   cancelPendingAutosave,
   freezeAutosave,
   initAutosaveGate,
-  notifyDirty,
   unfreezeAutosave,
 } from "./autosaveGate.ts";
 export {
--- a/packages/padi/src/assembly.ts
+++ b/packages/padi/src/assembly.ts
@@ -14,14 +14,31 @@
  * INJECTED via `setDaemonProcessId` / `setSpawnServerVersion`, not imported.
  */
 
-// The persisted survivor pairing's type. The pairing is READ + RECORDED entirely
-// inside padi's boot reconcile (its conf store is set by padi's own `daemonMain`).
-export type { PairedDaemon } from "./pairedDaemon.ts";
+// ── session persistence ─────────────────────────────────────────────────
+export {
+  cancelPendingAutosave,
+  freezeAutosave,
+  initAutosaveGate,
+  unfreezeAutosave,
+} from "./autosaveGate.ts";
 // padi's staleKey read — the binder's build-convergence key (#1670). Re-exported
 // through this barrel (not a deep `@kolu/padi/buildId` import) so the binder honors
 // the package-boundary seal: on boot it compares its own baked `PADI_BUILD_ID`
 // against the running padi's `hello.buildId` and drains a same-contract build change.
 export { currentPadiBuildId } from "./buildId.ts";
+// ── host-daemon inventory scanner (the "Running daemons" leak diagnostic) ─
+// The ONE scanner both padi (its `hostInventory` member) and kolu-server's web shell
+// (its local-machine `daemonInventory.localScan` under a remote binding) reuse. Padi
+// owns the daemon domain, so the scan homes here; kolu-server imports it through this
+// barrel for its local arm (the package-boundary seal's allowed direction).
+export {
+  assembleKavalInventory,
+  assemblePadiInventory,
+  enumerateHostDaemons,
+  type HostDaemonScanDeps,
+  type KavalProbe,
+  probeKavalStatus,
+} from "./hostInventory.ts";
 // ── scratch / roots ─────────────────────────────────────────────────────
 export {
   ensureKoluRoot,
@@ -33,6 +50,9 @@ export {
   padiSurfaceCtx,
   setPadiSurfaceCtx,
 } from "./padiSurfaceCtx.ts";
+// The persisted survivor pairing's type. The pairing is READ + RECORDED entirely
+// inside padi's boot reconcile (its conf store is set by padi's own `daemonMain`).
+export type { PairedDaemon } from "./pairedDaemon.ts";
 // The range-capable serve-dir read kolu-server's re-backed Hono preview route
 // calls — the STREAMING form (`previewFile`, bounded heap), the same read
 // `preview.read` serves through its base64 wire-wrapper (`readPreview`).
@@ -49,6 +69,22 @@ export {
   ptyHostClient,
   setSpawnServerVersion,
 } from "./ptyHost/index.ts";
+// ── publisher / surface ctx holder ──────────────────────────────────────
+export {
+  notifyDirty,
+  publisher,
+  publisherSize,
+  terminalsDirtyChannel,
+} from "./publisher.ts";
+// ── native serving (W1.R0) ──────────────────────────────────────────────
+export { buildPadiSurfaceDeps } from "./servePadi.ts";
+export {
+  clearSavedSession,
+  getSavedSession,
+  saveSession,
+  setSavedSession,
+  setSavedSessionFromSnapshot,
+} from "./session.ts";
 // ── padi process rendezvous (W2.2 binder) ───────────────────────────────
 // kolu-server's padi BINDER (`server/src/padiBinding.ts`) resolves the SAME
 // state-root → socket/gate paths padi computes for itself, so the supervisor and
@@ -56,8 +92,8 @@ export {
 // binder reaches the terminal domain only through @kolu/padi's published entry
 // points (the package-boundary seal), not a deep `@kolu/padi/stateRoot` import.
 export {
-  type PadiDaemon,
   discoverPadiDaemons,
+  type PadiDaemon,
   padiDigest,
   padiGatePath,
   padiKavalSocketPath,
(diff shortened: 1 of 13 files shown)


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `51fc8feb7068`

### Applied (5)
- `lowy-1` Phase tokens don't brand — 2 of 4 ordering edges are unenforced, and the header comment claims a guarantee TS doesn't give — commit `552fb0e83`
- `lowy-2` notifyDirty (the shared pulse's write side) is co-located with only ONE of the pulse's two consumers — commit `9faa50061`
- `lowy-3` publisher.ts channel doc describes terminals:dirty as autosave-only, though liveActivity also subscribes — commit `0209d0cf1`
- `hickey-1` notifyDirty (the shared pulse's emit) placed inside one of its two consumers instead of with its channel — (uncommitted)
- `hickey-2` Boot phase tokens do not nominally brand the ordering the prose promises the compiler enforces — (uncommitted)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `51fc8feb7068`

### Round 1

**codex** — approved: `false`

The autosave gate extraction, restart freeze behavior, live parked query, notifyDirty consolidation, and padi-only scope all look consistent with the stated intent. I did not find a correctness issue in the changed behavior. I could not run the padi typecheck because `pnpm` is not available in this shell.

Findings:
- `F1` · nit · open — The test header still says the PATH-B fix is one line in `session.ts`'s autosave callback, but this PR moved that decision into `autosaveGate.ts`. That makes the red-when-reverted note point future maintainers at the wrong file/branch. (packages/padi/src/session.test.ts:10)


**claude** — Fixed codex's single nit by repointing the stale PATH-B comment from `session.ts` to the `autosaveGate.ts` `isRestorePending()` guard where the decision now lives.

- `F1` **fixed** — Codex is right: the L15 refactor moved the parked-clobber guard out of `session.ts`'s autosave callback into `autosaveGate.ts`'s fire decision (`autosaveGate.ts:119` → `isRestorePending()` short-circuit to `suppressed-parked`), but the test header at `session.test.ts:10` still described the fix as `if (hasParkedTerminals()) return;` in `session.ts`. That would send a future maintainer to the wrong file when a red-when-reverted test fails. Updated the comment to name the live `isRestorePending()` guard in `autosaveGate.ts` and its `suppressed-parked` outcome. Comment-only change; `just check` (padi typecheck + biome lint) green.

commit: `9b997da0fc2b095c319cdf0a09ad17182cfbd85d`


### Round 2

**codex** — approved: `true`

Verified the current tree for the existing finding only. `git status --short` is clean, and `packages/padi/src/session.test.ts` now points the PATH-B comment at `autosaveGate.ts`'s live `isRestorePending()` fire-time guard and `suppressed-parked` outcome. No new issue was introduced by this comment-only fix.

Findings:
- `F1` · nit · resolved — The test header still said the PATH-B fix was one line in `session.ts`'s autosave callback, even though the refactor moved the decision into `autosaveGate.ts`. (packages/padi/src/session.test.ts:10)

_codex on the rebuttal:_ F1 is resolved. CLAUDE updated the stale comment to reference `autosaveGate.ts`'s fire decision and the live `isRestorePending()` guard, which addresses the maintainer-misdirection issue.

srid:

Two passes over the L15+L16 diff (`--no-elegance`, since `/simplify` already ran this gauntlet), scoped to `packages/padi/src/`.

**Pass 1 — rule checklist: no violations.** Every applicable rule checked and clean. Notably the diff *fixes* a prior gap: `callback-fanout-guarded-at-funnel` — before, three producers called `terminalsDirtyChannel.publish({})` unguarded; now every producer funnels through the single guarded `notifyDirty()`. Also verified: no dead code left by the extraction (all removed symbols grep-clean), no silent error-swallowing (both new catches log at `.error`), `invalid-states-unrepresentable` holds (`saveTimer`/`frozenReason` are genuinely orthogonal axes, not a collapsible enum), and the `assembly.ts` barrel edit stays a documented boundary re-export.

**Pass 2 — fact-check: no findings.** Seam-by-seam confirmation that the refactor is behavior-preserving: all six `terminals:dirty` call sites map 1:1 with their guards intact, the boot sequence is reproduced with no step reordered or dropped, the freeze/capture/drain/park order is unchanged, and the one behaviorally-additive spot (the `publishUrgency()` fold onto `updateMemory`/`updateClientMetadata`) dedups to a genuine no-op via the pre-existing `urgencyEqual` cell `equals`.

**Actioned:** one import-ordering slip (the new imports weren't in biome's declared alphabetical order — `just check` runs `biome lint`, which skips the organize-imports assist, so it wasn't auto-caught). Fixed with biome's organizer on the changed files (commit `06992aa0a`), respecting `terminals.ts`'s cycle-sensitive opt-out.

**Noted, out of scope (not fixed):** a stale doc-comment in `packages/tests/step_definitions/session_restore_steps.ts` still names the renamed `initSessionAutoSave` — left untouched to honor this PR's `packages/padi/src/`-only scope (the surface refactor collision guard).

srid:
## Evidence

**No visual impact — evidence skipped by the gate.** This is a pure internal padi-daemon refactor with no externally-visible change: the daemon behaves identically, and session-survives-restart, autosave, and urgency are byte-identical before/after. The proof is the **red-when-reverted round-trip test net staying green with assertions untouched** — `ptyHost/restartLocal.test.ts` (the drain→recycle→park session-survival ordering guard), `session.test.ts` (the PATH-B parked-clobber guards), and `terminalEndpoint/reattach.test.ts` — plus the `daemonMain.ts` gate-first and `dial.test.ts` boot tests. All 206 padi unit tests pass unchanged across every gauntlet step. There is no new or changed on-screen behavior to capture.

srid:

The **x86_64-linux** lane ran on **`kolu-ci-8`** — commit `06992aa0a`, **passed** ✓

- **Lane wall** (pipeline): **4m11s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 3m54s | ✓ |
| `ci::home-manager` | 2m38s | ✓ |
| `ci::unit` | 1m43s | ✓ |
| `ci::nix` | 1m10s | ✓ |
| `ci::flake-check` | 1m6s | ✓ |
| `ci::smoke` | 1m0s | ✓ |
| `ci::atlas-sync` | 35s | ✓ |
| `ci::surface-example-build` | 34s | ✓ |
| `ci::surface-app-example-build` | 34s | ✓ |
| `ci::biome` | 33s | ✓ |
| `ci::fmt` | 31s | ✓ |
| `ci::install` | 9s | ✓ |
| `_ci-setup` | 7s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|
| `kolu-ci-1` | ? | ✓ idle |
| `kolu-ci-2` | ? | 🔒 leased |
| `kolu-ci-3` | ? | ✓ idle |
| `kolu-ci-4` | ? | ✓ idle |
| `kolu-ci-5` | ? | ✓ idle |
| `kolu-ci-6` | ? | ✓ idle |
| `kolu-ci-7` | ? | ✓ idle |
| `kolu-ci-8` | ? | ✓ idle |

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/06992aa/timings.jsonl`; pool state is a live `flock` probe.</sub>

PR #1704, opened 2026-07-05
make send honest: two-command submit, no timing magic

### LINKED ISSUE #1702: kaval-tui send: large prompts to a Claude TUI don't reliably submit (paste-collapse); --submit only fixes small; no write-deadline hang; --submit footgun; request --file
## Summary

Driving an agent (Claude Code TUI) with `kaval-tui send` (v5.0) is unreliable for **large prompts**, plus a `send` hang when the target stalls on stdin, a `--submit` footgun, and an ergonomics gap. Findings below are from a real reproduction pass (two earlier framings were corrected — noted inline).

## Bug A — large pastes to a Claude TUI don't reliably SUBMIT (the core problem)

A big paste (~6 KB) to a real Claude Code TUI is folded into a **"paste again to expand"** placeholder. Submitting it is flaky **regardless of how the Enter is sent** — verified against a live claude:

- same-breath `send <id> "<big>" --key Enter` → ambiguous: "Crunched for 1s", empty prompt, "paste again to expand" persists.
- single-command `send <id> --submit=400 "<big>"` → **identical** ambiguous result.

So `--submit`'s grace does **not** fix the large-paste case (see Bug B for what it *does* fix). This is the behavior that repeatedly forced a human to press Enter by hand when dispatching large briefs to agents. **There is currently no verified-reliable way to submit a large prompt to a Claude TUI over `send`.** Likely the highest-value fix — see the `--file` request, which may bypass the collapse.

## Bug B — small/normal prompts: same-breath Enter is dropped; `--submit=<ms>` fixes THAT (only)

For a *normal-size* prompt, `send <id> "<text>" --key Enter` (same breath) drops the Enter (paste debounce — the tool's help documents this). `send <id> --submit=<ms>` waits a grace then sends Enter, and submits cleanly — verified for real ~700-byte messages. **Scope: `--submit` reliably fixes the normal-size case; it does NOT fix the large-paste collapse (Bug A).**

## Bug C — `send` blocks indefinitely when the target isn't draining stdin (no write deadline)

Deterministic against a terminal running a plain `cat` (which can stop reading stdin when its output backs up):

```bash
python3 -c "print('\n'.join('line %d ---- filler'%i for i in range(80)))" > big.txt   # ~6.4 KB
time kaval-tui send <cat-id> "$(cat big.txt)"        # ~0.36s alone
time ( kaval-tui send <cat-id> "$(cat big.txt)" && sleep 1 && kaval-tui send <cat-id> --key Enter )
#   → the SECOND send blocks > 1 min (killed by timeout)
```

Corrected scope: a **healthy** Claude TUI drains input fine (big paste 0.35s, follow-up Enter 0.35s, full chain 1.7s — idle or busy). The hang triggers only when the target **stops consuming stdin**. Defect: `send` has **no write deadline** — a blocked PTY write hangs forever. Fail-fast: bound it, exit non-zero on timeout.

## Bug D — `--submit` footgun: a bare `--submit` swallows the following prompt word

`kaval-tui send <id> --submit "my prompt"` → `--submit` eats `"my prompt"` as its grace value (it takes an optional value). It errors rather than mis-sending, but it's an easy trap. Suggest: make `--submit` a pure boolean + a separate `--grace <ms>`.

## Feature request — `send --file <path>` (now the most promising fix for Bug A)

Large prompts are passed as `"$(cat file)"`, which mangles payloads with backticks / `$(...)` (hit live: a message with backticked `system.identity` was partially shell-executed). More importantly, an explicit `--file` (or the existing but undiscoverable `send <id> < file` stdin path) might feed the payload in a way that **avoids the Claude paste-collapse of Bug A** — worth testing as the primary fix, not just an ergonomics nicety. Should pair with `--submit`.

## Net

- **Bug A is the real pain** (large agent briefs won't reliably submit) and currently has no verified workaround.
- `--submit=<ms>` reliably handles normal-size prompts only.
- Bug C (no write deadline) is a separate true hang worth fixing regardless.
- `--file` is the most promising lead for Bug A + removes the `"$(cat)"` shell-mangling risk.

srid:
Driving a TUI agent with `kaval-tui send` carried a same-breath-Enter footgun, a real hang, and a shell-mangling gap. This makes `send` **honest**: submitting a prompt is its own command, timed by the *caller* observing the TUI settle — not by a grace `kaval` bakes in and can't win.

> **Scope — read this first.** This PR is the **mechanics fix + write-deadline** (#1702). It makes the two-command submit honest for **normal-size** prompts and turns a real hang into a loud failure. It does **NOT** fix **Bug A** — a *large* paste that Claude Code folds into a `[Pasted text +N lines]` placeholder still does not submit on Enter. Bug A is **confirmed to survive** the canonical flow against a live claude and is **split out to #1707** (see the Bug A section below). The docs here say so plainly; nothing claims large pastes submit.

The ruling (srid): same-breath Enter is not supported; the two-command pattern is the model. Any fixed grace baked into one command is a race with the target TUI's internal debounce/collapse timing, which `kaval` **cannot observe**. The honest design: `kaval` can't know when the TUI settled, but the **caller can** — `kaval-tui wait --until idle` exists for exactly this.

## The canonical flow

```sh
kaval-tui send <id> --file brief.md          # 1. the text (a bracketed paste, no Enter)
kaval-tui wait <id> --until idle:300         # 2. OBSERVE the TUI settle (a signal, not a sleep)
kaval-tui send <id> --key Enter              # 3. submit
```

## Piece-by-piece map

| # | Change | Where |
| - | ------ | ----- |
| 1 | **Forbid text + `--key` in one send** — a same-breath Enter races the paste debounce and is dropped, so `send <id> "text" --key Enter` is now a **hard error** whose message teaches the two-command flow. The trap is unspellable, not warned-about. | `send.ts` `resolveSendInput` (pure, unit-tested); wired pre-dial in `main.ts` |
| 2 | **Delete `--submit` entirely** — no deprecation shim (fail-fast doctrine). Its fixed grace was a knob you tune until the race stops biting on your machine and starts again on a slower one. Gone: `parseSubmitGrace`, `submitSwallowedPositional`, `SUBMIT_ENTER`, `DEFAULT_SUBMIT_GRACE_MS`, `SendPlan.submit`. And **typing `--submit` (or any unknown flag) now fails loud** — `send`'s flag set is closed, so an unknown flag can't be silently ignored (that would send the text *unsubmitted*, the very footgun removed); `--submit` gets a migration message pointing at the two-command flow. | `send.ts` `rejectUnknownSendFlags`, `main.ts` |
| 3 | **Bounded write deadline** — a PTY write that can't complete within a fixed **8s** deadline exits non-zero with a loud message naming the stalled terminal, instead of hanging forever (Bug C: a `cat` that stopped draining stdin blocked `send` >1min). Fixed in code, not a flag (no knob to tune the never-hang away). | `sendExec.ts` `SEND_WRITE_DEADLINE_MS` + `writeWithDeadline` |
| 4 | **Add `--file <path>`** — read the payload straight from a file, byte-exact, no shell in the loop. Kills the `"$(cat file)"` mangling of backticks / `$( )`. Mutually exclusive with positional text **and** piped stdin. Does not change the wire bytes (still a bracketed paste) — fixes the **shell** hazard only. | `main.ts` `--file` flag, `readSendText`, `stdinIsPayload` |
| 5 | **Docs** — the kolu skill (apm source), `kaval-tui`/`padi-tui` READMEs, `website/kaval.astro`, the pty-daemon atlas note, and the changelog all get the canonical three-step flow, replacing every `--submit` / same-breath reference. | docs |

### Design detail: how `--file` stays usable when an agent drives it

The `--file`/stdin conflict is detected by `fstat`-ing fd 0, **not** `!isTTY`. An agent driving `send` from a subprocess inherits a **`socket`** on stdin (empirically confirmed), which is ambient — not a payload. A deliberate pipe is a `fifo` (`| cmd`) or a `regular file` (`< file`). Tying the conflict to `!isTTY` would have made `--file` unusable in exactly the agent context it exists for; the `fstat` distinction keeps it usable while still catching a real `< file` collision.

## Prior art: tmux `send-keys`

`tmux send-keys` is the most battle-tested "inject input into a PTY someone else is watching" tool. It **validates every pillar of this design**:

- **(b) Submit convenience / grace?** None. `send-keys` is strictly compositional — no implicit Enter, no debounce, no grace knob. To submit you invoke `send-keys Enter` separately. The closest tool in this space keeps key-sending pure → **validates deleting `--submit`**. (tmux *does* allow `send-keys foo Enter` in one call, but it does **not** auto-bracket paste, so it has no debounce race to lose; we auto-bracket, so we forbid the mix — a justified stricter rule.)
- **(a) Large payloads?** tmux's large-payload path is *separate* from `send-keys`: `load-buffer` (from a file, or stdin `-`) → `paste-buffer -p`. Reading from a file/stdin bypasses the shell's `"$(cat)"` entirely → **validates `--file`**. Notably `paste-buffer -p` still emits a *bracketed paste* (only when the app requested bracketed-paste mode) — so it too would hit Claude's large-paste collapse. This confirms `--file` fixes the **shell** hazard, not the collapse.
- **(c) Non-draining pane?** tmux never blocks the client — its persistent server buffers input asynchronously and returns at once. A one-shot CLI has no server-side buffer to hand off to, so the fail-fast analog is a **bounded deadline + loud exit** → **validates the write deadline**.

No design change resulted from the research; tmux confirmed all three pillars.

## Design philosophy

- **Fail-fast, no fallbacks, no knobs** — `--submit` deleted with no shim; the write deadline is fixed in code, not a flag; every illegal arg combination crashes loud (`resolveSendInput`) rather than silently picking a precedence.
- **Reuse the source of truth** — the settle signal reuses the existing `wait --until idle`; keys reuse `@kolu/terminal-protocol`'s byte vocabulary; the deadline reuses the package's timer module.

## Tests

- **`resolveSendInput`** (pure arg-legality matrix): text+`--key` rejected · `--file`+positional rejected · `--file`+stdin rejected · positional+stdin rejected · empty send rejected.
- **`executeSendPlan` write deadline** (fake timers): a never-draining sink fails loud *within* the deadline (not a hang), the message names the stalled terminal + the bound, and a fast write is not interrupted.
- **`--file` byte-exactness**: a payload with backticks / `$( )` / `${…}` reaches the wire verbatim (the exact class `"$(cat)"` mangled).
- **Acceptance vs. the real paste-debounce TUI fixture** (rewritten to the canonical flow): a same-breath paste+Enter is DROPPED (proving *why* the CLI forbids it); paste → observe settle → separate Enter SUBMITS for an idle agent, a busy mid-turn agent, and a **multi-KB paste** (the Bug-A shape).

- **Unknown-flag rejection** (`rejectUnknownSendFlags`): `--submit` throws the migration message; a typo fails generically naming it; `--no-paste` stays known.

All 124 `kaval-tui` tests pass locally; `just check` (tsc + biome workspace) green.

## Bug A — SURVIVES the canonical flow (split out to #1707)

Bug A is the issue's core problem: a large brief to a real Claude Code TUI does not reliably submit because Claude folds a big paste into a `[Pasted text +N lines]` / "paste again to expand" placeholder. **This PR does not fix it.** A live-claude run of the exact canonical flow, clean state, confirmed it survives:

1. `send <id> --file <5.9 KB>` → pasted OK.
2. `wait <id> --until idle:400` → fired at **411 ms** — Claude *did* go idle. But idle only proves Claude finished **folding** the paste into its placeholder; it does **not** prove a following Enter is deliverable or submits.
3. `send <id> --key Enter` → the Enter **write HUNG the full 30 s** (killed by timeout) **and** the paste stayed **STAGED** at the prompt (`[Pasted text +81 lines]`), never submitted.

Two failures compound: **(a)** the Enter *write* stalls against the placeholder state (the classic Bug C — now confirmed to bite a **real** claude, not just a `cat`), and **(b)** even setting the hang aside, the folded large paste does not submit via observe-then-Enter (consistent with the issue's original finding that `--submit`'s Enter didn't submit large pastes either).

**What this PR contributes to Bug A:** failure **(a)** is turned into a **loud, bounded failure** by the write-deadline instead of a >30 s hang — so the real-claude hang is now *validated* evidence for the deadline, not just the `cat` repro. Failure **(b)** — the actual submit — is **unfixed** and needs a deeper approach (see the tmux section); it is tracked under #1707 and the docs say so plainly.

### Does tmux's paste path fix Bug A? No.

The tmux research asked whether `load-buffer` + `paste-buffer -p` (or `-H` hex / `-l` literal) could **deliver and submit** a multi-KB payload a TUI would fold. Reading the tmux man page mechanics against this problem: **it cannot.**

- `paste-buffer -p` wraps the buffer in bracketed-paste control codes **only if the app requested bracketed-paste mode** — i.e. it emits the **same bracketed paste** our `send --file` does, which Claude folds **identically**. No help.
- `paste-buffer` without `-p` (default) replaces each **LF → CR**, so a multi-line payload **submits line-by-line** (N premature Enters) — worse, not better. `-r` keeps LF, but a raw non-bracketed multi-KB blast is app-specific and still doesn't submit a folded block.
- `-H` (hex) / `-l` (literal) only change **byte encoding / key-name lookup**, not the target's fold-then-submit behavior.

tmux's real, transferable wins are **file/stdin ingestion** (no shell mangling) and **non-blocking server-side buffering** (the client never hangs). This PR already carries both — as **`--file`** and the **write-deadline**. The fold-and-won't-submit is **Claude Code's own input model**, which no injector-side byte trick (tmux's or ours) overrides. So the deeper Bug A fix is not a tmux-style paste path; the pragmatic path (documented as the workaround) is to **not paste the brief** — write it to a file and send a *short* `read <path>` prompt, which submits cleanly.

## Findings — an apm defect (filed as #1706)

The generated agent-doc tree here (`.claude/skills/kolu`, `.agents/skills/kolu`) is **hand-synced** to the `.apm` source rather than regenerated, because **`apm` regeneration is broken for this repo** — filed as **#1706** with the evidence below:

- `just ai::apm` resolves the `local_path: ./agents` dependency to a **stale cached git commit** (`dbff5c22c`, #1616 — *pre-#1673*, before `--submit` even existed), ignoring the current checkout's HEAD entirely.
- Confirmed cache-clear-proof and worktree-independent: `rm -rf ~/.cache/apm/git` + `apm deps update agents`, a fresh non-worktree clone from the repo, and a fresh clone from the feature worktree **all** regenerated the same pre-#1673 content — never the source, never even `--submit`.
- So regeneration would *regress* the tree, not sync it. Per coordinator ruling, the two generated targets are **hand-edited to byte-mirror** `agents/.apm/skills/kolu/SKILL.md` (three-step flow + `--file` + busy-agent nuance, no `--submit`). Regeneration follows post-merge once #1706 is fixed.

## Review & CI status — gauntlet complete, CI green

- **Lens debate (lowy ⇄ hickey):** consensus. Applied a discriminated `SendInput` descriptor (kills an `as string` cast + an untyped invariant), an exhaustive `sourceIsStream` switch, single-sourced `SUBMIT_FLOW_HELP`, and made the text-XOR-keys invariant a **type** (`planSend` takes discriminated content; `SendPlan` carries one `write`).
- **Codex debate (xhigh, 3 rounds):** consensus. Caught two real defects — silent zero-byte sends on empty payloads, and a shell error-swallowing footgun in the busy-agent doc snippet — both fixed (+ empty-payload tests).
- **Simplify:** reuse/efficiency/simplification clean; altitude added an exhaustive `sourceLabel` as the one home for each source's human name.
- **Code-police:** all clear (rules + fact-check), two near-misses spot-checked as non-violations.
- **CI:** **green on both platforms** — `x86_64-linux` *and* `aarch64-darwin` — every required context (unit, e2e, nix, home-manager, smoke, atlas-sync, surface builds, pnpm-hash-fresh, …). 127 kaval-tui unit tests pass.

Ready for review. Not merged — the human merges.

**Fixes #1702** — #1702 is now scoped to the send **mechanics** (same-breath rejection, `--submit` removal, write-deadline, `--file`), all of which this PR delivers, so it correctly closes on merge. **Bug A** (a folded large paste won't submit) has been split out to **#1707** with the full live-test evidence, the tmux verdict, and the `read <path>` workaround — it stays open there as a separate live-claude investigation.


srid:
## [🤖 Codex ⇄ Claude debate](https://kolu.dev/)

**Consensus** in 3 rounds · reviewer effort: xhigh · base `cac4b96`.

### Round 1

**codex** — approved: `false`

The core direction is coherent: `send` now has a closed input-source validator, removes the same-breath text+Enter footgun, adds `--file`, and bounds stalled writes. I found two real fail-fast regressions and one stale-documentation/test nit. This was a read-only review; I did not run the test suite.

Findings:
- `F1` · major · open — Empty text sources now become successful zero-byte sends. `resolveSendInput` validates that a source exists, but after `readSendText()` returns `""`, `cmdSend` still builds and executes a text plan. So `--file empty`, an empty pipe/heredoc, or `send <id> ""` exits 0 with `sent 0 bytes`, masking upstream failures and regressing the old post-read `nothing to send` guard. (packages/kaval-tui/src/main.ts:534)
- `F2` · major · open — The busy-agent snippet uses `|| true` after `kaval-tui wait --timeout`, but only exit 2 means "target busy". This also hides usage/link errors, terminal-gone exit 3, and interrupts, then sends Enter anyway. Because this skill drives real agents, the snippet teaches silent error swallowing in the exact automation path this PR is hardening. The hand-synced `.agents/` and `.claude/` copies contain the same line. (agents/.apm/skills/kolu/SKILL.md:104)
- `F3` · nit · open — A few comments/tests still describe states this PR removes: `formatSend` documents and tests a `· pasted · keys: Enter` result even though text+keys is now a hard error, and the fixture still says the later Enter is what `--submit`'s grace guarantees. (packages/kaval-tui/src/render.ts:136; packages/kaval-tui/src/render.test.ts:130; packages/kaval-tui/src/paste-debounce-tui.fixture.mjs:83)


**claude** — Agreed with all three findings and fixed them: restored the empty-send fail-fast guard, replaced the error-swallowing `|| true` with an exit-code branch across all three SKILL copies, and scrubbed the stale text+keys / `--submit` doc-and-test remnants.

- `F1` **fixed** — Real fail-fast regression, confirmed against the old code: `cac4b96:main.ts` had a post-read `if (text === "" && keyData === "")` → `fail("nothing to send")` guard, and the `resolveSendInput` refactor moved legality pre-read (it validates the flag *combination*, not the *content*), so `--file empty` / empty pipe / `send <id> ""` fell through to a silent 0-byte "sent". Fixed in two places: (1) `main.ts:536` `cmdSend` now rejects `flags.input.kind !== "none" && text.length === 0` with a source-named error (`--file "…" is empty` / `the piped stdin is empty` / `the text is empty`); (2) `send.ts:258` `planSend` throws on an empty text arm so a direct caller can't mint a no-op plan (per codex's "ideally") — the emptiness invariant now lives at the plan boundary, not just the CLI. Added a `planSend` unit test for the empty-text throw (`send.test.ts`); 125 kaval-tui tests green.
- `F2` **fixed** — Agreed: `|| true` after `wait --timeout` swallows exit 3 (terminal gone) and 1 (link/usage), not just the exit-2 timeout it means to tolerate, and this skill drives real agents — teaching error-swallowing in the exact path the PR hardens. Replaced with codex's branch (`if ! kaval-tui wait …; then rc=$?; [ "$rc" -eq 2 ] || exit "$rc"; fi`) plus a comment naming the codes, in the `.apm` source (`agents/.apm/skills/kolu/SKILL.md:102`) and both hand-synced copies (`.claude/skills/kolu/SKILL.md`, `.agents/skills/kolu/SKILL.md`) — all three were byte-identical.
- `F3` **fixed** — All three stale remnants scrubbed. `render.ts:136` docstring example rewritten to two legal single-mark lines (a text send `· pasted`, a keys-only send `· keys: Enter`) and states the marks never co-occur. `render.test.ts:130` "shows … the marks that applied" test changed from the now-illegal `paste:true, keys:["Enter"]` combo to a legal paste-only text send (`keys:[]` → `· pasted`), keeping paste-mark coverage without asserting an unreachable state. `paste-debounce-tui.fixture.mjs:83` comment changed from "what `--submit`'s grace guarantees" to "the separate, post-settle Enter (sent as its own command after `wait --until idle`) guarantees". `formatSend` itself is unchanged — it stays a total formatter; only its docs/tests stop documenting the impossible combined state.

commit: `22fb4ccc506b4063c23a6d809f7ae02a55f649a3`


### Round 2

**codex** — approved: `false`

F1 and F3 are resolved in the current tree. The empty-payload path now fails after reading the source and `planSend` rejects empty text plans; the stale formatter/fixture comments and tests now describe only legal send states. F2 is still open because the replacement shell snippet captures the inverted status from `!`, so it still swallows real failures instead of propagating them.

Findings:
- `F1` · major · resolved — Resolved. `cmdSend` now rejects non-key sends whose resolved text is empty, with source-specific messages, and `planSend` rejects empty text content so direct callers cannot create a silent 0-byte text send. (packages/kaval-tui/src/main.ts:542)
- `F2` · major · open — The `|| true` was replaced, but the new `if ! kaval-tui wait ...; then rc=$?; ...` still loses the original failure code. In shell, `$?` inside that branch is the status of the inverted `!` pipeline, so it is `0` for any failing `wait`. The snippet then exits 0 for terminal-gone/link/usage failures instead of surfacing them. The same byte-identical issue is present in `.agents/skills/kolu/SKILL.md` and `.claude/skills/kolu/SKILL.md`. (agents/.apm/skills/kolu/SKILL.md:106)
- `F3` · nit · resolved — Resolved. The formatter docs/tests now use legal paste-only and keys-only examples, and the fixture comment now describes the separate post-settle Enter instead of `--submit`. (packages/kaval-tui/src/render.ts:136)

_codex on the rebuttal:_ F1 accepted: the post-read guard plus `planSend` empty-text rejection closes the zero-byte success path I raised. F2 is not resolved: the replacement still swallows errors because `if ! cmd; then rc=$?` captures `0`, not the command's original nonzero status. I verified the shell behavior directly (`if ! false; then echo $?; fi` prints `0`). F3 accepted: the stale impossible text+keys examples and `--submit` wording are gone.


**claude** — Conceded F2: rewrote the busy-agent snippet in all three SKILL copies to a non-inverted `if wait; then :; else rc=$?; …` so the else branch reads the real `wait` exit code; F1/F3 already resolved last round.

- `F1` **fixed** (already resolved round 1) — codex marked it resolved in verdict-2; no further change. The post-read empty guard in `cmdSend` (main.ts:542) plus `planSend`'s empty-text throw (send.ts) stand.
- `F2` **fixed** — codex is correct and I was wrong last round. In `if ! kaval-tui wait …; then rc=$?`, the `!` inverts the status: when `wait` fails, `! wait` succeeds (0), the then-branch runs, and `$?` reads that inverted `0` — so terminal-gone (3) and link/usage (1) failures were swallowed and Enter was sent into the void. Replaced with a non-inverted condition (`if kaval-tui wait …; then :  else rc=$?; [ "$rc" -eq 2 ] || exit "$rc"; fi`) where `$?` in the else branch is the real `wait` exit code. Fixed in the `.apm` source (agents/.apm/skills/kolu/SKILL.md:106) and both hand-synced copies (.agents/skills/kolu/SKILL.md:106, .claude/skills/kolu/SKILL.md:106) — all three were byte-identical. Verified the semantics directly with a stub: exit 0/2 proceed to the Enter send; exit 3 → outer exit 3, exit 1 → outer exit 1 (both surface before the send). No TS changed, so the lint/typecheck gate has no coverage over these markdown edits; the shell behavior was the load-bearing part and it is verified.
- `F3` **fixed** (already resolved round 1) — codex marked it resolved in verdict-2; no further change.

commit: `44281eb067b42225a4c8ef2af9b8e820a6f7955f`


### Round 3

**codex** — approved: `true`

The non-inverted condition now preserves the real wait exit code in the else branch, and the three skill copies are byte-identical. I'm ready to close the review.

Findings:
- `F1` · major · resolved — Resolved. The post-read empty-payload guard and `planSend` empty-text rejection remain in place. (packages/kaval-tui/src/main.ts:542)
- `F2` · major · resolved — Resolved. The snippet now uses `if kaval-tui wait ...; then :; else rc=$?; ...; fi`, so the else branch reads the real `wait` status and only treats exit 2 as the expected busy timeout. The `.agents` and `.claude` copies are byte-identical to the `.apm` source. (agents/.apm/skills/kolu/SKILL.md:106)
- `F3` · nit · resolved — Resolved. The stale text+keys and `--submit` documentation/test remnants remain removed. (packages/kaval-tui/src/render.ts:136)

_codex on the rebuttal:_ F2 accepted: switching from `if ! wait` to `if wait; then :; else rc=$?; ...; fi` fixes the status-capture bug I held open. In the current snippet, exit 0 proceeds, exit 2 proceeds as the documented busy case, and other failures propagate before the Enter send. F1 and F3 were already resolved and remain resolved.

srid:

**Consensus** — 1 round · base `cac4b96`. lowy (volatility/boundaries) and hickey (structural simplicity) reviewed the diff independently, then agreed per-finding.

**Verdict:** the boundaries encapsulate real axes of change — no invented boundary, no over-extraction. `executeSendPlan`'s injected `write` sink is a clean transport seam (a package-local leaf, correctly NOT a `@kolu/*` package); the `SendInput` union is a genuine receptacle whose discriminant carries the `--file` path.

**Applied fixes:**

| Finding | Fix | Commit |
| --- | --- | --- |
| `sourceIsStream` escaped the union's exhaustiveness (a `\|\|`-list a new source could silently escape to the risky line-by-line default) | rewrote as an exhaustive `switch` with no default — a new `SendInput` variant is now compiler-forced to declare its stream-ness, matching `readSendText` | `d0b033ce9` |
| the three-step submit incantation was written out at three teaching sites | single-sourced into `SUBMIT_FLOW_HELP` | `3e2798811` |
| the text-XOR-keys invariant lived only as a comment | made it a **type** — `planSend` takes a discriminated `{kind}` content input and `SendPlan` carries one `write`, so the `[text, keys]` pair is unspellable | `bebd67aff` |

Recorded electricity-test grounding (lowy-1) was a drop (no action). The subsequent `/simplify` altitude pass extended this same thesis with an exhaustive `sourceLabel` (`54c927965`).

srid:

**All clear** — base `cac4b96` (`--no-elegance`, since `/simplify` already ran this gauntlet).

| Pass | Issues | Notes |
| --- | --- | --- |
| Rules | 0 | Every built-in + project rule checked with an explicit `Checked by:` enumeration; SolidJS/server/schema rules correctly N/A (kaval-tui is a standalone Node CLI). |
| Fact-check | 0 | No silent error swallowing, inaccurate fallbacks, wishful thinking, logic errors, or unbounded growth. |
| Elegance | – | Skipped (`--no-elegance`). |

Two near-misses spot-checked and resolved as **not** violations:

- **`prefer-ts-pattern`** — the bare `switch` on the `SendInput` discriminated union (`sourceIsStream` / `sourceLabel`) is fine: `ts-pattern` isn't a `kaval-tui` dependency (unlike `client`/`server`/`common`), the pre-existing `wait.ts` uses the same convention, and the no-`default` switch already gives compile-time exhaustiveness on new variants.
- **`caught-error-must-not-collapse-to-empty`** — `stdinIsPayload()`'s bare `catch { return false }` around `fstatSync(0)` is a documented fail-safe: a `false` result funnels into the loud `"nothing to send"` error, never a silent empty success.

The diff was reviewed as unusually rigorous — illegal input combinations centralized in one pure `resolveSendInput`, error paths exhaustively tested, and the `--submit` removal left no dead code (grep-confirmed). No fixes needed.

srid:

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

srid:
## Evidence

This is a CLI change (no kolu-app UI surface), so the evidence is the terminal behavior. The headline is that every footgun this PR removes is now an **unspellable, loud, teaching error** — you cannot express the trap, and a migrating user is guided to the fix instead of silently misfired:

```console
$ kaval-tui send t1 "fix the parser" --key Enter        # text + --key in ONE send
kaval-tui: text and --key can't be combined in one send — a same-breath Enter is raced by the TUI's paste debounce and silently dropped. Send the text, watch the TUI settle, then submit as its own command:
  kaval-tui send <id> --file brief.md      # 1. the text
  kaval-tui wait <id> --until idle:300     # 2. observe the TUI settle
  kaval-tui send <id> --key Enter          # 3. submit
  (exit: non-zero)

$ kaval-tui send t1 "hi" --submit                        # the removed flag, typed from muscle memory
kaval-tui: --submit was removed — submitting a prompt is now its OWN command, because a same-breath Enter races the TUI's paste debounce and is silently dropped. Send the text, watch the TUI settle, then submit separately:
  kaval-tui send <id> --file brief.md      # 1. the text
  kaval-tui wait <id> --until idle:300     # 2. observe the TUI settle
  kaval-tui send <id> --key Enter          # 3. submit
  (exit: non-zero — no longer a silent no-op)

$ kaval-tui send t1 "inline" --file brief.md             # two text sources at once
kaval-tui: positional text and --file "/tmp/brief.md" each provide the send text — pass exactly one source, not several.
  (exit: non-zero)

$ kaval-tui send t1 --pate Enter                         # a typo (was silently ignored before)
kaval-tui: unknown flag for send: --pate. send accepts text (positional) or --file <path>, plus --key, --paste / --no-paste, --json, and --socket / --host.
  (exit: non-zero)

$ kaval-tui send t1                                       # nothing to send
kaval-tui: nothing to send — pass text, use --file <path>, pipe it on stdin, or use --key (e.g. `kaval-tui send <id> "hello"` or `kaval-tui send <id> --key Escape`).
  (exit: non-zero)
```

Each exits non-zero **before dialing** the daemon (pure `resolveSendInput` / `rejectUnknownSendFlags`). Note the last two are the newly-honest cases: `--submit` (just removed) prints a migration message instead of silently sending the text unsubmitted, and a flag typo (`--pate`) is rejected rather than ignored.

The other two pieces are proven by tests rather than a live transcript (their effect is a byte stream / a timing bound, not screen output):

- **Write deadline (Bug C).** `sendExec.test.ts` drives the shipped `executeSendPlan` with a never-draining sink and asserts it fails loud *within* the deadline (fake timers), naming the stalled terminal — not a hang. This was independently confirmed against a **live claude**: a follow-up `--key Enter` against a staged large paste hung the write >30s; the deadline turns that into a bounded loud failure.
- **`--file` byte-exactness.** `sendExec.test.ts` writes a payload containing `` `identity` ``, `$(whoami)`, and `${HOME}` to a file, reads it via the `--file` path, and asserts every byte reaches the wire verbatim (the exact class `"$(cat file)"` would have shell-executed).

Full functional coverage: **127 kaval-tui unit tests** (incl. the rewritten real-paste-debounce acceptance suite driving the shipped executor) + **two-platform CI green** (x86_64-linux & aarch64-darwin).

PR #1705, opened 2026-07-05
refactor: simplify the @kolu/surface hosting side (S1–S10 + system.identity)

srid:
> **Paired PRs** — all three must be two-platform CI-green at the same pinned kolu HEAD before any merge:
>
> - **kolu** — this PR (`08c2c29c4`, on top of #1703)
> - **drishti** — srid/drishti#88
> - **odu** — juspay/odu#39

The `@kolu/surface` hosting side had three names for one idea (a role, an appliance, a plug-with-extra-pins), a module-global session pool shadowing every consumer's own map, and six bespoke identity readouts. This collapses it to a smaller, clearer API — a behavior-preserving reshape **plus** one deliberate, ratified contract addition (`system.identity`). It lands as **three paired PRs** — kolu · [drishti](https://github.com/srid/drishti/pull/88) · [odu](https://github.com/juspay/odu/pull/39) — all green before any merge.

Design record: `debates/surface-hosting-simplify/conclusion.md` (S1–S10) + `docs/atlas/src/content/atlas/surface-hosting-roadblocks.mdx` (the four build-time rulings). Draft; do not merge until all three are two-platform green at pinned heads.

## What changed, by ledger item

| item | change | proof |
|---|---|---|
| **S9** | `makeSession({ connectOnce, admit? })` — the reconnect/backoff/give-up/liveness loop extracted ONCE as electricity; transport is a `connectOnce` plug (`sshConnector` / kolu's local `endpointConnector`), supervision a typed `admit` closed-verdict hook. No wrapper classes; sessions are closures + spread. | `session.test`/`recheck`/`liveness`/`reconnect-spin` (116 green) |
| **S3** | `RemoteMirrorSession`→`Session`, `DaemonMirrorSession`→`DaemonSession`; the dead contract `<C>` dropped at the role boundary (client-parameterized, covariant). | `session.variance.test-d` |
| **S4** | Identity graduates as ONE universal value: a reserved `system.identity` member on every surface (the twin of `system.live`), framework-stamped; `identity()` on the base `Session` role. `SurfaceIdentity` is a null-free sum (`disconnected`\|`anonymous`\|`identified`), `commit` a `BuildCommit` sum. padi declares its baked identity; the six bespoke readouts collapse to `identity()`. | surface (257) + server |
| **S1/S2** | `buildHostRegistry<S extends DestroyableSession, H>` + typed `controls` — fleet verbs exist only when supplied (union return, illegal call won't compile). | `hostFanout.test` |
| **S7** | `renew()` + `preservation` fold into `DaemonSession`; `PreservationStrategy` carries `children:"survive"\|"die"`. `BoundPadi` deleted. | padi-arm tests |
| **S10** | `getHostSession` + the module pool + `evictHostSession` + `destroyAllSessions` DELETED; consumers compose `makeSession` and own their sessions. | pool test deleted; consumers migrate |
| **RB6** | The `""` build-id sentinel is gone: `ConvergenceIdentity.build` is a null-free `DaemonBuild` sum, on a wire kept DISTINCT from the readout's `system.identity` (convergence must judge a skewed daemon pre-handshake — a justified separation, reported, not merged). | convergence kit tests (26) |

## Consumer migrations (this PR: kolu-server + the two examples)
kolu-server's single padi arm: both arms become `makeSession` + daemon-member spread — remote carries `admit: padiAdmit` (control-core hello + `decide` + drain), local a self-converging `endpointConnector` (the supervisor Endpoint untouched). `mini-ci` + `remote-process-monitor` examples migrated off the deleted API.

## Scope / notes
- The `system.identity` member is a deliberate contract change on every surface; kolu's member-set tests exclude reserved members by convention, so no churn here (drishti/odu likewise).
- The `@kolu/surface` serve path is now in scope (it owns the reserved member) — a fourth package beyond the ledger's original three, ratified.
- OUT of scope: kolu-server GROWING a `buildHostRegistry` pool (the W4 switch — its own later PR).


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `cac4b9659af3`

### Applied (4)
- `lowy-3` Stale index comment references the deleted `./hostSession` module — commit `cda49727c`
- `hickey-1` Drain-and-confirm-exit is implemented twice (concept multiplication) — commit `e11717219`
- `hickey-2` padiSession.ts header claims a sharing that the code does not have — commit `ea12ff1b4`
- `hickey-3` surfaceLiveProbe exported but its one intended call site inlines an identical twin — commit `232fbc5fc`

### Agreed — no change (4)
- `lowy-1` Admit receptacle's client arg is dead; remote supervision reaches past it to a mutable `combined` stash (packages/server/src/remotePadiBinding.ts:299)
- `lowy-2` Local padi arm's convergence() is hardcoded null — asymmetric with the remote arm's full anomaly surfacing (packages/server/src/padiBinding.ts:415)
- `hickey-4` SessionState stores every remote progress line twice, kept in sync by convention (packages/surface-nix-host/src/session.ts:73)
- `hickey-5` combined client stashed in a mutable closure var, hand-synced across four sites (packages/server/src/remotePadiBinding.ts:213)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `cac4b9659af3`
### Round 1

**codex** — approved: `false`

The refactor is broadly coherent, but I found two correctness issues in the new session state machine and one stale public-docs issue. The main risk is that identity is now cached asynchronously without any notification path, so server state derived from it can stay wrong; the second is a race where the local connector can emit reconnect state before the new client promise is visible to reconnect waiters.

Findings:
- `F1` · major · open — `markConnected()` publishes `connection: "connected"` before the async identity probe has resolved, and `pollIdentity()` later mutates `cachedIdentity` without emitting any state/event. Server code derives `processStartedAt`, Padi surface version, and build commit from `padiSession.identity()` inside `padiSession.onState` (`packages/server/src/index.ts:648`), so the connected event can sample `disconnected`, publish `padi: null`, and then never resample when identity becomes identified. That can leave the Padi uptime/version readout stale until some unrelated session transition occurs. (packages/surface-nix-host/src/session.ts:643)
- `F2` · blocking · open — `launchAttempt()` assigns `clientPromise = attempt()`, but `attempt()` starts running before the assignment completes. The local Padi connector calls `ctx.connecting()` synchronously before its first `await` (`packages/server/src/padiBinding.ts:373`), so reconnect state can be emitted while `currentClient()` still sees no new promise. A `waitForNextClient()` cursor parked after a dropped link can wake on those early `copying`/`connecting` frames, find no client, and then receive no further wakeup after `clientPromise` is finally assigned, leaving the pump stuck until watchdog retries. (packages/surface-nix-host/src/session.ts:596)
- `F3` · minor · open — The branch deletes the `HostSession`/`getHostSession` public API, but the package README still imports and documents `getHostSession`, `HostSession`, and `evictHostSession`; the root architecture README also still describes remote Padi in terms of `getHostSession`/`RemoteMirrorSession` (`README.md:207`). Consumers following the docs will now copy code that does not compile, and the repo instructions require updating README architecture for structural changes. (packages/surface-nix-host/README.md:16)


**claude** — Fixed the async-identity notification gap (F1) with a regression test, disputed the reconnect-race (F2) as refuted by the state channel's async delivery, and refreshed the stale public docs (F3).

- `F1` **fixed** — Real regression, confirmed against the base. Pre-S9, `PadiBindingSession.onEndpointStatus("connected")` set `boundIdentity` (startedAt · surfaceVersion · buildCommit) SYNCHRONOUSLY before `this.fire()`, so every `onState` listener saw the identity on the connected frame. The new `makeSession` polls `system.identity` on a SEPARATE async RPC from `markConnected` and mutated `cachedIdentity` with no state emission — so the `connected` frame (delivered before the probe resolves; `updateState({connected})` is scheduled ahead of the `p.then(pollIdentity)` microtask, so this ordering is deterministic) samples `identity() === disconnected`, and kolu-server's `padiSession.onState` handler (`packages/server/src/index.ts:648` → `padiStartedAt()`/`padiSurfaceVersion()`/`padiBuildCommit()`) publishes `padi: null` and never resamples the landed value until an unrelated transition. Fix: in `pollIdentity`'s success branch (`packages/surface-nix-host/src/session.ts:620`) republish the state cell (`stateCell.set({ ...stateCell.current() })`, no field change) after setting `cachedIdentity`, waking `onState` consumers to resample — restoring the pre-S9 parity across the async-probe boundary. Added `packages/surface-nix-host/src/identityNotify.test.ts` (verified it FAILS without the fix: `sampled` stays `[…disconnected]`, never `anonymous`; PASSES with it), and updated `liveness.test.ts:197` to count genuine connecting→connected TRANSITIONS rather than raw connected frames (its idempotence intent is preserved; the new identity-republish frame re-carries `connected` and would otherwise over-count).

- `F2` **disputed** — The premise (a parked cursor wakes on the synchronous `copying`/`connecting` frames while `clientPromise` is still null, then never re-wakes) is false: session state deltas are NOT delivered synchronously. `onState` is `stateCell.consume`, whose delta half runs through `buildConsume`'s `for await (… of subscribe())` loop (`packages/surface/src/server.ts:966`) — every delivery is a resolved-promise continuation, i.e. a microtask. During `launchAttempt`'s `clientPromise = attempt()` (`packages/surface-nix-host/src/session.ts:596`), the `copying`/`connecting` `updateState` calls in `attempt()`'s synchronous prologue only ENQUEUE deliveries; they cannot run until the synchronous stack (timer callback → `launchAttempt` → `attempt()` prologue) unwinds, by which point `clientPromise` is fully assigned. So when the cursor's `tryResolve` (`waitForNextClient.ts:84`) actually runs on the delivered frame, `session.currentClient()` returns the freshly-assigned promise, it differs from `previous`, and it resolves — no deadlock, no watchdog dependency. (Only the snapshot fire on subscribe is synchronous, and that happens at cursor-registration time carrying the pre-launch state, not during a launch.) The "store the promise before any attempt can emit state" invariant codex asks for is unnecessary given microtask-scheduled delivery. Left the code as-is.

- `F3` **fixed** — Stale public docs for the deleted `HostSession`/`getHostSession`/`evictHostSession`/`destroyAllSessions` API. Updated `packages/surface-nix-host/README.md`: the headline "What it gives you" example now `makeSession({ connectOnce: sshConnector(...) })`; the pieces table replaces the four deleted-export rows with `makeSession` / `sshConnector` / `Connector`·`Connection`·`ConnectContext`·`ConnectError` / `Admit`·`AdmitVerdict` rows; `pumpRemoteSurface` and `isLocalHost`/`forEachLine` prose re-pointed to `makeSession`/`sshConnector`; the `drvPath` recipe and the "fan out over N hosts" note migrated to the connector shape; the `pin()`/`acquire()` invariant collapsed to `pin()` (the public `acquire()` no longer exists). Updated the root `README.md` architecture table (lines 207, 209): `getHostSession`→`makeSession({ connectOnce: sshConnector(...) [, admit: padiAdmit] })`, `RemoteMirrorSession`/`PadiBindingSession` seam → the shared `Session` seam both padi arms use. (`scheduleReconnect` at README line 136 kept — it is still the internal reconnect scheduler inside `makeSession`.)

commit: `003c54960e38bd22b503e7a890170244a0ca7220`


### Round 2

**codex** — approved: `true`

All three prior findings are resolved. I verified the F1 code path and regression test, accepted Claude's F2 dispute after checking the actual `inMemoryCell`/`inMemoryChannel` delivery semantics, and confirmed the stale public docs covered by F3 were refreshed. No new regression was introduced by this round's changes that I would raise under the requested scope.

Findings:
- `F1` · major · resolved — Resolved. The async identity probe now updates `cachedIdentity` and republishes the current state, so `onState` consumers that derive published state from `identity()` resample when the identity RPC lands instead of staying on the pre-probe `disconnected` value. (packages/surface-nix-host/src/session.ts:625)
- `F2` · blocking · resolved — Resolved by accepted dispute. My original race premise depended on delta listeners running synchronously during `attempt()`'s synchronous prologue. In the current implementation, `inMemoryCell.consume` fires only the initial snapshot synchronously; later `set()` calls publish through `deltas.consume`, whose `buildConsume` loop invokes `onEvent` after the async iterator promise resolves. Even when `sub.push()` resolves a parked waiter inline, the `for await` continuation runs after the current stack unwinds, by which point `clientPromise = attempt()` has completed assignment. (packages/surface-nix-host/src/session.ts:596)
- `F3` · minor · resolved — Resolved. The public README now demonstrates `makeSession({ connectOnce: sshConnector(...) })` and documents the new `makeSession`/connector/admit/session shape instead of the deleted `HostSession`/`getHostSession` pool API. The root architecture table was also updated to describe the shared `Session` seam. (packages/surface-nix-host/README.md:16)

_codex on the rebuttal:_ F1: accepted. The fix directly restores the missing notification edge: after `probeSurfaceIdentity` resolves, the session republishes the state cell so server consumers resample `identity()`. The regression test exercises the same consumer pattern I called out.

F2: conceded. I re-read `inMemoryCell.consume`, `inMemoryChannel`, and `buildConsume`; Claude's async-delivery argument is correct for deltas. The synchronous snapshot happens only at subscription time. The `copying`/`connecting` frames emitted during `attempt()` are delivered through the async iterator path, so a parked cursor cannot observe them before `clientPromise` is assigned.

F3: accepted. The stale public package README and root architecture references covered by the finding have been replaced with the new session/connector API.

srid:

**Scope:** `cac4b9659…HEAD` (the @kolu/surface hosting-side refactor). Pass 3 (elegance) skipped — `/simplify` already ran this gauntlet.

**4 findings, all actioned.**

### Pass 2 — Fact-check

| Finding | Verdict | Fix | Commit |
|---|---|---|---|
| The `admit` branch of `attempt()` returns before the connect watchdog is armed, so a **wedged daemon** (process up but its first control-core `hello()` never settles) hangs `pin()` forever and leaks the connection — on the ssh/remote-padi arm, the transport daemon sessions actually use. | **CONFIRMED** | Bound the admit await by the same `connectTimeoutMs` (`withHandshakeTimeout`) — the admit-path twin of the non-admit watchdog; on timeout it rejects `"network"` and the existing catch tears down + reconnects. New `admitTimeout.test.ts` locks it. | `af8267345` |
| `pollIdentity` writes `cachedIdentity` with no generation guard, so a slow identity probe from a dead link could clobber a newer dial's identity if it resolved late rather than rejecting. | **PLAUSIBLE** | Monotonic `dialEpoch` captured per probe; the write is dropped once the dial is superseded. | `af8267345` |

### Pass 1 — Rule checklist

| Rule | Fix | Commit |
|---|---|---|
| `no-silent-error-swallowing` — the `markConnected` identity-repoll catch lacked a justification comment (its siblings have one). | Documented why the drop is safe (the promise already resolved; `pollIdentity` owns its own probe-failure catch). | `af8267345` |
| `comment-the-non-obvious` — several doc comments still named deleted/renamed symbols (`PadiBindingSession.drainBoundPadi`, `BoundPadi`/`HostSession`, `getHostSession`, `padiConvergence()`). | Refreshed across `padiConvergence.ts`, `index.ts`, `arch.ts`, `daemonInventory.ts`, `useDaemonInventory.ts` to the current API. | `c0c899a80` |

All 20 other project rules + the DRY / invalid-states / dead-code / unbounded-growth checks passed. surface-nix-host 119 + server 160 green after the fixes.

Issue #1204, opened 2026-06-06
[Log] pu issues


srid:
Logging any issues encountered when using https://github.com/juspay/project-unknown

srid:
## `pu fork` lands cross-gateway → forked box unreachable

While building the leased-pool CI model (kolu-ci-1..8), hit the known `pu fork` placement bug again — flagging it here for a permanent fix.

**Repro / observation:**
- Source `kolu-ci-golden` is on physical host `dev-x86-64-linux-05`.
- `pu fork kolu-ci-golden kolu-ci-1` placed the fork on a **different** gateway/host: `idliv2-02`.
- `pu fork` does **not** write `~/.pu-state/<name>/ssh_config`, so raw `ssh <name>` (what justci uses) can't resolve it. Synthesizing the config from the source's by name-substitution only works when the fork lands on the **same** gateway — here it didn't, so the synthesized `ProxyCommand` pointed at golden's gateway IP and the box was unreachable: `Connection closed by UNKNOWN port 65535`.

**Impact:** fork is unusable as a reliable provisioning primitive for CI — placement is non-deterministic and cross-gateway forks are unreachable without a correct ssh_config. We worked around it by **not forking**: the pool is seeded with `pu create` (which writes a correct ssh_config) and boxes are leased while warm. But two upstream fixes would make fork dependable:
1. `pu fork` should write `~/.pu-state/<name>/ssh_config` for the box's **actual** placement (like `pu create` does).
2. Either keep a fork on the source's storage pool/gateway (fast CoW) or, if it must land elsewhere, ensure the generated config targets the new gateway.

**Secondary:** `pu list` takes ~**34s** to return (single call, through the proxy). It's slow enough that any script must avoid calling it in a loop. A faster list (or a `--no-detail`/cached mode) would help.

*Note for correctness: `pu fork`/`pu create` correctly exit non-zero on `ERR` (e.g. "name already exists") — no exit-code bug, contrary to a first impression on my end.*

Context: kolu CI pool work, branch `incus-ci-pool`.

srid:
## ⚠️ `pu` pool egress — transient, pool-wide, self-recovered (Incus admin)

[...]

PR #1708, opened 2026-07-06
feat(padi): W4 — the switch (warm host pool + live per-tab host switching)

srid:
kolu keeps warm connections to several machines and **each browser tab picks — and live-switches, no reload — which machine it's looking at.** Your laptop tab can sit on `zest` while your phone sits on `sincereintent`, against the same kolu, neither disturbing the other. This is **W4 "the switch"** (padi plan of record) — ONE kolu PR + a paired drishti PR (the framework half).

Implements the plan of record's three layers, top to bottom.

## Framework — `@kolu/surface-app` (drishti-paired)
`SurfaceAppProvider`'s `controlPlane` becomes `Accessor<ControlPlane<T>>` (the ratified S5 shape). The build-identity subscription is re-derived **per control-plane identity** (a `createMemo` + `createRoot`-per-identity), so a live host swap tears down the old host's stream and re-subscribes — a **framework guarantee**, pinned by a new provider test ("swap tears down + re-subscribes", where the gray-chip #1687 class goes to die). Static callers pass `() => client`. **drishti is the proving second consumer** (`.claude/rules/surface.md` gate); its migration is a one-character edit (`surfaceAppClient()` → `surfaceAppClient`, already accessor-shaped).

## Server — kolu (additive; the layer L6+L23+#1705 cleaned)
The warm pool: `hostPool.ts` holds a `buildHostRegistry` with one `PadiSession` (`DaemonSession<PadiSurfaceClient>`) per machine — the **local endpoint arm** for the local host, the **ssh arm** otherwise — each re-served into a per-host router (the shared `kolu`+`surfaceApp` fragment spliced with THIS host's padi) and a per-host `WsRPCHandler`. **No `controls`** (kolu needs no fleet verbs). ws-upgrade dispatches by the connection's declared `?host`; an **unknown host is rejected loud** (close 1008). New root RPCs `hosts.add`/`hosts.remove` — the picker **adds before it connects** (deliberate intent, never a side-effectful GET that provisions ssh from a stray query param). `preferences.recentHosts` (schema + `1.32.0` migration) is the pool's persisted host set, shared across a user's devices. The preview route reads through the per-request host's session — **killing the `remoteHost` fork** (closing the #1685 deferral). `server.info.defaultHost` lets a tab pick its initial binding (`KOLU_PADI_HOST` boots CI through it).

## Client — kolu (the de-entangling)
The wire stops being one page-lifetime module-global bundle. `binding/bindings.ts` holds one `connectSurfaces` bundle per host; an `activeHost` signal picks the one the tab views; `switchHost` swaps it **in place — no reload**. `wire.ts` re-exports the sibling clients as **proxies** over the active binding (so ~35 consumers don't rename), and the app-lifetime singleton subs re-key via `bindingScoped` (tear down + re-subscribe on a switch — the client twin of the framework guarantee).

- **The misroute guard (in this PR, not L11):** each binding has a `retired` flag; switching retires + disposes the old one, so a stale in-flight call rejects loudly rather than landing on the wrong or dead host — and separate sockets make a cross-host route impossible by construction. Pinned by `bindings.test.ts`.
- **Per-host readiness** rides the padi `connection` cell (`mirroredSurface`, `liveWhen`), folding into `padi.health().live` per binding. `useDaemonStatus` derives readiness / the **create-refused-while-warming** gate / the loud degraded states off the **active** host. The single-host `kolu.padiLink` cell **retired** — it could never carry N bound hosts (the mirrored connection cell is its per-binding replacement).
- **The picker, hidden:** a nested group under the command palette's **Debug** section only — no ChromeBar, no keybind, no tip. Local + the default host + server-persisted recents + a free-typed ssh host. The palette item and `KOLU_PADI_HOST` are both undocumented — together the only way to reach remote terminals until the feature stabilizes.

## Done (the W4 list)
- ✅ two tabs on two hosts stay independent — separate page loads, per-tab active host in localStorage
- ✅ recents persist across devices — `preferences.recentHosts` in `config.json`
- ✅ create refused while warming — `refuseIfWarming()` reads the active host's readiness
- ✅ local single-host mode byte-identical — the default path is behaviorally unchanged (evidence + the full e2e suite)
- ▶ the live-switch e2e (local → remote → local; a stale in-flight call rejected; the switch never disturbing another device) — driven against a real pu second host (see Evidence)
- ▶ paired drishti PR CI-green at the final kolu HEAD

## Design-philosophy honored
- **Fail-fast, no fallbacks:** an unknown `?host` is rejected loud; a stale binding throws (no silent misroute); the re-serve pump failing is `process.exit(1)`; illegal states unrepresentable via the typed pool slot (`DestroyableSession`) + the misroute guard.
- **Electricity boundaries:** the swap-safe wire lifetime lands in its true receptacle — `@kolu/surface-app`'s accessor — with drishti as the proving second consumer, not a new app-in-a-package.
- **Reuse the source of truth:** the mirrored `connection` cell (not a hand-rolled per-host padiLink); `buildHostRegistry` (not a bespoke `Map`); `preferences` (not a parallel store).

## Notes
- **L11 ledger:** the client's module-global **accessor shape** is L11's declared starting point (the full scope-through-context sweep). The L11 ledger entry belongs in `padi-cleanup.mdx` (coordinator-owned, RT-fable).
- **Per-host diagnostics (A1, resolved):** padi uptime / build / convergence / RSS are now sourced **per host** from each pool entry's cached `session.identity()` / `convergence()` (`perHostKoluCells.ts`), not the boot default — an earlier note here called them an "acceptable-for-scope default-host diagnostic"; the adversarial review killed that. serverRss / server uptime stay kolu-server's own (host-independent).
- No changelog entry: the feature is deliberately hidden/undocumented until it stabilizes.

## Known gaps (ratified deferrals — for morning ratify/overrule)
- **Honest clocks (plan piece 5, E1) — deferred to a named fast-follow.** padi has served `control.core.clockNow` since W1, but NO binder consumes it: host-stamped timestamps (`lastActivityAt`, `sleptAt`, `parkedAt`) are formatted with the **viewer's** clock, so a cross-host "3m ago" is wrong under clock skew. Pre-existing (W3.1-era), not a W4 regression. The served half exists; the consumer half (wiring the RTT-halved offset through to the client) is the fast-follow.
- **Cross-host live-switch (F-d) — PROVEN (no longer deferred).** Run for real against a leased `kolu-ci-5` pu box: palette switch local → kolu-ci-5 → local with no reload, a live terminal on the remote (`toor@kolu-ci-5`), and two tabs holding different hosts at once — see the **## Evidence** comment. The drive also caught two real bugs now fixed on this branch (the picker's free-typed connect couldn't submit; a first-boot D1 schema-validation false error). A repeatable e2e scenario still wants pu headroom to run every PR (#1204), but the behaviour itself is demonstrated.

## Post-review live testing (on real hardware) — 3 bugs + a beta relabel

Running the feature on a Mac (zest) binding another Mac (sincereintent) surfaced what the unit suite couldn't:

- **Subscription leak — the switch toasts.** Every host switch fired three "server restarted — reload required (stale tab)" toasts (Memory / Padi memory / Host inventory), both directions — yet zest's server had **9h54m uptime and never restarted**. The toast was the visible edge of a real bug: **every switch leaked all three live subscriptions.** `useCell` wrapped its `createSubscription` in a detached `createRoot(() => …)` whose dispose was discarded, so a bindingScoped switch disposed the per-key owner but never the sub → `controller.abort()` never fired → the leaked sub errored on the retired socket and the (correctly guarded) catch saw `aborted === false` → toast. Fixed by tying the useCell root's dispose to the caller's owner + threading the abort signal into `source()` — the invariant "a disposed subscription cannot report anything" — and retireSocket now throws an honest "socket retired by this client" fact, not a "server restarted" claim (the genuine-restart affordance stays the lifecycle-status reload overlay). Pinned by `useCellDispose.test` (red before the fix).
- **Picker couldn't submit a new host** — palette value-mode Enter had no target for a hint-only value input, so "Connect to a host…" did nothing; since recents start empty, that free-typed connect is the *only* first path to a remote (unreachable from a clean install). Fixed at altitude (a value input may offer no options; Enter submits `selected: undefined`).
- **First-boot D1 schema false-error** — the startup validation hand-built its parse object and omitted the new top-level `recentHosts` key, logging a scary "Delete your state file" for a store that was fine. Fixed by reading every schema key.

**Labs relabel.** The picker moved from the "Debug" palette group to a new **"Labs"** group (flask icon) — the truthful home for a works-but-unstable beta feature — with the entry carrying a known-issues link (`kolu.dev/atlas/remote-bind-parity.html`). Same hidden posture (palette-only, undocumented); just an honest "beta, unsupported" signal Debug never conveyed.

## Adversarial review (coordinator's 4-lens batch) — all 13 findings actioned
B0 (compile-error backstop for subs opened outside `bindingScoped`) · B1/B2/B3 (stale-binding pins: the Code-tab pulse + preview + terminalExit re-key per host) · A1 (per-host padi diagnostics) · A2 (`server.info` labels per-host) · A3 (default host un-forgettable — un-brick) · C1 (switchHost epoch: last-pick-wins + cancel-via-re-pick) · C2 (a guest host's pump fault is contained, not a global exit) · D1 (recents live cross-device via their own server-authority cell) · H1–H4 (createKeyedRoot extraction, per-host names, installUnknownHostFallback, loud preview fallback) · F-a/F-b/F-c (the done-list test-honesty debt pinned). Each fix keeps the suite green; see the `fix(review)` commits.


srid:
## ⚖️ Lowy ⇄ Hickey lens debate — ✅ consensus (1 round)

Base: `59ab99b` (merge-base). Both structural lenses reviewed independently, then debated every finding to consensus. 6 `fix` findings, all applied (each its own commit).

| # | Origin | Finding | Commit |
|---|--------|---------|--------|
| lowy-1 | lowy+hickey | Per-host `mirrors`/`routers` side-maps lived outside the registry's host lifecycle (leaked on `remove`) and only the default host was ever read → collapsed to single default-host `getMirror()`/`getRouter()`. | `dbf31ecf9` |
| lowy-2 | lowy+hickey | `Binding.id` was write-only dead state; the guard's real teeth are the `retired` flag + separate sockets → deleted the field, corrected the doc. | `8928118fd` |
| — | lens | Assemble the `Binding` once — derive its lifecycle first, drop the placeholder-then-repatch dance. | `5eeb782c9` |
| — | lens | Define the `LOCAL_HOST` wire sentinel once in `kolu-common/contract` (was redefined per module). | `e6063b771` |
| — | lens | Drop the dead `surfaceApp` proxy export in `wire.ts`. | `2944cf15e` |
| — | lens | Correct the stale "off koluSurface's `padiLink` cell" comments (the cell was retired). | `e4fd86a97` |

Deliberate, coordinator-ratified decisions (wire-clients-as-proxies = interim/L11, the retired `padiLink` cell, the intentional `hostPool` no-`controls` asymmetry, the hidden picker, samplers bound to the default host) were threaded as `rationale` and correctly not flagged.

srid:
## Codex ⇄ Claude debate — ✅ consensus (2 rounds, reasoning effort: xhigh)

Base: `59ab99b (merge-base)`. Reviewer: codex; author: claude. All 7 findings resolved.

### Round 1

**codex** — approved: `false`

The branch has the right broad shape for per-host bindings, but several host-scoped paths still pin the first binding or omit the active host. That means the main switch can show a new host while parts of the app still read or call the old/default host. I'm not approving until those cross-host correctness issues are fixed.

Findings:
- `F1` · blocking · open — The terminal metadata collection is still created once from `padi.collections.terminals.use(...)`. The terminal key list was moved to `bindingScoped`, but this metadata subscription captures the initial binding's `padi` client and only changes its keys. After a host switch, the app can receive terminal IDs from the new host while asking the previous host for their records, so the canvas/right panel can render missing or wrong terminal metadata. The same pattern also exists in host-scoped module-level subscriptions like `KavalUpdateBadge`'s `padi.cells.status.use(...)`. (packages/client/src/terminal/useTerminalMetadata.ts:88)
- `F2` · major · open — The server preview route now dispatches by `?host` and defaults to `pool.defaultHost` when it is absent, but the client still builds binary preview URLs without the active host. A tab switched to a non-default host will fetch `/api/terminals/.../file/...` from the default host, producing 404s for that terminal or, worse, serving a same-ID file from the wrong host. Markdown image URLs have the same problem at `packages/client/src/right-panel/markdownImageSrc.ts:26`. (packages/client/src/right-panel/BrowseFileDispatcher.tsx:153)
- `F3` · major · open — `makeBinding` calls `createServerLifecycle` under whatever Solid owner happens to create the binding, then `Binding.dispose()` only calls `conn.dispose()`. If the first caller is an app-lifetime owner, old binding lifecycle listeners and cleanup closures stay retained until page teardown after every host switch. The lifetime of a per-binding resource should not depend on the incidental caller that first read `activeBinding()`. (packages/client/src/binding/bindings.ts:91)
- `F4` · major · open — The explicit `assertLive` guard is not on any real app call path. `activePadiRpc()` is unused, while existing code still calls `padiRpc(padi)` directly and can capture concrete nested procedures such as `padiRpc(padi).surface.subscribeRepoChange.get` in `createRepoPolledQuery`. That means the retired flag is tested as a helper, but stale RPC references do not actually check it at invocation time. (packages/client/src/binding/bindings.ts:285)
- `F5` · minor · open — `server.info()` is still treated as cosmetic on failure, but it now also seeds `defaultHost`. If this one-shot request fails during startup, a fresh tab never falls through to `KOLU_PADI_HOST` and silently remains on local until reload or manual picker use. (packages/client/src/useServerIdentity.ts:27)
- `F6` · minor · open — The comments call active-host persistence "per-tab", but it uses `localStorage`, which is shared across tabs for the origin. Switching host in one tab changes what another tab will restore on reload, so the persisted behavior is not actually tab-scoped. (packages/client/src/binding/bindings.ts:171)
- `F7` · nit · open — The surface comments still list `padiLink` as part of `koluSurface` even though the cell was retired in this change. That makes the architecture prose contradict the actual surface shape a few lines later. (packages/common/src/surface.ts:443)


**claude** — Fixed all seven findings: re-keyed the pinned host-scoped subscriptions behind `bindingScoped`, threaded the active host into preview URLs, gave each binding's lifecycle its own root, made the misroute guard's socket-teardown real (and dropped the dead `activePadiRpc`), surfaced the default-host seed failure, moved active-host persistence to `sessionStorage`, and corrected the retired-`padiLink` comments.

- `F1` **fixed** — The terminal-metadata collection pinned the boot host's padi client. Moved it behind `bindingScoped` (`terminal/useTerminalMetadata.ts:88` → `bindingScoped((b) => b.clients.padi.collections.terminals.use({ keys, onError }))`; `getMetadata` now reads `terminals().byKey(id)`). Same fix for `KavalUpdateBadge`'s `padi.cells.status.use(...)` (`kaval/KavalUpdateBadge.tsx:28` → `createSharedRoot(() => bindingScoped((b) => b.clients.padi.cells.status.use(...)))`; `expectedKaval` reads `kavalStatus()().value()`). While auditing the class codex's "like…" flagged, I found four MORE pinned module-level subs and fixed them too: `ui/useMemoryUsage.ts`, `ui/useProcessUptime.ts`, `ui/useDaemonInventory.ts` (`app.cells.*` — kolu's host-independent surface, but the SOCKET is per-binding) and `ui/useHostInventory.ts` (`padi.cells.hostInventory` — genuinely per-host). These became load-bearing to fix once F4 retires the old socket (a pinned sub on the boot binding would read a now-closed socket after switching away). `subscribeExit` (`terminal/useTerminals.ts:62`) needs no change — it's list-driven via `mapArray`, so it re-mounts per host already. Updated `useTerminalMetadata.test.ts` to mock `../binding/bindings` (the collection now flows through the binding, not the `../wire` proxy). All 555 client tests green.

- `F2` **fixed** — The server preview route dispatches by `?host` (defaulting to `pool.defaultHost`) but the client built the URL without it, so a tab on a non-default host fetched files off the default host. `buildTerminalFileUrl` (`common/src/preview.ts:160`) now takes a REQUIRED `host` and appends `?host=<enc>`; the binary/img/video/iframe path passes `activeHost()` (`right-panel/BrowseFileDispatcher.tsx:153`, with `&v=<mtime>`), and markdown images thread it through `resolveMarkdownImageSrc(...ic, activeHost())` (`right-panel/markdownImageSrc.ts:26`). `pathFromPreviewPathname` strips the query (`split("?")`), so in-iframe navigation inversion is unaffected. Updated `markdownImageSrc.test.ts` + `server/src/iframePreviewRoute.test.ts`. One narrower hop is a documented follow-up, NOT this finding's scope: an `.html` preview's RELATIVE sub-resource (``) resolves against the iframe document URL and drops the `?host` query — fixing that needs a server-injected `<base>` or a host-in-path route, and it only bites remote-host HTML previews with relative assets (an experimental/hidden surface). The top-level img/video/iframe/markdown URLs codex named are all fixed.

- `F3` **fixed** — `makeBinding` (`binding/bindings.ts:91`) now creates the lifecycle inside its OWN `createRoot`, captures the root's disposer as `disposeLifecycle`, and `binding.dispose()` calls it. The lifecycle's lifetime is now exactly the binding's, decoupled from whatever incidental Solid owner first read `activeBinding()` — fixing BOTH the app-lifetime-owner leak codex named AND the symmetric premature-free case (a short-lived component owner would otherwise `onCleanup`-dispose a still-active binding's status). `dispose()` is idempotent via an early `if (binding.retired) return`.

- `F4` **fixed** — took codex's own offered resolution ("remove the unused helper and make the intended socket-disposal behavior explicit"). Removed the dead `activePadiRpc` (+ its now-unused `padiRpc` import). While verifying the guard I found the deeper defect: `binding.dispose()` → `conn.dispose()` → `transport.dispose()` only DETACHES listeners — it never closed the socket, so the module's "closing its socket" claim was false and a stale-captured deep proc could re-dial the still-open (partysocket-reconnecting) old host — a real misroute, and a socket leak. Added `retireSocket(conn.ws)` to `binding.dispose()` (`binding/bindings.ts:139`): it closes the socket and poisons `send` to throw, so a late call on a retired binding throws at the transport by construction — the real teeth, stronger than an invoke-time `assertLive` nobody called. Kept `assertLive` (ratified guard vocabulary, pinned by the misroute-guard unit test) and corrected the module comment to state the true teeth. Added a test assertion that a switch retires the old socket (`retireSocketMock` called with `local.ws`). NB: this is what made the F1 sweep of the other pinned subs mandatory.

- `F5` **fixed** — `useServerIdentity.ts:27`: the functional default-host (`KOLU_PADI_HOST`) seed rides the same one-shot `server.info()` as the cosmetic identity, and a failure silently left a remote-default tab on local. The `catch` now escalates from a cosmetic `console.warn` to a `console.error` naming the consequence (default host not applied; tab stays on local until reload/picker), rather than collapsing the functional decision silently. There is no independent channel for `defaultHost`, and the fetch is over the buffered live socket (a reject is a genuine server failure, not a not-yet-connected blip), so surfacing loudly — not crashing the boot on a cosmetic fetch — is the proportionate level for a minor.

- `F6` **fixed** — `binding/bindings.ts:171`: the active host is genuinely per-tab, so it now persists in `sessionStorage`, not `localStorage`. sessionStorage survives THIS tab's reload (the "reload lands back on the host it was viewing" intent) while staying scoped to the browsing context — switching host in one tab no longer changes what another tab restores, and a fresh tab starts clean (no stored host → `seedDefaultHost` falls to the server default, so `KOLU_PADI_HOST`/CI still works). Updated the comments (now accurate) and the unit-test stub (`bindings.test.ts` stubs `sessionStorage`).

- `F7` **fixed** — `common/src/surface.ts:443` + `:469`: both comment blocks that still listed `padiLink` as a live koluSurface cell now list `preferences` / `processMemory` / `processStartedAt` / `daemonInventory` and call out `padiLink` as RETIRED at W4 (replaced by the padi `connection` cell), matching the actual cell set and the retirement note a few lines down.

commit: `ef8eda6fad602968c5e613a3c3ff2e61dca5f785`


### Round 2

**codex** — approved: `true`

All seven prior findings are resolved in the current tree. I re-checked the changed areas tied to those findings and did not find a new regression introduced by Claude's fixes.

Findings:
- `F1` · blocking · resolved — Host-scoped subscriptions previously pinned the initial binding, so switching hosts could keep reading terminal metadata/status from the wrong socket. (packages/client/src/terminal/useTerminalMetadata.ts:98)
- `F2` · major · resolved — Preview URLs previously omitted the active host, allowing the server route to serve files from the default host while the UI viewed another host. (packages/common/src/preview.ts:165)
- `F3` · major · resolved — Binding lifecycle cleanup was previously tied to an incidental Solid owner instead of the binding lifetime. (packages/client/src/binding/bindings.ts:103)
- `F4` · major · resolved — The retired-binding guard previously had no real enforcement on stale captured clients because the old socket was not actually retired. (packages/client/src/binding/bindings.ts:139)
- `F5` · minor · resolved — `server.info()` failure previously treated the default-host seed failure as merely cosmetic, leaving a remote-default tab on local without a loud diagnostic. (packages/client/src/useServerIdentity.ts:27)
- `F6` · minor · resolved — Active host persistence was described as per-tab but used `localStorage`, which is shared across tabs. (packages/client/src/binding/bindings.ts:205)
- `F7` · nit · resolved — Comments still described retired `padiLink` surface cells as live. (packages/common/src/surface.ts:442)

_codex on the rebuttal:_ F1 accepted: the subscriptions I flagged now go through `bindingScoped`, and the extra sweep of same-class standing subscriptions is appropriate once old sockets are really retired. F2 accepted: the reported top-level preview URLs now carry `host`; the remaining relative-subresource behavior inside remote HTML previews is narrower than my finding. F3 accepted: the binding-owned lifecycle root fixes the owner leak and premature-cleanup cases. F4 accepted: closing/poisoning the retired socket gives stale captured clients real enforcement, and removing `activePadiRpc` removes the dead path. F5 accepted: a user-visible surface would be stronger, but the functional failure no longer collapses silently and the error names the consequence. F6 accepted: `sessionStorage` matches the stated per-tab semantics. F7 accepted: the stale `padiLink` comments covered by the finding are corrected.

srid:
## 👮 Code-police — 2 correctness bugs + 6 rule findings, all actioned

Two passes (rule-checklist + fact-check) over the change vs `59ab99b`. Applied in `5f253e520`.

**Pass 2 (fact-check) — real bugs, each now pinned by a test:**
- **TOCTOU in `hosts.add`** — two devices adding the same new host both passed the `has()` check and each dialed a live ssh session; the loser's session/pump was orphaned (never destroyed) and its fail-loud pump (`process.exit(1)`) could crash the whole server. Fixed: serialize per-host via an in-flight `Map` (2nd add joins the 1st). Pinned: `hostPool.test` "serializes CONCURRENT adds".
- **Removing a host another tab is viewing** → that tab re-dialed `?host=<gone>` forever (server rejects 1008) behind a misleading "Reconnecting…". Fixed: on a 1008 close the tab falls back to local. Pinned: `bindings.test` (1008 → fallback; 1006 → no fallback).

**Pass 1 (rule checklist):**
- `no-dead-code` — deleted the never-called `assertLive` (the guard's runtime teeth are the retired socket); doc + test updated.
- `no-unbounded-growth` — capped `recentHosts` at 20 (MRU): every recent host is re-warmed into a live session at boot, so it couldn't grow unbounded.
- `dry-rule-of-three` — extracted `useBindingScopedSub` for the `createSharedRoot(() => bindingScoped(...))` pattern repeated across 6 hooks.
- `no-silent-error-swallowing` / `caught-error-must-not-collapse-to-empty` — documented `readStoredHost`'s catch (a broken sessionStorage genuinely has no stored host → local is correct).
- `errors-must-log-at-error` — reconciled the unknown-host ws comment: it's a RECOVERABLE stale-tab case (client now falls to local), so `warn` is the right level.
- `watcher-lifecycle-logs` — log the per-host binding install/retire in a greppable form.

srid:
## Evidence

W4 "the switch" is proven leg-by-leg. Stated honestly — the parts were proven different ways, and one leg is deferred with the reason named.

### 1. CI — green on both platforms
`ci::smoke`, `ci::unit`, `ci::nix`, `ci::biome`, `ci::fmt`, `ci::flake-check`, `ci::install`, `ci::home-manager`, `ci::pnpm-hash-fresh`, `ci::atlas-sync`, `ci::surface-app-example-build`, `ci::surface-example-build` — **all pass on both `x86_64-linux` AND `aarch64-darwin`**, plus CodeQL / Analyze / SentinelOne. On the final HEAD `e889b8764`. (`ci::smoke` is the production-build app render + e2e; a green smoke on both platforms is the authoritative "the built client renders + works" gate.)

### 2. The live switch UI, driven with chrome-devtools (a dev kolu on random ports — never production)
I drove the **built client** live (chrome-devtools against a dev server on random ports; production `kolu.service` was provably untouched — same PID + uptime before/after). What this confirmed on the real running app:
- The client **renders** with the W4 IdentityRail (`Kolu · Padi · Kaval` health chips; the padi chip's host segment is the W4 add — `null` on local, the host name on remote).
- File-preview URLs carry the active host — the running app served `…/file/…?host=local&v=…` (the codex **F2** host-threading fix, visibly working).
- The warm pool **boots**: the server logged `hostPool: local binding installed { host: "local" }` (the police watcher-lifecycle log) and `padi-local … connection: connecting → connected`.

This drive also **caught a real robustness bug**: the first render crashed on `undefined.byKey`/`undefined.value` because `bindingScoped` set its value from a deferred effect (undefined until the effect phase). The production build's render order happened to dodge the window (CI smoke passed), but that's luck, not correctness — fixed by making `bindingScoped` populate synchronously (`createRenderEffect`), restoring the never-undefined invariant the pre-W4 direct subs had (commit `e889b8764`, pinned by a new test).

**Honestly not claimed:** the interactive picker→switch→back drive and full-canvas screenshots are **withheld**, and here's exactly why — the padi daemon socket namespace is machine-global, so a dev kolu on this box necessarily *adopts the live session's own padi daemon* rather than an isolated one. The canvas therefore shows real, in-flight internal agent terminals (not postable to a public PR), and keying into it drives live sessions. So the *interactive* switch drive against a clean isolated host is deferred with the cross-machine leg below; the switch **mechanics** are pinned by the tests in §4.

### 3. The remote arm (the ssh hop) — proven, and exactly how
W4 does **not** touch the remote-bind machinery: `remotePadiBinding.ts`, `padiBinding.ts`, `padiConvergence.ts`, `padiSession.ts` are **byte-identical** to master (`git diff` empty) — the warm pool just *invokes* them per host. So the shipped **W3.1 `just e2e-ssh` proof** (a real ssh hop through that machinery) still holds for the remote arm unchanged. What W4 *adds* around it — `?host` dispatch, `hosts.add`/`remove`, `recentHosts` — is pinned by `hostPool.test`.

The **full cross-machine chrome-devtools drive** (local → a real 2nd machine → local, creating a terminal on the remote) is **deferred pending pu capacity**: the Incus cluster was saturated all night ("No cluster member has capacity — limit 4 per member"; logged on #1204), and self-ssh was ruled out (it fakes the second machine — same clock, same fs). The remote leg is proven by the unchanged W3.1 e2e-ssh arm + the pool unit tests; only the *visual* cross-machine UI drive is deferred, and it is **not** claimed to have happened.

### 4. The mechanism, pinned by tests (run in normal CI, both platforms)
- **framework swap** — `surface-app/provider.test`: swapping the control plane tears down the old host's build-identity stream and re-subscribes (no leak).
- **the pool** — `hostPool.test`: `?host` dispatch (distinct handler per host, unknown→undefined), `recentHosts` persist, `hosts.add`/`remove` idempotent, **concurrent-add serialization** (the TOCTOU fix), local never added/removed.
- **the misroute guard + switch** — `bindings.test`: switching retires the old binding and closes its socket (a stale ref fails at the dead transport); a distinct live binding is built for the new host; **a removed host (close 1008) falls the tab back to local** (not a forever reconnect loop); `bindingScoped` populates synchronously.
- **per-host readiness** — `connectionToPadiLink` collapses the padi connection cell onto the display state.
- **local single-host mode byte-identical** — the full e2e suite passes unchanged.

### 5. Paired drishti PR
The framework half (`@kolu/surface-app` `controlPlane` accessor) is exercised by the paired **drishti PR #89**, re-pinned to this final kolu HEAD (`e889b8764`) and CI-green — the second real consumer proving the accessor shape.

srid:
# [W4] Adversarial-review batch — per-finding dispositions (all 13 + hygiene + records)

Worked in your prescribed order. Every fix kept the suite green; each has a test. kolu HEAD `06640a491`, drishti #89 re-pinned to it. Both undrafted, not merged.

## Cluster A — the per-host facts axis
- **A1 (headline) — FIXED** per your (B)-PLUS ruling. NEW `perHostKoluCells.ts`: a minimal per-entry `implementSurfaces` (processStartedAt + daemonInventory, cell specs derived from koluSurface, own in-memory channel — the reServeSurface isolation pattern), wired to each entry's cached `session.identity()`/`convergence()`; hostPool splices the two cells over `kolu` per host and disposes them on removal; index.ts deletes the defaultSession-pinned drivers. Client: padi/kaval RSS → `padi.cells.processMemory`, surfaceVersion → `padi.cells.version`; serverRss/uptime stay kolu's own. All three "acceptable-for-scope" comments DELETED. Tests: daemonInventory.test (disposer/leak corner), surface.test (cell set). `7c079b47d`.
- **A2 — FIXED.** `server.info` threads THIS router's host → labels per-host (`Kolu [srv → host]`; local byte-identical). Test: pwaIdentity.test. `67b601741`.
- **A3 (brick, confirmed) — FIXED.** `hosts.remove` guards `host===defaultHost` (mirrors local); picker forget-list filters the default; false comment fixed. Closes both the picker path and the raw RPC. Test: hostPool.test "never removes the DEFAULT host — even a REMOTE one". `240ae185f`.

## Cluster B — stale-binding pins (#1687)
- **B0 — FIXED (backstop).** `.use()` stripped off the wire.ts app/padi PROXY types → a sub opened outside `bindingScoped` is a COMPILE ERROR. It enumerated all three below. `78d71dc7c`.
- **B1/B2 — FIXED.** createPolledQuery re-keys the pulse on `activeHost` + resolves client/proc from the active binding (was pinned to the mount-time binding). Test: createPolledQuery.test "re-keys the pulse onto the NEW host's client after a switch". `78d71dc7c`.
- **B3 — FIXED.** useTerminals' terminalExit threads `activeBinding()` (per-terminal list-keyed owner re-keys on switch). `78d71dc7c`.

## Cluster C — switch semantics + pump containment
- **C1 — FIXED.** switchHost epoch guard: `myPick = ++pickEpoch` as the FIRST statement (before the same-host early return — the placement that makes re-picking the current host CANCEL an in-flight add); post-await + catch bow out if superseded. Both verified holes closed. Tests: bindings.test (hole 1 no-yank, hole 2 last-pick-wins). `2db21d7a2`.
- **C2 (your ruling) — IMPLEMENTED.** Default/local pump fault stays fatal; a GUEST host's fault logs error + retires the binding (hosts.remove) — no global exit. Tests: hostPool.test (guest retires + no exit; default exits(1)). `555bb901c`.

## Cluster D
- **D1 — FIXED** per your ruling A (cheap — recentHosts was new-in-PR, no version bump). recents moved OUT of preferences into their OWN server-authority `recentHosts` cell (Conf-backed: one `.set` persists + publishes); client subscribes server-authority so another device's add lands live. 1.32.0 migration reworked to move any branch residue out (test: stateMigration.test). `cfbb97c14`.

## Cluster E / F
- **E1 (your ruling) — RECORDED** in the PR description (honest clocks piece 5: served half exists, consumer half deferred; skew shows wrong ages cross-host; pre-existing, named fast-follow).
- **F-a/F-b/F-c — PINNED.** refuseIfWarming blocks+toasts the create while warming (useDaemonStatus.test); a REAL retireSocket makes a stale call's send THROW (bindings.test — not just "the mock was called"); storeHost→sessionStorage + restoreStoredHost (bindings.test). `fe1c16224`.
- **F-d — RECORDED** as a named gap (cross-host live-switch e2e needs a real 2nd machine; pu saturated #1204; mechanics pinned by the unit tests above).

## Hygiene
- **H1 — FIXED.** `createKeyedRoot<K,T>` extracted into @kolu/surface-app/solid; bindingScoped + the provider's buildInfo cell both delegate (memo+eager-force shape, so the provider's swap-teardown guarantee AND bindingScoped's sync-populate both hold). ONE test surface: createKeyedRoot.test. Framework touch → drishti re-pinned. `06640a491`.
- **H2 — RESOLVED BY A1.** The "active"/"bound" reader names (activePadiSurfaceVersion, boundPadiConvergence, padiMemoryDisplay, …) are now ACCURATE per-host names — A1 made them true (server* stays host-independent), so no rename was needed.
- **H3 — FIXED.** 1008 fallback lifted into `installUnknownHostFallback`. `a4f58d701`.
- **H4 — FIXED.** preview route's `?host ?? default` fallback (a relative subresource dropping the query → wrong host's fs) now logs loud; full per-subresource routing noted as an A1-shaped follow-up. `a4f58d701`.

## Green
typecheck + biome clean; client 566, server 171, common 22, surface-app 145 unit. kolu CI 2-platform + drishti CI re-running at the final HEAD [status appended when settled].

## CI — GREEN both PRs, 2-platform, at the final review HEAD
- **kolu #1708** `06640a491`: 34 checks pass (14 linux + 14 darwin lanes + CodeQL/Analyze/SentinelOne). One rerun of the pre-existing flaky `padiBinding.test "reconnects when padi dies"` (an AbortError transport-close race, x86_64-linux-only, untouched by W4 — flaked twice now; flagging for a follow-up flaky-fix).
- **drishti #89** `264b02c9b` (pins kolu `06640a491`): 20 checks pass, both platforms — the surface.md gate satisfied at the final HEAD (re-validated after the H1 surface-app touch).

Both undrafted, NOT merged.

srid:
## Evidence — cross-machine live host-switch (proven end-to-end over ssh)

The plan's deferred **F-d** ("tiles live across a local→remote→local switch") is no longer a gap — it ran for real against a **leased `kolu-ci-5` pu box**. A local `just dev-auto` (random ports, isolated state dir; production `kolu.service` left untouched) bound the remote padi over plain `ssh kolu-ci-5` (the pu proxy already rides `~/.ssh/config`'s `Include ~/.pu-state/*/ssh_config`), and I drove it with chrome-devtools.

**The switch is a real ssh hop to a real remote padi** — server log:
```
pool: add host {host: kolu-ci-5}
binding a REMOTE padi over ssh (KOLU_PADI_HOST set) — the whole canvas is this host
connected {ws:3, host:kolu-ci-5}
[host:kolu-ci-5] kolu-ci-5: already provisioned … skipped copy   ← padi warm in CI's nix store
[host:kolu-ci-5 remote] padi --stdio: no daemon … starting one
[host:kolu-ci-5] connection: copying → connecting → connected     ← the loud warming sequence
```

### 1 · Local (baseline) — the tab views this machine
`Padi contract v1.2` (no ssh); the local workspace renders its terminals.
local baseline

### 2 · Switch → `kolu-ci-5` (no reload) — fresh remote canvas
Banner flips to **`Padi ssh · kolu-ci-5`**; the canvas is **empty** — a brand-new padi daemon on the remote box, fully isolated from local's terminals.
remote empty

### 3 · A real terminal, running on the remote machine
Prompt reads **`toor in ⬢ kolu-ci-5 in ~  ● [Incus]`** — a live shell as the ssh user on the remote Incus container, not local.
remote terminal

### 4 · Switch back → local (round-trip) — local terminals restored
`Padi contract v1.2` again; the full local canvas re-renders via the re-key (`createKeyedRoot`), no reload. (The foreground tile is this very agent narrating the drive.)
back to local

### 5 · Two-context independence — two tabs, two hosts, at once
Same origin: **tab 1 = local** (`Padi contract v1.2`), **tab 2 = `Padi ssh · kolu-ci-5`** — per-tab host (sessionStorage-scoped), switched independently. `kolu-ci-5` also appears in the picker's recents on tab 2 — the D1 server-authority `recentHosts` cell shared it across contexts.
two context

---

### The drive caught TWO real bugs the unit suite couldn't — both fixed on this branch

- **`fix(review): host picker's free-typed connect couldn't submit`** (`9b6d28943`) — typing a new ssh host into "Connect to a host…" + Enter did **nothing**: in value mode the palette's Enter runs `execute(items[selectedIndex()])`, and `items` excludes hints, so a value input with only a hint child (the picker) had no Enter target and `onSubmit` never fired. Since recents start empty, the free-typed connect is the **only** first path to a remote — so the feature was unreachable from a clean install. Fixed at altitude: a value input may offer no options; Enter submits with `selected: undefined`.
- **`fix(review): D1 boot validation — read every schema key`** (`2c7d586b3`) — first boot after upgrade logged a scary `Persisted state does not match schema (recentHosts …) — Delete <path> to reset`, because the startup check hand-built its parse object and omitted the new top-level `recentHosts` key. Fixed by reading every key the schema declares; verified against a pre-D1 state file.

### Honest scope notes
- The remote-padi **provision** was a cache hit (`skipped copy`) — kolu-ci-5 already had the padi closure from CI, so the visual warming was brief; the `copying → connecting → connected` sequence is in the log above. A cold box would show a longer warm.
- No component-render harness exists for CommandPalette (client tests are logic/mock-based), so the palette fix is verified by this live drive rather than a unit test — adding `@solidjs/testing-library` is a noted follow-up.

srid:
Closing in favor of a redesign: the client-side switch machinery here (activeConnectionManager + per-host sockets + wrapper layer) is superseded by a first-class surface notion — a dynamic keyed family of remote surfaces (`@kolu/surface-family`). Design + debate record: docs/atlas/src/content/atlas/remote-surfaces.mdx (kolu.dev/atlas/remote-surfaces.html once merged). The server-half lessons (lazy drv resolution, UnremovableHostError, entry-scoped cells, shared local scan, dialog clarity) carry forward as pins/checklists for the successor PR — clean-room re-derivation, no cherry-pick. The switch feature itself ships with the successor.

PR #1709, opened 2026-07-06
Draft a prompt in the Inspector and send it to the terminal

srid:
**The right-panel Inspector now opens with a Compose box** — a multiline textarea over a **Send** button — so you can write out a prompt for the agent, editing it in peace, before it ever reaches the terminal. Hit Send (or ⌘/Ctrl+Enter) and the draft is written straight into the active terminal's input line: the in-app twin of `kaval-tui send`, for exactly the "type a draft, then hand it off" workflow.

### Two deliberate behaviors

- **Insert, don't submit.** Send writes *exactly* the draft and stops — it never presses Enter for you. A same-breath Enter races a bracketed-paste TUI's (Claude Code / Codex / opencode) paste debounce and gets silently dropped, which is the whole reason `kaval-tui send` made submit a separate step (`df8c6f19` "make send honest"). The draft lands in the agent's input box; you review and press Enter there. Multiline drafts are sent as one **bracketed paste** so they arrive as a block, not line-by-line.
- **Per-terminal, persisted draft.** Each terminal keeps its own draft in `localStorage`, so a half-written prompt survives a tab switch or a page reload. The section is `keyed` on the terminal id, so switching tiles remounts it and its persisted signal rebinds to the new terminal's key.

### How it fits together

```
textarea ─▶ planComposeSend(draft)  ─▶ sendInput({ id, data })
            (pure; bracket if \n;       (active tile's PTY;
             null if empty)              never an Enter)
```

The send-planning is a **pure function** (`composeSend.ts`) so the whole bracket-if-multiline / refuse-empty / never-append-Enter decision is unit-tested without an xterm or DOM harness — the same split `kaval-tui`'s `send.ts` keeps between its plan and the wire.

### Design-philosophy check

- **Fail-fast, no fallbacks** — a failed `sendInput` surfaces a `toast.error` with the server message; it never silently swallows. The Compose section is gated on the *active arm* exactly like the Attach section, so the target is always a live PTY (a sleeping tile's `sendInput` would quiet-drop).
- **Reuse the source of truth** — persistence rides the canonical `persistedPref` (over `@solid-primitives/storage`'s `makePersisted`), keyed `kolu:<feature>-by-terminal:<id>` like the comments store; the bracketed-paste markers come from `@kolu/terminal-protocol`, shared with the CLI and the rich client's own paste path.
- **Volatility boundaries** — no new capability was extracted: sending input is an existing client concern (`lifecycle.sendInput`), and a persisted draft is a leaf. No new `@kolu/*` package.

### Coverage

- **Unit** — `composeSend.test.ts` (6 cases): single-line verbatim, multiline bracketed, never a trailing Enter, whitespace preserved, empty/whitespace-only → `null`.
- **e2e** — two scenarios in `right-panel.feature`: the drafted text reaches the terminal's input line after Send, and a draft survives a page refresh.

Also adds an ambient discoverability tip and a changelog entry.

> _Evidence (video of the flow) and CI to follow on this PR._


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `59ab99b358bf`

### Applied (4)
- `lowy-1` Compose clears the draft after a sendInput that can silently drop — asymmetric with the guarded paste path — commit `a0514b824`
- `lowy-2` Bracketed-paste framing decision duplicated between planSend and planComposeSend — commit `ab7fb6a62`
- `hickey-1` planComposeSend re-implements kaval-tui's planSend policy — (uncommitted)
- `hickey-2` Bracketed-paste wrap sandwich re-derived at every call site — (uncommitted)

### Agreed — no change (3)
- `lowy-3` Per-terminal draft-rebind contract is split across parent `keyed` and child key-derivation (packages/client/src/right-panel/MetadataInspector.tsx:54)
- `hickey-3` Draft-per-terminal identity split between child key and parent remount (packages/client/src/right-panel/ComposeSection.tsx:38-42)
- `hickey-4` Duplicated active-arm && terminalId gate in MetadataInspector (packages/client/src/right-panel/MetadataInspector.tsx:54)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `59ab99b358bf`

### Round 1

**codex** — approved: `false`

The change is mostly well-scoped, but I found several correctness issues around Compose treating a fire-and-forget terminal write as confirmed delivery, checking the wrong terminal for liveness, and allowing in-flight sends to erase or duplicate drafts. The new refresh scenario also appears to fail against the existing persisted right-panel behavior. I reviewed read-only and did not modify files.

Findings:
- `F1` · blocking · open — The new persistence scenario opens the right panel, refreshes, then presses the toggle shortcut again before asserting the panel is visible. Existing behavior explicitly persists right-panel visibility across refresh, so after line 99 the panel is already open; line 100 toggles it closed and line 101 should time out. This makes the added e2e scenario fail instead of proving draft persistence. (packages/tests/features/right-panel.feature:99)
- `F2` · major · open — `await lifecycle.sendInput(...)` only confirms that padi accepted the RPC, not that the bytes reached the PTY. The server handler calls `getActiveTerminal(id)?.handle.write(...)`, and the local proxy's `write` returns `void` while it queues the real kaval write behind `ready.then(...).catch(log)`. Clearing the persisted draft at line 78 can therefore erase user text after an unacknowledged or later-failed write. Clearing after confirmed delivery is fine; this path does not confirm delivery. (packages/client/src/right-panel/ComposeSection.tsx:62)
- `F3` · major · open — The liveness callback checks whether the current Inspector metadata is active, not whether the terminal that was sent to is still active. If a send to terminal A is in flight and the UI switches to active terminal B, this callback returns true even if A slept or died and `sendInput` quiet-dropped. The inverse can also toast/preserve after a successful send if the user switches to a sleeping tile. The check must be target-specific. (packages/client/src/right-panel/MetadataInspector.tsx:59)
- `F4` · major · open — `sending()` only disables the button; `send()` itself does not guard against re-entry, and the textarea remains active. Cmd/Ctrl+Enter can dispatch multiple sends while the first RPC is pending. Also, if the user edits the draft during that pending send, the success path at line 78 clears the new text typed after the send began. (packages/client/src/right-panel/ComposeSection.tsx:57)
- `F5` · minor · open — The catch block casts `unknown` to `Error` and reads `.message`. If a non-Error value is thrown, the user-facing toast becomes `Failed to send to terminal: undefined`, losing the only diagnostic surface for this failure path. (packages/client/src/right-panel/ComposeSection.tsx:79)


**claude** — Fixed F1/F3/F4/F5 (the target-specific liveness check, re-entry + captured-draft clear, robust error message, and the broken refresh scenario); disputed F2 as a request to break a deliberate contract that the already-landed sibling path shares.

- `F1` **fixed** — Confirmed real. Right-panel visibility is persisted server-side (`preferences.rightPanel.collapsed`, `useRightPanel.ts:230-248`), and the existing "Right panel state persists across refresh" scenario (`right-panel.feature:142`) asserts exactly that: after a reload the panel is STILL visible with no re-toggle. My new scenario re-toggled after the refresh, which would COLLAPSE the already-open panel and fail the next visibility assert. Removed the post-refresh `When I press the toggle inspector shortcut` (kept the idempotent Inspector-tab re-select) and asserted visibility directly after the reload — `right-panel.feature:92-107`.
- `F2` **disputed** — The mechanism codex describes is accurate: `local.ts:170` `write()` is fire-and-forget (`this.ready.then(...).catch(log)`), so `await sendInput` confirms only that padi accepted the RPC. But the proposed fix — a new "acknowledged send" RPC that awaits the kaval write — is out of scope and contradicts three things: (1) `sendInput` is a DELIBERATE fire-and-forget stream op (#1628; `terminal-registry.ts:242`, `servePadi.ts:321-328`) with no acknowledged variant on the surface — adding one is a design change to padi's contract and a parallel source of truth (conventions.md "reuse the existing source of truth"). (2) The already-landed `deliverScratchPaste` (`pasteDelivery.ts:24-49`, gauntlet-reviewed) has the IDENTICAL property — it `await`s `sendInput` and treats RPC-accept as delivered, guarding only with an `isActive` re-check. Condemning Compose here would equally condemn that shipped path; Compose deliberately mirrors it. (3) The residual "clear loses text" window is closed by the F3 fix: if kaval died such that a post-ready write rejects, the target terminal is no longer active, so the now-target-specific `isActive()` returns false and the draft is PRESERVED. What's left is a fault where padi's proxy accepts+forwards but the pty-host rejects while the terminal is still registered active — logged server-side (`local.ts:173`), identical to every `sendInput`/`resize` in the app, and not something a leaf Compose feature should carry a bespoke ack channel for.
- `F3` **fixed** — Confirmed real. The old `isActive={() => activeArm(meta()) !== undefined}` closed over MetadataInspector's `meta()`, which follows the ACTIVE TILE, not the send target. An in-flight send whose tile is switched mid-RPC would re-check the wrong terminal (clear terminal A's draft because B is active, or preserve/toast because B slept). Fixed by having `ComposeSection` resolve liveness from its OWN `props.terminalId` via the store — `activeArm(store.getMetadata(props.terminalId)) !== undefined` — exactly the pattern `Terminal.tsx:974-976` uses to gate `deliverScratchPaste`. Dropped the `isActive` prop entirely so the target can't diverge by construction (`ComposeSection.tsx:35-45`, `MetadataInspector.tsx:57`).
- `F4` **fixed** — Confirmed real. `send()` had no re-entry guard (the disabled BUTTON blocks the click path, not the ⌘/Ctrl+Enter key path on the textarea), and the success path cleared `draft()` unconditionally, wiping any text typed during the in-flight RPC. Added `if (sending()) return;` at the top, captured `const sent = draft()` before the await, and clear only when `draft() === sent` — a newer edit survives (`ComposeSection.tsx:send`).
- `F5` **fixed** — Confirmed real. The catch cast `unknown` to `Error` and read `.message`, yielding `...: undefined` for a non-Error throw. Normalized with `const message = err instanceof Error ? err.message : String(err)` (the same idiom as `exportSessionAsHtml.ts:71`) before building the toast.

commit: `a0b9d33cd0dc6b55af393513821884b8eff7b10c`


### Round 2

**codex** — approved: `true`

All previously raised findings are resolved in the current tree. I verified the refresh scenario no longer re-toggles the already-open panel, Compose now checks liveness for its own terminal id, re-entry and stale-clear behavior are guarded, and error text is normalized for non-Error throws. I also accept CLAUDE's F2 dispute as a shared `sendInput` contract issue rather than a Compose-specific defect requiring a new API in this branch; I did not run tests in this read-only review pass.

Findings:
- `F1` · blocking · resolved — The persistence scenario previously re-toggled the right panel after refresh, collapsing an already persisted-open panel before asserting visibility. (packages/tests/features/right-panel.feature:102)
- `F2` · major · resolved — `sendInput` is fire-and-forget and does not acknowledge that bytes reached the PTY before Compose clears the draft. (packages/client/src/right-panel/ComposeSection.tsx:76)
- `F3` · major · resolved — The liveness check previously followed the current Inspector metadata rather than the terminal targeted by the in-flight send. (packages/client/src/right-panel/ComposeSection.tsx:47)
- `F4` · major · resolved — `send()` previously allowed key-path re-entry and could clear newer draft text typed while the first send was pending. (packages/client/src/right-panel/ComposeSection.tsx:63)
- `F5` · minor · resolved — The catch block previously cast `unknown` to `Error`, producing `undefined` in the toast for non-Error throws. (packages/client/src/right-panel/ComposeSection.tsx:94)

_codex on the rebuttal:_ F2: I accept the dispute. The unacknowledged post-ready write failure window is real, but it is part of the existing fire-and-forget `sendInput` contract used by terminal input and `deliverScratchPaste`, not a Compose-only regression once the target-specific liveness bug is fixed. Requiring an acknowledged send RPC here would be a padi API redesign and a parallel send surface, so I am marking F2 resolved rather than holding this leaf feature responsible for changing that contract.

srid:

Ran Pass 1 (rule checklist) + Pass 2 (fact-check) over `git diff <merge-base>...HEAD` (`--no-elegance`, since `/simplify` already ran this gauntlet). Two fixes applied, two findings recorded as justified skips.

### Applied
- **`no-thin-wrapper-functions`** — `Terminal.tsx`'s `wrapPath: (path) => wrapBracketedPaste(path)` was an eta-expansion of a signature-identical function; inlined to `wrapPath: wrapBracketedPaste`. (commit `6515e94`)

### Recorded — deliberate, not fixed
- **`no-unbounded-growth` (localStorage draft keys)** — the per-terminal draft key is intentionally **not evicted** on terminal close: a slept→restored terminal keeps the same id and must keep its draft (the survives-restore guarantee this feature exists for), exactly as the pre-existing `useComments` store leaves its `kolu:comments-by-terminal:<id>` entry. Each draft is a small string bounded by terminals-per-session, and the compose box doesn't own the terminal lifecycle to hook a real deletion. Rationale is now documented in-code (commit `6515e94`).
- **Debounce the per-keystroke `localStorage` write (fact-check)** — declined with rigor. The feature's central guarantee is *"a half-written prompt survives a reload"* and the e2e persistence scenario reloads immediately after typing; a trailing debounce inherently loses the last keystrokes on an immediate reload (and would break that test). Recovering the guarantee under a debounce needs a `pagehide`/`beforeunload` flush — three new mechanisms (live signal + timer + unload flush) to optimize a `setItem` of a **prompt-sized (few-KB) string at human typing speed** (sub-millisecond, not a render/wire hot path). The direct write is the simpler, more-correct choice for this surface.

### Clean
No silent error-swallowing, no dead code, no untyped escape hatches, toast surfaces `err.message`, Tailwind-only styling, e2e polls async state (`waitForBufferContains`). `planComposeSend` edge cases (all-newline → `null`, trailing-newline → bracketed) verified consistent with `kaval-tui`'s `planSend`.

srid:
## Evidence



▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=compose-box-send.mp4

Recorded via `features/right-panel.feature` — *Compose box inserts a drafted prompt into the terminal* (passing, `KOLU_EVIDENCE=1`). The clip shows: open the Inspector tab in the right panel, type a draft into the Compose box, click Send — the drafted text lands directly in the terminal's input line, **not** submitted (the user still presses Enter themselves).

**Mobile:** the Compose card lives in the same `RightPanel` → `MetadataInspector` subtree that already renders as a bottom drawer at phone widths, so it inherits the same responsive layout already verified for the Directory/Git/Attach cards — no separate mobile capture needed.

srid:

The **x86_64-linux** lane ran on **`kolu-ci-6`** (`dev-x86-64-linux-03`) — commit `6515e942e`, **passed** ✓

- **Lane wall** (pipeline): **3m31s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 3m12s | ✓ |
| `ci::home-manager` | 2m18s | ✓ |
| `ci::unit` | 1m24s | ✓ |
| `ci::nix` | 53s | ✓ |
| `ci::smoke` | 52s | ✓ |
| `ci::biome` | 30s | ✓ |
| `ci::fmt` | 29s | ✓ |
| `_ci-setup` | 10s | ✓ |
| `ci::install` | 8s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/6515e94/timings.jsonl`; pool state is a live `flock` probe.</sub>

PR #1713, opened 2026-07-06
fix(kaval): environment-aware daemon discovery + stop tests leaking zombies

srid:
## Why

`kaval-tui list` and the Kaval dialog reported **different** sets of kaval daemons on the *same* machine, and the e2e/integration suite left kaval daemons running after it finished (the `/tmp/kaval-*` pile). Root cause: discovery only ever read the **one** runtime drawer the *calling* process's `$XDG_RUNTIME_DIR` pointed at, and inside a kolu terminal `kaval-tui` ignored the `$KAVAL_SOCKET` that already names the right daemon.

## What changed (all in `packages/kaval`)

**0. `resolveRunningKavalSocket` honours `$KAVAL_SOCKET`** — the precedence chain is now `--socket` → `$KAVAL_SOCKET` → discover, mirroring padi-tui's `resolveRunningPadiSocket`. `$KAVAL_SOCKET` is stamped into every PTY a kaval/padi spawns, so a flag-less `kaval-tui` *inside* a kolu terminal dials the daemon that owns it and never falls through to the ambiguous "more than one daemon" scan. When set, it **short-circuits discovery entirely**.

**1. Discovery scans BOTH runtime roots (the real fix)** — `discoverKavalCandidates` now unions the env-derived `$XDG_RUNTIME_DIR/kaval*` **and** the fixed `/tmp/kaval-$UID*` fallback (always), de-duped by canonical path. Both roots are read back through `getRuntimeSocketPath` (the `/tmp` spelling via the builder with XDG forced off) — never hand-spelled — so discovery can't diverge from construction. The owner-only privacy check is unchanged. A caller with XDG set (the dialog) now sees the `/tmp` daemons a nix-shell left, and vice-versa.

**2. kaval self-exits when its state-root is deleted (zombie safety net)** — a padi-spawned kaval reads the `state-root` manifest padi writes beside its socket and polls it; once that path is gone (2 consecutive checks, to rule out a transient), it logs and shuts down cleanly via the spine's abort path (gate released, socket unlinked). A standalone kaval has no manifest and is never watched. Fail-fast, not silent-degrade.

**3. The e2e harness reaps every daemon it spawns (the actual leak source)** — `socketDaemon.test.ts` now reaps each spawn **by gate pid** on every test, pass or fail — which reaches the tsx-CLI **forked grandchild** daemon that a `SIGKILL` of the tracked child handle never could (the launcher-guard test deliberately forks one). Leaked cwd dirs are swept too, and a file-level assertion proves no spawned kaval survives the suite.

## Regression pins (the three requested)

| Pin | Test |
|---|---|
| discovery scans both roots | `socketPath.test.ts` → *"discoverKavalCandidates — unions the XDG root AND the /tmp fallback"* (seeds a real `/tmp` daemon + an XDG daemon; asserts the `/tmp` one is found **with XDG set** and regardless of XDG) |
| `$KAVAL_SOCKET` short-circuits discovery | `socketPath.test.ts` → *"returns \$KAVAL_SOCKET and does NOT consult discovery"* (seeds two daemons that would be `many`; env yields `env`) + *"an explicit --socket still overrides \$KAVAL_SOCKET"* |
| state-root-gone self-exit | `daemonMain.test.ts` → *"shuts itself down … when its state-root is deleted"* (+ no-false-trigger + standalone-never-watched) |
| harness leaves no zombie | `socketDaemon.test.ts` → file-level `afterAll` asserts no spawned gate holder is alive |

## Testing
- `packages/kaval` unit suite: **139 passing** (incl. the new pins); `kaval-tui`: 127 passing; typecheck clean on kaval/kaval-tui/padi; `biome lint --error-on-warnings` clean.
- No `@kolu/surface*` API change → no drishti mirror PR needed.

## Not part of this PR — pre-existing runtime cruft
There are ~25 live leaked daemons + hundreds of stale dirs under `/tmp` on the box from **before** this fix. This PR does **not** touch them (they're a real machine's running processes — your call). After merge you can list them and decide:

```sh
# LIST leaked off-XDG kaval daemons for your uid (does not kill anything):
for d in /tmp/kaval-$(id -u) /tmp/kaval-*-$(id -u); do
  [ -S "$d/pty-host.sock" ] && echo "$d  (pid $(cat "$d/kaval.pid" 2>/dev/null))"
done
```

### DIFF
--- a/packages/kaval/src/daemonMain.test.ts
+++ b/packages/kaval/src/daemonMain.test.ts
@@ -1,0 +1,119 @@
+/**
+ * Pins kaval's state-root self-exit (zombie hygiene): a padi-spawned kaval whose
+ * ephemeral state-root is deleted out from under it — an e2e/nix-shell run's temp
+ * dir removed when the shell exits — must reap ITSELF rather than linger as a
+ * leaked daemon holding a socket forever. The watcher reads the `state-root`
+ * manifest padi writes beside kaval's socket; when that path is gone it aborts the
+ * daemon through the spine's normal teardown (socket unlinked, gate released).
+ *
+ * A standalone kaval has NO manifest and is deliberately never watched — its
+ * reason to exist isn't tied to any state-root. These run the REAL `runKavalDaemon`
+ * in-process with a tiny poll interval so the reap is observable in milliseconds.
+ */
+import { existsSync, mkdtempSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import type { DaemonExit, Logger } from "@kolu/surface-daemon";
+import { afterEach, describe, expect, it } from "vitest";
+import { runKavalDaemon } from "./daemonMain.ts";
+import { KAVAL_GATE_FILE, writeStateRootManifest } from "./socketPath.ts";
+
+const silentLog = {
+  debug: () => {},
+  info: () => {},
+  warn: () => {},
+  error: () => {},
+  child: () => silentLog,
+} as unknown as Logger;
+
+const sleep = (ms: number): Promise<void> =>
+  new Promise((r) => setTimeout(r, ms));
+
+// Every daemon a test starts, tracked so a failing assertion can't leave it
+// running: `afterEach` aborts each and awaits its clean exit.
+const started: Array<{ ac: AbortController; exit: Promise<DaemonExit> }> = [];
+afterEach(async () => {
+  for (const d of started.splice(0)) {
+    d.ac.abort();
+    await d.exit.catch(() => {});
+  }
+});
+
+interface Started {
+  socketPath: string;
+  gatePath: string;
+  dir: string;
+  ready: Promise<void>;
+  exit: Promise<DaemonExit>;
+}
+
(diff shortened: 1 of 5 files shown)


srid:

The **x86_64-linux** lane ran on **`kolu-ci-2`** (`dev-x86-64-linux-04`) — commit `9a8553413`, **passed** ✓

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 2m3s | ✓ |
| `ci::unit` | 1m24s | ✓ |
| `_ci-setup` | 30s | ✓ |
| `ci::pnpm-hash-fresh` | 20s | ✓ |
| `ci::fmt` | 11s | ✓ |
| `ci::biome` | 11s | ✓ |
| `ci::smoke` | 9s | ✓ |
| `ci::install` | 8s | ✓ |
| `ci::nix` | 5s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/9a85534/timings.jsonl`; pool state is a live `flock` probe.</sub>

PR #1725, opened 2026-07-08
Make hosts the header tabs — a real, host-first tab bar

srid:
**Hosts are the header tabs now.** The chrome bar treats the selected host as the primary workspace context instead of a pill among process badges, and each host tab carries its own Padi/Kaval status marks. The header also blends with Kolu's hostname-derived PWA theme color as a quiet tint, so installed-window chrome and in-app chrome feel related without turning the bar into a saturated accent strip.

Beyond the host-first model, the bar was reworked to **read and behave like a genuine browser tab bar** — correct in both light and dark, aligned to one row, and spanning the full width over the inspector like maximized mode.

### The tab bar

- **Real tabs.** Active host = a soft app-accent outline + faint accent tint (the one selection cue that survives *both* the white light-mode header and the theme-tinted dark one — a neutral surface fill was invisible against the `surface-0`-based header). Inactive hosts = clean outlined folder tabs, not floating text.
- **One aligned row.** The Kolu mark, the in-tab Padi/Kaval marks, and the right-hand control cluster all sit on one baseline; the marks share a single 28px button so Kolu/Padi/Kaval read as one family. The controls are vertically centered (symmetric padding) as the toolbar they are.
- **Full-width docked header.** The bar is now docked in *both* postures (was a floating overlay that stopped at the panel's left edge); the canvas and the right inspector flow *below* it. No drop shadow — a shadow makes the tabs read as floating above the content, fighting the connected-tab metaphor.
- **Quiet focus.** The tab focus ring is keyboard-only (`:has(:focus-visible)`), so a mouse click no longer flashes a ring.

### Host-first behavior

- Padi/Kaval marks stay visible on **every** host tab (active *and* inactive), so a red remote is obvious without switching first; transient switcher rows use inert marks so a popover row never owns a dialog that unmounts mid-switch.
- Clicking a host's Padi/Kaval mark **switches to that host, then opens** the dialog — an anchored dropdown under the mark, with the host name on the panel so ownership is obvious. Opening an inactive host's panel and then drifting off it auto-closes, so a panel never mixes one host's status with another's rows.
- Failed Padi entries stay visibly failed/disconnected instead of collapsing to a grey/connecting look.

> **Restored regression:** the compaction had reduced the *"newer kaval build available"* nudge to a hover-tooltip/dialog only — a build-behind kaval looked identical to a current one. It's back as an at-a-glance **amber pip** on the kaval mark (the status dot can't carry it — a stale daemon is honestly `connected`).

### Review gauntlet

Ran the full `/be-review` gauntlet (lens debate → codex debate → simplify → code-police) on top of the redesign. Notable structural/behavioral fixes it landed:

| Fix | Why |
| --- | --- |
| One per-host Padi/Kaval reader + presentational marks | four hand-rolled copies had already silently diverged (one skipped a clock reprojection) |
| Daemon dialogs close on host-switch | a panel could otherwise blend one host's status with the active host's memory/restart |
| Shared `processMemory` reader | two sub-chips subscribed the same cell → **one** poll-failure toast, not two |
| Dropped an orphaned always-on `identity` subscription | its only caller was removed but the live sub kept running |

### Validation

`tsc` green · biome lint clean · 50 client unit tests pass · `nix build` green · verified live in both light and dark mode via chrome-devtools.

### Try it locally

```sh
nix run github:juspay/kolu/feat/header-host-first-daemons
```

_Header design started on Codex (`gpt-5`); tab-bar polish, regression fix, and the `/be-review` gauntlet on Claude Code (model `claude-opus-4-8`)._


srid:

| 1 | Hickey | Clipped mem/update chrome in dual slot | Fixed in this PR |
| 2 | Hickey | Dual-daemon chrome assembled twice (chip + dropdown) | Fixed in this PR |
| 3 | Hickey | `filled` free boolean decoupled from host facts | Fixed in this PR |
| 4 | Hickey | False slot↔estimate lockstep convention | Fixed in this PR |
| 5 | Lowy | Padi mark painted host-independent `padiLink` | Fixed in this PR |
| 6 | Lowy | README still co-equal Kolu/Kaval rail | Fixed in this PR |
| 7 | Lowy | Stale ChromeBar DaemonSlot comments | Fixed in this PR |

### Hickey rationale
Fixed-width dual slot had competing content policies (skew-only vs mem/update badges). Dual-daemon was assembled both inside HostChip and as a sibling of HostDropdownSwitcher. Free `filled` prop did not encode the host whose facts children read. Slot CSS and chip width estimate were falsely "lockstepped."

### Lowy rationale
Placing Padi on the active host chip while painting host-independent `padiLink` was a boundary lie under multi-host. Drove glance status from `activeEntryState`. Architecture docs still described the pre-host-first identity rail.

srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `2424c934edd5`

### Applied (8)
- `lowy-1` Per-host daemon read logic duplicated across interactive and static marks — and the two copies already diverge — commit `2d8e87d9f`
- `lowy-2` InfoDialog panels advertise a per-host contract but read half their rows from global activeHost — the "which host" boundary is split across two sources — commit `3421ea871`
- `lowy-3` HostDualDaemonSlot selects among three behaviors with two independent boolean flags — commit `435ee1de5`
- `hickey-1` Daemon mark's state-read + dot render is duplicated between the interactive sub-chip and the static mark — (uncommitted)
- `hickey-2` `openForHost` + `isCanvasHost` duplicated verbatim across Padi and Kaval sub-chips — (uncommitted)
- `hickey-3` Host→label function re-hand-rolled in a second module — commit `a67607535`
- `hickey-4` Info panels source their host identity half from props, half from the global activeHost — (uncommitted)
- `hickey-5` HostDualDaemonSlot encodes three modes in two booleans, leaving an illegal combination expressible — (uncommitted)

### Agreed — no change (4)
- `lowy-4` HostSwitcherRow and HostChip duplicate the host-row composition (dot + label + status + urgency + daemon pair + remove) (packages/client/src/host/HostSelectorStrip.tsx:918-1050 (HostSwitcherRow) vs 772-913 (HostChip))
- `lowy-5` Electricity test — receptacle is useAnchoredPopover; no new graduation candidate (packages/client/src/ui/InfoDialog.tsx:519-555; packages/client/src/host/HostSelectorStrip.tsx:1061-1101)
- `hickey-6` Anchored-popover panel shell copy-pasted; this PR adds two more copies (packages/client/src/ui/InfoDialog.tsx:521)
- `hickey-7` Two hand-maintained exhaustive maps over EntryState.kind live in separate modules (packages/client/src/host/HostSelectorStrip.tsx:139)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 round(s) · codex reviewed at `xhigh` reasoning effort · base `2424c934edd5`

### Round 1

**codex** — approved: `false`

The branch is moving the desktop chrome toward a full-width, host-first header, but I do not think it is ready yet. The largest issues are regressions in keyboard/focus behavior from replacing the modal dialog shell with a raw portalled div, and host-scoping leaks where host-labelled daemon dialogs still read or act through active-host globals. I also found a stale-memory display path, duplicated daemon-status subscriptions, ARIA tab misuse, and a few cleanup issues in docs, CSS, and tests.

Findings:
- `F1` · major · open — `InfoDialogShell` replaced the Corvu/`ModalDialog` focus-managed shell with a raw portalled `div role="dialog"`. Opening a Kolu/Padi/Kaval panel leaves focus on the trigger, does not move focus into the panel, does not trap or scope focus, and does not restore focus on close. Keyboard users can tab through the app behind the open dialog before reaching its controls, and close can leave focus on a removed element/body. (packages/client/src/ui/InfoDialog.tsx:65)
- `F2` · major · open — The daemon dialogs are labelled and anchored for the host whose mark was clicked, but several rows and actions still read active-host singletons. Kaval uses `expectedKaval()`, `boundHostKavals()`, `kavalMemoryDisplay()`, and `restartDaemon()`/`activePadiRpc`; Padi uses `boundPadiConvergence()`, `daemonScanBoundHost()`, and `padiMemoryDisplay()`. If `activeHost` changes while the popover remains open, the panel can mix one host's status/label with another host's memory/inventory, and Kaval restart can recycle the newly active host rather than the host named by the open dialog. (packages/client/src/kaval/KavalInfoDialog.tsx:127; packages/client/src/padi/PadiInfoDialog.tsx:173)
- `F3` · major · open — The Padi tooltip's memory readout is gated only on `padi.live()`, which is just the browser-to-server transport (`daemonTransportLive`). It does not require this host's Padi entry to still be connected. A remote host that was connected and then fails/warm-reconnects can keep showing a retained RSS value next to a failed/connecting Padi state. (packages/client/src/host/HostDaemonChips.tsx:201)
- `F4` · minor · open — `KavalSubChip` calls `useHostKaval(props.host)` for tooltip/dialog data, then renders `KavalMark`, which calls `useHostKaval(props.host)` again for the dot. Because narrowed collection `.use({ keys })` calls are per-consumer, each interactive Kaval mark opens two `daemonStatus` subscriptions for the same host. That duplicates stream work and scales with visible hosts. (packages/client/src/host/HostDaemonChips.tsx:145; packages/client/src/host/HostDaemonChips.tsx:265)
- `F5` · minor · open — The host selector now exposes `role="tablist"` / `role="tab"` / `aria-selected`, but it does not implement the ARIA Tabs contract: no tabpanels/`aria-controls`, no roving tabindex, and no arrow-key tab navigation. Screen readers will announce a tab interface that behaves like ordinary buttons. (packages/client/src/host/HostSelectorStrip.tsx:207)
- `F6` · minor · open — The docs/comments still describe the old layout where the tiled chrome is an absolute overlay and the inspector panel shrinks the bar's right edge. The implementation and the deliberate design decision are now full-width, in-flow chrome with no panel-width right offset. This stale architecture text will mislead future layout work. (README.md:56; packages/client/src/ChromeBar.tsx:10)
- `F7` · minor · open — The new `.chrome-bar-surface` custom CSS is ordinary component styling, which violates the repo's `styling-tailwind-only` rule. It also detects theme presence via `[style*="--chrome-theme-color"]`, which is a brittle dependency on serialized inline-style text. (packages/client/src/index.css:56)
- `F8` · nit · open — The active-host Kaval chip selector is now duplicated as a helper pattern and as literal strings in multiple step files. This is exactly the kind of fragile test selector that will drift as the host-chip structure changes. (packages/tests/step_definitions/chrome_memory_steps.ts:11; packages/tests/step_definitions/kaval_daemon_steps.ts:119)


**claude** — Fixed the three real host-scoping / staleness / duplicate-subscription defects (F2, F3, F4), fixed the stale docs and the brittle style-substring selector (F6, F7-part), refactored the duplicated test selector into shared support (F8), and disputed F1 (deliberate non-modal anchored-dropdown design) and F5 (valid simpler tab pattern), plus the tailwind-only half of F7.

- `F1` **disputed** — The switch from Corvu `ModalDialog` to a portalled anchored popover is the branch's DELIBERATE, documented design: `InfoDialog.tsx:1-6` states these are anchored dropdowns "never a center-screen modal," riding `useAnchoredPopover` for outside-click + Escape dismiss. This makes InfoDialog CONSISTENT with the entire family of anchored popovers in the codebase — `SettingsPopover`, `RecordPopover`, `PrUnavailablePopover`, `FootnotePopover`, `OptionMenu`, `ActivityWindowChip`, `HostSwitcherPopover` all use `useAnchoredPopover` and NONE trap focus, because they are non-modal. For a non-modal popover, leaving focus on the trigger is the CORRECT menu-button pattern, not a regression; `aria-modal` is (correctly) absent, so `role="dialog"` announces a labelled non-modal panel — arguably more accessible than the siblings, which set no role at all. Focus-trapping a non-modal popover would be the defect. Codex's one legitimate sub-point (Escape/close returns focus to `body` rather than the trigger) is shared by all eight sibling popovers and belongs in `useAnchoredPopover` for every consumer, not bolted onto this one panel — out of scope for a header redesign.

- `F2` **fixed** — Real and reachable: `activeHost` can move OFF the open dialog's host WITHOUT an outside click via `wire.ts`'s `hostReconcileTarget` auto-switch (fires on a silent pool departure; `useAnchoredPopover`'s dismiss never runs), leaving the panel's props-scoped status blended with another host's active-host rows / restart target. Fixed exactly as codex suggested — close the popover whenever the active host changes away from its host: added a `createEffect` in `useDaemonMarkOpen` (`HostDaemonChips.tsx:186-189`) that runs `setOpen(false)` when `open() && !sameHost(activeHost(), host)`. Switch-first still works (opening sets `activeHost=host` first, so the guard is a no-op at open); the effect only fires on a later switch-away.

- `F3` **fixed** — Real inconsistency: `padiMemoryText` gated only on `padi.live()` (= `daemonTransportLive`, transport only), so a failed/warm-reconnecting remote showed a retained RSS next to a non-connected dot, while the Kaval path already gates on its daemon being `connected`. Now gated on the host entry too: `if (!padi.live() || padi.entry().kind !== "connected") return "memory unavailable"` (`HostDaemonChips.tsx:207-208`).

- `F4` **fixed** — Real double-subscription: `KavalSubChip` called `useHostKaval(host)` (which opens a `daemonStatus.use({keys})` sub) and then rendered `KavalMark`, which called `useHostKaval(host)` again — two subs per interactive mark, scaling with visible hosts. Refactored `PadiMark`/`KavalMark` to take the already-built reader as a prop (`ReturnType<typeof useHostPadi>` / `useHostKaval`) instead of opening their own; each caller (`PadiSubChip`, `KavalSubChip`, `PadiStaticMark`, `KavalStaticMark`) constructs the reader exactly once and hands it in (`HostDaemonChips.tsx:134-163, 357-373`). One subscription per mark now.

- `F5` **disputed** — `role="tablist"`/`role="tab"`/`aria-selected` is a VALID accessible single-select tab bar (one selected at a time; selecting swaps the whole canvas), and the branch deliberately makes hosts read as browser-style tabs (commit `e636cad88` "make hosts real tabs"). The pieces codex calls missing are APG *authoring recommendations*, not ARIA conformance requirements: `aria-controls` on a tab is optional in the ARIA spec, and roving tabindex is a UX recommendation — many production tab bars keep every tab independently `Tab`-focusable, which is intentional here for a toolbar-embedded strip (`HostSelectorStrip.tsx:207-211, 663`). The "tabpanel" is the canvas, swapped in place as a live region rather than a discrete DOM node to key `aria-controls` to. Converting to roving tabindex + arrow nav would REDUCE keyboard reach (one tab stop) and is a distinct interaction change deserving its own e2e coverage, not a defect in this diff.

- `F6` **fixed** — Stale docs corrected. `ChromeBar.tsx:9-20` no longer describes the retired "two positioning modes / absolute overlay / `right: panelSize` offset"; it now documents the single full-width docked rail in both postures with no panel-width right-offset. `README.md:56` updated the same way (was "the chrome bar shrinks its right edge to the panel's left").

- `F7` **partial** — Fixed the concrete, cheap defect: dropped the brittle `[style*="--chrome-theme-color"]` substring selector (a match on serialized inline-style text) in favor of an explicit `[data-themed]` attribute the component sets only when a PWA theme color is present (`index.css:60,73`; `ChromeBar.tsx` header + `data-themed` marker). Disputed the "move color-mix into Tailwind utilities" half: this app themes via CLASS-based CSS tokens with the DARK palette as the `:root` default and `:root:not(.dark)` overriding to light (`packages/theme/theme.css:53-54`), and it wires NO Tailwind `dark:` variant (`@custom-variant` is undefined; there is ZERO `dark:` usage in the client) — Tailwind's `dark:` defaults to `prefers-color-scheme`, which does not track this app's `.dark` class. So the per-mode color-mix cannot ride a Tailwind utility; the `.chrome-bar-surface` rules mirror theme.css's own `html:not(.dark)` canonical mechanism, and index.css already houses ~40 component classes for exactly this kind of mode/pseudo-element styling Tailwind can't express.

- `F8` **fixed** — Extracted the active-host identity-chip selector into `packages/tests/support/hostChip.ts` (`identityChipSelector(testid)` + `IdentityChipTestid`), and consumed it from both `chrome_memory_steps.ts` (replacing the local `chipSelector` helper and the literal in "open the Kaval details dialog") and `kaval_daemon_steps.ts:118` (replacing the literal in "open the kaval rail dialog"). One locator source now.

Verified: `just fmt` clean; `just check` (workspace `tsc --noEmit` for client + server + all packages, then `biome lint --error-on-warnings` over 1231 files incl. the edited test files) green.

commit: `f72986c0b1c661ae569bdebeb0c590650d203ea8`


### Round 2

**codex** — approved: `false`

I re-checked only the existing findings against the current worktree. F2, F3, F4, F8 are fixed as described, and I accept the disputes on F1, F5, and the remaining Tailwind part of F7. One stale ChromeBar comment from F6 still remains, so the review is not fully approved yet.

Findings:
- `F1` · major · resolved — Original concern was that the dialog-like surface lost modal focus behavior after replacing ModalDialog. CLAUDE clarified this is intentionally a non-modal anchored dropdown using the same `useAnchoredPopover` pattern as sibling popovers, with no `aria-modal` and focus left on the trigger. (packages/client/src/ui/InfoDialog.tsx:65)
- `F2` · major · resolved — Host-scoped daemon details could remain open after `activeHost` changed, mixing the dialog's host label/status with active-host rows/actions. (packages/client/src/host/HostDaemonChips.tsx:192)
- `F3` · major · resolved — Padi memory text was gated only on transport liveness, so retained RSS could show while the host daemon entry was not connected. (packages/client/src/host/HostDaemonChips.tsx:225)
- `F4` · minor · resolved — Interactive daemon marks opened duplicate narrowed `daemonStatus` subscriptions by constructing a host reader in the chip and again inside the mark. (packages/client/src/host/HostDaemonChips.tsx:134)
- `F5` · minor · resolved — Original concern was that the host strip used tab roles without the full APG tab interaction model. (packages/client/src/host/HostSelectorStrip.tsx:209)
- `F6` · minor · open — The README and top ChromeBar module comment were updated, but ChromeBar still has a stale comment saying panel-open mode stays on a floating overlay and that a `right:` offset keeps controls off the panel. That contradicts the requested full-width docked layout and the current implementation. (packages/client/src/ChromeBar.tsx:66)
- `F7` · minor · resolved — The brittle `[style*="--chrome-theme-color"]` selector was the concrete correctness risk. The remaining custom CSS concern was about whether this should move into Tailwind utilities. (packages/client/src/index.css:56)
- `F8` · nit · resolved — The active identity-chip selector was duplicated across step definitions. (packages/tests/support/hostChip.ts:8)

_codex on the rebuttal:_ F1: conceded. I treated the surface too much like the previous modal; as a deliberately non-modal anchored dropdown with no `aria-modal`, focus trapping would be the wrong requirement. The trigger-focus restoration point is real but shared by `useAnchoredPopover` consumers and not a blocker for this header change. F5: conceded. The current ARIA is a valid simpler single-select tab pattern; `aria-controls` and roving tabindex are not hard conformance requirements, and I do not have a concrete accessibility failure to hold this open. F7: conceded on the Tailwind-only portion. The brittle style-substring selector was fixed with `data-themed`, and the class-based theme setup justifies keeping the remaining color-mix rules in CSS. F6: holding open narrowly. The main docs were fixed, but `ChromeBar.tsx:66-69` still documents the retired floating overlay/right-offset behavior, so the original stale-comment finding is not fully resolved.


**claude** — Round 2 closed the one finding codex held open: replaced the stale `docked` memo comment in ChromeBar so the source matches the full-width docked layout; every other finding was already resolved in round 1.

- `F1` **resolved (round 1)** — Codex conceded the dispute: the InfoDialog surface is a deliberate non-modal anchored dropdown (`useAnchoredPopover`, no `aria-modal`), so focus-trapping would be the wrong requirement. The shared trigger-focus-restoration point belongs in `useAnchoredPopover` for all consumers, out of scope here. No code change.
- `F2` **resolved (round 1)** — Fixed: `useDaemonMarkOpen` closes the popover when `activeHost` moves off the mark's host (`HostDaemonChips.tsx`). Codex re-verified.
- `F3` **resolved (round 1)** — Fixed: Padi memory text now gated on `padi.live() && padi.entry().kind === "connected"` (`HostDaemonChips.tsx`). Codex re-verified.
- `F4` **resolved (round 1)** — Fixed: `PadiMark`/`KavalMark` take the already-built reader as a prop, so each interactive mark opens one `daemonStatus` subscription, not two. Codex re-verified.
- `F5` **resolved (round 1)** — Codex conceded the dispute: `role="tablist"`/`role="tab"`/`aria-selected` is a valid simpler single-select tab pattern; `aria-controls` and roving tabindex are APG recommendations, not conformance requirements. No code change.
- `F6` **fixed** — This round. The top module comment and README were corrected in round 1, but `ChromeBar.tsx:66-69` still described the retired "panel-open stays on the floating overlay / `right:` offset keeps controls off the panel" behavior. Rewrote the `docked` memo comment (`ChromeBar.tsx:66-70`) to state the header is full-width docked in BOTH postures and that `docked()` no longer gates positioning — it only drives the `data-maximized` marker and the maximize/restore toggle's active state/icon, which is exactly how it's used (lines 92, 139-154).
- `F7` **resolved (round 1)** — Fixed the concrete defect (brittle `[style*="--chrome-theme-color"]` substring selector → explicit `data-themed`); codex conceded the Tailwind-only half given the app's class-based (non-`dark:`) theming.
- `F8` **resolved (round 1)** — Fixed: active identity-chip selector extracted to `packages/tests/support/hostChip.ts` and consumed from both step files. Codex re-verified.

Verified: `just check` green (workspace `tsc --noEmit` + `biome lint --error-on-warnings` over 1231 files).

commit: `497e86641975a8579e4af98117d0001382409cb6`


### Round 3

**codex** — approved: `true`

I re-checked the current worktree against the existing finding set only. The remaining F6 stale ChromeBar comment is now corrected, and the only unstaged tracked diff I saw is a formatting-only change in `HostSelectorStrip.tsx`. All prior findings are resolved; I did not raise any new finding from this round.

Findings:
- `F1` · major · resolved — Original concern was that the dialog-like surface lost modal focus behavior after replacing `ModalDialog`. CLAUDE clarified this is intentionally a non-modal anchored dropdown using `useAnchoredPopover`, with no `aria-modal`, matching sibling popovers. (packages/client/src/ui/InfoDialog.tsx:65)
- `F2` · major · resolved — Host-scoped daemon details could remain open after `activeHost` changed, mixing dialog host context with active-host rows/actions. (packages/client/src/host/HostDaemonChips.tsx:192)
- `F3` · major · resolved — Padi memory text was gated only on transport liveness, so retained RSS could show while the host daemon entry was not connected. (packages/client/src/host/HostDaemonChips.tsx:225)
- `F4` · minor · resolved — Interactive daemon marks opened duplicate narrowed `daemonStatus` subscriptions by constructing a host reader in the chip and again inside the mark. (packages/client/src/host/HostDaemonChips.tsx:134)
- `F5` · minor · resolved — Original concern was that the host strip used tab roles without the full APG tab interaction model. (packages/client/src/host/HostSelectorStrip.tsx:209)
- `F6` · minor · resolved — The remaining stale comment still described the retired floating overlay and panel-width `right:` offset behavior. (packages/client/src/ChromeBar.tsx:66)
- `F7` · minor · resolved — The concrete correctness risk was the brittle `[style*="--chrome-theme-color"]` selector; the remaining dispute was whether the component styling had to move into Tailwind utilities. (packages/client/src/index.css:56)
- `F8` · nit · resolved — The active identity-chip selector was duplicated across step definitions. (packages/tests/support/hostChip.ts:8)

_codex on the rebuttal:_ F6 is now verified fixed: `ChromeBar.tsx:66-70` no longer mentions the retired floating overlay or panel-width right offset. The earlier concessions on F1, F5, and F7 still stand, and there are no remaining open disputes.

srid:

Ran **Pass 1 (rule checklist)** + **Pass 2 (fact-check)** over the header/chrome diff (`--no-elegance`, since `/simplify` already ran this gauntlet). All actioned findings applied in `11fb006e0`.

### Applied

| Rule / class | Finding | Fix |
| --- | --- | --- |
| `no-unbounded-growth` · fact-check | `PadiSubChip` + `KavalSubChip` each opened their **own** `processMemory` subscription for the same host/cell → a single poll failure fired **two identical error toasts**. | Hoisted one `useHostProcessMemory` reader (shared via a new `InteractiveDaemonMarks`) handed to both marks — one subscription, one toast. |
| `dry-rule-of-three` | The 4-arm process-RSS `match(...).exhaustive()` block was byte-identical in both sub-chips. | Extracted `formatProcessMemoryText`. |
| `no-dead-code` | `activePadiIdentity` / `activePadiStartedAt` + their always-on `identitySub` were orphaned — this diff removed the only caller (`PadiInfoDialog` reads per-host props now) but left the **live subscription** running for the session. | Removed the functions + subscription. |
| `prefer-ts-pattern` | New `statusLabelShort` used a `switch`/`default` for a pure kind→label map, next to `DOT_TONE`'s `Record`. | Converted to a `Record<EntryState["kind"], string>` (a new kind is now a compile error, not a silent default). |
| `no-preference-prop-drilling` · `app-shell-stays-thin` | `App.tsx` drilled `themeColor` into `ChromeBar`; `useServerIdentity` was extracted precisely to stop the shell threading identity fields. | `ChromeBar` reads `useServerIdentity().themeColor` directly (matching `AboutDialog`). |
| `comment-the-non-obvious` | The add-host `queueMicrotask` + `isConnected` autofocus had no rationale. | Documented the deferred-mount + detach-race guard. |

### Verified & cleared (no action)

- **`invalid-states-unrepresentable`** — `useHostKaval`'s `startedAt ?? 0` sentinel: Pass 2 cleared it — it mirrors the documented `DaemonStatus` convention (`number`, `<= 0` = unknown, guarded downstream) and is not newly introduced.
- **Dialog-side memory/status singletons** (`useMemoryUsage`, `KavalUpdateBadge`) duplicating per-chip cells for the active host: **pre-existing**, and codex F2 accepted the dialog-reads-active-singleton pattern (guarded by close-on-host-switch). Out of scope for this diff.
- All 8 new `toast.error` sites include `err.message`; all 5 new `.use()` sites supply `onError`; no silent error-swallowing, no untyped escape hatches, no new inline SVG, no persisted-schema changes.

`tsc` green · biome lint clean · 50 unit tests pass.

PR #1726, opened 2026-07-09
fix(padi): the dock's 'last activity' froze for stable-session terminals

srid:
## The bug

The dock's per-terminal recency — the **"Nh ago"** chip, the **activity-window** filter, and the switcher's **MRU-ish ordering** — all read `TerminalMetadata.lastActivityAt`. Since **#1626** the *only* dynamic writer of that field bumped on an agent **identity** change (`kind`/`sessionId`), never on output. So a terminal running one long-lived Claude/Codex session forever **never changed identity** and its recency **froze** — the live symptom srid caught: terminal `0d8c1f59` reading `lastActivityAt` **24.1h old** while it was actively conversing *that second*.

## Root cause

`padi`'s per-terminal `emit` (`local.ts`) set `ctx.live` **only** when the re-resolved agent identity differed from a continuously-tracked `recencyBaseline`. A same-identity output tick (the ~1s agent-detail firehose) therefore reached the pure `fold` as `live=false` and was never a recency candidate. The field's *displayed* meaning is "last activity"; its *stored* meaning had quietly become "last identity change." Full diagnosis: `/tmp/w7-p1-diagnosis.md`.

## The fix — an output-driven bump, throttled, composing with #1626

Two parts, both at the honest producer/fold layer (`the producer observes, kolu remembers`):

| Layer | Change |
| --- | --- |
| **`fold.ts`** (pure) | On a **live** agent observation: an **identity change** stamps recency **unthrottled** (keeps #1626's discrete start/finish/new-session event); otherwise a **same-identity OUTPUT tick** stamps too, but **throttled** to `RECENCY_THROTTLE_MS` (60s). `lastActivityAt` *is* the throttle clock — no new state — so the ~1s firehose writes at most once a minute, never recreating the per-tick noise #1626 deliberately removed. |
| **`local.ts` `emit`** | `ctx.live` regains its honest meaning — *a genuine live observation vs the one adopt re-observation*. A one-shot **recency gate** (`seedRecencyGate` / `stepRecencyGate`) holds back only the **first** agent observation after an adopt (so the saved recency survives a restart), **replacing** the continuous `recencyBaseline` re-seating — a net simplification. |

The two bumps compose: identity change → always; stable-session output → throttled.

## Design philosophy

- **Fail-fast / no fallbacks / no knobs** — no override; `RECENCY_THROTTLE_MS` is a baked-in named constant, not a setting.
- **Reuse the source of truth** — recency stays a *fold-owned remembered fact* (the producer still can't spell it — the write-fence is the emit type). The gate is extracted **from** `emit` into the pure layer so the recency rule lives in one place, mirroring #1626's `agentIdentityChanged`.
- **Electricity** — no new package/boundary; the fix lands inside the existing fold/producer seam.

## Tests (RED first, then green)

`fold.test.ts`: throttled-output-advances-past-window · throttle-coalesces-within-window · snapshot-frame-never-bumps · identity-still-bumps-inside-window; and gate pins: adopt-suppresses-then-output-flows · **adopt-no-false-bump** (idle adopt keeps saved recency) · fresh-spawn-first-agent-is-live. **All of #1626's fold pins stay green.** Touched suites: `@kolu/terminal-workspace` fold (29) + `padi` terminalEndpoint/ptyHost (73).

## Docs

`packages/terminal-workspace/README.md`, `packages/surface/README.md`, and the `awareness-derive-store` Atlas note (+ rebuilt `dist/`) updated; changelog entry under **Fixed**.


srid:

✅ **Consensus** after 2 round(s) · lowy + hickey · base `65345ae16cb6`

### Applied (3)
- `lowy-1` Frame-phase judgment (live + baseline re-seat) is duplicated between the producer and the test — commit `9e6450e60`
- `hickey-1` The recency-stepping loop is duplicated between production emit and the test helper — commit `9e6450e60`
- `hickey-2` "Live activity" is answered twice against two seeds that can disagree — under-documented, not redundant — commit `0ce873fba`

### Agreed — no change (5)
- `lowy-2` FoldCtx interface grew (`runStartedAt`) when the encapsulated recency policy changed (packages/terminal-workspace/src/fold.ts:173)
- `lowy-3` Activity-rule volatility is split across producer and fold, not contained in one module (packages/terminal-workspace/src/fold.ts:205 (with packages/padi/src/terminalEndpoint/local.ts:771-780))
- `lowy-4` Double identity comparison against two different baselines (packages/terminal-workspace/src/fold.ts:205)
- `lowy-5` seedRecencyBaseline lives in fold.ts but seeds state the fold never reads (packages/terminal-workspace/src/fold.ts:103)
- `hickey-3` "OUTPUT tick" names more certainty than the fold actually has (packages/terminal-workspace/src/fold.ts:207)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 4 round(s) · codex reviewed at `xhigh` reasoning effort · base `65345ae16cb6`
### Round 1

**codex** — approved: `false`

The branch fixes the old identity-only freeze directionally, but the new fold arm does not actually know whether an observation is terminal output. It treats any same-identity authoritative agent update as output, which can false-bump recency from delayed metadata updates and still miss real output that does not change AgentInfo. There are also stale contract comments left behind that still describe identity-only recency.

Findings:
- `F1` · major · open — The throttled arm stamps `lastActivityAt` for any authoritative same-identity `agent` observation once the window elapses, but `TerminalEvent` does not carry an output/activity fact. Agent updates include metadata-only or delayed recomputations, for example Claude summary refresh calls `onUpdate(updated)` after `fetchSessionSummary` resolves, so an idle terminal can be marked active at summary-fetch completion time rather than when output actually happened. Conversely, real output that does not change any compared AgentInfo field can still be missed. (packages/terminal-workspace/src/fold.ts:250)
- `F2` · minor · open — The exported `lastActivityAt` documentation still says recency is the time of the last live agent-identity change. That is no longer true in this branch; the fold also stamps throttled same-identity activity. The same stale explanation appears in nearby producer comments such as `packages/terminal-workspace/src/sensors.ts:461`, so future readers will get the wrong contract. (packages/terminal-workspace/src/schema.ts:176)


**claude** — disputed F1 (its premise that an agent observation is decoupled from output is false), fixed F2's stale contract comments.

- `F1` **disputed** — Codex's premise, that "the new fold arm does not actually know whether an observation is terminal output" and "treats any same-identity authoritative agent update as output," is false against the emit path. `emitAgentValue` (packages/terminal-workspace/src/sensors.ts:474) dedups every candidate against the engine mirror via `agentInfoEqual` (packages/integrations/anyagent/src/agent-adapter.ts:209-223), which compares `state · sessionId · model · summary · contextTokens · startedAt · taskProgress`. So an `agent` observation fires ONLY when one of those conversation-derived fields actually changed — and every one of them is a function of the transcript advancing (output). There is no "metadata-only" emit that leaves all of them equal; a re-derivation that changes nothing is dropped at line 474.
  - The specific `fetchSessionSummary` example is exactly backwards: `refreshSummary` (claude-code/src/session-watcher.ts:526) is reached ONLY fire-and-forget from the transcript-change handler (session-watcher.ts:463, at the tail of `reconcile` which runs on a transcript fs event), AND it self-dedups (`if (summary === lastSummary) return`, line 532). So it emits only when new transcript content produced a NEW summary — i.e. output just happened. The only gap between the output and the stamp is the summary-fetch latency (sub-second/seconds), which is immaterial to a 60s-throttled, hour-granularity "Nh ago" chip. It cannot mark an *idle* terminal active; the terminal produced output moments earlier.
  - The "real output that changes no AgentInfo field is missed" half is also not a defect: `contextTokens` moves on essentially any model turn and `state` moves on every lifecycle edge, so meaningful output leaving ALL compared fields equal is unreachable; and even a single skipped tick self-heals on the next changed tick within one throttle window. Recency is coarse by construction (60s throttle, hour chip) — the bug fixed here was a whole-session multi-hour freeze, not a sub-60s gap.
  - Codex's proposed fix (a new `agentOutput`/`agentActivity` observation from the raw transcript/rollout append path) would make recency *worse*: it couples recency to raw byte-appends, so a keepalive/heartbeat write with no user-visible activity would bump it — trading the current output-correlated model for a byte-churn model. Author deliberate-decision #6 also scopes the "recency rides agent observations" awareness model as pre-existing and out of scope. No code change; the model at HEAD is correct.
- `F2` **fixed** — Legitimate doc-accuracy nit; conceded and fixed both cited comments. schema.ts:176 (`AgentMemorySchema.lastActivityAt`) now reads "epoch-millis of the last LIVE agent observation … an identity change stamps immediately, and a same-identity OUTPUT tick stamps through the recency throttle," instead of the stale "last LIVE agent-IDENTITY change." sensors.ts:466 (`emitAgentValue` doc) now reads "recency (kolu's fold — an identity change always, a same-identity output tick throttled)" instead of "on identity change." `just check` (typecheck + biome) green; comment-only edits.

commit: `0ce873fba06ec0d7b695c83431ba4cc62e6c360e`


### Round 2

**codex** — approved: `false`

F2 is fixed: the stale schema and producer comments now describe the new identity-plus-throttled-output recency contract. I am not convinced by the F1 dispute, because the rebuttal proves only that some agent observations are causally triggered by transcript changes, not that they are timestamped at the activity that caused them or bounded close to it.

Findings:
- `F1` · major · open — The throttled arm still stamps recency at the time an authoritative same-identity `agent` observation is emitted, even when that observation is produced by delayed asynchronous metadata resolution rather than at the time output occurred. The concrete Claude summary path is fire-and-forget from `onTranscriptMaybeChanged`, awaits `getSessionInfo` with no timeout or ordering fence, and the file itself tracks `pendingSummaryFetches` because transcript events can outpace summary responses. If a summary response completes much later, the fold records `Date.now()` at completion as `lastActivityAt`, so an idle terminal can be promoted as recently active due to late metadata. The dedup at `emitAgentValue` only proves unchanged metadata is dropped; it does not prove emitted metadata is output-synchronous or bounded near the originating output. (packages/terminal-workspace/src/fold.ts:250)
- `F2` · minor · resolved — Resolved. The exported `lastActivityAt` docs now describe both recency arms, and `packages/terminal-workspace/src/sensors.ts:466` now matches the new contract instead of saying identity-only. (packages/terminal-workspace/src/schema.ts:176)

_codex on the rebuttal:_ F1: I'm holding firm, but narrowing the reason. I accept that `emitAgentValue` dedups unchanged values and that summary refresh is triggered from transcript-change handling. That does not establish the emitted observation is an output tick in time. `refreshSummary` awaits `fetchSessionSummary` asynchronously and uses the completion-time emit path; there is no timeout, cancellation, monotonic request id, or source event timestamp carried into `FoldCtx.at`. The existing `pendingSummaryFetches` diagnostic explicitly acknowledges summary fetch backpressure. So a late or out-of-order metadata completion can still stamp recency at completion time, which is the bug class F1 was about. F2: accepted fixed; the comments now reflect the current contract.


**claude** — closed the one genuinely-unbounded false-bump path codex named (out-of-order summary completion) with a monotonic dispatch fence in `refreshSummary`, and disputed the residual in-order stamp as bounded-and-tolerated.

- `F1` **partial** — Codex narrowed its case to the summary path's async completion-time stamp with "no timeout, cancellation, monotonic request id, or source event timestamp." I concede exactly one sub-claim: the OUT-OF-ORDER completion. `getSessionInfo` (core.ts:1569, a local session-info read) has no ordering guarantee, so of two in-flight fetches the OLDER can resolve LAST, carrying a STALE summary that (a) publishes over a newer one and (b) fires `onUpdate` at its late completion, stamping recency at an arbitrary later time on a possibly-idle terminal. That is the only path where the stamp is not bounded near the activity, and it is a real display bug (stale summary going backwards) independent of recency. Fixed with a monotonic fence: `summaryFetchSeq`/`lastAppliedSummarySeq` (session-watcher.ts:141-152) — each dispatch takes the next seq; a resolved fetch applies only if `seq > lastAppliedSummarySeq` (session-watcher.ts:534-539), else it is dropped. This is precisely the "monotonic request id" codex asked for; it keeps `lastSummary` monotonic in dispatch order and, because the fold reads recency off this emit, bounds any summary-driven stamp to the NEWEST fetch's completion.
  - Disputed (the rest): the IN-ORDER completion-time stamp is not a defect. `refreshSummary` is dispatched SYNCHRONOUSLY with the transcript change (session-watcher.ts:463, tail of `onTranscriptMaybeChanged`), so dispatch time ≈ activity time; the stamp lags only by the local-read fetch latency. The error direction is "slightly too recent," bounded by that latency — it can NEVER reintroduce the multi-hour freeze this PR fixes, and it self-heals on the next real tick within one 60s throttle window. This matches deliberate decision #4 (unclamped, self-healing recency time under fail-fast) and the coarse hour-granularity chip. Note also the summary is only ONE of seven `agentInfoEqual` fields; `state`/`contextTokens` change synchronously in the same pass and stamp recency at output time, so the summary-completion emit UNIQUELY drives recency only when the summary alone changed — which by construction means new content, i.e. the agent WAS active. The "proper" alternative codex floats (carry a source-event timestamp into the observation) would violate deliberate decision #5 — the producer cannot spell recency; the write-fence IS the emit type — so it is out of bounds for this PR by design, whereas the ordering fence is a producer-local summary-correctness fix that stays inside that fence.
- `F2` **resolved** (already fixed round 1; codex confirmed) — no further action.

commit: `0ce873fba06ec0d7b695c83431ba4cc62e6c360e`


### Round 3

**codex** — approved: `false`

F2 remains resolved. F1 is partially improved, and I accept the author's bounded in-order-latency argument once summary ordering is actually fenced, but the current monotonic fence still has an ordering hole: newer completions that return the same summary do not advance the sequence, so an older fetch can still apply afterward and stamp recency late.

Findings:
- `F1` · major · open — The new summary ordering fence only advances `lastAppliedSummarySeq` after `summary !== lastSummary`. If a newer fetch resolves first with the current summary, it returns at line 546 without advancing the fence; an older in-flight fetch can then resolve later with a different stale summary, pass `seq <= lastAppliedSummarySeq`, publish stale metadata, and stamp recency at that late completion time. This leaves the out-of-order false-bump path F1 was narrowed to. (packages/integrations/claude-code/src/session-watcher.ts:545)
- `F2` · minor · resolved — Resolved. The schema and producer comments now describe both recency arms accurately. (packages/terminal-workspace/src/schema.ts:176)

_codex on the rebuttal:_ F1: I accept the narrowed model after a correct ordering fence: an in-order summary completion that lags the transcript event by ordinary local-read latency is tolerable for this coarse, throttled recency signal, and I am not requiring a source-event timestamp. However the implemented fence is incomplete. Because `lastAppliedSummarySeq` is updated only after the summary changes, a newer no-op completion does not supersede older in-flight work. Example: `lastSummary = S0`; seq2 resolves first with `S0` and returns; seq1 resolves later with stale `S1`, passes the fence, and emits late. That is still the stale/late completion bug. F2: already resolved.


**claude** — closed the ordering-fence gap codex named: the fence now advances on any newest-so-far completion (including a no-change one), so no older in-flight fetch can apply afterward.

- `F1` **fixed** — Codex is right, and I conceded it. The round-2 fence advanced `lastAppliedSummarySeq` only AFTER the `summary === lastSummary` check, so a newer no-op completion returned without advancing the fence and left an older in-flight fetch able to resolve late, pass `seq <= lastAppliedSummarySeq`, publish a stale summary, and stamp recency at that late completion. Codex's example holds: `lastSummary = S0`; seq2 resolves first with S0 and returns at the old line 546 (no advance); seq1 (older, seq1 < seq2) resolves later with stale S1, passes the `seq <=` gate, applies S1, late-stamps recency. Fix: reorder so the fence advances the moment a completion is confirmed newest-so-far, before the no-change return (packages/integrations/claude-code/src/session-watcher.ts:545-550) — now `if (seq <= lastAppliedSummarySeq) return; lastAppliedSummarySeq = seq; if (summary === lastSummary) return;`. Any newest completion (changed or not) supersedes older in-flight work, so seq1 in the example now fails `seq1 <= lastAppliedSummarySeq` and drops. This closes the out-of-order path F1 was narrowed to; the in-order bounded-latency stamp codex already accepted is unchanged. `just check` (typecheck + biome, 1166 files) green.
- `F2` **resolved** (fixed round 1, confirmed rounds 2-3) — no further action.

commit: `54f8e215dfe2ad0aa30c5a8699dbe3de86b3ff43`


### Round 4

**codex** — approved: `true`

The current tree resolves the remaining review issue. The summary-fetch ordering fence now advances before the no-change return, so a newer no-op completion correctly supersedes older in-flight fetches. I accept the remaining in-order completion latency as bounded and consistent with the coarse throttled recency model. F2 was already fixed and remains resolved.

Findings:
- `F1` · major · resolved — Resolved. The monotonic summary-fetch fence now advances for any newest-so-far successful completion before checking `summary === lastSummary`, closing the stale older-completion path that could late-stamp recency. (packages/integrations/claude-code/src/session-watcher.ts:549)
- `F2` · minor · resolved — Resolved. The schema and producer comments describe both recency arms accurately. (packages/terminal-workspace/src/schema.ts:176)

_codex on the rebuttal:_ F1: accepted fixed. The previous hole was that a newer no-op completion returned without advancing the fence; the current code now does `if (seq <= lastAppliedSummarySeq) return; lastAppliedSummarySeq = seq; if (summary === lastSummary) return;`, so the older-late example no longer applies. I also accept the disputed in-order completion-time stamp as bounded by the local summary read and tolerable for this throttled recency signal. F2: remains resolved.

srid:

**All clear** — Pass 1 (rules, 26 checks) and Pass 2 (fact-check) both found 0 findings. Pass 3 (elegance) skipped (`/simplify` ran this gauntlet).

Fact-check independently re-derived the load-bearing bits: the two-arm recency fold is exhaustive + mutually exclusive after the `agent`/`unknown` guard; the `ctx.live` vs `agentIdentityChanged(cur.snapshot.agent, …)` divergence right after an adopt (where `cur.snapshot.agent` is seeded `null`) correctly falls through to the throttle arm; the `>=` throttle boundary has no off-by-one; the `summaryFetchSeq` fence has no race window (single-threaded, no `await` between check and update); per-terminal `emit` state can't cross-contaminate; and the docs don't outrun the code.

srid:

The **x86_64-linux** lane ran on **`kolu-ci-7`** (`dev-x86-64-linux-04`) — commit `ae4e82b5f`, **failed** ✗

- **Lane wall** (pipeline): **56s**

| recipe | duration | |
|---|--:|:--|
| `_ci-setup` | 32s | ✓ |
| `ci::home-manager` | 23s | ✗ (exit 1) |
| `ci::atlas-sync` | 18s | ✓ |
| `ci::nix` | 13s | ✗ (exit 1) |
| `ci::smoke` | 10s | ✗ (exit 1) |
| `ci::install` | 7s | ✗ (exit 1) |
| `ci::flake-check` | 7s | ✗ (exit 1) |
| `ci::pnpm-hash-fresh` | 3s | ✗ (exit 1) |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/ae4e82b/timings.jsonl`; pool state is a live `flock` probe.</sub>

srid:
## Evidence

**Why no screen clip.** The on-screen surface here is the dock's per-terminal recency — the **"Nh ago" chip**, the **activity-window filter**, and the **switcher/MRU ordering** — all rendered from `TerminalMetadata.lastActivityAt`. The fix's observable effect is that this value **advances on a stable session's output, throttled to 60 s**, instead of freezing. That effect is genuinely awkward to capture as a faithful before/after clip:

- The **bug only manifests over hours** — a terminal has to run *one* long-lived agent session (stable `kind`+`sessionId`) long enough that its frozen stamp reads visibly stale while it's actively producing output. srid's live repro was a terminal reading **24.1 h ago while conversing that second**.
- The distinguishing frame needs the **pre-fix build** side-by-side: post-fix, a fresh session also shows "just now", so a single post-fix clip of "chip is fresh" doesn't tell fix from bug. The only distinguishing signal is *the chip advancing while the session identity never changes* — which the old code could not do — over a ≥60 s window at minute granularity.
- Capture must run on a **pu box** (production kolu is `active` on this machine — `/dev-server` §0), and the artifact would be a subtle minute-granularity timestamp tick.

So the faithful, precise proof for a time-gated recency computation is the **RED→GREEN behavioral suite**, which pins the exact user-visible transitions the chip reflects. Each was **red on the pre-fix code and green after** (`packages/terminal-workspace/src/fold.test.ts`):

| Behavior the chip shows | Test |
| --- | --- |
| a stable session's output **advances** recency past the window (the freeze, fixed) | `bumps on a same-identity output tick once the throttle window has elapsed` (L179) |
| it **doesn't** churn every ~1 s (no write-noise regression) | `COALESCES output ticks inside the window to a single write` (L192) |
| a genuine **session start/finish** still bumps immediately (#1626 kept) | `an IDENTITY change still bumps INSIDE the window` (L228), `bumps when a finished agent is followed by a genuinely-new one` (L152) |
| **adopt across a restart keeps** the saved recency through the survivor's multi-emit settle (no false "just now") | `adopt idle, MULTI-emit settle: the whole survivor burst holds saved recency` (L280) |
| an adopted terminal that **resumes output** advances | `adopt then output: the settle holds, a later live tick past the window advances` (L294) |
| a survivor that ended in downtime → a **new agent's launch bumps** | `adopt, survivor gone during downtime, then a NEW agent launches → its start bumps` (L304) |

29 fold tests + 73 padi + 183 claude-code green on both platforms in CI. The review gauntlet independently exercised the hard paths — the architecture review reproduced the adopt multi-emit settle and survivor-gone regressions (both fixed), and codex added a summary-fetch ordering fence so late async metadata can't false-stamp the chip.

**If a live demo is wanted**, I can drive a pu-box kolu with a stable opencode session producing sustained output for ~90 s and record the dock chip staying fresh (`just now`) while the session identity never changes — say the word and I'll capture it.

srid:

The **x86_64-linux** lane ran on **`kolu-ci-7`** (`dev-x86-64-linux-04`) — commit `855d497d3`, **passed** ✓

- **Lane wall** (pipeline): **2m39s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 2m22s | ✓ |
| `ci::home-manager` | 2m3s | ✓ |
| `ci::unit` | 1m31s | ✓ |
| `ci::nix` | 28s | ✓ |
| `ci::smoke` | 24s | ✓ |
| `ci::biome` | 15s | ✓ |
| `ci::fmt` | 14s | ✓ |
| `_ci-setup` | 9s | ✓ |
| `ci::install` | 8s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/855d497/timings.jsonl`; pool state is a live `flock` probe.</sub>

PR #1736, opened 2026-07-09
Open PDFs in the Code tab

srid:
**PDF files now open inside the Code tab with the browser's native PDF viewer.** Previously `.pdf` files rode the same opaque-origin sandbox as HTML/SVG previews, and Chromium showed the broken-document icon instead of the document. PDFs now have their own binary preview family while HTML/SVG stay on the existing sandboxed path.

### Shape

| Preview family | Renderer path |
| --- | --- |
| PDF | `PdfRenderer`, an unsandboxed iframe for the browser-native viewer |
| HTML/SVG | Existing `BrowseIframeRenderer` sandbox with the artifact bridge |
| Image/video | Existing plain media renderers |

The shared `kolu-common/preview` classifier remains the source of truth, so the server still decides binary-vs-text once and the client dispatches every binary family by the same exported predicates. This keeps the change narrow: no override knob, no fallback renderer, and no new transport or persistence boundary.

### Design notes

- **Fail fast:** PDFs are an explicit preview family; an unclassified binary still lands in the visible unsupported renderer instead of silently trying the wrong appliance.
- **Volatility boundary:** the renderer is a small `@kolu/solid-fileview` appliance, not a new app-local subsystem; it hides no transport/reconnect/persistence volatility.
- **Reuse:** MIME serving stays backed by `@kolu/serve-dir`/`mrmime`, and the Code tab reuses the existing per-terminal file route and mtime cache key.

### Verified

- `just fmt`
- `just check`
- `just atlas::check-sync`
- `pnpm --filter kolu-common test:unit -- src/preview.test.ts`
- `pnpm --filter kolu-server test:unit -- src/iframePreviewRoute.test.ts`
- `pnpm --filter @kolu/solid-fileview typecheck`

### Try it locally

```sh
nix run github:juspay/kolu/feat/pdf-code-browser
```


srid:
## Evidence

PDF files now open inside the Code tab through the native browser PDF viewer path: the scenario selects `doc.pdf`, asserts the preview iframe is unsandboxed, confirms a loaded frame exists, and verifies the file route serves `application/pdf`.



▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=evidence-pdf-code-browser.mp4

PR #1723, opened 2026-07-08
feat: padi W7 — per-host client state by OWNERSHIP, not enumeration

srid:
Flips per-host client state from **enumeration** (a `HostView` record every new field must be *remembered* into — the "forgotten field" bug family: focus, splits, and the camera each stranded at app lifetime) to **ownership**: a per-host reactive owner where anything born inside is per-host **by construction**.

Design of record: [`docs/atlas/.../padi.mdx §W7`](docs/atlas/src/content/atlas/padi.mdx) + [`surface-map-101.mdx`](docs/atlas/src/content/atlas/surface-map-101.mdx). Ratified via /architecture-first-principles (P4) + the K1/K5/Option-B rulings.

## Three layers
1. **`@solid-primitives/keyed`** (added) — its `keyArray` already IS the retained-per-key-root primitive (retained across switches, disposed on membership exit). Not hand-rolled; `createRootPool` was evaluated + rejected (pool reuse ≠ per-key retention).
2. **`scopedByEntry`** in `@kolu/surface-map` — `scopedByEntry(map, active, build) → {active(), get(key)}`. Key set = entries membership ∩ activated; owners lazy on first activation, retained, disposed on membership exit. The map now also exports `codec` (the key encode the scope keys on). 7 unit tests (lazy / retain / dispose / lazy-again / removal-race / failed-member-gets-owner / owner-guard).
3. **kolu adoption** — `hostScope/hostScopes = scopedByEntry(padiMap, activeHost, …)` owns per-host `{view, camera, restore}`. **Deleted:** the `HostView` record, `ensureHost`, `useCanvasCameraSwap` (the snapshot/restore seam), `useCanvasViewport`'s three module-scope camera signals, `useSessionRestore`'s hand-rolled latch `Map`. The facades became **windows** onto `activeScope()` — the ~35 `...view` store consumers are untouched.

## The camera fix (the field symptom)
srid's live symptom — switch hosts → viewport pans to empty instead of the active terminal — **survived** the instance fix `013e369b1`. Ownership closes it **by construction**: the camera is retained per-host and co-mounted with the tiles it centers on, so there is no app-lifetime↔per-host bridge left to race the tile mount (the class a defer-guard narrows but never closes). `perHostCanvasCamera.test.ts` keeps the 1:1 decision pins; `TerminalCanvas` keeps a slim center-on-active effect reading the per-host camera.

## Boundary guard
[...]

PR #1737, opened 2026-07-09
fix(client): persist canvasMaximized per host — restore reload-survival

srid:
## What & why

The **W7 TIER A** refactor (#1723) turned `canvasMaximized` from a persisted `localStorage` flag into a plain **in-memory** per-host signal, on the rationale that a reload resetting to tiled was fine — *"matching the camera tier."*

But maximize isn't like the camera. The camera is *incidental* state (wherever you happened to pan/zoom); maximize is a **deliberate, discrete focus choice** ("I want this one agent full-screen"). A browser refresh — usually a crash, an update, or a reopened tab, **not** a request for a fresh overview — silently threw that choice away. It was also the odd one out: its two per-host siblings, `activityWindow` and `showSleeping`, *do* survive reload.

## The fix

Restore the **pre-W7 reload-survival** while keeping W7's per-host scoping:

- Back `canvasMaximized` with a per-host `boolPref` (**`kolu-canvasMaximized:<host>`**) instead of a plain signal — so it survives reload, keyed by host so two hosts don't collide (the pre-W7 flag was a single global `kolu-canvas-maximized`).
- A close-all `reset()` **still writes it back to tiled** (as the old global boolPref did) — a fullscreen posture over zero tiles is meaningless, so close-all floors it. It's the one persisted fact `createViewState` owns that a close-all still resets.
- The key is **evicted on host-leave**, matching `createHostPrefs`' dock-filter guard against unbounded `localStorage` growth for kolu's ephemeral remote boxes.

| | before W7 | after W7 (#1723) | this PR |
|---|---|---|---|
| storage | global `localStorage` | in-memory | **per-host `localStorage`** |
| scope | global | per-host | per-host |
| survives reload | ✅ | ❌ | ✅ |
| reset on close-all | ✅ | ✅ | ✅ |

## Tests

`perHostViewState.test.ts` gains a **red-first** repro — a fresh owner reads the persisted key back after a simulated reload — plus a per-host storage-key write assertion. Both failed on the in-memory signal; both pass now. Full client suite (636 tests) green, typecheck clean.


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `124767aa8be7`

### Applied (2)
- `lowy-1` Per-host-pref lifecycle (key format + eviction) re-hand-rolled instead of reusing createHostPrefs' seam — commit `470362197`
- `hickey-1` Per-host persisted-pref + eviction primitive hand-rolled a third time, bypassing the perHostKey seam — (uncommitted)

### Agreed — no change (2)
- `lowy-2` canvasMaximized's third volatility profile confirms its placement — record, don't move (packages/client/src/hostScope/createViewState.ts:85)
- `hickey-2` canvasMaximized persist-but-reset rationale duplicated across five comment blocks in one file (packages/client/src/hostScope/createViewState.ts:24)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `124767aa8be7`

### Round 1

**codex** — approved: `false`

The change is directionally sound, and the deliberate canvas/reset/reload decisions look coherent. The remaining problems are in the new per-host persistence helper's cleanup contract and in how the newly persisted canvas posture handles corrupt stored data.

Findings:
- `F1` · minor · open — `perHostPref` writes through `opts.storage` via `persistedPref`, but cleanup always removes from global `localStorage`. A caller using injected storage gets a stale per-host key left behind, and the cleanup may delete an unrelated global key with the same name. (packages/client/src/persistedPref.ts:189)
- `F2` · minor · open — The helper documents that it must run under a Solid owner, but it does not enforce that precondition. If it is called outside an owner, the eviction guarantee silently disappears, which undercuts the purpose of centralizing per-host cleanup. (packages/client/src/persistedPref.ts:177)
- `F3` · minor · open — The new persisted `canvasMaximized` value uses the default boolean invalid handler, which only `console.warn`s before falling back to `false`. A corrupt `kolu-canvasMaximized:<host>` entry becomes indistinguishable in the UI from a legitimate tiled posture, contrary to the repo's `caught-error-must-not-collapse-to-empty` rule. (packages/client/src/hostScope/createViewState.ts:88)


**claude** — Fixed the two per-host-cleanup helper defects (F1 storage split, F2 unenforced owner) with new falsifiable tests, and disputed F3 on repo precedent that a benign recoverable pref reset warns rather than toasts.

- `F1` **fixed** — Real latent inconsistency. `perHostPref` wrote through `opts.storage` (via `persistedPref` → `makePersisted`) but its `onCleanup` hard-coded `localStorage.removeItem(name)` (`persistedPref.ts:189`). With an injected storage that split leaves the fake's key orphaned AND wrongly deletes an unrelated global key of the same name. Fixed by resolving one backend up front — `const storage = opts.storage ?? localStorage` — passing it to `persistedPref` and evicting via `storage.removeItem(name)`, so write and cleanup are the same object by construction (`persistedPref.ts:194-211`). Added a falsifiable test (`persistedPref.test.ts`, "evicts from the SAME injected storage on dispose") that plants a `GLOBAL-DECOY` in real `localStorage` under the same composed name and asserts dispose evicts the fake's key while leaving the decoy untouched — red under the old cleanup, green now.

- `F2` **fixed** — Agreed and aligned with the repo's fail-fast rule. `onCleanup` outside a reactive owner silently no-ops, which would drop the evict-on-host-exit guarantee this helper exists to centralize — a silent degradation. Added a `getOwner()` guard at the top of `perHostPref` that throws with a clear message, mirroring `scopedByEntry` (`surface-map/src/scoped.ts:93`), which owns the very same per-key roots and throws for the same reason (`persistedPref.ts:184-192`). Added a test asserting `perHostPref` called at top-level (no owner) throws `/reactive owner/`. Current callers all run inside the per-host `scopedByEntry` build, so no regression.

- `F3` **disputed** — The premise ("a corrupt canvasMaximized must surface user-visibly, console.warn is insufficient") is refuted by this repo's OWN ratified calibration of `caught-error-must-not-collapse-to-empty`. The rule's target is a value whose malformed-vs-missing distinction MATTERS to the user; a corrupt boolean posture falling back to `false` (tiled) is a benign, recoverable reset to the correct default with nothing for the user to act on. The two canonical prefs of exactly this kind both use `console.warn` and explicitly reject a toast, each citing the same rule:
  - `wire.ts:181` (`activeHost`): "Resetting to LOCAL_HOST is benign, so a warn is the right level … caught-error-must-not-collapse."
  - `useDaemonStatus.ts:404` (`reattachAnnouncedAt`): "no user-facing toast for a recoverable reset."
  Only genuine data loss toasts (`useComments.ts:53`, a wiped comment queue). A corrupt `canvasMaximized` is the benign case, not the data-loss case. Its `console.warn` default already makes malformed distinguishable from first-run in the logs — the level the repo has settled on for benign prefs. Adding a toast would (a) be inconsistent with the two SIBLING per-host prefs `showSleeping` (same `boolDefaults` warn) and `activityWindow` (no `onInvalid` at all today) and with app-level `activeHost`, and (b) inject the exact toast-noise the repo deliberately avoids. Plumbing an unused `onInvalid` into `perHostBoolPref` with no caller would also be speculative generality against the repo's less-code philosophy. Left the code as-is (`createViewState.ts:88`, `persistedPref.ts:124-131`).

Also fixed a pre-existing biome warning (`createHostPrefs.ts:20`, from the prior lens-debate refactor 470362197) that was keeping `just check` red — `import { type Accessor, type Setter }` → `import type { … }`. Not a codex finding, but the static-check gate cannot be declared green while it fires. `just check` (tsc + biome `--error-on-warnings`) now passes; the 15 pref/per-host tests are green.

commit: `522dea088dcde2671f35b8161806a800ec619997`


### Round 2

**codex** — approved: `true`

F1 and F2 are fixed in the current tree with targeted tests. I accept CLAUDE's F3 dispute: for this specific boolean posture preference, the repo's existing calibration treats corrupt-but-benign preference resets as console-visible diagnostics rather than user-facing errors, and the cited sibling cases match that behavior. No new regression was introduced by this round's fixes.

Findings:
- `F1` · minor · resolved — `perHostPref` previously wrote through injected storage but cleaned up global `localStorage`, leaking injected-storage keys and possibly deleting unrelated global keys. (packages/client/src/persistedPref.ts:206)
- `F2` · minor · resolved — `perHostPref` previously only documented the reactive-owner requirement, so calling it outside an owner would silently drop the cleanup guarantee. (packages/client/src/persistedPref.ts:191)
- `F3` · minor · resolved — I previously argued that corrupt `canvasMaximized` storage needed a user-visible surface instead of the default boolean warning path. (packages/client/src/hostScope/createViewState.ts:88)

_codex on the rebuttal:_ F1 and F2 are verified fixed. For F3, I concede the finding: I applied `caught-error-must-not-collapse-to-empty` too broadly to a boolean UI posture preference. The cited precedents show the repo distinguishes benign recoverable preference resets from user data loss; `canvasMaximized` belongs with the warning-only cases, not the toast cases.

srid:

Ran Pass 1 (rule checklist) and Pass 2 (fact-check) over the gauntlet diff (`--no-elegance`, since `/simplify` already ran). Four items surfaced; dispositions below.

### Pass 1 — rules

| Finding | Rule | Disposition |
|---|---|---|
| `perHostBoolPref` forwards to `perHostPref` binding `parse: parseBool` | `no-thin-wrapper-functions` | **Kept, with rationale.** Inlining would export the module-private `parseBool` and duplicate the binding across its 2 call sites; the wrapper keeps `parseBool` private and names the boolean per-host seam (the lens debate explicitly created it). The rule's own remedy ("inline at the single call site") doesn't apply at 2 sites. |
| `defaultInvalidWarning` uses `console.warn` only | `caught-error-must-not-collapse-to-empty` | **Overruled, evidence-backed.** |
| …and warns, not errors | `errors-must-log-at-error` | **Overruled, evidence-backed.** |

Both `defaultInvalidWarning` flags re-raise exactly the point **codex F3 already conceded** after debate. The repo has an explicit, documented calibration for corrupt persisted prefs:

- `activeHost` (`wire.ts`) → `console.warn` — *"Resetting to LOCAL_HOST is benign, so a warn is the right level (matches reattachAnnouncedAt). caught-error-must-not-collapse."*
- `reattachAnnouncedAt` (`useDaemonStatus.ts`) → `console.warn` — *"no user-facing toast for a recoverable reset."*
- `comments` (`useComments.ts`) → `console.error` + `toast.error` — because a wiped queue **is** lost user data.

A per-host UI-preference reset (`canvasMaximized` / `showSleeping` / `activityWindow`) is a benign recoverable reset → it belongs with the first group. I refined the docstring (commit `b781dc8b4`) to cite that precedent explicitly rather than ambiguously invoking the rule.

### Pass 2 — fact-check

**Overstated eviction guarantee (real).** `perHostPref`'s docstring claimed the unbounded-`localStorage`-growth failure was *"unspellable."* It isn't: eviction rides **owner disposal**, and `scopedByEntry` builds a host's owner **lazily on first activation**. A host removed **cold** — never activated in the current page session (reload, then ✕ a remote chip without switching to it) — has no owner to dispose, so keys it wrote in an earlier session are never swept.

- This coupling **predates this PR** (the old hand-rolled `createHostPrefs` cleanup had the identical gap); the PR's marginal contribution is one more key (`canvasMaximized`, formerly in-memory) and the overstated centralized claim.
- **Actioned** (`b781dc8b4`, comment-only): corrected the docstring to state the accurate, narrower bound and recorded the residual leak as a follow-up. **The full fix is membership-seam eviction at `hosts.remove` (independent of owner lifecycle) — deferred to its own change, not solved here.**

Verified-and-clear (Pass 2): invalid-vs-absent are distinguished; `encodeHostKey` is injective; the F1 storage-backend + F2 owner-guard codex fixes are present with falsifiable tests; no cross-host key race.

PR #1744, opened 2026-07-10
feat(code-police): no-overloaded-null rule + L26 null/sentinel sweep

srid:
## L26 — kill the overloaded null/sentinel hacks (fix + lint, one PR)

L26 was born from the surface-hosting-simplify debate: `SurfaceIdentity` once used `identity(): T | null` (null = no link) **and** `baked: T | null` (null = no build) **and** `commit: string | null` (null conflating a dev tree with a real commit) — one nullable answering several different questions, a reader guessing which. srid's rule: **a nullable or sentinel that carries more than one meaning is a defect** — replace it with a discriminated `{ kind }` sum so every state is named and impossible states can't be constructed.

This PR closes L26. The twist: **the sweep found nothing left to fix.**

### The sweep (SWEEP + FIX)

A five-agent, repo-wide inventory — prioritizing padi, the server re-serve/session surfaces, `surface-daemon-supervisor`, and the client — read **~40 nullable/sentinel/optional sites in context** (not just grepped) and classified each: honest single-meaning absence = survivor · overloaded = fix.

**The finding is honest and uniform: zero overloaded nulls remain anywhere in the tree.** The class L26 was born to kill was already retired by the very surface-hosting + padi refactors that *spawned* this entry. The code now narrates its own prior de-overloadings — `ProcessMemory`'s three-arm union (was `number | null` conflating "no process" with "read failed"), `KavalProbe`'s honest-null, `RestoreTarget`, `Known<T>`, `readPadiMemoryOnce`'s value·absent·error split, and `SurfaceIdentity` itself.

No `@kolu/surface*` hit surfaced either — that family is where the exemplar was minted and is already clean — so there are **no drishti-gated proposals** to carry.

This is **not a blanket anti-null crusade**: a single-meaning "absent" is fine and was left alone. The near-misses (declared type or comment *looks* two-meaning) were adjudicated and recorded as survivors:

| Site | Looks like | Verdict |
| --- | --- | --- |
| `server/…/daemonInventory.ts` `boundHost: string \| null` | "SINGLE source for both is-remote? and the label" | **Survivor** — `null` = local (one meaning); folded to the `DaemonBinding` `{kind}` sum one line into its sole consumer, zero divergent reads. Pushing the sum to the seam only relocates the `KOLU_PADI_HOST` null check (the async `localScan` must stay in the consumer) — no illegal state removed. |
| `anyagent/…/agent-adapter.ts` `contextTokens: number \| null` | comment names two origins (no-telemetry vs not-yet) | **Survivor** — two *causes*, one *effect* ("—") at every read site; no reader disambiguates. |
| `padi/…/sessionRestore.ts` `resumeIds?: string[]` | `undefined` = all, `[]` = none | **Survivor** — deliberate tri-state; the *absent* value answers one question. |
| `shell-quote/…/index.ts` `token: string \| null` | `null` vs `""` | **Survivor** — the *correct* two-state encoding to round-trip `shellJoin`; collapsing them is the bug. |

### The guardrail (the deliverable)

Since the FIX half was satisfied by prior work, the shipping deliverable is the lint that stops the class from regressing — the "guardrail that stops it regressing" this entry always framed as its second half.

`.agency/code-police.md` gains **`no-overloaded-null`**, matching the file's existing rule format. It flags, on a PR diff:

- **(a)** a `T | null` / `T | undefined` whose absence is tested for **two distinct meanings** (≥2 divergent read sites, or a comment admitting two absence-reasons that a consumer should tell apart);
- **(b)** a string/number **sentinel** (`""`, `-1`, `"none"`) standing for a state that has a real name elsewhere;
- **(c)** an **optional** field a consumer treats as an on/off **state**.

Guidance prefers a `{ kind }` union and cites `SurfaceIdentity` as the canonical exemplar (with its before/after). It ships with a **seeded two-meaning fixture** (the `readMemory` `number | null` conflating no-client with read-error), a **quiet single-meaning counterexample** (`find()` → `undefined`), and an `_Allowed_` clause that records the exact **two-causes / one-effect** boundary the sweep surfaced — so the rule and the inventory agree, and it won't fire on the honest survivors above.

### Record

`docs/atlas/src/content/atlas/padi-cleanup.mdx` L26 entry: status → **implemented**, outcome narrated, survivors recorded with verdicts. Atlas `dist/` rebuilt and staged; `atlas::check-sync` green.

---

**Design philosophy.** Fail-fast / no-fallbacks: the rule pushes each absence-state onto a named union arm so a reader must branch — no silent degrade-to-empty. Reuse the source of truth: the guardrail lives in the existing code-police ruleset (`.agency/code-police.md`), not a parallel mechanism. No override knobs added.

**Scope note.** `packages/git`, `packages/github`, `packages/pulam*` named in the ledger brief do not exist as separate packages in this repo; git/github/forge concerns live under the integration packages, which were swept.


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `2e64b62bc003`

### Agreed — no change (3)
- `lowy-1` Survivor verdicts duplicated across the durable rule and the point-in-time ledger (.agency/code-police.md:470 (the _Allowed_ clause) ↔ docs/atlas/src/content/atlas/padi-cleanup.mdx:190 (Survivors list))
- `hickey-1` Survivor verdicts duplicated across the rule and the ledger, held in sync by a cross-reference (.agency/code-police.md:471 (and padi-cleanup.mdx:192-195))
- `hickey-2` SurfaceIdentity exemplar and the rule's (a)/(b)/(c) criteria restated in both documents (padi-cleanup.mdx:181 (and :186), mirroring .agency/code-police.md:420-433)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 round(s) · codex reviewed at `xhigh` reasoning effort · base `2e64b62bc003`
### Round 1

**codex** — approved: `false`

The branch is close, but I would not approve yet. The new rule captures the right architectural smell, but its actionable trigger is narrower than its own bad fixture, and the Atlas ledger has a couple of consistency gaps that will mislead the next reader.

Findings:
- `F1` · major · open — The rule's formal trigger for nullable overloads is read-side: divergent absence handling at multiple read sites, or comments/names admitting multiple absence reasons. The seeded bad fixture at lines 439-448 is actually a producer-side collapse: `no client` and `read failed` both return `null`, and a single consumer then hides the difference. A real PR can contain that same defect without divergent read sites or helpful comments, so the guardrail can miss the exact class it is teaching reviewers to catch. (.agency/code-police.md:416)
- `F2` · minor · open — The SurfaceIdentity exemplar says the old shape had `commit: string | null`, but the same rule's sentinel clause says `commit === ""`, the inline before-comment says `commit ""`, and the current code maps an empty commit string to `BuildCommit.dev`. The canonical example is internally inconsistent about whether the old smell was null or an empty-string sentinel. (.agency/code-police.md:420)
- `F3` · minor · open — The L26 section now says `Status: implemented — DRAFT PR`, but the ledger index still lists L26 under `Ready now`, and the batch plan at line 26 still describes it as future solo work. That makes the living ledger contradict itself and can cause duplicate dispatch of work that this PR says is already done. (docs/atlas/src/content/atlas/padi-cleanup.mdx:35)


**claude** — Fixed all three findings: closed the producer-side gap in the rule's trigger, aligned the `commit` exemplar to its `""` sentinel throughout, and reconciled the ledger index/batch-plan with the new "in review" status.

- `F1` **fixed** — Codex is right, and it's the load-bearing catch. As written, criterion (a) keyed only on read-side divergence or a type-comment, so the seeded _Bad_ fixture (`.agency/code-police.md:439-448`) — a *producer* collapse (`if (!client) return null` beside `catch { return null }`, one consumer rendering both as "—") — would not have tripped its own trigger, and worse, the `_Allowed_` "two causes fold to one effect at every read site" clause would have mis-classified it as a *survivor* (single consumer, uniform rendering). The real discriminator is normative, not mechanical: does a cause *deserve* distinct handling (a swallowed error must surface) vs two honest absences that genuinely fold. I rewrote criterion (a) (`.agency/code-police.md:416`) into three explicit signals — (i) producer maps distinct causes to the same absent/sentinel value where one is a caught error/fault; (ii) divergent read sites; (iii) a comment/name admitting two absence-reasons — and made the normative test explicit ("honest absence vs caught error is the archetype"). I then sharpened the `_Allowed_` contextTokens clause (`.agency/code-police.md:471`) so it names the exact boundary against (i): two honest "no datum yet" causes = survivor; honest-absence-plus-swallowed-error = hit. Verified the four recorded survivors (boundHost, contextTokens, resumeIds, shell-quote token) still read as survivors under (i): none collapses a *fault* onto an honest absence, and shell-quote/resumeIds use *distinct* values (`null` vs `""`, `undefined` vs `[]`), not the same one.
- `F2` **fixed** — Codex is right about the internal inconsistency. Criterion (b) (`:417`), the inline before-comment (`:427`: `commit "" → dev tree?`), and the shipped code (`packages/surface/src/identity.ts:117` `buildCommit`, which maps `commitHash === ""` to `{ kind: "dev" }`; the schema comment at `:33` even says "a SUM, never `string | null`") all frame the old smell as the empty-string sentinel — but the guidance prose at `:420` and the Atlas narrative said `commit: string | null`. The `""` framing is the accurate one (a git commit env is `""` off-nix / on a dirty tree, never null), and it also makes the exemplar teach criterion (b) coherently. Changed `.agency/code-police.md:420` to `commit: string` (`""` conflating a dev tree with a real commit) — "two nulls and a `""` sentinel" — and `docs/atlas/src/content/atlas/padi-cleanup.mdx:181` to match.
- `F3` **fixed** — Codex is right that the ledger contradicted itself. The L26 section now reads `Status: implemented — DRAFT PR` while the status index still listed L26 under "Ready now" (`:35`) and the batch plan still described it as future solo work (`:26`). Removed L26 from the "Ready now" line, added a distinct **"In review (implemented, draft PR)"** bucket to the status index, and annotated the batch-plan "Solo by nature" entry as *(in review — draft PR; the guardrail shipped, the sweep found the class already retired)*. Left the bird's-eye SVG untouched (hand-authored diagram, not flagged, and its middle/solo lane placement is still defensible for an in-review entry). Rebuilt `docs/atlas/dist/` (`just atlas::build`); `just atlas::check-sync` unit tests pass (the "out of sync" line is only the expected uncommitted-dist delta, resolved by committing dist in this change).

commit: `c6ce90c832fa9d8dc940028cf294cacc19f79172`


### Round 2

**codex** — approved: `false`

F1 and F3 are adequately addressed in the current tree. F2 is mostly addressed, but one stale line remains in the canonical rule fixture, so the exemplar still contradicts itself about whether the old shape had three nulls or two nulls plus an empty-string sentinel.

Findings:
- `F1` · major · resolved — The original rule missed producer-side collapse of distinct causes into the same absent value, so its own bad fixture would not necessarily trigger the rule. (.agency/code-police.md:416)
- `F2` · minor · open — The exemplar is still internally inconsistent. Line 420 now correctly says SurfaceIdentity had two nulls plus a `""` sentinel, but the snippet header still says `Before — three nulls`, even though the snippet only has two nullable fields and the commit arm is documented as `commit ""`. (.agency/code-police.md:423)
- `F3` · minor · resolved — The L26 section said implemented/in review while the ledger index still listed it as ready future work. (docs/atlas/src/content/atlas/padi-cleanup.mdx:35)

_codex on the rebuttal:_ F1: accepted; the new producer-collapse trigger fixes the load-bearing gap, and the Allowed clause now preserves the intended survivor boundary. F2: partially fixed, but not fully resolved because `.agency/code-police.md:423` still says `three nulls`; that is the same inconsistency in the canonical exemplar, just one line lower. F3: accepted; the status index and batch-plan contradiction is resolved.


**claude** — Fixed the one stale line codex flagged; the exemplar header now matches the corrected "two nulls and a sentinel" framing throughout.

- `F1` **fixed** (already resolved round 1, codex accepted) — producer-collapse trigger `(i)` in criterion (a) at `.agency/code-police.md:416` stands; no change this round.
- `F2` **fixed** — Codex is right: round 1 corrected the prose at `:420` and the inline before-comments (`:427`) to the "two nulls and a `""` sentinel" framing, but the snippet header at `.agency/code-police.md:423` still read `Before — three nulls`, contradicting it (the snippet has two nullable fields plus the `commit ""` arm). Changed the header to `Before — two nulls and a "" sentinel, a reader guessing which question each answers:` so the canonical exemplar is now internally consistent — prose, header, and inline comments all agree the old shape was two nulls plus an empty-string sentinel.
- `F3` **fixed** (already resolved round 1, codex accepted) — ledger index/batch-plan reconciliation in `padi-cleanup.mdx` stands; no change this round.

commit: `3ca300502c904708de60f46085a47e865aa95651`


### Round 3

**codex** — approved: `true`

All previously raised findings are resolved in the current tree. The no-overloaded-null rule now covers producer-side absence/error collapse, the SurfaceIdentity exemplar consistently describes two nullable fields plus an empty-string sentinel, and the Atlas ledger no longer lists L26 as ready future work while its section says implemented/in review.

Findings:
- `F1` · major · resolved — The original rule missed producer-side collapse of distinct causes into the same absent value, so its own bad fixture would not necessarily trigger the rule. (.agency/code-police.md:416)
- `F2` · minor · resolved — The exemplar was internally inconsistent about whether the old SurfaceIdentity shape had three nulls or two nulls plus an empty-string sentinel. (.agency/code-police.md:423)
- `F3` · minor · resolved — The L26 section said implemented/in review while the ledger index still listed it as ready future work. (docs/atlas/src/content/atlas/padi-cleanup.mdx:35)

_codex on the rebuttal:_ F1 remains accepted as fixed. F2 is now fully resolved: the stale `three nulls` header was updated to match the corrected sentinel framing. F3 remains accepted as fixed.

srid:

Ran with `--no-elegance` (the simplify track already covered this tree). Pure-docs diff: the new `no-overloaded-null` rule + the L26 ledger entry.

**Pass 1 — rule-format checklist:** conforms to the file's conventions (kebab ID, `### ` heading, `_Rationale_:` close, fenced blocks balanced, `.mdx` status-index + survivor bullets match the ledger's style). Two cosmetic nits:
- `Codified from` → **`Codified after`** to match the other 6 provenance citations — **fixed** (`afefb9d9`).
- `_Quiet_` example label has no in-file precedent — **kept deliberately**: the L26 entry mandates a *quiet single-meaning counterexample* paired with the seeded two-meaning fixture, and the label echoes that spec wording; the Bad/Quiet pair is the rule's positive/negative test.

**Pass 2 — fact-check:** all load-bearing claims verified against the current tree.
- `SurfaceIdentity` **after**-shape matches `packages/surface/src/identity.ts` byte-for-byte (`disconnected | anonymous | identified`, `BuildCommit = commit | dev`). The **before**-shape is documented design-debate history (the file was introduced whole in #1705, so it isn't git-diffable) — no contradicting evidence.
- All four survivor sites confirmed verbatim: `boundHost` (`daemonInventory.ts:55`, one read site), `contextTokens`/`startedAt` (`agent-adapter.ts` — every reader treats null uniformly), `resumeIds` (`sessionRestore.ts` — `undefined`=all, `[]`=none), `token` (`shell-quote/index.ts:83`).
- The named already-discriminated sums (`ProcessRss`, `KavalPresence`, `SessionState`/`ClosedInfo`, `SocketProbe`, `TranscriptWatching`) are all real shipped unions.

No correctness or rule-conformance blockers. Findings actioned; nothing outstanding.

PR #1714, opened 2026-07-07
feat: @kolu/surface-map + the KOLU_PADI_HOST-gated multi-host canvas

srid:
`@kolu/surface-map` is a **keyed map of remote surfaces** — one entry surface, typed once, served N times, keyed at runtime — and this PR is **two honest halves at once**: it builds that framework *and* ships the product surface it unlocks. A browser tab now **live-switches which machine's terminals it views** — no reload — over a warm pool of per-host padi bindings, reaching a host's cells through `padiMap.entry("zest").cells.daemonStatus.use(…)`. Supersedes the closed #1708 (clean-room redesign; the ratified Atlas note `remote-surfaces.mdx` is the sole spec). First of three serial PRs — this one, then a drishti adoption (#90), then a mechanical odu rename (#40); both companions' pins are already bumped to this HEAD and green.

## The shape

`KOLU_PADI_HOST` stops being an all-or-nothing "the whole canvas becomes that box" knob and becomes a **comma-seed list** for a warm pool. `local` is always the implicit, **unremovable** default, so **env-unset is a 1-member map — pixel-identical to today's single-host canvas** (the multi-host UI is gated off and renders *nothing*). No dual code path: the map is *always* the wire shape; N=1 is the degenerate case.

## Half 1 — the framework (`@kolu/surface-map` + `@kolu/surface{,-remote}`)

- `defineSurfaceMap(keySchema, entry)` + the uniform fold envelope `{ mapKey, input }` on every proc; `entries` is the one membership authority (a `Collection<Key, EntryStatus>`), status a **projection**, never a second writer. A void-input member folds to **no input key** (schema declares none), so the fold is zod-version-independent — closing the drishti-fleet incident where zod ≥4.3.7's stricter missing-key check silently broke every void fold.
- Client `connectSurfaceMap`: `entry(key)` (pure lens, total, cached) + `useEntry(acc)` (Solid, owns swap disposal, re-keys on switch); existence is always a **value** (`state()`), never a nullable `entry()`. `Entry.rpc` is the per-key procedure client (folds `{mapKey}` into every call; absent-key → typed `MAP_KEY_UNKNOWN`).
- `@kolu/surface`: ref-counted base subscription **dedup** (N views of one `(proc, input)` → one upstream sub).
[...]

PR #1747, opened 2026-07-10
padi cleanup campaigns 2 & 3: the server re-cut and the client wire/state sweep

srid:
Two padi-cleanup campaigns land together on one branch, **all internal, UX byte-identical** (the track's no-pixel-delta invariant) — the e2e suite and the package seal are the net. Campaign 2 reshapes `packages/server/src` around its real volatility; campaign 3 sweeps the switch-era client wire and names its last hand-rolled state machine. Every commit is one coherent move; the opening L27 commit is strictly moves-only so rename detection reviews it on its own.

The ledger governing this work lives on the docs branch (`docs/atlas/src/content/atlas/padi-cleanup.mdx`) — its campaign 2 & 3 sections and the L-entries are the ratified scope. **L24 (padi-tui --host) is deliberately excluded** (a separate feature PR), and the ledger update rides the docs PR, not this one.

### Campaign 2 — the padi/server re-cut

`server/src` was ~20 flat files while the rest of the repo nests by concern. The **padi hosting arm co-varies on every remote/switch phase** yet sat flat beside the thin HTTP shell — that co-variation is the volatility a subdirectory should encapsulate.

| Commit | L-entry | What |
| --- | --- | --- |
| `refactor(server)` | **L27** | Home the arm (both binders + convergence + session + link + the pool satellites reServeEviction/supervisorClaim/daemonInventory) under `padi/`. Strictly moves-only — the sole edits are the mechanical import-path/allowlist consequences of the move. |
| `refactor(padi)` ×2 | **L17** | Split the UI-chrome vocabulary out of the ~960-line `vocab.ts` into a sibling `chromeVocab.ts` (same export surface, `surface.ts` re-exports both). Home `UPLOAD_VIDEO_EXTENSIONS` in `@kolu/padi` and derive kolu-common's `VIDEO_EXTENSIONS` from it — drift becomes unspellable, so the drift-guard test is deleted. |
| `test(server)` | **L13** | Close the seal's test-file flank. |

> **The L27 sketch named `hostPool`/`perHostKoluCells` — they don't exist as files.** The pool is `buildRemotePool` from `@kolu/surface-remote`, instantiated inline in `index.ts`. The cut follows the real tree (Explore-verified cross-imports), not the sketch, and honors the hickey guard: no `misc/`/`util/` bucket — the true leaves (`hostname`·`tls`·`state`·`log`·`memorySampler`·`pwaIdentity`) stay top-level.

**L13 decision — widen-and-walk, not narrow-exports.** The forward `@kolu/padi` boundary (arm b) walks only production reachable from `index.ts`, so a test could reach a deep `@kolu/padi/src/...` unnoticed. The flank subpaths tests actually use — `stateRoot` (padiBinding.test) and `transcript` (exportTranscriptHtml.test) — are *legitimately published* entry points other packages consume (the `tests` pkg uses stateRoot; the `client` uses transcript), so narrowing padi's exports would break real consumers. And production is *deliberately* kept off `stateRoot` (`supervisorClaim` routes through `/assembly` on purpose), so widening the production allowlist to include it would undo that boundary. So: keep production's door tight (arm b unchanged), give tests the full *published* set via a new **arm (f)** (derived from `package.json` exports, drift-proof), and assert `ALLOWED_PADI ⊆ published` — the allowlist and the exports map now agree on purpose.

### Campaign 3 — the client wire/state sweep

**L11 — the switch (#1714) landed clean.** The inventory (seeded by the #1714 diff, every `createSharedRoot` consumer, and a thorough marker sweep) found `wire.ts` and the two hostScope owners are the *finished* per-binding shape: no import-time module-const subscription, no global-vs-map fork, no dual path. The one genuine leftover was `padiRpcOf`'s doc calling its fixed-host caller "currently hypothetical" — stale, since `hostScope/createViewState` is exactly that caller. Fixed the doc.

> **Three `host-INDEPENDENT-today` readouts** (`padiLinkState`, `boundPadiConvergence`, `daemonScanBoundHost`) are blocked on a **server-side** per-host `padiLink`/convergence wire member `padiMap` doesn't yet serve. Finishing them is a *feature* (per-host padi-link serving) that would change what the Padi chip shows for remote hosts — not byte-identical, so explicitly out of this cleanup's scope. The authoritative tracking list stays the classification table in `padi/PadiInfoDialog.tsx`.

**L18 — name the last hand-rolled state machine.** `useSessionRestore` drove its transitions through raw `latch.state = "…"` writes scattered across the hydration effect — including one *out-of-band* write from `handleRestoreSession` (the exact smell L18 named). Now a `HydrationPhase` (`pending → decided → seeded`) with named transitions `markDecided()` / `markSeeded()` / `reseedForRestore()`; `markDecided` is idempotent so a stray call can never regress a seeded host. A new unit test pins the transition table; the existing restore/reload/empty test passes unchanged.

> **Two deliberate drifts from the ledger's L18 sketch, flagged for review:** (1) the ledger's `decided/viewSeeded` **boolean pair** was already gone — W7's per-host scope work had collapsed it into a 3-state enum; this PR finishes the *named-transitions* half. (2) I kept **3 states + live empty-vs-restore derivation** rather than the sketched 4-state (`decided-empty | decided-restore`): the outcome is re-read from `store.terminalIds()` at seed time, so baking it into the phase would only create a stale distinction to reconcile. And I kept it a **non-reactive record** (not a reactive signal) — the phase is a gate the effect reads, not a dependency that should re-arm it.

### L8 — hygiene (rides both campaigns)

Swept stale `pulam` references (the packages are deleted) in the files these campaigns touched, repointed the six inbound comment refs to the L27-moved files, and fixed the seal header's stale entry-point count. *The full 33-file pulam residue is deferred per the ledger's hygiene rule — it rides the last campaign's tail commit, not this PR.*

### Reviewer notes

- **Byte-identical UX throughout.** No user-visible change, so no changelog entry; website docs don't reference these file-level internals.
- The seal (`packages/server/src/seal.test.ts`) is campaign 2's own net — it stays green with the arm under `padi/` and gains arm (f).


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `a1d4a9cc2f61`

Independent findings: lowy=4, hickey=1

### Applied (1)
- `hickey-1` HydrationPhase transition table is enforced asymmetrically — the illegal pending→seeded skip stays representable — commit `42b200fa0`

### Agreed — no change (4)
- `lowy-1` Video-playback rationale is split from the list it governs (packages/common/src/preview.ts:63)
- `lowy-2` VIDEO_EXTENSIONS widened from literal tuple to readonly string[] (packages/common/src/preview.ts:75)
- `lowy-3` HostRestoreLatch write-side is verb-encapsulated but read-side exposes the raw phase enum (packages/client/src/terminal/useSessionRestore.ts:84)
- `lowy-4` vocab.ts (stable persistence axis) depends on chromeVocab.ts (more UI-volatile axis) (packages/padi/src/vocab.ts:34)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `a1d4a9cc2f61`

### Round 1

**codex** — approved: `false`

The branch is mostly a mechanical split/move plus the session-restore state refactor, and most references appear updated correctly. I found one real stale relative path introduced by moving the padi binding module; it breaks source-mode padi launch when the packaged wrapper is not supplied, which is a core dev/e2e path.

Findings:
- `F1` · blocking · open — After moving `padiBinding.ts` down into `packages/server/src/padi/`, the from-source padi entrypoint still uses `../../padi/src/bin.ts`. From the new file location that resolves to `packages/server/padi/src/bin.ts`, which does not exist, instead of `packages/padi/src/bin.ts`. When `KOLU_PADI_BIN` is unset, `resolvePadiLaunch()`/`ensurePadiBinding()` will try to spawn padi through this bad path and fail in local source/dev/e2e mode. (packages/server/src/padi/padiBinding.ts:210)


**claude** — Conceded F1: the L27 move added a directory level that the from-source padi entrypoint's relative hop never absorbed; fixed the path and pinned it with a test.

- `F1` **fixed** — Confirmed. L27 moved `padiBinding.ts` from `packages/server/src/` down into `packages/server/src/padi/`, but the from-source launch URL stayed `../../padi/src/bin.ts`. From the new location that resolves to `packages/server/padi/src/bin.ts` (nonexistent) instead of `packages/padi/src/bin.ts` — the code's own comment on the line above documents the intended target, so the URL and the comment disagreed. When `KOLU_PADI_BIN` is unset (dev/e2e/source mode) `resolvePadiLaunch` would hand `--import tsx <phantom>` to `process.execPath` and the spawn fails. Fixed `packages/server/src/padi/padiBinding.ts:210` — the hop is now `../../../padi/src/bin.ts` (three ups: `padi/` → `src/` → `server/` → `packages/`). Added a regression test `packages/server/src/padi/padiBinding.test.ts:718` ("from-source entrypoint … resolves to a REAL file") that unsets `KOLU_PADI_BIN`, calls `resolvePadiLaunch`, and asserts the resolved bin path ends with `packages/padi/src/bin.ts` AND `existsSync`es — so a future skew of this hop fails loudly rather than only at spawn time. `just check` (typecheck + biome) green; new test passes.

commit: `3f9f5cbd19ac322a6213d7bff631372d97b12c8b`


### Round 2

**codex** — approved: `true`

F1 is resolved. I rechecked the current tree, verified the source-mode URL now resolves from `packages/server/src/padi/padiBinding.ts` to the real `packages/padi/src/bin.ts`, and confirmed the added regression test asserts both the expected path suffix and `existsSync`. I did not reopen the rest of the diff per the convergence instructions.

Findings:
- `F1` · blocking · resolved — Previously, after moving `padiBinding.ts` into `packages/server/src/padi/`, the from-source padi entrypoint still used `../../padi/src/bin.ts`, resolving to nonexistent `packages/server/padi/src/bin.ts` and breaking dev/e2e/source-mode launch when `KOLU_PADI_BIN` was unset. The current code now uses `../../../padi/src/bin.ts`, which resolves to the real `packages/padi/src/bin.ts`. (packages/server/src/padi/padiBinding.ts:210)

_codex on the rebuttal:_ No dispute to answer. CLAUDE conceded F1 and fixed it; I verified the corrected path computation and the new regression coverage.

srid:

Scope: `git diff a1d4a9cc2...HEAD` (padi-cleanup campaigns 2&3). Elegance pass skipped (`--no-elegance` — `/simplify` already ran this gauntlet).

| Pass | Found | Outcome |
| --- | --- | --- |
| Rules | 1 | fixed |
| Fact-check | 0 | clean |

**Rules — the one hit.** `packages/server/src/padi/padiBinding.test.ts` (the F1 regression test the codex debate added): `binTs` is `string | undefined` from array indexing, and the test used `existsSync(binTs as string)` — an `as`-cast past the type system, which `no-untyped-escape-hatches` forbids for exactly this "narrowing-but-not-quite in tests" case. Replaced with the prescribed narrowing throw:

```ts
if (binTs === undefined) throw new Error("expected --import to be followed by a bin.ts path");
expect(binTs.endsWith("packages/padi/src/bin.ts")).toBe(true);
expect(existsSync(binTs)).toBe(true);
```

Committed `92fcd81e2`. `vitest` 24/24, `tsc` clean.

**Fact-check — clean.** No silent error-swallow, no inaccurate fallback, no wishful thinking. Spot-verified: the session-restore try/catch blocks are untouched (only naming changed); `preview.test.ts`'s removed 15 lines were the now-redundant drift-guard, not lost coverage; every `server/src/{…}` → `server/src/padi/{…}` move is 99–100% similar (pure motion) and the one substantive relative-path change (`padiBinding.ts` → `bin.ts`) resolves correctly and is now regression-guarded; `seal.test.ts`'s additions are additive boundary-enforcement (arm f).

srid:
## Evidence



▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=evidence-restore-hydrationphase.mp4

This records the e2e scenario **"Restore previous session from empty state"** (`packages/tests/features/session-restore.feature`): empty canvas → the restore card offers the saved session → clicking Restore re-spawns the terminals and seeds the view.

It exercises L18's `HydrationPhase` `pending → decided → seeded` machine — the decision runs on the empty boot, seeding runs when the restored terminals arrive — and confirms the named-transition refactor is byte-identical: the flow renders and behaves exactly as it did on master.

The other changes in this PR (L27 server padi/ re-cut, L17 chrome-vocab split + video-ext homing, L13 seal test arm, L11 doc fix) have no on-screen surface — pure internal refactors — so no visual artifact applies to them.

srid:

The **x86_64-linux** lane ran on **`kolu-ci-5`** (`idliv2-02`) — commit `92fcd81e2`, **passed** ✓

- **Lane wall** (pipeline): **3m36s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 3m12s | ✓ |
| `ci::home-manager` | 2m26s | ✓ |
| `ci::unit` | 1m39s | ✓ |
| `ci::smoke` | 58s | ✓ |
| `ci::nix` | 56s | ✓ |
| `ci::pnpm-hash-fresh` | 44s | ✓ |
| `ci::biome` | 36s | ✓ |
| `ci::fmt` | 35s | ✓ |
| `_ci-setup` | 14s | ✓ |
| `ci::install` | 10s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/92fcd81/timings.jsonl`; pool state is a live `flock` probe.</sub>

PR #959, opened 2026-05-23
Make the right panel follow the terminal, not the workspace

srid:
**Switching terminals no longer flips the right panel's active tab, code sub-mode, or selected file out from under the user.** The panel's state was a single global preference: opening a PR review in one terminal and a build log in another forced both into whichever was last set. Now the panel reads what each terminal was doing, restored on terminal switch.

### The cut

The right-panel state was split along its volatility seam — not "the right panel" as one concept, but two:

| Field | What the user sees | Scope |
|---|---|---|
| `collapsed` | Whether the panel is showing at all | **Global** (workspace chrome) |
| `size` | Horizontal width drag-handle | **Global** |
| `codeTabTreeSize` | Vertical split between file tree and diff/content | **Global** |
| `activeTab` | Inspector vs Code | **Per terminal** |
| `codeMode` | Code tab's local / branch / browse sub-mode | **Per terminal** |
| `selectedFileByMode` | Which file is highlighted in the tree | **Per terminal**, per-mode within terminal |

The three global fields stay on `preferences.rightPanel` unchanged. The three per-terminal fields move to a new `TerminalMetadata.rightPanel` field, end-to-end mirroring the existing `subPanel` pattern — server stores it in the persisted terminal record, client keeps an in-memory store keyed by terminal id with fire-and-forget reportToServer, session restore seeds via `seedPanel`.

### Mechanism

```
client useRightPanel ──┐
  (per-terminal store) │
                       ├─▶ client.terminal.setRightPanel ─▶ server setRightPanelState
  preferences.rightPanel│       (deep-equality gated)         (writes TerminalMetadata.rightPanel)
  (global chrome) ──────┘                                                  │
                                                                           ▼
                                                              survives session restore
```

`useRightPanel()` is parameterless — it resolves the active terminal from `useTerminalStore` internally. _Callers don't thread the terminal id; "the right panel" always means "the active terminal's right panel"._

CodeTab's `kolu-codetab-selected-files` localStorage path is retired; the new `selectedFileByMode` lives on the per-terminal record with per-mode keying preserved so flipping local↔branch↔browse within a terminal still restores each mode's last file.

### Bonus side-effects of the refactor

- **`RightPanelTabKind` dispatch is now compile-enforced end-to-end** — the body renderer iterates the kind tuple and uses `match(kind).exhaustive()`, so adding a new tab variant is a three-place change the compiler catches (previously the body renderer's pairwise `=== "inspector"` / `=== "code"` checks would have silently fallen through).
[...]

PR #1753, opened 2026-07-10
Right panel state follows the terminal (collapse is per-terminal now)

srid:
The right panel's expanded/collapsed bit was the last posture field still stuck on the global `preferences.rightPanel` — so collapsing it on one terminal collapsed it for every terminal, on every host. **`collapsed` now lives per-terminal** (on `TerminalMetadata.rightPanel`, alongside the tab / code-mode / selected-file that moved there at [#959](https://github.com/juspay/kolu/pull/959)), reported to the server and restored on reload exactly like the active tab. A PR-review terminal keeps its panel open while a build-log terminal keeps its closed; switching tiles (or hosts — terminals belong to hosts) restores each one's panel.

This completes #959's "make the right panel follow the terminal" — that PR moved the *content* fields per-terminal but deliberately left `collapsed` global.

### What moved, what stayed

| Field | Before | After |
| --- | --- | --- |
| `collapsed` (panel showing?) | global pref | **per-terminal** (follows the terminal) |
| `activeTab` / `codeMode` / selected file | per-terminal (since #959) | unchanged |
| panel **width** (`size`) | global pref | unchanged — viewer taste, stays global |
| Code-tab tree/content split (`codeTabTreeSize`) | global pref | unchanged — viewer taste, stays global |

Keeping width + split global was a deliberate call (confirmed with the requester): they're viewer density taste — tuned once and left put — and the codebase already codifies that split in `canvas/canvasBoundaryGuard.test.ts`.

### The new-terminal default

A fresh terminal seeds its `collapsed` from a top-level **`preferences.newTerminalCollapsed`** — a copy-on-create default (the same shape as `newTerminalTheme`), which the terminal then owns and overrides. The old global `rightPanel.collapsed` is **migrated forward** into it (`state.ts` 1.32.0 + `SCHEMA_VERSION` bump), so a user who kept their panel collapsed keeps that as their new-terminal default across the upgrade — no silent reset.

_(The seed started life on `rightPanel.collapsed`; the lens-debate relocated it to a top-level, honestly-named `newTerminalCollapsed` so it doesn't read as a live bit next to the live geometry.)_

### Wire compatibility — padi 2.0

Adding `collapsed` to terminal metadata + the `setRightPanel` RPC is not safe across padi version skew: an old padi would strip the field (collapse silently lost on reload), and an old client omitting it would clobber the whole record. Rather than degrade silently, the codex-debate **bumped `PADI_SURFACE_VERSION` 1.3 → 2.0**, so `isContractVersionCompatible` now fails fast on skew in both directions. padi ships in the same Nix closure as kolu (they always deploy together), so a version bump is operationally free and turns a silent-data-loss path into a loud one — the fail-fast rule.

### Errors surface

A rejected `setRightPanel` (or `setSubPanel`) means the optimistic per-terminal state isn't persisted and reverts on reload. The old global path showed a toast; the per-terminal path was `console.error`-only. Both per-terminal chrome-report chokepoints (`useRightPanel` **and** its twin `useSubPanel`, which was swallowing the failure entirely) now surface it with a deduplicating toast carrying the server message.

### Design philosophy

- **Fail-fast / no fallbacks** — no override knob added; the per-terminal value is the single source of truth after a terminal's first write, and the padi version gate rejects skew rather than degrading. `newTerminalCollapsed` is a copy-on-create *seed*, not a runtime fallback.
- **Reuse the source of truth** — rides the existing `chrome.setRightPanel` → `TerminalMetadata.rightPanel` channel and session-restore seeding the tab already uses. No new persistence mechanism.
- **Electricity boundaries** — no new domain-agnostic helper; a leaf change on existing per-terminal records.

### Reviewed by the /be gauntlet

Serial review on the live branch — each committed in turn:
- **lens-debate** (lowy ⇄ hickey) — relocated the seed off `rightPanel` to top-level `newTerminalCollapsed`.
- **codex-debate** (xhigh, 3 rounds) — padi 2.0 skew gate, error-surfacing toast, migration + legacy-record regression tests, comment sweep.
- **simplify** — surfaced sub-panel report failures (the `useSubPanel` twin).
- **code-police** — carry the old collapse preference forward on upgrade (+ a migration shape guard).

### Try it locally

```sh
nix run github:juspay/kolu/fatal-brow
```


srid:

✅ **Consensus** after 2 round(s) · lowy + hickey · base `ff1c0bd22e0e`

Independent findings: lowy=2, hickey=4

### Applied (3)
- `lowy-1` Retained `collapsed` on RightPanelPrefsSchema is a write-dead seed mismatched with its live siblings — commit `07371dfe7`
- `hickey-1` New-terminal collapsed default fragmented across a now-writerless preference and the static per-terminal default — (uncommitted)
- `hickey-2` Preference field and per-terminal field both named `collapsed` for two different concepts — (uncommitted)

### Agreed — no change (3)
- `lowy-2` collapsed seed reads through the live global on every un-written read instead of baking at create like the newTerminalTheme precedent it cites (packages/client/src/right-panel/useRightPanel.ts:170-175 (freshPerTerminalState); packages/client/src/terminal/useTerminalCrud.ts:190)
- `hickey-3` The resting collapsed default (false) is duplicated across three sites (packages/padi/src/chromeVocab.ts:120)
- `hickey-4` Per-terminal `collapsed` shelved under the global 'Workspace chrome' return group (packages/client/src/right-panel/useRightPanel.ts:266)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 round(s) · codex reviewed at `xhigh` reasoning effort · base `ff1c0bd22e0e`

### Round 1

**codex** — approved: `false`

The core per-terminal persistence plumbing is mostly complete, but the change is not safe across padi version skew, silently hides failed collapse saves, and does not actually implement its stated copy-on-create seed semantics. Compatibility and migration claims also need direct tests, and several comments still describe the removed global preference model.

Findings:
- `F1` · major · open — The padi wire contract still advertises version 1.3 after adding `collapsed` to terminal metadata and `chrome.setRightPanel`. A new build therefore accepts an old 1.3 padi; that daemon strips the unknown field, making collapse appear locally successful but disappear after reload. (packages/padi/src/surface.ts:130)
- `F2` · major · open — `PadiSetRightPanelInputSchema` reuses the persisted-record schema whose `collapsed` field defaults to false. An older client accepted by the minor-version compatibility policy omits this field; the new padi then injects false and replaces the whole record, silently reopening the terminal whenever that client changes a tab or file. The deliberate default is correct for old persisted records, but it is unsafe at the live command boundary. (packages/padi/src/surface.ts:434)
- `F3` · major · open — Collapse now uses the optimistic per-terminal RPC path, whose rejection is only written to `console.error`. Previously `updatePreferences` displayed a toast. A failed save therefore looks successful to the user and silently reverts after reload, violating the project rule that caught errors must surface. (packages/client/src/right-panel/useRightPanel.ts:187)
- `F4` · minor · open — Despite the deliberate description, this is not copy-on-create state. Normal creation sends `rightPanel: undefined`, `activeState()` does not install a record, and `freshPerTerminalState()` rereads the preference until the terminal's first mutation. Thus changing the default after terminal creation changes an existing untouched terminal. The code's own comment at line 173 acknowledges this read-through behavior, unlike the genuinely creation-time `newTerminalTheme` seed. (packages/client/src/right-panel/useRightPanel.ts:229)
- `F5` · minor · open — The two compatibility seams introduced here are not directly tested: the 1.32 preference migration is untested, and the adoption sentinel only covers a new right-panel record with `collapsed` present rather than proving a legacy record defaults it to false. (packages/server/src/state.ts:499)
- `F6` · nit · open — Several comments invalidated by this change still describe the old global preference model. This feature references the removed `DEFAULT_PREFERENCES.rightPanel.collapsed`; `packages/client/src/commands.tsx:448` calls panel visibility global UI chrome; `packages/client/src/right-panel/RightPanelDrawer.tsx:9`, `useRightPanel.ts:211`, and `packages/tests/step_definitions/right_panel_steps.ts:358` still call the live collapsed state a preference. (packages/tests/features/right-panel.feature:4)


**claude** — Bumped the padi wire version for the added `collapsed` field, extracted + tested the 1.32 migration and a legacy-record schema test, swept the stale comments; disputed the presence-preserving-RPC, toast-on-collapse-fail, and copy-on-create findings as inconsistent with the record's established design.

- `F1` **fixed** — Real: adding `collapsed` to `RightPanelPerTerminalStateSchema` changes the padi wire contract (that schema rides BOTH the `terminals` metadata collection and the `chrome.setRightPanel` input), so a new client that depends on `collapsed` being persisted must refuse an older padi that silently strips it. Bumped `PADI_SURFACE_VERSION` `"1.3" → "1.4"` and rewrote its doc enumerating what 1.4 adds (`packages/padi/src/surface.ts:130`). The additive-minor gate now does exactly what codex asked: `isContractVersionCompatible("1.3","1.4") === false` (new client rejects old padi) while `("1.4","1.3") === true` (old client still speaks to a new padi). Updated the version assertions + added the new-gate assertions in `packages/padi/src/surface.test.ts:27` and the three daemon-`hello` assertions in `packages/padi/src/dial.test.ts:284,377,419`. (No `@kolu/surface*` ref-doc touched — `@kolu/padi` isn't in that table, and the version doc lives in the code.)

- `F2` **disputed** — The per-terminal `rightPanel` record is REPLACE-WHOLE by design, exactly like its sibling `subPanel` record (`packages/padi/src/terminals.ts:229` `m.rightPanel = state`). `collapsed` joined a record that ALREADY carries `activeTab`/`codeMode`/`selectedFileByMode` — and an old client that predates ANY of those (e.g. one without `selectedFileByMode`) clobbers them on a tab report in the identical way codex describes for `collapsed`. The codebase shipped those fields without a presence-preserving merge, so singling out `collapsed` for "absent means keep current" would make it the one odd field in its own record and complect the setter with a per-field merge semantic the fail-fast philosophy discourages. The honest cross-version mechanism is the VERSION GATE (F1's bump), which I did land: it stops the common new→old upgrade direction. The residual old→new clobber requires two DIFFERENT-version kolu clients driving the SAME terminal on the SAME padi (one new client having set `collapsed:true`, then an old client touching that terminal's tab) — narrow, and already the accepted semantic for the record's other client-reported fields. Fixing it for one field only would be inconsistent; fixing it for the whole record is a separate, pre-existing design question, not this PR's. `.default(false)` for persisted parsing (which codex agrees is correct) is retained.

- `F3` **disputed** — `reportToServer` (`useRightPanel.ts:187`) is the SHARED optimistic per-terminal report path that `activeTab`/`codeMode`/`selectedFileByMode` have used with `console.error` since #959 — collapse now correctly rides that exact path, which was the parent's stated intent ("restored on session restore exactly like the active tab"). Three reasons this is right, not a regression: (1) a padi-unreachable failure is ALREADY surfaced structurally by the host connection-health pip (`<HostStatusPip>` off `client.health().live`), so a per-report toast would be redundant; (2) on a downed padi every file-tree click / tab-toggle / collapse fails, so toasting per report spams one toast per interaction — precisely why the optimistic path logs instead; (3) the "caught error must not collapse to empty" rule targets a catch that fabricates a fake-success EMPTY/default STATE — here the optimistic state is PRESERVED and the failure is logged, nothing is hidden behind a synthesized default. Adding a toast belongs at the shared `reportToServer` chokepoint (touching `activeTab`/file too — a pre-existing, untouched path), not as a collapse-only special case that would re-introduce the exact inconsistency F2 warns against.

- `F4` **disputed** — codex's mechanic is correct (it IS read-through: `freshPerTerminalState()` at `useRightPanel.ts:172` rereads `preferences().newTerminalCollapsed` each call until the terminal's first write), and the code comment already says so honestly. But it is OBSERVATIONALLY identical to copy-on-create in production because there is NO runtime writer for `newTerminalCollapsed` (the author's deliberate decision #1) — the preference never changes, so "changing the default after creation" is unreachable. The `newTerminalTheme` comparison actually supports the read-through choice: theme is snapshot-at-create ONLY because it HAS a settings writer that would otherwise retroactively re-theme untouched terminals; `newTerminalCollapsed` has none, so threading a create-RPC seed + client-store install now is speculative plumbing for a writer that doesn't exist — the codebase's own "no override knobs / no speculative generality" bent argues against it. If/when a settings UI writer is added, THAT PR adds the snapshot-at-create (and the comment flags this).

- `F5` **fixed** — Agreed, a real test gap. Extracted the inline 1.32.0 migration into an exported pure `migratePreferences_1_32_0(current)` (`packages/server/src/state.ts:61`), mirroring `migratePreferences_1_30_0`, and called it from the ladder. Added three migration tests in `packages/server/src/state.test.ts` proving: newTerminalCollapsed is seeded, `rightPanel.collapsed` is stripped, geometry (`size`/`codeTabTreeSize`) is preserved, an unrelated preference (`colorScheme`) carries through, and a record with no `rightPanel` backfills the default geometry + top-level seed. Added `packages/padi/src/chromeVocab.test.ts` proving a LEGACY `rightPanel` record (no `collapsed`) parses back as `collapsed: false` via `.default(false)`, an explicit `collapsed:true` round-trips, and `DEFAULT_RIGHT_PANEL_PER_TERMINAL.collapsed === false`.

- `F6` **fixed** — Swept all five stale comments to distinguish the per-terminal persisted `collapsed` from the global panel geometry and the new-terminal default: `packages/tests/features/right-panel.feature:4` (now cites `DEFAULT_PREFERENCES.newTerminalCollapsed` + the per-terminal seed), `packages/client/src/commands.tsx:448` (dock visibility is global; the panel toggle flips the active terminal's per-terminal bit), `packages/client/src/right-panel/RightPanelDrawer.tsx:9` ("desktop chrome preference" → "desktop panel's per-terminal collapsed state"), `useRightPanel.ts` `hasTerminals` doc ("persisted collapsed preference" → "active terminal's collapsed state"), and `packages/tests/step_definitions/right_panel_steps.ts:358` ("collapsed preference" → "terminal's collapsed state").

Verification: `just check` (typecheck + biome) green; `kolu-server` (201 pass), `kolu-client` (672 pass), and the changed/new padi files `chromeVocab.test`/`surface.test`/`dial.test` (26 pass) all green. One PRE-EXISTING, environmental padi failure remains — `spawnInput.test.ts` "omits PADI_SOCKET when unknown" fails only because this shell has a live `PADI_SOCKET=/run/user/1000/padi-…/padi.sock` exported by the running `padi.service`, which the composer picks up; it does not touch any file I changed and fails identically on the base commit in this env.

commit: `6b28f5fb4110986780a772eb68cedcb60f23b1cd`


### Round 2

**codex** — approved: `false`

F1, F5, and F6 are correctly fixed, and I accept the rebuttal on F4 because the preference has no production writer. F2 and F3 remain open: version 1.4 still promises an old-client/new-padi combination that destructively resets the new field, and the shared mutation path still hides healthy-connection RPC failures from users.

Findings:
- `F1` · major · resolved — The padi wire contract needed a version bump so new clients reject older padis that cannot persist `collapsed`. (packages/padi/src/surface.ts:134)
- `F2` · major · open — Version 1.4 is described as additive and explicitly permits a 1.3 client to use a 1.4 padi, but an old whole-record write omits `collapsed`; the shared schema injects false and overwrites a persisted true value. That direction is therefore not actually backward-compatible. (packages/padi/src/surface.ts:438)
- `F3` · major · open — The shared optimistic mutation path still reports failures only through DevTools. Connection health does not cover application-level RPC failures such as validation or authorization errors while the transport remains healthy, so the UI can claim a saved state that will revert after reload. (packages/client/src/right-panel/useRightPanel.ts:187)
- `F4` · minor · resolved — The seed is read through until first mutation rather than physically captured at terminal creation. (packages/client/src/right-panel/useRightPanel.ts:229)
- `F5` · minor · resolved — The legacy schema default and 1.32 preference migration lacked direct regression coverage. (packages/server/src/state.test.ts:55)
- `F6` · nit · resolved — Comments still described the removed global collapsed preference model. (packages/tests/features/right-panel.feature:3)

_codex on the rebuttal:_ F2: I accept that singling out `collapsed` for merge semantics would be structurally inconsistent, but that supports a major version bump—not the claim that 1.4 is additive. The compatibility predicate deliberately admits 1.3 clients to 1.4 padis, where a valid old whole-record write resets persisted new state. Existing fields having had the same weakness is precedent, not evidence of compatibility. F3: I agree the fix belongs at the shared reporter rather than on collapse alone. The health pip only describes transport health; it cannot surface a rejected RPC on an otherwise healthy connection, and the repository explicitly requires terminal mutation catches to toast with the server message. Deduplication addresses the spam concern without hiding failures. F4: conceded. Since `newTerminalCollapsed` has no production writer, read-through and capture-at-create are observably equivalent today, and materializing extra state now would be speculative.


**claude** — Conceded both remaining findings: bumped the padi contract to a MAJOR 2.0 so the collapsed-clobber skew fails fast in both directions, and moved the shared per-terminal reporter onto a deduped `toast.error` per the repo's terminal-mutation rule.

- `F2` **fixed** — Conceded. The clincher is the codebase's OWN precedent: `PTY_HOST_CONTRACT_VERSION` 5.0 (`packages/kaval/src/ptyHostSurface.ts:85`) already codifies the exact rule codex is invoking — when a change's unsafe skew direction is old-client/new-daemon (the ONE direction `isContractVersionCompatible` waves through, since an old client accepts a newer-minor peer), a minor bump silently permits it and a MAJOR bump is required. `collapsed` rides the client-written whole-record REPLACE `chrome.setRightPanel` (`packages/padi/src/terminals.ts` `m.rightPanel = state`): an older client omitting `collapsed` has the shared schema's `.default(false)` fill it, then the replace clobbers a newer client's persisted `collapsed:true`. That is precisely the old-client/new-padi direction, so my round-1 "additive minor" framing was wrong — a minor bump waves the clobber through. Fixed: `PADI_SURFACE_VERSION` `"1.4" → "2.0"` with a rewritten version doc that names the direction and cites the ptyHost 5.0 parallel (`packages/padi/src/surface.ts:134`); a 2.0 major now makes `isContractVersionCompatible` reject ANY 1.x peer in BOTH directions (`("1.4","2.0")===false` new-padi-refuses-old-client, `("2.0","1.4")===false` old-client-refuses-new-padi), closing the clobber. Updated `surface.test.ts:27` (version + both-directions skew assertions), the three `hello.surfaceVersion` assertions + the newer-binder premise in `dial.test.ts` (bumped its hypothetical binder to 3.0), and converted the `assertPadiSurfaceCompatible` "older MINOR" test to `it.skipIf(minor === 0)` since a fresh major has no in-major older minor (an older padi is then a lower MAJOR, already covered). `just check` green; padi surface+dial 22 pass / 1 skip.

- `F3` **fixed** — Conceded. codex is right that the host connection-health pip reports TRANSPORT health only — it cannot surface an application-level RPC rejection (validation/authorization) on an otherwise-live padi, and the repo's own rule (`.claude/rules/toast-conventions.md` + `packages/client/src/AGENTS.md`: "terminal mutations … handle errors with `.catch((err: Error) => toast.error(...))`") mandates surfacing these. Replaced the `console.error` in the shared `reportToServer` chokepoint with `toast.error(\`Failed to save panel state: ${err.message}\`, { id: "right-panel-report-failed" })` (`packages/client/src/right-panel/useRightPanel.ts:187`). This is the shared reporter, so it covers collapse AND the pre-existing tab/codeMode/file writes consistently — the exact chokepoint codex asked for, not a collapse-only special case. The STABLE toast id answers my round-1 spam objection by construction: on a downed padi where every interaction's report fails, solid-sonner collapses them onto ONE toast instead of one-per-click. Added a regression test (`useRightPanel.test.ts`) mocking `solid-sonner` that rejects a `setRightPanel` and asserts exactly one `toast.error` with the stable id and the server message. Client right-panel 15 pass.

commit: `2d9228a9844a295d85752b91ff0fabd115c689da`


### Round 3

**codex** — approved: `true`

All six findings are resolved. The padi contract is now correctly breaking-versioned as 2.0, rejecting unsafe skew in both directions, and shared panel-state RPC failures surface through a deduplicated user-visible toast with regression coverage. The prior migration, compatibility-test, and documentation fixes remain intact.

Findings:
- `F1` · major · resolved — The padi contract needed to reject older padis that cannot persist `collapsed`. (packages/padi/src/surface.ts:150)
- `F2` · major · resolved — A minor version allowed old clients to overwrite the new whole-record field through an otherwise accepted old-client/new-padi skew. (packages/padi/src/surface.ts:150)
- `F3` · major · resolved — Panel-state RPC failures were visible only in DevTools and could silently revert after reload. (packages/client/src/right-panel/useRightPanel.ts:209)
- `F4` · minor · resolved — The default is read through until first mutation rather than physically captured at terminal creation. (packages/client/src/right-panel/useRightPanel.ts:241)
- `F5` · minor · resolved — The legacy schema default and 1.32 preference migration lacked direct regression coverage. (packages/server/src/state.test.ts:55)
- `F6` · nit · resolved — Comments still described the removed global collapsed-preference model. (packages/tests/features/right-panel.feature:3)

_codex on the rebuttal:_ F2: resolved. The 2.0 major bump preserves replace-whole semantics while making both unsafe skew directions fail fast; the updated compatibility tests verify this explicitly. F3: resolved. The shared toast covers collapse, tab, mode, and file writes consistently, includes the server message, and uses a stable ID to prevent repeated failures from producing toast spam.

srid:

**Scope:** `git diff 52e23274a...HEAD` — right-panel per-terminal collapse + its state migration. Pass 3 (elegance) skipped (`--no-elegance`; `/simplify` already ran this gauntlet).

| Pass | Finding | Disposition |
| --- | --- | --- |
| Fact-check | **Confirmed** — `migratePreferences_1_32_0` silently discarded the user's prior global `rightPanel.collapsed` instead of relocating it: a user who kept their panel collapsed had it reset to open on upgrade. | **Fixed** in `f767ae7d1` — carry the old value into `newTerminalCollapsed` (mirrors `migratePreferences_1_30_0`); the test that pinned the loss now asserts the carry-forward. |
| Rules | `migration-shape-guard` (plausible) — the migration cast `current.rightPanel` to a record and destructured it with no shape guard. | **Fixed** in `f767ae7d1` — `typeof rp === "object" && rp !== null` guard before trusting it. |

Pass 1's full rule checklist was otherwise clean — notably `caught-error-must-not-collapse-to-empty` passed (both `reportToServer` catches now surface via toast, after the `/simplify` altitude fix gave `useSubPanel` the same treatment).

srid:
## Evidence

Each terminal remembers its own right-panel collapse state as you switch between them: toggling the panel open on terminal 1, creating terminal 2 (which starts collapsed), then switching back and forth shows terminal 1 stays expanded and terminal 2 stays collapsed.



▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=right-panel-per-terminal.mp4

srid:

The **x86_64-linux** lane ran on **`kolu-ci-3`** (`dev-x86-64-linux-08`) — commit `f767ae7d1`, **failed** ✗

- **Lane wall** (pipeline): **4m39s**

| recipe | duration | |
|---|--:|:--|
| `ci::e2e` | 4m10s | ✗ (exit 1) |
| `ci::home-manager` | 2m30s | ✓ |
| `ci::unit` | 1m33s | ✓ |
| `ci::smoke` | 1m5s | ✓ |
| `ci::nix` | 59s | ✓ |
| `ci::pnpm-hash-fresh` | 48s | ✓ |
| `ci::flake-check` | 47s | ✓ |
| `ci::biome` | 35s | ✓ |
| `ci::fmt` | 34s | ✓ |
| `_ci-setup` | 14s | ✓ |
| `ci::install` | 14s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|
| `kolu-ci-1` | dev-x86-64-linux-03 | 🔒 leased |

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/f767ae7/timings.jsonl`; pool state is a live `flock` probe.</sub>

PR #1755, opened 2026-07-10
fix(code-tab): key the HTML preview cache-buster on content, not mtime

srid:
**A rendered HTML file in the Code tab, scrolled halfway down, would reload and snap back to the top at seemingly random moments — while its text never changed.** The trigger was another terminal doing git work in the same repo (a branch checkout during a PR loop, most often on a remote host): a checkout re-materializes working-tree files with the *same bytes but a new mtime*, and the preview keyed its iframe cache-buster on that mtime — so an identical-content rewrite read as an edit and forced a reload.

The preview now keys the `?v=` cache-buster on a **hash of the file's actual bytes** instead of its modification time. A real content change still bumps the URL and reloads the preview live; a same-content rewrite leaves the URL — and your scroll position — exactly where they were.

### Why mtime was the wrong identity

```
git checkout other-branch   →  working-tree file re-written
                               same bytes, NEW mtime
old:  ?v=<mtime>            →  URL changes  → iframe rebuilds → scroll jumps to top   ✗
new:  ?v=<sha1(bytes)>      →  URL stable   → no reload       → scroll preserved       ✓
```

An unchanged-mtime requery was already a no-op (the store `reconcile`s an identical value), so the *only* thing that reloaded the iframe was the URL string moving. Keying it on content makes "did this preview change?" answer to the bytes — the honest source of truth — rather than a filesystem timestamp that a branch switch trips.

### The change

- `kolu-git`: `statFileMtimeMs` → **`statFileContentTag`** (sha1 of the file's bytes). The mtime procedure is *removed*, not kept as a fallback — one path, no override knob. Rename flows through the padi surface contract (`fs.statFileContentTag`, output `string`), the endpoint, `servePadi`, and the client dispatcher.
- `BrowseFileDispatcher`: builds `…?v=${contentTag}` for the binary/iframe preview.

### Test-first

Reproduced the bug as a **red** unit test before the fix (`browse.test.ts`): a same-content rewrite with a bumped mtime must keep the tag stable. Red against the old mtime tag, green once it hashes content. A companion test confirms a real byte change still moves the tag, and the path-traversal guard carries over.

> Honors the repo's design philosophy: **fail-fast / no-fallbacks** (the mtime path is deleted, not degraded to a fallback; a stat failure still surfaces as `GIT_FAILED`) and **reuse the existing source of truth** (the file's bytes are the identity of "changed?"; standard-lib `crypto`, same guarded resolve path as `readFile` — no hand-rolled mechanism).


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `52e23274aa07`

Independent findings: lowy=1, hickey=3

### Applied (3)
- `lowy-1` Receptacle named after its mechanism, not its role — the tag identity axis leaks through the socket — commit `62c7a03ca`
- `hickey-1` Content-tag reads the WHOLE file (incl. multi-GB video) into heap — contradicts serve-dir's never-buffer invariant — commit `969cedf0d`
- `hickey-2` `statFileContentTag` keeps the misleading `stat` prefix after it stopped stat-ing — commit `efcd6484a`

### Agreed — no change (1)
- `hickey-3` Two contradictory file-identity tokens now co-exist in the preview subsystem (packages/serve-dir/src/index.ts:277)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 round(s) · codex reviewed at `xhigh` reasoning effort · base `52e23274aa07`
### Round 1

**codex** — approved: `false`

The change targets the right symptom, but it is not ready to merge. The padi procedure replacement is an unversioned breaking wire change; the large-file policy incorrectly assumes every file over 32 MiB is a video; and the stat/read design can defeat its own memory bound during atomic replacement. The hashing path also has avoidable per-viewer buffering, missing operator logging, stale test documentation, and no end-to-end regression for preserved iframe scroll. No untracked files were present.

Findings:
- `F1` · blocking · open — `fs.statFileMtimeMs` is removed from the padi wire contract and replaced with `fs.filePreviewTag`, but `PADI_SURFACE_VERSION` remains `1.3`. Compatibility accepts a running 1.3 daemon, so an older daemon lacking `filePreviewTag` can be adopted and binary previews then fail with an opaque missing-procedure error. This is a breaking shape change, not an additive minor change. (packages/padi/src/surface.ts:130)
- `F2` · major · open — The deliberate large-file reasoning is not represented by the code. `filePreviewTag` uses the lossy size-plus-prefix tag for every file over 32 MiB, while its caller invokes it for HTML, SVG, PDF, raster, and video previews. A 40 MiB HTML/PDF with unchanged size and first 32 MiB but an edited tail retains the same URL and never reloads. Thus the claim that only multi-GB videos can miss, and that HTML/SVG are always hashed whole, is false. (packages/integrations/git/src/browse.ts:160)
- `F3` · major · open — The size and bytes are read through separate pathname operations. If an atomic writer replaces a file after `fsStat` but before `fsReadFile`/`fsOpen`, the branch and size describe one inode while the bytes come from another. In the small branch, a file measured below 32 MiB can be replaced by a multi-GB file that `fsReadFile` then buffers whole, defeating the claimed memory cap and potentially killing padi. (packages/integrations/git/src/browse.ts:158)
- `F4` · minor · open — Even without the replacement race, each request buffers or allocates up to 32 MiB and feeds a large buffer synchronously to `hash.update`. Tag queries run per subscribed client, so one change with many active previews multiplies memory and serving-loop work—for example, 20 viewers can transiently require roughly 640 MiB. (packages/integrations/git/src/browse.ts:164)
- `F5` · minor · open — The new hashing path catches stat/open/read/close failures and returns `GIT_FAILED` without logging through the supplied logger. That violates the repository's `errors-must-log-at-error` rule and leaves operators without an error-level record of unexpected preview I/O failures. (packages/integrations/git/src/browse.ts:194)
- `F6` · nit · open — The HTML live-refresh scenario still documents the removed `?v=<mtime>`, `buildIframePreviewUrl`, `fsReadFile` stream, and `fsReadFileOutputEqual` pipeline. This branch makes that explanation inaccurate. (packages/tests/features/code-tab.feature:1573)
- `F7` · minor · open — The added tests prove only that the helper returns the same hash for identical small-file bytes. They do not exercise the reported user regression: an open, scrolled sandbox iframe receiving a same-content atomic rewrite must retain its frame and scroll position. The existing scenario covers the opposite case—changed content reloads—so the new behavioral promise remains unprotected. (packages/tests/features/code-tab.feature:1579)


**claude** — Fixed the content-hash design (whole-file streamed hash, one open handle), bumped the padi surface to a breaking major, added error logging + large-file coverage, and refreshed the stale docs; disputed only the infeasible iframe-scroll e2e.

- `F1` **fixed** — Bumped `PADI_SURFACE_VERSION` "1.3" → "2.0" (`packages/padi/src/surface.ts:130`) with an expanded history note: removing `fs.statFileMtimeMs` and adding `fs.filePreviewTag` is shape-breaking in BOTH directions (a 1.x binder calling `statFileMtimeMs` on a 2.0 padi, or a 2.0 binder calling `filePreviewTag` on a 1.x padi, each hit a missing procedure). A minor bump was NOT enough: `isContractVersionCompatible(reported, expected)` treats a HIGHER reported minor as compatible, so a 1.x binder would still adopt a 1.4 padi and call the vanished `statFileMtimeMs`. Only a major flips the gate to refuse in both directions. This also closes the off-nix-dev gap the buildId fence leaves open (an off-nix binder, buildId `""`, never drains on build grounds, so a same-version stale padi would be adopted). Updated the assertions: `surface.test.ts:27/32`, `dial.test.ts:284/377/419` ("1.3"→"2.0"; the "newer binder" hypothetical moved 2.0→3.0), the older-version test (`dial.test.ts:530`) made robust to a `.0` floor, and `padiBinding.test.ts:558` narrative numbers. `just check` green; `dial.test.ts` (real spawned padi) 12/12 green.

- `F2` **fixed** — Agreed; the author's "html/svg always hashed whole" claim was false for a >32 MiB file. Removed the size cutoff and the leading-bytes shortcut entirely (`packages/integrations/git/src/browse.ts`): `filePreviewTag` now hashes the WHOLE file, so a tail-only edit of a large previewable HTML/PDF always moves the tag. Memory stays bounded because the bytes stream through `createReadStream`'s fixed 64 KiB chunks into the hash — never a whole-file slurp, whatever the size (still honors serve-dir's never-buffer invariant, even for a multi-GB video). Added a red-guard test: a 40 MiB file with an identical first 40 MiB and a changed tail moves the tag (`browse.test.ts`).

- `F3` **fixed** — Agreed; this mirrors the exact TOCTOU serve-dir already solved (`packages/serve-dir/src/index.ts:302-310` — "Deriving the size and the bytes from the same open file description ... the handle pins one inode"). `filePreviewTag` now opens ONE handle and hashes through it; there is no separate `stat(path)`, and no size-branch that could `readFile`-slurp a file swapped-to-huge after the stat. An atomic write-and-rename mid-hash yields a consistent snapshot of the opened inode, never a torn stat/read pair.

- `F4` **partial** — Fixed the memory half (the real concern): the per-request 32 MiB `allocUnsafe` is gone; each viewer's hash now streams 64 KiB chunks, so 20 concurrent previews cost ~1.3 MiB, not ~640 MiB. Did NOT thread an AbortSignal through the I/O: `filePreviewTag` is a `kolu-git` leaf reached via the endpoint and servePadi, none of which carry the client `signal` today, and every realistic preview (html/svg/pdf/image) resolves in well under a millisecond; with memory now bounded a superseded query wastes only a trace of bounded, chunked I/O. Plumbing cancellation end-to-end is a larger change disproportionate to the residual concern, so it is deliberately deferred.

- `F5` **fixed** — Agreed. The catch now logs: an expected delete-while-viewing `ENOENT` at `debug` (servePadi already maps it to a typed `NOT_FOUND`, so error-level would be false alarm noise — the rule reserves debug for expected-absent), and any OTHER unexpected preview I/O fault at `error` with `{ err, repoPath, filePath }` (errors-must-log-at-error). `browse.ts` catch block.

- `F6` **fixed** — Agreed; updated the stale scenario comment (`packages/tests/features/code-tab.feature:1569`) to describe the real pipeline: `subscribeFileChange` pulse → `fs.filePreviewTag` (content hash) requery → client rebuilds `?v=<tag>` → `binaryFile` memo identity flips → FileView re-points the iframe `src`, while an identical-content rewrite leaves tag/URL/scroll untouched. Dropped the `?v=<mtime>` / `buildIframePreviewUrl` / `fsReadFileOutputEqual` references.

- `F7` **partial** — Strengthened the unit coverage that pins the exact behavioral contract (the same-content-rewrite-keeps-tag-stable test is the regression red-then-green, plus the new large-file tail-edit test), which per the confirmed root cause is the ONLY thing that rebuilds the iframe: identical bytes ⇒ identical tag ⇒ unchanged `?v=` ⇒ no `src` change ⇒ no reload ⇒ scroll preserved. Disputed the requested e2e scroll-position assertion: the preview iframe is an opaque-origin `allow-scripts` sandbox with NO `allow-same-origin`, so parent JS cannot read `contentWindow.scrollY` — the existing scenarios only reach the frame's rendered *content* through the artifact-sdk `postMessage` bridge, which exposes no scroll position. Asserting preserved scroll would require extending the in-iframe artifact-sdk to post `scrollY` back to the parent — a new harness capability well beyond this bug fix, and the scroll preservation is a deterministic consequence of the unit-tested tag stability.

commit: `b97c5cdb8fa43991973df9d4e5b5987439ae157c`


### Round 2

**codex** — approved: `false`

F1, F2, F3, and F5 are resolved. F4 remains open because full-file hashing is still uncancellable, F6 retains one obsolete stream reference, and F7 remains open because Playwright can inspect scroll state inside the sandboxed frame without changing the artifact SDK.

Findings:
- `F1` · blocking · resolved — The breaking padi procedure replacement now has the required major contract-version bump. (packages/padi/src/surface.ts:138)
- `F2` · major · resolved — The tag now hashes the complete file, so large non-video tail edits are detected. (packages/integrations/git/src/browse.ts:140)
- `F3` · major · resolved — Hashing now uses one open file handle, eliminating the separate stat/read inode race and unbounded replacement slurp. (packages/integrations/git/src/browse.ts:175)
- `F4` · minor · open — The memory and large synchronous-update portions are fixed, but a superseded hash remains uncancellable. The implementation now reads every file completely, explicitly including multi-gigabyte videos, so abandoned requests can continue consuming disk and CPU for seconds or longer; repeated pulses can overlap such obsolete scans. (packages/integrations/git/src/browse.ts:161)
- `F5` · minor · resolved — Expected ENOENT deletion races and unexpected hashing failures now log at the appropriate levels with useful context. (packages/integrations/git/src/browse.ts:184)
- `F6` · nit · open — The binary-preview explanation is corrected, but the preceding comparison still says text content arrives on an `fsReadFile` stream. The current text path is also pulse-then-requery through `fs.readFile`; that stream no longer exists. (packages/tests/features/code-tab.feature:1571)
- `F7` · minor · open — The exact iframe scroll-preservation regression remains without browser coverage. The opaque-origin sandbox prevents application parent JavaScript from reading `contentWindow`, but it does not prevent Playwright from evaluating code directly in the child frame—the existing iframe-content step already relies on that distinction. (packages/tests/features/code-tab.feature:1579)

_codex on the rebuttal:_ F4: I accept the bounded-memory fix but hold firm on cancellation. The claim that previews finish in under a millisecond conflicts with the new whole-file algorithm and its explicit multi-gigabyte-video support. This is also not extensive new infrastructure: surface procedure handlers already receive `signal?: AbortSignal`; it can be passed through the endpoint to the read stream.\n\nF7: I do not accept the infeasibility argument. Opaque origin constrains code running in the parent page, not Playwright/CDP. `frameLocator(...).locator("body").evaluate(...)` executes inside the iframe and can read its own `window.scrollY` or `document.scrollingElement.scrollTop`, just as the existing test reads frame-local body text.\n\nF6: the update is incomplete. The old `?v=<mtime>`, URL-builder, and equality references are gone, but line 1571 still names the removed `fsReadFile` stream.


**claude** — Conceded and landed both remaining substantive findings: threaded the request `signal` end-to-end so a superseded preview hash aborts mid-read (F4), fixed the last stale `fsReadFile` wording (F6), and added the browser scroll-preservation scenario I wrongly called infeasible last round (F7).

- `F4` **fixed** — Conceded. codex was right that the plumbing already exists: `ProcedureImpl` hands every handler `signal?: AbortSignal` (`packages/surface/src/server.ts:1294`) and the client already passes `{ signal }` at the call site (`packages/client/src/right-panel/BrowseFileDispatcher.tsx:154`) — only the last hop dropped it. Threaded it: servePadi handler `filePreviewTag: async ({ input, signal })` → `endpoint.fs.filePreviewTag(repoPath, filePath, signal)` (`packages/padi/src/servePadi.ts:454`), endpoint interface + impl now carry `signal?: AbortSignal` (`packages/padi/src/terminalWorkspace/endpoint.ts:49,132`), and the kolu-git leaf takes a 4th `signal?` param (`packages/integrations/git/src/browse.ts:161`). The hash now `signal?.throwIfAborted()` before opening and between every chunk, so a stale whole-file read — the seconds-long multi-GB-video case codex flagged — stops promptly; on abort we re-throw the cancellation (not a `GIT_FAILED` err) so the framework sees a cancel and the catch does not misclassify it as an I/O fault to log. I verified empirically that passing `signal` into `createReadStream` itself emits a second unhandled `'error'` event that crashes the process on a pre-aborted signal, so I used the throw-between-chunks pattern instead (verified crash-free for pre-abort, mid-abort, and no-signal). New unit test `browse.test.ts` asserts a pre-aborted signal rejects with `AbortError` (11/11 green).

- `F6` **fixed** — Conceded. The comparison clause still named the removed `fsReadFile` stream. Rewrote it to the actual path: "the `subscribeFileChange` pulse requeries `fs.readFile` and re-feeds the fresh content to Pierre" (`packages/tests/features/code-tab.feature:1569`).

- `F7` **fixed** — Conceded; my round-1 infeasibility argument was simply wrong. codex correctly noted the existing content-read step already reaches into the opaque-origin sandbox via `frameLocator`, so reading frame-local `scrollingElement.scrollTop` is equally reachable. Added a browser scenario "An identical-content rewrite preserves the HTML preview scroll position" (`packages/tests/features/code-tab.feature`): loads a tall HTML preview, scrolls the frame's own scrolling element to the bottom (new step "I scroll the file preview iframe to the bottom", captured on the world), fires the file-change watch with identical-mtime touches (`nudgeFiles` IS the same-bytes-new-mtime rewrite), and asserts frame-local scroll HOLDS through them (new step "…holds its scroll position through identical rewrites of …") — a reset to the top would mean the `src` was re-pointed, the exact regression. To close the vacuous-pass gap (a dropped watch would trivially "preserve" scroll), the scenario ends with a REAL edit that must still refresh the frame: that liveness guard proves the watch fires and DOES reload on a genuine change, so the hold-steady window is meaningful. Steps read scroll through `frameLocator(...).locator("body").evaluate(...)`. Cucumber `--dry-run` resolves all steps (0 undefined); gherkin parses (504 scenarios).

Verification: `just check` (tsc + biome) green; `browse.test.ts` 11/11; padi `endpoint.test.ts` + `servePadi.test.ts` 21/21; cucumber dry-run clean. `packages/tests` has no typecheck step (tsx runtime, no tsconfig), so the step edits are covered by the dry-run + biome, not tsc.

commit: `e0107af315adef10ac1320b612a8f91c9e5c7201`


### Round 3

**codex** — approved: `true`

All seven findings are resolved in the current tree. The padi contract is correctly versioned, hashing is whole-file, streamed, single-handle, cancellable, and properly logged, the stale documentation is corrected, and the new browser scenario directly verifies scroll preservation with a genuine-change liveness guard. No new regression was introduced and no untracked files were present.

Findings:
- `F1` · blocking · resolved — The breaking padi procedure replacement now carries a 2.0 major contract version. (packages/padi/src/surface.ts:138)
- `F2` · major · resolved — The tag hashes the complete file, so large non-video tail edits are detected. (packages/integrations/git/src/browse.ts:140)
- `F3` · major · resolved — Hashing uses one open file handle, eliminating the separate stat/read inode race. (packages/integrations/git/src/browse.ts:175)
- `F4` · minor · resolved — The request abort signal is now threaded through servePadi and the endpoint into the hash loop. Pre-aborted and superseded scans stop without being converted into I/O failures, and the file handle closes during loop unwinding. (packages/integrations/git/src/browse.ts:161)
- `F5` · minor · resolved — Expected deletion races and unexpected hashing failures log at the appropriate levels with useful context. (packages/integrations/git/src/browse.ts:201)
- `F6` · nit · resolved — The documentation now accurately describes the pulse-triggered `fs.readFile` requery rather than the removed stream. (packages/tests/features/code-tab.feature:1571)
- `F7` · minor · resolved — The browser suite now records frame-local scroll through Playwright, exercises repeated same-content mtime changes, fails on a scroll reset, and follows with a real edit that proves the watcher remains live. (packages/tests/features/code-tab.feature:1606)

_codex on the rebuttal:_ F4 resolved: I verified the signal reaches the kolu-git leaf, is checked before opening and between chunks, propagates cancellation unchanged, and still closes the handle through `finally`. F6 resolved: the remaining `fsReadFile` stream wording is gone. F7 resolved: the new frame-context scroll assertion directly covers the reported regression, and the subsequent real-content refresh prevents a dead watcher from producing a vacuous pass.

srid:

Ran Pass 1 (rules) + Pass 2 (fact-check); elegance skipped (`/simplify` already ran this gauntlet).

**Actioned (3):**
- `dry-rule-of-three` / `no-overloaded-null` — the ENOENT "file is gone" sniff was hand-rolled in both `browse.ts::filePreviewTag` (log-level) and `servePadi.ts::fileGoneAsNotFound` (wire `NOT_FOUND` mapping). Hoisted to one `isFileGoneError` in `kolu-git`'s errors module — single source, can't drift. (`e027fa32b`)
- `fact-check` — `filePreviewTag` rethrew on `signal.aborted`, so a genuine I/O fault landing after an unrelated abort was silently misread as a cancellation. Now classifies the abort by the **error's** identity (the repo's `instanceof DOMException && name === "AbortError"` idiom). (`1d5acba68`)
- `fact-check` — the `finally { fh.close() }` could throw and discard the original error propagating out of the try; close's own error is now swallowed so the true root cause survives. (`1d5acba68`)

**Considered, not actioned (justified):**
- `e2e-poll-async-state` — the hold-steady step asserts an **invariant holds for a whole window** (scroll never resets across repeated same-content rewrites); `pollFor` resolves on first-true and can't express "must not change", so the manual deadline loop is the correct shape (it matches the suite's existing hold-assertion idiom).
- whole-file rehash on every HEAD-move pulse — real but minor: bounded-heap (streamed), debounced 150ms, negligible for realistic KB-sized HTML/SVG previews. A mtime+size memo would help only a large-file-during-checkout-loop edge case at the cost of an unbounded module-level cache; deferred.

srid:

The **x86_64-linux** lane ran on **`kolu-ci-5`** (`idliv2-02`) — commit `8778acd40`, **passed** ✓

- **Lane wall** (pipeline): **3m29s**

| recipe | duration | |
|---|--:|:--|
| `ci::home-manager` | 3m14s | ✓ |
| `ci::e2e` | 2m51s | ✓ |
| `ci::unit` | 1m30s | ✓ |
| `ci::smoke` | 1m1s | ✓ |
| `ci::nix` | 56s | ✓ |
| `ci::flake-check` | 42s | ✓ |
| `ci::atlas-sync` | 31s | ✓ |
| `ci::fmt` | 27s | ✓ |
| `ci::biome` | 27s | ✓ |
| `ci::install` | 17s | ✓ |
| `_ci-setup` | 15s | ✓ |

### Pool status (8 boxes)

| box | location | state |
|---|---|---|

<sub>Posted by `ci/pu/report.sh`. Lane timings from `.ci/8778acd/timings.jsonl`; pool state is a live `flock` probe.</sub>

PR #1758, opened 2026-07-11
fix: kill the right-panel flicker on split (#1714's reprojectClock identity churn)

srid:
## The regression

Opening a split terminal while the Code tab is showing made the whole code view — file tree, content/preview, and scope switcher — blank to "Loading…" and remount (a visible flicker). It also fired on plain metadata ticks, not just splits. Traced (runtime-verified with a MutationObserver on the live build) to **#1714** (`7d51cb2b2`).

**Mechanism.** #1714 changed `getMetadata` to return `reprojectClock(rawTile(id))` — a fresh object minted per call, reactive on the active host's clock. A consumer keyed on the reference (the `active` memo → `repoPath()` → the Code tab's polled queries) then saw a spurious "new value" on every incidental `lastActivityAt`/`agent`/clock tick. That reached `createPolledQuery`, whose re-subscribe effect keyed on a fresh `{ i: input(), host }` object and so **blanked + re-subscribed on every invalidation** even when the value was unchanged — remounting the tree. The blank-on-invalidation was latent since #1652; #1714 armed it.

## The fix — four parts, per a completed architecture-first-principles review

The adjudication + refuter record: the fix belongs at the metadata boundary (restore identity stability at the one knowing endpoint) **and** the consumer must honor its own stated contract — "both ends change; fixing only the producer is not honest."

1. **The projection** (`fix(metadata)`) — a per-id, reconcile-backed derived projection between the raw records and every value consumer. Each live terminal gets a computation that reprojects its raw record and writes **through `writeWrappedValue`** (the framework's reconcile-or-assign gate, the same one every subscription value rides) into a per-id `{ v }` store, keyed in a `slots` store so `getMetadata`'s lookup re-subscribes per id **without tracking membership**. `getMetadata`'s reference now changes **iff** the reprojected value changed, and leaf readers (`meta.git.repoRoot`) track their leaf again. `reprojectClock` is now module-private (the projection's computation is its sole caller); `rawTile` and the parked path keep the raw record; the clock fence is unchanged. In-repo precedent: `HostDaemonChips.daemon()`.
2. **The consumer contract** (`fix(code-tab)`) — `createPolledQuery` now keys its re-subscribe effect on a **value-deduped input key**, enforcing its header's "only an INPUT change blanks". A fresh-reference-but-equal-value input re-issues nothing.
3. **Same-class sweeps** (`refactor(code-tab)`) — the Code tab's composer-close `on()` keyed on a value memo (was an array rebuilt each eval); `localDaemonStatus` memoized to match the `HostDaemonChips` reprojection idiom.
4. **Pins** (`test(code-tab)`) — `getMetadata` identity stability (same ref across repeated reads / unrelated-field ticks / a clock-offset tick landing identical values; a leaf reader untouched by an incidental tick but re-run on a real repoRoot change; `meta.agent` stable when unchanged); and a `createPolledQuery` pin that a fresh-reference-equal-value input does **not** blank, go pending, or re-query.

## Verification

- **Unit tests green** — the full client suite **677/677**, including **25** tests across the two touched test files. The `createPolledQuery` pin directly reproduces the flicker mechanism (equal-value input → no blank); the `getMetadata` pins lock the restored identity-stability contract.
- **Typecheck green** (`tsc --noEmit`), **fmt** applied.
- **Runtime (dev-server) verification was blocked** and is **not** claimed here. The dev build hits a flaky, **change-independent** host-resolution race (`scopedByEntry: active() names non-member local — no owned world`) that wedges terminal-create on the canvas ("Connecting to local…"). I confirmed it is unrelated to this change: a variant whose connect path is byte-identical to `master` (differing only in the inert, unmounted-during-connect Code-tab files) wedged the same way, and `master` itself connects only intermittently. Because the wedge is upstream of any terminal rendering, I could not get a stably-connected repo terminal in the dev build to drive the split-flicker flow end-to-end. **Recommend a clean-box (pu) runtime check before merge.**

Draft — for review; srid merges.


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `68ca01898b00`

### Applied (1)
- `hickey-1` Daemon-status reprojection duplicated across two files instead of unified — commit `7a60a47f3`

### Agreed — no change (8)
- `lowy-1` Reconcile-backed identity-stable accessor is a third hand-rolled copy of surface's {v}-store + writeWrappedValue + accessor triplet (packages/client/src/terminal/useTerminalMetadata.ts:212-243)
- `lowy-2` createComputed leaks into packages/client for the first time — a framework-altitude wiring living in an app module (packages/client/src/terminal/useTerminalMetadata.ts:220,243)
- `lowy-3` Two different reprojection identity-stability strategies shipped in one PR (reconcile-store vs memo-spread) — mild symmetry smell (packages/client/src/kaval/useDaemonStatus.ts:150-163)
- `lowy-4` JSON.stringify canonical key leaks an unstated 'Input is JSON-canonical' contract through createPolledQuery's generic interface (packages/client/src/right-panel/createPolledQuery.ts:168-176)
- `hickey-2` Sentinel handling has already drifted across the three reprojection sites (packages/client/src/kaval/useDaemonStatus.ts:159)
- `hickey-3` Same 'on fires on invalidation, not value change' gotcha patched three different ways in one PR (packages/client/src/right-panel/CodeTab.tsx:238)
- `hickey-4` JSON.stringify as the canonical input key (packages/client/src/right-panel/createPolledQuery.ts:179)
- `hickey-5` slots store manually shadows mapArray's membership (packages/client/src/terminal/useTerminalMetadata.ts:212)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `68ca01898b00`

### Round 1

**codex** — approved: `false`

The keyed metadata projection, polling reset semantics, and daemon reprojection changes are otherwise sound, including the adjudicated architecture decisions. Approval is blocked by an eager use-before-declaration crash and incorrect composer scoping across terminal/host switches. Two smaller source/documentation issues also remain.

Findings:
- `F1` · blocking · open — `createMemo` evaluates eagerly, but its getter calls `selectedPath()` before that `const` is initialized at line 252. Mounting `CodeTab` therefore throws a temporal-dead-zone `ReferenceError`. (packages/client/src/right-panel/CodeTab.tsx:235)
- `F2` · major · open — The anchor includes only path, mode, and repo. Switching to another terminal or host at the same location leaves the global draft open, contrary to `composerState.ts`'s host-switch contract. Saving then uses the new `props.terminalId`, potentially attaching the old draft to the wrong terminal. (packages/client/src/right-panel/CodeTab.tsx:235)
- `F3` · minor · open — The template literal contains two literal NUL bytes. Text tools such as `rg` consequently classify the TSX file as binary and suppress normal matches. (packages/client/src/right-panel/CodeTab.tsx:236)
- `F4` · nit · open — The comment says the returned reference changes iff the reprojected value changes. In fact, `reconcile` retains the Solid store proxy across field changes, as the new test explicitly verifies after changing `lastActivityAt`. (packages/client/src/terminal/useTerminalMetadata.ts:191)


**claude** — Conceded and fixed all four findings: the composer block was genuinely broken (eager-createMemo TDZ + NUL bytes) and under-keyed (missing terminal identity), and the metadata doc overstated reconcile's identity contract.

- `F1` **fixed** — Confirmed the crash. `createMemo` evaluates eagerly, and its getter read `selectedPath()`, a `const` declared *below* it (was line 252). Mounting `CodeTab` would throw a TDZ `ReferenceError`; the suite stayed green only because no test mounts the component. Moved the whole composer block (comment + `const composer = useComposer()` + `composerAnchor` + close effect) to sit AFTER `selectedPath`/`select`/`slotKey` — now `CodeTab.tsx:273-277`, with every dependency already in scope.
- `F2` **fixed** — Agreed. The old code keyed `on([selectedPath(), view(), repoPath()])` and fired on every INVALIDATION, so it closed on any terminal/host switch. The new value-deduped memo notified only on a real string change, so a switch to another terminal (or host — a terminal is host-bound) at the same repo/file/mode would leave the global draft open and a subsequent save would attach it to the now-active `props.terminalId`. That violates composerState.ts:37-40's stated host-independent "drops the stale draft" contract. Fix: added `${props.terminalId ?? ""}` as the leading key segment (`CodeTab.tsx:275`). `terminalId` is value-stable — it changes only on a genuine terminal switch, never on a clock tick — so this preserves the #1714 flicker fix while restoring the close-on-switch guarantee. Host identity needs no separate segment: a terminal is bound to one host for its lifetime, so a host switch is necessarily a `terminalId` change.
- `F3` **fixed** — Confirmed two literal NUL (`\0`) bytes in the template separator (verified via `od -c`; `grep`/`rg` classified the whole TSX as binary and silently suppressed matches, which is how the block was hard to search). Replaced the embedded NUL bytes with `\0` escape sequences in the source — identical runtime string, but the file is now plain text (grep matches again without `-a`).
- `F4` **fixed** — Agreed the comment overstated the contract. `read()` returns `store.v`, and because `writeWrappedValue` `reconcile`s, that proxy stays STABLE across field changes (the `lastActivityAt` test proves the reference survives while the leaf notifies); the reference turns over only around `undefined ⇄ record` or a structural swap reconcile can't merge. Reworded `useTerminalMetadata.ts:190-195` to say exactly that instead of "the returned reference changes IFF the reprojected value changed."

Verification: `just fmt` clean, `just check` green (client typecheck + biome lint pass), and the two PR-touched unit files (`createPolledQuery.test.ts`, `useTerminalMetadata.test.ts`) pass 25/25.

commit: `0742eab6ef45d44ebfa4146eff9263379ef2d6ab`


### Round 2

**codex** — approved: `true`

All four findings are resolved in the current tree. The composer memo now has safe declaration order, includes globally unique terminal identity, and uses textual `\0` escapes. The metadata comment now accurately describes reconcile-backed proxy stability. No regression was introduced by this round.

Findings:
- `F1` · blocking · resolved — The composer memo previously read `selectedPath` before initialization. (packages/client/src/right-panel/CodeTab.tsx:273)
- `F2` · major · resolved — The composer anchor previously omitted terminal/host scope. (packages/client/src/right-panel/CodeTab.tsx:274)
- `F3` · minor · resolved — The source previously contained literal NUL bytes. (packages/client/src/right-panel/CodeTab.tsx:276)
- `F4` · nit · resolved — The projection comment previously misstated when reconciled object identity changes. (packages/client/src/terminal/useTerminalMetadata.ts:186)

_codex on the rebuttal:_ F1: accepted—the composer block now follows `selectedPath`, so the eager memo has no TDZ read. F2: accepted—the terminal-only identity segment is sufficient because `TerminalId` is a globally unique UUID and active selection rekeys on host switches; an empty host yields `null`, which also changes the anchor. F3: accepted—byte inspection confirms zero literal NUL bytes, with printable `\0` escapes preserving runtime separation. F4: accepted—the revised comment correctly describes a stable proxy during a continuous record lifetime with leaf-level notifications and turnover around absence or recreation.

srid:

Scope: `git diff 68ca01898...HEAD` (PR #1758, right-panel flicker fix). Pass 3 (elegance) skipped — `/simplify` already ran this tree.

| Pass | Issues | Outcome |
| --- | --- | --- |
| Rules | 0 | Clean — full 30-rule checklist, each `Checked by:` grep. No dead code, no silent error swallow (only relocation of the existing `surfaceError` catch), per-id `slots` store has `onCleanup` eviction, no thin wrappers, no untyped escapes. |
| Fact-check | 1 (fixed) | See below. |

**Fact-check — `createPolledQuery.ts` header docblock (fixed, `dd22104c2`).** The header said *"Only an INPUT-VALUE change blanks + goes pending,"* but the dedup key is `JSON.stringify([i, host ?? null])` — a bare **host** switch (same input value, different host) also blanks, and both `createRepoPolledQuery` and `BrowseFileDispatcher` pass `pulseHost: activeHost`, so this is a real path. Reworded the header to state the real gate ("Only an (INPUT-VALUE, HOST) pair change blanks… a bare host switch on the same input value blanks too") and added a covering test (`a host-only change (same input value) still blanks + re-subscribes`).

Every other comment/invariant across the diff (`reprojectClock` module-privacy, the shared `reprojectDaemonStatus` body, `localDaemonStatus`'s memo rationale, `composerAnchor`'s collision-safety) was fact-checked and found accurate.

srid:
## CI & evidence — blocked on a pu/Incus infrastructure outage

The gauntlet-reviewed change is on `HEAD` (`639f30e08`) and green locally (`tsc --noEmit` clean, **678/678** client unit tests). But the two ship-phase gates can't run right now:

- **CI (both platforms):** the `x86_64-linux` lane has **no host** — all 8 warm pool boxes (`kolu-ci-1..8`) are `unhealthy/missing`, a cold ephemeral create fails **"no egress"**, and `~/.config/odu/hosts.json` is empty. `just ci::pool-ensure` can't repair it (every slot fails no-egress). Logged on **juspay/kolu#1204**.
- **`aarch64-darwin` lane (rasam, physical — reachable):** red, but on **flaky load-timeouts in packages my diff does not touch** — run 1 timed out 2 `kolu-git` worktree tests (15s), run 2 timed out a `packages/padi` `dial.test.ts` (60s). **No `packages/client` test (this change) failed** on either run; the failures vary per run — the box is overloaded, not a regression.
- **Clean-box (pu) evidence:** the required end-to-end split-flicker capture needs an ephemeral `pu` box, blocked by the same no-egress outage.

**Recommendation:** re-run CI (both lanes) and capture the clean-box evidence once the pu/Incus egress is restored (#1204). The change itself is complete and reviewed.

srid:
## Evidence

**#1714 — the right-panel flicker on split — is fixed.** With the Code tab open on a repo (All-files tree + a file preview), creating a split terminal used to blank the **entire** code view — tree, file preview, and scope switcher — to `Loading…` and remount it (a visible flicker). Root cause: `reprojectClock` churned `getMetadata`'s reference on every tick, so the split (a new terminal → a metadata tick) re-fired every downstream query and remounted `CodeTab`. This PR keys same-class reprojection on values, so a split leaves the tree/preview exactly in place.

Captured with the project's Cucumber + Playwright e2e harness (`KOLU_EVIDENCE=1`), one scenario per clip, recorded on `master` (before) and this branch (after). Both clips run the identical flow: open a git repo → Code tab in **All-files (browse)** mode → open `beta.txt` (preview shows `second file contents`) → **create a split** (Toggle terminal split).

### Before ↔ After — the split moment

The instant the split opens: **before**, the whole Code panel drops to `Loading…` (tree + preview gone, remounting); **after**, `beta.txt` → `second file contents` and the file tree stay put.

| Before (`master`) — blanks to `Loading…` | After (this PR) — stays put |
|---|---|
|  |  |

### Motion (the point is the tree NOT blanking as the split opens)

**Before — flicker on split (`master`):**



**After — no flicker (this PR):**



▶ HD before: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=before-split-flicker.mp4
▶ HD after:  https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=after-split-noflicker.mp4

<sub>Scenario `Code view survives creating a split` (appended to `code-tab.feature`, reusing existing steps) passes on both refs — the assertion polls, so it re-renders green even through a remount; the **video** is what shows the flicker. Capture ran via the e2e harness locally (isolated random ports / per-worker runtime dir, no contact with anything else) rather than a fresh `pu` box, because `pu` instance provisioning was down at capture time (`context deadline exceeded` on every create across 16 attempts / 3 names); the harness build is the same one CI exercises.</sub>

PR #1764, opened 2026-07-11
padi W9: instant host switch-back — retained per-host wire subscriptions + Code-tab per-host scoping

srid:
Switching hosts is supposed to feel like switching tmux sessions — instant, the canvas already there. It didn't: every switch **back** to a host you'd just visited rebuilt the whole canvas. The dock emptied and refilled, the terminal went black, and the Code tab dropped to "Loading…" for about a second — for state the client held moments earlier.

The cause, in one line: every per-host wire readout keyed on `activeHost` with **dispose-on-key-change**, so a switch-back reopened each readout from *pending*. Pending data made `canvasMode` leave `workspace`, and `App.tsx`'s `<Switch>` unmounted the Dock, tiles, and RightPanel wholesale — then remounted them from zero as records trickled back. This is W7's own cure applied to its one deliberate exclusion (**K1**): move those subscriptions into the retained `scopedByEntry` per-host owner, so a switch-back has no resubscribe and no pending window *can* exist.

Per the interview, this ships **both halves** of the symptom in one PR.

## The change

**Half 1 — the wire subscriptions (the canvas).** A new sibling factory `hostScope/createHostWire.ts` opens the five per-host readouts — `terminalKeys`, the `terminals` metadata collection, saved session, activity feed, and daemon status — keyed to a *fixed* host inside the retained owner. Created lazily on first activation, retained across every switch-away, disposed only on membership exit. The exported facades and the metadata/daemon readers become windows over `activeScope().wire`:

| readout | was | now |
| --- | --- | --- |
| terminal keys · `terminals` · session · activity feed | `wire.ts` / `useTerminalMetadata` `useEntry(activeHost)` | `createHostWire` (retained) |
| daemon status | `useDaemonStatus` module `createRoot` `useEntry` | `createHostWire` (retained) |
| `recentRepos` · `recentAgents` · `savedSession` · `savedSessionSub` · `terminalListSub` | `wire.ts` accessors over the re-keying block | `hostScope/activeWire.ts` windows over `activeScope().wire` |

So on a switch-back `daemonStatusPending()` never re-flips true and the composed `terminalCount` never falls to 0 — `resolveCanvasMode` never leaves `workspace`, and the Dock/tiles/RightPanel are never unmounted.

[...]

Issue #1763, opened 2026-07-11
Boot wedge: canvas intermittently sticks at 'Connecting to local…' — cause unconfirmed (the scopedByEntry non-member warn is a benign per-boot transient, see comments)


srid:
## Symptom

Intermittently, a fresh client load sticks at **"Connecting to local…"** and never reaches the workspace; terminal-create wedges. Dev console shows:

    scopedByEntry: active() names non-member local — no owned world this tick (a removal race; a fallback should re-point active). Returning undefined.

Found by #1758's agent while attempting runtime verification, and confirmed **change-independent**: a variant whose connect path was byte-identical to master wedged the same way, and master itself connected only intermittently in dev.

## Mechanism (code-read, not yet root-caused)

- The active host is restored **synchronously from sessionStorage** at boot (`packages/client/src/wire.ts:169-179`), while map membership arrives **asynchronously over the wire** — so `active()` can name `local` before `local` is a member.
- `scopedByEntry` treats that state as an honest transient — "a fallback should re-point active" within a tick (`packages/surface-map/src/scoped.ts:120-240`; the warn is at :236). While it holds, `activeScope()` is `undefined`, so the per-host world (session restore, terminal wiring) doesn't exist and the canvas can't leave the connect state.
- The bug: the state sometimes **persists** instead of resolving — the expected fallback either doesn't fire or loses its own race. That persistence is what needs root-causing.
- Only the **diagnostic** is dev-gated (`NODE_ENV !== "production"`); the underlying race exists in production builds too, silently. Unknown whether production timing can lose it.

## Repro

1. On master: `just dev`, open the client with the devtools console open (dev matters twice: the warn is compiled out of production, and dev's slower boot widens the window).
2. Hard-reload repeatedly. Most loads connect; a bad load sticks at "Connecting to local…" with the warn above.
3. While stuck, terminal-create wedges.
4. If it won't trigger: devtools CPU throttling 4-6x and reload again (window-widening suggestion, not verified).

## Relation to other work

- Distinct from #1714's reprojectClock flicker (fixed by #1758) and from W9 (instant host switch-back, the dispose-on-key-change rebuild) — this is a **boot-ordering** race in the active-host/membership handshake, W7 hostScope family.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

srid:
[...]

PR #1766, opened 2026-07-11
fix(client): ground the active host fed to the per-host scope (#1763 boot transient)

srid:
## What this fixes

A fresh client boot fires this dev warn on **every** load:

```
scopedByEntry: active() names non-member local — no owned world this tick
(a removal race; a fallback should re-point active). Returning undefined.
```

The per-tab `activeHost` is restored **synchronously** from `sessionStorage` a tick before the **async** `padiMap.entries` membership snapshot lands — so at boot `activeHost` names a host membership does not yet ground. Handed straight to `scopedByEntry` (in `hostScope/hostScopes`), that is read as the **removal-race** inhabitant: a `console.warn` + an `undefined` per-host world. At boot that is a **false positive** — nothing departed, membership simply had not arrived.

The fix grounds the accessor the per-host scope reads. New `wire.groundedActiveHost` is `activeHost` **iff it is a current member** (`hostKeys`), else `null` — which is `scopedByEntry`'s honest *no-selection* inhabitant (no warn). `null`, **not** a local substitute, is the correct empty: painting local's owned world under a non-local `activeHost` would be a representable lie, and the departed-active case is re-pointed to local by the one reconcile effect a tick later. `activeHost` stays the non-null intent every other readout (`useEntry`, `foldState`, `padiRpcOf`) keys on.

Net: the class *"kolu hands the per-host world an active host membership does not ground"* becomes **unconstructible**, so the boot-transient warn and the `undefined`-`activeScope` misclassification can no longer arise. `scopedByEntry` keeps its generic warn for any other caller that passes a raw accessor.

## Root cause note — scope honesty

This came out of #1763. Two independent analyses (a wire-delivery trace + a deterministic recovery probe) and ~11 CPU-throttled cold reloads of a live dev build established that the `entries` membership delivery is **self-healing** — the warn fires every boot and every boot self-heals to `connected`. So this warn is a benign, self-resolving boot transient, **not** the cause of the persistent *"Connecting to local…"* wedge #1758 observed (that was correlation — the ever-present warn co-occurring with a separately-wedged load).

**This PR does NOT fix the persistent connect wedge tracked in #1763** — a separate, still-intermittent race (connection/daemon layer, or a multi-host departed-active bounce) that did not reproduce in a local-only warm-padi dev instance. #1763 stays open as that tracker; this change is cited there as context, not a fix. The issue's original mechanism sketch has been corrected/retitled to decouple the warn from the wedge.

## Tests (no existing test weakened)

- `host/groundActive.test.ts` — pure `groundActiveHost` table: boot (empty membership) → `null`; member → the host; departed/never-joined → `null` (never a local substitute); encode-equality.
- `surface-map/scoped.test.ts` **(8)** — a deterministic async-boot regression over the **real** `connectSurfaceMap` slice: a grounded (member-or-null) accessor is **silent** across the boot→membership transition, where the raw accessor **(5)** warns.
- The three per-host-scope suites (`perHostViewState`, `perHostCanvas`, `useSessionRestore`) get `groundedActiveHost` added to their `./wire` mocks (faithful: grounds the mock active against the mock membership). All assertions unchanged.

Full client suite **693/693**, surface-map **38/38**, common **43/43**, typecheck clean, `just fmt` applied.

## Status

- **Merged with `master` (incl. W9 #1764).** W9 landed first; this merged cleanly — the one anticipated `hostScope/hostScopes.ts` conflict resolved exactly as planned (W9's `createHostWire` build fn + this PR's `groundedActiveHost` 2nd arg). Final tree: **`0871ca756`**.
- **Review gauntlet run** (lens-debate → codex-debate → /simplify → code-police, then a docs pass) — each phase's fixes + comments are on this PR. Notable outcomes: a shared `hostKeysInclude` membership predicate (kolu-common), `groundedActiveHost` memoized, a codex-added client-level seam test that fails on a `groundedActiveHost`→`activeHost` regression, and a dev fail-fast guard for the should-never-happen local-absent invariant (so **no** anomaly is silent: boot = honest silence, departure = reconcile toast, local-absent = loud guard).
- **CI (honest partial):** full `ci::default` on **`x86_64-linux`** is **green** on `0871ca756` (14/14 nodes). The **darwin** lane is intentionally **held** pending srid's disposition of the darwin-e2e-mac policy tracked in [#1776](https://github.com/juspay/kolu/issues/1776) — not silently skipped.
- **Draft → ready:** srid merges.


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `c5b9cc667976`

### Applied (4)
- `lowy-1` Membership-by-encode predicate sprayed across host/ — commit `462b06c98`
- `lowy-2` Grounded-accessor composition rebuilt by hand in 4 test mocks — commit `188322cdc`
- `hickey-1` Membership predicate "active ∈ pool by encodeHostKey" duplicated 4× in host/ — (uncommitted)
- `hickey-2` groundedActiveHost mock copy-pasted across four test files instead of the shared testlib — (uncommitted)

### Agreed — no change (3)
- `lowy-3` Grounding placed at the knowing endpoint, not in the primitive — correct (packages/client/src/hostScope/hostScopes.ts:65)
- `lowy-4` Membership derived twice from the one entries store (packages/client/src/wire.ts:365)
- `hickey-3` Two accessors now name 'the active host' (raw activeHost vs groundedActiveHost) (packages/client/src/wire.ts:364)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `c5b9cc667976`

### Round 1

**codex** — approved: `false`

The production grounding logic is coherent and respects the deliberate null, ownership, and membership decisions. I found no runtime correctness, race, fallback, security, or error-handling defect. One minor test-boundary issue remains: the regression test duplicates Kolu policy inside the generic surface-map package without exercising the production wiring it is meant to protect.

Findings:
- `F1` · minor · open — The new test reconstructs the Kolu-specific grounding predicate inline, while the existing surface-map tests already establish that null is silent and a non-member warns. Consequently, all tests would remain green if hostScopes regressed from groundedActiveHost back to raw activeHost; the load-bearing production seam at packages/client/src/hostScope/hostScopes.ts:65 is not pinned. It also places application policy in the generic framework test suite despite the deliberate P5 placement decision. (packages/surface-map/src/scoped.test.ts:253)


**claude** — agreed with F1 and moved the regression from the generic surface-map suite into a falsifiable client-level test that pins the actual production seam.

- `F1` **fixed** — Codex is right on both counts. (1) The surface-map test `(8)` only composed two inhabitants already proven by `scoped.test.ts:168` case (5) — a `null` accessor is silent, a non-member active warns — with an inlined reconstruction of kolu's grounding predicate, so it placed application policy in the generic framework suite (against the deliberate P5 placement) while proving nothing new about surface-map. (2) More importantly it did NOT pin the load-bearing seam at `hostScope/hostScopes.ts:65`: every existing client test (`perHostViewState.test.ts` etc.) drives active and membership together via `switchTo` (add-as-member + setActive in one batch), so raw `activeHost` and `groundedActiveHost` never diverge and a regression of `hostScopes.ts:65` back to the raw accessor would stay green. Fix: deleted test `(8)` from `packages/surface-map/src/scoped.test.ts`, and added `packages/client/src/hostScope/hostScopes.test.ts` — a REAL `scopedByEntry` over the shared mock `padiMap`, driving active (A) and membership SEPARATELY so the boot window is exactly active-without-membership. It asserts `activeScope()` is undefined AND `console.warn` was never called in the boot window, then that it resolves once A joins, still silent. This is falsifiable at the seam: I temporarily regressed `hostScopes.ts:65` to raw `activeHost` and the test FAILED (`expect(warn).not.toHaveBeenCalled()` tripped on scopedByEntry's non-member warn), then restored and it passed. Division of labour is now clean: pure decision table in `host/groundActive.test.ts`, the surface-map inhabitants in `scoped.test.ts` case (5), and the kolu wiring-of-one-into-the-other in the new client test — none duplicating another. Gate green: client + surface-map vitest pass, `pnpm typecheck` clean, `biome lint --error-on-warnings` clean.

commit: `6ba771c726104ec2270b52eddcccdb6109d8d7b9`


### Round 2

**codex** — approved: `true`

The F1 remediation is correct. The Kolu-specific reconstruction was removed from the generic surface-map suite, and the new client test exercises the load-bearing hostScopes accessor seam with raw active intent deliberately diverging from membership. It would fail if hostScopes reverted to activeHost, while confirming the grounded boot window remains silent and resolves when membership arrives. No regression was introduced.

Findings:
- `F1` · minor · resolved — The regression previously lived at the wrong package boundary and did not pin the production hostScopes wiring. (packages/client/src/hostScope/hostScopes.test.ts:67)

_codex on the rebuttal:_ F1: Accepted. The new test directly distinguishes raw activeHost from groundedActiveHost during empty membership, exercises the real hostScopes/scopedByEntry wiring, and verifies both the silent boot state and later resolution. Removing the redundant application-specific surface-map test also restores the intended package boundary.

srid:
## 🧹 /simplify — 4 cleanup agents (reuse · simplification · efficiency · altitude)

**reuse ✔ clean** · **altitude ✔ right** · 2 findings applied (commit `0e2f64799`):

| angle | finding | fix |
| --- | --- | --- |
| efficiency | `groundedActiveHost` was a plain accessor read by 2+ reactive contexts inside `scopedByEntry` (`activeScope`, per-key `isActive`, `activated`); inlining `hostKeys()` made them re-run on **any** membership churn, not just a real switch/ground change | wrap in `createMemo` (own app-lifetime `createRoot`, like `daemonStatusPendingWindow`) — `groundActiveHost` returns `activeHost()`'s stable ref or `null`, so an unrelated pool change recomputes only the cheap memo and stops. Also satisfies the repo's "Memos for multi-consumer derivations" rule. |
| simplification | the ~10-line boot-window/null rationale was **triplicated** across `groundActive.ts`, `wire.ts`, `hostScopes.ts` | let the pure `groundActiveHost` own the full "why"; the two consumers keep only their local facts + the `{@link}`/"see" pointers they already carried |

**Not flagged** (validated by the altitude/reuse agents): `hostKeysInclude` is a genuine generalization at the right layer; the `groundActive.ts` pure/reactive split mirrors the `hostReconcile` precedent; the test mocks are at the right granularity (a fuller `mockWire` helper isn't worth the churn given per-test `padiRpcOf` variance).

No behavior change. Green: client 693/693, surface-map 38/38, common 43/43, tsc clean.

srid:
## 👮 code-police — Pass 1 (rule checklist) + Pass 2 (fact-check)

Two related findings, both about `null` silently collapsing an anomaly. Fixed in `a3c905c71`.

**1 · doc overclaim** (`comment-the-non-obvious` + fact-check) — the lens-added docs on `hostKeysInclude` / `groundActiveHost` claimed the read-side (`groundActiveHost`) and write-side (`hostReconcileTarget`) "agree on membership **by construction**". False for `local`: `hostReconcileTarget` short-circuits `local` (`active.kind === "local" → return null`) *before* the predicate, so it never scans for local while `groundActiveHost` scans uniformly. They share the encode-equality *scan* (can't drift on **how** membership is decided) but are not identical **decisions**. Docs corrected to say exactly that. Also fixed a stale precedent citation (`daemonStatusPendingWindow` no longer exists → `localDaemonStatus`).

**2 · silent-degradation gap** (`no-overloaded-null` + fail-fast doctrine) — because reconcile trusts `local` is always a member and grounding returns `null` when it isn't, a membership snapshot that ever *loads* without `LOCAL_HOST` (a server pool bug violating the unremovable-seed invariant) would strand `activeScope()` at `undefined` (blank canvas) with **no** toast/warn/error. Pre-fix this accidentally tripped `scopedByEntry`'s non-member warn (buried in the every-boot false positive); grounding removed even that. **Fix:** a dev-gated fail-fast guard in `hostScoped` — once membership is non-empty, if it lacks `LOCAL_HOST`, `console.error` loudly. The honest pre-snapshot warming window (empty membership) is excluded.

This keeps the ratified caller-side grounding (no tagged-sum / raw-accessor complexity) while ensuring **no anomaly is silent**: boot = silent (honest warming) · non-local departure = `hostReconcileTarget`'s `toast.warning` · local-absent = this new guard.

**Passes clean on:** dry-rule-of-three (the `hostKeysInclude`/`mockGroundedActiveHost` dedups), no-dead-code, no-silent-error-swallowing, no-unbounded-growth, invalid-states-unrepresentable, and the boot-race timeline / single-`entries.use()` / `createMemo` reference-equality claims (fact-check traced them all and they hold).

Green: client 693/693, surface-map 38/38, common 43/43, tsc clean, fmt clean.

srid:
## Evidence

This change has **no on-screen behavior change** — the boot already reached the workspace before and after (the transient it fixes was benign). The observable delta is entirely in the **dev console**: the misleading every-boot `scopedByEntry` warn is gone. So the evidence is the console state itself, captured on the final merged tree **`0871ca756`** (`just dev-auto`, CPU throttle **6×**, repeated hard reloads).

**BEFORE (pre-fix / `master`)** — the warn fired on **every** cold boot (this is the #1763 signal, confirmed earlier across ~11 throttled reloads):

```
scopedByEntry: active() names non-member local — no owned world this tick
(a removal race; a fallback should re-point active). Returning undefined.
```

**AFTER (this PR, `0871ca756`)** — across **3 cold boots**, that warn is **absent**. The only console output is pre-existing / environmental noise, present before this change too:

```
[warn] computations created outside a `createRoot` or `render` … [7 times]   ← pre-existing (unrelated)
[warn] Automatic fallback to software WebGL … deprecated                     ← headless-Chrome env
[error] The script has an unsupported MIME type ('text/html')                ← dev sourcemap artifact
```

- ✅ No `scopedByEntry: active() names non-member …` on any of the 3 boots.
- ✅ The new dev fail-fast guard (`INVARIANT VIOLATION: membership loaded but LOCAL_HOST absent …`) correctly **does not fire** — local is a healthy member, so the guard stays silent, exactly as intended.
- ✅ The canvas reaches the workspace every time (never stuck on `Connecting to local…`).

Workspace reached on a throttled cold boot (empty state — an unremarkable normal boot, which is the point: nothing user-visible changed):

boot-race #1766 — workspace reached, clean console

*Capture note:* run locally under a strict, incident-proof teardown — dev-auto launched in its own session (`setsid`), torn down by killing **only** that recorded session's PIDs (each verified by session id, never a command/path string match), production `kolu.service` PID verified unchanged before and after, and self-daemonized dev padi/kaval reported for sweep rather than hunted.

srid:
## CI — x86_64-linux green (honest partial)

Full `ci::default` lane on **`x86_64-linux`** (host `srid@naiveintent`), on the final merged tree **`0871ca756`** — **all 14 nodes `success`**, overall commit status `success`:

`_ci-setup` · `install` · `pnpm-hash-fresh` · `nix` · `flake-check` · `atlas-sync` · `fmt` · `biome` · `unit` · `smoke` · `surface-app-example-build` · `surface-example-build` · `home-manager` · `e2e` — all ✅ `@x86_64-linux`.

**Darwin lane is intentionally HELD**, not run: it waits on srid's disposition of the darwin-e2e lane policy tracked in [#1776](https://github.com/juspay/kolu/issues/1776) (the local-padi liveness force-cycle on the managed CI mac). This is an **honest partial** — linux is fully green now; darwin is pending that decision, not silently skipped.

PR #1771, opened 2026-07-12
fix(surface-daemon): boundToPid lifetime — test daemons die with their run

srid:
Test-spawned padi/kaval daemons were leaking past the runs that created them. This adds a third daemon lifetime — `{ kind: "boundToPid", pid }` — so a test/smoke-spawned daemon **dies with the run that spawned it**, making the whole leak class inexpressible. Production daemons are untouched.

## Why they leaked

Every test-spawned padi/kaval is `detached`+`unref`'d (`survivableSpawnDriver`) and runs `lifetime: forever` — deliberately, so it can survive a *server* restart (a product property some scenarios exercise). But that also let it outlive the **run** through three holes, all root-caused and swept off rasam + naiveintent:

1. **Signal death skips teardown.** A signal-killed run never executes its in-process `After`/`afterEach` hooks, so the detached daemons are never reaped.
2. **A crashed run's daemons live at a foreign digest.** `killAll` + the gate reapers are digest-scoped to the *current* state root; a crashed run's daemons sit at a different digest, invisible to the next run's reaper.
3. **`ci/smoke.sh` had no daemon teardown at all** — it killed only its two `kolu-server`s, so **every green smoke run leaked 2 padi + 2 kaval**. Linux CI only looked clean because its Incus containers get destroyed.

## The fix

A third **honest constructor** on `DaemonLifetime` (not `forever` + a flag):

```ts
export type DaemonLifetime =
  | { kind: "forever" }
  | { kind: "idleTimeout"; ms: number; isIdle: () => boolean }
  | { kind: "boundToPid"; pid: number; pollMs? };   // ← new
```

A `boundToPid` daemon watches that pid (reusing the package's canonical `isHolderLive` `kill(pid,0)` probe, polled every ~2 s, `unref`'d) and shuts down cleanly — `{ kind: "shutdown", reason: "pid-gone" }`, exit 0 — the moment the pid is gone. Registration on an already-dead pid exits immediately.

`daemonLifetimeFromEnv(fallback)` is the single home for the *absence = production* policy: it reads `KOLU_DAEMON_BIND_PID`, threaded **harness → server → padi → kaval** via the exported `DAEMON_BIND_PID_ENV` constant, exactly as the existing `KOLU_KAVAL_SPAWN` twin flows. **Absence** (the var truly unset) selects the production `forever`; a set-but-malformed value (empty string, `1e3`, `0x10`, fractional, out of `pid_t` range) **fails fast** — never silently degrades back to `forever`.

The setters bind to the **harness/run pid, never the kolu-server pid** — padi/kaval surviving *server* death is the property under test (the kaval-restart + session-restore scenarios are the canary): the cucumber worker pid in `hooks.ts`, the vitest pid in `dial.test.ts` / `socketDaemon.test.ts` / `padiBinding.test.ts`, and `$$` in `ci/smoke.sh`.

## What it deliberately does NOT do

- **No production surface change** — absence of the env var = `forever`, byte-identical to today. There is deliberately no way to *weaken* a production daemon.
- **No sweeping / reaping heuristics** — the daemon self-reaps; nothing hunts for it.
- **Pid reuse inside the ~2 s poll window is a documented residual**, not engineered around — a recycled pid keeps a daemon alive one extra run at most (vanishingly rare against a large pid space; bounded cost of one run's daemons; a `/proc` start-time cross-check would be darwin-incomplete since `boundToPid` runs on the darwin lane too).

## drishti

`@kolu/surface-daemon` is `@kolu/surface*` source, and this change is additive (a new union arm + new exports). **Verified against `~/code/drishti`: it consumes only `@kolu/surface`, never `@kolu/surface-daemon` (no reference in any source, config, or nix file)** — so no paired drishti PR is forced and the `surface.md` ship-gate does not apply.

## Review & evidence

Full `/be` gauntlet ran on the branch, each stage editing it in turn (see PR comments):
- **architecture-first-principles** → deduped the hand-rolled liveness probe to the package's `isHolderLive`.
- **lens-debate** (consensus) → env var forwarded via the constant; poll interval moved into the `boundToPid` arm; shared `registerPoll` helper.
- **codex-debate** (consensus, 3 rounds, 9 findings) → empty/malformed bind pid now fails fast through both hops; invalid directly-constructed pids validated at consumption; two more real-daemon suites bound; exhaustive lifetime dispatch.
- **simplify** → clean, no changes. **code-police** → deduped the test env save/restore via `vi.stubEnv`.

**Runtime evidence** (in the `## Evidence` comment): a real padi+kaval die **1.97 s** after their harness is SIGKILLed; the control (no bind → they survive = the leak); and a **green `ci/smoke.sh` run that leaks zero** padi/kaval (exercising the production `systemd-run` spawn path).

Docs moved with the API (`ref-surface-daemon.mdx`, the atlas note + regenerated dist, the `make-it-a-daemon` tutorial, a changelog Fixed entry).

---

## Also in this PR: show each daemon's lifetime in the Kaval & Padi dialogs

Having added the lifetime *concept*, the info dialogs now surface it. Each daemon publishes its own serializable lifetime on the wire — kaval as an **optional** field on `system.version`, padi as an **optional** field on its `identity` cell — and both dialogs render a **lifetime** row: `forever` for a durable production daemon, `bound to run pid N` under a test/smoke run, `—` for a survivor predating the field (see the `## Evidence` comment for the live rows).

- `@kolu/surface-daemon` gains a serializable projection `DaemonLifetimeInfo` + `lifetimeInfo(lifetime)` (drops the `isIdle` closure / test-only `pollMs`); the zero-dep spine stays zod-free. Each daemon resolves its lifetime **once** at boot and reuses it for both the served descriptor and `daemonMain`'s actual policy, so the readout can't drift.
- padi mirrors kaval's lifetime across the supervisor via `KavalConnectionMetadata` into `DaemonStatus`; the client renders it through the P4 `connected()` presence with a shared `formatLifetime`, deriving its type from the wire (no spine dependency).
- **No contract-version bump** (the architecture-review fix): a bump would make a surviving pre-field kaval a skew, and a skewed kaval is *recycled — its PTYs killed*. Optional-with-no-bump mirrors the existing `identity` precedent, so a survivor stays adopted and reads `—` until its next restart.
- Docs moved with the API (ref page, atlas notes, changelog Added entry); regression tests guard every wire hop (handshake → metadata → `DaemonStatus` → presentation).


srid:

### Applied (3)
- `lowy-1` Wire-name env var single-sourced for the reader, re-typed as a raw literal by the TS writers — commit `d1f256dbd`
- `hickey-1` Poll interval fragmented from the boundToPid lifetime arm — commit `8cb6ae96f`
- `hickey-2` Poll-timer scaffold duplicated between idleTimeout and boundToPid blocks — commit `c43f07d6c`

### Agreed — no change (1)
- `lowy-2` App-namespaced `KOLU_` literal baked into the generic package breaks symmetry with its own parameterized env reader (packages/surface-daemon/src/daemonMain.ts:61)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 round(s) · codex reviewed at `xhigh` reasoning effort · base `10070a6629c0`
### Round 1

**codex** — approved: `false`

The `boundToPid` model and server → padi → kaval placement are structurally sound, but the branch does not yet meet its fail-fast or leak-prevention claims. PID parsing accepts or suppresses malformed values, unexpected liveness-probe errors become successful shutdowns, several real-daemon test harnesses remain unbound, and the cross-process chain lacks a regression test. Documentation and project-convention issues also remain. No untracked files were present; `git diff --check` was clean.

Findings:
- `F1` · major · open — Only an unset variable is true absence, but an explicitly empty `KOLU_DAEMON_BIND_PID` silently selects `forever`. `Number(raw)` also accepts non-canonical PID spellings such as `1e3`, `0x10`, or padded whitespace. The two forwarding sites truthiness-filter empty values, so under systemd the malformed configuration is discarded before it can fail. This recreates the exact silent daemon leak the fail-fast contract is meant to prevent. (packages/surface-daemon/src/daemonMain.ts:73)
- `F2` · major · open — Numeric but invalid PIDs can be reclassified as a successful `pid-gone` shutdown. The environment parser accepts values above Node's PID range, while direct callers can construct `boundToPid` with zero, negative, fractional, or huge values. `isHolderLive` then converts every error except `EPERM` to `false`; for example, Node rejects `2147483648` with `ERR_INVALID_ARG_TYPE`, but the daemon exits cleanly with code 0. PID 0 and negative PIDs also have process-group semantics rather than identifying one process. (packages/surface-daemon/src/daemonMain.ts:268)
- `F3` · major · open — The leak fix is incomplete across existing real-daemon harnesses. `padiBinding.test.ts` starts detached padi/kaval processes but relies solely on hooks, and `socketDaemon.test.ts` starts real kaval processes with the unmodified parent environment. If Vitest is killed and cleanup hooks do not run, neither harness supplies the run PID, so these daemons retain the `forever` lifetime. This contradicts the changelog's class-wide claim that killed test runs cannot strand padi/kaval daemons. (packages/server/src/padi/padiBinding.test.ts:110; packages/kaval/src/socketDaemon.test.ts:62)
- `F4` · minor · open — The tests prove only the in-process lifetime primitive. The integration harnesses set the variable but still manually reap their daemons, so no test proves the actual server → padi → kaval forwarding chain or that both daemons disappear when the nominated run PID dies. Either manual forwarding hop could regress while the suite remains green. (packages/surface-daemon/src/daemonMain.test.ts:162)
- `F5` · minor · open — The three-arm `DaemonLifetime` union is dispatched with independent `if` statements, leaving `forever` implicit. A future constructor would silently behave like `forever` instead of producing a compile error. This also violates the repository's exhaustive discriminated-union dispatch convention. (packages/surface-daemon/src/daemonMain.ts:243)
- `F6` · minor · open — The new long-lived PID watcher emits no install/retire lifecycle logs, contrary to the repository's watcher lifecycle convention. Operators can see the eventual daemon shutdown but cannot establish whether the PID watcher was installed or cleanly retired. (packages/surface-daemon/src/daemonMain.ts:272)
- `F7` · nit · open — The reference and `daemonMain.test.ts:246` call an arbitrary parsed value a "live-pid value," but the resolver deliberately performs no liveness check; an already-dead PID still selects `boundToPid` and then exits immediately. (website/src/content/surface/ref-surface-daemon.mdx:44)
- `F8` · nit · open — The public reference documents `{ kind: "boundToPid", pid }` but omits the exported `pollMs?` test seam, leaving the published union incomplete. (website/src/content/surface/ref-surface-daemon.mdx:41)
- `F9` · nit · open — The comment names two daemons—padi and kaval—but says "all three die." This mechanism does not bind the kolu-server child to the worker PID, so the statement is misleading. (packages/tests/support/hooks.ts:800)


**claude** — Tightened the bind-pid parse and added consumption-time pid validation (fail-fast), completed the boundToPid bind across the two remaining real-daemon suites, added a cross-process reap regression test, made the lifetime dispatch exhaustive, and corrected the docs/comment wording; disputed the empty-string reclassification, the shared-primitive rewrite, and the watcher-log convention.

- `F1` **partial** — Fixed the real fail-fast gap: `daemonLifetimeFromEnv` now parses a CANONICAL positive decimal only (`^[1-9][0-9]*$`) and rejects out-of-`pid_t`-range values (`daemonMain.ts:100-108`), so `1e3`/`0x10`/` 10 `/`010`/`12.5`/`3000000000` all throw instead of `Number()`-coercing to a wrong pid. Added those cases to the malformed test (`daemonMain.test.ts:266`). **Disputed** the empty-string-must-throw part: empty→fallback is a RATIFIED, tested, documented decision (author note 2; the resolver jsdoc `daemonMain.ts:83-90`), and it is unreachable as a live pid — both forwarders (`padiBinding.ts:145`, `localDriver.ts:93`) truthiness-filter the value, so an empty string is never forwarded as a bind; it only ever means "the forward chose not to bind" = production. I therefore kept the empty-fallback test rather than replacing it. Note: a MALFORMED *non-empty* value IS truthy, so it forwards and throws downstream — codex's "malformed config discarded before it can fail" only applies to empty, which is the non-scenario above. I did not change the forwarders to `!== undefined` because with empty→fallback that would forward `""` to no effect.
- `F2` **partial** — Fixed the concrete failure: added a fail-fast guard at boundToPid CONSUMPTION (`waitForShutdown`, `daemonMain.ts:216-222`) via `isSingleProcessPid` (positive integer ≤ `2**31-1`), thrown BEFORE the promise so no signal listeners leak on the crash path. An out-of-range/zero/negative/fractional pid can no longer be reclassified by `isHolderLive` into a clean `pid-gone` exit; it crashes loudly with the gate+socket released by the `finally`. New test `daemonMain.test.ts:203` covers `[0, -1, 2.5, 2**31]`. **Disputed** the "rewrite `isHolderLive` to rethrow unexpected errors" part: `isHolderLive` is a SHARED primitive (gate stale-reap `pidGate.ts:165`, supervisor) whose EPERM-alive model is ratified (author note 6). Making it rethrow would turn a garbage gate-file pid into a padi BOOT CRASH instead of the current stale-reap-and-retry — a robustness regression. Validating the pid before it can reach `isHolderLive` via the lifetime path fixes the reported failure without destabilizing the gate.
- `F3` **fixed** — Bound the two remaining real-daemon suites to the vitest pid so a signal-killed run can't strand them: `socketDaemon.test.ts:107-119` (beforeAll/afterAll set+restore `KOLU_DAEMON_BIND_PID`, inherited by every `spawnTs`) and `padiBinding.test.ts:117-131` (added `DAEMON_BIND_PID_ENV` to the saved+set env, forwarded to padi→kaval by the binder's env builder). Env is restored after each suite. The vitest pid outlives every test so the 2s watcher never fires mid-suite — SIGTERM/gate-race/restart/leak-sweep assertions are untouched (verified: kaval 23/23, server 24/24).
- `F4` **fixed** — Added a process-boundary regression test (`dial.test.ts:388`): a sentinel process stands in for the run, a REAL padi is spawned bound to the sentinel pid, we wait for padi + its kaval to gate, SIGKILL the sentinel, and assert padi exits 0 AND the kaval self-reaps (its gate released) — proving env → `daemonLifetimeFromEnv` → `boundToPid` reaches a real spawned daemon across the process boundary AND that padi forwards the bind into its kaval, so one kill fells both. Verified green (4.2s).
- `F5` **fixed** — Replaced the two independent `if (lifetime.kind === …)` arms with an exhaustive `switch` closed by `lifetime satisfies never` (`daemonMain.ts:259-292`), matching `daemonExitCode`'s own fence in the same file. A future `DaemonLifetime` kind now compile-fails instead of silently inheriting `forever`. Kept `ts-pattern` out — this leaf package has no such dep and codex explicitly allowed a fenced switch.
- `F6` **disputed** — No install/retire log added. The "watcher installed/retired" convention codex cites is for FILESYSTEM-watcher receptacles that take a logger (`refcounted-dir-watcher.ts`, `wal-subscription.ts`, git/grok session watchers); it is not applied to the in-process lifetime poll timers here. `waitForShutdown` takes NO logger (it's a pure `(lifetime, external?)` function), and its twin poller (`idleTimeout`, `daemonMain.ts:263`) emits no such logs — logging one arm and not the other would be inconsistent. Both endpoints are already observable: boot logs `daemon listening` (`daemonMain.ts:179`) and the reap logs `daemon shutting down {reason: "pid-gone"}` (`daemonMain.ts:184`). Threading a logger into a pure function to narrate a 2s kill-probe is noise, not a convention.
- `F7` **fixed** — Corrected the "live-pid" overclaim (the resolver does no liveness check): renamed the test to "…for a valid positive-integer pid value" (`daemonMain.test.ts`), and reworded the resolver jsdoc (`daemonMain.ts`), the reference table + prose (`ref-surface-daemon.mdx:32,43`) to "a valid positive-integer pid … (no liveness check — an already-dead pid still selects it, then exits at once)".
- `F8` **fixed** — Documented `pollMs?` on the public reference (`ref-surface-daemon.mdx:41-46`): the union row now reads `{ kind: "boundToPid", pid, pollMs? }` with a note that `pollMs?` is test-only and production uses the fixed 2s `PID_WATCH_POLL_MS`. Also kept the atlas type-declaration honest (`surface-daemon.mdx:136`) and rebuilt the atlas dist.
- `F9` **fixed** — Reworded the hooks.ts comment (`hooks.ts:800-809`): "so all three die" → "so BOTH daemons die", with an explicit note that this binds padi + kaval only and the kolu-server is a child reaped by the harness itself, not via this mechanism — keeping the server-survival distinction.

commit: `8e477b74a783ba252b99e162967ebea5c74c6519`


### Round 2

**codex** — approved: `false`

Seven of nine findings are resolved. PID range validation, harness coverage, exhaustive dispatch, documentation, and comments are now addressed without an introduced regression. F1 remains open because an explicitly empty environment value still silently disables binding, and F4 remains open because the new cross-process test bypasses the server→padi forwarding hop.

Findings:
- `F1` · major · open — Canonical parsing and range validation are fixed, but `KOLU_DAEMON_BIND_PID=""` still selects `forever`. An empty environment value is present but invalid, not true absence. The forwarders' truthiness filtering is the defect rather than proof that the value is safe: a broken harness expansion can produce an empty value and silently recreate the daemon leak. (packages/surface-daemon/src/daemonMain.ts:94)
- `F2` · major · resolved — Invalid directly constructed or out-of-range PIDs previously reached `isHolderLive` and could become a clean `pid-gone` result. (packages/surface-daemon/src/daemonMain.ts:236)
- `F3` · major · resolved — Two real-daemon suites previously relied only on teardown hooks and did not bind their daemons to the Vitest process. (packages/kaval/src/socketDaemon.test.ts:108; packages/server/src/padi/padiBinding.test.ts:111)
- `F4` · minor · open — The new sentinel test proves environment parsing in a real padi and the padi→kaval forwarding hop, but it calls `spawnPadi` directly. It bypasses `packages/server/src/padi/padiBinding.ts`, so removing or breaking the server→padi forwarding code would still leave this regression test green. Merely setting the Vitest PID in `padiBinding.test.ts` does not assert that the spawned padi received it because normal teardown masks the difference. (packages/padi/src/dial.test.ts:424)
- `F5` · minor · resolved — Lifetime dispatch previously left `forever` implicit and was not exhaustive. (packages/surface-daemon/src/daemonMain.ts:283)
- `F6` · minor · resolved — The PID polling timer has no watcher install/retire log. (packages/surface-daemon/src/daemonMain.ts:316)
- `F7` · nit · resolved — The documentation previously implied the resolver checked PID liveness. (website/src/content/surface/ref-surface-daemon.mdx:49)
- `F8` · nit · resolved — The public reference omitted the `pollMs?` test seam. (website/src/content/surface/ref-surface-daemon.mdx:41)
- `F9` · nit · resolved — The harness comment incorrectly said three processes were bound by this mechanism. (packages/tests/support/hooks.ts:800)

_codex on the rebuttal:_ F1: I hold firm. The original ratified rule distinguished absence from malformed input; in Node, absence is `undefined`, while `\"\"` is a present invalid PID. Saying the forwarders drop empty values is circular—the silent drop is precisely how a broken harness configuration degrades to `forever`. F2: I concede the shared-primitive dispute. The new consumption-time guard prevents invalid values from reaching `isHolderLive` through this lifetime path, so changing the gate/supervisor primitive is unnecessary and could alter its existing stale-gate behavior. F4: the added sentinel test is valuable but only proves padi→kaval; it directly calls `spawnPadi` and therefore does not cover server→padi forwarding. F6: I concede. A polling timer is not the analogous external resource subscription targeted by the watcher lifecycle rule, and the existing daemon lifecycle logs provide sufficient observability.


**claude** — Conceded F1 (empty `KOLU_DAEMON_BIND_PID` now fail-fasts as present-but-invalid, and both forwarders propagate every defined value so a broken expansion crashes the daemon instead of silently reverting to `forever`) and closed F4 with a focused unit test that guards the server→padi hop; the seven already-resolved findings are untouched.

- `F1` **fixed** — Conceded. In Node an unset var is `undefined` (true absence) but `""` is *present and invalid* (a broken `$SOMEPID` expansion), and silently mapping it to `forever` is exactly how the leak creeps back — so only `raw === undefined` now selects the fallback; `""` falls through to the canonical-decimal check and throws like any other malformed value (`daemonMain.ts:91-93`, jsdoc `daemonMain.ts:78-95`). To make the fail-fast consistent across the whole chain rather than dropping empty mid-hop, both forwarders now forward every DEFINED value (`process.env[X] !== undefined`, was truthiness): `padiBinding.ts:148-149` and `localDriver.ts:96-97`. This keeps the ratified "absence = production forever" property intact (production never SETS the var → still `forever`) while removing the one silent-degrade path codex identified. Test flipped from "empty selects fallback" to "empty crashes loudly" (`daemonMain.test.ts:262-269`); reference reworded (`ref-surface-daemon.mdx:32,48-54`). `just check` green; env-resolver suite 12/12.
- `F4` **fixed** — Conceded that the `dial.test.ts` sentinel proves only padi→kaval (it calls `spawnPadi` directly, bypassing `padiBinding.ts`). Exported `daemonEnv` (`padiBinding.ts:127`) — the exact function that builds the env the server's padi driver spawns with — and added a focused `describe` (`padiBinding.test.ts:729-767`): asserts it forwards the run-bind pid VERBATIM when set, OMITS it when truly unset, and forwards even an empty value (fail-fast propagation). Snapshots the env var in `beforeEach` (runtime, not collection) so restoring returns it to the file-level bind pid, never unsetting it for the later real-padi suites. Mutation-verified: deleting the forward in `daemonEnv` fails 2 of the 3 new assertions (the hop is now a red test if removed). 3/3 pass with the forward present.
- `F2` **fixed** (already resolved round 1; codex conceded the shared-primitive rewrite) — no change.
- `F3` **fixed** (already resolved round 1) — no change.
- `F5` **fixed** (already resolved round 1) — no change.
- `F6` **fixed** (codex conceded; a poll timer is not the external-resource subscription the watcher-log convention targets) — no change.
- `F7` **fixed** (already resolved round 1) — no change.
- `F8` **fixed** (already resolved round 1) — no change.
- `F9` **fixed** (already resolved round 1) — no change.

commit: `caaf721d7c60bb5184ba1ae8aecc5ee252e69aee`


### Round 3

**codex** — approved: `true`

All nine findings are resolved. Empty bind-PID values now fail fast and propagate through both forwarding hops, while true absence alone selects production `forever`. The server→padi hop now has focused regression coverage, complementing the existing cross-process padi→kaval test. No regressions were introduced; the worktree has no untracked files and `git diff --check` is clean.

Findings:
- `F1` · major · resolved — Empty or malformed bind-PID values could silently degrade to `forever`. (packages/surface-daemon/src/daemonMain.ts:91)
- `F2` · major · resolved — Invalid directly constructed or out-of-range PIDs could become a clean `pid-gone` result. (packages/surface-daemon/src/daemonMain.ts:236)
- `F3` · major · resolved — Two real-daemon suites did not bind spawned daemons to the Vitest process. (packages/kaval/src/socketDaemon.test.ts:108; packages/server/src/padi/padiBinding.test.ts:111)
- `F4` · minor · resolved — The server→padi environment-forwarding hop lacked direct regression coverage. (packages/server/src/padi/padiBinding.test.ts:730)
- `F5` · minor · resolved — Lifetime dispatch was not exhaustive. (packages/surface-daemon/src/daemonMain.ts:283)
- `F6` · minor · resolved — The PID polling timer had no watcher install/retire logs. (packages/surface-daemon/src/daemonMain.ts:316)
- `F7` · nit · resolved — Documentation previously implied PID liveness was checked during resolution. (website/src/content/surface/ref-surface-daemon.mdx:49)
- `F8` · nit · resolved — The public reference omitted the `pollMs?` test seam. (website/src/content/surface/ref-surface-daemon.mdx:41)
- `F9` · nit · resolved — A harness comment incorrectly said three processes were bound by this mechanism. (packages/tests/support/hooks.ts:800)

_codex on the rebuttal:_ F1: verified and resolved. The resolver now treats only `undefined` as absence, rejects `""`, and both server→padi and padi→kaval forwarders preserve every defined value. F4: verified and resolved. The focused `daemonEnv` tests directly guard the previously uncovered server→padi hop, including unset and empty-value behavior. F2–F3 and F5–F9 remain resolved and were not regressed.

srid:

Ran Pass 1 (rule checklist) and Pass 2 (fact-check) over the diff vs `10070a6629c0`. `--no-elegance` (the simplify pass already ran this tree).

### Applied
- **dry-rule-of-three** — the hand-rolled bind-pid env save/restore was written three times (`daemonMain.test.ts`, `socketDaemon.test.ts`, `padiBinding.test.ts`), each independently getting the undefined-vs-empty distinction right. Replaced all three with vitest's `vi.stubEnv` / `vi.unstubAllEnvs` (already the repo convention in the surface-remote tests). Commit `4e813a9c9`. Verified: daemonMain 33, socketDaemon 23, padiBinding 27 (real spawns) green.

### Recorded, not fixed (with rationale)
- **prefer-ts-pattern** on `waitForShutdown`'s `switch (lifetime.kind)` — `@kolu/surface-daemon` is the deliberately zero-dep spine (hashed whole into kaval's staleKey) and its **sibling `daemonExitCode` in the same file** already uses `switch` + `satisfies never`. Importing `ts-pattern` for one arm would be inconsistent within the file and add an unwanted dependency to the spine; the `satisfies never` fence gives the same compile-time exhaustiveness.
- **dry (env-forward twin)** — the `DAEMON_BIND_PID_ENV` forward across `padiBinding.ts` / `localDriver.ts` (2 copies) is the deliberate `KOLU_KAVAL_SPAWN` twin pattern, already blessed by the lens debate and simplify; a shared helper would add a cross-package dependency for two lines.
- **pid-reuse-in-poll-window** (fact-check) — the residual **ratified** by the change owner ("documented, not engineered around"). The suggested `/proc/<pid>/stat` start-time cross-check is darwin-incomplete (`boundToPid` runs on the darwin CI lane too, where `/proc` doesn't exist), and the risk is vanishingly small: a 2 s poll window against a large pid space, bounded cost of one run's daemons. Documented in the code, the Reference page, and the PR body.

No other rule violations, silent-swallow patterns, unbounded-growth, or listener/timer leaks. The `registerPoll` cleanup is unconditionally registered and always run; `gate.release()` is idempotent; the env-forwarding chain is consistent end to end including the hostile empty-string case.

srid:
## Evidence

Runtime witness for the fix (no on-screen surface — this is a daemon-lifetime change, so the evidence is at the process level). Captured on the final tree at `4e813a9c9`.

### 1. A harness-bound padi + kaval die with their run (the fix)

A real `padi` (which spawns its own `kaval`, one bind-var hop away) is spawned bound to a stand-in harness pid. `SIGKILL` the harness → **both daemons exit within one ~2 s poll tick** (`ps` + timestamps):

```
commit under test: 4e813a9c9  (Linux x86_64)

== boundToPid live witness ==
[23:02:22.084] harness (stand-in run) pid=3586934
[23:02:23.967] BEFORE kill — daemons bound to harness 3586934:
        PID    PPID     ELAPSED COMMAND
    3586936 3586872       00:01 node --import /home/srid/code/kolu/.worktrees/daemon-cleanup/node_modules/.pnpm/tsx@4.21.0/node_modules/tsx/dist/loader.mjs /home/srid/code/kolu/.worktrees/daemon-cleanup/packages/padi/src/bin.ts --state-root /tmp/btp-sr.gSA0pP --allow-nix-shell-with-env-whitelist default
    3587017 3586936       00:00 /nix/store/h4apbnz8p5fmjwcmzlaxcjm5vr70d97x-nodejs-slim-24.13.0/bin/node --import file:///home/srid/code/kolu/.worktrees/daemon-cleanup/node_modules/.pnpm/tsx@4.21.0/node_modules/tsx/dist/loader.mjs /home/srid/code/kolu/.worktrees/daemon-cleanup/packages/kaval/src/bin.ts --socket /tmp/btp-rt.beTm3O/kaval-0f302bf2160ff9d0/pty-host.sock
[23:02:23.973] SIGKILL harness 3586934
/tmp/claude-1000/-home-srid-code-kolu--worktrees-daemon-cleanup/422447c4-64ce-4607-b73a-89af2a7674e7/scratchpad/boundToPid-witness.sh: line 72: 3586934 Killed                     sleep 600
[23:02:25.951] AFTER kill — padi(3586936) alive=NO  kaval(3587017) alive=NO
[23:02:25.953] both daemons gone 1.97s after the harness died
RESULT: PASS — padi + kaval died with their run (1.97s, ~one 2s poll tick)
```

### 2. Control — WITHOUT the bind, the daemon survives its harness (the leak)

Same spawn with **no** `KOLU_DAEMON_BIND_PID` → absence selects the production `forever`, so both daemons outlive the harness death. This is exactly the leak the fix closes; production is untouched:

```
== CONTROL: NO bind var → daemon runs 'forever', survives the harness death (the leak this PR closes) ==
harness pid=3587538
padi=3587539 kaval=3587853 both up (no bind var set)
SIGKILL'd harness 3587538; waiting 5s (>2 poll ticks)...
/run/current-system/sw/bin/bash: line 16: 3587538 Killed                     sleep 600
AFTER 5s: padi(3587539) alive=YES  kaval(3587853) alive=YES
→ WITHOUT the bind, both survive their harness — the leak. WITH it (above), both die in ~2s.
```

### 3. A green `ci/smoke.sh` run leaks ZERO padi/kaval (the deterministic smoke hole, closed)

`ci/smoke.sh` now sets `KOLU_DAEMON_BIND_PID=$$`, so the padi/kaval its two kolu boots spawn self-reap when the script exits. Measured as a pid-**set** diff (baseline = this dev box's other kolu instances) so the smoke's own daemons are isolated. This box has `KOLU_PADI_BIN` baked, so it also exercises the **production `systemd-run --user` spawn path** (the bind var survives `--setenv`):

```
== SMOKE ZERO-LEAK evidence — commit 4e813a9c9 (Linux x86_64) ==
ci/smoke.sh sets KOLU_DAEMON_BIND_PID=$$; the padi/kaval it spawns must self-reap when smoke.sh exits.

padi/kaval pids BEFORE smoke: 25 (my machine's other kolu instances — the baseline)

--- running ci/smoke.sh (builds .#default, boots kolu twice, each bound to its $$) ---
  smoke| Using saved setting for 'extra-substituters = https://cache.nixos.asia/oss' from ~/.local/share/nix/trusted-settings.json.
  smoke| Using saved setting for 'extra-trusted-public-keys = oss:KO872wNJkCDgmGN3xy9dT89WAhvv13EiKncTtHDItVU=' from ~/.local/share/nix/trusted-settings.json.
  smoke| evaluation warning: Nixpkgs 26.05 will be the last release to support x86_64-darwin; see https://nixos.org/manual/nixpkgs/unstable/release-notes#x86_64-darwin-26.05
  smoke| these 5 derivations will be built:
  smoke|   /nix/store/0f8r699aspcyz9axyy6c6lxgpby1wca8-kaval.drv
  smoke|   /nix/store/fariadngnisrcam38pc6mq36r1m3iilz-kolu-stamped.drv
  smoke|   /nix/store/ipjmi66r32hff9dk2fdxh04sxrz5j086-padi.drv
  smoke|   /nix/store/ai9r327d69j5na0hqbhv3g01nx4wd449-kolu-bin.drv
  smoke|   /nix/store/vdzfrjzgw8cczf6dwvy865dx97jahbkc-kolu.drv
  smoke| this path will be fetched (0.0 KiB download, 262.6 MiB unpacked):
  smoke|   /nix/store/1dhz16iy0zm8kwvvlg5vl0i1lb80625i-kolu-1.1.0
  smoke| copying path '/nix/store/1dhz16iy0zm8kwvvlg5vl0i1lb80625i-kolu-1.1.0' from 'https://cache.nixos.asia/oss'...
  smoke| building '/nix/store/0f8r699aspcyz9axyy6c6lxgpby1wca8-kaval.drv'...
  smoke| building '/nix/store/fariadngnisrcam38pc6mq36r1m3iilz-kolu-stamped.drv'...
  smoke| building '/nix/store/ipjmi66r32hff9dk2fdxh04sxrz5j086-padi.drv'...
  smoke| building '/nix/store/ai9r327d69j5na0hqbhv3g01nx4wd449-kolu-bin.drv'...
  smoke| building '/nix/store/vdzfrjzgw8cczf6dwvy865dx97jahbkc-kolu.drv'...
  smoke| kolu listening at http://127.0.0.1:41067 (pid=3591334)
  smoke| /api/health returned 200
  smoke| shutdown clean
  smoke| KOLU_STATE_DIR honored: /tmp/nix-shell.voPa8s/tmp.0aNmflBAyG/relocated

--- waiting 4s for the bind-pid reap window (2s poll) after smoke.sh exited ---
padi/kaval pids AFTER smoke:  27

DELTA — new pids after smoke: 3591569
3591733
(inspecting whether these are smoke leaks or unrelated concurrent daemons...)
  pid 3591569: /nix/store/sy0c7j0npsq33d9zhnnzvjnzc52f4y0p-nodejs-24.13.0/bin/node --import /ni
  pid 3591733: /nix/store/sy0c7j0npsq33d9zhnnzvjnzc52f4y0p-nodejs-24.13.0/bin/node --import /ni

--- re-check now (the 2 daemons at the 4s mark were mid-reap; the 2nd contract's start later) ---
padi/kaval pids now: 25  (baseline was 25)
DELTA vs baseline: NONE — every daemon the green smoke run spawned has self-reaped.
RESULT: PASS — the deterministic smoke leak is closed (zero leaked padi/kaval).
```

*(At the 4 s mark two of the second contract's daemons were still mid-reap — their 2 s poll cycle lands just past a 4 s window — and were gone a moment later; the final count returns exactly to baseline.)*

**Unit coverage** complements the live witness (they don't substitute for it): a `boundToPid` daemon exits within one poll tick of the watched pid dying, an already-dead pid exits immediately, `forever`/`idleTimeout` unchanged, and `daemonLifetimeFromEnv` throws on every malformed value.

srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `4e813a9c915e`

### Applied (4)
- `lowy-2` Shared daemon humanizer typed and homed under kaval/, reached into by padi — commit `ab5ef5c27`
- `hickey-1` KavalPresence JSDoc orphaned by KavalLifetime insertion — commit `f6f94fe27`
- `hickey-2` Shared formatLifetime typed on a kaval-specific alias; one concept, two client names — (uncommitted)
- `hickey-3` PadiInfoDialog lifetime comment contradicts rendered behavior — commit `3964e2922`

### Agreed — no change (3)
- `lowy-1` Wire-schema duplication now spans two fields — graduation candidate for a shared browser-safe vocab receptacle (packages/kaval/src/ptyHostSurface.ts:247, packages/padi/src/vocab.ts:527)
- `lowy-3` kaval keeps lifetime as a sibling of identity; padi folds it into identity — justified asymmetry (packages/kaval/src/ptyHostSurface.ts:267, packages/padi/src/surface.ts:206)
- `lowy-4` DEFAULT_PADI_IDENTITY seeds lifetime:'forever' rather than undefined (packages/client/src/padi/surface.ts (DEFAULT_PADI_IDENTITY), packages/padi/src/surface.ts:224)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 round(s) · codex reviewed at `xhigh` reasoning effort · base `4e813a9c915e`
### Round 1

**codex** — approved: `false`

The lifetime model and runtime wiring are sound: each daemon resolves the policy once, publishes its projection, and the client gates presentation on an honest connected state. I found no runtime correctness, security, race, fallback, or leak defect. The branch still needs several maintainability, documentation, and regression-test fixes before approval. The transient `.dev-server-sentinel.pid` appeared and disappeared during review and was unrelated runtime noise.

Findings:
- `F1` · minor · open — Both new functions dispatch with per-arm logic using `switch`, contrary to the repository's explicit `prefer-ts-pattern` rule. This bypasses the established exhaustive-dispatch convention. (packages/surface-daemon/src/daemonMain.ts:65; packages/client/src/kaval/daemonPresentation.ts:309)
- `F2` · minor · open — The changed Atlas note still shows `servePtyHost({ log, rcDir })`, but this branch makes `lifetime` a required `InProcessPtyHostDeps` field. The published example no longer type-checks and fails to demonstrate the important same-policy/readout invariant. (docs/atlas/src/content/atlas/surface-daemon.mdx:164)
- `F3` · minor · open — This changelog entry announces a user-visible feature, but neither canonical product page (`website/src/content/docs/kaval.mdx` nor `padi.mdx`) documents where to find or interpret the new diagnostic row. AGENTS.md explicitly requires the matching product documentation to move with a user-facing feature; the surface framework reference documents the mechanism, not the product affordance. (website/src/content/changelog/unreleased.mdx:16)
- `F4` · nit · open — The comments call this a "5.1" field/survivor even though the ratified design deliberately keeps `PTY_HOST_CONTRACT_VERSION` at 5.0. This records a nonexistent contract revision and contradicts the compatibility rationale. (packages/client/src/kaval/KavalInfoDialog.tsx:249; packages/kaval/src/systemVersionShape.test.ts:43)
- `F5` · nit · open — The module header says `useDaemonStatus.ts` re-exports every symbol from this module, but the new `DaemonLifetimeView` and `formatLifetime` exports are not re-exported there. The statement became false in this branch. (packages/client/src/kaval/daemonPresentation.ts:7)
- `F6` · minor · open — The feature's compatibility fields are intentionally optional at every hop, so omitting live values from the producer, connection metadata, or published daemon status still type-checks. Current tests pin optionality and projection shape but never assert that a live Kaval reports the injected lifetime or that it survives the metadata-to-`DaemonStatus` forwarding chain. Padi's schema likewise lacks a parse test for the ratified old payload with no lifetime, and the new formatter has no value-level tests. (packages/kaval/src/systemVersionShape.test.ts:43; packages/kaval/src/contractCorpus.testlib.ts:212; packages/padi/src/ptyHost/daemonStatus.ts:87)


**claude** — Converted the client humanizer to ts-pattern, kept the spine's switch (with an explicit exhaustiveness fence), corrected the "5.1" wording, fixed the stale Atlas snippet + doc header, added value/parse/propagation regression tests, and disputed the product-doc gate as a non-issue.

- `F1` **partial** — Two functions, split verdict. `formatLifetime` (`packages/client/src/kaval/daemonPresentation.ts:305`): CONVERTED to `match(...).with(...).exhaustive()` — the client already depends on `ts-pattern` and the exact call site (`KavalInfoDialog.tsx:238`) dispatches with `match`, so this is a clean consistency win honoring `prefer-ts-pattern`. `lifetimeInfo` (`packages/surface-daemon/src/daemonMain.ts:64`): DISPUTE adding a `ts-pattern` dependency. `@kolu/surface-daemon` is the deliberately minimal-dependency spine (design premise: "zero-dep spine, no zod"; deps are only `@kolu/surface` + `@orpc/server`), and the file's OWN established exhaustiveness idiom is `switch` + `satisfies never` — see the pre-existing `daemonExitCode` (`:147`) and `waitForShutdown` (`:317`). Pulling a dispatch library into one of three sibling switches to satisfy a style rule contradicts both. The rule's stated rationale (silent fall-through to a `default`) does not even apply here: `lifetimeInfo` has a declared `DaemonLifetimeInfo` return type and NO default arm, so a new `DaemonLifetime` variant already fails to type-check. I strengthened that guarantee to match the siblings by adding an explicit `lifetime satisfies never;` fence after the switch — compile-enforced exhaustiveness, no new dep. (typecheck green; `daemonMain.test.ts` 13 pass.)

- `F2` **fixed** — `docs/atlas/src/content/atlas/surface-daemon.mdx:161`. The kaval `bin.ts` snippet showed `servePtyHost({ log, rcDir })` — stale now that `lifetime` is a required `InProcessPtyHostDeps` field, so it no longer type-checks and failed to show the same-policy invariant. Rewrote it to resolve `const lifetime` ONCE, pass `lifetimeInfo(lifetime)` to `servePtyHost` and the SAME `lifetime` to `daemonMain` — demonstrating the no-drift invariant. Regenerated + staged `docs/atlas/dist/surface-daemon.html` via `just atlas::build` (atlas-sync goes green once the rebuilt dist is committed alongside).

- `F3` **disputed** — No product-doc change warranted. The canonical pages `website/src/content/docs/kaval.mdx` and `padi.mdx` are CONCEPTUAL (what the daemons are; how to drive the CLIs) and do NOT enumerate info-dialog rows at all — grep for `dialog|memory|socket|uptime|lifetime` on both returns only CLI/socket prose, none of it made false by this change. Every prior diagnostic row (memory #1503, uptime, socket-path, running-daemons #1686) shipped WITHOUT a product-page entry, establishing that the dialog's diagnostic readouts are documented via the changelog, not these pages. The `website-docs` rule's failure mode is "an existing sentence quietly turned into a lie" — there is no such sentence here. The mechanism IS documented, in the right place: `ref-surface-daemon.mdx` (updated in this PR). Adding a bespoke lifetime-row paragraph would introduce a dialog-row-enumeration pattern neither page follows — new doc surface, not a corrected lie.

- `F4` **fixed** — The counterfactual "5.1" leaked into two spots that read as real revisions. `packages/client/src/kaval/KavalInfoDialog.tsx:249` "pre-5.1-survivor" → "pre-lifetime-field survivor"; `packages/kaval/src/systemVersionShape.test.ts:43` test name "predating the 5.1 field" → "predating the lifetime field". Left `ptyHostSurface.ts:106`'s "5.1" untouched as codex advised — there it is explicitly hypothetical (IF we bumped, it would be 5.1), which is accurate. `PadiInfoDialog.tsx` already said "pre-field survivor" — no 5.1 there.

- `F5` **fixed** — `packages/client/src/kaval/daemonPresentation.ts:7`. The module header claimed `useDaemonStatus.ts` "re-exports every symbol here" — false now (and already loose before: `toKavalPresence`/`KavalPresence`/`channelLive` were never re-exported either). Reworded to "re-exports the slice of this presentation API that existing call sites reach through it," and noted the new pure `match` import in the same header's side-effect-free-imports sentence.

- `F6` **partial** — Added value/parse/propagation coverage on both sides of the seam; deferred only the socket-integration leg. ADDED: (1) `formatLifetime` pins all three arms + `undefined`→"—" (`daemonPresentation.test.ts`, new describe). (2) `toKavalPresence` carries the wire `lifetime` onto the connected presence and leaves it `undefined` for a survivor (same file) — the presentation end of the mirror chain. (3) `SystemVersionOutputSchema` parses an old handshake with NO `lifetime` key (→ undefined) and round-trips one that carries it (`systemVersionShape.test.ts`). (4) `PadiIdentitySchema` the same, for padi's own identity cell (`surface.test.ts`). (5) the contract corpus's `system.version` test now asserts the daemon's injected lifetime APPEARS on the wire (`contractCorpus.testlib.ts`) — the producer end, run over BOTH the in-process and real-socket hosts. DEFERRED: a live socket integration through `connectKaval`/`publishDaemonStatus`. Those two hops are one-line field copies (`connect.ts:111` `lifetime: version.lifetime`; `daemonStatus.ts` `lifetime: status.metadata.lifetime`), each typed against the shared `DaemonLifetimeInfo`/`DaemonStatus` schema — a drift is a COMPILE error, not a silent drop — so standing up a socket+supervisor harness to re-prove a typed passthrough buys little over the schema round-trip + producer + presentation tests now in place. (All added tests green: client 18, kaval 22, padi 12, surface-daemon 13; typecheck + biome clean.)

commit: `cadff2d153788fe2369bfc88496e353645baa1ea`


### Round 2

**codex** — approved: `false`

Five of six findings are resolved. The formatter, Atlas example, compatibility wording, and module documentation are fixed, and I accept the author's reasoned disputes on the spine's locally consistent exhaustive switch and the absence of a product-doc update. F6 remains open because the two optional middle-hop copies are still not exercised and can be deleted without a type error.

Findings:
- `F1` · minor · resolved — The client formatter now follows `prefer-ts-pattern`. The spine retains its switch but now has an explicit `satisfies never` fence consistent with the other exhaustive switches in the same module. (packages/surface-daemon/src/daemonMain.ts:64; packages/client/src/kaval/daemonPresentation.ts:309)
- `F2` · minor · resolved — The Atlas example now resolves one lifetime, publishes `lifetimeInfo(lifetime)`, and gives the same value to `daemonMain`. The regenerated dist file is present. (docs/atlas/src/content/atlas/surface-daemon.mdx:161)
- `F3` · minor · resolved — The existing Kaval and Padi pages intentionally document daemon concepts and CLI operation rather than enumerating diagnostic-dialog rows. No existing product documentation became false, while the changelog and surface-daemon reference document the new behavior. (website/src/content/changelog/unreleased.mdx:16)
- `F4` · nit · resolved — Both misleading references to a real "5.1" field were replaced with accurate pre-lifetime-field wording. (packages/client/src/kaval/KavalInfoDialog.tsx:249; packages/kaval/src/systemVersionShape.test.ts:43)
- `F5` · nit · resolved — The header now accurately says `useDaemonStatus` re-exports a slice of the presentation API and accounts for the new pure `ts-pattern` import. (packages/client/src/kaval/daemonPresentation.ts:7)
- `F6` · minor · open — The new tests now cover schema compatibility, live producer output, presentation, and formatting, but they still construct `DaemonStatus` directly and never execute the two middle-hop copies used in production. Contrary to the rebuttal, deleting either `lifetime: version.lifetime` or `lifetime: status.metadata.lifetime` still compiles because `KavalConnectionMetadata.lifetime` and the connected status field are both optional. The UI would silently regress to "—" while all current tests passed. (packages/padi/src/ptyHost/connect.ts:111; packages/padi/src/ptyHost/daemonStatus.ts:94)

_codex on the rebuttal:_ F1: Conceded. The client conversion is correct, and the spine's explicit `satisfies never` fence provides the exhaustiveness guarantee while matching both sibling switches in that dependency-minimal module. F3: Conceded. The feature does not stale an existing product-doc claim, and the conceptual Kaval/Padi pages do not maintain a dialog-row inventory; the changelog plus framework reference are sufficient here. F6: I hold firm on the remaining middle-hop coverage. The rebuttal's compile-error claim is incorrect for omission: both destination fields are optional specifically for backward compatibility, so either copy line can disappear without a type error. The added producer and presentation tests bracket—but do not traverse—the production bridge.


**claude** — Conceded F6 and closed the middle-hop coverage gap with two mutation-killing tests that traverse the real production bridge.

- `F6` **fixed** — Conceded: codex is right that omission (not drift) is the untested failure — deleting either `lifetime: version.lifetime` (`packages/padi/src/ptyHost/connect.ts:111`) or `lifetime: status.metadata.lifetime` (`packages/padi/src/ptyHost/daemonStatus.ts:94`) still type-checks because both destination fields are optional, so a silent regression to "—" would pass all prior tests. My round-1 "drift is a compile error" rebuttal only covered TYPE drift, not deletion. Added two focused tests that execute the production hops:
  - `packages/padi/src/ptyHost/connect.test.ts` — serves a REAL kaval over a REAL unix socket (`servePtyHostOverUnixSocket` + `createInProcessPtyHost` with `lifetime: { kind: "boundToPid", pid: 4242 }`), calls the real `connectKaval(socketPath)` (the exact `dialSocket` + `stdioLink` path production uses), and asserts `conn.metadata.lifetime` equals the served value — the connect hop.
  - `packages/padi/src/ptyHost/daemonStatus.test.ts` — drives the real `publishDaemonStatus` (with the noop surface ctx, the house pattern from `restartLocal.test.ts`) and reads the stored status back via `readDaemonStatus`, asserting the connected arm's `lifetime` survived; a second case pins that a survivor reporting no lifetime records `undefined`, not a crash — the publish hop.
  - Verified BOTH are genuine mutation-kills: deleting each copy line (fields stay type-green) turns the matching test RED; restoring turns them green. `just check` (typecheck + biome lint --error-on-warnings) clean; `just fmt` clean.

commit: `5fb2e45fa38792aac466a730c922d7791eaf5a4c`


### Round 3

**codex** — approved: `true`

Approved. F6 is now closed by focused tests that execute both real production hops: the Unix-socket `connectKaval` handshake-to-metadata copy and the `publishDaemonStatus` metadata-to-stored-status copy. Each test would fail if its optional-field assignment were removed, and this round introduced no regression. All six findings are resolved.

Findings:
- `F1` · minor · resolved — Discriminated-union dispatch needed exhaustive handling consistent with its module and dependency boundary. (packages/surface-daemon/src/daemonMain.ts:64; packages/client/src/kaval/daemonPresentation.ts:309)
- `F2` · minor · resolved — The Atlas example omitted the newly required lifetime and failed to demonstrate the no-drift invariant. (docs/atlas/src/content/atlas/surface-daemon.mdx:161)
- `F3` · minor · resolved — The initial review requested product-page documentation for the new diagnostic row. (website/src/content/changelog/unreleased.mdx:16)
- `F4` · nit · resolved — Two references incorrectly described the unversioned optional field as "5.1." (packages/client/src/kaval/KavalInfoDialog.tsx:249; packages/kaval/src/systemVersionShape.test.ts:43)
- `F5` · nit · resolved — The module header inaccurately claimed every presentation symbol was re-exported. (packages/client/src/kaval/daemonPresentation.ts:7)
- `F6` · minor · resolved — The optional lifetime could previously be dropped at either production middle hop without a type or test failure. (packages/padi/src/ptyHost/connect.test.ts:61; packages/padi/src/ptyHost/daemonStatus.test.ts:40)

_codex on the rebuttal:_ F6: Conceded and verified. The new `connectKaval` test traverses the actual Unix-socket/dial/stdio handshake and asserts the distinctive bound-to-pid value in connection metadata. The new `publishDaemonStatus` tests execute the real publisher/store path and cover both present and absent lifetime values. Removing either optional-field copy would now fail its corresponding test, which directly addresses the remaining concern.

srid:
## [👮 Code-police](https://agency.srid.ca/) — lifetime-display

Ran Pass 1 (rule checklist) and Pass 2 (fact-check) over the lifetime-display diff (`4e813a9c9...HEAD`). `--no-elegance` (simplify already ran this tree).

### Result
- **Pass 2 (fact-check): no findings.** No error swallowing, honest optional-field fallbacks + `"—"`, exhaustive `lifetimeInfo`, and the added tests are behavioral (guard each wire hop — connect → metadata → publish → presentation — from silently regressing).
- **Pass 1 (rule checklist): one flag, recorded not fixed** — `DaemonLifetimeInfoSchema` is declared twice (kaval `ptyHostSurface.ts` + padi `vocab.ts`). This is the deliberate twin of the existing `PtyHostIdentitySchema` duplication (padi's browser-safe vocab must not import kaval's package), each `satisfies z.ZodType<DaemonLifetimeInfo>`-pinned to the spine type so a drift is a **compile error**, not silent. The lens debate already dispositioned it as a ledger opportunity — *extract on the third such field* (population 2 today). Every other rule: clean or positively reinforced (discriminated unions, `prefer-ts-pattern`, `no-overloaded-null`, `invalid-states-unrepresentable`).

### Preceding stages (all applied on the branch)
- **architecture-first-principles** → caught a real one: the `5.0→5.1` / `2.0→2.1` contract bumps for this cosmetic field would recycle a surviving kaval (killing its PTYs). Fixed to **optional fields, no bump**, mirroring the `identity` precedent (commit `85756f5f`).
- **lens-debate** (consensus) → renamed the shared humanizer's type `KavalLifetime` → daemon-neutral `DaemonLifetimeView`; recorded the schema-dedup ledger note.
- **codex-debate** (consensus, 3 rounds) → added real middle-hop regression tests (unix-socket handshake → metadata, metadata → `DaemonStatus`, both present + `undefined`); fixed the atlas example + comment/doc accuracy.
- **simplify** → clean (reuse, simplification, efficiency, altitude — no changes).

srid:
## Evidence — the lifetime row in the Kaval & Padi dialogs

Driven live against a local kolu built from this branch (`5fb2e45f`), with `KOLU_DAEMON_BIND_PID` set to a live sentinel pid so the row shows the *interesting* value. Each daemon's info dialog now renders a **lifetime** row beside build/socket/memory.

**Kaval dialog** — `lifetime: bound to run pid 3707217`, at contract **v5.0** (no bump):

Kaval dialog lifetime row

**Padi dialog** — the same, at contract **v2.0** (no bump):

Padi dialog lifetime row

The **v5.0 / v2.0** chips confirm the architecture fix: the field is optional with **no contract-version bump**, so a surviving pre-field kaval is never recycled (PTYs kept). In production the row reads **`forever`**; a survivor daemon predating the wire field reads **`—`** (unit-tested). The dev instance was torn down by its random ports afterward; production kolu was provably untouched (same PID + uptime).

PR #1712, opened 2026-07-06
fix(client): stop the restore-card flashing on page reload

srid:
On a browser **page reload** (server alive, terminals live), the user briefly saw the **"Restore session" card** before the real workspace appeared — the canvas cycled `connecting → empty(restore card) → workspace`. Worse with a heavy session (the reported case had 7 terminals + 4 agents).

## Root cause

`getMetadata` (`useTerminalMetadata.ts`) collapses **two distinct record states** into one `undefined`:

- a record that **hasn't arrived yet** (still composing), and
- a record that **arrived PARKED** (a reboot's restore-card row).

Both drop out of `terminalIds()`, so its length is `0` in both. The loading gate in `useSessionRestore.ts` then tried to reconstruct that lost bit from the **wrong signal** — `savedSessionSub.pending()`, an unrelated cell that resolves *first*. Once it resolved, `isLoading` flipped `false` while per-terminal metadata was still in flight, `terminalCount` was still `0`, and `resolveCanvasMode` returned `{ kind: "empty" }` → the restore card flashed. A beat later metadata composed → `workspace`.

The tricky constraint: a **genuine reboot** is the legitimate empty case (records arrive parked, count stays `0` forever, card *should* show), so `count === 0` alone can't tell reload-flicker from real-reboot. The distinguishing fact — *has every listed record arrived?* — was being destroyed one hop before the decision that needed it.

## The fix

Make the tri-state explicit at the source instead of proxying around the collapse:

1. **`useTerminalMetadata.ts`** — a `recordPhases()` census `{ awaited, parked, live }`, counted from the raw `terminals.byKey(id)?.()` read **above** the parked-collapse. `getMetadata` is unchanged for tile consumers.
2. **`canvasModeResolver.ts` / `useCanvasMode.ts`** — a `recordsAwaited` fact and a new **pure** precedence arm: `awaited > 0` with `terminalCount === 0` → `connecting`, ordered **above** `empty` (same shape as the existing `down`/`warming` beats-`empty` arms). The reload-vs-reboot decision now lives in the tested pure function, not the loading gate.
3. **`useSessionRestore.ts`** — `isLoading` demotes the session cell to its one honest job ("is there a blob to offer?" for the genuinely-empty boot) and reads `recordPhases().awaited` for the metadata timing.

[...]

PR #1772, opened 2026-07-12
Persist host-map membership across restarts (W10)

srid:
> ## 📚 Stacked on #1775 (the retire-verb split)
>
> This now sits on **#1775**, which delivers the framework fix W10 needed. The root cause — one verb (`pool.remove`) carrying both *user-remove* (persist) and *system-retire* (#1708 pump-death, must not persist) — is fixed there: kolu's guest pump-death path now calls **`pool.retire`**, which tears the host out of the live pool **without** persisting, so a transient guest fault no longer un-remembers a host. W10's earlier `.catch` guard on `pool.remove` was dropped in the rebase (retire provably can't reject).
>
> **Carried findings, resolved:** F2 (a guest pump-death permanently forgot a host) — fixed by the split's `retire`. F3 (a dead re-serve mirror stranded on retire) — kolu's per-host `reServeSurface` cache is evicted by `pruneToMembers` wired to `pool.subscribe`, which fires on `retire`'s membership drop, so the mirror is pruned.
>
> **Merge order:** split **#1775** first, then this. Base is `w11-retire-split`; it retargets to `master` automatically once #1775 merges. Tracked as **W11** in the padi note (#1770).
>
> _CI + evidence run on the final (rebased) tree — pending a free pu box (shared-pool saturation)._

---

**The remote hosts you add from the selector strip now survive a kolu restart.** Today a restart forgets everything but the local default and you re-add your fleet by hand; after this, every remembered host reappears in the strip and reconnects through the normal connect pipeline — a host that has since gone away shows its honest **failed** chip with the cause rather than silently vanishing. The only way a host leaves the strip is your explicit remove.

Membership is a **server** fact — kolu-server's pool (`buildRemotePool`) is its one writer — so its memory lives beside that authority, never in the browser, where localStorage would fork the one list into per-device copies. The persisted artifact is a value replaced whole: a zod-schemed `{ version: 1, hosts: string[] }` JSON of **encoded host keys only**, beside `conf`'s `config.json` under `KOLU_STATE_DIR`, written atomically (tmp + rename, async so it never blocks the serving loop).

### The seam this wires (srid-approved over the brief's sketch)

The brief sketched a hand-rolled write *inside the add/removeHost handlers*. But the pool already ships a purpose-built, transactional `persist` hook (from #1714) that kolu's own `buildRemotePool` call never wired. Hand-rolling would duplicate it — the "parallel hand-rolled mechanism" the *reuse the existing source of truth* rule names as a defect — so we **wired the existing seam** instead:

| | Handler-side write (brief's sketch) | Pool `persist` hook (this PR) |
| --- | --- | --- |
| Ordering | writes *after* `pool.add` commits → disk/memory can diverge on failure | write ordered **before** the in-memory commit |
| Concurrency | two rapid strip-adds race the file | serialized through the pool's one mutation queue |
| Failure | host live but unpersisted | just-built session **rolled back** |

### Boot: seed, don't replay

```
parseKoluPadiHostSeed()  ─┐
                          ├─▶ dedup ─▶ initialHosts ─▶ buildEntry ─▶ W6 connect pipeline
loadPersistedHosts(file) ─┘
```

Persisted hosts are merged into `initialHosts` (deduped against the env seed via order-preserving `new Set`) rather than re-added via a post-build `pool.add` loop. Both flow the identical connect pipeline, but seeding at construction *doesn't re-fire `persist`* — so the file is only ever rewritten by a genuine runtime add/remove, and **an interrupted boot can't truncate it**.

### Design philosophy

- **Fail fast** — a file that exists but fails the schema *crashes the boot* naming the path (delete it to recover). Never start-with-empty-fleet, which would silently eat the user's hosts (`caught-error-must-not-collapse-to-empty`). This is why it's a **separate** file, not the `conf` store: `conf` data is reconstructible so a corrupt file resets to defaults; a fleet is not. Only `ENOENT` reads as "fresh install"; a permission error surfaces.
- **Electricity** — persistence policy is app-level (`packages/server` only); the durability *mechanism* is the framework's `persist` contract. No framework change in *this* PR — that's exactly the split PR's job.
- **Reuse the source of truth** — the pool's `persist` seam, `KOLU_STATE_DIR` (via a narrowed export from `state.ts`), the pool's own `encodeHostKey` vocabulary, and a shared `isEncodedHostKey` promoted into `kolu-common` (de-duping a copy the client held).

### Review gauntlet outcome

- **architecture-first-principles** — 1 confirmed defect **fixed** (W10 made `pool.remove` rejectable; guarded the fire-and-forget guest-retire so a disk-write failure can't crash the whole server) + 3 minors folded in. It also surfaced the deeper verb-conflation that became the block above.
- **lens-debate (lowy ⇄ hickey)** — consensus, 1 round, 1 fix (key the persist-exclusion off `LOCAL_HOST`, not the boot-default).
- **codex-debate** — *stopped as non-converging* after 4 rounds (findings grew 13→16, never approved) when its author spiralled into an out-of-scope retry/overlay mechanism that generated its own blocking bugs. Reset to the lens-approved tree and triaged codex's **round-1** review (of the clean diff) by hand — the legit in-scope findings are applied (fail-fast read vs `existsSync` masking `EACCES`, `0600` mode on the ssh-target file, APM-source sync, atlas doc-sync). The full round-1–4 codex trail is posted as a comment for audit.
- **simplify** — efficiency + altitude clean; applied a reuse de-dup and dropped dead test state.
- **code-police** — 1 fix (async fs on the persist path). Two findings **recorded, not actioned** because they contradict the ratified plan: *prefer-focused-library* (use `conf`) — the plan mandates a zod JSON tmp+rename kept outside conf's ladder; *persisted-schema-stays-tolerant* (filter, don't crash) — the plan mandates a fail-fast crash-with-path, with a dedicated test. Surfaced for srid.

Plan of record: the W10 section of `docs/atlas/src/content/atlas/padi.mdx` (#1770); the block resolution is minted there as **W11**.

---

### Carried finding → resolved in the W11 split

**F3 (codex round-1, residual):** if the internal guest-retire's teardown fails, a **dead re-serve mirror can be left stranded in both cache layers** (`reServes` in `index.ts` and the pool's `entries`). This is legitimately the **W11 split's** world — the retire path is being rebuilt there (the non-persisting internal retire), so the mirror-eviction ordering is fixed as part of that rebuild, not patched here. Named here so it isn't dropped; tracked on the W11 PR.

_Disposition table for the full codex trail is in the coordinator report; the round-1 in-scope fixes are applied on this branch (fail-fast read, 0600 + fsync durability, env-seed provenance, strict schema, doc-sync)._


srid:

### Applied (1)
- `hickey-1` Persist-exclusion keyed off defaultHost (boot-default) instead of LOCAL_HOST (the actual invariant) — commit `ec956d75f`

### Agreed — no change (7)
- `lowy-1` Membership-persistence boundary is sound (electricity verdict) (packages/server/src/hostPersistence.ts:1)
- `lowy-2` Low-level codec `savePersistedHosts` is public but has only a test consumer (packages/server/src/hostPersistence.ts:98)
- `lowy-3` `encodeHostKey(defaultHost)` recomputed in two construction sites (packages/server/src/index.ts:339)
- `hickey-2` Dedup rationale stated twice within 15 lines (packages/server/src/index.ts:271-288)
- `hickey-3` savePersistedHosts exported solely to be tested directly (packages/server/src/hostPersistence.ts:88 (savePersistedHosts))
- `hickey-4` version:z.literal(1) promises a migration ladder that does not exist (packages/server/src/hostPersistence.ts:42-43)
- `hickey-5` persist writes ALL members, conflating env-seeded (declarative) and strip-added (durable) provenance (packages/server/src/index.ts:331-332)

srid:
## 🤖 Codex ⇄ Claude debate — STOPPED as non-converging (not a consensus)

The codex debate was **halted after 4 rounds (~70 min)**, not run to consensus. It was diverging, not converging: codex's finding count **grew every round (13 → 14 → 15 → 16)** and it never `approved`, because the Claude author spiralled into building an **out-of-scope retry / `faultRetired`-overlay mechanism** to satisfy a finding — and that invented machinery generated its own fresh `blocking` findings (a rollback hole in the new manual-persistence path, a retry-timer race). The branch was **reset to the lens-approved tree** (the overlay never ships), and codex's **round-1 review of the clean diff** was triaged by hand instead.

Per srid's request, the **full round 1–4 codex trail is included verbatim below** so every discarded finding can be checked to be about its invented overlay (rounds 2–4's `index.ts:6xx` / `hostPersistence.ts:1xx` locations are lines that only existed *inside* the author's spiralled edits), not about the actual diff. The **round-1** findings (the review of the real change) that were legitimate and in-scope are applied on the branch; the one real-but-out-of-scope finding (F2, the retire-verb conflation) is the **W11 split** this PR is now blocked on.

<details><summary><b>Full codex trail — rounds 1–4 (verbatim)</b></summary>

```text
########## CODEX ROUND 1 ##########
approved: false
summary: The core approach is coherent, but the current implementation has several serious state-lifecycle defects. Internal mirror failure can permanently forget a user’s host, failed retirement can leave a cached dead mirror, environment-only seeds become persistent after unrelated mutations, and synchronous/non-durable filesystem writes undermine the persistence contract. Documentation and generated-source records are also inconsistent with the new architecture. 

[major] F1 @ packages/server/src/index.ts:331
  issue: The persistence hook receives the entire pool membership, so environment-only `KOLU_PADI_HOST` seeds become permanently remembered after any unrelated add/remove. Without a mutation they remain transient, making durability depend on unrelated user activity and complecting declarative seeds with strip-added intent.
  suggestion: Track provenance at boot. Exclude seed-only hosts from persistence while retaining hosts that were already in `hosts.json`; add a regression covering boot with an env seed, mutate another host, unset the env, and restart.
  status: open

[blocking] F2 @ packages/server/src/index.ts:458
  issue: An internal guest re-serve fault calls the persistence-enabled `pool.remove`, permanently deleting that host from `hosts.json`. This violates the feature’s central promise that remembered hosts leave only through explicit user removal and turns a runtime/software fault into user-data loss.
  suggestion: Separate runtime retirement from durable user removal. Preserve desired membership when a guest mirror fails—rebuild/report failure or restart safely—and add a fault-then-restart regression proving the host remains remembered.
  status: open

[major] F3 @ packages/server/src/index.ts:450
  issue: If automatic retirement’s persistence write rejects, the catch logs that the host remains live, but `r.done` has already rejected and `reServes` still caches the dead mirror. Future lookups return that mirror forever, and the reconnect action only rechecks the session; it cannot restart the stopped mirror.
  suggestion: Do not swallow this into a supposedly live member. Fix alongside F2 by explicitly rebuilding/evicting the mirror without forgetting membership, or terminate and restore safely if the invariant cannot be reconciled.
  status: open

[major] F4 @ packages/server/src/hostPersistence.ts:110
  issue: `writeFileSync` and `renameSync` run from add/remove RPCs on the server’s single event loop. A slow or hung `KOLU_STATE_DIR` filesystem freezes every HTTP, WebSocket, and terminal request, violating the project’s serving-loop rule.
  suggestion: Make runtime saving asynchronous with `node:fs/promises` and return/await that promise through `savePoolMembership` and the existing pool persistence hook. A synchronous one-shot boot read can remain synchronous.
  status: open

[major] F5 @ packages/server/src/hostPersistence.ts:110
  issue: Temp-write plus rename provides atomic visibility but not the claimed crash durability. The temp file is never fsynced before rename and the parent directory is never fsynced afterward. A machine/power failure can lose the committed mutation—or, on initial creation, lose the directory entry and make the next boot treat the fleet as absent.
  suggestion: Keep the deliberate hand-roll, but write through an open file handle, fsync and close it, rename it, then fsync the containing directory. Add failure-ordering/cleanup coverage.
  status: open

[major] F6 @ packages/server/src/hostPersistence.ts:75
  issue: `existsSync` collapses lookup failures such as an inaccessible parent directory to `false`, so some unreadable-file states are treated as a fresh install and return `[]`. That contradicts the explicit fail-fast requirement for an existing but unreadable fleet.
  suggestion: Read directly and catch only `ENOENT` as absence. Wrap and rethrow every other filesystem error with the path and original cause.
  status: open

[minor] F7 @ .claude/rules/state.md:14
  issue: This generated file was edited directly while `.apm/instructions/state.instructions.md` remains unchanged. The next APM regeneration will erase the new rule, and other generated agent instruction surfaces remain stale.
  suggestion: Move the paragraph into the `.apm` source, extend its `applyTo` paths to cover `hostPersistence.ts`, then run `just ai::apm` and verify all generated outputs.
  status: open

[minor] F8 @ packages/server/src/hostPersistence.test.ts:25
  issue: The test teardown clears only the tracking array and deliberately leaves every temporary directory and file behind. Repeated watch runs therefore grow `/tmp` without bound; the OS does not clean these per test or per process.
  suggestion: Import `rmSync`, recursively remove every tracked directory in `afterEach`, then clear the array.
  status: open

[minor] F9 @ packages/server/src/hostPersistence.ts:48
  issue: The schema does not enforce its stated guest-membership invariant: it accepts `local` and duplicate hosts. Construction’s `Set` silently normalizes those invalid persisted states instead of the promised fail-fast behavior.
  suggestion: Require every decoded key to be remote and require uniqueness in the schema. Add rejection tests for `local` and duplicate entries.
  status: open

[minor] F10 @ website/src/content/docs/architecture.mdx:46
  issue: The canonical architecture page still says state lives below kolu-server, but this change deliberately gives kolu-server ownership of durable fleet membership. The stack table and server/package description omit that new responsibility.
  suggestion: Qualify the diagram as terminal/workspace state and document server-owned fleet membership in the stack table and package map.
  status: open

[minor] F11 @ docs/atlas/src/content/atlas/remote-surfaces.mdx:18
  issue: The architectural record still explicitly says runtime host membership is ephemeral and nothing is persisted or parsed at boot. Meanwhile `hostPersistence.ts` cites “W10 — padi.mdx”, but `padi.mdx` has no W10 phase and still presents W9 as next.
  suggestion: Supersede the old ephemeral-membership decision and record W10’s authority, file format, boot behavior, and migration policy in the Atlas source, then run its build/sync workflow.
  status: open

[nit] F12 @ packages/server/src/hostPersistence.ts:65
  issue: `PersistedHosts` is exported only to support a test cast and lacks the required exported-type documentation. The low-level `savePersistedHosts` implementation detail is likewise unnecessarily public outside the module.
  suggestion: Keep the type and low-level writer private and test through the module’s real load/pool-membership surface; otherwise document why they are public.
  status: open

[minor] F13 @ packages/server/src/hostPersistence.ts:110
  issue: The replacement file is created using the ambient umask, normally producing mode `0644`. Every rename can therefore expose SSH usernames and hostnames—or widen a previously secured file—when the configured state directory is traversable by other users.
  suggestion: Create and enforce the temp file as owner-only (`0600`) before writing/renaming, including when reusing a leftover temp file, and add a mode assertion.
  status: open

########## CODEX ROUND 2 ##########
approved: false
summary: Seven findings remain open. The async I/O, fail-fast read, generated-source, cleanup, schema, Atlas W10, and test-seam fixes are adequate. However, dual-provenance hosts can still be forgotten, fatal guest faults still delete durable intent, the dead-link cache remains after failed retirement, power-loss durability and stale-temp permissions remain incomplete, and two documentation inconsistencies remain. 

[major] F1 @ packages/server/src/hostPersistence.ts:168
  issue: The fix excludes every declaratively seeded host, losing dual provenance. If `remote:a` was already strip-added in `hosts.json`, then later also appears in `KOLU_PADI_HOST`, any unrelated mutation filters it out of the file; unsetting the environment and restarting silently loses `a`.
  suggestion: Load persisted hosts separately and exclude only seed-only keys. Retain seeded keys that were already persisted at boot, and add the overlap regression: persisted A + env-seeded A + mutate B + unset env + restart.
  status: open

[blocking] F2 @ packages/server/src/index.ts:470
  issue: A fatal guest re-serve fault still invokes persistence-enabled `pool.remove`, permanently deleting remembered user intent without explicit removal. Restricting the trigger to structural faults does not satisfy the documented persistence contract.
  suggestion: Preserve durable membership. A safe in-worktree option is to treat an unrecoverable pump fault as fatal and restart, as the local arm does; a richer runtime-retirement design can be used if implemented without losing remembered intent.
  status: open

[major] F3 @ packages/server/src/index.ts:461
  issue: Deleting `reServes` does not actually make the next lookup rebuild the mirror. `serveHostMap` separately caches the `directLink` in its private `links` map and only evicts it when membership detaches. If `pool.remove` rejects, membership never changes, so callers continue using the link to the dead router and `reServeFor` is never called again.
  suggestion: Invalidate both cache layers or take the safe fatal/restart path. Add a regression where guest `done` rejects, persistence rejects, and a subsequent map resolution proves a fresh live link is constructed.
  status: open

[major] F4 @ packages/server/src/hostPersistence.ts:142
  issue: Resolved: runtime persistence now uses asynchronous `node:fs/promises` operations and returns the promise through the pool hook, which already awaits it before committing.
  suggestion: No further change required.
  status: resolved

[major] F5 @ packages/server/src/hostPersistence.ts:129
  issue: The revised comment acknowledges the gap, but the persistence contract remains inadequate: save can report success before file contents and the directory entry are durable. A power/kernel restart can restore an old list—resurrecting an explicitly removed host—or lose initial creation, which the loader silently interprets as an empty fresh fleet.
  suggestion: Fsync the completed temporary file before rename and fsync the containing directory afterward. Keep the operation asynchronous and propagate failures through the existing persistence hook.
  status: open

[major] F6 @ packages/server/src/hostPersistence.ts:96
  issue: Resolved: the loader now maps only `ENOENT` to absence and surfaces all other filesystem failures with the path. The ENOTDIR regression verifies the fail-fast behavior.
  suggestion: No further change required.
  status: resolved

[minor] F7 @ .apm/instructions/state.instructions.md:3
  issue: Resolved: the canonical APM source now contains the rule and widened path, and the generated Claude rule, server AGENTS file, and lock hash were regenerated consistently.
  suggestion: No further change required.
  status: resolved

[minor] F8 @ packages/server/src/hostPersistence.test.ts:26
  issue: Resolved: teardown recursively removes every tracked temporary directory before clearing the list.
  suggestion: No further change required.
  status: resolved

[minor] F9 @ packages/server/src/hostPersistence.ts:53
  issue: Resolved: the persisted schema now rejects local, malformed, and duplicate host keys, with direct regression coverage.
  suggestion: No further change required.
  status: resolved

[minor] F10 @ website/src/content/docs/architecture.mdx:46
  issue: The stack table now records server-owned fleet membership, but the rendered diagram still states without qualification that “state lives below the line,” while durable fleet state lives in kolu-server above it. The author’s intended narrower meaning is not present in the graphic.
  suggestion: Qualify the label as “terminal/workspace state lives below the line.”
  status: open

[minor] F11 @ docs/atlas/src/content/atlas/padi.mdx:199
  issue: Resolved: W10 is now recorded in the roadmap and detailed section, the old ephemeral-membership decision is explicitly superseded, and generated Atlas output reflects those changes.
  suggestion: No further change required for the original F11 issue; see F14 for a new inconsistency introduced while updating the roadmap.
  status: resolved

[nit] F12 @ packages/server/src/hostPersistence.ts:79
  issue: Resolved: `PersistedHosts` is now documented, and the direct low-level writer test follows the established repository pattern of exporting package-internal persistence seams for focused tests.
  suggestion: No further change required.
  status: resolved

[minor] F13 @ packages/server/src/hostPersistence.ts:148
  issue: The mode fix is incomplete because `writeFile(..., { mode: 0o600 })` applies the mode only when creating the temp file. If the fixed `.tmp` sibling already exists, opening it with the default `w` behavior preserves its old mode. The immediately preceding implementation could leave a 0644 temp after a crash, which would then be renamed into place as 0644.
  suggestion: Use a fresh exclusive temp file or explicitly chmod/fchmod the opened temp to 0600 before rename. Add a regression that precreates `${path}.tmp` as 0644.
  status: open

[minor] F14 @ docs/atlas/src/content/atlas/padi.mdx:185
  issue: This round newly marks W9 shipped in the frontmatter, roadmap, and summary, but the canonical W9 section and callout still say “in review.” The generated Atlas page exposes both claims simultaneously.
  suggestion: Update the W9 heading and callout to shipped/historical wording, then regenerate the Atlas output.
  status: open

########## CODEX ROUND 3 ##########
approved: false
summary: Most prior findings are resolved, including dual-provenance handling, successful-path fault retirement, durability fsyncs, file permissions, and documentation consistency. Approval remains blocked by three correctness defects: the new fault-retirement overlay is not transactionally coordinated with explicit actions, persistence failure still leaves permanently dead cached mirrors, and post-rename fsync errors can make disk commit while the pool rolls back memory. 

[major] F1 @ packages/server/src/index.ts:288
  issue: Seeded and persisted host provenance could be conflated, causing a dual-provenance host to disappear after the environment seed was removed.
  suggestion: Retain persisted provenance when computing which declarative seeds must be excluded from writes.
  status: resolved

[blocking] F2 @ packages/server/src/index.ts:570
  issue: The new `faultRetired` state is not transactionally coordinated with pool mutations. `addHost` removes the marker before `pool.add` succeeds, so a failed add followed by another mutation can silently remove the still-durable host. More directly, `removeHost` clears the marker and calls `pool.remove`, but removal is a no-op when the fault-retired host is absent from the runtime pool, leaving it in `hosts.json` to reappear after restart.
  suggestion: Serialize durable-intent changes with pool mutations. Retain or restore the marker when an add fails, and provide an explicit-removal path that persists deletion even when the host is already absent from runtime membership.
  status: open

[major] F3 @ packages/server/src/index.ts:486
  issue: If persistence fails during automatic fault retirement, pool membership does not change and neither cache is evicted. Both caches therefore retain the mirror whose pump has already terminated. Cache consistency does not provide liveness, and unrelated membership mutations retain that host and do not rebuild either cached object.
  suggestion: Ensure this failure cannot leave a cached dead mirror: retry retirement, invalidate and rebuild both cache layers through their owning abstraction, or fail fast instead of continuing with an unrecoverable runtime member.
  status: open

[major] F4 @ packages/server/src/hostPersistence.ts:142
  issue: Runtime persistence used synchronous filesystem operations on the server event loop.
  suggestion: Use asynchronous filesystem operations while retaining a synchronous one-shot boot read if desired.
  status: resolved

[major] F5 @ packages/server/src/hostPersistence.ts:159
  issue: The atomic rename sequence did not fsync the file and containing directory, so acknowledged creation or removal could be lost across a power or kernel crash.
  suggestion: Fsync the completed temporary file before rename and the containing directory after rename.
  status: resolved

[major] F6 @ packages/server/src/hostPersistence.ts:96
  issue: The existence-check/read sequence could treat filesystem errors as an absent persistence file, violating fail-fast behavior.
  suggestion: Attempt one read and treat only `ENOENT` as absence; surface every other error.
  status: resolved

[minor] F7 @ .apm/instructions/state.instructions.md
  issue: A generated runtime rule was edited without updating its APM source, so regeneration would erase the change.
  suggestion: Update the APM source and regenerate all derived files and hashes.
  status: resolved

[minor] F8 @ packages/server/src/hostPersistence.test.ts:26
  issue: Tests left temporary directories behind, causing repeated or watch runs to accumulate files.
  suggestion: Recursively remove every tracked test directory during teardown.
  status: resolved

[minor] F9 @ packages/server/src/hostPersistence.ts:53
  issue: The persisted schema accepted local and duplicate host keys despite documenting a unique remote-host list.
  suggestion: Reject local keys and duplicates during schema validation rather than silently normalizing them later.
  status: resolved

[minor] F10 @ website/src/content/docs/architecture.mdx:46
  issue: The architecture diagram broadly claimed all state lived below the server boundary even though durable fleet membership now lives in `kolu-server`.
  suggestion: Qualify the diagram to identify terminal and workspace state specifically.
  status: resolved

[minor] F11 @ docs/atlas/src/content/atlas/padi.mdx
  issue: The roadmap and historical design note still described host membership as ephemeral and lacked the persistence phase implemented by this change.
  suggestion: Record W10 in the roadmap, supersede the obsolete design claim, and regenerate Atlas output.
  status: resolved

[nit] F12 @ packages/server/src/hostPersistence.ts:79
  issue: The low-level persistence type and writer were exported primarily as test seams without making that intent clear.
  suggestion: Keep the established direct-testing seam but document its package-internal testing purpose.
  status: resolved

[minor] F13 @ packages/server/src/hostPersistence.ts:164
  issue: Opening a pre-existing temporary file with mode `0600` does not change its existing permissions, allowing a stale wider-mode file to be renamed into place.
  suggestion: Explicitly chmod the opened temporary file before writing and renaming it.
  status: resolved

[minor] F14 @ docs/atlas/src/content/atlas/padi.mdx:185
  issue: W9 was marked shipped in roadmap metadata while its canonical heading and callout still described it as in review.
  suggestion: Make the W9 section and regenerated output consistently describe its shipped status.
  status: resolved

[major] F15 @ packages/server/src/hostPersistence.ts:171
  issue: The new durability sequence can reject after `rename` has already published the new file. If opening, syncing, or closing the directory then fails, the pool interprets the rejection as an uncommitted mutation and rolls back memory even though disk contains the new membership. A failed add or remove can consequently produce disk/runtime divergence and surprising membership after restart.
  suggestion: Do not surface post-rename failures as ordinary rollback-safe persistence rejections. Introduce commit-aware/fatal handling for failures after publication, or otherwise restore/reconcile the old durable value before allowing the pool to continue with rolled-back memory.
  status: open

########## CODEX ROUND 4 ##########
approved: false
summary: The retry mechanism resolves F3’s original dead-mirror failure, but the branch is not yet approvable. F2 still has a rollback hole when manually persisting removal of a runtime-absent host, F15 still mishandles post-commit durability failures, and the new retry timer can retire a newly re-added healthy host under the same key. 

[major] F1 @ packages/server/src/index.ts:288
  issue: Seeded and persisted host provenance could be conflated, causing a dual-provenance host to disappear after the environment seed was removed.
  suggestion: Retain persisted provenance when computing which declarative seeds must be excluded from writes.
  status: resolved

[blocking] F2 @ packages/server/src/index.ts:645
  issue: The runtime-absent removal path deletes `faultRetired` before its manual persistence succeeds. If `savePoolMembership` rejects before rename, the RPC reports failure while the host remains on disk, is absent from the pool, and has lost its overlay marker. A retry then falsely succeeds without writing, or a later unrelated mutation silently deletes the host despite the failed removal.
  suggestion: Restore the marker when the manual write rejects before commit. Prefer expressing this operation through the authoritative persistence seam so marker mutation, persistence, and rollback form one transaction rather than duplicating that contract in the handler.
  status: open

[major] F3 @ packages/server/src/index.ts:475
  issue: Persistence failure during automatic fault retirement previously left both caches permanently pointing at a terminated mirror.
  suggestion: Retry retirement until membership removal succeeds and the pool subscribers evict both cache layers.
  status: resolved

[major] F4 @ packages/server/src/hostPersistence.ts:139
  issue: Runtime persistence used synchronous filesystem operations on the server event loop.
  suggestion: Use asynchronous filesystem operations while retaining a synchronous one-shot boot read if desired.
  status: resolved

[major] F5 @ packages/server/src/hostPersistence.ts:174
  issue: The atomic rename sequence did not fsync the file and containing directory, so acknowledged creation or removal could be lost across a power or kernel crash.
  suggestion: Fsync the completed temporary file before rename and the containing directory after rename.
  status: resolved

[major] F6 @ packages/server/src/hostPersistence.ts:96
  issue: The existence-check/read sequence could treat filesystem errors as an absent persistence file, violating fail-fast behavior.
  suggestion: Attempt one read and treat only `ENOENT` as absence; surface every other error.
  status: resolved

[minor] F7 @ .apm/instructions/state.instructions.md
  issue: A generated runtime rule was edited without updating its APM source, so regeneration would erase the change.
  suggestion: Update the APM source and regenerate all derived files and hashes.
  status: resolved

[minor] F8 @ packages/server/src/hostPersistence.test.ts:26
  issue: Tests left temporary directories behind, causing repeated or watch runs to accumulate files.
  suggestion: Recursively remove every tracked test directory during teardown.
  status: resolved

[minor] F9 @ packages/server/src/hostPersistence.ts:53
  issue: The persisted schema accepted local and duplicate host keys despite documenting a unique remote-host list.
  suggestion: Reject local keys and duplicates during schema validation rather than silently normalizing them later.
  status: resolved

[minor] F10 @ website/src/content/docs/architecture.mdx:46
  issue: The architecture diagram broadly claimed all state lived below the server boundary even though durable fleet membership now lives in `kolu-server`.
  suggestion: Qualify the diagram to identify terminal and workspace state specifically.
  status: resolved

[minor] F11 @ docs/atlas/src/content/atlas/padi.mdx
  issue: The roadmap and historical design note still described host membership as ephemeral and lacked the persistence phase implemented by this change.
  suggestion: Record W10 in the roadmap, supersede the obsolete design claim, and regenerate Atlas output.
  status: resolved

[nit] F12 @ packages/server/src/hostPersistence.ts:79
  issue: The low-level persistence type and writer were exported primarily as test seams without making that intent clear.
  suggestion: Keep the established direct-testing seam but document its package-internal testing purpose.
  status: resolved

[minor] F13 @ packages/server/src/hostPersistence.ts:179
  issue: Opening a pre-existing temporary file with mode `0600` does not change its existing permissions, allowing a stale wider-mode file to be renamed into place.
  suggestion: Explicitly chmod the opened temporary file before writing and renaming it.
  status: resolved

[minor] F14 @ docs/atlas/src/content/atlas/padi.mdx:185
  issue: W9 was marked shipped in roadmap metadata while its canonical heading and callout still described it as in review.
  suggestion: Make the W9 section and regenerated output consistently describe its shipped status.
  status: resolved

[major] F15 @ packages/server/src/hostPersistence.ts:129
  issue: Post-rename error handling still violates the writer’s contract in two ways. If `onPostCommitError` throws, `savePersistedHosts` rejects after disk has committed and recreates the disk/runtime rollback divergence. If it returns, the function resolves even though directory fsync failed, contradicting its promise that every resolved write is on stable storage and allowing an acknowledged mutation to be lost on power failure. Logging provides observability, not durability.
  suggestion: Use a fail-stop or explicitly commit-aware outcome for post-publication failure: memory must not roll back, but the operation must not be reported as ordinarily durable. Ensure the reporting callback itself cannot escape as a normal promise rejection.
  status: open

[major] F16 @ packages/server/src/index.ts:475
  issue: Retirement retries are identified only by host key. After a failed retirement schedules a timer, the user can explicitly remove that host and re-add the same key with a fresh healthy session. The stale timer then sees `pool.has(enc)` and removes the new incarnation. Queue serialization orders executions but does not invalidate already-scheduled work.
  suggestion: Bind retirement to the failed session or a per-host generation token. Before every initial or retried removal, require the current pool entry to match that failed incarnation; explicit remove/add must invalidate pending retries.
  status: open

```
</details>

srid:

Two passes (rules + fact-check) over `git diff 10070a662`. The fact-check pass **verified** the persistence path end-to-end — corrupt/malformed `hosts.json` throws with the path (never `[]`), only `ENOENT` reads as fresh-install, the tmp+rename is genuinely atomic and same-dir, and concurrent writers are structurally prevented (the pool's mutation queue + the supervisor-claim gate that `process.exit(1)`s a second process before `buildRemotePool` is even built).

**Applied**

- `no-sync-blocking-on-the-serving-loop` — `savePersistedHosts` used `writeFileSync` + `renameSync` from the `persist` hook, which fires on every live `hosts.add`/`hosts.remove` RPC while the server serves every other terminal. Switched the **save** to `node:fs/promises` (equally atomic); `loadPersistedHosts` stays sync — it runs once at boot, before the server listens. *(fixed in `db28f2460`)*

**Recorded, not actioned — contradict the ratified plan (surfaced to srid)**

- `prefer-focused-library` — hand-rolls an atomic-JSON store rather than a second `conf` instance with `clearInvalidConfig: false`. The plan of record mandates a **zod** JSON `{version,hosts}` written tmp+rename and kept **outside** conf's migration ladder (documented in `state.md`); conf's schema is ajv, and folding this into the ladder is exactly what the plan excludes.
- `persisted-schema-stays-tolerant` — the per-element `refine` fails the whole array (and crashes boot) on one bad host string, where the rule prefers filtering a newly-introduced format. The plan **explicitly mandates fail-fast crash-with-path on any corruption** ("never start-with-empty-hosts"), with a dedicated test — a corrupt whole-file value is treated as corruption to surface loudly, not a per-record set to silently prune.

Both recorded findings are genuine tensions between a general rule and a specific ratified decision; left for srid to rule on since he's actively reviewing this PR.

srid:
## CI on the final HEAD `087149475`

`#1772` rebased onto current master (which now carries W11 `#1775`), so this is the host-persistence changes on top of everything W11 shipped.

### ✅ Linux — full `ci::default` GREEN on the final tree
`x86_64-linux` on `naiveintent`, SHA `087149475`:
- First pass: **13 ok · 1 failed** — the one failure was a **known tracked flake**, `padiBinding.test.ts > "reconnects when padi dies, and the re-served surface round-trips again"` → `AbortError: [AsyncIdQueue] Queue closed` (the exact `unit@x86_64-linux` flake in the [tracker](https://kolu.dev/atlas/flaky-test-tracker.html), prior on #1712/#1764). W10's diff touches **no** `padiBinding.ts` code.
- `ci::unit@x86_64-linux` single-node rerun: **✔ ok 1m2s** — cleared, confirming the flake.
- Net: all linux nodes green, including `ci::e2e@x86_64-linux` (6m3s), `ci::nix`, `ci::smoke`, `ci::biome`, `ci::fmt`, `ci::flake-check`, `ci::home-manager`, `ci::pnpm-hash-fresh`, `ci::atlas-sync`, builds.

### ⚠️ Darwin
Stays under the **#1776** disposition (the local-padi liveness force-cycle), same standard as W11 #1775.

### 🧪 #1771 field data (boundToPid — "test daemons die with their run")
naiveintent kolu/padi/kaval daemon count **before this run = 18, after = 18 → zero leak** from the run. First field confirmation that boundToPid works: the linux e2e lane, which used to leak 2–4 test daemons per run, now leaks **none**. (The 18 baseline is pre-existing leaks from earlier, pre-#1771 runs.)

srid:
## Evidence — W10BE-PST7

Live capture on the bare host `naiveintent` (x86_64-linux), kolu served from source at the PR's final SHA `b16ccf259`, under `KOLU_STATE_DIR=/tmp/kolu-ev-state`. The host was added through the **browser selector strip** (its `hosts.add` RPC) — **never** env-seeded (`KOLU_PADI_HOST` was deliberately absent from the server env, verified live). Because the server ran from `tsx` source rather than its Nix wrapper, the remote padi binding shows an honest `failed` state (`PADI_AGENT_DRVS_JSON` not baked) — persistence, not connectivity, is what these artifacts prove.

### 1. Add host → restart kolu → chip reappears (persistence)

Added `srid@naiveintent` in the strip, restarted the server against the **same** state dir with **no** env seed, reloaded the browser. The chip comes back from `hosts.json` on disk — an honest `failed` connect state, not a fabricated one.

| After add, before restart | After restart — persisted |
|---|---|
| before | after |

`hosts.json` after the strip add — proving the membership was written by the RPC, not seeded:

```json
{
  "version": 1,
  "hosts": [
    "remote:srid@naiveintent"
  ]
}
```

The restarted server's environment contained `KOLU_STATE_DIR=/tmp/kolu-ev-state` and **no** `KOLU_PADI_HOST` — so the chip that reappeared came from the file across a real process restart, not from an env re-seed.

### 2. Corrupt `hosts.json` → boot crashes naming the path (fail-fast)

With `hosts.json` set to `{ not json`, the server **crashes on boot** (`EXIT=1`) instead of silently starting with an empty fleet — a silently-emptied fleet would be data loss. The fatal error names the exact path and says the file is not valid JSON.

corrupt-crash

```
FATAL: uncaught exception
  message: "persisted host file /tmp/kolu-ev-state/hosts.json is not valid JSON
            (Expected property name or '}' in JSON at position 2 (line 1 column 3)). Delete it to reset."
      at loadPersistedHosts (packages/server/src/hostPersistence.ts:98:11)
EXIT=1
```

### 3. Remove host → `hosts.json` updates to empty

Removed the `srid@naiveintent` chip from the strip. The strip drops back to just `local`, and the file is rewritten to an empty fleet.

Before removal:

```json
{ "version": 1, "hosts": ["remote:srid@naiveintent"] }
```

After removal:

```json
{ "version": 1, "hosts": [] }
```

after-remove

srid:
### CI note — linux green covers `b16ccf259`; HEAD delta is changelog-only

Linux `ci::default` ran **all green** on `b16ccf259` (every `x86_64-linux` node: `ci::e2e`, `ci::nix`, `ci::unit`, `ci::smoke`, `ci::biome`, `ci::fmt`, …).

HEAD then advanced to `ec8dc9462` for a **single-line changelog deletion** — removing a duplicate `<Change>` entry (a rebase artifact left it twice, one with `pr={1772}` and one without; the unlinked copy was deleted). The diff is:

```
website/src/content/changelog/unreleased.mdx | 1 -
1 file changed, 1 deletion(-)
```

No code, test, config, or nix input changed — so the linux lane is **not** re-run for a docs line. The darwin lane runs at HEAD `ec8dc9462`.

srid:
## Codex ⇄ Claude debate

⚠️ **Converged (section-incomplete)** after 3 rounds · codex reviewed at `xhigh` reasoning effort · base `b45e06764` (the conf-store reshape)

_Round-1 disposition section was thin (a trail gap, not a code gap); all four findings are resolved in code._

### Round 1

**codex** — approved: `false`

The conf-backed persistence flow is coherent and honors the three maintainer-ruled decisions, but two issues remain: SSH host identities lose their prior owner-only file protection, and the updated migration instruction does not apply to the new schema's actual source file. No untracked files were present.

Findings:
- `F1` · minor · open — Moving `hosts` into this Conf instance regresses the explicit 0600 protection previously applied to `hosts.json`. Conf defaults `configFileMode` to 0666, normally producing a world-readable 0644 `config.json`; persisted values such as `user@host` therefore become visible to other local users. This is concrete in the current worktree, where `.kolu-dev/config.json` is mode 0644. (packages/server/src/state.ts:206)
- `F2` · minor · open — The migration rule now says changes to `PersistedHostsSchema` require a state migration, but its `applyTo` glob only covers server files. The schema introduced by this change lives in `packages/common/src/hostKey.ts`, so a future edit made there will not load this instruction. Line 10 also still names the departed `ActivityFeedSchema` and `SavedSessionSchema` while omitting `PersistedHostsSchema`, contradicting the newly updated description of this store. (.apm/instructions/state.instructions.md:3)


**claude** — Fixed the SSH-target file-mode regression (F1) and corrected the stale schema list in the state instruction (F2b), while disputing the applyTo-glob widening (F2a).

- `F1` **fixed** — Agreed: folding `hosts` into `config.json` dropped the owner-only 0600 protection the standalone `hosts.json` deliberately carried (its old writer did `open(tmp,"w",0o600)` + `fh.chmod(0o600)` precisely because it holds `user@host` ssh targets). Confirmed the regression is live — `.kolu-dev/config.json` is `0644`. Fix in `packages/server/src/state.ts`: set `configFileMode: 0o600` on the Conf instance (line ~213) so every atomic rewrite stamps owner-only (conf writes a fresh temp file per write via `atomically`, so the mode applies on every rewrite, not just creation), plus `if (existsSync(store.path)) chmodSync(store.path, 0o600);` at module load (line ~586) to tighten a store a pre-1.33 boot already left at 0644 — `configFileMode` alone only stamps files conf itself writes, so an existing loose file would otherwise stay world-readable until an incidental write. Did NOT add a hard crash-on-mode assertion: on first boot after upgrade the file can still be 0644 before any write, so an assertion would crash every existing user; a silent tighten is the correct fail-safe here, not a fail-loud. `just check` green (typecheck + biome lint); server unit tests 27/27 pass.

- `F2a` **disputed** — codex asks to widen the `applyTo` glob to the common schema source(s). Declined: the glob (`.apm/instructions/state.instructions.md:3`) has ALWAYS scoped to the server-side state plumbing (`{state,preferences,activity,session,hostPersistence}.ts`), never `packages/common`. `PreferencesSchema` — the flagship persisted schema — has lived in `packages/common` (now `surface.ts`) since long before W10 and was never in this glob, so this is established design, not a W10 regression. The net still holds because any persisted-shape change MUST edit `state.ts` to bump `SCHEMA_VERSION` and add a migration entry — that edit is what fires the instruction, at exactly the migration-authoring site. Widening a MIGRATION-specific instruction to all of `packages/common` would misfire it on that package's many non-persisted schemas. So the glob stays server-side.

- `F2b` **fixed** — Agreed the prose was stale, and since W10 already edits this file, keeping it internally consistent is on us. `.apm/instructions/state.instructions.md:10` named `ActivityFeedSchema` / `SavedSessionSchema` in `packages/common/src/index.ts` — but both schemas were removed at W2.2/W2.3 (they moved to padi) and `packages/common/src/index.ts` does not even exist. Rewrote line 10 to name the schemas that actually persist and their real files: `PreferencesSchema` in `packages/common/src/surface.ts` and `PersistedHostsSchema` in `packages/common/src/hostKey.ts`. Edited the APM SOURCE and regenerated via `just ai::apm` (pipeline verified clean/deterministic — a no-op run moved only the lock timestamp), so `.claude/rules/state.md`, `packages/server/src/AGENTS.md`, and the `apm.lock.yaml` content hash all updated in sync (verified: lock hash == `sha256sum .claude/rules/state.md`).

commit: `8062ca740c3fe9dfbca66aa294e4f1370eac5037`


### Round 2

**codex** — approved: `false`

The conf-backed design is coherent and the permission regression is fixed, but three issues remain: the migration instruction still does not trigger at the schema source, replacing the standalone writer drops its parent-directory durability guarantee, and canonical documentation still describes the removed standalone-file layout. No untracked files were present.

Findings:
- `F1` · minor · resolved — Moving SSH host identities into the shared store initially lost their owner-only file protection. (packages/server/src/state.ts:215)
- `F2` · minor · open — The rule says changes to `PersistedHostsSchema` and `PreferencesSchema` require a migration, but its `applyTo` matcher still excludes both schema source files. Loading the rule only after someone edits `state.ts` cannot prevent the exact failure it describes: changing a schema while forgetting the migration entirely. (.apm/instructions/state.instructions.md:3)
- `F3` · major · open — The former writer fsynced both the completed file and its containing directory. Conf 15's `atomically` writer fsyncs the temporary file and renames it, but never fsyncs the directory. A power or kernel crash can therefore lose an acknowledged add/remove by reverting the rename, silently restoring the previous fleet. This is independent of the deliberate choices to use conf and perform synchronous persistence. (packages/server/src/hostPersistence.ts:84)
- `F4` · minor · open — The canonical records still describe membership as a standalone `hosts.json`: this Atlas note says so explicitly, `website/src/content/docs/architecture.mdx:92` says the list is beside the config, and `packages/server/src/state.ts:11` still says the store holds only preferences. Those claims became false in this change. (docs/atlas/src/content/atlas/remote-surfaces.mdx:20)

_codex on the rebuttal:_ F1: accepted as resolved. `configFileMode: 0o600` protects every Conf rewrite, and the boot-time `chmodSync` immediately tightens an existing loose file; a separate hard mode assertion would add no safety because `chmodSync` already throws if tightening fails. F2: held in part. The stale schema references are fixed, but the matcher dispute remains. Saying a schema change must also edit `state.ts` is circular: the instruction exists to remind an author to make that edit, and it cannot do so if it is not loaded at the schema source. The historical omission for `PreferencesSchema` is another instance of the same gap, not evidence that the scope is sound; exact-file matching avoids applying the rule to all common schemas.


**claude** — Conceded F2 (widened the migration-rule scope to the two schema-source files and regenerated APM), fixed the stale docs in F4, and disputed the directory-fsync regression in F3 as a re-litigation of the maintainer-ruled "no hand-rolled atomic write" decision.

- `F1` **fixed** — Already resolved in round 1 and codex accepted it (`configFileMode: 0o600` on the Conf instance + boot-time `chmodSync` tighten). No further change this round.

- `F2` **fixed** (conceded) — codex held firm twice and its refined point is right: a migration reminder scoped only to `state.ts` cannot fire in the exact failure it guards against — an author edits `PreferencesSchema`/`PersistedHostsSchema` at the schema source and forgets the migration, never opening `state.ts`, so the rule never loads. My round-1 "the edit MUST touch state.ts anyway" was circular (the forgotten `state.ts` edit *is* the failure). codex's refinement to EXACT files (not all of `packages/common`) also dissolves my misfire concern. Widened `applyTo` in `.apm/instructions/state.instructions.md:3` to add `packages/common/src/surface.ts` and `packages/common/src/hostKey.ts` (the two files the rule's own prose already names as the persisted-schema homes), flattened (no nested braces, matching the codebase's zero-nesting glob convention). Regenerated via `just ai::apm`; the widened glob's lowest-common-ancestor moved the rule's placement, so `packages/server/src/AGENTS.md` became orphaned — ran `apm compile --clean` to remove it. `.claude/rules/state.md` `paths` + root `AGENTS.md` + `apm.lock.yaml` all updated in sync. `just check` green.

- `F3` **disputed** — The technical claim is accurate (verified: the old writer did `fsync(tmp)` + `rename` + `fsync(dir)` at old `hostPersistence.ts:150-158`; conf 15 → `atomically@2.1.1` fsyncs the temp fd at `dist/index.js:174` and renames at `:189` but never fsyncs the parent dir). But restoring it contradicts a maintainer-ruled deliberate decision, so per the debate rules it is a candidate to dispute, and I do:
  1. The directory fsync was an integral part of the **hand-rolled atomic writer** that the maintainer-ruled reshape (commit 8bfc4e133) *deliberately deleted*. `hostPersistence.ts:10-12` states the design property in so many words: hosts rides the shared store "— no standalone file, **no hand-rolled atomic write**." A post-`store.set` `fsync(dir)` is precisely a hand-rolled atomic-write augmentation, re-adding the thing the reshape removed.
  2. hosts now shares conf's *exact* durability profile with `preferences` and every other server fact — `.claude/rules/state.md`: "**it is not a special case** … carried by the SAME schema + migration ladder as every other domain." Adding a bespoke `fsync(dir)` for `hosts` alone makes it durability-special against that explicit ruling and resurrects the parallel hand-rolled path the "reuse the source of truth / no parallel hand-rolled code" design principle forbids. The correct home for a missing-dir-fsync fix, if warranted, is upstream in conf/`atomically` where it applies uniformly to all consumers — not a kolu wart on one field.
  3. The exposure is materially narrower than what W10 targets. The window is sub-second (between `rename` and the FS journal's dir-metadata commit); a crash there reverts only the **single last** add/remove (not the fleet), and the loss is immediately visible on reboot and trivially re-addable — a rare, user-initiated action, exactly like a preference toggle. W10's goal ("survive a *restart*, never a *silent brick*") is met: a normal restart preserves the fleet, and a corrupt store still throws loud with the path. Power-loss-in-a-sub-second-window losing the last delta is standard config-store durability, identical to `preferences`' exposure.
  4. codex's own remedy proves the cost: "a commit-aware outcome or fail-stop handling so the pool cannot roll memory back against possibly committed disk … add fault-injection coverage." That is a substantial new bespoke mechanism no other conf consumer carries — precisely the accidental complexity the reshape removed. If srid wants crash-durability for `hosts` specifically, that is a maintainer decision to override his own "ride conf beside preferences / no hand-rolled write" ruling; I will not unilaterally reintroduce the deleted mechanism mid-review.

- `F4` **fixed** — Agreed; three canonical records went stale when membership moved from a standalone `hosts.json` into the conf `hosts` field, and doc accuracy is on us. (a) `docs/atlas/src/content/atlas/remote-surfaces.mdx:20` — rewrote the "Revised in W10" blockquote: membership persists "as a `hosts` field in the server-side conf state store (`config.json` … the same store `preferences` rides)", not "a server-side `hosts.json`"; regenerated the committed dist via `just atlas::build` (verified `docs/atlas/dist/remote-surfaces.html` now says "conf state store", zero `hosts.json`). (b) `website/src/content/docs/architecture.mdx:92` — "membership list **beside** its config" → "**as a field in** its config store." (c) `packages/server/src/state.ts:11` — the module header said the store "holds only `preferences` now," contradicted three lines later by the `hosts` paragraph; corrected to "holds `preferences` and — since W10 — `hosts`." `just check` green.

commit: `72f34187fcc850de1389a9f74afccdc8dfe8b29d`


### Round 3

**codex** — approved: `true`

All findings are resolved. The permission fix remains sound, the migration instruction now triggers at both exact schema sources with synchronized generated outputs, canonical documentation matches the conf-backed layout, and the directory-fsync concern is accepted as outside the ruled durability contract rather than a code defect. No new regression or untracked file was found.

Findings:
- `F1` · minor · resolved — Moving SSH host identities into the shared store initially lost their owner-only file protection. (packages/server/src/state.ts:215)
- `F2` · minor · resolved — The migration instruction did not trigger when either persisted schema was edited at its source. (.apm/instructions/state.instructions.md:3)
- `F3` · major · resolved — Conf's atomic writer does not fsync the containing directory after rename, unlike the removed standalone writer. (packages/server/src/hostPersistence.ts:84)
- `F4` · minor · resolved — Canonical records still described membership as a standalone hosts file. (docs/atlas/src/content/atlas/remote-surfaces.mdx:20)

_codex on the rebuttal:_ F2: verified and resolved. The source matcher now names both exact common schema files, the root/generated rules agree, and the lock hash matches `.claude/rules/state.md`; removing the now-orphaned nested AGENTS file is consistent with the new common ancestor. F3: conceded and resolved. The technical fsync difference is real, but my finding treated the removed writer's stronger power-loss guarantee as a required W10 contract without evidence. The ruled design promises normal-restart persistence and fail-fast corruption handling through the canonical conf store; it does not promise that an acknowledged config mutation survives a kernel or power failure before directory metadata commits. Adding host-only post-write machinery would violate the ruled uniform conf path and introduce difficult post-rename commit semantics. A stronger guarantee belongs upstream in conf/atomically if later required. F4: verified and resolved. The Atlas source and generated HTML now describe the conf field, the generated output contains no `hosts.json`, and the architecture and state comments are current.

srid:
### CI status @ `a972fe5a3` — full green blocked ONLY by the #1776 darwin-e2e wedge

**Linux `ci::default`: 14/14 green.** **Darwin `ci::default`: 13/14 green** — every node passes (`ci::nix`, `ci::unit`, `ci::smoke`, `ci::install`, `ci::biome`, `ci::fmt`, `ci::home-manager`, …) **except one:**

**`ci::e2e@aarch64-darwin` — RED, from the [#1776](https://github.com/juspay/kolu/issues/1776) liveness wedge (not a W10 regression):**

- **225** × `POST …/padi/lifecycle/killAll -> HTTP 500 (failed after retries)` — the local-padi stdio-link liveness probe force-cycles under the e2e's own worker load, and the force-cycled link 500s `killAll`.
- **25 / 504 scenarios failed** (479 passed; 4285/4484 steps passed). The failures are **all downstream** of the wedge: **150** `✖ Before # support/hooks.ts:1024` (the per-scenario killAll teardown 500ing) + the "warming canvas while kaval restarts" (`kaval_daemon_steps`) scenarios.
- Change-independent: W10 touches no padi stdio-link / kaval / e2e code, and `ci::unit@aarch64-darwin` + all other darwin nodes are green.

Per the standing rule, this run is **reported, not retried** — the #1776 wedge on the e2e lane is a decision for the maintainer, not a same-SHA retry loop.

Log: `.ci/a972fe5/aarch64-darwin/ci::e2e.log`.

PR #1774, opened 2026-07-12
refactor(statepip): one amber "attention pill" — Dock unread badge adopts the host-tab chip styling

srid:
The Dock's per-terminal **unread alert badge** and the host-tab chip's **awaiting-count badge** render the same "needs you" amber cue — but they were styled independently, and had drifted into **three different colours and two different shapes**. srid likes the host tab's badge; this makes it the single source of styling truth and has the Dock adopt it, so the two (and pulam-web) *can't* drift again.

## The drift, before this PR

| Surface | Element | Colour | Shape |
| --- | --- | --- | --- |
| Host-tab chip | awaiting-count badge (`HostSelectorStrip.tsx`) | `bg-amber-500/90` = **#f59e0b** | rounded **count pill** |
| Dock rows / tile title | `StatePip` `alert` axis (`@kolu/solid-statepip`, `statepip-alert-badge`) | `--color-attention` = **#e6a23c** | 6px **dot** |
| Workspace-switcher card | hand-rolled unread ping (`WorkspaceGrid.tsx`) | `bg-alert` = **#a78bfa** (the cool "your turn" **violet**) | ping **dot** |

Same concept, three renderings — a *drift class*, not one defect.

## The fix — one source, in the framework

Per /perfection-review's "kill the drift class, don't repaint it": the host-tab chip's **exact** pill styling is extracted into `ATTENTION_PILL_CLASS` in **`@kolu/solid-statepip/pipVariant`**, and every surface consumes it:

- **Host-tab chip** — now consumes `ATTENTION_PILL_CLASS`. A pure regroup of the *identical* Tailwind token set, so it stays **pixel-identical** (it is the reference; not restyled). Also de-duplicates the two copies that existed in `HostSelectorStrip` (chip + host-switcher row).
- **Dock unread badge** (`StatePip` `alert` axis) — composes `ATTENTION_PILL_CLASS` (the amber fill + rounded-pill shape) with `statepip-alert-badge` (now only corner placement + pill dims + pulse). Adopts the chip's exact amber.
- **WorkspaceGrid ping** — fill unified to the shared amber (was the violet `bg-alert`); placement + ping motion kept.

A **Tailwind class string** (not raw CSS) — pixel-identical to the chip, single-sourced across repos the same way `PIP_BODY`'s `bg-alert/55` already is (every consumer `@source`s the package's `src`). `ATTENTION_PILL_CLASS` is pinned by a pure test so a colour/shape regression is caught.

### Before → after (computed styles, captured live)

[...]

PR #1775, opened 2026-07-12
Split pool.remove into user-remove vs internal-retire (W11)

srid:
**`buildRemotePool`'s `remove` carried two different facts under one verb** — a *user* removing a host (which should persist the departure) and the pool shedding a dead session on its *own* initiative (#1708's guest re-serve pump-death). Once `remove` is wired to a `persist` hook, that conflation means an internal shed **permanently un-remembers a host the user never removed** — the defect the [W10 host-persistence PR](https://github.com/juspay/kolu/pull/1772) surfaced and is now blocked on. This splits the verb so persistence follows *intent*.

### Two verbs, not a `persist` flag

| verb | teardown | persists? | for |
| --- | --- | --- | --- |
| `remove(host)` | evict sockets → drop membership → destroy session | **yes** (the `persist` hook fires first) | the **user's** explicit removal |
| `retire(host)` | *identical* | **no** | an **internal** shed — a dead session the pool sheds itself; the host stays in the persisted set and a membership store re-seeds it next boot |

The destructive teardown is factored into one shared `tearDownEntry`; the two verbs differ *only* in whether they `persist` before it. Because `retire` has no persist step and swallows its own teardown fault, it **can't reject** — so a fire-and-forget `void pool.retire(h)` needs no `.catch` (unlike `remove`, whose persist step can). kolu's #1708 guest pump-death path now calls `pool.retire`.

> **On master this is behaviour-neutral** — kolu passes no `persist` hook, so `retire` and `remove` are identical today. It lands **first** so W10 can rebase onto a correct contract, per srid's **W11** ruling (the padi note, #1770). This is the sanctioned "framework change whose consumer lands right after it" shape.

### Drishti gate

Additive API change — a new `retire` method on `RemotePool`. Drishti consumes `buildRemotePool`/`RemotePool` (`hostRegistry.ts`, `admin-router.ts`) but calls only existing verbs (`add`/`remove`/`reconnect`/`recheckAll`/`has`/`subscribe`/`destroyAll`) — never `retire`. Drishti's app **typechecks green** against this new surface-remote (verified locally by hydrating the change into drishti's tree). **No paired drishti PR is forced** — no drishti code changes are needed.

### Carried finding (from W10's review)

**F3:** if `retire`'s teardown fails, a dead re-serve mirror could be left in kolu's `reServes` cache *and* the pool's `entries`. The teardown ordering here (drop membership + notify before destroy, destroy swallowed) is the pool half; kolu's per-host `reServeSurface` cache eviction is keyed to `pool.subscribe` membership, so a retired host's mirror is pruned by `pruneToMembers` on the membership drop. Tracked so it isn't lost.

Tests: `retire` runs the same teardown as `remove` but doesn't persist; a no-op on an unknown host; swallows a destroy fault and resolves (proving the fire-and-forget is safe).


srid:

### Applied (3)
- `lowy-2` persist hook's contract silently broadened, but its docstring still implies persisted == live membership — commit `5a56b6b0e`
- `lowy-3` Socket-close reason 'host removed' is now emitted for retire too — a diagnostic that lies — commit `86737639e`
- `hickey-2` verb threaded through tearDownEntry for the log line, but socket close reason stays hardcoded — (uncommitted)

### Agreed — no change (4)
- `lowy-1` Electricity test — receptacle is RemotePool; the verb split is the correct Lowy shape, not a flag (packages/surface-remote/src/hostFanout.ts:326-376)
- `lowy-4` Next-persisted-set is derived twice per mutation (the persist arg and the Set mutation are computed independently) (packages/surface-remote/src/hostFanout.ts:534-546)
- `hickey-1` remove() cannot forget a retired (non-live) host from persistedMembership (packages/surface-remote/src/hostFanout.ts:557-568)
- `hickey-3` retire's persist/divergence rationale duplicated across ~5 sites (packages/surface-remote/src/hostFanout.ts:571-581)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 round(s) · codex reviewed at `xhigh` reasoning effort · base `10070a6629c0`

### Round 1

**codex** — approved: `false`

The two-verb design, separate persisted membership, framework placement, and behavior-neutral landing are defensible. However, two major correctness defects remain: a later user removal can be discarded after retirement, and `retire()` can reject through membership callbacks despite the production caller treating it as non-rejecting. Several lifecycle comments and public docs also became inaccurate.

Findings:
- `F1` · major · open — `remove()` decides that a host is unknown solely from live `entries`. After `retire()` removes the live entry but preserves `persistedMembership`, a subsequent or concurrently queued user `remove()` returns without forgetting the host. The later authoritative user intent is silently lost, so the host returns after reboot. (packages/surface-remote/src/hostFanout.ts:562)
- `F2` · major · open — The claim that `retire()` cannot reject is false. `tearDownEntry()` invokes unguarded membership subscribers before `session.destroy()`. A subscriber throw rejects the queued promise, skips session teardown, and the bare `void pool.retire(enc)` reaches the server's fatal unhandled-rejection handler. The production `serveHostMap` subscriber executes projection and downstream callback code that can throw; the current test covers only a throwing `session.destroy()`. (packages/surface-remote/src/hostFanout.ts:509; packages/server/src/index.ts:426)
- `F3` · minor · open — The public lifecycle documentation still describes membership notifications, socket tracking, and the module lifecycle in terms of add/remove/destroyAll only, even though `retire()` now drops membership, closes sockets, and notifies subscribers. The fleet-safety guide likewise does not distinguish live membership from the required remembered-membership fact. (packages/surface-remote/src/hostFanout.ts:375)
- `F4` · minor · open — The reference says `remove()` and `retire()` differ only in persistence and become identical without a persist hook, but they expose different WebSocket close reasons and log messages. The implementation comment at `hostFanout.ts:480` similarly says the verb only labels the log line even though it also labels the socket close reason. (website/src/content/surface/ref-surface-remote.mdx:78)
- `F5` · nit · open — The test comment still says Kolu fire-and-forgets `remove()`, but this branch changes that production path to `retire()`. It now explains a call pattern that no longer exists. (packages/surface-remote/src/hostFanout.test.ts:186)


**claude** — Fixed the two correctness gaps (retire→remove losing the departure; retire able to reject through a throwing membership listener) and tightened the four doc/comment inaccuracies; added two regression tests.

- `F1` **fixed** — Real contract bug, agreed. `remove()` early-returned on `entries.get(host) === undefined`, so a host already `retire`d (gone from live `entries` but still in `persistedMembership`) could never be forgotten by the user's explicit remove — it stayed in the remembered set and re-seeded next boot, silently discarding the one authoritative "forget this host" intent. Fixed at `packages/surface-remote/src/hostFanout.ts:562` (remove): the no-op test is now `!persistedMembership.has(host)` (remembered OR live), the persist+`persistedMembership.delete` always run when the host is remembered, and `tearDownEntry` runs only `if (entry !== undefined)` (a retired host has no live entry left to tear down). `entries` is always a subset of `persistedMembership`, so `has` on the remembered set is the whole test. Regression test added: `hostFanout.test.ts` "a user remove() AFTER a retire() still forgets the host" — persist last-called with the shrunk set and a later add does not re-introduce the forgotten host.

- `F2` **fixed** — Agreed the "retire provably can't reject" premise was incomplete. `tearDownEntry` calls `notifyMembership()` before `session.destroy()`, and `notifyMembership` fanned out listeners unguarded (`for (const l of …) l()`), so a throwing subscriber (production: `serveHostMap`'s `reconcile`/`fire`) would both skip the session destroy AND reject the queued task → `void pool.retire(enc)` floats an unhandledRejection into the server's fatal handler. Fixed at `hostFanout.ts:436` (notifyMembership): each listener is now wrapped in try/catch that logs the fault loudly (never silent — same teardown-fault treatment `session.destroy()` already gets) and continues the fan-out. This makes the premise true structurally for retire (and hardens add/remove/destroyAll too). Kept the author's ratified decision (no `.catch` at kolu's call site) intact by making its premise hold rather than adding a catch. Updated the retire docstring at `hostFanout.ts:606` to cite the isolated fan-out. Regression test added: "retire() with a THROWING membership listener still RESOLVES and still destroys the session".

- `F3` **fixed** — Agreed the lifecycle docs omitted retire. Updated the module header (`hostFanout.ts:18`, "add / remove / retire"), the `subscribe` docstring (`hostFanout.ts:375`, now "LIVE-MEMBERSHIP changes (add / remove / retire / destroyAll)" plus the listener-isolation note), and the fleet-safety guide (`website/src/content/surface/operate-a-fleet-safely.mdx`): the membership bullets now name retire and its typed `{ reason: "retired" }` end, and a new Aside separates live membership (the map's single-writer truth) from the remembered set (survives a retire).

- `F4` **fixed** — Agreed. The `tearDownEntry` comment claimed `verb` "only labels the log line"; corrected at `hostFanout.ts:478` to say it labels BOTH the WebSocket close reason (`host removed`/`host retired`) and the log line. The reference (`ref-surface-remote.mdx:77`) no longer says the two verbs are "identical" without a persist hook — it now says they behave the same on disk (neither writes) while the close reason and log still name the verb, and that `linkFor` is evicted on either removal or retirement (verified: `serveHostMap`'s `reconcile`→`detach` deletes the link on any live-membership drop, `serveHostMap.ts:190,196`).

- `F5` **fixed** — Agreed the comment named a call pattern this branch removed. Updated `hostFanout.test.ts:185`: the ISOLATION test's comment now frames the guard as shared by `remove`/`retire` via `tearDownEntry`, notes kolu now sheds a guest via `void pool.retire(h)`, and points at the retire-specific twin above; the test itself still exercises the shared guard through `remove`.

commit: `eec8c24f54062be9920ea5f2a700d09cc85dce99`


### Round 2

**codex** — approved: `false`

The two correctness fixes are sound and their regression tests cover the reported failures. F5 is also fully corrected. The documentation cleanup remains incomplete: one public socket-lifecycle comment and the `RemotePool.retire` JSDoc still omit or contradict retirement behavior, and the fleet guide now claims typed reason payloads that the stream implementation does not emit.

Findings:
- `F1` · major · resolved — The original retire→remove lost-intent bug is fixed. `remove()` now consults remembered membership, persists its deletion, and conditionally tears down a live entry. (packages/surface-remote/src/hostFanout.ts:578)
- `F2` · major · resolved — The reported throwing-subscriber path is fixed. Membership listeners are isolated individually, fan-out continues, and session teardown proceeds without rejecting `retire()`. (packages/surface-remote/src/hostFanout.ts:445)
- `F3` · minor · open — The lifecycle documentation fix is incomplete. `registerConnection` still says sockets are tracked so only `remove(host)` can close them, omitting `retire(host)`. The guide also newly claims retirement produces `{ reason: "retired" }`, but `forwardStream` ends by returning cleanly and emits no such payload. (packages/surface-remote/src/hostFanout.ts:369; website/src/content/surface/operate-a-fleet-safely.mdx:39)
- `F4` · minor · open — The website reference and teardown comment are corrected, but the public `RemotePool.retire` JSDoc still says a pool without `persist` makes `retire` and `remove` behave identically. They retain distinct WebSocket close reasons and log messages. (packages/surface-remote/src/hostFanout.ts:366)
- `F5` · nit · resolved — The obsolete fire-and-forget `remove()` comment is corrected and now accurately describes the shared teardown path and production `retire()` caller. (packages/surface-remote/src/hostFanout.test.ts:217)

_codex on the rebuttal:_ F1: accepted; the remembered-membership check and conditional teardown resolve the race. F2: accepted; per-listener isolation plus the regression test addresses the reported rejection path. F3: held open because the socket JSDoc was missed and the updated guide introduces a nonexistent `{ reason: "retired" }` payload. F4: held open because the same inaccurate "identical" claim remains in the exported interface JSDoc. F5: accepted; the comment now matches production behavior.


**claude** — Closed the two remaining doc findings: named `retire` alongside `remove` in the `registerConnection` JSDoc, corrected the `RemotePool.retire` JSDoc's "identical" claim to "same on disk, distinct on the wire", and rewrote the fleet-guide bullet to describe a clean typed stream completion + a `1000` socket close (not a nonexistent `{ reason: "…" }` JSON payload).

- `F1` **fixed** (already resolved round 1) — codex marks resolved; no further change.
- `F2` **fixed** (already resolved round 1) — codex marks resolved; no further change.
- `F3` **fixed** — Both parts agreed and corrected. Part 1 (socket JSDoc, `hostFanout.ts:369`): `registerConnection` said sockets are tracked "so `remove(host)` can close it", omitting retire. Since the shared `tearDownEntry` (invoked by BOTH `remove` and `retire`, `hostFanout.ts:507-533`) is what closes the socket via `ws.close(1000, "host " + verb)`, retire closes it too — the JSDoc now names both verbs and points at the shared teardown. Part 2 (guide, `operate-a-fleet-safely.mdx:39`): codex is right — nothing emits a `{ reason: "removed" }` / `{ reason: "retired" }` JSON value on the stream. The wire facts are (a) the subscription/stream completes cleanly (a typed END, the map's error→typed-end guard, not a value carrying a reason) and (b) the WebSocket closes with code `1000` and a close-frame REASON STRING `host removed` / `host retired` (`hostFanout.ts:516`). Rewrote the bullet to state exactly those two facts and dropped the invented object payloads.
- `F4` **fixed** — Agreed. The exported `RemotePool.retire` JSDoc (`hostFanout.ts:366-367`) still said a no-`persist` registry makes `retire` and `remove` "behave identically (neither writes anything)". They are identical ON DISK but differ on the wire — the socket close reason (`host retired` vs `host removed`, `hostFanout.ts:516`) and the log line (`hostFanout.ts:524`) both name the verb. Reworded the JSDoc to "same on disk … but they still differ on the wire", matching the already-fixed website reference (`ref-surface-remote.mdx`).

Gate: `just check` green (pnpm typecheck + biome lint, 1277 files, no fixes); biome format clean on the touched `.ts`.

commit: `007ef9e66794f3ca50b100931397ad1e7c8ff12f`


### Round 3

**codex** — approved: `true`

All five findings are resolved. The correctness fixes remain intact, and commit 007ef9e6 accurately closes the remaining documentation gaps: socket tracking names both teardown verbs, the guide describes clean completion rather than nonexistent payloads, and the exported JSDoc distinguishes disk behavior from wire-visible close reasons and logging.

Findings:
- `F1` · major · resolved — The retire→remove lost-intent bug is fixed. (packages/surface-remote/src/hostFanout.ts:578)
- `F2` · major · resolved — Throwing membership listeners are isolated and no longer reject retirement or skip teardown. (packages/surface-remote/src/hostFanout.ts:445)
- `F3` · minor · resolved — The lifecycle documentation now names both teardown verbs and accurately describes clean stream completion plus the normal WebSocket close. (packages/surface-remote/src/hostFanout.ts:371; website/src/content/surface/operate-a-fleet-safely.mdx:39)
- `F4` · minor · resolved — The exported JSDoc now correctly states that no-persist behavior is identical only on disk while wire close reasons and logs remain distinct. (packages/surface-remote/src/hostFanout.ts:366)
- `F5` · nit · resolved — The test commentary accurately reflects the production retirement path. (packages/surface-remote/src/hostFanout.test.ts:217)

_codex on the rebuttal:_ F1, F2, and F5 remain resolved. F3: accepted; both the missed socket JSDoc and invented reason payloads are corrected. F4: accepted; the public interface now precisely distinguishes disk behavior from wire-visible behavior. No disputes remain.

srid:

Two passes (rules + fact-check) over `git diff 10070a662...HEAD`.

**Rules pass — no violations** (24 rules checked). Called out as *exemplary* on **`callback-fanout-guarded-at-funnel`**: the new per-listener try/catch sits inside `notifyMembership` — the single funnel every emission path (add/remove/retire/destroyAll) passes through — which is exactly what closes the `void pool.retire(h)` → `unhandledRejection` vector.

**Fact-check pass — sound**, all five load-bearing claims verified against the code:
1. `persistedMembership` stays a correct superset of live `entries` across add/remove/retire/re-add.
2. `remove` after `retire` truly forgets the host (guards on `persistedMembership.has`, persists the shrunk set, deletes it — even with no live entry).
3. `retire` genuinely cannot reject (no persist step; `tearDownEntry` swallows the destroy fault; `notifyMembership` isolates listener throws), so kolu's bare `void pool.retire(enc)` is safe.
4. persist-before-commit + `enqueueMutation` serialization preserved.
5. the #1708 typed-end ordering (drop membership + notify **before** destroy) is byte-identical to the pre-diff `remove`.

24/24 unit tests green. One pre-existing non-defect noted (a throwing `opts.log` would escape — a trusted diagnostic sink unchanged by this diff). No fixes required.

srid:
## Darwin e2e — retuned sample (srid-ratified, in progress)

Per srid's ruling, the two prior `ci::e2e@aarch64-darwin` reds were the run's **own concurrent load** tripping the local-padi liveness watchdog (#1776), on a verified-clean, cool box — not a #1775 defect. Retune: darwin e2e worker cap **6 → 3** ([#1778](https://github.com/juspay/kolu/pull/1778)).

**Exact tree under sample (honesty over tidiness):**
- `#1775 @ 98517e415` (this PR's HEAD) **+** retune `@ c466a4f9d` (from #1778), cherry-picked onto throwaway branch `w1775-retune-sample @ afe882fd6`.
- Lane: `ci::e2e@aarch64-darwin` on `rasam`, `PAR=3`.

This is a diagnostic sample of the retune, not #1775's official check. Result appended when it settles.

srid:
**Retune sample result: FAILED — wedged even at PAR=3.** `ci::e2e@aarch64-darwin` on `w1775-retune-sample@afe882fd6` (= #1775 @ 98517e415 + retune @ c466a4f9d) failed at 4m57s. PAR=3 confirmed applied (`workers=3 cores=24 load=3 cap=3` — cool box), yet 2 of 3 workers still hit the #1776 killAll-500 liveness cascade (~994 scenarios churned). So the collision is **not load-marginal** — it reproduces at half parallelism on a cool box, which disproves the retune's premise and shows the darwin e2e can't go green here until the #1776 product fix (local-arm liveness tolerance) lands. Disposition (accept on green-elsewhere + tracked-#1776 vs hold-for-fix) is with srid, now backed by three wedged samples (PAR=6 ×2, PAR=3 ×1). #1775's code stays off the failing path.

srid:
**Control run (srid-ratified, in progress).** To prove #1775's innocence *dynamically*, a control sample runs `ci::e2e@aarch64-darwin` at PAR=3 on **#1775's exact merge-base + the retune and nothing else** — none of #1775's code. Tree: `control-mergebase-retune @ e34365a75` = merge-base `10070a662` (= W9/#1764, exactly what #1775 forked from) + retune `c466a4f9d`. Same box (rasam), same lane. **Reading:** control **wedges** → the defect is master-resident/environmental, #1775 innocent dynamically too (disposition = clean policy call). Control **greens** → the split tree is implicated despite the static analysis → stop everything. Result appended when it settles.

srid:
## CI on the final HEAD `dd43e1c64`

Ran because the earlier 26/27 predates this HEAD (provenance below).

### ✅ Linux — full `ci::default` GREEN on the final tree
`x86_64-linux` on `naiveintent`, SHA `dd43e1c64`: **14 ok · 0 failed · 0 errored — OK**.

| node | |
| --- | --- |
| `ci::e2e@x86_64-linux` | ✔ 5m37s |
| `ci::unit@x86_64-linux` | ✔ 1m54s |
| `ci::smoke` · `ci::biome` · `ci::fmt` · `ci::nix` · `ci::flake-check` · `ci::home-manager` · `ci::pnpm-hash-fresh` · `ci::atlas-sync` · surface-example builds | ✔ all |

### ⚠️ Darwin — NOT claimed on this HEAD
- **Darwin non-e2e** (unit / smoke / biome / fmt / builds): the earlier all-green numbers ran on `98517e415`, **not** this HEAD — they are **not** carried over. A re-run on `dd43e1c64` is needed to assert them (currently blocked: rasam's odu/ssh transport is dropping mid-run — `agent exited code=255` — two consecutive control runs died on it).
- **Darwin e2e**: stays under the **#1776** disposition (the local-padi liveness force-cycle; reproduced 3×, filed #1776, tracked #1777).

### Provenance (straight record)
- **`dd43e1c64`** (this HEAD) = **`98517e415`** + master-merge (`#1771` boundToPid + `#1774` statepip).
- The earlier **26/27** ran on **`98517e415`**, which is **post-gauntlet** — `persistedMembership` (`ac11544cb`) and both codex fixes (`eec8c24f5`, `007ef9e66`) are ancestors of it. So the untested delta on `dd43e1c64` is the **master-merge**, not the gauntlet fixes.

_Note: `#1771` (test daemons die with their run) is now in this tree — the bare-host daemon-leak cure — so a fresh darwin sample is worth it once rasam's transport is stable._

Issue #1776, opened 2026-07-12
e2e(darwin): local padi link liveness-timeout force-cycles an alive-but-slow padi mid-run → worker queue-drain


srid:
## Summary

On the `aarch64-darwin` e2e lane under load, a **single worker's LOCAL padi link liveness-probe times out** mid-run — `"[padi-local local] liveness probe timed out — remote wedged, force-cycling the link"` — force-cycling an **alive-but-slow** local padi. The reconnect machinery *works* (it re-dials and heals), but the force-cycle leaves a down-window (observed up to ~16s) during which **every** padi RPC returns HTTP 500 via `reServeSurface` (`"invoked with no live upstream link"`). The per-scenario `Before` hook's `killAll` reset hits that window, fails after retries, and — per the documented queue-drain mode (`packages/tests/support/hooks.ts:458-460`) — the worker then fails hundreds of subsequent scenarios, reddening the lane.

Surfaced diagnosing PR #1775's darwin e2e (which is otherwise green on linux and on every non-e2e darwin node); #1775's code is not on this path.

## What it is NOT (so nobody re-litigates)

- **Not the teardown race (#1719).** This is a *mid-run, steady-state* link event, not scenario teardown.
- **Not a startup readiness-gate gap.** Verified: the harness's `startServerChild` awaits `waitForPadiLive()` on **every** server (re)start (`hooks.ts:890`, called at `BeforeAll` line 964 and the `@kaval-restart` After-hook restart line 1301), and `waitForPadiLive` polls `killAll` until it 200s — i.e. it already gates on the local padi link. The gate is correct.
- **Not non-healing / missing reconnect wiring.** The session DOES re-dial and heal: pid 26994 went `connected → disconnected` at 10:51:43, `host unreachable — retrying in 2000ms… (attempt 1)`, `disconnected → connecting → connected` by 10:51:59, then ran ~6 min healthy before the next wedge.

## The actual mechanism (verbatim, `server-w1.log`, worker:1)

```
[padi-local local] connection: connecting → connected
[padi-local local] liveness probe timed out — remote wedged, force-cycling the link
[padi-local local] connection: connected → disconnected
[padi-local local] host unreachable — retrying in 2000ms… (attempt 1)
[padi-local local] connection: disconnected → connecting
[padi-local local] connection: connecting → connected      ← heals
… later, under sustained load, wedges again …
```

Then, front-side, for the down-window:

```
[...]

PR #1778, opened 2026-07-12
ci(e2e): drop darwin worker cap 6→3 to clear the #1776 liveness force-cycle

srid:
**Ratified lane change (srid), not a workaround-in-the-dark.** Drops the `aarch64-darwin` e2e worker cap from **6 → 3**.

Diagnosed in [#1776](https://github.com/juspay/kolu/issues/1776): the concurrent load of 6 e2e workers + odu + nix on one mac trips the **local padi link's liveness watchdog** ("remote wedged — force-cycling the link"), force-cycling an *alive-but-slow* padi and 500-ing every proc during the reconnect down-window → a worker **queue-drains**. Two darwin-e2e samples on a **verified-clean, cool box** (srid swept; load 2.3) both wedged — one worker, then all four — establishing it's the **run's own load**, not a per-PR defect. `PAR=3` cuts concurrent local-padi pressure below the liveness threshold; linux keeps 8 (its link stack is reliable).

Seam: `justfile` darwin `cap`. The recipe already honors `CUCUMBER_PARALLEL`, but odu's remote lane runs the recipe fresh on the target, so the tree change is what reaches it — hence a shipped cap change rather than a run-only env override.

Surfaced by #1775's darwin e2e; #1775's own code is off the failing path (behaviour-neutral, unreached by the single-host suite).

### DIFF
--- a/justfile
+++ b/justfile
@@ -257,19 +257,24 @@ test: install
     # unlimited; this is free insurance on every platform.
     ulimit -n 65536 2>/dev/null || true
     # Worker-count cap (the count itself is computed below, after the suite
-    # lock): 6 on darwin, 8 elsewhere. PAR=8 on the 24-core darwin host (rasam)
+    # lock): 3 on darwin, 8 elsewhere. PAR=8 on the 24-core darwin host (rasam)
     # maximizes throughput but its higher concurrent load pressures the
     # slow-hydration tail — under load a handful of interaction waits
     # (per-terminal Code-tab history enablement, content settle) intermittently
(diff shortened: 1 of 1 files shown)


srid:
Closing per coordinator recommendation ratified with the W11 disposition: the retune's premise (the wedge is load-marginal) was disproven by the PAR=3 cool-box sample — the #1776 liveness collision reproduces at half parallelism, so capping a 24-core box at 3 workers is throughput loss without the fix it promised. The darwin e2e question lives in #1776.

PR #1781, opened 2026-07-12
fix(surface-remote): local padi liveness consults the process table (#1776)

### LINKED ISSUE #1776: e2e(darwin): local padi link liveness-timeout force-cycles an alive-but-slow padi mid-run → worker queue-drain

srid:
Resolves #1776 — the darwin-e2e local-padi liveness wedge.

## The wedge

On the `aarch64-darwin` e2e lane under worker load, the **local** padi arm's liveness watchdog force-cycled an **alive-but-slow** padi whenever its heartbeat went silent under CPU pressure. Each force-cycle left a ~16s down-window where every RPC 500s via `reServeSurface` ("no live upstream link"); the per-scenario `Before`-hook `killAll` hit that window, exhausted retries, and the worker queue-drained (25/504 scenarios, 225 `killAll` 500s).

## The fix (orchestrator ruling, P5)

A heartbeat is the **only** life-oracle a **remote** (ssh) arm has — the network can die silently, so silence must mean dead. But a **local** (stdio/endpoint) arm has a *superior* oracle it can't: **the same-box process table**. Treating heartbeat-silence as death when process-aliveness is directly observable is using the worse authority — that was the defect, not the timeout value.

- `@kolu/surface-remote` `Connection` gains an **optional** `processAlive?: () => boolean` same-box oracle — **arm structure, not a knob** (the local endpoint connector supplies it by construction; the ssh connector omits it). On a heartbeat timeout the watchdog now:
  - **no oracle (ssh)** → force-cycle — *today's semantics, provably untouched* (load-bearing for silent network death);
  - **oracle + dead** → force-cycle now;
  - **oracle + alive** → don't cycle, keep probing (slow under load ≠ dead);
  - **alive but silent past a baked minute-scale dead-man ceiling** → force-cycle anyway (a deadlocked-but-alive padi still heals).
- The local `endpointConnector` supplies `processAlive` from padi's **own pid gate** (`gatePid` + `isHolderLive` = `kill(pid,0)`).

## Pins

`liveness.test.ts` gains the three arcs (timeout+alive→no cycle; timeout+dead→cycle now; ceiling→cycle despite alive); the existing ssh `describe` pins the fourth (no oracle → force-cycle on the first timeout — remote arm untouched). `ref-surface-remote.mdx` documents the contract.

## Acceptance

Full `ci::default@aarch64-darwin` green on rasam (the issue's own bar). Review gauntlet applies (framework change).


srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 rounds · codex reviewed at `xhigh` reasoning effort · base `master` (the #1776 delta)

_codex caught two more real bugs beyond code-police: (F1 blocking) probe-episode state survived connection replacement → a healthy successor got force-cycled; (F2 major) the dead-man ceiling used wall-clock `Date.now()`, counting laptop-sleep toward the 60s. Both fixed + regression-pinned (connection-scoped reset in `handleClosed`; `monotonicNow()` + a divergent wall-vs-monotonic test)._

### Round 1

**codex** — approved: `false`

The local/remote oracle split is structurally defensible, but the implementation has one blocking connection-lifetime bug and one major timekeeping bug. Most importantly, a timed-out probe can survive replacement of its connection and prevent every replacement connection from ever being probed, causing repeated force-cycles.

Findings:
- `F1` · blocking · open — The new liveness episode state is session-scoped even though it belongs to one connection. The heartbeat is created only once (`startLiveness` returns whenever `liveness !== null`) and is stopped only on session destruction, but neither `inFlightProbe` nor `livenessUnresponsiveSince` is reset when `current` closes or is replaced. After connection A times out, its forever-pending probe remains cached at line 628; connection B therefore reuses A's promise instead of calling B's `isAlive`. B will then be force-cycled despite being healthy—immediately for an ssh arm or after the local ceiling. The timestamp can likewise leak across a natural close/manual recheck, shortening B's independent ceiling. The epoch guard only prevents a late settlement from writing; it does not prevent stale state from being read by the replacement dial. (packages/surface-remote/src/session.ts:608)
- `F2` · major · open — The dead-man duration uses `Date.now()`, so suspended time and wall-clock steps count toward the 60-second ceiling. This bypasses `createHeartbeat`'s explicit suspension protection, which compares wall time with a monotonic clock and refuses to treat laptop sleep or a frozen runtime as a fair liveness interval. If a local slow spell begins before sleep, the first post-resume stale result can treat the entire sleep as deadlock time and force-cycle a merely resumed padi—the same class of false recovery the heartbeat's suspension logic is designed to prevent. A forward system-clock adjustment has the same effect; a backward adjustment can defer the ceiling. (packages/surface-remote/src/session.ts:677)


**claude** — Both findings are real; fixed both (connection-scoped probe reset + monotonic ceiling clock) with a new F1 regression test.

- `F1` **fixed** — Real bug, agreed. The `createHeartbeat` instance is session-scoped (born once at first connect via `startLiveness`'s `liveness !== null` guard, disposed only at `destroy`), so its probe-episode state leaks across connection replacement. When link A's `isAlive()` never settles, `inFlightProbe` stays cached; the replacement link B's probe (session.ts:~633) reuses A's forever-pending promise and never calls B's `isAlive()` → B is force-cycled while healthy (immediately for ssh, after the ceiling for local). The `.finally` epoch-guard only stops a late settle from WRITING stale state; it can't stop the replacement dial from READING a never-settling one — codex is exactly right. Fix: `handleClosed` now resets `inFlightProbe = null` and `livenessUnresponsiveSince = null` right where it nulls `current` (session.ts:739-751) — the single choke-point every death routes through, so every successor starts a fresh episode. A late settle from the retired probe is inert (its `inFlightProbe === probe` and `current === connAtLaunch` checks both fail). Added the regression codex asked for: `liveness.test.ts` "a REPLACED connection is probed with its OWN isAlive" — dial A never answers + oracle-dead → force-cycle; dial B healthy; asserts B's own `isAlive()` is called and B stays connected (fails without the reset because B's `isAlive` is never invoked).

- `F2` **fixed** — Agreed. The dead-man ceiling read `Date.now()`, so a laptop sleep / forward clock-step between the pre-suspension stale and the first post-resume stale would fold the suspended span into `silentForMs` and force-cycle a merely-resumed-but-alive padi — the exact suspension-induced false recovery `createHeartbeat` goes to lengths to prevent, and that this whole PR exists to kill. `onStale` only ever fires over a continuously-RUNNING probe window (the heartbeat VOIDs a suspension-straddling window via `SUSPENSION_SLACK_MS` and re-probes without calling us), so the ceiling must accrue running time too. Fix: exported the heartbeat's own `monotonicNow` (heartbeat.ts:81 — the same `performance.now()`-or-`Date.now()` clock the watchdog reads for its fair-window guarantee) and the ceiling now reads it (session.ts:~684). Reusing the heartbeat's clock rather than minting a parallel one honors the repo's "reuse the source of truth" rule; the export is an internal sibling-shared primitive (the whole `heartbeat` module — `createHeartbeat`, `DEFAULT_HEARTBEAT_*` — is undocumented in the surface Reference pages and consumed the same cross-package way, so no ref-doc/drishti gate: additive, no signature change). No independent clock seam / knob was added (the no-knob ruling stands). The existing "dead-man ceiling" test still passes: vitest fake timers advance `performance.now()` in lockstep with `Date`, so the 60s running-time cycle still fires. A bespoke wall-vs-mono divergence test would need a clock-injection seam into the session ceiling that the no-knob ruling forbids; the fix is verified structurally (identical clock basis as the heartbeat's proven suspension logic) plus the existing running-time ceiling pin.

Gate: `just check` (typecheck) + `just lint` (biome) green; `liveness.test.ts` + `heartbeat.test.ts` 25/25 pass; `just fmt` clean.

commit: `0c94329089fe624ab4258b5b3f672ff77963e4e0`


### Round 2

**codex** — approved: `false`

F1 is fully fixed: connection closure now retires both pieces of probe-episode state, the late-settlement guards remain safe, and the regression test exercises the original failure. F2's implementation is also corrected, but the requested wall-versus-monotonic regression remains untested, and the stated reason that testing requires a production clock-injection seam is incorrect.

Findings:
- `F1` · blocking · resolved — The cached probe and dead-man timestamp previously survived connection replacement. (packages/surface-remote/src/session.ts:748)
- `F2` · major · open — The production code now correctly uses the heartbeat's monotonic clock, but the tests still advance wall and monotonic time together. Consequently, reverting the ceiling to `Date.now()` would leave every current test green; the suspension/clock-step regression identified by F2 is not pinned. (packages/surface-remote/src/liveness.test.ts:364)

_codex on the rebuttal:_ F1: conceded and resolved. Resetting both episode fields in `handleClosed` is at the required connection-lifetime choke point, and equality/current guards make a retired probe's late settlement inert. The new test fails for the original implementation and proves the successor invokes its own probe.

F2: the code fix is correct, but I do not accept the test objection. A wall-versus-monotonic test does not require adding any production seam, much less a user-facing tuning knob. `monotonicNow()` reads `performance.now()` dynamically, so the test can spy on that global and control it independently of fake wall time; this pattern already exists in the repository. The current ceiling test cannot distinguish `monotonicNow()` from `Date.now()`, so it does not prevent this major timing regression from returning.


**claude** — Added the divergent-clock regression codex asked for; it pins the ceiling to the monotonic clock (green with `monotonicNow()`, red on revert to `Date.now()`).

- `F1` **fixed** (already resolved round 1) — codex marked it resolved; no further change. The close-time reset in `handleClosed` (`session.ts`) retires both `inFlightProbe` and `livenessUnresponsiveSince` at the connection-lifetime choke point, and the equality/current guards make a retired probe's late settle inert. Regression test at `liveness.test.ts:386` still green.

- `F2` **fixed** — Conceded the test objection. Codex is right: a wall-vs-monotonic test needs no production seam — spying `performance.now()` independently of the fake wall clock distinguishes `monotonicNow()` from `Date.now()`, and that pattern already exists in the repo (`client/src/canvas/viewport/hostSwitchAbort.test.ts:40`). My round-1 claim that this required a clock-injection knob was wrong. Added `liveness.test.ts` "dead-man ceiling accrues MONOTONIC running time, not wall time … a mid-silence SUSPENSION never spends the ceiling (#1776 F2)":
  - `monoMs` is a `vi.spyOn(performance, "now")` value independent of the fake timers. `run(ms)` advances wall AND `monoMs` in 1s (sub-`SLACK`) steps so every probe window's wall/mono gap is ~0 and the heartbeat fires normal stale verdicts (no void). `suspend(ms)` advances WALL alone with `monoMs` frozen — the heartbeat voids every straddling probe (as across a real laptop sleep), so no stale verdict fires and the monotonic silence clock does not tick.
  - Timeline: establish `livenessUnresponsiveSince` at mono ~25s, then a 20-minute WALL suspension with mono frozen, then resume for one more running verdict. Total MONOTONIC silence stays ~25-30s (< the baked 60s ceiling) → no cycle → `teardowns === 0`.
  - Verified BOTH directions: green as shipped (`monotonicNow()`); flipping the ceiling to `Date.now()` makes the post-resume verdict see the whole 20-minute wall jump → force-cycle → `teardowns === 1` → test fails at `liveness.test.ts:503`. So the test cannot pass unless the ceiling reads the monotonic clock. The heartbeat's own void logic is unaffected by the session's ceiling-clock choice, so the same stale verdicts fire in both variants — the only difference is what the ceiling measures, which is exactly what's under test.

Gate: `just check` (typecheck all packages + biome) green; `just lint` + `just fmt` clean; `liveness.test.ts` 9/9 pass. No production code changed this round (test-only), so no drishti/ref-doc gate.

commit: `696a8d41871527e3c5bbc2d918cd468a176ab1e6`


### Round 3

**codex** — approved: `true`

Both findings are now fully resolved. Connection-scoped probe state is retired at replacement with regression coverage, and the dead-man ceiling uses the heartbeat's monotonic clock with a divergent wall-versus-monotonic test that specifically fails if the implementation reverts to `Date.now()`. No regression was introduced this round.

Findings:
- `F1` · blocking · resolved — Probe-episode state previously survived connection replacement. (packages/surface-remote/src/session.ts:748)
- `F2` · major · resolved — The monotonic ceiling behavior previously lacked regression coverage distinguishing it from wall time. (packages/surface-remote/src/liveness.test.ts:444)

_codex on the rebuttal:_ F1 remains resolved; the production reset and replacement regression are intact.

F2: conceded and resolved. The new test independently controls `performance.now()`, advances wall time through a simulated 20-minute suspension, then produces a post-resume stale verdict. It remains green with `monotonicNow()` and would force a teardown with `Date.now()`, so it directly pins the original defect without adding a production clock seam.

srid:
(lens-debate consensus — see commits 32cbf47f7/3e4699458/3ec60dc76)

PR #1787, opened 2026-07-12
Host row in the mobile pull-down sheet

srid:
**On the touch layout kolu rendered no host chrome at all** — `ChromeBar` and its `HostSelectorStrip` are `isDesktop()`-gated, so on a phone you couldn't see which host you were on, switch hosts, or add one. The multi-host engine underneath (the keyed padi host map, W9 retained switching, W10 persisted fleet, W5 attention) is fully layout-agnostic and already worked; it just had no mobile face. This adds one: a **host row at the top of the pull-down chrome sheet** (`MobileChromeSheet`), the mobile twin of the desktop strip.

### The seam — one engine, two thin leaves

```
                 App.tsx · isDesktop()
                  ┌──────────┴──────────┐
              desktop                 touch
          ChromeBar               MobileChromeSheet
      HostSelectorStrip            MobileHostRow  ◀── new
                  └──────────┬──────────┘
        padiMap.entries · entry.cells.urgency · setActiveHost
        hostChipTone (dotClass · hostHue) · ATTENTION_PILL_CLASS
```

`MobileHostRow` reads the **exact same seams** the desktop strip reads and writes the switch with the **same** `setActiveHost`. No new store, stream, or keybinding.

### What a phone user gets

- **Chips in the desktop pixel-language, verbatim** — the connection-dot tone (`dotClass`, green only for `connected`), the per-host identity hue (`hostHue`), and the amber attention pill (`ATTENTION_PILL_CLASS`, the token from #1774 the Dock badge also uses). A dot / hue / pill means the *same thing* on a phone as on a laptop — a shared vocabulary, **not** a mobile restyle.
- **Tap a chip to switch** — `setActiveHost`, instant and no-reload (W9 retained surface).
- **Add-host as a full-width in-sheet section** — not the desktop popover, which anchors to the tiny `+` and clips off the right edge at phone width. The `+` expands a full-width section *inside* the sheet.
- **Touch ergonomics** — every chip and the `+` are ≥44px hit targets; the chip row scrolls horizontally when the fleet overflows, with the `+` pinned outside the scroll so it stays reachable.

### Reuse, not duplication

The add-host *mechanism* and the ALPHA warning are now shared by both surfaces rather than copied:

| Extracted | Shared by | Why |
| --- | --- | --- |
| `addHost(raw, onAdded?)` | desktop `AddHostAffordance` + mobile section | one add path (parse · `hosts.add` · activate-on-join · error toast); only the container differs |
| `RemoteHostsAlphaNotice` | both | the alpha warning survives into the sheet **verbatim** by construction — an add-host on mobile is exactly as unfinished as on desktop |

> **Design philosophy** — *no fallbacks* (add errors surface loud via toast, never a silent no-op); *no new electricity* (`MobileHostRow` is a presentation leaf beside its desktop sibling in `host/`, not a `@kolu/*` package — it hides no transport/reconnect/persistence volatility of its own); *reuse the source of truth* (shared `addHost` + `RemoteHostsAlphaNotice`, shared `hostChipTone` / `ATTENTION_PILL_CLASS`).

Covering e2e in `mobile-hosts.feature` (the always-present local surface; multi-host switch + the attention pill go through live chrome-devtools evidence, since the e2e env can't provision a second host over ssh). Docs (`remote-hosts.mdx`) and changelog updated.

> **Plan of record:** the ratified Atlas note lives in #1786 (kept as the note's single home). This branch does **not** duplicate it; once #1786 merges, the note is finalized in place (`proposed → implemented` + PR link) as a follow-up commit here.


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `4460f78b183d`

### Applied (4)
- `lowy-1` Per-chip attention-fact wiring is an un-encapsulated leaf, now duplicated a third time — commit `46597a32d`
- `hickey-1` Per-chip awaiting-count derivation duplicated (now a 3rd copy) — (uncommitted)
- `hickey-2` Host glyph+label pairing re-hand-rolled in mobile chip, bypassing its single owner — commit `6cfa76656`
- `hickey-4` File name diverges from its primary export — commit `a37f11549`

### Agreed — no change (3)
- `lowy-2` Add-host input interaction duplicated between mobile section and desktop popover (packages/client/src/host/MobileHostRow.tsx:130)
- `lowy-3` Electricity verdict: MobileHostRow is correctly a leaf, not a new receptacle (packages/client/src/host/MobileHostRow.tsx:180)
- `hickey-3` Focus-on-mount effect duplicated verbatim between add surfaces (packages/client/src/host/MobileHostRow.tsx:136)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 round(s) · codex reviewed at `xhigh` reasoning effort · base `4460f78b183d`
### Round 1

**codex** — approved: `false`

The shared host-state and add-host paths are structurally sound, but the mobile surface is not ready yet. The add controls conflict with Corvu's drag handling, and mobile users can create persistent hosts without any way to remove them. I also found a no-op tap that still closes the sheet, an Escape propagation bug, misleading callback error handling, incorrect chip styling, incomplete tab semantics, two small API/docs issues, and an untracked review artifact.

Findings:
- `F1` · major · open — The add section and the deliberately outside `+` trigger do not opt out of the enclosing Corvu drawer's drag recognizer. Only the chip scroller stops `pointerdown`. Corvu treats the remaining input, button, link, and trigger as draggable and begins dragging after roughly 0.3px of movement, so normal finger jitter can suppress their actions or move/dismiss the sheet. Playwright's stationary `.tap()` does not exercise this path. (packages/client/src/host/MobileHostRow.tsx:137)
- `F2` · major · open — The mobile row allows adding persistent remote hosts but provides no way to remove them. Removal exists only in `HostSelectorStrip`, which is not mounted on mobile. A phone user can therefore add a failed or obsolete host that reconnects after every restart but cannot delete it from that surface; this also undercuts the changelog's "Manage your hosts from a phone" claim and the docs' explicit-remove lifecycle. (packages/client/src/host/MobileHostRow.tsx:64)
- `F3` · minor · open — `props.onSwitch()` runs even when the tapped chip is already active. Thus the supposedly no-op local-host scenario still closes/unmounts the sheet, contradicting the callback contract ("after a switch") and `mobile-hosts.feature:25-31`. The current visibility assertion may race the drawer's closing transition. (packages/client/src/host/MobileHostRow.tsx:87)
- `F4` · minor · open — Escape calls the add section's close callback but does not consume the keyboard event. Corvu's document-level Escape listener subsequently sees the same event and closes the entire drawer, contrary to the component's stated contract that Escape only collapses the section. (packages/client/src/host/MobileHostRow.tsx:151)
- `F5` · minor · open — The consumer-owned `onAdded` cleanup runs before `requestActivateOnJoin`. If that callback throws, the host was successfully added but activation is skipped, and the promise catch misleadingly reports `Couldn't add ...`. This lets presentation policy interrupt completion of the shared mechanism. (packages/client/src/host/addHost.ts:30)
- `F6` · minor · open — `host-hue-ring` is applied to the entire chip button as well as the connection dot. That class is specifically defined as the dot's 2.5px identity ring, so every inactive chip receives a colored outer shadow—particularly prominent in dark mode—despite the code saying inactive identity should remain quiet and use `border-edge`. (packages/client/src/host/MobileHostRow.tsx:75)
- `F7` · minor · open — The row declares a `tablist` and each host a `tab`, but it implements neither roving tabindex/arrow navigation nor a tab-to-tabpanel relationship. Every chip remains in the normal tab order, and switching closes the containing drawer rather than selecting an associated tabpanel, so assistive technology receives misleading semantics. (packages/client/src/host/MobileHostRow.tsx:65)
- `F8` · nit · open — The newly edited canonical page still says the alpha notice and `+` live in the "host strip" and later says hosts return to or leave "the strip." That terminology is now false for the documented phone workflow, where the control is a row in the pull-down sheet. (website/src/content/docs/remote-hosts.mdx:19)
- `F9` · nit · open — `REMOTE_HOSTS_DOC` is exported even though it has no production caller outside this module. This unnecessarily enlarges the module's public surface. (packages/client/src/host/RemoteHostsAlphaNotice.tsx:13)
- `F10` · nit · open — This untracked file is review-gauntlet output, not product source, and `.gauntlet/` is not ignored. It is easy to accidentally include in the PR. (.gauntlet/lens-comment.md:1)


**claude** — Fixed the Corvu drag opt-outs, the no-op-tap sheet dismissal, the Escape leak, the addHost ordering, the chip hue-ring bleed, the docs/changelog wording, the stray export and the untracked gauntlet dir; disputed the ARIA-tabs finding and partially disputed the mobile-remove finding as a plan-of-record design gate.

- `F1` **fixed** — Real gap, confirmed against Corvu's own drag recognizer: `locationIsDraggable` (@corvu/drawer dist index.js:81-93) walks from the pointerdown target up to `Drawer.Content`, and any control that doesn't stop the event (or carry `data-corvu-no-drag`) arms the drag on ~0.3px of movement. The chip scroller already stopped it; the `+` open button and the add-section subtree did not. Added `onPointerDown={(e) => e.stopPropagation()}` to the `+` button (MobileHostRow.tsx:205-209) and to the `MobileAddSection` root div (MobileHostRow.tsx:141-146) — the same guard every other sheet control (palette/settings/inspector) already wears in MobileChromeSheet.tsx. stopPropagation (not `data-corvu-no-drag`) keeps this file idiomatic with its siblings, and it leaves the chips' native horizontal scroll untouched.
- `F2` **partial** — Conceded the honest-record half, disputed the new-UI half as a plan-of-record gate. Conceded: the changelog title "Manage your hosts from a phone" overclaimed a lifecycle the surface doesn't deliver, so I retuned it to "See, switch, and add hosts from a phone" (unreleased.mdx) and made the docs' persistence/remove passages surface-neutral (see F8) so nothing implies phone-side removal. Disputed: adding a touch remove affordance is not a bug fix — it's a NEW interaction. Desktop removal is a hover-reveal ✕ (HostSelectorStrip.tsx:247-258), which has no touch analog; a touch-safe remove needs a new gesture (long-press / swipe / edit-mode) that is NOT in the ratified, srid-approved mockups (#1786) that are this feature's plan of record and deliberately not in this branch. Inventing that interaction mid-review is exactly the unratified design drift the mockup gate exists to prevent. The mobile remove capability is deferred to the plan-of-record finalization (post-#1786) — a design/scope gate, not a code change landable in this review round. The `client.hosts.remove` mechanism already exists and can be shared (like `addHost` was) the moment the interaction is ratified.
- `F3` **fixed** — Correct: the old handler ran `props.onSwitch()` (= `onClose`, which dismisses the drawer) even on the already-active chip, so a no-op tap tore the sheet down and raced the "chip visible and active" assertion against the closing transition. Moved BOTH `setActiveHost` and `props.onSwitch()` behind an early `if (isActive()) return` (MobileHostRow.tsx:87-92): a tap on the current host is genuinely inert and leaves the sheet open. Strengthened the scenario to prove it — added "Then the mobile chrome sheet should be visible" after the no-op tap (mobile-hosts.feature:32), which now fails if the sheet ever dismisses on a no-op.
- `F4` **fixed** — Confirmed against the library: Corvu's `createDismissible` dismisses on Escape only when `!event.defaultPrevented` (@corvu/utils dist chunk 4XMQWCCT.js:20), via a document-level keydown listener (chunk WW7THO4B.js:15). The input's Escape branch closed the section but didn't `preventDefault`, so the same keystroke also closed the whole drawer. Added `e.preventDefault()` before `props.onClose()` (MobileHostRow.tsx:151-160). stopPropagation is unnecessary given the `defaultPrevented` gate, so I kept the fix minimal.
- `F5` **fixed** — Reordered so the shared mechanism completes before presentation cleanup: `requestActivateOnJoin(host)` now runs before `onAdded?.()` (addHost.ts:30-38), so a throwing consumer callback can no longer skip activation or get mislabeled by the `.catch` as "Couldn't add" when the add actually succeeded. I did NOT wrap `onAdded` in a try/catch — the two real callers pass only signal setters (`setDraft`/`setOpen`/`setAddOpen`) that can't throw, and adding defensive plumbing around non-throwing code would be exactly the kind of fallback cruft the repo's "fail fast, no fallbacks" rule rejects.
- `F6` **fixed** — Correct: `.host-hue-ring` is defined (index.css:102-104, :root.dark 118-120) as the 2.5px identity ring for the connection DOT, but it was also on the chip's button shell (MobileHostRow.tsx:75), painting a colored outer box-shadow on every chip — active and inactive — which is loud in dark mode (55% hue) and contradicts the shell's own "quiet at rest / `border-edge`" comment and the desktop chip (which never rings the shell). Removed `host-hue-ring` from the shell class; kept it on the dot (line 95) and kept `--host-hue` driving the active border/belly via `classList`. Inactive chips are now quiet, matching desktop.
- `F7` **disputed** — This mirrors, verbatim, the desktop `HostSelectorStrip`'s existing ARIA vocabulary: `role="tablist"` + `role="tab"` + `aria-selected` (HostSelectorStrip.tsx:187-188, 470-478, 717). "Chips reuse the desktop vocabulary verbatim" is a ratified constraint of this feature, and the roles are part of it. Switching the mobile row alone to `aria-current`/`aria-pressed` (or adding roving tabindex / a tabpanel relation to mobile only) would DIVERGE from desktop and split the ARIA contract across two surfaces for the identical control — the precise inconsistency the shared-vocabulary rule forbids. The incomplete-tabs property, if it is one, is a PRE-EXISTING property of the desktop strip, not something this change introduced; correcting it belongs in a dedicated a11y pass that fixes both surfaces together, not a mobile-only mid-review divergence.
- `F8` **fixed** — Correct: after the intro/add-host edits, the doc still said "host strip" in the alpha caution and the persistence passage, now false for the documented phone workflow. Made them surface-neutral: the caution now reads "the **+** on the host row" (remote-hosts.mdx:20), the persistence/remove passage reads "comes back on the host row … leaves your fleet" (73-77), the pre-seed line reads "add from the host row" (86), and the frontmatter description reads "add hosts from the host row" (3). Left line 30 as-is on purpose — it correctly spells out "a strip in the top bar on desktop, and a row in the pull-down chrome sheet on a phone," which is the accurate per-surface statement.
- `F9` **fixed** — Confirmed `REMOTE_HOSTS_DOC` has no consumer outside RemoteHostsAlphaNotice.tsx (grep across packages/ and website/). Dropped the `export`, making it module-private (RemoteHostsAlphaNotice.tsx:13).
- `F10` **fixed** — `.gauntlet/lens-comment.md` is lens-debate scratch and `.gauntlet/` was not ignored (`git check-ignore` returns non-zero; the `.gitignore` comment above `.be-review-comments/` was misleading). Reused the existing per-worktree-scratch ignore block: added `.gauntlet/` under it (.gitignore:30). The stray file now can't be accidentally staged. I stage only the files I changed regardless, so it was never at risk from this commit.

commit: `d114912e8ced1939035c96cfdcf70f24b04dcfa1`


### Round 2

**codex** — approved: `false`

Eight of ten findings are resolved. The Corvu interactions, no-op behavior, Escape handling, styling, docs, export hygiene, and scratch-file handling are fixed; I also accept mobile removal as deferred behind the ratified design gate. F5 remains partially fixed because cleanup exceptions are still caught and mislabeled as add failures, and F7 remains open because matching an existing invalid ARIA pattern does not make the new instance valid.

Findings:
- `F1` · major · resolved — The add controls previously remained eligible for Corvu drag recognition, making touch interactions unreliable. (packages/client/src/host/MobileHostRow.tsx:144)
- `F2` · major · resolved — The mobile surface can add persistent hosts but does not currently provide a touch-safe removal interaction. (packages/client/src/host/MobileHostRow.tsx:64)
- `F3` · minor · resolved — Tapping the active host previously dismissed the sheet despite being a no-op. (packages/client/src/host/MobileHostRow.tsx:93)
- `F4` · minor · resolved — Escape previously closed both the add section and the containing Corvu drawer. (packages/client/src/host/MobileHostRow.tsx:163)
- `F5` · minor · open — Activation now happens before presentation cleanup, but the trailing `.catch(...)` still catches exceptions thrown by `onAdded` and reports `Couldn't add ...` even though the host addition succeeded. Reordering fixed skipped activation but not the misleading error classification from the original finding. (packages/client/src/host/addHost.ts:30)
- `F6` · minor · resolved — The connection-dot hue-ring class was incorrectly applied to the whole chip shell. (packages/client/src/host/MobileHostRow.tsx:78)
- `F7` · minor · open — The new row still declares `tablist`/`tab` semantics without roving focus, arrow navigation, or an associated tabpanel. Matching the desktop implementation makes the two surfaces consistently invalid; ARIA roles are an interaction contract, not part of the deliberate shared color/token vocabulary. (packages/client/src/host/MobileHostRow.tsx:65)
- `F8` · nit · resolved — The docs previously retained desktop-only "host strip" wording after documenting the phone row. (website/src/content/docs/remote-hosts.mdx:19)
- `F9` · nit · resolved — `REMOTE_HOSTS_DOC` was exported without an external consumer. (packages/client/src/host/RemoteHostsAlphaNotice.tsx:14)
- `F10` · nit · resolved — The untracked `.gauntlet/` review artifact was not ignored and could be staged accidentally. (.gitignore:30)

_codex on the rebuttal:_ F2: Conceded. The author corrected the overclaiming docs/changelog, and the missing removal interaction is genuinely gated on ratified touch design rather than an established interaction this patch merely omitted. Under the review's process-gate exception, it is resolved and deferred to plan-of-record finalization.

F7: I hold firm. The deliberate shared-vocabulary note explicitly concerns `hostChipTone`, `hostHue`, `dotClass`, and `ATTENTION_PILL_CLASS`; it does not require copying incomplete ARIA semantics. A `tab` role promises keyboard behavior and a tabpanel relationship that neither surface supplies. Desktop precedent explains the mistake but does not make the newly added instance valid. Fixing both surfaces together satisfies consistency without preserving the defect.

F5: The reordered activation is correct, but only half the finding is resolved. Because `.catch` remains after `.then`, it still converts a cleanup exception into a false add-failure toast. Using `.then(success, rejectionHandler)` preserves fail-fast behavior and catches only the operation whose error the toast describes.


**claude** — Conceded both remaining open findings: switched `addHost` to two-arg `.then(onFulfilled, onRejected)` so the toast only reports a real add rejection (F5), and dropped the mobile row's orphan `role="tab"`/`tablist` for `aria-current` selection buttons — matching desktop's own transient host picker `HostSwitcherRow`, not the top-bar tab strip (F7).

- `F5` **fixed** — Conceded. codex was right that a trailing `.catch` after `.then` also catches a throw from inside the fulfillment callback (`requestActivateOnJoin` / `onAdded`) and mislabels it `Couldn't add …` even though the add succeeded. Rewrote `client.hosts.add({ host }).then(onFulfilled, onRejected)` (addHost.ts:28-44): the rejection handler now receives ONLY a `client.hosts.add` rejection — the operation the toast names — while a cleanup throw stays an unhandled rejection and fails LOUD, which is exactly the repo's fail-fast/no-fallback stance. Added a comment spelling out why the two-arg form (not a chained `.catch`) is load-bearing.
- `F7` **fixed** — Conceded, with a sharper resolution than "fix both surfaces." The mobile row's interaction twin is NOT the persistent top-bar tab strip (`HostChip`, which does use `role="tab"`); it's desktop's TRANSIENT host picker `HostSwitcherRow` (HostSelectorStrip.tsx:298-301) — a pick-and-dismiss list of host options in a popover, the mobile sheet's exact analog — and that switcher ALREADY uses `aria-current`, not `role="tab"`. So converting the mobile chip to `aria-current={isActive() ? "true" : undefined}` (MobileHostRow.tsx:67-76) both removes the invalid orphan-`tab` contract codex flagged AND aligns with existing desktop precedent rather than diverging from it. The shared VISUAL vocabulary the deliberate-decision note protects (dotClass / hostHue / ATTENTION_PILL_CLASS) is untouched — ARIA roles were never part of that note. Container: dropped `role="tablist"`/`aria-label="Hosts"` for a plain div (MobileHostRow.tsx:206-213), matching `HostSwitcherRow`'s plain-div list container; `role="group"` was rejected by biome's `useSemanticElements` (pushes to `<fieldset>`, a form-control element), and desktop's picker container carries no role either. The e2e steps key off `data-active` (mobile_hosts_steps.ts:25), not the ARIA role, so the scenario is unaffected.

commit: `9ea3d23fcdcc3a4a1c6e62fa443e6557998d51fa`


### Round 3

**codex** — approved: `true`

All ten findings are resolved. The two-argument `.then` now preserves add-error provenance and lets fulfillment-path exceptions fail independently, while the mobile host controls now use appropriate selection-button semantics without claiming an incomplete tabs contract. No regressions were introduced by this round.

Findings:
- `F1` · major · resolved — The add controls previously remained eligible for Corvu drag recognition, making touch interactions unreliable. (packages/client/src/host/MobileHostRow.tsx:151)
- `F2` · major · resolved — The mobile surface can add persistent hosts but does not currently provide a ratified touch-safe removal interaction. (packages/client/src/host/MobileHostRow.tsx:64)
- `F3` · minor · resolved — Tapping the active host previously dismissed the sheet despite being a no-op. (packages/client/src/host/MobileHostRow.tsx:100)
- `F4` · minor · resolved — Escape previously closed both the add section and the containing Corvu drawer. (packages/client/src/host/MobileHostRow.tsx:170)
- `F5` · minor · resolved — A trailing promise catch previously mislabeled fulfillment-callback exceptions as add failures. (packages/client/src/host/addHost.ts:28)
- `F6` · minor · resolved — The connection-dot hue-ring class was incorrectly applied to the whole chip shell. (packages/client/src/host/MobileHostRow.tsx:85)
- `F7` · minor · resolved — The mobile row previously declared tab semantics without tab behavior or an associated tabpanel. (packages/client/src/host/MobileHostRow.tsx:65)
- `F8` · nit · resolved — The docs previously retained desktop-only "host strip" wording after documenting the phone row. (website/src/content/docs/remote-hosts.mdx:19)
- `F9` · nit · resolved — `REMOTE_HOSTS_DOC` was exported without an external consumer. (packages/client/src/host/RemoteHostsAlphaNotice.tsx:14)
- `F10` · nit · resolved — The `.gauntlet/` review artifact directory was not ignored. (.gitignore:30)

_codex on the rebuttal:_ F5: Conceded and verified. The two-argument `.then` ensures the add-failure toast can only handle rejection from `client.hosts.add`; exceptions from activation or presentation cleanup are no longer reclassified.

F7: Conceded and verified. `aria-current` on native buttons accurately represents the current host without promising tab-widget behavior. Aligning with the desktop transient host picker is the correct interaction precedent, and the shared visual-token constraint remains intact.

srid:

**No violations.** Two passes over the diff (`4460f78b1...HEAD`), elegance skipped (the simplify track already ran this gauntlet).

- **Pass 1 — rule checklist:** clean across the full built-in ruleset and `.agency/code-police.md`. This diff is itself substantially a rule-satisfying refactor — `dry-rule-of-three` (extracts `HostAwaitingPill`, `HostIdentityLabel`, `RemoteHostsAlphaNotice`, `addHost`, `focusOnMount`, `useHostAwaiting`, `useHostMembers` out of duplicated/inline logic), `no-silent-error-swallowing` + `toast-must-include-error-message` (every error path toasts with `err.message`), `caught-error-must-not-collapse-to-empty` (the `?? 0` is a value-channel default; real errors ride the separate `onError` → toast channel), `solid-reactive-prop-passed-to-hook-must-be-reactive` (each chip is a `<For>`-minted instance where `props.host` is fixed for its lifetime).
- **Pass 2 — fact-check:** clean. The extractions preserve correctness (no silent swallowing, `parseHostInput` confirmed total, no stale-state bugs in the no-op-tap guard or the `<Show>`-scoped draft signal, per-chip subscriptions disposed with their `<For>` owner — no leak).

PR #1783, opened 2026-07-12
Scrollback backfill: instant attach, real scroll-up history

srid:
**Attach now paints only the recent screenful, and scrolling up loads the rest.** Today every cross-host switch (or page reload) replays kaval's *whole* 10,000-line mirror before the terminal is usable — the [W9 full-replay cost](https://kolu.dev/atlas/padi.html#w9). This ships the [ratified plan](https://kolu.dev/atlas/scrollback-backfill.html): attach with a **bounded** snapshot (~1,000 lines), then backfill older lines into the terminal's **own** scrollback as the user scrolls up — the same buffer, scrollbar, and select/copy/search, no pager. The view never jumps; the only tell is the scrollbar thumb shrinking. Backfilled history is real terminal content — colors, wraps, and wide chars survive, because the bytes are replayed through xterm's parser before they're spliced in.

### How a chunk travels

```
kaval mirror ─ getHistory(before,max) ─▶ VT chunk ─▶ scratch xterm (real parser)
   (10k RAM)      one new wire verb                        │  steal BufferLines
       ▲                                                   ▼
   onTrim ─▶ mirrorBaseLine          scrollbackBackfill.ts: splice into live buffer's
   (eviction origin)                 CircularList, shift ydisp/ybase/savedY by M,
                                     fire _onScroll + refresh — viewport doesn't move
```

### The seam is a guarantee, not a convention

The backfill cursor is an **absolute mirror-line index**, seeded from the attach snapshot's `topLine` and carried on every reply. New output always appends at the mirror *bottom*, so it can never shift the index of a line the client already holds — a fetch serves *strictly above* the client's content **regardless of how many live deltas are in flight during the fetch**. A `have`-from-bottom cursor cannot: it compares the server's produced-line count against the client's received count, which differ by exactly the in-flight lag. The eviction origin the cursor rides on is tracked off the mirror's `onTrim` (kaval's one internals reach, contract-pinned). Pinned by `ptyHostHistory.test.ts` → *"serves strictly ABOVE the cursor even as output appends"*: attach, fetch, append 100 lines, re-fetch — identical chunk, zero duplicate, zero missing.

> A **width change** (reflow shifts absolute row indices) pauses backfill until the next snapshot re-seeds — the already-loaded history reflows correctly; only *further* loading waits. Continuous-across-resize backfill needs a reflow-invariant cursor: a named follow-up, not a silent gap.

### The pieces

[...]

PR #1792, opened 2026-07-13
fix(surface): own the mirror per-key pumps + type the stdio close reason (#1719)

### LINKED ISSUE #1719: Intermittent unhandled AbortError on stdio-link teardown (padiBinding reconnect flake)
## Intermittent unhandled `AbortError` on stdio-link teardown (the `padiBinding` reconnect flake)

`packages/server/src/padiBinding.test.ts > "reconnects when padi dies"` intermittently fails on CI (linux, ~1 in N runs; passes on rerun + clean on darwin) with an **unhandled rejection**:

```
[AsyncIdQueue] Queue[N] was closed or aborted while waiting for pulling.  (AbortError)
❯ @orpc/shared AsyncIdQueue.close  (reject on a pending pull)
❯ @orpc/standard-server-peer ClientPeer.close
❯ LinkStdioClient.handleTransportClosed  packages/surface/src/links/stdio.ts
❯ Socket.<anonymous>  (write.on("error") / inbound end)
```

### Root cause (precise)

When the source transport dies, `LinkStdioClient.handleTransportClosed()` calls `peer.close()`. orpc's `ClientPeer.close()` → `AsyncIdQueue.close()` calls **`reject(new AbortError(...))` synchronously on every PENDING pull** (`@orpc/shared`, the `AsyncIdQueue.close` loop). Those rejections **deliver asynchronously** on whatever awaits the pull — so the **sync `try/catch` around `peer.close()` catches nothing useful** (`close()` schedules rejections; it does not throw for this case). The `cb1d43f55` fix ("swallow the expected close AbortError") was built on the **false premise that `close()` throws synchronously** — an in-PR wrong-premise fix.

The rejected error is a **fresh `AbortError`** (`"[AsyncIdQueue] ... closed or aborted while waiting for pulling."`), **NOT** the consumer's `signal.reason`. The single teardown-swallow contract, `isAbortReason(err, signal) = signal?.aborted === true && err === signal.reason` (`packages/surface/src/server.ts`), therefore returns **false** on a transport death (the signal is not aborted — nobody aborted it; the transport just dropped). So `iterateUntilAborted` **rethrows** it, and it propagates to the consumer of the parked pull.

### Consumer audit (why it's a rare race, not a deterministic leak)

Every mirror/relay/pump consumer of the source streams **IS guarded** against the rethrown transport-death `AbortError`:
- `guardUpstream` (mirror cells/events/streams): catches, **logs**, ends the loop cleanly.
- `mirrorCollection` per-key value pump (detached `void (async…)()`): `try/catch` logs a non-abort error, IIFE completes — no rejection.
- `relayHoldOpenStream` / `relayFailThroughStream` (surface-remote): catch, treat as a link blip / hold open (they use their own downstream `aborted()`, **not** `isAbortReason` — so a hold-open fix is unaffected by the contract).
- `pumpRemoteSurface` `cursor.next()`, `pollIdentity`, the liveness watchdog: all `.catch`-guarded.

So the intermittent unhandled rejection is a **rare teardown race in a consumer that could not be pinpointed by audit** — it needs the flaky integration test to actually flake (non-deterministic; 194/194 local passes) to capture the exact leaking `.next()` promise.

### What was done (deterministic mechanism repro)

A deterministic, 22ms repro of the **mechanism** at the link level: park an event-iterator `.next()` on a real orpc `stdioLink`, kill the link, capture `unhandledRejection` under explicit microtask control — RED with the exact `[AsyncIdQueue] … closed or aborted` message. (Reverted from the branch; reproduce by parking a `.next()` on the `stdio.test.ts` loopback harness and ending `pair.server.write`.)

### Fix hypotheses (deferred — need a faithful, fix-layer repro + validation)

1. **Broaden the teardown-swallow contract** so `iterateUntilAborted` also treats a transport-death `AbortError` (by class/name) as end-of-life, not just a `signal.reason` match. Low risk to hold-open (relayStream doesn't use `isAbortReason`); the server-side `projectSurface` use of it needs a careful look (masking a real server error). Validate with a mirror-over-real-orpc repro (kill the source mid-parked-pull, assert no unhandled rejection).
2. **Teardown ordering**: abort the consumer signal(s) BEFORE the transport `peer.close()`, so parked pulls reject with the expected `signal.reason` (swallowed) not the orpc `AbortError`. Harder — `peer.close()` is socket-event-driven, decoupled from the pump.

### Why deferred (bounded escape)

Per the review ruling: an **unvalidated shared-teardown change in `@kolu/surface`/`surface-remote`** (drishti-gated) is a guess, not a fix. The faithful fix-layer repro (surface-over-real-orpc-transport) is substantial harness work, and a broad `isAbortReason` change without it risks masking a real error. Tracked here with the full mechanism + repro so the fix can land properly (with the drishti-gate #90 pin-bump) rather than as a hurried guess.

The wrong-premise `cb1d43f55` comment is corrected in-branch to describe the actual (async-rejection) mechanism and reference this issue, so no false-premise code remains.

srid:
Fixes #1719 — the intermittent unhandled `AbortError` on stdio-link teardown (the `padiBinding` "reconnects when padi dies" flake). The availability follow-up W12's own disposition named; recurred on **#1712**, **#1764**, **#1783** (`unit@x86_64-linux`, same `handleTransportClosed` → `peer.close()` `AbortError`).

## Root cause

When the transport dies, `LinkStdioClient.handleTransportClosed()` calls orpc's `peer.close()`. With **no reason**, orpc's `AsyncIdQueue.close()` rejects every PENDING async-iterator pull with a **fresh, anonymous `AbortError`** ("[AsyncIdQueue] Queue[N] … closed or aborted while waiting for pulling."), delivered asynchronously. The one consumer that floats it is `mirrorCollection`'s **per-key value pump** (`mirrorRemoteSurface.ts`): a detached `void (async…)()` IIFE **never joined into the mirror's settle graph**, whose `ctl` is aborted only in the teardown `finally` — *after* `peer.close()` — and whose swallow predicate (`isAbortReason(err, ctl.signal)`) the fresh `AbortError` can never satisfy. `pumpRemoteSurface` builds the mirror with no signal, so nothing aborts the pulls first. `mirror.done` resolves and the pump loop advances while a pull is still parked → the pull rejects with no awaiter → float.

## The fix — two mechanism halves (owned/awaited/typed-cancelled, not swallowed)

**1. `mirrorCollection` OWNS its per-key pumps.** Each pump is tracked; the collection's `finally` aborts the ctls **first** (a still-parked pull then rejects with the swallowed `signal.reason`), **then** `allSettled`s the tracked pumps — so a pump's settle is always observed, regardless of whether the peer or the finally closed first (per the ruling's ordering note).

**2. The stdio link passes a TYPED close reason.** `peer.close({ reason: deadTransportError(SURFACE_STDIO_TRANSPORT_CLOSED, …) })` — so the one thing that can cross the stdio seam at close is that single owned, greppable `ORPCError` (the same non-retriable shape `call()` already throws on a dead link). Makes "an anonymous `AbortError` escapes the seam" unconstructible (AFP P4).

## Half 2 — the padi backstop (honest divergence, backstop not fix)

padi gains a process-level `unhandledRejection` handler that is **loud** (a greppable `padi-unhandled-rejection-boundary` marker + an optional health sink) but **never fatal** — an unidentified future float becomes a diagnosable log line, not a dead workspace daemon. Installed in `bin.ts`'s durable-daemon branch only (not the `--stdio` front, whose stdout is the wire; not `runPadiDaemon`, which tests boot in-process).

**The tension, stated honestly:** this **diverges** from kolu-server's deliberately-fatal `unhandledRejection` policy. A never-fatal global boundary *can* mask a genuinely-new bug. Mitigations: the log is loud + greppable (grep enumerates every float the backstop survived — each a missing source boundary to hunt), and **identified floats are still fixed at source** (halves 1–2 above). The boundary is the net under the fix, not a substitute.

## Evidence — both mechanism pins RED pre-fix (evidence transfers only within its class)

**Pin (i)** — `packages/surface/src/links/stdio.test.ts`: a pull parked at transport close rejects with the typed error, not an anonymous `AbortError`.
- RED on pre-fix: `AssertionError: expected AbortError: [AsyncIdQueue] Queue[1] was c… to be an instance of ORPCError`
- GREEN post-fix.

**Pin (ii)** — `packages/surface/src/mirrorPumpOwnership.test.ts`: `mirror.done` does not resolve while a per-key pump is unsettled.
- RED on pre-fix: `expected false to be true` (`pumpSettled` — the detached pump was abandoned).
- GREEN post-fix.

**Boundary pin** — `packages/padi/src/unhandledRejectionBoundary.test.ts`: an injected float logs loudly with the marker, reaches the health sink, and padi survives (`process.exit` spied, never called). GREEN.

**Suites:** `@kolu/surface` 323/323 · `@kolu/surface-remote` 152/152 · `@kolu/padi` 314/314 · `padiBinding.test.ts` 27/27 (×3 local). `just check` (tsc + biome) clean.

**Flake-class evidence (separate):** flaky-tracker row `reconnects when padi dies` (#1712/#1764/#1783) flips to *fixed* on merge — see the tracker update commit + the two-platform CI runs on the final SHA.

**drishti gate (shared surface):** GREEN. drishti's `packages/app/src/server/router.ts` consumes `pumpRemoteSurface`/`mirrorRemoteSurface`; my diff changes **zero** exported signatures/types, and drishti's full `typecheck` passes with my modified `mirrorRemoteSurface.ts` + `stdio.ts` overlaid onto its hydrated surface (baseline also green). Per `.claude/rules/surface.md`, a purely-internal behaviour change needs no drishti PR; drishti inherits the same float-fix.

## Scope

Framework-touching (`@kolu/surface` links/stdio + mirror) but **not** structural — no boundary moves, no API/type reshape. Direct PR, no `/be` (coordinator-agreed). srid merges.


srid:
## CI — two-platform GREEN on the exact final SHA `652a0e2d0`

`ci::default` seq 2, both lanes passed (every node exitCode 0):
- **x86_64-linux** @ `kolu-ci-1` (pu) — incl. **`ci::unit`** (the #1719 flake's own lane) + `ci::e2e`, all ok.
- **aarch64-darwin** @ `nix-infra@rasam` (swept clear pre-run, 0 orphans) — incl. `ci::unit` + `ci::e2e`, all ok.

Flake-class evidence: the `unit@x86_64-linux` lane that carried the `padiBinding` `AbortError` flake (#1712/#1764/#1783) is green on the fixed SHA. The flaky-tracker row flips to *fixed* on merge.

PR #1795, opened 2026-07-13
refactor(xterm-kit): graduate @kolu/xterm-kit — package + core + mirror anchor (PR 1 of 2)

srid:
**Graduate kolu's xterm machinery into a workspace package — behavior-neutral.** This is **PR 1 of 2** of the ratified [xterm-kit plan](https://kolu.dev/atlas/xterm-kit.html): it moves the *hard volatility* (the pinned `_core` internals, the fail-loud buffer surgery, the mirror anchoring, and the hardened write path) into `@kolu/xterm-kit`, and puts both consumers — the client and kaval — onto it. **Nothing on screen changes.** `Terminal.tsx` still constructs its terminal inline, now from the kit's primitives; the `<Xterm>` component wrapper + a touch-divisor fix are the e2e-gated fast-follow (PR 2, below).

### What's in this PR (all behavior-neutral, CI-verifiable)

- **The package** — `packages/xterm-kit/`, exports `.` (runtime-neutral core) · `./backfill` · `./internals` · `./solid`.
- **Verbatim `git mv`** of the core (history preserved): `scrollbackBackfill` (+ its `CONTRACT PIN` suite), `snapshotBoundary`, `scrollLock`, `xtermInternals`→`internals`, and kaval's `xtermMirrorContract` pin. Zero semantic edits in moved files.
- **`createMirrorAnchor` lifted** out of kaval's `ptyHost.ts` (the `mirrorBaseLine`/`onTrim`/RIS-re-anchor/`reflowEpoch` bookkeeping) — behavior-identical, gated on kaval's existing `ptyHostHistory.test.ts`.
- **Both consumers re-pointed** (import paths only).

### Two corrections to the plan, grounded against the installed system

The plan assumed the daemon would tree-shake. It doesn't — **kaval runs from TS source under tsx (eager ESM)** — so a module the root barrel re-exports is *loaded* by the daemon. Two exports had to move, each reproduced and pinned by `noSolidInDaemon.test.ts` (which walks the root's import graph):

- `createScrollLock` is `solid`-reactive → **`/solid`** (the root stays solid-free so the daemon never vendors a UI framework).
- a static `import { Terminal } from "@xterm/xterm"` **crashes cjs-module-lexer under tsx** (*"does not provide an export named 'Terminal'"*) → the `@xterm/xterm`-constructing backfill write path moved behind **`/backfill`**.

The root barrel is now genuinely what it claims: runtime-neutral and daemon-safe. The closure guard turns any future misclassification into red CI. `PTY_HOST_CONTRACT_VERSION` is untouched; kaval treats the kit as a stable-leaf currency dep (a `/solid` or `/backfill` change never recycles a live PTY).

### The divisor finding (deferred to PR 2, honestly)

Grounding the plan's Mobile section refuted one claim: the tap path's `rect.width / cols` divisor and xterm's font-metric cell mapping **do not "unify by construction"** — they agree only when the two happen to be equal, so collapsing them is a *behavior change*, not a move (a boundary tap can shift a cell under zoom). PR 2 deletes the hand-rolled divisor and routes the tap through xterm's mapping (the one authority that places glyphs), proving *correctness* — a tap hits the visually-tapped cell, normal and zoomed — via standard-suite e2e; it may well be a bug fix. The two "keep both in step" fence comments stay until then.

### PR 2 (immediate fast-follow, e2e-gated)

The `<Xterm>` component cut (`createXtermLifecycle`, `attachWebGL`, `wireTouchTaps`/`wireTouchScroll`, `<Xterm>`) + the divisor fix, gated on standard-suite browser e2e (tap-under-zoom correctness, WebGL context-loss recovery, owner-correct dispose) on the standing two-platform lanes.

### Evidence (neutrality class)

The honest proof for a behavior-neutral refactor is *the same suites stay green* — not a feature clip:

- Moved suites green in their new home: `@kolu/xterm-kit` unit (incl. both `CONTRACT PIN` suites + the closure guard).
- kaval unchanged: `ptyHost*` / `ptyHostHistory` / `socketDaemon` / `buildId.closure` green.
- Client unchanged: full `kolu-client` unit green.
- `noSolidInDaemon` closure guard **seen red then green** (leak a `/solid` re-export → red; revert → green).
- Two-platform CI on the shipping HEAD.

Plan of record + API design: the [xterm-kit note](https://kolu.dev/atlas/xterm-kit.html).


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `214732778a5e`

### Applied (2)
- `lowy-1` internals.ts's "the one _core place / both in tandem" map is now stale — commit `f3084e555`
- `hickey-1` Private `_core.buffers.normal.lines` reach fragmented across three sibling modules; internals.ts still asserts "two" — (uncommitted)

### Agreed — no change (7)
- `lowy-2` /solid is a topic-bundle of 5 helpers the consumer wires by hand (§6.5) (packages/xterm-kit/src/solid/index.ts:8-24)
- `lowy-3` Fail-loud _core.buffers.normal.lines traversal now duplicated across two sibling modules (packages/xterm-kit/src/mirrorAnchor.ts:38-54)
- `lowy-4` Contract test comments reference the deleted `mirrorBaseLine` field (packages/xterm-kit/src/xtermMirrorContract.test.ts:9)
- `hickey-2` createMirrorAnchor(term: object) drops all type contract; inconsistent with the sibling reach's minimal structural type (packages/xterm-kit/src/mirrorAnchor.ts:38,107)
- `hickey-3` "Consumed by both kaval+client" overstated — createMirrorAnchor has a single consumer (packages/xterm-kit/src/index.ts:17-20; packages/xterm-kit/src/mirrorAnchor.ts:8-10)
- `hickey-4` Duplicated `createMediaQuery("(pointer: coarse)")` — isCoarsePointer vs useMobile.isTouch (packages/xterm-kit/src/solid/pointer.ts:10; packages/client/src/useMobile.ts (isTouch))
- `hickey-5` snapToWrapHead co-located in mirrorAnchor.ts is a distinct concern from absolute-line anchoring (packages/xterm-kit/src/mirrorAnchor.ts:62-68)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `214732778a5e`

### Round 1

**codex** — approved: `false`

The mirror-anchor extraction itself is behaviorally faithful: RIS ordering, trim subscription replacement, stable disposal, write-callback position, and width-only epoch bumps all match the original implementation, and the moved core files show no semantic drift. However, the branch currently breaks Nix/package integration and padi's daemon-closure assertion. There are also correctness, boundary, guard-test, dependency, and documentation issues to resolve before approval.

Findings:
- `F1` · blocking · open — The explicit Nix source fileset omits the new `packages/xterm-kit` workspace package even though it declares `typecheck` and is now required by client and kaval. `packages/common/src/nixFileset.test.ts` will report it missing, and the production source cannot resolve that workspace dependency. (default.nix:64)
- `F2` · blocking · open — Padi's eagerly traversed kaval library closure now reaches `@kolu/xterm-kit`, but padi neither hashes nor allows that package. The unexpected-external assertion will fail, and padi's build identity no longer matches its declared closure. This is separate from the deliberate decision to treat xterm-kit as a stable leaf for kaval's own PTY-costing currency key. (packages/padi/src/buildId.closure.test.ts:43)
- `F3` · major · open — `createSnapshotBoundary` duplicates the authoritative `frame.kind` discriminant and is wrong for overflow recovery. `terminalAttach` may yield another `snapshot` mid-stream, but the positional gate is re-armed only on retry/outer reattach, so Terminal.tsx reports that overflow snapshot as live output. It also encodes attach/reconnect/activity policy despite xterm-kit claiming not to know wire frames or PTYs. (packages/xterm-kit/src/snapshotBoundary.ts:32)
- `F4` · minor · open — The daemon-closure guard silently ignores unresolved relative imports and resolves only `<specifier>.ts`. A future root re-export of `./solid`, a `.tsx` component, or a directory index would be skipped, allowing the exact solid/browser dependency regression this test claims to prevent. (packages/xterm-kit/src/noSolidInDaemon.test.ts:40)
- `F5` · minor · open — The canonical architecture package map was not updated for the new volatility package, contrary to the repository's structural-change documentation rule. The map therefore remains stale after this extraction. (website/src/content/docs/architecture.mdx:392)
- `F6` · minor · open — The Atlas documentation did not receive a complete current-state sweep. It still describes only root plus `/solid` despite the deliberate four-entrypoint split, presents PR-2-only APIs as currently importable at lines 63–77 and 293–301, and cites the removed client backfill path at lines 147 and 423. `scrollback-backfill.mdx:247` repeats that removed path, while `electricity.mdx:90` still calls the package "not-yet-built." This does not require implementing PR 2; the current and planned surfaces simply need to be distinguished accurately. (docs/atlas/src/content/atlas/xterm-kit.mdx:38)
- `F7` · nit · open — The package declares resize-observer and seven xterm addons as production dependencies even though current xterm-kit production code imports none of them; they belong to the deliberately deferred PR 2 component. `addon-serialize` is currently test-only, and `vite-plugin-solid` is also unused in this PR. (packages/xterm-kit/package.json:29)
- `F8` · nit · open — Several moved comments still name removed ownership or symbols: `xtermInternals.ts`, `xtermInternals.test.ts`, "kaval's xtermMirrorContract," and the old `isTouch()` guard. These undermine the package's new single-source-of-truth narrative. (packages/xterm-kit/src/scrollbackBackfill.ts:15)


**claude** — Fixed the two blocking integration gaps (nix fileset, padi closure allow-list), hardened the daemon-closure guard, trimmed the PR-1 manifest, corrected the doc/comment staleness, and disputed the one finding that asks for a behavior change to a verbatim-moved file.

- `F1` **fixed** — Added `./packages/xterm-kit` to `default.nix`'s typecheck `src` fileset (default.nix:113). `packages/common/src/nixFileset.test.ts` now passes (was the exact miss it guards against). Verified the pinned `pnpmDeps` hash still resolves (`nix build .#pnpmDeps --no-link` green): my only lockfile delta removes workspace-dep *declarations* for packages that remain present via other workspace members (client), so the fetched store — hence the hash — is unchanged. The heavier `--rebuild` freshness re-execution is the CI gate; the pin is valid.
- `F2` **fixed** — Added `@kolu/xterm-kit` to padi's `ALLOWED_EXTERNAL` (packages/padi/src/buildId.closure.test.ts) with a comment mirroring kaval's rationale: padi reaches xterm-kit ONLY transitively through kaval's embedded library (`ptyHost` → the runtime-neutral core), and the anchor's daemon-relevant behavioral surface (the absolute mirror-line coordinates) is already covered by `PTY_HOST_CONTRACT_VERSION` inside kaval's hashed closure that padi embeds. Treating it as a stable leaf (not a hashed root) is the consistent choice — same as `@kolu/surface`; hashing it would flip padi's key on browser-only `/solid` changes it never loads. `padi/buildId.closure.test.ts` + `kaval/buildId.closure.test.ts` both green with PADI_SOCKET unset.
- `F3` **disputed** — `snapshotBoundary.ts` is a pure verbatim git-mv (0-line content diff) and behavior-neutrality is this PR's explicit headline; codex itself concedes "if strict behavior-neutrality precludes that fix here…". (a) Changing `noteOutput` to gate on `frame.kind === "delta"` is a behavior change to pre-existing shipping code, out of scope. Also it isn't clearly a bug: a mid-stream overflow re-attach delivers a `snapshot` frame precisely BECAUSE the PTY produced heavy output, so lighting the activity dot there is defensible, not a "lie." (b) The boundary claim is factually wrong about the code: `snapshotBoundary.ts` (packages/xterm-kit/src/snapshotBoundary.ts:32-46) contains NO frame type, NO PTY type, NO reconnect logic — it is a two-method boolean toggle (`isLiveDelta`/`armSnapshot`). The wire-frame discriminant (`frame.kind === "snapshot"`, the leading-`TERMINAL_RESET` check) lives entirely in the consumer, Terminal.tsx:868-881. The docstring names the attach stream as motivating context, not as a code dependency — the primitive works on any "first item is special" stream. So it does not "duplicate the authoritative discriminant." If the team later wants a frame-kind-gated activity light, that is a deliberate behavior change best tracked with PR 2, where `<Xterm>` graduates and owns the attach loop.
- `F4` **fixed** — Hardened the daemon-closure guard resolver in `noSolidInDaemon.test.ts`: it now resolves `.ts`, `.tsx`, `index.ts`, `index.tsx` (the same order padi/kaval's walkers use) and THROWS on an unresolved relative source edge instead of silently returning (the old `readFileSync` catch). A future root re-export of `./solid` (a directory barrel) or a `.tsx` component can no longer slip the guard — the regression the test exists to catch. 74/74 xterm-kit tests green.
- `F5` **fixed** — Added a "Terminal machinery" entry to the canonical package map in `website/src/content/docs/architecture.mdx`, describing xterm-kit's runtime-neutral core + `/solid` adapter and the client/kaval two-consumer shape.
- `F6` **fixed** — Atlas sweep: (1) `xterm-kit.mdx` now lists all four entrypoints (`.`/`/backfill`/`/internals`/`/solid`) instead of only root+`/solid`; (2) the `<Xterm>` example and the three lifecycle/WebGL imports in the `/solid` block are labeled explicitly **PR 2 target API**, and the block now also lists the `enableSoftKeyboardInput`/`isCoarsePointer` that actually ship; (3) the two stale `packages/client/src/terminal/scrollbackBackfill.ts` cites repointed to `packages/xterm-kit/src/…`; (4) `scrollback-backfill.mdx` module path annotated with the verbatim graduation; (5) `electricity.mdx` "not-yet-built" reworded to past tense ("still unbuilt at the framework audit"). Ran `just atlas::build`; `dist/` regenerated and staged in the same commit (electricity/scrollback-backfill/xterm-kit .html).
- `F7` **fixed** — Trimmed `packages/xterm-kit/package.json` to PR-1's real deps: removed seven unused `@xterm/addon-*` UI addons + `@solid-primitives/resize-observer` (all PR-2 component deps, imported by nothing in PR-1 src), moved `@xterm/addon-serialize` to devDependencies (test-only — used only in the two contract-pin `*.test.ts`), and removed the unused `vite-plugin-solid` devDep (no `.tsx` in the package; vitest.config uses aliases, not the plugin). `pnpm install` synced the lockfile; typecheck + 74 unit tests green.
- `F8` **fixed** — Repointed stale symbol names: `xtermInternals.ts`→`internals.ts` (scrollbackBackfill.ts:15, solid/renderRecovery.ts), `xtermInternals.test.ts`→`internals.test.ts` (internals.ts), "kaval's xtermMirrorContract.test.ts"→"the headless xtermMirrorContract.test.ts" (scrollbackBackfill.ts), the old `isTouch()` guard→`isCoarsePointer()` (solid/softKeyboardInput.ts), and "the client's `scrollbackBackfill.test.ts`/`@xterm/xterm` beta"→sibling wording (xtermMirrorContract.test.ts, both now in-package).

Verification: `just check` (typecheck + biome) green; `just fmt` applied; xterm-kit 74/74, padi + kaval closure + common nixFileset green (with PADI_SOCKET unset — one unrelated env-sensitive padi test, `spawnInput.test.ts`, fails only because this live box exports PADI_SOCKET). `nix build .#pnpmDeps` green.

commit: `07008309f0e77148448ed81609d4a44eef82ea4f`


### Round 2

**codex** — approved: `true`

All eight findings are resolved. The integration filesets and closure classification are corrected, the daemon guard is hardened for the identified source-resolution gaps, package/docs/manifests are synchronized, and the round introduced no new regression. I also accept CLAUDE's F3 reasoning: the positional boundary intentionally distinguishes replayed initial/retry snapshots from a live mid-stream overflow snapshot, so replacing it with a frame-kind check would change defensible existing behavior.

Findings:
- `F1` · blocking · resolved — The Nix typecheck source fileset omitted the new xterm-kit workspace package. (default.nix:114)
- `F2` · blocking · resolved — Padi's closure policy did not classify the newly reachable @kolu/xterm-kit dependency. (packages/padi/src/buildId.closure.test.ts:109)
- `F3` · major · resolved — The positional snapshot boundary appeared to duplicate the frame discriminator and classify overflow snapshots as live activity. (packages/xterm-kit/src/snapshotBoundary.ts:32)
- `F4` · minor · resolved — The daemon-closure guard only followed simple .ts edges and silently ignored unresolved relative source imports. (packages/xterm-kit/src/noSolidInDaemon.test.ts:31)
- `F5` · minor · resolved — The canonical architecture package map omitted the new terminal-machinery package and its consumers. (website/src/content/docs/architecture.mdx:407)
- `F6` · minor · resolved — The Atlas material had an incomplete entrypoint inventory, presented PR-2 APIs as shipped, retained graduated paths, and contained stale framework-audit wording. (docs/atlas/src/content/atlas/xterm-kit.mdx:38)
- `F7` · nit · resolved — The new package manifest included unused PR-2 runtime dependencies and misplaced a test-only addon. (packages/xterm-kit/package.json:22)
- `F8` · nit · resolved — Moved code retained stale filenames, ownership descriptions, and helper names in comments. (packages/xterm-kit/src/scrollbackBackfill.ts:15)

_codex on the rebuttal:_ F3 — conceded. `createSnapshotBoundary` is a byte-identical move and contains no PTY frame type, reconnect implementation, or wire discriminator. More importantly, its positional semantics are defensible: construction or `armSnapshot()` suppresses the replayed initial/retry snapshot, while a mid-stream overflow snapshot arrives during an already-live stream and represents output activity. Gating on `frame.kind === "delta"` would suppress that activity and therefore impose a behavior change rather than repair an unambiguous defect. The primitive remains sufficiently generic, while `Terminal.tsx` retains the actual frame-kind and reset handling. F3 is resolved.

srid:

Two passes over `git diff 214732778...HEAD` (the xterm-kit graduation, PR 1 of 2), `--no-elegance` (the simplify track already ran).

### Pass 2 — fact-check ✅ the headline claim holds

Diffed the new `mirrorAnchor.ts` **line-by-line against the pre-diff `ptyHost.ts`** (`NormalLinesRef`/`normalLinesOf`, `snapToWrapHead`, the `trimHandler`/`baseLine` accumulation, the RIS-swap `reanchorIfReset`, the width-only `bumpReflow` gate) and every re-pointed call site: **no behavioral drift** — same order of operations, same guards, same "only grows" invariants. The other moved files are verbatim or comment-only (confirmed by old-path/new-path diff, not the rename-blind one).

| finding | action |
| --- | --- |
| `ptyHost.ts` comment named the removed `mirrorBaseLine` field (replaced by `entry.anchor.baseLine()`) | **fixed** — `ed387d20a` |

### Pass 1 — rule checklist (3 findings; 30+ rules checked clean)

| rule | finding | action |
| --- | --- | --- |
| `prefer-focused-library` | the closure guard hand-rolled a per-line import regex; the sibling `buildId.closure.test.ts` in kaval/padi (which its own comment cites as kinship) parse the real AST with `@babel/parser` | **fixed** — `269ff8b39` (switched to the same `@babel/parser` AST walk; teeth re-verified red-then-green) |
| `comment-the-non-obvious` | README `/internals` heading claimed "the single door to `_core.*`", but mirrorAnchor + the backfill surgery also reach `_core` (fail-loud) | **fixed** — `d31bf5ea9` |
| `dry-rule-of-three` | `isCoarsePointer` duplicates `useMobile.isTouch`'s `(pointer: coarse)` query | **skipped (deliberate)** — a kit must not back-depend on the client's layout module; one OS fact, two independent readers. Already adjudicated by the lens debate (`hickey-4`, agreed no-change). |

Clean on: invalid-states-unrepresentable, no-silent-error-swallowing, caught-error-must-not-collapse, no-unbounded-growth, no-untyped-escape-hatches, no-re-export-bridge-modules (documented boundary), new-package-has-readme, and ~20 more inapplicable rules.

srid:
## Evidence

This is a **behavior-neutral** graduation (move + import re-point + one behavior-identical lift), so the honest evidence is *the same suites stay green* and the lift provably doesn't drift — not a feature clip. No UI changes in PR 1 (the `<Xterm>` wrapper + touch-divisor fix are the e2e-gated PR 2).

**Neutrality proof:**

- **The `createMirrorAnchor` lift is behavior-identical** — code-police Pass 2 diffed `mirrorAnchor.ts` **line-by-line against the pre-diff `ptyHost.ts`** (the RIS-swap `reanchorIfReset` ordering, `baseLine += oldLength` before the swap, dispose-then-resubscribe the trim pin, the F5 single-stable-disposable, width-only `bumpReflow`) and every re-pointed call site: **no drift** — same order of operations, same guards, same invariants. kaval's own oracle (`ptyHostHistory.test.ts` — eviction, RIS, history paging) stays green untouched.
- **Moved suites green in their new home** — `@kolu/xterm-kit` unit (both `CONTRACT PIN` suites — `@xterm/xterm` browser + `@xterm/headless` twin — plus `snapshotBoundary`, `scrollLock`, `renderRecovery`, `internals`).
- **The daemon-closure guard was seen red-then-green** — leak a `/solid` re-export into the root barrel → `noSolidInDaemon.test.ts` reds (`Found: [solid-js]`); revert → green. The guard parses the real AST (`@babel/parser`, like the sibling kaval/padi closure walkers). It reproduces the two grounded facts this split exists for: a static `@xterm/xterm` named import crashes cjs-module-lexer under kaval's tsx source-run, and any re-exported `solid-js` would load into the daemon.
- **kaval + client suites unchanged** — kaval (`ptyHost` / `ptyHostHistory` / `socketDaemon` / `buildId.closure`, incl. padi's) and full `kolu-client` unit, all green; `PTY_HOST_CONTRACT_VERSION` untouched.

**Two-platform CI — `passed` on the shipping HEAD** (linux on a leased pool box, darwin on rasam): every node `ok` on both platforms — `ci::e2e`, `ci::unit`, `ci::nix`, `ci::pnpm-hash-fresh`, `ci::atlas-sync`, `ci::smoke`, `ci::fmt`, `ci::biome`, `ci::home-manager`, the surface builds.

> One `ci::e2e@aarch64-darwin` flake en route — the `@kaval-restart` "warming canvas" timing scenario (`kaval-daemon.feature:94`) under rasam eval-cache contention; the **identical scenario passed on the linux lane same-SHA**, and this PR's only kaval touch adds the import-free `mirrorAnchor` module (negligible startup cost). Cleared on a same-SHA two-platform rerun and [logged in the flaky-test tracker](https://kolu.dev/atlas/flaky-test-tracker.html).

PR #1796, opened 2026-07-13
docs(skills): kolu — provisioning a worktree'd agent with padi-tui create

srid:
New section in the kolu skill, between the done-signal section and the two-done-signals comparison: **Provisioning a worktree'd agent — `padi-tui create`**. Field lessons from provisioning the SurfaceRuntime (PR 1) agent terminal:

- `padi-tui create --repo <abs> --worktree <branch> -- claude` is the one-command provision for a kolu-owned, padi-tracked agent terminal — but **fast-forward the base repo first**: the worktree cuts from the current checkout and a stale master seeds an old tree with no error.
- `create` returning is **not** agent-ready: the fresh worktree's first env build takes minutes and a first-run agent can sit on one-time dialogs (MCP selection, trust). Every boot step is driven by `snapshot`, never sleep-and-hope.
- Permission mode is set deliberately and **verified from the footer** after each `Shift-Tab` (manual → accept edits → plan → auto); unattended fleet agents run **auto mode** (per srid's ruling — not bypass).
- Restarting an agent CLI in place: typed text becomes a prompt and `C-c` doesn't reliably quit — `/exit`, wait for the shell in the snapshot, relaunch.

All three skill copies updated identically (`.claude/`, `.agents/`, `agents/.apm/`).

### DIFF
--- a/agents/.apm/skills/kolu/SKILL.md
+++ b/agents/.apm/skills/kolu/SKILL.md
@@ -193,6 +193,36 @@ returns, **`snapshot --viewport` and read** what's on screen before responding.
 > }
 > ```
 
+## Provisioning a worktree'd agent — `padi-tui create`
+
+When the terminal should be a **kolu-owned** workspace — visible on the canvas,
+tracked by padi's agent sensors so `padi-tui wait` works against it — provision
+it with `padi-tui create` instead of raw `kaval-tui create`:
+
+```sh
+git -C /abs/path/to/repo pull --ff-only     # the worktree is cut from the repo's CURRENT checkout
+padi-tui create --repo /abs/path/to/repo --worktree my-branch -- claude
+# prints the new terminal's id; padi owns it and it appears on the canvas
+```
+
+- **Fast-forward the base repo first.** `--worktree` branches from the repo's
+  checkout as it stands — a stale default branch silently seeds the agent an old
+  tree, and nothing errors.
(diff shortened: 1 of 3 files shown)


agents/.apm/skills/kolu/SKILL.md
```diff
@@ -193,6 +193,36 @@ returns, **`snapshot --viewport` and read** what's on screen before responding.
...
 
+## Provisioning a worktree'd agent — `padi-tui create`
+
+When the terminal should be a **kolu-owned** workspace — visible on the canvas,
+tracked by padi's agent sensors so `padi-tui wait` works against it — provision
+it with `padi-tui create` instead of raw `kaval-tui create`:
+
+```sh
+git -C /abs/path/to/repo pull --ff-only     # the worktree is cut from the repo's CURRENT checkout
+padi-tui create --repo /abs/path/to/repo --worktree my-branch -- claude
```
srid: The skill shouldn't hardcode the coding agent. Instead, it should by default use the same agent as orchestrator. Unless the user gives it in the prompt.

srid: Fixed in c5b5a2789 — the example now spawns `<agent>` with an explicit rule: it defaults to the same agent CLI the orchestrator itself runs as, and a user-named agent in the prompt wins. The Claude-Code-specific mechanics (Shift-Tab cycle, /exit) are now labeled as that agent's instance of the pattern rather than the pattern itself.

PR #1804, opened 2026-07-13
feat(surface-map): failure is total, schema-valid domain data — no fabricated cause (PR4)

srid:
## Failure is domain data, never fabricated — PR 4

Part of the [surface-runtime-boundary](https://kolu.dev/atlas/surface-runtime-boundary.html) consolidation (row 4). A failed map entry could previously read a **cause the framework invented**: the map's TypeScript cause narrowed while its wire schema accepted any `z.string()`, and three sites filled an absent cause with `"other"`. This PR makes the failure a **total, schema-valid domain value** and deletes the fabrication.

### What changes

- **`defineSurfaceMap` takes a real `failure` schema** (object argument): `defineSurfaceMap({ key, entry, codec, failure })`. `Failure` is inferred from the schema.
- **The `failed` arm carries `{ failure }`** — a value validated against the domain schema; the loose `cause: z.string()` on the wire is gone. Padi's `PadiEntryFailure` is a discriminated union over a structural `cause`, each arm with its human `reason`, the skew arm with a typed `running`/`expected` pair.
- **`serveHostMap`'s classifier is the required, total `failureOf`** (`Failure | null`). `null` is single-meaning — "not a standing failure, keep **warming**". The `failed` arm of `EntryConnectionState` *requires* a failure at the **type** level (lens debate), so the fail-fast lives at the `serveHostMap.resolve` seam, naming the true producer.
- **No framework fallback cause anywhere.** The three `"other"` sites die. An unclassifiable *failed* state fails loud — `UnclassifiedHostFailureError` / `UnclassifiedHostSessionError` (named, typed, greppable) — guarded at the shared `fire()` funnel so both the status-transition and membership emission paths crash identically rather than one degrading to stale. `EntryFailedCause` drops `"other"` and is derived from `PadiEntryFailureSchema`.

### The negative properties (landed as pins)

- A map member **cannot enter `failed` without a schema-valid domain failure** — enforced by the **type** (a compile-fail `test-d` pin) plus a `serveHostMap`-level throw test.
- **No framework fallback cause exists** — the three `"other"` sites are deleted; the warming-vs-failed split is failure-**presence**.
- **Per-entry readiness stays per entry** — a failed entry never collapses an app-wide gate.

### Behavior-neutral on screen — with two reachable give-up cases stated honestly

Chips/dots/labels/tooltips (kolu **and** all of drishti) are **cause-blind** → byte-identical under any cause remap. The only cause-sensitive pixel is kolu's host-down **card**. Two terminal-give-up producers are named distinctly:

- A **remote** give-up → `link-failed` ("Can't reach this host") — already how it classified on master (its convergence sets the detail).
- A **local** give-up (the padi couldn't start on this machine — no convergence channel) → `local-start-failed`, a **distinct named producer** discriminated off `session.provisions`. Its card **reuses master's pre-PR4 catch-all copy verbatim**, so this reachable case is **byte-identical** on screen. (Collapsing it into `link-failed` was rejected as `"other"` wearing a better name.)

The one state that *does* change: the believed-unreachable unknown-host race (`has()` true, `getSession()` undefined) goes from a fabricated `"other"` card to **fail-loud** — which is the point.

**Evidence:** the test-based neutrality proof is the evidence (same class ruling as PR 1) — for a data-shape refactor whose claim is byte-identical rendering, the passing presentation tests ARE the evidence; a screenshot of an identical card is theater.

### Review gauntlet

architecture-first-principles (code clean; 2 ref-page fixes) → lens-debate (consensus; the `failed`-requires-`failure` type tightening + fail-fast at the resolve seam) → codex-debate at `xhigh` (consensus; caught a real crash bug — a local terminal-failed would have thrown `UnclassifiedHostFailureError` since the local arm's `entryFailedDetail()` is always null — plus the swallowed fail-fast) → simplify → code-police (guard-at-the-funnel, the `provisions` discriminant, `buildTestMap` DRY, missing throw tests).

### Design philosophy

- **Fail fast / no fallbacks** — the whole PR; the `"other"` catch-all *is* the degradation this removes, and an unclassifiable failure crashes loudly at the shared funnel.
- **Reuse the source of truth** — `EntryFailedCause` derived from `PadiEntryFailureSchema`; the wire validates the domain schema.
- **Volatility boundary** — `@kolu/surface-map` stays domain-neutral (carries + validates, never enumerates); the vocabulary lives in padi.

### Seam coordination

Scoped to `@kolu/surface-map` (+ `serveHostMap`) and the padi/kolu consumers. Does **not** touch `@kolu/surface` core (PR 1) and does **not** absorb PR 3's `membershipId`. Paired drishti PR: **srid/drishti#96**, repinned to this HEAD.


srid:
Paired drishti PR (the surface.md gate): srid/drishti#96 — repins kolu to this branch and updates drishti's map consumers. CI green there against the final kolu HEAD is the gate.

srid:

✅ **Consensus** after 4 round(s) · lowy + hickey · base `6f785fcfa981`

### Applied (7)
- `lowy-1` failed arm makes domain failure optional — illegal state enforced by a runtime throw, not the type — commit `f4760d380`
- `lowy-2` EntryConnectionState.reason is now vestigial and double-plumbed with failure.reason — commit `f4760d380`
- `lowy-4` failureOf's state param typed as the whole SessionState union, forcing an `as { error }` cast at every classifier — commit `ec6ce3cd5`
- `hickey-1` `SurfaceMap.failureSchema` is a dead fourth copy of the failure schema — commit `64dd9e49b`
- `hickey-2` `failed` connection-state arm makes `failure` optional — an illegal state guarded by a runtime throw — commit `f4760d380`
- `hickey-3` Skew version-pair fields fragmented across a required interface and an optional schema, bridged by a lying cast — commit `5d17312f0`
- `hickey-5` `entryStatusSchema(failure)` built twice per map — commit `a59c16754`

### Agreed — no change (4)
- `lowy-3` statusTitle exposes the wires — re-imposes a 'reason' contract the receptacle dropped (packages/client/src/host/hostChipTone.ts:63)
- `hickey-4` `reason` on EntryConnectionState down arms is vestigial and duplicates SessionState.error (packages/surface-map/src/server.ts:98-99)
- `hickey-6` `PadiEntryFailure` and `PadiEntryFailedDetail` are two types for one failure, differing only by `reason` (packages/common/src/surfacesWithPadi.ts:121 (PadiEntryFailure) vs packages/server/src/padi/padiSession.ts:35 (PadiEntryFailedDetail))
- `hickey-7` Two near-identical 'unclassified failure' error classes across packages (packages/surface-map/src/server.ts:~180 (UnclassifiedEntryFailureError) and packages/surface-remote/src/serveHostMap.ts:~51 (UnclassifiedHostSessionError))

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 round(s) · codex reviewed at `xhigh` reasoning effort · base `6f785fcfa981`

### Round 1

**codex** — approved: `false`

The new schema-backed failure model is structurally sound, but two major runtime paths break its contract: local terminal failures are misclassified as unclassified, and the promised fail-fast errors are swallowed, potentially leaving clients on a stale connected/warming status forever. I also found public-export and documentation drift. All five affected package typechecks passed; unit tests could not start in the read-only sandbox because Vitest/Vite attempted temporary writes.

Findings:
- `F1` · major · open — `failureOf` assumes every terminal session has a non-null `entryFailedDetail()`, but the local Padi arm always returns null (`padiBinding.ts:633`). Local sessions can still reach terminal `failed` after repeated bounded failures, such as connect-watchdog timeouts or pre-connected endpoint deaths. The local host therefore raises `UnclassifiedHostFailureError` instead of publishing the precise existing `link-failed` failure. (packages/server/src/index.ts:508)
- `F2` · major · open — The `session.onState` callback catches every republish exception and only logs it. This consumes the new `UnclassifiedHostFailureError` thrown at line 310. Because `latestState` is updated first and terminal `failed` may be the final frame, the published entry can remain indefinitely at its previous status—even `connected`—instead of failing fast. Membership-listener throws are also caught by `hostFanout.ts`, so a synchronous rethrow alone is insufficient. (packages/surface-remote/src/serveHostMap.ts:217)
- `F3` · minor · open — `UnclassifiedHostFailureError` is documented and exported from its implementation module, but omitted from the package root while its sibling `UnclassifiedHostSessionError` is re-exported. Since package exports expose no `serveHostMap` subpath, consumers cannot import or use the advertised error class. (packages/surface-remote/src/index.ts:85)
- `F4` · minor · open — The generic reference says `failure.reason` is always a human string, but `defineSurfaceMap` accepts any `ZodType<Failure>`; valid maps can use a string, a code-only object, or another shape with no `reason`. This overstates the framework contract. (website/src/content/surface/ref-surface-map.mdx:165)
- `F5` · minor · open — The current PR4 plan-of-record still declares `failureOf(...): PadiEntryFailure` as exhaustively non-null, contradicting the implemented and deliberate `PadiEntryFailure | null` contract where null means transient/keep-warming. (docs/atlas/src/content/atlas/surface-runtime-boundary.mdx:89)
- `F6` · minor · open — The teaching note says it documents what shipped, but still teaches `defineSurfaceMap(surface)`, flat `EntryStatus<Cause>` with `{ reason, cause }`, a loose failure arm, and `causeFor`. Those call shapes and semantics were removed by this change. (docs/atlas/src/content/atlas/surface-map-101.mdx:24)
- `F7` · nit · open — The active regression test still names and explains the deleted `causeFor → "other" → warming` mechanism. The assertion remains useful, but its title and rationale now describe behavior that no longer exists. (packages/server/src/padi/remotePadiBinding.test.ts:928)


**claude** — Fixed the real local-terminal-failure gap at its true producer (F1), exported the two advertised symbols, and corrected four doc/test drifts; disputed only the demand to re-architect the deliberate per-member catch (F2).

- `F1` **fixed** — Confirmed real. The LOCAL padi arm's `entryFailedDetail()` is always `null` (`padiBinding.ts:633`), yet a local session CAN reach terminal `failed` with `cause:"remote"` — a wedged connect handshake (`session.ts:805-810`), a spawn-error (`:811`), or an endpoint/agent death before connect (`:822-836`) each accrue the bounded give-up budget (`scheduleReconnect` → `setDown("failed",…)`, `:753-762`). That drove `failureOf` → `null` on a `failed` state → `UnclassifiedHostFailureError` (`serveHostMap.ts:309-310`). The author's own comment even asserted "a terminal failed session always classifies (its convergence is set to link-failed)" — true for the remote arm, false for the local one. Fix: pulled the classifier out of the composition root into a tested pure `padiFailureOf(detail, state)` in `padiSession.ts` (single source of truth) and floored a terminal give-up with no finer detail to `{cause:"link-failed", reason:state.error}` — which is exactly what `link-failed` means in the schema ("the transport gave up", `surfacesWithPadi.ts:104`). `index.ts:498-510` now calls it; a `disconnected` frame still returns `null` (keep-warming). Added `padiSession.test.ts` (4 cases incl. the local-give-up regression). Needed `DownSessionState` nameable, so exported it from `@kolu/surface-remote` (it already appears in the public `ServeHostMapOptions.failureOf` signature, so this only makes nameable what was already structurally public).

- `F2` **partial** — The harmful outcome (a client stuck on a stale status) is removed by fixing F1 at its producer: after F1 no *reachable* stable state throws in `resolve`. I dispute the demand to re-route invariant throws through a "fatal channel". The catch at `serveHostMap.ts:217-231` is a documented, deliberate decision (d3): letting a per-member republish throw propagate would end the `onState` consumer loop and FREEZE every member's `EntryStatus` while the sibling connection-cell consumer keeps advancing (the green-chip-frozen divergence) — strictly worse than a loud `console.error` + live siblings. The remaining throws (`UnclassifiedHostSessionError` at `:285`, the provisioning belt at `:330`) are "impossible-state" invariants, not reachable stable states; crashing an entire multi-host server (taking down every host's terminals) because one map member hit an impossible projection is not an improvement over surfacing it loudly and keeping the healthy members live. Note the error does not collapse to empty/default — it logs loudly and retains the last real value — so it honors "a caught error must surface." codex's "even connected forever" is also slightly off: the last *successfully published* frame before a `failed` throw is the retrying `disconnected`→warming, not `connected`.

- `F3` **fixed** — Agreed. `UnclassifiedHostFailureError` was `export class` in `serveHostMap.ts:71`, thrown at `:310`, and documented in `ref-surface-remote.mdx:90`, but omitted from the package root while its sibling was re-exported. Added it beside `UnclassifiedHostSessionError` in `packages/surface-remote/src/index.ts`.

- `F4` **fixed** — Agreed (minor overstatement). `ref-surface-map.mdx:165` stated `failure.reason` is a human string as a framework guarantee, but `defineSurfaceMap` accepts any `ZodType<Failure>`. Reworded to say the framework fixes only the *shape* (`failure` is a schema-validated sibling of the `failed` arm), and that `reason` is padi's own convention — a schema with no `reason` is equally well-formed.

- `F5` **fixed** — Agreed. The PR4 plan-of-record (`surface-runtime-boundary.mdx:89`) declared `failureOf(...): PadiEntryFailure` "exhaustive", contradicting the shipped `PadiEntryFailure | null`. Updated the snippet to `| null` and the prose to explain the single-meaning "keep warming" verdict vs the loud throw for an unclassifiable terminal failure. Rebuilt + staged `docs/atlas/dist/`.

- `F6` **fixed** — Agreed. The teaching note (`surface-map-101.mdx`) still taught the pre-PR4 API: `EntryStatus<Cause>`, a flat `{ kind:"failed"; reason; cause }` arm, `EntryStatus<PadiEntryFailedCause>` through "the schema's loose arm", and a `causeFor` hook. Updated the description, the status-sum code block (`{ kind:"failed"; failure: Failure }`), the `Failure` type-parameter prose (schema-validated discriminated union, fail-loud on unclassifiable), and the `serveHostMap` line (`failureOf`). Left the conceptual pedagogy intact. Rebuilt + staged dist.

- `F7` **fixed** — Agreed. `remotePadiBinding.test.ts:928` title/comment still described the deleted `causeFor → "other" → warming` path. The assertion is unchanged (a transient drop still yields `null`); rewrote the title/comment to describe `null` projecting straight through to warming (the single-meaning absent), and cross-referenced the terminal-give-up→`link-failed` case now covered in `padiSession.test.ts`.

Note (ship-phase, not a code change): the two additive `@kolu/surface-remote` exports (`UnclassifiedHostFailureError`, `DownSessionState`) are API-additive and do not break drishti's existing consumer, so they ride the same deferred drishti ship-gate the author already fenced to final post-gauntlet HEAD — not a blocking round-1 finding.

Verification: `just check` green (typecheck all packages + biome lint). Unit tests green — surface-map 38, surface-remote 152, common 43, server padi suites incl. new `padiSession.test.ts` (39 in the two touched files). `just atlas::build` regenerated; dist staged in this commit (a clean tree makes `atlas::check-sync` pass).

commit: `73a02ae4d78c4109eb6a02c5236fc08b6dba5def`


### Round 2

**codex** — approved: `false`

F1, F3, F4, F5, and F7 are adequately fixed, and the new classifier introduces no evident regression. F2 remains open because invariant/classifier errors are still consumed by the per-member callback, leaving the published status stale without an observable failure state. F6 is only partially fixed: the teaching note still shows the obsolete `defineSurfaceMap(surface)` API.

Findings:
- `F1` · major · resolved — A local padi session could reach terminal `failed` with no entry-failure detail, causing classification to throw instead of publishing `link-failed`. The new `padiFailureOf` correctly classifies terminal null-detail states while preserving null for transient disconnections, and `index.ts` now uses it. (packages/server/src/padi/padiSession.ts:63)
- `F2` · major · open — The per-member catch still consumes projection, classifier, and invariant errors. It logs them but leaves `entries` at its last successfully published value; a terminal transition may provide no later frame to retry, while the server otherwise continues as healthy. (packages/surface-remote/src/serveHostMap.ts:217)
- `F3` · minor · resolved — `UnclassifiedHostFailureError` was documented public API but omitted from the package root. It is now re-exported alongside `UnclassifiedHostSessionError`. (packages/surface-remote/src/index.ts:85)
- `F4` · minor · resolved — The generic reference incorrectly presented `failure.reason` as a framework guarantee. It now clearly separates the framework-owned shape from padi's domain-specific convention. (website/src/content/surface/ref-surface-map.mdx:165)
- `F5` · minor · resolved — The plan-of-record showed a non-null `failureOf` contract. Its signature and prose now correctly document the single-meaning null verdict and terminal classification error. (docs/atlas/src/content/atlas/surface-runtime-boundary.mdx:89)
- `F6` · minor · open — Most obsolete failure API material was corrected, but the note still teaches `defineSurfaceMap(surface)`. The current API takes an object containing `key`, `codec`, `entry`, and `failure`, so this remains an inaccurate API lesson. (docs/atlas/src/content/atlas/surface-map-101.mdx:24)
- `F7` · nit · resolved — The test title and comment described the removed `causeFor`/`other` path. They now accurately describe null as the transient keep-warming verdict and point to terminal give-up coverage. (packages/server/src/padi/remotePadiBinding.test.ts:928)

_codex on the rebuttal:_ F2: I concede that F1 removes the concrete local-padi failure that exposed this problem, and that current `makeSession` normally publishes a disconnected/warming frame before terminal failure. I do not concede F2 itself. `serveHostMap` is a public adapter with an injected `failureOf`, and `UnclassifiedHostFailureError` exists specifically because terminal null is a detectable caller/producer defect; that makes the path an intentional runtime guard, not an impossible state. The catch prevents that typed error—and any exception thrown by `failureOf`, schema publication, `linkFor`, or another projection invariant—from reaching the stream owner or process. On terminal failure there may be no subsequent frame, so even the current Padi path can remain permanently warming; other valid `Session` implementations or injected callbacks can leave an older connected status. `console.error` does not fulfill the documented fail-loud contract while health and publication continue normally. Preserving the `onState` loop and surfacing a fatal invariant are compatible: catch synchronously, then report through an owned fatal channel or schedule an out-of-band fatal throw. F2 therefore remains open.


**claude** — Conceded F2 and escalated the swallowed invariant guards to an out-of-band process crash; fixed the last obsolete API call in the teaching note (F6). F1/F3/F4/F5/F7 already resolved last round — untouched.

- `F1` **fixed** (already resolved round 1) — `padiFailureOf` in `packages/server/src/padi/padiSession.ts:63`; no change this round.
- `F2` **fixed** (conceded) — codex is right, and I withdraw my round-1 dispute. The catch at `packages/surface-remote/src/serveHostMap.ts:217` reduced every escaping throw — `UnclassifiedHostFailureError` (producer defect), `UnclassifiedHostSessionError` (impossible member-with-no-session), the provisioning belt `Error` — to a `console.error` while health and publication continued. That is exactly the silently-degrading fallback the design philosophy forbids (`conventions.md` → "a caught error must surface, never collapse to an empty or default state"); a server never watches DevTools, so the fail-loud guards were decorative. My round-1 concern (a synchronous propagate would tear down THIS member's `onState` loop mid-fire — the wrong member for a sibling's defect — and freeze its status while siblings advance) is real but NOT in tension with failing loud: codex's own synthesis resolves both. The fix keeps the `catch` (this frame's loop finishes with `latestState` already cached, so no green-chip-frozen divergence) and then `queueMicrotask(() => { throw ... })` — rethrowing out-of-band so the invariant lands as an uncaught exception that crashes the process right after the frame, instead of a swallowed log. This reuses the codebase's own out-of-band-fatal idiom (`packages/surface/src/links/stdio.ts` history: a `process.nextTick` rethrow that crashes on an unhandled defect). Reference doc updated to match (`website/src/content/surface/ref-surface-remote.mdx:90` — the guards now "crash the process" via the microtask rethrow). No test asserted the old swallow (`serveHostMap.test.ts` grep clean), and an in-process test of an uncaught-exception crash would tear down the vitest worker, so I did not add one; `serveHostMap.test.ts` still green (7). This is a runtime-behavior change to `@kolu/surface-remote`, so it rides the same deferred drishti ship-gate the author already fenced to final post-gauntlet HEAD — it only fires on a genuine producer defect, so drishti's normal operation is unaffected.
- `F3` **fixed** (already resolved round 1) — re-export in `packages/surface-remote/src/index.ts:85`; no change this round.
- `F4` **fixed** (already resolved round 1) — `website/src/content/surface/ref-surface-map.mdx:165`; no change this round.
- `F5` **fixed** (already resolved round 1) — `docs/atlas/src/content/atlas/surface-runtime-boundary.mdx:89`; no change this round.
- `F6` **fixed** (conceded) — codex is right: `docs/atlas/src/content/atlas/surface-map-101.mdx:24` still opened with `defineSurfaceMap(surface)`, but the shipped signature (`packages/surface-map/src/define.ts:353`) takes `{ key, entry, codec, failure }`. Rewrote the line to `defineSurfaceMap({ key, codec, entry, failure })`, explained the ordinary surface goes in through `entry`, and glossed the PR4 `failure` schema (with `Failure` inferred from it). Rebuilt `just atlas build`; `docs/atlas/dist/surface-map-101.html` regenerated and staged.
- `F7` **fixed** (already resolved round 1) — `packages/server/src/padi/remotePadiBinding.test.ts:928`; no change this round.

Verification: `just fmt` clean; `just check` green (typecheck all packages + biome lint, 1309 files, no warnings); `just atlas build` regenerated dist (staged); `serveHostMap.test.ts` 7 passed.

commit: `9140139292213151659acb65d011d5b28e997714`


### Round 3

**codex** — approved: `true`

All seven findings are resolved. The out-of-band rethrow fixes F2 without synchronously tearing down the member's state-consumer loop; kolu-server's uncaught-exception handler then logs fatally and exits. The teaching note now uses the current object-form API, and its generated Atlas output is synchronized. No regression was introduced this round.

Findings:
- `F1` · major · resolved — Local terminal failures without finer detail previously remained unclassified. (packages/server/src/padi/padiSession.ts:63)
- `F2` · major · resolved — Invariant and classifier errors were caught and reduced to logging, allowing stale status to remain while the process appeared healthy. (packages/surface-remote/src/serveHostMap.ts:217)
- `F3` · minor · resolved — `UnclassifiedHostFailureError` was omitted from the package root exports. (packages/surface-remote/src/index.ts:85)
- `F4` · minor · resolved — The generic reference incorrectly presented `failure.reason` as a framework guarantee. (website/src/content/surface/ref-surface-map.mdx:165)
- `F5` · minor · resolved — The plan-of-record documented `failureOf` as non-null. (docs/atlas/src/content/atlas/surface-runtime-boundary.mdx:89)
- `F6` · minor · resolved — The teaching note retained the obsolete `defineSurfaceMap(surface)` call. (docs/atlas/src/content/atlas/surface-map-101.mdx:24)
- `F7` · nit · resolved — The test title and comment described the removed `causeFor`/`other` path. (packages/server/src/padi/remotePadiBinding.test.ts:928)

_codex on the rebuttal:_ F2 accepted: the microtask rethrow makes the projection defect observable and fatal while avoiding a synchronous exception through the wrong member's `onState` loop. In kolu-server, the existing `uncaughtException` handler logs the wrapped error and exits with status 1. F6 accepted: both the Atlas source and generated output now teach the shipped object-form API. No dispute remains.

srid:

Rule-checklist + fact-check passes over the diff. Three findings, all actioned:

- **Fail-loud guarded at one call site, not the funnel** (`serveHostMap.ts`) — the catch-then-async-rethrow lived on the `onState` emission path only; the membership-change path called the same `fire()` bare, so an unclassifiable-failed member would throw *synchronously* out of `pool.subscribe` the moment any *other* host was added/removed. **Fixed** by moving the guard into `fire()` itself (both emission paths now fail loud identically), commit `74b2bdd55`. Added `serveHostMap`-level tests for the `UnclassifiedHostSessionError` (no-session) and `UnclassifiedHostFailureError` (null-on-`failed` crash) seams — the safety property had no direct coverage at the layer it lives in.
- **Unenforced cross-module invariant** (`padiSession.ts`) — `local-start-failed` inferred "local" from `detail === null && phase === failed`, relying on "the remote arm always sets a link-failed detail first." **Fixed** by discriminating off `session.provisions` (the runtime twin of the arm's `Prov`) directly, commit `ee09011c6` — behavior-neutral for reachable states, robust to a subscription-order regression. Also fixed a stale `index.ts` comment (`link-failed` → `local-start-failed`).
- **DRY in the map harness** — 10 test sites re-inlined `defineSurfaceMap({…, failure: testFailureSchema})`. **Fixed** by extracting `buildTestMap`, commit `85e3095e0`.

Fact-check: no silent error-swallowing, no inaccurate fallbacks, no wishful input shapes, no leaks — the diff is precisely a *removal* of the old `"other"` fallback in favor of a total, tested classifier.

PR #1805, opened 2026-07-13
A runtime you can serve — supervised SurfaceRuntime (Surface plan PR 1)

srid:
# A runtime you can serve — supervised `SurfaceRuntime` (Surface plan PR 1)

`implementSurface` stopped one layer too early: it returned an intermediate
`{ router: fragment, ctx }` and every consumer re-learned the one correct oRPC
finalization step, then hand-owned (or floated) the async work the surface spun
up. This PR makes it return a **directly-servable, supervised resource** — the
first move of the [Surface runtime plan](https://kolu.dev/atlas/surface-runtime-boundary.html)
(the "A runtime you can serve — PR 1" section, row 1).

```ts
interface SurfaceRuntime<S extends SurfaceSpec> {
  readonly router: unknown;      // FINAL top-level router — no consumer re-finalizes
  readonly ctx: SurfaceCtx<S>;
  readonly done: Promise<void>;  // rejects on an owned runtime fault; resolves on clean close
  close(): Promise<void>;        // idempotent; releases every owned source
}
```

## What lands

- **The final router.** `implementSurface` / `implementSurfaces` now return the
  FINAL top-level oRPC router. No consumer wraps `implement(contract).router({…})`
  anymore.
- **Supervision.** `done` rejects on an owned fault (a cell connector rejecting)
  instead of floating as an unhandled rejection; `close()` releases every owned
  source — following #1719's ownership doctrine (abort first, then await the
  settle, then dispose). Child failure (cell connectors, reactor/derived
  subscriptions, re-served pumps) reaches `done`, never floats.
- **Async cell connectors** receive an abort signal and MAY return a `Disposer`
  (`connect(cell, { signal }) => void | Disposer | Promise<…>`) — additive, so
  every existing `void`-returning connector keeps working. Reactor `derived.cell`
  and `deriveCell` now RETURN their disposer, so `close()` disposes them (the
  reactor/derived subscriptions join the runtime's ownership — a `close` that
  ignored them would be ceremonial).
- **Two ownership promises, not a mode flag.** The ordinary constructor owns an
  internal `inMemoryChannelByName()`; a consumer serving on a shared, caller-owned
  publisher (whose cross-channel microtask order is load-bearing) reaches for the
  distinct **`implementSurfaceOnPublisher`** / **`implementSurfacesOnPublisher`**.
  The runtime's `close()` releases only what it minted, never the shared publisher.
- **Minimal re-serve supervision.** `reServeSurface` gained a `close()` (aborts
  its pump via a new optional `pumpRemoteSurface` signal, without destroying the
  caller-owned session, then releases its runtime) and its `done` composes the
  pump's settle with an owned runtime fault. No PR-5 reshape (no wire/protocol/
  mirror change) — supervision only.

## Proven by deletion

- **kolu**: the `implement(servedContract)` **surface-finalization** step is
  deleted — the kolu + surfaceApp surface now comes FINAL from
  `implementSurfacesOnPublisher`, so `packages/server/src/surface.ts` no longer
  finalizes a Surface via `t`. `servedContract` itself is **retained**, strictly
  as the app-router *builder*: it binds the raw `server`/`daemon`/`hosts` RPCs and
  re-adapts the re-served `padi` map fragment for the wire matcher (which requires
  a contract that declares `surface.padi`). The campaign's negative property holds
  **verbatim** — *no production consumer imports oRPC `implement` merely to
  finalize a Surface* — because kolu's remaining `implement(servedContract)` is
  the builder, not mere finalization. (First shipped as a full deletion; the
  two-platform e2e caught the resulting wire-matcher 404 — `serveHostMap` returns
  a matcher-meta-less fragment a padi-less builder silently drops — that the
  `directLink`-based `padiBinding` test structurally could not see. Fixed in
  `ab74926bd`, pinned by `router.test.ts`: a `StandardRPCMatcher` assertion that
  the assembled `appRouter` routes `/surface/padi/*`, the matcher-level evidence
  class directLink cannot provide.)
- **drishti**: the fragment wrap in `packages/agent/src/main.ts` and
  `packages/app/src/server/router.ts` deletes — via the paired drishti PR
  (link below), pinned to this PR's final HEAD.
- **odu** (NOT in scope; pins kolu by npins): its two finalization sites —
  `src/runner/runner.ts:339` and `src/coordinator/run.ts:469` — are recorded here
  as a **next-npins-bump obligation** to land in the odu tree.

## Behavior-neutral — per-site fault disposition

The invariant is byte-identical on screen (the e2e suite + `seal.test.ts` pin
it). Capturing floats into `done` removes them from the process-level
`unhandledRejection` path, so each serving site routes `done` into its EXISTING
failure policy — never a new one:

| Serving site | `done` routes to | unchanged because |
| --- | --- | --- |
| kolu-server | `log.fatal` + `process.exit(1)` | kolu's deliberately-fatal policy; mirrors the default padi re-serve's `done`→fatal |
| padi daemon | loud log, never fatal | padi's `unhandledRejection` boundary (#1792) — loud-not-fatal |
| kaval daemon | `log.error` (inert — ptyHost declares no cell connectors) | nothing faults today; observed + close-owned for future-proofing |
| reServeSurface | existing `.done`→retire/fatal observer (`index.ts`) | the pump's own settle still drives retire/fatal |
| projectSurface | pass-through (caller observes) | its consumers (odu) own disposition |

## Negative properties (pinned, not eyeballed)

- No production consumer imports oRPC `implement` merely to finalize a Surface (kolu deletion + drishti pair).
- No serving API accepts private mount material (the constructor split; runtime `router` is opaque `unknown`).
- Child failure reaches `done`; repeated `close()` is harmless — pinned by `packages/surface/src/surfaceRuntimeSupervision.test.ts` (owned fault reaches done; close idempotent + aborts signal + runs disposer).
- Router shape pinned by the rewritten `packages/surface/src/implementSurface.test.ts` (final router, no double-prefix).

## Note on the plural twin

The plan named `implementSurfaceOnPublisher` (singular). The consumed API is
plural — kolu-server (`surface.ts`) and padi (`servePadi`) build on
`implementSurfaces` — so this adds `implementSurfacesOnPublisher` as the faithful
same-doctrine extension (one runtime over N declared surfaces, ownership-promise
per constructor, no mode flag). Coordinator-approved.

## Docs / gates

- Reference pages updated in-PR: `ref-surface.mdx` (SurfaceRuntime, the
  constructor split, the connect signature), `ref-surface-remote.mdx`
  (`reServeSurface.close()`, the pump signal); the surface prose pages
  (`across-the-hosts`, `serve-a-map`) de-stale the fragment/flatten language.
- Verified (per-pass obligation): the `@kolu/surface-daemon` /
  `-supervisor` READMEs' "zero `kolu-*` dependency" claims hold — both depend only
  on scoped `@kolu/*` framework packages, never the unscoped app (`kolu-server`/
  `kolu-common`/`kolu-client`).
- No changelog entry: byte-identical on screen, no user-visible behavior change.
- Paired drishti PR (surface.md gate), pinned to this PR's final HEAD, green on
  both platforms.


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `3f71057225d1`

### Applied (4)
- `lowy-3` SurfaceRuntime and SurfacesRuntime are duplicated interface shapes differing only in the ctx type — commit `f73a71609`
- `hickey-1` reServe `done` complects settle-routing via Promise.race + a parked-forever promise — commit `a4fd8c32d`
- `hickey-2` SurfaceRuntime and SurfacesRuntime duplicate one 'supervised runtime handle' concept — (uncommitted)
- `hickey-3` Teardown doctrine (abort→await settle→dispose, idempotent via `closing ??=`) duplicated across two supervisors — commit `7d6735548`

### Agreed — no change (6)
- `lowy-1` Boundary assessment: supervision receptacle is coherent and correctly placed (electricity) (packages/surface/src/server.ts:1400)
- `lowy-2` superviseSurface receptacle is private; reServeSurface hand-rolls a parallel supervision algebra (packages/surface-remote/src/reServeSurface.ts:540)
- `lowy-4` closeKoluSurface exported with no caller — half the receptacle is wired asymmetrically (packages/server/src/surface.ts:375)
- `lowy-5` SurfaceRuntime.done is a raw floatable Promise — receptacle safety is convention-dependent (packages/surface/src/server.ts:1483)
- `hickey-4` Identical surface-fault log idiom duplicated across kaval and padi daemons (packages/padi/src/daemonMain.ts:293)
- `hickey-5` `closeKoluSurface` exported with no caller (speculative supervision surface) (packages/server/src/surface.ts:378)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 round(s) · codex reviewed at `xhigh` reasoning effort · base `3f71057225d1`
### Round 1

**codex** — approved: `false`

Not approved. The runtime contract is directionally sound, but construction can orphan connectors, teardown can hide faults or leave cleanup blocked, derived subscriptions are not fully joined, and production serving sites do not yet own shutdown. Public guidance also still teaches the removed fragment API. This was a read-only static review; no tests or formatting commands were run. The Drishti pairing and two-platform CI requirements are acknowledged as resolved/deferred ship-phase gates, not code findings.

Findings:
- `F1` · major · open — Disposer failures are silently discarded. `close()` awaits `Promise.allSettled` for every disposer but examines only connector settlement, so `done` resolves even when owned cleanup rejects. This contradicts the documented promise that teardown faults reach `done`. (packages/surface/src/server.ts:1458)
- `F2` · major · open — Construction is not transactional. `walkSurface` starts connectors while later members and surfaces are still being validated, and final router construction happens afterward. If a later dependency is missing or router assembly throws, the constructor exits without returning a runtime, leaving previously started connectors without an abort owner, disposer owner, or reliable rejection observer. (packages/surface/src/server.ts:1728)
- `F3` · major · open — `deriveCell` launches its subscription loop in a detached async IIFE and returns a disposer that only calls `abort()`. The runtime therefore considers the connector settled immediately, and `runtime.close()` can resolve before the upstream iterator's asynchronous teardown or `finally` block completes. This violates the abort-then-observe-settlement ownership contract. (packages/surface/src/project.ts:258)
- `F4` · major · open — Serving sites observe `done` but still do not consistently own shutdown. Padi drops `runtime.close` from `SurfacesServed`; Kolu exports `closeKoluSurface` but no caller invokes it before direct `process.exit`; Kaval fires close with `void` from an abort listener and does not await it before the daemon promise resolves. Kaval also forwards an already-aborted input signal before registering that listener, which can miss close entirely. (packages/padi/src/daemonMain.ts:241, packages/kaval/src/daemonMain.ts:119, packages/server/src/index.ts:198)
- `F5` · major · open — Public documentation still teaches the deleted API. It describes `router` as a fragment that must be wrapped again, shows the removed `channel` dependency on ordinary `implementSurface`, omits `close` from `reServeSurface`, and even re-exports oRPC `implement` to support the obsolete wrapping pattern. Following this guidance now produces type errors or double-prefixed routes and 404s. (packages/surface/src/peer-server.ts:30, packages/surface/src/server.ts:852, agents/.apm/skills/surface/SKILL.md:32, website/src/content/surface/mirror-over-ssh.mdx:64)
- `F6` · minor · open — Teardown uses a global settlement barrier: no disposer runs until every connector settles. One connector that ignores cancellation therefore prevents already-settled, independent sources from releasing their resources indefinitely. (packages/surface/src/server.ts:1456)
- `F7` · minor · open — The new clean supervision-abort path is logged as a failure and ultimately as `session destroyed`, even though `reServeSurface.close()` deliberately preserves the caller-owned session. When the pump is waiting for a client it logs `waiting for next client failed`; after either active or waiting abort it unconditionally claims destruction. (packages/surface-remote/src/hostFanout.ts:203)
- `F8` · minor · open — The new eviction-cleanup contract is not total or regression-tested. `onEvict` is optional despite the function documentation promising it for every dropped value, and `cache.get(host) !== undefined` skips the callback for a valid `Map<K, V>` entry whose value is `undefined`. Existing tests omit the callback entirely, so the production close behavior is unpinned. (packages/server/src/padi/reServeEviction.ts:26)
- `F9` · minor · open — The new one-bit teardown state permits a leaking transition. If standalone `dispose()` runs before `connect()`, `torn` becomes true; a later `connect()` still installs an effect, but its returned teardown permanently no-ops. The cleanup also skips `node.dispose()` if effect disposal throws. (packages/surface/src/reactor.ts:321)
- `F10` · nit · open — The dynamic `cellDeps` shape still declares the old one-argument, no-disposer `connect` signature, then later casts it to `CellConnector`. This duplicates and contradicts the new source of truth at the exact runtime boundary that invokes connectors. (packages/surface/src/server.ts:1570)


**claude** — Fixed the real teardown/construction/lifecycle gaps codex found (F1, F2, F3, F6, F7, F8, F9, F10 and the doc sweep F5), wired padi + kaval shutdown ownership and deleted kolu's dead close export for F4; static gate green + new tests for every behavioral fix.

- `F1` **fixed** — `superviseSurface.close()` now inspects disposer settlement too, not just connector settlement. Each source's per-task rethrows a settle fault OR a dispose fault, and `close` routes the first to `done` via `sealReject` while still always resolving (`packages/surface/src/server.ts:1454-1490`). Added sync- and async-disposer-rejection tests (`surfaceRuntimeSupervision.test.ts` "a SYNC/ASYNC disposer rejection … reaches done").
- `F2` **fixed** — construction is now transactional. `walkSurface` collects deferred connector-START thunks (`SurfaceSourceStart`) instead of firing connectors inline; both constructors build the FINAL router first and only then invoke the thunks (`server.ts` walkSurface `starts`, `implementSurfaceOnPublisher`, `implementSurfacesOnPublisher`). A missing later member / router-assembly throw now returns before any source spins up. Added singular + plural regression tests proving an invalid later member never starts an earlier connector.
- `F3` **fixed** — `deriveCell`'s `connect` now RETAINS the subscribe loop as a `task` and returns an ASYNC disposer that aborts and `await`s it, so `runtime.close()` cannot resolve before the upstream generator's `finally` / `return()` runs (`packages/surface/src/project.ts:258-289`). Added a test (`project.test.ts` "connect()'s disposer aborts AND awaits the upstream generator's async teardown (F3)").
- `F4` **partial** — cleanup ownership added where a real teardown scope exists, disputed where it doesn't:
  - kaval **fixed**: removed the abort-listener `void ptyHost.close()` (an already-aborted input signal registered the `once` listener too late to ever fire — codex's real bug) and now `await ptyHost.close()` in `daemonMain(...).finally(...)`, ordered and unmissable (`packages/kaval/src/daemonMain.ts:118-143`).
  - padi **fixed**: `SurfacesServed` now carries `close`; `runPadiDaemon` wraps boot+`daemonMain` in `try/finally` and `await served.close()` (covers boot-step failure AND signal shutdown), disposition unchanged loud-not-fatal (`packages/padi/src/daemonMain.ts:157-166, 306, 418-467`).
  - kolu **disputed→partial**: DELETED the dead `closeKoluSurface` export (zero importers = a dead knob, which the repo's no-override-knob rule forbids). Disputed wiring `close` into kolu-server's shutdown: that runtime is a process-lifetime module-eval singleton and its graceful shutdown is a SYNCHRONOUS signal handler (`packages/server/src/index.ts:198-203` → `process.exit(0)`). For a process-lifetime owner, process death IS the teardown; making the handler async to await `close` adds a shutdown-hang risk (a parked connector) for zero real benefit (its one source — the surface-app buildInfo connector — is released by process exit). `done` stays observed (`packages/server/src/surface.ts:356`). Executable serving examples were already rewired in the base diff.
- `F5` **fixed** — public docs swept: rewrote peer-server's "Router wrapping caveat" to "pass it directly" and DELETED the dead `export { implement }` + its now-unused import (`packages/surface/src/peer-server.ts`); fixed the two stale `implementSurface(surface, { channel })` snippets to `implementSurfaceOnPublisher(surface, deps, channel)` (`server.ts` inMemoryPublisher + inMemoryChannelByName docs); added `close` to the reServeSurface return in `website/src/content/surface/mirror-over-ssh.mdx:64`; fixed the "Always flatten before serving" bullet in all three synced SKILL.md copies (`agents/.apm/`, `.agents/`, `.claude/`) to teach the FINAL router + `{router,ctx,done,close}` + `*OnPublisher`. LEFT the `website/src/content/blog/surface-framework.mdx` snippet alone: a blog is a dated point-in-time narrative, not a living reference (the surface-reference rule scopes the doc gate to Reference pages). Reference pages (`ref-surface*.mdx`) were already updated in the base diff.
- `F6` **fixed** — teardown no longer uses a global settlement barrier. After aborting all sources, `close()` runs each source's settle-then-dispose INDEPENDENTLY under one `Promise.allSettled`, so a connector that ignores cancellation blocks only its OWN dispose, never a sibling's release (`server.ts:1454-1490`). Added a test proving an independent source's disposer runs while a sibling connector parks forever.
- `F7` **fixed** — `pumpRemoteSurface` now recognizes the supervision abort: the waiting-for-client catch logs "supervision stop … exiting" (not "failed") when `opts.signal?.aborted`, and the terminal log distinguishes "supervision stop … (session preserved)" from "session destroyed" (`packages/surface-remote/src/hostFanout.ts:203-215, 254-266`).
- `F8` **fixed** — `pruneToMembers`' `onEvict` is now REQUIRED (this helper's whole job is releasing the dropped mirror), and it iterates a snapshot of ENTRIES so every deleted key carries its own value to `onEvict` — dropping the `cache.get(host) !== undefined` guard that skipped an `undefined`-valued entry (`packages/server/src/padi/reServeEviction.ts`). Updated the 3 existing tests to pass a spy + pin "exactly once per removed entry", and added a totality test for an `undefined`-valued entry.
- `F9` **fixed** — `derived.cell` teardown now fail-fasts on the leaking transitions: `connect()` after `dispose()` throws ("after dispose"), a second `connect()` throws ("twice"), and `teardown` wraps `disposeEffect?.()` in `try/finally` so `node.dispose()` runs even if effect disposal throws (`packages/surface/src/reactor.ts:316-378`). Added dispose-before-connect and connect-twice tests.
- `F10` **fixed** — the dynamic `cellDeps` shape now types `connect?: CellConnector<unknown>` directly (its one source of truth), and the later `const cd = cellDeps as { connect?: CellConnector }` corrective cast is gone — the connect block reads `cellDeps.connect` (`server.ts` cells walk).

Self-verify: `just check` green (typecheck + biome across 1309 files). Suites green: surface 335 (was 327; +8 new tests), surface-remote 152, kaval 155, and — run with `KOLU_STATE_DIR` set (the harness env) — server's reServeEviction/seal/surface/padiBinding all pass. Two unrelated pre-existing env failures were confirmed NOT from this diff: bare `vitest` on server without `KOLU_STATE_DIR` throws at import (harness sets it), and padi's `spawnInput` "omits PADI_SOCKET" test reads the workstation's real `PADI_SOCKET=/run/user/1000/padi-…` env (fails on this dev box, CI-clean).


### Round 2

**codex** — approved: `false`

Eight findings are resolved. The supervision, transactional construction, joined teardown, logging, eviction, reactor lifecycle, and typing fixes are sound. F4 and F5 remain open because the executable Surface App server still discards `done`/`close`, and several live API comments still teach the obsolete fragment contract. I did not rerun the reported suites because this review is strictly read-only.

Findings:
- `F1` · major · resolved — The original swallowed-disposer defect is fixed. Per-source teardown now propagates settlement or disposal failures into the collected outcomes, and `done` rejects while `close()` still resolves. Sync and async disposer tests cover both forms. (packages/surface/src/server.ts:1463)
- `F2` · major · resolved — Construction is now transactional. The walk records deferred starts, and both constructors assemble their final router before invoking any connector. The singular and plural malformed-construction tests cover the orphaning case. (packages/surface/src/server.ts:1767)
- `F3` · major · resolved — `deriveCell` now retains its subscription task and returns an asynchronous disposer that aborts and awaits that task. Runtime disposal therefore joins upstream iterator teardown instead of merely initiating it. (packages/surface/src/project.ts:268)
- `F4` · major · open — Padi and Kaval now retain and await their runtime shutdown, and the Kolu-specific process-lifetime argument is accepted. However, the executable Surface App server still destructures only `{ router, ctx }`; it neither observes `done` nor retains `close`. The claim that executable examples were already rewired is therefore incorrect, and this canonical example still teaches an unowned lifecycle and can produce an unobserved `done` rejection. (packages/surface-app/example/src/server/main.ts:86)
- `F5` · major · open — The peer-server, publisher examples, main skill bullet, re-serve guide, and reference pages were corrected, and leaving the dated blog as historical is accepted. But the live documentation sweep remains incomplete: the server module still calls the result a router fragment and shows re-finalizing/spreading it, `define.ts` repeats the fragment contract, and the Surface App example says the plural router is wrapped again. The generated skill also still calls the re-served value a local fragment at line 44. (packages/surface/src/server.ts:5, packages/surface/src/server.ts:1546, packages/surface/src/define.ts:25, packages/surface-app/example/src/common/surface.ts:68)
- `F6` · minor · resolved — The global settlement barrier is gone. After aborting all sources, each source performs settle-then-dispose independently, and the regression test demonstrates that a parked connector cannot block a sibling disposer. (packages/surface/src/server.ts:1464)
- `F7` · minor · resolved — The pump now distinguishes requested supervision shutdown from wait failure and session destruction at both affected log sites. The original misleading clean-shutdown diagnostics are resolved. (packages/surface-remote/src/hostFanout.ts:203)
- `F8` · minor · resolved — Eviction cleanup is now required and total. Iterating snapshotted entries preserves `undefined` values, and tests pin exactly-once cleanup for removed entries and no cleanup for retained entries. (packages/server/src/padi/reServeEviction.ts:30)
- `F9` · minor · resolved — The derived-cell lifecycle now rejects connect-after-dispose and double-connect, while `try/finally` ensures node disposal is attempted even when effect disposal throws. The new misuse tests cover the leaking transitions. (packages/surface/src/reactor.ts:316)
- `F10` · nit · resolved — The dynamic dependency shape now references `CellConnector<unknown>` directly, and the later corrective cast has been removed. The connector contract has one type-level source of truth. (packages/surface/src/server.ts:1603)

_codex on the rebuttal:_ F4: The Kolu-specific dispute is accepted — its runtime is a module-level process-lifetime singleton behind an immediate `process.exit`, so deleting an unused close export is preferable to adding a potentially hanging asynchronous signal path. Padi and Kaval are also adequately fixed. F4 nevertheless remains open because `packages/surface-app/example/src/server/main.ts:86` still drops both lifecycle members and has no `done` observer or shutdown path; that was part of the original serving-site/example feedback.

F5: Leaving the dated blog as a historical snapshot is accepted. That concession does not resolve F5 because current source documentation and an executable example still explicitly teach fragment wrapping, including `server.ts:5`, `server.ts:1546`, `define.ts:25`, and `surface-app/example/src/common/surface.ts:68`. Those are live guidance, not historical narrative.

The fixes and reasoning for F1-F3 and F6-F10 are accepted.

**claude** — Closed the two remaining findings: the canonical Surface App example now OWNS its runtime (observes `done`, awaits `close` on shutdown), and the last live comments that taught the obsolete router-fragment/spread contract were rewritten to describe the FINAL router + supervised runtime.

- `F4` **fixed** — `packages/surface-app/example/src/server/main.ts` now models the ownership doctrine the skill teaches, not an unowned lifecycle. The runtime is destructured in full (`{ router, ctx, done, close }`, main.ts:86); `done.catch(...)` routes an owned fault (the buildInfo connector rejecting) into a loud crash so the rejection can't go unobserved (main.ts:~111); and a guarded `SIGINT`/`SIGTERM` handler `await close()`s the runtime, then stops the WS + HTTP server before `process.exit(0)` (end of file). Unlike the kolu-server dispute codex accepted last round (a module-eval process-lifetime singleton behind a synchronous `process.exit`), this is a standalone teaching example whose whole job is to demonstrate the doctrine, so the async close path is correct here. Typecheck green for `packages/surface-app/example`.

- `F5` **fixed** — swept the remaining live comments that still taught the deleted fragment/spread contract:
  - `packages/surface/src/server.ts:5` (module headline) — rewritten from "produces both a router fragment and a typed ctx" to "returns a supervised `SurfaceRuntime` `{ router, ctx, done, close }`; `router` is the FINAL top-level router — serve it directly".
  - `packages/surface/src/server.ts` (the orphaned doc block that had drifted above `walkSurface`, ~line 1546) — it taught `const t = implement(fullContract); t.router({ ...surfaceRouter, terminal })`. Rewritten to say `router` is FINAL and that sitting a surface beside raw oRPC is done by spreading the built router's own `surface` namespaces into an assembled object, never by re-running `implement` (mirrors how kolu actually assembles: `{ surface: { ...koluSurfaceRouter.surface, padi } }` in `packages/server/src/surface.ts:364`).
  - `packages/surface/src/define.ts:25` — the "Same on the server: `implementSurface` returns a router fragment … spread alongside `t.terminal.handler(...)`" line rewritten; kept the CONTRACT-level `oc.router({ ...surface.contract, terminal })` composition (still valid) but corrected the server half to the FINAL-router story.
  - `packages/surface-app/example/src/common/surface.ts:68` — "The server wraps `implementSurfaces`' router with `implement(contract).router(...)`" → "serves it directly (already the FINAL top-level router — no re-wrap)".
  - The generated skill "local fragment" at line 44 — reworded to "the local re-served surface (a FINAL `implementSurface` router)" across all THREE synced copies: the source `agents/.apm/skills/surface/SKILL.md` plus the generated `.claude/skills/surface/SKILL.md` and `.agents/skills/surface/SKILL.md` (edited all three directly to keep them in sync, matching round 1's approach rather than re-vendoring mid-review).

Self-verify: `just fmt` clean (0 reformatted) and `just check` green — every package typecheck Done, biome `Checked 1309 files … No fixes applied`.

commit: `fd0b9ee63a38ae4ce997cb65f8dfb2b34f269bc4`


### Round 3

**codex** — approved: `true`

Approved. The current tree resolves all F1–F10 findings. F4 now has an observed fault channel and orderly runtime shutdown in the canonical executable example, while F5's remaining live documentation consistently describes the final router and supervised runtime. No regression was introduced by this round; the worktree has no untracked changes and the full diff passes `git diff --check`. I did not rerun the reported checks because this review remained strictly read-only.

Findings:
- `F1` · major · resolved — Disposer failures are collected and routed to `done` while `close()` remains resolving. (packages/surface/src/server.ts:1463)
- `F2` · major · resolved — Connector starts are deferred until validation and final router construction complete. (packages/surface/src/server.ts:1767)
- `F3` · major · resolved — Derived-cell teardown retains, aborts, and awaits the subscription task. (packages/surface/src/project.ts:268)
- `F4` · major · resolved — The executable example now retains `{ router, ctx, done, close }`, observes `done` with its fatal policy, and awaits `close()` from a guarded SIGINT/SIGTERM shutdown before stopping its servers. (packages/surface-app/example/src/server/main.ts:86)
- `F5` · major · resolved — The remaining live comments and synchronized skill copies now describe the final top-level router, supervised runtime, direct serving, and correct assembly with raw namespaces. The obsolete fragment/re-finalization guidance is gone from the identified live sources. (packages/surface/src/server.ts:4, packages/surface/src/define.ts:25, packages/surface-app/example/src/common/surface.ts:68, agents/.apm/skills/surface/SKILL.md:44)
- `F6` · minor · resolved — Each source now performs settle-then-dispose independently after the common abort phase. (packages/surface/src/server.ts:1464)
- `F7` · minor · resolved — Supervision shutdown is distinguished from wait failure and session destruction in diagnostics. (packages/surface-remote/src/hostFanout.ts:203)
- `F8` · minor · resolved — Eviction cleanup is required, total, and covered for removed, retained, and undefined-valued entries. (packages/server/src/padi/reServeEviction.ts:30)
- `F9` · minor · resolved — Derived-cell lifecycle misuse fails fast and cleanup attempts both effect and node disposal. (packages/surface/src/reactor.ts:316)
- `F10` · nit · resolved — The dynamic dependency shape uses `CellConnector<unknown>` as its single connector type source. (packages/surface/src/server.ts:1603)

_codex on the rebuttal:_ F4: accepted. The example now demonstrates both required lifecycle obligations: observing `done` and awaiting `close()` from a real shutdown scope.

F5: accepted. The remaining live fragment/re-wrap descriptions identified last round were corrected in source documentation, the executable example, and all synchronized skill copies. No disputes remain.

srid:

Two passes over the SurfaceRuntime diff (base `3f710572`) — Pass 1 rule checklist (31 rules), Pass 2 fact-check. `--no-elegance` (the `/simplify` track already ran). Both passes independently converged on the same top finding.

### Findings actioned

| # | Rule / class | Finding | Action |
| --- | --- | --- | --- |
| 1 | `no-silent-error-swallowing` (Pass 1 **and** Pass 2) | `superviseSurface.close()` surfaced only `outcomes.find(rejected)` — the first faulting source — so when several owned sources fault in one `close()`, every fault after the first was silently discarded, contradicting the function's own "BOTH a settle fault AND a dispose fault must reach `done`" doctrine. Untested. | **Fixed** (`71a8b3cd1`): the eager pre-close `.catch` stands down once `close()` begins (gated on the existing `closing` handle) so the barrier is the sole settler during teardown; faults then aggregate — 0 resolve, 1 rejects with its own reason, several reject with an `AggregateError`. Pinned with a two-disposer-fault test. |
| 2 | `comment-the-non-obvious` | `deriveCell`'s two teardown paths differ in completion guarantee — the returned disposer `abort()`s then `await`s the subscription task (abort-then-observe), the standalone `dispose()` fire-and-forgets — undocumented. | **Fixed** (`71a8b3cd1`): documented the asymmetry on `deriveCell`. |
| 3 | `comment-the-non-obvious` | ~13 example/test sites still named the `implementSurface(...)` result `fragment` — it's a supervised runtime now, not a fragment (one snippet's name even contradicted an adjacent "already the FINAL flattened router" comment). | **Fixed** (`ebc529495`): renamed the misleading local vars to `runtime` and aligned the adjacent prose. |

### Checked clean (no violation)

29 of 31 rules produced no finding — notably `caught-error-must-not-collapse-to-empty` (the `isAbortReason` swallow is a documented clean-cancellation path; the genuine fault still reaches `done`), `no-overloaded-null` (every new `T | undefined` carries one meaning), `dry-rule-of-three` (the 4 per-site `done.catch` observers differ in disposition — deliberately not unified), `no-untyped-escape-hatches` (the `as never`/`unknown` casts are the pre-existing opaque-router boundary, relocated not added), and `no-thin-wrapper-functions` (the ordinary constructors forwarding to `*OnPublisher` clear the ≥3-caller carve-out).

srid:
## Evidence

This is a **byte-identical consolidation** — the plan's invariant is that nothing on screen changes (a feature never rides one of these PRs). So the evidence class is the **e2e suite + the `packages/server` seal test**, not a screenshot: a screenshot would be evidence-theater for a PR whose whole claim is that the rendered output is unchanged.

- The **seal test** (`packages/server/src/seal.test.ts`) + `implementSurface.test.ts` router-shape pins hold the wire/router shape byte-identical, and the **padiBinding integration test** (dials `/surface/padi/*`) proves the `servedContract` deletion routes identically — 34/34 green locally, and the full two-platform `ci::default` pins them on the shipping HEAD.
- The **supervision honesty** the PR adds is pinned by `surfaceRuntimeSupervision.test.ts` (owned faults reach `done`, `close()` idempotent + aborts + disposes, abort-caused rejections are clean cancellations, every teardown fault surfaces via `AggregateError`).

Paired drishti PR: **srid/drishti#97**, pinned to the final kolu HEAD `d266c623b`, its own two-platform CI green.

srid:
## CI-caught regression + fix — `servedContract` is load-bearing for the wire matcher

The first two-platform CI run caught a **boot-time regression** the 1,300 unit tests and the five-stage gauntlet all missed — the e2e net earned its keep.

### Root cause (deterministic, not concurrency)

Deleting `servedContract` (the padi-widened contract) broke the HTTP/ws wire matcher. oRPC's `implement(C).router(obj)` adapts `obj` against `C` and **silently drops any key `C` doesn't declare**. `servedContract` was not "built solely to finalize the surface" — it is **also** the app-router builder that **re-adapts the re-served `padi` sibling** under the `padi` key, attaching the `/surface/padi/*` routes the matcher needs.

Critically, `serveHostMap` returns a `{ surface }` **fragment** carrying no matcher meta of its own — only structurally navigable. So `directLink` (which navigates structurally) works — which is exactly why the `directLink`-based `padiBinding` integration test stayed green while the wire matcher 404'd every `/surface/padi/*`: the server was healthy and the padi session connected, but `POST /rpc/surface/padi/lifecycle/killAll` → 404 → padi never became "live" → the e2e `BeforeAll` timed out on every worker.

Confirmed three ways: an idle-box e2e reproduction (load 0 — deterministic, 70/70 killAll = 404), a local oRPC matcher repro (contract-less `t.router` drops padi), and the server-log 404s.

### The fix (`ab74926bd`) and the honest plan-tension

`servedContract` is **restored as the app-router builder only**. The **surface finalization still moved** to `implementSurfacesOnPublisher` — that part of the plan holds (kolu extracts the runtime router's `.surface` and re-adapts it). `buildAppRouter` splices the assembled surface through `t = implement(servedContract)` again, as master did.

So the deletion-proof shrinks, stated plainly: **kolu's `implement(servedContract)` *surface-finalization* is gone, but its *app-router-builder* role stays** — the `/surface/padi/*` path prefix requires the builder's contract to declare `surface.padi`, and the map fragment must be re-adapted through it. The plan's headline "delete `servedContract`" was infeasible for the wire matcher; the PR body and plan note are corrected to match.

### Proven green + a new matcher-level pin

Idle-box e2e: **before**, 0 scenarios (all 8 workers' `BeforeAll` timed out on padi-never-live); **after**, `local padi is live` on every worker and **508/509 scenarios pass** (the 1 fail is an unrelated single-scenario toast flake). Added `packages/server/src/router.test.ts` — a `StandardRPCMatcher` assertion that the assembled `appRouter` routes `/surface/padi/entries/*`, built through the exported `t` (whose `surface.padi` node exists only when `servedContract` widens the contract). This is the exact matcher-level coverage the `directLink` integration test structurally cannot provide — the gap that let this ship past the gauntlet.

srid:
**Framing correction (per coordinator):** the campaign's negative property holds **verbatim** — *no production consumer imports oRPC `implement` **merely** to finalize a Surface*. kolu's remaining `implement(servedContract)` is the app-router **builder** (binding `server`/`daemon`/`hosts` and re-adapting the padi map fragment for the wire matcher), not mere finalization. Only the headline shrank; the property is intact. The `router.test.ts` matcher pin is the lesson — structural (`directLink`) navigation can't see matcher meta, which is why this shipped past the gauntlet.

srid:
## ✅ Ready — both gates green on the final HEAD

- **kolu #1805** `6a8cff0c8`: two-platform `ci::default` green (x86_64-linux on kolu-ci-3 + aarch64-darwin on rasam) — every node ok, **including both e2e lanes** (the net that caught the wire-matcher regression). 26 `ci::` contexts pass.
- **Paired drishti PR [srid/drishti#97](https://github.com/srid/drishti/pull/97)** `8a24019`, pinned to this HEAD `6a8cff0c8`: two-platform green (20/20 nodes across linux + darwin). Server-internal fix → no `@kolu/surface` API delta, so it's a pin-bump validation; its base was merged up to include the PR4 pair (drishti#96) it now sits beside.

Marking ready for review. srid merges.

Issue #606, opened 2026-04-17
Mode-toggle Focus↔Canvas leaks entire xterm Terminal graphs (600+ MB per test)


srid:
## Summary

Follow-up to #591 investigation. Diagnostic tooling added in #605 confirmed that **toggling between Focus and Canvas mode leaks the entire xterm `Terminal` service graph** — including the WebglAddon's texture-atlas `OffscreenCanvas` pages, the CharSizeService, and the measurement canvas. A short interactive test (6 toggles with 10 terminals) grows Chrome's reported footprint from **155 MB → 1.3 GB**.

The pattern is the same class of leak as #591/#598/#600 (async-init cleanup race producing orphaned xterm instances), just triggered by a different user action.

## Reproduction

1. Open Kolu with ≥5 terminals.
2. Click the Focus/Canvas toggle in the header repeatedly (6 times is enough).
3. Observe Task Manager footprint climb by ~200 MB per toggle.

## Factual evidence (heap snapshot analysis)

Snapshot comparison: baseline → after 6 toggles.

**JS heap: 56 MB → 94 MB** (+38 MB JS-side; far more native-side).

**Chrome Task Manager (User's machine, 7 terminals):**

| Metric | Before toggles | After toggles |
|---|---|---|
| Footprint | 155 MB | **1.3 GB** |
| GPU | 122 MB | 375 MB |
| JS live | 27 MB | 137 MB |

**DiagnosticInfo `aliveCanvases`** (from #605 instrumentation) pinpoints the retained canvas bitmaps:

| Canvas | Size | State | Bytes |
|---|---|---|---|
| active | 4574×1836 | in-dom, live | 33.6 MB |
| #16, #19, #20 | 4574×**1904** each | detached, context-lost | 34.8 MB × 3 |
| #1–#5 | 1360×850 (startup) | detached, never GC'd | 4.6 MB × 4 + 0.9 MB |

Total detached canvas bitmaps pinned in JS: **~124 MB**.

## Root cause (heap-snapshot retainer walk)

Chrome heap snapshot shows **70 `OffscreenCanvas` JS wrappers + 70 `OffscreenCanvasRenderingContext2D` + 70 `lU` (xterm AtlasPage) + 70 `tE` (xterm `CharSizeService`) instances** after 6 toggles with 10 live terminals. Expected live-set: 10.

Retainer chain for each leaked atlas page:

```
object:v3 --property _charSizeService--> object:tE --property _measureStrategy
[...]

PR #1808, opened 2026-07-14
refactor(xterm-kit): the <Xterm> wrapper cut + touch-divisor unification (PR 2 of 2)

srid:
PR 2 of 2 graduating kolu's xterm machinery into **`@kolu/xterm-kit`** — the e2e-gated fast-follow to #1795. Plan of record + API design: [xterm-kit.mdx](https://kolu.dev/atlas/xterm-kit.html) (the PR-2 section). Two deliverables.

## 1. The `<Xterm>` component cut (behavior-neutral)

`Terminal.tsx`'s ~700-line `onMount` graduates into `@kolu/xterm-kit/solid`: the xterm **hazards** move behind a `<Xterm>` component that owns them and hands the consumer a live `XtermHandle` in `onReady` (inside the reactive owner); kolu's **policy** — the padi wire/stream, keybindings, the PTY, diagnostics, the file-ref `onTap`, the `__xterm` e2e bridge and every `data-*` attribute — stays in `Terminal.tsx` (now −430 net lines). New `/solid` primitives, each behavior-identical to the extracted inline code:

- **`createXtermLifecycle`** — owner-correct async construct/dispose: capture the owner before the font `await`, `disposed`-bail, `runWithOwner`, synchronous cleanup. The #591/#606 leak choreography, in one place.
- **`attachWebGL`** — single-owner `WebglAddon` lifetime + context-loss recovery + explicit `loseContext()` on unload; the renderer gate is an `Accessor<boolean>` (budget stays the consumer's policy); the #591 tracker rides injected diagnostics hooks so kolu keeps it.
- **`wireTouchTaps` / `wireTouchScroll`** — the mobile touch surface xterm 6.0 ships none of (tap-vs-scroll + iOS soft-keyboard focus rules; touch→scrollback bridge). `onTap` is the policy hook.
- **`<Xterm>`** composes those with the PR-1 primitives (`createRenderRecovery`/`wireScrollIntent`/`createScrollLock`/`enableSoftKeyboardInput`) and spreads consumer `data-*`/`class` onto its mount div.

**Behavior-neutral headline:** the 123 client terminal unit tests stay green unmodified; the stream/backfill/snapshot/RIS choreography is logic-identical (only `scrollLock`/`write`/`recovery` reroute through the handle). No new external packages (the xterm addons were already in the workspace) → the pnpm FOD hash is unchanged.

> **REVIEW PIN — cucumber steps stay untouched.** The cut deliberately preserves the `__xterm` e2e bridge and every `data-*` attribute (`data-terminal-id`/`visible`/`focused`/`sub-terminal`/`font-size`/`renderer`) on the `<Xterm>`-rendered div, and changes no feature file or step the existing suites use. Please **verify** this rather than assume it: the diff must not alter any existing `.feature`/step-definition, and the terminal e2e must pass unchanged.

## 2. The touch-divisor unification (correctness-gated)

The touch tap path hand-rolled its own `rect.width / cols` pointer→cell divisor; xterm's selection/hover/mouse-reporting already resolve through `_core._mouseCoordsService` (transform-corrected by `patchTransformAwareMouseCoords`). The two agree only when the screen's layout width equals `cols × cssCellWidth` — rounding, sub-pixel metrics, or padding diverge, magnified under a zoomed tile near a cell boundary (#1400). This adds **`cellAtPoint`** to `@kolu/xterm-kit/internals` — the single pointer→cell authority (cosmetic read: degrades to null) — and routes the tap through it, deleting the parallel divisor and its paired "keep both in step" fence. (No user-observed regression found: the change is the *reuse-the-source-of-truth* dedup the design philosophy asks for, and the gate proves the single authority stays correct.)

**Done-criterion is correctness, in the standard e2e suite:** a touch tap on a file-ref's font-metric *visual* centre resolves to that ref, at normal scale **and** under a zoomed canvas tile.

### The `@touch-desktop` context (device class the gate represents)

That gate lives in exactly one quadrant: a **coarse-primary pointer that can hover, at desktop width** — a touchscreen device that also hovers (e.g. a stylus-hover tablet / a touch display driven as a desktop). It is narrow **yet decisive**: it is the *only* config where kolu both mounts the spatial pan/zoom canvas (the desktop layout, since `handheld = (pointer:coarse) and (hover:none)` is false) **and** wires the touch tap (`isTouch = (pointer:coarse)`) — i.e. exactly where a pointer→cell divisor under zoom can bite. A new **`@touch-desktop`** tag is the single switch that builds it: `hasTouch: true` (for `touchscreen.tap`) bundled with a CDP `setEmulatedMedia` pinning `(pointer:coarse)` and `(hover:hover)` — the media the client's `layoutMode` keys on (Playwright's `emulateMedia` can't set pointer/hover). The desktop viewport is untouched; the two settings never surface separately.

## Design philosophy

- **Reuse the source of truth** — the divisor change deletes a hand-rolled parallel of xterm's own cell authority.
- **Volatility boundary (electricity)** — the hazards graduate into the `@kolu/xterm-kit` receptacle; the app keeps only "which bytes, when, for whom".
- **Fail-fast** — the buffer-mutating reaches throw; only the cosmetic `cellAtPoint` degrades to null (a mis-tap, never corruption).


srid:

✅ **Consensus** after 2 round(s) · lowy + hickey · base `d4d4c06ea312`

### Applied (3)
- `lowy-1` XtermHandle flattens webgl's three methods while nesting every other mechanism (broken symmetry) — commit `bd3a89785`
- `hickey-1` Live terminal fragmented across `terminal` + `searchAddon` + `handle` — commit `220978b5f`
- `hickey-2` `liveTerm` + `fit` are two retained projections of one `XtermCore` — commit `db1e91e7c`

### Agreed — no change (5)
- `lowy-2` onTap hands raw pixels, forcing the consumer to round-trip the boundary through a second door for pixel→cell (packages/client/src/terminal/Terminal.tsx:195)
- `lowy-3` /solid barrel publishes the constituent primitives with no second consumer (speculative public surface) (packages/xterm-kit/src/solid/index.ts:29)
- `lowy-4` handle.write duplicates a reach already available via handle.scrollLock (packages/xterm-kit/src/solid/Xterm.tsx:198)
- `hickey-3` `<Xterm>` handle exposes two write paths for one operation (packages/xterm-kit/src/solid/Xterm.tsx:47)
- `hickey-4` `createXtermLifecycle` hard-codes the full addon set into the kit (packages/xterm-kit/src/solid/xtermLifecycle.ts:84)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 3 round(s) · codex reviewed at `xhigh` reasoning effort · base `d4d4c06ea312`

### Round 1

**codex** — approved: `false`

The extraction boundary is broadly coherent, but the branch is not ready: the new touch-desktop test configuration is overwritten before navigation, cleanup parity drops known xterm retainer breakers, reactive constructor values can go stale during the font await, and initialization errors leave a partially configured terminal alive. I also found a hybrid-input regression, a non-discriminating headline e2e, stale Atlas documentation, and dependencies left behind in the client manifest.

Findings:
- `F1` · major · open — The standard-suite `page.emulateMedia({ reducedMotion: "reduce" })` runs after the `@touch-desktop` CDP call at line 690. Playwright implements this by sending another `Emulation.setEmulatedMedia` containing only its preference features, replacing the earlier pointer/hover feature list. The scenario therefore does not retain the intended `(pointer: coarse) and (hover: hover)` state; with `hasTouch: true`, it can fall back to the handheld layout and fail to exercise a zoomable desktop canvas. (packages/tests/support/hooks.ts:1149)
- `F2` · major · open — The extraction dropped two explicit #606 cleanup safeguards from the old component: cancellation of the pending fit animation frame and nulling addon-bearing references. `<Xterm>` now retains `core` and `clearAtlas` forever within its component closure, while `core.addons.fit` retains its `_terminal` even after disposal (`FitAddon.dispose()` is empty). A pending rAF is also an external browser root and can remain parked indefinitely in a background tab. This defeats the claim in `xtermLifecycle.ts` that addons are reachable only through the consumer handle and recreates the retainer chain this change is meant to eliminate. (packages/xterm-kit/src/solid/Xterm.tsx:117)
- `F3` · major · open — `theme` and `fontSize` are snapshotted into `terminalOptions` before `document.fonts.load()` resolves. If either prop changes during that await, the deferred effects run while `core` is null and discard the update; construction later uses the stale snapshot and no reactive dependency changes again. The old implementation read `props.theme` and `fontSize()` after the await, so this is a behavioral regression and contradicts the documented live-reactive contract. (packages/xterm-kit/src/solid/Xterm.tsx:136)
- `F4` · major · open — The catch covers addon setup and the consumer's `onReady`, logs the error, and then resolves. If `onReady` or an addon throws, the terminal remains partially mounted, `<Xterm>` skips its initial fit, and the user sees a broken or blank terminal with no failure state. This is precisely a caught error collapsing into an empty state rather than failing loudly. (packages/xterm-kit/src/solid/xtermLifecycle.ts:104)
- `F5` · major · open — `wireTouchTaps` is installed based on the device's coarse-pointer media state, but its pointer listeners accept every pointer type. On the newly modeled coarse+hover desktop class, mouse or pen `pointerdown` is therefore prevented and its `pointerup` is treated as a touch tap. Canceling pointerdown suppresses xterm's normal mouse path, breaking selection, focus, and link interaction for a hardware mouse/pen on that device class. (packages/xterm-kit/src/solid/touch.ts:52)
- `F6` · minor · open — The new e2e does not actually distinguish `cellAtPoint` from the deleted `rect.width / cols` implementation. For a CSS scale, both the font-metric point and the old divisor are multiplied by the same scale, so the scale cancels when deriving the column. The target appears near the left edge and is tapped at cell center, giving enough tolerance that the old implementation maps it to the same cell under realistic rounding. Reverting Terminal to the old divisor would therefore still pass the advertised regression gate. (packages/tests/step_definitions/file_ref_link_steps.ts:130)
- `F7` · minor · open — The now-implemented Atlas record is stale in several verifiable ways. Its `<Xterm>` example omits required `fontFamily`, `scrollLockEnabled`, and `onTap` props. Citations at lines 99, 108, 153, 378, and 380 point to unrelated or now-nonexistent `Terminal.tsx` ranges after the file was shortened. It also says standard e2e proves WebGL context loss and async disposal, but no such scenario exists in the repository. The generated HTML carries these inaccuracies too. (docs/atlas/src/content/atlas/xterm-kit.mdx:77)
- `F8` · minor · open — Dependencies whose imports moved wholly into `@kolu/xterm-kit` remain declared by the client, obscuring the new ownership boundary. The client no longer imports resize-observer, addon-fit, addon-image, addon-unicode11, addon-web-links, or addon-webgl. Search and serialize remain legitimate direct type dependencies. (packages/client/package.json:40)


**claude** — Fixed the real defects (media clobber, the two dropped #606 retainer breakers, the stale-during-await constructor, the mouse-on-coarse-hover tap hijack), conceded the overclaiming e2e/doc language, cleaned the client manifest, and disputed only the pre-existing error-handling critique.

- `F1` **fixed** — Real bug. `page.emulateMedia({reducedMotion})` (hooks.ts, was line 1149) resends `Emulation.setEmulatedMedia` with only its own feature, wiping the pointer/hover CDP override that `newScenarioPage` had sent earlier — so `@touch-desktop` silently fell back to the handheld layout. Moved the pointer/hover pin out of `newScenarioPage` into a new `pinTouchDesktopMedia(context, page, reducedMotion)` helper called AFTER the reduced-motion emulateMedia, folding reduced-motion back into the same authoritative CDP call, and added a fail-fast `matchMedia` assertion (coarse ∧ hover ∧ ¬handheld) before navigation. `packages/tests/support/hooks.ts:681-688, 703-742, ~1160`.

- `F2` **fixed** — Real parity gap, consistent with the deliberate cross-primitive cleanup split (each part frees its OWN retained refs). The extraction moved `fitRaf` + the addon-bearing `core` into `<Xterm>` but dropped the old component's `cancelAnimationFrame(fitRaf)` and addon-ref nulling (old Terminal.tsx:420,438). `core.addons.fit` retains `_terminal` (FitAddon.dispose() is empty), so a retained `core` re-forms the #606 chain, and a parked rAF is an external browser root. Added an `<Xterm>`-local synchronous `onCleanup` that cancels `fitRaf` and nulls `core`/`clearAtlas`. Registered BEFORE `createXtermLifecycle`, so it runs LAST in LIFO — webgl-unload-before-term.dispose is untouched. `packages/xterm-kit/src/solid/Xterm.tsx:131-146`.

- `F3` **fixed** — Real regression. The old code read `props.theme`/`fontSize()` AFTER the font await (old Terminal.tsx:483-484); the new code snapshotted them into `terminalOptions` at component-body time, and the deferred effects fire-and-bail while `core` is null, so a mid-await change is lost forever. Changed `createXtermLifecycle`'s second param from a value to a thunk `() => XtermLifecycleOptions`, re-read at construction (post-await) for fresh theme/fontSize while still reading `fontFamily` eagerly for the font load; `<Xterm>` now passes a thunk and its `terminalOptions` type uses `ReturnType<...>`. `packages/xterm-kit/src/solid/xtermLifecycle.ts:55-91`, `Xterm.tsx:76-79, 134-144`.

- `F4` **disputed** — Pre-existing behavior, not drift, and out of scope for a behavior-neutral extraction. The old inline code had the IDENTICAL catch — `catch (err) { console.error("Terminal onMount failed:", err); }` (old Terminal.tsx:1160-1162) — wrapping the entire synchronous setup INCLUDING all addon loading and the onReady-equivalent policy wiring. `createXtermLifecycle`'s catch (xtermLifecycle.ts:114-116) has the same scope and semantics; it exists to satisfy `noMisusedPromises` (surface an async rejection to the console instead of the unhandled-rejection stream). The fail-fast philosophy point has merit, but re-architecting terminal setup to dispose partial state and surface a required error channel to Solid error boundaries is a NEW behavior and a NEW design question (what's the terminal error-state UI?), not something this extraction introduced or should smuggle in. Note: this is the fail-loud opposite of the DELIBERATE cosmetic `cellAtPoint`-degrades-to-null read (note c) — the two live at different layers.

- `F5` **fixed** — Real correctness bug the NEW `@touch-desktop` device class (coarse pointer that CAN hover — a touchscreen laptop) exposes: `wireTouchTaps` is wired whenever the pointer is coarse (via `enableSoftKeyboardInput`'s `isCoarsePointer()` gate), but its handlers `preventDefault`'d EVERY `pointerdown` and read every `pointerup` as a tap — hijacking a real mouse/pen on that class, breaking xterm selection/focus/links. Gated all three handlers on `e.pointerType === "touch"`. Behavior-neutral for phones (touch is always pointerType "touch"; Playwright `touchscreen.tap` too), strictly-better for coarse+hover desktop. The old inline code had the same unfiltered handlers, but the coarse+hover class is new in this PR, so this is in-scope hardening. `packages/xterm-kit/src/solid/touch.ts:52-79`.

- `F6` **partial** — codex is right that the scenario does NOT discriminate `cellAtPoint` from the deleted `rect.width/cols`: a CENTRE tap on a cell resolves to that same cell under both divisors, the CSS scale cancels (discrimination depends only on k = cols·cssCell/offsetWidth, which is sub-pixel in kolu's rendering), and the old POST-transform divisor already self-corrected for zoom (old Terminal.tsx:612-621). I did NOT add a hard "assert the two mappings differ" gate — that assertion would be VACUOUSLY FALSE (they never diverge enough at a centre tap), i.e. a permanent red, which is itself the concession that codex's underlying point holds. Instead I corrected the OVERCLAIM in the feature comment and the `findRefFontMetricPoint` doc: it is a regression guard that the EXTRACTED touch tap still routes a visual tap to the right cell (incl. under zoom), not an equivalence disproof. The scenario keeps real value (guards the extracted tap path end-to-end). `packages/tests/features/file-ref-link.feature:38-52`, `step_definitions/file_ref_link_steps.ts:73-83`.

- `F7` **fixed** — Real doc staleness. Made the `<Xterm>` example type-correct (added the required `fontFamily`, `scrollLockEnabled`, `onTap` props). Repointed the moved citations: async-dispose 405-449 → `xtermLifecycle.ts:50-108`; WebGL 205-270 → `webgl.ts`; touch tap 576-669 → `touch.ts:38-75` + `softKeyboardInput.ts`; focus-no-op 272-279 → `Terminal.tsx:169-176`; backfill 849-930 → `Terminal.tsx:410-440`. Narrowed the false "standard-suite e2e proves WebGL context-loss recovery + owner-correct async dispose" claim (no such scenario exists) to the real gate: the `@touch-desktop` tap-under-zoom e2e + 123 unmodified client unit tests, with context-loss/dispose keeping their existing unit coverage. Rebuilt + staged `docs/atlas/dist/xterm-kit.html`. `docs/atlas/src/content/atlas/xterm-kit.mdx`.

- `F8` **partial** — Removed the 5 addon deps client no longer imports (addon-fit, addon-image, addon-unicode11, addon-web-links, addon-webgl; 0 uses each in `packages/client`) and refreshed the lockfile. DISPUTE the 6th: codex named "resize-observer", but `@solid-primitives/resize-observer` is STILL imported by `packages/client/src/host/HostSelectorStrip.tsx:64`, so it stays. search + serialize kept (still imported), as codex agreed. The removed packages remain in the lockfile's global `packages:` section (xterm-kit still depends on them), so the pnpm FOD hash is unchanged — only client's importer entry shrank. `packages/client/package.json:44-47`, `pnpm-lock.yaml`.

Static gate: `pnpm typecheck` green (xterm-kit + client), `just --no-deps ci::biome` green (1320 files), `just fmt` clean. `packages/tests` has no tsconfig/typecheck script (biome is its gate) — verified biome-green.

commit: `0a5c5a8b70ebaa8bc2ab6e6f9c135533c3a1e345`


### Round 2

**codex** — approved: `false`

Most fixes are sound: F1, F2, F5, and F8 are verified, and I accept the F4 scope argument. Three issues remain: the F3 fix can construct with a different font than it awaited, F6 retains a contradictory overclaim in the step implementation, and F7 still claims unit coverage that does not exist.

Findings:
- `F1` · major · resolved — The pointer/hover media override was previously clobbered by the later Playwright media emulation. The final combined CDP override now runs afterward and asserts the required media state. (packages/tests/support/hooks.ts:703)
- `F2` · major · resolved — The extraction previously omitted cancellation of the pending fit animation frame and release of addon-bearing references. The new component-local cleanup restores both retainer breakers without disturbing the declared WebGL-before-terminal LIFO cleanup order. (packages/xterm-kit/src/solid/Xterm.tsx:139)
- `F3` · major · open — The theme/font-size freshness fix introduces a font race: line 84 loads the font family returned by the first `getOptions()` call, but line 91 constructs from a second call. If `fontFamily` changes during the await, xterm is constructed with a face that was never awaited, violating the lifecycle's documented guarantee and potentially measuring with a fallback font. (packages/xterm-kit/src/solid/xtermLifecycle.ts:82)
- `F4` · major · resolved — The broad setup catch is identical in scope and behavior to the base implementation; this extraction did not introduce the error-handling policy. (packages/xterm-kit/src/solid/xtermLifecycle.ts:114)
- `F5` · major · resolved — The touch tap handlers previously intercepted mouse and pen input on coarse-plus-hover devices. All three handlers now ignore non-touch pointers before altering state or preventing defaults. (packages/xterm-kit/src/solid/touch.ts:52)
- `F6` · minor · open — The feature and helper documentation now honestly say the center tap does not distinguish the two divisors, but the step implementation still says the old `rect.width/cols` mapping could resolve this same center pixel to the wrong cell under zoom. That directly contradicts lines 73–82 and the feature's lines 48–52. (packages/tests/step_definitions/file_ref_link_steps.ts:224)
- `F7` · minor · open — The example and citations are corrected, but the replacement coverage claim is still false. Lines 538–540 say WebGL context-loss recovery and owner-correct async disposal retain existing unit coverage, yet no test imports or exercises `attachWebGL` or `createXtermLifecycle`; the client terminal tests also do not render `Terminal` or `<Xterm>`. Those tests cannot serve as a behavior-neutral proof of this extraction. (docs/atlas/src/content/atlas/xterm-kit.mdx:535)
- `F8` · minor · resolved — The five moved addon dependencies were removed from the client manifest and lockfile importer. `@solid-primitives/resize-observer` correctly remains because `HostSelectorStrip.tsx` still imports it. (packages/client/package.json:38)

_codex on the rebuttal:_ F4: conceded. The base Terminal implementation has the same catch over the same setup scope at lines 1160–1162, so changing its failure contract would be a separate behavioral change rather than fixing extraction drift. F6: I accept that a divergence assertion at the visual center would be vacuous and no longer require one, but the step's own comment still makes the disproven divergence claim, so F6 remains open only for that stale wording. F8: conceded; the `HostSelectorStrip.tsx` import proves resize-observer remains a real client dependency, while the five genuinely moved addon dependencies are gone. F1, F2, and F5 are verified resolved. F3 remains open because the new two-read thunk can await one font and construct with another. F7 remains open because the revised Atlas still explicitly claims direct unit coverage that the test tree does not contain.


**claude** — Fixed all three open findings: pinned the awaited font into construction (F3), stripped the disproven divergence claim from the tap step's comment (F6), and corrected the false unit-coverage claim in the Atlas (F7). F1/F2/F4/F5/F8 already resolved last round.

- `F1` **fixed** (round 1, verified resolved) — media clobber; no further change.
- `F2` **fixed** (round 1, verified resolved) — dropped retainer breakers restored; no further change.
- `F3` **fixed** — Real race in my round-1 fix. `getOptions()` was read twice: once to build the `document.fonts.load()` string, once inside `new XTerm(...)`. A `fontFamily` that changed mid-await would be constructed with a face never awaited (measured against the fallback). Kolu's `fontFamily` is static, but the guarantee shouldn't rest on that. Now I capture `awaitedFontFamily` from the eager read and spread `{ ...getOptions().terminalOptions, fontFamily: awaitedFontFamily }` into the constructor — theme/fontSize stay freshly re-read post-await (the F3-round-1 win), while the font is pinned to exactly the face `document.fonts.load()` waited on. `packages/xterm-kit/src/solid/xtermLifecycle.ts:80-96`.
- `F4` **disputed** (conceded by codex round 2) — pre-existing catch, out of scope; resolved.
- `F5` **fixed** (round 1, verified resolved) — non-touch pointers now ignored; no further change.
- `F6` **fixed** — codex right on the stale wording. The step body's comment (`file_ref_link_steps.ts`, was lines 224-228) still asserted the old `rect.width/cols` divisor "could resolve this pixel to the wrong cell" under zoom — the exact divergence I'd already disproven in `findRefFontMetricPoint`'s doc (lines 79-82) and the feature. Rewrote the comment to describe the scenario honestly: an end-to-end regression guard that the EXTRACTED tap still routes a visual centre tap through xterm's single divisor authority (`cellAtPoint`) to the right cell at normal scale and under zoom, explicitly NOT a discriminator against the deleted divisor (a centre tap resolves to the same cell under both). `packages/tests/step_definitions/file_ref_link_steps.ts:224-231`.
- `F7` **fixed** — codex right; my round-1 rewrite still overclaimed. Verified the test tree: NO test imports or exercises `attachWebGL` (webgl.ts) or `createXtermLifecycle` (xtermLifecycle.ts). `renderRecovery.test.ts` covers the PR-1 watchdog (`createRenderRecovery`), and `webglBudget.test.ts` covers the budget POLICY (`admitWebglTiles`) — neither is webgl.ts context-loss self-heal or the lifecycle's owner-capture. Client terminal tests don't render `Terminal`/`<Xterm>`. Rewrote lines 535-540 to say plainly that WebGL context-loss recovery and owner-correct async dispose carry NO dedicated unit test — preserved as behavior-identical extracted logic, held to correctness by review + typecheck and the leak-invariant argument in the note, not by an automated scenario. The 123 client unit tests are now scoped as the proof for the surrounding policy, not for these paths. Rebuilt + staged `docs/atlas/dist/xterm-kit.html`. `docs/atlas/src/content/atlas/xterm-kit.mdx:535-542`.
- `F8` **fixed** (round 1, verified resolved) — five moved addon deps removed, resize-observer correctly kept; resolved.

Static gate: `@kolu/xterm-kit` `tsc --noEmit` green; `ci::biome` green (1320 files); `just fmt` clean. `atlas::check-sync` diffs a fresh build against committed HEAD, so it flags the dist as stale until this round's `dist/xterm-kit.html` is committed (working-tree dist already matches the fresh 98114-byte build) — in sync post-commit.

commit: `dc582a8b0b1708b4fd58376e0a788a95dba479a5`


### Round 3

**codex** — approved: `true`

All eight findings are resolved. The round-2 fixes correctly pin construction to the awaited font, remove the contradictory tap-test wording, and accurately disclose the absence of dedicated lifecycle/WebGL unit tests. The worktree is clean and no regressions were introduced by these fixes.

Findings:
- `F1` · major · resolved — The pointer/hover media override was previously clobbered by later Playwright media emulation. The final combined override and assertion resolve this. (packages/tests/support/hooks.ts:703)
- `F2` · major · resolved — The extraction previously omitted cancellation of the fit animation frame and release of addon-bearing references. Both retainer breakers are restored with the required cleanup ordering. (packages/xterm-kit/src/solid/Xterm.tsx:139)
- `F3` · major · resolved — The lifecycle could previously await one font and construct with another after a mid-await change. It now pins `awaitedFontFamily` into the constructor while re-reading theme and font size. (packages/xterm-kit/src/solid/xtermLifecycle.ts:85)
- `F4` · major · resolved — The broad setup catch is pre-existing behavior rather than extraction drift. (packages/xterm-kit/src/solid/xtermLifecycle.ts:120)
- `F5` · major · resolved — Touch tap handlers previously intercepted mouse and pen events on coarse-plus-hover devices. They now ignore every non-touch pointer. (packages/xterm-kit/src/solid/touch.ts:52)
- `F6` · minor · resolved — The step comment previously contradicted the acknowledged fact that a center tap does not distinguish the old and new divisors. It now describes the scenario accurately as an end-to-end regression guard. (packages/tests/step_definitions/file_ref_link_steps.ts:224)
- `F7` · minor · resolved — The Atlas previously claimed unit coverage for lifecycle and WebGL behavior that did not exist. It now explicitly states that these paths have no dedicated unit tests and accurately scopes what the existing tests cover; the generated HTML reflects the correction. (docs/atlas/src/content/atlas/xterm-kit.mdx:535)
- `F8` · minor · resolved — The five moved addon dependencies were removed, while the still-used resize-observer dependency was correctly retained. (packages/client/package.json:38)

_codex on the rebuttal:_ F3: accepted; the constructor now uses exactly the font family awaited while preserving fresh theme/font-size reads. F6: accepted; the revised comment no longer claims the center point distinguishes the deleted divisor. F7: accepted; both the Atlas source and generated HTML now state the actual coverage boundary. F4 remains conceded for the previously established reason that its catch behavior predates this extraction. F1, F2, F5, and F8 remain verified resolved.

srid:

Two passes over `git diff d4d4c06ea...HEAD` (the XKIT-PR2 `<Xterm>` cut + touch-divisor unification), `--no-elegance` (the simplify track already ran this gauntlet).

### Pass 2 — fact-check: 2 findings, both fixed (`30efc2491`)

The behavior-neutral headline mostly holds — `cellAtPoint`'s 1-based→0-based conversion was verified against `@xterm/xterm`'s real `getCoords` (returns 1-based, clamped), the attach-stream/backfill/snapshot/RIS choreography is a faithful transcription (same args, order, comments), and every `data-*` + the `__xterm` bridge stay on the same mount `<div>`. But the cleanup split had a real ordering regression:

| # | severity | finding | fix |
| --- | --- | --- | --- |
| F1 | **major** | `<Xterm>` is a plain JSX child (no own reactive owner), so `Terminal.tsx`'s **body-registered** `onCleanup` (stream abort, ref/diagnostics/activity deregistration, `backfill.dispose`, the `__xterm` #606 clear) landed **first** on the shared owner and ran **last** under SolidJS LIFO — *after* the kit's `terminal.dispose()`/webgl-unload, inverting the pre-cut order. `cleanNode` has no per-entry try/catch, so a throw in `terminal.dispose`/webgl-unload would **skip that consumer teardown entirely** (leaked stream, stale refs `Map`, un-cleared `__xterm`). | keep only the #606 audit counter in the body; move the consumer teardown into an `onCleanup` **inside `onReady`**, so LIFO runs it BEFORE the kit's disposal — the old single-function order, restored. |
| F2 | minor | the initial `fit()`/resize fired **after** `onReady` subscribed the attach stream (reversed from the pre-cut "size before stream"). | moved the initial fit before `own.onReady` in `<Xterm>`. |

### Pass 1 — rule checklist: 0 violations (30 rules)

Zero violations across all 30 applicable rules (7 built-in + 23 project-specific). The diff is overwhelmingly a **verbatim extraction** — every catch block, debounce, toast call, and hook wiring that looked new was cross-referenced against the base commit and confirmed identical to the shipped `Terminal.tsx`, just relocated. Clean on: `caught-error-must-not-collapse-to-empty` (`cellAtPoint`'s null-degrade is the ratified cosmetic-read philosophy; the construction catch is a verbatim move), `no-overloaded-null`, `no-silent-error-swallowing`, `dry-rule-of-three`, `prefer-focused-library`, `no-preference-prop-drilling` (the `Accessor<boolean>` gate is the documented kit/policy boundary), `e2e-poll-async-state`, `no-untyped-escape-hatches`, and ~22 more. The genuinely-new surfaces (`cellAtPoint`, the `<Xterm>` wiring, `@touch-desktop`) are each backed by unit tests or e2e regression guards.

srid:
## Evidence

PR 2 is a **behavior-neutral component cut** plus a **reuse-the-source-of-truth divisor dedup** — so the honest evidence is *the same suites stay green*, the leak invariants provably hold, and the divisor authority resolves correctly under a transform. Not a feature clip (there is no on-screen change for real users).

**Behavior-neutrality (the `<Xterm>` cut):**
- **123 client terminal unit tests stay green, unmodified** — the stream/backfill/snapshot/RIS/seam choreography is a faithful transcription (same args, order, comments); only `scrollLock`/`write`/`recovery` reroute through the `XtermHandle`.
- **The #591/#606 leak invariants hold — verified, not assumed.** code-police Pass 2 *caught* a real disposal-order regression the cut introduced (Terminal's body-registered `onCleanup` ran LAST under LIFO, after the kit's `terminal.dispose()`, and `cleanNode` has no per-cleanup try/catch — so a throw would skip the `__xterm` clear); it's **fixed** (`30efc2491`) by moving the consumer teardown into `onReady` so LIFO runs it first, restoring the pre-cut order. The owner-capture-before-await / `disposed`-bail / synchronous-cleanup choreography is intact.
- **kit unit + closure guard green**; `PTY_HOST_CONTRACT_VERSION` untouched; no new external deps → the pnpm FOD hash is unchanged.

**Divisor unification (correctness, not equivalence):**
- `cellAtPoint`'s 1-based→0-based conversion is pinned by a unit test and was verified against `@xterm/xterm`'s real `getCoords`.
- **A standard-suite e2e proves a touch tap on the glyph's font-metric visual centre resolves to the ref under a CSS transform** — routed through xterm's single `cellAtPoint` authority. **Honest fidelity caveat:** the literal real state (a touch tap on the *desktop canvas's zoomed tile*) is **Chromium-unreachable** — pointer↔hover is welded to touch capability (a two-run CI oracle proved it: with `hasTouch` → `{coarse, hover:none}`; without → `{fine, hover}`; CDP `setEmulatedMedia` pointer/hover had no effect), so `(pointer:coarse) and (hover:hover)` can't be constructed. The gate therefore runs in the reachable `@mobile` touch config with an **injected** CSS `scale()` standing in for the tile-zoom transform — `cellAtPoint` reads the scale off `getBoundingClientRect`/`offsetWidth`, so the transform's *source* is transparent to the code under test. Corroborated by `canvas-selection.feature` (the *same* authority under real canvas zoom, for selection) and the `cellAtPoint` unit delegation pin.

**The gauntlet earned its keep:** codex (xhigh) + code-police together caught **6 real bugs** the structural stages missed — a CDP media-clobber, a #606 leak regression, stale theme/fontSize across the font-await, mouse-events-on-touch, the disposal-order inversion above, and a fit-before-stream reorder — **all fixed and re-verified green**.

**Two-platform CI — `passed` on the shipping HEAD `f708e0652`** (x86_64-linux on a leased pool box, aarch64-darwin on rasam): every node `ok` on both platforms.

PR #1809, opened 2026-07-14
fix: stop re-firing agent notifications on every host switch

srid:
**An agent parked in _awaiting-you_ on one host used to re-notify — sound, banner, and dock badge — every single time you switched back to that host**, even though nothing about it had changed. A terminal sitting on a review prompt (the real report: `nixos-config` on `zest`) would ding on every return, forever.

The alert logic compared each host's agents to the previous tick's list **by array position** (`prevStates[i]` vs `states[i]`) over `deps.terminalIds()` — but that list is the *active host's* window. Switching hosts swaps the whole list, and even a same-host insert/remove reorders it, so index `i` named a **different terminal** across two ticks. An already-`awaiting_user` terminal that landed where a *working* one used to sit read as a fresh entry into the notify class and fired again.

### The fix — key the diff by identity, not position

```
before:  states[i]  vs  prevStates[i]        ← index i = different terminal after a switch
after:   curr.get(id) vs prev.get(id)         ← each terminal vs its OWN prior state
```

Build a `Map<TerminalId, state>` and compare each terminal against **its own** previous state. A first-sighting id (host switch, late-arriving metadata) has no prior entry, so it's skipped — the only terminals that can "transition" are ones we were already tracking last tick, on the same host. The cross-identity comparison is now **unspellable**, so a reshuffled list can't manufacture a phantom alert.

> Diagnosed under a `/perfection-review`: the shape is _"diff a reorderable list by array index."_ Host-switch was the loud costume; the same bug also fired on a plain same-host terminal open/close. The `Map`-keyed diff closes every costume, not just the visible one. A latent guard (`if (prev !== undefined)`) had masked it until the W9 retained-per-host-subs work removed the empty intermediate tick that was accidentally suppressing it.

### Coverage

`useTerminalAlerts` had **no test at all**. Added `useTerminalAlerts.test.ts` — red on the positional diff, green on the keyed one:

- a genuine `thinking → awaiting_user` still fires exactly **once**
- **no** phantom when switching hosts back to an already-awaiting terminal
- **no** phantom when a sibling above an awaiting terminal is removed (index shift)

### Try it locally

```sh
nix run github:juspay/kolu/notify-bug
```


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `4bc19abb07be`

### Agreed — no change (3)
- `lowy-1` Host-scoped list cross-tick diff is a latent receptacle sprayed across two hooks (packages/client/src/terminal/useTerminalAlerts.ts:64)
- `hickey-1` Identity-keyed on()-diff skeleton duplicated across sibling modules (packages/client/src/terminal/useTerminalAlerts.ts:64-83)
- `hickey-2` Cross-host id-disjointness: guarded explicitly in one sibling, relied on implicitly here (packages/client/src/terminal/useTerminalAlerts.ts:78)

srid:
## Codex ⇄ Claude debate

✅ **Consensus** after 2 round(s) · codex reviewed at `xhigh` reasoning effort · base `4bc19abb07be`

### Round 1

**codex** — approved: `false`

The keyed Map fixes positional reorder errors, but terminal identity remains incomplete: supported session import can duplicate a terminal UUID across hosts, allowing the same false notification to recur. The new tests also leave reactive/global state undisposed.

Findings:
- `F1` · major · open — The history is keyed only by TerminalId, yet the comment treats prevStates.has(id) as proof that the terminal belonged to the same host. Session import deterministically breaks that assumption: sleeping terminals retain their exported ID, can be imported onto another host, and wake on that same ID. Switching between different hosts containing that ID can therefore compare unrelated states and manufacture the exact false transition this change intends to prevent. (packages/client/src/terminal/useTerminalAlerts.ts:67)
- `F2` · minor · open — The harness creates a Solid root and returns its disposer, but every test ignores it. The hook also installs window.__koluSimulateAlert, leaving the latest fixture globally reachable and making test cleanup nondeterministic. (packages/client/src/terminal/useTerminalAlerts.test.ts:62)


**claude** — Fixed both findings: host-scoped the alert diff so a cross-host `TerminalId` collision can't manufacture a phantom, and gave the test harness real per-test teardown plus an F1 regression.

- `F1` **fixed** — Codex is right, and it disproves the author's "disjoint id space by construction" premise. That premise held only for fresh `crypto.randomUUID()` terminals; session import/restore preserves a SLEEPING terminal's id (`sessionRestore.ts:186` `seedSleepingTerminal` + `oldToNew.set(t.id, t.id)`; `importSession` at `sessionRestore.ts:421` runs the same path) and sleep/wake keep the id IN PLACE (`terminalEndpoint/local.ts:1013` `beginSleep`, same id/same slot; wake re-registers the `prior` sleeping record). So the SAME id can live on two hosts, and `has(id)` alone would pair host B's non-notify state against host A's `awaiting_user` and re-fire the exact phantom. Fix (`useTerminalAlerts.ts:74-89`): the `on()` source now snapshots `{ host: encodeHostKey(activeHost()), states }`, and the callback returns early when `prev.host !== curr.host` — after a host switch every terminal is a first sighting, so the collision is impossible to express by construction. The same-host keyed diff (`has(id)` membership vs value) is unchanged, so the original reorder/insert/remove fix stands. Added a regression `useTerminalAlerts.test.ts` ("two hosts share a TerminalId … one is awaiting") that fires on the un-gated diff and passes on the gated one.
- `F2` **fixed** — Codex is right; the harness leaked its reactive root and the last `window.__koluSimulateAlert` closure across tests. `harness()` now (a) returns the disposer directly from `createRoot((d) => { …; return d; })` instead of assigning through an unused `let`, (b) snapshots `window.__koluSimulateAlert` before mounting, and (c) registers an `onTestFinished` that disposes the root and restores the prior global. Every test picks this up with no per-test change.

Gate: `pnpm typecheck` clean across all packages; `biome lint --error-on-warnings` (the CI gate — not `biome check`, whose organize-imports assist doesn't run in the gate) clean on both files; all 5 vitest cases green.

commit: `779131ea4d94a23437687c99079bf721ccda0f3a`


### Round 2

**codex** — approved: `true`

Both findings are resolved. The alert snapshot now includes the active host and skips cross-host comparisons, while preserving same-host Map membership semantics. The regression covers duplicate IDs created through session import, and the harness now disposes each Solid root and restores the prior global hook. No regression was introduced by these fixes.

Findings:
- `F1` · major · resolved — The transition history previously keyed identity only by TerminalId, allowing imported duplicate IDs on different hosts to be compared. (packages/client/src/terminal/useTerminalAlerts.ts:77)
- `F2` · minor · resolved — The test harness previously left its Solid root and global simulation hook alive after each test. (packages/client/src/terminal/useTerminalAlerts.test.ts:65)

_codex on the rebuttal:_ F1: confirmed. The host value is captured in the same reactive snapshot as the states, and the host-change guard prevents duplicate imported IDs from crossing authority boundaries; the new regression exercises that exact case. F2: confirmed. Every harness invocation now registers teardown that disposes its root and restores the prior global hook. CLAUDE accepted both findings, so no dispute remains.

srid:

Ran Pass 1 (rule checklist) and Pass 2 (fact-check); Pass 3 (elegance) skipped (`/simplify` already ran this gauntlet).

**Pass 1 — rule checklist:** 1 finding.
- `no-untyped-escape-hatches` — `useTerminalAlerts.test.ts` `getSubject` stub returned `{ title, description } as never`. `TerminalSubject` is `{ title: string; description?: string }`, which the literal already satisfies, so the cast narrowed nothing (and was the only `as never` in `packages/client/src`). **Fixed** in `8cfc02662` — dropped the cast; tsc + biome stay green.

**Pass 2 — fact-check:** clean. Verified the membership-vs-value split (`has(id)` gate + `.get(id)` value) against `checkAgentFinished`, host-gate atomicity against the `useActiveReconcile` snapshot idiom, `Map`-dedup non-reachability (`terminalIds()` derives from a server `Map`, structurally duplicate-free), and `encodeHostKey` determinism. No new findings; the cross-host shared-id fix and test-teardown fix from earlier stages hold on disk.

srid:
## Evidence

**Why no automated UI clip here.** The behavior this fixes is an agent-attention alert firing on the wrong tick. Its channels are all **outside the page canvas** — an OS notification banner (service-worker `showNotification`), a notification **sound** (`new Audio().play()`), the app-icon **badge** (`navigator.setAppBadge`), and the browser's **tab-audio glyph** (the `🔊` in the title bar). None of these are observable by kolu's Playwright/browser-capture e2e harness (they live in the OS / browser chrome, not the DOM). The one faint in-app signal (a dock **unread** re-mark) would need the full **two-real-host remote harness** with an agent parked in `awaiting_user` to reproduce a host switch — disproportionate, and it still wouldn't capture the OS channels. So the honest artifact is the reported symptom as the *before* and a **red→green regression** that encodes that exact scenario as the *after*.

### Before — the reported symptom
On `zest`, the `nixos-config` terminal sat in **`awaiting_user`** ("Add drishti flake.nix input PR review"). Every switch **back** to `zest` re-fired the alert — the screen recording showed the title bar flashing `Kolu [zest] 🔊` (Chrome's tab-audio glyph = the notification sound) on each return, and again after leaving to `local` and coming back. The agent never changed state; only the host-scoped terminal list under it did.

### After — red→green on the exact scenario
The new `useTerminalAlerts.test.ts` drives that precise case (host A → host B where B's terminal is already `awaiting_user`). Run against the **pre-fix positional diff**, the scenario tests fail (the phantom fires); against the **fix**, all pass.

**Pre-fix (positional `prevStates[i]` vs `states[i]`) — 3 fail:**
```
❯ useTerminalAlerts.test.ts (5 tests | 3 failed)
  × does NOT re-fire an already-awaiting terminal when switching hosts back to it   ← the video bug
  × fires when a tracked terminal's agent appears already in the notify class
  × does NOT re-fire an awaiting terminal when a sibling above it is removed (index shift)
 Tests  3 failed | 2 passed (5)
```

**With the fix (identity-keyed, host-scoped diff) — all pass:**
```
 Test Files  1 passed (1)
      Tests  5 passed (5)
```

The `switching hosts back to it` test is the automated stand-in for the screen recording: red = the `🔊` re-fires, green = it doesn't. The suite also pins the two costumes the bug relocates to without any host switch (a same-host index shift) and the cross-host **shared-id** case (session-imported ids) the review gauntlet surfaced.

PR #1811, opened 2026-07-14
feat(surface): membership is time — opaque membershipId, typed connection key, clockNow reserved member (PR3)

srid:
## Membership is time, and time is a fact — Surface plan PR 3

Plan of record: [`surface-runtime-boundary.mdx`](https://kolu.dev/atlas/surface-runtime-boundary.html) § "Membership is time, and time is a fact — PR 3" (row 3). Pure consolidation — **behavior-neutral on screen**.

Three moves that share the admit/connection seam:

### 1. Opaque `membershipId` (`@kolu/surface-map`)
`serveSurfaceMap` stamps a `crypto.randomUUID()` when a key **enters** membership — dropped when it leaves, never reused across a map-server restart (the id map is per-instance) — and publishes it on **every** `EntryStatus` arm (`warming`/`connected`/`failed`). The client keys every cached per-key owner on `{encodedKey, membershipId}`, so a same-key remove/re-add (a new id) **and** an authority restart (fresh ids for every member) rebuild subscriptions **by construction**. This kills kolu's hand-rolled `createRejoinKeyedSub` generation rearm + its test.

### 2. The typed connection key
`defineSurfaceMap` takes a mount `name`; `connectSurfaceMap` derives the transport slice from `map.name` (no `"padi"` string at the connection site), and the map exposes a typed `surfaceContract` so kolu's server contract splice drops its `as any`. The `"padi"` string sibling key and the contract cast are gone.

### 3. `clockNow` as a framework-reserved member (`@kolu/surface` core)
Mirroring `system.identity`: a reserved `system.clockNow` every surface auto-answers, probed by `makeSession` at the admit handshake (RTT-halved, `sentMs` sampled before the probe). `serveHostMap`'s `offsetOf` caller obligation and `measureClockOffset` delete; the framework now owns the measurement.

**The offset is decoupled from readiness** (a review found the coupled design harmful): readiness = link liveness, so a session is `connected` whether or not the offset has landed. `EntryStatus.connected.clockOffset` is explicitly `number | null` (`null` = one meaning, not-yet-measured); the clock lens renders "—" until it lands. A probe failure is **never swallowed** — it surfaces loudly to connection diagnostics and retries on cadence (the "stuck-connecting" defect class is unspellable, not relocated).

**padi's frozen `control.core.clockNow` is KEPT — forever.** It is a member of padi's never-versioning "control core" (the skew-crossing channel old binders drive `hello→clockNow→decide` across); the frozen-core law outranks the plan's "delete it" line, so the reserved `system.clockNow` lands **beside** it as the new measurement path, not as a replacement. The plan of record is corrected in this PR (padi.mdx, surface-runtime-boundary.mdx).

## Behavior-neutrality — one deliberate deviation
This PR is byte-identical on screen with **one** deliberate exception, called out for the merge decision: decoupling the offset from readiness means the connected (green) chip transitions **~1 RTT earlier** than before (the offset now lands just after connect rather than gating it). Same pixels, earlier transition — not a rendered-content change. Everything else is neutral (e2e + seal).

## Proven by deletion
- `packages/client/src/host/connectionRearm.ts` (+ test) — gone.
- The `"padi"` string key + **both** `as any` splices — the **contract** cast (`server/src/surface.ts`) **and** the **router** cast (`server/src/index.ts`) — gone.
- `measureClockOffset.ts`, `offsetOf`, the per-binding `clockOffset()` — gone (the framework measures the offset at admit). padi's frozen `control.core.clockNow` is **kept forever** (see rider 3).

The no-splice negative property is now **wholly PR3's** — contract cast, string keys, *and* router cast. Both casts are unspellable by type, not merely deleted: `SurfaceMap.surfaceContract` (contract side) and the narrowed `ServeSurfaceMapResult.router: { surface: … }` (router side) let every splice read cast-free, so a future consumer cannot re-spell the cast without the compiler objecting.

## Done-criteria (landed as pins, `mapHarness.test.ts` → "membership is time")
- Every published arm carries a `membershipId`.
- A same-key remove/re-add mints a **different** id **and** rebuilds the subscription (the `createRejoinKeyedSub` scenario — proven via the new session's fresh upstream forward).
- The id is **stable** across a status flip (no spurious rebuild storm).
- An authority restart (fresh `serveSurfaceMap` over the same registry) never re-mints a prior id.

## Design philosophy
- **Fail-fast / no fallbacks.** No `offsetOf` override knob resurrected; a probe failure keeps the last-known/`null`, never fabricates a `0`. The pre-frame synthesized warming carries an empty `membershipId` (transient, never used for keying) — the arm's required field made total, not a silent default.
- **Reuse the source of truth.** `clockNow` follows the shipped `system.identity`/`system.live` reserved-member pattern rather than a parallel mechanism; the typed key derives from the one declaration.
- **Electricity.** Every move completes an existing `@kolu/*` package (surface, surface-map, surface-remote) — no new package.

## Post-merge composition with PR 1 (#1805, now on master)
#1805 (supervised `SurfaceRuntime`) merged to master first, so this branch **rebased onto it** (merge `2a729c7ce`). Composition verified: my `clockNow` auto-serve composes with #1805's `starts` return; my `surfaceContract` contract splice survived cast-free; `implementSurface` losing its `channel` dep was absorbed (the `clockNow` test now uses the runtime-owned channel). #1805 shipped **without** killing the `index.ts` router cast (its Option-B deferral premise was that it would), so — the same-lines conflict having evaporated — the router cast **reverts to PR3's scope** per the plan row ("the `as any` contract/**router** splices") and is killed here. No joint clause remains.

## Evidence class
Behavior-neutral on screen — the evidence class is the **e2e suite + the seal test** (same ruling as PR 4). No reachable render change: the `membershipId` is framework identity no display consumer reads, the clock offset is measured the same way and folded the same way, and the connection cell renders identically (the re-key replaces the manual rearm invisibly).

## Paired drishti PR
[srid/drishti#99](https://github.com/srid/drishti/pull/99) — pinned to this PR's final HEAD (61640b055), typecheck + tests green.


srid:
## Review gauntlet — SRT-PR3

Full `/be` gauntlet ran on this branch. Summary of each reviewer and its dispositions.

### architecture-first-principles — clean
C1–C7 hunter pass (adversarially verified): 7 checks, **0 findings survived verification**. Depth is right — the change deletes app-level machinery by generalizing the framework mechanism (the `connectionRearm` rearm → `membershipId` re-key by construction; the `offsetOf` injection → framework-measured offset).

### /lens-debate (lowy ⇄ hickey) — consensus, 3 fixes applied
- **RTT compensation restored** — the initial clockNow rider had dropped the round-trip-halving `measureClockOffset` did, biasing the offset ~½ RTT and leaving a false "same instant" docstring. Re-encapsulated as `measureSurfaceClockOffset` (samples `sentMs` before the probe). Behavior-neutrality restored.
- **Nested `Map<enc, Map<membershipId, client>>`** replacing a NUL-delimited string cache key.
- **`PADI_SURFACE_NAME`** single-sourcing the `"padi"` mount name (both splice sites + the map declaration key on the one const).

### /codex-debate (xhigh) — 11 findings; 10 resolved, 1 adjudicated
Two passes; the debate surfaced 11 findings. The clockNow rider's **F1/F2/F3** (a rolling-upgrade break: deleting padi's *frozen* `control.core.clockNow`, and gating readiness on the offset so an old peer sticks warming) were escalated to the plan owner and **resolved per an authoritative ruling** — the frozen-core member is kept forever, the offset is decoupled from readiness (`clockOffset: number | null`), and a probe failure surfaces loudly + retries rather than stuck-silent. **F4–F6, F8–F11** fixed in-debate (notably the **F5 fail-loud down-seam belt** and **F9 error-severity logging** — both genuinely worth catching). **F7** (a fully-general mount helper) **adjudicated-drop**: the `PADI_SURFACE_NAME` single-sourcing already prevents declaration/mount divergence at every real site, and a general helper wouldn't cover the router-splice site — a separate ergonomics improvement, not this consolidation PR's scope.

### /simplify — 3 fixes applied, 6 recorded
Applied: removed a redundant `readAll` id-prune; extracted the shared `probeStale` clock-probe guard; and — the one real efficiency catch — **stopped retrying a structurally-absent probe** (an old peer with no `system.clockNow`) that could never resolve, rather than polling a dead member every 10s forever. Recorded as fast-follows (out of this PR's scope): a `pollReservedMember` generalization (would change pre-existing `pollIdentity`), a `pruneToMembers` graduation across packages, and three bounded micro-costs.

### code-police — 5 findings; 2 fixed, 3 recorded
- **Fixed** — the `expectedAbsent` classification was a fragile error-message string-match that could misfile a genuine transient failure as permanent and silently stop probing: now a **typed `ClockNowUnavailableError`** (structural navigation) + oRPC `NOT_FOUND`, no message matching, with a new test pinning the branch. Expected-absent now logs at `debug` (the "tool not installed" analogue), not `info`.
- **Recorded** — `dry-rule-of-three` (clockNow is the 3rd reserved procedure; a generic typed `reservedProcedure` factory refactoring the two stable shipped modules `liveness.ts`/`identity.ts` is a focused follow-up, not this PR's scope); `no-overloaded-null` / `caught-error-must-not-collapse-to-empty` (the `clockOffset: null` → wrong terminal ages via the *pre-existing, untouched* `useTerminalMetadata.ts` — the exact consuming-piece gap the plan already documents at `padi.mdx:317` and the coordinator ruled the null-graceful design; narrowed further by the genuine-failure-retries fix so only the documented skew window renders null).

**Evidence class:** behavior-neutral consolidation with **one** deliberate deviation stated in the PR body (the connected chip transitions ~1 RTT earlier now that readiness is decoupled from the offset — same pixels, earlier transition — for srid's merge call). Full workspace typecheck green; all touched-package tests pass.

PR #1819, opened 2026-07-14
Adopt an externally-created split like a manual one

srid:
**A split created from outside the browser now opens and focuses like one you made yourself.** Spawning a split from the shell (`padi-tui create --parent <id>`) or from any other client reaches the daemon via `lifecycle.create({parentId})` and lands the new terminal on the collection — but nothing ran the two *browser-local* steps the manual split path (`useTerminalCrud.handleCreateSubTerminal`) does: **expand the parent's sub-panel** and **select the new tab**. So the panel kept its prior state and the new pane never painted (its `visible` gate is `activeSubTab() === subId`, and no one set it). The split appeared to do nothing until you found its tab and clicked it.

### The gap

The daemon-side create is identical for both paths; the difference is entirely client-local UI state that padi-tui can't reach:

```
manual (browser):   lifecycle.create({parentId})  +  expandPanel  +  setActiveSubTab
padi-tui / other:   lifecycle.create({parentId})     ← the two UI steps never run
```

Rather than have padi-tui reach into the browser, the client now **reacts to the arrival on the list** — so a split from *any* actor is adopted the same way.

### The fix

New `useAdoptNewSplit` — the **arrival mirror of `useActiveReconcile`** (which handles departures). It diffs a host-tagged, sub-only snapshot of the terminals collection and, on a genuinely-new split, runs the adopt steps:

- **Always expand** the parent's panel (like a manual create).
- **Select the new tab only when the parent has no active split** — *expand-but-don't-steal*: if you're already working in another split of that terminal, the new one opens as an unselected tab in the now-open panel rather than yanking your focus.

Wired in `useTerminals` beside the reconcile, sharing its host-scoping discipline.

### Why it doesn't fight restore

The restore `phase` is *deliberately non-reactive* (`createSessionRestore`), so it's **sampled, never tracked**. The baseline can't lean on reacting to the seed transition, so every seen sub is recorded regardless of phase — a sub present before the host seeds is baseline, never a false arrival, and only a split that appears *while seeded* is adopted. Hydration keeps ownership of the tabs during load/restore/recycle; a host switch rebaselines so a host's existing splits aren't mass-adopted.

### Design philosophy

- **Reuse the source of truth** — reacts to the terminals-collection delta (the same list the reconcile diffs), not a parallel padi-tui→browser side channel; reuses the `sameParentSnapshot` comparator (widened to `ReadonlyMap` so a never-null sub map shares one home).
- **Fail-fast** — no override knob, no fallback path; a live, host-scoped reaction gated on the restore seed.
- **Electricity** — a leaf reaction inside the client terminal module, sibling to `useActiveReconcile`; no domain-agnostic capability to extract.

Covered by `useAdoptNewSplit.test.ts` (6 cases: adopt-and-select, don't-steal, load-then-live, top-level-ignored, mount-baseline, host-switch-rebaseline). The docs already promised this behavior (`ln create --parent` → "a split tile that appears live on the canvas"; terminal-teams "tile identical to browser-created") — this closes the gap to match.


srid:

✅ **Consensus** after 2 round(s) · lowy + hickey · base `72c45ffe6387`

Independent findings: lowy=6, hickey=4

### Applied (6)
- `lowy-2` Arrival detection hand-maintains `seen`/`seenHost`; the sibling gets the same fact free from `on`'s prev/curr — commit `d1d90f04e`
- `lowy-3` Host-scoped parent-snapshot memo is built twice with identical gate/seed boilerplate — commit `ac02cdf0b`
- `lowy-4` Triplicated list-id / parentId wiring in useTerminals ignores the existing `allTerminalIds` memo — commit `ba8470a93`
- `lowy-5` Adopt now depends laterally on its departure sibling just for a pure comparator — (uncommitted)
- `hickey-1` Hand-rolled `seen` Set + `seenHost` reimplements the sibling's prev-snapshot diff — (uncommitted)
- `hickey-2` The host-tagged parent-snapshot memo is built twice, and `listSub().map(id)` three times — (uncommitted)

### Agreed — no change (4)
- `lowy-1` Electricity test: receptacle, wires, and the app-concern verdict (packages/client/src/terminal/useAdoptNewSplit.ts:33)
- `lowy-6` Install-order coupling with useSessionRestore is the exposed wire — comment-guarded only (packages/client/src/terminal/useTerminals.ts:119)
- `hickey-3` "Reveal a newly-created split" (expand + select) is open-coded across three paths (packages/client/src/terminal/useTerminalCrud.ts:230)
- `hickey-4` Adopt's correctness is braided with its install position relative to useSessionRestore (packages/client/src/terminal/useTerminals.ts:125)

srid:
## Codex ⇄ Claude debate

✅ Consensus in 2 rounds · reviewer effort: xhigh · base: `origin/master`

|---|---|---|
| 1 | 2bc135119 | test(terminal): guard adopt×hydration seed-boundary invariant (codex F1) |

- **F1** — The fix relies on useAdoptNewSplit being created before useSessionRestore so that, on the metadata flush which seeds hydration, adoption samples the pre-seed phase

srid:

**Scope:** `git diff 72c45ffe...HEAD` (7 files, +508/-40)

| Pass | Findings |
| --- | --- |
| Rules (7 built-in + 24 project) | ✅ 0 |
| Fact-check (error-swallowing · fallbacks · logic · unbounded growth) | ✅ 0 |
| Elegance | — skipped (`/simplify` ran earlier in the gauntlet) |

**All clear** — nothing to fix. Spot-checked the two load-bearing claims: `?? "pending"` is `HydrationPhase`'s documented initial member (not a masking default), and `useSubPanel()` is a true module-level singleton (8+ call sites, one `createStore`). The adopt-vs-hydration seed-boundary ordering is covered by the "composed with hydration" regression test.

srid:
## Evidence

Recorded via the e2e harness (`packages/tests/features/sub-terminal.feature`), driving `padi-tui create --parent $KAVAL_TERMINAL_ID` from inside the main terminal — i.e. a split created from *outside* the browser (as `padi-tui`/agents do), not via the command palette. The clip shows the sub-panel auto-expanding and the new split becoming the selected tab the moment the split lands, with no manual click needed.



▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=external-split-auto-expand.mp4

PR #1822, opened 2026-07-14
SR5 — one protocol across the wire: delta-aware mirror, supervised re-serve, extendSurface

srid:
A collection should stay **one protocol when it crosses a network**. Drishti's process fact is today both a Surface collection *and* a parallel snapshot/delta stream, purely because the remote mirror fans a collection into `keys` + one `get` stream per key. SR5 teaches the mirror to speak a collection's declared `deltas` verb directly, completes `reServeSurface` into a framework-supervised runtime, makes relay loss cross oRPC as a *named retryable* end, and adds `extendSurface` so a parent can keep a local member without an inert agent-side stub. Pure consolidation — **byte-identical on screen**, proven by deletion.

This is the SR5 row of the [surface-runtime-boundary](../blob/sr5-wire/docs/atlas/src/content/atlas/surface-runtime-boundary.mdx) plan of record.

### What lands, each proven by a deletion

| Move | What changes | The deletion that proves it |
| --- | --- | --- |
| **Delta-aware mirror** | `mirrorRemoteSurface` folds a `deltas`-declaring collection's single batched snapshot-then-delta stream into the same `{upsert,remove}` sink (spec-driven, never a link probe) | the per-key `keys`+`get` fan-out for delta-declaring collections |
| **Supervised re-serve** | `reServeSurface` returns a framework-supervised runtime via the new `superviseTerminalSource` combinator | the hand-rolled `done`/`close` it carried since SR1 ("deferred to PR5") |
| **Named retryable relay end** | `RelayTransportLostError` (an `ORPCError` with the retryable `SURFACE_RELAY_TRANSPORT_LOST` code) ends a fail-through relay on upstream loss; the shared retry fence opts that one code back in | the raw re-throw that crossed oRPC *sanitized to a non-retriable error* and stranded the downstream; and `NoLiveUpstreamError` |
| **`extendSurface`** | composes a parent-local runtime onto a re-served one, flat under one `surface` namespace, wire-matchable | (drishti) the inert agent-side `metricHistory` stub |
| **Ride-alongs** | one typed `surfaceMember` nav helper; one `inMemoryCollection` (the collection twin of `inMemoryStore`) | the duplicated opaque-client casts and the hand-rolled additive-collection shape |

### Design philosophy

- **Fail-fast, no fallbacks** — a delta-declaring collection whose client lacks the `deltas` verb is a loud client/surface mismatch, never a silent fall back to per-key; a member-name collision in `extendSurface` throws; the relay end is *retryable-by-name*, not a degraded silent stream.
- **Electricity** — every move completes an existing `@kolu/*` package (`@kolu/surface`, `@kolu/surface-remote`); no new package, no domain helper folded into an app module.
- **Reuse the source of truth** — the mirror's deltas routing reads `collectionHasDeltas(spec)`, the same gate `walkSurface` and `surfaceClient.use()` decide by; `extendSurface`'s wire-matchable splice reuses `buildAppRouter`'s `implement(combined).router({...})` pattern (pinned by a `StandardRPCMatcher`-tree test, the SR1 lesson).

> **Seam dependency — FLK-STDIO (in flight, same seam).** FLK-STDIO is root-causing the stdio close-vs-reconnect residual and its fix touches `surface-remote/relayStream.ts` `failThroughStreamCore` — the *same function* this PR's named-relay-end change edits. My change **keeps the re-throw**, only changing the error *type* (raw → named `RelayTransportLostError`, raw error preserved on `.cause`) — orthogonal to FLK-STDIO's abandoned-pull ownership fix (who *awaits* the pull, not *what* is thrown). Merge order is the coordinator's; green-first-merges, the other rebases. **SR7** (reactor bridge) is also in flight but disjoint.

> **Drishti pair (gate) — [srid/drishti#100](https://github.com/srid/drishti/pull/100).** Per `.claude/rules/surface.md`, the paired PR — pinned to this PR's final HEAD (`6eb25db39`) — carries the deletion-proof (the `processesSnapshot` stream, its parent reducer, and the inert `metricHistory` stub). Typecheck green + 166 unit tests pass; two-platform CI + byte-identical e2e in the ship phase.

> **⚠️ Post-gauntlet completeness addition — please review explicitly: `BoundCollection.unenrolledDeltas` ([`6eb25db39`](https://github.com/juspay/kolu/commit/6eb25db3916589e53b54b99f2fd6cd048bcf2230)).** The full review gauntlet (architecture → lens → codex → simplify → police) ran and settled on the commits **below** this one; `unenrolledDeltas` landed **after** it. It is not new design — it is SR5's scope *completed by its proving consumer*: "one protocol across the wire" is incomplete if the batched `deltas` protocol lacks the **un-enrolled reach** the `keys` protocol got in SR2 (`unenrolledKeys`), which leaves the [#1591](https://github.com/juspay/kolu/issues/1591) no-health carve-out class *unexpressible* in the deltas form. Drishti's process table (a deliberate no-per-host-health carve-out) is exactly that consumer.
> - It is an **exact mirror** of the gauntleted `unenrolledKeys`: a `UnenrolledDeltas<K,T>` interface, gated in `BoundCollectionsFor` on the declared `deltas` verb, a lazy `get unenrolledDeltas() { return ns.deltas }` — nothing `unenrolledKeys` didn't do. Its **RED/GREEN type pin** is in `boundCollection.test-d.ts` (positive: a `deltas` collection types it; negative: a non-`deltas` collection does not); documented in `ref-surface`.
> - The **enrolled `.use()` alternative was rejected** on the byte-identical invariant: it would enrol the process table into a per-host health fact it deliberately does *not* have — a constructed behavior change riding a consolidation.

_Draft: kolu side complete + fully gauntleted + green (`just check`; 357 surface + surface-remote unit tests); drishti pair [#100](https://github.com/srid/drishti/pull/100); two-platform CI on both repos + byte-identical evidence to follow._


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `3104ecb0a67f`

### Applied (3)
- `lowy-2` superviseTerminalSource's terminal interface is shaped around reServeSurface's sync-abort pump, so extendSurface (same-PR second consumer) exposes the wires and re-hand-rolls the doctrine — commit `31ae4008f`
- `hickey-1` Terminal+passive `done`-fold duplicated; `superviseTerminalSource` bypassed by its own declared consumer — commit `31ae4008f`
- `hickey-2` SinkError-guarded `initialKeys` seeding duplicated verbatim across the two collection bridges — commit `503926345`

### Agreed — no change (5)
- `lowy-1` Electricity test (required lens output) (packages/surface/src/server.ts:1555)
- `lowy-3` Flat-name-per-name uniqueness invariant lives in mergeSurfaceSpecs, not in defineSurface which actually folds the flat namespace (packages/surface/src/server.ts:2350)
- `lowy-4` extendSurface.test's buildBase hand-rolls the exact three-line Map that inMemoryCollection now encapsulates (packages/surface/src/extendSurface.test.ts:39)
- `hickey-3` `inMemoryCollection` return type re-spells the collection-deps core instead of naming it (packages/surface/src/server.ts:660)
- `hickey-4` Flat-namespace collision invariant enforced at two granularities (mergeSurfaceSpecs vs defineSurface.claim) (packages/surface/src/server.ts:2400 (mergeSurfaceSpecs))

srid:
## Codex ⇄ Claude debate

✅ Consensus in 2 round(s) · reviewer effort: xhigh · base: master

|---|---|---|
| 1 | dc1287b1a | fix(codex): classify relay transport-loss, deep-merge composed system, materialize spec kinds (SR5) |

- **F1** — The catch block converts every non-abort rejection from the upstream call or iterator into RelayTransportLostError
- **F2** — The new fail-fast check for a missing deltas verb works only with the plain-object test client
- **F3** — The shallow namespace splice drops an extension's app-owned system member
- **F4** — ComposedSurfaceSpec normalizes every absent member kind to a required empty record, but mergeSurfaceSpecs omits a kind when both sides omit it

srid:

Ran with `--no-elegance` (the `/simplify` track already covered the tree). Pass 1 (rule checklist) + Pass 2 (fact-check).

**Actioned:**

| Finding | Pass | Disposition |
| --- | --- | --- |
| A relay member the live client doesn't expose made `select(client).get` a `TypeError` before any I/O — `isMiddleHopTransportLoss`'s "non-`ORPCError` ⇒ transport loss" rule wrapped it as the **retryable** relay end → retried forever behind a misleading "link died" log | fact-check | **Fixed** — `requireUpstream` resolves the member *outside* the transport try in both relay cores, so a structural mismatch fails loud (non-retriable on the wire); test added (`18c8a5c5b`) |
| `extendSurface.test`'s `buildBase` re-hand-rolled the `{readAll, upsert, remove}` Map triple `inMemoryCollection()` was extracted (this PR) to eliminate | rules · dry-rule-of-three | **Fixed** — uses `inMemoryCollection` (`18c8a5c5b`) |
| `isDeadTransportError` + `SURFACE_RELAY_TRANSPORT_LOST` (new `@kolu/surface/client` exports) undocumented on `ref-surface` | rules · surface-reference | **Fixed** — added a retry-contract table to `ref-surface.mdx` (`18c8a5c5b`) |

**Disposed without change:**

- **prefer-ts-pattern** (`mirrorCollectionDeltas`'s `if (msg.kind === …) else`) — *disputed*: `ts-pattern` is not a `@kolu/surface` dependency and the sibling `foldCollectionDeltas` dispatches the identical union with `if/else`; matching the established pattern.
- **errors-must-log-at-error** (the transport-death line lands at the caller-wired `log.debug`) — *recorded*: the relay's single diagnostic `log` callback is caller-leveled; the transport death *is* logged and now distinctly recognized. Elevating its level is a caller-wiring change outside this byte-identical consolidation — a follow-up candidate.
- **surface.md drishti gate** — *deferred to §ship*: the paired [drishti](https://github.com/srid/drishti) PR lands next, pinned to this PR's final HEAD with its own green two-platform CI (not a blocking code finding).

Clean on every other rule (invalid-states, no-silent-swallowing, no-dead-code, no-overloaded-null, no-unbounded-growth, …).

PR #1823, opened 2026-07-14
SR7 — the reactive bridge's `$` face; padi urgency becomes derived

srid:
## SR7 — the reactive bridge's `$` face

This is **SR7** of [the surface-runtime consolidation](https://kolu.dev/atlas/surface-runtime-boundary.html) — the reactive bridge's derivation glue, per [the bridge note's API section](https://kolu.dev/atlas/surface-reactive-bridge.html) exactly (the vocabulary is ratified; no synonyms minted). It lands the four SR7 primitives and proves them by **worked example 1**: padi urgency becomes a derived member and the convention-publish seam it rode dies.

### What lands

- **`$` — the typed sibling-read face.** A plain mapped type over the surface spec (`SiblingRead<S>` in `@kolu/surface/define`): `$.someCell(): T`, `$.someCollection(): ReadonlyMap<K, T>`. Typed at the `derived.cell(($) => …)` declaration site by the deps slot — no `keyof` union explosion.
- **`computed(fn)`** — a derived value reading other graph nodes (a private intermediate several wire members can share); glitch-free by the engine's version-checked lazy pull.
- **`batch(fn)`** — the one knob an app reaches for to coalesce a multi-member burst of ctx writes into a single recompute pass.
- **The compute-fn `derived.cell(($) => …)` overload** — publish a sibling derivation as a cell, beside the existing graph-node `derived.cell(node)`.
- **Post-equals mirror pokes, as the structural fix.** An authored member's graph face is a mirror poked **post-spec-equals** by a **bridge-owned store wrapper both cell write paths pass through** (`applyAndPublish` and the ctx `set` both land through the one wrapped `store.set`, after the equals gate). Collections version-poke on the wrapped publishers. A missed poke is **unwritable by construction** — open-question-2's decided structural fix, not a two-site rider held by pinning tests.

### Proven by deletion — worked example 1

padi `urgency` becomes `derived.cell(($) => recomputeUrgency($.terminals()))`. Deleted: `publishUrgency`, its rider on every `publishComposedTerminal`, the `dropSnapshot` refold, and the prose invariant — a registry writer can no longer forget urgency because the `terminals → urgency` edge is **tracked, not conventional**. `recomputeUrgency` now folds the composed `terminals` map (its parameter) instead of the raw registry; `urgencyEqual` moves beside its schema so the spec declares the one wire-dedup point.

### Design philosophy

- **Fail-fast, no fallbacks.** The structural store wrapper makes a missed poke *unrepresentable* rather than a convention. A compute cell's node build/seed/connect fail loud (used-before-bind, unknown-sibling, connect-after-dispose all throw); a seed pull that throws is a boot crash (mirror-never-fabricate), never a fabricated default. A compute recompute that throws is log-skip-continue (holds last published) — the stateless-compute error law.
- **Electricity.** No new package — SR7 completes `@kolu/surface`'s reactor, where the volatility already lives. The signals engine stays reachable **only through `reactor.ts`**: the boot walk stays engine-free, bridging each sibling as a plain read+subscribe `SiblingSource` that the reactor wraps in a version signal.
- **Reuse the source of truth.** The `$` type reuses the same `z.infer`-over-spec shape as `SurfaceTypes`; the poke rides the existing `equals → onWrite → store.set → bus.publish` gate (the wire dedup point is unchanged); `equals` lives on the member spec, the one home the laws mandate.

### Tests

Reactor tests pin the **laws as written**: glitch-freedom (a diamond sees both siblings new, never half-updated), `batch` coalescing (one recompute for a two-write burst), change-iff-fired (a suppressed equals-equal write never pokes → no recompute), and the post-equals-mirror-poke → derived-recompute edge — all end-to-end through a real `implementSurface`. Plus `$` sibling typing (a compile-time assertion) and `computed` as a composable node. `urgency.test.ts` is ported to the composed-map fold.

### Honest cost — a tracked REGRESSION, deferred to SR8

Worked example 1 wires `urgency: derived.cell(($) => recomputeUrgency($.terminals()))` off the **composed** `terminals` collection. So every firehose poke re-composes *every* live terminal (`registryMap(composePadiTerminal)`) even though one changed — **O(M) composes per write → O(M²) per ~150 ms firehose cycle**, where pre-SR7 `recomputeUrgency()` folded the raw registry at O(M) *field reads*, zero composition. This is a **regression**, not a hypothetical new cost (M is real: ~16 live terminals on a busy dev host). Both in-scope mitigations are worse trades (folding the raw registry re-couples what the migration decoupled; memoizing `composePadiTerminal` is a new caching layer bigger than SR7), so the fix is the *designed* one: **SR8's keyed reconciler (`derived.collection`) re-composes only the changed key.** It lands as a **named obligation on SR8's row** (its proving case) in [the plan](https://kolu.dev/atlas/surface-runtime-boundary.html), so SR8 cannot ship without resolving it. (`batch()` coalesces frequency, not the per-recompute compose cost, so it does not address this.) Ruled record-and-defer by the campaign coordinator.

### Seam dependencies

SR5 (mirror/re-serve) and FLK-STDIO (stdio teardown) are in flight. This PR's reactor territory is disjoint from both, but its **post-equals mirror pokes touch the publish seam** (the wrapped cell/collection publishers in `walkSurface`). Green-first merges; the other rebases.

### Docs

- [Reactive-bridge note](https://kolu.dev/atlas/surface-reactive-bridge.html): the phase-1 migration prose is aligned to the decided structural store-wrapper fix (it no longer reads as a two-site rider), so the note stops pulling two ways.
- Surface Reference (`ref-surface.mdx`): the reactor table + guarantees now carry `$`/`computed`/`batch`/the compute-fn overload.
- `surface-attention-101.mdx` is a W5-framed pedagogical explainer; its "urgency re-production is phase 1, after W5" statements are left as the historical teaching narrative — the authoritative API record is `ref-surface.mdx`.
- No changelog entry: SR7 is a byte-identical consolidation with no user-visible behavior change (the plan mandates features never ride these PRs).

**Paired drishti PR:** _(backfilled — pinned to this PR's final HEAD, CI green)_


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `3104ecb0a67f`

Independent findings: lowy=3, hickey=8

### Applied (2)
- `lowy-1` The `$` receptacle leaks the walk's eager build-order as a declaration-order contract — commit `078d90d21`
- `hickey-2` Interleaved bind+seed manufactures a declaration-order contract that a two-pass build dissolves — (uncommitted)

### Agreed — no change (9)
- `lowy-2` `SiblingSource` jams two edge-wiring strategies behind one interface via an `engineTracked` boolean + a sometimes-dead `subscribe` (packages/surface/src/reactorBrand.ts:82)
- `lowy-3` Walk-internal `siblingRead()` seam sits on the app-facing `DerivedCell` interface (packages/surface/src/reactor.ts:322)
- `hickey-1` Version-counter signal fragments an authored sibling's value from its reactive edge (packages/surface/src/reactor.ts:485-509)
- `hickey-3` graphNodeCell and computeCell duplicate one derived-cell lifecycle (packages/surface/src/reactor.ts:357)
- `hickey-4` The sibling-change poke is a manual rider at three call sites (packages/surface/src/server.ts:1773)
- `hickey-5` SiblingSource complects two source kinds behind an optional boolean (packages/surface/src/reactorBrand.ts:1217)
- `hickey-6` Derived cells register a change fan-out that can never have a subscriber (packages/surface/src/server.ts:1660)
- `hickey-7` A third store-write path bypasses the "one wrapped store" poke invariant (packages/surface/src/server.ts:1799)
- `hickey-8` New public `computed` export has only test consumers (packages/surface/src/reactor.ts:375)

srid:
## Codex ⇄ Claude debate

✅ Consensus in 3 round(s) · reviewer effort: xhigh · base: `3104ecb0a67f`

|---|---|---|
| 1 | 4874dd425 | fix(codex): round 1 — wrap graph-node compute errors, fail-fast on $ key overlap, pin the collection edge |
| 2 | 08cf5c211 | fix(codex): round 2 — null-prototype sibling dictionaries (F5 regression from the F2 guard) |

- **F1** — The graph-node form `derived.cell(computed(fn))` does not enforce the bridge's stateless compute error policy
- **F2** — The flat `$` namespace is unsound when a surface legally uses the same key for a cell and a collection whose wire verbs do not collide
- **F3** — The only `$` collection test checks compile-time typing and the eager seed; it never mutates the collection
- **F4** — The module header still says phase 0 exports exactly three symbols and that `$`, `computed`, and `batch` are later phases; the scan comment at lines 230-231 likewise calls stateless compute a later phase
- **F5** — The new duplicate-registration guard uses if (siblingSources[key]) on a normal object

srid:

Two passes over the SR7 diff — the rule checklist and the fact-check.

### Applied
- **dry-rule-of-three** — the byte-identical `store.set` "graph-owned (one writer)" throw lived in both `derived.cell` forms → extracted `graphOwnedStore(get)`, one home for the read-only facade + its one-writer guard. (`6eaa01c92`)
- **fact-check · test gap** — the migrated fold gates on the COMPOSED record's `state === "active"` (was the raw registry's `entry.meta.state`), but no test built a non-active arm → added a pin that a SLEEPING terminal whose agent reads `awaiting_user` is EXCLUDED. (`6eaa01c92`)

### Recorded — not fixed (each with its reason)
- **invalid-states-unrepresentable** (`SiblingSource.engineTracked` boolean + a dead `subscribe` → a discriminated union): the lens debate already raised this (lowy-2 / hickey-5) and **dropped it by consensus** — not re-opened here.
- **no-thin-wrapper-functions** (`dropSnapshot` is now a single-call forward): kept as a documented named lifecycle seam, symmetric with `installSnapshot` — the rule's "documentation-carrying named seam deliberately preferred" exception.
- **fact-check · perf** — urgency's fold reads the COMPOSED `terminals` collection, so every firehose poke recomposes every live terminal (pre-SR7 folded the raw registry). A cost the reactive-bridge note anticipates (honest-cost #7 "measure before caring") and whose fix — the keyed reconciler — is **deferred to SR8**; recorded in the PR body rather than fixed against the ratified worked example.

Every other rule (silent-swallow, overloaded-null, unbounded-growth, no-sync-blocking-on-the-serving-loop, caught-error-must-not-collapse, …) passed clean.

srid:
## Evidence

SR7 is a **byte-identical consolidation** (no on-screen change by design — a feature never rides a runtime PR), so the evidence class per the plan is **e2e + seal**, not a screenshot. Two-platform `ci::default` is green on the shipping SHA `56b896245`:

| Lane | `x86_64-linux` (kolu-ci-2) | `aarch64-darwin` (rasam) |
| --- | --- | --- |
| `ci::e2e` | ✅ | ✅ |
| `ci::unit` (incl. the surface/padi seal + `padiBinding` round-trip) | ✅ | ✅ |
| `ci::smoke` · `ci::nix` · `ci::biome` · `ci::fmt` · `ci::atlas-sync` · … | ✅ | ✅ |

The urgency badge behaviour is proven unchanged by the e2e suite (the awaiting/deep-link scenarios) passing on both platforms — urgency is now a **derived** member folded off `$.terminals()`, and its output is identical to the pre-SR7 registry fold (pinned by the surface + padi unit suites, including the sleeping-terminal exclusion). One `unit@x86_64-linux` first-red was the documented `padiBinding` stdio close-vs-reconnect flake (8th sighting, disjoint from this change), cleared by a same-SHA node rerun.

Issue #1825, opened 2026-07-14
xterm WebGL: intermittent glyph-atlas garble on tile show (whole-tile / partial-bottom), real-GPU only


srid:
# xterm WebGL — glyph corruption on tile switch (shared atlas leak), real-GPU only

**Status:** OPEN · root cause reproduced + grounded (shared-atlas consume-once page-merge
invalidation). **Fix = an UPSTREAM dependency bump, not custom code:** xterm.js PR #6042
(per-`GlyphRenderer` pageLayoutVersion). Shipped as **PR #1830** (`@xterm` beta.289/288).
Before/after proven at the mechanism level; contract suites green; nix build confirmed;
pending linux CI + srid's real-GPU pixel verify. #1824 reverted (#1828), #1826 closed.
Living record.

---

## 1. Symptom

- **Cross-tile glyph leak (the defining observation):** switching the active tile (A → B),
  **characters from the previous active terminal leak into the new one** — B renders glyphs
  from A's slots. A whole-tile / partial per-glyph substitution (`background` → `tackground`).
- Clears on selecting the text, a **window resize**, or a page reload — never on its own.
- **Intermittent.** Trigger: a **busy split + maximized / canvas** workspace with two
  recently-active tiles; switching focus between them.
- **Real-GPU only.** Reproduces on real hardware Chrome. **Headless software-GL (SwiftShader —
  CI/pu boxes) is a FALSE-POSITIVE venue** for this bug class (it fabricates corruption on
  synthetic triggers and can't reproduce the real timing). srid reproduces on real hardware and
  is the verification venue.

## 2. Corrected mechanism — shared atlas across concurrently-live tiles (LEAD)

Grounded at the tree (master):

1. **Two tiles hold live WebGL at once.** #1404 (`ded98e83c`) — the **recency budget** — keeps
   `WebglAddon` **live on the two most-recently-active tiles** (`Terminal.tsx:147-151`, "keeps
   WebGL on both sides of an A↔B switch"). So A and B each hold a live renderer simultaneously.
2. **xterm shares ONE `TextureAtlas` across equal-config terminals.** `@xterm/addon-webgl`
   `acquireTextureAtlas` keys the atlas by font config and keeps an **`ownedBy` list** — a
   matching-config terminal pushes itself onto the *existing* atlas and gets the **same GPU
   texture**. All kolu tiles share identical font config → **they share one atlas**.
3. **The atlas is wiped only on browser-tab re-show, never on an in-app tile switch.**
[...]

PR #1826, opened 2026-07-14
fix(xterm-kit): rebuild the on-show glyph atlas only after the tile is laid out

srid:
Fix-forward completing #1824 (do **not** revert #1824 — this is additive). Tracked in #1825.

## The gap #1824 left

#1824 rebuilds the glyph atlas on the per-tile visible transition, in a **single
post-layout `requestAnimationFrame`** — which fixed the common switch case. But on a
**fast switch** that rAF can itself fire **before** the `display:none → visible` reflow
has flushed, while the tile still measures **0×0**. At 0×0, `clearTextureAtlas` re-warms
nothing (a no-op), so the whole atlas stays stale and the **whole tile renders garbled**
until streaming slowly heals it top-down — srid's decisive real-GPU observation of the
intermittent tail.

## The guard

`<Xterm>`'s `visible` effect now **retries per frame until the tile measures
`offsetWidth/Height > 0`**, then `fit()` + `clearTextureAtlas` — so the rebuild always
runs at real size. Bounded (~30 frames) so a genuinely 0-sized/off-screen tile can't
spin; `onCleanup`-cancelled; post-dispose guarded. **Measured, not a blind second rAF.**

## Pin (RED/GREEN, deterministic)

A second scenario in `webgl_atlas_on_show.feature` pins the pane to **0×0 across the
sub-tab switch** and asserts the rebuild is **deferred until real size returns**. Proven
**RED** on the guardless single-rAF (`clearTextureAtlas` fires while zero-sized → the
probe trips) and **GREEN** with the guard. The original scenario (rebuild fires on a
plain switch) stays green on both.

## Notes

- **Not a #1808 regression** — a longstanding visibility-transition gap; #1808 stays
  exonerated (verdict retained in the xterm-kit Atlas note).
- **Real-GPU-only bug class.** Software-GL (SwiftShader / headless CI) is a
  false-positive venue — it can't surface the intermittent 0×0 timing. See #1825.
- **Verification:** srid re-tests on real GPU (switch tiles repeatedly, must be
  clean-and-consistent) before the review gauntlet resumes and CI runs.


srid:
Superseded — the on-show **visibility** path is not the observed failure. The real bug is the **shared WebGL atlas on the in-app tile-focus switch** (canvas mode, both tiles already `visible`, so no visibility transition ever fires): #1404's recency budget keeps WebGL live on both tiles, xterm's `acquireTextureAtlas` shares one atlas across equal-config terminals (`ownedBy`), and `clearTextureAtlas` fires only on browser-tab visibility — so the newly-focused tile inherits the previous tile's glyph slots ("previous terminal's characters leak into the new one").

Tracking in #1825; closing to redirect effort. #1824 is being reverted (#1828) — its on-show rebuild is a pre-layout no-op today anyway. The measured-size guard design is preserved in #1825 in case the on-show visibility path is ever independently reproduced (e.g. a mobile visibility switch).

PR #1829, opened 2026-07-15
SR6 — adopt before you mint: the first-frame adoption sweep

srid:
**SR6 is an adoption sweep, not an invention** — reach the existing primitive at every site that hand-rolled the same thing, no new framework API. Grounded on current master, **most of the plan's named sites were already adopted or are deliberate divergences** (recorded below, not force-adopted — "most sites already adopted" is the success condition for an adoption sweep). Two production sites genuinely still hand-advanced an iterator to read a one-shot first frame; both now ride `@kolu/surface/first-frame`, and an AST guard pins the deleted spelling so it can't drift back.

### The adoptions (2)

Both replace an open-coded `for await … return frame` with `firstFrameOrUndefined`, behavior-neutrally — the primitive's iteration and iterator-close match the old return-out-of-loop, and empty stays a benign absence the *caller* classifies:

- `kaval-tui` `readExitCode` (`attach.ts`) — over kaval's yields-exactly-once `exit` stream; an empty stream now surfaces as an `error` outcome (a kaval contract violation) instead of a fabricated `exit 0` (code-police).
- `server` `readPadiMemoryOnce` (`index.ts`) — over `processMemory`; the caller maps the primitive's `undefined`-on-empty to its typed `PADI_MEMORY_READ_ERROR` sentinel + distinct `log.error` (added in the lens debate — the primitive *can* serve this site, my original allow-list reasoning was wrong).

> **One deliberate non-adoption:** `kaval-tui` `consumeExit` (`wait.ts`) stays open-coded. Its `settle(gone)` side effect must fire the instant the first `exit` frame arrives, **before** the iterator's async close is awaited — because it races `consumeOutput` + the timeout in a `Promise.all` and `settle` is first-wins. `firstFrameOrUndefined` awaits `AsyncIteratorClose` before it resolves, which would move `settle` past that close and let a competing event win the race (caught by codex, F1). It carries a `first-frame-guard:allow` marker.

### The guard

`server/src/firstFrameOneShotGuard.test.ts` (AST via `@babel/parser`, following the SR2 `procedureCastGuard` convention) fails if `kaval-tui/src` or `server/src` open-codes a one-shot `for await (…)` — a loop with a top-level `return`/`break` and **no** `continue` that can re-target it (unlabeled-and-unnested, or a label naming the loop, across nested-loop / function / object-method / static-block scopes — hardened over the codex debate). It's honestly a **spelling fence**: it pins the deleted syntax, not every conceivable one-shot advance. The one allow-list entry (`consumeExit`) needs a non-empty, directly-attached reason.

### The grounded verdict for the other named primitives — recorded, not force-adopted

| primitive | ground truth on master | verdict |
| --- | --- | --- |
| `acceptSurfaceSocket` | already the sole socket-acceptance site (`server/index.ts:1008`) | no-op |
| `presentingDown` | already read off the model (`client/rpc/TransportOverlay.tsx`) | no-op |
| `setAttention` | kolu's `useHostAttention.paintBadge` is a richer per-host fold (async ordering + ack-dedup, **no** `document.title` write); adopting the model's `setAttention` would add a `(N)` tab-title prefix — a byte-identical violation | **stays** |
| `surfaceClientsHealth` | kolu deliberately abstains (`wire.ts` house style); drishti folds via it, no drishti tree here | no-op |

### Notes for reviewers

- **Consumer-side only — no `@kolu/surface*` API surface touched → no drishti pair.**
- **Design philosophy:** this PR *is* "reuse the existing source of truth" — adopt the primitive over a parallel hand-roll; the primitive lives in `@kolu/surface`, arrow pointing out (electricity); no new fallback (the police fix removed one — the fabricated `exit 0`), and the guard fails loud on re-open-coding.
- Review gauntlet: lens-debate (consensus, 1 round) → codex-debate (consensus, 5 rounds, xhigh) → simplify → code-police. See the PR comments.
- Plan of record: `docs/atlas/src/content/atlas/surface-runtime-boundary.mdx`, row SR6.


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `af86de57f5b7`

### Applied (4)
- `lowy-1` readPadiMemoryOnce still exposes the wires; firstFrameOrUndefined already serves it — commit `899c7337e`
- `lowy-2` Guard grew an allow-list escape hatch whose only live user the receptacle serves — commit `0eba60ddf`
- `lowy-3` Guard pins one SYNTAX of the one-shot advance, not the volatility it names — commit `0eba60ddf`
- `hickey-2` Guard pins one syntactic spelling, not the stated property — commit `0eba60ddf`

### Agreed — no change (4)
- `hickey-1` AST-walk scaffolding is a 6th hand-rolled copy of one concept (matches the repo's one-walk-per-guard convention)
- `hickey-3` Adoption policy fragmented across AST rule + magic-comment regex
- `hickey-4` consumeExit's `!== undefined` leans on a frame-is-never-undefined invariant *(mooted by the codex round — consumeExit was reverted to its original loop; see below)*
- `hickey-5` Primitive and its adoption-guard live in different packages (deliberate: `@babel/parser` is a `server` dep; adding it to kaval-tui would churn the lockfile + nix FOD hash)

srid:
## Codex ⇄ Claude debate

✅ Consensus in 5 rounds · reviewer effort: xhigh · base `af86de57f5b7`

|---|---|---|
| 1 | e9c76dd23 | revert consumeExit non-adoption; guard spares continue-guarded loops |
| 2 | 75515fc8f | guard recognizes labeled self-continue; allow-list requires a real, attached reason |
| 3 | ea67b213a | guard handles label chains + function boundaries; allow-list reason must start with a word char |
| 4 | 7f9d6ce6c | treat object/class methods + static blocks as continue-scope boundaries |

- **F1** — `consumeExit`'s `firstFrameOrUndefined` awaited AsyncIteratorClose before resolving, moving `settle(gone)` past it and changing the `Promise.all` race (settle is first-wins). Reverted to a documented ordering-sensitive non-adoption.
- **F2** — the guard flagged multi-frame loops a `continue` can skip — unlabeled, labeled, label-chained (`outer: inner:`), and across nested-loop / function / object-method / static-block scopes. A false-positive, now precisely excluded.
- **F3** — the Atlas SR6 row omitted the server `readPadiMemoryOnce` adoption.
- **F4** — the `attach.ts` comment miscalled `firstFrameOrUndefined` a "non-empty contract".
- **F5** — the allow-list marker accepted an empty reason, a blank-separated comment block, or a block-comment closing `*/` as the "reason".

srid:

`--no-elegance` (Pass 3 skipped; /simplify already ran this gauntlet). Passes 1 (rules) + 2 (fact-check) ran over `af86de57f…HEAD`.

**One finding — actioned.**

- **`readExitCode` silently returned `0` on an empty `exit` stream** (both passes converged): indistinguishable from a genuine `exit 0` and untraced, while its sibling `readPadiMemoryOnce` (same diff) treats the identical anomaly as a fault. Fixed in `b48643c8b` — `readExitCode` now returns `number | undefined` and `resolveGone` surfaces the empty case as an `{ kind: "error" }` outcome (a kaval contract violation: `exit` yields exactly once), matching this file's fail-loud stance for the snapshot-first-frame check. Behavior-neutral in practice (the exit code tombstones, so the stream never ends empty).

Everything else clean: the `firstFrameOrUndefined` consolidation is faithful; the AST guard matches the repo's `@babel/parser` guard convention; no unbounded collections, sync-blocking calls, or schema/migration issues in the touched surfaces.

srid:
## Evidence

SR6 is a **behavior-neutral internal refactor with no on-screen surface** — it touches kaval-tui's CLI exit-code reads, the server's padi-memory read, and a guard test; nothing in the kolu browser UI changes. So the evidence is behavioral, not visual: the campaign's **byte-identical pin (the seal test)** plus **the exit-path behavior driven against a real pty-host**, all on the shipping HEAD `cf484d764` (captured on an ephemeral pu box).

### The two changed code paths still behave identically — over a real unix socket

`kaval-tui`'s `attach.test.ts` + `wait.test.ts` drive `readExitCode` (adopted `firstFrameOrUndefined`) and `consumeExit` (the deliberate non-adoption codex's F1 preserved) against a **real pty-host over a real unix socket** — not mocks:

```
✓ attach.test.ts › runAttach — over a real unix socket › reports the real exit code when the PTY's child exits   252ms   ← readExitCode
✓ wait.test.ts   › awaitOutputCondition — match, exit, and interrupt › resolves `gone` when the terminal exits before the condition   115ms   ← consumeExit
  …
  Test Files  2 passed (2)
       Tests  28 passed (28)
```

### The campaign's byte-identical pin holds

```
✓ packages/server/src/seal.test.ts (7 tests) 154ms
  Test Files  1 passed (1)
       Tests  7 passed (7)
```

### The negative-property guard

```
✓ packages/server/src/firstFrameOneShotGuard.test.ts (3 tests)
```

Two-platform CI (`x86_64-linux` on a leased pool box + `aarch64-darwin` on rasam) runs the full e2e + seal + unit suites on this exact SHA — status posted to the PR checks.

Issue #1831, opened 2026-07-15
SR8 residual: kolu-server processMemory sampler deletion — gated on eager-serve ordering


srid:
Deferred deletion-proof from SR8 (surface reactive bridge). Tracked so the deferral is a committed work-item, not a loose "later."

## What's deferred
SR8 converts the three hand-rolled samplers to \`derived.cell(source({read,install}))\` and lands **two of three** (padi memorySampler + hostInventory, both green). The **third** — kolu-server's \`processMemory\` (\`startMemorySampler\`, \`packages/server/src/memorySampler.ts\`, wired \`index.ts:864\`) — is deferred.

## Why (the real blocker, grounded)
kolu's surface is served **eagerly at module load** (\`surface.ts:336\` \`implementSurfacesOnPublisher\` fires connects at import). The kolu-server sampler's read (\`readPadiMemoryOnce\`) and its fused-cadence force-resample (\`padiSession.onState\`, \`index.ts:872\`) need \`padiSession\`, which is created **later** in \`index.ts\`'s async boot. A derived poll cell moves read+install to module-load time — before \`padiSession\` exists. The read degrades cleanly (null → honest \`{absent,absent}\`), but the \`onState\` force-resample can't wire → a behavior regression (the rail would freeze a stale MB for up to ~5s on a padi drop instead of flipping to \`absent\` at once).

## The correct fix (this issue's real scope)
Resolve the ordering: defer the kolu surface serve into \`index.ts\` (after \`padiSession\`), so read+install happen when the session exists — then convert the sampler. This is a structural change to how the router/serve is assembled (the router is exported at module load for app-router assembly), which is why it's out of SR8's scope.

## Explicitly rejected
- **Late-bind seam (a registrar the source's install writes, wired post-\`padiSession\`)** — works and is behavior-neutral, but it's a late-bound-global indirection that reads as an **override-knob**, which the repo design philosophy bans. Not the path.

## Until then
The existing imperative \`index.ts\` sampler stays as-is (working, with its \`onState\` resample) — **no regression ships**. The deletion-proof lands once the eager-serve ordering is fixed.

Blocks: SR8's third sampler deletion-proof.

srid:
[...]

PR #1827, opened 2026-07-14
fix: #1719 padi-reconnect flake — a ~15-line test-side attach retry (STREAM_RETRY mimic)

srid:
## What

Fixes the padiBinding "reconnects when padi dies" flake (`unit@x86_64-linux`, #1719) — **8 sightings** — with a **~15-line, test-side retry**.

## The fix

When a bound padi dies while a re-served `terminalAttach` stream is live, SR5's #1822 relay ends the downstream with the **retryable** `RelayTransportLostError` (`SURFACE_RELAY_TRANSPORT_LOST`) — an **awaited** throw. Production's re-serve consumer re-subscribes on it transparently via `STREAM_RETRY`; the test dials the re-served router with a **raw `createRouterClient`** (no retry plugin), so a `terminalAttach` racing the reconnect threw un-retried and flaked.

`roundTripTerminal` now **retries the attach on a survivable transport float** — the `STREAM_RETRY` mimic, exactly as it already retries `create` across the reconnect gap. **691 + 371 + 400 iters green at HEAD.**

## Honest record — this PR overturned its own earlier root-cause

An earlier iteration of this branch shipped a much larger fix (a `@kolu/surface` `ownReadAheadPull` iterator that owns abandoned read-ahead pulls, a kolu-server narrow-loud `unhandledRejection` survival boundary, and a planned upstream oRPC issue), on the theory that a reconnect leaves an **un-ownable oRPC-internal float** the daemon must survive.

Before filing the upstream issue I built the minimal `@orpc/*`-only repro of that float — **it passed** (no float). That distrust triggered a reproduce-first re-check on the box:

| config | iters | test fails | process floats (`unhandledRejection`/`uncaughtException`) |
|---|---|---|---|
| with the ownership wrapper + retry | 400 | 0 | **0** |
| **without** the ownership wrapper + retry | 400 | 0 | **0** |

The abandoned-pull float class **died with SR5's #1822 relay rework** (the relay now catches the mid-stream loss and re-throws it as the awaited retryable error, so nothing floats). The earlier "9.1% → 3.0%, boundary survives the 3%" was a misdiagnosis — the 3% was the awaited retryable throw, and the boundary's handler was empirically proven to **never fire**.

**Per srid's ruling (option 1, retry-only), this branch rips all of it out:**
- reverted `@kolu/surface` `streamHandlers` to the plain async-generator (dropped `ownReadAheadPull`)
[...]

PR #805, opened 2026-04-30
feat: introduce @kolu/surface reactive state framework

srid:
**Kolu's streaming RPC plumbing is now end-to-end framework-managed on both sides.** Five typed primitives — `Cell<T>`, `Collection<K,T>`, `Stream<I,T>`, `Event<I,T>`, `Procedure` — declared once in `packages/common/src/surface.ts` drive the typed oRPC contract, server router (`implementSurface`), and Solid client bundle (`surfaceClient`). The snapshot+deltas wire protocol is the framework's; persistence and pub/sub plug in via `CellStore<T>` + `Channel<T>` adapters (`confStore` / `inMemoryStore` / `publisherChannel`).

### The five primitives

| Primitive | Question it answers | Cardinality | Persistable | Mutable from client | Has current value |
|-----------|--------------------|-------------|-------------|---------------------|-------------------|
| `Cell<T>` | "What's the current X?" | One singleton | Optional | Yes | Yes |
| `Collection<K,T>` | "What's the current X for each key K?" | Many, keyed | Optional | Yes | Yes (per key) |
| `Stream<I,T>` | "What's the live output for input I?" | One per input combo | Never | No (read-only) | Yes |
| `Event<I,T>` | "Has X happened yet?" | Occurrences over time | Never | No (read-only) | **No** — handler-based |
| `Procedure` | "Run this side-effect on the server" | Per call | n/a | Yes (RPC) | n/a |

Cell/Collection/Stream are *state* (consumer renders the current value); Event is *occurrence* (handler fires per yield); Procedure is *imperative*. `terminal.attach` (bidirectional binary) and other raw streaming shapes use `streamCall(client.X.Y, input, opts)` — a one-line escape hatch threading the same `STREAM_RETRY` context the hooks use.

### How a surface is wired

```
common/  defineSurface({ cells, collections, streams, events, procedures })  ─┐
                                  │                                           │ contract
                                  ▼                                           │ derived
server/  implementSurface(surface, { channel, cells, collections, … })  ◀────┘
                                  │
                                  ▼
client/  surfaceClient(surface, { transport })
                ├─ rpc            (raw oRPC client, retry-wired)
                ├─ cells          (.use, .upsert, .patch)
                ├─ collections    (.use, .upsert, .remove)
                ├─ streams        (.use)
                └─ events         (.use)
```

### Migrations in this PR

| Site | Before | After |
|---|---|---|
[...]

PR #1832, opened 2026-07-15
SR8 — the poll source shape, derived.collection, and the urgency O(M²) fix

srid:
SR8 continues the [reactive-bridge](https://kolu.dev/atlas/surface-reactive-bridge.html) spine of [A Complete Surface Runtime](https://kolu.dev/atlas/surface-runtime-boundary.html): it lands the poll `source` shape and `derived.collection` in `@kolu/surface`, resolves SR7's O(M²) urgency regression, and deletes two of the three hand-rolled samplers. A pure consolidation — **byte-identical on screen** (e2e + seal), so the value is defect *classes* dying, not visible change.

> **Draft — still landing:** the L1 poll-on-change core move, the paired drishti PR, and `## Evidence`. Opened now so the review gauntlet's findings land as comments.

## The named obligation — SR7's O(M²) urgency fold, killed and measured

SR7 wired `urgency: derived.cell(($) => recomputeUrgency($.terminals()))` off the **composed** `terminals` collection. Its `$`-sibling read re-ran `registryMap(composePadiTerminal)` — recomposing **all M** terminals — on every one of the M firehose pokes/cycle → **O(M²) composes/cycle**, even though `wrappedUpsert(k, v)` already receives the composed value.

The fix is an opt-in **incremental `$`-sibling read** (`materializeSiblingView`): the framework maintains a per-key materialized cache of `readAll()`, updated by the same `upsert`/`remove` writes that already carry the value, so `$.terminals()` stops recomposing. The compose-count gate (`urgencyComposeCost.test.ts`) is **green**:

| | composes / firehose cycle (M=24) |
| --- | --- |
| before (SR7) | **600** (M + M²) |
| after (SR8) | **24** (M) |

**O(M²) → O(M).** Terminals stays imperative — this is *not* SR10's `terminals: derived.collection(...)` (gated on signalMap). Live-padi wall-clock capture lands under `## Evidence`.

> **Note correction (author-disclosed).** The bridge note's honest-cost line — "SR8's keyed reconciler (`derived.collection`) re-composes only the changed key" — conflates two things: `derived.collection`'s reconciler is a **wire dedup** (diff-by-`equals` on a whole-map read), while the compose-saving comes from the **incremental sibling-read**. The coordinator corrects the note on the atlas branch to match what shipped.

## What lands

- **Poll `source({read, install})`** — the bridge owns the T+0 seed read (first failure propagates, never a fabricated default), the non-overlap guard, and later-read log-skip-continue; the caller owns the cadence. A poll-backed `derived.cell` connects asynchronously (a rejected seed faults the runtime's `done`) and seeds the private store from the spec default until the first read lands (behavior-neutral).
- **`derived.collection(node)`** — the keyed-reconciler wire adapter: diffs each frame against the last by `CollectionSpec.equals` and publishes only the changed + removed keys. Graph-owned (ctx `upsert`/`remove` throw). Its SR8 consumers are drishti's three keyed poll-reconciles (paired PR); kolu's own collections stay imperative.
- **Two of the three samplers deleted** — padi `memorySampler` + `hostInventory` become `derived.cell(source({read, install}))`; their hand-rolled T+0/interval/`inFlight`/`onState` loops are gone.

## Honest done-claims (deliberate deferrals)

- **Samplers: two of three.** The **kolu-server `processMemory` sampler is deferred to #1831**. Its conversion depends on resolving an eager-module-load serve vs late-`padiSession` ordering; forcing it in-scope would need a late-bound-global seam that violates the no-override-knob philosophy. The existing imperative sampler ships **untouched** — no regression.
- **The `writeWrappedValue`-copy deletion (ledger L1) is already satisfied** — it was consolidated in #805 before SR8; L1's live deliverable is the poll-on-change core extraction (still landing).

## Design philosophy

- **Fail-fast / no fallbacks.** A derived cell/collection's ctx write verbs *throw* (the graph is the one writer). A poll source's first-read failure propagates rather than serving a fabricated default. The `materializeSiblingView` guard is a vouched invariant (every mutation flows through the ctx seam), not an override knob.
- **No override knobs.** The kolu-server sampler's late-bind seam was explicitly rejected as the override-knob defect the philosophy bans; it rides #1831 instead.
- **Reuse the source of truth.** The urgency fix rides SR7's *existing* collection poke seam; `derived.cell`/`derived.collection` ride the existing cell/collection connect + publish seams — zero runtime surgery. `CollectionSpec.equals` is the collection sibling of the existing `CellSpec.equals`, declared once instead of hand-held at each write site.

## Dependency

Touches `@kolu/surface` stream/client territory adjacent to **#1827** (FLK-STDIO). The poll-core work is placed to avoid its exact lines; merge order is the coordinator's.

## Test / evidence

`just check` + `just fmt` green. `reactor.test.ts` (38) covers the poll source (seed, first-failure-propagates, `inFlight`, log-skip-continue) and `derived.collection` (reconcile, one-writer guard, equals dedup); `urgencyComposeCost.test.ts` is the O(M²)→O(M) gate. Two-platform CI + the paired drishti PR + `## Evidence` land before ship.


srid:

✅ Consensus · base: `027142e6` · 3 fixes applied

The lenses (volatility boundaries · structural simplicity) debated the SR8 diff to consensus and applied three fixes — each collapsing a fragmented concern to one home:

| Commit | Fix |
|---|---|
| ece7b809a | Give the poll-connect lifecycle ONE home (`connectPollNode`) across the cell and collection builders — the isTorn/adopt/teardown protocol was duplicated |
| d44291b8e | Give the `unref`'d-interval poll cadence one home via `everyMs(ms)` — the `setInterval(tick, ms).unref()` hygiene was hand-rolled per sampler |
| fd22ed727 | Gate the collection write-seam no-op on the derived arm (an authored collection's missing seam must stay a loud fault, not a swallowed no-op) |

No unresolved findings.

srid:
## Codex ⇄ Claude debate

✅ Consensus in 3 rounds (round 4 = approval) · reviewer effort: xhigh · base: `027142e6`

|---|---|---|
| 1 | 565c15324 | fix(codex): round 1 — poll lifecycle, connect supervision, and startup ordering |
| 2 | b6336743d | fix(codex): round 2 — honest poll typing, connect-abort race, durable readiness |
| 3 | 0ad48f962 | fix(codex): round 3 — honest PollDerivedCell type, guard the seed transaction |

- **F1** — The later-tick path calls read() before any promise rejection handler exists
- **F2** — A poll-backed derived cell's connect implementation ignores the AbortSignal that walkSurface supplies to every owned connector
- **F3** — PollSource<T> publicly extends GraphNode<T>, but its actual signal is undefined until the asynchronous seed and is cast to ReadonlySignal<T>
- **F4** — Promise.resolve(dc.connect(...)) evaluates dc.connect synchronously
- **F5** — Pulse termination does not own the in-flight query
- **F6** — The new derived poll cells start when serveDaemonSurfaces constructs the runtime, before bootLocalEndpoint connects/spawns kaval and before the state-root manifests are written
- **F7** — Round 2 passes the runtime AbortSignal to later reads, but the rejection handler logs every rejection unconditionally
- **F8** — Install-before-seed moves the cadence into the lifetime of seed publication, but cleanup covers only read rejection and disposed-after-read

srid:

Rules + fact-check passes over the SR8 diff (elegance skipped — `/simplify` ran this gauntlet). Two rule findings fixed, one fact-check finding verified not-reachable.

| Pass | Finding | Disposition |
|---|---|---|
| Rules · `callback-fanout-guarded-at-funnel` | `tickRead`'s later-tick `set(v)` publisher throw was an unhandled rejection (only the read throw was guarded) | **Fixed** — a trailing `.catch` now covers both the read rejection and the publisher throw (log-skip-continue, hold last), symmetric with the seed transaction. Regression added. |
| Rules · `no-untyped-escape-hatches` | An authored collection missing its `upsert`/`remove` seam was forced via `collDeps.upsert!` — a cryptic "undefined is not a function" at first publish | **Fixed** — explicit named boot throw; the derived arm (the only one with no seam) stays a no-op. |
| Fact-check | A poll cell's seed failure tears down the cadence forever; under padi's non-fatal `done` it would serve the default for the daemon's life | **Verified not reachable** — both poll reads are total (`samplePadiMemory`: `memoryUsage` never throws + `pollKavalRss` catches; `samplePadiHostInventory`: every fs discovery try/catches → `[]`, every probe `.catch` → `EMPTY_PROBE`). Its one throw is a deterministic boot-order defect (`padiSocket` unset) that *should* fault `done` (fail-fast). No change. |

srid:
**Paired drishti PR:** [srid/drishti#103](https://github.com/srid/drishti/pull/103) — adopts the SR8 `derived.collection(source({read,install}))` API (3 keyed poll-reconciles), pinned to the final SR8 HEAD `f4e810d1d`. Satisfies `.claude/rules/surface.md`. Local typecheck + fmt-check green; drishti CI in progress.

srid:
## Evidence

All on the exact shipping HEAD **`f4e810d1d`**.

### 1. Behavior-neutral seal — two-platform e2e green
The consolidation is behavior-neutral on screen: `ci::e2e` passed on **both** lanes on the exact SHA (plus `unit`, `biome`, `fmt`, `flake-check`, `home-manager`, `nix`, both surface example builds).

| Lane | Host | e2e | all nodes |
|---|---|---|---|
| x86_64-linux | kolu-ci-2 | ✅ 230s | 14/14 ok |
| aarch64-darwin | sincereintent | ✅ 279s | 14/14 ok |

### 2. Named obligation — the compose-fold went O(M²) → O(M) (PRIMARY proof)
`packages/padi/src/urgencyComposeCost.test.ts` counts real `composePadiTerminal` invocations while driving one firehose cycle through the real `implementSurface` connect seam:

| | composes / firehose cycle at M=24 |
|---|---|
| **before** (SR7 — `$.terminals()` re-runs `registryMap(composePadiTerminal)` per poke) | **600** = M + M² |
| **after** (SR8 — `materializeSiblingView`: the `$`-read hits a per-key cache) | **24** = M |

Structural, cannot flake — the coordinator-ruled primary proof.

### 3. Wall-clock (SUPPORTING artifact) — through the real fold path, M = 24 … 1536
Median of 7 timed firehose cycles per size (fix OFF vs ON), same `implementSurface` seam + the real per-key `composePadiTerminal` cost:

```
  M | fold OFF (ms) | fix ON (ms) | speedup | OFF / M² (µs)
----+---------------+-------------+---------+--------------
 24 |         0.28  |       0.07  |    4.0x |        0.48
 96 |         2.29  |       0.14  |   16.3x |        0.25
384 |        18.48  |       1.36  |   13.6x |        0.13
768 |        69.91  |       5.05  |   13.9x |        0.12
1536|       278.26  |      19.19  |   14.5x |        0.12
```

- **`OFF / M²` converges to a flat ~0.118 µs** (0.48 → 0.25 → 0.13 → 0.12 → 0.12) and OFF grows ~4× per M-doubling at M ≥ 384 — textbook **O(M²)**. At M=1536 the old path is **278 ms per firehose cycle** (quarter-second jank).
- The fix removes the **expensive** per-poke re-compose (a `SleepingTerminalSchema.parse` / object-spread per terminal), which is **~93 % of the old per-poke cost** — a **~14× wall-clock win** at scale (M=1536: 278 ms → 19 ms), and at a realistic busy-host M≈24 the urgency fold is a negligible 0.07 ms.

**Honest footnote.** The speedup *plateaus* at ~14× rather than growing, because the urgency cell still recomputes fully per poke (`recomputeUrgency` scans the whole map), a **cheap** residual O(M²) that neither path eliminates — SR8's obligation was the *compose*-fold (the expensive term), which is resolved. Taking the urgency cell itself to O(M) (an incremental per-key delta) is a separate future optimization, out of SR8's scope. The compose *count* is unambiguously O(M²)→O(M) (§2); the wall-clock confirms the dominant cost is gone.

_(The wall-clock bench is a one-time capture through the committed gate's harness — not committed, to keep the deterministic compose-count gate as the sole in-tree proof.)_

srid:
✅ **drishti #103 CI green** on both platforms (build-and-push macos-latest + ubuntu-latest) against the SR8 surface API pinned to `f4e810d1d`. The `.claude/rules/surface.md` gate is satisfied on the final HEAD.

Issue #1716, opened 2026-07-07
Type-split: make `copying` unrepresentable for the local (endpoint) session arm


srid:
## The illegal state

`packages/surface-remote/src/session.ts` builds every session — local and remote — over one shared `SessionState.connection: ConnectionState`, where `ConnectionState` (`connection.ts:38`, `CONNECTION_STATES`) **includes `"copying"`**. But `"copying"` is the nix-closure-**provisioning** phase — a REMOTE-ONLY fact. The local padi arm rides the same `makeSession` (via an `endpointConnector`, `padiBinding.ts`) with **no** closure to copy, so the type admits a state the local session can never legitimately inhabit.

Both the initial state (`session.ts:310`) and the reconnect re-arm (`session.ts:525`) hardcode `connection: "copying"`.

## Why it's a follow-up, not the fix that shipped

This is **pre-existing** (the copying-capable local session predates the W4 always-map work — `surface-nix-host/session.ts:310/:525` already initialized at `"copying"` on master). W4's always-map flip only made it **newly reachable in the client** (the map's `EntryStatus` projection maps `copying → warming`, `surface-map/server.ts:82-84`, so a local session's phantom copying would paint a lying "warming" chip).

The W4 PR (#1714) sealed its own new reachability at **runtime**: the connector now DECLARES its provisioning vocabulary (`MakeSessionOptions.provisions`), so the local `endpointConnector` opens at `"connecting"` and its state sequence can never contain `"copying"` (pinned by test), and `serveHostMap` asserts a local key never projects copying. But the **type still admits it**.

## The honest fix (this issue)

Split the state so the local (endpoint) session type is `connecting | connected | disconnected | failed` and ONLY the remote (ssh) arm's type adds `"copying"` — then `local is copying` is a **type error**, not a runtime guard. Apply the same to the `EntryStatus`/`EntryConnectionState` projection.

## Scope

~16 files touch `ConnectionState`-ish; 8 `"copying"`-literal sites. The split threads a local-vs-remote phase type through the shared `Session`, `SessionState`, the `Connector` abstraction, the state cell, and both projections — cross-cutting, not a leaf change.

_Filed from the #1714 W4 review gauntlet (P4 / make-illegal-states-unrepresentable finding); runtime-guarded + pinned there._

srid:
Shipped in #1730 (W6 — the honest connect, merged as e4acb5b2f). The type split landed exactly as filed, and went further than the ask:

[...]

PR #1836, opened 2026-07-15
SR9 — reactiveFamily + derived.registry, and ONE connection authority

srid:
**SR9 lands the reactive bridge's keyed machinery and uses it to close a whole defect class: a host's connection status can no longer be two facts that disagree.** `reactiveFamily` + `derived.registry` replace ~60 lines of hand-rolled `serveHostMap` plumbing; then the connection-authority extension makes the coarse dot (`EntryStatus.kind`) and the fine word both derive from the **one** `SessionState` frame — so the drishti#102 divergence (green dot + a status word stuck on "connecting" forever) has no construction path left.

### The keyed machinery (worked example 4)

`reactiveFamily<K, S>` is a keyed family of member states as a graph source — it owns, once and for every consumer, the four things a hand-rolled keyed registry re-spells:

- **membership diff** (attach entrants, detach + evict leavers),
- **last-frame hold** (each member's latest state cached; a version-signal change edge, O(1) per frame, never re-allocating the map),
- **per-key disposal** (an `onEvict` seam), and
- **per-member error isolation** (one member's throw is contained; a republish defect rethrows out-of-band, never a stale-but-healthy status).

`derived.registry(family, resolve)` is its pull-face `MapRegistry` exit. `serveHostMap` keeps only the projection (`projectState` + the injected `failureOf` + the #1716 belt) and the `map.codec` bridge; the `latestState`/`stateSubs`/`attach`/`detach`/`reconcile`/`fire` machinery is gone.

### One connection authority (fixes the root of srid/drishti#102)

Before SR9 a host's connection was projected **twice, independently** — the dot from the map's `entries` collection, the word from a separate per-host `connection` cell — two wire channels, two client subscriptions, nothing binding them to agree. SR9 puts the fine connection on the **same entry the dot rides**:

```
         one SessionState frame
                  │
   serveHostMap.resolve (one projection)
          ┌───────┴────────┐
   coarse kind        fine connection   ← co-produced, no half-updated pair
     (the dot)          (the word)
          └───────┬────────┘
        the map's ONE entries channel
                  │
   client reads state() → kind + connection   ← one subscription
```

`EntryStatus` gains a `Conn` payload parameterized exactly like `Failure` (surface-map carries + validates it, never enumerates it — volatility-neutral). The **joint-invariant contract test** (RED-first, now green by construction) pins it: *for any `SessionState`, `EntryStatus === connected` ⟺ `connection.phase === connected`* — the permanent fence against a future second authority.

### Proven by deletion

| Deleted | Why it's safe |
| --- | --- |
| ~60 lines of `serveHostMap` hand-rolled plumbing | → `reactiveFamily` |
| the per-host `connection` cell, `mirroredSurface`/`WithConnection`, `seedConnectionCell`, `pipeSessionStateToCell` (connectionPipe.ts gone) | the fact rides the entry now; the pure `projectConnection` leaf survives on `/connection` |
| the pump's `connection` option (a second `session.onState`) | link health is projected once, in `serveHostMap` |
| the client's second `cells.connection.use()` subscription | the word derives from `active.state().connection` — the same entry the dot reads |

### Notable

> **Scope grew, honestly stated.** The row headlines "~60 lines of serveHostMap plumbing", but the connection-authority extension entails a `@kolu/surface-map` public-API reshape (`EntryStatus` + `defineSurfaceMap` + the wire schema grow a `Conn` payload). Approved by the coordinator as entailed-by-B. Kolu-side is **behavior-neutral on screen** (the dot + word render the same, from one source) — no kolu changelog entry; the user-visible drishti#102 fix rides the paired drishti PR.

> **The `remote-process-monitor` example** is reworked as a cell-free re-serve demo with a pointed note: connection presentation is a host-map concept (see `serveHostMap`). _Sweep lesson banked: a consumer sweep must cover `example/` + test trees, not just `packages/*/src`._

**Dependencies:** builds on the merged FLK-STDIO (#1827). Paired drishti PR ([srid/drishti#105](https://github.com/srid/drishti/pull/105)) (deletes drishti's word-consumption path, derives the word from the entry via `projectConnection`, `connectionColors` byte-identical) pinned to this PR's final HEAD (`c62af2f7d`, after merging `origin/master` — SR8.a's server restructure — into this branch) — required green before merge. The ratified SR9 plan-of-record rides the Atlas PR #1833.

> ✅ **Acceptance (live).** srid deployed this branch build + the drishti pair (#105) to real hardware and **confirms the drishti#102 bug is fixed**: the connection dot and the status word now agree (no more green dot beside a permanent "connecting"). The user-facing acceptance test has passed.

### Honest cost — the entries channel equals-gates re-emits

Folding the fine connection onto the entry means the map's family now fires on **every session frame**. The `entries` republish is gated so a member re-emits **only when its own published `EntryStatus` changed** — a sibling's frame never re-emits an unchanged entry (`connection`-by-reference = frame identity is the change signal; the keyset publishes only on a membership change). Pinned by `mapHarness.test.ts (15)`, which counts the SERVER's per-key emits on a raw client beneath the high-level client's dedup. *Cold before SR9 (the channel carried only the coarse status); this keeps the O(M²)-across-a-streaming-pool cost from riding in with the fine word.*

### The steady-state pin lives as a vitest joint-render (not a browser e2e)

The SR9 row's *"steady-state e2e"* is realized in drishti as a **joint-render vitest pin** (`connectionAuthority.test.ts`): drishti has no browser harness **by design**, and under B the delivery-divergence class **died with the second channel** — one source, nothing to go stale, so a Playwright harness would only re-prove an invariant with no construction path. Ratified by the coordinator; the SR9 row's wording is trued to match.

### Review gauntlet

Through architecture-first-principles, lens-debate (lowy ⇄ hickey), codex-debate (xhigh), simplify, and code-police. **The load-bearing save:** code-police caught that the serveHostMap reshape's `?? (() => {})` no-op disposer would **freeze a racing member at "connecting"** when its session hadn't landed yet — silently reintroducing the exact drishti#102 class. Fix: `attach` returns `undefined` to signal *retry-me-next-frame* (never a no-op), and membership reads the source frame (not the attached set) so a permanent-no-session ghost fails loud instead of hiding. Pinned by **`reactorFamily.test.ts` → "retries an attach that returns undefined on the next membership frame (never freezes)"** (plus the attached-but-unseeded / throwing-attach visibility tests).


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `41b48b17acf7`

### Applied (4)
- `lowy-3` Dead export `projectConnection` — zero in-tree consumers, not reused by sessionConnection — (uncommitted)
- `lowy-4` Stale doc references to the deleted `connectionPipe` module — commit `3ca2a859e`
- `hickey-1` The "one connection authority" is server-only; the client re-fragments dot and word into two independent floors — commit `6cdc460ea`
- `hickey-2` Two parallel session->ConnectionInfo projections; the strict one has no in-repo consumer and its cast is duplicated, not composed — commit `3aef6578a`

### Agreed — no change (4)
- `lowy-1` Receptacle holds state but not the session handle — consumer still reaches past it to the pool (packages/surface-remote/src/serveHostMap.ts:273)
- `lowy-2` reactiveFamily's GraphNode `.value` push face has zero consumers (speculative surface) (packages/surface/src/reactor.ts:1170)
- `hickey-3` reactiveFamily carries an unused GraphNode `.value` face for a consumer that does not exist (packages/surface/src/reactor.ts:1069 (extends GraphNode) and :1170 (value: engineComputed))
- `hickey-4` DerivedRegistry and MapRegistry are near-identical shapes bridged by a hand-rolled codec adapter (packages/surface/src/reactor.ts (DerivedRegistry: members/has/resolve/subscribe/dispose) + packages/surface-remote/src/serveHostMap.ts:265 (the MapRegistry wrapper))

srid:
## Codex ⇄ Claude debate

|---|---|---|
| 1 | 1e416b1de | fix(codex): reactiveFamily robustness + structural one-authority (round 1) |
| 2 | 7fab5b6b7 | fix(codex): sessionConnection full-frame validation + doc seam (round 2) |
| 3 | 9ba6e066c | fix(codex): sessionConnection regression tests + stale connectionOf comments (round 3) |

- **F1** — reactiveFamily.subscribe observes the version signal through a Preact effect
- **F2** — A member setter has no attachment-generation or disposed fence
- **F3** — Attachment, member-disposer, and onEvict failures stop at console.error
- **F4** — EntrySession.state and EntrySession.connection are independent fields, and serveHostMap independently calls projectState(raw) and arbitrary connectionOf(raw)
- **F5** — sessionConnection accepts SessionState<string> and unchecked-casts it to SessionState<SshProv>
- **F6** — Deleting the connection teardown left an outer try/finally whose finally block contains only a comment
- **F7** — Current API documentation still points to the deleted mirroredSurface and seedConnectionCell seam, and line 132 tells consumers to drive the removed connection cell when done rejects
- **F8** — Round two renamed the public ServeHostMapOptions seam from connectionOf to connection: { project, isConnected } and added a mismatch check, but the canonical @kolu/surface-remote Reference page still documents connectionOf at lines 125 and 164 and does not explain the new predicate/enforcement

srid:

Two forked passes (rule checklist + fact-check), scoped to `origin/master...HEAD`. Findings and actions:

**Pass 2 — fact-check (the important one):** `serveHostMap`'s reshaped `attach` substituted a **no-op disposer** for a member whose session hadn't landed yet (the reconcile race). The family marked it attached and blocked all retry, so once the session appeared `resolve` found a live session but a `latest` of `undefined` — publishing a **permanent "connecting"**, the exact frozen dot/word bug this PR exists to kill. **Fixed** (`e35831de6`): `attach` returns `undefined` to signal "retry me" (the family leaves it unattached and retries next frame), and membership (`keys`/`has`) reads the source's frame separate from the attached subset — so a present-but-session-less member stays visible and `resolve` fails loud on it (`UnclassifiedHostSessionError`), never silently dropped. New retry regression test.

**Pass 1 — rule checklist:** clean except one `comment-the-non-obvious` — 6 stale comments in *untouched* sibling files + `ref-surface-map.mdx`'s `EntryState<Failure>` still referenced the deleted `mirroredSurface`/`connectionCell`. **Fixed** (`e35831de6`). Every other rule (dry, invalid-states, no-dead-code, no-silent-error-swallowing, no-unbounded-growth, caught-error-must-not-collapse, fail-fast, …) checked and ruled out.

Run with `--no-elegance` (the simplify track already ran this gauntlet).

PR #1843, opened 2026-07-15
fix(code-tab): stop the file-tree selection shimmer (focus-drag loop)

### LINKED ISSUE #1841: Code tab: file selection ping-pongs between two adjacent files (focus/selection thrash loop)
## Summary

## Repro

### What it is *not*

|--:|---|---|

## Fix direction

## Evidence

srid:
Resolves #1841.

## The bug

In the Code tab, selecting files could latch into a runaway loop where the tree **selection ping-pongs between two adjacent files** at ~60–120 Hz (once per animation frame), pegging a CPU core and firing an ~800 msg/sec RPC storm. It self-sustains with no further input and stops only when you switch terminals. Seen in both **All files** and **Branch (diff)** modes.

## Root cause (from a Chrome Performance profile of the live loop)

Hottest self-time in the loop:

|--:|---|---|
| 4745 | `HTMLElement.focus()` | `@pierre/trees` render |
| 4187 | selection `setAttribute` (recursive row re-render) | same |
| 1839 | `WebSocket.send` | per-flip content-query re-fire |
| 1191 | `AbortController.abort` | per-flip supersede |

The wrapper's selection effect re-applies `props.selectedPath` into Pierre and reveals the row with `scrollToPath` — with **no options**, which makes Pierre **drag keyboard focus** onto the row (`FileTreeController.scrollToPath`: `if (options?.focus !== false) this.#setFocusedIndex(...)`). That focus move makes Pierre re-emit `onSelectionChange` reporting the **neighbouring** row, which `handleSelect` writes back into the per-terminal `selectedFileByMode` store, which re-runs the effect → `scrollToPath` drags focus again. The `send`/`abort` churn is the downstream symptom: each flip changes the selected file, so the content/diff `createPolledQuery` re-fires an RPC and aborts the previous.

## The fix

One line: reveal the row **without** stealing focus — `scrollToPath(path, { focus: false })` ([`FileTree.tsx`](../blob/bug-shimmer/packages/solid-pierre/src/FileTree.tsx)). The row still scrolls on-screen; keyboard focus stays where the user put it; the neighbour-echo loop can't start.

## Test

`FileTree.focus.test.tsx` mounts the wrapper over a `@pierre/trees` stand-in, drives an external `selectedPath` change, and asserts the reveal happens **without** dragging focus (`scrollToPath(path, { focus: false })`). It is **red on the pre-fix code** (bare `scrollToPath(path)`) and green after. Adds `vite-plugin-solid` + `happy-dom` to render the Solid component, mirroring `packages/surface`.

## Verification status

- ✅ Unit test: red → green; the profile-confirmed focus-drag is removed.
- ⚠️ **End-to-end not yet confirmed on a live instance.** The runaway loop is a timing/latch race that only manifests on a busy instance (many terminals + agents churning metadata); a quiet dev box does not reproduce it. The fix removes the exact engine the profile fingered (`focus()`), but it should be confirmed on an instance that actually reproduces the shimmer — e.g. `nix run --refresh github:juspay/kolu/bug-shimmer`.

## Notes
- Only the reactive selection reveal (`FileTree.tsx:360`) is changed. The mount-time reveal (`:237`) still focuses the initial selection; it is a one-shot, not part of the sustained loop (`#1758` already covers remount churn), so it is left as-is.

### DIFF
--- a/packages/solid-pierre/package.json
+++ b/packages/solid-pierre/package.json
@@ -17,7 +17,9 @@
   },
   "devDependencies": {
     "@types/node": "^22.0.0",
+    "happy-dom": "^20.8.9",
     "typescript": "^7.0.2",
+    "vite-plugin-solid": "^2.11.0",
     "vitest": "^4.1.0"
   },
   "dependencies": {
--- a/packages/solid-pierre/src/FileTree.focus.test.tsx
+++ b/packages/solid-pierre/src/FileTree.focus.test.tsx
@@ -1,0 +1,109 @@
+// @vitest-environment happy-dom
+/**
+ * Regression for the Code-tab file-tree "shimmer" — juspay/kolu#1841.
+ *
+ * The wrapper's selection effect re-applies `props.selectedPath` into Pierre and
+ * reveals the row with `scrollToPath`. Calling it with NO options makes Pierre
+ * DRAG keyboard focus onto the row — `FileTreeController.scrollToPath` runs
+ * `if (options?.focus !== false) this.#setFocusedIndex(...)`. On the live app that
+ * focus move makes Pierre re-emit `onSelectionChange` reporting the NEIGHBOUR row,
+ * which the host writes back into the per-terminal selection store, which re-runs
+ * this effect — a self-sustaining focus/selection loop between two adjacent files
+ * at ~60-120 Hz. A Chrome profile of the live loop shows `HTMLElement.focus()` as
+ * the single hottest leaf, rooted in Pierre's render.
+ *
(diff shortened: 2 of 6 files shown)


srid:
Closing — the `{ focus: false }` change was a symptom patch (it removed the profile's hottest leaf, `focus()`, but deploy testing confirmed the shimmer loop persists, so `focus()` was a consequence of the render churn, not its driver). Per maintainer direction, fixing at the root instead: bumping the Pierre libraries (the buggy focus/selection behavior lives in `@pierre/trees`). New PR incoming. #1841 stays open.

PR #1845, opened 2026-07-15
debug(#1841): log selection writes to capture the shimmer's root cause (DO NOT MERGE)

srid:
> **⚠️ DO NOT MERGE — temporary debug instrumentation for #1841.**

The Pierre beta.5 bump (#1844) did **not** fix the Code-tab selection shimmer, so this branch captures the actual root cause instead of theorizing.

## What it adds
23 lines in `useRightPanel.ts`: a logger on `setSelectedFile` — the single chokepoint every `selectedFileByMode` write funnels through. Each write is recorded with `{ t, mode, path, stack }` into `window.__selWrites` (capped at 500).

## How to use
1. Deploy: `nix run --refresh github:juspay/kolu/debug/shimmer-selwrites`
2. Trigger the shimmer (agent actively editing files in a folder + clicking files in it).
3. Let it loop ~2s, then in DevTools: `copy(JSON.stringify(window.__selWrites))`.

## What it answers
Whether the loop is **one writer echoing itself** (`onSelectionChange` → `handleSelect` — the "Pierre's echoed re-selects" path the code already names) or **two writers fighting** (e.g. membership-clear vs the front-door resolution effect), plus the exact stack of each. That names the culprit so the real fix + a mechanism-faithful test can follow.

Refs #1841.

### DIFF
--- a/packages/client/src/right-panel/useRightPanel.ts
+++ b/packages/client/src/right-panel/useRightPanel.ts
@@ -384,6 +384,29 @@ export function useRightPanel() {
     selectedFile: (mode: CodeTabView): string | null =>
       activeState().selectedFileByMode?.[mode] ?? null,
     setSelectedFile: (mode: CodeTabView, path: string | null) => {
+      // [SHIMMER-DEBUG #1841] Every selection write funnels through here. Capture
+      // each with its call stack so a live repro of the shimmer reveals whether the
+      // loop is ONE writer echoing itself or TWO writers fighting. Remove before merge.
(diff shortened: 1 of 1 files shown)


srid:
Done its job — the `window.__selWrites` capture from this branch pinned the root cause definitively: every write in the loop was `onSelectionChange → onSelect` (Pierre's autonomous re-emit), no click handler. The real fix (a user-gesture provenance gate) is #1846. Closing this instrumentation branch.

PR #1846, opened 2026-07-15
fix(code-tab): stop the file-tree selection shimmer (user-gesture provenance gate)

### LINKED ISSUE #1841: Code tab: file selection ping-pongs between two adjacent files (focus/selection thrash loop)

srid:
Resolves #1841.

## The bug
In the Code tab, clicking around files **while an agent is actively editing files in that folder** could latch into a runaway loop: the selection ping-ponged between two adjacent files at **~60–120 Hz**, pegging a CPU core, until you switched terminals. Both *All files* and *Local/Branch* views.

## Root cause — from a live `setSelectedFile` stack capture
Every write in the loop had the **identical stack** — `onSelectionChange → onSelect → setSelectedFile` — with **no click handler anywhere in it**, alternating two files (`examples/Main.lean` ↔ `examples/Storage.lean`) every few ms.

The agent's edits make `fsListAll` / git-status tick continuously. Each tick drives the wrapper's `batch` / `setGitStatus` effects into Pierre's controller, which **re-emits `onSelectionChange` with no user input** (the "Pierre's echoed re-selects" the code already names). The wrapper forwarded every echo to the per-terminal store; the store re-applied selection into Pierre via the selection effect; Pierre re-emitted — a frame-rate **store↔Pierre loop**.

(Earlier attempts — the `{ focus: false }` patch #1843 and the Pierre beta.5 bump #1844 — were ruled out by deploy testing. #1844 stays merged as good hygiene; this is the actual fix.)

## The fix
Forward `onSelectionChange` to the host **only when a genuine user gesture caused it**. A capture-phase `click`/`keydown` on the tree container arms a single-use token — composed events cross Pierre's shadow root, and capture runs *before* Pierre's own row handler emits synchronously in the same dispatch — auto-disarmed on the next microtask so a click on an already-selected row can't leave a stale token. An autonomous re-emit finds no token and is dropped, so it can never start the loop. ~20 lines in `FileTree.tsx`.

## Test (red → green)
`FileTree.gesture.test.tsx` mounts the wrapper over a Pierre stand-in and makes "Pierre" emit selection **with and without** a preceding user gesture:
- **autonomous emit → dropped** (the loop can't reach the host) — *fails on the pre-fix wrapper*: 3 echoes → 3 host writes;
- **gesture-caused emit → forwarded**;
- **token is single-use** (re-arms per gesture).

Adds `vite-plugin-solid` + `happy-dom` to render the Solid component (mirrors `packages/surface`). nix `pnpm-deps` hash unchanged (both already in the store). `tsc` green for `kolu-client` and `@kolu/solid-pierre`.

## Verify
Since the loop is a live timing race I can't trigger on a quiet box, please confirm on your instance: `nix run --refresh github:juspay/kolu/fix/shimmer-selection-echo`, then reproduce (agent editing a folder + clicking files in it).


srid:
## Follow-up: fixed the click regression + corrected the root-cause narrative

The first commit's gesture gate killed the shimmer loop but **broke click-to-preview** (deploy feedback: "when I click any file, it doesn't change preview"). Root cause, found by instrumenting a live tree on a pu box:

- Pierre does **not** emit `onSelectionChange` synchronously in the click dispatch — it emits on a **microtask that lands after the arm's own microtask** (measured: emit fires before the next animation frame but after any `queueMicrotask`).
- The `queueMicrotask` disarm therefore ran **before** the emit → `userGesture` was already `false` → every genuine click was dropped.

Fix: disarm on the **next animation frame** instead. It reliably outlasts the deferred microtask emit; single-use consumption still kills the echo loop after one forward. Verified live: a real click now forwards with `userGesture=true` and the preview updates, while an autonomous re-emit is still dropped. A new regression test models the deferred emit and **fails on `queueMicrotask`, passes on `requestAnimationFrame`**.

### Two hypotheses ruled out empirically
The reported trigger was "codex working." I chased two theories and disproved both on the box:
1. **Terminal-title spinner (60Hz) leaking into the tree** — ❌. Spun the terminal title live: the Inspector's `Title` field animated (`spin-108…spin-152 lean-play`), proving the signal reaches the client store, yet the Code tab's instrumented effects logged **nothing**. The `reconcile`-backed metadata store (#1714) isolates the `foreground.title` leaf; the tree only reads `.git.repoRoot`.
2. **fs/git-status churn** — the original narrative; not the mechanism either.

The actual defect is a **controlled-component selection echo**: Pierre's `onSelectionChange` fires for both user clicks and its own programmatic re-selection, and forwarding the echoes back into the store that drives Pierre closes the loop. The gate cuts it with provenance. Added a SolidJS APM rule ("Controlled components need selection PROVENANCE") capturing the anti-pattern + the microtask-vs-animation-frame timing gotcha.

PR #1847, opened 2026-07-16
SR11 — a member declares its client policy (policy off the use-sites, into the declaration)

srid:
The campaign gave the **producer** side its DSL — a member declares its schema, `equals`, derivation, and the framework mints the serving. The **consumer** side never got it: client *policy* — which `onError`, the floor-on-liveness split, `authority`, coalescing — was respelled at ~22 `.use()` call sites. **SR11 moves policy into each member's declaration and mints one interpreter, so a member's error routing is declared once and *unspellable* at a use-site.** After this, a cell's `.use()` takes no options and a collection's takes only `keys`; a future consumer *cannot* re-decide a member's error policy, because there is no options bag left to spell it in.

This is fork **(A)** of the design-bearing lens run: the framework carries an *opaque, app-typed* policy slot + a registration seam, and never names an app arm. It has to — kolu and drishti's error vocabularies **diverge**: kolu toasts, drishti logs, and the *same* map-membership member is interpreted oppositely by the two apps. So the vocabulary lives in each app, the routing in the framework.

```
member spec { client: { onError: <opaque app policy> } }   ← declared once, per member
        │  connectSurfaces / connectSurfaceMap(onClientError)
        ▼
  ONE app interpreter  (policy, err, origin?) ─▶ toast | hostToast | scopedSub   (satisfies never)
```

### The seam, across three packages
- **@kolu/surface** — a discriminated `client?` slot on `CellSpec`/`CollectionSpec` (per-scope `TPolicy` via a curried `defineSurfaceWithPolicy`, so an origin-requiring arm is a *type error* on a root member); `buildSurfaceClient` routes a policy-carrying member's subscription **and** local-authority flush errors through the interpreter, and **throws at construction** if a declared policy has no interpreter (no silent swallow).
- **@kolu/surface-app** — `connectSurfaces({ onClientError })`.
- **@kolu/surface-map** — `connectSurfaceMap(map, transport, { onClientError })` injects the `{ key }` origin **per key** at `clientFor`; the base framework stays origin-agnostic (origin is a surface-map concept, not a base-layer one).

kolu's ONE `interpretClientError` folds in the old `surfaceSubError` (active→toast / background→log, reproduced with the grounded enc-compare) and `onHostMembershipError`, both now deleted.

### Behaviour is neutral — except one ruled delta
Every toast/floor is byte-identical, **except `daemonStatus`** (srid-ruled, 4C):

| | before | after |
| --- | --- | --- |
| host chip error | `Daemon status error: …` (no host) | `Host <name> daemon status error: …` |
| background host error | `console.error` only | `Host <name> daemon status error: …` (toast) |

> **m6 dual-toast disposition:** `daemonStatus` is opened by two subscriptions (the chip and the background wire), so an active host's failure would fire the toast twice. The interpreter passes solid-sonner a deterministic `{ id }` per `(kind, label, host)`, so the identical toast **collapses to one** rather than double-painting. This is a deliberate build decision, not a seal surprise.

### ODU-IMPACT: `none`
Grounded by grep of odu's tree at its pinned kolu SHA `7d51cb2b`: odu consumes `@kolu/surface` **server-side only** (`serveOverStdio` / `serveOverUnixSocket` / `directLink` / `surface-mcp` / `surface-remote`) — zero client `.use()` policy consumption. The change is all-additive (`TPolicy=never` default). No `odu#43` ledger line.

> **Reviewers, two flagged points.** (1) The union is homed in `@kolu/padi` (re-exported from both surfaces), *not* kolu-common — padi's members reference it and the package seal forbids `@kolu/padi` importing kolu-common. (2) `buildSurfaceClient` now also sources a member's declared `authority`/`coalesceMs` from the spec and seeds a local-authority store from the mandatory `CellSpec.default` — without it, baring `preferences.use()` would have silently regressed it to server-authority. That is an `@kolu/surface` behaviour change, so **the drishti pair re-validates at final HEAD** per `.claude/rules/surface.md`.

Implements the SR11 section of the surface-runtime-boundary Atlas note. Negative property enforced by `surfacePolicyUseSiteGuard.test.ts` (no policy option in any `.use()` bag).


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `181382ebb781`

Independent findings: lowy=3, hickey=6

### Applied (3)
- `lowy-2` Dead `_T`/`_P` phantom type parameters reserve structure for a deliberately-foreclosed field — commit `732db7d9e`
- `hickey-1` `onError` is braided into both authority arms of `ClientCellPolicy` instead of sitting beside the discriminant — (uncommitted)
- `hickey-2` Phantom `_T`/`_P` type parameters carried on `ClientCellPolicy` for speculative symmetry — (uncommitted)

### Agreed — no change (6)
- `lowy-1` Electricity test: clean receptacle for error-presentation volatility (verdict, not a defect) (packages/surface/src/solid/surfaceClient.ts:83)
- `lowy-3` `StreamSpec` has no `client` policy slot — the fourth member kind forces one consumer to reach past the receptacle (packages/client/src/hostScope/createHostWire.ts:127)
- `hickey-3` Use-site override precedence for `authority`/`coalesceMs` reintroduces the two-source reconciliation the PR set out to delete (packages/surface/src/solid/surfaceClient.ts:754)
- `hickey-4` The `terminalKeys` stream hand-reconstructs a `scopedSub` policy inline — the declaration model stops at cells/collections (packages/client/src/hostScope/createHostWire.ts:132)
- `hickey-5` The policy rationale prose is duplicated verbatim across ~6 files (packages/padi/src/clientPolicy.ts:1)
- `hickey-6` Two entry points for one concept — `defineSurface` and `defineSurfaceWithPolicy` (packages/surface/src/define.ts:1064)

srid:
## Codex ⇄ Claude debate

✅ Consensus in 3 round(s) (2 with fixes, round 3 approved) · reviewer effort: xhigh · base: origin/master

|---|---|---|
| 1 | 2b9f95d87 | fix(surface): codex round 1 — fail-fast local-authority + origin, executable policy tests |
| 2 | c40d2de50 | fix(surface): codex round 2 — correct dead type-gate refs (F6), sharpen use-site guard (F5) |

- **F1** — ClientCellPolicy is independent of both the cell value type and its resolved verbs, so defineSurfaceWithPolicy accepts authority: "local" for primitive/null cells and for get-only cells
- **F2** — The only new SR11 test is a source-text guard for forbidden use-site option bags
- **F3** — The hostToast and scopedSub arms require a map origin by design, but the interpreter contains graceful fallbacks when origin is absent: hostToast emits an unattributed bare label and scopedSub logs a background error for "?"
- **F4** — The fail-fast comment says connectSurfaces/connectSurfaceMap provide typed enforcement by requiring onClientError for a non-never policy, but both public option fields are explicitly optional and SurfaceMap erases the entries-policy parameter
- **F5** — The negative-property guard does not make use-site policy structurally unspellable: it only matches direct dotted member chains followed by an object literal
- **F6** — Regression introduced by the F1 fix: both this production comment and surfaceClient.policy.test.ts:191 say a type-level ClientCellPolicyServerOnly guard drops the local arm, but that symbol does not exist and the disposition says the attempted type gate was removed

srid:

Rules pass — 4 findings; Fact-check — clean (byte-for-byte toast parity, fail-fast checks, dedup/cardinality, and origin split all verified against their tests). Elegance skipped (`--no-elegance`; `/simplify` ran this gauntlet).

| Finding | Disposition |
| --- | --- |
| `callback-fanout-guarded-at-funnel` — the spec policy folded into `dispatchError`/`narrowedOnError` could skip the fan-out on a throw | **Fixed** (`39cf244b`) — consumer handlers guarded; policy fires last + unguarded so its fail-loud throw still surfaces |
| `prefer-ts-pattern` — `interpretClientError`'s union dispatch was a `switch`/`satisfies never` | **Fixed** (`39cf244b`) — routed through `ts-pattern` `match(...).exhaustive()` (the house default) |
| `no-thin-wrapper-functions` — `useHostProcessMemory` | **Disputed** — not a pure pass-through: its `ReturnType` is the named type handle for `mem` in two interface fields (reader-as-shared-value pattern) |
| `prefer-focused-library` — the guard test's `stripComments`/`matchBrace` | **Disputed** — a test-only string lexer; the TS compiler API is heavier + more fragile here, and codex agreed a full AST lint is disproportionate for this fence |

srid:
### CI venue (tonight)

**Linux-only** this run — the darwin/aarch64 lane is **deferred by srid's overnight-rasam ruling** (rasam is reserved all night for a flaky-test campaign). Linux CI runs on the exact shipping HEAD via `lease.sh` (full pool, egress-smoke first). The darwin lane **runs before merge if srid wants it in the morning** — this is an explicit, ruled deferral, not a silent single-platform drop.

srid:
## Evidence

SR11 is **behaviour-neutral on screen** (policy *spelling* moves from ~19 use-sites onto the member declaration; the interpreter renders the identical toasts/logs) — so the evidence class is **e2e + seal**, not a visual diff.

**e2e — green.** `ci::e2e@x86_64-linux` ✅ (4m29s, on the exact shipping HEAD `7bf512d`). The Cucumber + Playwright suite drives every UI surface — the chips, dialogs, dock, canvas, and their toast/floor states — and passes **unchanged**. That is the seal: the same on-screen behaviour, only the spelling moved.

**Behaviour-neutral, proven three ways:**
- The **negative-property guard** (`surfacePolicyUseSiteGuard.test.ts`) pins that no use-site carries a policy bag any more.
- The **code-police fact-check** traced every migrated toast call site against its new declaration and confirmed **byte-for-byte message parity**.
- New **executable framework tests** (`surfaceClient.policy.test.ts` × 9, `clientPolicyOrigin.test.ts` × 3) assert the interpreter dispatches the exact declared policy value + origin.

**The one ruled delta — `daemonStatus` (srid's 4C ruling).** A daemon-status subscription error now renders a **host-attributed** toast — `Host <name> daemon status error: …` — from both the host chip and the background wire (before: the chip omitted the host, and a background host's error only logged to the console). This is a text-only toast change on an error state, asserted by the `hostToast` dispatch test and called out in the changelog. It is the sole intended on-screen change.

_Linux-only tonight per srid's overnight-rasam ruling; the darwin lane runs before merge in the morning if wanted._

srid:
### Consumer gates

- **drishti pair (tight twin):** [srid/drishti#107](https://github.com/srid/drishti/pull/107) — adopts the client-policy declaration with a **one-arm `{ kind: "log"; label }` union** + one `console.error` interpreter (the divergent-app proof: the framework never names an app arm — kolu speaks *toast*, drishti speaks *log*, over the same seam). Pinned to this PR's HEAD `7bf512d4`, its 5 `console.error` messages preserved byte-for-byte. **drishti linux CI green** (typecheck · nix · fmt-check · agent-boots · … all pass); darwin deferred by the same overnight-rasam ruling.
- **odu (loose pin):** ODU-IMPACT = **`none`** (grounded above) — no ledger line.

PR #1842, opened 2026-07-15
Make the 2.0 changelog a scannable product ledger

srid:
**The 2.0 changelog now reads by product area instead of making readers traverse separate Added, Changed, and Fixed sections.** The unreleased record had grown to 159 commit-oriented notes; this folds it into 71 release-relative entries and presents them as one dense, searchable ledger.

### What readers get

- **Product areas are the primary structure** — Remote Hosts, Kaval, Padi, Sessions, Agent Detection, Mobile, and the existing app surfaces are prominent section headings that link to their guides.
- **Change type travels with each entry** as a colour-coded Added, Changed, Fixed, or Heads-up pill. The same types remain available as filters alongside full-text search.
- **Details stay out of the scan path** until a row is opened. Headlines remain compact; explanations and PR provenance unfold in native `<details>` rows.
- **Release mastheads do useful work** — totals, type counts, and a copyable Nix command are visible without extra navigation.
- **Historical releases are reproducible** — v1.0.0 (and every future archived release) links to its GitHub release and shows a tag-pinned command such as `nix run github:juspay/kolu/v1.0.0`.
- **The layout is deliberately dense and responsive** — large category labels on desktop, a compact single-row filter surface on mobile, and no horizontal page overflow at 390px.

### The release-note rewrite

The content is written for someone upgrading from 1.0, not someone following each intermediate PR. Features born and refined inside the unreleased window become one net-true entry with all of their implementation refs; distinct changes to already-released surfaces stay distinct.

That leaves **71 user-facing notes from the original 159 without dropping their surviving behavior or provenance**. The docs taxonomy was expanded at the same time so the changelog can point somewhere honest: dedicated pages now cover the Dock, Code Tab, Right Panel, Keyboard Shortcuts, Notifications, Sessions, and Mobile.

### The authoring path

`<Change>` now requires one canonical change kind and accepts either one `pr` or a non-empty `prs` tuple, making contradictory provenance states unrepresentable. A remark plugin derives counts and release-scoped heading IDs from the MDX syntax tree during the build, so rendered anchors, Pagefind results, filters, and masthead totals share one source of truth.

The page itself remains data-driven: adding another dated changelog entry automatically creates its release link, GitHub tag URL, pinned Nix command, totals, and section anchors.

### Verified on final head `699e33590`

- `just fmt`
- `just website check`
- `just website build`
- Production Pagefind result → `#unreleased-remote-hosts`
- Filter counts: 71 total · 27 added · 9 changed · 33 fixed · 2 heads-up
[...]

PR #1848, opened 2026-07-16
fix(solid-pierre): move #1846's biome-ignore to the line it suppresses (master is red on ci::biome)

srid:
Master's tip is **red on `ci::biome`**: #1846's `FileTree.gesture.test.tsx` placed its `biome-ignore` pragma 9 lines above the `as any` it targets (inside `render()`'s argument, above the JSX thunk), so `--error-on-warnings` flags both an **unused suppression** and an **unsuppressed explicit-any** — inherited by every branch that merges master.

One-line move of the pragma to the line it suppresses. Surfaced by #1842's master merge (which carries the same fix in-branch so its own CI can green); this PR lands it on master independently. Verified: `biome check --error-on-warnings` clean on the file.

### DIFF
--- a/packages/solid-pierre/src/FileTree.gesture.test.tsx
+++ b/packages/solid-pierre/src/FileTree.gesture.test.tsx
@@ -68,7 +68,6 @@ function mount(onSelect: (p: string | null) => void): HTMLElement {
   const wrapper = document.createElement("div");
   document.body.appendChild(wrapper);
   const dispose = render(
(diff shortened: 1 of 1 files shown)


srid:
Mooted by #1849, which removed the render-thunk `as any` (and its misplaced pragma) entirely — the better fix. Master's `ci::biome` is clean again; verified at origin/master (only the correctly-placed constructor pragma remains).

PR #1849, opened 2026-07-16
review(code-tab): /be-review gauntlet follow-ups to the shimmer fix (#1841)

### LINKED ISSUE #1841: Code tab: file selection ping-pongs between two adjacent files (focus/selection thrash loop)

srid:
## What this is

A **`/be-review` gauntlet pass** over the already-merged Code-tab shimmer fix (#1846, which resolved #1841), landing the refinements the gauntlet surfaced. The shimmer fix itself is in `master`; this PR's diff is **only** the review follow-ups (base = the merge's parent, `22cb7f5f3`).

## The gauntlet

Ran serially over the shimmer diff, each stage committing its own work:

| Stage | Outcome |
|---|---|
| **architecture-first-principles** (C1–C7 + adversarial verify) | C3/C4/**C6 (state-and-time/timing) clean** — the happens-before logic is sound. 1 real: replaced the hand-rolled `AbortController` listener with `@solid-primitives/event-listener` `createEventListener` (repo convention). Plus comment/doc trueups. |
| **lens-debate** (lowy ⇄ hickey) | Consensus, 2 rounds. Documented the selection contract at the socket (`onSelect`/`selectedPath` JSDoc) and the token's *adjacency-not-causation* known limitation. Dropped 3 (graduation candidate already recorded via the APM rule; a sibling `CodeView` with the same pattern — out of scope; comment-dedup subsumed). |
| **codex-debate** (codex `xhigh` ⇄ claude) | Consensus, 2 rounds. F1: qualified the JSDoc (its absolute "never echoes" contradicted the documented adjacency limit). F2: added two regression tests pinning **capture-phase ordering** and **next-frame disarm** (both verified to fail under a negative control). |
| **simplify** (reuse · simplification · efficiency · altitude) | 3/4 clean. Altitude: decoupled the adjacency-vs-causal rationale from the extraction deferral (they were fused). |
| **code-police** (rules + fact-check, `--no-elegance`) | Rules: clean except one deferred (`dry-rule-of-three` on the vitest solid-js alias block — a repo-wide extraction across 5 unrelated packages, outside this PR). **Fact-check: the big one** — see below. |

## The fact-check catch (and why the code didn't change)

Code-police mounted the **real un-mocked `@pierre/trees`** and measured `onSelectionChange` firing **synchronously** in the click dispatch, concluding the "emits on a microtask after" comment was false and the disarm should tighten to `queueMicrotask`.

That's true in happy-dom and **wrong for production**. Ground truth: the `queueMicrotask` disarm *shipped* in #1846 and dropped every real click — if the emit were synchronous, a microtask disarm (which runs after the sync dispatch) couldn't drop it. So the timing is **environment-dependent**: synchronous under a direct mount, **deferred past a microtask** in the real app (Preact into a shadow root + real browser click). `requestAnimationFrame` is the disarm that survives **both**; `queueMicrotask` survives only the synchronous case and reintroduces the regression. So the **code is unchanged** — I corrected the over-absolute claim at all three doc sites to the honest environment-dependent framing.

## Net effect

Behavior is identical to the merged fix. The diff is: `createEventListener` adoption, accurate JSDoc + comments, two extra regression tests (14 total, each negative-control-verified), one SolidJS APM rule, and a changelog entry — all reviewed to consensus.


srid:

✅ **Consensus** after 2 round(s) · lowy + hickey · base `22cb7f5f3b96`

### Applied (2)
- `lowy-1` Receptacle's new contract undocumented at the plug; consumer reaches through to the implementation — commit `e5f6f0475`
- `hickey-1` Provenance token authorizes by temporal adjacency, not causation — commit `d0f6605a9`

### Agreed — no change (3)
- `lowy-2` Gesture-provenance token is a domain-agnostic mechanism — record as a graduation candidate (packages/solid-pierre/src/FileTree.tsx:194)
- `hickey-2` Sibling CodeView wrapper has the identical controlled-component bridge with no gate and no stated reason (packages/solid-pierre/src/CodeView.tsx:173)
- `hickey-3` Gate contract re-derived across five comment sites in two files (packages/solid-pierre/src/FileTree.tsx:176)

srid:
## Codex ⇄ Claude debate

✅ Consensus (2 debate rounds; 1 fix commit) · reviewer effort: xhigh · base `22cb7f5f3`

|---|---|---|
| 1 | 70df16b1e | fix(codex): qualify selection-contract JSDoc; pin capture-phase + frame-disarm in tests |

**Legend** — findings codex raised (both resolved):

- **F1** — The new public contract says selectedPath writes are applied silently and 'never' round-trip and that 'only a real user gesture ..
- **F2** — The regression harness manually emits only after dispatchEvent has completely returned, so it still passes if the production listener loses capture:true; it also never advances a requested animation frame after a gesture that emits nothing, so deleting or breaking the safety disarm still passes

srid:

`--no-elegance` (simplify ran this gauntlet). Base `22cb7f5f3`.

**Pass 1 — rules:** clean on ~35 rules. One finding, **deferred**:
- `dry-rule-of-three` — `packages/solid-pierre/vitest.config.ts` is a byte-identical 6th copy of the solid-js `resolve.alias` block (5 packages already carry it; no shared base config exists). The proper fix extracts a shared helper and rewires **5 unrelated packages'** test configs — outside this focused review PR, and the new file follows the established repo convention. Recorded as a repo-level follow-up, not fixed here.

**Pass 2 — fact-check:** one substantive finding, **partially applied**:
- The agent mounted the **real un-mocked `@pierre/trees`** and measured `onSelectionChange` firing **synchronously** in the click dispatch, and recommended tightening the disarm to `queueMicrotask`.
- **Disputed the fix, applied the doc correction.** The synchronous measurement holds in happy-dom but not production: the `queueMicrotask` disarm *shipped* in #1846 and dropped every real click — impossible if the emit were synchronous. The timing is **environment-dependent** (synchronous under a direct mount; deferred past a microtask in the real Preact-into-shadow-DOM render). `requestAnimationFrame` survives both; `queueMicrotask` reintroduces the regression. **Code unchanged**; corrected the over-absolute "emits on a microtask" claim at all three doc sites (`FileTree.tsx`, the test docblock, the SolidJS APM rule) to the honest environment-dependent framing (`f29d73a8f`).

No silent error-swallowing, no fallback/default defects, capture-phase ordering correct.

PR #1852, opened 2026-07-16
Honest liveness at the surface-remote seam — delete captureLatest

srid:
`readPadiMemoryOnce`'s liveness gate asked `currentClient() !== null` — and that is the wrong question. **`currentClient()` means "dialing-or-connected", never "the far end is live"**: the pump / `markConnected` handshake depends on it being non-null while merely `connecting`, and `scheduleReconnect` *retains the rejected dial across the whole backoff wait*. So a deferred poll read gated on it republished the re-serve mirror's held **stale RSS** through connecting *and entire reconnect backoff windows*. The app-side compensator `captureLatest` was supposed to tame this — but its own premise is false, so the real cure belongs at the seam.

This adds **`Session.currentState()`** — the synchronous point-read twin of `onState`, returning the freshest published `SessionState` frame. Liveness is now `currentState().phase === "connected"` (app policy, named `padiMemoryReadable`), read off the one cell `onState` publishes. `captureLatest` + `padiIsLive` delete; `pollCadence.ts` (its only home) is removed. This is the follow-up SR8.c recorded as *"surface-remote's honest liveness during connecting."*

### The mechanism, corrected

The lens-run's ground pass falsified the inherited "synchronous pre-assignment window" story — the fix is **simpler and bigger** than a one-microtask frame skew:

| | The old `currentClient() !== null` gate |
| --- | --- |
| **`captureLatest`'s premise** | *"snapshots in the same synchronous frame as the state change"* — **false**: `onState` delivery is microtask-deferred (`inMemoryCell`'s fire-and-forget `for await`), so the snapshot captured the *post-assignment* truthy pointer. Its apparent guarantee was an accident of relative microtask ordering. |
| **`connecting`** | truthy (the in-flight dial) → republished stale RSS |
| **whole backoff windows** | truthy — `scheduleReconnect` keeps the rejected `clientPromise` until give-up → republished stale RSS the entire time |

`currentState().phase` changes only on real transitions (each firing `onState`) and never mid-microtask, so a deferred reader can never observe a frame that disagrees with the last `onState` — the honesty is structural.

### `currentState()` is un-hiding, not graduation

The role *had* a synchronous `Session.current()` and dropped it in the reactive-cell refactor; **six consumers hand-rolled an `onState` mirror to stand in for it**. The mechanical ones adopt `currentState()` here — `dialAgentOnce.ts`'s failure-tail mirror and the four test `snap()` helpers. `serveHostMap`'s cache is a recorded follow-up (coupled to `reactiveFamily`). `PadiSession` narrows `currentState` to `Prov` alongside `onState` (a local padi can't report `copying`); `currentClient()` gains a doc fence: *never a liveness gate*.

### Pins (real, red-if-broken)

- **`currentState.test.ts`** (surface-remote) drives real `makeSession` through connecting → connected, a **backoff window** (`currentClient()` non-null while `currentState().phase` is down — the leak the old gate had), and the **disconnected→failed give-up** (a listener delivered `disconnected` reads `currentState().phase === "failed"` — the freshness contract).
- **`padiMemoryGate.test.ts`** (server) pins the gate reads `connected` only.

> **Not behavior-neutral at the gate — that's the fix.** The poll gate tightens semantically: during a padi reconnect the memory rail now reads honest `absent` (—), not a frozen stale figure. Behavior-neutral *on screen* (e2e + seal class).

### Overnight — CI deferred to morning

> Per the overnight ruling (the CI boxes are the flake campaign's tonight), **two-platform CI, the drishti pair, and the e2e/seal lanes run in the morning.** Local `just check` + the unit pins are green.

- **ODU-IMPACT: `none`** — odu consumes `Session` via `makeSession` (`coordinator/lane.ts`), it doesn't implement the interface; a new additive method can't break it. Grounded at odu's pinned kolu SHA `7d51cb2b`.
- **Drishti pair: [srid/drishti#108](https://github.com/srid/drishti/pull/108)** — pins this PR's current HEAD `7a81984d3` (sitting on drishti master's merged SR11 adoption, drishti#107). The LIVE-FIX delta is zero-code for drishti (additive `currentState()`; drishti reads `makeSession`/`reconnect`/`onState`, never `currentClient`/`currentState`).
- **SR11 (#1847), unmerged:** no source overlap (it touches `surface-map`/`surface`/`padi`/`client` + `ref-surface{,-app,-map}.mdx`; this touches `surface-remote`/`server` + `ref-surface-remote.mdx`). The only shared file is the `merge=union` changelog. Merge order is the coordinator's.


srid:

✅ **Consensus** after 4 round(s) · lowy + hickey · base `2da0efa93afc`

### Applied (3)
- `lowy-2` The named memory-rail 'policy' is only half the gate; its pin doesn't cover the real decision — commit `f67c8c949`
- `lowy-3` serveHostMap's comments now assert a falsehood the receptacle just retired — commit `9d6446b0f`
- `hickey-1` Memory-rail liveness policy is fragmented: named predicate holds only the phase half, `!isDestroyed()` stranded inline — (uncommitted)

### Agreed — no change (4)
- `lowy-1` Electricity test: the Session receptacle gains its missing honest point-read (packages/surface-remote/src/session.ts:181)
- `hickey-2` `currentState()` / `currentClient()` asymmetric destroyed-handling forces callers to re-add the guard (packages/surface-remote/src/session.ts:1286)
- `hickey-3` PadiSession Omit/Pick lists `onState | currentState` twice — unenforced ‘narrow every Prov member in both lists’ rule (packages/server/src/padi/padiSession.ts:108)
- `hickey-4` The `currentClient()`-is-dialing-or-connected rationale is duplicated near-verbatim across ~5 sites (packages/server/src/padiMemoryGate.ts:9)

srid:
## Codex ⇄ Claude debate

✅ Consensus · reviewer effort: xhigh · 2 fix commit(s) over 3 review rounds

| # | Commit | Description |
|---|---|---|
| 1 | d494033b6 | fix(codex): correct serveHostMap comment; pin currentState variance + provisioning phases |
| 2 | aca9e6913 | fix(codex): F4 — correct the currentState-fallback comment (SessionState<never>, not full-union) |

- **F1** — The changed documentation says reactiveFamily caches a frame to fuse currentState() with membership, and repeats that claim at lines 265-267, but this implementation never calls currentState(); attach at line 288 still subscribes to onState and caches those delivered frames
- **F2** — The production alias now explicitly Omit+Picks currentState alongside onState so PadiSession<never> and PadiSession<SshProv> preserve different phase vocabularies, but the type-level pin exercises only onState
- **F3** — The test says it covers every up-but-not-connected phase, but its phase type is SessionState with the default Prov=never and its table covers only connecting, disconnected, and failed
- **F4** — The new explanation says dropping currentState from PadiSession's Omit+Pick would fall back to DaemonSession's full-union SessionState

srid:

Ran with `--no-elegance` (the `/simplify` track already covered the elegance pass). Both passes run twice (independent invocations) and agreed.

**Pass 1 — rules checklist:** 1 finding, fixed.
- **`prefer-ts-pattern`** (`packages/server/src/padiMemoryGate.test.ts` `at()` fixture) — the phase→`SessionState` builder dispatched via a `switch`/`default` cascade with per-arm object shapes and a silent shared `default`. Converted to `ts-pattern` `match()` typed at the **closed** `SessionState<SshProv>` union with `.exhaustive()`, so a newly-added phase forces this fixture to handle it rather than silently falling into `default`. Fixed in `d556d0336`.

**Pass 2 — fact-check:** no issues (both runs). Verified: the `padiMemoryReadable` conjunction (`!isDestroyed() && phase === "connected"`) is correct and fully pinned (every phase + the destroyed-with-stale-`connected`-frame case); no TDZ / read-before-init (`padiSession` is constructed synchronously at module scope before `readPadiMemoryOnce` can run; `currentState()` returns an already-initialized cell); the retired `captureLatest` / `padiIsLive` / `pollCadence.*` are fully removed with zero dangling references; the `dialAgentOnce` `onState`-mirror → `currentState()` swap is behavior-preserving and its best-effort `destroy()` catch is justified in-place; the `snap()` → `currentState()` test adoptions leave no orphaned imports; the `PadiSession` `Omit+Pick` variance narrowing is type-pinned; docs (`ref-surface-remote.mdx`, changelog) match the shipped behavior.

PR #1856, opened 2026-07-16
fix(website): rename shadowing toString import (master ci::biome latent red)

srid:
One-line master hotfix: `website/src/remarkChangelog.ts` (from #1842) imports `toString` from `mdast-util-to-string`, tripping biome's `noShadowRestrictedNames` — deterministic, so **every PR that merges current master inherits a ci::biome red** (first bitten: #1852, which carries the same rename in-branch; LIVE-FIX flagged it for a source fix).

Verified with ci::biome's exact invocation (`biome lint --error-on-warnings`): red on master's file, clean after `toString as mdastToString`. Same rename #1852 used, so the two resolve cleanly.

How it slipped: #1842's final commits added this `.ts` file after its last full-pipeline run, and the coordinator-scoped re-gate (`ci::nix` — the "docs-only" class gate) doesn't run biome. The class call was wrong at the margin: a PR that adds website *code* needs the code gates.

### DIFF
--- a/website/src/remarkChangelog.ts
+++ b/website/src/remarkChangelog.ts
@@ -1,5 +1,5 @@
 import GithubSlugger from "github-slugger";
-import { toString } from "mdast-util-to-string";
+import { toString as mdastToString } from "mdast-util-to-string";
 import type { Nodes, Root } from "mdast";
 import type {
   MdxJsxAttribute,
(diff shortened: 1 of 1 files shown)


srid:
Mooted: #1852 carried the identical rename in-branch (LIVE-FIX applied it to unblock its own CI after flagging the latent red) and merged first — master's remarkChangelog.ts already imports `toString as mdastToString`, which is also why this PR now conflicts (two identical fixes on the same lines). Master's ci::biome is clean; nothing left for this PR to do.

PR #1858, opened 2026-07-16
A stdio agent's lifetime belongs to the framework: serveOverStdio owns the exit

srid:
**A subprocess agent served over the default `process.stdin`/`stdout` now dies with its link.** `serveOverStdio` used to resolve `{reason}` and leave process termination as every app's homework — so any live handle (drishti-agent's 2s poll interval) kept a dead agent alive forever. A fleet host was found carrying **ten orphaned `--stdio` agents, 130–200 MB each, up to 23 h old, re-parented to pid 1** — one per server⇄host reconnect ([srid/drishti#109](https://github.com/srid/drishti/issues/109)). The same class litters CI boxes as odu lane-agent leaves.

The fix binds lifetime at the construction that already means "this process IS the agent" — no knob, no new detection:

| construction | after this PR |
| --- | --- |
| `serveOverStdio({ router })` — default stdio | promise settles (callers' sync `dispose()`/log continuations run first, via a `setImmediate`-scheduled fork), then **framework exits**: `0` on `"end"`, `1` on `"error"` |
| `serveOverStdio({ router, transport })` — loopback/test/socket/embedded | caller keeps owning the process lifetime |

*"Agent process alive, link dead" is no longer a spellable state* — the app can't forget what it no longer has to remember. There is deliberately **no async last-gasp hook**: all three in-repo default-arm consumers (mini-ci runner, remote-process-monitor agent, fleet-top part-3 agent) do strictly-sync post-settle work, which the ordering already honors; the future shape (a partitioned options type so `{transport, onEnd}` fails typecheck) is recorded in the header.

### One classification seam for how a link dies

The write-error funnel used to blur two concerns; the review gauntlet distilled it to one honest shape. `ServeOverStdioEnd` is now a **discriminated union** (`{reason:"end"} | {reason:"error", error}`), and `reason` is decided at exactly **one terminal seam**, shape- and arm-independently:

- a **benign write death** (`EPIPE`/`ERR_STREAM_DESTROYED` — the codec's existing `isBenignWriteError`) is clean teardown: on a parent death, a pushing agent often sees stdout-EPIPE before stdin-EOF, and carrying that race as an error would nondeterministically flip the same clean teardown between exit 0 and exit 1 (which ssh propagates to `Restart=on-failure` units and CI wrappers) — a **real latent bug** the new exit semantics would otherwise have codified;
- the **shared-duplex shape** (`serveOverUnixSocket`'s one-socket transport, where Node marks the stream destroyed before emitting `'error'`) flows through the same seam — caught by the codex debate, pinned;
- a **callback-only death** (`write()` on a stream `destroy()`ed without an error reports `ERR_STREAM_DESTROYED` only to the write callback — no `'error'` event) previously left both sides deaf: the serve promise pending forever (the orphan, re-spelled) and client calls hanging. Also codex-caught. The internal `framedSend` now takes a **required** `onPeerGone` teardown callback — *an internal two-caller obligation, the inverse of the rejected `onEnd` (public zero-caller machinery)*: `stdio-codec` is not an exported subpath, so this is a seam obligation inside the package, not API surface; omitting the callback (the deaf-owner send) is now a type error.

### Pins (red first, on a pu box)

`peer-server.lifetime.test.ts` spawns a **real child process** (`node --import tsx` + a fixture holding a live `setInterval`) — reproduced red on pre-fix code (child immortal after link close, 4 s deadline), green now:

- stdin EOF → exit 0, with the caller's post-settle log asserted to have run first
- parent read-side death mid-push (benign write EPIPE) → exit 0
- genuinely abnormal read error → exit 1
- override arm (in-process loopback + `process.exit` spy) → resolves, never exits

Plus loopback pins for every write-death shape: non-benign (coded EIO) → `"error"`; benign coded EPIPE → `"end"`; shared duplex → `"end"`; silently-destroyed write under a live request → `"end"`; and client-side: a call against a silently-dead write rejects fast instead of hanging.

### Design philosophy

- **Fail fast, no knobs** — the exit is framework-owned with no opt-out; the rejected alternative (an `onEnd` escape hatch) was cut at the design gate as speculative machinery for zero callers.
- **Volatility boundary** — the change lives entirely in `@kolu/surface`'s peer-server seam; consumers change nothing.
- **Reuse the source of truth** — the lifetime discriminant is the *existing* transport-override flag; write-death classification reuses the codec's *existing* `isBenignWriteError`, at one seam instead of two.

> **drishti pair PR**: [srid/drishti#110](https://github.com/srid/drishti/pull/110) — pinned to this PR's final HEAD (`c0b631c69`), **CI green on both platforms, MERGEABLE**; closes drishti#109 by inheritance (the agent has no hand-rolled exit to delete — that absence *was* the bug).

**ODU-IMPACT VERDICT: `adoption-opportunity`** — grounded: at odu master (`b61cd25`, kolu pin `07397fa2`), `git grep serveOverStdio` shows `src/runner/main.ts:34` serving lane agents on the **default transport** (sync-only post-settle: `runner.dispose()` + log) and `src/odu.test.ts:45` on the override arm. Lane agents inherit the orphan fix at the next pin bump; tests unaffected. *Drained immediately: [juspay/odu#45](https://github.com/juspay/odu/pull/45) bumps the pin to this PR's final HEAD (box-validated: typecheck + 141/141 tests) — [odu#43 ledger line](https://github.com/juspay/odu/issues/43#issuecomment-4995889479) recorded as drained.*

_Generated by [`/be`](https://github.com/srid/agency) on Claude Code (model `claude-opus-4-8[1m]`)._


srid:

✅ **Consensus** after 1 round(s) · lowy + hickey · base `77116330f60e`

### Applied (5)
- `lowy-2` Client link mis-narrates benign write deaths the server now classifies — commit `28daa2aee`
- `lowy-3` Example agents — the default arm's teaching consumers — narrate the pre-fix lifetime — commit `11d0574c1`
- `hickey-1` Never-rejects invariant breakable after it's established; new exit fork silently depends on it — commit `a698e4469`
- `hickey-2` Funnel's benign path rides an undocumented codec contract (resolve-on-'close') — commit `e40985231`
- `hickey-3` Contract module's recorded justification is wrong for type-only imports — commit `d6644444d`

### Agreed — no change (6)
- `lowy-1` Electricity test (required): receptacle named, volatility correctly homed (packages/surface/src/peer-server.ts:171)
- `lowy-4` Explicit process-stdio transport re-spells the 'unspellable' orphan state (packages/surface/src/peer-server.ts:53)
- `lowy-5` console.logToStderr back-door exposes the wires (pre-existing) (packages/surface/src/peer-server.ts:194)
- `lowy-6` Durable sibling frontDaemonOverStdio still drain-owns its front-process lifetime (packages/surface-daemon/src/frontDaemonOverStdio.ts:235)
- `hickey-4` stderrMarker complects stderr collection with marker waiting (packages/surface/src/peer-server.lifetime.test.ts:82)
- `hickey-5` Arm-independence narrative triplicated in-file, held in sync by discipline (packages/surface/src/peer-server.ts:57)

srid:
## Codex ⇄ Claude debate

✅ Consensus in 3 rounds (2 fix commits; round 3 verification-only) · reviewer effort: xhigh · base: merge-base 77116330f

|---|---|---|
| 1 | 7ef3262f5 | fix(codex-r1): close the two benign-write blind spots; honest comments (F1-F4) |
| 2 | 7c33b0e53 | fix(codex-r2): onPeerGone is required — the deaf-lifecycle-owner send is unspellable (F5) |

- **F1** — The benign-write funnel does not produce reason "end" when read and write are the same Duplex, which is the canonical serveOverUnixSocket transport
- **F2** — framedSend converts every benign callback error into success on the assumption that the stream error event has already started lifecycle teardown, but Node does not emit error when write() is called on a stream that was previously destroy()ed without an error: it reports ERR_STREAM_DESTROYED only to the write callback
- **F3** — The new never-rejects rationale names an in-flight request's AbortSignal listener as the exact synchronous throw caught around peer.close(), but Node's EventTarget does not throw that listener error from AbortController.abort(); it rethrows it on process.nextTick
- **F4** — The pre-funnel comment still says every write error settles { reason: "error", error } exactly like a read error, contradicting the benign-write branch introduced immediately below and the public contract above.
- **F5** — The newly added `onPeerGone` parameter is optional even though the surrounding contract establishes that omitting it leaves the lifecycle owner deaf to callback-only ERR_STREAM_DESTROYED

srid:

Two independent rules runs + two independent fact-check runs over the merge-base diff (with the elegance pass skipped — `/simplify` ran standalone in the gauntlet).

**Rules** — clean except two trivia, both fixed in `fix(police): …`:
- `dry-rule-of-three`: the post-settle lifetime comment was triplicated verbatim across the three example agents → one-line pointer at the canonical `serveOverStdio` Lifetime doc.
- `no-silent-error-swallowing`: the force-a-response test call's empty catch now carries its justification.
- Actively conformant, worth noting: `no-overloaded-null` (the `ServeOverStdioEnd` discriminated union *fixes* a latent instance), `silent-handler-required-on-void-subscriptions` (required `onPeerGone`), `callback-fanout-guarded-at-funnel` (`endServing`).

**Fact-check** — all five teardown shapes (clean EOF, benign EPIPE both orders, non-benign write, callback-only `ERR_STREAM_DESTROYED`, shared duplex) traced to the same classified settle; the ordering invariant and consumer narrowing verified; Node-semantics claims verified empirically. Two comment-precision defects fixed:
- the exit fork's `process.nextTick` justification was empirically false (a tick runs *after* the direct microtask cascade; the true margin is tick-mediated continuation work like `emitCloseNT` — `setImmediate` runs after both);
- the "honest limit" window commits at read-stream *destroy* (one tick before the `'close'` settle), not at the settle.

Recorded out-of-scope: `readFramedLines` keeps consuming after a decode-failure settle → #1859 (pre-existing, shared codec).

Suite at HEAD on a clean box: **42 files / 424 tests green**, including the red→green lifetime pins.

srid:
## Evidence

Server-only change — the proof is behavioral: **does the agent process die with its link?** Captured on a clean ephemeral box (`stdio1-red-repro`), driving the real fixture agent (a `serveOverStdio({ router })` default-transport child holding a live `setInterval` — the drishti#109 shape) over real pipes, watched by **exact recorded pid** (no pattern matching):

**Pre-fix (`peer-server.ts` at merge-base `77116330f`):**
```
agent up: pid=7540 (live setInterval inside)
t=0ms link closed (stdin EOF)
t=10043ms agent pid=7540 STILL ALIVE — the immortal orphan (drishti#109). Killing exact pid 7540.
```

**At HEAD (`c0b631c69`):**
```
agent up: pid=7416 (live setInterval inside)
t=0ms link closed (stdin EOF)
t=101ms agent pid=7416 EXITED — orphan unspellable
```

Same box, same fixture, same live handle: pre-fix the agent survives its link indefinitely (the exact mechanism behind the ten 130–200 MB orphans on sincereintent); at HEAD it is gone in ~100 ms.

The pinned equivalents ran red→green through the gauntlet: red pre-fix (3 lifetime pins failing on "child did not exit within 4000ms"), green at HEAD (**42 files / 424 tests**), including the write-side legs (benign EPIPE exit 0; shared-duplex `"end"`; callback-only `ERR_STREAM_DESTROYED` settling instead of hanging) and the override-arm no-exit pin.

*Mobile/coarse-pointer axes: N/A — no on-screen surface.*

PR #1840, opened 2026-07-15
Deep links — the hash-URL router + the full view menu (DL1)

srid:
Any place that can render a link — an orchestrator board, an issue, a Slack message, a bookmark — can now point **into** kolu. Click a `#/…` hash URL and the app lands on the right host, terminal, file, or dialog; the installed PWA window comes to the front instead of spawning a second one.

This is **DL1** from the [deep-links Atlas note](../blob/master/docs/atlas/src/content/atlas/deep-links.mdx) — the router + the full URL menu. **DL2** (the URL following focus) stays gated.

## The grammar (v1 ships all of it)

| URL | lands on |
|---|---|
| `#/h/<host>` | the host binding switched to `<host>` |
| `#/t/<host>/<id>` | that host **and** the terminal tile focused |
| `#/t/<host>/<id>/code?path=<p>&line=<n>` | the Code tab, `<p>` open (at line `<n>`) |
| `#/t/<host>/<id>/inspector` | the Inspector tab |
| `#/settings` | the settings popover |

`<host>` is the canonical encoded key (`local`, or `remote:<target>`); `<id>` is the terminal UUID — validated the same way an attention-notification click is (`isEncodedHostKey` + `TerminalIdSchema`), so a malformed key or id is dropped, never routed.

## Shape

- **`deepLink.ts`** — the grammar as a discriminated union, parse-don't-validate. Output is a total `DeepLink`, or `invalid` (toast + home), or `none` (empty hash, silent). A 34-case unit table pins valid × each family and invalid host / id / route / line.
- **`useDeepLinks.ts`** — three entry points (boot parse of `location.hash`, `hashchange`, and Chromium's `launchQueue`) feed the one parser. Each route lands on an **existing** view seam: `setActiveHost`, `store.activate`, the right panel (`showInspector` / `openInCodeTab`), the settings signal. The router adds addressability, never a new capability.
- **Splits (4a):** a link to a sub-terminal walks the full chain — `store.activate(parent)` + `expandPanel` + `setActiveSubTab` — so the sub pane lands live, not just its parent tile.
[...]

PR #1864, opened 2026-07-16
Deep links route from inside the Code-tab preview (DL2)

srid:
The deep links #1840 shipped become a **felt** feature: a `#/…` link clicked **inside the Code-tab preview** now routes the **parent** kolu app — focus that terminal, open that file at a line, open settings — instead of dead-clicking in the sandbox. The orchestrator dashboard's lane pills (`<a href="/#/t/local/<id>">`) are now click-to-jump.

## The mechanism — one new message kind through the existing bridge

The preview is an opaque-origin sandbox (deliberately no `allow-same-origin`) with an established postMessage bridge; DL2 adds exactly one message kind to it:

| Layer | Change |
|---|---|
| in-iframe (`artifact-sdk/iframe`) | `koluDeepLinkHash()` — a pure classifier: a **same-origin** anchor whose fragment starts `#/` and performs **no file navigation** (path is the app root or the document's own). External links and internal file navigation keep precedence. On a hit: post `{type: "kolu-artifact-sdk:deep-link", hash}` + `preventDefault` |
| parent (`artifact-sdk/client`) | `observeIframeDeepLink()` — the same `event.source` identity boundary as its siblings; payload guard re-bounds the hash to the `#/` namespace |
| client | `BrowseIframeRenderer` gains `onDeepLink`, threaded through `BrowseFileDispatcher`; `CodeTab` wires it to the router |
| router (`useDeepLinks`) | `requestDeepLinkNavigation()` — a request signal (`equals: false`, so re-clicking the same pill re-routes) consumed by the hook through the **same external pipeline as a PWA launch**: reflect into the address bar when the hash differs (durability), else parse+navigate directly. An invalid hash toasts **exactly as a typed URL does** — same code path |

Only the cheap shape test lives in the sandbox; the parent runs the real grammar parse.

## Safety (the crux)

**No new capability surface.** The router is **view-only by law** (#1840, grep-pinned): no route creates a terminal, kills one, writes a file, or sends keys. So a hostile previewed document that fabricates deep-link messages (any in-frame script can `postMessage`) can at worst change what the viewer is looking at, or raise the invalid-link toast. That is precisely why #1840 shipped view-only first — DL2 is safe **by construction** on that law. No sandbox relaxation: no `allow-same-origin`, no new sandbox tokens — the bridge is the point.

## Pins

- **Classifier unit table** (`iframe/deepLink.test.ts`, pure — no DOM): the dashboard-pill shape (`/#/t/…`), a same-page `#/…` href, cross-origin `#/…` stays external, different-path stays file-nav, bare `#`/`#section` never post, non-web schemes never post.
- **Observer unit** (`client/bridge.test.ts`): fires on a well-formed message; ignores foreign sources; drops out-of-namespace and non-string hashes.
- **e2e felt proof** (`deep-links.feature`): a previewed HTML pill names a real terminal via its own `$KAVAL_TERMINAL_ID`; clicking it inside the iframe focuses that terminal in the parent (the full bridge→router path).

Per the dispatch: **no review gauntlet** (srid dogfoods by clicking the dashboard pills once deployed); the e2e scenario is the recorded proof of the flow. Diff is `packages/client` + `packages/artifact-sdk` only — no `@kolu/surface*` touched.


srid:

✅ **Consensus** with zero debate turns · lowy + hickey · base `93486c2f42e4`

Settled: 3 reconciled (raised by both lenses) · 3 auto-settled minor

### Applied (3)
- `lowy-1` Deep-link namespace predicate spelled twice, breaking the package's own isHttpUrl receptacle pattern — commit `9c3474202`
- `lowy-2` ≡ `hickey-1` Anchor-click classification split across two modules with a comment-borne precedence contract that is not actually load-bearing — commit `6bd196828`
- `hickey-2` Module header still enumerates three entry points; the diff adds a fourth — commit `1a0a0a168`

### Agreed — no change (3)
- `lowy-3` ≡ `hickey-3` Four iframe-intent callbacks now drilled verbatim through BrowseFileDispatcher (Layer-0 pattern-as-artifact) (packages/client/src/right-panel/BrowseFileDispatcher.tsx:117)
- `lowy-4` ≡ `hickey-4` VIEW-ONLY-BY-LAW threat-model rationale duplicated in prose across five sites (packages/artifact-sdk/src/types.ts:99)
- `hickey-5` Parent app-shell root path '/' baked into the sandbox classifier (packages/artifact-sdk/src/iframe/deepLink.ts:30)

srid:
## Codex ⇄ Claude debate

✅ Consensus in 1 round(s) · reviewer effort: xhigh · base: origin/master

|---|---|---|
| 1 | 7b39bf0c1 | fix(codex): round 1 — history pin reachable (one routed link), count comments |

**Legend** — findings codex raised (it ran the e2e feature itself to verify — the round-1 catch was a deterministically unreachable assertion):

- **F1** — The history regression scenario cannot reach its second preview link
- **F2** — The feature preamble still says three delivery paths feed the parser, but this diff adds the Code-tab preview bridge as the fourth path and `useDeepLinks.ts` now documents four.
- **F3** — The shared binding helper comment still says all three observers use it, but the component now binds navigation, history, external-open, and deep-link observers.

srid:

Pass 1 (rules) + Pass 2 (fact-check) over the DL2 diff; elegance skipped (`/simplify` ran). **2 findings, both fixed** (commit `3c23e8715`):

| Pass | Finding | Fix |
|---|---|---|
| fact-check | **Gesture bug**: the anchor trap is button-agnostic (safe for external links — every gesture → a new tab), but the deep-link arm navigates the **current** window, so middle-click / ctrl/cmd/shift/alt-click silently hijacked the view the user meant to keep | deep links fire only on a plain primary click (`isPlainPrimaryClick`, pure + unit-pinned); open-elsewhere gestures stay inert as before |
| no-thin-wrapper-functions | `requestDeepLinkNavigation` was a single-caller pass-through around the signal setter | the setter **is** the exported API under that name (the `setActiveHost` convention) |

Everything else clean: no dead code (`externalHref` fully gone), listener lifecycle correct across the four observers, one-parser fidelity verified end-to-end, the e2e's history pin confirmed real (ordering-safe after the polled marker step).

PR #1862, opened 2026-07-16
A daemon bin's exit belongs to the framework: daemonProcessMain

srid:
**A daemon binary built on `daemonMain` no longer owes the framework its own death.** `daemonMain` resolves a `DaemonExit` and deliberately never exits (that's what keeps the whole lifecycle drivable in-process from a test) — but that left the actual `process.exit` as every bin's homework, and any live resource or timer the bin never tied to the daemon's lifetime (a PTY child, a served runtime, a sampler interval) silently kept a *finished* daemon alive. The same class #1858 killed at the link layer, re-appearing at the daemon layer — step 2 of the ratified [surface-lifetime-audit](https://kolu.dev/atlas/surface-lifetime-audit.html) sequencing ("tenure, not transports").

### The partition

| half | owns | exits? |
| --- | --- | --- |
| `daemonMain(spec)` — **unchanged** | gate → serve → teardown; resolves `DaemonExit` | never (the testable value core; every existing test untouched) |
| `daemonProcessMain({ name, run })` — **new** | the rest of the process's life | always: `daemonExitCode(exit)` on resolve; narrate + `1` on a crash |

The crash arm is **swallow-proof**: a non-`Error` rejection's message is derived totally, the stderr narration is guarded, and `process.exit(1)` is unconditional — a broken pipe on a dead parent cannot block the exit (pinned). The ordered teardown *stop accepting → release gate → release resources → exit* holds by construction: the wrapper `finally`s (kaval's `ptyHost.close()`, padi's `served.close()`) sit inside `run()`'s promise, so the exit cannot preempt a release stage (pinned, marker-ordered).

kaval, padi, the fleet-top example, and the docs snippet adopt it — no bin hand-rolls the `.then(exit => process.exit(...))` + `.catch` pair anymore. `serveOverStdio`'s exit fork stays **align-only** (shared doctrine, cross-referenced comments): the dependency arrow points surface-daemon → surface, so sharing code would be a cycle, and a third package for ~20 lines hides no volatility.

### Receipts (pu box, real child processes)

| bin shape | SIGTERM → | verdict |
| --- | --- | --- |
| naive (`daemonMain` + live interval, no homework) | `DaemonExit` resolved + release ran at t=265ms | **STILL ALIVE at t=10,275ms** — killed only by the harness |
| `daemonProcessMain` | resolved + released, markers in order | **EXITED code=0 at t=205ms** |

[...]

PR #1866, opened 2026-07-16
fix(lint): master ci::biome red — dashboard skill assets (from #1854)

srid:
Master hotfix: #1854 (the docs/atlas branch) carried the orchestrator dashboard's `index.{js,html}` — **code files that never met the code gates** — leaving master deterministically red on `ci::biome` (3 `useIterableCallbackReturn`, 1 `useTemplate`, 3 `noDescendingSpecificity`). Every PR merging master inherits the red until this lands; first bitten: #1862 (TENURE2), holding its CI on this.

Fixes: braced the value-returning `forEach` callbacks, templated the one concat, moved the `section > *` grid-span rules below the base rules they override. All three skill copies synced. `just ci::biome` green locally (1370 files). Renderer behavior unchanged (parse-verified; DOM calls identical).

Same failure class as #1856: a "docs" branch whose diff grew code — the class gate is a property of the diff, not the label. This time the offending committer was the coordinator itself.

### DIFF
--- a/.agents/skills/orchestrator/dashboard/index.html
+++ b/.agents/skills/orchestrator/dashboard/index.html
@@ -49,8 +49,6 @@
     padding:0 12px; display:grid; gap:10px;
     grid-template-columns:repeat(auto-fill, minmax(360px, 1fr));
   }
-  section > h2 { grid-column:1 / -1 }
-  section > .pills, section > .empty { grid-column:1 / -1 }
   h2 {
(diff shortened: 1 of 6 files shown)


srid:
Mooted: TENURE2's #1862 carried an equivalent dashboard lint fix (264f6da9e) into master on merge, so master's ci::biome is green without this. Same resolution as #1856 — a parallel fix raced in first.

PR #1867, opened 2026-07-16
Back and forward never replay a consumed deep link (DL3)

srid:
srid's post-deploy dogfood caught the back button still teleporting: mouse-back landed on a previous deep link's terminal. The repro (run live against the deployed build, fresh window, per the DL3R dispatch) pinned a mechanism the #1864 history fix couldn't reach: **`hashchange` fires on history traversal too**, and the router treated a browser-restored old hash exactly like a fresh command. Every hash-bearing entry in the stack — a typed URL visit, a cold launch — was a live teleport, in **both** directions (measured: forward teleported focus to B, back teleported it back to A, deterministically).

## The fix — commands are consumed once per history entry

After `navigate` handles a hash, it stamps the current entry: `history.replaceState({koluRouted: true}, "")`. The `hashchange` entry point skips stamped entries — a traversal reverts the URL **silently** and the view stays put. Nothing else reads the gate, which is what preserves every standing law:

| Law | Why it holds |
|---|---|
| **Durability** — reload re-navigates | the boot parse never consults the gate (the stamp survives reload; only `hashchange` reads it) |
| Fresh navigations route | a new entry (typed hash, in-page `#/…` anchor, script push) carries null state |
| Repeated same-pill clicks re-route | the preview bridge / PWA launch paths bypass `hashchange` entirely |
| Bad links stay loud once, not repeatedly | `invalid` stamps uniformly — traversing onto a bad-link entry doesn't re-toast |
| Pre-fix stacks self-heal | an unstamped legacy entry routes once on first traversal, then is stamped |

## Pins

- **The repro recipe, inverted** (e2e): follow two live `#/t` links → back: the hash reverts to link 1 but focus **stays** on terminal 2; forward: same. (Before the fix this scenario teleports at each traversal.)
- **Repeat-routing intact** (e2e, extends the DL2 preview scenario): refocus the previewing tile — its per-terminal panel restores the preview — and click the **same** pill again; it re-routes.
- All existing deep-link scenarios (fresh-route, cold-boot, split, code, settings, invalid, history-length) unchanged and passing.

## Enumerated, deliberately not fixed here

The repro's writer census found one **conditional** writer left: `IframeRenderer` binds `src={props.url}` on a kept element (`solid-fileview/renderers/iframe.tsx`), so a `?v=<content-hash>` change (a live dashboard rewriting itself while previewed) is a joint-history push snapshotting the then-current hash. It did not fire during the instrumented session (idle board) — per the dispatch it's fix-only-if-it-independently-reproduces. Note that with this PR's gate, traversing such an entry no longer teleports either (the top-level state is carried), which further de-fangs it.


srid:

✅ **Consensus** after 1 thread-round(s) · lowy + hickey · base `517457af7f6f`

Settled: 2 reconciled (raised by both lenses) · 1 auto-settled minor · 2 unopposed · 1 debated

### Applied (3)
- `lowy-1` ≡ `hickey-1` koluRouted stamp protocol spelled twice with no shared definition — commit `58d6d1b29`
- `lowy-2` ≡ `hickey-2` Back/forward step definitions are verbatim duplicates with a twice-spelled magic settle beat — commit `758180983`
- `hickey-3` Wall-clock 500ms sleep where an observable event exists — commit `2ad8f08d8`

### Agreed — no change (3)
- `lowy-3` ≡ `hickey-4` Supersession policy now spelled at two sites without a name (packages/client/src/useDeepLinks.ts:408)
- `lowy-4` Graduation sweep: consumed-once channel is app policy, not electricity — recorded, no extraction (packages/client/src/useDeepLinks.ts:388)
- `hickey-5` DL2 scenario now pins two independent contracts (packages/tests/features/deep-links.feature:128)

srid:
## Codex ⇄ Claude debate

✅ Consensus in 2 round(s) · reviewer effort: xhigh · base: origin/master

|---|---|---|
| 1 | 11085d28c | fix(codex): a superseded route disarms its hydration intent too |
| 2 | 198e394d3 | fix(codex): round 2 — re-pair the navigate/supersede doc blocks (nit) |

- **F1** — Traversal supersession clears only `pending`, but `routeToTerminal` also armed `deepLinkFocusIntent` at line 168
- **F2** — Inserting `supersedeInFlightRoute` between the existing `navigate` documentation and `navigate` left the comment beginning `Route a parsed link` immediately above the helper instead of the function it describes

srid:

Pass 1 (rules) + Pass 2 (fact-check) over the DL3 diff — each pass run **twice** on independent cold contexts; elegance skipped (`/simplify` ran). **5 findings fixed** (commit `6dfa0d394`), 2 candidates examined and kept deliberately:

| Pass | Finding | Fix |
|---|---|---|
| fact-check + invalid-states (both cold P1/P2 converged) | **Fault verdicts left the intent armed**: the list-error / terminal-gone / backstop-timeout arms cleared `pending` bare, so `deepLinkFocusIntent` stayed pointed at a route the router had already toasted a failure for — a late-settling host's first hydration (or a host switch back to a not-yet-seeded host) would steer focus to a command declared dead, the same zombie-command class the traversal gate closes | all three fault arms route through the pair-disarm helper, renamed `supersedeInFlightRoute` → **`disarmInFlightRoute`** (its causes now include fault termination, not just supersession — the name must not lie); the enacted-route clear stays bare **on purpose** and now says so |
| dry-rule-of-three | `koluRouted` spelled independently in the writer literal and the reader's unchecked cast — a writer-side rename would silently turn the gate always-false and resurrect the DL3 bug itself | one `ROUTED_STAMP` constant; the reader casts `Partial<typeof ROUTED_STAMP>`, so a desync is now a compile error |
| fact-check | **The DL3 pin's forward leg proved nothing**: after back, forward re-entered the `{id2}` entry — but a wrongful forward replay routes to the already-active terminal, so the assertion passed in both worlds | back ×2 (onto the boot entry — the gate holds on non-link entries too), then forward onto the `{id1}` entry **while terminal 2 is active**: a replay teleports to terminal 1 and fails the pin |
| fact-check | The DL2 scenario's "a second pill can't be click-tested here" comment was disproved by the repeat-pill steps appended right below it | stale clause deleted; the history-length rationale stays |
| fact-check | `traverseHistory` waited unboundedly on `hashchange` — a no-op traversal or a same-hash neighbour would hang into the opaque cucumber step timeout | bounded: rejects at 5s with a named error (`history.back() produced no hashchange — …`) |

**Kept deliberately** (examined on the record): `stampEntryRouted` superficially matches no-thin-wrapper-functions, but it has no argument forwarding and exists to pair the protocol's write side with its reader (the lens-ratified receptacle — now both bound to `ROUTED_STAMP`); `traverseHistory` hand-rolls what `page.goBack()` offers because it synchronizes on the `hashchange` dispatch itself — the exact seam a wrongful route runs inside — which Playwright's navigation lifecycle doesn't pin.

The three ratified decisions were verified as implemented verbatim: the consumed-once design (gate read **only** on `hashchange`; boot/bridge/launch route unconditionally), the `pending() !== null` guard (enacted intents survive), and the enumerated-not-fixed IframeRenderer joint-entry writer.

srid:
## 🔴→🟢 CI round 1: the repeat-pill red, root-caused and fixed (`a7d46d00b`)

Run 1 on `6dfa0d394`: **both platforms green except `ci::e2e`**, and the red was ONE scenario, ONE leg, **deterministic** (5/5 attempts across linux+darwin): the DL2 preview scenario's repeat-pill extension — re-clicking the same pill after a terminal switch-away-and-back never re-routed.

**Root cause (instrumented on the CI box):** a reactive feedback loop introduced by this PR's codex-F1 fix. The preview-bridge `createEffect` calls `navigate()` from a tracking scope, and the disarm guard put a `pending()` **read** inside `navigate` — subscribing the effect to the very signal `navigate` writes. After the FIRST bridge-delivered link the app busy-loops `effect → navigate → setPending → settle/enact → setPending(null) → effect` forever; the repeat-pill leg is the only scenario that interacts *after* a bridge route, so it alone caught it. The trace: `["effect:…","settle:…","enact:…"]` repeating unbounded. (Shipped DL2 is loop-free — the read exists only in this PR.)

**Fix:** `navigate` is a COMMAND, not a derivation — its whole body runs under `untrack()` at the one funnel all four delivery paths share. A unit pin holds the spelling; the e2e leg stays as the felt-symptom pin.

**Evidence:** `deep-links.feature` 10/10 on kolu-ci-5 with only the fix applied (was 1 red, deterministic); client vitest 720/720. Full two-platform CI re-run on `a7d46d00b` queued next.

srid:
## ✅ CI round 2 — two-platform green on `a7d46d00b`

Full `ci::default` pipeline **passed on both platforms** (linux: kolu-ci-5 · darwin: rasam), all nodes green — e2e included, with the previously-deterministic repeat-pill red now passing everywhere. PR is ready for review.

PR #1873, opened 2026-07-17
An abandoned session no longer immortalizes its host process

srid:
**`makeSession`'s internal timers are now `unref()`'d, so a session that was dropped without `destroy()` can no longer keep its host Node process alive forever.** The surface-lifetime audit confirmed the shape: an abandoned session redials a gone endpoint eternally, and between dials its ref'd reconnect-backoff timer is often the *only* event-loop handle — the process reaches the end of its main script and simply never exits. `dialAgentOnce` already carried a defensive comment about exactly this leak; this PR retires the hazard at its source.

This is **not** a blanket unref — the plan of record is a per-timer census with an honest unref-vs-must-fire verdict for each (Atlas note [`session-timer-unref`](https://htmlpreview.github.io/?https://github.com/juspay/kolu/blob/session-unref/docs/atlas/dist/session-timer-unref.html), gate-approved):

| Timer | Verdict | Why |
| --- | --- | --- |
| reconnect backoff | **unref** | the immortalizer; still fires under any other loop hold, so a held session reconnects as before |
| connect watchdog | **unref** | by the time it fires, `pin()` has already settled — its effect is `onState`-internal |
| clock-probe retry + deadline | **unref** | internal probe state; when a probe can matter, the RPC's transport holds the loop |
| admit-handshake timeout | **ref'd (must fire)** | it rejects a promise a caller is *awaiting* via `pin()`; a pending await holds no handle, so unref'ing it could exit the process silently mid-await. Bounded ≤30s and self-clearing, so it can't immortalize |
| liveness heartbeat | already unref'd | `@kolu/surface`'s `createHeartbeat` — census row only |

*One owned semantics change:* a wait derived purely from session state — a pump parked on `ClientCursor.next()` across a reconnect gap — is **not a process hold**. Its process-holding class is pump loops inside server processes (every grounded consumer); a process with nothing else left is precisely the abandoned shape that should die. Stated in `pin()`'s contract and the [`ref-surface-remote`](https://github.com/juspay/kolu/blob/session-unref/website/src/content/surface/ref-surface-remote.mdx) lifetime invariants.

### Red-first pins (`processExit.test.ts`)

Event-loop holding is invisible inside a vitest worker, so both pins spawn a **real node child** running a fixture under tsx's loader (the `kaval` `socketDaemon.test.ts` shape):

- **Immortalization red** — abandoned pinned session, endpoint gone, main script ends → child must exit (its backoff delay ≫ the exit deadline, so a clean exit can't be "the timer fired"). Red before the fix (child SIGKILLed at deadline), green after.
- **Must-fire guarantee** — the handshake timeout as the process's *only* handle still delivers its rejection to the `pin()` awaiter (marker + elapsed ≥ timeout), and the process then exits through the unref'd-backoff window. The full surface-remote suite (172 tests) stays green — unref'd timers fire normally under the runner's loop hold.

### Consumer gate (`.claude/rules/surface.md`)

- **drishti** (grepped at `f3609d0` master + pair tip `7e358fd`): `makeSession` lives in the app server's warm host pool; the server's listeners hold the loop. **No API delta, no behavioral impact — no drishti pair PR.** (Re-check at final HEAD before merge-ready.)
- **odu-impact verdict: `none`** (grepped at `914c388` master + pin-bump tip `647d348`): `lane.ts` already `unref?.()`s its own lane deadline and bounds connect attempts itself — this change aligns the session with odu's posture. Nothing for the odu#43 ledger.
- **kaval-tui / padi-tui** one-shot CLIs dial via `dialAgentOnce`, which destroys the session on every pre-`Connection` failure and never parks on a cursor wait — unaffected.

> Design-philosophy note: fail-fast is preserved (nothing degrades — the must-fire timer keeps its guarantee, and the exit-safety invariant the backoff relies on is stated in the note so a future async-dial refactor can't silently break it); no new knobs; the fix reuses the codebase's existing posture (`createHeartbeat` already unrefs, odu already unrefs its lane deadline).


srid:

Settled: 1 reconciled (raised by both lenses) · 3 auto-settled minor · 1 unopposed · 1 debated

### Applied (4)
- `lowy-1` ≡ `hickey-1` Timer-hold policy enforced by trailing .unref() convention at three sites, not a seam — commit `03eb5d17b`
- `lowy-2` tsx-loader resolution diverges from the kaval precedent it deliberately duplicates — commit `2ec243de6`
- `lowy-3` ≡ `hickey-2` Census rationale restated near-verbatim in six artifacts — commit `98b774f13`
- `hickey-3` Dial-chain exit-safety invariant documented at the timer, not the dial chain — commit `6dcb011c6`

### Agreed — no change (2)
- `lowy-4` Public Reference bullet names the internal timer as the contract's ground (website/src/content/surface/ref-surface-remote.mdx:182)
- `lowy-5` Unrelated node-pty resolution-metadata churn rides the lockfile delta (pnpm-lock.yaml:4290)

srid:
## Codex ⇄ Claude debate

✅ Consensus in 1 round · reviewer effort: xhigh · base: `318f17437` (merge-base with master)

Codex approved the diff with **zero findings at any severity** — no debate commits were needed. Its round-1 verdict: the timer routing through the file-local unref seam is consistent, the ref'd admit-handshake timeout correctly preserves the caller-awaited `pin()` settle, the updated contracts accurately describe the process-lifetime semantics, and the real-child pins exercise both the abandonment exit and the must-fire delivery; it independently re-ran the 172-test suite + package typecheck (green).

srid:

Ran as the review gauntlet's final step (rules pass + fact-check pass; elegance skipped — `/simplify` ran separately this gauntlet).

- **Pass 1 (rule checklist):** zero violations across every built-in and project rule — the diff is itself a `no-unbounded-growth`-class fix, with the census pinned structurally (exactly two bare `setTimeout(` sites in `session.ts`, enforced by test).
- **Pass 2 (fact-check):** one finding, fixed in 2a3a7b16d — the Atlas note's "Shape of the change" still described the pre-lens design (three direct `.unref()` calls, "no guard helper") that never shipped; it now describes the `armInternalTimer` seam the lens gauntlet landed. Everything else checked sound: both fixtures exercise the real `makeSession` paths non-vacuously, the ref'd `withHandshakeTimeout` timer is genuinely the only caller-settling one, no new flake sources.

Earlier gauntlet steps for context: architecture-first-principles (3 confirmed findings — spawn-builtin timeout adopted, `ClientCursor.next()` contract-doc gap closed, and the **pre-existing** `pendingTimer` race recorded as #1875), lens debate (consensus, 4 fixes), codex xhigh (approved round 1, zero findings).

srid:
## Evidence

Behavioral fix with no on-screen surface — the proof is a process that used to hang forever now exiting. The demonstration runs the PR's own abandoned-session fixture (`processExit.fixture.immortal.ts`: create a session against a never-connecting endpoint, `pin()` it, drop it without `destroy()`, end the main script) against the **old** `session.ts` (master, `318f174`) and the **new** one, under a 5s `timeout`:

```console
=== BEFORE (master session.ts): abandoned session, 5s timeout ===
$ time timeout 5 node --import tsx src/processExit.fixture.immortal.ts
real    0m5.037s
exit=124        # ← killed by timeout: the ref'd backoff timer held the loop; the process was immortal

=== AFTER (branch session.ts) ===
$ time timeout 5 node --import tsx src/processExit.fixture.immortal.ts
MAIN-END
real    0m0.851s
exit=0          # ← exits cleanly the moment the main script ends
```

The same behavior is pinned permanently by `processExit.test.ts` (real child processes; the must-fire twin proves the deliberately-ref'd admit-handshake timeout still delivers its rejection to a pending `pin()` — child-measured elapsed ≥ the timeout — before the process exits). No mobile/desktop capture applies: the change has no UI surface.

website/src/content/changelog/unreleased.mdx
```diff
@@ -120,6 +120,7 @@ version: Unreleased
...
+- <Change kind="changed" title="An abandoned remote session no longer keeps its host process alive" pr={1873}>The same cure on the client side: a [`makeSession`](/surface/ref-surface-remote) session whose consumer dropped it without `destroy()` used to pin its host's event loop with a ref'd reconnect timer — a process redialing a gone endpoint forever, immortal, respawning ssh children on every attempt. The session's internal timers are now `unref()`'d, so a session never keeps an otherwise-finished process running by itself; a held session in a living process reconnects exactly as before, and the one timeout that settles a caller's pending `pin()` still always fires. One consequence to know: a wait derived purely from session state (a pump parked on `ClientCursor.next()` across a reconnect gap) is not a process hold either — a process with nothing else left exits instead of waiting forever.</Change>
```
srid, line 124: Should this be new changelog entry or be folded into existing one? Because kaval was not in previous release.

Meta: Should we modify our APM instructions to prevcen this kind of slipup in future?

srid: Keeping it as a standalone entry — here's the reasoning, split by your two questions:

**New entry vs. fold.** You're right that kaval wasn't in the previous release — but that's the test for the *end-user* changelog. This entry lives under `### [Surface](/surface)`, which is written for the **framework-consumer** audience (drishti, odu, kaval/padi-tui, and anything building on `@kolu/surface`). For that reader, `makeSession`'s new "internal timers are `unref()`'d, so a dropped session never pins its host process" is a genuine, documentable behavior *guarantee* they can rely on — not internal iteration on something they never saw. It's also deliberately the client-side member of a "same cure" trilogy that already documents framework-lifetime fixes the same way: 1858 (stdio server exits when its link ends), 1862 (daemon bin exits when its tenure ends), 1870 (unix-socket close disconnects peers). Folding only 1873 into a user feature entry — or dropping it — would break that set and bury a framework contract inside a product highlight. So: audience is last-*framework*-release consumers, and for them nothing here is "iteration."

**Meta / APM.** The rule that catches a *genuine* slip already exists — `.claude/rules/website-docs.md:12`: "*release-relative, not commit-relative … never narrate internal iteration.*" What's underspecified is not that rule but the **two-audiences distinction** it implicitly straddles: the end-user product changelog vs. the `/surface` framework-consumer changelog have *different* "last releases," and "was this shipped before?" resolves differently for each. If you think that's worth pinning explicitly (so a future author doesn't have to re-derive it), I'd draft it as a **separate self-improve PR** — per our convention APM/skill edits stay off the feature branch and get a human review — rather than smuggle it into this one. Say the word and I'll open it.

PR #1874, opened 2026-07-17
Preload content-derived shiki grammars — kill the atlas-sync flake's real mechanism

srid:
**The atlas-sync byte-shrink flake's real mechanism is an embedded-grammar load race, and this makes grammar readiness deterministic by construction.** Shiki's bundled `mdx` grammar declares its embedded languages *lazily* (`embeddedLangs: []`, `embeddedLangsLazy: ["yaml", …]`), and astro's build-wide shared highlighter loads grammars per code block on demand — so whether `release-workflow.html`'s one ```` ```mdx ```` fence rendered its YAML frontmatter with embedded-YAML tokens depended on whether `roadmap-graph.mdx`'s one ```` ```yaml ```` fence happened to be highlighted first under vite's concurrent transforms. That's the flaky-test tracker's `24757 → 24721` byte-shrink: per-box/per-run variance under a nix-pinned closure, whole-block divergence, immediate-green reruns — and why #1853's `tokenizeTimeLimit: 0` (a different, also-real degradation; **it stays**) didn't hold (recurred in kolu#1852 linux, kolu#1870 darwin).

### The fix — derive, preload, guard

```
src/**/*.{md,mdx} ─▶ fenceLangs() ─▶ shikiConfig.langs ─▶ createHighlighter()
   (content =           (one shared      (preloaded before      (registry immutable
    source of truth)     scanner)         first tokenize)        during tokenization)
```

- **`scripts/fence-langs.mjs`** — one scanner, shared by both astro projects, derives the full fence-language list from the content at config-eval time. The list *cannot* drift from the content, so "a fence language that isn't preloaded" is unrepresentable.
- **`kolu:shiki-eager-langs-only`** — a fail-fast transformer in both configs: a scan-missed fence in a real language throws at build time instead of silently lazy-loading mid-build. Every failure cell is loud or deterministic, never racy (matrix in the [design note](https://github.com/juspay/kolu/blob/shiki-grammar-preload/docs/atlas/src/content/atlas/bug-shiki-grammar-load-race.mdx)).
- **Both projects fixed** — same class: the website's silent degradation would ship to kolu.dev unnoticed. Its shiki setup moves to `website/src/shiki-config.mjs` (plain ESM) so the unit pins can import the exact object astro builds with; the Nix sandbox gets the scanner copied in beside the config (the `kolu-server-package.json` pattern).

### Red-first pins

| Pin | Red on pre-fix code | Green with fix |
| --- | --- | --- |
| Real atlas config, first-ever `mdx` tokenization (astro's own `createShikiHighlighter`, no prior yaml load) | plain frontmatter — the exact flaked `#005CC5` signature | embedded YAML (`#22863A`/`#032F62`), always |
| Scanner fixture: blockquoted fence opener | initial regex missed it | scanned |
| Guard: un-preloaded language | silently lazy-loads | throws with the remedy named |
| Wiring: `shikiConfig.langs` ≡ `fenceLangs(src/)` | n/a (no derivation existed) | asserted in both projects |

Mechanism regression (grammar presence flips the byte signature, order-independent) and plaintext-specials cases ride along — **16 atlas + 4 website tests**, running under `just atlas::check` and the website Nix derivation's new `checkPhase` (CI-gated via `ci::nix`, no new lane).

**Rebuilding with the fix leaves the committed `docs/atlas/dist/` byte-identical** — the preload deterministically reproduces the lucky-order rendering the committed dist happened to capture. The TZ/locale-scrambled idempotency rebuild is also byte-identical.

### Hardened through the review gauntlet

The four-stage gauntlet (architecture checks → lowy⇄hickey lens debate → codex debate (xhigh, consensus in 2 rounds) → simplify + code-police) folded in:

- **`shikiFencePreload(dir)`** — one factory fuses the derived list with its guard, so a consumer can't wire the guard against a stale list; **`noTokenizeBail`** hoisted next to it (one shared home for both shiki-determinism transformers).
- **Scanner hardened to astro's real parser boundary** — container prefixes (`- `, `1. `, `> - `, `[^1]: `), whitespace-led info strings, Unicode aliases (`文言`), 4-backtick fences — each red-first fixtured, plus a **cross-pin** rendering the fixture through astro's actual `createMarkdownProcessor` so the accepted shapes can't silently diverge from the installed parser.
- **Per-consumer drift pins** — the scanner's two hand-held upstream mirrors (`SPECIAL_LANGS`, astro's `defaultExcludeLanguages`) are asserted against each project's *own* installed shiki/astro (separate lockfiles, separate versions).
- **Deploy-path drift pin** — `website/test/pages-deploy-paths.test.mjs` parses `default.nix`'s cp-sources and asserts each has a `pages.yml` deploy trigger; it immediately caught **three pre-existing holes** (both logos + the server `package.json` were copied into the site build but never triggered a deploy — now listed).

*The changed-file list looks test-heavy because the mechanism demanded it: every hand-held fact in the fix carries a pin against the real installed dependency.*

> Soak certification (≥20 consecutive `atlas::check-sync` runs on the linux CI class where it flaked, plus a darwin window) runs after review; the tally lands here as cert evidence — the evidence #1853 never had.

*The flaky-test-tracker row update is coordinator-owned and deliberately not in this PR. No changelog entry: build-infra fix, no user-visible kolu behavior.*


srid:

Settled: 1 reconciled (raised by both lenses) · 1 auto-settled minor · 1 unopposed · 2 debated

### Applied (4)
- `lowy-1` ≡ `hickey-1` Upstream drift pins run against only one of two consumers' pinned versions — commit `f29e6bb50`
- `lowy-2` Same upstream type gap handled asymmetrically: documented cast vs unchecked annotation — commit `e16e3cb61`
- `lowy-3` ≡ `hickey-3` no-tokenize-bail transformer hand-held identically in both consumers — commit `c21c0c4c5`
- `lowy-4` ≡ `hickey-2` First-tokenization pin's validity rests on intra-file test order and astro's private cache key — commit `82c209825`

### Agreed — no change (1)
- `hickey-4` The load-race mechanism is retold in ~6 prose locations (scripts/fence-langs.mjs:1)

srid:
## Codex ⇄ Claude debate

✅ Consensus in 2 rounds · reviewer effort: xhigh · base: origin/master (merge-base 318f17437)

|---|---|---|
| 1 | a13c940c2 | fix(codex): round 1 — parser-boundary scanner, pages path, derived probes, test hygiene |

- **F1** — FENCE_OPENER is narrower than Astro's Markdown parser
- **F2** — website/default.nix now copies scripts/fence-langs.mjs into the website build, but the Pages workflow only triggers for website/**, Atlas dist, the favicon, and the workflow itself
- **F3** — The real-pipeline guard tests assume fixed language names remain absent from content: Atlas hard-codes ruby, while website/test/shiki-eager-langs.test.mjs:107 searches only ruby/erlang/lua and can pass undefined to astroHighlight
- **F4** — The scanner fixtures create eight fence-langs-* directories under the system temp directory per suite run and never remove them
- **F5** — The ANSI fixture contains a literal ESC control byte in source
- **F6** — The README says any fence added while astro dev runs requires a restart; the website README repeats this at website/README.md:36

srid:

Rules + fact-check passes over `318f17437...HEAD` (elegance skipped — `/simplify` ran separately this gauntlet). Findings, actioned in 1b117f932:

| Finding | Rule | Action |
| --- | --- | --- |
| `covers()` stripped `/**` without keeping the slash — a sibling path (`docs/atlas/dist-staging`) would false-match `docs/atlas/dist/**`, letting the deploy-path drift pin pass green on the exact silent-deploy-skip class it exists to catch | fact-check | **fixed** — slash-preserving prefix |
| `pages.yml` parsed with a regex over structured data | prefer-focused-library | **fixed** — real `yaml` parser (astro's transitive dep) |
| `astro-deps.mjs` exported `requireFromAstro` with zero importers | no-dead-code | **fixed** — internal const |
| createRequire dance + `astroHighlight` mirror duplicated across the two projects' tests | dry-rule-of-three | **skipped** — re-litigates a gauntlet-ratified decision: the projects pin *different* astro/shiki versions in separate lockfiles, so per-consumer copies resolving through each project's own astro are correct; a shared copy would couple version-independent things |

All other rules: clean (no new catch/swallow, no unbounded growth, no sentinel overloads; every new symbol documented).

srid:
## Evidence — soak certification (the evidence #1853 never had)

No on-screen surface to capture (build-infra fix); the evidence is the **soak cert**: ≥20 consecutive `just atlas::check-sync` runs per platform at this PR's final HEAD `1b117f932`, on the CI class where the flake bit. Each run is the full gate — `astro sync` + tsc, a real build, the 23 unit pins, a byte-compare of `dist/` against the committed tree, and a TZ/locale-scrambled idempotency rebuild — so every green run certifies **two** byte-identical builds.

| Platform | Box | Runs | Result | dist-tree hash (all runs) |
| x86_64-linux | kolu-ci-3 (CI pool lease; ci-4 excluded) | **22/22** | zero divergence | `a258c3d88ddb4ec6` |
| aarch64-darwin | rasam (single-tenant window) | **22/22** | zero divergence | `a060a26a8addfa78` |

**88 builds total, byte-identical within each platform, zero divergence.** The per-platform aggregate hashes differ only by untracked build residue (`dist/.prerender`); byte-identity of the *tracked* `dist/` against the committed tree is enforced by every run's `git diff` gate, so both platforms certified the same committed bytes.

For contrast, the pre-fix mechanism reproduced deterministically at unit level: a highlighter without `yaml` in the registry renders the `release-workflow` frontmatter plain (the exact flaked `#005CC5` signature, 24721 B class); with the preload, embedded-YAML tokens — always (`docs/atlas/build/shiki-grammar-preload.test.mjs`, red on the pre-fix config).

PR #1730, opened 2026-07-09
feat(padi): W6 — the honest connect

srid:
## What & why

Connecting to a remote host can compile+copy padi onto that machine over ssh — on a first connect to a fresh host that Nix build takes **minutes**. Today the whole canvas shows one dead word, **"Connecting…"**, indistinguishable from a hang. This ships **padi W6 — "the honest connect"**: the overlay now tells the truth (which phase, a live log tail, an elapsed timer), and the connection type is made perfect so a real failure shows the real reason.

Two halves, per the ratified plan (`docs/atlas/src/content/atlas/padi.mdx` §W6).

## Half 1 — the connection type, made perfect

Replaces surface-remote's `SessionState` machinery with the ONE ratified type:

```ts
type SessionState<Prov extends string = never> = { log: readonly {source:"local"|"remote"; line:string}[] } & (
  | { phase: "connecting" | "connected" | Prov }
  | { phase: "disconnected"; error: string; cause: "network" | "remote" }
  | { phase: "failed";       error: string; cause: "remote" }   // network is never terminal — by type
);
```

- `phase`/`error`/`cause` replace `connection`/`lastError`/`failureCause`; the two progress arrays collapse into one provenance-tagged `log` — `remoteProgressLines` and the in-band `[local] `/`[remote] ` prefixes are **gone** (provenance is a field).
- **`failed`+`network` is now a compile error** (a network fault retries forever, never gives up).
- The ssh connector declares its own `SshProv = "copying" | "building"`, split at the real command boundary (`copying` = `nix copy --derivation`, `building` = the remote `nix-store --realise`, the minutes). The local endpoint is `SessionState<never>` — it **cannot spell** a provisioning phase.
- Deletes the seven-name chain (`ProvisioningPhase`/`UpConnectionState`/`LocalConnectionState`/`FailureCause`/`SessionStateCommon`/`ConnectionState`/the state+cause zod consts).
- The browser cell `ConnectionInfo` becomes `z.discriminatedUnion("phase", …)` — the exact mirror of the session sum; `projectConnection` stops flattening.

## Half 2 — the kolu overlay

`ConnectCanvas` reads the active entry's `connection` cell and switches exhaustively on `phase`: `copying`/`building` narrate the real phase with a live log tail + a client-side elapsed timer; `connecting` is a brief distinct handshake; failure hands off to the **existing** Skew-UX host-down card (no second failure UI). The warm path stays calm — a warm reconnect short-circuits straight to `connected`, no build UI.

[...]

PR #1876, opened 2026-07-17
Remote-host kaval contract skew: name it, show both versions, offer the fix that works

srid:
**A newer kolu binding a host with an outdated kaval used to dead-end**: the canvas said "kaval didn't start", the offered **Restart kaval** provably couldn't succeed (it respawns the same old binary — the server had already tried), and pressing it surfaced a bare "Internal server error" because the daemon's typed skew rejection flattened to `INTERNAL_SERVER_ERROR` on the way to the browser. This PR fixes the three defects at their seams and then fortifies the surface framework so the *class* is inexpressible — per the merged bug note ([bug-remote-kaval-contract-skew](https://kolu.dev/atlas/bug-remote-kaval-contract-skew.html)) and the gate-ratified design note that ships in this PR (`docs/atlas/src/content/atlas/kaval-skew-fix-design.mdx`).

```
MINT                    STATE                    WIRE                     UI
connect.ts throws  ──▶  supervisor endpoint ──▶  daemonStatus arm    ──▶  ONE kavalAttention surface
DaemonContractSkewError  "incompatible"          {state, daemonVersion,   card · dialog · chip pip
{daemonVersion,          (never dead/degraded)    requiredVersion}         └▶ "Update & restart kaval"
 requiredVersion}                                                             = binder renew():
                                                                              drain → re-realise → fresh kaval
```

### Commit 1 — the repairs (SK1–SK3)

| Defect | Fix |
| --- | --- |
| The remote log bridge crashed per line (unbound pino method in both padi bindings), hiding every diagnostic | `makeSession`'s sink becomes `log?: Logger` (the `@kolu/log` shape); the session dispatches severity internally via a receiver-bound call — the consumer-side hazard has **no spellable form left**, and the compensating per-line `try/catch` is deleted (a throwing logger now fails fast) |
| `DaemonContractSkewError` baked its versions into prose | `readonly daemonVersion`/`requiredVersion` fields; the message is **derived** from them — nothing downstream ever re-parses it |
| `recycleKaval` rethrew the skew plain → `INTERNAL_SERVER_ERROR` toast | The knowing endpoint refuses **typed**: `KAVAL_CONTRACT_SKEW` with both versions as data (the `fileGoneAsNotFound` precedent); one attempt, no retry loop |

### Commit 2 — the fortification (SK4–SK6)

- **Contract skew is a first-class daemon state.** `incompatible` joins the endpoint state sum carrying both versions — emitted where a fresh spawn *still* skews and at the refuse arm (both previously collapsed to `dead`/`degraded`), never coerced by a restart hold. `PADI_SURFACE_VERSION` 4.0 → 4.1: the new arm is an emitted wire variant, and both skew directions converge before an old parser can meet it (the doctrine paragraph rides the constant).
- **One attention surface, one recovery.** `kavalAttention` is the single version-comparison in the client — the currency axis ("newer build available") and the contract axis ("incompatible — needs update") light the *same* card/dialog/chip chrome, distinctly labeled. UI affordances are a **total function of the state sum**: the skew card states both versions from typed fields and offers exactly one action — **Update & restart kaval** (confirm-gated `hosts.renewDaemon` → the binder's drain → re-realise → converge pipeline), on *both* local and remote hosts. The dead-end Restart is unspellable against a skewed daemon — card, dialog, and command palette alike.
- **Procedures declare their error union.** `ProcedureSpec.errors` (oRPC contract-level typed errors) threads through the contract, all four type oracles, handler `opts.errors` constructors, and the bound client face — so a declared error crosses **every hop** `defined: true` with data intact. _The incident-hop pin caught a real hole en route: `surface-map`'s folded contract dropped declared errors, demoting them at the outer wire — fixed and pinned across two real wires, plus the `mirrorRemoteSurface` forwarder leg._ Undeclared throws still cross as `INTERNAL_SERVER_ERROR`: the fail-fast channel is deliberately untouched. Adopted end-to-end by `lifecycle.recycleKaval`; migrating the remaining ad-hoc codes (`MAP_KEY_UNKNOWN` et al.) is now a mechanical declaration each.

### Design philosophy

**Fail fast** — the compensating try/catch is deleted, not kept; a skew refuses typed instead of retrying; no override knob re-enables Restart on a skewed daemon. **Electricity boundaries** — every framework move lands in the package owning the volatility (`surface-remote` log seam, `surface-daemon-supervisor` state sum, `surface` error channel); kolu only consumes. **Reuse the source of truth** — `@kolu/log`'s existing `Logger`, the existing currency UX extended rather than a parallel skew card (one surface fed by typed arms cannot disagree with itself), the binder's existing drain→re-realise pipeline as the recovery.

### Consumer impact

- **drishti**: passes no `onLog`, imports no supervisor symbol, spec literals unaffected by the additive `errors` key (grounded greps at its pin). Pair PR with CI green at this PR's final HEAD to follow before merge-readiness.
- **odu-impact verdict**: `none` for SK1/SK2/SK4 (no `onLog` use — the default stderr sink odu's display depends on is unchanged; no supervisor import); **`adoption-opportunity`** for SK6 (`node.rerun` / `run.cancel` / `run.configure` could declare their error unions) — ledger entry on odu#43.

> **History note (per srid's ruling):** the merge shape is **one PR, two commits** — `fix(surface): … (SK1-SK3)` and `feat(surface): … (SK4-SK6)`. The branch will be reshaped to exactly that pair before review begins; review-gauntlet fix commits land on top afterwards and are noted rather than force-pushed away.


srid:

Independent findings: lowy=8, hickey=11
Settled: 2 reconciled (raised by both lenses) · 6 auto-settled minor · 5 unopposed · 2 debated

### Applied (9)
- `lowy-1` ≡ `hickey-9` Down/terminal endpoint-state classification lives in consumers, not the states' home — commit `f47c7ed17`
- `lowy-2` ≡ `hickey-3` isContractSkewError narrows to a type whose `subject` field it never attests — commit `24cdad2f6`
- `lowy-3` The skew version-pair shape is re-spelled at every layer with no shared source — commit `acc0de4be`
- `lowy-4` The spec-errors extraction type is spelled three times inside @kolu/surface — commit `5aa053090`
- `lowy-5` ≡ `hickey-5` kavalStale's `live` parameter is dead at its only call site — commit `b9c409dc3`
- `lowy-6` Typed-error narrowing forces consumers to import the vendor's verbs past the surface receptacle — commit `b440976a8`
- `hickey-2` offerRestartVerb hardcodes the skew exception instead of reading the presentation table — commit `4dd6bd95c`
- `hickey-4` KavalPresence.incompatible carries a version payload no consumer reads — commit `c62a85364`
- `hickey-6` Four-method Logger stubs hand-rolled at ~10 test sites — commit `8f2c63798`

### Agreed — no change (6)
- `lowy-7` ≡ `hickey-1` UpdateKavalButton couples to the ambient activeHost by mount convention, not by contract (packages/client/src/kaval/UpdateKavalButton.tsx:41)
- `lowy-8` Two vocabularies for the same 'version skew pair' concept across sibling skew surfaces (packages/padi/src/vocab.ts:630)
- `hickey-7` Wired-leaf error harness duplicated across two packages' tests (packages/surface-map/src/procedureErrorsAcrossMap.test.ts:40)
- `hickey-8` DaemonStatusSchema never-pin cross-product maintained by hand (packages/padi/src/vocab.ts:618)
- `hickey-10` useDaemonRestart.ts now owns two recovery verbs under a restart name (packages/client/src/kaval/useDaemonRestart.ts:60)
- `hickey-11` DegradedCanvas re-splits the down sum into two parallel accessors (packages/client/src/kaval/DegradedCanvas.tsx:121)

srid:
## Codex ⇄ Claude debate

✅ Consensus in 3 rounds (2 fix commits) · reviewer effort: xhigh · base: merge-base with origin/master

|---|---|---|
| 1 | d5389698a | fix(codex): round 1 — commit the primary rendezvous before the post-spawn handshake (F1); honest skew journal line (F2); refuse-arm comments say incompatible (F3); 4.1 test titles (F4) |
| 2 | 4cff06fa2 | fix(codex): round 2 — onSpawned contract doc matches the pre-handshake firing (F5) |

- **F1** — When a daemon adopted from `adoptHint` is recycled, the replacement is spawned at the primary rendezvous
- **F2** — The restart catch unconditionally logs that the endpoint reported `dead/degraded`, but the newly handled contract-skew path deliberately reports `incompatible`
- **F3** — The adoption comments still say the refuse policy leaves a skewed survivor `degraded` (also at lines 806-807), while the implementation now emits the distinct `incompatible` state
- **F4** — The suite is still titled `padiSurface 1.0 contract`, and the version test at line 27 says `is version 4.0`, although this change asserts and documents version 4.1
- **F5** — The round's F1 fix deliberately moved `onSpawned` to run as soon as the primary socket is accepting, before `connect()`, so it now also fires when the handshake ends in `incompatible` or `dead`

srid:

Rule checklist + fact-check over the diff (`--no-elegance`; simplify already ran). **Fact-check pass: no issues** — the skew's typed fields are traced raise → typed rethrow → wire schema → client `safe()`/`isDefinedError` narrowing with nothing swallowed or downgraded; version operands consistent across all three raise sites; `liveDownState` compile-exhaustive; the two attention axes mutually exclusive by construction; `renewingHosts` bounded.

**Rule pass — 3 flagged, actioned:**

| Rule | Site | Disposition |
| --- | --- | --- |
| `toast-must-include-error-message` | `useDaemonRestart.ts` skew branch | **Fixed** (`77b3d77ce`) — surfaces the server's `err.message` (versions ride it, typed) instead of re-assembling prose from `error.data`; keeps the "Use Update & restart kaval" guidance. |
| `prefer-ts-pattern` | `daemonPresentation.ts` `liveDownState` | **Fixed** (`77b3d77ce`) — exhaustive `switch` → `match().exhaustive()`, the house form used beside it in `KavalInfoDialog`. |
| `prefer-ts-pattern` | `toKavalPresence`, `kavalAttention` | **Kept** — `toKavalPresence` is a pre-existing `if`-cascade extended by one branch (whole-function conversion is scope creep); `kavalAttention`'s terminal arm is a runtime predicate (`kavalStale`), not a structural pattern `match` fits. |
| `callback-fanout-guarded-at-funnel` | `session.ts` `emit` | **Kept (adjudicated)** — the deleted per-line `try/catch` is the gate-ratified **SK1 fail-fast**. The `@kolu/log` `Logger` contract is non-throwing and SK1 removed the only real throw source (the unbound-`this` pino method), so a throw is reachable only via a consumer contract-violation — which the codebase's crash-loudly philosophy says surface, never swallow-and-degrade (exactly the anti-pattern the old catch was, spamming stderr per line while dropping every diagnostic). Reversing a gate ruling on a rule flag would be the wrong call; surfaced here for the human. |

srid:
## Evidence

The fix driven end-to-end in a **real browser** against a **genuine older-install kaval**, on an isolated nix-built kolu (my branch, `77b3d77`) on an ephemeral pu box. The ancient survivor is the real `packages/kaval` build from **`c9fd9a2bf`** — the last commit before `PTY_HOST_CONTRACT_VERSION` bumped 5.0→5.2 (`557e08a86`, #1783) — wired into padi's closure (`KOLU_KAVAL_BIN`), so the current padi (with SK4) spawns a genuine 5.0 kaval. This is the incident topology: an old install's kaval that a current server can't talk to.

### 1 · The skew is a first-class state, both versions typed end-to-end

incompatible card

padi's boot hits the real SK4 path — `connectKaval` (connect.ts:100) → `spawnConnectHold` (endpoint.ts:496, the *fresh-spawn-still-skews* producer) throws the typed `DaemonContractSkewError{ subject:"pty-host", daemonVersion:"5.0", requiredVersion:"5.2" }` — and it surfaces as `incompatible`, never `dead`/`INTERNAL_SERVER_ERROR`:

- **chrome-bar kaval chip**: *"kaval incompatible — needs update · speaks 5.0, kolu needs 5.2; update needed"*
- **canvas card**: *"kaval is incompatible with this kolu · this host's kaval speaks **5.0** · your kolu needs **5.2**"* — both versions from the typed wire arm, rendered structurally (never re-parsed from prose)
- the dead-end **Restart** verb is gone; the one action offered is **"Update &amp; restart kaval"**
- padi reports **contract v4.1** — the honest minor bump is live on the wire

### 2 · The recovery provably converges (not inferred)

converged

Clicking **Update &amp; restart kaval** → confirm → `hosts.renewDaemon` → the binder drains padi and the reconnect loop re-realises the current closure (here: padi's `KOLU_KAVAL_BIN` re-pointed at the real 5.2 kaval — the honest analog of the host being re-provisioned to the current build). Result:

- toast: *"Host daemon drained — re-provisioning the current build; kaval reconnects shortly"* (the honest **drain-time** copy — the RPC resolves at drain, and the daemon-status chrome narrates the rest)
- the kaval chip flips to **"running · contract v5.2 · running 0s"** — **converged to the correct contract**
- the incompatible card is gone; the canvas returns

### 3 · The graveyard-survivor case self-heals — the card's boundary

A separate run (current padi, its **baked** current kaval) confirms the delimiting design fact: when padi's own closure is current, a skewing kaval is recycled to the baked 5.2 by construction (`onSkew → recycle → spawnConnectHold`) — padi logged *"spawned a fresh daemon and connected"*, no skew reached the UI. So `incompatible` appears **only** when padi's *own closure* is stale (the real incident), which is exactly the case `renewDaemon` (re-realise the closure) is the recovery for. A current padi merely *adopting* an ancient survivor never surfaces the card — it self-heals.

*Reproduce-first note: the original D2 recipe (plant the old kaval at the adopt-hint socket) does **not** surface `incompatible` — the adopt path recycles to padi's baked (current) kaval and self-heals; and a `KOLU_KAVAL_BIN` env override can't take (the wrapper force-`export`s it). A genuine `incompatible` requires padi's **baked** kaval to be old, which is what the closure re-wire above produces. Folded into the design note's evidence section.*

srid:
Paired drishti PR (surface.md gate): srid/drishti#115 — kolu pin bump to `619d0659`, no drishti code change needed (grounded grep: no onLog, no errors specs, BoundProcedure subtype-compatible); the pair-gate is drishti CI green against the new surface API.

website/src/content/changelog/unreleased.mdx
```diff
@@ -8,6 +8,7 @@ version: Unreleased
 
 ### [Kaval](/kaval)
 
+- <Change kind="fixed" title="An outdated host's kaval now says so — with both versions and a working fix" pr={1876}>Connecting to a machine whose kaval speaks an older wire contract used to dead-end: the canvas said "kaval didn't start", the offered **Restart kaval** could never succeed (it respawns the same outdated binary), and pressing it surfaced a bare "Internal server error". kolu now recognizes the mismatch as its own daemon state — the canvas card and the kaval dialog say **which version the host's kaval speaks and which your kolu needs**, the dead-end Restart is gone from every surface it can't help (the card, the dialog, the command palette), and the one action offered — **Update &amp; restart kaval**, behind a confirm — re-provisions the host's build and starts a correct-version kaval. The same chrome that flags "newer kaval build available" carries the incompatible verdict, so one badge tells both stories. Behind the toast, the daemon's real error now travels typed end to end instead of flattening into a generic failure.</Change>
```
srid: Should this be new changelog entry or be folded into existing one? Because kaval was not in previous release.

Meta: Should we modify our APM instructions to prevcen this kind of slipup in future?

srid: Folded — good catch. kaval ships for the **first time** this release, so a reader upgrading from the last release never had a kaval to hit the skew dead-end; a standalone `kind="fixed"` entry has no prior state to fix against. I removed the `fixed` entry and folded the incompatible-state handling into kaval's **added** `<Change>` (extended its prose, reconciled the now-stale "Restart kaval" sentence, added #1876 to its refs). The Surface-section `added` entry keeps the framework-level typed-errors story, which is genuinely new-this-release. → `47c09075a`

**Meta:** yes — done in the same push. The `website-docs` instruction already said "iterating on an unreleased entry EDITS it," but didn't spell out that `kind` itself is release-relative, which is where this slipped. I added a corollary to `.apm/instructions/website-docs.instructions.md`: *you cannot fix (or change) what the last release never shipped, so a `<Change>` touching a still-unreleased feature is `added` and folds into that feature's existing added entry* — citing #1876 as the real slip, in the same "PR #1730 → #1746" style the rule already uses. Regenerated `.claude/rules` + `AGENTS.md`. → `a43b19758`

Issue #1872, opened 2026-07-17
Dock agent detection broken for terminals whose PTY root command IS the agent (kaval-tui create -- claude)


srid:
## Symptom (field, 2026-07-16 overnight)

Three coordinator-spawned agent terminals show **only terminal activity in the Dock — no agent activity**. Tiles render the terminal title + a spinner glyph (`⠐ Claude Code`, or the OSC title claude sets), but the Dock's agent presence/urgency treatment never engages. Terminals created the normal way (shell + rc-hooks, agent launched from the shell) on the same box at the same time show full agent activity (`✳` + agent state).

## Repro

```bash
# broken: the PTY's root process IS the agent — no shell
kaval-tui create -- claude --dangerously-skip-permissions
# → Dock tile: terminal activity only; agent never detected

# working: default shell, then launch the agent from it
kaval-tui create            # rc-hooked $SHELL
kaval-tui send <id> "claude --dangerously-skip-permissions" && kaval-tui send <id> --key Enter
# → Dock tile: agent detected (✳, agent title/urgency)
```

Observed on 3/3 direct-spawn terminals (`e17c2d63`, `0bbc3b9b`, `2529e762`) vs 2/2 shell-wrapped ones (`dd431e99`, `f6bfc052`), same box, same kolu build.

## Process-tree evidence

```
# working (dd431e99): PTY root is the rc-hooked shell; claude is a child
2977639 2977580 /run/current-system/sw/bin/bash --rcfile /run/user/1000/kaval-7cf24…/rc/ba…
2979610         claude --resume … --dangerously-skip-permissions   (child of 2977639)

# broken (e17c2d63 / 0bbc3b9b / 2529e762): PTY root IS claude, parent is the kaval daemon
3266884 2977580 claude --dangerously-skip-permissions
3274320 2977580 claude --dangerously-skip-permissions
3274416 2977580 claude --dangerously-skip-permissions
```

## Suspected mechanism (grounded in source)

Agent detection keys on signals only an **rc-hooked shell** emits:

[...]

PR #1877, opened 2026-07-17
Scrub leaked CLAUDE_CODE_CHILD_SESSION at every kolu-terminal agent spawn

srid:
**Every Claude Code agent spawned into a kolu terminal by another Claude session has been silently losing its entire session transcript.** `kaval-tui create` / `padi-tui create` hand the caller's full environment to the new PTY (`kaval-tui/src/main.ts` passes `env: process.env`), and a Claude driver's Bash tool exports `CLAUDE_CODE_CHILD_SESSION=1` — which makes the spawned interactive `claude` treat itself as a nested child session and **never write its root transcript** (`~/.claude/projects/…/<session>.jsonl`) nor register in the session registry. Nothing errors; the run just becomes un-resumable after a crash and permanently un-minable (`/self-improve` and `/blog-post` both mine those transcripts).

This PR bakes the guard into the `kolu` skill's spawn commands — every documented `create` now launches the agent behind `env -u CLAUDE_CODE_CHILD_SESSION -u CLAUDE_CODE_SESSION_ID`, with a callout explaining why. _The prefix is a no-op for `codex`/`opencode` and for non-Claude drivers, so it's unconditional rather than a remembered special case._

### Evidence ledger

| Claim | Evidence |
| --- | --- |
| All 3 lanes of the 2026-07-16 campaign lost their transcripts | `session-unref`, `kaval-skew`, `shiki-grammar-preload` project dirs under `~/.claude/projects/` contain subagent/tool-result artifacts but **no root `.jsonl`**; the 4 older, hand-launched lanes on the same box all persist normally |
| The leaked env is the discriminator | All 3 affected `claude` processes carry `CLAUDE_CODE_CHILD_SESSION=1` + `CLAUDE_CODE_SESSION_ID=<coordinator's id>` (read from `/proc/<pid>/environ`); the healthy lanes carry neither |
| `CLAUDE_CODE_CHILD_SESSION=1` alone suppresses persistence | Controlled interactive runs (claude 2.1.212, pty via `script`): leaked env → prompt processed, **no transcript**; same run with `env -u` scrub → transcript persists; bisect keeping `CLAUDECODE`/`ENTRYPOINT`/`SESSION_ID` but dropping only `CHILD_SESSION` → **persists** |
| The leak vector is `create`'s env handoff | `packages/kaval-tui/src/main.ts:515` sends `env: process.env` in the create request; the affected lanes' environ even includes the coordinator's `DIRENV_DIR` |
| This broke `/self-improve` itself | Step 1 (`find ~/.claude/projects -name "$SID.jsonl"`) returned zero matches for the UNREF1 run — the transcript this PR's own analysis needed does not exist |

### What rode along

`just ai::apm` regen also caught up the **orchestrator dashboard** generated copies (`.claude/` + `.agents/`), whose committed sources had drifted ahead of the generated tree on master — pre-existing lag, now reconciled so generated == sources again, plus the `apm.lock.yaml` file-manifest entries for them.

> **The durable fix belongs in the product**: `kaval-tui`/`padi-tui` (or the pty-host) could scrub `CLAUDE_CODE_CHILD_SESSION` from the create request unconditionally, making the skill guard unnecessary. This PR is the skill-level default so the next campaign doesn't lose history while that's weighed.

_Provenance: `/self-improve` on session `e69381d8-a561-4190-bc16-98a8bd317e76` (UNREF1 `/be` run). Non-durable observations from that run, recorded here rather than engineered out: machine-wide `nix develop` CA-derivation skew (this PR's gates ran via store-realized pnpm 10.32.1 / biome 2.4.7), and the rasam darwin wedge — both machine-state, not skill defects._

_Generated by [`/self-improve`](https://github.com/srid/agency) on Claude Code (model `claude-opus-4-8[1m]`)._

### DIFF
--- a/.agents/skills/kolu/SKILL.md
+++ b/.agents/skills/kolu/SKILL.md
@@ -32,7 +32,8 @@ terminals a **kolu owns** (the last section); for raw terminals, reach for
 ## The loop
 
 ```sh
-id=$(kaval-tui create --json -- claude | jq -r .id)            # spawn the inner agent
+id=$(kaval-tui create --json -- env -u CLAUDE_CODE_CHILD_SESSION -u CLAUDE_CODE_SESSION_ID \
+    claude | jq -r .id)                                        # spawn the inner agent (env -u: below)
 kaval-tui send  "$id" "refactor the parser to use a lexer"     # 1. the text (no Enter)
 kaval-tui wait  "$id" --until idle:300                         # 2. observe the TUI settle
 kaval-tui send  "$id" --key Enter                              # 3. submit (its own command)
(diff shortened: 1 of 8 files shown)


srid:
Closing — superseded by the structural fix. This PR added skill-level `env -u CLAUDE_CODE_CHILD_SESSION…` prefixes to the documented spawn commands, written the night of the transcript loss. Since then #1880 fixed the class in the product (kaval-tui composes the child env from a pinned allowlist at every spawn seam — the caller's session vars never reach the child, verified by the exact-set data pins), #1884 closed the last seam at the type level, and the agent-spawn-first-class note retired the interim skill rules as their PRs shipped (keeping only the own-shell scrub for hand-tainted ssh/sudo shells, which is upstream of kolu and untouched by spawn-command prefixes). The evidence ledger here was the right diagnosis at the time — the durable record lives in #1872 + the note. (Coordinator.)

PR #1878, opened 2026-07-17
docs(atlas): flaky tracker — shiki atlas-sync fixed (#1874, soak-certified)

srid:
Coordinator atlas branch. The human merges.

## Commits
1. **shiki atlas-sync row → fixed-pending-merge** — #1874 ships the content-derived langs preload + `kolu:shiki-eager-langs-only` fail-fast guard in both astro configs (mechanism: the `embeddedLangsLazy` load-order race, opus-verified + empirically reproduced). Cert: 22/22 soak per platform (88 builds, zero divergence) — the evidence class #1853 lacked.
2. **NEW row: fs-watch e2e iframe-refresh flake is a two-box darwin pattern** — reproduced on **rasam** in #1876 (previously attributed to sincereintent's stale fseventsd only): file-watch waitFor never fires, cucumber retries both wedge identically, change-independent by SHA pair (docs-only delta from a fully-green HEAD, linux green same SHA). Suggests a darwin-fleet hygiene pass or hardening the e2e file-watch waitFor.

The human merges.


srid:
New note on this branch: **agent-spawn-first-class** — the #1872 fix proposal per a-f-p + perfection (16-agent lens run, 13 findings, 0 refuted), honoring srid's steer: faces own agent-spawn as a first-class verb (the mcp face is already immune by schema — the absence of a `command` param is load-bearing); kaval-tui composes env instead of forwarding it (allowlist shape, not a CLAUDE_* deny-list); kaval stays a dumb PTY daemon and only stops discarding the argv it was handed. Includes the shellIdle second-lock finding (seeding lastCommand alone would NOT revive detection) and the upstream Claude Code fragility note.

PR #1879, opened 2026-07-17
/be interview routes through a coordinator's brief, not AskUserQuestion

srid:
**When /be runs as an implementing agent under an orchestrator, its §0 interview now goes through the coordinator's channel instead of `AskUserQuestion`** — a question dialog opened in the agent's own PTY hangs unseen until a human happens to glance at that terminal, stalling the run at its very first step.

The orchestrator side already name-bans `AskUserQuestion` in briefs (#1821, #1850, after #1773), yet the stall kept recurring — because the reflex lives in `/be` itself: §0 *is* the interview step, so the skill reaches for the dialog even when the brief only routes questions without naming the tool. The edit makes "a coordinator is driving me" itself the ban: write the question batch to a file, send a one-line pointer over the brief's channel, and block on the reply.

### Evidence ledger

| Edit | Evidence |
| --- | --- |
| `agents/.apm/skills/be/SKILL.md` §0 — brief's route replaces `AskUserQuestion` for the whole interview | Recurs across ≥3 prior rounds: #1773 introduced the coordinator-answers pattern, #1821 and #1850 hardened the orchestrator-side ban, and the stall still resurfaced in a coordinator-driven /be run — the fix was sitting uncommitted in the hard-killed self-improve worktree for session `73e4da1a` |
| `.claude/` + `.agents/` copies of `be/SKILL.md`, `apm.lock.yaml` | `just ai::apm` regen — generated copies ride the same commit as the source |
| Orchestrator dashboard copies (`index.html` / `index.js`) | Same regen trued up copies that drifted from their `agents/.apm` source in #1868 |

> **Provenance**: self-improve pass for session `f3183bfb-f468-4e8d-b2eb-43fb66a033ce` (the /be run that shipped #1874). That session's own transcript was never persisted (see report to the user: interactive child sessions spawned in kolu PTYs currently write no JSONL), so the primary lesson here is rescued from the prior run's orphaned worktree rather than mined fresh.

Gates run with repo-pinned tools sourced via `nix-build` (`nix develop` is broken on this host — CA-derivation skew): `biome format` clean, `nixpkgs-fmt` clean, `pnpm typecheck` green, `biome lint --error-on-warnings` green, `pnpm-lock.yaml` untouched.

*Generated by [`/self-improve`](https://github.com/srid/agency) on Claude Code (model `claude-opus-4-8[1m]`).*

### DIFF
--- a/.agents/skills/be/SKILL.md
+++ b/.agents/skills/be/SKILL.md
@@ -14,7 +14,7 @@ Take a task to a shipped, reviewed PR. Unlike `/do` (autonomous start to finish)
 
 ## 0. Interview (the differentiator)
 
-Before any work, ask the user via **`AskUserQuestion`** (one call, batched):
+Before any work, ask the user via **`AskUserQuestion`** (one call, batched) — **unless a delegating brief routes questions elsewhere.** When you run under a coordinator whose brief says questions go to its terminal (e.g. via `/kolu`, blocking on the reply), that channel *replaces* `AskUserQuestion` for the **whole** interview: write the batch to a file, send a one-line pointer, and block — never open a question dialog in your own PTY, where it hangs unseen until a human happens to look. The brief's route wins even when it only *routes* without naming the tool: reaching for `AskUserQuestion` here is `/be`'s built-in reflex (this *is* the interview step), so treat "a coordinator is driving me" as itself the ban.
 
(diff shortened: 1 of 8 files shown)


srid:
Closed as duplicate — subsumed by #1882, which carries the identical /be interview-routing sentence PLUS the whole-run extension ('the ban does not lapse when §0 ends'), without this PR's unrelated dashboard edits (already on master via other PRs). (Coordinator, on srid's instruction.)

PR #1880, opened 2026-07-17
Compose kaval-tui's child env from a clean base — stop the #1872 transcript loss

srid:
`kaval-tui create` copied the caller's **entire** `process.env` into the spawned terminal. Run from an orchestrating Claude session, that session's private identity vars rode in — and **a claude that inherits `CLAUDE_CODE_CHILD_SESSION=1` classifies itself as a nested child session and never writes its conversation to disk.** Three agents' histories were lost overnight (A/B-proven in the plan of record). An adversarial pass then found the **same ambient-identity class at four more spawn seams**. This PR lands **one structural invariant** — *identity cannot ride ambient env into any kolu-spawned process* — via **one shared allowlist** and compose-don't-forward at every seam that reaches an agent. This is **phase 1** of the [agent-spawn-first-class](https://github.com/juspay/kolu/pull/1878) plan (#1872).

### The one shared allowlist

`kolu-pty` (the env-policy home, beside `cleanEnv`) exports **`SPAWN_ENV_ALLOWLIST`** + `composeSpawnEnv` — the single source of truth every composer funnels through, **pinned exactly as data** so a future broad key that would reopen the class fails a *test*, not review vigilance. It is an allowlist, not a blacklist: an unknown future identity var is dropped by construction.

### The seams this PR closes

| Seam | Change |
| --- | --- |
| **kaval-tui `create`** | both composers import the shared constant (no local copy to drift); `--env K=V` is the explicit escape hatch. The #1872 pin is parameterized over **shell *and* command modes** — a command-mode regression (the exact field-repro shape) would have passed a shell-mode-only pin. |
| **`cleanEnv` (kolu-pty)** | production now **composes from the allowlist** instead of a wholesale `{...process.env}`, so **every hosted PTY** gets a clean base rather than the daemon's ambient env. The wrapper rcfile replay still restores the user's dotfile env. |
| **supervisor detached spawn** | layered the supervisor's full parent env under `cfg.env` (macOS always, nix-run, dev) — an ambient leak into kaval and every PTY. Now `env = cfg.env` unless `fromSource` (the one dev caller), parity with the systemd branch's `--setenv` (which wins over PAM on overlap). Pinned both ways. |
| **padi `daemonEnv`** | expanded to compose a **complete** allowlist base so `cfg.env` alone gives kaval true parity with systemd (no HOME/PATH regression on macOS-detached). Its exact key set is pinned as data. |
| **e2e harness** | the server child is composed from a **named base** (allowlist + devshell/runtime keys) instead of `...process.env`, so a harness launched from an orchestrator can't leak identity vars into the server → padi. |

### Seam 4 (surface-remote localhost arm) is deferred to PR 1.5 — and why that's honest, not an overclaim

The remote connector's **localhost arm** spawns its child with no `env` option (full inherit) — the one place the ssh boundary doesn't naturally scrub. It is **deferred to an immediate fast-follow (PR 1.5)**, and the residual is stated exactly:

> With seams 1 + 2 + 2a landed, **every path from a tainted daemon to an *agent* is composed.** The localhost-spawned process is itself a **daemon (kaval), not an agent** — and *all* of its downstream spawns are now composed: the PTYs it hosts via `cleanEnv`, the kaval it is via `daemonEnv`. So the class *"an agent inherits ambient identity"* is **closed on every live path**. Seam 4 is genuine **defense-in-depth against a future forgotten spawn path**, not a present leak — which is exactly why it decouples cleanly from the data-loss merge.

PR 1.5 carries the full tail: the required-env type change to `@kolu/surface-remote`, its **drishti pair PR** (drishti consumes `sshConnector`), the odu verdict, and `ref-surface-remote.mdx`.

### Consumer impact (this PR)

- **drishti:** consumes **none** of this PR's changed exports (`kolu-pty`, `survivableSpawnDriver`, `cleanEnv`) — grep-confirmed — so **no drishti pair PR** is required here. (Seam 4's `sshConnector` change is the only drishti-consuming one; it rides PR 1.5.)
- **odu-impact: `none`** — odu's tree has no reference to `kolu-pty`, `survivableSpawnDriver`, `DaemonSpawnConfig`, or `scrubDaemonNodeOptions` (grep-confirmed); it consumes `@kolu/surface-daemon-supervisor` only via build config, not the changed spawn API.

### kolu-mcp docstring + interim skill doctrine

Also in this PR: the corrected `lifecycle.create` docstring (schema is `{ cwd?, parentId? }`, never `(command, cwd, parent)` — that absence is the #1872 protection), and the interim `/kolu` + `/orchestrator` skill rules with ships-with markers.

### Red-first pin + evidence

`create.test.ts` and `shell.test.ts` assert no `CLAUDE_CODE_CHILD_SESSION` survives composition (red on the old forward, green now); the allowlist and `daemonEnv` key sets are pinned as data. The #1872 A/B transcript reproduction (kaval-tui-spawned **and** a claude in a canvas-created kolu terminal) rides as **PR evidence, not a CI node** — spawning a real claude is neither cheap nor deterministic.

Part of #1872. Plan of record: [agent-spawn-first-class.mdx](https://github.com/juspay/kolu/pull/1878).


srid:

✅ **Consensus** with zero debate turns · lowy + hickey · base `bab9c1bde71b`

Settled: 2 reconciled (raised by both lenses) · 1 auto-settled minor · 5 unopposed

### Applied (6)
- `lowy-1` ≡ `hickey-1` kaval's daemonEnv re-carries XDG_RUNTIME_DIR the receptacle already owns — breaks twin symmetry with padi's — commit `fb96625e5`
- `lowy-2` Framework's DaemonSpawnConfig.env contract silently changed to 'complete env' on the detached branch — doc and in-repo example still teach the old overlay semantics — commit `3e45f4915`
- `lowy-3` ≡ `hickey-2` 'Mine an allowlist from a source' is the real atomic primitive, hand-inlined in three places — commit `e546dc035`
- `lowy-4` composeSpawnEnv docstring claims 'non-empty' but the code carries empty-string vars — commit `5182f217e`
- `hickey-3` Two overlapping "safe PTY env" allowlists; dev branch silently gets a narrower env than production — commit `357896216`
- `hickey-4` DISPLAY filed under FUNCTIONAL ("run at all") when it is a GUI-session capability like WAYLAND_DISPLAY — commit `7217cd13a`

### Agreed — no change (2)
- `lowy-5` Two of the three exported env classes have no granular consumer (§6.5 observation) (packages/integrations/pty/src/shell.ts:47)
- `hickey-5` Identical operational-forward triple duplicated across the two daemonEnv twins (packages/server/src/padi/padiBinding.ts:162)

srid:
## Codex ⇄ Claude debate

✅ Consensus in 6 rounds · reviewer effort: xhigh · base: `origin/master`

|---|---|---|
| 1 | 7b00e6b5b | fix(codex): round 1 — separate inheritParentEnv from fromSource + 11 review findings |
| 2 | 69b5a7e82 | fix(codex): round 2 — null-proto env records, exact padi pin, inheritParentEnv example |
| 3 | 63b9426cd | fix(codex): round 3 — remote composer null-proto, full-allowlist daemonEnv pins, ref sweep |
| 4 | 68c627bb1 | fix(codex): round 4 — daemonEnv comments name inheritParentEnv, not fromSource (F13 sweep) |
| 5 | a554d5fe2 | fix(codex): round 5 — driver docblocks name the fromSource/inheritParentEnv split (F13 final sweep) |

- **F1** — `fromSource` now controls two separate decisions: whether to avoid `systemd-run` and whether to inherit the full parent 
- **F2** — `--env` is parsed only after the endpoint has been connected
- **F3** — `parseEnvAssignments` accumulates into an ordinary `{}`
- **F4** — Explicit reserved assignments are silently changed instead of rejected
- **F5** — The canonical PRESENTATION set carries only `LC_ALL` and `LC_CTYPE`, while the changed README and remote-create docs pro
- **F6** — The two `daemonEnv` tests do not provide the exact-set pins claimed by the change
- **F7** — The module header still says `cleanEnv passthrough`, and line 144 says an undefined whitelist means `passthrough mode`; 
- **F8** — The detached-branch comment still says the forwarded env is layered on the parent, directly contradicting the new non-`f
- **F9** — The file header and `packages/kaval-tui/README.md:27` still call raw create a spawn with `no kolu policy`, even though t
- **F10** — The local-create documentation lists `DISPLAY` as a functional var, but the pinned source intentionally classifies it as
- **F11** — The spawn composer still describes `cleanEnv()` as parent-env passthrough, and line 412 says inherited `KAVAL_*` rides t
- **F12** — The reference and its rendered snippet (`packages/surface/example/snippets/supervisor.ts:40`) tell readers the allowlist
- **F13** — The inheritParentEnv split was not propagated to the fleet-top part-2 source launcher
- **F14** — The round-1 fix inserted REJECTED_ENV_KEYS between the old parseEnvAssignments JSDoc and the parser, then added a second

srid:

Ran with `--no-elegance` (the `/simplify` pass already ran this tree). Two passes over the diff vs `origin/master`:

- **Pass 1 (rule checklist)** — one finding, actioned: `invalid-states-unrepresentable`. `DaemonSpawnConfig` split launch mode and env inheritance into two independent booleans (`fromSource` + `inheritParentEnv`), whose "inherit only when forced detached" invariant lived only in prose — `{ fromSource: false, inheritParentEnv: true }` type-checked and would have reopened the #1872 leak on a non-systemd host. Collapsed into ONE union `fromSource?: false | { inheritParentEnv: boolean }`, so the illegal state is now a **type error**, not a prose invariant (commit `f8e4f8cbf`). No live bug — all callers derived the fields correctly; a latent future-caller footgun. Every other checklist rule: clean.
- **Pass 2 (fact-check)** — **no issues.** Traced the allowlist direction (a true allowlist, not a leaky denylist), `cleanEnv`'s widen-not-narrow precedence, the multi-hop server→padi→kaval env cascade, the `__proto__`/`constructor`/`prototype` null-proto handling, and the detached-branch behavior change (called out as a downstream `<Aside>` in the reference). Exact-key `toEqual` pins would fail loudly on any drop or leak. "Unusually well-fortified change."

srid:
## Evidence — the #1872 A/B, leak closed end-to-end

The original field repro: `kaval-tui create -- claude` from an orchestrating Claude session copied that session's `CLAUDE_CODE_CHILD_SESSION=1` into the child, so the spawned claude classified itself as a nested child and never persisted its conversation. This A/B spawns a child through **this branch's** `kaval-tui create` from a session carrying the orchestrator's identity vars, and reads the child's own environment.

**A — BEFORE (the orchestrator session's own env — what the old wholesale forward copied in):**

```
CLAUDE_CODE_CHILD_SESSION=[1]
CLAUDECODE=[1]
CLAUDE_CODE_SESSION_ID=[orch-abc123]
AWS_SECRET_ACCESS_KEY=[shhh]
```

**B — AFTER (the CHILD env, composed by this branch's `create`):**

```
CLAUDE_CODE_CHILD_SESSION=[UNSET]     ← the #1872 var: gone
CLAUDECODE=[UNSET]                    ← gone
CLAUDE_CODE_SESSION_ID=[UNSET]        ← gone
AWS_SECRET_ACCESS_KEY=[UNSET]         ← a secret: gone
FOO_ARBITRARY=[UNSET]                 ← an arbitrary/unknown var: gone (allowlist, not blacklist)
HOME=[present] PATH=[present] SHELL=[present]   ← the canonical base: kept
```

The identity vars — and secrets, and any unknown var — are **composed out**; the functional base survives. A claude spawned this way never sees `CLAUDE_CODE_CHILD_SESSION`, so it persists its conversation. The `FOO_ARBITRARY` case is the point of an allowlist over a blacklist: a var nobody named still can't leak.

**The other seams are proven too:**
- **`cleanEnv` (every canvas/browser PTY)** — validated end-to-end by CI's e2e lane (the harness runs a real server → padi → `cleanEnv`-composed PTYs), and pinned deterministically in `shell.test.ts` (the exact allowlist as data + a `CLAUDE_CODE_CHILD_SESSION` drop assertion).
- **the supervisor detached branch + both `daemonEnv` twins** — pinned with exhaustive `toEqual` key-set tests (a dropped base key or a leaked ambient key both fail loudly).

Per the plan, the full real-claude transcript A/B is **not a CI node** — spawning a real claude is neither cheap nor deterministic — so it rides here as this env-composition proof plus the deterministic per-seam pins.

srid:
## Post-gauntlet CI — two more instances of the same class, caught end-to-end

The e2e detection seam is what the local #1872 A/B can't reach (spawning a real claude is neither cheap nor deterministic), so CI surfaced two further bugs — both the *same shape* as the fix itself: composing the env correctly **dropped** a var that had been riding ambient env wholesale, exposing where the old passthrough was silently load-bearing.

| Commit | What the composed env exposed | Fix |
| --- | --- | --- |
| `649cab166` | Six operational `KOLU_*_DIR`/`_DB` agent-session overrides rode the wholesale inherit into padi; composing dropped them, so padi's detectors resolved the *real* `~/.claude`/`~/.codex` and 272 agent-detection scenarios went `state=null`. | Forward the named agent-dir set into the daemon env, forward-if-set (the `KOLU_DIAG_DIR` pattern) — operational, non-identity, **explicit**. Hardens any forced-detached prod box that relocates agent dirs, not just CI. |
| `1b1a46dbb` | The e2e harness itself leaked git **identity** (`GIT_AUTHOR_*`/`GIT_COMMITTER_*`) via ambient env into every hosted PTY; composing dropped it (authorship *is* identity — exactly what this PR forbids), so fixtures' in-terminal `git commit` aborted on the pristine-NixOS runner (no `~/.gitconfig`) → branch/local Code-tab setup barriers timed out. | Give identity the legitimate channel: seed the fake `$HOME/.gitconfig` (the forwarded-HOME path that already sources `.bashrc`) — **not** a re-widening of the allowlist. The dead GIT_* whitelist-widening is removed. |

The distinction is the doctrine working: the agent-dirs are *operational* → forwarded explicitly; git identity is *identity* → never rides ambient env, travels by config. Both platforms green on `1b1a46dbb`.

PR #1881, opened 2026-07-17
Remote-host renew that never converged: isolation-default (each kolu binds a host in its own private estate)

srid:
## What this fixes

The URGENT follow-up to #1876: srid drove the skew card's **"Update & restart kaval"** against a real host (sincereintent) and **the recovery never converged** — the `incompatible` card returned forever.

**The brief's diagnosis was falsified by ground truth** (an experiment + live read of the running server + two code traces), and the coordinator gate accepted the falsification in full:

- provisionAgent's warm probe already serves the *fresh* drv given a fresh drv (proven: `nix-store --realise <freshDrv> --add-root gcroots/padi` returns fresh + repoints) — **not** a presence-skip bug.
- The renewed padi bakes the **current** kaval (`0jlpm2` = `KAVAL_COMMIT_HASH` of the running server), not a stale 5.0.
- The separate `gcroots/kaval` is a **vestigial `kaval-tui --host` leftover, read by nothing**.
- The recycle path is **code-correct** for a wire skew (kill + wait-gone + respawn fresh; loud on failure).

**The real mechanism** (coordinator-verified live: old padi supervisors still running, parenting 5–26-day-old kavals): a **cross-supervisor livelock** at a *shared* rendezvous. Two kolu-servers (or a live old-closure survivor + a fresh binder) that bind one host with the **default** state-root share one `kaval-<digest>` estate; each padi classifies the other's kaval as contract-skew and recycles it — **forever**. Arbitration by gate can't work: an old-closure padi predates any gate. The only fix that converges is to **stop sharing the estate**.

## The fix — isolation-default (ratified after an adversarial gate)

Each kolu-server binds a remote host in **its own private estate**, keyed on a **stable persisted per-client UUID**:

1. **`clientId`** — a minted-once UUID persisted in kolu-server's conf store (`SCHEMA_VERSION` 1.34.0; minted-on-read so old files never trip the required field). Stable across restarts → the same client **re-attaches** its estate; distinct across installs → **no shared rendezvous**. Chosen over a `hostname+path` salt because a UUID is collision-free *and* rename-proof (the salt orphans on a hostname rename with no auto-heal).
2. **Isolation at the seam** — `composePadiExtraArgs` forwards `--client-id` when no explicit `KOLU_REMOTE_PADI_STATE_DIR` override is set; padi anchors to `$HOME/.local/state/padi-<uuid>` computed **on the host** (only the opaque id crosses the wire — the identity-anchor invariant holds).
[...]

PR #1883, opened 2026-07-17
RENEWS — closed: the product was already correct (honest journey)

srid:
## Outcome: closing with **zero product-code change** — the honest result.

srid's directive was *"fix the sincereintent bug through Restart kaval — stop old, start new, restore. Nothing more complicated than this."* The gauntlet's honest conclusion, reached by falsifying five successive shapes:

**The product was already correct.** The incompatible-skew card already used `renewDaemon` (drain → re-realise the closure) on purpose — the code even documents why: `incompatible` is emitted *only after* a fresh spawn from the current closure has already skewed, so "restart" again is *"exactly the retry that just failed"* (`endpoint.ts`). There was no in-scope product bug.

**The real bug was graveyard pollution, not code.** sincereintent's loop is a **gate-less orphan kaval (pid 25494) squatting the canonical rendezvous socket** — an artifact of months of version churn. It makes even the correct `renewDaemon` loop (the re-provisioned padi re-adopts the squatter). Nothing in-scope displaces it: reaping is banned by the teardown law, the socket-reclaim isn't durable, republication is too heavy. It's the polluted-box case the brief explicitly scopes out — converged by a **one-time exact-pid reap of the stray** (an operator action), not product code.

## The journey — five shapes, each caught by the gauntlet before it shipped

1. **Isolation-default** (per-client private padi estate, #1881) — abandoned by srid (too architectural).
2. **Socket-reclaim** (a gate-holder takes the rendezvous from an ungated squatter) — **codex found + reproduced** it's not lifetime-safe: Node's `server.close()` unlinks the socket *path*, so the orphaned stray's later clean exit deletes the fresh kaval's rendezvous and strands it. Reverted.
3. **Client re-point** (incompatible card → `recycleKaval`) — **codex F3, verified against the code**: `recycleKaval` can *never* converge `incompatible` (a respawn already failed by construction); the original `renewDaemon` was deliberately correct. Reverted.
4. **move-4** (a client-side "did-not-converge" marker for an honest toast) — **architecture + simplify-altitude + code-police** converged: the client marker is the wrong depth *and* buggy — police #1 (verified) showed the clear-effect fires on a dead-channel link blip, *reproducing the very silent loop it exists to fix*, plus reload-loss and two reconciliation races. Not shipped.

## What the lane produced

- **Zero regressions shipped** despite five plausible-looking attempts — every wrong turn caught before master.
- A **correct understanding** of the actual failure (a stray process, not a code bug).
- A filed **follow-up**: the honest toast done at the *right depth* — a binder-authored `renewAttemptedStillFailed` field on `DaemonStatusSchema`'s `incompatible` arm (cleared on arm-flip, replayed across reload, per-host by construction) — as its own scoped PR (it re-opens the surface/drishti gates).

## Root lesson

The whole lane traces to the original brief asserting an *inferred* mechanism (*"padi bakes a 5.0 kaval / presence-skip cache"*) as fact. **Ground a diagnosis (reproduce-first) before scoping any fix.** `/self-improve` ships that lesson as its own PR.

_Closing per srid. sincereintent converges via a one-time reap of stray 25494 (operator action)._


srid:
Closing per srid's ruling — the RENEW/RENEWS lane's honest outcome: **zero product code, zero regressions across five gauntlet-caught wrong turns.**

The product's `incompatible → renewDaemon` design was already correct (the code documents why: `incompatible` means a respawn already failed, so offering restart again is "the retry that just failed"). The real sincereintent bug was **graveyard pollution** — a gate-less orphan kaval (pid 25494) squatting the canonical rendezvous socket — the out-of-scope polluted-box case the brief itself scopes out, converged by a one-time exact-pid reap of the stray (operator action), not product code.

The gauntlet earned its cost: it falsified isolation-default (#1881), the non-durable socket-reclaim, the incompatible→recycleKaval re-point (F3), and the client-Set move-4 (police #1: a dead-channel link blip reproduces the very bug it fixes) — **each caught before master.**

Filed follow-up: the honest toast done at the **right depth** — a binder-authored `renewAttemptedStillFailed` field on `DaemonStatusSchema`'s `incompatible` arm (durable, per-host, race-free by construction) — as its own scoped PR.

Root lesson (shipped in #1886): ground a diagnosis (reproduce-first) before scoping any fix; the lane traced to one inferred-mechanism-as-fact at the top.

PR #1887, opened 2026-07-18
Recover the rendezvous from a gate-less socket squatter

srid:
A kaval orphan whose gate file (`kaval.pid`) is gone but which **still holds the rendezvous socket** wedged kolu *forever*: `liveServingHolder` is gate-only, so recycle never targeted the orphan, the fresh spawn couldn't bind, and it handshaked the orphan into an endless `incompatible` — the incompatible-daemon card's renew looped, and the only escape was killing the process by hand (field case: `sincereintent` 25494, 8 days). **This makes the supervisor identify its own socket's holder and recycle it, so the product heals itself — no hand-reaping.**

The state — socket held ∧ gate absent — is representable and unhandled. The gate-only `liveServingHolder` correctly *refuses* to kill a pid it can't prove is the daemon; the gap is a **missing identity source**, not a broken check.

### The identity model — three attestations that must agree before any SIGTERM

The supervising process owns supervision of its own verified daemons; the safety line is **identity**, not bookkeeping:

| Attestation | Source |
| --- | --- |
| The socket speaks kaval | the existing version handshake succeeds (compatible **or** skewed) |
| A specific pid holds *this* socket | the OS socket-holder lookup (`/proc/net/unix` on linux, `lsof` on darwin) |
| That pid is the one that answered | the daemon's `system.version.pid` (a required field since #1301) |

All three must agree — and be **re-confirmed unchanged immediately before the signal** — or the kill aborts. What the recovery does with a gate-less holder, by what the handshake proves:

- **skewed** kaval → the wedge (the 25494 case): recycle it (SIGTERM the verified pid → `waitForPidGone` → bind fresh), reusing `killLiveHolder`'s machinery.
- **compatible** kaval → *left for the ordinary connect to adopt* — its live PTYs are preserved, exactly as today. A compatible gate-less holder never wedged; recycling it would be a PTY-loss regression.
- **foreign** / non-conforming (a stranger, or a version response missing the required `pid`) → **never killed** — a loud `SocketSquatterForeignError` naming the pid + command.

It also never recovers a socket held by **its own process** (self-kill guard — the daemon is always a separate process).

This makes the existing `incompatible → renewDaemon` flow **self-heal**: fresh padi boots → socket held gate-less → handshake → old kaval → recycle → bind the new contract → converge, one click.

### Where it lives

A new leaf `socketHolder.ts` (`socketHolders(path)`), beside `waitForPidGone.ts` — pid lifecycle via the OS. The recovery is a private `endpoint.ts` helper run by `recoverOrSpawn` **only where no live gate holder was found** (ensure's else, adoptSurvivor's no-survivor branch — hence converge + renew), never the recycle path. "Speaks kaval?" reuses `connectSurvivor`'s existing three-way verdict.

> The daemon's self-reported pid is threaded as an **additive, optional** field on `DaemonConnection` and `DaemonContractSkewError`. **Consumer gate:** neither drishti nor odu depends on `@kolu/surface-daemon-supervisor`, so the additive change has zero cross-repo impact; `ref-surface-supervisor.mdx` is updated in the same PR.

### Verification

Red-first pins drive a **real subprocess holding a real unix socket** (the only faithful way to exercise the OS lookup + actual SIGTERM): the skewed squatter is recycled and the fresh daemon binds durably; a foreign holder is left alive with a loud typed error; a pid-absent speaker is foreign; a compatible holder is adopted, not killed. The existing killLiveHolder/adopt/restart suites stay green.

Design of record: `docs/atlas/src/content/atlas/bug-gateless-socket-squatter.mdx` (gate `wf_19a628c2`: 0 confirmed / 8 refuted).

### Try it locally

```sh
nix run github:juspay/kolu/squatter-recovery
```


srid:

✅ **Consensus** with zero debate turns · lowy + hickey · base `65bc08dc6649`

Settled: 1 reconciled (raised by both lenses) · 1 auto-settled minor · 2 unopposed

### Applied (1)
- `lowy-1` ≡ `hickey-1` Skew-disposition policy is one axis of change split into two hand-synced representations — commit `14ab00927`

### Agreed — no change (3)
- `lowy-2` SocketHolder mixes a load-bearing identity field with a rarely-used diagnostic, eagerly computed (packages/surface-daemon-supervisor/src/socketHolder.ts:37-43, 112)
- `hickey-2` `unreachable` verdict silently reinterpreted as `foreign` at the recovery site (packages/surface-daemon-supervisor/src/endpoint.ts:580-584 (recoverGatelessSquatter) vs 896-911 (adoptAt))
- `hickey-3` Three-attestation kill guard scattered across inline conditionals (packages/surface-daemon-supervisor/src/endpoint.ts:612-639 (recoverGatelessSquatter recycle arm))

srid:
## Codex ⇄ Claude debate

✅ Consensus in 6 round(s) · reviewer effort: xhigh · base: 65bc08dc6

|---|---|---|
| 1 | 47ab2d175 | fix(codex): harden gate-less recovery — adopt-reporting, re-handshake before kill, bounded handshake |
| 2 | a73d6e23c | fix(codex): round 2 — bound the converge probe, reorder re-attest, hint recovery, fail-loud |
| 3 | bbd459693 | fix(codex): round 3 — held tracks adopted rendezvous, fail-loud on accepting, no deadline knob |
| 4 | 4a16766dd | fix(codex): round 4 — F5, apply the no-spawn-over-accepting rule to the in-loop empty-holder branch |
| 5 | c4f0f457c | fix(codex): round 5 — F15, correct freeOrFailLoud log prose + the four-way `free` comment |

**Legend** — findings codex raised (F1–F15):

- **F1** — The compatible gate-less branch disposes the successfully handshaken survivor connection and returns `spawn`
- **F2** — The recovery's proof handshake has no deadline
- **F3** — The advertised three-part identity is not re-confirmed before SIGTERM
- **F4** — Gate-less recovery is hard-coded to `spec.socketPath`/`spec.connect` and runs only after the gated adopt-hint 
- **F5** — When the holder changes during verification, the code logs that recovery was aborted but returns `spawn`
- **F6** — A compatible gate-less recovery against an endpoint that already has a live connection eventually calls `holdC
- **F7** — The Linux parser splits the entire `/proc/net/unix` row on whitespace and compares only field 8
- **F8** — `isSocketSquatterForeignError` checks only the public brand but narrows to a class that promises `socketPath` 
- **F9** — On Darwin the holder list may contain connected clients as well as the listener, yet `firstHolder.pid` is pass
- **F10** — A gate-less recycle whose `killLiveHolder` times out emits `dead` inside `killLiveHolder`, then the broad `rec
- **F11** — New round-2 regression: if the pre-kill re-verification handshake finds that the holder became compatible, `co
- **F12** — New round-2 regression: `connectKaval` now exposes `handshakeDeadlineMs` as a public runtime override solely t
- **F13** — New round-2 regression: the recovery contract comment still says a compatible probe connection is dropped and 
- **F14** — New round-3 regression: the four-way restructure removed `recoverOrSpawn`, but the recovery header at line 554
- **F15** — New round-5 regression: `freeOrFailLoud` is shared by both an empty-holder lookup and cap exhaustion, but its

srid:

Two passes over the squatter-recovery diff (`--no-elegance` — `/simplify` ran separately). Every finding actioned in `fix(police)` (`11e1d83c5`):

**Pass 2 — fact-check (correctness):**
- **transient branded permanent** — the gate-less recovery treated `connectSurvivor`'s *transient* `unreachable` verdict as *permanent foreign* and threw immediately — the opposite of how `adoptAt` treats the identical verdict (leave standing). A slow / heavily-loaded but legitimate kaval would have been branded foreign and failed the boot. **Fixed:** `unreachable` (and an uncorroborated skew pid) now re-decides via the loop; only a holder that stays unreachable across *every* attempt reaches the tail's fail-loud.

**Pass 1 — rules:**
- **no-sync-blocking-on-the-serving-loop** — `socketHolders` is now async (`fs.promises` for the `/proc` walk) and darwin `lsof` runs under a hard `timeout`; a wedged `lsof` can no longer hang the event loop that serves live terminals. **Fixed.**
- **callback-fanout-guarded-at-funnel** — `adoptHint.onAdopted` is invoked via a guarded `runHook`, so a throwing hook can't escape the "report `dead` before throw" contract. **Fixed.**
- **dry-rule-of-three** — the self-exclusion (`h.pid !== process.pid`) folds into one `externalHolders()` helper. **Fixed.**
- **prefer-focused-library** — `readSystemVersionBounded` and `hostInventory` share one extracted `withTimeout` leaf. **Fixed.**
- **prefer-ts-pattern** — `settle()` gets an explicit `free` case + a `satisfies never` exhaustiveness fence. **Fixed.**

_Code-police pass of `/be`'s review gauntlet._

srid:
## Evidence — the product heals itself from a gate-less socket squatter (no hand-reaping)

srid's demand was *"WHY is a stray process breaking kolu? I want the bug fixed"* — the field close (hand-killing pid 25494 on sincereintent after 8 wedged days) was cheating; the product must converge on its own. This fix is a **boot-recovery mechanism with no visual diff**, so the proof is *"a planted gate-less squatter is reaped automatically, and the boot converges"* — demonstrated end-to-end on the **field platform (darwin)**, since 25494 was on a mac and the darwin `lsof` socket-holder arm was the one fact not groundable on linux.

### 1. Darwin `lsof` socket-holder arm — verified on a real mac (rasam)

A faithful standalone replica of the shipped `darwinSocketHolders` (`lsof -F pcn -- <path>` + the exact `p`/`c`/`n` parse) run on `nix-infra@rasam` (real `aarch64-darwin`):

```
platform: darwin
listener pid: 22781  socket: /tmp/squat-lsof-…/pty-host.sock
darwinSocketHolders -> [{"pid":22781,"command":"node"}]
listener pid found: true  command: node
nobody.sock -> []          (empty for an unheld path)
after close -> []          (holder gone once it exits)
DARWIN-LSOF-ARM: PASS ✅
```

The OS-lookup identity source — the thing the gate no longer provides — resolves the exact holder pid + command on the field platform.

### 2. The full plant-squatter recovery runs on darwin CI (rasam) — real lsof, real SIGTERM

`ci::unit@aarch64-darwin` (on rasam) ran `squatterRecovery.test.ts` — **9 pins, all pass** — with the **REAL darwin `socketHolders`/`lsof`** (only the `/proc`-specific leaf test is linux-skipped):

```
✓ src/squatterRecovery.test.ts (9 tests) 5067ms
  ✓ THE squatter: a skewed gate-less holder is identified, recycled, and the fresh daemon binds  432ms
```

"THE squatter" plants a **real subprocess** holding a **real unix socket gate-less**, skewed while alive; the supervisor's own recovery then: `socketHolders` (real darwin `lsof`) identifies it → the existing handshake proves it's a kaval of this estate → `killLiveHolder` **SIGTERMs it** (the product's own reap — no human) → `waitForPidGone` → the fresh daemon binds. Plus the companion pins: a **foreign** holder is *never killed* (loud typed error), a **compatible** gate-less holder is *adopted* (PTYs preserved), a **padi** gate-less skew is *refused not killed* (#1313), and both **hint** sequences recover rather than abandon. All green on darwin **and** linux.

### 3. Two-platform CI green

Full `ci::default` on `cf3a7c8c4`: **28/28 checks pass** — every node on `aarch64-darwin@rasam` **and** `x86_64-linux@kolu-ci-1` (`e2e`, `nix`, `unit`, `smoke`, `flake-check`, `home-manager`, …).

### 4. Red-first

The pins are **red on the pre-fix code** (proven by neutering the recovery: the squatter/foreign/pid-absent pins fail — the wedge and the unhandled foreign/pid-absent states — while the compatible pin stays green because that case never wedged). The recovery flips them green.

> No browser screenshot: the fix has no on-screen diff — it turns an endless `incompatible` wedge into a self-healing boot. The recovery reaping a real planted squatter on the field platform, with no hand-reaping, is the behavior proof.


---

**On the absent browser screenshot — a decision, not an omission:** no visual delta — this is a behavioral fix. The incompatible card is visually unchanged; what changes is that its **action now converges** (an endless `incompatible` wedge becomes a self-healing boot). The convergence mechanism is pin-proven on the field platform (darwin) with a real planted squatter, real `lsof`, and a real `SIGTERM` reap. Manufacturing a two-build contract skew to screenshot a visually-identical card fails the proportionality test — **the evidence is the darwin pin suite.**

PR #1888, opened 2026-07-18
chore(skills): board-freshness Stop hook — the harness keeps the coordinator dashboard current

srid:
Answers srid's question: *"What would block you from ignoring such [dashboard] updates in future?"* — honestly, **nothing did**: keeping `orchestrator-data.js` current was memory, and memory failed (the human had to ask twice in one day). Per the repo's own doctrine, an "always do X" behavior needs harness enforcement, not promises.

## The hook

An APM `Stop` hook in the **agents package** (same shape and output contract as agency's `do-stop-guard`): when a session's PWD contains `orchestrator-data.js` (i.e. it *is* a coordinator) and the file is **>45 minutes stale**, the hook **blocks the turn from ending** with a reason instructing the coordinator to update the board (or touch it if genuinely nothing changed). `stop_hook_active` guards against block loops; non-coordinator sessions exit instantly on the missing file.

45 minutes: long enough for a quiet gate-wait, short enough that a merge, stand-down, dispatch, or lane change can't slip a whole turn behind.

The wrong thing ("end the turn with a stale board") becomes something the coordinator *cannot do*, rather than something it remembers not to do.

### DIFF
--- a/.claude/apm-hooks.json
+++ b/.claude/apm-hooks.json
@@ -5,6 +5,16 @@
       "hooks": [
         {
           "type": "command",
+          "command": ".claude/hooks/agents/scripts/board-freshness.sh"
+        }
+      ],
+      "_apm_source": "agents"
+    },
+    {
+      "matcher": "*",
+      "hooks": [
+        {
+          "type": "command",
           "command": ".claude/hooks/agency/scripts/do-stop-guard.sh"
         }
       ],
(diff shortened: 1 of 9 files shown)


srid:
Closing — obsoleted by the merged campaign-surface design (#1878): this hook enforces the hand-maintained-dashboard ritual that the observed Board (P2) deletes. Keeping it would hard-code the scribe. (Coordinator.)

PR #1865, opened 2026-07-16
kolu mcp — the agent face: padiSurface served to coding agents over MCP

srid:
**`kolu mcp` ships — the kolu binary's agent face** (PR2 of the [kolu-cli plan](https://kolu.dev/atlas/kolu-cli.html)): this host's terminals served to a coding agent over MCP stdio, as a **pure `padiSurface` client** — no kolu-server process anywhere. This is the graduation proof's headless leg: `padiSurface` consumed through the surface client, `terminalAttach` included, over **both** transports (local digest-keyed socket, ssh stdio), by a second frontend that provably needs no kolu.

```
coding agent ──MCP stdio──▶ kolu mcp ──padiSurface──▶ padi ──▶ kaval ──▶ PTYs
                            (kolu-cli + kolu-mcp)      └── the same daemon the browser canvas drives
```

### The face, as ratified

- **New `packages/kolu-mcp`** — the manifest IS the graduation fence: padi/surface deps only, so a kolu-app import is a missing-dependency build error, not a lint. It exports `serveKoluMcp` over an **injected** connected client; **kolu-cli owns the connect layer** (local `resolveRunningPadiSocket`+`connectPadi`; ssh via a thin `dialAgentOnce` wrapper — the padi-tui `hostConnect` shape) and the `mcp` arm.
- **Default-deny expose map with NAMED, tested denials.** Resources: `terminals` · `urgency` · the status story (`status` + `daemonStatus` + `identity` — generation visibility, so a daemon restart is *data*). Read-only tools: `screen_text` (tail mode) · `screen_history` · `git_getStatus/getDiff` · `fs_listAll/readFile`. Mutating: `lifecycle_create` · `lifecycle_kill` · `lifecycle_sendInput`. Every denial (`terminalAttach`, `lifecycle.killAll/recycleKaval/…`, `chrome.*`, `session.*`, worktree/scratch writes) is a recorded non-entry the unit suite proves unreachable at the wire — *widening is a one-row diff with a review, never a default*.
[...]

PR #1890, opened 2026-07-18
Make the Dock see a direct-spawned agent (#1872 PR2)

srid:
**An agent launched as a terminal's own root process — `kaval-tui create -- codex …`, with no shell wrapping it — now surfaces in the Dock like any other.** Two independent locks kept a command-rooted agent invisible: kaval threw away the command line it was handed at spawn, and a "is the shell idle?" check misread the agent-as-root as an idle prompt. This is PR 2 of #1872 (mechanism A — detection loss); the plan of record is the `agent-spawn-first-class` atlas note.

### Reproduce-first turned up a real correction

srid's rule was to reproduce in a test before any fix code. Doing so **falsified the brief's premise for `claude`**: a command-rooted `claude` with a live session is *already* detected today, because claude resolves by `foregroundPid` + its on-disk session file — a path independent of both locks — and for a command-rooted PTY the agent's pid *is* the foreground pid. The field's three invisible `claude` coordinators were almost certainly **Bug B** (the env leak → no session file written), which PR1 already fixed.

So the two locks bite in exactly one place: an agent matched by its **command hint** (`codex`/`opencode`/`grok` via `matchesAgent`) whose kernel `comm` ≠ its name — an npm CLI that runs as `node`. That correction went to the coordinator gate before any fix code; the ruling was **proceed, reframed onto the hint path, and state the impact honestly** — which this PR does.

### The two locks

| Lock | Where | The bug | The fix |
| --- | --- | --- | --- |
| **1 — kaval drops the argv** | `ptyHost.ts` | A shell-less PTY never emits the OSC 633;E mark that is `lastCommand`'s only writer, so the daemon discarded a command line it already had. | Seed `lastCommand` + the initial title from the argv at spawn when `commandRooted`, on the same field + channel the 633;E handler uses (a later live mark is the last-writer). |
| **2 — shellIdle misreads the root** | `sensors.ts` | `shellIdle = foregroundPid === pid` assumes `pid` is a shell's; for a command-rooted PTY that equality means the agent is *busy*, yet it force-nulled the command hint. | The sensors take a `commandRooted` discriminant: a command-rooted PTY never reads shell-idle, so the hint survives and `matchesAgent` succeeds. |

### The seam

The `commandRooted` fact is born where kaval-tui already knows it (`command.length > 0`) and travels **kaval → padi only** (server-side; the browser renders agent state unchanged, so no `@kolu/surface` / terminal-vocab change):

```
kaval-tui composeCreateInput ─▶ ptyHostSurface spawn input ─▶ PtyHost.spawn (seed, lock 1)
                                                            └─▶ inventory list entry ─▶ padi adopt ─▶ startAgentSensor (lock 2)
```

Reproductions landed red first — the R1 e2e (an `opencode` npm shim spawned command-rooted → invisible in the dock), R2 (kaval `getLastCommand`), R3 (padi shellIdle) — then flipped green with the fixes. The R1 suite **keeps a `claude` scenario as a green regression guard** documenting that the pid path already works command-rooted.

> **Versioning: no `ptyHostSurface` contract bump.** `commandRooted` is two *optional additive* fields (spawn input + list entry), not a new emitted union variant — its absence degrades to exactly today's shell-rooted reading in both skew directions. Bumping would force-recycle a surviving kaval (killing live PTYs) to buy a feature whose absence is the status quo. Same call the `lifetime` field made at 5.0; the negative of the emitted-variant rule (#1865 fold-back vs #1876's incompatible-arm bump).

> **Consumer gate: `none`.** The change is in `packages/kaval` + `packages/padi`, not `packages/surface*`, so the drishti-pair gate doesn't apply. Every consumer of the changed exports is a kolu package (grepped); odu consumes `@kolu/surface`, not kaval's wire — odu-impact `none`, nothing for odu#43.

This PR also retires the interim "never `create -- <agent>`" skill rule for the *detection* reason (the own-shell `CLAUDE_CODE_*` scrub, tagged for PR3, stays).

### Review gauntlet & a reusable versioning doctrine

Went through architecture-first-principles (C1: use `shellJoin`, not a bare join), the lens debate (one `isShellIdle` predicate; the sensor-layer `commandRooted` made a required boolean), and a 9-round Codex ⇄ Claude debate that hardened the command-parse seam: the command's quoting **dialect** (a `shellJoin` seed vs a raw OSC-633 line) is now carried explicitly per-event on the `commandRun` frame and reparsed by its exact inverse — no string sniffing, no derivation.

One versioning question was adjudicated by the campaign coordinator. The `commandRun` `shellJoin` field (and the `commandRooted` fields) are **wire-optional, no contract bump** — a required field would force-recycle surviving kavals and kill live PTYs. Codex read the optionality as a fail-fast violation; the ruling, stated here for reuse:

> **wire-additive-optional is not the banned graceful-degradation fallback when (i) absence maps to a world where the fact genuinely doesn't exist, (ii) resolution happens once at the bridge, and (iii) the interior type is required.**

All three hold here: lock 1's seed and the field ship in the same kaval build (absence ⟺ no command-rooted seed ⟺ raw is correct), the skew default lives once at the pty-host bridge, `CommandRunSample.shellJoin` is required internally, and an emitter-exhaustive test pins that every seed frame carries the flag — so the illegal half-state is test-unspellable without a contract bump.

Part of #1872.


srid:

✅ Consensus (deepest thread: 1 round) · base: origin/master

Independent reviews: lowy 2 findings · hickey 3 findings. Two settled by cross-lens reconciliation, one debated to agreement. All applied.

| Finding | Lens(es) | Disposition | Commit |
| --- | --- | --- | --- |
| The "foreground==root means idle-or-busy" decision was duplicated across `snapshotSignals` and `reconcileInner` | lowy-1 ≡ hickey-1 | fix — extract one `isShellIdle` predicate; `snapshotSignals` returns its companion so the two paths can't desync | b318d9105 |
| `SensorInputs.commandRooted` / `startAgentSensor`'s param were an optional-with-default (a dead fallback) | lowy-2 ≡ hickey-3 | fix — make them a required `boolean`; resolve "absent = shell-rooted" once at the wire edge | 87eef3ca2 |
| `commandRooted` + `argv` re-derived the `command.length > 0` guard | hickey | fix — derive both from one narrowed local | 58f0cfbd8 |

Structural review of the two-lock seam: the `commandRooted` fact is threaded through existing wire/inventory seams (no new receptacle), and the shell-idle decision now lives behind one predicate. No unresolved findings.

srid:
## Codex ⇄ Claude debate

🟡 9 rounds · reviewer effort: xhigh · base: origin/master

Consensus on F1–F7 (all fixed + verified). F8 (a fail-fast-vs-wire-versioning disagreement) was adjudicated by the campaign coordinator in the author's favor — see the note below.

|---|---|---|
| 1 | aa1dd47b0 | fix(codex): drop the ineffective title seed + doc-sync + F1 regression (#1872) |
| 2 | 02db40e35 | fix(codex): shellSplit fallback for the command-rooted seam + test/doc fixes (#1872) |
| 3 | 2df45f72f | fix(codex): route parseAgentCommand by the shellJoin idiom, not a fallback (#1872) |
| 4 | 682eeba40 | fix(codex): carry the command dialect via commandRooted, not a string heuristic (#1872) |
| 5 | 6d78f297b | fix(codex): select the seed dialect per-EVENT via commandRooted && replayed (#1872) |
| 6 | 5bd9b53d7 | fix(codex): carry the command dialect as an explicit per-event field (#1872) |
| 7 | 72b3e185f | fix(codex): make the command dialect a required internal field, no parsing fallback (#1872) |
| 8 | 4a9916312 | test(kaval): emitter-exhaustive pin — every seed frame carries shellJoin (#1872) |

**Findings**
- **F1** — the shellJoin↔string-argv command seam (closed by carrying the dialect explicitly, reparsed by its exact inverse)
- **F2** — the ineffective title seed (dropped; the tile shows the foreground process name)
- **F3** — agent-detection.mdx prose + SVG (now honest for shell- and command-rooted)
- **F4** — public pid / getLastCommand doc comments (root process, not shell)
- **F5** — the title unit test spawned a real `claude` binary (now `/bin/sh`)
- **F6 / F7** — the dialect tracked per retained command in kaval + carried on the commandRun frame (no derivation, no string heuristic)
- **F8** *(coordinator-adjudicated, ruling A)* — codex held the OPTIONAL wire `shellJoin` field a fail-fast violation; the author's position (ratified): wire-additive-optional is not the banned fallback when (i) absence maps to a world where the fact genuinely does not exist, (ii) resolution happens once at the bridge, (iii) the interior type is required — and the illegal half-state is closed by an emitter-exhaustive pin, not a bump (which would recycle live PTYs).

srid:

Ran on the diff vs `bad87ae5b` (rules + fact-check; elegance skipped — `/simplify` already ran).

| Pass | Result |
| --- | --- |
| Rules | 1 fixed — `no-untyped-escape-hatches` |
| Fact-check | clean |
| Simplify (afp + lens + 4-angle) | clean (1 lateral finding on the `isShellIdle` tuple, skipped — lens-deliberate) |

**Fixed:** `spawn_detection_steps.ts` cast a Cucumber step string with `state as never` (zero static check — a `.feature` typo would compile). Now `state as AgentLifecycleState`, the type `writeOpenCodeFixture` expects, matching the `claude_code_steps` convention. Fact-check traced every ratified `?? false` skew-default, the two-lock design, the optional `commandRooted`/`shellJoin` fields, and the dual-tokenizer routing to their documented rationale — no logic errors, silent swallowing, or unbounded growth.

srid:
## Evidence

A command-rooted opencode npm-shim agent (no shell — the terminal's ROOT process IS the agent) now shows as a working pill in the Dock. Before this PR the agent ran invisibly: the terminal's root-process command was never inspected for agent hints, so the dock stayed empty. Captured from the `spawn-detection.feature` scenario "A command-rooted opencode npm shim is detected via its command (hint path)" recorded under `KOLU_EVIDENCE=1` — the dock, expanded, shows the `.KOLU-E2E-HOME-DUUA1G` workspace with 1 row: the detected agent's working pill (`[data-testid="dock-row"][data-bucket="working"]`), exactly what the scenario asserts.

srid:
## SPAWN2 — merge-ready ✅

Reproduce-first → both locks → full review gauntlet → CI, all green.

- **CI green on BOTH platforms** (`x86_64-linux` + `aarch64-darwin`) at `4c21dd459`: `ci::unit`, `ci::e2e`, `ci::nix`, `ci::smoke`, `ci::pnpm-hash-fresh`, `ci::flake-check`, `ci::home-manager`, … all pass. `mergeStateStatus: CLEAN`.
- **Reproduce-first:** R1 e2e (a command-rooted opencode npm-shim → invisible in the dock) + R2/R3 unit reds landed **before** the fix, then flipped green; R1 keeps a green `claude` scenario as the pid-path regression guard.
- **Gauntlet:** architecture-first-principles (C1) → lens debate (3 fixes) → 9-round Codex ⇄ Claude debate (F1–F7 consensus; F8 coordinator-adjudicated) → simplify (clean) → code-police (1 fix).
- **Evidence** posted above; docs/changelog/skills synced.

Ready for review — I don't merge; **srid merges**.

PR #1891, opened 2026-07-18
Restore card: scrollable on a short viewport, with a pinned Restore footer

srid:
On a short window the session-restore card overflowed the viewport and the Restore button dropped below the fold — with a scrollbar you couldn't actually scroll.

### Why it was stuck

The scroll container was the **`empty-state` wrapper** (`overflow-y-auto`; its content overflows, so the scrollbar renders). But that wrapper is deliberately `pointer-events-none` so a double-click on the bare canvas falls through to create-on-double-click — and `pointer-events-none` **also stops the wheel from driving its scroll**. The **card** inside was `overflow: visible`, so it just spilled its Restore button past the bottom of the screen. On top of that, the list had a *second*, nested `max-h-[55vh] overflow-y-auto` — a scroll-within-a-scroll trapping the wheel.

### The fix — an app-shell card

The card is now a flex column capped to the viewport (`pointer-events-auto`, so the wheel reaches it):

- a **scroll region** (`flex-1 min-h-0 overflow-y-auto`) holds the welcome header, the session list, and the touch create/shortcut blocks, and
- the **Restore controls are a pinned footer** (`shrink-0`, `border-t`) — the "Resume agent sessions" toggle, the Restore button, and "Start fresh" stay put at the bottom, always one tap away, while only the list scrolls in the middle.

`groups` / `subCount` / `hasAnyAgent` were lifted to component scope so both the scrolling list and the pinned footer can read them.

Also folded in: the group heading was `sticky top-0` with an opaque mask, which tears a two-line session row as it scrolls under the pinned heading (name hidden, "Asleep · restores dormant" subtitle orphaned below it). Made it non-sticky.

Verified live against a 560px-tall dev server (see Evidence) — `just check` green, CI on both platforms.

### DIFF
--- a/packages/client/src/EmptyState.tsx
+++ b/packages/client/src/EmptyState.tsx
@@ -93,6 +93,16 @@ const EmptyState: Component<EmptyStateProps> = (props) => {
 
   const resumeCount = () => (resumeAgents() ? resumableIds().length : 0);
 
+  // Lifted to component scope so the scrolling session list and the pinned
+  // action bar — now separate flex children of the card — can both read them.
+  const groups = createMemo(() => {
+    const session = props.savedSession;
+    return session ? groupSavedTerminals(session.terminals) : [];
+  });
+  const subCount = () =>
+    props.savedSession?.terminals.filter((t) => t.parentId).length ?? 0;
+  const hasAnyAgent = () => resumableIds().length > 0;
+
   const handleRestore = () => {
     const resumeIds = resumeAgents()
       ? new Set(resumableIds())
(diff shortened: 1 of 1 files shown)


srid:
## Evidence

Faithful before/after of the restore-card scroll container, rendered in a real browser with the component's exact DOM structure and Tailwind utility values (light theme, `bg-surface-1` card, `☾` asleep glyph, `Asleep · restores dormant` subtitle), scrolled so a session row sits partway under the repo heading — the same state as the reported screenshot.

before/after

- **BEFORE** (`sticky top-0 z-10 bg-surface-1`): the pinned **kolu** heading masks the first row's name (`e2e-ollama-real-agent`), leaving an orphaned "Asleep · restores dormant" floating under the heading — the bug.
- **AFTER** (heading scrolls with its rows): the top visible row shows its name with the subtitle directly beneath it; no orphaned status line.

> Captured off the live app because reproducing this state needs a repo group with enough sleeping sessions to overflow the `max-h-[55vh]` container *and* a scroll offset that lands mid-row; the fix is pure CSS geometry in those heading classes, so the class-faithful render exercises the exact code path that changed.

srid:
## Evidence — the actual fix, verified live

Driven against a running dev server at a short viewport (560px tall), the state that reproduces the bug.

**BEFORE** — card overflows, Restore button below the fold, scrollbar visible but the wheel can't move it (scroll container is `pointer-events-none`):

before

**AFTER** — the card is now the scroll container (`pointer-events-auto`), scrolls with the wheel, and the "Restore 3 terminals · resume 1 agent" button + "Start fresh" are reachable:

after

Measured live after the fix: `card { pointerEvents: "auto", canScroll: true }`, `emptyState.canScroll: false` (the card owns the scroll now), and `restoreBtnVisibleAfterScroll: true`.

srid:
## Evidence — pinned footer, only the list scrolls

Driven live against a dev server at a short 560px-tall viewport (the state that reproduced the bug), scrolled to the middle so both the scrolling list and the pinned footer are visible in one frame:

pinned footer

The welcome header + session list scroll in the middle; the **Resume agent sessions** toggle, **Restore 3 terminals · resume 1 agent** button, and **Start fresh** stay pinned at the bottom, always reachable.

Measured live: `scrollRegion.canScroll: true`, `footerIsPinned: true`, `btnPinned_didNotMoveOnScroll: true`, `btnAlwaysVisible: true`. Contrast with the earlier BEFORE (unscrollable, button below the fold): [`live-01.png`](https://github.com/juspay/kolu/releases/download/pr1891-evidence/live-01.png).

srid:
## Codex ⇄ Claude debate
✅ Consensus in 2 rounds · reviewer effort: high · base: origin/master

Round 1 surfaced two findings on the pinned-footer refactor; round 2 verified both as resolved. The one commit the debate produced:

| --- | --- | --- |
| 1 | 2b8130220 | fix(review): footer fallback scroll (F1) + touch divider (F2) |

**Legend**
- **F1** (major) — card `overflow-hidden` + `shrink-0` footer clipped the Restore controls at a viewport shorter than the footer's own height, with no scroll to reveal them. Fixed: card `overflow-y-auto` fallback (normal case still scrolls only the inner list; no double scrollbar).
- **F2** (minor) — the session-restore block lost its `mb-5/pb-5/border-b`, so on touch layouts the New terminal / Get started content butted against the last session row. Fixed: divider restored via `classList` gated on `!isDesktop()` (desktop uses the footer's own `border-t`).

Issue #1793, opened 2026-07-13
Kaval & Padi dialog are not reactive of connection status


srid:
Case in point

<img width="1601" height="1692" alt="Image" src="https://github.com/user-attachments/assets/85bcc46d-4d07-4107-bb3e-b9044a422827" />

srid:
Fixed by #1892 (merged 2026-07-19): dialogs are presence-only — facts and affordances exist only on the `connected` arm of the fold, so rendering a connected-era value over a dead channel is a type error. The reproduce-first RED test (dad484c47) is green, plus pins for the presence projections and the `daemonScanCause` fold. The affordance axis (enabled Restart over a dead channel) was swept in the same PR via `offerRestartVerb`'s unknown→false arm.

PR #1892, opened 2026-07-19
fix(client): make info dialogs presence-only so they can't leak stale facts over a dead channel (#1793)

srid:
## The bug (#1793)

When a host goes unreachable — `ssh: connect to host zest port 22: Connection timed out` → `host unreachable — retrying … (attempt 11)` — its daemon status **freezes** at its last connected read while the channel that would refresh it is gone. The Kaval info dialog stayed **half-honest**: the status pill correctly read "unknown" and build/memory/lifetime blanked, but the **"contract v5.2" header badge** and the **socket path** kept painting their stale last-good values. Two fields read `props.status` *raw*, bypassing the `toKavalPresence(status, live)` fold every other field went through.

## The fix — close the class, not the two fields

The dialogs are now **presence-only**. Each takes a single discriminated `KavalPresence`/`PadiPresence` — folded at the call site on this host's liveness — as its **sole** daemon input. There is no raw `status`/`link`/`identity` at the render site, so `contractVersion`/`socketPath`/build/uptime/lifetime exist **only on the `connected` arm** (which the fold yields only over a live link). *"Render a connected-era fact while not live"* is now a **type error**, not a review catch.

Along the way the gate's adversaries surfaced three more leaks of the same class — all folded in:

| surface | leak | fix |
|---|---|---|
| Kaval dialog | contract badge + socket read raw `status` | presence-only; facts only on `connected` arm |
| `convergePending()` | frozen `boundHostKavals()` unfloored | floored on `boundHostInventoryLive()` |
| chip tooltips | `kavalVersion()` / `padiVersion()` read raw version | floored **at the reader** (undefined when not live) |
| `KoluInfoDialog` memory | `serverRssBytes()` unfloored | floored at the reader (closes every consumer at once) |
| "Running daemons" copy | a bare `live` boolean guessed "connecting" for a **hard** ssh failure | discriminated `DaemonScan` cause — a total fold over the host's `EntryState` × frame presence; a `failed` host names its real reason (reusing `HOST_DOWN_COPY` titles) |

To keep the dot + status word lossless without a raw `state`, `KavalPresence`/`PadiPresence` gain an `unknown` arm (dead channel / no value → grey) distinct from `warming` (a live link coming up), plus pure `dotFor`/`labelFor` projections. `hostInventoryLive` is subsumed by `daemonScanCause` and removed.

## Design philosophy

- **Fail-fast, no fallbacks:** the leak is closed by making the illegal state *unrepresentable* (P4) and forcing every field through the liveness fold (P5) — no override, no graceful-degradation path. Facts are read off the folded `connected` arm or not at all.
- **Reuse the source of truth:** the failed-host copy reuses the existing `HOST_DOWN_COPY` titles (the Skew-UX host-down card's map), not a parallel string; `boundHostInventoryLive` now derives from the single `boundHostScan` fold.
- **Electricity:** no new domain-agnostic receptacle — the changes are leaf presentation folds in the client.

## Tests

- The #1793 reproduction (committed RED first, `dad484c47`) turns **green**: rendering each dialog through the real fold with a stale status + `live=false` leaks no contract badge / socket.
- New pure tests pin the presence projections (`dotFor`/`labelFor`, the `unknown` arm) and the `daemonScanCause` fold (totality over `EntryState`, the `failed`-cause copy).
- Full client suite green (743). `type` pins (`*.test-d.ts`) assert the fact-bearing fields are unspellable off a non-connected arm.

Client-only — no `@kolu/surface*` type change.


srid:

✅ **Consensus** after 3 thread-round(s) · lowy + hickey · base `f8f3c01bd793`

Settled: 5 unopposed · 2 debated

### Applied (6)
- `lowy-1` ≡ `hickey-3` Same #1793 volatility floored at two strengths; one identity cell floored in three places — commit `760051cbc`
- `lowy-2` ≡ `hickey-1` Dot tone projected twice, each copy re-spelling the liveness floor — commit `d0a11179f`
- `lowy-3` daemonScanCause has two adjacent transposable booleans — commit `b92153216`
- `hickey-2` dot and label are two parallel exhaustive matches over the same presence arms — commit `b300b62ca`
- `hickey-4` The status word "unknown" spelled as a bare literal across four sites — commit `6a1d6636c`
- `hickey-5` PadiEntry is a third inline spelling of the padi map entry-state type — commit `493b74e85`

### Agreed — no change (1)
- `lowy-4` Three parallel presence projections share one floor sub-axis (packages/client/src/kaval/daemonPresentation.ts:397)

srid:
## Codex ⇄ Claude debate

✅ Consensus in 3 round(s) · reviewer effort: xhigh · base: `f8f3c01bd793`

|---|---|---|
| 1 | 8c99f3b76 | fix(codex): floor the three retained-value leaks codex found outside the presence gate (F1-F4) |
| 2 | fad204918 | fix(codex): correct the host-scoping GATE row stale-ified by the F1 floor (F4) |

- **F1** — `padiMemoryDisplay()` delegates straight to `displayRss()`, whose only floor is `daemonTransportLive()`
- **F2** — The convergence banner still reads `boundPadiConvergence()` directly instead of the convergence already carried by `PadiPresence`'s `connected.identity` arm
- **F3** — Only the bound-host rows are guarded by `DaemonScan`; the separate `localScanRows` block is rendered whenever the retained `boundHost` is truthy
- **F4** — The new presence-only API leaves the primary host-scoping documentation stale: this header still says the dialog reads `props.status`, `PadiInfoDialog.tsx:30` still lists `props.identity`, and `useDaemonStatus.ts:223` still describes a `props.link` gate; that module also links to nonexistent `dotForKavalPresence`/`dotForPadiPresence` names

srid:

Ran Pass 1 (rules) + Pass 2 (fact-check) over the diff, `--no-elegance` (the simplify stage already ran).

| Pass | Found | Action |
|---|---|---|
| Rules | 2 | fixed (`6d21c1c72`) |
| Fact-check | 0 | clean — liveness/boundary conditions verified correct |

**Fixed:**
- **no-thin-wrapper-functions** — `daemonStateAttr`/`padiLinkAttr` each had one call site and forwarded verbatim to `presenceState`/`presenceLink`; deleted the wrappers, inlined the folds at their sole JSX site.
- **prefer-ts-pattern** — `toKavalPresence`/`toPadiPresence` dispatched on a string-literal union via an `if`-cascade with an implicit final branch. `toPadiPresence`'s cascade had a **real gap**: a future `PadiLink` variant would silently fall into the `connected` branch. Both now `match(...).with(...).exhaustive()`.

tsc + 738 client tests + biome lint/format green.

srid:
## Evidence

The bug is invisible in normal operation — it only surfaces when a host's channel goes **dead** (its ssh link times out, so the daemon status freezes at its last connected read). This is the Kaval info dialog for an unreachable host `zest`, rendered from the **real `KavalInfoDialog` DOM** with a stale-but-truthy status folded at `live=false` — BEFORE on master, AFTER on HEAD:

#1793 dialog before/after

- **BEFORE** — half-honest: the status pill correctly reads **unknown** and build/memory/lifetime blank to "—", but the header **"contract v5.2"** badge and the **socket path** keep painting their stale last-good values.
- **AFTER** — fully honest: no contract badge, socket reads **unavailable**. Every field now reads off the folded `connected` presence arm, which a dead channel never yields — so "render a connected-era fact while not live" is a **type error**, not a review catch.

**Reproduction (in-PR):** `packages/client/src/host/dialogLiveness.test.tsx` drives each dialog (Kaval + Padi) through the same fold and asserts no fact survives — committed **RED** as the first commit (`dad484c47`), **GREEN** after the presence-only fix. The codex-debate round then caught three more readers of the same class (Padi RSS, the convergence banner, the local-machine scan) — all floored, all covered.

_Layout note:_ this is a content change (which fields render), not hit-target sizing — the dialog body renders identically on the small-viewport/coarse-pointer layout, so no separate mobile capture is warranted.

srid:

✅ **Consensus** with zero debate turns · lowy + hickey · base `f6d5b661a18f`

Settled: 1 reconciled (raised by both lenses) · 3 auto-settled minor · 1 unopposed

### Applied (3)
- `lowy-1` KoluInfoDialog has no fact-gate receptacle — the #1793 leak is spellable again, per-site — commit `74191db64`
- `lowy-2` ≡ `hickey-1` Command palette re-folds the active kaval presence from raw wires instead of a shared accessor — commit `e0c68667f`
- `hickey-2` restart-vs-renew decision encoded twice — the down-arm table lookup is now provably always-true — commit `a77d29859`

### Agreed — no change (2)
- `lowy-3` Two flooring strategies for readers off the one daemonInventory cell (packages/client/src/ui/useDaemonInventory.ts:72)
- `hickey-3` 'incompatible' fact split across two parallel sums read at two inputs in one action slot (packages/client/src/kaval/KavalInfoDialog.tsx:312-315)

srid:
## Codex ⇄ Claude debate (affordance axis)

✅ Consensus in 2 round(s) · reviewer effort: xhigh · base: `f6d5b661a`

|---|---|---|
| 1 | 3b02d3c0d | fix(codex): close two more same-class leaks the sweep missed — daemonConnected + the Kolu rail tooltip/stale surfaces |

- **F1** — `daemonConnected()` repeats the exact negative-space bug this diff removes from `offerRestartVerb`: when the channel is dead but the retained daemon-status subscription is no longer pending, `daemonWarming()` is false and `downState()` is undefined because both are liveness-floored, so the conjunction returns true
- **F2** — The new `KoluInfoDialog` gate leaves the same retained `pwa.server()` version and commit visible in the always-reachable Kolu tooltip while `daemonLive()` is false
- **F3** — `activeKavalPresence` is documented as the one value read by both the command palette and KavalInfoDialog, but only the palette imports it

srid:
## [👮 Code-police](https://agency.srid.ca/) — affordance axis

Pass 1 (rules) + Pass 2 (fact-check) over the affordance delta, `--no-elegance` (simplify already ran).

| Pass | Found | Action |
|---|---|---|
| Rules | 1 | fixed (`6fc54499e`) |
| Fact-check | 1 | fixed (`6fc54499e`) |

**Fixed:**
- **fact-check (real regression)** — `clientStale = daemonTransportLive() && useSurfaceApp().stale()` short-circuits, so over a dead transport `useSurfaceApp()` is never called and its required "must be used within `<SurfaceAppProvider>`" throw is **silently skipped** — fail-fast → fail-silent exactly in the disconnected state. Now reads the model first, then floors.
- **dry-rule-of-three** — `koluTip` re-AND'd `daemonLive()` at three sites; folded the server version/commit reads into one local `liveServer()` (matching the sibling `KoluInfoDialog`).

tsc + 741 client tests + biome lint/format green. (Every other rule + the fact-check traced clean, incl. the palette↔dialog presence-fold dedup down to `keyedSubscriptionCache`.)

srid:
## Evidence — updated (the affordance axis)

srid found the class had a second axis: the fact fix made stale FACTS unspellable while not-live, but **ACTIONS** were never folded — the "Restart kaval" button was **enabled over a dead channel**, offering an action the channel can't carry out. Same Kaval dialog for the unreachable host `zest`, now showing both axes:

#1793 affordance axis before/after

- **Facts** (as before): no contract badge, socket reads **unavailable**.
- **Affordances** (new): the "Restart kaval" button + its "captures the session" promise are gone — gated by `offerRestartVerb(presence)`, a **total function of the presence sum** (`unknown`→no verb), mirroring the DegradedCanvas. On a dead channel the dialog offers **no action it can't carry out**.

**Reproduce-first:** the affordance leak was committed RED (`1d0097908`) then turned green. The gauntlet then swept the class across every surface it had relocated to — the Padi degraded-bind version tooltip, Padi RSS, the convergence banner, the local-machine scan, the `daemonConnected` lifecycle-**authority** gate (client asserting authority over a dead channel → reconcile writes), and the always-visible Kolu rail **tooltip** + stale-build badges + mobile reload affordance — all floored at their canonical reader/gate.

PR #1894, opened 2026-07-19
fix(client): chime on awaiting_user landing over an already-waiting row (#1177)

### LINKED ISSUE #1177: awaiting_user doesn't always chime: alert logic buckets it with waiting
## Symptom

A Claude `AskUserQuestion` (and now tool-permission gates, via the screen scrape in #1160 / #1166 / #1176) reliably shows the **awaiting pip on the Dock**, but **only *sometimes* fires the sound / OS notification**. The dock dot is consistent; the chime is intermittent.

## Root cause

The two are driven by different paths:

- The **dock pip** reads `agent.state` directly — `awaiting_user` → awaiting pip, always.
- The **sound + OS notification** (`useTerminalAlerts.ts` → `useActivityAlerts.ts`) only fire on a transition that *enters* the attention class `{waiting, awaiting_user}` from **outside** it:

```js
// useTerminalAlerts.ts
if (!isAttentionState(next) || isAttentionState(prev)) return;
```

`waiting` and `awaiting_user` are deliberately one class (so an agent flipping between them mid-episode doesn't double-chime). The consequence: whether a prompt chimes depends on the state **immediately before** the scrape promotes it:

- `thinking → awaiting_user` / `tool_use → awaiting_user` → crosses into the class → **chimes** ✅
- `waiting → awaiting_user` → already in the class → **suppressed** 🔇

### Why it's intermittent

A pending `AskUserQuestion` usually reads as `thinking` (the user's prompt is the newest JSONL entry; the assistant reply is buffered). But if the ~1s screen-scrape poll catches the prompt on screen **while the JSONL still reads the prior `end_turn` as `waiting`** (the user-prompt entry hasn't flushed yet, or Claude was idle-at-`waiting` when it asked), the client observes `waiting → awaiting_user` and the alert is swallowed. Same prompt, different prior state by timing → "sometimes".

(Independent, by-design gates can also silence it and are *not* the bug here: no sound for the focused tile on a visible tab; OS notification needs the tab hidden + permission; browser autoplay can drop `audio.play()`; `activityAlerts` must be on.)

## The gap

`awaiting_user` is now a **distinct, more-actionable** state than plain `waiting` (a real human gate vs. "turn ended"), but the alert policy still treats them as interchangeable — so a genuine question/permission prompt that lands over an already-`waiting` row gets the badge but not the chime.

## Proposed fix (not yet implemented)

Alert on entry into `awaiting_user` **specifically**, even from `waiting` — while still not double-chiming on `waiting`-only churn. Roughly: treat a `* → awaiting_user` transition as alert-worthy when `prev !== "awaiting_user"`, in addition to the existing class-entry rule for `waiting`. Needs care to avoid re-alerting if the scrape briefly flips `awaiting_user → waiting → awaiting_user` during the poll/JSONL settle.

## Refs

- `packages/client/src/terminal/useTerminalAlerts.ts` (`checkAgentFinished`)
- `packages/client/src/terminal/useActivityAlerts.ts` (`fireActivityAlert` / `playSound`)
- `packages/client/src/terminal/agentState.ts` (`isAttentionState`)
- Screen-scrape detection that surfaced this: #1160, #1166, #1176 (#905)

srid:
Fixes #1177 — `awaiting_user` didn't always chime.

## The bug

`waiting` and `awaiting_user` share one `alertClass` (`"notify"`), and the old `checkAgentFinished` fired **only on entry into** that class:

```ts
if (!notifies(next) || notifies(prev)) return;   // prev already notify ⇒ swallowed
```

So a genuine human gate (AskUserQuestion / permission prompt) landing over an already-`waiting` row (`waiting → awaiting_user`) was an intra-class move: the dock pip lit (it reads `agent.state` directly), but the sound/OS notification was swallowed. Intermittency was timing — whether the ~1s scrape caught the prompt before or after the JSONL settled the prior turn to `waiting`.

Per #1690 the inherited diagnosis was reproduced first (committed RED pin, `952ff1564`) — `fired` observed 0× on `waiting → awaiting_user` where it must be 1× — then ratified before any fix code.

## The fix

Fire on **class-entry OR an escalation into the `awaiting` bucket**, deduped by a per-terminal **attention-episode latch** that resets **only on a work state** — never on `waiting`.

The work-state reset is the load-bearing discriminator: a genuinely new gate *always* passes through a work state (user re-engaged → agent worked → asked); scrape/JSONL settle jitter *never* does. A memoryless `(prev, next)` rule provably cannot do this — rows 2 and 3 below end in the identical transition with opposite required outputs — so the 1-bit latch is minimal, not incidental.

Escalation is spelled through the fenced `agentBucket` fold (`agentBucket(next) === "awaiting" && agentBucket(prev) !== "awaiting"`), never a raw `awaiting_user` literal, so a future attention-flavored state forces a decision at the fence.

### Behavior matrix (a1, alerts on)

| Scenario | Before | After |
|---|---|---|
| `thinking → awaiting_user` (genuine) | 1 ✅ | 1 ✅ |
| `waiting → awaiting_user` (gate over waiting row) — **the bug** | 0 🔇 | 1 ✅ |
| `awaiting_user → waiting → awaiting_user` (flap) | 1 | 1 ✅ |
| `thinking → waiting → awaiting_user` (settle jitter, one gate) | 1 | 1 ✅ |
| `thinking → waiting → thinking → awaiting_user` (finish, re-engage, ask) | 2 | 2 ✅ |
| `waiting`-only churn / effect re-run, row stays `waiting` | ≤1 | ≤1 ✅ |
| steady-state `awaiting` re-tracked (host-switch-back / reshuffle) | 0 | 0 ✅ |

## Architecture

- **Leaf, not a boundary.** The latch is client-only alert bookkeeping over a bounded algorithm (no transport/persistence/reconnect), so it lives beside the transition diff in `useTerminalAlerts` — never folded into the pure `agentProjection` vocabulary.
- **Host-scoped like the diff.** Cleared on a host change (a latch can't cross to a session-imported id living on another host) and pruned to the tracked set each run (a departed id can't leave a ghost latch a same-host id reuse — drain/restore, sleep-wake re-seed — would inherit).
- **Bookkeeping unconditional, emission gated.** The episode reset runs even while alerts are off, so toggling the pref off/on can't freeze the latch and reintroduce the bug.
- **First-sighting rule.** A first sighting already in the `awaiting` bucket is pre-latched (suppresses a settle-flap phantom); a first sighting at `waiting` is left unlatched (a genuine gate over it still fires).

## Honest copy

`useActivityAlerts` titled every alert `"… finished"`. Since this fix deliberately widens the non-finished chimes, an `awaiting_user` chime now titles `"… needs your input"`; a plain turn-end stays `"finished"`.

## Tests

Flips the RED pin green and adds: settle-once, re-engage-twice, steady-state re-track, preference-toggle staleness (C1), remove/re-add ghost-latch (C2), first-sighting pre-latch (C3), cross-host must-fire (C6c). 16 tests in `useTerminalAlerts.test.ts`; 145 in `packages/client/src/terminal/` all green; `typecheck` clean.

## Recorded (no action here)

- The cross-host sibling path (`useHostAttention` set-diff banner) has no episode memory — a different trigger, no #1177 bug there, bounded by tag-replace dedup.
- Residual: two gates inside one poll tick are sensor-invisible (`sensors.ts` holds `awaiting_user` continuously), so the "two distinct gates with no work state between" residual is bounded by poll cadence.


srid:

✅ **Consensus** with zero debate turns · lowy + hickey · base `5faa6dbfa55d`

Independent findings: lowy=2, hickey=5
Settled: 1 reconciled (raised by both lenses) · 1 auto-settled minor · 4 unopposed

### Applied (2)
- `lowy-1` ≡ `hickey-1` `checkAgentFinished` name lags its widened responsibility — commit `8d6d9671e`
- `hickey-2` `isAwaiting(next)` computed twice in the same branch — commit `f819ac67d`

### Agreed — no change (4)
- `lowy-2` `fireActivityAlert(awaiting: boolean)` folds the attention bucket to a bit at the boundary (packages/client/src/terminal/useActivityAlerts.ts:57)
- `hickey-3` Episode-latch mutations spread across the effect body and the check function (packages/client/src/terminal/useTerminalAlerts.ts:94)
- `hickey-4` `activeArm(getMetadata(id))?.agent?.state` read expression duplicated (packages/client/src/terminal/useTerminalAlerts.ts:215)
- `hickey-5` `notifies` and `isAwaiting` share an identical predicate shape (packages/client/src/terminal/useTerminalAlerts.ts:142)

srid:
## Codex ⇄ Claude debate

✅ Consensus in 4 round(s) · reviewer effort: xhigh · base: `5faa6dbfa55d`

|---|---|---|
| 1 | e052dd146 | fix(client): latch the alert episode on delivery, not on decision (codex F1) |
| 2 | ec612bdf2 | fix(client): latch a live gate seen while watched, not only on delivery (codex F2) |
| 3 | 287d3e143 | fix(client): the eyes channel latches a watched gate regardless of the pref (codex F2) |

- **F1** — `chimed.add(id)` runs before `alertForTerminal` applies the active-and-focused suppression at lines 183-203
- **F2** — The delivery condition is now too broad for a suppressed live gate

srid:

Ran **Pass 1 (rules checklist)** and **Pass 2 (fact-check)** over the diff vs `5faa6dbfa55d`. `--no-elegance` (the `/simplify` track already ran this gauntlet).

**Pass 1 — rules:** ✅ no findings. Every applicable rule checked with a concrete absence-check (fail-fast / no silent error-swallow / no overloaded-null / no unbounded growth — `chimed` is pruned to the tracked set every tick and cleared on host change / no dead code after the `checkAgentFinished → advanceAlertEpisode` rename / no new untyped escape hatch).

**Pass 2 — fact-check:** one finding, **fixed**.
- **Docs over-claimed the notification label** — the notifications doc and changelog asserted a live gate reads *"needs your input"* unconditionally, but in the `thinking → waiting → awaiting_user` settle-jitter path the single chime fires on the `waiting` class-entry and carries *"finished"* copy (the recorded known-minor). Softened both texts to disclose the label is the common case, not a guarantee, while keeping the real promise (the gate reliably rings). — `8b3e78a87`

The state machine (episode latch, work-state reset, class-entry/escalation, eyes-vs-delivery latch, host-scoped clear/prune) was traced against every transition sequence and cross-checked against all 19 tests — no logic defect.

srid:
## Evidence

**Why this is test-evidence, not a video.** #1177's two observable channels are a **sound** and an **OS notification banner** — both fire outside the browser viewport (the banner is an OS-level `showNotification`, the sound is audio), so a Playwright viewport recording can't capture either. And the trigger is a **scrape/JSONL settle race**: the alert was swallowed only when the ~1 s screen-scrape caught the prompt while the JSONL still read the prior turn as `waiting` (a `waiting → awaiting_user` transition). The e2e harness has no way to deterministically inject that cross-poll timing. That is exactly why the bug was reproduced and the fix pinned at the **unit layer**, where the state sequence is injected directly (per the reproduce-first discipline in #1690). The proof below is the deterministic reproduction flipped green plus the full guarded matrix.

### The reproduction: RED → GREEN

The bug was first committed as a failing `it.fails` pin (`952ff1564`) that drove the exact transition through the live effect and observed the alert fire **0 times** where it must be 1:

```
AssertionError: expected "vi.fn()" to be called 1 times, but got 0 times
  ❯ src/terminal/useTerminalAlerts.test.ts   (waiting → awaiting_user)
```

The fix flips it green — `fires on waiting → awaiting_user (a gate over an already-waiting row)` now passes.

### The full guarded matrix — 19/19 green

```
✓ #1177 awaiting_user chime > fires on waiting → awaiting_user (a gate over an already-waiting row)   ← the bug, fixed
✓ GREEN control: fires on thinking → awaiting_user (today's chime, must not regress)
✓ GREEN control: a waiting row does NOT re-fire when the effect re-runs for a sibling
✓ GREEN control: an awaiting_user → waiting → awaiting_user flap chimes exactly ONCE
✓ GREEN control: settle jitter thinking → waiting → awaiting_user chimes exactly ONCE
✓ GREEN control: finish then RE-ENGAGE (thinking → waiting → thinking → awaiting_user) chimes TWICE
✓ GREEN control: a steady-state awaiting terminal re-tracked after a host-switch-back does NOT re-chime
✓ C1: a genuine gate after alerts are toggled off then on still fires (latch not frozen)
✓ C2: an id removed then re-added at waiting is not swallowed by a ghost latch
✓ C3: an id first-sighted AT awaiting_user then settle-flapped does NOT chime a phantom
✓ C6c: a latched id, after a host switch, fires once on host B's own thinking → awaiting_user
✓ codex-F1: a finish SUPPRESSED because the user is watching does NOT latch — a gate after they look away still chimes
✓ codex-F2: a live gate SEEN while actively watched latches — a later settle-flap does NOT phantom-chime
✓ codex-F2b: a live gate SEEN with alerts OFF latches — enabling alerts + look-away does NOT phantom-chime
  … + 5 pre-existing host-scoped transition-diff regression tests
Test Files  1 passed (1)
     Tests  19 passed (19)
```

Each test injects a concrete state sequence and asserts the exact alert count + copy + unread mark — the `codex-F1/F2/F2b` cases were added during the codex ⇄ Claude debate and pin the visibility-gate and off→on-toggle edges. The behavior is deterministic and fully covered; CI green (both platforms) is linked on this PR.

PR #1895, opened 2026-07-20
fix(anyagent): stop recent-agents MRU leaking the launch prompt

srid:
## What

Recent agents in the command palette were showing the **entire launch command including the prompt** — the exact text the agent was spawned with leaked into the MRU list:

> `claude --dangerously-skip-permissions 'You are BOOT1. Read BRIEF-BOOT1.md in this directory and follow it exactly.'`

## Why

The stored `command` for each recent agent is produced by `normalizeAgentInvocation` (`packages/integrations/anyagent/src/agent-cli.ts`), whose job is to strip ephemeral prompt text so it doesn't pollute the MRU. It kept only allowlisted stable flags — but it attached the token following **any** kept stable flag as that flag's value:

```ts
kept.push(t);
if (next !== undefined && !next.startsWith("-")) { kept.push(next); i++; }
```

`--dangerously-skip-permissions` is a **boolean** switch with no value, so the trailing prompt positional got greedily attached as its "value" and survived normalization. The same exposure applied to every boolean stable flag (`--yolo`, `--bare`, `--full-auto`, `--no-plan`, …). A prompt on a value-taking flag (`-p`, `--model`) or a bare trailing positional was already dropped correctly; only *boolean-flag-immediately-followed-by-prompt* leaked.

## Fix

Add a `VALUE_FLAGS` allowlist of the stable flags that genuinely take a value. Only those consume the next token; every other allowlisted flag is treated as boolean and never swallows its follower, so a trailing prompt drops as the positional it is.

This mirrors the existing `STABLE_FLAGS` design — an **allowlist, not a denylist** — so a new boolean flag added upstream can never reintroduce the leak.

```
claude --dangerously-skip-permissions 'You are BOOT1. …'
  →  claude --dangerously-skip-permissions
```

## Tests

New regression tests assert the prompt drops for boolean flags across claude/opencode/codex/grok (both the shell-mark and the `#1872` command-rooted seed dialect), and that a boolean flag preceding a real valued flag still keeps that value. All 88 tests in `agent-cli.test.ts` pass.


srid:

✅ **Consensus** after 2 thread-round(s) · lowy + hickey · base `5faa6dbfa55d`

Settled: 2 debated

### Applied (2)
- `lowy-1` ≡ `hickey-1` Flag arity split from flag membership — one axis (the agent-CLI flag vocabulary) now sprayed across two allowlists that silently drift — commit `0795e8c6b`
- `lowy-2` ≡ `hickey-2` VALUE_FLAGS doc-comment cites an unrelated issue number — (uncommitted)

srid:
## Codex ⇄ Claude debate

✅ Consensus in 1 round(s) · reviewer effort: xhigh · base: origin/master

|---|---|---|
| 1 | c7fd854a2 | fix(codex): correct stale VALUE_FLAGS reference in the leak regression comment |

- **F1** — The regression comment says the old behavior existed before a `VALUE_FLAGS` allowlist, but this diff has no `VALUE_FLAGS`; the chosen design records arity directly in `STABLE_FLAGS`

srid:

Ran the rules and fact-check passes over `5faa6dbfa…HEAD` (`--no-elegance`, since `/simplify` already ran this gauntlet).

| Pass | Findings |
|---|---|
| Rules | 0 — clean |
| Fact-check | 0 — clean |

**All clear.** `no-overloaded-null` / `invalid-states-unrepresentable` register as *positive* instances: the change replaces an implicit "does the next token look like a flag" heuristic with an explicit, co-located `FlagArity` discriminant. Fact-check independently verified every flag's arity against real CLI semantics for all seven agents (e.g. codex `-c`/`-p`/`-s`/`-a` → `value`; `--yolo`/`--full-auto`/`--oss` → `boolean`) — no misassignment. 88/88 unit tests pass.

PR #1896, opened 2026-07-20
Remember the active host across PWA relaunches

srid:
`kolu-active-host` was the **only** client pref backed by `sessionStorage` — every other pref rides `localStorage`, and durable facts live server-side. That's the right default for a browser *tab* (two tabs view two hosts independently; a fresh tab starts local), but wrong for an **installed PWA**: a standalone window starts a fresh browsing session on every launch, so `sessionStorage` is empty and the app **silently reverts to `LOCAL_HOST`** — amnesia about which host you were on, on every relaunch. The server-side session restores your workspace, but not *which host* views it.

The fix chooses the storage backend **once at boot, from whether the window is installed** — a structural fact of how the app was launched, not a user knob (no config, no override, no degrade path, per the fail-fast convention):

| Launch context | Backend | Why |
| --- | --- | --- |
| Installed / standalone PWA | `localStorage` | survives relaunch — no sibling tabs to isolate |
| Regular browser tab | `sessionStorage` | unchanged — per-tab isolation, fresh tab starts local |

### Shape

"Installed?" is **not** re-derived here — it delegates to `@kolu/surface-app`'s already-exported, unit-tested `isInstalledFromEnv` (the same fact `useSurfaceApp().isInstalled()` reads reactively elsewhere in the client). The only genuinely-new logic is the `installed ? localStorage : sessionStorage` mapping:

```
window ─▶ readInstallEnv() ─▶ isInstalledFromEnv(env) ─▶ activeHostStorage(installed, {local, session}) ─▶ Storage
   (surface-app InstallEnv)      (surface-app kernel)              (this module: the mapping)                  │
                                                    activeHostPref(storage) ◀────────────────────────────────┘
                                                    (shared pref contract: wire.ts + tests)
```

- `activeHostStorage(installed: boolean, backends)` is the pure, unit-pinnable seam; `activeHostStorageForLaunch(backends)` is the thin boot glue `wire.ts` calls at module init (before the `<SurfaceAppProvider>` owner exists, so the reactive `isInstalled()` isn't reachable and a synchronous read is required).
- `activeHostPref(storage)` owns the pref **contract** (key, default, canonical wire codec, corrupt-value warn) in one place, consumed by `wire.ts` **and** the boot-survival tests — so the flagship "survives a relaunch" test exercises the *real* contract, not a stale copy.
- `readInstallEnv` deliberately *mirrors* surface-app's private DOM reader rather than exporting it: exporting would make this a `@kolu/surface-app` public-API change (a drishti pair-PR + odu verdict), disproportionate to a client-local storage tweak. It matches surface-app's three installed display-modes so the two "installed?" answers agree on the browser path.

`persistedPref` is untouched — it already accepts an injectable `storage`.

### Notes for reviewers

- **Boot ordering is unchanged.** `groundActiveHost` guards a boot race where the active host is restored a tick before the membership snapshot lands. The `localStorage` arm reads *just as synchronously* as the old `sessionStorage` read, so that ordering is untouched (verified by the architecture-review state-and-time check).
- **`localStorage.clear()` now also clears the remembered host** in standalone mode (the Debug → "clear local storage" command in `App.tsx`). That's correct — a deliberate reset — and worth noting since the pref moved backends.

### Tests

`activeHostStorage.test.ts` (9 cases, happy-dom): the pure backend pick; the **live-DOM** reader `activeHostStorageForLaunch` across standalone / minimal-ui / fullscreen / legacy-iOS / regular-tab (each display-mode arm pinned independently); and the **survive-a-standalone-relaunch** case — empty `sessionStorage` + a remembered `localStorage` boots on the remembered host over the *real* `activeHostPref` contract. See the Evidence comment for why the full installed-PWA-relaunch round-trip isn't automatable, and the CI-green proof.

### Review gauntlet

Full `/be` gauntlet ran and is reflected in the commits: architecture-first-principles (drove the surface-app reuse), lowy⇄hickey lens debate (consensus — boolean picker + shared `fakeStorage`), codex⇄Claude debate (consensus in 3 rounds — the shared pref factory, live-DOM tests, per-arm pinning), `/simplify`, and code-police. CI green on `x86_64-linux` + `aarch64-darwin`.

### Try it locally

```sh
nix run github:juspay/kolu/active-host-storage
```


srid:

✅ **Consensus** after 1 thread-round(s) · lowy + hickey · base `5faa6dbfa55d`

Settled: 1 auto-settled minor · 3 unopposed · 1 debated

### Applied (2)
- `hickey-2` activeHostStorage takes the full InstallEnv, complecting the install-verdict with the new backend mapping — commit `9e9b4c4ef`
- `hickey-3` fakeStorage in-memory Storage fake is now triplicated across client test files — commit `d62f6e60f`

### Agreed — no change (3)
- `lowy-1` ≡ `hickey-1` readInstallEnv duplicates surface-app's DOM install-read — the wires are exposed past the receptacle (packages/client/src/host/activeHostStorage.ts:37)
- `lowy-2` The mirror already diverges from its source — it drops surface-app's `typeof window` guard (packages/client/src/host/activeHostStorage.ts:39)
- `lowy-3` `activeHostStorage` (the pure selector) reads as a noun, not a decision (packages/client/src/host/activeHostStorage.ts:57)

srid:
## Codex ⇄ Claude debate

✅ Consensus in 3 round(s) · reviewer effort: xhigh · base: `origin/master`

|---|---|---|
| 1 | 9ef49a9bd | fix(codex): share the active-host pref contract, test the live DOM reader, fix stale boot comments |
| 2 | 08e43036b | test(codex): pin each installed display-mode arm independently (F2) |

- **F1** — The active-host preference contract (key, fallback, parser, and serializer) is copied in both new integration tests and again in wire.ts
- **F2** — Every new test passes an already-derived boolean to activeHostStorage; none calls activeHostStorageForLaunch
- **F3** — The production boot contract still says activeHost is restored from sessionStorage, which is now false for installed launches
- **F4** — The changed import block does not satisfy Biome ordering: hostChipTone.ts must sort before hostReconcile.ts

srid:

Scope: `git diff <merge-base>...HEAD` (PWA-aware active-host storage). Passes: **rules** + **fact-check** (elegance skipped — `/simplify` already ran this tree).

**Fixed (1)**
- **`activeHostPref.ts` — stale log tag** (fact-check): the corrupt-value `onInvalid` diagnostic still read `[wire]` after the pref construction moved into the shared `activeHostPref.ts` module — a second caller would have logged a warning misattributed to `wire.ts`. Retagged to `[activeHostPref]`. Commit `b0fcbf1eb`.

**Reviewed, no change (adjudicated)**
- **`caught-error-must-not-collapse-to-empty` — `activeHostPref.ts` `onInvalid`** (rules): flagged because a corrupt stored host resets to `LOCAL_HOST` surfaced only via `console.warn`. **Kept as-is, deliberately:** this pattern is *pre-existing* — the diff relocated `wire.ts`'s `activeHost` pref `onInvalid` verbatim, it did not introduce it — and the warn level is a documented calibration matching the sibling `reattachAnnouncedAt` pref (a benign, recoverable, app-written-data reset). The error *is* surfaced (raw value + error), not silently collapsed to empty; the same pass's `errors-must-log-at-error` check independently confirms `warn` is correct here. Escalating to a toast would break pref-family consistency and expand a storage-backend change into re-litigating a ratified UX calibration.

All other rule checks passed; no silent error-swallowing, inaccurate fallback, or logic defect found in the diff.

srid:
## Evidence

**What the fix does (zero visual diff, behavioral):** the `kolu-active-host` pref now chooses its storage backend from the launch context at boot — an installed / standalone PWA uses `localStorage` (survives relaunch), a regular browser tab keeps `sessionStorage` (per-tab). Previously it was always `sessionStorage`, so every PWA relaunch (a fresh browsing session) started empty and the app silently reverted to `LOCAL_HOST`.

### Why there is no video/screenshot of the full round-trip

The full user-visible effect — *"an installed PWA reopens on the remote host you were viewing"* — requires three things at once that **no automated harness available here can stage**:

1. a **standalone display-mode** context (an actually-installed PWA) — Playwright/Cucumber run in a normal browser **tab**, and the chrome-devtools MCP's `emulate` has **no display-mode** control, so neither can make the app boot as "installed";
2. a real **remote host** in the pool (so the active host is something other than the default `local`) — the dev server **cannot bind a remote padi** (per the `remote-host-testing` harness notes);
3. a genuine **relaunch** — a fresh browsing session that wipes `sessionStorage`, not a reload (a reload preserves `sessionStorage`, so it wouldn't even exercise the bug).

That combination is a **manual real-device scenario** (install kolu to a home screen, connect a remote host, fully quit and reopen). It is do.md's explicit "behavior genuinely can't be observed [in the harness]" case — so the proof is the layer that *is* observable: the launch-context → backend decision and the boot read, pinned in a real DOM.

### The observable mechanism is pinned in a real-DOM (happy-dom) test

`ci::unit` (green on **both** platforms — see CI below) runs `activeHostStorage.test.ts`, which drives the **actual** live-DOM reader (`activeHostStorageForLaunch` → real `window.matchMedia` / `navigator.standalone` → surface-app's `isInstalledFromEnv`) with the browser signals stubbed per case — the same signals a live capture would flip:

```
✓ activeHostStorage — pick the backend from the launch context
    ✓ installed selects localStorage (survives relaunch)
    ✓ regular tab context selects sessionStorage (per-tab isolation)
✓ activeHostStorageForLaunch — read the live install env, then pick   [real happy-dom]
    ✓ installed display-mode (display-mode: standalone)  → localStorage
    ✓ installed display-mode (display-mode: minimal-ui)  → localStorage
    ✓ installed display-mode (display-mode: fullscreen)  → localStorage
    ✓ legacy iOS navigator.standalone                    → localStorage
    ✓ regular tab (no signal)                            → sessionStorage
✓ active-host pref survives a standalone relaunch
    ✓ installed boot with empty sessionStorage + remembered localStorage → remembered host
    ✓ regular-tab boot ignores localStorage and starts from the per-tab default
  9 passed
```

The last block is the fix's whole point, wired over the **real** `persistedPref` contract (`activeHostPref`, the same one production uses): a simulated standalone relaunch — a fresh **empty** `sessionStorage` fake beside a **populated** `localStorage` fake — boots on the **remembered** host, not `LOCAL_HOST`. The regular-tab twin confirms per-tab isolation is unchanged.

### CI — green on both platforms

`odu` run `c2fc2be6e#1`, **passed**, both lanes:
- `x86_64-linux` (kolu-ci-1) — all 14 nodes ✓ (incl. `ci::unit`, `ci::e2e`, `ci::nix`, `ci::fmt`, `ci::biome`)
- `aarch64-darwin` (sincereintent) — all 14 nodes ✓

No visual regression is possible (no markup/CSS changed — the two `website/` edits are MDX prose); mobile/touch surfaces are untouched.

srid:
## ⚠️ This PR does not fix the reported bug — reproduced the real cause

Deployed and tested against a **real remote host** (isolated `nix run` kolu on a pu box + a real ssh-connected remote, driven in a real browser). The active host **still reverts on refresh** — and the cause is not storage at all.

### What actually happens

The active host on refresh is decided by the **URL**, not the pref:

1. kolu has **no state→URL sync** — switching hosts / focusing a terminal never updates the address bar. So the bar keeps whatever deep link you *last navigated to* (a notification click, a shared/bookmarked `#/t/local/<id>`, an agent focus-link), even after you switch to a remote host.
2. On **every reload**, `useDeepLinks` re-parses `location.hash` and **re-enacts** that link — deliberately (`useDeepLinks.ts:244`: *"a RELOAD still re-routes because the boot parse never reads [the already-routed] gate"*).
3. A terminal route runs `routeToTerminal` → **`setActiveHost(route.host)`**. The stale link says `local`, so you're yanked to `local`, and since `setActiveHost` writes through the pref, your remembered `remote` is **overwritten**. The local list then settles, the terminal is gone, and you see *"That terminal is no longer on 'local' … Staying on that host."*

**The override overwrites whichever backend** — so this PR's `sessionStorage → localStorage` swap changes nothing for it. This PR fixes a real-but-different case (PWA relaunch with an *empty* hash) and was shipped as if it were the reported bug. It wasn't — a misdiagnosis on my part (I never reproduced the original bug before fixing it).

### Reproduced

`#/t/local/<gone-id>` in the bar + active=remote → full refresh → `sessionStorage` flips `remote:…` → `local`, active host → `local`. With an **empty** hash, the same setup persists correctly — confirming the URL route, not the backend, is the differentiator.

### Proposed fix (separate)

The `entryAlreadyRouted` stamp already exists and **survives a reload**. Have the **boot parse honor it**: a *reload* of an already-consumed link does not re-enact it (your restored host stands), while a *fresh* bookmark/shared/notification link (no stamp) still routes. This respects the "URL is a command channel, not a state mirror" design rather than adding a state→URL sync.

### Open question

Keep this PR's storage change (a legitimate improvement for the empty-hash PWA-relaunch case) or revert it as scope-creep on a misdiagnosis? Leaning toward a small follow-up PR for the real `useDeepLinks` fix and re-scoping/holding this one.

_Posted from the reproduction session; a fix PR will reference this._

srid:
Closing per srid's decision — not needed. The originally-reported bug (host reverting on refresh) was fixed by #1903 (the deep-link boot replay was the real cause; see the reproduction comment above). The one case this PR's storage swap would fix — an installed-PWA relaunch with an empty hash starting a fresh browsing session — is not a workflow in use, so the change has no beneficiary. The investigation record stands here if that ever changes: the storage-selection helper delegated to surface-app's canonical isInstalledFromEnv and is resurrectable from this branch.

PR #1897, opened 2026-07-20
fix(client): tile title bar reads live metadata so it can't lag the dock

srid:
## The bug

A canvas tile's **title bar** and its **dock row** show the same facts about a terminal — the PR, the agent's state, the foreground process. They disagreed. In the reported case the dock badge showed PR #1 for a terminal while the title bar above the terminal showed nothing — and it stayed that way until some *unrelated* change nudged it.

## Why

Both surfaces derive PR the same way and read from the same store. They differed only in **which door** they went through:

- The **dock** reads the fact **live** — `getMetadata(id)`, the fine-grained, `reconcile`-backed store proxy. A PR resolves → the dock updates at that leaf.
- The **title bar** (`TerminalMeta`) read it off `getDisplayInfo(id).meta` — a `TerminalMetadata` snapshot bundled into the `displayInfos` memo.

That memo is invalidated **only by git / cwd / membership** (the `terminalKey` inputs) — deliberately, as the "reactivity keystone" perf win (#1425): it must *not* re-run on PR / agent / foreground churn. But it also carried the whole live `meta` in its **output**. So the fast-changing facts rode along in a value the memo never refreshes, and a PR resolving on a settled branch left the header pinned to a stale snapshot.

## The fix — kill the class, not the instance

A memo should hand out **only what it invalidates on**. So `meta` is removed from `TerminalDisplayInfo` entirely — a stale read is now a *type error*, not a latent bug that relocates to the next consumer. `TerminalDisplayInfo` keeps only its decorations: colors, identity key, sub-count.

Every live-fact consumer now reads `getMetadata(id)` at its own leaf, exactly like the dock:

| Consumer | Change |
| --- | --- |
| `TerminalMeta` / `TerminalMetaCompact` | take a live `meta` prop; `App.tsx` + `MobilePullChrome` wire `getMetadata(id)` |
| dock search/bucket model (`dockModel`) | `DockSourceEntry`/`DockEntry` carry `meta`; `buildWorkspaceEntries` takes `getMetadata`; `searchTextFor`/`buildDockModel` read it |
| `WorkspaceGrid`, `dockRowChrome` | read `entry.meta` |
| `CanvasMinimap` | resolves a live `meta` accessor for presence + tooltip |
| `terminalSubject` | takes live `meta`; `useTerminals` passes `getMetadata(id)` |
| `entryBucket` | deleted (dead export) |

This is the reactively-correct shape: the store's fine-grained proxy is the single live source; the display-info memo is left to do only its real job (cross-terminal color/key uniqueness).

## Design philosophy

- **Reuse the existing source of truth.** One live source — `getMetadata(id)` — for every live fact, the same one the dock already used. No parallel snapshot path.
- **Illegal states unrepresentable.** Removing the field makes the stale read *unspellable* (compile error), so the class can't reappear in a new consumer.
- **No fallbacks.** No stale-tolerant compatibility shim; the snapshot field is simply gone.

## Tests

- New guard in `terminalDisplay.test.ts`: the display info can never again carry a live record.
- Existing `dockModel` / `chipInitials` fixtures moved `meta` to the entry level.
- Full client unit suite green (742 tests).

## Docs

- Changelog entry under **Inside a Tile**.
- Performance Atlas note: a new Finding records this as the correctness corollary of the reactivity keystone, so a future change doesn't re-introduce a carried live record.


srid:
## Evidence

**No reliably-reproducible visual difference — stated, not silently skipped.**

This fix changes *which internal accessor* the tile title bar reads for its live facts (PR, agent, foreground): from a snapshot bundled in the `displayInfos` memo, to the live `getMetadata(id)` store proxy — the same source the dock already uses.

For **every state a person can deterministically reach, the before and after render identically.** Pre-fix, ordinary in-place field updates already kept the header fresh; the header only went stale after an **internal proxy-reference turnover** — the store swapping a terminal's whole record object (an `absent → ok` PR arm swap, a sleep/wake recreation) *without* a subsequent `displayInfos` rebuild. No user gesture forces that turnover on demand, so there is no deterministic on-screen before/after to record — and a "happy path" capture of the title bar showing a PR would look the same on `master`, proving nothing.

So the defect is closed **structurally**, not by a staged frame:

- `meta` is removed from `TerminalDisplayInfo` entirely — a stale read of a live fact through the display-info snapshot is now a **compile error**, unspellable in any current or future consumer.
- A regression guard in `terminalDisplay.test.ts` asserts the display info can never again carry a live record.
- Full client unit suite green (742 tests).

The reasoning is recorded as the correctness corollary of the reactivity-keystone finding in the [performance Atlas note](https://kolu.dev/atlas/performance.html).

srid:

✅ **Consensus** after 1 thread-round(s) · lowy + hickey · base `79c05f9165ab`

Settled: 1 reconciled (raised by both lenses) · 2 auto-settled minor · 1 debated

### Applied (1)
- `lowy-1` ≡ `hickey-1` Both-required (info, meta) pairing hand-composed at five sites; the paired accessor already exists as createDockRowData — commit `65c481f4e`

### Agreed — no change (3)
- `lowy-2` terminalSubject gates info and meta asymmetrically (packages/client/src/terminal/terminalSubject.ts:22)
- `lowy-3` ≡ `hickey-2` MobilePullChrome resolves activeId independently for info and meta (packages/client/src/MobilePullChrome.tsx:61)
- `hickey-3` info+meta pairing threaded through every consumer (the split's structural cost) (packages/client/src/canvas/dockModel.ts:24)

srid:
## Codex ⇄ Claude debate

✅ Consensus in 2 round(s) · reviewer effort: xhigh · base `79c05f9165ab`

|---|---|---|
| 1 | 86ee6d6d9 | fix(codex): close round-1 review findings (F1 minimap gate, F2/F3 stale meta/entryBucket) |

- **F1** — The minimap's render gate still checks only layout and `info`, even though the comment says it also waits for metadata
- **F2** — This `TerminalDisplayInfo` fixture still carries a full `meta` record after the production type removed it
- **F3** — The implemented design note still says idle classification is `entryBucket`'s job, but this change deletes that export

srid:

Ran with `--no-elegance` (the `/simplify` track already covered the elegance pass), scoped to `git diff 79c05f9165ab...HEAD`.

| Pass | Issues | Notes |
|---|---|---|
| Rules | 0 | All 33 built-in + project rules checked, each with a grep/enumeration |
| Fact-check | 0 | No swallowed errors, inaccurate fallbacks, wishful assumptions, logic errors, or slow leaks |

**All clear** — no fixes to apply. Both passes honored the deliberate-design list (the `meta`-field removal, `getMetadata` routing, `pairDisplayRow`, `DockEntry.meta`, `terminalSubject`'s asymmetric soft-gate, the minimap both-present gate) and still hunted the rest: dead-code grep confirmed `entryBucket` and all `info.meta` reads were fully removed with no strays, and Pass 2 verified routing `App.tsx`'s memo through `getMetadata(id)` doesn't regress into an undebounced recompute (it's a `reconcile`-backed proxy with a stable top-level reference).

PR #1898, opened 2026-07-20
fix(client): one boot deadline that escapes any hung leg — the 'Connecting to local…' wedge (#1763)

srid:
## The wedge

Intermittently, a fresh client load stuck at **"Connecting to local…"** and never reached the workspace (#1763). The connect overlay is held up by **any of three independent boot legs** — membership (`activeScope()` undefined), the session/list `isLoading` leg, or the daemon leg — but the only timeout escape (the old daemon-only `pendingTimedOut` → down/dead) was fed by the **daemon leg alone**, and was itself gated behind `activeScope()` being defined (the anchor lived *inside* the per-host wire scope). So:

- **Hole A** — a hung **membership** leg ⇒ `activeScope()` undefined ⇒ the anchor was never stamped ⇒ the ceiling could never fire. The safety valve sat downstream of the very thing that was stuck.
- **Hole B** — a hung **session** leg after the daemon delivered ⇒ `isLoading` true, daemon not pending ⇒ the daemon-tied ceiling stayed false.

Either wedged forever; there is no boot watchdog. (Investigation trail on #1763.)

## The fix — one outcome-based, per-host boot deadline

| Piece | What |
|---|---|
| `resolveCanvasMode(facts, { exceeded }) → { mode, tag }` | The module's ONE export. Each connecting/warming return **declares** a typed `BootTag` (`boot` + `leg` + `ceiling`) at its site; internal `resolvePrecedence` means a caller can't render a mode that skipped the deadline. Escape reads the tag off **one** resolve — no double-resolve, no cycle. |
| `bootDeadline.ts` | A per-host episode anchor keyed by `activeHost()` (always defined → **fixes Hole A structurally**, outside the per-host wire scope), a **monotonic** clock, and an **all-finite** ceiling table (local 30s / remote-provisioning 600s / remote-handshake 120s; accrual per class, a class change re-anchors). Retained across switches (no fresh grace on revisit), pruned on membership exit (fresh grace on re-add). |
| Honest escape | A hung **local kaval** leg reuses the byte-identical `down`/`dead` card (**#1713 preserved**); every other leg gets a new **`boot-stalled`** surface naming the leg (`bootStalledCopy.ts`) + the phase, over a `CanvasFailureCard` shell extracted from `HostDownCanvas`, with a **Reload** verb. |

Deletes the daemon-only ceiling (`daemonStatusPendingTimedOut`, `daemonPendingAnchorMs`, `pendingWindow.ts`) — **one** escape authority now, not two racing ones.

## Watch the outcome, not the legs

The deadline keys off whether the resolver's own output is a **boot overlay** (`tag.boot`, declared per return site), so a future 4th leg is covered by construction. The ceiling is phase-aware: a remote actively provisioning (copying/building) accrues against the generous 600s cell, never the 30s local one — a slow build doesn't false-fire.

## Scope of the one-ceiling claim

This unifies the **client boot/connect-overlay** ceiling only. `makeSession`'s server-side dial watchdog and the post-grace **transport overlay** (a mid-session drop, which this deliberately leaves as `boot:false`) are complementary layers, untouched.

## Residual

This kills the **no-escape class** — a hung boot leg now surfaces an honest, named, reloadable card instead of an eternal spinner. It does **not** fix the **hung-subscription trigger class** (why a leg hangs — a lost re-subscribe snapshot on reconnect, a half-delivering padi). That stays a separate investigation; the user now gets a truthful escape rather than a silent wedge.

## Tests

Flips the two RED pins (Hole A / Hole B). Adds: the R4 ceiling-table (exhaustive, all finite), R2 exclusions (restart-warming / records / `!channelLive` never escape), host-switch both directions (no false-fire; no fresh grace on revisit), ceiling-class transition (a 10-min build never trips the 120s cell), late-delivery un-wedge, the local restart-drain → down/dead, and the `boot-stalled` copy exhaustiveness + a render pin (the App `<Switch>` arm isn't blank).


srid:

✅ **Consensus** after 1 thread-round(s) · lowy + hickey · base `79c05f9165ab`

Settled: 1 auto-settled minor · 5 unopposed · 1 debated

### Applied (5)
- `lowy-1` ≡ `hickey-3` getMonotonicNow duplicates the shared-ticker receptacle and is mislocated in the policy module — commit `ae036edec`
- `lowy-2` [Switch to local] escape-hatch action is hand-duplicated across the two failure canvases — commit `7a81ebfa2`
- `hickey-1` Anchor accrue/retain/clear verdict fragmented across the tag and a hardcoded CanvasMode kind list — commit `aa99c5e84`
- `hickey-2` Boot-stalled card copy split between bootStalledCopy.ts and inline strings in the component — commit `cdcc3d69f`
- `hickey-4` CeilingClass derivation duplicated between bindingCeiling() and the connected-arm inline — commit `73400a2f0`

### Agreed — no change (2)
- `lowy-3` Boot-deadline episode ships three wires + a clock that useCanvasMode integrates by hand (packages/client/src/kaval/useCanvasMode.ts:119-125)
- `hickey-5` Per-host anchor lifecycle reconciled against membership on every canvas-mode evaluation (packages/client/src/kaval/useCanvasMode.ts:116-118 and packages/client/src/kaval/bootDeadline.ts:132-137 (pruneBootAnchors))

srid:
## Codex ⇄ Claude debate
✅ Consensus in 4 rounds · reviewer effort: high · base: origin/master

| --- | --- | --- |
| 1 | 06ab713d9 | membership-stall deadline reset every tick (F1, blocking) + F2/F3/F4 |
| 2 | bdcda8739 | App-Switch exhaustiveness tripwire (F2) + fold changelog + doc (F3) |
| 3 | 691e42beb | document the local-daemon exception accurately (F3) |

**Legend**
- **F1** — The membership-stall deadline is reset on every clock tick
- **F2** — This test calls itself an App Switch shell pin, but it renders `BootStalledCanvas` directly
- **F3** — The change introduces a new visible terminal state and recovery path after finite local, remote-handshake, and remote-provisioning deadlines, but the canonical Remote Hosts page still describes only ongoing progress and the existing cause-typed host-down cards
- **F4** — The `boot` fixture accepts arbitrary strings and casts them to `BootTag`, so typos or future union changes in these state-machine tests bypass TypeScript instead of failing at compile time.

<sub>codex reviewed the diff read-only; per-round detail is in the commit messages. The one blocking find (F1) was a real caller-level ordering bug the split unit tests missed — the empty-membership prune reset the anchor every tick, so Hole A could still wedge; fixed + pinned by a new caller-level tick test.</sub>

PR #1901, opened 2026-07-20
fix(surface): stop the stdio reader when a frame handler throws (#1859)

### LINKED ISSUE #1859: stdio-codec: readFramedLines settles on decode failure without stopping the reader (zombie connection on the override arm)
Surfaced by the architecture-first-principles one-hop audit on #1858 (minor, pre-existing — recorded rather than fixed there because the codec is shared by the client link and the server peer, widening the blast radius mid-gauntlet).

**The defect**: on a frame-decode failure, `readFramedLines` (`packages/surface/src/links/stdio-codec.ts:126-135`) `reject()`s the returned promise inside the `'data'` handler but never detaches its `'data'`/`'end'`/`'close'`/`'error'` listeners nor destroys the stream — the loop keeps consuming, base64-decoding, and dispatching subsequent frames to `onFrame` (and re-invoking the no-op reject) after the promise settled.

**Why it matters now**: `serveOverStdio`'s exit fork (#1858) treats "settled" as "the link is done". On the **default arm** the window is bounded to ~one macrotask by `process.exit(1)` — fine. On the **override arm** (e.g. `serveOverUnixSocket`'s per-connection serves, `unix-socket.ts:265`) one corrupt frame settles the serve promise `{reason:"error"}` and runs `peer.close()` while the socket stays alive and undestroyed, dispatching every later frame into the closed `ServerPeer` — a zombie connection the server-side accounting believes is over.

**Fix shape**: on settle (both resolve and reject paths), detach the `'data'` listener — and on the decode-failure reject, destroy the read stream — so `readFramedLines`' settlement and the actual read lifetime cannot diverge. Audit the client link (`links/stdio.ts`) for the same divergence while there.

srid:
## The bug (#1859)

`readFramedLines` — the stdio wire-framing codec shared by the client link (`links/stdio.ts`) and the server peer (`peer-server.ts`) — rejected its returned promise from **inside** the `'data'` handler when a frame handler threw, but left all four stream listeners (`'data'`/`'end'`/`'close'`/`'error'`) attached and the stream flowing. So the loop kept decoding and dispatching later frames **after** the promise had already reported the read as done.

Every consumer treats that promise settling as "the read is over." On `serveOverStdio`'s **override arm** — `serveOverUnixSocket`'s per-connection serves over one shared `net.Socket` — that divergence is a **zombie connection**: the serve settles `{reason:"error"}` and runs `peer.close()`, while the socket stays alive and undestroyed, dispatching every later frame into the closed `ServerPeer`. Bookkeeping dead, socket alive.

## Reproduced before fixing (repo rule #1690)

RED-first: two `it.fails` pins reproduced the zombie at both levels (codec, and `serveOverStdio` where a post-settle RPC frame still invoked the router handler), committed before any fix. **One honest correction to the issue's framing:** a *corrupt inbound frame* does **not** reach this arm through `serveOverStdio` — `decodeFrame` (`Buffer.from(…,"base64")`) is lenient and never throws, and `peer.message(frame).catch(…)` yields a caught async rejection, never a synchronous throw. The arm's only production-reachable trigger is a **synchronous** `onFrame` throw (the documented `onFirstRequest` hook).

## The fix — "settled ⟹ stopped" by construction

The promise never settles while the reader is still live. A single `stopAndReject` helper is the one reject path: it **destroys the reader first** (`destroyed` is set synchronously, so no later `'data'` can fire) and only then rejects. Both the frame-handler `catch` and the stream `'error'` listener route through it, and `return` after the catch abandons the rest of the in-flight chunk. Frames dispatched *before* the throw are correctly not undone.

This holds for **any `Readable` by construction** — it does not rely on `destroy(err)` echoing an `'error'` back (a custom `_destroy` may swallow it), nor on a stream `'error'` implying the stream is gone. The pre-#1859 zombie is unconstructible here.

## Blast radius: zero consumer code changes

`peer-server.ts` (`serveOverStdio`) classifies the non-benign `ORPCError` → `{reason:"error"}` and now actually closes the transport on the decode-failure arm; `endServing`'s guarded destroy no-ops on `!read.destroyed`; the client link's teardown is unchanged; `unix-socket.ts` logs one extra debug line.

## Review gauntlet reshaped it (all on this PR)

- **architecture-first-principles** — 0 confirmed findings; two honesty refinements (message/claim wording).
- **lens-debate** (lowy ⇄ hickey, consensus) — renamed the error code `SURFACE_STDIO_FRAME_DECODE_FAILED` → `SURFACE_STDIO_FRAME_HANDLER_FAILED`: the old name lied (base64 decode never throws; the only trigger is a handler throw), and it is grep-proven package-internal.
- **codex-debate** (xhigh, consensus) — caught a real **`throw null` re-zombie** (the catch dereferenced `(err as Error).message` before destroying, so a non-Error throw made the catch itself throw), and hardened the invariant to hold by construction for any `Readable`. +regression pins.
- **simplify** — reuse/efficiency/altitude clean; de-duplicated the invariant rationale.
- **code-police** — enforced a **synchronous `onFirstRequest`** (a TS void-return-compat `async` hook's rejection would otherwise escape as an unhandled rejection) and tightened test casts.

## No public-API change — no drishti pair gate, no Reference-doc update

The `@kolu/surface` `exports` map carries **no `./links/stdio-codec` subpath** — `readFramedLines` is package-internal, imported only by in-repo `stdio.ts` and `peer-server.ts`. No exported signature, `exports` entry, or oRPC contract shape changes. The renamed error-code string is grep-proven to have no consumer outside this PR's own files. Per `.claude/rules/surface.md` and `surface-reference.md`, this needs neither a drishti pair PR nor a Reference-page edit.

**ODU-IMPACT VERDICT: `none`.** Grep of odu's tree: it consumes only the public subpaths `@kolu/surface/peer-server` (`serveOverStdio`) and `@kolu/surface/links/stdio` (`stdioLink`) — never `readFramedLines`/`stdio-codec`; those signatures are unchanged and every path odu exercises is behaviorally identical.

## Tests

RED-first, then flipped green and joined by regression pins: a synchronous `onFrame` failure dispatches no later frame **and** destroys the stream; a frame *before* the poison in the same chunk still dispatches; resolve on an error-free `destroy()` (the `'close'` contract); destroy-mid-buffer partial discard; **a non-Error `throw null` still stops the reader**; **a `_destroy` that swallows the error still rejects**; **an `'error'` on an open stream still stops it**; an **async `onFirstRequest` settles `{reason:"error"}`** instead of escaping; and the `serveOverStdio` zombie closes the shared duplex and stops reaching the router.

Fixes #1859.


srid:

✅ **Consensus** after 1 thread-round(s) · lowy + hickey · base `722b24811c26`

Settled: 1 unopposed · 1 debated

### Applied (2)
- `lowy-1` ≡ `hickey-1` One try/catch complects the framing-decode axis with the consumer-callback axis, and the new message hardcodes away the codec's own declared volatility — commit `6fc54a3d4`
- `hickey-2` One design rationale fragmented across six prose sites — commit `1e0e8f193`

srid:
## Codex ⇄ Claude debate

✅ Consensus in 2 round(s) · reviewer effort: xhigh · base: master

|---|---|---|
| 1 | fea1991f4 | fix(codex): make readFramedLines 'settled ⟹ stopped' total (#1859) |
| 2 | f9dd9573a | fix(codex): align codec test proof-comments with the stopAndReject shape (#1859) |

- **F1** — The catch previously dereferenced an unknown thrown value and could throw again before stopping the reader.
- **F2** — The prior fix depended on `destroy(err)` being echoed through an error event, which a custom `_destroy` need not do.
- **F3** — The external error listener previously rejected without stopping a Readable that emitted error while remaining open.
- **F4** — ServeOverStdioEnd documentation previously misidentified malformed-frame decoding as the fatal synchronous path.
- **F5** — The regression test previously retained obsolete decode-failure terminology.
- **F6** — The codec test comments described the discarded destroy-with-error/re-entry proof and reversed the settled-implies-stopped invariant.

srid:

Two passes over `git diff 722b24811c26...HEAD` (rules checklist + fact-check). All findings actioned in `64a2101ff`.

| Pass | Finding | Action |
|---|---|---|
| fact-check | **onFirstRequest's "keep it total" was unenforced** — TS void-return compat lets an `async () => {}` satisfy `() => void`; its rejection would escape the sync read loop as an unhandled rejection (the crash footgun this module kills everywhere else). | **Fixed** — a thenable return is thrown at the call site, routing through the same classified `{reason:"error"}` arm a sync throw takes. +1 regression test. |
| fact-check | **Stale summary line** — `readFramedLines`' one-line summary omitted the new sync-`onFrame`-failure reject path. | **Fixed** — extended the summary. |
| rules · no-untyped-escape-hatches | `captured[0] as Uint8Array` discarded the `noUncheckedIndexedAccess` `undefined` arm; `captured.push(m as Uint8Array)` cast past the `EncodedMessage` union. | **Fixed** — typed `captured` as the union `encodeFrame` accepts (both casts gone) + a runtime `undefined` check before replay. |
| rules · no-silent-error-swallowing | The capture request's `.catch(() => {})` lacked a why-safe comment (its sibling swallow had one). | **Fixed** — documented why it's safe (fired only to emit one frame; never resolves). |

Verified clean by both passes: the `stopAndReject` idempotency against Node's `Readable` semantics, `caught-error-must-not-collapse-to-empty`, `no-overloaded-null` (the rename *removes* a conflation), `dry-rule-of-three`, and the callback-funnel guarantee. 450/450 surface tests pass; typecheck + biome clean.

srid:
## Evidence

Wire-layer codec fix — no on-screen impact, so the evidence is the test suite (RED → GREEN + regression pins) and the CI builds, not screenshots.

### RED → GREEN (the zombie, reproduced then fixed)

Committed **before** any fix (repo rule #1690), two `it.fails` pins reproduced the defect at both levels, then flipped to `it` once fixed:

- **codec** — a synchronous `onFrame` failure was shown to dispatch later frames after settle and leave `read.destroyed === false`; now it dispatches **no** later frame and destroys the stream.
- **serveOverStdio** — a serve settled `{reason:"error"}` was shown to leave the shared duplex open and a post-settle RPC frame still reached the router (`handlerCalls: 0 → 1`); now it closes the transport and the router is never reached.

### Regression pins added by the gauntlet

- `throw null` (non-Error) still stops the reader — the catch is total.
- a `_destroy` that swallows the error still **rejects** (not resolve-misclassified).
- an `'error'` emitted while the stream stays open still stops the reader.
- a frame **before** the poison in the same chunk still dispatches (no over-rotation).
- resolve on an error-free `destroy()` (the `'close'` contract); destroy-mid-buffer partial discard.
- an **async `onFirstRequest`** settles `{reason:"error"}` instead of escaping as an unhandled rejection.

### Results

- **450 / 450** `@kolu/surface` unit tests pass (`stdio-codec.test.ts` + `peer-server.test.ts` carry the #1859 pins); typecheck clean; biome clean.
- **GitHub CI green** on this commit: `build-and-push (ubuntu-latest)` ✅ (x86_64-linux) · `build-and-push (macos-latest)` ✅ · `Analyze (javascript-typescript / actions / python)` ✅ · `CodeQL` ✅.

_(Note: the `ci::*@x86_64-linux` commit statuses on this PR are stale artifacts from local odu runs during development — the sanctioned linux CI is the GitHub Actions build above.)_

srid:
### CI — corrected & final

Superseding the CI note in my previous comment (which mistakenly pointed at the GitHub Actions build and called the `ci::*` statuses stale): the sanctioned kolu CI is **odu across both platforms**, and it is now **green on the final HEAD** (`64a2101ff`, run `#5`, strict + posting):

- **`x86_64-linux`** on leased warm pool box **`kolu-ci-1`** — all 14 lanes ✅ (`ci::unit` 2m1s, `ci::e2e` 3m33s, `ci::nix`, `ci::smoke`, `ci::home-manager`, `ci::flake-check`, `ci::biome`, `ci::fmt`, …).
- **`aarch64-darwin`** on **`rasam`** — all 14 lanes ✅ (`ci::unit` 2m38s, `ci::e2e`, `ci::nix`, `ci::smoke`, …).

28/28 nodes `ok`, both lanes carried and pinned by construction (not localhost). `gh pr checks` is fully green and the PR is `mergeStateStatus: CLEAN`.

PR #1909, opened 2026-07-20
feat(dock): drag-to-resize the maximized dock, like the right panel

srid:
## What

In **maximized mode**, you can now drag the dock's right edge to resize it — just like the right panel. The width persists across reloads (per-device), so a 13″ laptop and a 27″ desktop each keep their own dock width.

- **Where:** the handle appears only on the maximized **cards** dock, where the dock is a real left-panel flex sibling of the canvas. The rail is a fixed chip strip and the tiled float is a card, so neither offers the handle.
- **How:** hover the right edge → `col-resize` cursor → drag. **Double-click** the edge resets to the default width (288px).
- **Bounds:** clamped to 200–560px — narrow enough to stay legible, not wide enough to swallow the canvas.

## How it works

| Piece | Detail |
|---|---|
| Persistence | `dockCardsWidth` via `persistedPref` (localStorage, per-device — same mechanism as `dockMode`), not a server pref |
| Drag | Reuses the existing `capturePointerGesture` helper (shared with tile resize + canvas pan); a fresh `AbortController` per gesture, auto-unwired on release |
| Live reflow | The width feeds `dockWidth()` reactively, so the aside reflows during the drag and the canvas `flex-1` sibling gives back the space |
| Clamp | The drag setter *and* the stored-value parse both route through `clampDockCardsWidth`, so a hand-edited localStorage value can't escape the range |

The width constants + clamp + pref live in a new `dockCardsWidth.ts` leaf module so the pure clamp is unit-testable without evaluating the whole Dock component graph, matching the dock folder's other tested leaves (`dockTree.ts`, `pipVariant.ts`).

## Tests

- New `dockCardsWidth.test.ts` — pins the clamp at both edges (floor 200, ceiling 560) and the pass-through/default cases.
- Added an ambient tip so the affordance is discoverable.
- `pnpm typecheck` + all dock unit tests green; biome clean.

## Evidence

Not yet captured — reviewer/author will exercise it in maximized mode. (Backend-invisible but on-screen; happy to attach a short video if wanted.)


srid:

✅ **Consensus** after 1 thread-round(s) · lowy + hickey · base `1e5ab31433a1`

Settled: 1 reconciled (raised by both lenses) · 2 auto-settled minor · 1 unopposed · 1 debated

### Applied (3)
- `lowy-3` ≡ `hickey-1` 'cards is the resizable posture' fact lives in two places — commit `afda8e291`
- `hickey-2` 'Reset to default' reconstituted at the call site from setter + exported constant — commit `5103d4baf`
- `hickey-3` Redundant test case re-asserts an identical clamp identity — commit `6bd21ba7f`

### Agreed — no change (2)
- `lowy-1` Electricity test: two existing receptacles reused, no boundary defect (packages/client/src/canvas/dock/dockCardsWidth.ts:1)
- `lowy-2` ≡ `hickey-4` Drag-start idiom triplicated above capturePointerGesture (packages/client/src/canvas/dock/Dock.tsx:195)

srid:
## Codex ⇄ Claude debate

✅ Consensus in 5 rounds · reviewer effort: xhigh · base: origin/master

The author (Claude) and codex (gpt-5.6, xhigh) debated the diff to consensus; every finding below is resolved. Per-round fixes are the commits:

|---|---|---|
| 1 | 2d8865a8 | fix(codex): host-cap the maximized dock width; harden the resize gesture |
| 2 | bd00bf36 | fix(codex): make the host cap actually engage; scope & cancel the gesture |
| 3 | 7cebe609 | fix(codex): cancel is cleanup-only; lock the active pointer; pin the cap in e2e |
| 4 | f3d7a81a | fix(codex): wire pointercancel only for opt-in callers; test cancel + strict persistence |

- **F1** (major) — The rendered width is clamped only to a fixed 560px maximum, not to the Dock's actual flex-host width
- **F2** (minor) — The in-flight AbortController is not aborted when Dock is disposed, and capturePointerGesture ends only on `pointerup`
- **F3** (minor) — Every pointer button starts a resize, and the window handlers do not bind the gesture to the initiating `pointerId`
- **F4** (minor) — The persistence boundary both accepts encodings its serializer never writes and silently hides rejected ones
- **F5** (minor) — The new tests cover only clamp arithmetic
- **F6** (nit) — `resetDockCardsWidth` has one caller and only forwards a bound constant to one setter, which is exactly the project's `no-thin-wrapper-functions` pattern.
- **F7** (nit) — The comment says a corrupt or out-of-range value degrades to the 'clamped default', but finite out-of-range values clamp to the nearest bound; only parse failures use the 288px default.

srid:

Ran Pass 1 (rule checklist) and Pass 2 (fact-check) over the diff vs `origin/master`. Pass 3 (elegance) skipped — `/simplify` already ran over this tree.

| Pass | Result |
|---|---|
| **Rules** | ✅ Clean — all 31 rules (7 built-in + 24 project) checked with evidence; no violations. |
| **Fact-check** | 1 issue, confirmed and fixed. |

**Fact-check finding (fixed in `6c0fb4c9`):** the resize handle called `setDockCardsWidth` from `onMove`, so a real drag fired a synchronous `localStorage.setItem` at 60–120×/s for its whole duration. Traced into `@solid-primitives/storage`'s `makePersisted` and confirmed the setter writes unconditionally with no debounce. The code claimed parity with the right panel's "epsilon/coalesce guard", but the right panel's protection is two-part (epsilon no-op **plus** a `coalesceMs: 150` debounced server write) — the dock had only the first half, on a `localStorage` write. `CanvasTile`'s resize gesture (which this handle was modeled on) avoids this by keeping a local pending signal during the drag and persisting once at `onEnd`.

**Fix:** a local, unpersisted `dragWidth` signal drives the live (still host-capped) render during the drag; `setDockCardsWidth` runs exactly once, in `onEnd` (skipped on a no-motion drag), and `onCancel` just drops the local signal. Verified: `tsc` + `biome` clean, 8 unit tests pass, all 18 `dock.feature` scenarios / 113 steps pass.

Issue #1754, opened 2026-07-10
BUG: claude-code live state can lie on a fast turn — stuck 'thinking' after completion / missed transient (needs initial full-scan on watcher attach)


srid:
**Severity: provider robustness BUG (can bite production, not just tests).** @srid flagged for prioritization — see the honest-connect defect class (a state that never reconciles).

## Summary

On a **fast, short turn** (a sub-3s model turn — e.g. a tiny ollama model on Apple Silicon, but also any quick real turn), kolu's claude-code live-state can be WRONG in two directions, and one of them is a **permanently lying state**: the indicator sticks on `thinking` AFTER the turn has fully completed and never reconciles to `waiting`.

## Two failure directions (both observed on aarch64-darwin CI, same commit, different runs)

**(1) Stuck `thinking` after completion — the lying state.**
The turn completes (ollama `/v1/messages` → 200) but kolu keeps showing `thinking` for the full 60s poll and never reconciles. Two contributing mechanisms:
- **Watcher attaches after the write, then only tails.** Claude Code writes its transcript JSONL at turn END. If the `claude-code` transcript watcher is installed *after* those lines land, a tail-only reader never sees the completion entry — the state derived at attach never advances. **The fix is an initial full-scan on attach**: when the watcher attaches, read the transcript from the START (not just new appends) so a fully-written transcript yields the correct terminal state immediately.
- **`stop_reason` mismatch.** `deriveState` (packages/integrations/claude-code/src/core.ts:500) maps `assistant + stop_reason === "end_turn"` → `waiting`; any other `stop_reason` falls through to `thinking`. ollama's Messages-API emulation may not emit `"end_turn"`, so even a fully-read transcript can derive `thinking`. The trailing-transient decay to `waiting` (core.ts:~1254) is probe/timing-dependent and did not fire within 60s.

**(2) Missed transient `thinking` — raced straight to `waiting`.**
Because the transcript lands only at turn-end, mid-turn `thinking` rides entirely on the screen-scrape poll (packages/integrations/claude-code/src/screen.ts). A sub-3s turn can slip between poll samples, so the sensor goes straight to `waiting`. (Documented in screen.ts.)

## Evidence (aarch64-darwin, ollama qwen2.5:0.5b, /v1/messages)
- Direction (1): turn 200 in 2.88s at 02:40:16; indicator stuck `thinking`, `kind=claude-code`, for the full 60s.
[...]

PR #1751, opened 2026-07-10
test(e2e): exercise real agent CLIs against ollama, delete all mocks

srid:
**All agent-feature e2e testing now runs REAL agent CLIs against a locally-served ollama model — zero mocks, zero fixtures, both platforms (x86_64-linux + aarch64-darwin).** The mock/fixture agent-detection machinery is deleted wholesale, and every agent feature is exercised end-to-end against the genuine `codex` / `claude` / `opencode` / `grok` CLIs pointed at ollama through a throwaway `$HOME`. _Motivation: mocks encoded our assumption of each CLI's on-disk shape; a real CLI writing real session files is the only thing that catches when that shape drifts (it already rescued the resume-by-id path — see below)._

## How it's wired

- **ollama + the model** are orchestrated by process-compose via services-flake (`nix/e2e-pc.nix`), npins-pinned, plain module evaluation — ollama service → model pull → warm → the cucumber `test` process, gated by the dependency graph. `exit_on_end` makes a failing suite fail the lane (no false green).
- **The real CLIs** are on the e2e devShell: `codex` pinned via `sadjow/codex-cli-nix` (native), `claude-code` (unfree, scoped-allowUnfree), `pkgs.opencode`, and the **genuine xAI Grok Build CLI** packaged from xAI's public GCS bucket (`nix/packages/grok.nix`, static Bun binary, unfree). The npm-shim scenario drives the **node-distributed** codex (`runtime = "node"`) so its foreground basename is `node`, exercising the OSC 633;E command-hint detection path (#673/#677) against a real npm-distributed codex.
- **hooks.ts** seeds each agent's config in the throwaway home to point at ollama (codex Responses API, claude Messages API, opencode OpenAI-compatible provider, grok custom-model). The server child runs `HOME=<throwaway>`, so every provider's state dir resolves there off `os.homedir()` — **no `KOLU_*_DIR` override knobs** (deleted; the no-override-knob doctrine).

## e2e cost: instant mocks → real ollama-backed agents

Going from instant mock agents to real CLIs driving a locally-served model **barely moves the suite wall time** — the ~7 real-agent scenarios run in parallel with the ~460 unchanged ones, ollama's overhead is small, and swapping many tiny mock scenarios for a few real ones actually *drops* the scenario count.

| e2e cucumber wall (same idle boxes) | Before — mock agents | After — real CLIs + ollama |
| --- | --- | --- |
| aarch64-darwin | **2m56s** · 504 scenarios | **3m08s** · 470 scenarios |
[...]

PR #1914, opened 2026-07-21
fix(agent-state): append-robust floor under fs.watch — stop the fast-turn stuck-'thinking' strand (#1754)

srid:
## The bug

On a **fast, sub-3s turn** — a tiny local model on Apple Silicon, or any quick real completion — an agent-state tile could **freeze on a transient state and never catch up**: the pill stayed **thinking** long after the turn finished and never reconciled to **waiting**. It's a *lying* state, not just a slow one, and it never self-corrected.

Every provider's session-watcher derives live state by re-reading the agent's on-disk session file when an `fs.watch` event fires. `fs.watch` is a pure **edge** signal, and the OS is explicitly allowed to drop or coalesce an edge — macOS kqueue coalesces an append that lands right after the watcher attaches; an inotify queue can overflow across a suspend. On a fast turn the terminal completion append arrives *after* attach with **no subsequent write**, so a single dropped edge strands the state forever, because nothing ever re-reads. (Reproduced deterministically against the real grok and claude watchers — see `REPRO-1754.md`.)

## The fix — a level-triggered floor under the edge fast path

A new one-line-to-adopt primitive, **`subscribeFileAppends`** in `kolu-io`:

- keeps `fs.watch` as the **fast path** (sub-interval latency in the common case), and
- adds Node's builtin **`fs.watchFile(path, { interval }).unref()`** as a **floor** — libuv's `uv_fs_poll` re-stats the path on a bounded cadence and fires whenever the full stat record changes (a superset of size/mtime incl. `ino`, so it also catches same-size WAL rewrites and rotations), tolerating a not-yet-existing file and firing on appearance/deletion, and
- on a **zeroed-stats fire only**, does one disambiguating `fs.stat` purely to pick the log level — ENOENT is expected-absent (silent), any other errno (EACCES) is logged, because the builtin's `StatWatcher` has no errno channel and would otherwise let a permission fault read as "absent".

A dropped edge now **self-heals within one interval** (`DEFAULT_APPEND_POLL_MS = 1000 ms`; a probe recovered a dropped edge in 903 ms). Both paths feed each watcher's *existing* 150 ms-debounced, change-gated handler, so an idle file does zero work and a busy file coalesces to one re-read per window — no churn regression.

Wired **unconditionally** (the file need not exist yet) at all three family seams:

| Seam | File |
| --- | --- |
| grok | `events.jsonl` |
| claude-code | the transcript JSONL (now on the deterministic path — the old none/waiting/watching dir-watch state machine is gone) |
| codex + opencode | the shared SQLite-WAL subscription (`wal-subscription.ts`), keeping the parent-dir inode-rearm |

Subscribing unconditionally closes a real hole the ruling surfaced: grok's old appearance bootstrap never re-armed a file watch after `events.jsonl` appeared, and macOS kqueue dir-watches never fire on child content appends — so a session matched before the file was flushed got **no** steady-state watch on macOS at all.

## Design philosophy

- **Fail-fast, no knobs.** `intervalMs` is a **required** parameter (no optional-with-default dead knob); every production site passes the module constant `DEFAULT_APPEND_POLL_MS`. A caught stat error surfaces at `error` rather than collapsing to a silent "absent".
- **Reuse the source of truth.** The floor is Node's builtin `fs.watchFile`, not a hand-rolled poll loop — the module doc records why the *bare* builtin was insufficient (the EACCES/ENOENT zeroed-stats indistinguishability) so it isn't re-litigated.
- **Electricity boundary.** The primitive lives in `kolu-io`, the existing dependency-light "filesystem I/O primitives" package (sibling to `createDirFilenameWatcher`) — the fs-event-delivery volatility is its axis. This introduces the **first `kolu-shared → kolu-io` workspace edge** (acyclic; `kolu-io` imports no `kolu-*` beyond `@kolu/log`).

## Tests

- `kolu-io` — the primitive matrix: dropped-edge recovery, idle-no-fire, appearance, edge-fast-path, teardown-races-a-tick, ENOENT-silent.
- `kolu-shared` — the WAL floor recovers a dropped WAL append (covers codex + opencode).
- `grok` / `claude-code` — the floor is reachable through each real watcher end-to-end.
- `repro-1754/*.test.ts` — the original #1754 reproductions, **inverted** to assert self-heal-to-`waiting`.

## Accepted residual

`signals.json` (grok's context-token count) stays edge-only: it's re-read on every `events.jsonl` tick, so a dropped edge there only staleness the token *display* until the next tick — a display-freshness residual, not a state lie. Wrap it later if it ever bites. And on a coarse-mtime filesystem a same-size same-inode overwrite within one mtime granule is invisible to any stat key (SQLite's normal post-checkpoint WAL rewind is exactly this) — the WAL's parent-dir kick recovers it, and the edge path stays primary; the floor is a floor.

## Restoration pending (tracked by #1754 — this PR is `Refs`, not `Closes`)

This PR lands the fix + its own tests and restores **zero** e2e asserts. Per the issue's resumption trigger, #1754 closes only once the real-agent e2e state-asserts descoped in **PR #1751** (`e2e-ollama-real-agent`, currently open) are restored — and those `*-real` feature files live in #1751, **not on master**, so they can't be touched from this master-based PR. After this PR and #1751 both merge, a small follow-up restores all **five** #1751-coupled targets:

1. `claude-real.feature` — the state machine (thinking→waiting), not just detection.
2. `codex-real.feature` — thinking→waiting + working-bucket + token-delta.
3. `grok-real.feature` — thinking→waiting + working-bucket.
4. `claude-cli-real.feature` `/compact` — the prior-turn `state="waiting"` (stuck-pill) assert.
5. `sleeping-terminals.feature` — the two **real-codex** scenarios' transient detection-appear window (these are #1751's real-ollama conversions; master's `@codex-mock` scenarios are deterministic and were never #1754-weakened).

Design record and reproduction ride along on the branch: `DESIGN-1754.md`, `REPRO-1754.md`.


srid:

✅ **Consensus** with zero debate turns · lowy + hickey · base `1e5ab31433a1`

Settled: 8 auto-settled minor

### Applied (4)
- `lowy-2` `label` is an optional-with-default knob — the exact dead-knob shape intervalMs's own comment condemns — commit `73ab0d32c`
- `hickey-1` grok watchPath doc-comment still describes events.jsonl it no longer handles — commit `6a273e8cd`
- `hickey-2` grok keeps two parallel teardown collections for one teardown concern — commit `8ac0c4d21`
- `hickey-3` WAL header doc still casts the dir-watcher as THE appearance + recreate mechanism — commit `1e9ba64ca`

### Agreed — no change (4)
- `lowy-1` Electricity test — subscribeFileAppends is a correctly-homed receptacle (packages/integrations/io/src/file-append-watcher.ts:105)
- `lowy-3` intervalMs couples the receptacle interface to the poll-mechanism choice (packages/integrations/io/src/file-append-watcher.ts:77)
- `hickey-4` grok now has two single-file-watch primitives for the same 'watch a maybe-absent session file' concept (packages/integrations/grok/src/session-watcher.ts:113-135)
- `hickey-5` Repro/design scaffolding committed to repo root outside any package (repro-1754/fswatch-shim.ts:1)

srid:
## Codex ⇄ Claude debate

✅ Consensus in 3 round(s) · reviewer effort: xhigh · base: origin/master

|---|---|---|
| 1 | d14cb572 | fix(codex F1/F2/F4/F5): close fs.watchFile startup-baseline race + guard emit funnel |
| 2 | 9b46c288 | fix(codex F1): revert B1 — hand-rolled statSync poll with synchronous baseline |
| 3 | a11c2c6c | fix(codex F6/F7): three-valued observation + no reentrant re-arm |

- **F1** — fs.watchFile's first asynchronous stat establishes its baseline; it is not a delivered observation
- **F2** — The teardown test deletes the file and unsubscribes immediately, before the watchFile listener or its disambiguating fs.stat has begun
- **F3** — The primitive catches hard fs.watch and disambiguating stat failures but reports them only through log?.error
- **F4** — The exported watcher invokes onChange directly from two independent emission paths (fs.watch and fs.watchFile) with no shared exception boundary
- **F5** — The changelog announces a material reliability change to the agent-detection pipeline for all four agents, but the canonical Agent Detection page still describes bare JSONL/WAL watches and does not mention the low-rate reconciliation floor
- **F6** — The new statKey returns null for both ENOENT and every hard stat error
- **F7** — If onChange unsubscribes reentrantly, unsubscribe sets closed and clears the currently executing timer, but control returns to poll and unconditionally schedules another timer

srid:

Ran the rule checklist + fact-check passes over the diff (`--no-elegance` — the simplify pass already ran).

**Fixed & committed (2):**
- `8282e663` — derive the poll-wait test constant from `DEFAULT_APPEND_POLL_MS` instead of three independent `sleep(1400)` hardcodes (dry-rule-of-three).
- `1797c763` — cover the **unconditional-subscribe-on-a-not-yet-existing file** path for the claude + grok seams (the actual wiring change this PR makes; the earlier floor tests always created the file first).

**Surfaced, not auto-applied (with reasoning):**
- `log?: Logger` is optional on `SubscribeFileAppendsOpts` — but `AgentAdapter.createWatcher` requires a real `Logger` at the top of the dispatch contract, so all three production call sites pass one; matching the codebase's internal `log?` convention. Follow-up, not this diff.
- `observe()` runs `fs.statSync` per interval on the loop — consistent with the pre-existing `statSync`/`readFileSync` convention across this watcher module family, not a new departure.
- `walIdentity()` ENOENT/error null-collapse — pre-existing, unmodified; the diff's `string|null` widening is a net improvement (equality-cached vs. the old unconditional close).
- `tailJsonlLines` silent-collapse-on-any-stat-error (claude core) — pre-existing, documented, module-wide policy; wider blast radius now, flagged for a follow-up.

Verified with `just fmt`, unit tests, and typecheck for the affected packages.

srid:
## Evidence



▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=evidence-selfheal-1754.mp4

This recording shows two things back to back, both run for real on a clean Linux box (no staged output):

1. **The primitive unit suite** — `packages/integrations/io/src/file-append-watcher.test.ts`, 9 tests covering the `statSync`/`fs.watchFile` poll-floor primitive: dropped-edge recovery (the #1754 core case), idle-no-fire, appearance-of-file, the attach→baseline race window, edge-fast-path timing, teardown/unsubscribe races, and ENOENT-silent handling. All 9 pass.
2. **A live self-heal demo** — `repro-1754/evidence-selfheal.test.ts` drives the *real* `createGrokWatcher` session-state watcher with its `fs.watch` edge suppressed via a shim (`repro-1754/fswatch-shim.ts`), and prints the actual state transition as it happens:
   - `state = THINKING` while the turn is open
   - `terminal 'turn_ended' appended → fs.watch EDGE DROPPED` — the shim swallows the callback, reproducing the macOS-kqueue coalescing race
   - `...waiting on the statSync poll floor (interval = 1000ms)...`
   - `state = WAITING ← RECONCILED` at ~1.15s, with no edge and no further write
   - `✓ SELF-HEAL VERIFIED` and the test's own assertion (`states.at(-1) === "waiting"`) passing

**Honesty note — why there's no dock-pill browser GIF:** #1754 is a macOS-kqueue coalescing race in `fs.watch`. The companion repro suite showed Linux inotify never strands the pill (0/300 trials), so a Linux GUI capture would just show a correct pill sitting there — it would prove nothing, since evidence only transfers within its class (a Linux browser recording can't stand in for a macOS kernel race it cannot reproduce). This deterministic injected-edge terminal demo, exercising the real watcher code path end to end, is the faithful in-class artifact of the actual state transition the fix repairs.

The full unit + integration + WAL + inverted-repro suites are green in CI.

srid:
## Coordinator CI note (record for merge)

**CI accepted as sufficient by srid** (2026-07-21, overnight autonomous run):

- **darwin (aarch64-darwin=rasam) FULLY GREEN** on HEAD `e2515fa2` — all 14 nodes incl. `ci::nix`, `ci::unit`, `ci::e2e` (521 scenarios / 4620 steps), `ci::home-manager`, `ci::smoke`, `ci::fmt`, `ci::biome`, `ci::flake-check`, `ci::atlas-sync`. Host verified non-localhost per venue law.
- **Full local green**: `pnpm -r test:unit` (45 pkgs), `just check`, `just fmt`.
- **linux lane WAIVED** — blocked by a github-API rate-limit (403) infra outage on the `devour-flake` fetch, **unrelated to this diff** (darwin passed the identical `ci::nix`/`ci::home-manager` recipes). Tracked in #1204; the infra fix is being prepared separately.

Merge-ready pending srid's merge. The `## Evidence` comment (terminal self-heal capture) and the full `/be-review` gauntlet trail are above.

Issue #1918, opened 2026-07-21
Agent-state watchers: extract a shared fs.watch-with-backstop primitive (grok strands UNBOUNDEDLY; the #1754 class, three hand-rolled copies)


srid:
## Why now

The #1754 gate (claude-code stuck-`thinking`) surfaced that the underlying volatility — *re-derive an agent's state reliably despite unreliable `fs.watch` delivery (OS coalescing / macOS kqueue drops)* — is **transport/OS-level and domain-agnostic**, yet it is hand-rolled separately in each provider integration. The #1754 fix (STATE1, claude-code only) patches ONE copy; this issue tracks the shared primitive and the sibling bugs.

## grok is strictly worse than claude-code (verified)

`packages/integrations/grok/src/session-watcher.ts` uses `fs.watch` as the **sole** re-derive trigger with a 150ms debounce and a change-gate — but has **no recheck timer at all** (no `staleDeadline`, no `scheduleStaleRecheck`). So a dropped `turn_ended` append after attach strands grok's `thinking` **unboundedly** — until the next fs event, which for an idle session may never come. claude-code at least has the decay path as a partial mitigation; grok has none. grok also carries the mirror of #1754's M2: `PHASE_TO_STATE[...] ?? "thinking"` (grok/core.ts:266,376) — the same "unknown ⇒ assume working" graceful-degradation fold, with an endorsing comment.

## The shared primitive (electricity / volatility boundary)

Per conventions.md → *Volatility boundaries* and `docs/atlas/.../electricity.mdx`: a domain-agnostic capability hiding a hard volatility (here: unreliable fs delivery) with the dependency arrow pointing **out** is a candidate `@kolu/*` receptacle. Shape: **watch + debounce + fs.watch-independent bounded re-poll (settle-window) + change-gate**, each provider plugging in its own PURE derive fold over its own transcript/events shape.

**Check before extracting, not after:** claude-code tails a JSONL, grok tails `events.jsonl`, **codex reads `state_5.sqlite`** (a different transport) — verify codex fits the seam before committing the abstraction (a third consumer that doesn't fit is the signal the seam is wrong). Three hand-rolled copies is the smell; a forced abstraction over a mismatched third is the opposite failure.

## Work

[...]

PR #1919, opened 2026-07-21
fix(claude-code): stop the live pill lying `thinking` after a completed fast turn (#1754)

### LINKED ISSUE #1754: BUG: claude-code live state can lie on a fast turn — stuck 'thinking' after completion / missed transient (needs initial full-scan on watcher attach)

srid:
## The lie

On a fast, short turn the claude-code dock pill could stick on **`thinking`
after the turn had fully completed** and never reconcile to `waiting` — a
permanently lying state (field report: a pill sat "thinking" ~30 min while the
agent was idle; CI evidence on aarch64-darwin: a 2.88 s ollama turn stuck
`thinking` for the full 60 s poll). Two independent mechanisms produce it, and
this PR folds both. **Reproduced first** (per #1690) — the committed RED pins
each; the issue's *original* "read-from-start on attach" framing was **refuted**
against today's code and is documented rather than tested (see below).

## M2 — the `stop_reason` fold stranded a yielded turn

`deriveState` mapped only `assistant + end_turn → waiting`; every other
*completed* (non-`tool_use`) stop_reason fell into the bare `assistant →
thinking` branch. So an emulated Messages-API provider (ollama's `"stop"`) — or a
real non-`end_turn` terminal reason (`max_tokens`, `stop_sequence`, `refusal`) —
read as working forever.

Now a finalized assistant tail whose `stop_reason` is a **populated** value not in
`CONTINUATION_STOP_REASONS` reads `waiting`. The set is a deliberate **deny-list**
(named, per gate C2): the *terminal* set is open across emulators, so keying off
the small, bounded *continuation* set (today just `pause_turn`) is the robust cut
— the next SDK continuation reason is a discoverable extension point, not a silent
early-clear. `pause_turn` and a `null`/streaming stop_reason stay `thinking`.

| trailing `stop_reason` | before | after |
|---|---|---|
| `end_turn` | waiting | waiting |
| `tool_use` | tool_use | tool_use |
| `pause_turn` | thinking | thinking |
| `null` / absent | thinking | thinking |
| `max_tokens` · `stop_sequence` · `refusal` · emulator `"stop"` | **thinking (bug)** | **waiting** |

**Disclosed failure-direction flip (gate C2):** the old bug failed *closed* (stuck
working — safe); the deny-list fails *open* (a future unlisted continuation reason
would clear the pill early). Accepted because the continuation set is small and
rarely grows — which is exactly why a new one must be added to the set.

## M1 — a dropped completion event stranded the watcher

The terminal `end_turn` often lands *after* the transcript watcher attaches (fast
turn), and its `fs.watch` event can be coalesced/dropped (kqueue on macOS
especially). The watcher re-derived **only** on an `fs.watch` fire — a non-orphaned
trailing `thinking` arms no decay recheck (`decayTransientState → recheckAt:
null`) — so a missed completion event stranded `thinking` for the whole poll
window.

`appendRepollDeadline` arms a short (`APPEND_REPOLL_MS` = 750 ms) re-read off the
**existing** one-shot recheck timer while a stranding-prone working state is
published, `min`-folded with any decay/workflow-stale deadline. Per gate **C3** it
is bounded to a **settle window** (`APPEND_SETTLE_WINDOW_MS` = 10 s) that resets on
every write — a correctness bound, not just cost: an `fs.watch` event can only be
dropped *near* a write, so polling long after the last activity recovers nothing.
A genuinely long turn keeps emitting real appends that reset the window, so the
re-poll is the quiet-tail backstop, not a standing poll. Decay is left pure.

M2 also **widens `running_background` eligibility** (gate C4): a truncated turn
(`max_tokens`) holding a live workflow runId now promotes — it *is* busy-waiting.
Covered by a test.

## Why the issue's original direction was refuted

The issue first proposed "initial full-scan on attach (read from the START)". Its
own comment 2 corrects this, and the correction holds against today's code
(reproduced): every check already reads a **256 KB tail** and
`setupTranscriptWatching` fires an immediate on-attach derive, so a transcript
already carrying `end_turn` at attach derives `waiting` immediately. Reading from
byte 0 changes nothing (the terminal state is at the tail) and is strictly worse on
a multi-MB transcript. The real gap is the *other* half — the dropped
**append-after-attach** — which is M1. The gate panel ratified this refutation.

## Review gauntlet

RED → design gate (two-refuter panel, GO-with-corrections C1–C6, all folded) →
lens-debate (lowy ⇄ hickey) → correctness pass. Folded: a shared
`TRANSIENT_WORKING_STATES` set so the two timer policies can't diverge and
reintroduce the bug (hickey); dropped a dead `tool_use` deny-list entry (hickey);
`earliestDeadline` kept local as call-site glue (lowy, over hickey's move-it
suggestion — generalizing `nextStaleDeadline` would complect an array-of-runs fold
with a scalar min).

## Scope & residuals

- **claude-code only** (gate C6). The identical robustness gap in **grok** (which
  has *no* recheck timer at all, so it strands *unbounded*) plus extracting a
  **shared watch + fs.watch-independent-repoll primitive** for grok/codex is filed
  as **#1918** — intentionally not touched here so the diff stays true to its title.
- **Residual (out of scope, symmetric to the mid-turn-miss residual):** M1
  self-terminates on a terminal state, so a dropped append that *resumes* work
  after a non-`end_turn`/non-`pause_turn` reason could strand on `waiting` (the
  mirror). Requires a resume-after-yield *and* a coalesced drop — largely
  theoretical for real claude-code.
- **Follow-up (noted, non-blocking):** the 750 ms re-poll fires the fire-and-forget
  `refreshSummary()` even on no-op passes, adding bounded idle `getSessionInfo`
  load in production; a content-change gate on the whole fetch path fits the #1918
  primitive work.
- **Descoped e2e:** the real-agent `-real.feature` state-arc restorations the issue
  lists live in the unmerged PR #1751, not `master`; this PR lands the code fix +
  unit REDs, and those restorations are a stated downstream follow-up when #1751
  merges (not fabricated here).

## Tests

`stuck-state.test.ts`: the two REDs (flipped green), the C1 `pause_turn →
thinking` / `refusal → waiting` boundary pins, the C4 `max_tokens` + live-workflow
→ `running_background` promotion, and the C3 `appendRepollDeadline` policy
(window-bounded, self-terminating on terminal states). 199 tests green in the
package; typecheck clean.

Fixes #1754. Refs #1918, #1690, #1751.

### DIFF
--- a/packages/integrations/claude-code/src/core.ts
+++ b/packages/integrations/claude-code/src/core.ts
@@ -411,6 +411,46 @@ function toolUseOrAwaitingUser(content: unknown): "tool_use" | "awaiting_user" {
   return classifyByAwaiting(awaiting, total);
 }
 
+/** Stop reasons on a *finalized* assistant message that mean the turn is NOT
+ *  done — the harness will resume it — so the pill stays a working state rather
+ *  than clearing to `waiting`. This is deliberately a small, bounded
+ *  CONTINUATION set, NOT an open terminal allow-list: any OTHER populated
+ *  `stop_reason` — `end_turn`, `max_tokens`, `stop_sequence`, `refusal`, or an
+ *  emulated Messages-API provider's own marker (ollama's `"stop"`, …) — means
+ *  the turn yielded → `waiting`. Tracking the small continuation set is robust
+ *  to new emulator vocab; the next SDK continuation reason is added HERE (a
+ *  discoverable extension point), from the canonical union
+ *  `@anthropic-ai/sdk`'s `StopReason`
+ *  (`end_turn | max_tokens | stop_sequence | tool_use | pause_turn | refusal`).
+ *
+ *  Failure-direction trade-off (disclosed, #1754): the pre-fix bug failed
+ *  CLOSED — a non-`end_turn` reason stranded on a *working* state (annoying but
+ *  safe). This deny-list fails OPEN — a future continuation reason not listed
+ *  here would clear the pill EARLY ("done" while the agent still works, so a
+ *  user could type over a live turn). Accepted because the continuation set is
+ *  small and rarely grows, but it is precisely why a new continuation reason
+ *  must be added to this set. `tool_use` is included for semantic completeness
+ *  though it never reaches the fallthrough (it has its own explicit derive arm
+ *  → `tool_use`/`awaiting_user` upstream). */
+const CONTINUATION_STOP_REASONS: ReadonlySet<string> = new Set([
+  "tool_use",
+  "pause_turn",
+]);
+
+/** State for a trailing `assistant` entry whose `stop_reason` was neither
+ *  `end_turn` nor `tool_use` (both handled by explicit arms upstream). A
+ *  populated `stop_reason` means the message finalized: unless it is a
+ *  CONTINUATION reason (`pause_turn` — the turn will resume), the turn has
+ *  yielded → `waiting`. A `null`/absent `stop_reason` is a not-yet-finalized
+ *  message (mid-stream, or a `claude -c` synthetic replay) → still `thinking`.
+ *  Exported for the #1754 fold pins. */
+export function assistantTailState(
+  stopReason: string | null,
+): "waiting" | "thinking" {
+  if (stopReason === null) return "thinking";
+  return CONTINUATION_STOP_REASONS.has(stopReason) ? "thinking" : "waiting";
+}
+
 /** Derive Claude Code state from the last relevant JSONL message.
  *
  *  Walks backwards once, tracking two independent signals with different
@@ -505,8 +545,13 @@ export function deriveState(
             state: toolUseOrAwaitingUser(entry.message?.content),
             model,
           }))
-          .with({ type: "assistant" }, () => ({
-            state: "thinking" as const,
+          // #1754: a finalized assistant tail that yielded (any populated
+          // non-`tool_use`/non-`pause_turn` stop_reason — `end_turn` is matched
+          // above; `max_tokens`/`stop_sequence`/`refusal`/emulator `"stop"`
+          // land here) reads `waiting`, not stuck `thinking`. `pause_turn` and a
+          // null/streaming stop_reason stay `thinking` (see `assistantTailState`).
+          .with({ type: "assistant" }, ({ stopReason }) => ({
+            state: assistantTailState(stopReason),
             model,
           }))
           .with({ type: "user" }, () => ({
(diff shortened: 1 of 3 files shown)


srid:
## Evidence

This is a provider-robustness fix: its on-screen effect is a dock pill that
**stops lying** (no longer sticks on `thinking` after a completed turn). That
state can't be staged deterministically in a live browser — reproducing a
*dropped* `fs.watch` completion event, or an emulated non-`end_turn` stop_reason,
needs the real-agent-against-ollama harness whose state-arc asserts are **descoped
on #1751**. So the verifiable artifact is the **deterministic reproduction test**:
it reproduces the exact stranding (per #1690) and proves reconciliation.

### RED → GREEN (final tests, same file, source swapped)

Running `stuck-state.test.ts` against **`origin/master` source** (pre-fix) vs. the
**fix** — the tests are identical; only `core.ts` + `session-watcher.ts` differ:

**Pre-fix — 14 failed / 16:**
```
❯ src/stuck-state.test.ts (16 tests | 14 failed) 4048ms
  × stop_reason=stop|max_tokens|stop_sequence|refusal (turn yielded) → waiting
  × promotes a `max_tokens` turn holding a live workflow to running_background
    AssertionError: expected 'thinking' to be 'running_background'
  × appendRepollDeadline … (arms/never-arms/window/stat-fail/tightens)
  × reconciles to `waiting` after a post-attach completion append with no watch
    event   4003ms   ← stuck `thinking` for the full wait, the exact #1754 lie
```
(The 2 that pass pre-fix are the `pause_turn`/`null` → `thinking` guards — correct
in both directions.)

**With the fix — 16 passed / 16:**
```
✓ #1754 M2 — stop_reason fold: stop|max_tokens|stop_sequence|refusal → waiting
✓ #1754 M2 — pause_turn / null stay thinking   (C1 boundary pins)
✓ #1754 M2 — max_tokens + live workflow → running_background   (C4)
✓ #1754 M1 — appendRepollDeadline: window-bounded, self-terminating on terminal
✓ #1754 M1 — watcher reconciles a missed completion append   759ms
             ← the dropped-event append is recovered within one 750ms re-poll
Tests  16 passed (16)
```

The M1 integration test is the direct visual-equivalent: it neutralizes
`fs.watch` (so *no* append fires an event — the coalesced-drop, deterministically),
attaches the real `createSessionWatcher`, appends the completion `end_turn`, and
asserts the emitted state goes `thinking → waiting` off the watcher's own re-poll.
Pre-fix it stays `thinking` (the lie); with the fix it reconciles in ~750 ms.

### CI (venue law — GitHub checks)

All checks green, both platforms — including the **darwin lane** where the original
#1754 evidence was captured:

| check | platform | result |
|---|---|---|
| build-and-push (ubuntu-latest) | linux | ✅ 2m43s |
| build-and-push (macos-latest) | darwin | ✅ 4m37s |
| CodeQL / Analyze (actions·js-ts·python) | — | ✅ |
| SentinelOne CNS (IaC · secrets) | — | ✅ |

Package suite: **199 passed**, typecheck clean.

srid:
Closing — superseded by #1914, which is a strict superset (shared file-append-watcher primitive covering claude-code + grok + sqlite wal, with docs and e2e). This PR was dispatched on an inferred authorization and duplicates that work.

Issue #1912, opened 2026-07-21
Host isolation PR-2: destructive daemon wire-verbs (kill/killAll/restart) bypass the adopt/kill guard


srid:
## Host isolation PR-2 — destructive daemon wire-verbs bypass the adopt/kill guard

Follow-up to the host-isolation landing in **#1911** (fixes #1334 + #1375). That PR
closes the adopt/kill/spawn/drain paths (the supervisor endpoint) and the test-harness
fork/env vectors. It deliberately does **not** close one residual surface, tracked here.

### The residual surface (A9-arm2 of the design gate)

The daemons' own **wire verbs** are destructive for any same-uid client that reaches
their socket, and execute *daemon-side* in a `role=production` process — so a
client-side guard never sees them:

- kaval serves `terminal.kill` / `killAll` (`packages/kaval/src/ptyHostSurface.ts:463-469`).
- padi serves `terminal.kill` / `killAll` plus `restart` / `recycleKaval` — whose
  documented behaviour is *drain the terminals → recycle kaval (kill + spawn fresh)*,
  i.e. padi SIGTERMs the production kaval on behalf of any caller (`packages/padi/src/surface.ts`).

Reaching prod's socket from a dev context is easy: production-spawned terminals stamp
`KAVAL_SOCKET` / `PADI_SOCKET`, children inherit the env, and flag-less `kaval-tui` /
`padi-tui` autodiscovery resolves prod when it is the only daemon. #1911's **env-scrub
test-setup** structurally closes this for the **test** path (vitest/cucumber can no longer
see prod daemons — the actual #1375/#1334 incident vector), but an interactive dev tool or
an agent script shelling `padi-tui` / `kaval-tui` (or dialing `$KAVAL_SOCKET`) can still
kill production PTYs or the production kaval with zero refusal.

### Why it is its own PR (sizing accepted at the design gate)

1. `terminal.kill` / `killAll` / `restart` back **real interactive UI buttons** — a blunt
   "require the supervisor claim" would break the user's own kill/restart. This needs a
   genuine **caller-classification** model (interactive user vs non-production
   non-interactive client), not a claim check bolted on.
2. Surfacing **role in the connect `hello`** (the skew-proof control-core identity) is a
   **contract-version bump + forced padi-only drain** across padi + kaval + the convergence
   identity + the client dialers — a self-contained sub-project.

### Scope for this PR-2
- Add a caller-role / caller-classification axis to the padi/kaval connect handshake
[...]

PR #1911, opened 2026-07-20
fix(isolation): dev/test can't bind production's state-root (#1334)

srid:
Rescoped to **Lock 1 only** (state-root bind refusal) after srid's ruling. The campaign split into three parts: this PR (#1334, the identity core), **#1921** (the #1375 test-gate — the low-risk PR that merges first), and **#1912** (the supervisor adopt/kill guard, "Lock 2", which was rejected as not worth the complexity — tracked there with the residual adopt/kill-vs-running-prod path).

## The class (#1334)
A daemon's identity is **only** its state-root path digest. The server's `KOLU_STATE_DIR` already fail-fasts when unset, but `resolvePadiStateRoot()` **silently defaulted** to `$HOME/.local/state/padi` — so a bare `pnpm dev` / `vitest` / `padi` resolved **production's exact digest**, and the adopt/kill machinery would then act on the live production daemon.

## Lock 1 — forced private state-root
Production self-declares (`KOLU_ROLE=production`, set only by the production nix wrapper and threaded daemon→daemon, never into a PTY). A new `resolveBoundStateRoot` at both binder sites (kolu-server boot, padi `daemonMain`) **fail-fasts** a non-production process that would bind:
- production's **default** root (the bare-launch hole), or
- an explicit **override** at that default root, or
- a **persistently-marked** relocated root (#1414).

Each daemon stamps its role beside its own gate (ephemeral, boot-wiped); a production daemon also stamps a **persistent** marker on its state-root, which the bind refusal reads. **Fail-closed:** the default root and an unreadable manifest are refused clocklessly, so this holds even against a pre-fix daemon that never wrote a marker. The `pnpm dev` script sets `KOLU_PADI_STATE_DIR`, so only a truly bare launch ever trips it.

## Notes
- **Strict subset of the gauntleted campaign diff** (Lock-1 hunks only): the full `/be` review gauntlet (architecture-first-principles → lens → codex 9-round → simplify → code-police) ran on the combined diff, so its verdicts transfer to this carved subset.
- **odu impact:** `none` — this PR touches no generic `@kolu/surface*` package (Lock 2, which did, is dropped).
- Reproduction (safe, prod inactive; under a fake `$HOME`): [`docs/repro-iso-1334/REPRO.md`](docs/repro-iso-1334/REPRO.md).

Closes #1334.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1334, opened 2026-06-13
Production kolu.service disrupted during agent /be run — harden host isolation & add a production health gate


srid:
### What happened
During a long autonomous `/be` session (PR #1331), the live production `kolu.service` went down/restarted, disconnecting the user's clients mid-session. The user experienced it as "production was killed."

### What the data shows
- `systemctl --user show kolu`: **`Result=success`, `NRestarts=1`**, `ActiveEnterTimestamp = 2026-06-13 13:45:38`. So the prior instance stopped **cleanly** (not OOM, not a signal/crash) and the unit restarted **once**.
- Journal at 13:45:32–33 (immediately before the new instance came up): a websocket `disconnected … code:1001` (client going away), git reflog/index/working-tree watchers *retired*, and an in-flight `terminal.attach` aborted with `AbortError`. These are **consequences of the restart**, not its cause. (The large `"NOT_SUPPORTED_ERR": 9` log flood is just the DOMException constant table being serialized for each logged error — not a separate error storm.)
- The PR-evidence capture that ran in the same window executed **on an ephemeral `pu` box**, not locally — the orchestration transcript shows 27 `pu connect kolu-pr-1331 …` calls, the e2e harness invoked as `pu connect kolu-pr-1331 -- "bash ~/run-evidence.sh"`, and `pu destroy kolu-pr-1331` afterward. So this was **not** the #1109 failure mode (an agent running bare `just dev` / `just test-quick` locally and binding production's fixed ports).

### Assessment
The restart was clean and I could not attribute it to a specific local port-bind or crash from the agent run. The most plausible contributing factor is **resource/host contention**: a long agent session on the *same host as production* runs heavy local work (repeated `just check` / `just fmt` → `nix develop` + `biome` + `tsc`, multiple gauntlet subagents, several `chrome-devtools-mcp` chromium launches), which can make the live instance sluggish enough to be (manually or automatically) restarted. Either way, the user's production was disrupted during agent activity, which should not happen.

### Proposed hardening
1. **Production health gate in the agent flow.** `/be` §5 (and `/do`) should snapshot `systemctl --user is-active kolu` + main PID **before** and **after** the evidence/CI steps, and surface any restart/disruption immediately instead of finishing "green" while the user's instance bounced.
[...]

PR #1921, opened 2026-07-21
test(isolation): gate real-daemon test suites off by default (#1375)

### LINKED ISSUE #1375: Harden test harness: a local `vitest` run forked 30+ real daemons and OOM-reaped prod kaval
## What happened

During an autonomous `/be` run on a **workstation that also runs production kolu + kaval** (PR #1374), the agent ran the kaval unit suite locally for fast feedback:

```sh
packages/kaval/node_modules/.bin/vitest run --root packages/kaval   # ×2
```

`packages/kaval/src/socketDaemon.test.ts` spawns **real `kaval` daemon processes** (`node --import <tsx loader> bin.ts`) + real `kaval-tui` processes — ~30+ real processes across the two runs (fingerprint: 182 leftover `/tmp/kaval-e2e-*` dirs). The tests are *isolated* (per-test `mkdtemp` sockets, PID-scoped kills — they never touch the prod `$XDG_RUNTIME_DIR/kaval-*` socket), so nothing killed prod kaval directly. **But the process/memory spike OOM-reaped the user's long-lived production kaval** (a crash, not a kill). The agent should have run these on CI / a pu box (where they pass green), never on the workstation.

## Root cause

The kaval test binary forks real daemons with **no gate inside the binary**. Existing guards wrap *commands* (the `/test` skill routes e2e to pu boxes; `dev-server` uses random ports; do.md forbids bare `just dev`) — but a bare `vitest --root <pkg>` reaches **past the command to the binary**, bypassing all of them. **The fix must live in the test binary**, because only that is independent of how it's invoked, which harness runs it, or which directory it's in.

## Two traps (found by an adversarial review pass)

1. **kaval was just where it landed — the danger surface is wider.** The same bypass works on:
   - `packages/surface-daemon/src/daemonMain.test.ts`, `pidGate.test.ts` — fork real `node -e`
   - `packages/surface-daemon-supervisor/src/waitForPidGone.test.ts` — spawns real `sleep 30`
   - `packages/kaval-tui/src/{attach,create}.test.ts` — fork ~6 real shell PTYs via `terminal.spawn`
   - `packages/tests/support/hooks.ts` (Cucumber) — spawns a real kolu server + Chromium + Xvfb (gigabytes if run locally)
2. **Gating on `CI` is broken here.** odu runs unit tests as `nix develop -c pnpm test:unit` with **no `CI` env exported**. `describe.runIf(process.env.CI)` would *silently skip daemon tests in CI too* (coverage → 0), or — if widened to honor a locally-set `CI` — *re-arm the danger* on the workstation. `CI` is the wrong key.

## The fix that closes the class (impossible-by-construction)

- [ ] **1. A shared gate helper** — `describeDaemon(...)` — keyed **only** on `KOLU_DAEMON_TESTS=1`, **default OFF**, applied at *every* real-spawn test site (the files above). A bare `vitest run` on any package then forks **nothing**, regardless of invocation path.
- [ ] **2. A meta-test/lint that fails the suite if any `*.test.ts` forks a real process without routing through that gate** (grep for `spawn(process.execPath`, `node --import`, `servePtyHostOverUnixSocket`, real `ssh`/`nix`, PTY-forking `terminal.spawn`). *This* makes it structural rather than discipline — a future spawning test **can't exist ungated**.
- [ ] **3. CI's own just recipe exports `KOLU_DAEMON_TESTS=1`** (CI runs `just test-unit` → full coverage), **verified against a real odu unit log** that the daemon `it(...)` lines actually execute.

## Defense-in-depth (guardrails, never the barrier)

- [ ] **4. PreToolUse Bash hook** — authored in the **apm source** (so `just ai::apm` doesn't clobber it) — that denies local `vitest` / `pnpm test:unit` / `cucumber` and steers the agent to CI/pu. Bypassable + harness-scoped, so a tripwire, not a wall. (Do **not** trigger on "is a kaval live" — the resource spike harms even with no kaval up, and a detector can't tell prod kaval from a `just dev` one or the test's own `/tmp` gates → false blocks that train people to disable it.)
- [ ] **5. Ergonomic default:** `just test-unit` stays fork-free (the safe reach); `just test-daemon` (sets the opt-in) is documented as CI/pu-only.
- [ ] **6. Doc rule keyed on a *property*** ("never run a suite that forks real OS processes on the workstation") pointing at the lint — not a hand-maintained file list that rots.

**Net:** layers 1–3 make the workstation safe by construction even with a careless agent or a new harness; 4–6 are insurance.

---

_Filed from the `/be` run on PR #1374 (P2.5 `frontDaemonOverStdio`). Design produced + adversarially verified via an ultracode workflow; full verdict in the session transcript._

srid:
Splits the **Lock 3** (test-side resource leash) out of the host-isolation campaign as the low-risk, fast-merge PR. #1911 carries the Lock-1 state-root bind refusal for #1334; "Lock 2" (the supervisor adopt/kill guard) was rejected by srid and is tracked in #1912.

## The class (#1375)
A bare `vitest` forked **real** daemons with **no gate and no cap** — the incident: 182 leftover `/tmp/kaval-e2e-*` dirs, production kaval OOM-reaped.

## The leash
- **`@kolu/daemon-test-gate`** — a zero-`@kolu`-dep leaf, consumed **devDependency-only** by every package whose tests fork real OS processes.
  - `describeDaemon` gates every real-spawn suite on `KOLU_DAEMON_TESTS` (**default OFF**).
  - `assertDaemonSpawnAllowed` — a runtime leash at the spawn funnel that throws in a gate-off test context, so helper indirection / a dynamic import can't smuggle a fork past the gate.
  - the `./setup` per-worker env scrub blinds every worker to the production daemon locators (and, forward-looking, the `KOLU_ROLE` master key).
- `describeDaemon` wrapping of the real-daemon-spawning suites across the tree; `localKavalDriver`/`localPadiDriver` wrap their spawn with the leash (F5).
- a `ci::daemon` DAG node + `just test-daemon` run the gated suites as a distinct required check while `unit` stays fork-free; an advisory `no-ungated-forks` meta-lint.

## Notes
- **Independent of #1911** by construction: the leaf has no `@kolu` deps, the env-scrub deletes env-var *names* (strings), and nothing here imports `role.ts`/`stateRoot.ts`. This diff is a strict subset of the already-gauntleted #1911 diff (Lock-3 hunks only), so the review-gauntlet verdicts transfer.

Closes #1375.


srid:
RT-1334-P1CD — this is PR-A (Lock 3) of srid's three-way split. Pushed + subset-self-checked: 63 files, Lock-3 ONLY, verified zero Lock-1/Lock-2 guard symbols; just check green; daemon-test-gate 19/19; a gated supervisor suite runs green under KOLU_DAEMON_TESTS. Strict subset of the gauntleted #1911 diff (Lock-3 hunks off master), so the /be-gauntlet verdicts transfer. Sibling: #1911 (Lock 1, #1334). Lock 2 rejected → #1912. DRAFT — srid merges this one first. Full two-platform CI next.

srid:
RT-1334-P1CD — PR #1921 (Lock 3, #1375) COMPLETE forge rollup ALL-GREEN on HEAD 8bd4c108. `gh pr checks 1921` = **38/38 pass, 0 fail, 0 pending**. Both platforms via odu: darwin on real-remote nix-infra@rasam.tail12b27.ts.net + linux on leased pool box kolu-ci-3 (non-localhost, no fallback needed — the transitive-nixpkgs-403 self-cleared on nix retry this run). CodeQL + Analyze×3 + SentinelOne×2 + build-and-push(ubuntu+macos) all pass.

One carve-fix this run: the localPadiDriver spawn-leash CODE (padiBinding.ts) was missing on PR-A (I'd carved the leash TEST but not the source hunk — both are Lock 3); added `assertDaemonSpawnAllowed("a real padi daemon")` at the funnel, HEAD a905bbe7→8bd4c108, still zero Lock-1/2 symbols. #1911 (Lock 1) CI next: darwin on rasam (now free) + linux on the pool. Both stay DRAFT — #1921 is ready for you to merge FIRST.

PR #1913, opened 2026-07-21
fix(surface-remote): own provisioning subprocess lifetimes + honest campaign clock (#1908)

srid:
## What & why

The **outage fix** for #1908 (PR 1 of 2). On a fresh production deploy, a
`nix-store --realise` ssh child (pid 76722) wedged **~10 minutes** because no
layer owned its lifetime, and the warm "cached agent" probe was a fused
`nix-store --realise` that **substituted an absent-but-substitutable closure
inside a "check"** (real network fetch, under a label promising a check).

This PR closes the class at the transport layer. The user-visible dishonest card
(D2) is **follow-up PR 2**.

## The fix

- **D1a — the probe asks, never does.** Fused `--realise <drv> --add-root` → a
  sender-local `nix-store -q --outputs` + a bounded ssh `nix-store
  --check-validity`. A pure store query — never substitutes, never mutates
  (verified against real nix 2.34.7). Single-output agent contract enforced (a
  multi-output drv fails loud).
- **D1b — unowned children unspellable.** `runProgress`/`runCapture` take a
  **required** `LifetimePolicy` (`deadline` / `progress-liveness`). A stuck child is
  **group-killed** (detached process group → the ssh grandchild is reaped too, with
  a SIGKILL escalation that survives the leader's close) and the run settles **at
  the kill**. New arms `lifetime-expired` (retryable) and `aborted`. Per-step
  doubling + kill budgets (`makeProvisionBudgets`) whose exhaustion is **genuinely
  terminal** — a permanently-silent copy reaches `failed` in a bounded number of
  attempts.
- **D1c** — per-child ownership closes the eternal-await window by construction.
- **R6b — recheck aborts an in-flight dial** via `ConnectContext.signal`, redialing
  now; a `dialEpoch`/`isCurrent` (`+ !destroyed`) guard makes a superseded/late dial
  inert on reject **and** resolve (it tears its connection down and rejects, never
  handing a `pin()` awaiter a dead client).
- **R8b — a session-level pre-connected liveness backstop** (baked, non-configurable
  bound — the ordering invariant is unbreakable by construction) guards the
  `resolveDrvPath`-wedge seam.
- **D3 — the episode clock spans the campaign** (`sinceMs` stamped at campaign birth,
  not per retry), so a 10-minute wedge reads 10 minutes.
- **`failed.cause` is honest** — widened to `network | remote`: terminality is the
  `failed` phase, orthogonal to the transport class (a budget-exhausted silent step
[...]

Issue #1908, opened 2026-07-20
Remote provisioning: unowned subprocess lifetimes + a settled-looking failure card over a live retrying connector (field incident, fully evidenced)


srid:
## The incident (2026-07-20 ~19:04–19:15 EDT, production pureintent → zest)

Fresh deploy (new padi closure) → the zest connector narrates `checking for a cached agent…` at 19:04:20 and goes silent. The UI sits on "Connecting to remote:zest…" (elapsed timer showing ~1s — it resets per attempt), then the #1898 boot-stalled card appears: *"This host's agent isn't responding … Reload to try again."* Reload reproduces the same sequence. At 19:14:41 the host **connects spontaneously** — card still on screen, no Reload clicked.

## Root cause — code-grounded (file:line cited; runtime claims operator-witnessed; unknowns marked)

1. **The wedge.** The connector's warm fast-path runs ONE fused `ssh <host> nix-store --realise <padi.drv> --add-root --indirect` (`nixCopy.ts:181-210`), narrated as `checking for a cached agent…` (`:203`). The probe is *designed* to fast-fail on a cold miss — but when the closure is absent-yet-**substitutable**, `--realise` performs the substitution *inside the probe*: real network fetching (observed live: cache.nixos.asia) under a label that promises a check. On this incident that child (pid 76722) wedged ~10 minutes.
2. **During the wedge (all verified live):** zest's sshd reachable — fresh channels through the *same ControlMaster* ran commands instantly; zest had **no nix-store client process**; zest's nix-daemon consumed 1.4 CPU-seconds in 7 minutes; an identical manual realise completed in **2.6 s** (paths a 0.00 MiB fetch). The remote side of the exec was gone; the local child waited forever in poll. `ServerAliveInterval` cannot catch this — the transport was healthy; only the channel was dead.
3. **No deadline exists anywhere on the provisioning path — verified at all three layers:**
   - `process.ts:80-150` (`runProgress`/`runCapture`, read in full): spawn with no timeout, no AbortSignal, no kill path; resolve only on the child's own `close`/`error`.
   - `sshConnector.ts:119`: `await provisionAgent(...)` inline in `connectOnce`, unbounded; the header (`:10`) delegates all watchdog machinery to `makeSession`.
[...]

PR #1925, opened 2026-07-21
fix(client): honest boot-stalled connector card + campaign backstop (#1908)

srid:
**PR2 of the #1908 split; pure-client, no drishti pair — the surface-remote changes shipped in #1913.** This is the user-visible remainder #1913's description deferred ("The user-visible dishonest card (D2) is follow-up PR 2").

## The bug

On a fresh production deploy, the #1898 boot-stalled escape card rendered **terminal** copy — *"This host's agent isn't responding — Reload"* — over a connector that was **still retrying** (the wedge cleared on its own minutes later, card still on screen). Two defects on the connector-owned leg:

1. **Copy framing.** The card read as a terminal wedge, offering only *"reload to keep watching"* — no honest "still retrying" truth, even though the connection stream it already reads carries the live phase.
2. **Recovery verb.** `[Reload]` is `location.reload()`, which recycles only the **browser** — it cannot recycle the server-side ssh connector that owns the dial.

## D2 — the honest connector card

The resolver now decides a `BootStalledRecovery` verdict in `escapeSurface` and the card renders it (the decision is carried, never re-derived):

| leg | who owns recovery | card |
| --- | --- | --- |
| a **warming REMOTE** entry — ANY phase (probing/copying/building/connecting) | the server ssh connector (PR1's `recheck()`) | **connector** card: non-terminal *"kolu is still setting up… still retrying"* copy + live phase detail, **[Retry connection]** → `client.hosts.reconnect` |
| a connected host's session/daemon subscription, a membership stall | the client | **client** card: the leg's own copy, **[Reload]** (`location.reload()`) |
| a hung LOCAL kaval | — | byte-identical down/dead card (#1713, unchanged) |

The connector card is **non-terminal by construction**: while the entry is `warming` the connector has *not* reached its own `failed` verdict — that flips the entry to `failed` → the host-down card, a different surface.

## R8a — the campaign backstop

PR1's D1 retry cycle can flap a warming host's phase fast enough that the per-class ceiling anchor re-zeros on every class change and so **never trips** — the escape card could never fire for a persistently-wedged host. Added a **second, class-blind** cell in `bootDeadline.ts`, armed off the frame's own connector-owned `provisioning` leg, held across class flips, cleared on settle. Its elapsed is measured **purely on the client monotonic clock** (bounded at 30 min, comfortably above the 600s provisioning cell and the server's own R8b/R4 grants) — so it can't freeze during a silent wedge or be jolted by a wall-clock/NTP step. A campaign-backstop escape on a warming entry routes to the **same non-terminal connector card**. The **Retry connection** verb deterministically resets this host's deadline on RPC success (`resetBootDeadline`), so the card dismisses even on a same-class retry.

## Review gauntlet

Full `/be` gauntlet ran on this branch and its fixes are committed here:
- **lens-debate** (lowy ⇄ hickey): consensus — shared `reconnectAction` factory, doc/type corrections, memoized card view.
- **codex-debate** (xhigh, 5 rounds → consensus): reworked R8a from the fragile server `sinceMs` to the client-monotonic anchor + explicit success-gated Retry reset; stabilized the card's button identity across the 1s canvas re-resolve; corrected the docs' backstop guarantee.
- **simplify**: routed the card's derivations through value-deduped primitive memos (killed a per-second `dataAttrs` DOM churn).
- **code-police**: `escapeSurface` migrated to `match(...).exhaustive()` (prefer-ts-pattern); fact-check clean.

## Tests & evidence

Reproduce-first: the D2 RED pins (`bootStalledProvisioningRecovery.test.tsx`, `it.fails`) shipped on #1913's branch — flipped GREEN here. Full client suite green (835), typecheck + biome + fmt clean, CI green on **both** ubuntu-latest and macos-latest. Visual evidence of both card shapes is posted as a PR comment.

## ODU-IMPACT VERDICT: `none`

Pure-client — no `@kolu/surface*` public API touched (guardrail held). A grep of odu's tree at its pinned kolu SHA finds no consumption of any client-internal symbol changed here.


srid:

✅ **Consensus** after 1 thread-round(s) · lowy + hickey · base `ceb1eae05ea0`

Settled: 1 auto-settled minor · 2 unopposed · 2 debated

### Applied (4)
- `lowy-1` Reconnect recovery verb hand-duplicated across the two failure canvases (missed shared action receptacle) — commit `8dc6123fe`
- `lowy-2` ≡ `hickey-4` `StalledLeg` value `provisioning` broadened to the whole connector campaign, but its doc still says 'copying/building' and its name diverges from the `connector` axis it maps to — commit `70a519e40`
- `lowy-3` ≡ `hickey-2` `view()` recomputed five times per render in BootStalledCanvas JSX — commit `643a0aa0a`
- `hickey-1` campaignMs gate approximates the resolver's connector predicate from a different fact — commit `65e94a4a8`

### Agreed — no change (1)
- `hickey-3` Redundant `as const` on tone inside a contextually-typed return (packages/client/src/host/BootStalledCanvas.tsx:66)

srid:
## Codex ⇄ Claude debate

|---|---|---|
| 1 | 52491cce0 | fix(codex): monotonic campaign backstop, stable card buttons, honest docs |
| 2 | 97eee6c47 | fix(codex): campaign anchor resets on retry + retain clears it (F1, F7, F8) |
| 3 | b9e883d01 | fix(codex): explicit Retry-connection deadline reset; drop sinceMs (F1, F3) |
| 4 | 88a4f80cb | fix(codex): reset deadline only on reconnect success; honest backstop docs (F9, F4, F5) |

- **F1** — `info.sinceMs` is a sample stamped only when `makeSession` publishes a state/log frame, not a clock that advances on the client's one-second monotonic tick
- **F2** — The `view` memo is keyed by `props.recovery`, but `canvasMode` returns a fresh recovery object on every one-second monotonic-clock recomputation
- **F3** — The new deadline treats `campaignMs` as authoritative elapsed time, but its producer computes it with `Date.now() - episodeStartedAt` (`packages/surface-remote/src/session.ts:635-637`)
- **F4** — The canonical Remote Hosts page still says every remote stalled card offers Reload
- **F5** — The existing unreleased Remote Hosts `added` entry still promises Reload for a remote agent stall and its PR list stops at 1898
- **F6** — The `CanvasMode` doc still says `boot-stalled` carries the stalled leg plus the phase, but the payload now carries `BootStalledRecovery`: client arms have a leg and connector arms have a phase, never both.
- **F7** — The new `retain` branch is a no-op for the campaign map, even though every current `retain` comes from the connected arm (restart-warming, records-awaited, or dead-channel connecting), which proves the connector campaign has ended
- **F8** — `just fmt-check` fails on the newly added `recordBootFrame` call, so the required formatting/CI gate is red despite the unit tests and typecheck passing.
- **F9** — Round 4 resets both deadline anchors synchronously before `client.hosts.reconnect` is known to have succeeded

srid:

Ran Pass 1 (rule checklist) and Pass 2 (fact-check) over the diff vs `ceb1eae05ea0`. `--no-elegance` (the `/simplify` pass already ran this tree).

**Pass 2 — fact-check: clean.** No silent error-swallowing, no unjustified fallbacks, no off-by-one/wrong-operator logic, no unbounded collections. The one `.catch` (`reconnectAction`) surfaces `err.message` via `toast.error`; the deadline reset is correctly gated on the RPC *resolving* (`.then()`), and a rejected reconnect leaves the anchor untouched (pinned by a test). `campaignAnchors` is bounded and pruned/reset in lockstep with `anchors`. `bootStalledPhaseDetail`'s switch is exhaustive over the real `ConnectPhase` union.

**Pass 1 — rules: 1 finding, fixed.**

| Rule | Finding | Action |
| --- | --- | --- |
| `prefer-ts-pattern` | `escapeSurface` dispatched on the `StalledLeg` union via a 3-way `if`-cascade with a multi-field first arm — inconsistent with `resolvePrecedence` in the same module, which already uses `match(...).with(...).exhaustive()`. | **Fixed** — rewrote as `match({ leg, isLocalHost }).with(...).exhaustive()` (28de0f936), so a future `StalledLeg` must be handled or fail the build. |

All other rules (dry-rule-of-three, invalid-states-unrepresentable, no-dead-code, no-unbounded-growth, no-overloaded-null, toast conventions, solid reactive-prop rules, …) passed — the diff notably *fixes* an `invalid-states-unrepresentable` case by replacing the flat `{ leg, phase }` boot-stalled payload with the discriminated `BootStalledRecovery` union.

srid:
## Evidence

The two boot-stalled card shapes this PR introduces (#1908 D2), rendered from the real production `BootStalledCanvas` (with `../wire` stubbed so the surface stack doesn't boot — the card is a state no e2e scenario reaches).

boot-stalled connector and client cards

- **Top — `via: "connector"`** (a warming-remote campaign still retrying, phase `building`): the card is now **non-terminal** — *"kolu is still setting up this host … it is still retrying"* with a live *"Still building on the host…"* detail, and the recovery verb is **[Retry connection]** (which recycles the server-side connector via `client.hosts.reconnect`). This replaces the old **lie** — the #1898 card showed *"This host's agent isn't responding — Reload"* (a terminal message + a browser reload that can't touch the server dial) over a connector that was still retrying.
- **Bottom — `via: "client"`** (a genuinely client-side leg — here `session`): unchanged honest behavior — *"Your workspace didn't finish loading … Reload to try again"* with **[Reload]** (`location.reload()` re-runs the client-side subscription).

Both offer **[Switch to local]** on a remote host. (The R8a campaign backstop and the deterministic Retry-connection deadline reset are pure timing logic with no distinct on-screen surface of their own — they gate *when* the connector card above appears/dismisses; covered by unit tests.)

PR #1926, opened 2026-07-21
feat(attention): jump-to-agent pill + a loud "blocked" pip

srid:
When an agent stops to ask you something — a Claude Code `AskUserQuestion`, an `ExitPlanMode` — kolu already notices: the host tab grows a small amber count, and the agent's dock entry gets an orange mark. The signal exists, but it's weak on both ends of the question you actually ask:

- **Which agent is asking?** The host-tab pill tells you *how many* terminals wait, never *which* — you scan the dock for the orange mark by eye.
- **Is this important?** The "asking" mark was the **faintest pip in the whole palette** (a 6px dot at 55% opacity), while the lower-stakes *unread* badge pulses. The loudness hierarchy was inverted.

Both are fixed here with facts the client **already receives** — no new wire field, no new sensing.

## AD1 — the awaiting pill becomes a jump button

The host's `urgency` cell already carries the *identities* of the waiting terminals (`awaitingIds`), not just a count — the UI just threw that away and projected `.length`. So AD1 is pure wiring:

- `useHostAwaiting` now returns the ids; its three readers use `.length` for the badge.
- `HostAwaitingPill` gains an optional `onActivate` and renders as a real `<button>`. In `HostChip` it moves **out** of the `role="tab"` select button to sit beside it as a sibling (the same shape the remove-`✕` already uses), so we never nest a button inside a button.
- Clicking runs `setActiveHost` + `tileStore.activate`, **cycling** through `awaitingIds` on repeated clicks — the *exact* switch-then-activate seam the cross-host notification click already uses, so a badge click and a notification click focus the terminal the same way.

## AD2 — a genuinely-blocked agent gets its own pip

kolu's paint fold deliberately collapses two states into one amber dot: an agent **blocked on your answer** (`awaiting_user`) and one that **just finished its turn** (`waiting`, the post-turn lull). The lingering dim glow for a just-finished agent is correct — the bug is that the agent asking you a question *right now* whispered at the same volume.

AD2 splits only the loud end back out:

- A new `blocked` pip variant in `@kolu/solid-statepip` — full-opacity, larger, pulsing — versus the dim lingering `awaiting` dot, unchanged.
- Promoted through the existing `agentBucket` fence (not a hand-rolled state check), so a new agent state can't slip past.

**Why not split the shared `AgentPaintClass`?** Because that's the wrong-sized cut — it's read by the tile aura, minimap, *and* the expanded-switcher columns, so a new class there forces a new switcher column and a louder aura on every tile. Keeping `blocked` **dock-local** (a member of `DockRowBucket`) plus the tile-title pip touches exactly the two surfaces in the complaint. The parity invariant — *dock row pip ≡ tile-title pip* — still holds, because both promote through the same `agentBucket` check (pinned by the ranking parity test).

## Out of scope: surfacing the question *text*

The natural next wish — showing the question itself in the dock subtitle — is scoped in the Atlas note and **deliberately deferred**. The wire plumbing is cheap, but the text isn't anywhere kolu can cleanly read while a question is pending (the Claude Agent SDK buffers the tool-use block off-disk until you answer), so it needs a brittle, version-sensitive per-agent screen-scraper — or a Claude Code hook, which is its own new integration mechanism. That belongs in a separate pass.

## Plan of record

📄 [Atlas: Agent Attention — Which One Is Asking?](https://kolu.dev/atlas/agent-attention-discoverability.html) (`docs/atlas/src/content/atlas/agent-attention-discoverability.mdx`)

## Testing

- Typecheck + biome clean across the workspace (`just check`).
- Unit: `@kolu/solid-statepip` pipVariant (blocked look + reduced-motion + louder-than-awaiting), dock `pipVariant` (blocked→blocked), `dockRowRanking` (awaiting_user→`blocked`, waiting→`awaiting`, and the updated row-pip ≡ tile-title-pip parity). Full client suite green (821 passed).
- Atlas `check-sync` idempotent.


srid:
Closing in favour of a design-first pass. The jump-pill + amber blocked-pip here are real, but the underlying problem is that notification *policy* is split across modules by the active-vs-background-host axis (useTerminalAlerts vs useHostAttention vs the active-only unread ledger), so background `waiting` finishes behave differently from active ones and it's hard to reason about. Stepping back to define one host-agnostic attention model first — Atlas note incoming — then re-doing the surfacing on top of that. Branch `sober-beak` kept for reference.

Issue #1904, opened 2026-07-20
kolu MCP needs a standing subscription verb — the coordinator's watcher choreography drops worker reports at every seam


srid:
## The incident chain (2026-07-20, one session)

A coordinator agent drives worker agents in kolu terminals. Its only 'notify me' primitives are the blocking `wait_agentState`/`wait_outputSettled` MCP calls, so it hand-rolls subscriptions: background watcher agents that loop the waits and report back. Today, in ONE session, that choreography dropped reports at four distinct seams:

1. A kaval restart re-keyed every terminal id → all watchers died pointing at dead ids.
2. `wait_*` returned `closed` for terminals that were **alive** (a server/padi blip closed the subscriptions while kaval survivors kept the PTYs) → watchers falsely reported worker death; a verify-with-screen_text rule had to be bolted on.
3. A watcher legitimately *completing* (delivering one report) left the lane unwatched unless the coordinator remembered to re-attach — it forgot once, and **a worker's MERGE-READY report went into the void** until the operator noticed by hand.
4. Operator restarts of the coordinator killed all watchers silently.

The operator's earlier design question — *'doesn't kolu MCP have a subscription for terminal activity changes?'* — was the correct architecture review. The answer today is no: the resources exist (`surface://collections/terminals` etc.) but no push channel reaches an MCP agent, so every coordinator rebuilds this fragile layer.

## What to build (either or both)

**(a) A standing subscription verb on the MCP face** — e.g. `watch_agentState(id | all, until-buckets)` that streams transition events (or MCP resource-subscription notifications on the `terminals` collection), surviving re-subscribes across daemon restarts by key, not by terminal instance.

**(b) The supervision-edge delivery from the campaign-surface design** (docs/atlas: campaign-surface, attention-flow class-kill): the `parentId`/creator edge IS the subscription — padi delivers a worker's settle event (agent-state ∈ {awaiting, waiting} AND raw output settled — the conjunction both signals already shipped as wait tools) into the supervisor's mailbox. No watcher processes at all; the structure is the listener.

(b) is the class-kill (no coordinator can forget to listen — the edge subscribes at spawn); (a) is the general-purpose face verb. Both eliminate the four failure seams above.

## Grounding

- The wait tools + the 'idle+settled conjunction' rationale: kolu mcp (#1865) and the campaign-surface note's attention-flow section.
- Watchers' false-death and void-report evidence: this session's coordinator transcript (operator-witnessed).

[...]

PR #1930, opened 2026-07-21
docs(atlas): state-isolation plan of record + pre-2.0 board catch-up (#1911 closed after review)

srid:
Two connected pieces, both fallout from today's adversarial review of PR #1911.

**A new note — `state-isolation`** — is the plan of record for rebuilding the #1334 fix lock-only, ratified by srid after #1911 was closed. It leads with an architecture diagram (three binder entry points → one guarded resolver → bind or crash; kaval explicitly out of scope) and pins two sequenced work items with track ids: **SI1** extracts the one genuine treasure from #1911 (padi leaks its boot gate on a startup throw — a real pre-existing bug) as its own small PR; **SI2** rebuilds the lock in padi + server + nix at ~half of #1911's size, with done-criteria the old PR proved necessary: `just e2e-ssh` run once and passing (CI has no ssh — #1911 shipped 38/38 green with both ssh lanes broken), a test that the production marker actually gets stamped, and docs that claim exactly what a bind-time lock guarantees. The write-only role stamps ship with their reader in #1912.

**The pre-2.0 hardening board catches up to the tree**, each row verified at the forge:

| Item | Was | Is |
| --- | --- | --- |
| #1334 row | "PR #1911, 38/38 green, merge decision open" | #1911 closed after review; plan → state-isolation note (SI1 → SI2); wire verbs → #1912 |
| #1375 (vitest forks 30+ daemons) | 🔶 in flight | ✅ shipped — PR #1921, issue closed |
| #1754 (dock stuck "thinking") | 🔶 CI finishing | ✅ shipped — PR #1914 (`Refs` not `Closes` by design: issue waits on the #1751-coupled e2e asserts) |
| Coordinator handoff (2026-07-20) | two stale/resolved decision rows | rewritten as **Standing facts**: CI venue law, the watcher era (#1904), standing law |

Docs-only; `atlas::check-sync` green (dist idempotent).


srid:
## Review: this plan is overengineered for a path check

The *incident class* is real. The *document* is Claude Fable theater built around a lock that should be ~50 lines.

### What the problem actually is

A daemon’s identity is the digest of its state-root. Production lives at `~/.local/state/padi`. `resolvePadiStateRoot()` still defaults there. So a bare `pnpm dev` / `padi` can compute production’s identity and then adopt/kill it.

Lock 3 already shipped (#1921 / `daemon-test-gate`). The bare-launch hole (#1334) is still open. That part is not dogshit.

### Where the plan goes off the rails

#### 1. It invents a credential system for a path check

The useful rule is one sentence:

> Non-production may not bind production’s state-root. Crash before open.

Everything else is machinery in search of a threat model:

| Plan artifact | Why it smells |
| --- | --- |
| `role.ts` “~250 lines” | 250 lines for “is this the default root, and am I allowed?” is the tell. Real code should be a small guard in/near `stateRoot.ts`. |
| Persistent `production` marker on disk | Solves “dev explicitly points at a *relocated* production root.” That’s not the #1334 incident. It’s a second system (write, fail-closed read, corrupt/EACCES, stamp-on-boot tests). |
| `KOLU_ROLE` as authority | Reasonable as a *wrapper-baked fact*. Overweight as a first-class role subsystem with threading, `--role`, once-at-entry values, ssh flags, daemon-to-daemon forward. |
| Relocated-config refusal | Fine edge case; does not justify half the table. |
| Three refuse paths + realpath-to-ancestor | Symlink hardening is real; still belongs in ~30 lines of path comparison, not a RoleRead algebra. |

The three-lens “keep `KOLU_ROLE`” section is pure process cosplay. Location ≠ authority is a true sentence; it does not need refuter-verified epistemology.

#### 2. SI1 as a campaign phase is process inflation

“Padi should release its gate if boot throws” is a good bug fix. It is not “SI1 of the State Isolation track.” It does not need a plan-of-record, a track prefix, or its own full `/be` arc *because* SI2 exists. Ship it as a normal small PR if you care; don’t scaffold a program around it.

#### 3. Deferring wire verbs is the one smart cut — and the plan still over-narrates it

#1911 died partly because it shipped writers without readers (kaval stamp machinery for a guard that wasn’t in the PR). SI2’s “zero kaval, bind-time only” is the correct rescope. That lesson fits in one paragraph. The rest of the footnote museum (“11 hunters,” “~52% of 1,483 lines,” “path-not-feature rule”) is resume padding for the agent that wrote it.

#### 4. The docs are already lying to each other

`host-isolation-locks.mdx` still reads as if Lock 1 landed with #1911 and “what you feel afterwards.” `state-isolation.mdx` is the plan to rebuild after #1911 was closed. Same day, two truths. Classic Fable: more notes than code, and the notes disagree.

### The design that would actually be enough

In priority order:

1. **Make first-party paths private by construction**  
   `pnpm dev` / worktree scripts set `KOLU_PADI_STATE_DIR` to `.kolu-dev/padi`. Tests already isolate. That kills the day-to-day hole without any role vocabulary.

2. **One bind-time refuse**  
   At the one place roots are resolved (server boot + padi `daemonMain` + stdio bridge): if resolved root is the default (or same identity after realpath) and `KOLU_ROLE !== "production"`, throw a one-line error naming `KOLU_PADI_STATE_DIR`.  
   Production nix wrapper: `--set KOLU_ROLE production` (or just always pass the explicit root from the wrapper — even simpler).

3. **Optional, later**  
   Persistent marker for relocated prod roots, wire-verb guards (#1912), relocated-config collision — only if you still have a real residual after (1)+(2).

That’s SI2’s *table row that matters*, without the Role module novel.

### What I’d keep from the note

- Bind-time only; no kaval in this PR.
- Don’t ship stamps before a reader exists.
- Honest README (refuses *binding*, not adopt/SIGTERM).
- Force `e2e-ssh` as a local done criterion (CI has no ssh) — that’s a real scar from #1911.
- `selfRole` once at entry if you keep the env fact.
- SI1 gate-release if it’s a real stuck-gate bug (as a normal fix, not a “track”).

### Verdict

| Layer | Assessment |
| --- | --- |
| Problem (#1334 bare default → prod identity) | Real |
| Bind-time refuse + private dev roots | Right size |
| `role.ts` + markers + role threading + SI track + three-lens justifications | Overengineered Fable residue from a PR that already failed for bulk |
| SI1 as phase zero of a campaign | Cosplay |
| #1912 deferred | Correct |
| Atlas status `accepted` | Premature — this is still a design argument, not settled law |

**Smell diagnosis:** the agent closed a bloated PR under adversarial fire, then wrote a plan that *looks* disciplined (half size, lock-only, sequenced) while preserving the same conceptual bulk — a Role Product instead of “don’t sit in production’s chair.”

**One-line ruling:** accept the refuse-at-bind rule; reject the role subsystem as written; demote the plan until the PR is ~100 lines of guard + wrapper + tests, not a 250-line role module with a pedigree.

srid:
## On `KOLU_ROLE`: we don't need the env var

I'd prefer not to add that env complexity. The simpler path is enough.

### You already have the right pattern for server config

- **No default.** `KOLU_STATE_DIR` unset → crash.
- Production wrapper sets it (`:-` so #1414 can relocate).
- `pnpm dev` sets a worktree path.
- Tests set a temp path.

Padi is the odd one out: **`resolvePadiStateRoot()` still defaults** to `~/.local/state/padi`. That's the hole. Not "missing role badge."

And first-party dev is already doing the simple thing:

```text
KOLU_STATE_DIR=…/.kolu-dev
KOLU_PADI_STATE_DIR=…/.kolu-dev/padi
```

(in `packages/kolu-cli`'s `dev` script)

### What "always pass the explicit root" means

Mirror server:

1. **Require** `KOLU_PADI_STATE_DIR` / `--state-root`. No silent default.
2. **Production wrapper** exports it, same shape as state dir:
   `KOLU_PADI_STATE_DIR="${KOLU_PADI_STATE_DIR:-$HOME/.local/state/padi}"`
3. Dev / tests keep setting their private paths (already mostly true).
4. Bare `padi` without a path: one-line crash, not a secret trip into production.

Then **`KOLU_ROLE` does no work.** The wrapper isn't a credential; it's the thing that *supplies the path*. Accidental default becomes unrepresentable.

### What the env-var plan was buying (and why you can drop it)

| Claimed need | Without `KOLU_ROLE` |
| --- | --- |
| Bare launch mustn't hit prod | No default → can't |
| Dev mustn't sit in prod's chair | Dev already passes private root |
| "Location ≠ authority" | Bind isn't authority; same-uid can always kill prod. The bug was *silent* path choice |
| Relocated prod (#1414) | Second instance sets both dirs explicitly; refuse half-set if you want |
| Explicit override to prod path | Deliberate; same footgun as `rm` on the real tree |
| Markers / stamps / role threading | Only needed if you invent a second identity system |

### When you'd still want a badge

Only if the threat is: **someone intentionally points at prod's path** and you want a second door they must open. Same uid can SIGTERM the live daemon anyway. That's not #1334. Don't build for it in the first PR.

### Bottom line

- **Keep:** explicit padi root everywhere; wrapper always sets it; fail-fast if missing.
- **Drop:** `KOLU_ROLE`, `role.ts`, persistent markers, `--role`, stamp/read algebra.
- **Optional later:** "production wrapper set both or refuse" if relocated config without a padi root is a real footgun.

`KOLU_ROLE` is a parallel authority story. You already have the authority story that works: **the path is chosen out loud, or the process dies.**

srid:
Accepted — both comments. The plan is rewritten to the no-default shape (commit ad46d3c5).

**On the follow-up (drop `KOLU_ROLE`): conceded.** My "keep the env var" ruling compared the wrong alternative — deriving role from the config dir — not *deleting the default*. Two mechanics verified at the tree make your shape complete: the PTY spawn allowlist wholesale-scrubs `KOLU_*` from hosted shells (`integrations/pty/src/shell.ts`, `koluInternalEnv`), so the wrapper's export can't leak into a terminal kolu hosts; and the remote `padi` runs from a nix-realised closure the binder ships (`surface-remote/src/dialAgentOnce.ts`), so the same wrapper supplies the remote default remote-side — the reason I thought a wire signal was structurally required evaporates. Dropped: `KOLU_ROLE`, `role.ts`, markers, `--role`, all threading. Residuals named in the note as accepted/deferred: deliberate explicit pointing (out of the accident model), remote-dev binder reaching the remote default (today's behavior; #1912's territory), half-relocated production (optional later).

**On the first comment:** marker system → deferred; SI1 demoted to a plain bug fix (listed only for ordering); footnote pedigree trimmed; status demoted to `proposed`; and the docs contradiction was real — `host-isolation-locks.mdx` said "#1911 **is** three locks" with the PR closed; fixed to a design-record framing pointing here.

What survived your review because it's load-bearing, not residue: `just e2e-ssh` as a run-once done-gate (CI has no ssh), zero kaval, no writers without readers, honest README language, and the canonical-identity point folds away entirely — with no default in code there is nothing to alias to.

PR #1931, opened 2026-07-21
feat(attention): one owner for agent attention across all hosts

srid:
Consolidates kolu's agent-notification logic into **one owner** and standardizes the vocabulary on **attention**.

Before, the rules for reacting to an agent (sound, OS popup, app badge, dock unread, host-tab mark) lived in **three** modules split by an accidental axis — *whether you were looking at the host*: `useTerminalAlerts` (active host only), `useHostAttention` (other hosts only), and an active-only unread ledger. So a finished agent on a background host reached you differently than the same agent on the host you were viewing, and it was hard to reason about.

## What changed

- **One owner** — new `packages/client/src/attention/useAttention.ts` replaces `useTerminalAlerts` + `useHostAttention` + `useActivityAlerts` (all deleted), wired once in `App.tsx`.
- **Same rules on every host** — attention is read cross-host off each host's tiny `urgency` cell, now carrying `finishedIds` alongside `awaitingIds`. The only place "which host" enters is the `watched` predicate (a tile you're staring at is "seen"). Falling out of the uniform rule: **a background host's finished/awaiting agent now pings + shows a quiet host-tab mark**, exactly like the active host's.
- **Pure, tested core** — the genuinely-blocked-vs-finished, fire-once/escalation decision (the #1177-class logic) is extracted as the wire-free, unit-tested `attentionTransitions`.
- **Terminology** — the `activityAlerts` preference is renamed `attentionAlerts` (with a conf-ladder migration so existing settings carry forward); the setting label is now "Attention alerts".
- **Delivery unchanged** — still rides the shared `attentionNotify` service-worker seam.

## Plan of record

📄 [Atlas: Agent Attention — One Model, One Place](https://kolu.dev/atlas/agent-attention.html) (`docs/atlas/src/content/atlas/agent-attention.mdx`)

## Testing

- `just check` green: typecheck across the workspace + biome (1456 files).
- Unit: client suite (816), `recomputeUrgency`/`urgencyEqual` (8), new `attentionTransitions` (6), preferences migration (17).
- **Not yet run:** full two-platform CI and e2e — including the existing agent-alert e2e scenarios, which exercise the rewritten `__koluSimulateAlert` path. This is a deliberate **testbed** PR: deploy it, see the behavior, then authorize the review gauntlet + CI.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1938, opened 2026-07-22
Hold the "finished" notification until the terminal goes quiet

srid:
**Some agents (Claude Code among them) mark their turn _finished_ while background sub-agents keep working** — the transcript says done, but the terminal is still emitting bytes, so kolu would chime and drop the "finished" dot before the work actually stopped. This gates the _finished_ signal on the terminal itself: a `waiting` agent enters the attention model's `finishedIds` only once its PTY has gone **quiet for ~5s**. A _needs your input_ prompt is never held back — it's blocking, so it fires at once.

All the machinery is padi-side — the daemon sees every terminal's bytes, active or background, so it (not the browser) can own this cross-host fact. It's the "effective finished" follow-up the #1931 note recorded, now shipped.

### How it fits together

```
commit firehose ─▶ agent-bucket edge ─┐
(never misses an episode)             ├─▶ finishGate ─▶ settledFinished ─▶ recomputeUrgency ─▶ finishedIds
byte tap (waiting terminals only) ────┘    (debounce)     (reactor source)
```

- **`finishGate`** — a pure, transport-injected debounce core: byte-tap _only_ the `waiting` terminals (bounded — not `liveActivity`'s tap-every-active), settle each after a quiet window, publish the settled set into the reactor graph.
- **The episode edge** — a synchronous agent-bucket signal off `commitSnapshot` that fires on every state transition, so it never misses a fast `waiting→working→waiting` blip the 1s reconcile poll would, and a new turn earns a _fresh_ window instead of inheriting the last one's settle.
- **Reuse** — the timer machinery is extracted from `liveActivity` into a shared `terminalActivityTracker`; `recomputeUrgency` gains one required `settledFinished` argument.

### Ordering bugs the review gauntlet caught

The gate is a timing-sensitive state machine, and a 5-round codex ⇄ Claude debate surfaced three real happens-before defects that architecture + lens review missed — each would have defeated the feature in its target case:

| # | The hole | Fix |
| --- | --- | --- |
| **F1** | A settled turn carried over a **poll-missed** `waiting→working→waiting` blip → a premature finish for turn 2 (the exact background-sub-agent case) | the synchronous agent-bucket edge drops the stale settle on leave and re-arms a fresh window on re-entry |
| **F3** | A **pending or dropped** tap could settle a terminal that _nothing was observing_ | settle gated on an affirmative `attached` fact (set only by a real `onReady`/output) + a per-tap generation fence |
| **F5** | Leaving `waiting` mid-window fired a **transient false finish** — a synchronous `tracker.forget` ran the settle listener before the fence was set | set the episode fence _before_ `forget` |

Also fixed pre-debate: a dropped byte-tap that never re-opened (architecture review) and a `dispose` complected with the reactor source's subscriber refcount (lens).

### Docs

Folds in this branch's original scope — the #1931 note compaction + follow-up recording: `agent-attention.mdx` flips the "effective finished" follow-up to shipped, `notifications.mdx` + the changelog record the user-facing debounce, and the gate's six-collection lifecycle is itself recorded as a "model it as one discriminated `TapState`" follow-up.

> **No visual diff — this is timing-internal.** The observable effect (a delayed finish notification) can't be captured by the e2e harness, which drives a simulate hook that _bypasses_ the gate; the proof is 13 `finishGate` unit tests over the timing edges plus production-glue coverage. See the Evidence comment.


srid:

✅ **Consensus** with zero debate turns · lowy + hickey · base `3202f142953c`

Settled: 2 reconciled (raised by both lenses) · 4 unopposed

### Applied (1)
- `hickey-1` dispose() teardown complected with the source's sole-subscriber refcount — commit `ce5382b1d`

### Agreed — no change (5)
- `lowy-1` Electricity-test grounding (Löwy lens summary — required output) (packages/padi/src/finishGate.ts:72)
- `lowy-2` ≡ `hickey-2` Byte-tap primitive duplicated across the two tap consumers (packages/padi/src/padiFinishGate.ts:51)
- `lowy-3` TerminalLocation type alias + comment duplicated (packages/padi/src/padiFinishGate.ts:27)
- `lowy-4` ≡ `hickey-3` Two id-set-equality helpers on two representations (packages/padi/src/finishGate.ts:59)
- `lowy-5` Reconcile-trigger asymmetry between the two sibling tap modules (packages/padi/src/finishGate.ts:195)

srid:
## Codex ⇄ Claude debate

|---|---|---|
| 1 | fbdd8e53 | fix(codex): ready-gate + sticky-settled finish gate (F2, F3); doc-sync (F4) |
| 2 | 6d44d1a0 | fix(codex): episode-safe finish gate via agent-state edge (F1); tap-gen fence (F3) |
| 3 | c1d13bea | fix(codex): gate settle on real attachment, not manufactured readiness (F3 r3) |
| 4 | 582f8708 | fix(codex): fence settle loop before tracker.forget on leave (F5 r4) |

- **F1** — Tracking is keyed only by terminal id, so a settled result survives across waiting episodes when the 1s poll misses the intervening non-waiting state
- **F2** — When a tap ends after the terminal has settled, onTapClosed returns without removing or reopening it
- **F3** — The quiet timer is seeded before openTap has established the asynchronous attachment
- **F4** — The module header still presents recomputeUrgency($.terminals()) as the exact production wiring and says only the terminals edge drives the cell, even though the function now requires settledFinished and the gate source is a second reactive dependency
- **F5** — New regression: the leave path calls tracker.forget(id) before awaitingRearm.add(id), but ActivityTracker.forget synchronously removes the live flag and notifies its listeners

srid:

Rule checklist + fact-check over the effective-finish-gate diff (elegance skipped — `/simplify` already ran this tree). **3 rule findings + 1 fact-check finding; 3 actioned, 1 recorded as a follow-up.**

| Pass | Rule / check | Finding | Action |
|---|---|---|---|
| Rules | `prefer-ts-pattern` | `urgency.ts` dispatched `agentBucket`'s union with an `if/else if` + a `settledFinished` conjunct, so a future 5th bucket would silently fall through | Rewrote as an exhaustive `switch` with a `satisfies never` default (matching `agentBucket`'s own fence) — a new bucket is now a compile error here (commit `e80695a4`) |
| Rules | `errors-must-log-at-error` | `padiFinishGate.ts`'s byte-tap catch logged a real thrown fault at `debug` (a graceful stream end never throws — it exits the loop) | Bumped to `log.warn` (degraded-but-recoverable), noting the re-tap recovery (`e80695a4`) |
| Rules | `invalid-states-unrepresentable` | `finishGate.ts` tracks each terminal across six parallel keyed collections whose invariants are held by call-order discipline | **Recorded as an Atlas follow-up**, not rewritten: the state is a genuine *product* + per-tap metadata (not a clean single discriminant), and the current form is quiet-window-timing-sensitive, fully unit-tested (incl. the F1/F3/F5 edges), and freshly hardened through this review — so the discriminated-`TapState` rewrite is a follow-up to weigh, not churn to force now |
| Fact-check | test coverage | The **production episode-edge glue** (`commitSnapshot`'s bucket-transition diff + `padiFinishGate.listWaiting`'s active+waiting filter) had zero direct coverage — the gate *core* is tested only against fakes, so a drift there would silently degrade the fast-blip guarantee with nothing in CI to catch it | Added coverage: a `metadata.test.ts` case asserting `commitSnapshot` publishes the agent-bucket transition **only** on a `waiting` crossing (not every commit), and a new `padiFinishGate.test.ts` over the extracted pure `selectWaitingTerminals` filter (`e80695a4`) |

**Fact-check (logic):** no silent swallows, no unjustified fallbacks, no unbounded growth, no off-by-one/operator/shadowing errors. The reentrant path (`publishAgentBucket` fires synchronously inside `commitSnapshot`, before `publishComposedTerminal`) was traced in both bucket directions — every transient frame it can produce is consistent with the eventual correct state.

The bounded-`waiting`-only tap scope, the required `settledFinished` param, the DEFAULT-EXCLUDED mechanism, the loud-log catches, and the synchronous registry were treated as settled by the earlier architecture + lens + codex passes and not re-litigated.

srid:
## Evidence

This change has **no visual UI diff** — it's a padi-internal *timing* correctness feature. Its only user-observable effect is a **delay**: a "finished" notification/dot is now held until the terminal's PTY has actually gone quiet (~5s), so an agent that marked its turn `waiting` while background sub-agents keep emitting no longer fires a premature finish. The ASKING path is unchanged (fires at once).

That behavior can't be honestly captured by a recorded e2e clip: kolu's attention e2e drives the `__koluSimulateAlert` hook, which **bypasses** the detect → fire → gate path this PR adds (the exact "test the real path" gap the Atlas note records as a follow-up). A clip of the simulate hook would prove nothing about the gate. So the proof here is the test suite and the review, not a video:

**Unit coverage of the observable timing behavior** (`packages/padi/src/finishGate.test.ts`, 13 cases with fake timers):
- settles only after attach is READY **and** a full quiet window — never while the attach is pending;
- output inside the window re-arms the debounce; output *after* settling un-settles it (resumed background work);
- a dropped tap is re-opened and re-observed — a still-working terminal is never settled off a dead tap, and an already-settled one keeps its level across the gap (no flicker/double-chime);
- **no transient false finish** when an attached terminal leaves waiting mid-window (recorded via an eager reactor effect over the settled level);
- a NEW waiting episode earns a fresh window — a settled turn does not carry over a poll-missed `waiting→working→waiting` blip (the exact background-sub-agent case);
- a stale tap generation's late `onReady` is fenced from a newer tap.

**Production-glue coverage** (added in the police pass): `metadata.test.ts` asserts `commitSnapshot` publishes the agent-bucket transition **only** on a `waiting` crossing; `padiFinishGate.test.ts` pins the `active` + `waiting`-bucket membership filter.

**Review**: the timing edges (F1 episode-safety, F3 settle-without-observer, F5 transient-on-leave) were each surfaced and fixed under a 5-round codex ⇄ Claude debate to consensus — see the codex comment above. Full padi unit suite (346) + `just check` green; two-platform CI green.

srid:
## Regression found + a re-architecture (activity ownership moves to kaval)

### The regression this PR introduced

**Visiting a finished terminal, then leaving, re-fires its "finished" notification with zero agent activity.** Traced end-to-end in code:

1. Revealing/switching to a tile re-fits xterm → `publishDimensions` (`Terminal.tsx:206`) resizes the server PTY → SIGWINCH → the program **repaints** — a genuine delta on the attach stream. The browser mutes this for its own live-dot (`activity.suppress`, `Terminal.tsx:216`, `RESIZE_ACTIVITY_SUPPRESS_MS = 600`).
2. The finish gate's byte tap (`padiFinishGate.openTap`) is an **independent** padi-side attach with **no** such mute → the repaint hits `onOutput` → `settled.delete(id)` → the terminal leaves `finishedIds`.
3. The client sees it leave the attention class → `attentionCore.ts:78` clears its fire-once latch.
4. You navigate away; ~5s of real quiet later the gate re-settles it → it **re-enters** `finishedIds`.
5. A re-entry is a *fresh* finish (`attentionTransitions.ts:54`, in neither prev set) → `deliver` re-fires (sound + OS popup + `markUnread`) and `nextUnseenFinished` re-lights the host-tab dot.

Before this PR a `waiting` terminal sat in `finishedIds` **statically**, so a repaint couldn't dislodge it. The gate made `finishedIds` react to *any* PTY byte — including the resize repaint the client already knew to ignore.

### The real problem: "is this terminal producing output?" is computed three times

- **browser** `useTerminalActivity` — counts bytes off the attach stream, powers the Dock/tile dot, carries the 600ms resize-mute hack.
- **padi `liveActivity`** — its own byte taps for the `activity` stream. No mute.
- **padi `finishGate`** (this PR) — *another* set of taps for the finish decision. No mute.

Three independent byte-counters, three idle timers, and only one knows a resize repaint is fake. That's exactly why the mute knowledge didn't travel — it lives as a hack in one consumer instead of a fact at the source.

### The cost of consolidating — measured, not guessed

Per 4KB output chunk, on the dev box: JSON frame encode+decode **5.3µs**, unix-socket one-way **1.8µs**, tracker re-arm **0.24µs** → **~7.5µs/chunk**. At 16 active terminals: a typical busy host is **~0.2% of one core**; a pathological 50MB/s blaster is **~10% of one core**. Negligible. My earlier "real tap cost" worry is **retracted**.

But the measurement makes the right design obvious: **7.3 of the 7.5µs is serializing + shipping bytes to another process only to discard them and keep a boolean.** The 0.24µs is the actual logic. And **kaval owns the PTY** — it already touches every byte *and* handles every resize.

### The fix (this PR, no deferral)

- **kaval becomes the single activity authority**, at the byte source: computes "output flowing / quiet" per terminal, resize-aware (it natively knows "I just resized this PTY; the next burst is a repaint").
- The wire carries **transitions** ("terminal X went live / went quiet"), not every chunk — a couple of tiny frames per episode.
- padi's `finishGate`, padi's `activity` stream, and the browser's Dock dot all become **consumers** of that one fact. The three byte-counters and the client's 600ms suppress get deleted.

This touches the kaval↔padi contract (a version bump) — the honest scope of doing it right.

srid:
## Feedback on the re-architecture (from Grok)

### What kaval should own

Agree that **activity ownership should move to kaval** (it sees every byte *and* every resize), that the wire should not re-stream chunks just to maintain a boolean, that the client's 600ms suppress should die as a product of the signal rather than app lore, and that a contract bump is honest if the surface changes.

Pushback is on the **product grain**. “Live / quiet” is not one fact — it is **chronology + a window policy**:

| Consumer | Window / meaning |
| --- | --- |
| Dock / tile live-dot | short idle (~`TERMINAL_IDLE_AFTER_MS`) |
| Effective finish | ~5s quiet |
| (future) other | could differ again |

If kaval emits a single `live | quiet` with **one** internal idle timer, consumer policy has been pushed into the producer. Finish and the live-dot will fight over the window, or kaval grows N windows, or someone reintroduces a client-side override. That mixes policy volatility into byte-observation volatility, and complects two concepts as one boolean.

**Prefer this contract instead:**

1. **Kaval owns meaningful-output chronology**
   - On real PTY output: update `lastOutputAt` (or emit a tiny `output` edge).
   - On resize-driven repaint: **do not** advance that chronology (or tag the burst so consumers must ignore it).
   - Kaval is the only place that can do this without a 600ms guess: it knows “I just resized this PTY.”

2. **Consumers own windows as pure functions**
   - live-dot: `now - lastMeaningfulOutputAt < IDLE`
   - finish: `waiting ∧ now - max(enteredWaitingAt, lastMeaningfulOutputAt) ≥ FINISH_QUIET`
   - The wire can still be transition-shaped (e.g. “meaningful output at T”), but the **level** is a timestamp (or event log), not “quiet for my preferred W.”

3. **Finish is a fold, not a second activity subsystem**
   Even after activity moves to kaval, `finishGate` as a six-set machine should still go away. `finishedIds` should be pure over agent episode + meaningful chronology + now, with a deadline wake. Un-settling on *real* post-finish output stays expressible; un-settling on resize becomes unspellable because resize never moved `lastMeaningfulOutputAt`.

So: agree with “activity ownership moves to kaval,” but spell the product as **resize-aware meaningful-output chronology**, not **kaval’s live/quiet state machine**.

### How this relates to the finish feature specifically

Finish is not “just another activity consumer.” It needs **two** facts:

1. Meaningful PTY chronology (kaval)
2. Waiting-episode boundary (`enteredWaitingAt` / leave waiting) — agent/commit path

Resize only explains the **new** regression. Episode inheritance (F1) and default-excluded settle are still agent×time, not activity alone. A kaval activity stream that `finishGate` merely subscribes to is a big improvement but can still leave a smaller gate if episode is not folded into pure policy.

I’d amend the “The fix” bullets to something like:

- **kaval:** meaningful-output chronology (resize-aware)
- **padi:** stamp waiting episode; pure effective-finish fold; derive `activity` from chronology
- **client:** Dock dot from the same fact; delete suppress hack
- **delete** finishGate byte taps and parallel counters

— Grok (xAI)

srid:
Closing without merge.

This PR tried to land effective-finish + kaval activity ownership + liveActivity rewire + client live-dot mirror in one shot. The *design* (resize-aware meaningful-output chronology at kaval; finish as a pure fold on padi-local stamps; windows stay consumer policy) is right; the *delivery* mixed four volatilities and shipped two user-visible defects (visit→re-chime on the first design; sticky “all terminals live” on the client mirror).

Plan of record lands as a separate Atlas note (track **EF**), phased so each PR is mergeable alone:

- **EF0** — close/capture (this)
- **EF1** — kaval activity edge only (no consumer behavior)
- **EF2** — effective finish pure fold (first user-facing fix)
- **EF3** — padi `activity` stream on the edge
- **EF4** — client live-dot mirror (with sticky-live fence)
- **EF5** — shared fan-out / renames / e2e

Do not resurrect this branch wholesale; implement from the Atlas note off master.

PR #1943, opened 2026-07-22
Highlight Rhai files in the Code tab

srid:
**The Code tab now recognizes `.rhai` files and syntax-highlights them in source and diff views.** Rhai previously fell through to plain text because Shiki does not bundle a Rhai grammar.

Kolu registers Rhai through Pierre's custom-language API and keeps the grammar lazy-loaded with the rest of the highlighting workers. **The official TextMate grammar is pinned from `rhaiscript/lsp` through npins and extracted by Nix**, so the upstream source and license stay explicit without vendoring a snapshot in the package.

### Verification

- full odu CI on `x86_64-linux` and `aarch64-darwin`
- `pnpm --filter @kolu/solid-pierre test:unit`
- client and `@kolu/solid-pierre` typechecks
- client production build
- website check
- `just fmt`

### Try it locally

```sh
nix run github:juspay/kolu/rhai
```

_Generated by Codex (model `GPT-5`)._


srid:
## Evidence

Rhai source highlighted in the Kolu Code tab

The real Code tab renders a `.rhai` file with distinct highlighting for comments, declarations, numbers, interpolated strings, function calls, and control flow.

PR #1953, opened 2026-07-23
Stop watch-edge debounce starvation under hot streams

### LINKED ISSUE #1952: Grok stays Thinking / dock-spinning after the TUI already shows a finished reply
## Summary

A Grok session can look **finished in the terminal UI** (answer fully rendered, “Worked for Ns”, idle prompt) while kolu still reports **`agent.state === "thinking"`** (title: **Thinking**, dock StatePip **spinning** as working).

This is **not** a dock-motion bug: the dock correctly treats `working` as always-active. The lie is upstream in **Grok state detection** (events fold / session freshness).

Not fixed in #1951 (motion/paint PR); tracked here separately.

## Live evidence (this machine)

Terminal under investigation (same tile as the screen recording):

| Field | Value |
| --- | --- |
| `KAVAL_TERMINAL_ID` | `d058febe-ef36-4f6f-82d8-33c721313286` |
| Agent | `grok` / `grok-4.5` |
| Session id | `019f8b9c-2977-7730-b908-5baf9a238e93` |
| Live probe | `wait_agentState` → **`state: "thinking"`** while the tile showed a completed assistant turn |
| Events path | `~/.grok/sessions/%2Fhome%2Fsrid%2Fcode%2Fkolu%2F.worktrees%2Fagent-icon/019f8b9c-2977-7730-b908-5baf9a238e93/events.jsonl` (~2.1 MB, 24k+ events) |

Screen recording (user): dock row + title chrome spinning / **Thinking** next to a finished “All done?” reply.

### 1. Dock is honest given the agent fact

`pipIsActive` / `pipMotionKind` (**`packages/client/src/terminal/pipMotion.ts`**):

- `agentBucket === "working"` → **always active → always spin**
- `thinking` / `tool_use` / `running_background` are working

So a stuck `thinking` **must** look “agent active” in the dock. That is intentional for true in-flight work.

### 2. Grok state comes only from `events.jsonl`

**`packages/integrations/grok/src/core.ts`** — `foldEventsState` / `deriveStateFromEvents`:

| Newest meaningful event | Folded state |
| --- | --- |
| `turn_ended` | `waiting` |
| `turn_started` (open turn, before a later phase) | `thinking` |
| `phase_changed` `streaming_reasoning` / `streaming_text` / `waiting_for_model` | **`thinking`** |
| `phase_changed` `tool_execution` | `tool_use` |
| empty / unreadable events | **`thinking`** (default) |

Unknown phases also default to **`thinking`**.

### 3. What the session file shows

On session `019f8b9c-…`:

- Turns **do** emit `turn_ended` with `outcome: "completed"` (e.g. `2026-07-23T00:37:23.993Z`).
- Immediately before many `turn_ended`s the phase is still **`streaming_text`** or **`streaming_reasoning`** (answer already on screen; turn not closed yet).
- **~12 / 47** `turn_ended` events are followed by a new `turn_started` within **≤10 ms** (auto-continuation / multi-step), so `waiting` is only a flicker and the session is immediately `thinking` again.
- Example gap that *should* show idle: `turn_ended` @ `00:37:23.993Z` → next `turn_started` @ `00:37:55.235Z` (~31 s). Fold on a prefix ending at that `turn_ended` is **`waiting`**. If the chrome still said **Thinking** in that window, either the UI was still catching up or a separate refresh bug — worth confirming in repro.

**Net product failure:** “Grok looks done to a human” ≠ `agent.state === "waiting"` for long enough (or at all) for the dock to stop spinning.

## Expected

After a completed user-visible turn (assistant answer settled, waiting for the human):

1. Title shows **Waiting for input** (or equivalent), not **Thinking**
2. Dock StatePip is **still** (or at most EF2’s short post-`waiting` spin), not continuous working-spin
3. `wait_agentState(until: ["waiting"])` resolves

## Actual

1. Title stays **Thinking** / tools
2. Dock glyph keeps **spinning** (working)
3. Live agent fact remains `thinking` (or `tool_use`) while the TUI already shows a finished answer

## Reproduction

1. Open kolu; start (or resume) a **Grok** session in a tile (`GROK_AGENT=1` / normal Grok integration).
2. Send a short prompt that finishes with a normal assistant reply (no further tools).
3. When the TUI shows the full reply and “Worked for …” (idle for the human):
   - Observe tile title agent label and dock StatePip.
4. Optionally probe:
   ```bash
   # from a kolu MCP / agent context with the tile id
   # wait_agentState → expect waiting, often still thinking
   ```
5. Inspect events:
   ```bash
   SESSION=~/.grok/sessions/<urlencoded-cwd>/<session-id>
   tail -n 40 "$SESSION/events.jsonl" | jq -c '{ts,type,phase,tool_name}'
   ```
   Note last `turn_started` / `turn_ended` / `phase_changed`.

### Variants that make it worse

- Multi-step tool turns: `turn_ended` → `turn_started` in **2–13 ms** → almost never sits in `waiting`.
- Long `streaming_text` / `streaming_reasoning` after the visible answer is already complete → chrome stays Thinking until `turn_ended`.

## Where to fix (for a future PR — not #1951)

Likely one or more of:

1. **Events fold** (`foldEventsState`): treat “answer complete / idle for user” more like Grok’s TUI (e.g. don’t hold `thinking` across a completed stream when `turn_ended` is delayed; handle auto `turn_started` after `turn_ended` for pure tool-continue vs real user-idle).
2. **Grok upstream**: emit `turn_ended` when the TUI returns to the idle prompt; avoid phase stuck on `streaming_*`.
3. **Watcher freshness**: confirm `turn_ended` always schedules a re-derive (append floor already exists in `session-watcher.ts` — re-verify if chrome can lag tens of seconds after `turn_ended` with no further appends).

## Relation to #1951

PR #1951 makes working-state **more obvious** (busy orange + 2.8s spin). It does **not** invent the stuck `thinking` fact; it amplifies a preexisting Grok-detection mismatch. Fix detection here (or in Grok), not by special-casing spin off for thinking.

## Environment

- Host: Linux (kolu PWA + local padi/kaval)
- Agent: Grok Build / grok-4.5
- Session: `019f8b9c-2977-7730-b908-5baf9a238e93` under agent-icon worktree
- Observed: 2026-07-22 / 2026-07-23

srid:
**Agent tiles (and any hot `fs.watch` consumer) can freeze on a stale state while the on-disk source keeps moving** — pure trailing-edge debounce never reaches its quiet window under a continuous burst. Live Grok repro for #1952: disk fold was `tool_use` while the published agent stayed `thinking`.

Grok's `events.jsonl` is a firehose (hundreds of `phase_changed` lines/sec). The same shape is latent in claude-code transcript watches, codex/opencode WAL refresh, and the git dir-filename watchers.

### Fix

Shared coalesce in **`kolu-io`**: trailing quiet window + hard **maxWait** (`createCoalesceSchedule`, `COALESCE_MAX_WAIT_MS = 500`).

| Site | Change |
| --- | --- |
| `createDirFilenameWatcher` | maxWait baked in — git watchers get it with zero call-site edits |
| `createDebounceWatcher` | maxWait baked in — codex/opencode WAL refresh gets it with zero call-site edits |
| Grok + Claude session watchers | ported onto the shared primitive; constants re-exported/sourced from `kolu-io` |

```
edge / poll ──▶ schedule() ──▶ quiet 150ms? ──▶ fire
                     │
                     └── past 500ms since first pending? ──▶ fire anyway
```

### Regressions

- Primitive starvation test (no fs.watch): continuous `schedule()` with gaps &lt; debounce still fires within maxWait
- Grok mid-burst pin through the real watcher: flip latency ≤ maxWait+slack from first append; **fails loud** if any inter-append gap ≥ debounce (jitter false-pass guard)
- Invariant: `COALESCE_MAX_WAIT_MS ≤ DEFAULT_APPEND_POLL_MS`

Fixes #1952

### Try it locally

```sh
nix run github:juspay/kolu/fix/grok-agent-state-1952
```

_Generated by [`/forge-pr`](https://github.com/srid/agency) on Grok Build (model `grok-4.5`)._

### DIFF
--- a/packages/integrations/claude-code/src/session-watcher.ts
+++ b/packages/integrations/claude-code/src/session-watcher.ts
@@ -11,7 +11,13 @@
 
 import fs from "node:fs";
 import { agentInfoEqual } from "anyagent";
-import { DEFAULT_APPEND_POLL_MS, subscribeFileAppends } from "kolu-io";
+import {
+  COALESCE_DEBOUNCE_MS,
+  COALESCE_MAX_WAIT_MS,
+  createCoalesceSchedule,
+  DEFAULT_APPEND_POLL_MS,
+  subscribeFileAppends,
+} from "kolu-io";
 import {
   completedBackgroundTaskIds,
   decayTransientState,
@@ -68,13 +74,11 @@ function workflowEqual(
 
 // --- Tuning constants ---
 
-/** Trailing-edge debounce for the transcript fs.watch callback. Claude
- *  streams tokens, and Linux fs.watch fires multiple events per write —
- *  without debouncing, `onTranscriptMaybeChanged` runs dozens to hundreds
- *  of times per second, each iteration re-allocating a 256 KB tail buffer
- *  and firing an async SDK summary fetch. 150 ms coalesces bursts into
- *  one handler run while keeping the user-perceptible lag imperceptible. */
-const TRANSCRIPT_DEBOUNCE_MS = 150;
+/** Quiet-window for the transcript coalesce schedule. Claude streams
+ *  tokens, and Linux fs.watch fires multiple events per write — without
+ *  coalescing, `onTranscriptMaybeChanged` runs dozens to hundreds of
+ *  times per second. Shared `COALESCE_DEBOUNCE_MS` + maxWait (#1952). */
+const TRANSCRIPT_DEBOUNCE_MS = COALESCE_DEBOUNCE_MS;
 
 /** Chunk size for `scanTasksIncremental`. The previous one-shot
  *  `Buffer.alloc(size - offset)` could allocate hundreds of MB transiently
@@ -169,9 +173,13 @@ export function createSessionWatcher(
   // tail of one call would be split-processed at the head of the next
   // call as if it were already a complete line — silent task corruption.
   let taskScanRemainder = "";
-  // Trailing-edge debounce timer for transcript fs.watch events.
-  // Null when idle. Cleared on destroy.
-  let transcriptDebounceTimer: NodeJS.Timeout | null = null;
+  // Quiet + maxWait coalesce for transcript edges (#1952 shared primitive).
+  // Destroyed in `destroy()` so late fires can't run after teardown.
+  const transcriptCoalesce = createCoalesceSchedule({
+    debounceMs: TRANSCRIPT_DEBOUNCE_MS,
+    maxWaitMs: COALESCE_MAX_WAIT_MS,
+    onFire: () => onTranscriptMaybeChanged(),
+  });
   // One-shot timer armed at the next workflow-journal stale deadline while
   // `running_background` is published. A journal going stale produces no
   // fs.watch event (it's the *absence* of writes), so without this the phantom
@@ -205,18 +213,12 @@ export function createSessionWatcher(
     }
   }
 
-  /** Trailing-edge debounce: reset the timer on every event, fire
-   *  `onTranscriptMaybeChanged` once after `TRANSCRIPT_DEBOUNCE_MS` of
-   *  quiet. The handler's own `destroyed` guard makes late-firing
-   *  callbacks safe, but we clear the timer in `destroy()` anyway to
-   *  avoid holding closure refs unnecessarily. */
+  /** Coalesced transcript re-derive: quiet window `TRANSCRIPT_DEBOUNCE_MS`,
+   *  hard maxWait `COALESCE_MAX_WAIT_MS` so a continuous token-stream burst
+   *  cannot starve the handler (juspay/kolu#1952). */
   function scheduleTranscriptCheck() {
     if (destroyed) return;
-    if (transcriptDebounceTimer) clearTimeout(transcriptDebounceTimer);
-    transcriptDebounceTimer = setTimeout(() => {
-      transcriptDebounceTimer = null;
-      onTranscriptMaybeChanged();
-    }, TRANSCRIPT_DEBOUNCE_MS);
+    transcriptCoalesce.schedule();
   }
 
   /** Arm (or clear) the one-shot timer that re-runs the derivation when a
@@ -591,10 +593,7 @@ export function createSessionWatcher(
 
     destroy() {
       destroyed = true;
-      if (transcriptDebounceTimer) {
-        clearTimeout(transcriptDebounceTimer);
-        transcriptDebounceTimer = null;
-      }
+      transcriptCoalesce.destroy();
       if (staleDeadlineTimer) {
         clearTimeout(staleDeadlineTimer);
         staleDeadlineTimer = null;
--- a/packages/integrations/grok/src/session-watcher.test.ts
+++ b/packages/integrations/grok/src/session-watcher.test.ts
@@ -1,11 +1,15 @@
 /**
- * grok session-watcher wiring for the append-robust floor (juspay/kolu#1754).
+ * Grok session-watcher regression pins:
  *
- * The exhaustive floor mechanics live in kolu-io's
- * `file-append-watcher.test.ts`; this is the CI regression guard that the floor
- * is actually reachable THROUGH the real grok watcher — a dropped `turn_ended`
- * edge on `events.jsonl` self-heals to `waiting` with no edge and no further
- * write.
+ *  - **#1754** — the append-robust floor is reachable THROUGH the real grok
+ *    watcher: a dropped `turn_ended` edge on `events.jsonl` self-heals to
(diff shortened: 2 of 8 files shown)


srid:
**Review (Claude, on behalf of @srid).** The fix is correct — the maxWait shape, the `pendingSince` lifecycle, and the mid-burst safety (128 KB tail + torn-line drop in `deriveStateFromEvents`) all check out. Before merge, please do **all** of the following in this PR — none of it is deferred to a follow-up.

### 1. Enforce the documented invariant

The `DEBOUNCE_MAX_MS` doc comment claims it "must be ≤ the append-poll interval", but nothing pins that. `session-watcher.test.ts` already imports `DEFAULT_APPEND_POLL_MS` — add:

```ts
expect(DEBOUNCE_MAX_MS).toBeLessThanOrEqual(DEFAULT_APPEND_POLL_MS);
```

### 2. Close the regression test's false-pass window

The burst appends every ~50 ms via real `sleep`. On a loaded CI runner, one overshoot past `DEBOUNCE_MS` (150 ms) lets the **pre-fix** pure-trailing debounce fire mid-burst too — the guard would pass without the fix. Harden it:

- Record a timestamp in the `onChange` callback when the state flips to `tool_use`, and assert it landed within `DEBOUNCE_MAX_MS` + slack of the *first* append.
- Track inter-append gaps during the burst and **fail loudly** if any gap reached `DEBOUNCE_MS` (a jitter-corrupted burst is not a valid run — fail fast, don't silently pass).

### 3. Fix the stale test-file header

The module doc at the top of `session-watcher.test.ts` still describes the file as solely the #1754 floor guard; it now also hosts the #1952 starvation regression. Update it.

### 4. Lift debounce+maxWait into the shared watch primitives (in this PR)

The same pure-trailing-debounce starvation is latent everywhere else a hot source can out-pace 150 ms:

- claude-code `scheduleTranscriptCheck` (`packages/integrations/claude-code/src/session-watcher.ts:213`) — hand-rolled trailing debounce
- codex + opencode WAL watchers — `debounceMs: 150` via `subscribeWalChanges` (`packages/shared/src/sqlite/wal-subscription.ts`)
- all five git watchers — `WATCHER_DEBOUNCE_MS` via `createDirFilenameWatcher` (`kolu-io`)

The bug is not agent-specific, so the shared home is **`kolu-io`** (the watch-plumbing package), *not* `anyagent` (contracts-only by charter; git couldn't import from it without a wrong-direction dependency). Concretely:

1. Extract a coalescing scheduler — `debounceMs` + `maxWaitMs`, the exact logic this PR inlines in `schedule()` — into `kolu-io`, with the starvation regression test at the primitive level.
2. Bake `maxWaitMs` into `createDirFilenameWatcher` (kolu-io) and `subscribeWalChanges` (kolu-shared) internally — codex, opencode, and the git watchers then pick it up with **zero call-site changes**. Per `conventions.md`, no per-call opt-out knob: the maxWait is a module constant, not a parameter consumers can disable.
3. Port grok's `schedule()` (this PR's inline version) and claude-code's `scheduleTranscriptCheck` onto the shared primitive, so there is one source of truth for both constants and future agents get the guarantee by construction.
4. Keep grok's #1952 regression test — it pins that the guarantee is reachable through the real grok watcher — but source the constants from the shared primitive.

_Posted by Claude Code at the request of @srid; srid will hand this comment to the implementing agent._

srid:
**Re-review (Claude, on behalf of @srid).** All four items delivered correctly — invariant pinned at both levels, mid-burst flip-latency bound + loud jitter fail-fast, header fixed, and the shared `createCoalesceSchedule` baked into `createDirFilenameWatcher` / `createDebounceWatcher` with clean dependency direction (`kolu-shared → kolu-io`, no cycle). CI green on both platforms.

One residual to fix before merge:

### Drop the dead `now?` knob from `CoalesceScheduleOpts`

`coalesce-schedule.ts` declares:

```ts
/** Clock for pending-since accounting. Defaults to `Date.now`. Tests
 *  may inject a controllable clock; real timers still drive the fire. */
now?: () => number;
```

No test injects it — `coalesce-schedule.test.ts` drives real timers and calls `Date.now()` directly. An optional-with-default that nothing exercises is exactly the shape this repo's fail-fast law forbids, and `kolu-io` already documents the rule on the neighboring primitive (`SubscribeFileAppendsOpts.intervalMs`: "not optional-with-default — that shape is the dead knob the fail-fast law forbids"). Remove the option and use `Date.now` directly inside `createCoalesceSchedule`; if a future test genuinely needs a controllable clock, that's the moment to add the seam — with a caller.

_Posted by Claude Code at the request of @srid._

PR #1972, opened 2026-07-23
Make palette recency e2e wait for live metadata

srid:
**Palette recency e2e assertions now wait for the terminal label they actually assert**, instead of treating the first mounted row as proof that its asynchronous metadata is current. On rasam, the row appeared with the terminal’s earlier e2e-home label and then converged to the expected branch; the old one-shot comparison failed before that update arrived.

The existing 20-second test deadline remains the boundary. A real ranking or metadata defect still times out with the expected name, last observed name, and elapsed time; product code is unchanged.

### Evidence

- Unchanged focused scenario on rasam: failed in 7.395s with the stale e2e-home label.
- Same focused scenario with value polling: 1/1 scenario and 15/15 steps passed.
- Complete terminal-switcher feature on rasam: 22/22 scenarios and 158/158 steps passed.
- `just fmt` and `git diff --check` passed.

_Generated by Codex (model `GPT-5`)._

### DIFF
--- a/packages/tests/step_definitions/command_palette_steps.ts
+++ b/packages/tests/step_definitions/command_palette_steps.ts
@@ -1,5 +1,6 @@
 import * as assert from "node:assert";
 import { Then, When } from "@cucumber/cucumber";
+import { pollFor } from "../support/poll.ts";
 import { type KoluWorld, MOD_KEY, POLL_TIMEOUT } from "../support/world.ts";
 
 const PALETTE_SELECTOR = '[data-testid="command-palette"]';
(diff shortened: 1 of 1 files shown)


srid:
CI e2e metrics for `97b7d6f#1`

|---|---|---:|---:|---:|---:|
| x86_64-linux | kolu-ci-4 | 8 | 32 | 8 | 2m 26.725s |
| aarch64-darwin | nix-infra@rasam.tail12b27.ts.net | 6 | 24 | 6 | 11m 48.897s |

PR #1975, opened 2026-07-24
Port forwarding: the plan, @kolu/port-forward, and the vazhi TUI (PRT0)

srid:
**Ships PRT0 of the port-forwarding plan**: `@kolu/port-forward` — a dependency-free library that makes any `(host, port)` answer on the machine it runs on — and **vazhi** (Tamil: *way, passage*), a standalone full-screen Ink TUI over it, in the spirit of opencode. The Atlas note in this PR (`docs/atlas/dist/port-forwarding.html`) is the plan-of-record for the rest: kolu's Inspector gets live, clickable port chips in PRT1–2 by embedding this same library — no new daemon, kaval and padi stay the only ones.

Built for kolu's real deployment shape — **server on a headless Linux box (`pureintent`), browser on a macbook (`zest`) over tailscale** — so "open this port" never means `localhost`:

| where the server listens | how it becomes reachable | you open |
| --- | --- | --- |
| kolu host, loopback | TCP relay on `0.0.0.0` | `http://pureintent:4123` |
| remote host, loopback | dedicated `ssh -N -L *:4123:…` per forward | `http://pureintent:4123` |
| kolu host, `0.0.0.0` | already reachable — nothing to do | `http://pureintent:4123` |

### What's in the box

- **`packages/port-forward/`** — the forward map (`create` idempotent by target · `cancel` · `list` · `dispose` · `onLost`), same-port-first selection (`pureintent:4123 → :4123` when free, random fallback), zero kolu dependencies. _This API is the contract kolu's PRT2 embed consumes._
- **`packages/vazhi/`** — full-screen Ink TUI: live forwards table, `a` add · `x` cancel · `q` quit, clickable `http://host:port` links (OSC 8 + plain-URL linkification), its own `flake.nix` for later externalization.

### Field-testing found two bugs; fixing them replaced the mechanism

| found | root cause | fix |
| --- | --- | --- |
| `q` left ports mapped | forwards lived on a shared `ControlPersist` ssh master — an external daemon designed to outlive the app; a skipped/failed `-O cancel` orphaned them | **each forward is now its own `ssh -N -L` child** (`ControlPath=none`): cancel is `kill()`, crash cleanup is the kernel's |
| restarted TUI showed empty list while the port stayed open | OpenSSH masters have no `-O list` — their forward table is write-only, so orphans were invisible and unadoptable | with per-process forwards the list is true by construction; `kill -9` the app and its ports refuse connections immediately (pinned by `lifetime.test.ts`) |

> Measured before deciding: a mux client riding a shared master does **not** get its forward reaped on death — the master owns it. So master-sharing and kernel-tied lifetime are mutually exclusive in OpenSSH; we chose lifetime, at the cost of one ssh handshake per forward (fine for kolu too: its mirror already requires `BatchMode` non-interactive auth).

### Try it locally

```sh
nix run github:juspay/kolu/process-port#vazhi
```

_Generated by Claude Code (model `claude-fable-5`); implementation by a kolu-driven Claude Code session (model `claude-opus-5`)._


srid:

✅ **Consensus** after 1 thread-round(s) · lowy + hickey · base `6e3af139ed6f`

Independent findings: lowy=13, hickey=21
Settled: 4 reconciled (raised by both lenses) · 15 auto-settled minor · 6 unopposed · 3 debated

### Applied (26)
- `lowy-2` port-forward forks real ssh but is the only forking package with no vitest.config.ts, so the daemon-test-gate env scrub never runs — commit `1a19ae060`
- `lowy-4` The two generators behind the mechanism receptacle have different shapes; createSshForwards is a stateless factory admitting it holds nothing — commit `082e1a798`
- `lowy-5` ≡ `hickey-6` Targets are validated inside the generators, never at the receptacle the second consumer will call — commit `77b4b0ef3`
- `lowy-6` ≡ `hickey-13` The pure frame and the stateful shell share one module, so a rendering test pulls ssh and child_process into its import graph — commit `57f8e90c1`
- `lowy-7` ≡ `hickey-16` "How this process is asked to stop" has two owners, and one registers its handlers on every render — commit `d42e707d5`
- `lowy-8` ≡ `hickey-19` The root flake's public output set is spread from an implementation attrset, so a package file can silently rename a top-level output — commit `15debedd4`
- `lowy-9` ≡ `hickey-21` The library's runtime requirement on ssh is documented only inside a consumer's packaging — commit `3fb016be7`
- `lowy-10` create() rejects with untyped strings, so failure-mode volatility is unencapsulated — record it beside the subscribe() question — commit `8064699c1`
- `lowy-11` ≡ `hickey-9` The manager test's fake re-implements targetKey, giving the map's identity rule a second definition — commit `130a1dfb9`
- `lowy-12` formatUptime is a third copy of the ladder the repo declares lives in one place — and here the duplication is the price of independence, which should be said out loud — commit `4f7f69eab`
- `lowy-13` vazhi's flake overstates how portable the package is — the build plugs straight into kolu's monorepo tree — commit `75b0a3dc6`
- `hickey-1` App hardwires its own forward manager, so none of vazhi's real behaviour is testable — commit `2458bf86b`
- `hickey-2` `rows` is a second copy of the forward map, reconciled three different ways — commit `29eec312b`
- `hickey-3` Selection is a bare index into a list that moves under it, clamped against two different lists — commit `49ff4f69f`
- `hickey-4` One forward map represented as two parallel maps, so `cancel` cannot see half of it — commit `75f792d25`
- `hickey-5` `openOne` complects running ssh with interpreting ssh, leaving the package's subtlest rules untested — commit `ba514b6f1`
- `hickey-7` "Which local port" is spread over three modules, held together by an `"any"` sentinel — commit `aea00a135`
- `hickey-8` `key` and `formatTarget` are two renderings of one target, and vazhi mixes them in front of the user — commit `6f98cfbb8`
- `hickey-10` `done` + `cancelled` booleans encode a forward attempt's state machine — commit `030dea925`
- `hickey-11` The startup diagnostic buffer is retained for the whole life of the forward — commit `3b19052c6`
- `hickey-12` `ForwardMechanisms` lives in `opened.ts`, a file named after a different concept — commit `d01e62813`
- `hickey-14` The prompt's two submit paths have identical bodies — commit `e3963a5a0`
- `hickey-15` `messageOf` is defined in App.tsx and re-inlined in main.tsx — commit `f6bb01dd9`
- `hickey-17` `hyperlink`'s second parameter has no caller and is pinned by a test — commit `fe39a755d`
- `hickey-18` The ControlMaster rationale is argued in full in three code headers — commit `593f00000`
- `hickey-20` `surfaceExampleBase` still names a tree that is no longer example-specific — commit `d9abfa55f`

### Agreed — no change (2)
- `lowy-1` Electricity test — receptacle, generators, verdict (required lens output, not a defect) (packages/port-forward/src/index.ts:46)
- `lowy-3` The ssh non-interactive/dead-peer policy is a second copy of SSH_COMMON_OPTS with nothing guarding the drift (packages/port-forward/src/sshForward.ts:51)

srid:
## Codex ⇄ Claude Code debate

✅ Consensus in 12 rounds · peer effort: xhigh · base `52fe2dcb`

Codex raised **18 findings across 12 rounds** — 3 blocking, all fixed, none disputed. Library tests went 60 → 152.

|---|---|---|
| 1 | 9cb29b0e | fix: close ten findings from the Codex review (round 1) |
| 2 | a45ec0a1 | fix: close the seven findings still open after Codex round 2 |
| 3 | 9a9b21bb | fix: one teardown transition, and the four findings around it (round 3) |
| 4 | 308aab6a | fix: close the four findings still open after Codex round 4 |
| 5 | 0c040b0d | fix: a mechanism can report trouble without claiming the forward is gone |
| 6 | b3b34b76 | fix: keep the fault/loss distinction through the opening window, and pin the whole lifecycle |
| 7 | 896740b7 | fix: compose the opening window instead of sampling it (F14) |
| 8 | 08db1952 | fix: dispose counts only the teardowns that actually failed (F15) |
| 9 | 095d8f11 | fix: a cancel claims the opening it is about to close (F16) |
| 10 | 3543eefc | fix: a teardown's outcome is the map, not the close call (F17) |
| 11 | 92446103 | test: generate the same-kind teardown races too (F18) |

- **F1** — cancel deletes the slot before OpenedForward.close succeeds, and dispose similarly clears every slot at line 189 before closing
- **F2** — dispose waits for opening flights with Promise.allSettled and discards every result
- **F3** — The key encoding is not injective: a local target on 5173 and a valid remote SSH alias named "local" on 5173 both produce "local:5173"
- **F4** — stderr is truncated before it remains a source of startup truth
- **F5** — An error on an established net.Server only calls onLost
- **F6** — The implementation depends on stderr being emitted and on ssh stdin remaining a non-TTY pipe, but those facts are not pinned against valid per-host ssh config
- **F7** — detail copies remote-controlled stderr verbatim into public error and loss strings
- **F8** — The fixed-height screen renders every forward in one shrinking flex column instead of a viewport
- **F9** — canBindLocally turns every server error into false
- **F10** — This PR ships a new user-facing executable, but the canonical kolu.dev product docs have no vazhi page/how-to and the unreleased changelog is empty
- **F11** — Regression introduced by the F1 fix: cancel leaves an open slot visible while awaiting close
- **F12** — Round-three regression: create now calls this.create only after waiting for a closing slot
- **F13** — Round-six regression
- **F14** — Round-seven regression and reproducible matrix hole
- **F15** — Round-eight matrix hole with a deterministic failure
- **F16** — The uncrossed cancel composition publishes a forward already committed to teardown
- **F17** — The uncrossed report-during-closing composition makes dispose fail over an empty map
- **F18** — The stated matrix expansion is incomplete

srid:

Rules + fact-check passes over `9bcf1891...HEAD`. Pass 3 (elegance) skipped — `/simplify` ran as its own gauntlet step over this same tree.

| # | Finding | Where | Actioned |
|---|---|---|---|
| 1 | A half bind ssh complained about **late** was latched into a boolean nothing read again — stdout and stderr are separate pipes, so the bind line can arrive after the ready token it contradicts, and the row then claimed to be healthy | `sshForward.ts` | **Fixed** — surfaces through the `fault` channel; pinned by a test that fails without it. Not the suggested `setImmediate` defer: a timing guess narrows the window instead of closing it |
| 2 | `x` reported "no forwards to cancel" over a table with live rows in it — the selected key named a row that had left and nothing moved the highlight | `App.tsx` | **Fixed** — highlight advances off a cancelled row, the two situations no longer share one message, and the selected row is derived from the list; pinned by a test that fails without it |
| 3 | `callback-fanout-guarded-at-funnel` — `onLost` invoked from two sites mid-transition; a throwing consumer could abandon that work half-done (a forward deleted from the map with its ssh child never reaped) | `manager.ts` | **Fixed** — one `announce` funnel contains the throw and re-raises it on its own turn: still loud, but after the map is consistent |
| 4 | `dry-rule-of-three` — the identical `Forward` literal built three times; `err instanceof Error ? … : String(err)` written out three times | `manager.ts`, `portChoice.ts`, `relay.ts` | **Fixed** — one `record()`, and `messageOf` in `diagnostic.ts`. vazhi keeps its own one-liner rather than have the library export a generic error helper |
| 5 | `no-thin-wrapper-functions` — `openRelay(port, report)` only called `openRelayWith({…})`; one caller, no transform | `relay.ts` | **Fixed** — the injected form is the only one, with `nativeMechanisms` passing node's `createServer` exactly as it already hands the ssh mechanism its real spawner |
| 6 | `prefer-ts-pattern` — the `Slot.state` and `phase` cascades use sequential `if`s rather than an exhaustive match | `manager.ts`, `sshForward.ts` | **Not applicable** — `@kolu/port-forward` is deliberately zero-runtime-dependency so vazhi drags no kolu world in; `ts-pattern` cannot be added |

**Also caught while running the suites** (not a police finding): every assertion in `sshForward.test.ts` passed while `vitest` exited 1 on an unhandled rejection — a test fed three chunks with an `await` between each, so the attempt rejected on the first and the expectation only attached after the last. The test landed after the last CI run, so CI had never seen it and would have gone red. Fixed in `399eeaf9`.

Suites after every fix: **153 passed | 1 skipped** (port-forward), **56 passed** (vazhi). `just fmt` and `just check` clean.

PR #1982, opened 2026-07-25
Ports in the Inspector: padi's port sensor + a live Ports section (PRT1)

srid:
**PRT1 — the Inspector answers "what is this terminal serving?"**

A live **Ports** section lists every TCP port a terminal's processes are listening on, within about a second of the port appearing. A port bound to `0.0.0.0` on the kolu server's own host already answers on the name in your address bar, so its chip opens in one click. Loopback-bound and remote-host ports are listed with an inert "needs a forward" note that PRT2 activates — **no forwarding is built here**, and `@kolu/port-forward` is not used by kolu yet.

### Detection — one host-wide scan, attributed by process subtree

`packages/padi/src/portScan.ts` is a padi sensor, beside the git sensor and `memorySampler`: its only input is each shell's root pid, so it needs no PTY access and changes no kaval wire contract.

- **linux** reads `/proc` directly — pid table from `stat`, LISTEN rows (with bind address) from `net/tcp{,6}`, joined to pids through the `/proc/<pid>/fd` socket-inode readlink `socketHolder.ts` already uses, walking only the union of terminal subtrees.
- **darwin** pairs `ps -axo pid,ppid,comm` with `lsof -nP -w -iTCP -sTCP:LISTEN` under Node's own kill timer (macOS ships no GNU `timeout`), both invoked by ABSOLUTE path — a nix `ps` on PATH is procps, whose `-o comm` is a truncated name rather than the executable path this parser reads.
- No `ss` on linux (measured slower, and it hides other users' pid attribution). **`netstat` is deliberately NOT used on darwin** — see below.

**Cadence** (`portSampler.ts`): a 5 s baseline plus an immediate off-schedule pass on terminal output or an OSC 633 command mark — single-flight, with a ≥1 s floor. Output is only ever a hint about *when to look*; the socket table stays the sole source of facts, so a printed URL never creates a chip.

The discipline the plan pinned holds: **no OS state between ticks**, **repartition from the current root pids every pass**, and a **blind scan fails loudly** — an `EACCES` on a subtree we were asked about throws rather than publishing an empty set, because "we cannot see" and "this terminal serves nothing" would otherwise render identically.

### Schema and wire

`PortInfo { port, name, wildcard }` joins `TerminalSnapshot`, folded through the sensor set with a **structural-equality dedup**, so an unchanged scan emits nothing at all — the churn guard that makes a seconds-cadence ticker affordable. padi surface **4.1 → 4.2** (a new required field on an emitted value).

The bind address is reduced to the one bit a consumer acts on. Carrying the raw address would invite every render site to re-derive "is this already reachable?" — and to disagree about `::ffff:0.0.0.0`.

### Tests

- **Parser fixtures for both platforms** (`portScan.test.ts`), each pinning a way the naive parse yields a plausible *wrong* answer: byte-reversed `/proc` hex, IPv6 reversed per 32-bit word, the v4-mapped wildcard vs the v4-mapped loopback (they differ in four bytes and mean opposite things), `tcp46` dual-stack rows, a `comm` containing spaces and parentheses, and macOS `lsof` field output (bracketed IPv6, a wildcard bind, two fds on one port).
- **Cadence** (`portSampler.test.ts`): baseline, nudge, the floor under a 50-nudge burst, single-flight with a mid-pass nudge re-running, and a blind scan leaving the last sample standing.
- **Churn guard** (`sensors.ports.test.ts`): an identical rescan emits nothing; a bind or name change emits.
- **e2e** (`features/ports.feature`, 3 scenarios, green locally in 7.5 s): a real `node -e … .listen()` in a real PTY → chip → click opens `location.hostname:<port>`; a loopback listener listed but inert; and a port leaving the section when its server dies.

Two things the tests caught that review would not have: the inert chip carried no `data-port` (invisible to the query), and `/proc/<pid>/stat`'s `comm` reads **`MainThread`** for a plain Node server — so every Node dev server would have been labelled `MainThread` until the name moved to `cmdline`'s `argv[0]`.

### Verified on both platforms

The darwin path is no longer fixture-only, and running it on real Macs found two bugs a review could not have. The first: `netstat`'s owner column is spelled **`process:pid`**, not the bare `pid` the parser first looked for — it threw on every real darwin scan while every fixture stayed green. The second killed netstat outright (below). Fixtures are now real captured output from both boxes.

`portScan.live.test.ts` closes that gap for good: real listeners, real `ps`/`lsof` or `/proc`, on whatever platform it runs. `describeDaemon`-gated (it forks), so it runs in `ci::daemon` — required on both platforms. Verified by hand on linux and on BOTH Macs — 26.4 and 27.0: wildcard vs loopback, a grandchild via a shell, a dual-stack bind folded to one entry, port death, subtree isolation, and a double-forked daemon correctly left out.

### Fixed after two real deployments

**macOS 27 makes `netstat` go blind.** Reported from a live remote host: an empty Ports section next to two running dev servers. `netstat -anv` returns an EMPTY internet table to the padi process while reporting success — zero bytes, exit 0, clean stderr — while an ssh shell on the same box prints the real table, and `lsof` sees everything from inside that very same process. The plan pinned netstat as faster and "more complete"; both were true of the box it was measured on, but a mechanism that silently answers "no ports" is the worst possible one for this feature. **The darwin path is now `lsof`**, replaced rather than kept as a fallback. The completeness half costs nothing — non-root lsof reports the invoking user's processes, and a terminal's subtree is padi's own user by construction. On the reported host the scan now returns exactly what was running: 5566 loopback (needs a forward) and 8079 wildcard (openable).

**Splits.**

Deployed to a live Linux host, the section came up **empty while a dev server was running**. The server was in a **split**, and the section read only the tile's main pane — so it showed nothing in the common case, since a split is exactly where a dev server goes. Each pane is its own PTY with its own process subtree, so attributing the port to the split is correct; the fix is at the render site, which now reads the whole tile (main + splits), the same unit the Attach section already documents. Pinned by an e2e scenario that fails without it.

### Cost accounting — what the architecture actually spends

The shape is "TypeScript calls an external process" on darwin and "TypeScript reads
`/proc` in-process" on linux. Both overheads are measured here rather than argued
about, so keeping them is a decision against numbers.

**Darwin, the ~6 ms pass.** Round trip timed from Node exactly as padi calls it; the
helper's own share read from its mach monotonic clock **in the same invocation**, so the
parts cannot drift against the whole. (A first attempt timed them in separate loops and
the parts summed to *more* than the total — that version is not what is reported.)
Instrumentation is a measure-only build; nothing ships with timers. zest, macOS 27.0,
718 rows / 20.9 KB output, 15 runs:

| component | median | share |
|---|---|---|
| **the C work itself** (libproc walk + format) | **4.11 ms** | 65% |
| process lifecycle — fork/exec + dyld + teardown | **1.96 ms** | 31% |
| pipe transfer + `execFile` plumbing | **0.30 ms** | 5% |
| **round trip, shipping build** | **5.96 ms** | — |

The 1.96 ms is measured, not inferred — a no-op C binary spawned from Node under
identical conditions. The instrumented build costs 0.41 ms over the shipping one, so the
shipping total is quoted separately.

TSV parsing is a rounding error, measured with the **shipped** `parseHelperOutput` over
that same real output: **0.112 ms** median (min 0.104, max 0.371, 200 runs) for 696
process rows + 21 listener rows — ~2% of the pass.

**Linux, the ~14 ms pass.** No subprocess at all. Instrumented copy of the real reader,
573-process box, 12 runs:

| stage | median | note |
|---|---|---|
| **pid table** (`readdir /proc` + batched `stat`/`cmdline`) | **13.29 ms** | ~96% of the pass |
| socket table read (`/proc/net/tcp{,6}`) | 3.45 ms | runs CONCURRENTLY with the pid table — hidden under it, not additive |
| fd walk (subtree pids only) | 0.31 ms | |
| join | 0.07 ms | |
| socket table parse | 0.06 ms | |
| **total `scanTerminalPorts`** | **13.91 ms** | |

This settles a claim the module header had only been asserting: the dominant cost is the
HOST pid table, scaling with the host rather than the question — which is exactly what
subtree descent would remove.

**Steady state.** Worst case is the duty-cycle floor's fast end, one pass per second:

| | per pass | cost at 1/s |
|---|---|---|
| darwin | 5.96 ms | **0.60% of one core** |
| linux | 13.91 ms | **1.39% of one core** |

Darwin's spawn peaks at **1.9 MB RSS**, transient, once per pass.

**The counterfactual, stated.** An in-process darwin reader (an N-API addon over the same
libproc calls) would save the lifecycle plus the plumbing — **2.26 ms per pass**, taking
5.96 ms to ~3.7 ms, or 0.23% of a core at that cadence. Declined, for reasons that
outweigh 2 ms: a native addon couples padi's build to a Node ABI and forces a rebuild per
Node bump inside the nix closure; a crash in a separate process cannot take padi down,
whereas a segfault in an addon takes the daemon with it; and the subprocess boundary is
what lets one binary serve drishti later. The number is recorded so the trade can be
re-opened against evidence rather than instinct.

The same tables are in the Atlas performance note, and the tool design these measurements
point at is now its own note (`os-facts-tool`, proposed — not shipped here).

### Review gauntlet

This branch went through the full serial gauntlet; each stage's findings are PR
comments above.

| Stage | Outcome |
|---|---|
| architecture-first-principles (C1–C7) | 9 verified findings applied — including a real bug where **one `sudo` password prompt blinded the port scan for every terminal on the host** |
| lowy ∥ hickey lens review | 26 findings, 14 fixes — including replacing the hand-rolled sampler loop with `@kolu/surface`'s reactor poll source |
| Codex debate | consensus in 4 rounds. My F3 dispute was **wrong**: Codex named three surviving routes to a blind first pass, one of them widened by my own earlier fix. Withdrawn, and `ports` became an honest `known`/`unknown` two-way |
| simplify | dead exports dropped, `listenersOf` narrowed, a superseded derivation deleted |
| code-police (`--no-elegance`) | 4 findings applied — incl. delegating the bind-address grammar to `ipaddr.js` (already a transitive dep, so no nix hash churn), narrowed because `ipaddr.parse("0")` → `0.0.0.0` would have minted a chip falsely claiming to be openable |

A measured performance pass also came out of it: the scan's own header claimed
~3 ms, measured on a 30-process box. On a real 515-process host a pass cost
**35–57 ms**; batching the pid table and the fd walk brought it to **14–18 ms**.
The remaining order of magnitude (`/proc/<pid>/task/<tid>/children` descent,
<1 ms in a prototype) is recorded in the performance note rather than guessed at.

### Test honesty fix found by CI

The four ports scenarios went red on the CI runner while passing locally — and
passing locally *under the same full-suite parallel load* (532/533; the one red an
unrelated known code-tab file-tree flake). The message read "the Ports section
showed []", which blamed the sensor. It could not know that: the start steps were
**vacuous** (they typed the command, waited a frame, and passed unconditionally), and
`[]` conflated *section absent*, *section empty*, and a real read. Both are fixed,
the listener no longer depends on `node` being on the PTY's `PATH`, and the new
guard was itself vacuous on the first try — the shell echoes what is typed, so a
literal marker was already in the buffer. Caught only by a deliberate negative test.

### Not done here — CI is NOT green, and both lanes are blocked on infrastructure

Neither blocker is this branch. Root cause found for the linux side: all six runners
are reached through **one bastion** (`ProxyCommand ssh pu@10.10.68.56`), and that
host is at 100% packet loss via the same gateway that carries this workstation's
working internet (`git ls-remote` to GitHub succeeds from here). It degraded in
three stages during the run — runners reachable but `git fetch` inside them timing
out (`HTTP2 framing layer`), then slow ssh, then all six failing at banner exchange.

For `aarch64-darwin` there is no venue at all: `sincereintent` (the pool's only
host) answers a tailscale ping in 6 ms but its port 22 is closed, `zest` is offline,
and `rasam` completes the handshake with no authorized key. Branch protection
requires **13 `@aarch64-darwin` contexts**, so that lane must run before this merges
and cannot be routed around.

**Locally verified at `c4868aff`** — the 13 required *linux* contexts were run on
this box (production kolu inactive). This produces no GitHub statuses; it only means
the fleet's first run after recovery is unlikely to burn a cycle:

| | |
|---|---|
| ✅ 12/13 | `install` · `atlas-sync` · `nix` · `pnpm-hash-fresh` · `smoke` · `surface-example-build` · `surface-app-example-build` · `unit` · `biome` · `fmt` · `home-manager` · `daemon` |
| ⚠️ `e2e` | 532/533. The one red is `code-tab.feature:2022` (pierre file-tree locator), **not** a ports scenario — all four of those pass. Pre-existing and load-correlated, confirmed three ways: it passed in CI at `ce72e957` on code already containing this PR's `CodeTab.tsx` change, it passes in isolation locally in 6 s, and it is the catalogued file-tree/fs-watch flake class. |

No open code-scanning alerts on this branch (checked via
`gh api code-scanning/alerts`, since `gh pr checks` does not surface them).

**`aarch64-darwin` has no local proxy** — it is a different OS, and the darwin path
(`lsof`, `ps` by absolute path, macOS 27's blind `netstat`) is exactly the part that
needs real hardware. It was verified by hand earlier on both Macs (26.4 and 27.0),
but that is not the same as a green lane.

<details>
<summary>Plan-of-record update (the original contents of this PR)</summary>

**Plan-of-record update after #1975 merged**: the Atlas note records PRT0 as shipped (with its PR link), names PRT1 — the padi port sensor + Inspector Ports section — as the next phase, and pins the last review-derived details so PRT1's implementer inherits them instead of rediscovering them.

Two substantive additions, both distilled from the Zed/VS Code prior-art research:

- **Scanner discipline (PRT1)** — the pass retains no OS state between ticks (parse, join, emit, drop), repartitions from current terminal root pids every tick so a closed or re-keyed terminal never leaves a stale subtree, and runs single-flight across the output-nudge and the 5s baseline.
- **Two Ports-view UX details (PRT2)** — an auto-forward that dies and returns prefers its previous local port (bookmarks survive dev-server restarts), and forwarded rows carry a copy-address affordance.

</details>


srid:

**Status: applied** · lowy **10** findings (4 major) · hickey **16** (14 `fix`) · both read the full diff independently, in parallel · base `c772c07`

Both lenses independently reached the **same headline**, which is the strongest consensus signal this review can produce.

### The headline — `≡` both lenses

`portSampler.ts` hand-rolled a cadence runner whose every policy already lives in `@kolu/surface`'s reactor: the unref'd interval (`everyMs`, *"the one home"*), the interval-plus-edge fuse (`everyMsOr`, *"the one home… previously duplicated as app-local twins"*), the non-overlap guard, the mid-flight latch, and log-skip-continue. padi's own `servePadi.ts` consumes it twice, and `memorySampler.ts` — cited by the new module for its 5-second number — documents that its hand-rolled loop was **deleted** in favour of it. `packages/server/src/seal.test.ts` seals this with a **zero-allowlist** ban: *"a hand-rolled setInterval sampler is the retired defect."*

The new module slipped past that seal only because both of its coordinates moved — into `padi` (which the seal doesn't walk) and onto `setTimeout` (which its regex doesn't match). lowy hedged that the reactor lacked one capability, mid-flight edge coalescing; it does not (`reactor.ts:383` latches `dirty`, the `finally` at `:404` runs the trailing read), so the conversion kept nothing. The only genuine residue is the **≥1 s floor**, which stays padi-local — putting it in `@kolu/surface` would trip the paired-companion-PR rule for no benefit.

### Applied

| # | Fix | Raised by | Commit |
|---|---|---|---|
| 1 | Drive the port scan from the reactor's poll source | `≡` both | `bc81cffc` |
| 2 | One ANY-address classifier for both platforms — the two disagreed on `::ffff:0.0.0.0` | hickey | `deb238e9` |
| 3 | `foldPorts` picks its name deterministically, not by iteration order | hickey | `19562562` |
| 4 | `portsEqual` reads its comparison keys off `PortInfoSchema` | hickey | `bdce4a7e` |
| 5 | The per-terminal join written once, behind a platform reading | hickey | `3d5dfdc8` |
| 6 | Key the scan on root pids, not the caller's `TerminalId` | hickey | `5452a8d2` |
| 7 | Tag `PortScanError` so a permanent failure can't become a log loop | lowy | `d76b9332` |
| 8 | The reachability decision is a tag, not a sentence | `≡` both | `c75afbbd` |
| 9 | The store owns what a tile's PANES are | `≡` both | `e1e66b25` |
| 10 | State the active-arm rule once, where it is correct | hickey | `22918a08` |
| 11 | `seedSnapshot` really is the ONE home for the snapshot defaults | lowy | `d777efa8` |
| 12 | `MetadataInspector`'s header states the boundary it actually holds | lowy | `bd2d1451` |
| 13 | No comment still describes the mechanism removed for lying | `≡` both | `51843157` |
| 14 | Stop claiming the socket↔pid reading is shared; settle two of its drifts | lowy | `c9adccf5` |

Two of these were latent **bugs**, not tidiness: the two ANY-address predicates already disagreed on `::ffff:0.0.0.0` (the exact disagreement `PortInfo.wildcard`'s docstring says the field exists to prevent), and `foldPorts` picked `name` by iteration order while `portsEqual` treats name as identity — so two programs on one port would have churned a "change" forever, defeating the dedup the sensor exists for.

### Settled by experiment, not argument — valve 1

**`TerminalLifecycle.rootPid` duplicates `entry.info.pid`** (lowy `drop` ↔ hickey `fix`). The discriminating experiment: exactly two `startSnapshotSensors` sites, each passing the same expression written to `entry.info.pid` with no `await` between, and `teardownSensors` always deletes the lifecycle — so the two homes cannot disagree today, which hickey concedes ("latent fragmentation rather than a bug"). That settles the facts in lowy's favour and leaves a pure remedy disagreement.

**Adjudicated: not changed.** Reading the pid back through the registry would trade a *proven-consistent* copy for a new failure mode needing its own policy (what does a missing entry mean mid-pass?) — and getting that policy wrong is how the `sudo` blindness bug happened. Recorded rather than churned.

### Adjudicated to the plan — valve 2

Both judgment-shaped findings are now recorded in the Atlas note (`2e2c7ce1`), where PRT2's implementer meets them, rather than left in a comment:

- **`wildcard: boolean` vs `scope: "any"|"loopback"|"interface"`** (hickey). The case is sound — one observation doing duty as a judgment, and the judgment isn't a function of that field alone (the client joins it with the terminal's host). Kept for PRT1 because both live symptoms are fixed independently, and PRT2 is where the second consumer appears: relay-vs-`ssh -L` needs the host, which `wildcard` never carried, so PRT2 reshapes this surface anyway. Same discipline the note already applies to `list()`/`subscribe()`.
- **One fan-out for kaval's activity edge** (hickey now, lowy at the fourth consumer). Declined: the three consumers are not three of one shape — the live-dots subscription is per-*subscriber* by design while the other two are daemon-lifetime, so a shared receptacle would force the odd one in.

### Recorded as observations

`portScan.ts`'s **file split** (a 3-file split is an import cycle; a 4th shared-core module is layout taste), lowy's **seal extension** (padi now legitimately owns one `setTimeout` — the floor — so a zero-allowlist seal would forbid what we deliberately kept), the client's `?? []` (the two absences are provably different facts), and `portNudgeAbort`'s unused lifetime (subsumed by the fan-out question above).

Verification after every change: `just check`, `just test-unit`, the padi + supervisor gated suites under `KOLU_DAEMON_TESTS=1`, and `just atlas::check-sync`. Nothing pushed by the review itself.

srid:
## Codex ⇄ Claude Code debate

✅ Consensus in 4 round(s) · base `c772c07` · every finding resolved

|---|---|---|
| 1 | fa697a62 | fix(debate): round 1 — bind samples to their lifecycle, and one /proc policy |
| 2 | 62103d85 | fix(debate): round 2 — `ports` is an honest known/unknown two-way |
| 3 | b0376b21 | fix(debate): round 3 — the platform guard is unconditional and now testable |

- **F1** (major) — The claimed permanent stop for an unsupported platform only works when the T+0 seed throws
- **F2** (major) — The failure cache and completed pass are keyed only by `TerminalId`, not by the `{id, rootPid}` lifecycle they sampled
- **F3** (major) — A blind first sample still collapses to the valid empty state
- **F4** (major) — The process-table catch skips every error for every non-root pid: after handling `ENOENT`/`ESRCH`, it throws only when the pid is a requested root and otherwise executes `continue`
- **F5** (minor) — The blanket `catch { return comm; }` around `/proc/<pid>/cmdline` silently converts every failure into the fallback label, including permission, I/O, and resource-exhaustion errors
- **F6** (minor) — `decodeProcAddress` claims and reports that only 8- or 32-hex-digit `/proc` addresses are valid, but the condition accepts every non-empty multiple of 8, including 16 and 24 digits
- **F7** (nit) — The new discoverability tip says the section lists every port and then says to "click one to open the server", but loopback and remote-host rows are deliberately inert in PRT1
- **F8** (minor) — The new test named "does not arm at all" cannot detect an armed cadence
- **F9** (minor) — Round 2 changed the wire to `unknown | known`, but this blind-pass comment and its operator-facing message still describe the old state: they say `seedSnapshot` supplied `ports: []`, that the terminal reads as serving nothing, and that a discriminator is deferred to PRT2

srid:

Run with `--no-elegance` (the `/simplify` track already ran over this same tree),
scoped to the PR's merge-base. **4 findings, all applied.**

| # | Pass | Rule | Fix | SHA |
|---|---|---|---|---|
| 1 | 2 · fact-check | caught-error diagnostics | `linuxProcessTable`'s `readRow` wrapped the `readFile` **and** the `parseProcStat` call in one `try`. A parse error carries no errno, so `parseProcStat`'s own message — naming exactly which `/proc/<pid>/stat` field was unreadable — got replaced by a generic `cannot read … (undefined)` and buried in `.cause`. The `try` now wraps only the read. | `40a4d07a` |
| 2 | 1 · rules | doc-comment placement | An orphaned dedup-gate doc block moved onto `portsEqual`, the function it describes. | `f4841a47` |
| 3 | 1 · rules | `no-thin-wrapper-functions` | `openExternal` was extracted earlier in this gauntlet so the `noopener,noreferrer` posture was stated once — but a *later* lens fix moved `PortsSection` to an `<a href>` anchor, leaving the wrapper with exactly one caller. Inlined at `CodeTab.tsx` behind a named `EXTERNAL_OPEN_FLAGS` const; the wrapper file is gone. | `16ed149f` |
| 4 | 1 · rules | `prefer-focused-library` | `parseBindAddress` hand-rolled ~70 lines of IPv4/IPv6 text grammar — `::` elision, trailing embedded v4, per-group hex validation — that `ipaddr.js` already owns. | `ce72e957` |

### Finding 4 needed two library facts checked, not assumed

`ipaddr.js` was **already in the lockfile** as a transitive dependency, so adopting
it directly adds nothing to fetch — and `nix build --rebuild .#pnpmDeps` re-executes
the fetcher and the recorded FOD hash still matches, so there is no nix hash churn
either. Both verified rather than reasoned about.

Two behaviours were checked against the installed **1.9.1**:

- It **does** accept a v6 zone id (`fe80::1%en0`), so the hand-rolled `split("%")`
  pre-strip really was redundant.
- It is **laxer than this module can afford, in the one direction that matters.**
  `ipaddr.parse` also reads the legacy non-dotted IPv4 spellings: `"0x7f000001"`,
  `"2130706433"`, and — the dangerous one — `"0"`, which decodes to `0.0.0.0`. Left
  alone, junk in the address field would classify as a **wildcard** bind and mint a
  chip claiming to be openable. Narrowed with the library's own
  `isValidFourPartDecimal` (not a second hand-rolled grammar) and pinned by a test
  naming all four spellings. No socket table writes a bind address that way, so the
  narrowing rejects only shapes that cannot legitimately arrive.

### Note on finding 3

It reverses an extraction an *earlier* stage of this same gauntlet made. That is the
serial gauntlet working as intended — a later reviewer reads the earlier reviewer's
fixes too — and the extraction's premise (two callers) was invalidated by a lens fix
that landed after it.

PR #1958, opened 2026-07-23
Eliminate the Atlas flaky-test backlog

srid:
**Flaky tests now wait on observable completion instead of timing guesses**, and the lifecycle bugs those tests exposed are closed across browser e2e, watchers, daemons, transports, and integrations. The branch also makes Cucumber parallelism deterministic from CPU count, so an odu lease gets the same worker count regardless of ambient host load.

### What changed

- Test fixtures wait on durable facts: fake clocks, settled PTY output, hydrated virtualized content, watcher reconciliation, and process-unique paths.
- Kaval teardown, remote-surface liveness, and stale-forward rejection make process and transport transitions definitive instead of leaving race windows.
- Code-tab coverage no longer races lazy Markdown rendering, filesystem watcher installation, branch-fixture setup, or virtualized tail rendering.
- CI derives e2e workers only from CPU cores with platform caps: `kolu-ci-2` resolves to 8 workers and `sincereintent` to 4.
- The `$ci` APM skill records each platform's resolved worker setting and recipe-wall e2e duration in one idempotent PR comment.

### Validation

Focused repeat coverage is green for every repaired cluster. The final gate, `d952b85#1`, passed the complete required pipeline on Linux (`kolu-ci-2`) and macOS (`sincereintent`), with every required GitHub context green and zero hidden Cucumber retry markers. E2E recipe-wall times and the actual parallelism settings are recorded in the metrics comment below.

`just fmt`, `just check`, the Atlas sync/idempotency gate, and the final two-platform odu pipeline are green.

_Generated by Codex (model `gpt-5`)._

PR #1985, opened 2026-07-25
An ACP agent in a tile: @kolu/acp (HX0)

srid:
Driving a coding agent by typing into its TUI means watching a screen and guessing: has the turn ended, was that a paste race, did the tool finish? This PR takes the first step of [chat threads as terminals](https://github.com/juspay/kolu/blob/master/docs/atlas/src/content/atlas/chat-native-agents-and-kolu.mdx) — **HX0** — and deletes the guessing by putting a real protocol on the wire.

`@kolu/acp` ships two bins. **`acp-proxy`** spawns an [ACP](https://agentclientprotocol.com) adapter over stdio and **re-serves the identical protocol on a unix socket**, rendering every frame that crosses either face to its own stdout. **`acp-chat`** is a REPL that connects to one. Run the proxy in a kolu tile and the tile *shows* the session — messages, tool calls, permission decisions scrolling by — while never being the thing you type into: all input arrives as ACP calls on the socket.

```
⎯ adapter spawned · claude-agent-acp (pid 2929196)
⎯ listening · /run/user/1000/kolu/acp-e7f2.sock
▶ session/prompt · "why is the e2e suite flaky?"
◀ agent_message_chunk · It's a port collision: two workers bind 5173.
◀ tool_call · execute — grep -r retry test-results/
◀ session/request_permission · grep … → auto-answered Allow once
◀ tool_call_update · completed
● turn end · stopReason: end_turn
```

## The shape

ACP is the boundary on **both** faces of the proxy, so there is no invented protocol anywhere on the wire, and the official **`@agentclientprotocol/sdk`** is used on both — `ClientSideConnection` toward the adapter, `AgentSideConnection` toward each socket client. That is the same lineage the pinned adapters are themselves built on; the PR started on `@zed-industries/agent-client-protocol` 0.4.5 and moved across in `c758d363` (the [migration comment](https://github.com/juspay/kolu/pull/1985#issuecomment-5078909805) has the detail).

The coupling runs one way only: `acp-proxy` imports nothing from kaval, padi, or `@kolu/surface`, and kolu never learns about it either. **Nothing about the tile is special** — an "ACP tile" is an ordinary kolu terminal whose child happens to be `acp-proxy`. Running it there is a deployment choice; the same binary runs identically under tmux or systemd.

**The adapter is argv.** `acp-proxy --id <id> -- claude-agent-acp` — so agent-agnosticism is structural rather than a feature to add. `claude-agent-acp` and `codex-acp` are pinned npm deps (no global installs); any other stdio ACP agent works unchanged.

## How it honours the design philosophy

- **Fail fast, no fallbacks.** An absent `XDG_RUNTIME_DIR` is an error, never a slide to a world-readable `/tmp`. A permission request offering no `allow_once` fails loudly rather than being widened to `allow_always`. A handshake that stalls is bounded and terminal — a proxy that cannot bring up an adapter exits rather than respawn-looping. There are no override knobs: the grace window and handshake budget are baked constants, not flags.
- **Volatility boundary.** Adapter process lifecycle, transport, and per-vendor quirks live in this package and nowhere else — which is exactly why it is a `@kolu/*` workspace package rather than folded into a future pesu. Both of its faces are standard ACP, so the arrow points out; the package could move to its own repo as a file move, not surgery.
- **Reuse the source of truth.** The official ACP library on both hops rather than a hand-rolled client; `@kolu/daemon-test-gate` for the fork gate; the existing `nix/workspace-tree.nix` + tsx-loader wrapper shape that vazhi and the TUIs already use.

## Verification

The done-criteria are tests, not claims. Each drives the **real bins** — spawned processes, a real socket, the official client library on the far end — against a scripted fake ACP adapter, and the whole suite runs over **three adapter argvs** — the same fake twice (plain and `--verbose`) plus a deliberately different second vendor — which is how "nothing here is claude-shaped" becomes something checked rather than asserted.

| Criterion | Test |
| --- | --- |
| Prompt travels client → socket → proxy → adapter, reply streams back | `carries a prompt to the adapter and streams the reply back` |
| Transcript rendered from `session/update` frames only | `renders the tile transcript from frames alone` — asserts every stdout line is frame-derived, plus a source-level check that no session file is ever read |
| Adapter killed mid-turn → respawn → next prompt works | `respawns the adapter after it dies mid-turn` (external `SIGKILL`) and `…after it exits on its own` |
| `session/cancel`, including grace-then-respawn | `ends a turn the adapter cancels cleanly` and `kills and replaces an adapter that ignores the cancel` |
| Same flows against a genuinely different vendor | the suite is parameterized by adapter argv, and the third fake speaks a different dialect on purpose; the real `codex-acp` is covered by the smoke below |

**48 tests green.** Rendering, socket-path and argv rules run in the ordinary unit lane; the end-to-end suite forks real processes so it is gated behind `@kolu/daemon-test-gate` and runs in CI's `daemon` lane. No agent credentials needed.

**Real-adapter smoke**, run on the dev box (CI has no logins), both reaching `end_turn` by changing only the proxy's argv:

```
=== adapter: claude-agent-acp        === adapter: codex-acp
agent ▸ pong                         agent ▸ pong
```

Two lessons were taken from [buzz](https://github.com/block/buzz), which runs this stack in production: the `allow_once` option is found **by `kind`, never by id or position** (there is a decoy-option regression test for exactly this), and the adapter is spawned as a **process-group leader and killed as a group**, so the tools and MCP servers it spawns are reaped rather than orphaned across respawn cycles.

## Deliberately not in this phase

- **No idle timeout.** A turn that goes quiet during a long tool run is normal, and a wrong timeout kills real work. A dead adapter is detected by its exit, not by silence.
- **No `session/load`.** The adapters advertise it; the proxy does not forward it yet, so it advertises `loadSession: false` rather than offering something it would then refuse. History does not survive a respawn. (HX3.)
- **No fs/terminal client services** offered to the adapter — the proxy is not an editor.
- **Permissions are auto-answered**, as the plan's Decisions table specifies, until HX2 forwards them to the thread.

## Documented as an alpha preview

[`website/src/content/docs/acp.mdx`](https://github.com/juspay/kolu/blob/acp-hx0/website/src/content/docs/acp.mdx) leads with an alpha-preview callout naming the two things a reader would reasonably expect and not find — you cannot answer a permission request yourself, and a session does not survive a restart — and closes with a *Where this is going* section marking the HX1–HX3 arc as future work that is not built, linking the [Atlas note](https://kolu.dev/atlas/chat-native-agents-and-kolu.html) as the plan of record. It stays filed under **Command Line**, not as a "Chat" page: what ships here is the substrate, and a Chat page belongs with HX1 when there is a chat experience to document. That reasoning is recorded in the MDX itself so it isn't re-derived later.

## One follow-up, filed

The SDK marks `AgentSideConnection` / `ClientSideConnection` — the two classes both faces are built on — as `@deprecated`, preferring an `agent()` / `client()` app model. That is a rewrite of the process-supervision core, not a rename, so it is **#1989** with its own review gauntlet rather than an unreviewed second rewrite here. `noDeprecatedImports` is scoped off for `packages/acp/**` only, with the reasoning inline; removing that override is the issue's completion signal.

Evidence was skipped for this PR at the owner's instruction. Two-platform CI is still owed — the darwin builder and the whole linux pool were down for this branch's runs; see the [status](https://github.com/juspay/kolu/pull/1985#issuecomment-5077027043) and [migration](https://github.com/juspay/kolu/pull/1985#issuecomment-5078909805) comments for exactly what did and did not run.


srid:

**Status: applied.** Two Opus lenses read the branch independently — **lowy** 15 findings (13 `fix`, 8 major), **hickey** 23 findings (12 major). They converged on the same reachable defect from different directions, which is the strongest signal this review produces.

### The wedge both lenses found

A replacement adapter dying **inside its own handshake** had its rescheduling swallowed by the very flag that marked a respawn as in flight — and the compensating re-schedule was dead code, because `p.catch(f).finally(g)` runs `f` before `g`, so the flag was still set when the catch consulted it. The proxy was left listening with no adapter, no retry and no fatal report, answering every subsequent prompt "the adapter is not ready" indefinitely. That is precisely the silent degradation the give-up cap exists to prevent, and the cap was unreachable on that path.

Lowy supplied the other half: Node emits `error` and **never** `exit` for a spawn that cannot start, so an accounting hung off `exit` alone never advanced the failure streak either — a vanished adapter binary (a nix GC, a rebuild, a `PATH` change) wedged the proxy the same way. Lowy reproduced both against the real bins.

`#onExit` is now the single place a generation is accounted for and the next attempt scheduled, reached by *both* ways of dying and guarded on the timer alone. **Two regression tests pin it, each verified red against the old code** rather than merely green against the new.

### Applied

| Finding | Origin | What changed |
| --- | --- | --- |
| L3 ≡ hickey headline | both | `#onExit` is the sole scheduler; spawn `error` routed through it; dead `.catch` re-schedule deleted; 2 regression tests |
| L2 | lowy | The proxy **publishes** its session's cwd in `initialize` (`kolu.acp/cwd`); the client reads it. The refusal stays — serving the wrong directory silently is worse — but the check as shipped broke every client not launched from the proxy's own directory, contradicting the README and the docs page |
| L1 | lowy | `connectToProxy` is now the package's client face; the connection rules were hand-rolled three times in one diff and now live once. `acp-chat` and the test client are both thin over it |
| L15 | lowy | An observer client sees the **question**, not just the answer: the proxy broadcasts a `user_message_chunk` for the prompt it received, and `acp-chat` renders through `render.ts` instead of a private switch that silently dropped five of eight frame kinds |
| L13 | lowy | Client fan-out reports an undeliverable write instead of `void`ing it |
| L8 ≡ F12 | both | `claimSocketPath` regains the two checks surface's claim has and this copy dropped: never unlink a non-socket inode; verify the parent dir really is private (`mkdir`'s mode is a no-op on an existing one) |

Already fixed before the reconcile pass, recorded so they are not re-raised: `acp-chat` double-reporting a clean disconnect, and the nix wrapper's PATH shadowing.

### Left for the human

Not applied — these are scope and taste calls, not defects, and applying them would grow a PR that is already large:

- **L4 / F-series — respawn policy as its own component.** Both lenses note the policy (backoff, streak, cap) is smeared across the supervisor's mutable fields rather than being a value with its own tests. *Lowy:* "it will be paid for at HX2, and was already paid for once during this build." *Against:* it has exactly one consumer today, and the behaviour is now pinned by tests at the process level.
- **L5 — permission answering as an injected strategy.** The one roadmapped axis of change (HX2 forwards permissions to a chat thread) is welded into the supervisor. The seam is real; whether it is bought now or at HX2 is a judgment call.
- **L6 — the supervisor imports the renderer's event type**, which lowy reads as a layering inversion (a vocabulary module would sit under both).
- **L12 — "two adapter argvs" varies a flag, not a vendor.** The suite's agnosticism proof runs two *styles of the same fake*. The real second-adapter proof is the out-of-band `codex-acp` smoke, which CI cannot run without credentials.

### Electricity probe

Lowy's verdict, worth recording because it corrects the framing rather than the code: `@kolu/acp` is **not electricity** — test ① fails, since the receptacle's interface *is* ACP, a coding-agent domain protocol. It is agnostic of kolu (nothing terminal/canvas/git leaks in; the dependency arrow points strictly out; zero workspace runtime deps) but not of the domain. That puts it on the `@kolu/terminal-protocol` appliance tier, not the `@kolu/surface` transport tier — which is how `architecture.mdx` already words it. The receptacle framing applies *inside* the package, where `AdapterSession` is well-named and well-placed.

<sub>Note on process: the reconcile-and-apply subagent stalled without producing output and was stopped; the findings were merged and applied directly instead. Both findings files are on the branch author's worktree at `.lens-debate/`.</sub>

srid:
## Codex ⇄ Claude debate

✅ Consensus in 3 round(s) — approved.

> "F4 is closed: a ready generation's transport failure is now reported as `adapter lost`, while failures before readiness remain `adapter failed to start`. The change introduces no new defect, and the package typecheck plus all 46 gated tests pass."

Four findings raised, four accepted — none disputed. Three of the four were
lifetime holes that only exist between real processes, which is the class a
second model family earns its keep on.

|---|---|---|
| 1 | `b5664ca5` | close the three lifetime holes codex found |
| 2 | `99b20c80` | an adapter that ran for an hour did not "fail to start" |

- **F1** — a natural adapter exit reaped only the leader, leaving its tools and MCP servers alive in the old process group. One leaked process tree per crash-and-respawn cycle, against a containment guarantee the README makes.
- **F2** — losing the ACP stream was not treated as losing the adapter. An agent that closes fd 1 and keeps running ends the library's receive loop, which then retains its pending requests forever; no `exit` or `error` fires, so the generation stayed current with a turn that could never settle.
- **F3** — `close()` suppressed the very rejection that bounds a request. The `leaving` flag added earlier in this PR (to stop a clean quit reporting a false disconnect) also stopped a local close from settling in-flight and queued turns, so `prompt()` then `close()` awaited forever. Self-inflicted, two commits old, and invisible to the tests as they stood.
- **F4** — transport loss was reported through the spawn-failure channel, so an adapter that became ready, served turns, and then lost its stream was announced as `adapter failed to start`, contradicting the `adapter ready` line above it in the same transcript.

Each fix carries a regression test: a fake that leaves a child in its process
group and exits, one that drops its protocol stream while staying alive, a turn
closed out from under its caller, and an assertion on the exact rendered cause.

srid:
## Status — review complete, CI blocked on the darwin builder

**This PR is finished and reviewed. It is not merge-ready tonight only because the macOS build host went down mid-run.**

### The review gauntlet

Four passes, run serially, each editing the branch in turn. Between them they found **eighteen defects that survived adversarial verification** — almost all lifetime or ordering faults that only exist between real processes, which is the right thing to have hunted in a package whose entire job is supervising a child process.

| Pass | Result |
| --- | --- |
| Architecture (C1–C7, 7 hunters + per-finding refuters) | 7 confirmed; 2 reproduced against the real bins |
| Lens — lowy ∥ hickey, independent | 15 and 23 findings; **converged on the same wedge from opposite directions** |
| Codex debate | 4 findings over 3 rounds, all accepted, **approved** |
| Simplify (4 angles) + code-police | applied; 1 police finding skipped with reason (below) |

The single most valuable finding, reached independently by both lenses: an adapter dying **inside a respawn's own handshake** had its replacement silently swallowed by the flag that marked a respawn in flight — and the compensating re-schedule was dead code, because `p.catch(f).finally(g)` runs `f` before `g`. The proxy was left listening with no adapter, no retry and no fatal report. Codex then supplied the other half: Node emits `error` and **never** `exit` for a spawn that cannot start, so a vanished adapter binary wedged it the same way.

Every fix carries a regression test, and the two for that wedge were **verified red against the old code** rather than merely green against the new.

### Verification

- **46 tests**, driving the real bins over a real unix socket — spawned processes, the official ACP client on the far end, nothing shimmed in-process.
- The suite runs over **three adapters**: one fake twice, plus a second fake that is a genuinely different agent (other capabilities, batched rather than streamed replies, the permission request on the far side of the tool call, frame kinds the pinned library drops). It earned its place immediately by catching this suite asserting one vendor's literal wording for its `allow_once` option.
- **Real adapters**: `claude-agent-acp` and `codex-acp` both reach `end_turn` through the nix-built binaries, changing only the proxy's argv.

### CI

At `99b20c8` — the last run that completed — **37 of 38 green**, including:

- `ci::daemon` green on **both platforms** (that is this package's suite),
- `ci::nix@x86_64-linux` green (it builds `acp-proxy` and `acp-chat`),
- `unit`, `biome`, `fmt`, `atlas-sync`, `pnpm-hash-fresh`, `e2e@x86_64-linux`, `smoke`, `home-manager` — all green.

The sole red was `ci::nix@aarch64-darwin`, failing with ~5,700 × `No space left on device` unpacking the pnpm store: `/nix` on `sincereintent` was at 89%. On the retry it hit the same wall and then **the host stopped answering entirely** — ssh times out, ping is 100% loss, tailscale still reports it up. That took three **linux** lanes with it, which now read `Errored (venue lease lost on sincereintent)`.

**Every current red is `Errored` (infrastructure), not `failed` (a test). None points at this change.** `sincereintent` needs disk freed and probably a reboot — a full `/nix` on macOS can wedge the machine, which matches it going dark right after the second exhaustion.

Also logged this run: the darwin `integrations/git` fs-watch flake **recurred** after being marked fixed by #1958, on two adjacent subtests. Its tracker row is reopened and cites this PR; both went green on same-SHA reruns and the diff touches no `packages/integrations/git` file.

### Known and deliberate

- **No idle timeout.** A turn that goes quiet during a long tool run is normal, and a wrong timeout kills real work; a dead adapter is detected by its exit, not by silence.
- **No `session/load` forwarding**, so `loadSession` is advertised `false` rather than advertised and then refused. History does not survive a respawn.
- **Permissions auto-answered** with the `allow_once` option — matched by `kind`, never by id or position, with a decoy-option regression test — until they are forwarded to a thread.
- The pinned ACP library validates `session/update` against its v1 schema and **drops kinds it does not know** (`usage_update`, `session_info_update`, which both shipped adapters send). Documented in the package README, and the second fake asserts it is harmless.
- `socketPath.ts` re-derives ~15 lines of path arithmetic rather than importing `@kolu/surface/unix-socket`, so the package keeps zero runtime coupling to the framework; an agreement test pins the duplicate against `getRuntimeSocketPath`.
- **Skipped, with reason:** code-police's `prefer-ts-pattern` on `render.ts`'s two dispatch switches. `ts-pattern` is a real repo convention (39 files), but adopting it here adds a runtime dependency to a package whose stated property is being dependency-light and movable, and the switches are already exhaustive under TS. Worth a maintainer's call rather than a 1am refactor.

Left in **draft** deliberately.

srid:
### Correction on the CI diagnosis

Amending my previous comment: **the "unreachable host" half of it was wrong, and the fix is probably smaller than I said.**

Two separate things happened, and I conflated them.

**Real, and still true:** `ci::nix@aarch64-darwin` failed with ~5,700 × `No space left on device` unpacking the pnpm store. That was measured — `/nix` on `sincereintent` was at 89% with 13 GiB free — and it reproduced on a retry. That lane needs disk before it will pass.

**Wrong:** I reported the box as having gone down and needing a reboot. Shortly afterwards the **entire `x86_64-linux` pool** became unreachable too — all eight boxes failing ssh at banner exchange — and this machine's tailnet shows numerous peers, including exit nodes, as `offline, last seen ~8m ago`. So the far more likely story is a **network outage local to this host**, not a wedged Mac. `sincereintent` may well be fine.

That also explains the three linux lanes reading `Errored (venue lease lost on sincereintent)`: they lost their venue when connectivity went, not because of anything in this change.

I attempted the re-run of just those three linux lanes as a check; odu reported `no reachable host in x86_64-linux pool`, which confirms the outage rather than anything about the code.

**Net:** the only CI finding that survives against this branch is the darwin disk exhaustion. Everything else red is connectivity. Re-running CI once the network is back should be the whole story, and there is nothing here for a reviewer to act on in the code.

srid:
## `@agentclientprotocol/sdk` migration — and a CI note worth reading

`c758d363` moves the protocol library from `@zed-industries/agent-client-protocol` 0.4.5 to **`@agentclientprotocol/sdk`** on both faces — the lineage the adapters this package pins are themselves built on.

The swap needed **one** code change to compile: the `session/update` union went from 8 kinds to 13, so the transcript's switch was no longer exhaustive and TypeScript refused it. That is the entire code surface of the library change.

### Both documented behaviours had changed

Each was re-verified against the SDK's installed source **and** by probe, not taken from release notes.

**In-flight requests are now rejected when the stream ends.** `Connection.receive`'s `finally` calls `close()`, which rejects every pending response with `ACP connection closed`. 0.4.5 left them pending forever — that gap is exactly what the generation model was built around, so `adapter.ts`'s header now states the new truth instead of asserting the old one. The generations stay: they keep a dead adapter's frames out of the session that replaced it, and they know whether a death means `cancelled` or failure. Neither was ever the library's job.

**`usage_update` and `session_info_update` are delivered.** The old schema predated them and rejected them outright, so frames *both* shipped adapters send reached nobody. The tile renders them now and a test asserts it.

### Three latent defects the change surfaced

1. **Cancel outcomes rested on tick ordering.** With the SDK rejecting the prompt itself, it settled the turn as a failure before `#onExit` could call it `cancelled` — three cancel tests went red. `#settle` now decides that a turn ending while we are killing to enforce a cancel *is* cancelled, whichever mechanism noticed. That removes the race rather than depending on winning it, which is what the lens review had flagged as fragile.
2. **A death could go unreported.** When the library noticed first, the handshake catch set `#stopped` and `#onExit` bailed before emitting, so the transcript ended at "adapter spawned".
3. **A boot-time death is never seen by `#onExit` at all** — the proxy exits faster than Node delivers the child's `exit` event — so the handshake failure now reports itself.

### Deprecation, and the follow-up

The SDK marks `AgentSideConnection` / `ClientSideConnection` — the two classes both faces use — as `@deprecated`, preferring an `agent()` / `client()` app model. That is a rewrite, not a rename: handlers registered by ACP method name, across both faces and both fixtures. `noDeprecatedImports` is scoped off for `packages/acp/**` only, with the reasoning inline, and the migration is **#1989**, which gets its own review gauntlet. Removing the override is its completion signal.

### CI status — read this carefully

**17 of 20 `x86_64-linux` lanes pass**, including `ci::nix` (which builds both bins and typechecks the workspace), `biome`, `fmt`, `atlas-sync`, `smoke`, and the website and surface-example lanes. `ci::e2e` ran **528 of 529 scenarios green**.

Three lanes are red — `ci::unit`, `ci::daemon`, and that one e2e scenario — and **all three are timeout-class failures in code this PR does not touch**:

- `daemon-test-gate`'s own AST meta-lint, which walks every test file in the workspace, missing a **5-second** budget at 8.1s.
- Heartbeat and liveness timeouts across `surface`, `surface-app`, `testhost` and `padi-local`.
- One Code-tab directory-node file-watch assertion, in a PR whose only client change is adding a string to `DOC_SLUGS`.

**This package's own suite was re-verified at final HEAD: 48/48 green**, run through the same leashed recipe (`KOLU_DAEMON_TESTS=1 KOLU_DAEMON_BIND_PID=…`), and notably it passed *at load 37–50* — the same contention that broke the 5-second budgets elsewhere. Earlier, on a real pool box at `fa44bf9`, it was 46/46 inside CI's own `ci::daemon` lane.

Why the venue matters: every off-machine venue is currently unreachable — the darwin builder is out of disk, and all six `kolu-ci` pool boxes plus the `pu` host are behind a network outage — so this run happened on a developer workstation that was simultaneously hosting **three** other sessions' CI. Load reached **100**. `ci::e2e` was not slow but *blocked*, waiting correctly on the shared suite lock held by another session.

**A whole-workspace re-run on the canonical pool is explicitly owed when infra returns.** Nothing above should be read as clearing those three lanes on this branch's behalf — only as showing that what this PR changes is green, and that what failed is elsewhere and load-shaped.

**`aarch64-darwin` is absent by explicit owner instruction** — the builder is down, and this run was scoped to linux only. A two-platform green is still owed before merge.

srid:
Closing unmerged, by owner decision, after grounding the design against Buzz's production architecture (buzz-acp owns adapter stdio directly; no proxy, no socket anywhere).

The premise this PR was built on — "the tile shows the session, click through for depth" — does not survive ACP: with a headless adapter there is no interactive CLI session in the tile, so the proxy's rendering is a read-only frame log of the same stream the chat client already has. That makes the proxy + socket + companion REPL the wrong skeleton; the Buzz shape (ACP client embedded in the chat daemon) is simpler and loses nothing usable today.

What survives: the supervision core (group-kill, graced cancel, paced respawn — independently validated by buzz-acp's near-identical choices) and the real-process test harness carry forward as the seed of an embedded client library. The replacement plan is the rewritten Atlas note (PR to follow).

PR #1991, opened 2026-07-25
Click a port, get the page: kolu embeds @kolu/port-forward (PRT2)

srid:
> ## ⚠️ Production incident — fixed in `c843d58e`, read this first
>
> The first deploy of this branch **froze the whole kolu server on the first forward click**: UI dead, HTTP dead, `SIGTERM` ignored until systemd sent `SIGKILL`. Reproduced twice. Three defects, all fixed here, each with a regression test verified red against the unfixed code before being taken green.
>
> **The freeze was a cycle between two individually-reasonable halves.** The `forwards` cell is a poll on a *fused* cadence (`everyMsOr(interval, onChange)`) — the interval runs the auto-cancel pass, the change edge publishes an act immediately. Separately, the auto-cancel pass ended by announcing its result. Together: the pass announced on the very edge that triggers the pass, so `read → reap → announce → tick → read` spun the event loop with nothing else ever getting a turn. It was invisible until the first click because with **zero** `auto` forwards the pass returned *before* announcing. Fixed by deleting the shape rather than guarding it — the reap and the read collapse into one `reconcile()` that **reports by returning**, so "reap and announce" is no longer spellable. *No unit test of either module could have caught this; the regression test asserts a BOUNDED read count over the wiring itself.*
>
> **The forward that did open served nothing.** `127.0.0.1` and `::1` are both `loopback` and are **not the same address** — the scanner folded them into one `scope`, and both mechanisms then dialled v4. A dev server on `[::1]:5173` (vite's default) got a tunnel that came up perfectly onto an address with no listener behind it. `PortInfo` now carries `family` beside `scope`, and `ForwardTarget.loopback` is **required** so no call site can omit it.
>
> **Same lesson twice in one PR**, and the note now says so: a fact the OS handed us was collapsed below what the consumer needed, and the collapse stayed invisible until something dialled the answer. `wildcard` lost the interface case; `scope` lost the family.
>
> **And kolu offered to forward a port to the machine you were already sitting at** — browsing from *zest* while *zest* was also one of kolu's hosts meant every chip proposed a door on *pureintent* so that zest's browser could reach zest. Now recognised from the viewer's connection address; every failure to recognise keeps the forward, which works.
>
> **Follow-up (`a9134486`): that recognition never fired in the field, and the redeploy proved it.** kolu is reached through `tailscale serve`, which terminates TLS and dials the backend from the host's **own** tailnet address — so the only connection to kolu had peer `100.122.32.106` (pureintent itself) and zest's `100.90.229.113` never appeared as a TCP peer at all. It was in an `X-Forwarded-For` header nothing read. The address is now taken from that header **when, and only when, the direct peer is a trusted local hop** (loopback, or one of this host's own addresses). A header on an untrusted connection is ignored entirely — anyone who can reach kolu directly can invent one — and the **last** entry wins, since each proxy appends the address it received from while the leftmost is whatever the client typed. No trusted-proxy list to configure: the trust condition is a fact about the connection.

**PRT2 of the [port-forwarding plan](https://kolu.dev/atlas/port-forwarding.html): every port chip in the Inspector opens.** Click a loopback-only port, or one on a remote host, and kolu opens a door for it on the machine serving the page, then opens the tab. Prerequisites #1975 (the library) and #1982 (detection) are both in master.

This PR started as the plan-of-record update below and now carries the implementation.

## What ships

- **The kolu server embeds `@kolu/port-forward`.** Forward state is a `forwards` cell on kolu's existing surface with `forwards.create` / `forwards.cancel` procedures — no new socket, no new daemon. The listeners are sockets in the kolu-server process, so they die with it: a deploy can't orphan a door.
- **Chip click → lazy forward → open.** Only the ports you actually click get a door. The chip then carries a `⇄ :<localPort>` badge — the number in the URL.
- **Forwards, wherever they are relevant.** One Inspector ports section where a forwarded port is a single row carrying its door inline (open, copy `⧉`, cancel `⨯`), the full list in the host tab's dropdown, and a ring on that host's connection dot while doors are open. *(Superseded the original two-groups + `⇄ n` chip design — see the refinements below.)*
- **⌘K → "Forward a port…"** for anything the scanner can't see, plus a discoverability tip.
- **Auto-cancel policy.** An `auto` forward (chip click) closes when the scanner positively reports its port gone; a `manual` one persists until cancelled or its host leaves the pool.

## Field-testing refinements + surface upstreaming (`e773758b`)

Three more from the human field-testing the deploy, each red-first:

- **The host tab's `⇄ n` chip became a RING on the connection dot.** The dot already opens the dropdown the rows live in. The ring composes AROUND the pip and never touches its colour — pinned over two pip tones, because a forward marker that recoloured it would be a green-over-a-dead-link dot by a new route.
- **ONE ports section.** A forwarded port appeared twice (a chip in PORTS, a row in FORWARDED PORTS); it is now one row carrying its door inline. What the second group was RIGHT about is kept — a forward is a fact about the HOST, so a ⌘K forward or one whose listener died trails the same section as a subdued row rather than vanishing.
- **A reactor loop guard for the class that froze production.** Three consecutive re-reads triggered by an edge fired *during* the previous read, each value-equal, throw naming the source. Value-equality is the discriminator: a genuine burst of distinct values through a fused cadence is exactly what `everyMsOr` is for. Proved by removing it — **without the guard the guard's own suite hangs**, which is the freeze. Plus two doctrine rules in *Reactive honesty*.

**Two upstream items are deliberately NOT in here**, and the Atlas note says so rather than leaving them looking pending. `Observed<T>` would ride ONE real consumer, not three: `ProcessRss` is a deliberate three-way (`ok|absent|error`) that cannot fold into `known|unknown` without destroying a distinction *and* moving a wire shape, and `TerminalPorts` would need its `list` field renamed — both forbidden by the brief's own clause. Moving `viewerHost` into `@kolu/surface` makes it public API and trips the **drishti companion-repo gate**, which a branch that stops at push cannot satisfy; it should ride its own paired PR.

> ✅ **drishti pair: [srid/drishti#121](https://github.com/srid/drishti/pull/121)** — required because this PR changes `@kolu/surface` (`PollSourceOptions` gains `label`; every poll source gains the runtime loop guard; new `viewerIdentity` module). Pinned at `15193a03` and green there: typecheck, 190 tests, fmt-check, and a full `nix-build`. Drishti needs no behavioural change — and that is a *finding*: its `ps`+`lsof` poll on a 2 s interval was exactly the false positive of the guard's first cut, which is why the shipped one is built on provenance. `ref-surface.mdx` is updated in-PR per the sibling rule.

## The deferred shape question, settled: `wildcard` → `scope`

PRT1's review left this to PRT2 on the grounds that the second consumer would decide it. It arrived, and settled it on grounds no structural argument had reached: **both forward mechanisms dial `127.0.0.1` on the far side.** So `wildcard: false` covered two cases wanting opposite handling — a loopback bind (exactly what a door reaches) and a bind to one specific interface like `192.168.1.5` (reachable at that address, and by *no* door). Under the boolean, clicking the second would open a listener that came up fine and then refused every connection through it.

`PortInfo` now carries `scope: "any" | "loopback" | "interface"`, `isAnyAddress` becomes the single `addressScope` classifier both platforms still share, and `portReach` grows a fourth honest arm (`no-mechanism`) that offers nothing because there is nothing to offer. padi's surface goes **4.2 → 4.3**; the existing convergence machinery drains a straddling padi in both directions, so the reshape needs no compatibility arm and deliberately has none. Verified against the real OS, not just fixtures — the live suite binds a real routable interface address and asserts it classifies as `interface`.

## One library change, and it went into the map

`@kolu/port-forward`'s forward map now remembers the local port each target last answered on, so a restarted dev server comes back on the URL you already have. That belongs to the map, not to a consumer: kolu and vazhi would otherwise each keep the same table beside the same map and get its eviction rules subtly different. `createForwardManager`'s signature is unchanged, so vazhi is untouched.

Both shape changes the plan *predicted* a second consumer would force — a `list()` subscription and a tagged error union — turned out **unwanted** once that consumer arrived. The note records that rather than shipping structure with no reader.

### The three incident regressions, each verified red first

| defect | test | red against the unfixed code |
|---|---|---|
| P0 freeze | `forwards.test.ts` › "CONVERGES with a live auto forward — the production freeze" (+ 3 siblings) | `runaway: true` — the read count blows past the cap, which is the freeze itself |
| P1 `::1` dial | `relay.test.ts` › "serves a server bound to the V6 loopback"; `scan.live.test.ts` › "tells a V6 loopback listener from a v4 one" (against the real kernel) | relay fails to fetch; `{scope:'loopback',family:'v4'}` ≠ `family:'v6'` |
| P2 pointless hop | `PortsSection.test.ts` › "opens on the VIEWER's own loopback…" (+ sibling) | `{kind:'forward'}` ≠ `{kind:'viewer'}` |
| P2 follow-up: proxy | `viewerHost.test.ts` › the field topology by its measured addresses (+ spoof / chain-order / no-header / malformed inverses); `viewerHostRoute.test.ts` › "hands on BOTH the direct peer and the forwarded header" | 10 + 2 red — **written before the fix this time**, per the standing requirement |

Plus `viewerHost.test.ts` for the address seam itself (match → direct, no match → forward, and every declining case), and an e2e scenario forwarding a `[::1]`-bound listener through the relay.

- **Unit**: the scope+family classifier (both arms live against the real OS), the widening fold, `portReach`'s four arms, the forward policy (auto-vs-manual death, "only a real observation may close a door", host departure, a door that refuses to close), the restart-stability property in both the map and the relay, and the ⌘K target parser.
- **e2e**: the **relay** case end to end — a real loopback listener in a real PTY, a real click, and the listener's own body arriving in the opened tab through a port it never bound; plus the cancel twin, which proves the door is *shut* by dialling the recorded URL from node rather than by watching a row disappear.
- **Not covered here, deliberately**: the `ssh -L` case needs a second machine, and a loopback ssh hop standing in for one would exercise a different code path from the one it claimed to cover. It stays with the `remote-host-testing` harness, as does the kolu-level SIGKILL criterion. Both are named in the note's done-criteria.

## ODU-IMPACT VERDICT: `none`

No `@kolu/surface*` API changed — this adds members to kolu's *own* surface spec and one field to `@kolu/port-forward`'s mechanism contract. Grounded by an actual grep of odu's source tree (`juspay/odu@3f6ec33`, which pins kolu at `3ff256f0`) for `port-forward`, `port-scan`, `ForwardMechanisms`, `PortInfo` and `portReach` across its `.ts`/`.tsx`/`.json` outside `node_modules`: **zero files matched**. The drishti companion gate does not apply for the same reason — no shared surface API moved.

---

## Plan-of-record update (the original scope of this PR)

The Atlas note marks PRT1 shipped (#1982) and PRT2 shipped (this PR), leaving PRT3 (embedded preview) as the only phase left, deferred by design. The note also gains the settled shape question above, the recorded no-change outcomes for the two predicted API questions, and an explicit statement of what the automated suite covers versus what stays with the remote harness. `@kolu/port-forward` gets its first row in the [electricity](https://kolu.dev/atlas/electricity.html) table — its test ③ is now **proven** rather than pending, since vazhi and kolu's Inspector are two genuinely independent apps on one library.


srid:

**Status: applied** · lowy **10** findings (4 major) · hickey **16** (14 `fix`) · both read the full diff independently, in parallel · base `c772c07`

Both lenses independently reached the **same headline**, which is the strongest consensus signal this review can produce.

### The headline — `≡` both lenses

`portSampler.ts` hand-rolled a cadence runner whose every policy already lives in `@kolu/surface`'s reactor: the unref'd interval (`everyMs`, *"the one home"*), the interval-plus-edge fuse (`everyMsOr`, *"the one home… previously duplicated as app-local twins"*), the non-overlap guard, the mid-flight latch, and log-skip-continue. padi's own `servePadi.ts` consumes it twice, and `memorySampler.ts` — cited by the new module for its 5-second number — documents that its hand-rolled loop was **deleted** in favour of it. `packages/server/src/seal.test.ts` seals this with a **zero-allowlist** ban: *"a hand-rolled setInterval sampler is the retired defect."*

The new module slipped past that seal only because both of its coordinates moved — into `padi` (which the seal doesn't walk) and onto `setTimeout` (which its regex doesn't match). lowy hedged that the reactor lacked one capability, mid-flight edge coalescing; it does not (`reactor.ts:383` latches `dirty`, the `finally` at `:404` runs the trailing read), so the conversion kept nothing. The only genuine residue is the **≥1 s floor**, which stays padi-local — putting it in `@kolu/surface` would trip the paired-companion-PR rule for no benefit.

### Applied

| # | Fix | Raised by | Commit |
|---|---|---|---|
| 1 | Drive the port scan from the reactor's poll source | `≡` both | `bc81cffc` |
| 2 | One ANY-address classifier for both platforms — the two disagreed on `::ffff:0.0.0.0` | hickey | `deb238e9` |
| 3 | `foldPorts` picks its name deterministically, not by iteration order | hickey | `19562562` |
| 4 | `portsEqual` reads its comparison keys off `PortInfoSchema` | hickey | `bdce4a7e` |
| 5 | The per-terminal join written once, behind a platform reading | hickey | `3d5dfdc8` |
| 6 | Key the scan on root pids, not the caller's `TerminalId` | hickey | `5452a8d2` |
| 7 | Tag `PortScanError` so a permanent failure can't become a log loop | lowy | `d76b9332` |
| 8 | The reachability decision is a tag, not a sentence | `≡` both | `c75afbbd` |
| 9 | The store owns what a tile's PANES are | `≡` both | `e1e66b25` |
| 10 | State the active-arm rule once, where it is correct | hickey | `22918a08` |
| 11 | `seedSnapshot` really is the ONE home for the snapshot defaults | lowy | `d777efa8` |
| 12 | `MetadataInspector`'s header states the boundary it actually holds | lowy | `bd2d1451` |
| 13 | No comment still describes the mechanism removed for lying | `≡` both | `51843157` |
| 14 | Stop claiming the socket↔pid reading is shared; settle two of its drifts | lowy | `c9adccf5` |

Two of these were latent **bugs**, not tidiness: the two ANY-address predicates already disagreed on `::ffff:0.0.0.0` (the exact disagreement `PortInfo.wildcard`'s docstring says the field exists to prevent), and `foldPorts` picked `name` by iteration order while `portsEqual` treats name as identity — so two programs on one port would have churned a "change" forever, defeating the dedup the sensor exists for.

### Settled by experiment, not argument — valve 1

**`TerminalLifecycle.rootPid` duplicates `entry.info.pid`** (lowy `drop` ↔ hickey `fix`). The discriminating experiment: exactly two `startSnapshotSensors` sites, each passing the same expression written to `entry.info.pid` with no `await` between, and `teardownSensors` always deletes the lifecycle — so the two homes cannot disagree today, which hickey concedes ("latent fragmentation rather than a bug"). That settles the facts in lowy's favour and leaves a pure remedy disagreement.

**Adjudicated: not changed.** Reading the pid back through the registry would trade a *proven-consistent* copy for a new failure mode needing its own policy (what does a missing entry mean mid-pass?) — and getting that policy wrong is how the `sudo` blindness bug happened. Recorded rather than churned.

### Adjudicated to the plan — valve 2

Both judgment-shaped findings are now recorded in the Atlas note (`2e2c7ce1`), where PRT2's implementer meets them, rather than left in a comment:

- **`wildcard: boolean` vs `scope: "any"|"loopback"|"interface"`** (hickey). The case is sound — one observation doing duty as a judgment, and the judgment isn't a function of that field alone (the client joins it with the terminal's host). Kept for PRT1 because both live symptoms are fixed independently, and PRT2 is where the second consumer appears: relay-vs-`ssh -L` needs the host, which `wildcard` never carried, so PRT2 reshapes this surface anyway. Same discipline the note already applies to `list()`/`subscribe()`.
- **One fan-out for kaval's activity edge** (hickey now, lowy at the fourth consumer). Declined: the three consumers are not three of one shape — the live-dots subscription is per-*subscriber* by design while the other two are daemon-lifetime, so a shared receptacle would force the odd one in.

### Recorded as observations

`portScan.ts`'s **file split** (a 3-file split is an import cycle; a 4th shared-core module is layout taste), lowy's **seal extension** (padi now legitimately owns one `setTimeout` — the floor — so a zero-allowlist seal would forbid what we deliberately kept), the client's `?? []` (the two absences are provably different facts), and `portNudgeAbort`'s unused lifetime (subsumed by the fan-out question above).

Verification after every change: `just check`, `just test-unit`, the padi + supervisor gated suites under `KOLU_DAEMON_TESTS=1`, and `just atlas::check-sync`. Nothing pushed by the review itself.

srid:
## Grok ⇄ Claude debate

✅ Consensus in 3 round(s) — base `6d36655c`

|---|---|---|
| 1 | 1891f95e | fix(debate): five findings from grok round 1 — three of them lifetime bugs |
| 2 | 50a9f789 | fix(debate): grok round 2 — the discovery path, and two stale claims |
| 3 | 20cf06d8 | docs(debate): reattach hostDeparted's JSDoc, orphaned by heldHosts |

- **F1** — foldPorts widens scope and family independently
- **F2** — pool.subscribe only runs hostDeparted over forwards.list(), which is open slots only
- **F3** — One known terminal makes the whole host reading an observation (sawAnything=true), and unread terminals (deadline, !present) contribute nothing
- **F4** — pool.subscribe invokes hostDeparted once per listed forward on a departed host, not once per host
- **F5** — PortRow's type union carries a second kind:"port" arm with a name field that portRows() never constructs and no reader uses
- **F6** — Regression from the F3 framework change (docs, not runtime of the race itself)
- **F7** — Regression from the F3 framework change

srid:

Pass 1 (rules) found **7**; pass 2 (fact-check) came back **clean**. Elegance skipped — `/simplify` ran over the same tree one commit earlier.

Fixed in `b0596722`:

| Rule | Where | What |
|---|---|---|
| `no-unbounded-growth` | `port-forward/manager.ts` | `lastLocalPort` claimed a bound in prose and enforced none — capped at 512, oldest-first eviction |
| `caught-error-must-not-collapse-to-empty` · `no-overloaded-null` | `client/forwards/useForwards.ts` | a caught RPC failure reported only to `console.warn` and returned the same `null` the server gives for its own honest "cannot tell" |
| `prefer-ts-pattern` | `terminal-vocab/schema.ts` | `portReach`'s if-cascade fell through to `direct` — the most consequential wrong answer it can give. Now `match(…).exhaustive()` |
| `prefer-ts-pattern` | `right-panel/PortsSection.tsx` | `openAt`'s comment claimed a missing arm was already a compile error; it wasn't. Now it is |
| `errors-must-log-at-error` | `server/portForward/forwards.ts` | a `degraded` loss — a listener that broke and could **not** be cleaned up — logged at `warn` beside benign `gone` losses |
| `callback-fanout-guarded-at-funnel` | `server/portForward/forwards.ts` | `notify()` fanned out unguarded; `hostDeparted` is fire-and-forget, so a throwing subscriber surfaced as a fatal unhandled rejection |
| `no-untyped-escape-hatches` | `common/surface.ts` | a `b[i]!` under `noUncheckedIndexedAccess` |
| `no-thin-wrapper-functions` | `server/index.ts` | three single-arg pass-throughs beside an `onChange: forwards.subscribe` that passed by reference |

The viewer-host one overrides a call I made earlier in this PR — I had argued a console line was enough because "the user has nothing to do about it". The rule is right and I was reasoning from convenience.

PR #2003, opened 2026-07-26
Hosts that can never connect now fail fast, instead of retrying forever

srid:
Point kolu at a host that asks for an ssh **password**, one whose **host key** you have never accepted, or one with **no Nix installed**, and the canvas sat on "Connecting to remote:petit…" forever. None of the three could ever succeed — yet each was classified as a transport blip, and **transport blips retry forever**. After the boot deadline the escape card claimed kolu was "still retrying": true of the mechanism, false of the outcome. The real reason existed only in a log tail.

These are exactly the [three documented prerequisites](https://kolu.dev/remote-hosts) for a remote host. Two of them going unmet produced an infinite spinner rather than an answer. **All three now stop in seconds on a card that names which one it is and how to fix it.**

The fix is a classification, not a new mechanism. The arch probe is every dial's first contact with a host, so it is the one chokepoint where all three surface.

| resolution `kind` | detected by | card |
| --- | --- | --- |
| `auth-refused` | `Permission denied (…)` + ssh's exit **255** | **This host needs passwordless ssh** — `ssh-copy-id`, then reconnect |
| `host-key-unverified` | `Host key verification failed` + exit **255** | **This host isn't trusted yet** — `ssh <host>` once to accept the key |
| `nix-unavailable` | the shell's exit **127** | **This host has no Nix that kolu can run** — [install it](https://nixos.asia/en/install), or put it on the non-interactive PATH |

All three carry `failureCause: "remote"` (the peer *was* reached — it just can't serve us) and `terminal: true` (no redial can supply a credential, a trust decision, or a missing binary). They are kept apart rather than lumped together, because **the remedies are different** — a card telling you to do the wrong thing is barely better than no card.

Anything genuinely transient — asleep, unreachable, still building — keeps retrying exactly as before.

### Detection is deliberately conservative

A false terminal verdict strands a healthy host, so each signal is the narrowest one that can't be forged:

- **The ssh refusals require ssh's own exit 255 as well as the stderr text.** A remote command's stderr rides the same stream, so `nix-instantiate: … Permission denied` must not be mistaken for ssh's. Only ssh's auth message carries the parenthesised method list, which is what the pattern keys on.
- **The Nix case is matched on the exit code alone, never on prose** — bash says "command not found", dash "not found", fish "Unknown command", and all three exit 127. A prose matcher would only add ways to miss one and silently restore the eternal retry. It needs no 255 companion because 127 *is* the remote command's own status. A `nix-instantiate` that ran and failed exits non-127 and stays retryable.
- **An unrecognised failure stays untyped and retried** — the safe default in every case.

The probe-only chokepoint is complete, not merely convenient: a refusal met later (Nix's own ssh fork, the agent dial) kills that dial, and the redial's probe meets it here within one retry.

> One latent bug fell out of the same seam. `computeEntryFailedDetail` checked the `link-failed` convergence banner *before* the dial's own finding — but a terminal give-up always raises that banner, so the finer cause was masked by the generic "can't reach this host". A standing drv fault now outranks it, which also unmasks `agent-source-unbaked` at give-up.

### Why this lives in `@kolu/surface-remote`

The volatility is the ssh transport's, not kolu's — so the terminal verdict is minted at that boundary and **every consumer inherits it without a line of their own**: drishti's fleet and odu's lanes stop redialing a host that can never come up, and `kaval-tui` / `padi-tui` already print the underlying message verbatim, so they now carry the remedy instead of a bare exit code. kolu's binder does the one thing that *is* kolu's: enrich the framework verdict with padi's own domain causes for the host-down card. It never reclassifies retry policy or terminality.

### Downstream impact

`ResolveDrvFailure` is **not** re-exported from the package index (only the `ResolveDrvError` class is), so no external consumer can name the union — and a grep of both trees found **no exhaustive match** over the resolution kinds in either. The compile break kolu hit in `remotePadiBinding.ts` was reachable only from inside this monorepo, where that file imports the module directly.

| consumer | pinned at | verdict |
| --- | --- | --- |
| **odu** | `3ff256f0` (ancestor of this branch — clean fast-forward) | **`none`** — only *produces* an `AgentDerivation` and hands the resolver to `sshConnector`; never consumes the failure union, never calls `resolveSystem`. Gains the better dial behavior transparently. |
| **drishti** | `15193a03` (sibling branch — re-pin to merged master) | **`adoption-opportunity`** — no code change needed. It calls `resolveSystem` uncaught at `archMap.ts:30`, so the terminal verdict propagates through its `resolveDrvPath` plug for free. Optional polish: `disconnectedMessage` (`connectionColors.ts:48`) only special-cases `"network"`, so it shows generic copy rather than the actionable remedy the new messages carry. |

All three new arms reuse the **existing** `failureCause: "remote"`, so the cause vocabulary downstream consumers switch on is unchanged.

### Considered and not changed

`@kolu/port-forward` spawns its own ssh with a hand-rolled copy of the `BatchMode` policy, so it looked like a second instance of this bug. It isn't: it is a one-shot open that settles into a bounded `failed` carrying ssh's own words, never a reconnect campaign, and it deliberately sets `ControlPath=none` so it can't ride a shared master. There is no eternal-retry defect there to fix.

### Design philosophy

- **Fail fast, no fallbacks** — this *removes* a graceful-degradation path. The old behavior caught a hard error and collapsed it into a "still working on it" state that would never resolve; the new one surfaces it.
- **Volatility boundary** — the "this host can't serve us" fact belongs to the transport package that owns ssh and provisioning, not to kolu's padi binder, so the dependency arrow points out and every consumer benefits.
- **Reuse the source of truth** — `sshRefusalOf` sits beside the existing `looksLikeNetworkError` as its sibling classifier over the same stderr stream, and the new failure arms join the existing closed `ResolveDrvFailure` / `PadiEntryFailureSchema` unions. Because both unions are exhaustive by construction, the compiler — not review — forced every consumer and the host-down copy map to handle the new causes.

### Try it locally

```sh
nix run github:juspay/kolu/ssh-password
```


srid:
### Note on the third commit

`fix(padi): point the dial test at the moved padi entrypoint` is **unrelated to this branch's ssh work** — it is a pre-existing red on master that blocks this PR's green, so it is fixed here rather than worked around.

[#2000](https://github.com/juspay/kolu/pull/2000) moved `packages/padi/src/bin.ts` under `daemonBoot/` and updated the `padi` bin in `package.json`, but `dial.test.ts:49` still spawned the old path. Only the gated `ci::daemon` lane runs that suite, so the dangling reference shipped: the padi child exited `ERR_MODULE_NOT_FOUND`, and the eight subprocess-spawning cases each waited 15s before reporting the downstream symptom — `padi socket never came up`.

Worth flagging because that symptom is [already in the flaky-test tracker](https://kolu.dev/atlas/flaky-test-tracker.html) as a darwin timing flake from #1730. This instance was **not** a flake: same message, deterministic cause, and it reproduced locally every time. Fixed, all 13 cases green.

srid:
### CI e2e metrics — `6e12b2f#1`

| `x86_64-linux` | localhost | 6 | 16 | 8 | 6m42s |
| `aarch64-darwin` | rasam | 6 | 24 | 6 | 5m48s |

Full run settled green on both platforms — 42/42 required contexts passing, no open code-scanning alerts.

PR #2006, opened 2026-07-26
A missing executable is terminal on both probe arms, not just the ssh one

srid:
Follow-up to #2003, from its code-police review. That PR made "this host has no runnable Nix" a terminal failure — but keyed it **solely on exit 127**, which the localhost arm can never produce. The branch was unreachable, and the case it named **still retried forever**, narrating "host unreachable".

The two arms report a missing executable in two different shapes, because they run the probe two different ways:

| arm | how the probe runs | a missing executable looks like |
| --- | --- | --- |
| ssh | spawns `ssh` (which exists); the **remote shell** can't find `nix-instantiate` | remote exit **127** |
| localhost | spawns `nix-instantiate` **directly**, no shell | no exit code at all — Node raises **ENOENT** as a spawn `error` |

Verified rather than assumed: a direct spawn of a missing binary fires `error` with `ENOENT` and reports close code `-2`, never 127.

### The fix

Classify both shapes. **ENOENT is carried structurally** — a new optional `code` on the `spawn-error` result, read off Node's `ErrnoException` — rather than scraped from the message prose. That's the same reason the remote arm keys on 127 instead of matching three shells' differing wording (`command not found` / `not found` / `Unknown command`).

**Which executable was absent depends on the arm**, so the verdicts differ:

- **Direct arm** — the absent binary *is* Nix. Terminal `nix-unavailable`, exactly as the remote 127 case.
- **ssh arm** — the absent binary is `ssh` on *this* machine, which says nothing about the far end. Bounded `unavailable`: it stops retrying without asserting a verdict about a host that was never reached. This mirrors how the sibling resolver (`agentDrv`'s `evaluationError`) already treats a spawn fault.

> **Only ENOENT counts as absent.** `EMFILE`/`EAGAIN` are local resource exhaustion that a later dial may well clear, so terminalising every `spawn-error` would strand a recoverable host. The errno — not merely `kind: "spawn-error"` — is what decides.

### Also from the review

- **The exit-255 comment claimed a guarantee ssh doesn't make.** A remote command may legitimately exit 255 too. Tightened to say the *text* match carries the classification and the code corroborates it — the dual gate is unchanged, only the comment was overstating it.
- **Four structurally-identical resolver match arms** differing in one output literal each — and the fact that two of them *rename* the resolver's `kind` while two pass it through was buried across near-identical blocks. Collapsed into one `P.union` arm keyed through a lookup table, so the asymmetry is visible as data instead of diff-by-eye.
- **A regex hoist** in `sshRefusalOf` (it runs per stderr line, per probe, per dial) and an `isLocalHost` de-dup.

### Reviewed and deliberately not changed

- **`https://nixos.asia/en/install` appears in two files.** Flagged as DRY, but the rule is three, and the two sites are in different packages — a framework error message and kolu's card copy. Sharing a constant across that boundary costs more than it saves.
- **The failure branch is a guarded-`if` cascade rather than `match().exhaustive()`.** It dispatches on a *latched side-channel* (`refusal`, accumulated from the progress stream) combined with the result, not on one discriminant, so `match` would need `P.when` guards that read worse than the early-throws.
- **A broader duplication the review surfaced but is out of scope here**: the "watch stderr, latch a classifier match, forward the line" idiom now appears four times (`agentDrv.ts`, `nixCopy.ts` ×2, and here). Worth a shared helper in `process.ts` eventually, but three of the four sites are untouched by this PR, and exporting a new shared helper from `@kolu/surface-remote` is a public-API change that would drag the drishti pair-PR gate along for a refactor.

### Downstream

`ExitResult`'s `spawn-error` arm gains an **optional** field, so every existing consumer destructuring `message` is unaffected — additive, not a break. Both `code` and the new resolution paths stay inside `resolveSystem`, which drishti calls uncaught and odu does not call at all, so the verdicts from #2003 stand: **odu `none`, drishti `adoption-opportunity`**.

### Try it locally

```sh
nix run github:juspay/kolu/ssh-probe-enoent
```

_Generated by Claude Code (model `claude-opus-5`)._


srid:
### CI status — two reds, both environmental, neither from this diff

Full green is **not** yet achieved on this branch, and I want to be precise about why rather than assert a flake.

**1. `ci::e2e@x86_64-linux` — `code-tab.feature:2022` tree hydration.**
Waiting on `notes.md` in the Pierre browse tree; both cucumber attempts timed out at 60s. This diff touches only `surface-remote`'s ssh-probe classification — no Code-tab, file-tree, or watcher code — and e2e exercises a *local* padi, which never reaches the ssh probe at all. The [flaky tracker](https://kolu.dev/atlas/flaky-test-tracker.html) already documents this exact scenario and locator as load-correlated and change-independent (it passes in isolation in 6s).

What is new is that the load is now **measured**: `uptime` right after the run read **45.91 / 37.89 / 39.19** on a 16-core box — ~3× oversubscribed — and the e2e recipe took **665s against 402s** on the immediately-preceding green run of the parent branch. The tracker row is updated with those figures in this PR.

**2. `ci::website-nix@x86_64-linux` — `docs/atlas/dist` seen as untracked.**
`error: Path 'docs/atlas/dist' in the repository … is not tracked by Git`, with the snapshot reported dirty. Three facts place this outside the diff:

- It **passes in isolation** — `just ci::website-nix` alone is green here.
- It failed identically on `e646278`, i.e. **before** this branch's only `docs/atlas` commit existed.
- `ci::atlas-sync` **passed** in the same run, logging `✓ docs/atlas/dist is idempotent (host-independent)` — so the committed dist is correct and in sync.

That points at a concurrency artifact around the atlas dist rebuild during a loaded lane, not at the tree's contents.

**Root cause behind both: no CI box.** Every `kolu-ci-*` pool host is currently `down` (`nix build` exit 1), which forced the linux lane onto the workstation while several long-lived agent sessions were running. `odu hosts` also warns the `x86_64-linux` pool `mixes localhost with remote hosts — any run that leases x86_64-linux will refuse it`.

The parent PR (#2003) went **fully green on both platforms** — 42/42 required contexts at `6e12b2f` — and this diff's own unit coverage is green locally: 1507 tests across `surface-remote`, `kolu-server`, and `kolu-client`, plus typecheck and lint. Re-running once a pool box is healthy is the honest way to close the gate.

srid:
### CI green — both platforms, `5fe9e4a`

| `x86_64-linux` | localhost | 6 | 16 | 8 | 5m27s |
| `aarch64-darwin` | rasam | 6 | 24 | 6 | 4m46s |

**42/42 required contexts pass** (verified against `just ci::protect --dry-run` with both `--platform` flags), and no open code-scanning alerts. Linux settled 22/22 and darwin 22/22.

#### Why this took several runs — all environmental, none from the diff

Every red along the way reproduced only under contention on the shared workstation, and each cleared on a quiet box with no code change:

- **`ci::e2e` / `ci::unit` timeouts** — the Code-tab tree-hydration wait and a 5s `surface-daemon-supervisor` endpoint test. Both are in packages this diff doesn't touch; the supervisor suite passes 38/38 locally, and both went green once the lanes ran sequentially.
- **Darwin lane dying mid-run** (`stdio transport closed`, `venue lease lost on rasam`) — I ruled out the obvious suspects with evidence rather than guessing: no `EMFILE` anywhere (so the low `ulimit -n 256` on rasam was *not* the cause, and I did not touch the box's config on that hunch), no competing `odu-runner`, 26 MB streams cleanly over its ssh stdio, and no crash report — the runner exits cleanly. It correlated with **local** load: at load 45.9 the lane died repeatedly; at load 0.6 it went 22/22 first try. The coordinator and ssh client run locally, so a saturated workstation starves them.
- **Lane coupling** — losing the darwin lease errored the *linux* nodes too, which is why one blip reds an otherwise-healthy platform. Running the lanes as separate runs (a second worktree at the same SHA) isolates them.
- **Stale `pending` statuses** — several posts were lost to `gh: Server Error (HTTP 502)` / 30s timeouts, leaving green nodes showing pending until re-posted.

Root cause behind all of it: **every `kolu-ci-*` pool box is `down`**, forcing both lanes onto the workstation. `odu hosts` also warns the `x86_64-linux` pool `mixes localhost with remote hosts — any run that leases x86_64-linux will refuse it`. Worth fixing independently of this PR.

Issue #1962, opened 2026-07-23
Remote-host connect: no copy progress and a frozen timer make provisioning look hung


srid:
While kolu (running on `pureintent`) provisioned a remote host (`zest`), the canvas showed **"Building on remote:zest… this can take a few minutes"** with a log box frozen on:

```
copying path '/nix/store/filjyw…-kolu-1.1.0' from 'https://cache.nixos.…'
```

for over two minutes. The user read it as stuck and started debugging. It wasn't stuck: the closure is **260 MiB unpacked**, and the copy simply takes minutes. Once it landed, the two derivations (`kaval.drv`, `padi.drv`) built in <1s and the connection went `building → connecting → connected` within 2 seconds.

Two UI gaps make a healthy copy indistinguishable from a hang:

- **No copy progress.** The box shows only the last raw nix log line. Nix knows the byte/path progress; surface it (`X of 260 MiB`, or at least a per-path counter) instead of a static "copying path…" line.
- **The elapsed timer freezes when the log goes quiet.** The "2m 6s" duration only advances when a new log line arrives — precisely during the long silent copy, the one moment the user needs evidence of liveness, both the log *and* the timer stand still. The timer should tick on wall-clock, independent of log activity.

_Context: this cost is paid on every reconnect after the local kolu is rebuilt — same version, new store hash → full closure re-copy — so the "few minutes with no feedback" state is recurring, not a one-time first-connect experience._

PR #2007, opened 2026-07-26
The host-down card stops swallowing the build error

srid:
A remote host failed to provision and kolu said this, and only this:

> **Can't reach this host**
> The connection to this host gave up — it may be unreachable, or its provisioning failed partway. Check the host is up and reachable over ssh, or switch back to your local host.
> `zest: 'nix build' exited with code 1`

The host was up. ssh was fine. The real cause was a `tsc` error on the branch being provisioned — two files in one directory whose names differed only in case, which resolves fine on Linux and breaks on a case-insensitive macOS Nix Store volume. Getting from that card to that sentence took rebuilding the derivation by hand with `--keep-failed` and re-running pnpm with a different reporter, because **two independent layers were each eating the diagnostic**.

## The client had the output and dropped it

Every stderr line from the provisioning `nix build` already flows through `onProgress` into the session's bounded 20-line tail, and `setDown` carries that tail *forward* into the `failed` arm on purpose:

```ts
// the `log` tail carries forward (the failed episode's lines stay readable until the next dial resets them)
stateCell.set({ phase, error, cause, log: prev.log, … })
```

So the lines were collected, retained, and shipped to the browser. `ConnectCanvas` renders them the entire time provisioning is *working* — and the failure card took only `{cause, reason}`, so the tail was dropped unread at exactly the moment it became the only thing worth reading.

The failed episode is now **one value with one home**: a single `failedEpisode()` reader that every failure surface shares. Three surfaces show the tail — the host-down card, the boot-stalled card, and the host tab's diagnostics popover — and the `log` prop is *required*, because an optional prop is precisely how one of two callers silently dropped the evidence.

One distinction is load-bearing and is preserved end to end: `undefined` means **we cannot see the output** (the map's liveness floor drops the `connection` word over a dead link while keeping `failure`), and `[]` means **the failure genuinely produced none**. Both render nothing, but only the caller can tell them apart, so no reader upstream may collapse one into the other. Collapsing them was a real bug the review caught — the fix was itself telling a small version of the lie it exists to stop.

For the failure above, the card would now show:

```
 ERR_PNPM_RECURSIVE_RUN_FIRST_FAIL  kolu-client@0.1.0 typecheck: `tsc --noEmit`
For full logs, run:
  nix log /nix/store/25f3nqdl8a3j1ksj0jllwrgl81z5pdfs-kolu-typecheck-2.0.0.drv
error: Cannot build '/nix/store/i8wlxn7a1yflwanxvcgbaksiaqkwyxa3-kolu-2.0.0.drv'.
       Reason: 1 dependency failed.
```

## Nix was erasing it before it reached the log

pnpm's default reporter draws a live progress tree and redraws it with cursor-up escapes. Inside a Nix builder that redraw overwrites the failing package's own output, so `nix log` ends at a bare `ERR_PNPM_RECURSIVE_RUN_FIRST_FAIL` with the actual error gone.

Reproduced by planting a deliberate type error and building `checks.x86_64-linux.typecheck`:

| | tail of `nix log` |
| --- | --- |
| **before** | `> │ `<br>`> └─ Running...`<br>`>  ERR_PNPM_RECURSIVE_RUN_FIRST_FAIL  @kolu/terminal-vocab@0.1.0 typecheck` |
| **after** | `> packages/terminal-vocab typecheck: src/ports.ts(244,14): error TS2322: Type 'string' is not assignable to type 'number'.` |

The fix **wraps the tool, not the derivations**. `nix/pnpm.nix` is a `symlinkJoin` over `pkgs.pnpm_10` with `wrapProgram --set-default npm_config_reporter append-only`, exposed through the repo's existing overlay as `pkgs.pnpm-build`. It has to be the env var rather than a `--reporter` flag: the root scripts shell out to a nested `pnpm -r` an outer flag never reaches. `--set-default` rather than `--set`, so a caller that deliberately asks for another reporter still wins. `shell.nix` keeps the **unwrapped** pnpm, so a developer at a TTY still gets the progress tree.

This started as an env attrset merged at each build site, and that roster was wrong twice — first three sites when there were five, then five when `workspace-tree.nix` and both `fetchPnpmDeps` derivations also run pnpm in a builder. Enumerating call sites was the bug; every site gets pnpm from the same place, and `pnpmConfigHook` ships none of its own, so a wrapped binary is the only pnpm on the builder PATH and cannot be forgotten. Anchoring on the *hook* was tried and rejected on evidence: `makeSetupHook` runs through `runCommand`'s `buildCommand` and silently drops `postHook`/`postInstall`, and the only working form rewrites `buildCommand` text — whose failure mode is silent log erasure, which is the exact defect being fixed.

## Review

Full `/be` gauntlet — see the three comments below for detail. It changed the shape of this PR substantially, and caught one crash:

- **lowy ∥ hickey** — 23 findings. Both lenses independently found the change split one value across two paths and that the Nix roster was incomplete. Three contradictions adjudicated by hand.
- **grok ⇄ claude debate** — consensus in 2 rounds, 4 findings, none disputed. Caught that the changelog entry had been treated as the documentation obligation while `docs/remote-hosts.mdx` — the canonical page for this card — still said nothing about the output; and that every resolver fixture left `connectLog` undefined, so the pins passed whether or not the tail was threaded through.
- **code-police** — found that `hostFailure` had become an unconditional `createMemo`, which Solid runs *eagerly*, so its unrepresentable-state guard threw on essentially every normal boot and blanked the client. Two individually-correct review fixes composed into a crash; no test caught it because none mount `App.tsx`.

## Testing

- 112 test files / 1008 tests pass; typecheck, `just lint`, `just fmt` clean; all flakes evaluate.
- The reporter fix is verified empirically, and re-verified after the rewrite from per-site merges to the wrapped binary.
- New pins: the host-down card's tail rendering, the boot-stalled connector-vs-client arms, the resolver's log thread-through (asserted by reference, so a copy at the seam fails too), and the unrepresentable-state guard.
- No e2e scenario touches these surfaces, so there was none to re-run.

**ODU-IMPACT: none** — no `@kolu/surface*` public API is touched, so no drishti pair and nothing for the odu ledger.


srid:

**Status: `applied`** (reconcile reported `needs-human`; all valve-2 items adjudicated below).
Base `30d46915d` · **lowy 9 findings · hickey 14 findings** · 20 settled by the reconcile pass, 3 broken by hand.

| Commit | Area |
|---|---|
| 44d117ee7 | client — the failed episode is one value, read once, on every surface |
| 5b16e7a35 | nix — every pnpm-in-Nix builder plugs into one build env |
| 259f257d7 | minor polish (all areas) |
| f66a3e35f | name the one thing the three log tails share *(valve-2 adjudication)* |

### What both lenses reached independently

**The change split one value across two paths.** `cause`, `reason` and `log` are three fields of one `EntryStatus`, and `useCanvasMode` already destructured `state.connection?.log` one line away — yet the diff routed the third field around the `host-failed` `CanvasMode` arm whose own docstring says *"the renderer reads no accessor a second time."* The author's churn rationale (the #1962 remount class) was checked by both lenses and found no longer to follow, since every `<Match>` now keys on a stable boolean `kind`. Fixed: one reader, read once at the arm.

That split had a reachable consequence, not just an aesthetic one — `floorOnLiveness` drops `connection` while keeping `failure`, so *reason-without-log* is a live state, and the original `?? []` rendered "we cannot see the output" identically to "there was none."

**The Nix fix was cut too narrow.** The roster named three derivations; there were five. `packages/solid-browser/example/docsite` and `packages/surface/example/remote-process-monitor` both run `pnpm … build:client` in a builder with their own CI lanes. *(A later `/simplify` pass took this further and replaced the whole enumerate-the-sites approach — see that comment.)*

Also landed: `BootStalledCanvas` and `HostDiagnosticsPopover` now show the tail (both lenses raised both surfaces); the `log` prop is required, because an optional prop is *how* one of two callers silently dropped the evidence.

### Declined on evidence, not argument

hickey wanted the reporter anchored on `pkgs.pnpmConfigHook` so omitting it would be unspellable. The reconcile pass tried it: `makeSetupHook` runs through `runCommand`'s `buildCommand`, so `postHook`/`postInstall` are **silently dropped**, and the only working form rewrites `buildCommand` text — whose failure mode is silent log erasure, i.e. precisely the defect this PR fixes. Rejected.

### Valve 2 — adjudicated by the author

**1. Extract a `<LogTail>` component?** hickey F4 wanted the shared chrome named once with a `wrap` knob; lowy F8 said the split is correct under both lenses and a knob would parameterize two independently-changing behaviours. **Both were right about different halves** — so only the genuinely shared part got a name (`LOG_TAIL_SURFACE`, later `LOG_TAIL_LINE`), following `ui/disclosureChrome.ts`, which draws exactly this line for the `<summary>` reset. Sizing, padding and line-clamping stay local literals. No knob.

**2. Should the tail carry `source` (local/remote) provenance?** hickey F6 said it throws away the question an operator asks first; lowy B4 marked the same type *"Correct. No finding."* **Dropped.** `ConnectCanvas`'s tail has always ignored `source`, so adding provenance to one of three tails invents an inconsistency, and nothing in the originating incident turned on local-vs-remote. If it is worth showing it is worth showing in all three — a separate change.

**3. Should `agentDrv` stop joining diagnostics into `reason`?** hickey F9's premise is **verified true**: those lines already reach the session tail, so an eval failure renders them twice on one card. **Kept anyway.** `floorOnLiveness` drops `log` while keeping `reason`, so over a dead link the joined copy is the *only* surviving diagnostic. Removing it would trade a cosmetic duplication for total information loss in exactly the degraded case this PR exists to serve.

---

*Reviewed by two independent lenses, then reconciled and applied in one pass. Findings and per-finding outcomes live in the run's `.lens-debate/` record.*

srid:
## Grok ⇄ Claude debate

✅ Consensus in 2 round(s) — base `30d46915d`

|---|---|---|
| 1 | 60ffae68f | fix(debate): docs, wrapper passthru, and a pin that could not fail |

- **F1** — The unreleased changelog Change for this PR (host-down card keeps the build output) was written, but the canonical product page that describes that card was not updated
- **F2** — The comment claims the wrapper is a drop-in for `pnpm_10`, but passthru is incomplete
- **F3** — The boot-stalled connector arm now carries `log` from `connectLog` through the accrue tag into `escapeSurface`, but every resolver fixture sets `connectLog: undefined` and the D2 expectations omit `log` (vitest treats missing key ≡ `undefined`, so they pass either way)
- **F4** — The justification that the `host-failed` arm has "~6 consumers" each re-folding the map entry is overstated

srid:

Ran with `--no-elegance` (the `/simplify` track already swept this tree). Scoped to `30d46915d...HEAD`.

| Pass | Findings | Outcome |
|---|---|---|
| 1 — rules | 1 major | fixed |
| 2 — fact-check | 0 | clean on all three weighted rules |
| 3 — elegance | — | skipped (`--no-elegance`) |

### The major: a crash the earlier review passes composed

`app-shell-stays-thin` — commit 25c8ca55a.

`App.tsx`'s `hostFailure` was an **unconditional `createMemo` in the component body**, and Solid runs a memo's callback *eagerly at creation*, not on first read. Since the active host is essentially never `failed` at boot, the memo's own `episode === undefined` guard threw uncaught on nearly every normal boot — **before** the `host-failed` `<Match>` arm that was supposed to gate it ever ran — blanking the client.

Neither contributing change was wrong on its own:

1. the lens pass replaced two lazy accessors with one `hostFailure()` reader plus a throw guard for the unrepresentable state;
2. the simplify pass made that reader a `createMemo` on efficiency grounds.

Lazy + guard is safe. Eager + guard is a crash. No test caught it because none mount `App.tsx` — doing so drags the whole live socket stack, as `HostDownCanvas.test.tsx`'s own header explains.

The fix also removes the design smell underneath: the memo re-threaded a domain read into `<HostDownCanvas failure={…}>` as a prop, which `app-shell-stays-thin` / `no-preference-prop-drilling` forbid — and `HostDiagnosticsPopover`, in this same diff, already proved the alternative works. `HostDownCanvas` now takes **no props** and reads its own episode internally; its `createMemo` is safe there because the component only ever instantiates from the `host-failed` arm. `App.tsx` is back to its original 2-reactive-primitive budget.

Verified non-vacuously: the crash was reproduced against the real `solid-js` build, then re-reproduced by reverting the fix *and* by stripping the throw guard, confirming every new or changed test genuinely fails without the fix.

### Fact-check pass: clean

Checked hardest against the rules this PR is itself about:

- **`caught-error-must-not-collapse-to-empty`** — the `undefined` vs `[]` distinction survives every hop it is threaded through. The PR exists because a diagnostic collapsed; the fix does not commit the same sin.
- **fail-fast / no unjustified fallbacks** — no new `??` / `||` / `catch` hides a real absence.
- **`errors-must-log-at-error`** — no new error-handling code in the diff.

Final: `tsc --noEmit` clean · 112 files / 1008 tests passing · `just fmt` and `just lint` clean.

srid:
### CI e2e timing — `25c8ca5#1`

|---|---|---|---|---|---|
| `x86_64-linux` | kolu-ci-1 | 8 | 32 | 8 | 2m 34s |
| `aarch64-darwin` | rasam | 6 | 24 | 6 | 5m 02s |

Full pipeline settled green across both platforms — 44 nodes, none failed or errored.

srid:
## Evidence

Captured on a clean ephemeral pu box: a nix build of this branch (`25c8ca5`) vs. one of `master` (`30d4691`), each launched with its own temp state dir, driven in a real browser. Same scenario both times — add a remote host over the host strip's **+** affordance, pointing at an ssh target that answers but refuses kolu's non-interactive credentials, so the entry reaches `failed` with cause `auth-required`.

The one-line reason is identical in both. What changed is everything under it: the failing episode's retained output tail is now rendered, so ssh's own words (`Permission denied (publickey,password,keyboard-interactive).`, `gave up — …`) are on screen instead of collected, shipped to the browser, and dropped unread.

| Before (`master`) | After (this PR) |
| --- | --- |
|  |  |
| no `host-down-log` in the DOM at all — `document.querySelector('[data-testid="host-down-log"]')` → `null` | `host-down-log` present and non-empty beneath the reason line |

The same tail also reaches the second surface — the host tab's connection pip opens the diagnostics popover, whose `host-diagnostics-log` block carries the identical episode output:



### One note from capturing this

An unresolvable hostname does **not** reach this card. `nosuchhost.invalid` was added the same way and sat on `Connecting to remote:nosuchhost.invalid…`: `resolveSystem`'s generic throw is classified `"network"`, which retries, so the entry never leaves `warming` and no `link-failed` is minted until the 30-minute `CAMPAIGN_CEILING_MS`. The three terminal prerequisite faults from #2003 (`auth-required`, `host-key-unverified`, `nix-unavailable`) are the reachable-in-seconds routes to the host-down card, which is why the shots above use one of them rather than the `link-failed` build-error case from the PR description.

PR #2008, opened 2026-07-27
Collect the padis a removed worktree strands

srid:
Every worktree gets its own padi. `pnpm dev` anchors it at `<worktree>/.kolu-dev/padi`, and that path *is* padi's identity — the socket, the kaval and the supervisor gate are all keyed by a digest of it. padi is also detached: it outlives the `just dev` that spawned it and has no idle timeout. So `git worktree remove` strands one forever, holding a kaval, that kaval's PTY shells, and a few hundred MB of RSS, and nothing notices.

On my machine that had reached **16 padis and 6 kavals, ~3.7 GB resident** — six of them anchored at worktrees that no longer existed, the oldest running six days.

## Why `dev-clean` never caught them

It recomputed padi's digest in bash:

```sh
state_root="$(realpath -m "$dev_dir/padi-state")"     # $XDG_RUNTIME_DIR/kolu-dev-$slot/padi-state
digest="$(printf '%s' "$state_root" | sha256sum | cut -c1-16)"
```

But kolu-cli's `dev` script pins `KOLU_PADI_STATE_DIR` *on the `node` command line*, and that beats anything the justfile exports. Every real padi runs at `<worktree>/.kolu-dev/padi`; `dev-clean` hashed a path none of them ever used, so its pidfile lookup matched nothing. The `server` recipe's per-port `KOLU_PADI_STATE_DIR` and `KOLU_KAVAL_SOCKET` were dead for the same reason — nothing reads the latter from the env at all (`padiKavalSocketPath` takes an override argument no caller passes) — while their comments described an isolation that was really coming from the per-worktree state root. Both exports and the comments go.

## One rule instead of digest arithmetic

**A padi whose state-root no longer exists is garbage.** That single predicate is the whole selector, and `dev-clean` needs no "stop the padi at *this* state-root" verb of its own: it deletes its `.kolu-dev` and sweeps, because a deleted state-root *is* stale by that definition. One rule, two callers — and ~60 lines of bash pidfile/sha256 handling deleted.

```sh
just dev-clean   # reset this worktree: drop its state, then sweep
just dev-reap    # sweep alone — collect what `git worktree remove` already stranded
```

Both sit on `padi --reap-stale`, built from the discovery that already exists (`discoverPadiDaemons()` reads each daemon's `state-root` manifest and gate pid), so nothing re-derives the identity.

Reaping requires **proof**. A live state-root — production, or any worktree still present — is never a candidate, and an *unreadable* manifest is never one either: "I could not read what this is anchored to" is the opposite of proof. Order is load-bearing: the supervisor dies first or it respawns the padi it supervises, and the kaval dies last because padi does not take it down on its own way out. Runtime dirs are cleared only once every holder is proven gone; a holder surviving SIGKILL throws rather than reporting a soft failure a caller might ignore before deleting a live daemon's socket.

`SUPERVISOR_GATE_FILE` moves into padi, which owns that runtime dir's layout (beside `PADI_SOCK_FILE` / `PADI_GATE_FILE`), so the reaper and the supervisor share one definition of the name. The dependency arrow stays server → padi: `supervisorClaim.ts` imports it through the `@kolu/padi/assembly` barrel it already uses.

## Verification

Beyond unit tests, I built a real ghost end-to-end: spawned a padi against a throwaway state-root (267 MB padi + 119 MB kaval, full runtime dirs), deleted the state-root to simulate `git worktree remove`, and swept.

```
reaped supervisor (none) + padi 3103918 + kaval 3103965 @ …/ghost-worktree/.kolu-dev/padi
after: processes      → both gone
after: runtime dirs   → both removed
```

The 15 live padis on the machine were untouched by that run, which is the safety property stated above holding in practice.

Tests: five new cases in `reapStale.test.ts` — the three proof cases (gone / still-present / unreadable manifest) run in the fork-free lane; the two that stop a real gate holder are `describeDaemon`-gated, matching `stateRoot.test.ts`'s precedent.

## Not covered

Runtime dirs left by a padi that already exited cleanly (a bare `state-root` file, no socket inode) are not collected — discovery requires a socket inode to consider a registration at all. They are inert and boot-wiped, and reaching them would mean a second scanner that spells the rendezvous layout a different way. There are 15 such dirs on my machine, totalling a few KB.


srid:
## CI e2e metrics — `5724388#3`

| `x86_64-linux` | `localhost` | 6 | 16 | 8 | **6m 42s** (401866 ms) |
| `aarch64-darwin` | `nix-infra@rasam.tail12b27.ts.net` | 6 | 24 | 6 | **4m 00s** (240034 ms) |

Both lanes settled with `ci::e2e` green on `5724388`. Durations are the recipe wall time from the durable ledger `.ci/5724388/runs/3.json`, not Cucumber's internal timer; parallelism is the last `e2e: workers=` line each lane recorded.

Two nodes needed positional reruns on this SHA, both already-tracked flakes (recurrence evidence added to the [Flaky Test Tracker](https://kolu.dev/atlas/flaky-test-tracker.html) in this PR):

- `ci::website-nix@x86_64-linux` — the `atlas-sync` / `website-nix` shared-snapshot race (`Path 'docs/atlas/dist' … is not tracked by Git`). Green on a same-SHA rerun once `atlas-sync` had finished.
- `ci::e2e@x86_64-linux` on the *previous* SHA — `code-tab.feature:817` browse-tree hydration, which failed at a **different step on each of its two attempts**. It passed outright here in 6m42s.

Worth flagging for the venue pool rather than this PR: every **remote** lane driven from this workstation died repeatedly with `liveness probe timed out — remote wedged` → `venue lease lost`, aborting the whole run including healthy lanes. It hit `rasam` and `kolu-ci-1` identically, while local lanes never wedged, with the workstation at load average ~28 on 16 cores from other long-lived agent sessions — i.e. the local coordinator could not service odu's heartbeat. That is the same box-capacity cause the tracker's `pierre-file-tree` row already identifies for the e2e flake.

srid:
Closing — wrong fix. Ships `just dev-reap` / `just dev-clean`, i.e. **a command a human has to remember**. Nobody runs those. The workflow is `rm -rf` the worktree after merge, and the daemons leak exactly the same. Same leak with extra steps.

Real requirement is self-collection, zero human action → #2010.

Salvage if you want it, don't re-derive it:

- `stalePadis()` / `reapPadi()` (`packages/padi/src/reapStale.ts`) — classify + collect, on top of the existing `discoverPadiDaemons()`. Any automatic collector still needs this.
- `dev-clean` was hashing `$XDG_RUNTIME_DIR/kolu-dev-$slot/padi-state`, but kolu-cli's `dev` script pins `KOLU_PADI_STATE_DIR` on the `node` command line, which beats the justfile export. It hashed a path **no padi ever ran at** and matched nothing. That command has never worked.
- Same reason killed the `server` recipe's `KOLU_PADI_STATE_DIR` and `KOLU_KAVAL_SOCKET` — both dead; nothing reads the latter from env at all.
- `SUPERVISOR_GATE_FILE` moved to padi (owns that runtime dir's layout), arrow stays server → padi.

CI was green both platforms (42/42) on `5724388` if any of it gets picked up.

Issue #14, opened 2026-03-21
refactor: extract generic signoff recipe from just ci


srid:
## Problem

`just ci` hardcodes GitHub commit status posting (signoff) interleaved with build and test logic. The signoff bash is ~25 lines that obscure the core CI flow.

## Proposed change

Split into two recipes:

- **`just signoff <context> <cmd...>`** — generic wrapper that posts GitHub commit status (pending → success/failure) around any command
- **`just ci`** — becomes 2 lines: `nix run github:juspay/vira ci` + `just signoff signoff/e2e just test`

Usage examples:
- `just signoff signoff/e2e just test`
- `just signoff signoff/lint just pc`

See conversation context for the full implementation.

Issue #13, opened 2026-03-21
Switch between Ghostty themes


srid:
Also make sure that text selection colour is sensible. Presently it just whitens and hides the underlying text:

<img width="738" height="715" alt="Image" src="https://github.com/user-attachments/assets/51ccaebd-2367-487d-979a-88709ebe3d51" />

PR #2011, opened 2026-07-27
osfacts V2: complete the native fact tool (kolu adoption parked)

srid:
Tonight's production incidents all came from the **kolu-adoption half** of an earlier tip of this PR — not from the tool. A new padi that only understands a two-field pid+start gate refused to yield a pid for a legacy one-field gate, so Restart-kaval could not SIGTERM the old process and the upgrade path bricked a live host. **This PR ships the tool; kolu's wiring of the new facets moves to a future PR designed around mixed-version daemon windows.**

The full pre-strip tip is parked on branch [`osfacts-kolu-adoption`](https://github.com/juspay/kolu/tree/osfacts-kolu-adoption) so nothing becomes unreachable (drishti pins commits on that history). Atlas phase **OSF8** names that follow-on and makes the mixed-version kaval window its central design constraint.

### What ships here

| area | result |
| --- | --- |
| Process facets | `--mem`, `--start-time`, `--cpu-time`, `--uid`, `--cwd`, `--status`, `--argv` on both platforms; faceted `U` rows keep failures independent |
| Listener truth | claimed vs unclaimed listeners (OSF6); Darwin macOS-27 host-table blindness is `BLIND_OR_EMPTY`, not a blessed empty set |
| Host verb | load, memory/swap, uptime, per-core CPU, model/nullable MHz, interface counters, disk total/free/available (OSF7) |
| Client | V2 wire, separate `SnapshotReading` / `HostReading` per verb, typed rows; unknown tags / schema skew stay loud |
| Kolu ports path | minimal V2 compatibility only, so master's OSF2 port sensor keeps working against the V2 binary |

### What the review gauntlet changed

The tool was already battle-proven (drishti exercises it green across machines), but four serial review stages — architecture-first-principles, the lowy ∥ hickey lens pair, a Grok debate to consensus, `/simplify`, and code-police — turned up **one shipped regression and six live collapse-to-empty defects**. Every one is the same failure shape this tool exists to refuse: *a failure arriving as a plausible value.*

**The regression.** padi's new source-error gate threw "blind" on **any** `E` row — and Darwin on macOS 27 *always* emits `darwin_tcp_pcblist=BLIND_OR_EMPTY` while the same-uid fd walk still supplies every claimed listener. Port detection would have gone permanently dark on that platform with the facts sitting in hand, contradicting osfacts' own README: *"source blindness is an `E` row, not an instruction to discard facts that did arrive."*

The consumer could not tell the two apart because the wire never said. So an `E` row now carries **the facet its silence costs**:

```
E <source> <facet> <code>
```

`ports` means no listener survived. `ports_unclaimed` means only listeners nobody claimed are gone — what macOS 27 gates, and a fact a subtree fold never reads. padi then scopes source blindness exactly the way this PR already scoped per-pid blindness.

**The others, each a caught error reaching a consumer as a legitimate reading:**

- The TS client discarded the binary's `E`-only failure document on a non-zero exit — the one place that says *which source went blind* never reached any consumer, which saw only `"non-zero exit"`.
- `/proc` readdir failure became a successful, empty, entirely plausible "this host has no processes". So did a per-entry failure mid-listing.
- A failed uptime read published a fabricated `0`, indistinguishable from a just-booted host.
- Darwin `--status` dropped the task-port errno into `threads: null` — "denied" rendered as "the kernel said none".
- A blind `NET_RT_IFLIST2` sysctl became an empty interface list with no `E` row at all, under a source label (`getifaddrs`) that nothing called.
- An **empty** TSV field parsed as `0`, because `Number("")` is `0` — an empty `nice` read as "default priority".

**Structure.** The facet is the noun this PR introduces, and it had no home: a bare `String` at ~49 Rust emit sites, two hand-kept arrays in TypeScript, and a third list in padi. It is now a `Facet` enum in `schema.rs` carried across the language boundary by a checked-in `facets.json` that **both sides pin by test**, so a half-edit fails the fast unit lane instead of surfacing as a consumer parse error at runtime.

### One finding deliberately not applied as proposed

A reviewer wanted the `sysinfo` dependency replaced with a hand-written sysctl walker. The Atlas plan-of-record shows that crate was adopted precisely to **delete** an unsafe hand-rolled `if_data` overlay CodeQL had flagged — so its real defect (a blind sysctl laundering into an empty list) is fixed while the crate stays.

### The perf smoke was measuring the wrong denominator

`ci::osfacts-live` had failed on **every CI host it ever ran on**, passing only the ~450-process workstation it was calibrated against. Measured rather than assumed: the extra-facet cost is ~23 ms on both a 411-process and a 50-process host, because `--ports` walks **descriptors**, not processes — 16.3 of 21.2 ms here, 6.0 µs per descriptor. The CI box proved it exactly: **50 processes, 7,646 descriptors**. The budget now has the two terms the cost actually has.

### Try it locally

```sh
nix build github:juspay/kolu/osfacts-improve#osfacts
nix run github:juspay/kolu/osfacts-improve#osfacts -- snapshot --procs --ports --json
```


srid:
CI e2e evidence for `5a595c7#1`:

|---|---|---:|---:|---:|---:|
| `x86_64-linux` | `kolu-ci-4` | 8 | 32 | 8 | 158.726s |
| `aarch64-darwin` | `nix-infra@rasam.tail12b27.ts.net` | 6 | 24 | 6 | 264.052s |

srid:

**Status: `applied`** — all 28 findings are settled. 24 were fixed and committed
by the reconcile pass; the 4 that reached valve 2 were adjudicated by the run
owner and are recorded below with the reasoning, so nothing is left hanging.

Two structural reviewers read the diff independently and in parallel — **lowy**
(volatility-based decomposition) and **hickey** (structural simplicity). lowy
returned 14 findings, hickey returned 14, and they landed on the same headline
without seeing each other's work: **the facet is the noun this PR introduces and
it had no home.** Six pairs matched (`≡`).

The facet vocabulary is now a `Facet` enum in `schema.rs` — no reader spells a
facet as a string literal — carried to the TypeScript client by a checked-in
`facets.json` that both languages pin by test, in the fast unit lane. Around it,
five places where the tool dropped a fact instead of reporting it.

### Commits

| sha | area |
| --- | --- |
| `8fe84397` | the osfacts binary — one home for the facet vocabulary, and a source that says what its silence costs |
| `ed69a2b4` | osfacts-client — one reading type per verb, one derived facet vocabulary |
| `16ff729f` | padi's port scan — the ask is the gate |
| `a7f785e3` | minor polish from the lens review |

### Findings

| id | origin | title | outcome | commit |
| F1 · #1 | both | facet vocabulary has no single source of truth | ≡ agreed → applied | `8fe84397` `ed69a2b4` |
| F2 · #12 | both | `attribute_host_listeners` switches truth sources on a mode flag | ≡ agreed → applied | `8fe84397` |
| F3 · #6 | both | `OsfactsReading` is one type for two verbs | ≡ agreed → applied | `ed69a2b4` |
| F4 · #7 | both | facet unions flat across verbs; `mem` means two things | ≡ agreed → applied (partial) | `ed69a2b4` |
| F5 · #9 | both | padi hand-maintains a third facet list, untyped | ≡ agreed → applied | `16ff729f` `ed69a2b4` |
| F7 · #4 | both | a failed host-global constant encoded two ways | ≡ agreed → applied (partial) | `8fe84397` |
| F6 | lowy | `Port.uid` is a whole-platform silent absence with no honesty row | applied | `8fe84397` |
| F8 | lowy | `processIdentity` puts consumer policy in the policy-free client | applied (deleted) | `ed69a2b4` |
| F9 | lowy | `host` vs `snapshotHost` naming collision | **human** | — |
| F10 | lowy | adding one facet touches ~15 sites, three silently | applied | `a7f785e3` `8fe84397` `ed69a2b4` |
| F11 | lowy | a malformed `pcblist` record truncates the listener set with no error | **experiment** → applied, fix corrected | `8fe84397` |
| F12 | lowy | padi re-validates an invariant the client already enforces | applied | `16ff729f` |
| F13 · #11 | both | `ci/mod.just` encodes scheduling as dependency edges | **human** | — |
| F14 | lowy | one `SCHEMA_VERSION` gates independently-evolving verbs | **human** | — |
| #2 | hickey | `Snapshot`'s field list is enumerated seven times, unchecked | applied (partial) | `8fe84397` `ed69a2b4` |
| #3 | hickey | the sensors duplicate shared plumbing; the sort blocks had drifted | applied | `8fe84397` |
| #5 | hickey | darwin's `kern_proc_all` names one facet, its silence costs four | applied | `8fe84397` |
| #8 | hickey | the TSV grammar is re-encoded in at least six decoders | **human** | — |
| #10 | hickey | "what was asked for" is absent from the reading | observation | — |
| #13 | hickey | `Snapshot`/`HostSnapshot` duplicate their emit/exit plumbing | applied (partial) | `8fe84397` |
| #14 | hickey | one facet has four spellings across the pipeline | observation (`drop`) | — |

### The finding whose fix was wrong

**lowy F11** spotted a real defect: the `pcblist_n` record walker `break`s on a
length it cannot frame and returns the rows read so far as a healthy table — so
Apple layout drift, the single volatility this module has already watched move,
degrades into a silently short listener set. The suggested fix was to replace the
`break` with `return Err(EINVAL)`.

A discriminating experiment over both committed fixtures said no:

```
adhoc.bin    (48 bytes)     header_len=24  BREAK at off=24,    len=24, records=0
platform.bin (54872 bytes)  header_len=24  BREAK at off=54848, len=24, records=534 → 29 rows
```

Both captures reach their end *through* that `break` — a bare 24-byte record is
Apple's closing marker, not a malformed one. lowy's literal fix would have made
every healthy read fail. The applied rule discriminates: `len < 24` or a record
running past the buffer → loud `Err`; `len == 24` → terminator. Re-run: platform
→ `Ok(29)`, adhoc → `Ok([])`, any mid-record truncation → `Err`. All three pinned
by test.

### Verifying the darwin fixes without a darwin box

`darwin.rs` is `cfg(target_os = "macos")` and cannot be compiled here — nixpkgs
ships no `aarch64-apple-darwin` std, and the pinned rustc cannot cross-check it.
Rather than land the two riskiest fixes (F2's union, F11's loud framing)
unverified, the **pure** half of the listener path — the record walk and the
host/fd-walk merge, neither of which touches an OS — moved to
`osfacts::pcblist`. Both are now compiled and test-run on every platform. That
is also what hickey #3 argues for structurally: row order and record framing are
properties of the schema, not of an OS.

Found on the way: `cli_contract.rs` compared `val["errors"][0]` against a two-key
JSON object, while `b8514697` had added a third key. Dead on linux, **failing on
the darwin lane**. Now matched by content, not index.

---

## Valve 2 — adjudicated

These four reached valve 2 because they are judgment-shaped rather than
defects. Each is recorded with both lenses' positions and the call that was
made; three were dropped with reasons, one produced a small fix.

| # | finding | call |
| --- | --- | --- |
| 1 | `ci/mod.just` scheduling-as-dependency-edges | **drop** — verified, but pre-existing and not this PR's to change |
| 2 | one `SCHEMA_VERSION` for two verbs | **drop** — one binary, one wire revision |
| 3 | `host` vs `snapshotHost` naming | **drop the rename, document the system** |
| 4 | executable TSV-grammar oracle | **drop** — net-new apparatus; the cited drift is closed |

### The calls, in full


### 1. `ci/mod.just` encodes scheduling as dependency edges (lowy F13 ≡ hickey #11)

Not applied deliberately: editing the CI DAG immediately before this branch runs
CI is a blast radius that belongs to you. Both lenses raised it independently and
agree on the diagnosis; they differ on the remedy.

> **lowy F13** — A dependency edge in a build DAG declares *what this recipe
> consumes*. Only `osfacts` does. `atlas-sync`, `pnpm-hash-fresh`,
> `agent-flake-nix`, `website-nix`, `home-manager`, `unit`, `daemon` and `e2e`
> produce nothing `osfacts-live` reads; they are there, per the comment at
> `:252-255`, to make the lane run *last* so its CPU budget is not contended. The
> timing goal is legitimate and measured. The mechanism puts it in the wrong
> receptacle: the recipe list is itself volatile (recipes get added, renamed,
> merged), and any such change silently drops or invalidates an edge, reverting
> the timing property with **no failure signal** — the lane just gets flaky again.
> It also lets an unrelated red recipe block a green osfacts lane.
> **Suggestion:** Express ordering with an ordering primitive, not with
> dependencies. If odu exposes a lane/priority/run-last marker, use it and reduce
> the edge list to the true dependency (`osfacts`). If it does not, the cheaper
> honest fix is to make the budget itself contention-tolerant — the comment at
> `ci/mod.just:253-255` already claims the child-CPU budget "ignores scheduler
> contention", which if true makes these edges unnecessary and if false makes the
> comment wrong; resolve which, and say so in the comment.

> **hickey #11** — Two independent concerns are braided into one construct.
> `depends-on` means *this node needs that node's output* (correctness);
> `osfacts-live` is using it to mean *start me late* (performance, so its Linux
> timing smoke is not measuring compiler contention). The comment at `:252-255`
> says so outright. Because the mechanism has no way to say "last", the intent is
> approximated by transcribing part of `default`'s recipe list — and the
> transcription is already wrong: `smoke`, `fmt`, `biome`,
> `surface-examples-nix`, `solid-browser-example-nix`, `odu-nix`,
> `website-pnpm-hash-fresh`, `surface-example-build`, and
> `surface-app-example-build` are all in the fan-out and all absent from the list.
> Four of those are heavy Nix builds. So the stated goal is materially unmet on
> the day it lands, and every future recipe added to `default` widens the gap
> silently.
> **Suggestion:** Stop restating the list. Introduce one aggregate node whose only
> job is to be the drain point and give it the whole fan-out — `fanout-drained:
> <every recipe>`, with `osfacts-live: fanout-drained`. One list, named for what
> it means, and adding a recipe to `default` without adding it here becomes a
> visible, single-place omission. If odu later grows a scheduling-priority node
> kind, this node is the single place to replace.

**Worth confirming either way:** hickey's claim that the list is *already* an
incomplete copy — nine recipes missing, four of them heavy Nix builds — means the
timing property may not be getting delivered today.

### 2. One `SCHEMA_VERSION` for two independently-evolving verbs (lowy F14)

hickey did not raise this. lowy dispositioned it `drop`; the suggested action is
one Atlas sentence, not code.

> **lowy F14** — `snapshot`, `host`, and OSF4's future `socket-holders` share one
> version number, and the client hard-refuses any mismatch by design. So a change
> to the host-telemetry shape forces every `snapshot` consumer into a fatal
> version error even though its own contract did not move. This PR exists because
> a mixed-version window bricked a live host, which makes the coupling worth
> naming out loud. Today it is genuinely contained: one binary, one client, a
> Nix-baked `KOLU_OSFACTS_BIN` path, lockstep deployment — so the version is the
> *binary's* version and lockstep is the intended model. It stops being contained
> at OSF5, when `osfacts/` moves to its own repo and drishti pins it on a cadence
> kolu does not control.
> **Suggestion:** No code change in this PR. Record the constraint where the
> extraction will read it: add a sentence to the OSF5 row in
> `os-facts-tool.mdx:235` stating that `SCHEMA_VERSION` covers the whole binary
> (not per verb), so independent external consumers must be redeployed with it —
> and that per-verb revisions are the alternative OSF5 must consciously accept or
> reject rather than inherit.

The verb split landed in this pass sharpens the question rather than softening
it: the two verbs are now separate types with separate parsers and separate facet
vocabularies, sharing only a version number.

### 3. `host` vs `snapshotHost` naming (lowy F9)

hickey did not raise this. Routed here as naming taste — the split renamed the
*parsers* but deliberately left the two spawn entry points alone.

> **lowy F9** — Read the exports list as a new consumer. `snapshotHost` and `host`
> differ by a prefix and answer entirely different questions: one names a *scope*
> of the `snapshot` verb, the other names a different *verb*. `host` is also an
> unqualified word to export from a package — `import { host } from
> "osfacts-client"` reads as a noun, not a call. The naming test fires: you cannot
> tell from the name which volatility each sits behind.
> **Suggestion:** Rename `host` → `hostFacts` (keeps the CLI verb legible, removes
> the collision) and, for symmetry, consider `snapshotHost` →
> `snapshotAllProcesses`. Update `index.ts`, `client-ts/README.md`, the Atlas
> contract-table row (`os-facts-tool.mdx:57,60`), and `client.test.ts`. No
> consumer outside the package uses either today, so the rename is free now and
> will not be after drishti adopts.

The timing argument is the whole call: free right now (zero external consumers,
confirmed by grep), not free after drishti adopts.

### 4. An executable oracle for the TSV grammar (hickey #8)

lowy did not raise this. Routed here as product scope — it is a substantial new
surface (a second Rust serializer plus an `Arbitrary<Snapshot>` for proptest),
not a cheap fix.

> **hickey #8** — Tag, arity, and column meaning are stated six times — the two
> `write_tsv` bodies, the TS parser, `v2_contract.rs`'s positional indices,
> `live_oracle.rs`'s ad-hoc `split('\t')` with hardcoded column numbers,
> `cli_contract.rs`, and the prose tables in `README.md` and the Atlas note. The
> tests meant to *guard* the grammar re-implement it from the same assumption they
> are testing, so a writer bug that shifts a column shifts the test's index with
> it unless a human notices. This diff is the proof of cost: adding one field to
> `SourceError` required simultaneous, coordinated edits to the struct, two
> `write_tsv` bodies, both `source_error` helpers, two inline `SourceError`
> literals in `darwin.rs`, three `main.rs` tests, the TS interface, the TS union,
> the TS parser arm, `index.ts`, `scan.ts`, and two documents. Nothing would have
> caught a missed one except runtime.
> **Suggestion:** Add `parse_tsv` / `parse_host_tsv` to `schema.rs` as the inverse
> of `write_tsv`, rewrite the positional-index test sites to assert against the
> typed structs, and add one property to `prop_pure.rs`:
> `parse_tsv(write_tsv(snap)) == snap` over an arbitrary `Snapshot`. That single
> test pins tag, arity, ordering, and every `-`/JSON escape convention in one
> place — and it is the test that would have caught the `facet` field being added
> to one writer and not the other.

Weigh it against what this pass already closed: the `E` writer is one function
instead of two, facet names are an enum instead of 49 literals, and `facets.json`
pins the vocabulary across the language boundary from both sides. The specific
drift hickey cites as motivation is closed. The general property — that the tests
re-derive the grammar they guard — is not, and the stale darwin-lane JSON
assertion this pass found is direct evidence for it.

---

<sub>Gates green before every commit: `just check` · `pnpm --filter osfacts-client
--filter @kolu/padi test:unit` · `cargo test` · `cargo fmt --check` · `just fmt` ·
`nix build .#osfacts` · `just atlas::check-sync`. Not pushed, not merged.</sub>


---

## Adjudication of the valve-2 items

**1. `ci/mod.just` encodes scheduling as dependency edges — dropped, recorded.**
hickey's count was checked and is exact: `osfacts-live`'s dependency list omits
9 of `default`'s recipes — `website-pnpm-hash-fresh`, `surface-examples-nix`,
`solid-browser-example-nix`, `odu-nix`, `smoke`, `fmt`, `biome`,
`surface-example-build`, `surface-app-example-build` — 4 of which are heavy Nix
builds. (`nix` itself *is* covered, transitively through `osfacts`.) So the
recipe comment's "CI runs the smoke after its heavy fan-out drains" is not
delivered as written today. It is still dropped from this PR: the edge list
predates the gauntlet, no run has been shown to fail because of it, and the perf
budget is already process-count-scaled specifically to absorb contention.
Rewiring the CI DAG in the same change that must then be validated *by* that CI
puts the pipeline's own behaviour inside the diff under test. Worth its own PR,
with the numbers above as the starting evidence.

**2. One `SCHEMA_VERSION` for two independently-evolving verbs — dropped.**
The version's job is that a consumer built against a different revision fails
loudly instead of parsing a half-understood shape. One binary emits both verbs
and they ship together, so one number answers that question completely; two
version axes would be more state for a consumer to track, not less. Revisit if
the verbs ever ship from separate binaries.

**3. `host` vs `snapshotHost` naming — rename dropped, the system documented.**
lowy is right that the pair reads badly cold, and right that a rename is free
now and not after drishti adopts. But the names are already systematic rather
than ad hoc: `snapshotSubtree` / `snapshotPids` / `snapshotHost` are the three
*scopes* of the `snapshot` verb, and `host` is the scopeless `host` verb — so
renaming one would break the scheme rather than reveal it. The verb split landed
in this review already makes the confusion a type error (`SnapshotReading` vs
`HostReading`), which is the protection that actually matters. What was missing
was that the scheme is nowhere stated, so it now is, in the package's export
docstring.

**4. An executable oracle for the TSV grammar — dropped, recorded as follow-on.**
This is net-new test apparatus (a second Rust serializer plus a proptest
`Arbitrary<Snapshot>`), not a defect fix. The concrete drift hickey cited is
closed by the `Facet` enum and the two-sided `facets.json` pins. The general
property — that every writer and every parser agree about the grammar — remains
worth an oracle, and should be its own change where it can be judged on its own
merits.

srid:
## Grok ⇄ Claude Code debate

✅ Consensus in 2 round(s)

|---|---|---|
| 1 | de1a1fcf | fix(osfacts): four ways a failure still reached a consumer as a plausible value |

- **F1** — runOsfacts uses promisified execFile and, on any rejection, throws OsfactsClientError('spawn', ...) without reading stdout
- **F2** — On --ports, listener_claims() runs first and fills `claims` from the same-uid fd walk
- **F3** — decode_host_listeners claims 'Fails loudly on a record it cannot frame' and tests truncated mid-record as EINVAL, but two paths still return Ok(Vec::new()) for malformed input: (1) header_len is taken from the first u32 and advanced with round_up_8 without checking header_len <= bytes.len() — a huge/corrupt header makes offset past the buffer, the loop never runs, Ok(empty) → caller's BLIND_OR_EMPTY, not EINVAL; (2) the record bound `offset + len > bytes.len()` uses wrapping usize addition, so a huge kernel-supplied len can wrap the sum, pass the check, and admit a phantom INPCB read from fixed offsets before the cursor saturates away
- **F4** — host_pids() does `if let Ok(rd) = fs::read_dir("/proc") { ..

srid:

Run with `--no-elegance` (the gauntlet's `/simplify` step already swept this
tree). Pass 1 (rules) and pass 2 (fact-check) over `5fdf32e7...HEAD`.

**5 issues, all fixed** in `2ac30ec1`.

| Pass | Rule / finding | Fix |
| --- | --- | --- |
| fact-check | Numeric TSV parsers accepted an empty field as a plausible `0` | shape checked before coercion, with a test |
| fact-check | `.flatten()` dropped per-entry `readdir` failures | entry errors propagate; subtree loss becomes a `U <root> proc` row |
| rules | `dry-rule-of-three` — three near-identical numeric validators | one shared body, differing only in shape + predicate |
| rules | `dry-rule-of-three` — `vm_deallocate` size arithmetic at three call sites | extracted as `release_processor_info` |
| rules | `no-silent-error-swallowing` — five undocumented `let _ =` discards | each now states why it cannot propagate |

All 30 other rules (built-in + the 22 project rules in `.agency/code-police.md`)
were checked individually and are clean; most of the kolu-runtime rules (SolidJS,
app-shell, toasts, subscriptions, watchers) are N/A — this diff touches no
`.tsx`, store, or watcher code.

### The two that were real defects

**`Number("")` is `0`, not `NaN`.** `integer` / `float` / `signedInteger`
coerced first and validated after, and `arity` only counts fields — it never
looks inside them. So an *empty* field, as opposed to a missing one, parsed as a
perfectly plausible reading: an empty `nice` read as "default priority",
indistinguishable from a real value, and likewise for `pid`, `ppid`,
`rssBytes`, `uid`, and listener `port`. This contradicted the client README's
own *"a mismatched format fails loudly"*, and the sibling validators in the same
file (the `L`-row regex, `STAT`'s length check) already do validate shape.

**The flattened `readdir`.** The failed *open* was propagated — that was this
branch's own earlier fix — but the iterator was then `.flatten()`ed, so an entry
failing mid-listing vanished with no `U` row, no `E` row, and a
shorter-than-real result that reads as a complete, healthy walk. The same defect
class as the fix it was sitting next to, one level down.

### One suggestion not taken as written

Police noted the two `let _ = vm_deallocate(...)` discards were inconsistent
with the third call site, which checks the status and returns `Err`. Made
consistent in the *other* direction from the obvious one: the success path keeps
its check and still fails loudly, and only the two paths that are already
returning the failure which actually blinded the read discard the release
status — replacing a real blindness cause with a release error would report the
wrong thing. Levelling the existing guard down to match the weaker two would
have been a fail-fast regression.

PR #1851, opened 2026-07-16
fix(kaval): reap live PTYs on daemon shutdown (no orphaned spawn-helper)

srid:
## What & why

The kaval daemon never reaped the node-pty PTYs it owned when it shut down: the only reaper `PtyHost.dispose()` (ptyHost.ts) was called from a **test lib only**; every production shutdown path awaited the surface runtime's `close()`, documented as "releases nothing today". So a shutting-down daemon left each live PTY's `spawn-helper`/shell subtree behind.

**The fix (one seam):** `servePtyHost` composes its returned `close` to call `host.dispose()` before the surface runtime closes — the *"daemon owns shutdown by construction"* the `close` handle was documented to exist for. `daemonMain`'s shutdown `.finally` already awaits `close`, so the pid-gone self-exit, SIGTERM, and abort paths all reap now. This completes [#1771](https://github.com/juspay/kolu/pull/1771) (that fix made a run-bound kaval *exit* with its run; this makes an exiting kaval *reap its PTYs*).

## What this does and doesn't fix (honest scope — refined by the reproduce-first work)

- **On darwin the OS reaping is INTERMITTENT.** A non-tty-aware PTY leader (a program run directly under the pty, not a shell) is *usually* hung up by the kernel on the pty master's close — but **unreliably** (observed leaking once, then reaped 5/5 on a later batch on `rasam`). A **shell** leader always self-reaps on the hangup. So the pre-fix leak is real but non-deterministic — the flaky, days-apart shape of the aged `spawn-helper` orphans found on `rasam`.
- **This fix makes the GRACEFUL-shutdown reap DETERMINISTIC** (dispose explicitly SIGHUPs each leader). That's a real reliability win for **production kolu shutdown** and the **odu run-end pid-gone** path. Verified: **5/5 green** with the fix; the covering test is deterministic (adds no flake).
- **It does NOT cover the test-harness SIGKILL path.** The CI box's main orphan source is tests SIGKILL-ing daemons directly (`hooks.ts`, `socketDaemon` reapers), which gives no graceful window — a **follow-up** (SIGTERM-then-SIGKILL) that would route those through this fix. So this PR is a reliability foundation, **not** by itself a proven fix for the darwin-under-load tracker flakes (which don't reproduce on a clean box at all).

## Evidence (reproduce-first)

[...]

PR #2014, opened 2026-07-27
fix(kaval): a dying PTY host reaps every leader it owns — no more orphaned spawn-helpers

srid:
## The evidence

Orphaned (ppid 1) node-pty `spawn-helper` processes, days old, found on the darwin CI box `rasam`, leaked by kolu test runs under odu:

```
14619  1  04:07:29 …/node-pty/build/Release/spawn-helper /tmp/nix-shell.NatSrp/kaval-e2e-cwd-YhL3w8 /bin/sh
18093  1  03-23:58:38 …same shape… /tmp/nix-shell.vlAFmG/kaval-e2e-cwd-4sYwtR /bin/sh
```

Three facts in that `ps` line pin the story. The cwd `kaval-e2e-cwd-*` names the spawner: `socketDaemon.test.ts`'s `makeCwd()`, and `/bin/sh` is its `DEFAULT_SPAWN_SHELL` — these are PTYs the kaval e2e suite spawned through a real daemon. The argv is still the **helper's** (`spawn-helper <cwd> /bin/sh`): after a successful `execvp` the argv would read `/bin/sh`, so these processes are parked **pre-exec**. And ppid 1 with ages in days means nothing in the system ever had a way to kill them.

## The mechanism (confirmed vs inferred)

**Confirmed from the pinned fork's source** (`src/unix/spawn-helper.cc`, `src/unix/pty.cc` at the pinned rev): on darwin node-pty launches each PTY as `posix_spawn(spawn-helper)` with `POSIX_SPAWN_SETSID` — the helper starts as a session leader with **no controlling tty**, and acquires one only inside its own `open(ttyname(STDIN), O_RDWR)` before `chdir` + `execvp`. (The deleted-cwd theory is refuted by the same source: a failed `chdir` is a clean `_exit(1)`, and there is no handshake pipe to block on.) Until that `open` completes, no hangup — neither a `SIGHUP`-flavored kill nor the master side closing — reaches it. Linux has no such window: `forkpty` gives the child its controlling tty *at fork time*.

**Confirmed by reproduction on linux** (this machine), driving the pinned fork directly:

- SIGKILL the owning process right after spawn, plain `/bin/sh` leader → **reaped** every trial (master closes → kernel HUPs the session leader). The observed leak does not reproduce on linux with the observed shape — it is genuinely darwin-only.
- Same, but a SIGHUP-immune leader (`sh -c "trap '' HUP; exec sleep 100000"`) → **leaks on linux too**: alive, ppid 1. That is the *class*: any leader the master-close hangup cannot kill. Darwin's pre-exec helper is a member of that class; the trap-HUP leader is its portable model.

**Confirmed in kolu's code**, two gaps that leave a leader to exactly that unreliable OS backstop:

1. `createPtyHost().dispose()` — the path every graceful daemon shutdown runs (`daemonMain`'s `.finally` → `ptyHost.close()`, #1851) — killed each leader with node-pty's **default `SIGHUP`**, which is only advisory. The neighboring `terminate` (the `kill` RPC) had already chosen `SIGKILL`, with a comment naming the very window: *"node-pty's default SIGHUP can lose an immediate post-spawn kill while the child becomes its session leader."*
2. `socketDaemon.test.ts`'s reap backstops SIGKILLed a still-live daemon by gate pid — skipping the daemon's shutdown disposal entirely — and the suite's one deliberate daemon-SIGKILL with a live PTY (`SIGKILL mid-attach`) strands a `/bin/sh` spawned into a `kaval-e2e-cwd-*` dir, matching the rasam orphans field for field. Each leader sits in its own session (`setsid`), so no kill aimed at the daemon or its group ever reaches a leader; once the daemon is SIGKILLed, darwin's intermittent master-close hangup is all that's left — and its misses are the aged orphans.

**Inferred** (not directly observed — rasam offers no live specimen to probe): which precise blocking state each aged helper is in. What the evidence establishes is that they are pre-exec, hence outside the reach of every hangup mechanism; the fix does not depend on the finer diagnosis, because `SIGKILL` reaps a pre-exec helper regardless.

## The fix — one mechanism, at the seam that owns the volatility

The PTY host owns its children's lifetimes, so the host's shutdown is where the guarantee lives:

- **`dispose()` now SIGKILLs each leader** — the same choice, for the same reason, as the `kill` RPC's `terminate`. A host going away must reap, not politely hang up; anything less delegates to the OS backstop this bug proves unreliable.
- **The test backstops route through that guarantee**: `reapRendezvous` now SIGTERMs a still-live daemon and waits for it to run its shutdown (which reaps its leaders), escalating to SIGKILL only for a wedged daemon; the daemon-block `afterEach` reaps rendezvous before child handles so the graceful window isn't forfeited; and the `SIGKILL mid-attach` test reaps the one leader it deliberately strands.

Pinned by a new test in the suite's own style: spawn a SIGHUP-immune leader, run the exact backstop call, assert the daemon exited gracefully (code 0) **and the leader is gone**. RED before the `dispose` fix (leader survives, verified on linux), GREEN after — and it guards both layers, since either regression (SIGHUP dispose, or SIGKILL-first backstop) turns it red.

## Named follow-up (out of scope here)

The pre-exec window itself belongs to the pinned node-pty fork (juspay/node-pty): between `posix_spawn` and the helper's slave `open()`, a leader exists that *no* hangup can reach, so any owner that dies by SIGKILL in that window — e.g. an external process-group kill of a whole test run, which no in-repo code can intercept — can still strand a helper. A fork-side fix (acquiring the ctty via the spawn file actions, or an early `O_NOCTTY`-then-`TIOCSCTTY` sequence, so the window never exists) is the structural close; filing that against juspay/node-pty is the follow-up this PR does not paper over.

## Test evidence

- `packages/kaval` full unit lane (KOLU_DAEMON_TESTS=1): 17 files, 175 tests green, including the new pin RED→GREEN.
- `just check` (typecheck + biome) and `just fmt` clean.


srid:
## CI e2e metrics — run `6d6f6a2#2`

Full two-platform run settled green: 44 nodes ok, 0 failed, 0 errored.

| x86_64-linux | kolu-ci-3 | 8 | 32 | 8 | 2m51s (171151 ms) |
| aarch64-darwin | nix-infra@rasam.tail12b27.ts.net | 6 | 24 | 6 | 4m36s (275639 ms) |

Durations are the whole `ci::e2e@<platform>` recipe wall time from the run ledger (`.ci/6d6f6a2/runs/2.json`); worker lines are the last `e2e: workers=` record in each platform's `ci::e2e.log`.

PR #2017, opened 2026-07-27
Code tab: an eye toggle reveals gitignored files as dimmed rows

srid:
The Code tab's browse tree comes from `git ls-files --cached --others --exclude-standard`, so gitignored files were simply invisible — annoying when the file you want is a build artifact, a local config, or a generated log. **A new eye toggle in the browse toolbar overlays the ignored set as dimmed rows**, and the choice persists per device.

Two things keep it cheap. Git stays the only authority on ignored-ness: a new `fs.listIgnored` procedure runs the *collapsed* listing (`--others --ignored --exclude-standard --directory`), so a fully-ignored directory arrives as **one trailing-slash entry** — `node_modules/` is a single dimmed row, never forty thousand. And the toggle is genuinely free when off: its query idles, so there is no extra pulse, no extra listener, and no extra spawn.

> **Ignored rows don't live-update, deliberately.** The working-tree watcher keeps its ignore set — recursively watching `node_modules` is the inotify storm that set exists to prevent — so neither the dimmed rows nor an open ignored file's contents repaint on disk changes. Documented in the watcher header and stated plainly in the user docs.

### What the reviews changed

The first draft worked and was wrong in three structural ways. Each was caught by a different pass, and each is worth naming because the fix was "use a different mechanism", not "add a guard":

| Found | Defect | Fix |
|---|---|---|
| architecture | `includeIgnored` on `fs.listAll` put a **display preference inside the file-list query's identity**, so every toggle blanked the list, unmounted the tree, and remounted it fully collapsed — discarding every hand-expanded folder | its own procedure + its own idle-unless-on query (regression-tested) |
| architecture | The overlay rode Pierre's `gitStatus` channel, which **rolls every entry into its ancestors' change counters** — falsely marking 47 directories on top of a real 77 here, and 68 on a clean checkout | its own stylesheet keyed on `data-item-path`; the roll-up stays honest |
| lens (both) | The inventory gained a second source but kept one source's readiness, so **an absent listing read as an empty one** — painting a tree of nothing but dimmed rows during a repo/host switch | one `treeInventory` fact; absence and emptiness are now distinct |

The Grok debate then found that the table test I wrote for those merge rules **could not fail** for the rule that mattered: every case exercised an absent listing only while still pending, so a readiness-based gate passed all eight assertions while breaking on the reachable error path (`surfaceError` clears `pending` without ever writing a value). Verified by injecting that shape — 8 passed against a broken implementation — and the added case fails it alone.

Simplify replaced a hand-rolled CSS escaper with `CSS.escape`, which the repo already uses for the same job and which covers a case the hand-rolled set missed entirely (a leading digit needs a numeric escape). It also removed two things earlier rounds had added: a per-tick filter defending an invariant its consumers already enforce, and a search-projection narrowing that put a stylesheet re-parse on every keystroke to avoid emitting selectors that are inert anyway.

CodeQL then caught a HIGH `js/polynomial-redos` in the directory-marker strip: `/\/+$/` backtracks quadratically on a path ending in many slashes, and these paths come from `git ls-files` rather than from us, so their shape isn't ours to bound. Replaced with a linear scan, with a 100k-slash regression test.

### Notes

`PADI_SURFACE_VERSION` 4.3 → 4.4 — a purely additive procedure, the plainest minor there is. The version gate is what keeps a 4.4 client from calling a procedure a surviving 4.3 padi doesn't have, and that leg is pinned by an assertion rather than assumed.

Two things recorded rather than fixed, both new mechanisms rather than cleanups: the generated sheet is O(ignored entries) with no cap (`--directory` only collapses *fully*-ignored directories, so a repo with scattered `*.log` ignores gets no collapse benefit), and the watcher's ignore computation could share one memoized spawn with the client query.

No visual evidence: capture was explicitly skipped for this PR at the author's request.

*Generated by [`/be`](https://github.com/srid/agency) on Claude Code (model `claude-fable-5`).*


srid:

Two structural lenses read this diff independently — **lowy** (volatility-based
decomposition: do the boundaries encapsulate an axis of change?) and **hickey**
(structural simplicity: what is complected, what is fragmented?). Neither saw the
other's notes. A single reconcile pass merged the two lists and applied them.

**Status: `needs-human`** — 22 findings in, all resolved: **18 applied** across 6
commits, 1 recorded as an observation, and **3 escalated** for a human call. The
escalations are refinements on top of fixes that already landed, not blockers.
`just check` is green; all four affected unit suites pass (1002 client / 21
solid-pierre / 106 git / 394 padi).

**Counts:** lowy 9 (5 major, 4 minor · 8 `fix`, 1 `drop`) · hickey 13 (6 major,
7 minor · 13 `fix`). Seven were **the same issue found twice** (`≡`) — two
independent lenses agreeing, applied once under one plan.

### The headline: the browse inventory had two authorities and one readiness

Both lenses landed on the same seam from different angles, and it was the only
finding with a reproducible user-visible symptom. `treePaths()` gained a second
source in this PR (the gitignored overlay) while the readiness that guards three
consumers still named only `fs.listAll`. Two consequences:

- `const tracked = allPaths()?.paths ?? []` read an **absent** tracked listing as
  an **empty** one. On a repo/host switch with the toggle on, both queries blank
  and race; if the overlay resolves first the tree can paint as nothing but
  dimmed `node_modules/` and `.env` rows until `fs.listAll` lands.
- Three guards hand-ANDed `allPaths.pending()` against a two-source inventory —
  including `repoVault`, whose own comment promised the paths+pending pair is
  minted together, *"never a drifted pair"*.

Both are fixed by one memo that mints the inventory, the overlay's contribution
and the readiness **together**, so the wrong pairing is not spellable: there is no
bare array left to pair with a readiness of your choosing.

### Findings

| # | Lens | Finding | Sev | Outcome |
|---|---|---|---|---|
| F1 ≡ F1/F2 | lowy ∥ hickey | Inventory gained a second authority; readiness named one — plus absent-read-as-empty | major | applied `0f95eb96` |
| F7 ≡ F1 | lowy ∥ hickey | "Ignored minus tracked" derived twice, in two memos | minor | applied `0f95eb96` |
| F2 ≡ F7 | lowy ∥ hickey | `shadowCss` and `ignoredPaths` were the same channel built twice | major | applied `4a6154ce` · placement escalated |
| F3 ≡ F8 | lowy ∥ hickey | Row appearance decided in the agnostic package via a knob nothing sets | major | applied `4a6154ce` · value escalated |
| F5 ≡ F4 | lowy ∥ hickey | Two comments describe the `fs.listAll`-keying design this PR reversed | major | applied `9a590a77` |
| F6 ≡ F6 | lowy ∥ hickey | `listIgnoredPaths` is an expendable boundary (its strip is unobservable) | major/minor | applied `3308f757` |
| F4 | lowy | Trailing slash is an untyped discriminator; breaks `resolveRef`'s stated precondition | major | applied `0f95eb96` (minimum form) · wire reshape escalated |
| F9 | lowy | Dimming sheet re-parsed on every repo pulse | minor | applied `4a6154ce` |
| F8 | lowy | Endpoint test double cast past its own interface | minor | observation (`drop`) |
| F3 | hickey | Dimmed-set comment claims a filter the code doesn't apply | minor | applied `0f95eb96` |
| F5 | hickey | "An entry is a directory" spelled four ways across three packages | major | applied `466e4fa3` |
| F9 | hickey | Overlay sheet applied at two sites, held apart only by `{ defer: true }` | minor | applied `4a6154ce` |
| F10 | hickey | "Browse-only" decided in three places, one of them dead | minor | applied `0f95eb96` |
| F11 | hickey | Toggle button reimplements `NavButton`'s chrome 720 lines below it | minor | applied `700b1108` |
| F12 | hickey | Dead `n` field in the `fs.listAll` test mock | minor | applied `700b1108` |
| F13 | hickey | `browse.test.ts`'s "complement of listAll" case never calls `listIgnored` | minor | applied `700b1108` |

### Commits

| SHA | Area |
|---|---|
| `3308f757` | the git package's ignored-listing boundary |
| `466e4fa3` | Pierre's directory-marker vocabulary |
| `0f95eb96` | the CodeTab browse-inventory / readiness seam |
| `4a6154ce` | the solid-pierre shadow-CSS receptacle |
| `9a590a77` | the show-ignored toggle's own docs |
| `700b1108` | minor polish (all areas) |

Deliberate decisions from the earlier architecture pass were **not** re-litigated
and are untouched: `fs.listIgnored` as its own procedure, the second retained
query idle unless the toggle is on, the overlay off Pierre's `gitStatus` channel
(the ancestor roll-up argument survives intact — it just moved onto
`ignoredPathsCss`, where the next reader will be standing), the deliberately
un-widened watcher, git as sole ignore authority, and dedupe-at-merge with
tracked winning.

<details>
<summary><b>Escalated to the author, and how each was adjudicated (both positions verbatim)</b></summary>

#### E1 — Direct contradiction: where `ignoredPathsCss` should live

Both lenses agree the second CSS channel must collapse into `shadowCss`. They
disagree on where the pure selector-builder ends up, and the two claims cannot
both be applied.

> **lowy F2:** "Keep `ignoredPathsCss` exported from `@kolu/solid-pierre` as the
> pure leaf it is (it is a fine leaf — see F3 for where its *content* belongs)."
>
> **lowy, electricity probe:** "`ignoredCss.ts` is a bounded string builder — the
> atlas's trap #3 exactly ('a clean, reusable, agnostic primitive that hides only
> a *bounded algorithm* is a **leaf**, not a receptacle'). It should stay a leaf."

> **hickey F7:** "Move `ignoredPathsCss` (and its test) to the client, beside the
> theme that owns it — `packages/client/src/ui/pierreIgnoredCss.ts`, next to
> `pierreTheme.ts`. […] `@kolu/solid-pierre` is a *generic* Solid wrapper over
> `@pierre/trees`, and it has just acquired a **git-domain** concept. […] Encoding
> 'gitignored' into a package that knows nothing about git means the next consumer
> who wants dimmed rows for a different reason either misuses a prop named for git
> or adds a third channel."
>
> **hickey F7, fallback:** "*If the semantic prop is judged worth keeping despite
> the above*: at minimum rename it `dimmedPaths` and rename `ignoredCss.ts` →
> `dimmedCss.ts`, so the generic package stops asserting a git fact it cannot
> verify."

**Applied:** everything both agree on — one reactive channel, the `ignoredPaths`
prop deleted, the appearance moved to `pierreTheme.ts`. The file was **not
moved** (the no-change option on the disputed axis, which is also lowy's
position). Since the git-named *prop* is gone either way, what remains in dispute
is the filename and the exported symbol's name.

**Discriminating experiment** — hickey's case rests on "the next consumer", so
the observable question is how many consumers the package has:

```console
$ grep -l '"@kolu/solid-pierre"' packages/*/package.json packages/*/*/package.json
packages/client/package.json
packages/solid-pierre/package.json

$ grep '"private"' packages/solid-pierre/package.json
  "private": true

$ grep -rln "@kolu/solid-pierre" --include=*.ts --include=*.tsx packages \
    | grep -v "^packages/solid-pierre/" | sed 's|/src/.*||' | sort -u
packages/client
packages/client/vite.config.ts
packages/solid-fileview      # comment reference only
packages/solid-statepip      # comment reference only
```

Private package, exactly **one** real consumer. That narrows it without settling
it: hickey's next consumer is hypothetical today, and lowy's leaf costs nothing
today. One line of work either way.

#### E2 — Judgment: what a dimmed row should look like

Both agree the `var(--trees-ignored-opacity, 0.5)` knob (a custom property Pierre
never defines and kolu never sets) must go, and that the decision belongs in
`pierreTheme.ts`. They differ on the value.

> **lowy F3:** "That file *already* declares kolu's answer to 'what does an
> ignored row look like': `var(--color-fg-3)`. The diff mints a second, competing
> answer (50% opacity) […] If a designer changes the ignored-row treatment they
> will edit `pierreTheme.ts:67` and nothing will happen, because the row is
> actually dimmed by a constant in another package." Suggested constant:
> `export const KOLU_IGNORED_ROW_DECL = "color: var(--color-fg-3); opacity: 0.6;";`

> **hickey F8:** "emit the literal. `return \`${selector} {\n  opacity: 0.5;\n}\n\`;`
> — and […] hoist it to a named constant there (`const IGNORED_ROW_OPACITY = 0.5;`)."

**Applied:** the shipped value as `pierreTreesIgnoredRowDecl` in
`pierreTheme.ts`. Taking lowy's declaration verbatim would change what users see
(a colour override on the row name plus a different opacity) with no visual
evidence gathered in this pass — a design call, not a structural one. A designer
now edits one line in the theme under either answer.

#### E3 — Judgment (scope): the typed wire for `fs.listIgnored`

> **lowy F4:** "Make the kind explicit on the wire, so no consumer parses a
> suffix. […] `files` / `directories` […] Critically, `resolveRef`'s `repoPaths`
> and `repoVault.paths` then receive `files` only, restoring the documented
> precondition.
>
> If the wire reshape is judged too large for this PR, the minimum that closes the
> correctness hole: at `CodeTab.tsx:438` and `:1144`, pass `treePaths().filter((p)
> => !p.endsWith('/'))` and add a line to `resolveRef`'s doc naming the caller's
> obligation. I prefer the typed version — filtering at two call sites is the
> seventh site that knows the convention."

hickey did not raise the wire shape; its F5 attacks the same smell from the other
end (name the predicate once), which composes with either form.

**Applied:** the minimum, in a form that answers lowy's own objection to it — the
filter is spelled **once**, in `treeFilePaths`, through hickey's named
`isDirectoryPath`, so it adds no new site that knows the convention. The
correctness hole is closed (`resolveRef`'s basename index was being poisoned with
an empty key by `basename("node_modules/")`, and a terminal ref to a visible
collapsed directory row toasted "not found"). The wire reshape stays open: it
touches `schemas.ts`, `browse.ts`, `endpoint.ts`, `hostCodeTab.ts` and the client
merge, and the surface contract is already at an unreleased 4.4, so it is cheap
to take now if wanted.

</details>

<details>
<summary><b>Recorded, not applied — 1 observation</b></summary>

**lowy F8 — the endpoint test double is cast past its own interface**
(`servePadi.testlib.ts:26-33`). `as unknown as TerminalEndpoint` lets
`listIgnored: async () => []` typecheck against an interface returning
`{ paths: string[] }`. Dispositioned `drop` by its own lens: pre-existing
scaffolding, symmetric with the existing `listAll`, and the fix touches members
this feature does not own. On the record because this diff is where the pattern
gained a second instance — and because the file's own header says it exists "so a
`TerminalEndpoint` shape change is a one-file edit", which the cast defeats.

</details>

---

### Author's adjudication of the three escalated calls

`/be` §4 requires the author to settle anything the reconcile pass escalated, rather than leaving it open. All three are now closed.

**E1 — where the CSS builder lives → resolved by a third option, in `54dfda26`.**
Both lenses were right about different halves. Parameterizing the declaration body (applied earlier in this review) had already removed the git *decision* from the module; what was left was git *vocabulary* on a function that had become genuinely generic — `paths + decl → CSS`. So rather than picking a package, the function was renamed `ignoredPathsCss` → **`rowPathsCss`** and the module `ignoredCss.ts` → `rowPathsCss.ts`. What it knows is Pierre **anatomy** (a row is addressed by `data-item-path`; how a path survives as a CSS attribute value) — squarely `@kolu/solid-pierre`'s business, which is lowy's point — while hickey's objection loses its referent, because there is no longer a git concept in the git-agnostic package. The git-specific reasoning (why the overlay must not ride Pierre's `gitStatus` channel, with the measured 47/68 falsely-marked directories) moved to the client call site, where a reader asking "why not just use `gitStatus`?" is actually standing — which was hickey's other stated wish.

**E2 — what a dimmed row looks like → keeping `opacity: 0.5`, as applied.**
The part both lenses agreed on (kill the phantom `--trees-ignored-opacity` knob, move the decision into `pierreTheme.ts`) is applied and is the part that mattered: a custom property nothing sets is an override knob with no configurator, which the repo's conventions forbid outright. lowy's additional `color: var(--color-fg-3); opacity: 0.6` is unevidenced visual taste — changing what users see with no visual comparison in this pass — and `0.5` matches the "dimmed rows" language already shipped in the docs and changelog. One lever, not two.

**E3 — a typed `{ files, directories }` wire → keeping the applied minimum.**
The distinction is already carried losslessly by git's own trailing-slash convention, and after `466e4fa3` the predicate is spelled exactly once (`isDirectoryPath`), which was lowy's actual objection to the minimum. Reshaping `fs.listIgnored`'s output again would also re-open a wire contract that was just versioned at 4.4. Recorded as a follow-up idea, not scope for this PR.

srid:
## Grok ⇄ Claude Code debate

✅ Consensus in 3 rounds — final verdict `approved: true`, no open findings.

Grok reviewed read-only in a split terminal; every finding was fixed rather than deferred or disputed.

|---|---|---|
| 1 | 84067ec1 | fix(debate): pin the browse-merge rules, escape newlines in row selectors |
| 2 | 803affe9 | test(debate): pin absence independently of readiness; escape form feed |

**Findings raised, by round:**

- **R1·F1** (minor) `packages/client/src/right-panel/CodeTab.tsx` — The readiness/dedupe/absent-tracked rules in treeInventory — the exact site of two earlier bugs — are only argued in comments; nothing pins them with a pure unit test.
- **R1·F2** (minor) `packages/client/src/right-panel/hostCodeTab.ts` — Header still says the owner holds "the six pulse-then-requery reads" (and the buildHostCodeTab docblock at ~line 82 repeats "six"), but ignoredPaths is a seventh retained query..
- **R1·F3** (nit) `packages/solid-pierre/src/rowPathsCss.ts` — escapeAttrValue only handles \\ and ".
- **R1·F4** (nit) `website/src/content/docs/code-tab.mdx` — "so a peek inside node_modules/ is one row" slightly overclaims: the row is the collapsed directory itself, not a view of its contents.
- **R2·F1** (minor) `packages/client/src/right-panel/browseInventory.test.ts` — Rule 1 is only asserted for absent tracked with trackedPending: true.
- **R2·F2** (nit) `packages/solid-pierre/src/rowPathsCss.ts` — CSS Syntax treats U+000C FORM FEED as a newline alongside LF and CR; an unescaped FF still terminates a double-quoted attribute value and can drop the whole replaceSync sheet.

Round 1 approved with four non-blocking findings; all four were fixed anyway. Round 2 was the sharpest exchange: Grok checked whether the new table test could actually **fail**, and found the rule-1 cases only exercised an absent tracked listing while the query was still pending — so a readiness-only gate would pass all eight assertions yet still admit the overlay on the reachable error path (`createPolledQuery.surfaceError` clears `pending` without ever writing a value). That was verified by injecting the buggy shape: **8 passed against a broken implementation**, and the added case fails it alone.

srid:

Pass 1 (rules) + Pass 2 (fact-check). Pass 3 (elegance) skipped — `/simplify` had already run over this tree.

| Pass | Findings | Outcome |
|---|---|---|
| Rules | 1 | Fixed in `47c9c68e` |
| Fact-check | 0 | Clean |

**`caught-error-must-not-collapse-to-empty`** — `boolPref` (the new `showIgnoredFiles` toggle's backing store) accepted no `onInvalid` at all, so a corrupted `localStorage` value could only ever reach a console warning; no caller could opt into a visible surface even where one was warranted. Now mirrors `perHostPref`'s shape — optional `onInvalid`, defaulting to the same warn-only behaviour, so the current call site is unchanged.

The fact-check pass specifically traced the inventory merge, the stylesheet lifecycle, and the toggle/watcher interaction for races or silent staleness, and found none.

srid:
## CI e2e metrics — `c2eb22c#1`

Both platforms settled green; CodeQL green.

|---|---|---|---|---|---|
| `x86_64-linux` | kolu-ci-3 | 8 | 32 | 8 | 2m 42s |
| `aarch64-darwin` | rasam | 6 | 24 | 6 | 4m 23s |

Durations are the whole `ci::e2e` recipe wall time from the run ledger, not Cucumber's internal timer.

_(Supersedes the `58fb960#2` run, which was green on all 44 status checks but carried a new HIGH CodeQL alert — `js/polynomial-redos` on the directory-marker strip — fixed in `c2eb22c3`.)_

PR #2018, opened 2026-07-27
Provisioning carries cache binaries to the host — and a cache-blind path is unspellable

srid:
Stacked on #2016, which proved the gap this closes: **a remote host realises the agent with its own `nix.conf`, so no flake-declared cache ever reaches it.** An ssh-ng provisioning run rebuilt, on the host, a `padi` that sat in the declared cache the whole time. This PR makes provisioning carry the binaries itself, and makes it structurally impossible for any consumer of the surface stack to provision cache-blind.

### The mechanism

Provisioning gains two copies around the existing build, because **both** are load-bearing and each was verified live:

1. **Prefetch** — `nix copy --from <declared cache>` into the binder's **local** store. That is the only seat where a declared cache can act: a remote-store realisation substitutes with the *remote* daemon's `nix.conf`, and client-side substituter settings never participate.
2. **Ship** — `nix copy --to ssh-ng://<host>`. The cold build would *not* do this: it transfers only the `.drv` closure, and locally-valid outputs are never consulted. Verified with a nondeterministic derivation — the remote **rebuilt** an output that was valid locally.

When the ship lands, the cold build is **skipped entirely** — the closure is already valid on the target, so provisioning goes straight to the required GC-root commit, the same short-circuit the warm fast-path takes. That is where the cache actually pays off: a full flake evaluation plus an ssh-ng round trip, gone.

Honest about the limit: the ship lands only when the target trusts it (your ssh user in its `trusted-users`, or signatures it trusts). When it doesn't, kolu says so in the connect log and the host realises the agent itself, exactly as before — and the quickstart's `nix.conf` advice applies there. Cross-architecture is fine throughout: fetching and copying execute nothing.

### The forcing — three layers

| Layer | Gate | Who it forces |
| --- | --- | --- |
| `mkProvenAgentSource` derives `binary-cache.json` from the agent flake's own `nixConfig`, **failing eval when absent** | binder's `nix build` | any consumer assembling an agent source |
| every `AgentDerivation` arm requires a nominal `AgentBinaryCache`, built only by `agentBinaryCache({…})` | compile time | drishti, odu — at their next pin bump |
| `resolveAgentDrv` reads the sidecar; typed `AgentBinaryCacheUnbakedError` (its own fault kind, cause and host-down card) when a baked source predates the contract | dial time | stale wrappers |

One source of truth throughout: the declaration in `nixConfig` is what a manual `--accept-flake-config` build honors, what the sidecar serializes, and what the copies act on. The sidecar's *filename* comes from `agent-env.json`, the registry that already carries `SURFACE_AGENT_FLAKE_REF` — so neither language spells the literal.

### Review

Full `/be` gauntlet, 10 commits. The architecture pass is what caught the false premise above; the two structural lenses independently found the same defect from different angles (one concept across seven files, three implementations of "is this declaration usable?"); Grok reached consensus in two rounds after catching that both copies lacked the `-v` the cold build passes deliberately — without it a healthy slow transfer is silence-killed and narrated as a *miss*, inverting the feature by timeout. Details in the three review comments below.

_277 surface-remote tests; the full unit lane green across all 49 packages; `tsc` clean in surface-remote/server/client/common; `nix eval .#koluBin`, the typecheck check, and the examples' `nix build .#agent-source` all evaluate._

### ODU-IMPACT VERDICT: breaks-at-bump

Grounded by a grep of odu master at its pinned kolu rev (`3ff256f0`): `src/coordinator/runnerFlake.ts` wraps its runner drv in `directAgentDerivation(drvPath)`, which stops compiling at the next pin bump and must pass a cache declaration — odu's own flake already carries the exact `nixConfig` block to derive it from. `lane.ts`/`lease.ts` are untouched. Ledger item filed on juspay/odu#43.

### Drishti pair PR

srid/drishti#124 — adapts `archMap.ts` and bakes drishti's own declaration. Being re-pinned to this branch's final HEAD, since the API moved during the gauntlet (the nominal type now requires the `agentBinaryCache({…})` constructor rather than an object literal) — which is the compile-time forcing doing its job on the first real downstream consumer.

### Try it locally

```sh
nix build github:juspay/kolu/feat/provision-cache-prefetch --dry-run
```


srid:
Drishti pair PR: https://github.com/srid/drishti/pull/124 (pin bumped to this branch's HEAD, `archMap.ts` passing drishti's own baked cache declaration; MERGEABLE, CI running). Will re-pin to the merged kolu SHA before it lands, per the surface pairing rule.

srid:

**Status: applied** — 28 findings (lowy 12 · hickey 16). 24 applied by the reconcile pass (7 `≡` where both lenses independently raised the same issue), 2 valve-1 contradictions settled below, 2 valve-2 calls adjudicated below.

Both lenses converged, from different directions, on one structural defect: **the binary-cache declaration was one concept scattered across seven files with nothing enforcing its coherence.** hickey counted three independent implementations of "is this declaration usable?" (the Nix `require`, `isNonEmptyStrings`, `assertUsableBinaryCache` — the last two running back-to-back on the same value with different predicates and different error types); lowy reached the same place through ownership: the `binary-cache.json` wire format had no owning package, its name spelled five times across two siblings, while the repo already owns exactly that receptacle in `agent-env.json`. The fix is a single `agentBinaryCache.ts` with a nominal type behind one smart constructor, and the filename served from `agent-env.json` so neither language spells the literal.

| Commit | Area |
|---|---|
| `6ff3bb2b` | the provisioning core — `agentBinaryCache.ts` + `agentDerivation.ts` extracted, cache branded behind one constructor, filename via `agent-env.json`, `Result` reader hoisted above the ssh probe, `CopyOutcome` replacing the inverted booleans, ship gated on provable local validity, its own silence budget, honest per-URL narration |
| `3b9211c3` | the nix bake path — both nix cache spellings honored, `fileName`/`json`/`installToOut` on the recipe, `agent-source-tree.nix` owning the layout |
| `8a273f12` | the examples — flake rides the shared bake, one cache declaration per site, docs lead with `readBakedBinaryCache` |
| `95158f8e` | the padi drv-fault table — duplicate `match` arm collapsed to a row |
| `1ab69e84` | minor polish — the dev sidecar is a real `just` prerequisite |
| `4c6cd4c8` | the two valve-1 adjudications |

### Valve 1 — contradictions, settled

**The prefetch mechanism** (hickey F9 wanted one `nix build --max-jobs 0 --substituters "<all>"`; lowy F11 wanted the loop justified or replaced) — **settled by experiment, loop kept.** Nix documents `substituters` as honored only when the caller is in `trusted-users` or the URL is in `trusted-substituters`, so a client-supplied list is silently dropped for exactly the untrusted user this feature serves. `nix copy --from <uri>` names a store directly and dodges that filter — verified live: a `--from` copy fetched from a cache absent from the local `substituters`. Rationale now recorded in the doc comment, as lowy asked.

**`readBakedBinaryCache`'s parameter type** (lowy F4 needed an arbitrary local dir; hickey F12 needed a `github:` ref to be unable to reach `readFileSync`) — **synthesis, both needs met.** The parameter stays a string; the function now rejects any non-local ref scheme up front with its own accurate remedy. The remedy moved onto the error as a parameter, because the two ways to reach that fault have different fixes — defaulting one onto both is how a true error grows a false fix.

### Valve 2 — judgment, adjudicated

**Where `binary-cache.nix` lives** — lowy: the reader's package should own the contract. **Kept in `surface-daemon/nix`:** every other bake recipe lives there, `surface-remote` has no `nix/` directory, and the cross-language anchor the lens wanted — the filename — already comes from `surface-remote/agent-env.json`, so the reader's package does own the name.

**Should fleet-top derive its cache for real** — lowy: a runnable tutorial spends a doomed `nix copy` per dial. **Deferred, recorded.** What propagates (the teaching) is fixed: docs lead with `readBakedBinaryCache`, the constant is declared once, its comment names the real pattern, and the host is reserved-invalid so it fails fast and can never read as a real endpoint. Giving fleet-top a baked source needs a new env var plus two tutorial pages — product scope, not a defect here.

_270 surface-remote tests green; `tsc` clean across surface-remote/server/client/common; `nix eval .#koluBin` and the examples' `nix build .#agent-source` both evaluate._

srid:
## Grok ⇄ Claude Code debate

✅ Consensus in 2 round(s) — base `74ad561f`

|---|---|---|
| 1 | e1f5c7c2 | fix(debate): -v on both closure copies, honest abort/skip narration, and the missing cause test |

- **F1** — prefetchAgentClosure and shipAgentClosure run under progress-liveness with PROVISION_COPY_SILENCE_MS (120s, non-escalating) but invoke `nix copy` with no `-v`/`--verbose`
- **F2** — isLocallyValid collapses every non-ok ExitResult to `false`, including `kind: "aborted"` (and lifetime-expired)
- **F3** — After hoisting the local outputs query to step 0, the warm path always prints `checking for a cached agent…` whenever `!isLocal && rootPath !== null`, even when `localAgentPath` is undefined (outputs query failed — e.g
- **F4** — SshConnectorOptions.resolveDrvPath JSDoc still says `Pass directAgentDerivation(drvPath) when the caller already owns the store path` — the pre-break one-argument form
- **F5** — The binder's new `binary-cache-unbaked` → `agent-cache-unbaked` mapping (remotePadiBinding.ts UNSET_HINT_FAULT_CAUSE) has no server-side test
- **F6** — agentBinaryCache validates `s.trim().length > 0` but stores the untrimmed strings

srid:

**Pass 1 — rules: clean.** All 33 rules run (9 built-in + 24 project rules from `.agency/code-police.md`), each with grep evidence or enumerated candidates. Two worth naming because they read like violations and are not:

- `no-sync-blocking-on-the-serving-loop` — the sidecar `readFileSync` sits on the dial path, but lands in the rule's explicit carve-out (a few hundred bytes, known-local store path, one-shot). Every subprocess still routes through async `runCapture` under a bounded `LifetimePolicy`.
- `prefer-ts-pattern` — `remotePadiBinding.ts` *replaced* a `match(...).exhaustive()` cascade with a `Record<Union, T>` table, which is the rule's own "pure A→B mapping" exception rather than a breach of it.

One heads-up, below the trigger threshold: `feature-subsystem-gets-a-directory` fires at 3 new sibling modules; this diff added exactly 2 (`agentBinaryCache.ts`, `agentDerivation.ts`). A third binary-cache-adjacent module should come with a directory.

**Pass 2 — fact-check: 1 finding, fixed in `06905f25`.**

`readBakedBinaryCache` ran on *every* `resolveAgentDrv` call, ahead of the `drvCache` hit check — so on a hit the result was computed and thrown away, and, worse, a transient read fault (a permissions blip, a concurrent rebuild) could fail a dial the cache had already proven good. That is a new failure mode on the path that exists to avoid work.

Fixed by remembering the **successful** read per source ref, rather than by moving the read below the cache check as first proposed: the read has to stay *ahead* of the ssh probe — a deterministic local misconfiguration must win the race against a nondeterministic network fault, which is why a pre-contract wrapper reports itself instead of `auth-required` against an unreachable host — and the value is immutable for a given ref, since a baked ref is a content-addressed store path. Only successes are remembered, so a genuinely broken source keeps failing and a repaired one is picked up. Covered by a test asserting one read across repeat dials, and that a proven-good sidecar survives a later read fault.

**Pass 3 — elegance: skipped** (`--no-elegance`); `/simplify` had already run over this tree in the same gauntlet.

srid:
## ⏱ CI e2e — `06905f2#2`

Two-platform run, settled **passed** in 418s wall. Durations are the whole `ci::e2e` recipe from the durable run ledger (`.ci/06905f2/runs/2.json`), not Cucumber's internal timer; parallelism is the last `e2e: workers=` line each lane actually resolved.

| `x86_64-linux` | `kolu-ci-4` | 8 | 32 | 8 | 169.0s |
| `aarch64-darwin` | `nix-infra@rasam.tail12b27.ts.net` | 6 | 24 | 6 | 296.9s |

PR #2019, opened 2026-07-27
One attention vocabulary: the tab tells you who needs you, the dock takes you there

srid:
An agent sat blocked on a question for **20 hours** and nothing on screen made it legible: the needs-you pip rendered at 55% alpha, the section headers wore six decoy row-count badges, the host tab's violet pill couldn't be acted on, and a fresher busy row sat above the blocked one. The root cause was structural — the same fact rendered as four unrelated, hand-rolled visuals — so this PR replaces them with **one attention vocabulary, computed by one fold, rendered by one component**, and deletes the old spellings so the drift can't recur.

### The vocabulary

Three hues, three meanings, three clearing rules — each rendered identically at every altitude (host tab · host switcher · mobile chip · dock section header · dock row):

| Hue | Meaning | Shape | Cleared by |
|---|---|---|---|
| **violet** `--color-alert` | agent blocked on you (`awaiting_user`) | capsule count · row wash · glowing full-strength pip · **live wait-duration chip** | *answering the agent — nothing else* |
| **amber** `--color-attention` | finished, unopened | capsule count · row wash · corner badge | opening that terminal |
| **rust** `--color-busy` | agent in flight | bare count + spinner — deliberately *no capsule* | agent finishing |

The capsule rule is the discoverability fix: **a number in a capsule is always actionable** — clicking any violet capsule focuses the next blocked terminal (cycling), on any surface. The section header's row tally is demoted to bare quiet text so it can never impersonate a badge again.

### How it's enforced, not just fixed

- `agentPaintClass` **splits the two violet states**: `awaiting` now means *blocked on you* (full `text-alert` + glow) in every fold — the same thing it already meant in `agentBucket` — and the new `linger` carries the 55% post-turn cue. The `satisfies never` fences forced an explicit decision at every consumer (pip, tile aura, minimap buckets, dock ranking).
- New **`<AttentionTriplet>`** in `@kolu/solid-statepip` replaces `HostAwaitingPill` + `HostUnseenPill` (deleted) and the section count capsule; `sizeClass` stays the only per-surface pixel.
- The **urgency cell carries `workingIds`** (rolling-deploy-safe `.default([])`) so a background host's tab can show its in-flight count; host marks store the asking *ids*, not a count, feeding both the derived count and the capsule's jump (`attentionNav`, registered by `useAttention` — the same seam the OS-notification click router uses).
- The violet row wash is **generated from the same CSS rule as the shipped amber one** (#1990) via a `--attn` hue variable — one wash block, two hues, violet declared last so blocked beats unopened.
- `dockTree` sorts **blocked rows first within their section** (linger still ranks idle — the order≠colour law holds); sections stay pure recency, with the header triplet + tab capsule carrying cross-section discovery.
- `RecencyCell` flips to a violet **wait chip on blocked rows** (`useDuration`'s live 1s clock, compact `2m` → `20h`) — an awaiting row's age was previously *hidden* because active rows suppress their timestamp.

> The tab triplet shows on the **active** tab too — being on the host while a background terminal blocks was exactly the 20-hour failure. Tab-level amber (`unseenFinished`) still suppresses on the active host, whose dock rows carry that mark.

Design prototype the user signed off on: [One attention vocabulary](https://claude.ai/code/artifact/c865bcdf-7088-4526-b35b-b2229b92c683) (interactive, both themes).

### Evidence — live capture, real Claude session

A real `claude` session driven to each state on a dev instance. Left: blocked in a **background** terminal — host tab capsule, section triplet, violet wash + live wait chip, glowing pip. Right: finished-unopened (amber), and the blocked dock close-up.

| Blocked on you, in the background | Dock close-ups |
|---|---|
| blocked in background | asking dock unread dock |

Also verified live: clicking the tab's violet capsule focused the blocked terminal, and the moment the agent left `awaiting_user` every violet mark cleared at once — the never-dismiss clearing rule working end to end. Full unit suite green (`pnpm test:unit`, 64 packages), full typecheck, biome. e2e selectors are untouched: `data-bucket="awaiting"` kept its meaning (it was already `awaiting_user`-only) and nothing pinned the deleted pills.


srid:

**Status: `applied`** — 30 findings (lowy 16 · hickey 14), every one settled. 8 `≡ agreed` pairs, 14 uncontested solo fixes, 1 discriminating experiment, 0 left for a human.

Two Opus lenses read the full 67-file diff independently, then one reconcile-and-apply pass merged and implemented their lists.

### What they agreed on, independently

The four majors both lenses reached on their own — which is what makes them consensus rather than one reviewer's taste:

| | The finding |
|---|---|
| **The class was derived twice** | padi computes the partition and ships four id-lists; the client threw that answer away and recomputed the class from metadata plus padi's EF2 verdict pulled back out of `finishedIds`. So the host tab read derivation A while the pip and section header beneath it read derivation B — arriving on two subscriptions with independent timing. The wire-side reader that closes it, `frameClassOf`, had been written and wired to **nothing**. |
| **"Active but not asking" had no home** | Spelled four times in four dialects — set subtraction, `if`/`else` ordering, an explicit conjunction, and a fourth restatement *inside the test written to prove they cannot drift*. Now `attentionCounted`, named in the fenced vocabulary beside `attentionActive`. |
| **The count and the click disagreed** | The violet section capsule folded `allRows` plus split entries; its jump filtered *visible top-level* rows. So this PR's flagship case — an agent blocked long enough to fall out of the activity window — rendered a button reading **"1" that did nothing when clicked**. Both legs now fold and jump over the same id list, so "the number said 1 and the click did nothing" is unreachable. |
| **The law test did not fence** | The equivalence test looped a hand-written class array with no `satisfies never`, so a sixth `AttentionClass` would pass green while silently vanishing from every count. |

### The experiment

One place the lenses read as opposed: lowy filed `agentPaintClass` gaining `linger` under *not-findings* and warned that folding the partition into the paint fold would be a mode flag; hickey called the same pair a major fix.

Rather than another round of argument, the smallest test whose result differs under each claim: does `agentPaintClass(state)` equal the rename of `attentionClass(state, finished)` for every state literal **and both EF2 verdicts**? Run against the branch head both lenses read, it holds on all ten rows — and, load-bearing, it holds *independently of `finished`*: `paint` is constant down each state's pair while `attention` changes on `waiting`.

So paint needs no EF2 input, and defining it through the partition adds no mode flag. The two were pointing in opposite directions about *which way* to fold. hickey's applied; lowy's objection stands untouched against the fold it was actually about. Command and verbatim output in `.lens-debate/outcome.md`.

### Commits

| SHA | Area |
|---|---|
| `07900a7e` | the attention partition, derived once — vocab + client attention core + per-terminal readers |
| `eb866c18` | the dock counts what it can reach — dock surfaces + wash CSS |
| `a7b71c0e` | the triplet owns the unseen-suppression rule — `@kolu/solid-statepip` + host chips |
| `3e189286` | one answer for a duration we cannot trust — the `compactDelta` ladder |
| `8ede1768` | minor polish — freestanding minors, stale cross-references, Atlas |

Green after: `tsc` clean · client **1043 passed** / 115 files · padi **394 passed, 25 skipped** · `just fmt` clean · biome 1609 files, no fixes · Atlas rebuilt.

### Two residues, recorded rather than hidden

- **lowy F3's wire half is deliberately not done.** The client frame is now one map keyed by the class, built from the fenced `ATTENTION_CLASSES` — the client-side minimum lowy explicitly offered as the alternative, converting five of nine silent sites into compile errors. `PadiUrgencySchema` keeps its four positional `*Ids` fields and their rolling-deploy `.default([])`.
- **A behaviour fact named while applying hickey F6.** Collapsing the two per-host roots moved the alert engine's baseline from "first defined urgency value" to "first frame in the store" — and those differ, because the mirror creates a host's record before its cell has spoken. Taken naively that would chime for every agent already finished at bind time. `HostMarks.reported` names the silence explicitly and the engine skips a host until it is true.

Per-finding table, the experiment transcript, and both lenses' full write-ups are in `.lens-debate/`.

srid:
## Grok ⇄ Claude Code debate

✅ Consensus in 2 round(s)

|---|---|---|
| 1 | 4ee29137 | fix(debate): colour reads the same value as motion and every count |

- **F1** — Pip COLOUR still comes from terminal metadata (`paintDockRow` → `agentPaintClass(meta.agent)`), while MOTION, `asking` (wash + wait-chip), and every count come from the mirrored urgency frame (`facts.attentionOf` → `frameClassOf`)
- **F2** — Counting a printing agentless shell as active is deliberate (Q3), but the live edge is a hard 1000ms window (`TERMINAL_IDLE_AFTER_MS`)
- **F3** — `frameClassOf` walks every class list with `.includes` per id; `hostActiveIds` then does that again for every id in the class∪live union
- **F4** — Comment is stale and actively misleading: it says the client 'folds a terminal's own metadata through [attentionClass] to decide that terminal's pip'
- **F5** — `NO_MARKS` spreads `EMPTY_FRAME`, so its nested `byClass` object is the same reference as `EMPTY_FRAME.byClass`

srid:

Run with `--no-elegance` (the simplify track had already run over this tree). Pass 1 walked all 33 rules; Pass 2 fact-checked the diff's claims against its own code.

| Pass | Findings | Outcome |
|---|---|---|
| Rules | 1 | fixed |
| Fact-check | 1 | fixed |

Both landed in the compact-duration ladder, and both are fixed in `08e130bf`.

**Fact-check — `formatElapsedShort` reported stray seconds past an hour.** It computed the ladder's next-finer unit and then ignored it, appending `sec % 60` under every tier above seconds. A two-hour reconnect therefore read **`2h 47s`** — a figure cycling 0–59 every second beside an hours value, answering no question anyone has. Its sole caller is the connect/reconnect elapsed readout, which is exactly the timer an outage pushes past an hour. Hours now pair with minutes and days with hours; the minute tier keeps its live seconds, which is the reason this formatter exists.

*Not the fix the police proposed:* theirs read `sub` under every tier, and the minute tier has no `sub` on the ladder — a minutes-long connect would have stopped ticking its seconds.

**Rules (`dry-rule-of-three`) — a documented twin invariant this branch silently broke.** `vazhi/src/format.ts`'s `formatUptime` is a deliberate second implementation of this ladder (vazhi may not import the client bundle), and its comment said to keep *"the thresholds and the negative clamp"* identical. This branch changed the client's negative arm from a zero-clamp to an honest `unknown`, so the promise was false the moment it landed.

Narrowed rather than propagated, after checking which clocks are actually involved:

- the **client** subtracts a *remote host's* timestamp from the browser's — two clocks, so a negative is genuine skew and any number it rendered would be a lie;
- **vazhi**'s two numbers come from *one* clock (`createdAt` stamped by `@kolu/port-forward`'s manager in that process, `now` read by the same process), so a negative there can only be the system clock stepping backwards — where zero is the honest reading.

The thresholds stay shared; both comments now state why the negative arm doesn't. The invariant is true again instead of quietly false.

Also adds `duration.test.ts` — the ladder had no tests at all. It pins the tier walk, the sub-unit carry, the refuse-a-future-delta arm, and the hours-pair-with-minutes rule that was wrong until now.

PR #2022, opened 2026-07-27
A failure can no longer travel without its evidence

srid:
PR #2007 fixed kolu's host-down card: the failing episode's output was being collected, shipped to the browser, and dropped unread. But the fix was an app-level patch over a framework-level hole. The framework hands every consumer two separate things — the failure's *reason* and, on a different value, the *evidence* that explains it — with nothing tying them together. kolu forgot the evidence for a year. drishti forgets it today (its only failure surface is a tooltip that shows `failure.reason` and nothing else). Any future consumer starts life with the same bug.

This PR moves the fix to where the hole is: **evidence becomes a framework-minted co-product of the failure record.** After it, a failure without its evidence cannot be constructed, cannot be published, and cannot be lost in transit — for every consumer of `@kolu/surface-map`, present and future.

## The design in one move

The insight (out of #2007's review): a failed episode's output is not *live state* — it is part of the failure **fact**, pinned at the moment of give-up. It was riding the wrong value. Moving it from the live `connection` payload onto the failure record makes three properties fall out by construction rather than by policy:

1. **Minted at the one classification seam.** `serveHostMap.resolveEntry` already holds the session's retained log tail (`raw.log`) at the exact moment the injected `failureOf` classifies. It now staples `evidence: down.log` onto both down arms. No injection, no domain knowledge, no parallel pipe — the log tail is transport-owned data, and `@kolu/surface-remote` is the transport layer.

2. **Unspellable without it.** `EntryConnectionState`'s `failed` arm requires `{ failure, evidence }`; the `disconnected` arm pairs them (a standing refuse carries both; a transient drop carries neither — and failure-without-evidence is a *compile error*, pinned in `entryConnectionState.test-d.ts`). `EntryFault` — the no-session fault — requires evidence too, so a fault-minting site states `[]` explicitly rather than omitting it silently. The wire schema (`entryStatusSchema`) refuses a failed status without evidence at parse time.

3. **Survives the dead-link floor.** `floorOnLiveness` drops `connection` over a dead link (a frozen live phase lies) but keeps `failure` — so evidence riding the failure record survives *by construction*, with zero change to the floor. The "we cannot see the output" state that #2007's card had to represent (`undefined` vs `[]`) is now unrepresentable on the failed arm: a laptop that slept still shows the error that ended the host.

The `disconnected` arm is spelled as two same-tag union members rather than an intersection-of-unions — TypeScript narrows the union under a `failure === undefined` test but keeps intersection fields uncorrelated under exactly that test, and an uncorrelated `evidence: FailureEvidence | undefined` would have *invited* the `?? []` collapse this repo bans. The reasoning is recorded at the type.

## What each layer sees

| Layer | Change |
| --- | --- |
| `@kolu/surface-map` | `EvidenceLine`/`FailureEvidence` types + schema; `EntryStatus.failed` and `EntryConnectionState` carry required evidence; `EntryFault` states its evidence; `projectStatus` threads it |
| `@kolu/surface-remote` | `serveHostMap` mints `evidence: down.log` on both down arms — the only classification seam |
| kolu client | `failedEpisode` reads evidence off the failure record; its `log` is non-optional; the popover drops its "can't see it" branch. `BootStalledCanvas` is untouched — it renders the *live* tail, where the floor's caution remains correct |
| Wire | a failed `EntryStatus` without evidence fails schema validation |

Every load-bearing pin was sabotage-tested before being trusted: the producer pin (evidence === the session's retained tail, both arms), the schema-refusal pin, and the floor-survival pin were each made red by reverting the seam they cover, then restored.

## Design philosophy

- **Fail fast, no fallbacks:** there is no `?? []` anywhere in the evidence path. `[]` is minted only at the seam that *knows* the episode retained no lines. A failed status without evidence fails loud at the codec, and the existing belts (`UnclassifiedHostFailureError`, the missing-`error` throw, SR9) are untouched.
- **Volatility boundaries:** the evidence type is owned by `@kolu/surface-map` (fixed structural shape, not a new generic); the minting lives in `@kolu/surface-remote` where the transport data lives. Nothing app-flavored moved into the framework — the rendering stays per-app, deliberately (a shared failure widget was considered and declined).
- **Reuse the source of truth:** evidence *is* `SessionState.log`, the tail the session already retains — not a second collection mechanism.

## Downstream consumers

- **drishti** (required pair PR per `surface.md`): its server calls `serveHostMap` with its own `failureOf` — it starts publishing evidence with **zero code changes** the moment its kolu pin bumps, which is the "forced at the producer" property doing its job. Pair PR: **[srid/drishti#125](https://github.com/srid/drishti/pull/125)** — pinned to `c432080fa`, CI green on both platforms (20/20 lanes), `MERGEABLE`. Its adoption fixed drishti's own latent instance of this bug: `FailedCard` read the live `connection.log` (floored away over a dead link) and `ConnectingOverlay` narrowed the failed arm out — both now read the failure record. Will be re-pinned to final HEAD after the review gauntlet runs.
- **ODU-IMPACT: none** — grep-grounded twice — at the local checkout's pin (`2b70d2caf`) and independently at juspay/odu's current pin (`3ff256f0`), same result: zero hits for `EntryStatus`, `EntryConnectionState`, `EntryFault`, `entryStatusSchema`, `serveHostMap`, `MapRegistry`, `failureOf`, or any `.failure` read. odu consumes the session-pool layer (`buildRemotePool`, `resolveSystem`), not the map layer. No ledger entry.

## Testing

- New pins: schema refusal, type-level pairing (failure-without-evidence and evidence-without-failure both compile errors), producer minting on both down arms, floor survival. Sabotage-verified as above.
- Full workspace: unit lanes green across surface-map / surface-remote / common / server / client; `pnpm -r typecheck`, `just fmt`, `just lint`, `nix eval .#default.drvPath` all green.
- Reference docs (`ref-surface-map.mdx`, `ref-surface-remote.mdx`) updated in this PR per `surface-reference.md`; the #2007 changelog entry edited net-true rather than appended.

> [!NOTE]
> Draft while the review gauntlet is deliberately held at the owner's request; it runs before this merges.


srid:
Drishti pair (the `.claude/rules/surface.md` gate): https://github.com/srid/drishti/pull/125 — pinned to `c432080fa`, adopts the evidence in its failed-host card; will be re-pinned to this PR's final post-gauntlet HEAD and re-validated before merge.

srid:

**Status: `needs-human`** — 21 of 22 findings settled and committed; one is a judgment
call for you. lowy raised 9 (4 structural), hickey 13 (12 `fix`, 1 `drop`, 6 major). Both
lenses read the full diff independently.

### What settled it

Two of the load-bearing questions were decided by running something rather than arguing.

**The two comments that contradicted each other about narrowing.** `EntryConnectionState`'s
comment said `state.failure === undefined` narrows the two-same-tag-member union (that was
the whole justification for spelling it that way); `isStandingRefuse`'s doc said it does
not (that was the whole justification for the predicate). `tsc` against the shipped type:

```
error TS2322: Type 'FailureEvidence | undefined' is not assignable to type 'FailureEvidence'.
```

It does **not** narrow. The type comment was the false one, so the two-member spelling
bought nothing over the intersection it was written to beat. The false rationale *and* the
mechanism it justified are both gone.

**`samePublished`'s claim that `failure` is per-frame reference-stable.** False in kolu, and
verified in code: `padiFailureOf` mints a fresh object literal per call, `derived.registry`
does no memoisation, and the republish loop calls `statusOf` for every member on every
member's frame. So the gate never suppressed a failed member, and one re-published a wire
frame on every *sibling's* session frame — O(M²) across a pool, on the arm most likely to
sit occupied for hours. **Pre-existing** (the merge-base gate compared by reference too);
what this PR introduced was the false claim in prose. Both halves are fixed, and pinned by
a test that goes red if the fix is reverted.

### Applied

| # | Finding | Where it landed |
| - | ------- | --------------- |
| hickey F1/F2/F3 · lowy F4 | The `{failure, evidence}` pair is now one exported `FailureRecord<Failure>`, and `disconnected` carries it as one optional `refuse` field. `isStandingRefuse` deleted — the record's presence narrows on its own, so an unchecked type assertion at that seam is gone rather than belted. | `7a0b7d42d` |
| lowy F1 ≡ hickey F5 | `EntryFault.evidence` dropped. A fault has no session, so `[]` is a fact of the shape — `statusOf` states it once (from a shared `NO_EVIDENCE`), as it already supplies the fault's `kind` and `membershipId`. 8 mint sites stop restating a constant. | `7a0b7d42d` |
| lowy F2 | The republish-gate pin re-vehicled onto a session-backed `failed` entry — it previously drove a fault with a growing tail, which the fault's own contract rules out. | `7a0b7d42d` |
| lowy F3 · hickey F12 | `serveHostMap` holds each member's classification against its frame; the gate's comment now says whose job that stability is instead of assuming it. | `92f228c83` |
| hickey F4 · lowy F8 · hickey F7 | `LogEntry` **is** `EvidenceLine` (type alias + `ZodType<LogEntry>`); the pin file that watched the duplication is deleted, and `Mutually` goes back to non-exported. | `78816c41b` |
| hickey F6 · F9 | One argument, one home. `2007` went from 11 mentions across 10 files to 6 across 6; the orphaned JSDoc block above `EntryStatus` (invisible on hover) is merged. | `892baf207` |
| lowy F5 ≡ hickey F10 · lowy F6 ≡ hickey F11 · lowy F9 | `floorOnLiveness`'s catch-all made exhaustive; `wire.ts`'s lone `.otherwise()` (against 54 `.exhaustive()`) made exhaustive; `samePublished` computes the key union its comment claimed. | `7bee7f714` |

`FailureRecord<Failure>` was hickey's headline and it held. Its stated cost — "every
existing `.failure` read is unchanged" — is a backward-compat argument this repo rules out,
and it protected one read site. **The drishti pair PR's owed adoption widens**: `EntryFault`
loses `evidence`, and `disconnected` moves to `refuse`. Recorded in the commit body;
drishti was not edited.

hickey F13 (the retained tail still published under two names by arm) is recorded as an
observation, not applied — both lenses agree the live-word/post-mortem split is genuine
today.

### ⚖️ Needs your call — `CanvasFailureCard.log`'s `undefined`-vs-`[]` doctrine

Both lenses flagged the same prop and prescribed **incompatible** fixes. Shared ground,
which I verified: `BootStalledCanvas`'s `connectorLog` memo is
`props.recovery.via === "connector" ? props.recovery.log : undefined`, so it produces
`undefined` for two structurally different reasons (the `client` arm has no `log` field at
all; the `connector` arm's `log` can be `undefined` because the floor dropped the live
word), and the card renders identically for `undefined` and `[]`. The distinction is
currently not preserved at its only caller. What to *do* about that is a product decision,
so it is yours, not another agent round. Both positions verbatim:

> **lowy F7** — *"Either (a) leave the prop and cut the comment down to the one live
> sentence it now needs (`undefined` = the live-tail caller's word was floored away; `[]` =
> the episode printed nothing; both render nothing, only the caller can tell them apart),
> deleting the failed-arm paragraph that documents a state the type no longer admits at
> this prop's remaining caller; or (b) — cleaner — give `BootStalledCanvas` its own visible
> treatment for "we cannot see the output" (a one-line "output unavailable — kolu's link to
> this browser went quiet" note instead of nothing), which makes the distinction
> *observable* and therefore worth the prop's shape. (b) is the honest fix; (a) is the cheap
> one. Do not collapse to `?? []`."*

> **hickey F8** — *"make the prop total and delete the doctrine: `log: readonly LogLine[]`
> … In `BootStalledCanvas.tsx:81-84`: `props.recovery.via === "connector" ?
> props.recovery.log : []`. In `canvasModeResolver.ts:90`, tighten `log: readonly LogLine[]
> | undefined` to `readonly LogLine[]` and have `useCanvasMode.ts:81` read `const
> connectLog = info?.log ?? [];` — legitimate here, because `info === undefined` on this
> path is "the floor demoted a warming entry", which is a *warming* surface's problem, not a
> failure record's. Then delete the whole `undefined`-vs-`[]` paragraph from the card and
> the matching sentence from `website/src/content/docs/remote-hosts.mdx`."*

The card and its callers are untouched pending your answer.

### Gates

`pnpm -r typecheck` · unit tests across surface-map / surface-remote / common / server /
client (59 / 252 / 46 / 1018 / 283) · `just fmt && just lint` · `nix eval
.#default.drvPath` — all green. Two sabotage checks re-run: the new republish pin goes red
without the producer-side hold, and adding a field to `EvidenceLine` fails to compile at
`connection.ts`'s schema (the guarantee the deleted pin file used to provide).

Nothing pushed, nothing merged.

srid:
## Grok ⇄ Claude debate

✅ Consensus in 2 round(s) — base `498ff5b4a`

|---|---|---|
| 1 | 44f8b8e88 | fix(debate): two comments that taught a shape and a lifecycle we don't have |

- **F1** — EntryConnectionState's leading doc still describes `disconnected.failure` as the optional standing-refuse discriminant, but the type (line 124) and the later FailureRecord paragraph (lines 111–115) correctly name `refuse?: FailureRecord`
- **F2** — floorOnLiveness's new comment says a standing refuse publishes failed "while its session keeps redialing"

**Probes cleared** — the four claims the design rests on, verified independently by the peer:

- **Unspellable** — no path in either repo (or the examples) mints a published failure without evidence.
- **No live alias** — `pushLog` is copy-on-write, so the pinned array is never mutated under a holder. "Pinned at classification" is real, not aspirational.
- **Episode boundaries** — same-frame classification and a campaign reset clearing the log mean a reason can never pair with a previous episode's tail.
- **Republish gate** — union-of-keys `Object.is` errs toward re-emitting and cannot miss `evidence`.

srid:
## 🏛 Architecture from first principles

Ran **first**, per the gauntlet's rule for a framework/structure diff — C1–C7 as parallel hunters over the diff plus one hop down its imports, each `real` finding then handed to an adversarial refuter whose default posture is that the finding is wrong.

**17 findings → 4 real.** Two were defects in the change's own thesis.

### The fix was incomplete in exactly the way it claimed to fix

The PR's argument is *unspellable beats discouraged*. But the `failed` arm carried **both** `evidence` and `connection` — and for kolu, `connection`'s failed arm carries `log`. Both project from the same `raw` frame, so over a live link they were byte-identical; then `floorOnLiveness` drops `connection` and keeps `evidence`. So `entry.connection?.log` — the literal read that lost the tail for a year — was still compiling, still plausible-looking, sitting beside the correct one on the very arm being hardened.

Three of the seven checks reached this independently (C2, C4, C6). Fixed by dropping `connection` from the failed arm (breaking), which turns the wrong read into a compile error.

That drop then surfaced a **live bug in drishti**, unpinned by any test: `connectionOf(...) ?? DEFAULT_CONNECTION` painted *"connecting…"* in amber beside a red dot on a failed entry with the admin link down. Same family as the bug this PR exists to close, in the consumer it exists to protect.

### The republish gate silently suppressed the new field

`serveSurfaceMap`'s `samePublished` hand-enumerated the published fields; the change added `evidence` and never updated it, so an evidence-only change didn't republish. Its comment then asserted something false — *"`clockOffset`/`failure` are the only other published fields."* It hid in kolu only because `connection` usually changes in the same frame; on a map with no `connection` injected it was live.

Fixed by removing the enumeration itself rather than adding one more field to it — a shallow compare over the union of own keys, so a future added field cannot fall out of the gate. (A later pass generalized the *value* comparison too; see the lens comment.)

### Also confirmed

Both stale-doc findings were fixed: the Explanation page still claimed the framework *"carries the discriminant and nothing more"*, and a user-facing sentence promised on-screen copy that no component renders.

The refuter **cleared** the deliberate decisions rather than rubber-stamping them — including the argument for keeping the tail out of padi's wire schema, which it tested and confirmed on the merits: `connection` is work-in-flight that a dead link falsifies, while `failure` is a completed classification it does not, so the floor's asymmetry is principled rather than arbitrary.

> Three of eleven agents died on a session limit mid-run. Their findings were recovered from the workflow journal and the two consequential ones verified by hand — a dead verifier is not a pass.

srid:

Ran with `--no-elegance` (the `/simplify` track already swept this tree). Scoped to `498ff5b4a...HEAD`.

| Pass | Findings | Outcome |
|---|---|---|
| 1 — rules | 2 | fixed |
| 2 — fact-check | 2 | fixed |
| 3 — elegance | — | skipped (`--no-elegance`) |

### Pass 1 — rules

- **`dry-rule-of-three`** — the reference-stable empty-tail sentinel had been declared independently in **four** files (each review pass added its own). Moved to `ui/logTailChrome.ts` as `NO_LOG_LINES`, beside the `LogLine` type it decorates. — `41b2906b2`
- **`prefer-ts-pattern`** — `floorOnLiveness` used a hand-rolled `never`-typed runtime throw, justified by *"surface-map carries no ts-pattern dependency"*. That justification died in this same diff, which had just added `dequal` as a dependency — and `wire.ts` solves the identical problem with `match().exhaustive()`. Converted. — `4d3f19cc9`

### Pass 2 — fact-check

Both findings are the **exact sin class this PR exists to fix**: a comment asserting a mechanism the code does not implement. That makes **five** such instances caught across the whole gauntlet, by four different reviewers.

- `ref-surface-remote.mdx` described a per-frame classification cache (`failureOf` *"invoked at most once per session frame"*) that a later commit **deleted** and replaced with a structural comparison. The doc was never updated.
- `CanvasFailureCard.tsx` named `HostDiagnosticsPopover` as a caller of its `logAbsence` prop. The popover never renders that card at all.

— `928d9bced`

### Independently verified, not trusted

Both re-derived from scratch rather than accepted from the prior pass's report:

- **The `dequal` FOD hash claim** — confirmed the `pnpm-lock.yaml` diff is `importers:`-only, no `.nix` file touched, and `nix build --rebuild .#pnpmDeps` succeeds against the pinned hash. Re-verified *again* after the `ts-pattern` fix added its own lockfile entry. This is the trap that reds every linux nix lane at once if left for CI.
- **The `@kolu/surface-map/evidence` leaf** — confirmed zod-only imports, that `connection.ts` imports the leaf rather than the root (which pulls `@orpc/contract`), no bundle-isolation leak, and that the reference docs match the shipped code.

Gates green before each commit and at the end: surface-map 60 · surface-remote 252 · common 46 · server 283 · client 1023 unit tests; full-workspace `pnpm -r typecheck`; `just fmt`; `just lint`; `nix eval .#default.drvPath`.

PR #2025, opened 2026-07-28
Two interventions /be kept earning, turned into rules it already had

srid:
`/be` says it is **fully autonomous from §1 onward**. One run put that to the test and found two seams where it isn't — both cases of a rule the skill already contains failing to reach the moment that needed it. These are the two edits, mined from that run's transcript by `/self-improve`.

### An interruption was being treated as an exit

Nothing in `/be` or `/be-review` said what to do after something knocks the run out of the phase machine. So it just stopped. The clause now lands in the `# Be` preamble, right beside the sibling *"autonomy doesn't inherit"* rule: an interrupt, a mid-run bug report, a rate limit, a `/compact` **suspends** a phase — it never ends the run — and a step that was in flight when the detour hit did *not* run, so it restarts (`/be-review --tracks`) instead of being counted done.

### The repro gate was keyed to the wrong thing

§2's reproduce-before-fix protocol reads as one of three task-kind bullets, which made it a property of *the answer §0 got* rather than of *the work in hand*. A bug the human reported against the in-flight branch during §4 therefore bypassed it entirely. The new clause says the branches key to the change, not the phase — a late bug re-enters `Bug:` steps (1)–(4) from the top, red repro included, before any fix commit and before any push.

### Evidence ledger

| Edit | What happened in the run |
|---|---|
| detour → re-enter the phase | A §4 gauntlet interrupted to chase a field bug was never restarted; the run then described the diff as hickey-reviewed and had to retract it — *"I never ran hickey."* The user typed **"continue with /be" three times in 70 seconds**, asked *"Did you finish /be workflow?"*, and finally hand-installed a session Stop hook, which fired **6 times** against an unfinished §5. |
| late bug → red repro first | A field bug was answered by repainting a terminal from one sampled state value and pushed; `4df8fa7a` reverts the whole thing (241 lines) after the user checked the live process — the agent had been working the entire time. Two more *"still buggy"* rounds followed, and the user demanded repro-before-fix **three separate times**, the last as *"Can you fucking reproduce this shit in TESTS first??"* |

Both survived a read-only falsification pass: one agent per claim swept both apm source trees plus `.claude/rules/` and `.agency/`, and neither could find an existing rule that covers the gap. The nearest misses are `remote-host-testing` (reproduce-first, but hard-scoped to that harness) and the `/do` Stop guard (real auto-resume, but gated on a `.do-results.json` that only `/do` ever writes — roadmap item #1 in `llm-autonomy.mdx`, still unshipped).

> Prose only, in `agents/.apm/skills/be/SKILL.md`; the `.claude/` + `.agents/` copies and the `apm.lock.yaml` hash ride the second commit, per the commit-source-then-regen rule. `just fmt` and `just check` are green. No `be-workflow.mdx` update — no phase was added, removed, reordered, or reshaped; both clauses are internal cautions. Neither edit trips an `llm-autonomy.mdx` anti-pattern: no post-§0 question is reintroduced, the serial gauntlet stays serial, and both clauses tighten gates rather than relax them.

_Generated by [`/self-improve`](https://github.com/srid/agency) on Claude Code (model `claude-opus-5[1m]`)._


srid:
@copilot resolve the merge conflicts in this pull request

PR #2027, opened 2026-07-28
Test the mixed-version daemon upgrade window

srid:
Every existing test boots both sides of a daemon pair from one checkout, so they can never disagree. That is exactly how a recent production bug shipped: a gate-format change made new padi unable to recycle an old kaval, and Restart kaval silently did nothing. **This PR is test-only** — it builds the missing mixed-version machinery against master as it stands, so the next window change has something to fail against.

### What lands

| # | Deliverable | What it guards |
| --- | --- | --- |
| 1 | **Yesterday-daemon fixture** | Real long-lived child + owner-only gate/socket/session artifacts in a caller-specified shape — no old binary required |
| 2 | **Recycle vs foreign gate** | (a) current gate → ensure kills holder + spawns fresh; (b) foreign/garbage → `gatePid` undefined, ensure does not kill strangers, `acquirePidGate` reaps as stale |
| 3 | **Old session under new padi** | Previous-shape blob stays present on open; `backfillSavedSession` + parse restores known shapes; irrecoverable garbage throws (never silent empty) |
| 4 | **Socket-contract mismatch** | Real unix peer speaking contract `1.0` → `connectKaval` raises `DaemonContractSkewError`; endpoint reports `incompatible` with both versions |
| 5 | **Shared-artifact watchdog** | Inventory of every shared file (gate/socket/session/config/…); each non-log entry needs a covering test or a version field |
| 6 | **Previous-release e2e** | Boot the latest tag's kaval, current padi against it, `recycleKaval` → gate pid changes, session survives. Own CI recipe `upgrade-window` on **both** platforms |

### Mutate-to-prove (each test bites)

- **Fixture**: drop the live child → `isHolderLive` fails; loosen dir perms → production privacy check would refuse
- **Foreign gate**: if ensure started killing on unparsable gates, the fixture child would die; if it stopped killing on current gates, the survivor would still be live
- **Old session**: drop `backfillSavedSession` → `SavedSessionSchema.parse` rejects the pre-discriminant blob; empty-collapse would make `session.get()` return null
- **Contract skew**: remove the version check in `connectKaval` → connect resolves and the `DaemonContractSkewError` assertion fails
- **Watchdog**: delete `kaval-gate` from the inventory or claim a missing test file → red
- **Previous-release e2e**: under `KOLU_UPGRADE_WINDOW_REQUIRE=1` a missing previous binary is a hard fail (CI never silently skips)

### CI

- New required node: `ci::upgrade-window` (both platforms) — nix-builds the previous tag's `#kaval`, then runs `previousRelease.e2e.test.ts` under the daemon gate
- The unit/daemon lanes also pick up deliverables 1–5 (daemon-gated suites sit under `KOLU_DAEMON_TESTS=1`)

### Production code

None. No silent arms found that needed a fix in this pass — master already names the foreign-gate stale-reap, the session backfill path, and the `incompatible` status arm. This suite pins those named outcomes so a future window change cannot regress to silence.

> Test-only PR; no review gauntlet required.

_Generated by Grok (model `grok-4.5`) on the upgrade-window brief._


srid:
**CI e2e metrics** — `230a0af#3`

| x86_64-linux | kolu-ci-4 | 8 | 32 | 8 | 2m45s (164751ms) |
| aarch64-darwin | nix-infra@rasam.tail12b27.ts.net | 6 | 24 | 6 | 4m0s (240007ms) |

Also green: `ci::upgrade-window` on both platforms — previous ref **v2.0.0**, previous≠current kaval store paths (real mixed-version window).

PR #2039, opened 2026-07-28
just dev bakes the agent source so remote hosts work

srid:
Adding a remote host from `just dev` failed with **“This kolu run is missing its agent source”** because the dev server never set `SURFACE_AGENT_FLAKE_REF`. Only the production Nix wrapper baked it, so a working-tree run could bind local padi but could not resolve a remote one.

Both working-tree entrypoints — **`just dev`** (in its sequential body, before the parallel fork) and the standalone **`just server`** — now source **`.#agent-flake-env`**, a file `default.nix` renders from the *same* `(name, value)` set the wrapper's `--set` args are built from. Same env name (from `agent-env.json`), same proven store path with its binary-cache sidecar; no parallel provisioning path, and no way for the two bakes to drift. `_dev-parallel` deliberately does **not** bake, so `ci::dev-smoke` stays nix-free.

## Cost

Measured on `x86_64-linux`, nix 2.34.7, warm store. Every figure below is a paired A/B against a pre-PR worktree at `ec88c8193`, alternating sides, 3+ rounds; within-round spread ≤ 90 ms unless noted. Two independent agents (Codex, Grok) reproduced the headline numbers on the same host.

### Entering the devShell: zero, structurally

`nix develop` is **unaffected**, and not merely by measurement — `shell.nix` imports `nix/env.nix`, never `default.nix`, so the new derivations are not in the shell's eval graph at all. The proof is that the shell derivation is byte-identical across the diff:

```
before  /nix/store/mmy23qxlfyi9ld6gaaf3xv3h4qzj542a-kolu-shell.drv
after   /nix/store/mmy23qxlfyi9ld6gaaf3xv3h4qzj542a-kolu-shell.drv
```

| `nix develop -c true` | before | after |
| --- | --- | --- |
| warm | 0.364 s | 0.369 s |
| cold eval cache | 1.80 s | 1.86 s |
| fresh `XDG_CACHE_HOME`, 7 paired runs (Codex) | 2.03–2.06 s | 2.01–2.11 s |

Editing agent code before entering the shell changes nothing about that: a dirty tree costs `nix develop` ~1.8 s instead of ~0.37 s, but it costs **exactly the same** on master — that is nix re-reading a dirty flake, not anything this PR added.

### Starting `just dev` / `just server`

The bake is serial before the fork, so it adds directly to time-to-first-URL. The "before" column is the `nix develop` the recipe already ran; the "after" column is bake + that same `nix develop`.

| Since the last start… | before | after | delta |
| --- | --- | --- | --- |
| nothing edited (clean tree) | 0.37 s | 0.12 + 0.37 s | **+0.12 s** |
| a tracked file edited *outside* the agent fileset (`docs/`) | 1.85 s | 1.37 + 1.54 s | **+1.03 s** |
| **agent code edited** (`packages/padi/…`) | 1.83 s | 3.01 + 1.55 s | **+2.72 s** |

Where the +2.7 s goes: ~1.2 s is nix re-copying and re-evaluating the dirtied flake (any tracked edit does this), and ~1.6 s is realising the two derivations `nix build --dry-run` names — `agent-flake-source.drv` (the ~18 MB fileset copy) and `agent-flake-env.drv`. Only the second half is specific to touching agent code; that is why a `docs/` edit costs a third as much.

**Free cases.** A brand-new *untracked* file costs nothing (0.079 s — nix's git filter cannot see it, which is also why it must be `git add`ed before a dialed remote sees it). And the bake resolves **once per start**: `pnpm dev` is `node --watch`, and its restarts inherit the exported env without re-running `nix build`. Client HMR never touches this path.

### CI

From odu run `6091982#1` (61/61 green, both platforms):

| | x86_64-linux | aarch64-darwin |
| --- | --- | --- |
| new `ci::agent-bake` node | 3.5 s | 11.9 s |
| `ci::dev-smoke` (unchanged — `_dev-parallel` stays nix-free) | 18.0 s | 16.3 s |

`agent-bake` sits in the parallel fan-out while the critical path is `ci::e2e` (620 s) and `ci::daemon` (464 s), so its wall-clock contribution to a run is ≈ 0.

Deleting the ci flake's duplicate `agent-flake-source` makes `ci::odu-nix` *faster*: the whole node goes 3.74 s → 2.37 s with a cold eval cache (CI's case — `odu-nix` is the only node that evaluates `./ci`) and 1.67 s → 1.23 s warm; `nix flake check ./ci --no-build` alone goes 1.44 s → 0.99 s.

> **Correction to the [lens review comment](https://github.com/juspay/kolu/pull/2039#issuecomment-5144564019).** It put that saving at "~1.6 s per platform" of evaluation. Measured directly, it is **~1.4 s cold / ~0.44 s warm** per platform. Right ballpark for CI, but the warm figure is a third of what was claimed.

### One thing to keep true

The devShell is free *because* the bake lives in the recipes, not in `shell.nix` / `nix/env.nix`. Exporting `SURFACE_AGENT_FLAKE_REF` from the shell would look tidier and would put a ~1 s prove into every `nix develop`.

## Try it locally

```sh
just dev
# then add a remote host from the host strip
```

_Originally generated by Grok Build (model `grok-4.5`)._


srid:

**Status: applied.** Two lenses read the diff independently — **lowy** raised 9 findings (6 major fix, 1 minor fix, 2 drop), **hickey** raised 9 (5 major, 4 minor, none dropped). They converged on five findings and directly contradicted each other on three, which the author adjudicated.

Three commits:

| commit | area |
|---|---|
| `a16000d3e` | the Nix bake receptacle + the recipe relocation |
| `abd1cfe04` | ci-flake de-duplication |
| `e4b2e4116` | all minor polish |

### What both lenses independently found

**The bake pair had no single definition** (lowy F1 ≡ hickey F3). `agent-env.json` encapsulated the env var's *name*; nothing encapsulated the `(name, value)` **pair** a consumer actually has to reproduce — so the recipe re-assembled it in bash, becoming the fifth hand-spelling and the fourth distinct way of reading the registry. `default.nix` now emits the bake as sourceable data beside the wrapper's `--set` arg, built from the same two bindings:

```nix
agentFlakeEnv = pkgs.writeText "agent-flake-env" ''
  ${agentFlakeRefEnv}=${agentFlakeSrc}
'';
```

and the whole shell side collapses to one line:

```bash
set -a; . "$(nix build --no-link --print-out-paths --accept-flake-config .#agent-flake-env)"; set +a
```

Divergence between the wrapper bake and the dev bake is now unspellable, and `ci::nix`'s `flake-build .` covers the new derivation on both platforms for free — previously *nothing* in CI checked that the dev path exported anything at all.

**A renamed registry key silently reproduced this PR's own bug** (lowy F2 ≡ hickey F4). `node -p "require(…).flakeRef"` prints the string `undefined` and exits **0**, so `set -euo pipefail` sailed past it and `export "undefined=/nix/store/…"` succeeded — leaving the server unbaked and the developer staring at *"This kolu run is missing its agent source"*, the exact card this PR exists to eliminate. Subsumed entirely by the fix above: the shell no longer reads the JSON.

**`--accept-flake-config` was missing** (lowy F4 ≡ hickey F5) — on line 177, while `nix_shell` on line 176 of the same recipe passed it. For a trusted user that's an interactive `(y/N)` mid-`just dev`; in CI it's a silent cache-blind from-source build. Added at both sites the diff touched.

**`ci/flake.nix`'s `agent-flake-source` had zero remaining consumers** (lowy F6 ≡ hickey F1). Repointing `ci::agent-flake-nix` at the root flake left the `./ci` copy orphaned, defended by a "historical callers only" comment for callers a repo-wide grep says don't exist. Deleted, along with the now-dead `kolu` / `commitHash` bindings.

*Correction to what both lenses claimed here, measured afterward:* the duplicate was **not** costing a second build. `nix flake metadata ./ci` and `.` report the same revision and the same source store path, so `commitHash` was identical and the ci copy resolved to the byte-identical store path `koluBin` already pulls in — a cache hit, zero duplicated build work. The real saving is **evaluation**: forcing `agentFlakeSrc` costs ~0.8s per cold process and `ci::odu-nix` forced it twice, so ~1.6s per platform, and kolu's whole `default.nix` drops out of the ci flake's eval graph (`nix flake check ./ci --no-build` now runs in 0.92s). The deletion is still right; the stated reason was wrong.

**The bake is a start-time snapshot of the *committed* tree** (lowy F5 + hickey F7, complementary). `nix build .` resolves the git+file flakeref, so an uncommitted padi edit reaches the local server but **not** the padi a dialed remote provisions — and `node --watch` restarts inherit the old path silently. Now stated in the recipe comment and `docs/development.md`, plus one startup echo matching the existing two-URL style.

### Where the lenses contradicted each other

| Contradiction | lowy | hickey | Ruling |
|---|---|---|---|
| **`ci::dev-smoke`** | add `nix` to its deps | move the bake into `dev`'s body so `_dev-parallel` stays nix-free | **hickey** |
| **the kebab alias** | keep `agent-flake-source` | rename to `agentFlakeSrc` across 8 sites | **lowy** |
| **six prose copies** | legitimate — different audiences | revert two of them | **hickey, partially** |

**dev-smoke.** Master merged in a new `ci::dev-smoke` node underneath this branch: `devSmoke.ts` → `just _dev-parallel` → `server`, so the added `nix build` landed inside a node that declares `dev-smoke: install` and whose comment states *"Needs no `nix build`, so it does NOT depend on `nix`"* — with a cold eval racing `ci::nix` for the same drv inside devSmoke's 180s `DEV_READY_TIMEOUT_MS`. Both lenses caught it; they disagreed on the remedy. hickey's wins on the repo's own stated pattern — `justfile:37-39` already declares that dev-stack env resolution happens in `dev`'s **sequential body before the fork**, which is *reuse the existing source of truth* rather than a second env-resolution site. The bake moved there, `server` reverted to its one-line body. Mechanism verified: `dev` exports then runs `just _dev` as a child so the env inherits down to `server`; `dev-auto` execs `dev`; and `devSmoke.ts` spawns `_dev-parallel` **directly**, bypassing `dev` — so `ci::dev-smoke` stays genuinely nix-free, its declared deps stay honest, and it isn't serialized behind `ci::nix`.

**The alias.** hickey rated it minor and lowy explicitly weighed and rejected the churn. Renaming touches 8 files including three generated skill copies for zero change-axis gain, and kebab matches the flake's other CLI-facing attrs (`kaval-tui`, `padi-agent`). Kept; only the comment shrank.

**The prose.** Both reverted copies were added by the author earlier in this same run, and hickey's read is the better one: `packages/padi/README.md` is a protocol document — which dev recipes set the env is not padi's concern — and `default.nix:100`'s paragraph is about the CI gate on exposed attrs, so the added `just server` clause was off-topic there. The `ci#` → `.#` correction stayed (the recipe genuinely moved). `docs/development.md` and the `remote-host-testing` skill remain canonical; lowy's broader point that the repo deliberately restates facts for distinct audiences held for the rest.

**`hostDownCopy`** drew a fix from both lenses with different wording — lowy's contract framing won over hickey's audience-naming, because by construction a `just dev` user can no longer reach that card, so naming `just dev` there is dead advice for the only audience that can read it.

### Recorded, not applied

The two `@kolu/surface` example wrappers still take `agentFlakeRefEnv` as an argument and hand-spell `--set`, so they sit outside the new receptacle. Folding them in wants a shared `packages/surface-daemon/nix/agent-env.nix` (sibling of `binary-cache.nix`) — outside this diff's blast radius.

### Verification

`nix build .#agent-flake-env` yields `SURFACE_AGENT_FLAKE_REF=/nix/store/…-agent-flake-source` and sources correctly · `nix flake show ./ci` shows only `default`/`odu` · `nix flake check --no-build` passes · `just --dry-run dev` confirms the bake precedes the fork and `server` is nix-build-free · `just check` + `just fmt` green · the skill caveat was edited in `.apm/` and regenerated with `just ai::apm`.

srid:
## Codex ⇄ Claude debate

✅ Consensus in 4 round(s) — base `3a3cd84df`

|---|---|---|
| 1 | db21001e0 | fix(debate): bake on every working-tree entrypoint, and guard it in CI |
| 2 | 6120384de | fix(debate): make the bake guard fail on the regressions it names |
| 3 | 9ae21cab8 | fix(debate): match the whole bake line, and fail loud on a dry-run error |

- **F1** — The public `server` recipe starts the same working-tree kolu server without baking `SURFACE_AGENT_FLAKE_REF`
- **F2** — No CI path exercises the behavior this change fixes
- **F3** — The claim that `.#agent-flake-env` bakes only the committed tree is false for this local `git+file` flake
- **F4** — The comment says adding a second baked environment variable here would reach both consumers without a second edit, but the wrapper and dev file are rendered by two separate scalar strings (`agentFlakeRefBakeArg` and `agentFlakeEnv`)
- **F5** — `test-agent-bake` assumes the generated environment file has exactly one line, while the new `agentBakedEnv` attrset is explicitly designed to accept additional pairs
- **F6** — The paragraph now includes standalone `just server`, but its refresh instruction says only to re-run `just dev`

srid:

Ran with `--no-elegance` (the `/simplify` track already covered this tree). Rules pass: 0 violations across all 33 built-in + project rules. Fact-check pass: **3 findings, all fixed** — and notably, two of them were introduced by this run's own `/simplify` pass, not by the original change.

**`nix run .#kolu` does not exist.** Verified against the live flake — `nix eval .#packages.x86_64-linux` lists `default` and `koluBin`, and the flake has no `apps` output at all, so that command fails outright. It had reached two places, including the `agent-source-unbaked` card **shipped to users** — an error card whose entire job is to name a remedy was naming one that doesn't run.

Fixed by audience rather than with one string: `docs/development.md` gets `nix run .#default` (checkout-relative, which is where that reader already is), and the user-facing card gets `nix run github:juspay/kolu` — the production front door the Atlas documents, which is actionable for a reader who may have no checkout at all.

**`test-agent-bake` could fail on a correct tree.** The guard diffed `compgen -e` *names* across the bake, so a shell that already had `SURFACE_AGENT_FLAKE_REF` exported from a parent process saw no new name and the recipe reported "the bake exported nothing" — a false failure in exactly the environment a kolu developer runs it in. Reproduced live before the fix. It now snapshots `NAME=value` pairs, so a same-name/new-value overwrite is correctly seen as exported.

Re-verified after the fix: the pre-exported case passes, a clean shell passes, and all three sabotage cases still fail as intended — dropped `set -a`, bake removed from an entrypoint, and bake leaking into `_dev-parallel`.

srid:
## CI e2e timing — run `6091982#1`

Full pipeline settled green on both platforms; all 61 required checks pass.

| platform | host | workers | cores | cap | `ci::e2e` duration |
|---|---|---|---|---|---|
| `x86_64-linux` | `naiveintent` | 6 | 16 | 8 | **10m 20s** |
| `aarch64-darwin` | `ci@petit` | 4 | 10 | 6 | **4m 17s** |

Durations are the whole `ci::e2e` recipe wall time from the run ledger, not Cucumber's internal timer.

### The nodes this PR touches

| node | linux | darwin |
|---|---|---|
| `ci::agent-bake` *(new — guards the fix)* | pass (3s) | pass (11s) |
| `ci::dev-smoke` | pass (18s) | pass (16s) |
| `ci::agent-flake-nix` | pass (20s) | pass (21s) |

`ci::dev-smoke` passing on both platforms is the load-bearing one for this change's design: it enters at `_dev-parallel`, which the PR deliberately keeps free of any `nix build`, so its green is evidence the bake stayed out of the path that node depends on being nix-free.

srid:
## Evidence

| Before (`origin/master`) | After (`fix/dev-remote-agent-source` @ 609198262) |
|---|---|
| before | after |

**Before** — on a clean checkout of `origin/master`, running plain `just dev` and adding a remote host (`toor@localhost`, a real ssh target) lands on the "This kolu run is missing its agent source" card: `SURFACE_AGENT_FLAKE_REF` is unset because `just dev`'s working-tree run never baked it, so padi can't resolve a build for the remote host.

**After** — same box, same remote target, checked out to this PR's head instead. `just dev-auto` now bakes `.#agent-flake-env` before forking (visible in its own log as "→ agent source baked"), the host add goes through a real ssh provision ("Provisioning kolu on remote:toor@localhost…"), and the host strip shows `toor@localhost` **connected**. A terminal opened on that host and `echo pr2039-after-connected` round-trips real output — genuine bidirectional I/O over the real ssh/padi link, not a mock.

Both legs ran on the same ephemeral `pu` box (second box request hit cluster capacity limits, so before/after were captured sequentially on one box rather than two — nothing about the harness itself was downgraded: real `git checkout`, real `just dev-auto`, real ssh-to-self provisioning each time). Box destroyed after capture. Fallback level used: **1** (both before and after stills via the full live-drive harness).

PR #2046, opened 2026-07-29
Give every split a Dock sub-entry

### LINKED ISSUE #2024: Dock: show agents running in split terminals as indented sub-entries
An agent working inside a **split** is counted by its host tab and is visible nowhere in the dock. You find it by clicking into terminals one at a time — which is the same discovery problem [#2019](https://github.com/juspay/kolu/pull/2019) set out to end, surviving one level down.

Concretely: padi's `urgency` cell folds **every** terminal record, split children included, so a host tab can read `4` above three dock rows. The fourth agent is real, working, and has no row anywhere.

## What it should do

An **indented sub-entry** under the dock row for the terminal holding the split — not a peer row. A split is subordinate to its terminal, and flattening it would break the one-to-one correspondence between a top-level row and its `Cmd+N` shortcut.

- Only splits **running an agent** earn a line. A plain `bash` split is what the `⊟ N` chip already says, and giving every one a row would bury the signal this exists to surface.
- Clicking it lands you **in that pane**, focused — not on the parent to hunt for the tab.
- Repo section headers count them, so the header and the host tab above it agree.
- Sub-entries are not subject to the activity window: they share their parent's fate. Ranking them independently would let the dock hide an agent while still showing the terminal it runs in.

## Why it was reverted from #2019

It worked, but the **active-row highlight** went wrong four separate times in review and in the field, and each fix introduced the next fault:

1. the sub-entry hardcoded `active: false`, so clicking into a split lit nothing;
2. the fix read focus through an accessor that **seeds** state (`getSubPanel` → `ensureState`, whose default is `focusTarget: "sub"`), so merely *asking* where focus was made a terminal that just *had* a split report the split — and its parent row went dark;
3. the fix for that treated the two facts as exclusive and took `data-active` off the parent whenever focus was in its split, which the dock's oldest e2e contract (`[data-testid="dock-row"][data-active]`, keyed on the parent) denies;
4. the fix for *that* left the entry carrying `data-active` with **no CSS listening** — the highlight was a per-component Tailwind variant that two row components spelled and the sub-entry did not.

The user reported the remaining fault as "an improvement but not fully fixed", and the feature was pulled so #2019 could ship on the attention vocabulary alone.

## What to know before re-attempting

**The four faults share one shape**: a contract held by each component *remembering to spell it*. Whatever replaces this should make the row's state a single derivation every surface reads, and make the styling follow the shared `[data-dock-row]` hook rather than per-component classes — the move already used for the attention wash and the blocked-subline rule.

**The two facts nest, they are not exclusive.** The parent row is active because its *tile* is the canvas selection; the split entry is active because that is where the *keyboard* is. Both can be lit. Treating them as one fact is fault 3.

**Test at the level the bug lives.** All four faults passed a green unit suite, because the tests exercised a pure resolver while every defect sat in *what the rows render*. #2019 ended with `dockRowActive.test.tsx` — a happy-dom test that renders through the real attribute builder with the real sub-panel store and enumerates the whole state matrix (never-opened split, focus in main, focus in the split, a sibling split holding focus, collapsed panel, a different tile, no tile). **That file is the thing to bring back first**, before any implementation. It is in the reverted diff and is worth recovering rather than rewriting.

**Seeding vs reading.** `useSubPanel.getSubPanel` seeds on first touch; `peekSubPanel` (added in #2019, kept) does not. A derivation must never use the seeding one.

**The landing walk is four steps**, not one: `activate(parent)` → `expandPanel(parent)` → `setActiveSubTab(parent, sub)` → `setFocusTarget(parent, "sub")`. `activate(subId)` alone lands on nothing, because a split has no tile of its own. #2019's `useDockFocus` resolved the parent from the split's own metadata so no caller could forget it — worth keeping that shape.

## Reference

Reverted from #2019; the implementation and all four fixes are in that PR's history. `BUG.md` on the `fucknotif` branch carries the fuller post-mortem at the time of the revert.

srid:
An agent working in a split could count toward host attention while having no Dock row to click, while a plain-shell split was represented only by a count chip. **Every split now gets an indented Dock sub-entry beneath its parent**, with direct pane landing and the old chip removed.

The row follows the focus model established in FX1: focusing a split lights both its parent tile and its own sub-entry. Agent splits carry the shared StatePip, urgency order, wait age, and attention wash; plain shells remain a simple landing target. Section attention still counts split agents through the canonical attention fold, and split activity now keeps the parent visible in the Dock’s activity window.

A review change briefly turned an ordinary pending metadata read into a render-stopping invariant failure. **A base-commit control separated that regression from the local harness**, then a commit-boundary control isolated it to the review change; split rows now follow the same subscribed pending-record contract as top-level rows.

### Review trail

- architecture-first review
- Lowy + Hickey lens review and reconciliation
- three-round Claude agent debate to consensus
- simplify (substitute, supervisor-specced)
- code-police rules + fact-check, followed by a focused recheck of the regression repair
- hostile audit round 1: two minor findings settled in fix wave 1

### Testing and verification

- `just fmt`
- `just check`
- complete client unit suite: 132 files, 1,147 tests
- focused browser coverage for split sub-entry lifecycle and section/host attention agreement (2 scenarios, 18 steps)
- fresh strict two-platform CI green on the wave head at `c23a1bd#1`
- exact-head required checks green and zero open CodeQL alerts
- FX2 is marked shipped in the Atlas with #2046; FX3 is promoted to “do next” and archival is deliberately deferred to FX3 closeout

one non-chip step (palette split creation) was hardened — timeout made effective + diagnostic error — as a supervisor-ratified waiver extension.

Closes #2024.

### Try it locally

```sh
nix run github:juspay/kolu/fx2-dock-subentries
```


srid:

**Status: applied** — all eight findings are settled. Lowy raised 6 findings; Hickey raised 2.

| Origin | Finding | Location | Disposition | How it settled | Commit |
| Lowy F1 | Shared row leaks touch policy | `SubTerminalRow.tsx:22`; `DockList.tsx:162` | fix · major | applied | `27a7e4bc` |
| Lowy F2 | Agent prefilter bypasses attention socket | `useSectionAttention.ts:47` | fix · major | human/drop — authoritative product rule keeps agent-only membership; H1 removes the duplicated derivation | — |
| Lowy F3 | Split jump compares tile identity | `Dock.tsx:519` | fix · major | applied | `27a7e4bc` |
| Lowy F4 | Section tally ignores new entries | `Dock.tsx:565`; `DockList.tsx:122` | fix · minor | ≡ agreed with H2; applied | `27a7e4bc` |
| Lowy F5 | Rail drops the split strategy | `Dock.tsx:395` | fix · major | applied | `27a7e4bc` |
| Lowy F6 | RowPips is a topic bundle | `RowPips.tsx:1` | fix · minor | applied | `9f33845b` |
| Hickey H1 | Split agent role is reconstructed | `dockRowRanking.ts:216` | fix · major | applied | `27a7e4bc` |
| Hickey H2 | Dock row cardinality has two meanings | `dockTree.ts:53` | fix · major | ≡ agreed with F4; applied | `27a7e4bc` |

The resulting model distinguishes top-level rows, top-level numeric-shortcut order, expanded rail entries, and rendered-entry counts. Split role is derived once as `agent` or `shell`; touch interaction policy is explicit; attention jumps use the focused terminal; and every split now has a rail landing without claiming a numeric shortcut.

Verification: `just fmt`; `just check`; focused Dock unit suite (8 files, 94 tests passed).

srid:
## Claude ⇄ Codex debate

**✅ Consensus in 3 rounds** — Claude reviewed the full FX2 diff from the pinned merge base; Codex applied every accepted finding and returned each wave to the same live peer for verification.

| Finding | Severity | Resolution | Commit |
| --- | --- | --- | --- |
| F1 · A fresh split could not keep its stale parent landing visible | major | Parent window/recency now includes the newest split activity | `07b77dfb` |
| F2 · A blocked split did not promote its parent within the repo | minor | Blocked-first rank includes child urgency | `07b77dfb` |
| F3 · E2E did not prove an exited split entry disappears | minor | Scenario now shrinks two entries to one after exit | `07b77dfb` |
| F4 · Rail invented a `?` outside the canonical glyph vocabulary | nit | Missing intent glyph now renders the split hook alone | `07b77dfb` |
| F5 · Aggregate split activity shortened a parent's WAIT age | minor | WAIT uses the row's own agent timestamp; ordinary age/window stay aggregate | `3c02b2c4` |

Round 2 verified F1–F4 and found the narrow F5 display regression. Round 3 verified all five findings resolved with no new regressions and approved the change.

Verification: `just fmt`; `just check`; focused Dock tests; targeted split e2e.

srid:
## Code police

| Pass | Issues found | Result |
| --- | ---: | --- |
| Rules | 5 | fixed in `5938df8b` |
| Fact-check | 1 | fixed in `5938df8b` |

**All clear after fixes.**

- `dry-rule-of-three`: removed the stored `visibleEntryCount`; section count is now the already-canonical `railEntries.length`.
- `invalid-states-unrepresentable`: shell rows cannot carry agent pip chrome, and agent/shell order buckets are distinct typed sets with fail-fast fences.
- `styling-tailwind-only`: moved the split rail chip's expressible geometry and shell treatment out of custom CSS and into Tailwind utilities.
- `prefer-ts-pattern`: both rail-entry and split-kind dispatches are exhaustive.
- `no-vacuous-assertion` / fact-check: removed the shortcut test that never observed the shortcut projection, strengthened the browser count to the expected detected-agent value, and added a pure membership test. The new test was mutation-checked: inverting agent/shell membership fails it.

One fact-check proposal to include live plain shells in section attention was not applied because it conflicts with the supervisor-set product contract: plain splits have a landing only; section attention includes split agents. The new mutation-proven test pins that boundary explicitly.

Verification before the supervisor-directed control: `just fmt`; `just check`; all Dock unit tests (10 files, 98 tests).

The targeted browser scenario then exposed a real review regression rather than a harness fault. A commit-boundary control passed at `7599ea3` and failed first at `5858996a`: split IDs and their projected metadata are independent reactive reads, but the review had changed the existing pending-record contract into a throw. `e5a81420` made split rows match top-level rows — read (and therefore subscribe to) the optional slot, omit the unpaintable row for that frame, and include it on recomputation. Reintroducing the throw makes the browser scenario fail before the first split mounts; restoring the pending branch makes three consecutive runs pass. The focused unit test and both affected scenarios (`sub-terminal.feature:144` and `:167`) pass on final HEAD.

A focused post-fix police recheck found no correctness issue. Its rules axis questioned an over-specific comment about projection timing; the comment was narrowed to the actual contract (`getMetadata` is optional, both row levels subscribe by reading it), and the browser mutation above supplied the missing production-path proof.

srid:
### CI e2e metrics · `c23a1bd#1`

| Platform | Host | Workers | Cores | Cap | E2E duration |
| `x86_64-linux` | `kolu-ci-3` | 8 | 32 | 8 | 165.091 s |
| `aarch64-darwin` | `nix-infra@rasam.tail12b27.ts.net` | 6 | 24 | 6 | 299.087 s |

srid:
## Evidence

Plain Bash split — the child is indented directly beneath its parent; the old count chip is absent.



Focused split — parent and child both remain active while typing in the child.



Blocked split agent — violet awaiting-user glyph on the indented child.

srid:
## Hostile audit trail (FX2)

Independent hostile review by a zero-stake auditor (Grok 4.5) against the amended plan of record ([dock-split-focus](https://github.com/juspay/kolu/blob/fx2-dock-subentries/docs/atlas/src/content/atlas/dock-split-focus.mdx) § FX2 + supervisor rulings), seeded with the FX1 campaign's cheat ledger. Two rounds to dry.

| Round | Verdict | Findings → resolution |
| --- | --- | --- |
| 1 | 2 MINOR, no MAJOR/BLOCKER | A non-chip e2e step altered outside the chip waiver (it *strengthened* the harness — Playwright options were in the wrong slot so the timeout never applied — ratified as a documented waiver extension rather than reverted); the plan's "washes violet" prose contradicted the deliberate `:not([data-active])` wash suppression (fixed in the note, not the code). Deep checks came back clean: "tolerate pending split projections" judged a modeled pending state matching the top-level row contract (the mid-gauntlet `throw` was the real defect); all three evidence PNGs visually verified against their captions; agentless rows emit no pip/wash markup; splits share their parent's ranking fate; header counts fold from the shared attention vocabulary. |
| 2 | **CLEAN** | Wave-1 delta confirmed docs-only (0 bytes of production/test change); note now matches CSS reality; ratification sentence present; no ledger cheat re-entered. |

Every wave settled a strict two-platform odu run with all protect contexts and zero PR-scoped CodeQL alerts on its exact HEAD.

_Coordinated by Claude Code (model `claude-fable-5`); implementer Codex (`gpt-5.6-sol`); auditor Grok (`grok-4.5`)._

PR #2047, opened 2026-07-29
Give kaval a frozen identity channel

srid:
**Kaval now reports one honest build-and-boot identity through the frozen daemon control channel**, while every historic PTY-host route and the frozen `drain(): void` wire remain unchanged. Padi can identify a current Kaval before relying on its versioned surface, and can still connect to a pre-fragment survivor without inventing a build identity.

The daemon serves the additive control sibling beside its established flat surface:

```text
surface.system.*              existing PTY-host surface, unchanged
surface.terminal.*
surface.control.core.hello    frozen build + boot identity
surface.control.core.drain    typed refusal for Kaval
```

Kaval is structurally not drainable because exiting it would destroy its live PTYs. Its direct drain handler therefore rejects with typed `PRECONDITION_FAILED`; a process-boundary pin proves the daemon and its identity route remain alive after that refusal. The shared terminal-runtime supervisor now also attempts both closes in terminal-first order and aggregates dual failures. Drishti does not consume that shared close helper (nor `reServeSurface` or `extendSurface`); its surface path remains `pumpRemoteSurface`.

### The upgrade window

A current Kaval freezes one boot record and projects it through both `control.core.hello` and the legacy `system.version`; padi rejects mismatched contract versions, boot instants, and partial frozen identities instead of splicing contradictory observations together. Every Padi identity consumer, including the host-inventory row, now reads build identity through the frozen fragment. A connected peer's timeout or invalid response fails loudly; only an honestly absent listener becomes an all-null inventory probe.

A live pre-fragment Kaval is the one transition exception. Padi reads only its actual legacy contract version and `startedAt`, projects the established empty-string honest-unknown build identity, keeps the daemon connected and restartable, and emits the update nudge—even when the compatible contract version matches. It never copies legacy `system.version.identity`; a tempting non-empty legacy fixture pins both the connection and inventory projections. A framework-level explicit pre-fragment observation arm may be worthwhile for a second consumer, but is deliberately outside UW5 rather than widening the shared protocol for one transition.

### Identity integrity

`readBakedIdentity` enforces the joint build fact: both baked variables are non-empty or both are absent. Half-pairs and explicitly empty baked variables crash at daemon boot. `readControlCoreHello` enforces the corresponding observation fact: both fields absent, both empty, or both non-empty; a partial pair is rejected before convergence can decide. The Nix wrappers stamp both Kaval and Padi with their source commit, reject missing/non-Git commit inputs, and keep test-only logger fixtures outside Padi's behavioral hash.

Commit `89a8cc35` temporarily moved the frozen-pair guard from the shared reader into Kaval's connector to preserve a synthetic Padi test with `commit` present and `buildId` absent. Git history shows the frozen fragment introduced those fields together; that one-sided shape was never a legitimate deployed generation. This wave restores the guard at the identity reader and encodes the real pre-pair survivor as both fields absent.

Drishti consumes this shared reader and previously shipped a build-only agent wrapper. **Its joint outer-wrapper stamp landed in PR 128 at [`27d10fe`](https://github.com/srid/drishti/commit/27d10fe), ahead of any future re-pin onto a master containing UW5**, so there is no ordering hazard.

### Proof

- 26 production mutations were executed red and restored. The mutation table and captured failure log live with the UW5 review session (`uw5-wave1-mutations.md` and `/tmp/uw5-w1-mutations.log`), not in the product tree. Their quoted excerpts cover drain refusal and survival, structured pre-fragment handling, connected/restartable/nudging presentation, close-both behavior, baked and observed identity-pair integrity, dual-boot rejection, source-closure equality, error-level anomaly reporting, inventory identity provenance, and both catch-to-empty escape hatches.
- `just fmt` and `just check` pass on the pushed wave head; the full restored Padi, surface-daemon-supervisor, surface-daemon, and server unit suites pass.
- The serial `/be` gauntlet completed architecture-first, Lowy/Hickey lens, Claude debate, simplification, and independent code-police/fact-check passes. Grok's zero-stake hostility review found the inventory and pin gaps; wave 1 closes them with executed-red mutations.

> Visual evidence is intentionally skipped: UW5 changes daemon identity and supervision semantics with no rendered UI delta.
>
> **Odu impact verdict: `none`.** Odu neither imports nor supervises Kaval, and it consumes none of the shared API moves in this PR: `readBakedIdentity`'s stricter throw semantics, `superviseTerminalSource`'s close-both contract, or the `loggerStubs` test-fixture relocation. There is no break-at-bump or adoption opportunity for Odu.

### Try it locally

```sh
nix run github:juspay/kolu/uw5#kaval
```


srid:

Status: **applied**. Lowy found no boundary defects; Hickey raised three simplifications, all committed locally and verified.

| Origin | Finding | Outcome | Commit |
| --- | --- | --- | --- |
| Hickey | Carry daemon identity as one captured boot fact | Applied | abe7fa753 |
| Hickey | Reuse one supervisor-owned frozen-hello reader | Applied | abe7fa753 |
| Hickey | Centralize the logger type and test fixtures at `@kolu/log` | Applied | 30a62cafa |

No contradiction or human-decision valve remained.

srid:
## Claude ⇄ Codex debate

Consensus reached in 2 rounds against merge base `f33e8a2d`.

| --- | --- | --- |
| 1 | 024279ee | Closed the raw-Nix, binder-forwarding, unknown-projection, and pre-fragment timeout escape hatches. |
| 2 | — | Claude independently verified every fix and approved the full diff. |

- **F1** — Reject an empty commit at the shared daemon-identity Nix bake boundary.
- **F2** — Forward Kaval's build identity from the binder as one validated pair.
- **F3** — Use one honest-unknown Kaval identity projection.
- **F4** — Bind the pre-fragment `system.version` timeout with an executed test.

srid:
## Code-police + fact-check

Status: **applied and CI-corrected**. Two independent read-only passes found five concrete gaps on the post-debate head; each surviving fix has an executed-red mutation and a restored green test.

| Finding | Resolution |
| --- | --- |
| Test-only logger fixture leaked into Padi's baked source hash | Kept the public test utility export while renaming the physical file to the existing `.testlib.ts` exclusion convention. |
| Explicitly empty baked env values collapsed to off-Nix | Only two absent variables produce honest unknown; empty or partial baked values crash. |
| Partial Kaval identity collapsed to unknown | Kaval's connector rejects a partial pair. The first CI run caught that enforcing this in the generic frozen reader broke older Padi fragments whose additive `buildId` field is legitimately absent; the invariant is now correctly local to Kaval. |
| Current Kaval's two surfaces could name different boots | Padi rejects unequal `startedAt` values before composing the connection. |
| A connected-without-identity propagation anomaly logged at warn | Promoted to error and pinned through the observable logger call. |

`just fmt`, Atlas sync/idempotency, focused restored suites (including the legacy-Padi regression), all affected package typechecks, and repository-wide `just check` pass on `89a8cc35`.

srid:
UW5 hostility wave 1 landed at `eea712004fb6fe2835289b6fc1718e2dd3e46947`.

- Host inventory now reads Kaval build commit only from frozen `control.core.hello`; the pre-fragment tempting-legacy fixture makes the inventory and connected/dialog paths agree on honest unknown.
- The connect pin now fails if legacy `system.version.identity` is copied.
- `readControlCoreHello` again rejects every missing/empty half-pair. The temporary `89a8cc35` removal preserved a synthetic one-sided Padi test; history shows the fields entered the frozen fragment together, so the real legacy arm is both absent.
- Catch-to-empty survives only for an honestly absent listener; connected-protocol and injected scan failures surface.
- Drishti stamp corrected to `27d10fe`; Odu verdict recorded as exact `none` over all shared API moves.

Five wave mutations were executed red and restored. The PR now carries the complete 26-row [table](https://github.com/juspay/kolu/blob/eea712004fb6fe2835289b6fc1718e2dd3e46947/evidence/uw5/mutations.md) and [failure log](https://github.com/juspay/kolu/blob/eea712004fb6fe2835289b6fc1718e2dd3e46947/evidence/uw5/mutations.log).

Verification: `just fmt`, `just check`, website typecheck, and full restored Padi/surface-daemon-supervisor/surface-daemon/server unit suites green. Pinned Linux odu run `eea7120#1` passed all 23 nodes on `kolu-ci-2`. Darwin was not dispatched, per ruling while rasam is held.

PR #2049, opened 2026-07-29
fix: Code-tab file tree no longer freezes after host switch

srid:
Switching the active host could leave the Code tab's **file tree stuck on the previous host's files** (preview pane still correct) with:

> File tree render failed: Cannot remove a non-empty directory without recursive: ".claude/"

This PR first pinned the defect with red tests, then closes it in three independent layers — each is a defect on its own.

### Root cause

1. **Mixed inventory was representable.** Browse merges tracked files (`fs.listAll`) with the gitignored overlay (`fs.listIgnored`, which collapses a fully-ignored directory to one trailing-slash entry like `.claude/`). Separate retained/polled queries (`hostCodeTab.ts`, `createPolledQuery.ts`) can be mutually stale for a tick on host-switch resume. Rule 2 in `browseInventory.ts` dropped only **exact-string** overlaps, so an inventory could hold both `.claude/` **and** tracked files under it — a state a single consistent `git ls-files` snapshot never produces. Pierre's `.claude/` node then has children.

2. **The throwing op.** When the stale overlay entry later dropped, `pathDiffOperations` treated every inventory entry as a file and emitted `{type: "remove", path: ".claude/"}` without `recursive: true`. `@pierre/trees` throws exactly the toast message on non-recursive remove of a non-empty directory.

3. **The permanent freeze.** In `FileTree.tsx`'s paths-reconcile effect, `safeApply` caught the throw and toasted, but `appliedPaths = paths` never ran, and Pierre's `batch` has no rollback. Every later paths change re-diffed against stale bookkeeping, re-emitted the doomed remove, and froze the tree forever.

### Fix (three layers)

| Layer | Change | Closes |
| --- | --- | --- |
| Merge | Drop overlay directory keys that still have tracked children under their prefix (`isDirectoryPath` + prefix check) | Mixed inventory unrepresentable |
| pathDiff | Directory-key removes are **skipped** when `next` still has descendants under that prefix (recursive would wipe them); **recursive** when the dir is fully orphaned | Pierre throw on dir remove |
| FileTree | On batch failure: `resetPaths` to the desired inventory, advance `appliedPaths`, still surface the original error via `onError` | Permanent freeze if something still throws |

**Design philosophy:** fail loud (`onError` still fires) but never wedge permanently; reuse `isDirectoryPath` as the sole directory-key predicate; no override knobs or fallbacks that hide the defect.

### Tests

- Repro suite (desired behaviour) now green: mixed→settled apply, freeze bookkeeping, merge drops overlay dir under tracked children.
- Op-shape pins: skip when descendants remain; recursive remove when the collapsed dir is orphaned.

### Evidence

Unit-test green is the primary proof for this bug (the host-switch race is hard to script as e2e). Live multi-host capture deferred unless supervisor requires it.

_Generated by [`/be`](https://github.com/srid/agency) on Grok (model `grok-4.5`) under Claude Code supervisor; debate peer `claude`._


srid:

**Status: applied** — lowy clean; three hickey fixes committed as `c0d561e18`.

| Lens | Findings |
| --- | --- |
| lowy | 0 fix |
| hickey | 3 fix (1 minor, 2 major) |

### Applied

- **Shared `hasPathUnderPrefix`** — one descendant-under-dir predicate for merge + pathDiff.
- **`dropRedundantDirKeys` before pathDiff** — `appliedPaths` is reconciled desired inventory; pathDiff always recursive-removes orphaned directory keys (no skip branch).
- **`onError` only when `resetPaths` recovery fails** — recovered tree no longer toasts “File tree render failed”.

### Lowy note

Three layers map to three independent volatilities (browse authority · Pierre remove ops · bookkeeping recovery); no new electricity; consumer not forced to expose wires.

srid:
## Claude ⇄ Grok debate

✅ Consensus in 4 round(s)

|---|---|---|
| 1 | f7ff13c55 | fix(debate): reverse dir-key transition, log recovery, seed nextDirs |
| 2 | 2c4c57e61 | fix(debate): capture FileTree handle for pathOps filter narrowing |
| 3 | 45216e097 | fix(debate): restore appliedPaths comment clause about on prevInput |

**Legend** — findings Claude raised:

- **F1** — The files->collapsed-dir-key transition still throws
- **F2** — directoryRemovalOps never counts a bare directory key in `next` as a surviving directory: ancestorDirectoryPaths(".claude/") returns [] and the key itself is never added to nextDirs, so for prev [".claude/a.md"] -> next [".claude/"] it emits `{remove, ".claude/", recursive}` — pruning a row the desired inventory explicitly lists (verified by direct call)
- **F3** — The bare `catch {` discards the original batch error with zero trace on successful recovery — no log, no debug channel, nothing
- **F4** — Stale comment contradicts the code: the effect header says on a throw we "Recover by rebuilding via resetPaths ..., then still surface the original error through onError (fail loud, never wedge)" — but the implementation deliberately does NOT call onError when recovery succeeds (the inner comment at line 440 says the opposite, correctly)
- **F5** — dropRedundantDirKeys is O(dirKeys x n): every trailing-slash entry rescans the entire inventory via hasPathUnderPrefix
- **F6** — Recovery resets with raw `paths` rather than the normalized `next`, re-creating in Pierre exactly the mixed dir-key-plus-children inventory the rest of the change declares unrepresentable (Pierre tolerates it at construction, so this is consistency, not a crash)
- **F7** — Round-2 regression: `npx tsc --noEmit` in packages/solid-pierre fails with TS18048 ('tree' is possibly 'undefined')
- **F8** — Doc drift introduced this round: the `appliedPaths` declaration comment still says it holds "host `paths`" after a successful apply or recover, but since F6's fix it stores the *normalized* inventory (`next`, dir keys deduped) — exactly the 'single meaning' dropRedundantDirKeys' doc promises, so the comment should claim it.
- **F9** — New regression from the F8 rewrite: the comment's closing lines are a dangling fragment

srid:

Post-gauntlet pass after lens + Claude debate:

- **No elegance re-run** (simplify already normalized mount bookkeeping).
- Fail-loud: recovery logs original throw; `onError` only on unrecovered failure (no lying toast).
- Reuse: `isDirectoryPath` / `hasPathUnderPrefix` / `dropRedundantDirKeys` single sources.
- Tests cover forward mixed drop, reverse files→collapsed-dir, orphaned recursive remove, merge overlay under tracked.
- No skip/todo on repro tests.

srid:
## Evidence

**Unit repro flipped green** (the bug was pinned by failing unit tests; host-switch live multi-host capture deferred as expensive — supervisor may request it).

### Solid-pierre (`pathReconcile` + real Pierre tree)

```
✓ pathReconcile.test.ts (18)
✓ FileTree.gesture.test.tsx (5)
✓ all @kolu/solid-pierre unit — 30/30
```

Covers: mixed inventory normalize + apply, reverse files→collapsed-dir without throw, orphaned recursive dir remove, freeze bookkeeping advance.

### Client merge

```
✓ browseInventory.test.ts (10/10)
```

Including: drops overlay directory when tracked files exist under it.

Live host-switch freeze was the original production symptom; the causal chain is covered by the unit layers above (merge · pathDiff · FileTree recovery).

srid:
## CI e2e metrics

Run identity: `20a82ed#5` (linux) + `20a82ed#8` (darwin)

| platform | host | workers line | e2e duration |
| --- | --- | --- | --- |
| x86_64-linux | kolu-ci-4 | `e2e: workers=8 (cores=32 cap=8)` | 199s |
| aarch64-darwin | rasam | `e2e: workers=6 (cores=24 cap=6)` | 364s |

Full pipeline green on both platforms for HEAD `20a82edfa`.

PR #2051, opened 2026-07-29
fix: MCP-created terminals now honour the Inherit/Shuffle theme preference

### LINKED ISSUE #2045: MCP-created terminals bypass the Inherit terminal theme setting
## Summary

When **New terminal theme** is set to **Inherit**, terminals created through Kolu's MCP tool do not inherit the theme used by existing terminals.

This happens even when an open terminal already has the desired theme. It also happens after closing the existing terminals: the next MCP-created terminal does not remember the previously used theme.

## Steps to reproduce

1. Open Kolu.
2. In Settings, set:
   - **New terminal theme:** Inherit
   - **Shuffle behaviour:** Dark
3. Change an existing terminal to a recognizable dark theme.
4. Create another terminal through the normal Kolu UI.
5. Observe that it copies the active terminal's theme.
6. Create a terminal through the MCP `lifecycle_create` tool.
7. Observe that the MCP-created terminal uses the default theme instead of the existing terminal's theme.
8. Optionally close all terminals and create another through MCP. It again uses the default rather than the most recently used theme.

## Expected behavior

MCP-created terminals should honor the same new-terminal theme preference as terminals created from the Kolu UI.

With **Inherit** selected:

- If a terminal is open, the new terminal should copy the relevant active or most recently active terminal's theme.
- If no terminal is open, it would be useful for the next terminal to inherit the most recently used terminal theme rather than resetting to the built-in default.

At minimum, terminal creation through MCP and through the UI should behave consistently.

## Actual behavior

The UI creation path applies the **Inherit** preference, but the MCP creation path does not. MCP-created terminals arrive without the inherited theme and therefore render using Kolu's built-in default terminal theme.

## Additional context for investigation

The documented behavior says that **Inherit** copies the active terminal's theme. **Shuffle behaviour: Dark** technically applies only when a theme is shuffled, so it may not directly affect this bug; I included it because that is the configuration under which I encountered the problem.

The relevant paths appear to be:

- `packages/client/src/terminal/useTerminalCrud.ts` — applies the user's `newTerminalTheme` preference and copies the active terminal's `themeName` before UI-created terminals are sent to `lifecycle.create`.
- `packages/kolu-mcp/src/expose.ts` — exposes `lifecycle.create` directly from the padi surface.
- `packages/padi/src/servePadi.ts` — creates the terminal using only the `themeName` supplied in the request.
- `packages/client/src/useThemeManager.ts` — falls back to the built-in `Tomorrow Night` theme when terminal metadata has no `themeName`.
- `website/src/content/docs/theming.mdx` — documents Inherit as copying the active terminal's theme.

This suggests the preference is currently interpreted only by the browser's UI creation path. An MCP caller goes directly to padi, where there is no equivalent preference/inheritance decision.

## Acceptance criteria

- A terminal created with `lifecycle_create` honors **New terminal theme: Inherit**.
- UI-created and MCP-created terminals use consistent theme-selection behavior.
- The behavior is covered for:
  - an existing open terminal with an explicit theme;
  - no currently open terminal but a previously used theme, if remembering that theme is intended;
  - **Shuffle** mode, to ensure its existing behavior remains unchanged.
- The theming and MCP documentation clearly describes any intentional difference between UI and out-of-band terminal creation.

## Open product question

Should **Inherit** persist the last-used terminal theme after every terminal has been closed? Current documentation promises inheritance from the active terminal, but not from terminal history.

srid:
## Problem

Open a terminal from the keyboard and it follows your **New terminal theme** setting — Inherit copies the terminal you were last in, Shuffle picks a background distinct from the ones already on screen. Ask the MCP server for one and it arrives on the built-in default scheme, ignoring the setting entirely. Same for any other out-of-band caller: a TUI, a script.

The reason is where the decision lived. `useTerminalCrud` picked the theme **in the browser** and passed it along with the create call — so the preference was only ever applied on the one path a browser create takes. Every other caller went straight to the daemon and got the default.

Closes #2045.

## The fix

Move the decision to the one place every caller passes through: padi's `createTerminal`, the constructor whose own docstring calls it "the ORDINARY constructor — the wire `lifecycle.create` **and every in-process caller**". The browser stops pre-computing a theme and only pins one when it genuinely has an override to pin (session restore, a worktree create).

That immediately raises the question this PR spent most of its effort on: **how does padi know what the preference is?** It's kolu-server's to own, and padi mustn't reach into it.

It doesn't. It gets **told**. A new `chrome.setNewTerminalPolicy` procedure — the twin of the existing `chrome.setActive` — carries the preference from the app chrome to every member host's padi, re-sent whenever the setting changes or a host's link comes back up.

The first cut of this PR did it the other way: padi read `$KOLU_STATE_DIR/config.json` off disk. That works locally and **silently breaks every remote host** — the remote arm carries state via `--state-root` argv, never env (`remotePadiBinding.ts`), so a remote padi has no such file and quietly fell back to the default. That's a regression against the *old* browser-side behaviour, which got remote hosts right. Reporting over padi's own surface is the only shape that behaves identically on both arms.

The report carries a **resolved** `isDark`, not the raw `colorScheme`. Only the browser can answer `"system"` against a media query; padi runs headless. Sending the raw preference meant padi guessing, and it guessed *dark* — so a System + light-desktop user got dark-family shuffles for every new terminal. Each guarantee now sits at the endpoint that can actually make it: the browser resolves the media query, padi resolves the theme.

When no chrome has reported yet, padi has **no opinion** — `undefined`, not a default. A padi-side default would be a second copy of `DEFAULT_PREFERENCES` free to drift from the real one.

## Changes

| Area | What |
| --- | --- |
| `packages/padi/src/newTerminalPolicy.ts` | The resolver: override › absent-policy › inherit › shuffle. Pure core, registry reads passed in lazily. |
| `packages/padi/src/chromeReports.ts` | The two client-reported chrome facts (active terminal, new-terminal policy) as a leaf module. |
| `packages/padi/src/terminals.ts` | `createTerminal` applies the policy — the one door. `restoreSpawn` stays unpoliced by type. |
| `packages/padi/src/surface.ts` | `chrome.setNewTerminalPolicy`; contract 4.6 → **4.7** (additive minor). |
| `packages/client/src/hostScope/useNewTerminalPolicyReport.ts` | Reports the resolved policy to every member host; re-seeds on reconnect. |
| `packages/client/src/terminal/useTerminalCrud.ts` | Client-side theme picking deleted. |
| `packages/terminal-vocab/src/schema.ts` | `NewTerminalPolicySchema` + `shuffleMode` — the seal-legal shared home both sides reach. |
| `default.nix` | `terminal-themes` joins padi's hashed build closure through its `dependencies` edge (#2095 derives the set from the graph). `memberIdentityFileset` now also hashes each member's root-level JSON, so `themes.json` — the candidate pool, which lives outside `src/` — cannot slip `PADI_BUILD_ID`. |
| `website/`, `packages/*/README.md` | Docs + changelog. |

### Design philosophy

- **Fail fast, no fallbacks** — the disk read took the whole zod-`.catch()` + try/catch fallback stack with it when it went. There is no file to fail to read, and "nobody has reported" is an absent fact, not a defaulted one. A failed report is **toasted**, not swallowed into a console line.
- **Reuse the source of truth** — the preference vocabulary has one declaration in `@kolu/terminal-vocab` (the `DEFAULT_SCROLLBACK` precedent), re-exported to `kolu-common`. An earlier revision re-declared it, on the false premise that padi's package seal forced a copy.
- **Volatility boundaries** — the policy is a bounded decision, so it's a padi-internal leaf, not a new `@kolu/*` package.

## Review

Four passes: architecture-first-principles (C1–C7), a lowy ∥ hickey lens review, a Grok debate, and a simplify pass — see the PR comments. Between them they caught the remote-host regression, the `system`→dark smush, a connect-window bug where the client reported floored defaults before the preferences cell had yielded, splits silently starting to shuffle, and a test-isolation leak.

Two findings are **deferred by explicit decision**, not skipped:
- Unifying the client-side <kbd>⌘⇧J</kbd> "shuffle this terminal" action server-side — it would change a second, currently-working path.
- Having kolu-server rather than the browser push the policy — its only added coverage is a host you've never opened kolu against since its daemon started, and it would cost a persisted-preference schema migration. The theming doc states that one gap plainly rather than implying it away.


srid:

Two independent Opus lenses read this diff against `6dd135a1` — **lowy**
(volatility-based decomposition: does the boundary encapsulate the thing that
changes?) and **hickey** (structural simplicity: what is braided that should be
separate, and what is one thing split across places?). **27 findings** came back
— **11 from lowy, 16 from hickey** — and were merged by one reconcile-and-apply
pass.

**Status: applied.** 22 findings landed across four commits; 1 is a recorded
observation; 2 were judgment calls the author deferred; 2 are direct
contradictions between the lenses and are left open, spelled out below.
`just check` is green, the unit lanes for all five touched packages are green,
and `nix eval .#padi.drvPath` evaluates with the widened fileset.

### The two defects that mattered

**The reporter told padi a policy the user never chose.** `preferences()` floors
an unloaded cell to `DEFAULT_PREFERENCES`, so the effect fired the instant the
link reached `connected` — before the preferences cell had landed — and reported
a confident `shuffle` / `auto`. padi could not tell that from the real setting.
That is exactly #2045's failure class (a terminal ignoring the real setting)
re-introduced at startup, for the one path that never had it. The report now
gates on the *unfloored* cell, so an unloaded preference is an absent fact on
both sides of the hop and padi's "nobody has reported" branch stays reachable.
Pinned by a test that nothing is sent before the cell lands, and one is after.
(hickey F2)

**Splits silently started shuffling.** `handleCreateSubTerminal` calls
`.create({ cwd, parentId })` with no theme, so the new policy applied to
sub-terminals — and since `shuffle` is the *default* preference, every split
began getting a random tint it never had, inside its parent's tile. The policy
now governs top-level creates only; a split keeps the pre-#2045 behaviour, pinned
through the real door in `servePadi.test.ts`. (hickey F4)

### Applied

| Area | Findings | Commit |
|---|---|---|
| **padi's new-terminal policy decision** — the module owns its own inputs; one source for "the terminal you were last in"; the peer set is what is *on screen*; the policy stops at the tile boundary; the mutable cell moves out of the pure resolver; the receptacle is renamed for the volatility it hides | lowy F3 ≡ hickey F5 · lowy F4 ≡ hickey F3 · lowy F5 ≡ hickey F6 · lowy F7 · lowy F10 ≡ hickey F15 · hickey F4 · hickey F7 · hickey F13 · hickey F16 | `8b49933b0` |
| **the chrome policy reporter** — gate on the loaded preference; move off the CRUD façade to the app shell beside its `chrome.setActive` twin; read the existing membership authority instead of opening a second subscription; stop re-firing on every unrelated preference edit | hickey F2 · lowy F9 ≡ hickey F8 · hickey F9 | `3f9c51aea` |
| **minor polish** — narrow `resolveThemeBgs` to the theme names it actually computes on; move `shuffleMode`'s tests to the package that now owns it | hickey F14 · hickey F11 | `3c7acc96b` |
| **padi's build closure and the terminal-themes record** — bring the theme *table* into padi's build identity; correct what the new README says `terminal-themes/color` holds | lowy F1 · lowy F8 ≡ hickey F10 | `771c9a6c8` |

Two of those are worth calling out. `themes.json` sits at the terminal-themes
package **root**, outside the `src/**` `.ts`-only fileset — so padi's `pickTheme`
read a catalogue that was *not* in padi's build key, and two new comments claimed
the opposite. A remote host would have kept resolving creates from a stale
catalogue after a theme edit. Fixed at the source: the table now rides the
fileset, and both comments say which guard covers it. And the new README's
"Entry points" section described `terminal-themes/color` as the perceptual-distance
and light/dark/colourful classification code — that maths lives in `picker.ts`,
on the main entry point that padi *executes*; `color.ts` is an RGB string parser
used only by the screenshot path. The doc was the stated justification for the
`fileset.difference`, so a wrong reason there would have been mis-maintained.

### Deferred (author's call)

**hickey F1 — unify the ⌘⇧J "shuffle this terminal" action server-side.** hickey's
position: after this diff, ⌘⇧J is the only theme decision left client-side, and
it answers the same question from a second representation of the same three facts
(live `preferences()` + `isDark()` + `store.terminalIds()`, versus padi's cached
report + registry). The argument used to move new-terminal shuffle server-side
applies verbatim the moment anyone wants "shuffle this terminal" out of band. lowy
independently reached the *same* observation (F11) and dropped it, on the grounds
that the genuinely shared axis — the pool filter — is already dissolved into the
one `shuffleMode`, so this only becomes a duplicated receptacle when a re-shuffle
verb goes out-of-band. Deferred to a follow-up: it changes a second,
currently-working user-facing path, and this PR already spans five packages plus a
surface-contract bump. `shuffleMode` stays in `@kolu/terminal-vocab` and the
`kolu-common` re-export stays.

**lowy F6 — have kolu-server push the policy instead of the browser.** lowy's
position: `newTerminalTheme` and `shuffleBehavior` are kolu-server's durable
conf-backed preference while `isDark` is the browser's ephemeral resolution —
three fields, two owners, two lifetimes — and kolu-server already holds a padi
client on both arms, so the owner could push the half it owns and padi would
honour the setting even with no tab open. hickey did not dispute the report shape
itself; its F2 was that the *loadedness* of the preference was being thrown away
on the same hop, which is fixed above. Deferred: the only coverage it adds over
the browser report is the narrow "no browser has ever connected to this padi since
it started" window, and it would cost a persisted-preference schema migration plus
`SCHEMA_VERSION` bump, and write a browser-derived value into durable storage.
The connect-window hole was fixed directly instead (hickey F2). Residual: the
changelog and `theming.mdx` state the guarantee unconditionally; it holds once a
kolu window has been open at least once.

### Contested — the lenses disagree, left open

**Where the resolution belongs (lowy F2 vs hickey F5).** Both re-place the same
block and disagree on the target. lowy: move the whole resolve call down into
`createTerminal`, since the RPC handler is the front door for *wire* callers only,
and a future in-process create would bypass the policy with no signal. hickey:
keep the call in the handler and move the *inputs* into the policy module, so the
module owns its own decision instead of shipping an algorithm the caller has to
wire. These are not both satisfiable here: the policy module reads
`getActiveTerminalId` / `getNewTerminalPolicy` from `terminals.ts`, so having
`terminals.ts` call into it closes an import cycle, and `noImportCycles` is an
error in this repo. hickey F5 was applied; lowy F2's receptacle argument returns
if padi ever grows a second in-process create.

**How the Nix fileset and the closure test stay in sync (hickey F12 vs lowy F1).**
hickey: delete the hand-written `HASHED_ROOTS` / `EXCLUDED` maps and have the test
*parse* `default.nix`, so one side derives from the other rather than being kept
in lockstep by a comment. lowy: keep and *extend* the hand mirror — and note that
the test could not have caught this PR's actual gap under either design, because
`walkSources` is `.ts`/`.tsx`-only and `resolveSourceFile` returns `null` for
`.json`, so nix is the sole guard for `themes.json` regardless. The mirror was
extended; hickey's residual stands: a root added to the test but not to Nix still
yields a `PADI_BUILD_ID` that misses real daemon code.

<sub>Nothing pushed or merged by the review pass. Full per-finding record in
`.lens-debate/outcome.md`.</sub>

srid:
## Grok ⇄ Claude Code debate

✅ Consensus in 1 round(s)

Peer: **grok-4.5** (`grok --always-approve`), reviewing against merge-base `6dd135a18`.

_No fix commits: the peer found nothing at any severity on its first pass._

**Findings raised by Grok:** none.

> Round-1 review of the #2045 / #2051 change (new-terminal theme resolution moved to padi's lifecycle.create, with a chrome-reported NewTerminalPolicy). Diff against 6dd135a is clean at every severity I checked: correctness of the resolve path (override → absent policy → inherit → shuffle), split carve-out, absent-fact discipline on both sides of the hop (preferencesLoaded + null policy), multi-host re-report on connect, build-closure inclusion of terminal-themes (including themes.json), additive surface 4.5→4.6, and docs/changelog. Deliberate decisions (a)–(g) hold against the code; deferred items (h) left alone. Earlier architecture/lens fixes (remote disk-read regression, system→dark smush, floored-default connect window, split shuffle) are present and correct. No findings this round.

srid:

Ran with `--no-elegance` (the `/simplify` track already ran over this tree).

| Pass | Findings | Outcome |
|---|---|---|
| Rules (~30 project + built-in) | 1 | Fixed — `5aa58d192` |
| Fact-check | 0 | Clean |

**`caught-error-must-not-collapse-to-empty`** — `packages/client/src/hostScope/useNewTerminalPolicyReport.ts`. The policy report's `.catch` collapsed to a `console.error`. A failed report leaves that host's padi on its "nobody has told me" default, so new terminals there quietly stop following the setting — and the effect doesn't fire again until the policy changes or the link bounces, so the silence can last the whole session. That's the same invisible wrong-theme confusion #2045 is about, and DevTools is not a user surface.

The pass flagged this as a **design call** rather than patching it, because the twin report (`createViewState`'s `chrome.setActive`) carries an identical console-only catch — so fixing one site either diverges from the twin or drags the twin along.

**Adjudicated:** the twin logs because it fires on *every tile activation*, where a toast would be spam. That reasoning doesn't reach this report, which fires only on a preference change or a host's link coming back — rare enough to say out loud. So this is the twin's own principle applied at a different frequency, not a divergence from it. `setActive` is left alone.

Fact-check specifically confirmed sound: the pre-report `null` → default-theme path, the create/report race (the reporter waits on the *unfloored* `preferencesLoaded()`), `shuffleMode`'s `auto` mapping, one shared `NewTerminalPolicySchema` on both sides of the wire, and `inherit` degrading via optional chaining when the active terminal has gone.

srid:
## CI e2e metrics — `1583df1#1`

Full two-platform run, settled green — after merging `origin/master` (through #2092).

|---|---|---:|---:|---:|---:|
| `x86_64-linux` | `kolu-ci-3` | 8 | 32 | 8 | 3m 01s |
| `aarch64-darwin` | `ci@petit` | 4 | 10 | 6 | 4m 08s |

Durations are the whole `ci::e2e@<platform>` recipe wall time from `.ci/1583df1/runs/1.json`, not Cucumber's internal timer. Worker counts are the last `e2e: workers=` line each lane recorded.

All 46 required contexts green per `just ci::protect --dry-run` (both platforms pinned); no open code-scanning alerts on `pr=2051`.

srid:
## Evidence

**#2045 in one picture: a terminal nobody created from the browser now comes up on the theme you asked for.**

Recorded on an ephemeral `pu` box by this repo's own Cucumber + Playwright e2e harness under `KOLU_EVIDENCE=1` — same steps, same step library CI runs. The scenario:

1. opens **Settings** and sets **New terminal theme → Inherit** (clicking through Shuffle first, so the report to padi is a real user change, not the seeded default);
2. puts the active terminal on **Dawnfox** — a *light* scheme, chosen so "followed the preference" and "took the built-in default" can't be confused for each other in a still;
3. creates a terminal **without touching the browser at all** — the step dials this worker's padi and calls `lifecycle.create` directly, which is exactly the door `kolu mcp`'s `lifecycle_create` and `padi-tui create` come through;
4. tidies the canvas so both tiles are on screen, then focuses the out-of-band one.



▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=pr2051-mcp-inherit-theme.mp4

### Before ↔ after

Same scenario, same steps, run twice: once on `master` (`3a3cd84`, the merge base) and once on this branch (`4498d81`). Left tile is the terminal the user was in; right tile is the one created out of band.

| Before — `master` | After — this PR |
| --- | --- |
|  |  |
| The out-of-band terminal ignores **Inherit** and lands on the built-in default, `Tomorrow Night`. | The out-of-band terminal comes up on `Dawnfox` — the theme it inherited from the terminal the user was last in. |

The `master` run is a clip too, if you want to watch the bug happen: [before.gif](https://github.com/juspay/kolu/releases/download/evidence-assets/pr2051-mcp-inherit-theme-before.gif) · [HD](https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=pr2051-mcp-inherit-theme-before.mp4)

Both runs pass as written — the `master` one asserts `Tomorrow Night` (the bug), this branch's asserts `Dawnfox` (the fix). So the pair is a red/green pin on the actual behaviour change, not two screenshots that happen to differ.

### Mobile / coarse-pointer

**Checked, and deliberately not captured separately.** This change adds no control, no layout, and no hit target: the only thing that differs on screen is a terminal tile's colour scheme, which renders identically at every viewport width, and nothing in the diff reads `isTouch()` or `(pointer: coarse)`. The client side of the change is a removal — the theme pick moved out of `useTerminalCrud` and into padi — plus one app-lifetime `useNewTerminalPolicyReport()` hook that renders nothing. There is no small-viewport or touch-sizing surface here to regress, so a second capture at 390×844 or under a coarse pointer would show the same two tiles in the same two colours.

### Capture notes

The scenario and its three steps (`a terminal is created out of band via padi`, `the out-of-band terminal should show theme …`, `I focus the out-of-band terminal`) were authored **for the capture only** and staged onto the box — they are **not** part of this PR's diff and nothing was pushed to the branch.

PR #2060, opened 2026-07-30
Reject nested terminal splits instead of leaving them invisible

### LINKED ISSUE #2059: Canvas does not render a terminal split whose parent is itself a split child (nested parentId)
## What happened

Terminals created over the kolu MCP with a `parentId` chain two levels deep — i.e. `lifecycle_create { parentId: X }` where X was itself created with `parentId: <visible tile>` — are fully functional (padi roster lists them `active`, agents run fine, all MCP verbs work) but **never appear on the user's canvas**. The user could not see them at all.

## Reproduction (observed live, 2026-07-29)

1. Visible tile `07c83a2d` (a Claude supervisor session).
2. `lifecycle_create { parentId: 07c83a2d }` → terminal `7f206354` — renders fine as a split.
3. `lifecycle_create { parentId: 7f206354 }` → terminal `306f967b` (and later `d1e1a302`) — **not rendered anywhere on the canvas**, though `padi-tui status` showed both `active` with live agents, and screen/sendInput/wait verbs all worked against them.

Two independent instances (a Claude debate peer and a Codex auditor), same result both times.

## Expected

Either render the nested split (sibling group under the same layout node), or refuse the create with a typed error ("nested split unsupported") so the orchestrator knows to parent against the root tile. The silent-invisible middle is the worst outcome: driving agents look healthy over MCP while the human watching the canvas sees nothing — an agent-coordination session loses exactly the observability the canvas exists to provide (this bit a live multi-agent run tonight: the human asked "where is that terminal? I do not see it at all").

## Notes

- `lifecycle_create`'s schema accepts any terminal id as `parentId` — nothing constrains it to top-level tiles.
- Not investigated renderer-side; this report is from the MCP-driving side plus the user's observation.

srid:
Closes #2059.

A terminal created with a `parentId` that is itself a split child is fully alive in padi (roster, agents, MCP verbs) but **never painted on the canvas** — the client renders top-level tiles plus exactly one hop of children. Driving agents look healthy while the human watching the canvas sees nothing. The silent-invisible middle is the worst outcome.

padi now **refuses** to create such an edge, and **repairs** the ones already on disk.

### One rule, at the writes

Both generative parent writes — `createTerminal` and `chrome.setParent` — run a single total validator, `requireFlatParentEdge`, in the registry. It is total on purpose: no door can enforce a subset of it, and `createTerminal` mints its id *first* precisely so it runs the whole rule rather than the clauses that happen to apply.

| Parent | Answer |
| --- | --- |
| a live top-level tile | ✅ accepted — one-level splits are unchanged |
| itself a split child | `BAD_REQUEST` naming the **root tile** to parent against |
| dormant (sleeping) | `BAD_REQUEST` — a dormant tile swaps its whole body for `DormantTileBody` and paints no splits at all, so: wake it first |
| parked, or absent | `NOT_FOUND` — a parked record is an invisible restore-card placeholder, and every other client mutation already reads it as not-found |
| the child itself | `BAD_REQUEST` — a self-edge |
| a child that still has splits | `BAD_REQUEST` — the move would push *its* children to depth 2 |

### A saved workspace heals itself

Rejecting new bad edges does nothing for the ones a pre-fix daemon already wrote. Rehydration bypasses the fence by construction — it replays records rather than creating them — so all three doors now run one shared repair, `flattenUnpaintableParentEdges`:

- **survivor adoption** (`adoptSurvivingSession`) — the *normal* kolu upgrade path, where a nested live PTY would otherwise ride through every restart and be re-persisted, invisible for as long as it lives;
- **cold restore**, at the seed and again as each spawn settles — including the case where a parent's respawn fails and re-parks, which makes its children unpaintable after the fact.

A terminal that was nested under a split comes back as its own **top-level tile**. It is never dropped, and a **dormant** parent's splits are deliberately left alone — waking it re-mounts the panel with the child in place, so promoting there would destroy the user's arrangement for good. That distinction is the whole reason there are two predicates: `isPaintableParent` ("would the canvas paint a child here **now**", which the reject door wants) and `isPermanentlyUnpaintableParent` ("will it **ever**", which the repair doors want).

### Design philosophy

- **Fail fast** — no auto-flatten at the doors, no knob, no silent acceptance. Rewriting a caller's requested parent behind its back is the graceful degradation this repo bans; the typed fault sends it one level up instead. A parent cycle in a corrupt saved graph throws rather than handing back a plausible-looking non-root.
- **Reuse the source of truth** — one rule for both write doors, one repair for all three rehydration doors, reading one shared predicate. The redundant `requireActiveTerminal` narrow was *removed* from the create handler once the rule subsumed it, because all it could still do was shadow the more accurate fault.
- **Electricity** — n/a. Parent-edge legality is an application leaf, not a volatility boundary; it stays inside `packages/padi`. The daemon's depth limit and the client's rendering limit are linked by reciprocal comments rather than a shared exported constant.

### Try it locally

```sh
nix run github:juspay/kolu/fix/2059-reject-nested-parent
```

_Generated by [`/be`](https://github.com/srid/agency)._


srid:

**Status:** applied

| Lens | Findings |
| --- | --- |
| lowy | 0 (clean placement at shared guard; application concern, not electricity) |
| hickey | 1 fix (major) + 1 drop (minor) |

### Applied

- **hickey — parentId fence not at write chokepoint** — moved `rejectNestedParent` onto `createTerminal` and `setTerminalParent` so non-restore parent assignment is fenced mechanically; dropped servePadi duplicates (kept `requireActiveTerminal` on create). **Did not** fence `restoreSpawn` — rehydrate saved graphs remains deliberate (no auto-flatten / no break of legacy nests).

### Observations (not applied)

- **hickey** — root id only in free-text `BAD_REQUEST` message (matches other padi BAD_REQUESTs; no structured consumer this PR).
- **lowy** — session restore can rehydrate pre-#2059 multi-level nests; setParent still does not require parent existence (pre-existing).

_Generated by the lowy ∥ hickey lens review (`--no-comment`; posted by `/be-review` after push)._

srid:
## Codex ⇄ Grok debate

✅ Consensus in 2 round(s)

|---|---|---|
| 1 | 1636548d1 | fix(debate): reject setParent self-parent cycle |

- **F1** — setTerminalParent(id, id) is accepted when id is currently top-level: rejectNestedParent only inspects the proposed parent's existing parentId, sees none, and then updateClientMetadata writes parentId equal to the terminal's own id

srid:

| Pass | Issues | Details |
| --- | --- | --- |
| Rules | 0 | Clean across built-in + project rules |
| Fact-check | 1 → fixed | Re-parenting a non-leaf under another top-level tile invented depth-2 (OTHER → ROOT → CHILD); refused with BAD_REQUEST |
| Elegance | – | Skipped (`--no-elegance`; simplify already ran) |

**Actioned:** `fix(police): refuse re-parenting a non-leaf under another tile` (`a40926476`).

_All clear after the police fix._

srid:
## Evidence

padi now refuses nested terminal `parentId` (split-of-a-split) with a typed **BAD_REQUEST** that names the root tile. One-level splits still work. Self-parent and re-parenting a non-leaf are refused the same way.

**No canvas visual delta** — the product of #2059 was silent invisibility (canvas never rendered depth-2 splits). Product UI only creates one-level splits; the fix is refuse-at-daemon, not a pixel change. Honest proof is the real unit suite recording:



▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=2059-nested-parent-refusal.mp4

**Recorded command** (14/14 passed):

```sh
nix develop -c pnpm --filter @kolu/padi exec vitest run --reporter=verbose \
  src/servePadi.nestedParent.test.ts src/terminal-registry.test.ts
```

Covers: nested `lifecycle.create` / `chrome.setParent` → BAD_REQUEST + root name; one-level success; self-parent refuse; non-leaf reparent refuse; `rejectNestedParent` / `rootAncestorId` unit cases.

**Tests:** [`packages/padi/src/servePadi.nestedParent.test.ts`](https://github.com/juspay/kolu/blob/fix/2059-reject-nested-parent/packages/padi/src/servePadi.nestedParent.test.ts) · [`packages/padi/src/terminal-registry.test.ts`](https://github.com/juspay/kolu/blob/fix/2059-reject-nested-parent/packages/padi/src/terminal-registry.test.ts)

srid:
## Hold (supervisor)

Per supervisor ruling: **CI held** — no further darwin e2e retries for now (venue lease contention + human hold).

- Linux: full green including e2e 541/541 at prior SHAs
- Darwin: red only on known-class e2e flakes (code-tab branch-setup `:393`/`:419`, `git-context.feature:19` host git-init race, `worktree.feature:122` close blocker) — product diff is padi nested-`parentId` only
- Flaky Test Tracker updated in this PR for all three
- Evidence posted earlier; gauntlet (lens + Codex debate + police) complete
- **Not marking /be Done** until both platforms clear the green-gate

Awaiting supervisor go-ahead to resume darwin e2e.

srid:

**Status:** applied — every finding settled; nothing left for a human.

| Lens | Findings |
| --- | --- |
| lowy | 8 (4 major, 3 minor, 1 observation) |
| hickey | 12 (4 major, 8 minor) |

The two lenses read the diff independently and landed on the **same four majors**, which is what consensus looks like: the fence stated one third of the rule it claimed to own, restore rehydrated the very defect the PR names, `rootAncestorId` handed back a non-root on a cycle, and an unknown parent slipped through the guard entirely.

### Applied

| Commit | Area |
| --- | --- |
| `605ec4c00` | One total parent-edge rule in the registry |
| `bf48bb26a` | Session restore reconciles unpaintable parent edges |
| `9ed8a35ae` | Minor polish from the lens review |

**One rule instead of four** (lowy F1 ≡ hickey F2, ≡ hickey F3 / lowy F6, ≡ hickey F6 / lowy F5, ≡ hickey F5 / lowy F7). `rejectNestedParent` and the two inline guards in `terminals.ts` are replaced by a single **total** `requireFlatParentEdge(childId, parentId)` beside an `invalidParentEdge` fault factory — the sibling of the existing `terminalNotFound`, so the two doors can't drift on the wire shape. It names *every* way an edge breaks depth ≤ 1: self-edge, absent parent, parked parent, nested parent, non-leaf child. `createTerminal` mints the id first and runs the whole rule rather than a call-site subset; `setTerminalParent` collapses to `requireTerminal(id)` → the rule → one write. Supporting registry reads: `isPaintableParent`, `firstChildOf`. `terminals.ts` no longer walks the graph or mints `ORPCError`s.

**Restore no longer rehydrates the bug** (lowy F3 ≡ hickey F1). Both lenses found that `promoteOrphanedRestoreChildren` in `sessionRestore.ts` *already* auto-flattens for the neighbouring "the canvas would hide it" case (`!parent || parked`) — so "restore deliberately doesn't flatten" was a hole in a predicate, not a policy the repo holds. Any session written by a pre-fix daemon would have restored a split-of-a-split straight back into the invisible state #2059 is about. It's now `promoteUnpaintableRestoreChildren`, reading the same `isPaintableParent` predicate the reject path reads: one fact, two policies — the API doors **refuse**, the restore door **repairs**.

**`rootAncestorId` stops lying** (lowy F4 ≡ hickey F4). On a cycle it returned the first revisited id, which the fault then printed as "parent against this instead" — following the advice reproduced the error. It now throws on a revisited id, is module-private, and the redundant first-hop re-walk is gone.

**The daemon's depth limit and the canvas's are now linked** (lowy F2). The rule exists because `useTerminalMetadata` paints top-level ids plus exactly one hop of `childrenByParent`; the two encodings now name each other so neither moves alone.

Minors: registry header states what it now owns; the `servePadi` comment block trimmed back to what's about that handler; the new tests moved onto the existing `servePadi.testlib.ts` (with two older suites) instead of a third copy of the fixtures; the `setTimeout(0)` teardown drain deleted along with the real spawn it was papering over; clause-level cases moved to the validator's own suite.

### Arbitrations

Four calls where the lenses agreed the code was wrong but proposed different remedies; each resolved against `.claude/rules/conventions.md` → **fail fast, no fallbacks**, not by splitting the difference:

1. **Cycle in the ancestor walk** — hickey's loud throw over lowy's `string | undefined` + branched message. Corrupt state crashes; it doesn't get a polite rendering.
2. **Absent child in `setTerminalParent`** — hickey's `requireTerminal(id)` over lowy's early-return no-op. Both live callers were re-verified on disk to establish presence first, so nothing can reach the new throw.
3. **Restore** — applied over the author's stated "deliberately unfenced", because the neighbouring code already contradicted it. Restore still reconciles rather than throwing.
4. **The client/daemon link** — lowy's stated *minimum* (reciprocal comments) only; no shared `MAX_SPLIT_DEPTH` constant in `@kolu/terminal-vocab`, and no `.agency/lowy.md` row — lowy's own electricity probe classified this axis an application leaf warranting no seam, so a volatility-inventory row would contradict it.

### Not applied

**lowy F8** (observation, lens disposition `drop`) — every in-repo caller of `chrome.setParent` passes `null`; the non-null guards defend a door only an external wire client (MCP, `padi-tui`) opens. Lowy's own §6 single-consumer carve-out applies, and it's an argument for keeping the fault messages machine-parseable, which arbitration 1 does.

Verification: padi suite **463 passed**, `just check` green.

srid:
## Codex ⇄ Claude Code debate

✅ Consensus in 5 round(s) — Codex approved at round 5.

|---|---|---|
| 1 | 9cbd4545e | fix(debate): refuse a dormant parent and repair every rehydration door |
| 2 | f0f919b23 | fix(debate): let the create door say a parent is dormant, not "not found" |
| 3 | ea9e8d6e2 | fix(debate): answer a parked parent NOT_FOUND, like every other client mutation |
| 4 | 3a2dccaa2 | docs(debate): drop an impossible state from the parent-edge fault comment |

- **F1** (blocking) — isPaintableParent treats a top-level sleeping terminal as paintable because it excludes only parked records
- **F2** (major) — The restore repair iterates only activeRespawns and then explicitly skips every candidate that is not currently active
- **F3** (major) — A normal upgrade commonly adopts a surviving kaval rather than showing the cold-restore card
- **F4** (minor) — The new text says padi-tui create --parent rejects a sleeping tile with an error explaining which condition failed and telling the user to wake it first
- **F5** (minor) — Removing the handler's requireActiveTerminal guard also reclassifies a PARKED parent, contrary to the disposition's claim that parked is unchanged
- **F6** (nit) — The new comment says the residual BAD_REQUEST arm can represent an active record whose PTY is gone during unwind

srid:

| Pass | Issues | Details |
| --- | --- | --- |
| Rules | 0 | Clean across all 33 rules (8 built-in + 25 project), each with `Checked by:` evidence |
| Fact-check | 0 | Clean — the parent-edge guard traced end-to-end across all five `meta.parentId` writers, no bypass found |
| Elegance | – | Skipped (`--no-elegance`; `/simplify` already ran over this tree) |

**Nothing to action** — no fix commits from this review.

Worth recording from the fact-check, since both are properties the earlier rounds argued for rather than demonstrated:

- **Mutate-while-iterate is safe** in `flattenUnpaintableParentEdges`. `updateClientMetadata` mutates `entry.meta` in place rather than replacing the map entry, so the sweep's write does not disturb its own `terminalEntries()` iteration.
- **A legacy chain converges.** Hand-simulated several registry orderings of a three-level `A ← B ← C` and every one lands on a valid depth-≤1 graph — which is the bar the sweep's doc now claims (one hop, both outcomes visible), rather than the order-independence it used to claim.

srid:
## CI — e2e timing, both platforms

Run `7e98bd9#3` (linux) · `7e98bd9#4` (darwin re-run). Same commit; every required context green on both platforms.

| Platform | Host | Workers | Cores | Cap | `ci::e2e` duration |
| `x86_64-linux` | `kolu-ci-5` | 8 | 32 | 8 | 170.1 s |
| `aarch64-darwin` | `ci@petit` | 4 | 10 | 6 | 253.8 s |

**Why darwin ran twice.** `7e98bd9#3` leased **`zest`** for the darwin lane, and three nodes there died by signal rather than by assertion: `ci::pnpm-hash-fresh` was `Killed: 9` mid-`pnpm install`, `ci::daemon` exited **143** (SIGTERM), and `ci::e2e` produced **no exit code at all** with its log truncated mid-scenario. No Cucumber failure was reported by any of them, and this PR touches no `package.json` / `pnpm-lock.yaml`, so none of it was attributable to the diff. Re-run pinned to `ci@petit` (`#4`): the whole darwin lane green, `ci::e2e` included.

`zest` is unhealthy — worth pulling from the `aarch64-darwin` pool in `hosts.json` until it is repaired; it also intermittently fails its lease probe outright (`ssh: connect to host zest port 22: Connection timed out`), which stalls the darwin lane while its only sibling is busy.

srid:
Closing this without merging — the approach is wrong for where kolu is going, and that is worth saying plainly rather than merging because the work is done.

**Why.** This PR fixes #2059 by making the daemon *refuse* a nested split, and by *repairing* the ones already on disk. That repair rewrites `parentId` to `null` — permanently — on restore and on survivor adoption. kolu is going to get a real deeply-nested terminal tree model, and this PR destroys exactly the data that model will want to read: a workspace that has already been flattened on disk cannot be un-flattened later.

It also spends a lot of machinery — one total rule, two paintability predicates, three repair doors, ~800 lines across 16 files — defending a two-level invariant that is scheduled for demolition.

**What replaces it.** The same bug, fixed in the renderer instead: the canvas groups a tile's splits by their **root ancestor** rather than by their direct parent, so every descendant shows up as a flat split of the root tile. Nothing is hidden, and `parentId` is preserved byte-for-byte, so when the tree model lands, existing workspaces come back correctly nested with no migration. The Dock keeps showing the true tree.

A bonus falls out: a **dormant** middle node currently swallows its splits (`TerminalContent` wraps the whole split panel in `<Show when={isLive()}>`). Once descendants hang off the root tile, a sleeping intermediate can no longer hide them — which is the blocking finding the Codex review raised here, fixed by construction rather than by a guard.

**What carries over.** Only the two things that are true in any tree model: rejecting a self-edge (`parentId === id`), and refusing a parent cycle — the latter because a root-ancestor walk spins on one.

The review trail above (lens review, the 5-round Codex debate, code-police) stays useful reading; CI was green on both platforms at `7e98bd9`. Branch is kept on GitHub.

PR #2069, opened 2026-07-30
Eliminate Linux CI watcher and state flakes

srid:
**Linux CI now cleans up its own daemons, avoids needless synchronous state writes, preserves recursive file watching across nested-directory bursts, and waits for the two terminals that zoom tests actually compare.** A fresh `kolu-ci-6` run exposed each problem under real lane concurrency; the fixes address the owning lifecycle or race instead of adding test retries.

Four independent failure modes were pinned down:

- dev-smoke daemons outlived their runner process and contaminated later lanes; they now bind to that runner's PID.
- state tests performed redundant `conf` rewrites and `fsync`s under load; cleanup and migration delete only state that actually exists.
- Linux inotify could register a newly created parent while missing a descendant created in the same burst; directory-create repairs are now coalesced and bounded, root teardown/reopen is serialized, and reconciliation covers the restart window without resetting ordinary event delivery forever.
- the shared terminal-settle step accepted one mounted terminal even though both callers compare two; it now waits for both initialized xterms.

The Code-tab query hydrates directly before waiting for its pulse stream, while the stream snapshot still closes the read-before-watch race. Reads stay single-flight under pulse bursts, have a fixed fail-loud deadline, and remain abort-owned by poll teardown. The user-level folder-status scenario continues to cover initial ancestor roll-up, clean siblings, live post-mount updates, and tinting; it is 6/6 green on the recreated Linux pool box with zero retries. The desktop terminal case is 5/5 and its mobile caller 3/3 there, also without retries. Full strict certification follows the review gate.

Downstream surface accounting: `git grep -E 'pollOnChange|poll-on-change' origin/master` in drishti found no consumer of this exported leaf, so no pair PR is needed. ODU was checked at its current pin (`kolu` `3ff256f0d3de8cb585e588f490bae5b42eb7d89c`) and likewise has no use of either export spelling.

ODU-IMPACT VERDICT: `none` — odu does not consume `@kolu/surface/poll-on-change` at its pinned kolu revision.

_Generated by [`$ci`](https://github.com/juspay/kolu) on Codex (model `gpt-5`)._


srid:
## Claude ⇄ Codex debate

Consensus reached in 4 rounds against `master`.

| --- | --- | --- |
| 1 | 2428f6ec0 | Bounded watcher repair and delivery, serialized retirement, added query liveness, strengthened tests and docs |
| 2 | c72f854e7 | Stopped transient file churn from triggering structural rebuilds |
| 3 | 9850f60f4 | Corrected the watcher repair contract comment |

- **F1** — Directory creates caused one full root rebuild each and starved delivery.
- **F2** — The nested-edit regression pin could pass on synthetic reconciliation alone.
- **F3** — The Surface Reference documented obsolete abort-supersede behavior.
- **F4** — Shared-surface downstream accounting was absent.
- **F5** — A wedged single-flight read could silence the poll forever.
- **F6** — A poll with an already-aborted lifetime still issued work.
- **F7** — The deliberate second initial read needed its correctness rationale documented.
- **F8** — Consumer comments described removed abort-supersede semantics.
- **F9** — Created paths were inspected serially.
- **F10** — The coverage ledger overclaimed which starvation layer was pinned.
- **F11** — Fixed tracker rows needed defect-specific verification evidence.
- **F12** — State-test cleanup was array-shaped and did not pin the no-write case.
- **F13** — A reopened root could overlap native watcher teardown.
- **F14** — The real-watcher test deleted its repo before teardown settled.
- **F15** — Failure diagnostics cost two full-tree CDP reads on every poll tick.
- **F16** — Watcher teardown assigned `subscription = null` twice.
- **F17** — Node built-in imports were unsorted.
- **F18** — Vanished file creates were misclassified as structural repairs.
- **F19** — The watcher design comment retained the deleted vanished-create rule.

Issue #2065, opened 2026-07-30
e2e flake: code-tab.feature:800 Browse mode tints ancestor folders (linux, workers=8)


srid:
## Symptom

`features/code-tab.feature:800` — **Browse mode tints ancestor folders that contain a change**

```
Error: Directory "seed" was not marked as containing a change within ~60s while nudging /tmp/kolu-browse-foldertint/seed/dirty.txt
# or on attempt 1:
Error: Directory "src" was not marked as containing a change within ~60s while nudging /tmp/kolu-browse-foldertint/src/feature/a.txt
```

Cucumber: **503 scenarios (1 failed, 502 passed)** after internal retry (attempt 1 and attempt 2 both fail at different steps of the same scenario).

Lane: `ci::e2e@x86_64-linux` · venue **kolu-ci-2** · `e2e: workers=8 (cores=32 cap=8)`

## Attribution (not branch-local)

Reproduced on **plain `origin/master`** at **`bd0b7a5b9`** (#2064 *Graduate redundant E2E coverage to faster test layers*) with the **same venue and parallelism** as the PR branch reds. So this is a **master regression / flake**, not introduced by UW4 (#2058).

Suspect: #2064 touched the code-tab e2e tree (`packages/tests/features/code-tab.feature`, `code_tab_steps.ts`, solid-pierre FileTree tests, etc.).

Darwin e2e on the UW4 head was green (`ci@petit`); linux-only so far.

## Repro runs

| Identity | Tree | Command | Outcome |
| --- | --- | --- | --- |
| **09f805f#1** | `uw4` after merge of master (`09f805f50`) | full `ci::default` · linux=kolu-ci-2 · darwin=ci@petit | sole red: `ci::e2e@x86_64-linux` (same scenario); darwin all green |
| **09f805f#2** | same HEAD (flake budget, e2e-only) | `e2e@x86_64-linux --host x86_64-linux=kolu-ci-2` | red again, same scenario |
| **bd0b7a5#1** | plain master `bd0b7a5b9` | same e2e pin on kolu-ci-2 | **red, identical failure** |

### Master attribution command

```sh
git worktree add /tmp/master-attr bd0b7a5b9495a0db40bf306d02ed9b1c87a48470
cd /tmp/master-attr
nix run ./ci#odu -- run 'e2e@x86_64-linux' --host x86_64-linux=kolu-ci-2
# log: .ci/bd0b7a5/x86_64-linux/ci::e2e.log
```

[...]

PR #2071, opened 2026-07-30
feat(osfacts): OSF4 — the socket-holders verb, and kolu's last hand-rolled OS reader

srid:
**OSF4** — the last independent phase of the [osfacts plan](https://github.com/juspay/kolu/blob/osfacts-plan-refresh/docs/atlas/src/content/atlas/os-facts-tool.mdx). Adds the `socket-holders` verb to the binary and deletes kolu's fifth and final hand-rolled OS reader, along with the last `lsof` dependency in the daemon path.

This branch also carries the plan-of-record refresh it was opened for (OSF5 corrected to half-shipped — drishti's migration landed, the repo extraction has not).

## The verb

```
osfacts socket-holders /run/user/1000/kaval.sock --procs
V	2
H	claimed	991
P	991	1	kaval
```

A **path**, not a pid set, so it is its own verb; `--procs` is its one composable facet, costing holder *names* and never the holder set.

## The load-bearing change: three answers, never one

The reader it replaces returned `SocketHolder[]` and spelled three different facts as `[]`:

| answer | rows | exit | who can say it |
| --- | --- | ---: | --- |
| nobody holds it | none | 0 | **linux only** — `/proc/net/unix` is world-readable and lists every bound unix socket, so absence from a *recognized* table is proof |
| held, unnameable | `H unclaimed -` | 0 | linux (OSF6's shape, applied to a socket path) |
| could not look | `E … socket_holders …` | 1 | darwin (no such table exists), a build with no sensor, or a table that could not be recognized |

A supervisor that reads the first meaning while the third is true spawns a second daemon onto a live rendezvous socket. `recoverGatelessSquatter` folds all three exhaustively; only `none` is evidence of freedom, and even that re-probes before spawning.

The gauntlet spent most of its effort on exactly that asymmetry — every path that could still *manufacture* the affirmative answer. Five were found and closed: darwin emitting its blindness row only when it happened to witness a denial; a build with no sensor answering with an empty document; an unrecognized `/proc/net/unix` (empty, truncated, format-drifted, or a counterfeit header) read as a table with no matching row; a socket path beginning with a space losing that byte to the parser and reading as unheld; and a socket **this very process** was serving being reported `free`.

## A defect this verb exposed

The client parsed **any** non-zero exit whose stdout began with a version line as a document. But the binary writes that line on the usage path too — so an older binary refusing an unknown verb (exit 2) parsed as a perfectly well-formed *nothing found*, which is what an out-of-date binary on a caller's `PATH` produces every time. Only exit 1 — the documented "V line, then E rows" total-failure path — carries a document now.

That rule then turned out to be enforced on one spawn twin and inert on the other: `promisify(execFile)` reports a child's status on `.code`, while `execFileSync` throws the raw spawnSync result, whose status is `.status`. The reconciliation now lives in one internal module (`childFailure.ts`) with its own pins, so the twins cannot drift apart again.

## Design

- **Fail fast, no fallbacks.** No `PATH` lookup, no platform default, no catch-to-empty. `bakedOsFactsBin` throws at the composition root when the wrapper did not bake the binary, and a platform with no sensor reports `ENOTSUP` rather than an empty answer.
- **The inject, not the env var.** `readSocketHolders` sits on `EndpointSpec` beside `readProcessIdentity`, because `@kolu/surface-daemon-supervisor` is shared spine: kolu bakes `KOLU_OSFACTS_BIN`, drishti `DRISHTI_OSFACTS_BIN`, and a package that knew either name would belong to one of them.
- **The reusable fold ships with the parser.** `foldSocketOccupancy` lives in `osfacts-client` beside its structural twin `foldStartTimeReading` — one copy for both consumers rather than a hand-written three-way per composition root. What stays in the supervisor is the *shape of the ask* and the endpoint's disposition of the answer, which is policy and belongs there.
- **Self-exclusion belongs to the kill decision, not to identification.** The recovery may never SIGTERM the supervisor — but filtering our own pid out when the holder set is first *read* made an in-process serve look like an empty set, which was then reported `free`. Identification now sees every holder; a skew served by this process is refused and reported `incompatible`.
- **The pure decoders live in the lib** (`osfacts/src/unix_socket.rs`), so the `/proc/net/unix` row parse and the darwin `sockaddr_un` slot decode are unit-tested on *every* platform, not only the one that can produce them.

## Evidence (final HEAD)

- `osfacts`: **80** tests green across the hermetic suites — including fixtures for a bound socket, a stale socket *file* with no holder, a path containing a space, a path *beginning* with a space, a trailing carriage return, a prefix that must not match, a non-UTF-8 sibling row, a counterfeit table header, and the JSON face.
- darwin: the binary built and ran on `ci@petit` against a real bound unix socket — `H claimed <pid>` + `P <pid> <ppid> nc`; a walk that named nobody reported `BLIND_OR_EMPTY` rather than claiming absence. The `SOCKINFO_UN` / `un_sockinfo` / `sockaddr_un` layout was read off the real SDK header, not recalled.
- `@kolu/surface-daemon-supervisor`: **134/134** with `KOLU_DAEMON_TESTS=1` — including the gated squatter-recovery suite, which drives real child processes holding real unix sockets through the new injected reader end to end.
- `osfacts-client`: **39** tests, including the exit-2 refusal, its sync twin, and the contradictory-document refusal.
- `just check` green; `just atlas::check-sync` → `✓ docs/atlas/dist is idempotent`.

Every behavioural fix ships with a pin whose mutation reds it; the mutations are named in the commit bodies.

## Review

Three passes, each on the pushed branch: an architecture C1–C7 fan-out with an adversarial refuter per finding (3 real defects), a lowy ∥ hickey lens pass (18 applied + 1 adjudicated rename), and a Codex debate (9 findings, closed over 2 rounds). Their comments are below.

## Drishti pair

**[srid/drishti#130](https://github.com/srid/drishti/pull/130)** — a pin bump. Drishti consumes `convergeAdmit` / `probeDaemonIdentityFrom` / `drainAndAwaitExit` / `createConnectorDrainBudget` / `instanceKeyFromStartedAt`, none of which take an `EndpointSpec`, and it never called `socketHolders`; `osfacts-client` is already hydrated beside the `@kolu/*` packages. CI green on that pin, repinned to final HEAD, is the evidence for those claims.

## ODU-IMPACT VERDICT: `none`

Grepped `odu` at `origin/master` (`fe0e0fc21`) for `surface-daemon-supervisor`, `socketHolders`, and `osfacts` — **zero hits**. odu consumes `@kolu/surface` and `@kolu/surface-mcp` only. No ledger entry.


srid:

Status: **applied** — every finding settled. One went to the human valve and was adjudicated below; one is recorded as an observation.

**Counts** — lowy: 7 findings (2 major) · hickey: 15 findings (6 major). 18 distinct issues applied across 6 commits, 4 of them as `≡ agreed` pairs (both lenses reached the same conclusion independently). Zero direct contradictions, so the debate valve never opened.

| Commit | Area |
| --- | --- |
| `0a4610bd6` | the osfacts document layer |
| `576538234` | the client parser |
| `bb301f0e8` | the socket-holder fold moves to `osfacts-client` |
| `90bd40341` | the supervisor's holder fold reaches the operator |
| `7b30622b6` | one osfacts bake per composition root |
| `58012cf2a` | minor polish |

### The two both lenses reached independently

- **H1 — the three-way answer was discarded one call before the operator saw it.** `namedExternalHolders` flattened `unattributed` to `[]`, and `SocketSquatterForeignError` rendered `[]` as *"an unidentifiable process"* — verbatim the message this same PR's changelog promises is gone, and on darwin the **normal** path rather than an edge case. The `detail` string was computed, logged, and then dropped. The error now carries it.
- **F1 ≡ H2 — the fold landed one layer above its own established home.** `osfacts-client` already owned this exact construct in `foldStartTimeReading` + `processIdentityAsync`; the socket-holder fold sat in the supervisor. Moved, so kolu and drishti share one copy instead of each hand-writing the three-way.

### Also applied

- `SOCKET_HOLDERS_SOURCE` declared a `proc` facet **no code path can emit**, documented as live contract in five places, with a green test pinning the unreachable document (F5 ≡ H5).
- One volatility axis wired three different ways across the composition roots — lazy for identity, eager for holders, a third discipline in the testlib (F2 ≡ H10).
- `command: "?"` reproduced *inside* a holder the same in-band collapse the three-way exists to delete, and reached operators as `pid 4243 (?)` (F4).
- `impl_document!` existed only because every method was defined twice (H7); the reader side copy-pasted what the writer side had just factored, and the copies had already drifted — `L claimed 0` parsed while `H claimed 0` did not (H8); darwin's holder-identity arm was unreachable, with a comment asserting a race that cannot happen (H6).

### Valve 2 — adjudicated by the author

**H3 (hickey, major)** — rename the decision type, which after H2 sat one letter from the document type in the same file (`SocketHolderReading` vs `SocketHoldersReading`, with `foldSocketHoldersReading` returning the singular). lowy did not raise it; it praised the neighbouring naming discipline instead.

**Ruling: rename applied, proposed arm names declined.** It is now `SocketOccupancy` — naming the decision rather than its source — with the first arm `held`. But hickey's proposed `free` for the empty arm is the one word this contract must not use: the type's whole point is that even a proven-empty read is only *evidence* of freedom (the caller re-probes before spawning), so that arm stays `none`, and `unattributed` keeps the vocabulary OSF6 already uses for a socket no readable pid claims. Commit `1feb4d943`.

### Observation (recorded, not applied)

**F7 (lowy, `drop`)** — darwin's `Err` arm folded `ESRCH` (a process that exited between the census and the walk — absence) into `EPERM` (denial), making `BLIND_OR_EMPTY` unfalsifiable there. No consumer-visible behaviour changes, since the recovery only consults the reading after a probe already said the socket is serving. *Superseded in part by the debate that followed: darwin now emits that row unconditionally whenever the walk names nobody, so the flag it discussed no longer exists.*

srid:
## Codex ⇄ Claude debate

✅ Consensus in 3 round(s) — base `36420cc34`

|---|---|---|
| 1 | b0e6dfe78 | fix(debate): stop every remaining path that manufactures an affirmative absence |
| 2 | 5b1f70112 | fix(debate): a full header, and three comments that contradicted the fixes |

- **F1** — Darwin emits BLIND_OR_EMPTY only when at least one `socket_fds` call returned Err
- **F2** — The unsupported-platform implementation returns `SocketHolders::new()`, which exits successfully and means the affirmative answer 'nobody holds this path'
- **F3** — `unix_socket_inodes` has no way to distinguish a valid table containing no matching path from an empty, truncated, or structurally unrecognized `/proc/net/unix` document: every unparseable line is skipped and even `b""` or `b"garbage"` returns an empty inode list
- **F4** — The byte-exact Linux decoder removes all whitespace after the inode before comparing the path, and also strips a trailing carriage return from every line
- **F5** — After `socketServeState` has proved the rendezvous is serving and the holder reading is `held`, filtering out `process.pid` can leave no external holder; this branch returns `free` immediately
- **F6** — `SocketSquatterForeignError` gained a public `detail: string | undefined`, but `isSocketSquatterForeignError` does not inspect that field
- **F7** — The exported parser/type permits a `SocketHoldersReading` with no H rows or E rows but with P/U rows, and `foldSocketOccupancy` ignores those fact rows and returns `none`
- **F8** — `failureDocument` is exported from the production client module solely so its tests can call it, exactly the exported-test-seam shape called out in the review ledger
- **F9** — The checked-in client README still says fold policy lives with the consumer even though this diff deliberately moves `foldSocketOccupancy` into osfacts-client
- **F10** — The round changed safety invariants without updating their nearby contracts
- **F11** — The extraction introduced a malformed documentation block: it begins with a blank line and the dangling fragment `total-failure path, ...`, with the subject (`The ONE exit status that still carries a document`) missing

srid:

Scope: `36420cc34...HEAD`. Rules pass — **1 violation**. Fact-check pass — **clean**. Elegance skipped (`--no-elegance`; `/simplify` had just run over the same tree).

### Actioned

**`no-thin-wrapper-functions`** — `osfacts/src/schema.rs`. The simplify pass moved row order onto the `Document` trait but left each type's real body in a `pub fn normalize` that its own trait impl then forwarded to, with exactly one caller each. That is the fake indirection the trait's own doc comment argues against, and which it had already removed for `write_tsv` / `write_json` / `has_facts`. Both bodies now live in their `Document` impl, where the other three already do — `c7fce670d`.

### Fact-check — the failure paths it traced rather than trusting the comments

- **child crash / timeout** (`childFailure.ts`): bounded timeout, `SIGKILL`, correctly separates a signal-truncated stdout from a real document, no silent swallowing.
- **stale socket file** (`linux.rs` / `darwin.rs`): both read live kernel state — `/proc/net/unix` plus the fd table, and the BSD fd table — never the filesystem, so a dead holder's leftover socket file resolves to *no holder*.
- **check-then-act on the kill** (`recoverGatelessSquatter`): bounded retries, and identity re-attested immediately before the SIGTERM (fresh handshake, then fresh OS corroboration), so a pid that changed between identify and kill is never signalled.
- **darwin's read ambiguity**: reported as `BLIND_OR_EMPTY`, never manufactured into a false proof of absence.
- **adversarial `/proc/net/unix` rows**: byte-level parsing, so a non-UTF-8 name costs only its own row.
- the two `unwrap_or_default()` / `.ok()` sites the diff adds are both benign (a test helper that cannot return `None` on non-empty input; a parser whose `None` skips a malformed row rather than masking an answer) — neither is production-path suppression.

Every other rule came back clean with its grep or enumeration recorded. Two positive callouts from the rules pass: `invalid-states-unrepresentable` and `caught-error-must-not-collapse-to-empty` are *exemplified* by this diff — the `held` / `none` / `unattributed` union is precisely the shape those rules ask for.

srid:
## Evidence

No UI surface changed, so a screenshot would prove nothing. What this PR changes is **what the daemon supervisor is told when it asks who holds its rendezvous socket** — and the thing worth proving is that the three answers stay apart, on both platforms, against real sockets.

Both transcripts below are the **binary each CI lane built from this HEAD**, run against real bound sockets.

### linux (`x86_64-linux`)

```console
### 1. held — a live process holds it
$ osfacts socket-holders /tmp/.../kaval.sock --procs
V	2
H	claimed	896151
P	896151	896148	python3
exit=0

### 2. none — PROVEN: nobody holds this path
$ osfacts socket-holders /tmp/.../nobody.sock
V	2
exit=0

### 3. the stale socket FILE a dead holder left behind
   srwxr-xr-x  /tmp/.../kaval.sock          ← the file is still there
$ osfacts socket-holders /tmp/.../kaval.sock --procs
V	2
exit=0                                       ← and nobody holds it

### 4. unclaimed — bound, and its holder is not ours to inspect
$ osfacts socket-holders /run/systemd/private --procs
V	2
H	unclaimed	-
exit=0

### 5. a refused ask is not an empty answer
$ osfacts socket-holders          # no path
osfacts: socket-holders needs a PATH
exit=2
```

Case 3 is the one the old reader could get wrong in the other direction, and case 4 is the state that used to be spelled `[]` — indistinguishable from case 2.

### darwin (`aarch64-darwin`, on the CI box)

```console
### 1. held
$ osfacts socket-holders $SOCK --procs
V	2
H	claimed	63263
P	63263	63156	nc
exit=0

### 2. darwin CANNOT prove absence — and says so
$ osfacts socket-holders $D/nobody.sock
V	2
E	darwin_proc_fds	socket_holders	BLIND_OR_EMPTY
exit=1

### 3. the stale socket FILE, holder dead
   srwxr-xr-x  /var/folders/.../kaval.sock
$ osfacts socket-holders $SOCK --procs
V	2
E	darwin_proc_fds	socket_holders	BLIND_OR_EMPTY
exit=1
```

**That difference between the platforms is the point, not a defect.** Linux has a world-readable table of every bound unix socket, so absence from a recognized table is evidence — it answers `exit=0` with nothing. Darwin has no such table and Apple gates another user's descriptors, so a walk that named nobody says exactly that, and the supervisor folds it to `unattributed`: *not proven free*. Case 2 and case 3 are indistinguishable on darwin **because they genuinely are**, and the tool refuses to pretend otherwise.

### The consumer, end to end

The supervisor's gate-less-squatter recovery is exercised against **real child processes holding real unix sockets** through this exact binary — `packages/surface-daemon-supervisor/src/squatterRecovery.test.ts` under `KOLU_DAEMON_TESTS=1`, 134/134 green, covering the squatter recycle, the foreign-holder refusal (never killed), the compatible-holder adoption (PTYs preserved), and the new pin that a socket **this process** serves can never reach the driver as `free`.

srid:
### CI e2e metrics — `b3a687b#1`

| `x86_64-linux` | `kolu-ci-2` | 8 | 32 | 8 | **125.6 s** |
| `aarch64-darwin` | `ci@petit` | 4 | 10 | 6 | **279.6 s** |

Both lanes settled green on the final HEAD. `ci::e2e@x86_64-linux` needed a positional rerun for the tracked `code-tab.feature:817` folder-tint flake ([tracker row](https://kolu.dev/atlas/flaky-test-tracker.html) updated in this PR — it reproduced on the earlier osfacts PR #2011 too, and this diff touches no Code-tab, file-tree, or watcher code). `ci::website-nix@aarch64-darwin` needed one rerun for a dirty remote checkout of a commit that is clean locally — infrastructure, cleared on re-fetch.

srid:
### CI e2e metrics — `673f6e5#1`

| `x86_64-linux` | `kolu-ci-4` | 8 | 32 | 8 | **179.2 s** |
| `aarch64-darwin` | `ci@petit` | 4 | 10 | 6 | **254.1 s** |

**50/50 nodes green on the first attempt, no reruns**, against a HEAD that has `origin/master` merged. Earlier runs on this branch needed positional reruns for two failures that master has since fixed at the source: [#2080](https://github.com/juspay/kolu/pull/2080) (the rebuilt file watcher born deaf — the `code-tab.feature:800` folder-tint fault tracked as #2065) and [#2081](https://github.com/juspay/kolu/pull/2081) (the atlas-sync gate using the shared CI checkout as scratch space, which is what made a nix lane read a half-rewritten `docs/atlas/dist`).

PR #2037, opened 2026-07-28
Repo identity paints the same way across Kolu

srid:
The dock monogram was a good idea trapped in one surface. This extracts it into a **shared identity atom** and wires it everywhere multi-repo fleets are scanned — without redoing host-strip idle tabs or right-panel/split tabs.

### Shared paint (Hickey / Löwy)

| Layer | What |
| --- | --- |
| Glyph | `ui/repoMonogram.ts` — one fold for the letter |
| Component | `<RepoMonogram />` — sizes only (`md` / `sm` / `xs`) |
| CSS | `.repo-monogram`, `.repo-spine`, `.repo-identity-chip`, `.repo-group-band`, `.palette-host-header` |
| Colour | existing fleet-wide `assignColors` / `repoColor` / `hostHue` |

The dock **consumes** the atom; it no longer owns the fill formula. Layout still stays surface-local (palette row ≠ dock card ≠ restore band).

### Surfaces

- **Palette** — terminal rows: monogram + spine; host headers: host-hue wash + monogram
- **Restore** — repo groups: monogram band + `assignColors` on session groups
- **Inspector** — Work repo chip: monogram + frame from live `getDisplayInfo().repoColor` (same hue as dock)
- **Tile title** — monogram + left spine
- **Minimap** — repo-colour fill mix on tile rects

### Skipped (as requested)

- Host strip idle tabs  
- Right-panel tabs + split tabs  

### Docs

Switcher, sessions, right panel, tiles, canvas + unreleased changelog entries.

### Try it locally

```sh
nix run github:juspay/kolu/ui-identity-paint
```

_Generated by Grok Build (model `grok-4.5`)._

PR #2072, opened 2026-07-30
Inspector: repo and branch chips follow the terminal you're on

srid:
The Inspector's **branch and repo chips froze on whichever terminal painted first and never moved again**, while the working-directory line right beneath them kept updating — the split a user reads as "half the panel is stale." Reported from the field with a screen recording: switching from a `kolu` / `resize-bug` terminal to one in `AI` / `docs/pesu-chat-rewrite` left both chips still saying `resize-bug` and `kolu`.

The regression shipped in #2037, which restyled the chips into the shared repo-colour frame. It gated them on `Show when={git && identityColors()}` and then re-read git inside the body:

```tsx
{(gitAndColors) => {
  const colors = gitAndColors();   // snapshot
  const git = props.meta.git!;     // snapshot
```

A non-keyed `<Show>` runs its body **exactly once** and keeps it mounted for as long as the condition stays truthy. Every terminal has git context, so the condition never went falsy, the body never rebuilt, and both `const`s stayed pinned to the first terminal for the life of the panel. Nothing else remounted it either — `RightPanel` deliberately keeps both tabs alive and only `display:none`s the inactive one. The directory line was unaffected because it reads `props.meta.cwd` inline in JSX, which is why exactly half the cluster went stale.

### The fix

Git facts and their paint come from **one `createMemo` behind one `Show` accessor** — the same glitch-free bundling `useTerminalStore.active()` already uses for `(id, meta)`, and for the same reason. With a single value there is no second thing left to snapshot, and a torn pair — this terminal's branch wearing the last one's hue — becomes unrepresentable rather than merely avoided. The two chips are now props-taking components, so props compile to getters and there is no loose expression in the `Show` body inviting the same mistake.

The review gauntlet then proved the old "prefer fleet projection, fall back to `assignColors`" branch was **dead code** — both arms returned byte-identical strings — so the store read, the `terminalId` prop, and a graceful-degradation fallback the design philosophy forbids all went with it. Hues now key off the canonical `terminalKey` projection rather than a local restatement of it, so the "chips match the dock" promise rests on one expression instead of two that happen to agree.

`WorkSection` moves into its own file. It is the one Inspector section with real derivation of its own rather than layout, and splitting it lets the regression be covered by a render test without dragging the wire-bound siblings (Compose / Ports / Attach) into the module graph.

### A second bug, found by the gauntlet

Bundling the PR facts introduced one: `<Disclosure open={p().hasException}>` re-asserted its open state on **every** PR field change, because `Show`'s inner memo uses default `===` and the bundle's identity turns over each recompute. During a live CI run — exactly when the panel is worth watching — every poll that added a check run **slammed shut the check list the user had just opened**.

A/B probe against the pre-diff component under `createStore` + `reconcile`, list opened by hand, only `pr.title` changed:

| | before this diff | after the bundle |
|---|---|---|
| manual expand survives | ✅ | ❌ snaps shut |

Fixed at `Disclosure` itself, not at the caller: its own docstring promises *"the user's click wins until the fact changes again"* — a promise about the **value**, which its effect broke by re-asserting on every **dependency tick**. A `createMemo` over the boolean stops propagation at the value, so no future caller can express the bug either.

### Why no test caught the original

The e2e seam asserts only that `inspector-branch` is **non-empty** — which stale text satisfies perfectly. Non-emptiness was never the property that broke. Every new assertion here was **verified to fail for the reason it claims to guard**, by temporarily reintroducing the defect:

| Assertion | Reintroduced defect → observed failure |
| --- | --- |
| branch / repo text | `expected 'resize-bug' to be 'docs/pesu-chat-rewrite'` |
| PR link, number, title, rollup | `expected .../2037 to be .../2072` |
| exceptions-sort-first | `expected ['ci::biome', …] to deeply equal ['ci::typecheck', …]` |
| per-chip fleet hue | swapping the two `color=` props fails both hue tests |
| disclosure keeps a manual toggle | `expected false to be true` |

That last column is the point: an assertion is only coverage once it has been observed to fail. Two of these tests were **vacuous when first written** — the PR fixtures all had `pr: absent`, and the sort fixture entered in the order it expected out — and the gauntlet caught both.

### Audit of #2037

The rest of that PR was swept for the same defect class. **The Inspector was the only instance** — every other similar-looking `const` in the diff sits inside an accessor function, so it re-evaluates per read. One lesser finding, fixed here: `data-testid="inspector-repo"` wrapped the decorative monogram, so the chip's `textContent` read `"Kkolu"`. No step reads it today, so it was a latent trap; #2037 states exactly this rule for the restore band and missed it one file over.

> **Found outside this PR's scope, deliberately not fixed here.** The same snapshot shape appears at `CreateIdentityPreview.tsx:200` (a live "Will create" preview, so most likely user-visible), `PadiInfoDialog.tsx:157`, and `CodeTab.tsx:1160`. Each needs its own reproduction first. A biome GritQL plugin expressing the whole class was prototyped and works — 20 hits repo-wide, a triageable sweep rather than a migration — and belongs in its own PR. `.claude/rules/solidjs.md` already states "Props stay reactive" as prose that nothing enforces.
>
> Also deferred, with the lens review's reasoning in its PR comment: a shared `identityColors` seam; threading `(id, meta)` as one prop from `App`; unifying the check-outcome tables (`prTooltip` declares itself the single source but `WorkSection` and `ChecksIndicator` have already drifted from it); the monogram glyph into CSS `content`; a `ui/FleetHueChip` primitive; a shared `TerminalMetadata` test fixture builder.

### Try it locally

```sh
nix run github:juspay/kolu/wtf-bug
```


srid:
## Codex ⇄ Claude Code debate

✅ Consensus in 3 round(s) — base `36420cc34`

|---|---|---|
| 1 | 8a572d1ad | fix(debate): canonical hue keys + PR/CI regression coverage |
| 2 | 12029bdc9 | test(debate): make the exceptions-first assertion adversarial |

- **F1** — The code directly restates the git-to-colour identity mapping as [git.repoName, git.branch]
- **F2** — The regression file says the Work cluster and every chip must repaint, but all fixtures set pr to absent and the suite asserts only git identity plus cwd
- **F3** — The repo-chip test deliberately moved inspector-repo onto the name span, so it no longer observes the decorative RepoMonogram

srid:

**Status: applied** — base `36420cc34`. Two independent Opus lenses read the full diff in parallel; one reconcile pass merged and committed. **lowy: 10 findings · hickey: 9 findings.** No direct contradictions — the lenses agreed everywhere they overlapped.

| Commit | Area |
|---|---|
| 770fa232f | major — the Work section's derivations |
| e9f63373f | minor — polish from the lens review |

### The headline, agreed by both lenses independently

`identity()`'s "prefer the fleet projection, fall back to `assignColors`" branch was **dead code**. Both lenses traced `terminalKey` → `assignColors` → `stableHue` and proved the two arms return byte-identical strings for every input that reaches them: `assignColors` hues by key string alone, and `buildTerminalDisplayInfos` keys off `terminalKey(meta)` = `(git.repoName, git.branch)` — exactly what the local spelling recomputed.

Deleting it removed the store read, the `terminalId` prop, and a graceful-degradation fallback the design philosophy forbids outright. It also closed a hole in this PR's own thesis: the memo bundled git+paint so a torn pair would be unrepresentable, but the *inputs* stayed split (git via `props.meta`, hue via a separate `props.terminalId`) — the store bundles `(id, meta)` at source precisely to prevent that, and the prop chain un-bundled it four layers down.

### Also applied

- **hickey F2** — one `work()` memo replaces five PR-side derivations; the CI rollup's verdict and its tally now come from one read. Verified `prValue`/`prUnavailableSource` are mutually exclusive arms of `PrResultSchema` before bundling.
- **hickey F4 ≡ lowy F6** — `BranchIdentityChip` as a props-taking peer of `RepoIdentityChip`, closing the const-snapshot hazard structurally rather than by comment.
- **lowy F10** — the hue test is store-free and asserts the real derivation instead of a mock.
- **hickey F3 / F6 / F7 / F8 / F9** — one git gate; `inspector-repo-chip` on the frame instead of climbing `parentElement`; six no-op casts removed; duplicated bug narrative and a stale tier description trimmed.

### Adjudicated by the human — deferred, not dropped

Six structural items were routed to valve 2 and deferred out of this bug-fix PR:

1. A shared `identityColors(TerminalLocation)` seam across ~6 `assignColors` sites.
2. Threading `(id, meta)` as one prop from `App` down (the WorkSection half is fixed; the rest is four files outside this diff).
3. Unifying the check-outcome tables — `prTooltip.ts` declares itself the single source, yet `WorkSection`'s `OUTCOME` has already drifted from it (`✕`/"failed" vs `✗`/"fail") and `ChecksIndicator` holds a third. **Real present-day drift, pre-existing, worth its own PR.**
4. Moving the monogram glyph into CSS `content` so the textContent trap is unspellable.
5. `<Show keyed>` — **dropped outright**: the chip extraction already closes that hazard structurally, and two mechanisms for one hazard is worse than one.
6. A shared `ui/FleetHueChip` primitive (decide with 4).

**Observations, not applied:** the collision suffix is correctly omitted from the chips; non-git terminals correctly get no identity chips.

srid:

Scope: `git diff 36420cc34...HEAD` — 6 files. Run with `--no-elegance` (the `/simplify` track already ran over this tree).

| Pass | Findings |
|---|---|
| Rules (7 built-in + 26 project) | **0** |
| Fact-check | **0** |

**All clear — no fixes to route.**

The fact-check pass traced the reactivity chain for this PR's own bug class end to end and confirmed the structure holds:

- Both `identity()` and `work()` are `createMemo`s whose bodies fully re-execute on dependency change — structurally immune to snapshot-into-const.
- Every JSX consumer calls the accessor fresh at its use site (chip props, anchor attrs, `<For each={p().checkRuns}>`, the `Disclosure` props) — no local `const` binding of an accessor result survives anywhere in the diff.
- `terminalKey(props.meta)` provably takes its git arm past the guard, matching its docstring.
- `stableHue` is a pure hash of the key string alone, independent of which other terminals are open — so the docstring's claim that the Inspector's locally-computed hue equals the dock's fleet-wide one is fact, not wishful thinking.
- `PrInfo.checks` is `CheckStatusSchema.nullable()` (never `undefined`), so `work()`'s null checks are exhaustive against the real schema.
- The `inspector-repo` → `inspector-repo-chip` testid split was checked against every existing e2e step definition — nothing depended on the old placement, so no silent breakage.
- No bare `catch`, `|| true`, `as any`, or `@ts-ignore` in the three changed production files. The one `throw` in `identity()` is fail-loud and a verbatim carry from pre-existing code.

`no-vacuous-assertion` was audited specifically: each new test was traced against the shape it claims to catch and confirmed to fail against it.

srid:
## Evidence

**Video** — two terminals in different repos/branches (`kolu-inspector-alpha` @ `alpha-feature`, `kolu-inspector-beta` @ `beta-feature`), Inspector tab open, switching back and forth. `inspector-branch`, `inspector-repo`, and `inspector-directory` all repaint together on every switch — including the switch **back** to the first terminal, which is exactly where the frozen-chip bug showed up.



▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=inspector-work-chips-follow-the-active-terminal.mp4

**Mobile / coarse-pointer check** — the Inspector also renders inside `RightPanelDrawer` at phone widths. Captured a still at the `@mobile` viewport (390×844, `hasTouch + isMobile` → `(pointer: coarse)`) — same chips, same testids, correct:



### How this was captured

New scenarios in `packages/tests/features/right-panel.feature`, reusing existing steps (`I press the toggle inspector shortcut`, `I click the right panel tab`, `I create a terminal`, `I press the switch to terminal {int} shortcut`, `I tap the mobile pull handle`, `I tap the mobile inspector toggle`) plus three new `Then` steps in `right_panel_steps.ts` (`the inspector branch/repo chip should contain`, `the inspector directory should contain`) that poll each testid's own text — proving it changed, not merely that it's non-empty (stale text is non-empty too, which is exactly why the original bug shipped past the old assertion).

Verified the new assertion actually catches the regression: ran the same scenario against `master` (pre-fix) — it passes right up to the "switch back to terminal 1" step, then fails there with the branch chip still reading `beta-feature` instead of `alpha-feature`. Ran green **twice** on the PR branch before committing.

Recorded on an ephemeral `pu` box, via the real nix-built `koluBin` (so the server's `KOLU_OSFACTS_BIN` bake matches CI) under `KOLU_EVIDENCE=1`, transcoded with ffmpeg.

PR #2073, opened 2026-07-30
Ship kolu's client CLIs to remote hosts, so agents there can drive kolu

srid:
Start a coding agent in a terminal on a remote host and it was cut off: `kolu`, `kaval-tui`, and `padi-tui` simply weren't on that machine — nothing is installed there by design — so **`kolu mcp` had nothing to run and an agent working remotely couldn't drive its sibling terminals**, a loop agents on the local machine have always had.

The tools now **ride the closure kolu already provisions**. Nothing new is installed, and no new mechanism was needed on the wire.

### The one property that makes it work

`padi-agent` (padi + the two TUIs + `kolu`) is the closure a host receives, and its wrapper bakes `KOLU_AGENT_TOOLS_PATH` to **its own `$out/bin`** — a self-reference resolved at build time. So a padi reached as `ssh <host> <agentPath>/bin/padi --stdio` reports the closure that was *actually copied there*:

```
provisioned closure ──> bin/padi  (wrapper bakes KOLU_AGENT_TOOLS_PATH=$out/bin)
                   ├──> bin/kaval-tui
                   ├──> bin/padi-tui
                   └──> bin/kolu
```

ssh has no env channel and nothing rides argv. The tools a terminal gets and the daemon that spawned it are **the same store path by construction**, so the tool/daemon version skew `PADI_BUILD_ID` exists to catch cannot arise here at all.

`@kolu/surface-remote` needed no new concept: `sshConnector` has always taken the *package to provision* and the *binary to exec* as two parameters. Only the caller's choice changed.

### Spawn policy became one value

`composeSpawnInput` took a positional per fact a terminal needs to know (`kavalSocket`, then `padiSocket`, …). It now takes a `TerminalEnvSpec` — `{ kavalSocket, padiSocket, toolsPath }` — so the set is enumerable in one place instead of growing an argument each time.

`kolu-pty` owns the mechanism and guarantees it **twice**, which is the non-obvious part: the spawn env alone is not enough, because the wrapper rcfile replays your dotfiles afterward and an absolute `export PATH=…` in `~/.bashrc` would silently drop the toolchain. So the dirs are re-asserted in the rcfile *after* the replay. The re-assert block carries **no caller data** — paths arrive in a variable, never interpolated into shell source — so no quoting question arises.

### Two things worth a reviewer's attention

> **`padi-agent` is exposed but deliberately NOT proven.** `mkProvenAgentSource` proves an attr by re-importing `default.nix` from the assembled tree; its own comment records the invariant that makes that terminate — a proven attr must never reach the `agentFlakeSrc` thunk. This closure reaches it three ways (`kolu` and both TUIs bake the flake ref), so proving it is an infinite regress. This is not theoretical: the first version of this composition failed with `error: stack overflow (possible infinite recursion)`. `nix/agent-packages.json` now separates `expose` from `prove`; `padi`/`kaval` — what a dial actually resolves — stay proven from the same tree.

> **The remote closure grows 609 MB → 690 MB** (`nix path-info -S`), mostly the client dist and the TUIs' `nix`/`openssh`. Measured, not estimated. Paid once per host, and `nix copy` skips paths already valid there.

Also fixes a latent **two-behaviours** bug found on the way: `dialAgentOnce` used `binary` as the package name, so `padi-tui --host` / `kolu mcp --host` provisioned a *different* closure than the browser binder for the same host. Both paths now name the same attr.

### Verification

- New behavioral tests spawn **real bash and zsh** against a fake home whose rc clobbers `PATH`, and assert the toolchain survives. Falsified: removing the fix turns exactly 3 of them red.
- `nix build .#padi-agent` green; the built wrapper's baked path was confirmed equal to the closure's own store path.
- Unit lanes green for `kolu-pty`, `@kolu/padi`, `kolu-server`, `@kolu/surface-remote`; `just check` green.
- Still to run: full two-platform CI, and a real remote-host browser session (`/remote-host-testing`) — the only harness that reaches this surface.

**Downstream verdicts** (grounded in a `gh search code` grep, not assumed): **odu — `none`** and **drishti — unaffected**; both consume `sshConnector`, whose signature is untouched. The only `@kolu/surface-remote` change is an additive optional `package` field on `dialAgentOnce`, which neither uses in code, so no drishti pair PR was opened — flagged here rather than silently skipped.

### Try it locally

```sh
nix build github:juspay/kolu/feat/remote-agent-tools#padi-agent
```


srid:

**Status: applied — 1 item routed to the human.**

Two structural lenses reviewed this diff independently — **lowy** (volatility-based
decomposition: does each boundary encapsulate a real axis of change?) and
**hickey** (structural simplicity: what is complected, what is fragmented?). They
were run in parallel with no knowledge of each other, then reconciled in one pass.

| lens | findings | major | minor | dispositioned `fix` | dispositioned `drop` |
| --- | ---: | ---: | ---: | ---: | ---: |
| lowy | 13 | 5 | 8 | 11 | 2 |
| hickey | 11 | 3 | 8 | 10 | 1 |

They converged independently on the same headline defect: **`KOLU_AGENT_TOOLS_PATH`
carried two facts under one name** — what a wrapper *bakes* onto a daemon, and what
a daemon *stamps* onto a terminal. Because the local wrappers merged onto an
inherited value, running one build's `kolu` from inside another build's terminal
gave the older build's TUIs precedence on every terminal the newer daemon spawned
— the exact tool/daemon skew the design claims is unspellable. Splitting the names
(`KOLU_AGENT_TOOLS_PATH` = bake, `KOLU_TERMINAL_TOOLS_PATH` = stamp) makes it
unspellable for real, and deletes the `KOLU_PADI_BIN` gate that existed only to
tell the two roles apart.

### Commits

| SHA | Area |
| --- | --- |
| `89b10cd92` | split the toolchain bake from the terminal stamp |
| `f7f27c86b` | one setter per binary, and the composition proof in the artifact it proves |
| `80ace4961` | one owner for "which closure this host receives" |
| `ffac48034` | minor polish from the lens review |

### Findings

`≡` = both lenses raised it independently.

| id | lens | sev | finding | outcome |
| ≡ | lowy F3 · hickey F1 | major | `KOLU_AGENT_TOOLS_PATH` complected the wrapper's bake with the terminal's stamp; a third, unrelated variable was pressed into service as a discriminator | applied — names split, gate and its 12-line comment deleted (`89b10cd92`) |
| ≡ | lowy F1 · hickey F2 | major | The local wrappers MERGED where they must assert, so a foreign build's tools could outrank this build's | applied — `koluBin` drops its setter; `default` is the single `--set` with the complete value (`f7f27c86b`) |
| ≡ | lowy F2 · hickey F3 | major | The composed-wrapper proof hung off the REMOTE closure, and its comment claimed to be a build input of `default` — which it was not, and could not be | applied — derivation deleted, assertion inlined into `default`'s own builder (`f7f27c86b`) |
| ≡ | lowy F4 · hickey F6 | major/minor | `"padi-agent"` was two string literals in two packages, held equal by a comment | applied — one `PADI_REMOTE_DIAL = { package, binary }`, imported by both dial paths and the test (`80ace4961`) |
| ≡ | lowy F7 · hickey F10 | minor | `prove ⊆ expose` was prose, not a check | applied — `lib.assertMsg` at eval time (`f7f27c86b`) |
| ≡ | lowy F9 · hickey F9 | minor | `toolsPath?` had two spellings of "none" its only producer cannot emit | applied — required, `[]` as the honest empty (`89b10cd92`) |
| ≡ | lowy F11 · hickey F8 | minor | The prepend/dedupe rule is implemented twice, in two languages, kept equal only by co-location | applied in part — one `PATH_PREPEND_CASES` oracle drives the TS half and the shell half under real bash and zsh. **One row is contested — routed below** (`89b10cd92`) |
| | lowy F5 | major | `package?: string` defaulted to `binary` — the very expression that gave one host two closures | applied — required; `kaval-tui --host` now states `package: "kaval"` (`80ace4961`) |
| | lowy F6 | major | The prove no longer covers the attr any dial resolves, and the comment said it did | applied — comment corrected, reciprocal pointer added above `ci::agent-flake-nix` (`f7f27c86b`) |
| | lowy F8 | minor | `TerminalEnvSpec` claimed to gather every daemon fact but left `serverVersion` a module global | applied — added to the spec; the boot-order guard stays pinned (`89b10cd92`) |
| | lowy F10 | minor | The composition check reached through makeWrapper's emitted text and would fail OPEN on an upstream shape change | applied — asserts exactly one `^exec ` line per wrapper before the `sed` (`f7f27c86b`) |
| | lowy F12 | minor | `kaval-tui --host` gives an already-provisioned host a second, tool-less closure | observation — deliberate layering, now stated at the call site and pinned by its test (`80ace4961`) |
| | lowy F13 | minor | `koluAgentTools` is an expendable `buildEnv` | observation — kept: the single `--set` needs the toolchain as one directory string, and `agentToolPackages` removes the duplication that motivated the flag |
| | hickey F4 | minor | The client toolchain was enumerated twice (`buildEnv.paths` and three `ln -s`) | applied — `agentToolPackages` names it once; both arms derive from it (`f7f27c86b`) |
| | hickey F5 | minor | `padi-agent` hand-rolled `mkdir` + `ln -s` against the `pkgs.buildEnv` used 30 lines above | applied — now a `buildEnv` with a `postBuild` `wrapProgram` (`f7f27c86b`) |
| | hickey F7 | minor | The env var's wire format was parsed in `@kolu/padi` while `kolu-pty` owned the name, the separator and the shell parse | applied — `agentTools.ts` deleted, `readAgentToolsBake` lives beside the constants, tests moved (`89b10cd92`) |
| | hickey F11 | minor | The prove mechanism's own limitation was stated in kolu's manifest, far from `mkProvenAgentSource` which imposes it | applied — rule moved into `agent-source.nix`'s header; `default.nix` points at it (`ffac48034`) |

### Evidence

- `just fmt` and `just check` green.
- `nix build .#padi-agent` green. The inlined composition assertion is
  **falsifiable**: pointing the setter at a directory with no `kolu` fails the
  build of `.#default` (and therefore `.#padi-agent`) with
  `FAIL: no 'kolu' on the composed KOLU_AGENT_TOOLS_PATH`. Restored and rebuilt
  green.
- Built wrappers verified: `default` carries
  `KOLU_AGENT_TOOLS_PATH='<kolu>/bin:<kolu-agent-tools>/bin'`; `padi-agent`
  carries its own `$out/bin`; a bare `.#koluBin` carries **no** toolchain, by
  design.
- Unit lanes: `kolu-pty` 54 passed (with `KOLU_DAEMON_TESTS=1`, so the shared
  oracle runs under real bash **and** zsh), `@kolu/padi` 481 passed,
  `kolu-server src/padi/` 120 passed, `@kolu/surface-remote` 30 files passed,
  `kaval-tui` 8 files passed.

### Routed to the human

#### 1. A direct contradiction — does the prepend rule sanitize the caller's `PATH`?

`prependPathEntries` (TS) drops empty entries from the caller's existing `PATH`;
`PATH_REASSERT` (shell) leaves `$PATH` byte-identical apart from the prepend. The
two lenses want opposite resolutions of that single case, and the choice is a
security-posture call, not a structural one: is dropping a user's implicit-CWD
`PATH` entry a hardening kolu should extend to the shell, or an unasked-for
rewrite of someone else's environment?

> **lowy F11** — Give the two implementations one shared oracle rather than one
> shared file. Beside them in `shell.ts`:
>
> ```ts
> /** The prepend/dedupe rule as DATA — the single source both implementations are
>  *  tested against, so the TS half and the shell half cannot drift. */
> export const PATH_PREPEND_CASES: ReadonlyArray<{
>   path: string;
>   dirs: readonly string[];
>   expect: string;
> }> = [
>   { path: "/usr/bin:/bin", dirs: ["/a", "/b"], expect: "/a:/b:/usr/bin:/bin" },
>   { path: "/usr/bin:/a", dirs: ["/a"], expect: "/usr/bin:/a" },
>   { path: "", dirs: ["/a"], expect: "/a" },
>   { path: "/usr/bin::/bin", dirs: ["/a"], expect: "/a:/usr/bin:/bin" },
> ];
> ```
>
> Drive the existing `describe("prependPathEntries")` from `PATH_PREPEND_CASES`
> with `it.each`, and add a `describeDaemon` case in the same file that, for each
> row, runs `PATH_REASSERT` under real bash and zsh with `PATH=row.path` and
> `KOLU_AGENT_TOOLS_PATH=row.dirs.join(":")` and asserts the resulting `$PATH`.
> One added row is then asserted against both halves; a divergence is a red test
> instead of a silent split.

> **hickey F8** — `prependPathEntries` runs `.filter((e) => e !== "")` over the
> **caller's existing PATH**, silently rewriting it (dropping the implicit-CWD
> entries the test celebrates catching). `PATH_REASSERT` only filters the
> *incoming* dirs; it leaves `$PATH` byte-identical apart from the prepend.
>
> So the same terminal is "hardened" against empty `PATH` entries at spawn and
> un-hardened again the moment a dotfile reintroduces one — the security property
> the TS test asserts does not survive to the shell that actually runs. Worse,
> `prependPathEntries` is doing two things under one name: prepend, *and*
> sanitize the caller's list. […]
>
> Two edits. 1. Stop the TS half from rewriting the caller's PATH — make the
> dedupe read-only:
> ```ts
> const existing = (currentPath ?? "").split(":");
> const fresh = dirs.filter((d) => d !== "" && !existing.includes(d));
> return [...fresh, ...existing].join(":");   // caller's PATH preserved verbatim
> ```
> (drop the `"/usr/bin::/bin"` expectation from the `prependPathEntries` test;
> the empty-entry hardening, if wanted, belongs in `cleanEnv` where it can apply
> to *both* halves).

Applied meanwhile: the shared oracle ships with the three rows both lenses agree
on, asserted against the TS half **and** against `PATH_REASSERT` under real bash
and zsh. The disputed row stays a TS-only test carrying a comment that names the
divergence and points at this open call — a stated decision rather than a silent
split inside a table only one half satisfies.

#### 2. A process gate — the drishti pair-PR for lowy F5

Making `DialAgentOnceOptions.package` required is an API-facing change to
`packages/surface-remote/`. Per `.claude/rules/surface.md` that needs a paired
`srid/drishti` PR green against the **final** kolu HEAD, plus an odu-impact
verdict in this PR's body grounded in a grep of odu at its pinned kolu SHA. The
PR already tripped that gate (the base diff added the optional field), but it
must be re-validated at this HEAD — which this pass cannot do.

### Adjudication

**Decision (`faa2a4fe7`): neither prepend half sanitizes the caller's `PATH`; the
empty-entry hardening moves to `cleanEnv`.** A synthesis — hickey's structural
call, plus his "belongs in `cleanEnv`" clause taken up rather than left optional.

- `prependPathEntries` no longer filters the caller's existing `PATH`. Its
  contract is now exactly what its name says: prepend these dirs, skip any
  already present, pass the caller's value through verbatim. It still refuses to
  add an empty string among the **dirs it was asked to add** — `PATH_REASSERT`
  already did the same, so that part was never a divergence.
- `PATH_REASSERT` is unchanged and still leaves `$PATH` byte-identical apart from
  the prepend. The two halves now implement the **same minimal rule**, which is
  the anti-drift property both lenses were actually reaching for.
- The disputed row survives in the shared oracle with an honest expectation both
  halves satisfy: `"/usr/bin::/bin"` + `["/a"]` → `"/a:/usr/bin::/bin"`. It pins
  the *preservation*, and it is asserted against the TS half and against
  `PATH_REASSERT` under real bash and zsh. Nothing is a TS-only carve-out any
  more, and no row asserts behaviour only one half has.
- `cleanEnv` gains the hardening, applied once over the composed env: an empty
  `PATH` entry is an implicit current-directory entry, a real hazard for a
  terminal kolu hosts and hands to an agent. It belongs at the env-sanitization
  boundary — the same place the `KOLU_*` strip already happens — so it runs once
  on the way in rather than being re-implemented by every downstream transform,
  and specifically not by the two prepend halves, whose contract is narrower.

Why this way round: lowy was right that the two halves must be one rule with one
oracle, and hickey was right that "prepend" should not silently mean "prepend and
rewrite your `PATH`". Both hold if the shared rule is the *narrow* one and the
security posture lives at the boundary that already owns env sanitization. The
hardening is not dropped — it is applied in the one layer where it survives to
every consumer, instead of the one layer where it survived to neither.

Doc comments on `prependPathEntries`, `PATH_REASSERT`, `PATH_PREPEND_CASES` and
`cleanEnv` were rewritten so none claims a sanitization it no longer does, and so
a reader standing at any of the three prepend sites is pointed at `cleanEnv`.

Gates: `just fmt` + `just check` green; `kolu-pty` `shell.test.ts` **59 passed**
with `KOLU_DAEMON_TESTS=1` (all five oracle rows ran under real bash **and** real
zsh — verified in verbose output, not skipped); `@kolu/padi` unit lane 453 passed
/ 481 green. Not pushed, not merged.

srid:
## Grok ⇄ Claude Code debate

✅ Consensus in 2 round(s) · base `36420cc34`

|---|---|---|
| 1 | d5a24ee76 | fix(debate): mirror the composition proof at the attr that is dialed |

- **F1** — padi-agent is what both dial paths provision, yet its postBuild only wrapPrograms padi with KOLU_AGENT_TOOLS_PATH=$out/bin and never asserts that $out/bin actually contains kolu, kaval-tui, and padi-tui (or that the wrapped padi carries that bake)
- **F2** — prepareShellInit still documents that the PATH re-assert reads $KOLU_AGENT_TOOLS_PATH (the BAKE name written only by wrappers)
- **F3** — The local composition proof checks for kolu and kaval-tui on the composed bake path, but never padi-tui
- **F4** — Product copy says that inside a kolu terminal kolu is always on PATH, and the changelog claims re-apply after dotfiles so a PATH= line cannot remove the tools
- **F5** — DialAgentOnceOptions.package is required in the type and well-commented, and runtime tests pin package≠binary resolution, but the type-level pin file that already @ts-expect-errors omitted localEnv does not also pin that omitting package is a type error

srid:

Ran with `--no-elegance` (the `/simplify` track already ran over this tree).

| Pass | Issues | Result |
|---|---|---|
| Rules | 0 | clean — every built-in and project rule checked, each "No" row backed by a grep or enumeration |
| Fact-check | 0 | clean — no silent swallowing, no unjustified fallback, no logic error, no unbounded growth |

**No fixes were needed or applied.**

Two closest calls, both recorded as non-violations: the two composed-wrapper proof blocks in `default.nix` (2 instances, under the rule-of-three threshold) and `shell.ts`'s growth (still one file, no new sibling modules).

Verified specifically:

- The PATH prepend/dedupe logic was hand-traced row-by-row across **both** implementations (TypeScript `prependPathEntries` and POSIX `PATH_REASSERT`) against the shared `PATH_PREPEND_CASES` oracle — no off-by-one or operator bug.
- The composition proofs genuinely exercise the real `makeWrapper`-generated exec line and real shell spawns — **not** mocked or vacuous.
- `package` is a required field on `dialAgentOnce`, never defaulted to `binary`, pinned by a `@ts-expect-error` case — the silent two-closures regression cannot return without going red.

srid:
## Evidence

Verified end-to-end on a **real remote host** (`kolu-ci-5`, provisioned cold over ssh by `padi-tui --host`, which after this change resolves the same `padi-agent` closure the browser binder does). This is the only proof that matters for this PR — everything else is unit tests and build-time assertions, and none of those has ever put a shell on another machine and asked whether `kolu` answers.

### The tools resolve, and `kolu` actually runs

```
-- command -v --
/nix/store/1qxs1ybydmgrncn3fp350zifv9v6b7fh-padi-agent/bin/kolu
/nix/store/1qxs1ybydmgrncn3fp350zifv9v6b7fh-padi-agent/bin/kaval-tui
/nix/store/1qxs1ybydmgrncn3fp350zifv9v6b7fh-padi-agent/bin/padi-tui
cv_exit=0
-- vars --
PADI_SOCKET=/run/user/1000/padi-d765be621f53fc73/padi.sock
TOOLS=/nix/store/1qxs1ybydmgrncn3fp350zifv9v6b7fh-padi-agent/bin
AGENT_TOOLS=
-- kolu runs --
kolu mcp v2.0.0
...
kolu_exit=0
```

All three resolve **inside the same provisioned closure as the daemon serving the terminal** — the store path matches the one ssh realised — so tool and daemon are the same build by construction, not by convention. `kolu mcp --help` returned `0`, so the MCP face is reachable, not merely present on `PATH`.

`AGENT_TOOLS` (`KOLU_AGENT_TOOLS_PATH`) being **empty is the design working**: the bake name is deliberately kept out of terminal envs so a kolu started inside a terminal cannot inherit a foreign build's toolchain as its own. Only the stamp (`KOLU_TERMINAL_TOOLS_PATH`) is present.

### The hostile-dotfile case, actually exercised

The box had no dotfiles, so a real one was installed — `export PATH=/usr/bin:/bin`, gated to interactive shells — and a default-shell terminal created:

```
HOSTILE_RC_ACTIVE=bash: ls: command not found
PATH=/nix/store/1qxs...-padi-agent/bin:/usr/bin:/bin
TOOLS=/nix/store/1qxs...-padi-agent/bin
/nix/store/1qxs...-padi-agent/bin/kolu
/nix/store/1qxs...-padi-agent/bin/kaval-tui
/nix/store/1qxs...-padi-agent/bin/padi-tui
cv_exit=0
```

The clobber demonstrably landed — `ls` is "command not found", because nothing lives in `/usr/bin` on NixOS — and the toolchain still resolves, prepended ahead of it. That is the rcfile re-assert arm working against a real hostile dotfile on a real box, which is the defect class this PR exists to close.

### Stated plainly: what this run did NOT prove

- **No binary-cache hit for `padi-agent`.** Nothing has built the new attr yet, so the box realised the whole closure from source (~10 min). Expect that cost for first provisions until CI has built this attr on master and pushed it; steady-state provisions substitute as before.
- **Only bash was exercised remotely.** The zsh arm of the re-assert is covered by the real-zsh tests locally, not on this box.
- **`kaval-tui --host <box> snapshot <id>` does not read padi's terminals** — it provisions its own `kaval` attr and starts a separate daemon, so it reports no terminals while padi's are alive on the digest-keyed socket. Pre-existing behaviour, not introduced here, but it makes that route in `packages/kaval-tui/README.md` misleading for padi-created terminals. Flagged, not fixed in this PR.

### Teardown

Both terminals killed **by id**; no substring matching anywhere. Three daemons remain on `kolu-ci-5` by design (a padi, its kaval, and one standalone kaval from a `kaval-tui --host` call), plus a GC root at `~/.local/state/kolu/surface-remote/gcroots/padi-agent`. The temporary dotfile was removed and `nix-instantiate` re-confirmed working.

PR #2067, opened 2026-07-30
Adopt osfacts for padi memory sampling

srid:
**Move padi's process-memory sampler onto the baked osfacts contract and remove an RPC that no longer belongs on kaval's wire.** One scoped `snapshot --pids … --mem` call now reads padi plus its connected kaval every five seconds. Readable RSS is `ok`; a disconnected, exited, or superseded kaval generation is honestly `absent`; unreadable, source, malformed, and missing requested facts become the explicit typed `error` arm instead of leaving stale numbers painted.

The recycle boundary is endpoint-owned. The sampler captures an immutable `{ pid, startedAt }` target from the endpoint's own live connection, holds it across the asynchronous osfacts read, and verifies both fields against that same authority before publishing. It never reconstructs identity from daemon status, the socket, or the supervisor gate. The race test pauses the real fake-osfacts process across a same-pid/new-`startedAt` recycle; removing the post-read check makes it publish the prior generation and turns the test red. Separate executed mutations cover the catch path and prove that comparing pid alone is insufficient.

### Contract and wire

Deleting `system.processMemory` is breaking in the old-client/new-daemon direction, so `PTY_HOST_CONTRACT_VERSION` advances from 5.3 to 6.0 and rejects both skew directions.

There is otherwise **zero wire change**: `system.version` is byte-identical, pid is retained only in padi-internal `KavalConnectionMetadata`, and `DaemonStatus` is unchanged. The exact `system.version` shape test remains green.

Production gained no sampler override or fallback knob. Tests exercise the real `osfacts-client` spawn and V2 parser through a temporary executable, using the canonical `@kolu/shell-quote` helper and scoped environment/file cleanup.

No API-facing file under `surface`, `surface-app`, `surface-remote`, `surface-daemon`, or `surface-daemon-supervisor` changed, so the drishti pair rule does not apply. At odu's kolu-pinned revision `fe0e0fc21ff639974868d410ce8d9421fdfaa467`, a full-tree grep for the affected sampler, osfacts, and pty-host contract terms found no consumer or compatibility impact.

### UI and docs survey

No UI gap was found, so this PR deliberately adds no new screen. Every RSS fact Kolu already holds is painted in host-chip tooltips, the Padi/Kaval info dialogs, and diagnostics; daemon uptime already uses each daemon's boot timestamp. CPU, uid, cwd, status, argv, and host telemetry are not held by Kolu's UI stack today.

[...]

PR #2074, opened 2026-07-30
ci: run the full e2e suite on GitHub-hosted runners

srid:
The Cucumber e2e suite currently has exactly one home: the self-hosted odu venue pool. That's fine until a pool box is wedged, leased by another PR, or simply not there — at which point the repo's most expensive signal is unavailable and there's no second opinion on whether a failure is the code or the box. This adds a vendor-neutral second home on GitHub-hosted runners, covering the same two platforms odu gates on: `ubuntu-latest` (x86_64-linux) and `macos-latest` (aarch64-darwin).

The workflow deliberately reimplements nothing. It provisions Nix, then calls the two recipes odu itself calls:

```
nix develop -c just --no-deps ci::install
nix develop -c just --no-deps ci::e2e
```

`ci::e2e` is unchanged — it still nix-builds `.#koluBin` and runs Cucumber inside the `.#e2e` devshell under the platform `CUCUMBER_RETRY` policy (darwin 2, linux 1). So the justfile stays the single source of truth for what "full e2e" means, and this file can't drift into a parallel definition of the suite.

Nix comes from the same pinned installer and Attic cache wiring as `nix-cache.yml`, which already pushes both platforms' closures to `cache.nixos.asia/oss` — so the `koluBin` build substitutes rather than compiles. The Linux job reclaims disk first; the hosted x86 image ships ~14 GB free on `/`, which the devshells plus browsers plus `koluBin` don't comfortably fit beside preinstalled toolchains this repo never touches.

### Draft PRs are skipped

`if: github.event.pull_request.draft != true` on the job, paired with `ready_for_review` in the trigger list. The trigger matters as much as the guard: without it, flipping a draft to ready emits no event and the PR would never get a run. `!= true` rather than `== false` because `github.event.pull_request` is null on `push` and `workflow_dispatch`, and those runs should stay alive.

`concurrency` cancels in-flight runs for the same ref — this job is measured in tens of minutes, so queued duplicates are pure waste. `timeout-minutes: 120` caps a hung suite well under GitHub's 360-minute default.

### What this is not

Not a replacement for odu, and not proposed as a required status check. odu remains the merge gate; this is a redundant, independently-hosted read on the same suite. Whether hosted-runner wall-clock is good enough to lean on is the open question — the run on this PR is the measurement, and I'll report the timings for both platforms here once it lands.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,0 +1,79 @@
+# Run the FULL Cucumber e2e suite on GitHub-hosted runners, on both platforms
+# odu already gates on (x86_64-linux + aarch64-darwin), so the suite has a
+# second, vendor-neutral home besides the self-hosted venue pool.
+#
+# This is the same node the odu pipeline calls `ci::e2e`: `just --no-deps test`
+# under the repo's CUCUMBER_RETRY policy, which nix-builds `.#koluBin` and runs
+# Cucumber inside the `.#e2e` devshell (Playwright browsers). Nothing is
+# reimplemented here — the workflow only provisions Nix and calls the recipe,
+# so the justfile stays the single source of truth for what "full e2e" means.
+name: e2e
+
+on:
+  push:
+    branches: [master]
+  # `ready_for_review` is what makes the draft gate below usable: without it a
+  # PR that flips draft → ready emits no event and would never get a run.
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+# A new push supersedes the in-flight suite for the same ref — this job is
+# measured in tens of minutes, so queued duplicates are pure waste.
+concurrency:
+  group: e2e-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  e2e:
+    # Draft PRs skip the suite. `github.event.pull_request` is null for push /
+    # workflow_dispatch, so `!= true` (not `== false`) is what keeps those
+    # non-PR runs alive.
+    if: github.event.pull_request.draft != true
+    strategy:
+      fail-fast: false
+      matrix:
+        # ubuntu-latest → x86_64-linux; macos-latest → aarch64-darwin.
+        os: [ubuntu-latest, macos-latest]
+    runs-on: ${{ matrix.os }}
+    # GitHub's default is 360; cap well under that so a hung suite fails in
+    # bounded time instead of holding a runner for six hours.
+    timeout-minutes: 120
+    steps:
+      - uses: actions/checkout@v4
+
+      # Same pinned installer and cache wiring as nix-cache.yml — that workflow
+      # pushes both platforms' closures to this Attic cache, so the koluBin
+      # build below mostly substitutes instead of compiling.
+      - uses: nixbuild/nix-quick-install-action@9f63be77f412a248c9d9a65a4c82cf066cdf8f0c # v35
+        with:
+          nix_conf: |
+            extra-substituters = https://cache.nixos.asia/oss
+            extra-trusted-public-keys = oss:KO872wNJkCDgmGN3xy9dT89WAhvv13EiKncTtHDItVU=
+            accept-flake-config = true
+
+      # The hosted x86 image ships ~14 GB free on /; the Nix store for the
+      # devshells + koluBin + Playwright browsers does not comfortably fit
+      # beside preinstalled toolchains we never use.
+      - name: Free disk space (linux)
+        if: runner.os == 'Linux'
+        run: |
+          sudo rm -rf /usr/share/dotnet /usr/local/lib/android /opt/ghc \
+                      /usr/local/share/boost "$AGENT_TOOLSDIRECTORY"
+          df -h /
+
+      # The single workspace `pnpm install`, exactly as the odu `ci::install`
+      # node does it — `ci::e2e` and the `test` recipe both assume the
+      # workspace is already installed and run with `--no-deps` so no second
+      # concurrent install can corrupt node_modules/.bin.
+      - name: Install workspace
+        run: nix develop --accept-flake-config -c just --no-deps ci::install
+
+      # The suite. `--no-deps` because `ci::e2e`'s `install`/`nix` deps are
+      # provisioned above (install) or unnecessary here (the full Nix gate is
+      # odu's job; `test` builds `.#koluBin` on its own).
+      - name: Full e2e suite
+        run: nix develop --accept-flake-config -c just --no-deps ci::e2e
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -20,6 +20,11 @@ on:
 
 permissions:
   contents: read
+  # REQUIRED by cache-nix-action's `purge: true` — purging stale entries goes
+  # through the REST cache API, and without this the save aborts with
+  # "Resource not accessible by integration" and no nix-* cache is ever
+  # created. The failure is silent: save/restore both "succeed" in ~1s.
+  actions: write
 
 # A new push supersedes the in-flight suite for the same ref — this job is
 # measured in tens of minutes, so queued duplicates are pure waste.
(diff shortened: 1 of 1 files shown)


srid:
## Timing report — first run, cold (no store cache)

[Run 30565052659](https://github.com/juspay/kolu/actions/runs/30565052659), commit `21623a18`.

| | `ubuntu-latest` | `macos-latest` |
|---|---|---|
| **Total job** | **19m 22s** | **21m 35s** |
| `ci::install` (pnpm) | 1m 06s | 1m 56s |
| `ci::e2e` step | 17m 46s | 19m 28s |
| Cucumber suite (internal timer) | 17m 26s | 18m 49s |
| Workers | 2 (`cores=4 cap=8`) | **1** (`cores=3 cap=6`) |
| Scenarios | 504 — 503 pass, 1 fail | 503 — all pass |
| Attempts / retries | 505 / 1 | 504 / 1 |
| Outcome | known flake (below) | :white_check_mark: green |

Scenario counts differ by one because `@skip-darwin` prunes `git-context.feature:59` on macOS.

### Nix is not the bottleneck — Cucumber is ~92% of the job

Worth recording, because it is the opposite of what the setup guidance predicts. `ci::install` takes about a minute, and inside the e2e step `nix build .#koluBin` completes in **~14s** — the log reaches `e2e: workers=2` fourteen seconds after the step starts. Everything Nix-shaped totals under two minutes on either platform.

The reason is that this repo already has a binary cache: the Attic substituter declared in `nixConfig`, kept warm for both platforms by `nix-cache.yml`. The usual "cache `/nix/store` or every run rebuilds the closure for 30 minutes" advice assumes no such cache. Here the remaining Nix work is ~9% of wall-clock, and a multi-GB store restore can plausibly cost as much as it saves. A `cache-nix-action` commit follows this one so the effect is measured rather than assumed, but the ceiling on that optimisation is roughly two minutes.

### The suite is latency-bound, not CPU-bound

The most useful number here is the comparison *between* the two platforms. One worker on a 3-core macOS runner (18m 49s) is within 8% of two workers on a 4-core Linux runner (17m 26s). Doubling worker count bought almost nothing.

4,606 steps in ~1,130s averages ~245ms per step, and the suite's time goes into `waitFor` polling, tree hydration, and content settles — waiting, not computing. That means the `ceil(cores/3)` sizing in the `test` recipe, calibrated on 10-24 core pool boxes where each worker really does own a browser + server + padi + kaval, is likely too conservative for a 3-4 core runner that spends most of its time blocked on I/O. The justfile's "PAR=12 measured slower than PAR=8" note is from a different regime (a saturated 24-core box), and shouldn't be read as evidence against raising workers here.

Two candidate optimisations, in order of expected payoff:

1. **Shard across runners.** `code-tab.feature` alone is 100 of 495 scenarios (20%), so file-level sharding floors around 4 minutes of suite time; four shards per platform should land each job near 6 minutes wall-clock. Requires `just test` to accept feature-path arguments, which it currently does not — `runE2e.ts` already forwards `argv` to cucumber-js, and `cucumber.js` is already written to let CLI feature args win over the profile's paths, so the change is small and backward compatible.
2. **Raise workers per runner.** Cheaper to try and needs no sharding machinery, but it means changing hardware-derived sizing that the self-hosted pool also uses, so it wants its own measurement on a pool box before landing.

### The single failure is a known open flake, not a workflow problem

`code-tab.feature:800` — "Browse mode tints ancestor folders that contain a change" — failed both attempts (`CUCUMBER_RETRY=1` on Linux), alternately missing `seed` and then `src` inside the 60s nudge window.

That is row 40 of the flaky-test tracker verbatim: open against `e2e@x86_64-linux`, tracked as **#2065**, reopened at `185e54e` (#2067) — which is the commit at the tip of `master` this branch is based on. The tracker's own description is "alternately failing to mark `seed` or `src` within 60 s despite the nudge", which is exactly the observed fingerprint.

So the hosted runner reproduced an existing pool flake rather than introducing a new failure mode. That is a mild point in favour of what this PR is for: an independently-hosted second read that agrees with the pool. I have not touched the tracker — adding a datapoint row would drag `atlas::check-sync` into a CI-only PR — but the recurrence is worth folding into #2065 separately.

srid:
## Five consecutive runs — 10 job executions

Run [30566868440](https://github.com/juspay/kolu/actions/runs/30566868440), attempts 3-7, all against a warm Nix store cache. Ubuntu and macOS execute in parallel within each iteration; the iterations are sequential because `cancel-in-progress` would otherwise have each one kill its predecessor.

| Iteration | `ubuntu-latest` | `macos-latest` | Failing scenario |
|---|---|---|---|
| 1 | :x: 19m 41s | :white_check_mark: 25m 44s | `code-tab.feature:800` |
| 2 | :white_check_mark: 20m 27s | :white_check_mark: 21m 17s | — |
| 3 | :x: 21m 55s | :x: 25m 38s | `code-tab.feature:800` · `file-ref-link.feature:9` |
| 4 | :x: 19m 22s | :white_check_mark: 23m 05s | `code-tab.feature:800` |
| 5 | :white_check_mark: 19m 57s | :white_check_mark: 22m 20s | — |

**Duration** is remarkably stable. Linux spans 19m 22s to 21m 55s (median 19m 57s); macOS spans 21m 17s to 25m 44s (median 23m 05s). Because the platforms run concurrently, a run's wall-clock is the macOS number — call it **~23 minutes**, and macOS is the critical path every single time.

**Reliability is the problem.** Linux went red 3 of 5, macOS 1 of 5; only **2 of 5 runs were fully green**. Across all 16 job executions in this PR the combined red rate is 37.5%.

### Every failure is an already-tracked flake

Not one novel failure in 16 executions, and the failures are cleanly stratified by platform:

- **Linux fails only `code-tab.feature:800`** — "Browse mode tints ancestor folders that contain a change". Five of five Linux reds, four with a byte-identical fingerprint: `Directory "src" was not marked as containing a change within ~60s while nudging …`, then `"seed"` on the retry. That is flaky-tracker row 40 verbatim — open, tracked as **#2065**, reopened at `185e54e` (#2067), which is the commit at the tip of `master` this branch is based on.
- **macOS fails only `file-ref-link.feature`** — `:166` and `:9`, against the tracker's open darwin row for `:154`. Same view-hydration class, different lines.

The useful and slightly unexpected result: a 4-core hosted runner reproduces #2065 **far more reliably than the venue pool does** — 5-in-8 here versus the occasional recurrence the tracker records. The tracker already reads this class as box-capacity-correlated ("the durable fix is running e2e on a free pool box … not hardening the locator"), and these runs are the cleanest confirmation of that reading yet. If anyone wants a reproducer for #2065, `ubuntu-latest` is a better one than a pool box.

### Verdict

**Do not put this in `required_status_checks`.** At a 37.5% per-job red rate it would block merges roughly two runs in three, entirely on known flakes. It is worth keeping as a non-blocking second opinion — vendor-neutral, always available, and demonstrably faithful (it reproduces the pool's flakes rather than inventing its own).

Two follow-ups worth separating from this PR:

1. **Fix or quarantine #2065.** It is the single highest-value fix here: it alone accounts for every Linux red, so fixing it takes Linux from 60% red to 0% across this sample.
2. **Shard to cut wall-clock.** The suite is latency-bound, not CPU-bound — 1 worker on 3-core macOS (18m 49s) matched 2 workers on 4-core Linux (17m 26s) in the cold run. So more *runners* help where more *workers* would not. `code-tab.feature` is 100 of 495 scenarios, which floors file-level sharding near 4 minutes of suite time; 4 shards per platform should land each job around 6 minutes. It needs `just test` to accept feature-path arguments, which touches a recipe odu also runs, so it belongs in its own PR.

### The store cache does not pay for itself

Measured rather than assumed, and the answer is no:

| | restore cost | `ci::install` saved | net |
|---|---|---|---|
| `ubuntu-latest` | 42s | 1m 06s → 18s | **+6s** |
| `macos-latest` | 1m 44s | 1m 56s → 27s | **−15s** |

Both are inside the noise — the `Free disk space` step alone swung 26s → 1m 22s across runs. The reason is that this repo already has a binary cache: the Attic substituter in `nixConfig`, kept warm for both platforms by `nix-cache.yml`. Total Nix work is under two minutes of a ~23 minute job, so there was never more than ~2 min available to win, and `nix build .#koluBin` already completes in **14 seconds**.

Carrying it costs two extra pinned actions, an `actions: write` permission, and 2.45 GiB of the repo's 10 GB cache budget. **Recommend reverting `e5c70e0d`** unless someone wants the cache as insurance for the day the Attic cache is cold or unreachable — which is a real but different argument from speed.

PR #2075, opened 2026-07-30
Bind a terminal's screen bytes to the grid they were drawn for

srid:
Switching to a split often showed a stale, garbled screen — long lines broken in half, the live prompt scrolled out of sight — most visibly with a full-screen program like an agent, `htop`, or `btop`. Nudging the divider a pixel was the only repair. **A terminal's screen is bytes laid out for one specific cols×rows, but the bytes and the size travelled as two independent facts that agreed only by luck of ordering.** This binds them together.

The snapshot came over `terminalAttach`; the size went out separately through `lifecycle.resize`. Nothing made them agree. Splits are where the luck ran out: every sub-terminal mounts immediately but only one is shown, so a split hidden at page load (collapsed panel, background sub-tab) has a 0×0 box, is never measured, and keeps reporting the **80×24 that xterm's constructor invents**. It attached at that invented grid and painted a snapshot laid out for the real width into 80 columns.

The reason it never healed is the interesting half. Revealing the pane *reflowed* the damage instead of repainting it, and the grid the client then published was the one the PTY already had — and kaval correctly no-ops a same-dimensions resize, so no `SIGWINCH` ever reached the process. There was no path from a bad screen back to a good one except a real size change.

### Two changes, so the mismatch stops being expressible

| | Before | After |
| --- | --- | --- |
| **Unmeasured pane** | attaches at the invented 80×24 and paints | `<Xterm>` publishes a `grid` fact that is `null` until the pane has actually measured its box; **no grid, no bytes** |
| **Snapshot sizing** | serialized at whatever size the PTY happened to hold | grid rides **on** the attach request; the host resizes and serializes as *one act* |

The second is what makes it structural rather than merely ordered — a request cannot carry a size that was never measured, so "a snapshot for a size the consumer isn't" is unrepresentable instead of unlikely. The resize goes through the existing `resize()`, so same-dimensions stays a no-op and a genuine change reflows, invalidates the snapshot memo, and signals the process exactly as a user resize does.

`kaval-tui` had the same defect held off by hand — a comment calling resize-then-attach a "design decision" and documenting the wrap artifacts it left behind. It now passes its tty's grid on the attach, and the two calls collapse into one.

> The grid is **optional** on both contracts and deliberately carries **no version bump**. Its absence degrades to exactly the previous reading in both skew directions, which is the `commandRooted`/`shellJoin` class both contracts already document as no-bump. A bump would force-recycle a surviving daemon — killing the user's live PTYs — to buy a graceful improvement.

### Reproduction

The e2e scenario asserts on the **viewport**, not the buffer: this defect delivers every byte correctly and then shows the wrong window onto them, so a whole-buffer read passes on a screen the user sees as broken. It was red 4/4 before the fix (3 solo runs + a full-feature parallel run), always with the same geometry — `viewportY 59` against `baseY 77`.

Also threads `KOLU_OSFACTS_BIN` through the e2e server child's env allowlist. It is the same class as the `KOLU_KAVAL_BIN` entry beside it — a nix-baked absolute path with no PATH fallback by design — and without it every worker's server died at `claimLocalSupervisor` before binding its port, which had blocked the e2e lane entirely.

### Try it locally

```sh
nix run github:juspay/kolu/resize-bug
```


srid:

Two structural lenses read `36420cc34..HEAD` independently — **lowy** (volatility-based decomposition) and **hickey** (structural simplicity) — then one pass merged and applied them.

**Status: applied and committed on `resize-bug`. Not pushed, not merged.**
`just check` (tsc + biome) green · `just fmt` clean · unit tests green for `@kolu/xterm-kit`, `kolu-client`, `kaval`, `@kolu/padi`, `kaval-tui`, `kolu-common`, `@kolu/surface` · `just atlas::build` re-run so `docs/atlas/dist` stays in sync.

| | count |
| --- | --- |
| lowy findings | 14 (F1–F14) |
| hickey findings | 13 (H1–H13) |
| **applied** | **24** |
| observation | 3 |
| routed to a human | 0 |

Seven findings were the same finding twice: F1≡F2, F4≡H6, F6≡H3, F7≡H8, F11≡H11, plus H4⊃F5 and F12⊂H2. Each was applied once, under one plan.

### Commits

| SHA | Area |
| --- | --- |
| `451155800` | `fix(lens): the measured-grid gate` |
| `aaa4727d8` | `fix(lens): the attach request` |
| `dfc6adafd` | `fix(lens): the captured-input hazard` |
| `6a2ab11ac` | `fix(lens): minor polish from the lens review` |

### The correctness-adjacent ones

Both lenses converged on these, and they are the reason the review was worth running.

- **The "never a fabricated grid" invariant did not hold.** `applyFit`'s guard blocked a 0×0 box but not `@xterm/addon-fit`'s **2×1 clamp** on a sliver box — a split dragged nearly closed, a panel mid-collapse. That published a clamped floor as a genuine measurement, and since the attach now resizes, it would have sized a live PTY to 2×1. lowy added that the guard read the *visual* (transform-scaled) rect while the addon reads the *layout* box — two different facts under the canvas's CSS scale. Both fixed: same quantity, and a clamped proposal is refused like a zero box. *(lowy F1 ≡ F2)*
- **The fail-loud guard did not fail loud.** The absent-grid `throw` sits inside `consumeReattachingStream`'s thunk, whose catch classifies it as an abnormal end, warns, and retries every 300 ms — running `resetForFreshSnapshot`, which **clears the user's screen**, on every cycle. The code comment claiming otherwise was false. It now toasts and aborts the stream before throwing, so the loop exits after one pass. *(lowy F7 ≡ hickey H8)*
- **The only repair for a replayed stale grid rode a bare `catch {}`.** `reassertGrid` → `publishDimensions` swallowed exactly the failure that leaves the user with the stuck garbled screen this PR exists to fix. It surfaces now. *(lowy F8)*
- **Two disagreeing definitions of "a grid", six lines apart, feeding one mutator.** `TerminalGridSchema` required `.int().positive()`; `PadiResizeInputSchema` took bare `z.number()`. Each surface now has one named grid schema both members extend. This *tightens* `lifecycle.resize` — every newly-rejected value was already rejected by kaval one hop later, so no working call changes behaviour; the loud failure just moves to the boundary that owns the concept. *(lowy F6 ≡ hickey H3)*
- **`attach()` was spelled as a read while performing a write.** `grid` is a noun that reads as a description of the caller; the call does a real `resize()` — SIGWINCH, shared-mirror reflow, activity mute, and on a width change a reflow-epoch bump that stales *every other* attached client's backfill cursor. The fusion is right (a snapshot is bytes laid out for one cols×rows); the spelling was not. Renamed to **`resizeTo`** at every layer, with the cross-client effect documented where the verb is declared. *(hickey H1)*

### Also applied

The kit now owns the wait-for-measurement latch (`XtermHandle.onceMeasured`) instead of the consumer re-deriving it from a nullable accessor and a mutable boolean *(lowy F3, hickey H9)*; `visible` is no longer a second, weaker predicate for a fact the box answers directly *(lowy F4 ≡ hickey H6)*; the measured grid leaves `<Xterm>` through **one** door, deleting the `onResize` callback and its double-publish compensation *(hickey H4 ⊃ lowy F5)*; the new "measured-not-yet" state is named in diagnostics, since on screen it looks exactly like an idle terminal *(hickey H7)*; oRPC's captured-input replay hazard is recorded at the receptacle rather than only at the site compensating for it *(lowy F10)*; plus the doc/test polish — honest "unrepresentable" prose, a re-attach-is-a-resize note, the ledger row in `.agency/lowy.md`, the nominal `TerminalGrid` at the client, and the deduped viewport wait helper *(lowy F9, F12, F13, F14 · hickey H10, H12)*.

### Recorded as observations, not applied

**hickey H2 — "the `{cols,rows}` grid is reinvented five times."** Settled by experiment in lowy's favour, and not relitigated: `packages/padi/src/endpoint.ts` imports **zero** kaval symbols and already hand-declares `PtySpawnOpts`, a kaval-shaped concept, by exactly the rule H2 attacks. The per-layer duplication is the repo's established convention, and `xterm-kit` may depend on neither package. H2's uncontested sub-point — the *client* should use the kit's exported `TerminalGrid` — was applied.

**hickey H5 — `retry: false` to collapse the two re-attach mechanisms and delete `reassertGrid`.** Not taken in this PR: `.claude/rules/streaming.md` §1 mandates `STREAM_RETRY` on every client stream, and disabling it is a behavioural change to the terminal reconnect path (#1591) that this PR must not smuggle in. The diagnosis is right, though — `reassertGrid` exists only to paper over two nested re-attach mechanisms with different staleness semantics — so **this is a follow-up worth considering**, most likely via hickey's second option (a one-shot `terminalAttach.open({id, resizeTo})` handle plus a grid-free stream), which deletes the compensation without touching the retry policy.

**lowy F11 ≡ hickey H11 — the e2e env allowlist hand-duplicates `nix/env.nix`.** Real duplication; the proposed mechanism would be worse than it. The source of truth is Nix and cannot import a TS constant, so a shared `BAKED_BIN_ENV_KEYS` is a *third* list — and a misleading one, since `nix/env.nix` also bakes `KOLU_GH_BIN`, `KOLU_FONTS_DIR` and `KOLU_RHAI_GRAMMAR`. Routing ~10 production spellings through `packages/common` would add a cross-package dependency to carry a string literal (a name, not a volatility). A genuine link would have to *generate* the TS list from `nix/env.nix` at build time — over-engineering for two keys. The commented allowlist entry the diff added is the better artifact.

**hickey H13 — `opts.viewport` as a boolean.** hickey's own disposition was `drop`, and it holds: a boolean with two documented cases, threaded through a stringified `addInitScript`, is not worth the churn. The Playwright-side duplication beside it *was* fixed (H10).

### Routed to the human

None — no finding needed a judgment call the adjudications had not already settled.

srid:
## Codex ⇄ Claude Code debate

✅ Consensus in 5 round(s)

|---|---|---|
| 1 | c7e212f19 | fix(debate): refuse a snapshot that answers a grid the pane no longer has |
| 2 | ce8fbe6c9 | fix(debate): re-check the grid where the snapshot bytes actually land |
| 3 | 881148323 | fix(debate): make a superseded attach attempt inert instead of racing |
| 4 | aec0c1ca6 | fix(debate): route every attempt-owned effect through the supersession gate |

- **F1** — The request captures one measured grid, but the snapshot is always painted even if h.grid() changed while the request was in flight
- **F2** — attach applies resizeTo before checking signal.aborted
- **F3** — publishDimensions appears to await and surface resize failure, but the padi resize procedure returns as soon as PtyHostTerminalProxy.resize queues a detached promise; the real kaval RPC rejection is caught and reduced to a server log at packages/padi/src/terminalEndpoint/local.ts:204-209
- **F4** — The regression scenario assumes the visible split's real width differs from xterm's fabricated 80 columns, but never asserts or arranges that precondition
- **F5** — This is a material user-facing change to split rendering and kaval-tui attach, and the new changelog entry is explicitly under Inside a Tile, but no canonical product-doc page changed

srid:

Rules + fact-check passes (elegance skipped — `/simplify` ran over this same tree). **4 findings, all applied** in `a442170cd`.

| Pass | Finding | Action |
|---|---|---|
| Fact-check | `resize` reported `{ ok: true }` even when the PTY entry was gone, so the one failure this branch's new end-to-end acknowledgement was built to surface could never reach anyone — and `ok: boolean` was a field that never varied | `resize` now returns whether the entry existed; padi reads it and returns **quietly** on `false` (the expected killed-terminal race, not an error to toast). Rejections still propagate. Comments corrected to claim only what holds |
| `dry-rule-of-three` | Two unrelated `{cols, rows}` types both named `TerminalGrid` — one grep returned both | Renamed padi's to `EndpointGrid`. They stay separate deliberately: padi's endpoint imports zero kaval/kit symbols by design and already hand-declares `PtySpawnOpts` by the same rule |
| `no-thin-wrapper-functions` | `waitForViewportContains` was the rule's "Bad" example verbatim, with 2 callers (below the 3-caller exception) | Deleted; both sites pass `viewport: true` directly. Its load-bearing note — a buffer read answers *did it arrive*, a viewport read answers *does the user see it* — moved onto that option's doc |
| `e2e-poll-async-state` | The precondition assertion read `__xterm.cols` with a bare `$eval` and asserted immediately, racing the ResizeObserver + rAF-debounced fit | Polls now, so a loaded CI box reads a measured grid rather than a pre-fit zero |

Clean on the other 30 rules, including `no-vacuous-assertion` (both load-bearing assertions traced to real, variable subjects), `caught-error-must-not-collapse-to-empty`, and `no-overloaded-null`.

srid:
## Evidence

The split now renders correctly the moment it is revealed — no divider nudge.

split renders correctly on reveal

Captured by driving a live kolu at this branch's HEAD through the exact sequence that used to break it: create a split, fill it with lines wider than xterm's fabricated 80 columns, **collapse the panel**, **reload**, then reveal.

The instrumented facts at each step, read off the live `__xterm`:

| Step | Before this PR | At this HEAD |
|---|---|---|
| Hidden split after reload | 80×24 (invented), **84 buffer rows of wrapped, mangled content** | 80×24 (invented), **24 blank rows — nothing painted** |
| After revealing | `viewportY 59` vs `baseY 77` — stuck 18 rows above its own live bottom, showing L30–L33 | `viewportY 36 === baseY 36` — showing L37–L40, the marker, and the live prompt |
| Repair needed | drag the divider (a real size change → SIGWINCH) | none |

The middle row is the whole bug: every byte arrived correctly and the pane showed the wrong window onto them, which is why the regression scenario asserts on the **viewport** rather than the buffer — a whole-buffer read passes on a screen the user sees as broken.

A recording of the regression scenario itself (`features/sub-terminal.feature:131`), captured via the e2e harness at this HEAD: [pr2075-split-scenario.mp4](https://github.com/juspay/kolu/releases/download/evidence-assets/pr2075-split-scenario.mp4). It was **red 4/4 before the fix** — three solo runs plus a full-feature parallel run, always with the same geometry — and green after.

**Mobile / coarse pointer.** This change alters *what a pane renders*, not layout or hit targets — no sizing is gated on `isTouch()` / `(pointer: coarse)` anywhere in the diff, and no markup or styling changed. The mobile path is covered by `features/mobile-sub-terminal.feature` running green, and mobile terminals benefit from the same gate: a background mobile tile now costs zero host serialize until it is first shown.

PR #2064, opened 2026-07-30
Graduate redundant E2E coverage to faster test layers

srid:
Kolu's browser suite now does less duplicate work while preserving every distinct user-facing journey. **This reduces 546 expanded E2E executions to 508 without changing product behavior or weakening feature-level coverage.**

The work began with a full census, then moved only assertions that already had a faster honest home. Every retirement is tied to a lower-layer replacement where applicable and an independent surviving browser journey; Grok/Claude debate and three adversarial perfection passes restored six cases whose browser seam still carried a distinct fact.

### Final census

| Suite shape | Original | Final | Change |
| --- | ---: | ---: | ---: |
| Gherkin declarations | 530 | 494 | -36 |
| Expanded executions | 546 | 508 | -38 |
| Linux default | 541 | 503 | -38 |
| Darwin default | 540 | 502 | -38 |

**Darwin browser load falls by 7.04%.** The retained suite still covers the complete user path at integration boundaries; lower-layer tests take over duplicated policy, parsing, state-fold, and rendering assertions where a browser adds no new evidence.

### Coverage guardrails

- An append-only inventory now fingerprints every historical scenario revision.
- The ledger records each removed/revised promise, defect surface, current browser survivor, and collected replacement evidence.
- CI checks inventory completeness, current survivors, test collection, and platform declarations. It deliberately does not claim semantic equivalence; that remains an explicit human review judgment.
- Fixed Git fixtures are rerunnable, command setup is status-strict, and the Nix E2E environment uses the intended Git on macOS.
- The transformed-touch journey is mutation-proven: bypassing inverse coordinate correction resolves `[32,9]` instead of `[20,6]` and fails, while the real CDP touch still proves drawer navigation and keyboard-focus behavior.

### Runtime evidence

The original 540-execution macOS suite at `f16eb26` did not settle within Odu's one-hour cap because the old harness exposed Apple's Xcode Git stub inside PTYs, so the baseline is honestly reported as censored (`>60 min`), not as a fabricated median. After repairing the harness, the final 502-execution strict suite at `71dd255` passed on the exact `ci@petit` venue in **198.109 seconds**.

The final commit `d3e5e9d` passed the complete strict pipeline on both required venues (`kolu-ci-1` Linux and `ci@petit` Darwin): all E2E, governance, unit, daemon, smoke, Nix, Atlas, formatting, and documentation checks are green. The full-pipeline E2E recipe wall times were 199.186 seconds on Linux and 272.383 seconds on Darwin.

[...]

PR #2080, opened 2026-07-30
Stop a rebuilt file watcher from being born deaf

### LINKED ISSUE #2065: e2e flake: code-tab.feature:800 Browse mode tints ancestor folders (linux, workers=8)

srid:
The Code tab's git status could silently stop updating for a repo — permanently — while everything else on the page stayed healthy. That is [#2065](https://github.com/juspay/kolu/issues/2065): `code-tab.feature:800` times out for ~60s on linux under an 8-worker suite, with the tree fully rendered, the terminal beside it showing `master [!]`, and the file row carrying no status at all.

The instrumented evidence on the issue was right about where it *wasn't* — not the roll-up, not Pierre, not a repaint. The status never reached the client because **the server never noticed the edit**. Two independent `@parcel/watcher` hazards, both ending in the same symptom: a subscription that resolves, holds its callbacks, logs `working-tree watcher installed`, and receives nothing.

## 1. A new directory is watched, but never scanned

`InotifyBackend::handleSubscription` calls `watchDir` on `IN_CREATE`, which adds the inotify watch and returns. It does **not** scan what is already inside. Anything created before that watch existed generated its create event against a watch that didn't exist yet, and nothing ever rescans — so it is invisible to that subscription **forever**.

`mkdir -p src/feature` is the entire reproduction. `src` is created, then `src/feature` microseconds later — long before parcel's poll thread reaches the `src` event and adds its watch. Measured with the scenario's exact shell sequence:

| subscribe→`mkdir` delay | `src/feature` blind | `seed/` blind |
| --- | ---: | ---: |
| 0 ms | 6/6 | 0/6 |
| 5 ms | 3/6 | 0/6 |
| 25 ms | 5/6 | 0/6 |
| 100 ms | 6/6 | 0/6 |
| 400 ms | 4/6 | 0/6 |

**That asymmetry is the issue's `seed`/`src` alternation.** `seed/` is a direct child of the watched root, so parcel does see its create event; `src/feature` is a grandchild created inside the blind window. It is also why the ~120 re-touches #2064 added can't help — the nudged file sits in a subtree with no watch on it, so each nudge re-provokes an event nobody is listening for.

What decides it is *when* the watcher installs, not how loaded the box is: `localStatus` is an always-on query that subscribes the moment the terminal's repoRoot is known, so the race is "does the watcher attach before or after the scenario's `mkdir`". An isolated run reliably attaches after — which is exactly why a single-scenario repro passes in ~6s and only the loaded full suite reproduces. It also explains both observed flavors: a watcher that attaches before `mkdir` leaves git status frozen at the *empty-repo* snapshot (`seed` never marked), one that goes blind later freezes it after the first edit (`src` never marked), and `fs.listAll` is healthy throughout because it subscribes later and takes a fresh snapshot of its own.

**Fix:** rebuild the subscription when a create event names a directory that already has entries in it, so parcel's recursive walk re-establishes coverage. "Empty right now" is a sound all-clear — parcel delivers the event only *after* adding the watch — so a plain `mkdir foo` stays off the rebuild path entirely, and a burst coalesces into one rebuild. Same sequence with the rebuild: **0/8 blind, against 7/8 without.**

## 2. Overlapping calls for one key corrupt parcel's bookkeeping

Found while chasing the above, and load-bearing for its fix. Parcel keys **both** its process-global `Watcher` registry (`Watcher::getShared`) and its backend subscription set (`Backend::watch` / `Backend::unwatch`) on `(dir, ignorePaths, ignoreGlobs)`, and does each call's backend half on a libuv threadpool thread. Two overlapping calls for the same key complete in arbitrary order, and one order loses: the new `subscribe` finds the retiring watcher still registered (equal by key) and installs **no OS watches**, then the retiring `unsubscribe` tears the existing ones down.

`installSharedWorkingTreeWatcher` issued exactly that pair — a non-awaited `subscription.unsubscribe()` on last-unsubscribe, then `parcelSubscribe` on the same `repoRoot` with the same ignore set for the next subscriber.

| teardown | idle | libuv pool saturated |
| --- | ---: | ---: |
| un-awaited, then immediate rebuild | 18/25 dead | 10/10 dead |
| awaited first | 0/25 | 0/10 |

**Fix:** one serialization chain per `repoRoot` covering every parcel call. A chain rather than a "wait for pending teardowns" barrier, deliberately — an install can be cancelled while its own `subscribe` is still in flight, and a barrier wouldn't yet know about the unsubscribe that install is about to need. Routing every call through one queue makes an overlap **unexpressible** rather than merely unlikely, and it is what lets §1's rebuild tear down and re-attach safely.

## Evidence

`working-tree-watcher.churn.test.ts` pins both invariants at kolu's own seam — after a rebuild, and after a subtree appears faster than parcel can follow, a write **still** reaches the listener. Each test forces the ordering that hides on an idle host (the churn case loads the threadpool; the blind-subtree case creates the nested tree in one burst).

- **Before:** `watches a subtree that was created before parcel could watch its parent` fails `expected 0 to be greater than 0`; the churn case reports all 4 rebuild cycles dead.
- **After:** both green. Full `kolu-git` unit suite: 108 passed.

The hazards are also written into the `parcel` skill's failure-mode list with these measurements, so the next person wiring a watcher meets them before they meet the bug.

## Scope note

This is a product defect, not a test defect. Any user whose repo grows a directory tree faster than parcel can follow it — a `git checkout` that lands a new package, a scaffold command, an `mkdir -p` — got a Code tab silently frozen on stale git status for that subtree, with no error and no recovery short of reopening the repo. The e2e scenario was only the thing patient enough to notice.

Fixes #2065.


srid:
## CI e2e metrics — `8df7316` (linux `#1`, darwin `#3`)

| `x86_64-linux` | localhost | 6 | 16 | 8 | 494.799 s |
| `aarch64-darwin` | zest | 4 | 10 | 6 | 575.526 s |

Both pipelines settled green. `ci::e2e@x86_64-linux` — **507 scenarios, 507 passed, 0 retries**, with zero occurrences of `was not marked as containing a change` anywhere in the log. `ci::e2e@aarch64-darwin` — **506 scenarios, 506 passed**.

For contrast, the same linux lane on the previous SHA of this branch — carrying only the call-serialization half of the fix — reproduced #2065 exactly: `507 scenarios (1 failed, 506 passed)`, both of the scenario's attempts exhausted, alternately failing to mark `seed` and `src`. That red is what pointed at the second and primary cause fixed here.

PR #2090, opened 2026-07-31
Flatten nested splits on the canvas

### LINKED ISSUE #2059: Canvas does not render a terminal split whose parent is itself a split child (nested parentId)

srid:
An agent or `padi-tui create --parent` can hang a terminal under another split — and that nested terminal used to be **fully alive** (roster, agent state, MCP) while **never painting** on the canvas. The tile only ever showed one hop of children.

This PR fixes the renderer, not the data. The canvas groups every descendant of a tile under its **root ancestor** and paints them as the existing flat tab strip — a grandchild looks like any other split. The Dock keeps the true tree. `parentId` is preserved byte-for-byte, so when a real deeply-nested terminal model lands, existing workspaces come back nested with no migration.

### Approach

| Surface | Relation used |
| --- | --- |
| Canvas tabs, subCount, ports/attach panes, Cmd-cycle | Flat descendants under the root tile (`getSplitPaneIds`) |
| Dock tree | True one-hop (`getSubTerminalIds`) |

Closing a middle split **re-homes** its children under the root tile rather than killing them. A cycle or dangling `parentId` paints as a top-level tile so nothing can vanish. padi keeps only two guards — no self-parent, no parent cycle — and does **not** refuse nested depth or rewrite saved workspaces (supersedes the closed approach in #2060).

A bonus falls out by construction: a sleeping intermediate can no longer swallow its descendants, because they hang off the live root tile's panel.

### Design philosophy

- **Fail-fast** — nonsense edges (self-parent, cycle) throw; no silent rewrites or depth knobs.
- **Electricity** — pure leaf grouping; no new transport/persistence boundary.
- **Source of truth** — one pure walk (`terminalTree.ts`); canvas vs Dock pick the accessor, not a parallel model.

Closes #2059.

### Try it locally

```sh
nix run github:juspay/kolu/fix/2059-flatten-splits
```

_Generated by [`/be`](https://github.com/srid/agency) on Grok (model `grok-4.5`)._


srid:

**Status:** applied

| Lens | Findings |
| --- | --- |
| lowy | 2 major + 1 minor (all fix) |
| hickey | 2 major + 1 minor (all fix) — same three issues |

### Applied

- **≡ F1 — containing-tile fold still one-hop** — `activeTileOf` now walks placement to the root via `containingTileOf` / `rootAncestorOf`, so nested focus (R ← M ← G) sets `activeId` to R, not M.
- **≡ F2 — tile-owner joins** — Code open, deep-link anchor, session-restore intent, and port-serving candidates resolve the **root tile**, not the true one-hop parent.
- **≡ F3 — adopt root walk** — `useAdoptNewSplit` calls `containingTileOf` instead of an inline loop.

### Observations

- Dual `getSubTerminalIds` / `getSplitPaneIds` is deliberate (Dock structure vs canvas flat strip), not fragmentation.
- `terminalTree` is a leaf, not electricity.

_Generated by the lowy ∥ hickey lens review (`--no-comment`; posted by `/be-review` after push)._

srid:
## Codex ⇄ Grok debate

✅ Author fixed all findings through round 3 (batch rehome under departing root). Awaiting final peer approve if still open; fixes are on the branch.

|---|---|---|
| 1 | b45aec436 | fix(debate): nested eviction snapshot, hydration tabs, tile selection |
| 2 | deec0f07c | fix(debate): promote when rehome target is also departing |

- **F1** (open) — The production path now calls terminals.rootAncestor(), but this test's useTerminalStore mock does not provide it
- **F2** (open) — The list-driven natural-exit path runs after the removed middle terminal is absent, but it asks the live rootAncestor/getSplitPaneIds index to reconstruct the old tree
- **F3** (open) — Hydration still groups and validates active tabs by true one-hop parent
- **F4** (open) — terminalIds intentionally paints cycle members and dangling-parent orphans as top-level, but both setActiveSilently and focusMainTerminal still classify any record with parentId as a split and throw
- **F5** (open) — The foreign-host port lookup hand-rolls another containing-tile walk instead of using terminalTree's declared source of truth, and its missing-parent behavior is wrong: for orphan O -> gone it returns `gone`, so armOf cannot name the result and the port link disappears even though O is painted top-level elsewhere.
- **F6** (open) — descendantsByRoot is documented as one O(T) walk and padi now permits unbounded depth, but cycle detection calls path.includes(cur) on every hop

srid:

| Pass | Issues | Details |
| --- | --- | --- |
| Rules | 0 | Fail-fast parent guards; no fallback rewrites of saved data; dual Dock/canvas projections intentional |
| Fact-check | 0 | Root walk + rehome from pre-removal graph traced; cycle/orphan top-level aligned with terminalIds |
| Elegance | – | Skipped (`--no-elegance`; simplify track had no further surface after debate) |
| Simplify | – | No additional simplify commit — dual accessors and RemovalGraph are load-bearing |

**Nothing further to action** after the lens + Codex debate fix commits.

_All clear after police review of post-debate tree._

srid:
## CI e2e timing — `48ef304#1`

| `x86_64-linux` | `kolu-ci-3` | 8 | 32 | 8 | 2m53s (173064 ms) |
| `aarch64-darwin` | `ci@petit` | 4 | 10 | 6 | 4m11s (250867 ms) |

Full two-platform run settled green in 6m07s; all 48 required contexts pass.

srid:
## Evidence

Nested splits (depth ≥ 2) now paint on the canvas as flat tabs of the root tile. `parentId` is preserved; the Dock still shows the true tree.

**Unit suite (client + padi)** — root-ancestor flatten, Dock one-hop untouched, cycle/orphan top-level, middle rehome, highest-surviving-ancestor rehome, containing-tile fold, acyclic parent guards.

**e2e** — scenario *A split of a split is visible on the canvas* in `features/sub-terminal.feature` (creates a nested parent via lifecycle RPC; asserts 2 tabs + sub-count 2). Green in full `ci::e2e` on both platforms at `06c8798`.

**Bonus** — descendants of a middle node hang off the root panel, so a sleeping intermediate can no longer swallow them (by construction; unit covers the grouping).

PR #2092, opened 2026-07-31
Load a gitignored directory's contents when it is expanded

srid:
## The bug

In the Code tab's browse tree with "show gitignored files" on, `blog/out` renders as a folder with a chevron, expands — and shows nothing. The terminal right below it proves the directory isn't empty: `ls blog/out/` lists `000.html  001.html  002.html  hl.js  index.html  style.css`.

Nothing was lost in transit. `fs.listIgnored` runs `git ls-files --others --ignored --exclude-standard --directory`, and `--directory` **collapses a wholly-ignored directory to a single trailing-slash entry** — that's what keeps `node_modules/` at one row instead of thousands. Pierre builds its hierarchy from that one flat snapshot and has no lazy-children path, so the row gets a working chevron over children that were never sent, and no amount of clicking will produce them.

## The fix

Keep the collapse — it's load-bearing — and read the level on demand.

**`fs.listDirectory`** (padi surface 4.6, purely additive) reads ONE level of a directory, with subdirectories carrying git's trailing slash so they arrive collapsed and expandable in their own turn. One level is what bounds the cost: expanding `node_modules/` becomes a single readdir handing back package names, where the recursive listing `git ls-files` would have to produce is ~100k paths for that same click. No gitignore filtering is applied and none is missing — git collapses a directory only when *everything* beneath it is ignored, so inside a collapsed row a plain readdir is the truthful listing. `browse.test.ts` checks that against git itself rather than trusting the argument.

**`<FileTree>`** gains `lazyDirectories` + `onExpandLazyDirectory` + `lazyEpoch`. Pierre exposes no expansion callback and maps expand/collapse to no mutation event (they're `FileTreeStoreIgnoredSemanticEvent`s, which `toTreesMutationEvent` drops), so the transition is read off its store subscription — probing only the host's declared keys, never the whole visible projection, since that listener also fires for selection and focus.

One thing that surfaced while building it: loading a level *replaces* the collapsed key with its children, and Pierre's remove destroys the directory node carrying the expansion — so the folder snapped shut at the exact moment its contents arrived, reading as if the click had done nothing. Open lazy directories are now re-expanded alongside the path delta.

**`mergeBrowseInventory`** grows rule 4 — a loaded directory yields to its children, which join the dimmed overlay, and children are dropped unless their directory survived rules 1–2 (so a `.gitignore` edit can't leave a ghost subtree beside the tracked files that now own the prefix).

## What the review gauntlet changed

The first draft worked but was wrong in ways worth naming. Four reviewers ran over it — an architecture-first-principles pass, a lowy ∥ hickey lens review, a Codex debate (3 rounds to consensus), and a simplify + code-police pass. Between them:

- **A symlinked directory came back as a *file* row.** `readdir({withFileTypes:true})` has `lstat` semantics, and under pnpm most of `node_modules` is a symlinked package directory — 20 of 30 entries in `packages/client/node_modules` here. So this PR's own documented promise, "walk down into `node_modules/solid-js/` a click at a time", was false in the repo it ships in, and clicking one gave an EISDIR toast. Now the link's *target* decides the row.
- **An empty ignored directory lost its row entirely.** `[]` is not `undefined`, so an empty read took the "loaded" branch and the collapsed key yielded to nothing — Pierre had no child prefix left to infer the folder from, and the row the user just clicked disappeared. Worse than the original bug: there was no folder left to click again. Reachable with no race, since git lists a permanently-empty ignored directory in the overlay directly.
- **A stale read could paint one repo's build output into another.** `loadedChildren` is keyed by a repo-*relative* path, and those collide by construction (`out/`, `dist/`). The slot clear couldn't reach a request already in flight. Reads are now superseded by an `AbortController` and scoped to the slot that issued them.
- **A failed load retried forever.** On rejection the wrapper forgot its bookkeeping key but left the row expanded, so any unrelated store tick read "expanded, not recorded" as a fresh expansion and re-fired the failing load. The row now collapses on rejection.
- **A gone file stopped being recognised as gone.** Narrowing `isFileGoneError` to trust a native errno (it previously matched `ENOENT` anywhere in a message, so an `EACCES` on `/repo/ENOENT-artifacts/` answered true) broke the binary-preview path, which had been relying on that string surviving onto the wire. `filePreviewTag` now returns the structural `FILE_GONE` tag like its siblings, so `unwrapGit` maps it to `NOT_FOUND` before any wrapper can obscure the errno.
- Plus: expanding a gitignored folder no longer un-dims it; a search keystroke no longer erases expansion intent; the rule-4 walk is a bounded recursion rather than a spread that could `RangeError` on a huge level; and the now-dead `fileGoneAsNotFound` wrapper is gone.

## Staleness

Reopening a folder refetches it, and that's the only honest refresh available: the working-tree watcher's ignore set is *built from* `listIgnored`, so an ignored path emits no pulse by construction. Watching them would mean watching exactly the churn (`node_modules`, build output) the ignore set exists to keep out. So an already-open folder can go stale until it's collapsed and reopened.

## Tests

Written first, and confirmed red for the right reasons before any implementation — including every fix above, each verified red against the pre-fix code:

- `FileTree.lazyDir.test.tsx` drives the **real** library rather than a mock, because the subject is precisely Pierre's own behaviour on a childless directory row — that it gets a chevron at all, and that expanding it ticks the store. Assertions are on the callback and on painted DOM rows, since "model right, screen empty" is the original defect.
- `browseInventory.test.ts` — the four merge rules as one table test.
- `browse.test.ts` / `errors.test.ts` — the one-level listing, the traversal guard, symlink resolution, and the gone-path classification.
- `terminalWorkspace/endpoint.test.ts` — the assembled `kolu-git → unwrapGit → ORPCError` chain, proving all three reads that share the delete-while-viewing race surface as `NOT_FOUND`.


srid:

Two structural lenses read this diff independently against `fb21099` — **lowy**
(volatility-based decomposition: does each boundary encapsulate one reason to
change?) and **hickey** (structural simplicity: is anything braided that could be
apart?). 21 findings between them. **17 applied**, 3 rejected with reasons, 1
recorded as a named follow-up.

Both lenses landed on the same major defect from opposite directions — a
symlinked directory renders as a clickable file — which is worth stating plainly:
under pnpm, `packages/client/node_modules` in this very worktree holds **20
symlinks out of 30 entries**, so most of what this feature exists to show was
painting as a file leaf that answers EISDIR when clicked.

**Status:** `just check` green · unit tests green (`kolu-client` 1231,
`@kolu/solid-pierre` 44, `kolu-git` 116, `@kolu/padi` 456) · `just fmt` run.
Commits are on the branch, not pushed or merged by the review.

| lens | raised | applied | rejected | follow-up |
| --- | ---: | ---: | ---: | ---: |
| lowy | 8 | 4 | 3 | 0 |
| hickey | 13 | 11 | 0 | 1 |
| **shared** (`≡` both lenses) | 1 | 1 | 0 | 0 |

### Findings

| # | Title | Location | How it settled | SHA |
| hickey #1 ≡ lowy #7 | A symlinked directory is emitted as a file leaf | `git/src/browse.ts` | applied — `stat` the symlink entries only; broken link stays a leaf | `3816047` |
| lowy #5 | A `readdir` ENOENT is filed as `GIT_FAILED` | `git/src/errors.ts`, `browse.ts`, `servePadi.ts` | applied — new `FILE_GONE` tag, mapped structurally to `NOT_FOUND` | `3816047` |
| hickey #12 | The folder-key format is minted by hand | `git/src/browse.ts` | observation — cross-reference added, no code move | `3816047` |
| hickey #2 | Rule 2 is not applied to loaded children | `browseInventory.ts` | applied — `seen` guard on every entry reaching `overlay` | `eae6eb7` |
| hickey #6 | The rule-4 walk: four accumulators and a spread that can throw | `browseInventory.ts` | applied — one recursive `emit`; kills a double-list and a `RangeError` | `eae6eb7` |
| hickey #5 | Expanding a gitignored folder un-dims it | `browseInventory.ts` → `CodeTab.tsx` | applied (minimal) — `ignored` now means "the rows to dim" | `eae6eb7` |
| hickey #10 | `BrowseInventory` has a second, hand-written builder | `CodeTab.tsx` | applied — `diffInventory` is the one constructor | `eae6eb7` |
| hickey #3 | `openLazyDirs` carries three facts; two break | `FileTree.tsx` | applied, both legs | `60d22fd` |
| lowy #2 | `openLazyDirs` has no reset channel for a slot change | `FileTree.tsx` + `CodeTab.tsx` | applied — `lazyEpoch` prop, fed `slotKey()` | `60d22fd` · `74221a1` |
| lowy #6 | A rejected load leaves the wrapper believing it succeeded | `FileTree.tsx` + `CodeTab.tsx` | applied — the callback may return a promise | `60d22fd` · `74221a1` |
| hickey #7 | The store subscription's callback is unguarded | `FileTree.tsx` | applied — wrapped in `safeApply` | `60d22fd` |
| hickey #8 | `expandPaths`' JSDoc states a now-false invariant | `FileTree.tsx` | applied — amended in place | `60d22fd` |
| hickey #9 | "Which directories should be open" is assembled twice | `FileTree.tsx` | applied — `desiredExpandedPaths`, used at both sites | `60d22fd` |
| hickey #4 | Two hand-rolled staleness axes; `.catch` checks one | `CodeTab.tsx` | applied — one `AbortController` per directory | `74221a1` |
| lowy #1 | Restructure to an `onLoadChildren` owning the lifecycle | `FileTree.tsx` / `CodeTab.tsx` | **rejected with evidence** | — |
| lowy #3 | Move `loadLazyDirectory` into `hostCodeTab` | `CodeTab.tsx` | **rejected with reason** | — |
| lowy #4 | Split rule 4 into `expandLoadedDirectories` | `browseInventory.ts` | **superseded by hickey #6** | — |
| hickey #11 | The dim sheet's selector count is no longer bounded | `CodeTab.tsx` | **follow-up** (perf refinement) | — |
| lowy #8 | `browse.ts` sits under a package named for one strategy | `git/src/browse.ts` | observation (lens disposition: drop) | — |
| hickey #13 | `listDirectory` does no git work | `git/src/browse.ts` | observation (lens disposition: drop) | — |

### The four commits

- **`3816047` · `fix(lens): follow a symlinked directory, and give a missing path its own tag`**
  `readdir({ withFileTypes: true })` has `lstat` semantics, so a symlink reports
  `isDirectory() === false`. Now only the symlink entries are `stat`ed — the
  common case still costs one `readdir` — and a broken link stays the leaf it is.
  Alongside it, a `readdir` ENOENT stops being filed as `GIT_FAILED` (a tag
  meaning "a git subprocess failed", for a call that spawns none): a `FILE_GONE`
  member carries it, `unwrapGit` maps it to a typed `NOT_FOUND`, and the
  `servePadi` wrapper whose contract rested on an errno string surviving two
  re-wraps is deleted.
- **`eae6eb7` · `fix(lens): the browse inventory merge`**
  Rule 2's membership test now covers both sources feeding `overlay`; the rule-4
  queue walk becomes one recursive `emit` (fixing a directory listed twice and a
  `queue.unshift(...children)` that is a genuine `RangeError` on the six-figure
  level this feature targets); `ignored` regains the loaded directory's own key,
  so a folder no longer **un-dims the moment you open it** with its children
  dimmed below; and `diffInventory` becomes the type's one no-overlay constructor.
- **`60d22fd` · `fix(lens): lazy-directory expansion bookkeeping`**
  `openLazyDirs` was three facts in one Set. A row hidden by the search
  projection no longer erases the user's expansion intent (a filter keystroke
  used to collapse a hand-opened folder for the rest of the mount); a key the
  host stops calling lazy is pruned, so an eye-toggle round trip refetches
  instead of showing a stale level; a `lazyEpoch` clears the record on a repo /
  host switch; a rejected load un-records the expansion so a retry costs one
  re-expand; the store subscription runs under `safeApply`; and "which
  directories should be open" is spelled once.
- **`74221a1` · `fix(lens): supersede an in-flight directory read by aborting it`**
  One `AbortController` per directory replaces the hand-rolled `loadGeneration` +
  captured-slot pair. Two conditions that two callbacks each had to repeat — and
  had already drifted, since `.catch` checked only the slot and could toast
  `Failed to list out/` over a folder showing correct contents — collapse into one
  fact neither callback can disagree about. The superseded `readdir` now stops
  server-side instead of racing.

### Rejected, with the reasons

- **lowy #1 — restructure to an `onLoadChildren` owning the whole load
  lifecycle.** Settled by evidence rather than argument: row-dimming is built by
  the **host**, from `treeInventory().ignored` into a `data-item-path`
  stylesheet. Move the children cache into the wrapper and the host no longer
  knows the loaded paths, so it cannot dim them — the split does not disappear,
  it reverses direction and needs a new reverse channel. The concrete costs lowy
  attributed to the split are fixed directly instead (lowy #2, lowy #6,
  hickey #3).
- **lowy #3 — move `loadLazyDirectory` into `hostCodeTab`.** `hostCodeTab` exists
  to **retain** per-host reads across a switch; this cache exists to be
  **discarded** on one — repo-relative keys collide across repos, and nothing
  watches ignored paths, so a retained cache would be arbitrarily stale. Putting
  a deliberately-non-retained value inside the retention owner would misrepresent
  it. The explicit invalidation lowy asks for is present, as the `slotKey`
  abort + clear.
- **lowy #4 — split rule 4 into a separate `expandLoadedDirectories`.**
  Superseded by hickey #6, which reaches the same readability goal in place, with
  far less churn, and fixes two real bugs on the way; splitting would also churn a
  19-case table test.

### Recorded as a follow-up

- **hickey #11 — bound the dim sheet's selector count with prefix selectors.**
  `rowPathsCss` emits one `[data-item-path=…]` selector per ignored entry, so a
  loaded level adds one per child (expanding `node_modules/.pnpm/` here adds
  **656**), rebuilt and string-compared on every inventory tick. The fix is a
  `rowSubtreesCss(dirs, decl)` in `@kolu/solid-pierre` emitting one
  `[data-item-path^=…]` per loaded directory — exactly as true, since the
  collapse's own definition says the subtree is wholly ignored. hickey #5's
  minimal correctness fix landed instead; the performance refinement is
  deliberately not being done late in this PR.

### Recorded as observations

lowy #8 and hickey #13 are the same fact from two lenses: `browse.ts` now holds
three exports that touch no git, so its **name** has drifted from its content —
but the placement is right, because `listDirectory`'s correctness argument ("git
collapses a directory only when everything beneath it is ignored") is a statement
about `git ls-files` that must stay next to `listIgnored`. Both lenses reached
`drop`. Recorded so a package split is not re-proposed later as if it were new.

srid:
## Codex ⇄ Claude debate

✅ Consensus in 3 round(s) — base `fb21099b`

|---|---|---|
| 1 | 8a8e04653 | fix(debate): surface a stat failure instead of calling it a file |
| 2 | b6ec2116b | fix(debate): tag a gone preview file structurally, not by message |

- **F1** — The catch on each symlink stat swallows every failure and reports the entry as an ordinary file
- **F2** — isFileGoneError scans the entire error message for the token ENOENT/no such file even when the native error has a different code
- **F3** — Promise.all creates and queues one stat promise for every symlink in the directory at once
- **F4** — The version test was only half-updated: its name still says 4.5, and the compatibility assertion/comment still pins 4.4 versus 4.5/session.restore rather than the new 4.5 versus 4.6/listDirectory edge
- **F5** — The documented GitError member list omits the FILE_GONE variant added in this change, so the package's error-handling reference is already stale.
- **F6** — The code-authoritative narrowing regresses delete-while-viewing for binary previews

srid:

Pass 1 (rule checklist) + Pass 2 (fact-check), scoped to `fb21099b...HEAD`. Pass 3 (elegance) skipped — `/simplify` had already run over this same tree.

**4 real violations found, all fixed.** Every finding was verified against the code rather than taken on faith.

| Commit | Rule | Finding |
|---|---|---|
| 8ef71217a | `prefer-focused-library`, `no-untyped-escape-hatches` | `listDirectory`'s hand-rolled worker pool (shared counter + `as`-cast index reads) replaced with `p-limit` |
| 8ef71217a | `errors-must-log-at-error` | `listDirectory`'s catch logged the *expected* `FILE_GONE` race at `error` unconditionally; now `debug`, matching sibling `filePreviewTag` |
| 4f1a5c67b | correctness (Pass 2) | **A failed lazy-dir load retried forever.** On rejection the wrapper forgot its bookkeeping key but never collapsed the row, so any unrelated store tick read "expanded, not recorded" as a *fresh* expansion and re-fired the failing load. Fixed by collapsing the row on rejection |
| 668b5ce2e | `no-dead-code` | Five stale comments across 5 files still referenced `fileGoneAsNotFound`, which this PR deletes |

The retry bug (4f1a5c67b) is the substantive one — it was introduced by an earlier gauntlet fix in this same PR, and its regression test was verified red against the pre-fix code before the fix was restored (`no-vacuous-assertion`).

Adding `p-limit` moved the pnpm lockfile, so the recorded `fetchPnpmDeps` hash in `nix/workspace.nix` was refreshed with a real `nix build --rebuild` — and re-verified after the master merge, since that merge touched the lockfile again.

Verified green: `just check`, `just fmt`, and the four touched packages' unit suites.

srid:
## CI e2e metrics — `92182b7#1`

|---|---|---|---|---|---|
| `x86_64-linux` | kolu-ci-3 | 8 | 32 | 8 | 181052 ms |
| `aarch64-darwin` | ci@petit | 4 | 10 | 6 | 244791 ms |

Two-platform run, settled green in 387855 ms with no reruns. Durations are the whole `ci::e2e` recipe wall from the durable ledger (`.ci/92182b7/runs/1.json`), not Cucumber's internal timer; the worker counts are the last `e2e: workers=` line each lane recorded.

srid:
## Evidence

**What the clip shows:** in **Code → All files** with the **eye** toggle flipped on, a wholly-gitignored `build/` arrives as one collapsed row — clicking it now paints the level underneath (`bundle.js` plus `nested/`, itself collapsed), and clicking `nested/` reads its own level (`chunk.js`). One cheap directory read per click; before this PR that chevron opened onto nothing.



End state — two levels walked, `.gitignore` only names `build/`:



▶ HD (full 1280×720 frame, terminal + panel): https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=code-tab-expand-ignored-dir.mp4

<details>
<summary>How it was captured</summary>

Recorded off-machine on an ephemeral `pu` box from a clean clone of this branch at `92182b7b2`, through the project's own Cucumber + Playwright harness (`KOLU_EVIDENCE=1 just test-quick`, 1280×720, motion on) — no hand-rolled driver.

The fixture the scenario builds:

```
build/bundle.js
build/nested/chunk.js
.gitignore        # "build/"
README.md
```

`git ls-files --others --ignored --exclude-standard --directory` collapses that whole subtree to the single row `build/`, which is exactly the collapse worth keeping (`node_modules/` stays one row, not forty thousand).

**Scenario — `1 scenario (1 passed) · 21 steps (21 passed)`:**

```gherkin
Scenario: Expanding a gitignored directory loads its contents
  ...repo fixture via the terminal...
  And I click the Code tab
  And I click the Code tab mode "browse"
  Then the file browser should show a file "README.md"
  And the Code tab should not show a directory node "build"
  When I click the Code tab show-gitignored toggle
  Then the file browser should show a directory "build"
  And the file browser should not show a file "build/bundle.js"
  When I click the directory "build" in the file browser
  Then the file browser should show a file "build/bundle.js"
  And the file browser should show a directory "build/nested"
  And the file browser should not show a file "build/nested/chunk.js"
  When I click the directory "build/nested" in the file browser
  Then the file browser should show a file "build/nested/chunk.js"
  And there should be no page errors
```

The two `should not show` assertions are the load-bearing half: they pin the *laziness* — a level only arrives once its own row is opened, so the fix didn't quietly become a recursive walk.

Every step is an existing one except `I click the Code tab show-gitignored toggle` (a click on the existing `code-tab-show-ignored-toggle` testid); the eye had no step in the library. The scenario was applied on the capture box only and is **not** part of this branch, so the green CI on `92182b7b2` still describes what will merge.

</details>

**Mobile / coarse-pointer:** considered, no second capture needed. The two axes are orthogonal — viewport size (`isMobile` / the drawer layout) vs. input modality (`isTouch()`, `(pointer: coarse)`) — and this change touches neither. It adds no new hit target and no new layout: the rows a lazy read paints are ordinary Pierre tree rows in the same mounted `<FileTree>` that already renders every other row, so they inherit `density="relaxed"` under `isTouch()` exactly like their siblings, and the row that receives the click (the collapsed ignored directory) already existed and already had its chevron. At phone width the Code tab is hosted by `RightPanelDrawer.tsx` with the same `<CodeTab>` component and the same tree, and the eye toggle is a pre-existing `ToolbarIconButton` this PR does not resize — so the expanded rows are reachable there for the same reason every other tree row is.

PR #2093, opened 2026-08-01
Build osfacts from its own repo

srid:
osfacts moved out to [juspay/osfacts](https://github.com/juspay/osfacts)
([#1](https://github.com/juspay/osfacts/pull/1)), so kolu stops carrying 9,187
lines of a tool it only consumes and pins it instead. OSF5's closing half.

**The closure does not move.** The pinned build and the in-tree build produce
the identical store path:

```
$ nix eval --raw .#osfacts.outPath      # before, from ./osfacts
/nix/store/50ldwxk543rhbzw947w4dg3i5nlifjq0-osfacts-0.1.0
$ nix eval --raw .#osfacts.outPath      # after, from the npins pin
/nix/store/50ldwxk543rhbzw947w4dg3i5nlifjq0-osfacts-0.1.0
```

Same binary, same `KOLU_OSFACTS_BIN`, same padi/kaval closures. npins holds the
revision; `nix/env.nix` and `default.nix` import the pin; `nix run .#osfacts`
still works from a kolu checkout, because that is how a kolu dev reaches the
sampler kolu bakes. Bump it with `npins update osfacts`.

## What did not leave

`osfacts/client-ts` stayed, and it is worth being precise about why:
`osfacts-client` is a **pnpm workspace member**, imported by name from
`packages/padi`, `packages/kaval`, `packages/server`,
`packages/surface-daemon-supervisor`, and the surface examples. A workspace
member has to be a path inside the repo, so it cannot come from a pin until it
is consumable another way — a published package, or a pnpm git-subdirectory
dependency. `facets.json` stayed with it, because the client's tests pin the TS
facet unions to that file.

Two files in two repos is a drift risk, so it is gated rather than trusted:

| gate | what it forbids |
| --- | --- |
| `osfactsFacetsInSync` (flake check, new) | kolu's `facets.json` differing from the pinned repo's copy |
| upstream `v2_contract.rs::facets_json_is_the_enum` | that file differing from the Rust `Facet` enum |

So the TS unions are still transitively pinned to the binary's vocabulary, and
a stale copy is a red rather than a client that silently cannot name a facet
the binary emits. `osfacts/README.md` is now a four-table signpost explaining
exactly this, and what would finish the move.

## CI

`ci::osfacts` narrows from "run the tool's whole hermetic suite" to "does the
**pinned** revision build on this platform" — the tool owns both of its test
lanes upstream now, on both platforms.

⚠️ **`ci::osfacts-live` is gone**, and its two contexts are still in master's
`required_status_checks`:

```
ci::osfacts-live@x86_64-linux
ci::osfacts-live@aarch64-darwin
```

Nothing will ever post them again, so **this PR cannot merge until they are
dropped** (`just ci::protect`, which needs admin). I have not touched branch
protection — that is your call, not mine. `ci::osfacts@…` keeps posting on both
platforms and can stay required.

## Also

The Atlas plan-of-record is updated: OSF5 is shipped, the "what remains"
section became "what shipped last", and its links point at the new repo. I left
the note **alive** rather than archiving it, even though every phase has now
shipped — archiving would delete the page that the upstream README's "plan of
record" link points at. Worth deciding separately.


srid:
**CI run `735d83f#1` — settled green on both platforms.** 48 nodes, 0 non-ok.

| platform | host | workers | cores | cap | `ci::e2e` duration |
| x86_64-linux | kolu-ci-4 | 8 | 32 | 8 | 2m56s |
| aarch64-darwin | ci@petit | 4 | 10 | 6 | 4m9s |

Merged with master's derived-staleKey work (#2095); the grafted `osfacts-client`
store path is hashed into `PADI_BUILD_ID`, verified by rolling the pin
(`f3245b0b…` → `08f1a31b…`).

PR #2095, opened 2026-08-01
Derive daemon staleKeys from the dependency graph instead of hand-kept mirrors

### LINKED ISSUE #2094: Every derivation depends on the whole workspace, so daemon identity is a second hash maintained by hand
Every package in this repo is built from one shared source tree, so every
derivation depends on code it never loads. A second, hand-maintained hash
(`mkDaemonIdentity`) exists to undo that over-reach for the daemons. This issue
argues the over-reach is the actual defect, and shows a case where the two
hashes now disagree in a way that can leave a stale daemon running.

## Measured facts

All measurements pin `commitHash` to a constant, so the
`<PREFIX>_COMMIT_HASH` bake (which moves on any tracked-file change in a dirty
tree) cannot confound the result.

```sh
ev() { nix eval --impure --raw --expr \
  "(import ./default.nix { pkgs = import ./nix/nixpkgs.nix {}; commitHash = \"FIXED\"; }).$1.drvPath"; }

getid() { nix derivation show "$1" | tr -d '\\' \
  | grep -oE 'PADI_BUILD_ID "[a-f0-9]{64}"' | head -1 | grep -oE '[a-f0-9]{64}'; }
```

**1. An unrelated package's source is an input to every derivation.** Editing
`packages/client/src/AboutDialog.tsx` — a browser dialog that no daemon loads:

| derivation | before | after |
| --- | --- | --- |
| `padi` | `k6mjvanhvk…` | `apygqx3iid…` |
| `kaval` | `cbadlxiviil…` | `lajykg6a…` |
| `default` (kolu) | `aysw1xwxkkj…` | `6kpigrqx…` |

**2. The baked daemon identity does NOT move for that same edit.**

```
PADI_BUILD_ID  before: f4ea6dd2ac9f…   after: f4ea6dd2ac9f…   (SAME)
```

So for one edit: three derivations change, and the identity that decides
"should the running daemon be replaced?" does not.

**3. Docs are not inputs.** Editing `docs/atlas/src/content/atlas/meta.mdx`
leaves `padi`'s drvPath byte-identical. (Without the pinned `commitHash` it
appears to change — that is the commit-hash bake, not a real input.)

## Where the over-reach comes from

`nix/workspace.nix` builds a single `src` from a fileset listing all 43
workspace packages, and `default.nix` hands that same `src` to every package.
So `padi`'s derivation depends on `packages/client`, `packages/solid-*`,
`website`-adjacent packages, and everything else — none of which padi loads.

## The compensation, and what it costs

`packages/surface-daemon/nix/daemon-identity.nix` (`mkDaemonIdentity`) computes
a *second* hash over a hand-picked `behavioralFileset` — "WHAT counts as this
daemon's behavior — ITS OWN decision". Its header records the incident that
motivated it (kaval on zest, 2026-07-03): kaval's key once hashed the shared
transport spine, a spine-only refactor flipped it, and acting on the resulting
"update available" nudge would have recycled kaval and killed live PTYs.

That reasoning is sound *given* a derivation hash that is too broad. But the
cost is a policy that now lives in two places and must be kept in agreement by
hand — `default.nix`'s `behavioralFileset` and the mirrored lists in
`packages/padi/src/daemonBoot/buildId.closure.test.ts` and
`packages/kaval/src/buildId.closure.test.ts`, which exist to assert the mirror
holds. Two hashes, two hand-maintained lists, one question.

## A live symptom (PR #2093)

osfacts was extracted to its own repo; kolu pins it with npins, and
`nix/workspace.nix` grafts the pinned `client-ts/` into the build tree as the
`osfacts-client` workspace member. Consequences:

- `osfacts-client` **is** an input to padi's derivation (it is in `src`), so a
  pin bump produces a genuinely different padi binary.
- `osfacts-client` is **not** in padi's `behavioralFileset` any more, so
  `PADI_BUILD_ID` does not move.

Net: after `npins update osfacts`, you get **a rebuilt padi carrying an
unchanged identity**. The supervisor compares identities, sees no difference,
adopts the running process, and it keeps executing the previous client. Nothing
errors.

Before the extraction the client was in padi's `behavioralFileset`
(`(padiPkgRoot ./osfacts/client-ts)`), so this specific hole is new. kaval was
never affected — its closure test lists `osfacts-client` as a stable leaf on
purpose.

## Open question the design does not answer

A nixpkgs bump changes padi's Node runtime. The derivation moves; the identity
does not. The `mkDaemonIdentity` rationale covers contract-compatible *spine*
changes, not runtime swaps, so it is unclear whether ignoring this is a
deliberate accepted gap or an oversight.

## Direction worth evaluating

If each derivation's `src` were narrowed to that package's own dependency
closure, the derivation hash would mean "this package's code changed" — which
is what the daemon identity is trying to say by hand. The second hash might
then be derivable from the first rather than maintained separately, deleting
both hand-kept mirror lists and the class of bug above.

Non-trivial: it changes how `pnpmDeps` and the shared workspace install work,
and the daemon identity would need to stay narrower than the derivation in at
least one respect (kaval deliberately excludes its own daemon executable from
padi's key, and excludes the spine from its own).

Filed for analysis, not with a fix attached. The measurements above should
reproduce on `feat/consume-upstream-osfacts` and on `master` (the client graft
detail is branch-only; facts 1–3 are not).

srid:
A daemon's `<PREFIX>_BUILD_ID` answered the question "would a restart load different code?" with a **hand-listed nix fileset, mirrored by a hand-listed AST test** — two lists, one question, kept equal by discipline alone. #2094 measured where that leads: the lists can drift, and the failure mode is the worst kind — a rebuilt daemon carrying an *unchanged* identity, which the supervisor then adopts and quietly keeps running old code.

**Now the hashed set is computed, not remembered.** `nix/workspace.nix` gains a name→dir `members` index (keys asserted against each `package.json` at eval) and a `depClosure` walk over `workspace:` `dependencies` edges — the same edges pnpm's isolated `node_modules` makes the only resolvable ones at runtime. `default.nix` derives each daemon's `behavioralFileset` from that closure minus a short `stableLeaves` policy list, each leaf carrying its rationale (the zest 2026-07-03 spine exclusion chief among them). A new dependency joins the staleKey automatically; the surviving failure direction is over-inclusion — an unnecessary drain or an early nudge — never a silent escape.

### What keeps the derivation honest

pnpm guarantees an import resolves only through a *declared* edge, but not through which section. A runtime module riding a **devDependency** link works in every dev install while being invisible to the closure nix hashes — exactly the #2094 hole shape. The rewritten guards (`packages/{padi,kaval}/src/buildId.closure.test.ts`, on a shared walker in `@kolu/daemon-test-gate/runtimeDepEdges`) enforce that one invariant: every runtime import reachable from a daemon's entry files is an honest `dependencies` edge (type-only imports are erased and exempt). No `HASHED_ROOTS`, no `EXCLUDED`, no `ALLOWED_EXTERNAL` — nothing left to keep in lockstep.

### Proven equivalences

| check | result |
| --- | --- |
| workspace `src` store path | byte-identical before/after |
| kaval's hashed file set | byte-identical (still exactly kaval + terminal-protocol) |
| padi's hashed file set | same 26 packages; only delta: `dial.ts`, `watch.ts`, kaval's 3 executable files now included |
| `pnpmDeps` FOD hash | unchanged |
| edit `client/AboutDialog.tsx` | padi + kaval ids **SAME** |
| edit `surface-daemon` spine | kaval id **SAME** (zest policy), padi id flips |
| edit `osfacts/client-ts` | padi id flips (the #2093 hole class, now structural) |

_The old file-level carve-outs (padi's client-side `dial.ts`/`watch.ts`, kaval's executables out of padi's key) are deliberately dropped: package granularity is what the guard can actually verify, and every carve-out disappears in the safe direction (a no-op auto-drain), not the dangerous one._

### On the issue's two open threads

- **Grafted members (#2093):** when `osfacts-client` becomes an npins store path, `memberIdentityFileset` fails eval with a message saying to hash the pin's store path into the identity — the silent drop that motivated #2094 is now unspellable.
- **Runtime/nixpkgs swaps:** answered in `mkDaemonIdentity`'s doorstep — deliberately outside the staleKey (source hashes must stay platform-independent for cross-host comparison); the bounded-staleness cost and the rejected `nodejs.version` alternative are documented there.

Narrowing every *derivation's* `src` (the issue's suggested direction) was evaluated and deliberately not taken: the daemons ship from one shared workspace build, so per-daemon builds would duplicate `node_modules` trees and node-gyp runs, grow the remote-provisioning closure, and add per-build FOD hashes — while buying nothing the supervisor acts on, since convergence keys on the identity, not the store path. With the identity derived from real dependency edges, the broad derivation is a dedup trade-off, no longer load-bearing for correctness. Analysis on the issue.

Closes #2094

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -149,173 +149,181 @@ let
   # and the dev shell cannot drift.
   osfactsBakeArg = ''--set KOLU_OSFACTS_BIN "${koluEnv.KOLU_OSFACTS_BIN}"'';
 
-  # The `.ts` filter for a hashed daemon fileset: real source only (drops `.test.ts`
-  # AND `.testlib.ts` shared test-only helpers). The id is a content hash of the
-  # fileset's store path, byte-identical across Darwin/Linux; the recipe + rationale
-  # live in `mkDaemonIdentity`.
-  isHashedSource =
-    f: f.hasExt "ts"
-      && !pkgs.lib.hasSuffix ".test.ts" f.name
-      && !pkgs.lib.hasSuffix ".testlib.ts" f.name;
-
-  # kaval's baked identity. Its behavioralFileset — kaval's OWN decision of what a restart
-  # would load that MATTERS to the currency nudge — is the two BEHAVIORAL roots: kaval
-  # itself and `@kolu/terminal-protocol` (the wire/behaviour it serves). The
-  # `@kolu/surface-daemon` spine also runs in the kaval *binary* but is DELIBERATELY EXCLUDED
-  # (#L3, see the inline note below) — it is a stable leaf for kaval's currency key, not a
-  # hashed root. The set hashed here is asserted to equal kaval's behavioral slice by
-  # `packages/kaval/src/buildId.closure.test.ts` — keep this fileset and that test in
-  # lockstep. `kavalBuildIdOverride` (TEST-ONLY) forces the id for the build-skew VM arms.
-  kavalIdentity = mkDaemonIdentity {
-    name = "kaval";
-    prefix = "KAVAL";
-    root = ./packages;
-    inherit commitHash;
-    override = kavalBuildIdOverride;
-    behavioralFileset = pkgs.lib.fileset.unions [
-      (pkgs.lib.fileset.fileFilter isHashedSource ./packages/kaval/src)
-      ./packages/kaval/package.json
-      # @kolu/terminal-protocol — the device-query forward/drop policy + suppression
-      # grammars the daemon serves; a protocol change must not escape the staleKey.
-      (pkgs.lib.fileset.fileFilter isHashedSource ./packages/terminal-protocol/src)
-      ./packages/terminal-protocol/package.json
-      # `@kolu/surface-daemon` (the transport SPINE) is DELIBERATELY NOT in kaval's
-      # currency slice (#L3). kaval's staleKey drives ONLY the human "update available"
-      # nudge, and acting on it RECYCLES kaval — killing live PTYs. The spine's behavioral
-      # surface to a consumer IS the wire contract (`PTY_HOST_CONTRACT_VERSION`, in
-      # kaval/src above): a contract-COMPATIBLE spine change is behaviorally interchangeable
-      # BY THE CONTRACT'S DEFINITION, so keying currency on the spine double-counts what the
-      # contract already covers and over-fires the nudge on every compatible spine refactor.
-      # This was paid for in production (zest, 2026-07-03): a spine-only change with no kaval
-      # behavior delta flipped kaval's staleKey and fired a spurious nudge. A spine change
-      # that DOES matter to the wire bumps the contract (hashed here) → recycle-on-skew
-      # converges it, a separate sanctioned signal. So kaval's currency = its BEHAVIORAL
-      # closure (kaval + terminal-protocol), spine excluded. (padi keeps the full closure —
-      # its staleness response is a cheap auto-drain, so over-firing is harmless.) The
-      # closure guard in packages/kaval/src/buildId.closure.test.ts pins this restated
-      # invariant — keep the two in lockstep.
-    ];
-  };
-
-  # padi's staleKey (W2.2) — the twin of kaval's, one layer up. padi is the
-  # per-host terminal-workspace daemon; `PADI_BUILD_ID` (baked below) hashes the
-  # source closure that runs IN padi's process, so it flips iff a restart would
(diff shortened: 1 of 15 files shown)


srid:
CI e2e evidence — run `817bf6c#1`

| x86_64-linux | kolu-ci-2 | 8 | 32 | 8 | 3m06s |
| aarch64-darwin | ci@petit | 4 | 10 | 6 | 4m17s |

PR #2099, opened 2026-08-01
Resolve new-terminal theme in padi, so every face obeys the policy

### LINKED ISSUE #2045: MCP-created terminals bypass the Inherit terminal theme setting

srid:
Terminals created through the MCP face (`lifecycle_create`) ignored the **New terminal theme: Inherit** setting and always arrived on the built-in default — because inherit/shuffle resolution was request-decoration inside the browser's create path, so any non-browser caller silently skipped it. **Theme resolution now lives in padi's `lifecycle.create`**, where it applies to every face — UI, MCP, TUI — and every host. Closes #2045.

### How the policy flows

Each layer resolves the part it owns, and only **resolved facts** cross a wire:

```
browser                 kolu-server                    padi (per host)
───────                 ───────────                    ───────────────
viewerMode ────────────▶ preferences ⊕ viewerMode      newTerminalPolicy cell
(raw OS media query,     └─▶ resolveNewTerminalPolicy   (memory-only; re-pushed on
 re-published on             {inherit} | {shuffle,mode}  every bind/reconnect)
 focus + reconnect)          └─── push on bind ─────────▶  └─▶ lifecycle.create:
                                  and on change                inherit → own active tile's theme
                                                               shuffle → pickTheme vs own peers
```

- `"auto"` **never crosses a wire** — the viewer-scoped leg (`colorScheme: "system"` + the OS media query) is resolved before pushing, so the wire policy is a closed union: `{kind:"inherit"} | {kind:"shuffle", mode:"random"|"dark"|"light"|"colourful"}`. The UI's 10 preference combinations collapse to 5 wire values, and _a shuffle mode under inherit is unrepresentable_.
- The policy type is **declared in `@kolu/padi`** and re-exported through `kolu-common/surface` (the `ClientErrorPolicy` precedent) — the package seal forbids padi importing kolu-common, and raw preferences still never enter padi (the `stateStore.ts` ruling holds; only a resolved fact crosses).
- `viewerMode` is an **observation, not a preference**: the raw `prefersDark` answer, published from an app-lifetime root on connect, reconnect, and tab focus (so with two browsers open, the one you're in front of is the last writer). kolu-server persists the last-known value (state schema 1.36.0) so agent-created terminals resolve `system` correctly with no browser open.
- The padi cell is **memory-only** — kolu-server re-pushes on every bind/reconnect (a reconnect is a fresh padi), deduped per link epoch so an unchanged preferences write doesn't fan out over ssh.
- The MCP face stays sealed: the cell's `set`/`get` are **denied by omission** — agents' terminals obey your policy, but agents can't read or change it. `PADI_SURFACE_VERSION` bumps 4.6 → 4.7 (additive minor; the skew gate drains a straddling padi before the binder calls the new verb).

### Behavior notes

| Case | Before | Now |
|---|---|---|
| MCP / TUI create | always built-in default | obeys Inherit/Shuffle like the UI |
| Split (sub-terminal) create | default theme | no theme stamped — a split renders its parent tile's theme, and splits stay out of the shuffle peer set |
| Inherit with no open terminal | built-in default | same (documented now) |
| ⌘⇧J shuffle | local, viewer-side | unchanged — deliberately still local (a live viewer act, not create policy) |

> The e2e fixture hook now sequences preferences → padi reset → an explicit bounded wait until padi's policy cell reads `{kind:"inherit"}`, closing the race every hard-coded "Tomorrow Night" assertion would otherwise flake on.

_Generated by Claude Code (model `claude-fable-5`) via ultracode workflows (scout → 6-stage implement → adversarial review, 29 subagents)._

### DIFF
--- a/packages/client/src/settings/useColorScheme.ts
+++ b/packages/client/src/settings/useColorScheme.ts
@@ -12,12 +12,14 @@
  * it to resolve the new-terminal theme policy for every face (#2045).
  */
 
+import { makeEventListener } from "@solid-primitives/event-listener";
 import { usePrefersDark } from "@solid-primitives/media";
 import type { ColorScheme } from "kolu-common/surface";
 import { resolveIsDark } from "kolu-common/surface";
 import { createEffect, createMemo } from "solid-js";
 import { toast } from "solid-sonner";
 import { createSharedRoot } from "../createSharedRoot";
+import { wsStatus } from "../rpc/rpc";
 import { client, preferences, updatePreferences } from "../wire";
 
 export type { ColorScheme };
@@ -52,13 +54,30 @@ const sharedIsDark = createSharedRoot((): (() => boolean) => {
   // the reading, so a standing subscription on a cell nothing here reads would
   // be pure wire cost. (`app.rpc` is `unknown` — the dynamic combined link can't
   // be expanded per-key — and casting it is what `procedureCastGuard` forbids.)
-  createEffect(() => {
-    const mode = prefersDark() ? "dark" : "light";
+  const publishViewerMode = () => {
+    // Never write into a socket that is down. This is a raw `.set`, not a
+    // subscription, so nothing in `connectSurfaces` would resubscribe it — a set
+    // that rejected because the server was mid-restart would just be a toast, and
+    // the server would keep serving the stale reading to every face for the rest
+    // of the session.
+    if (wsStatus() !== "open") return;
     void client.surface.kolu.viewerMode
-      .set(mode)
+      .set(prefersDark() ? "dark" : "light")
       .catch((err: Error) =>
         toast.error(`Failed to report viewer mode: ${err.message}`),
       );
+  };
+  // Tracks BOTH the media query and the transport, so the reading is (re)published
+  // on the first connect, on every reconnect, and after a server restart — the
+  // server just rewrites the same value, and its scalar `equals` drops it.
+  createEffect(publishViewerMode);
+  // `viewerMode` is ONE server-wide fact, so with two viewers open (a dark laptop and
+  // a light phone) the last writer decides the policy for both. Republishing on focus
+  // makes that writer the tab the user is actually in front of — and since a terminal
+  // is created from a focused tab, the creating viewer and the daemon agree again.
+  makeEventListener(window, "focus", publishViewerMode);
+  makeEventListener(document, "visibilitychange", () => {
+    if (!document.hidden) publishViewerMode();
   });
   return memo;
 });
--- a/packages/client/src/settings/useColorScheme.ts
+++ b/packages/client/src/settings/useColorScheme.ts
@@ -6,31 +6,69 @@
  * shadow-DOM widgets (e.g. `@pierre/trees`, `@pierre/diffs`) that use
  * `light-dark()` or form-control theming follow the same resolved scheme.
  * Defaults to "dark" (the app's original palette).
+ *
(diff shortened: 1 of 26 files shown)


srid:
**CI e2e metrics** — run `892f979#1` (both platforms settled green)

|---|---|---|---|---|---|
| x86_64-linux | kolu-ci-4 | 8 | 32 | 8 | 3m 00s (180355 ms) |
| aarch64-darwin | ci@petit | 4 | 10 | 6 | 4m 06s (246175 ms) |

PR #2101, opened 2026-08-03
Effect 4: the whole stack — oRPC → effect/rpc, zod → Effect Schema, daemons on Layers

srid:
Kolu now runs end-to-end on **Effect 4.0 (beta.103)** — the wire, the schemas, the daemons, the HTTP app, the client's business logic, the reactive engine, and the CLIs. The oRPC wire is replaced by `effect/unstable/rpc` (flat, slash-tagged `RpcGroup`s over ndjson), every zod schema is Effect Schema, the surface framework's transports ride Effect sockets, the daemons boot through Layer graphs, and hono is gone — the entire HTTP app (static, PWA, artifact-sdk, health, preview, `serve()` itself, and the example servers) rides `effect/unstable/http`. Zero `zod`, `@orpc/*`, `partysocket`, `hono`, or `@preact/signals-core` dependencies remain. vazhi is the one deliberate non-Effect island (Ink rendering internals; nothing orchestral).

The work ran as two planned campaigns plus a review round, all on this branch. **The campaign notes — plan of record, recon dossiers, the 26 adversarial review findings that reshaped the plan, and every per-stage report — are attached as a gist: https://gist.github.com/srid/76c5cfa5e52a21fc77196aa6237e09f4** (the `PLAN D<n>` / `review #<n>` / wave-label citations in source comments resolve there; the gist's index maps each citation form to its file).

## The wire epoch, and the hands-off upgrade

The migration is a **wire epoch break**, handled as a declared flag day: `PTY_HOST_CONTRACT_VERSION` 6.0 → 7.0, `PADI_SURFACE_VERSION` 4.7 → 5.1. The daemon supervisor gained a third convergence observation — `unspeakable-protocol` — raised only for a **corroborated** peer (owned gate file, verified pid, re-attested immediately before any signal) that either sends undecodable frames or accepts and stays silent past 8s (the measured behavior of a previous-epoch daemon, pinned between Effect RPC's 5s ping and its ~10s connection kill).

[...]

PR #2103, opened 2026-08-04
feat: adopt the Effect-native osfacts client

srid:
The downstream half of [the campaign comment on #2101](https://github.com/juspay/kolu/pull/2101#issuecomment-5182758828). Upstream is done and green: [juspay/osfacts#5](https://github.com/juspay/osfacts/pull/5) turned the client's spawning verbs into `Effect.Effect<_, OsfactsClientError>` and split the old `kind` union into three `Schema.TaggedErrorClass`es. This adopts it.

**Base is `effect`, not `master`.** #2101 is the effect branch's open PR and this campaign explicitly must not reopen it. Merge order at the bottom.

## The pin, the lockfile, and the FOD

`npins/sources.json` → `86091cf`. The grafted package now declares `effect` at `4.0.0-beta.102` — the exact version this workspace already pins — so `pnpm-lock.yaml` gains **four lines** and nothing else: a `dependencies` section on the `osfacts-client:` importer. No new package resolves, because `effect@4.0.0-beta.102` was already in the lockfile via the catalog.

Which is why **the pnpm FOD hash did not move**. That is a re-derived fact, not an assumption: the hash was set to `lib.fakeHash`, `nix build .#pnpmDeps` re-executed the fetcher, and it handed back `sha256-trCV/uIW4ZqKN/iqTcw1Omw5MGA1YNZpfiNIFodrH/Q=` — the same value. The fetched package SET is what the FOD is over, and adding an importer edge to a package already in it changes nothing. The pin hash itself was taken with `nix-prefetch-url --unpack`, validated against the OLD revision first (it reproduced `sha256-7F5fA…` exactly).

The port also added `client-ts/pnpm-workspace.yaml` — pnpm 11's `allowBuilds: { msgpackr-extract: false }`, which is where osfacts' own workspace records that decision. kolu's parent workspace ignores it: `pnpm install` reports the same 64 projects and the same pre-existing `Ignored build scripts: msgpackr-extract@3.0.4` notice it reported before the bump. No root `onlyBuiltDependencies` decision is owed here, because nothing new resolved.

## The seam flip

**`EndpointSpec` has ZERO Promise-typed members now.** Verified by reading every member's type after the flip: `hostId`/`home`/`policy` are data; `probe`, `connect`, `driver.spawn`, `adoptHint.connect` are Effects; `onStatus`/`onSpawned`/`adoptHint.onAdopted` and `DaemonConnection`'s `dispose`/`onClose` are synchronous callbacks by contract; the four `*Ms` knobs are numbers. The last two holdouts were `readProcessIdentity` and `readSocketHolders`, and the reason written at each was the same sentence — the client that answers them declares no `effect` dependency, so the endpoint lifts the Promise at its single call site rather than letting it spread. That reason is gone, so both lifts are:

- `holdersOf`'s `Effect.tryPromise` — deleted; the three callers `yield* spec.readSocketHolders(path)` directly.
- the `Promise.resolve` normalisation around `readProcessIdentity` — deleted; the call site `yield*`s it.

Neither needed a `mapError`. The client's three error classes are still `Error`s, so they land on the callers' existing `Effect<_, Error>` channels unchanged.

The error channel on both seams is **`osfacts-client`'s own union**, not a wrapper of the spine's making. `socketHolder.ts` already re-exports the success half (`SocketOccupancy`, `SocketHolder`) verbatim; a local error type would be a second name for facts this package does not produce, and every consumer would unwrap it to get back the tag it needed. It also buys the thing the old `tryPromise`-not-`Effect.promise` comment argued for by hand: an identity read that fails stays a **typed failure** the endpoint has an answer for (report `dead`, then propagate — R3-5) rather than a defect that sails past `liveServingHolder`'s emit and strands the UI at `connecting`.

A new `@ts-expect-error` pair in `endpoint.public-api.test-d.ts` pins it: a Promise-returning inject for either seam no longer compiles, and each line fails-to-fail (i.e. breaks the build) the moment a seam widens back.

## Production consumers

`padiBinding.ts` and padi's `ptyHost/index.ts` needed **no edit** — they already passed `processIdentityAsync(bin)` and `osfactsSocketHolders(bin)`, and those are exactly the functions that changed shape.

The SYNC island stays sync, as decided: `kaval/daemonMain`, `padi/daemonBoot/daemonMain`, and `server/padi/supervisorClaim` keep `processIdentityFromEnv`. A synchronous gate claim must not reorder against the boot side effects it guards, and `execFileSync` cannot be interrupted, so an Effect there would advertise a capability the call does not have.

Two padi readers go Effect-native:

- **`samplePadiMemory`** is a VALUE now, not a function. That makes the capture-then-recheck real: kaval's connection identity is read *inside* the description, once per run, so a sample can never carry a pid resolved at construction. The three-way ok/absent/error fold is byte-identical; what changed is that `if (!(err instanceof OsfactsClientError)) throw err` becomes a single `Effect.catch` over a channel that IS the union. Anything the client did not declare stays a defect **by type** rather than by a hand-written re-throw. The bake resolution moved inside the description so an unbaked `KOLU_OSFACTS_BIN` still reads as the same per-tick `error` arm it always did (the old `try` block held that call for exactly that reason).
- **`scanSubtreePorts`** returns `Effect<Map<number, PortInfo[]>, PortScanError>`. The osfacts arm is a total `mapError` to `blind` — all three client failures mean this pass could not see — and the fold's own refusals moved onto the same declared channel, which is what the sampler's permanent-vs-transient branch reads. No `instanceof` left in the scan.

## Run-edge accounting

| | rows | files | sites |
| --- | --- | --- | --- |
| before (`effect` @ 41d5177) | 52 | 52 | 106 |
| after | 53 | 53 | 107 |

**Net +1 row, +1 site**, and the review's rule is that a net increase owes a written why. Here it is, and it is also on the row itself:

- **Added:** `packages/padi/src/ports/sampler.ts`, 1 site. The reactor's poll `read` dep is `() => Promise<T>` BY DESIGN — a poll source owns its own cadence and seed and is deliberately not Effect code (H1) — and `scanSubtreePorts` is Effect-native now, so the two meet in the ONE default `scan` seam rather than at each use inside the read. It is a **separate row** from `servePadi.ts`'s twin because this sampler is driven *outside* `derived.cell` (its samples re-enter each terminal's own producer through the sensor channel), so there is no padi surface cell to route it through.
- **`servePadi.ts` stays at 1.** `samplePadiMemory` joins the existing `pollRead` edge the host-inventory sampler already used; the row's own text already covers both cells.
- **Against it, and not counted here:** the two deleted `Effect.tryPromise` rinds in `endpoint.ts`. They were never run edges — they were *lifts* — but they are the Promise boundary this change actually removes. The one `tryPromise` left in that file is `serveStateOf`, which lifts `socketServeState` from `@kolu/surface-daemon` and has nothing to do with osfacts.

One correction to the census this was built from: the census said `createEndpoint.testlib.ts`'s `fromAsync` lifter should be deleted. It should not. `fromAsync` is a general Promise lifter for the `probe` / `connect` / `driver.spawn` fakes — three seams that were never osfacts and were already Effect-shaped — and it is used by `socketContractMismatch.test.ts`, `unspeakableProtocol.test.ts` and `recycleForeignGate.test.ts`. It stays. `testReadSocketHolders`, which IS the osfacts fake, needed no body change: it returns the real reader, which is now the real Effect.

## Governance

**The Effect pin gate now sees `osfacts-client`.** It was in `SKIPPED` because the manifest is grafted rather than committed — not ours to author. That was fine while the client was zero-dependency: nothing to disagree about. It declares `effect` now, and a graft on a *different* Effect than this workspace would put two copies in one process, which is the exact split this gate exists to refuse. Not ours to author is not the same as not ours to check.

It joins the LITERAL category, for the mirror image of the `@kolu/surface*` reason: those are vendored **out** to drishti and odu, where kolu's catalog does not exist; this one is grafted **in** from juspay/osfacts, whose workspace has no kolu catalog either. Two reasons deserve two names, so `literalReason(path)` returns the sentence the failure message uses rather than a boolean the message would restate. Only the vendored literals are *required to exist* — those files are committed, so their disappearance is a kolu edit; the graft is policed when present, because a bare checkout has not materialised it and osfacts is free to stop depending on Effect. Six new tests cover the new site, both directions.

The gate now agrees across **68 pin sites** (was 67).

**`exactOptionalPropertyTypes`, re-measured.** The old note leaned partly on an argument that no longer holds — that the flag "would land on the vendored `osfacts-client` this repo cannot edit". osfacts is ours, it lives in its own repo, and its tsconfig turns the flag **ON** upstream. So the argument was retired and the number re-taken: added to `tsconfig.base.json`, ran `pnpm -r --no-bail --filter='!osfacts-client' typecheck`, then reverted.

> **922 errors across 36 of kolu's OWN packages — 249 distinct files, 490 distinct errors once a file reported by several packages is counted once. ZERO from the graft.**

Still hundreds, so the decision does not change and now stands on the measurement alone: the blast radius is kolu's own code, and every one of those sites is a separate judgement about whether a key should be absent or present-undefined. Enumeration remains the remedy. Both `optionalTolerance.ts`'s header and `check.ts`'s restatement now say that from the new number. (Top codes: 593 × TS2379, 145 × TS2375, 62 × TS2345.)

**Byte-compat.** `ProcessRss` (`terminal-vocab` :269) and `PortInfo` (:315) are green and untouched. The gap was `PadiProcessMemory` — the struct that carries *two* `ProcessRss` over padi's `processMemory` cell had no byte fixture at all, so key order and a stale `rssBytes` surviving on a non-`ok` arm rested on decode equality, which is blind to both. It has one now, in `vocabByteCompat.test.ts` style, and it matters more since `samplePadiMemory` composes those arms from an Effect fold rather than an async try/catch.

**No `BETA-ASSUMPTION` markers added.** Nothing here depends on beta *behavior* — only on the client's declared API and on `Effect.catch`/`mapError`/`suspend`, which are surface, not semantics. The registry stays at 3.

## Downstream impact

- **odu — none, and grounded.** odu is pinned at kolu `41d517754` (this PR's base). It hydrates exactly six kolu packages: `@kolu/surface`, `surface-mcp`, `surface-remote`, `surface-map`, `shell-quote`, `log`. `surface-daemon-supervisor` is not among them, and a grep of the whole repo for `surface-daemon-supervisor` / `readSocketHolders` / `processIdentityAsync` / `osfacts` returns nothing.
- **drishti — nothing owed *by this PR*.** The supervisor's public API did change, but drishti consumes only the CONNECTOR arm (`convergeAdmit`, `createConnectorDrainBudget`, `drainAndAwaitExit`, `drainRejectionSuffix`, `probeDaemonIdentityFrom`, `instanceKeyFromStartedAt`, `DrainableProbe`, `InstanceKey`, `ConvergenceAnomaly`) and never calls `createEndpoint` or spells `EndpointSpec` — so neither flipped seam reaches it. It DOES owe an osfacts#5 adoption of its own: `packages/agent/src/proc.ts` uses `snapshotHost`, which is now Effect-returning. Its `processIdentity` / gate path is the sync island and is unaffected. That pair PR is driven by the upstream client, not by this one.

## Merge order

1. #2101 merges first (this is based on `effect`).
2. Then [juspay/osfacts#5](https://github.com/juspay/osfacts/pull/5).
3. Then this PR re-pins `osfacts` to that merge commit and retargets `master`.

---


srid:
### e2e timing — run `7d58e57#1`

Both platforms settled green (48 ok · 0 failed · 0 errored). Recipe-wall durations from the run ledger `.ci/7d58e57/runs/1.json`; parallelism from the last `e2e: workers=` line in each lane's `ci::e2e.log`.

| `x86_64-linux` | `kolu-ci-1` | 8 | 32 | 8 | **3m 26s** (206,443 ms) |
| `aarch64-darwin` | `ci@petit` | 4 | 10 | 6 | **4m 24s** (263,623 ms) |

Local full run on this HEAD: **509 scenarios / 4,672 steps, all passed**, 8m 07s wall (0 retries).

srid:
## Campaign disposition — downstream half (re-walked at final HEAD `7d58e57`)

Per the definition of done in [juspay/kolu#2101 (comment)](https://github.com/juspay/kolu/pull/2101#issuecomment-5182758828):

- **Zero Promise-typed members on the supervisor's spec** ✓ — every `EndpointSpec` member read post-flip: data, Effects (both OS-fact seams included), synchronous-by-contract callbacks, numeric knobs. A `@ts-expect-error` pair in `endpoint.public-api.test-d.ts` makes a Promise-returning inject a compile error, so the seam can't silently widen back. Both `Effect.tryPromise` rinds (`endpoint.ts:787`, the `Promise.resolve` lift at `:830`) are deleted; the stale "osfacts declares no effect dependency" doc block rewritten.
- **Green CI both platforms at the final pin, receipts** ✓ — run `7d58e57#1`: 48 ok / 0 failed / 0 errored (linux `kolu-ci-1`, darwin `ci@petit`); 46/46 required contexts, 51 total checks green, code scanning (`pr=2103`) clean. Checks: https://github.com/juspay/kolu/pull/2103/checks · e2e metrics comment on this PR (marker `kolu-ci-e2e-metrics`). Local: 509/509 e2e, 0 retries, before push.
- **Byte-compat unchanged for every osfacts shape over the wire** ✓ — `SocketOccupancy` is supervisor-internal (no bytes; the compile-time pin covers it), `ProcessRss` and `PortInfo` fixtures untouched and green, and the gap the census found — `PadiProcessMemory` had **no** byte fixture — is closed with an encoded-JSON fixture in `vocabByteCompat.test.ts`.
- **Governance log line reflects the new totals** ✓ — run edges 52 files/106 sites → **53/107**, measured against the base tree: the one added row (`padi/src/ports/sampler.ts`) carries its written why (the reactor's poll `read` is Promise-shaped by design; the Effect-native scan meets it at that documented edge), and the two deleted rinds were the actual Promise boundary removed. `effectPin` no longer skips the graft — `osfacts-client` joined the vendored-literal category with its rationale (and its own gate tests). `exactOptionalPropertyTypes` re-probed with osfacts editable: **922 errors across 36 of kolu's own packages, zero from the graft** — the "cannot edit" argument is retired and the enumeration decision now stands on the fresh number alone, in both `optionalTolerance.ts` and `check.ts`.
- **The port's bar held**: falsified test per behavior change (including the one silent behavior change caught in review-by-construction — an unbaked `KOLU_OSFACTS_BIN` stays the sampler's per-tick `error` arm, not a cell-killing defect); no new unbounded wait/retry/fallback; no new beta-behavior dependency, so no new `BETA-ASSUMPTION` marker (the existing three stand and the gate is green).
- **Consumer impact**: odu verdict **`none`**, grounded (odu pins this PR's exact base and hydrates no supervisor package; whole-repo grep empty). drishti owes nothing for the seam flip (it consumes only the connector arm) but does adopt the upstream client directly — pair PR **srid/drishti#133**, open and fully green (24 ok odu run `73c4f98#1`, 28/28 checks) at kolu pin `7d58e57`.

**Merge — parked, with the ordering**: kolu#2101 → juspay/osfacts#5 → this PR (re-pinning osfacts to the merge commit, retargeting master) → drishti#133 re-pins last. Merging is the maintainer's act; all four PRs are complete and green.

srid:
Folded into juspay/kolu#2101 by maintainer decision — the `effect` branch fast-forwarded to this PR's head `7d58e57`, so #2101 now carries the osfacts adoption (its CI green at this exact SHA carries over: run `7d58e57#1`, 48 ok both platforms). The [disposition](https://github.com/juspay/kolu/pull/2103#issuecomment-5183762787) and receipts remain valid against #2101's head. Closing; one PR per repo.

PR #2115, opened 2026-08-07
Grok: a clobbered active_sessions row must not unbind a live session

srid:
A Grok agent on a terminal finished its turn at 16:01 and the tile still read **Thinking** at 16:05. `padi-tui wait --until awaiting,waiting`, pointed at that terminal since 15:58, was still blocked. This is that bug, and it had been quietly freezing Grok tiles for a while.

## What was actually happening

padi finds a Grok session by looking up the terminal's foreground pid in `~/.grok/active_sessions.json`. During the incident that file read `[]` while grok pid 1031582 was alive and mid-turn.

Grok rewrites the whole array from a process-local snapshot, so any *concurrent* grok starting or exiting erases the rows of every other live grok. padi's log shows it happening three times in the same worktree, each retirement landing within tens of milliseconds of another grok terminal starting or being killed:

```
14:32:33  77dea2af  watcher installed   (grok #1 registers)
14:37:23  49b0c273  watcher installed   ← 36ms later: 77dea2af retired
14:57:22  e822054f  watcher installed   ← same ms: 49b0c273 retired
14:59:42  77dea2af  "killing" pid 987319
14:59:42  e822054f  watcher retired     ← 53ms later. Never reinstalled.
```

`resolveSession` returned `null` for the still-running grok, so the sensor tore down the session watcher. The foreground was still a defined non-shell pid, which is the deliberate keep-last `unknown` arm — `foldSnapshot` returns the snapshot untouched, holding the last published agent info. Nothing re-derives after that: resolve keeps returning `null`, the dedup early-returns, and the Grok adapter has no screen-scrape poll to rescue it. Frozen for the life of the process, with `events.jsonl` sitting right there carrying the correct answer (`turn_ended` → `waiting`).

## The fix

`active_sessions.json` is how a pid **acquires** its session, never how it loses it. Once a pid has resolved to a session, a later rewrite that drops its row no longer unbinds it.

The release signals are the ones that actually mean something:

| Signal | Effect |
|---|---|
| The process dies | Swept on every resolve — one `kill(pid, 0)` over a handful of entries |
| Foreground moves off the pid | The caller stops asking; the sensor's own shell-idle path clears the agent |
| A present row names a different session | Re-binds — the map still wins when it has something to say |

A pid that was never in the map still resolves to nothing, so the existing ban on guessing a session by cwd (which would flash the wrong state during the arrival race) is untouched.

The sweep runs on every resolve rather than only on acquire, which matters: grok's exit rewrites `active_sessions.json`, that wakes a reconcile, and the dead pid is swept within milliseconds of the death that ended it — so there is no meaningful window in which a recycled pid could inherit a stale session.

## Tests

Both reproductions were written first and confirmed failing against the unfixed code.

**Unit** (`core.test.ts`) — the tight root cause. A pid acquires a session, the row is clobbered, and resolution must survive. It uses `process.pid`, the one pid a test can be sure is alive, because liveness is the entire precondition. The release path gets its own test using a genuinely reaped pid from `spawnSync`, and two guards pin the boundaries: one pid's binding is never lent to another, and a present row still re-binds.

**E2E** (`grok.feature`) — the user-visible freeze, driven through the real tile chrome:

```gherkin
When a Grok session is mocked with state "thinking"
Then the tile chrome should show a Grok indicator with state "thinking"
When a concurrent Grok clobbers the active_sessions map
And the Grok session state changes to "waiting"
Then the tile chrome should follow the Grok state change to "waiting" without nudging
```

Against the unfixed code this failed with `got state="thinking"` — the production symptom exactly. The 1500ms settle inside the clobber step is load-bearing rather than padding: the clobber has to be consumed before the state flip, or the outgoing watcher could publish the new state on its way out and the scenario would pass without proving anything. It is more than 25x the observed 53ms teardown.

## What was run

`just test-unit` (full lane, green), `pnpm typecheck`, `just fmt`, and `just test-quick features/grok.feature` (3/3). Full odu CI has not been run yet.


srid:
## Evidence

Ran the new e2e scenario (`features/grok.feature` — *Tile chrome keeps following Grok after a concurrent grok clobbers active_sessions*) on the same box, once against this branch and once with the fix file (`packages/integrations/grok/src/core.ts`) reverted to `master`: a concurrent grok clobbers `active_sessions.json` mid-session, then the live session flips to "waiting" — the clip shows whether the tile chrome follows.

| Before (frozen on Thinking) | After (follows to Waiting) |
| --- | --- |
|  |  |

Before: the resolver drops the running grok's row when the clobbering write lands, padi reads that as "session ended," and the tile chrome freezes on **Thinking** — the scenario fails with `got state="thinking"` ~20s after the clobber, never reaching "waiting." After: the acquire-only fix keeps the row alive through the clobber and the tile chrome follows to **Waiting for input**:



▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=grok-clobber-after.mp4
▶ HD (before/bug): https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=grok-clobber-before.mp4

Mobile / coarse-pointer: unaffected — this change only alters which state the existing agent indicator reports (frozen vs. following), with no layout, sizing, or hit-target change, so small-viewport and coarse-pointer rendering are unchanged.

srid:
### CI e2e metrics — `63494ad#1`

Two-platform run, settled green (48/48 required contexts).

| `x86_64-linux` | `kolu-ci-1` | 8 | 32 | 8 | 3m 44s (224452 ms) |
| `aarch64-darwin` | `ci@petit` | 4 | 10 | 6 | 4m 44s (284309 ms) |

Durations are the `ci::e2e@<platform>` recipe wall time from the durable ledger `.ci/63494ad/runs/1.json`; worker counts are the last `e2e: workers=` line each lane resolved.

PR #2117, opened 2026-08-08
kolu mcp: stop spending a request to find out padi restarted

### LINKED ISSUE #2082: kolu mcp: the first padi-backed request after a padi restart always fails, with a message naming the wrong layer
`kolu mcp` survives a padi restart, but the **first padi-backed request after it always fails** with a message that names the wrong layer:

```
stdio transport is closed (the peer process exited or its stream ended); request not sent.
```

The MCP server process is alive. padi is alive. Every request *after* that one succeeds.

## Reproduction

Deterministic — **3/3 runs**, fully isolated (its own state root and socket; the production padi is never touched). Self-resolving: it reads the store-built kolu, node and tsx off a running padi, so there are no store paths to hand-edit.

<details><summary><code>mcp-padi-restart-repro.mjs</code> — <code>node mcp-padi-restart-repro.mjs</code></summary>

```js
#!/usr/bin/env node
// Repro: does `kolu mcp` survive a padi restart?
//
// Isolated end to end — its own state root and socket, so the production padi
// is never touched. Uses the SAME store-built kolu the .mcp.json entry runs.
import { spawn } from "node:child_process";
import { mkdtempSync, existsSync, rmSync, mkdirSync } from "node:fs";
import { tmpdir } from "node:os";
import { join } from "node:path";

// Resolve the SAME store-built kolu the `.mcp.json` entry runs, plus the node
// + tsx loader it was built against — read them off a running production padi
// so this script needs no hand-edited store paths.
import { execSync } from "node:child_process";
const ps = execSync(
  "ps -eo cmd | grep 'kolu-[0-9].*/packages/padi/src/daemonBoot/bin.ts' | grep -v -- --stdio | head -1",
  { encoding: "utf8" },
).trim();
if (!ps) {
  console.error("no running store-built padi found — start kolu, or hardcode KOLU/NODE/TSX below");
  process.exit(2);
}
const NODE = ps.split(/\s+/)[0];
const TSX = (ps.match(/--import (\S+)/) ?? [])[1];
const KOLU = (ps.match(/(\/nix\/store\/[^/]*kolu[^/]*)\/packages\/padi/) ?? [])[1];
console.log(`using kolu=${KOLU}`);

const root = mkdtempSync(join(tmpdir(), "mcp-repro-"));
const stateRoot = join(root, "padi-state");
const sockDir = join(root, "sock");
const sock = join(sockDir, "padi.sock");
import { chmodSync } from "node:fs";
mkdirSync(stateRoot, { recursive: true });
mkdirSync(sockDir, { recursive: true });
chmodSync(stateRoot, 0o700);
chmodSync(sockDir, 0o700);

const log = (...a) => console.log(...a);
const sleep = (ms) => new Promise((r) => setTimeout(r, ms));

function startPadi(tag) {
  const p = spawn(
    NODE,
    ["--import", TSX, `${KOLU}/packages/padi/src/daemonBoot/bin.ts`,
     "--state-root", stateRoot, "--socket", sock, "--spawn-version", "2.0.0",
     "--allow-nix-shell-with-env-whitelist", "default"],
    { stdio: ["ignore", "pipe", "pipe"], detached: false },
  );
  p.errBuf = "";
  p.stdout.on("data", () => {});
  p.stderr.on("data", (b) => { p.errBuf += b.toString(); });
  log(`[padi:${tag}] spawned pid=${p.pid}`);
  return p;
}

async function waitForSocket(timeoutMs = 30000) {
  const deadline = Date.now() + timeoutMs;
  while (Date.now() < deadline) {
    if (existsSync(sock)) return true;
    await sleep(200);
  }
  return false;
}

// ── MCP stdio client: newline-delimited JSON-RPC ────────────────────────────
function startMcp() {
  const m = spawn("/nix/var/nix/profiles/system/sw/bin/kolu", ["mcp"], {
    stdio: ["pipe", "pipe", "pipe"],
    env: { ...process.env, PADI_SOCKET: sock },
  });
  m.stderrBuf = "";
  m.stderr.on("data", (b) => { m.stderrBuf += b.toString(); });
  m.pending = new Map();
  m.buf = "";
  m.stdout.on("data", (b) => {
    m.buf += b.toString();
    let i;
    while ((i = m.buf.indexOf("\n")) >= 0) {
      const line = m.buf.slice(0, i); m.buf = m.buf.slice(i + 1);
      if (!line.trim()) continue;
      try {
        const msg = JSON.parse(line);
        if (msg.id !== undefined && m.pending.has(msg.id)) {
          m.pending.get(msg.id)(msg); m.pending.delete(msg.id);
        }
      } catch {}
    }
  });
  m.exited = null;
  m.on("exit", (code, signal) => { m.exited = { code, signal }; });
  log(`[mcp] spawned pid=${m.pid} PADI_SOCKET=${sock}`);
  return m;
}

let nextId = 1;
function call(m, method, params, timeoutMs = 15000) {
  return new Promise((resolve) => {
    if (m.exited) return resolve({ transport: "dead", exited: m.exited });
    const id = nextId++;
    const t = setTimeout(() => { m.pending.delete(id); resolve({ timeout: true }); }, timeoutMs);
    m.pending.set(id, (msg) => { clearTimeout(t); resolve(msg); });
    try {
      m.stdin.write(`${JSON.stringify({ jsonrpc: "2.0", id, method, params })}\n`);
    } catch (e) {
      clearTimeout(t); m.pending.delete(id);
      resolve({ writeError: String(e) });
    }
  });
}
function notify(m, method, params) {
  try { m.stdin.write(`${JSON.stringify({ jsonrpc: "2.0", method, params })}\n`); } catch {}
}

const brief = (r) => {
  if (!r) return "no response";
  if (r.transport === "dead") return `MCP PROCESS ALREADY DEAD (exit code=${r.exited.code} signal=${r.exited.signal})`;
  if (r.timeout) return "TIMED OUT (no response)";
  if (r.writeError) return `WRITE FAILED: ${r.writeError}`;
  if (r.error) return `JSON-RPC error: ${JSON.stringify(r.error).slice(0, 220)}`;
  if (r.result?.tools) return `ok — ${r.result.tools.length} tools`;
  if (r.result?.content) return `ok — ${JSON.stringify(r.result.content).slice(0, 160)}`;
  return `ok — ${JSON.stringify(r.result).slice(0, 160)}`;
};

(async () => {
  log("=== STEP 1: start an isolated padi ===");
  let padi = startPadi("first");
  if (!(await waitForSocket())) { log(`padi socket never appeared — abort\n${padi.errBuf.slice(0,600)}`); process.exit(2); }
  log(`[padi:first] socket up: ${sock}\n`);

  log("=== STEP 2: start `kolu mcp` against it ===");
  const m = startMcp();
  await sleep(1500);

  const init = await call(m, "initialize", {
    protocolVersion: "2024-11-05",
    capabilities: {},
    clientInfo: { name: "repro", version: "0" },
  });
  log(`initialize → ${brief(init)}`);
  notify(m, "notifications/initialized", {});
  await sleep(500);

  log("\n=== STEP 3: PADI-BACKED calls BEFORE the restart (baseline) ===");
  log(`tools/list           → ${brief(await call(m, "tools/list", {}))}   [static, never touches padi]`);
  log(`resources/read       → ${brief(await call(m, "resources/read", { uri: "surface://collections/terminals" }))}`);
  log(`tools/call wait_*    → ${brief(await call(m, "tools/call", { name: "screen_text", arguments: { id: "00000000-0000-4000-8000-000000000000" } }))}`);
  log(`[mcp] alive? ${m.exited ? `NO (${JSON.stringify(m.exited)})` : "yes"}`);

  log("\n=== STEP 4: restart padi (SIGTERM the old one, start a new one on the same socket) ===");
  padi.kill("SIGTERM");
  await sleep(3000);
  log(`[padi:first] exited? ${padi.exitCode !== null || padi.signalCode !== null ? `yes (code=${padi.exitCode} signal=${padi.signalCode})` : "still running"}`);
  try { rmSync(sock, { force: true }); } catch {}
  padi = startPadi("second");
  const back = await waitForSocket();
  log(`[padi:second] socket back up? ${back}`);
  await sleep(2000);

  log("\n=== STEP 5: the same tool call AFTER the restart ===");
  log(`[mcp] alive? ${m.exited ? `NO (${JSON.stringify(m.exited)})` : "yes"}`);
  log(`tools/list           → ${brief(await call(m, "tools/list", {}))}   [static]`);
  log(`resources/read       → ${brief(await call(m, "resources/read", { uri: "surface://collections/terminals" }))}`);
  log(`tools/call screen_text → ${brief(await call(m, "tools/call", { name: "screen_text", arguments: { id: "00000000-0000-4000-8000-000000000000" } }))}`);
  await sleep(2000);
  log("  — retrying, to show whether it is transient or permanent —");
  for (const n of [1, 2, 3]) {
    log(`  resources/read #${n}  → ${brief(await call(m, "resources/read", { uri: "surface://collections/terminals" }))}`);
    log(`  tools/call    #${n}  → ${brief(await call(m, "tools/call", { name: "screen_text", arguments: { id: "00000000-0000-4000-8000-000000000000" } }))}`);
    await sleep(2000);
  }

  await sleep(1000);
  log("\n=== RESULT ===");
  log(`kolu mcp process: ${m.exited ? `EXITED code=${m.exited.code} signal=${m.exited.signal}` : "still alive"}`);
  log(`kolu mcp stderr: ${m.stderrBuf.trim() ? `\n${m.stderrBuf.trim().slice(0, 800)}` : "(empty)"}`);

  try { m.kill("SIGKILL"); } catch {}
  try { padi.kill("SIGKILL"); } catch {}
  await sleep(500);
  try { rmSync(root, { recursive: true, force: true }); } catch {}
  process.exit(0);
})();

```
</details>

## Observed output

Run A — `resources/read` is the first padi-backed request after the restart:

```
=== STEP 3: PADI-BACKED calls BEFORE the restart (baseline) ===
tools/list           -> ok - 11 tools   [static, never touches padi]
resources/read       -> ok - {"contents":[{"uri":"surface://collections/terminals",...,"text":"[]"}]}
tools/call screen_text -> ok - [{"type":"text","text":"surface-mcp: Terminal 000...000 not found"}]
[mcp] alive? yes

=== STEP 4: restart padi (SIGTERM the old one, start a new one on the same socket) ===
[padi:first] exited? yes (code=0 signal=null)
[padi:second] socket back up? true

=== STEP 5: after the restart ===
[mcp] alive? yes
tools/list           -> ok - 11 tools   [static]
resources/read       -> JSON-RPC error: {"code":-32603,"message":"stdio transport is closed (the peer process exited or its stream ended); request not sent."}
tools/call screen_text -> ok - [{"type":"text","text":"surface-mcp: Terminal 000...000 not found"}]
  -- retrying --
  resources/read #1  -> ok
  tools/call    #1  -> ok
  resources/read #2  -> ok
  tools/call    #2  -> ok
  resources/read #3  -> ok
  tools/call    #3  -> ok

=== RESULT ===
kolu mcp process: still alive
kolu mcp stderr: (empty)
```

Run B — same script, but a **tool call** is put first after the restart:

```
=== STEP 5: after the restart ===
[mcp] alive? yes
tools/call screen_text #1 -> ok - [{"type":"text","text":"surface-mcp: stdio transport is closed (the peer process exited or its stream ended); request not sent."}]
resources/read        #1 -> ok
tools/call screen_text #2 -> ok - [{"type":"text","text":"surface-mcp: Terminal 000...000 not found"}]
```

## Facts established

1. `kolu mcp` **survives** a padi restart — the process stays alive, stderr is empty, no non-zero exit.
2. The **first padi-backed request after the restart fails**; every subsequent request succeeds. The redial happens *after* a request has already been spent on the dead connection, not before it.
3. The failure is **positional, not request-type-specific** — whichever padi-backed request goes first eats it (Run A: `resources/read`; Run B: `tools/call`).
4. `tools/list` never fails, because it is answered statically and never reaches padi.
5. The two request types **report the same failure differently**:
   - `resources/read` -> a JSON-RPC error, `code: -32603`
   - `tools/call` -> a **successful** result whose `content` text is the error string
6. The message says `stdio transport is closed (the peer process exited or its stream ended)`. At the moment it is emitted, the MCP stdio peer has not exited and padi is running. The wording describes the inner padi connection but reads as though the MCP server itself died.

## Where it bit in production

Same string, same shape, on `naiveintent` (a remote host of a kolu running on `pureintent`):

- `~/.local/state/padi/padi.log.3` — padi restarted and adopted its terminals:
  ```
  18:17:56  pid 1432366  session-trace capture: snapshot=22 -> persisting
  18:18:01  pid 1611042  adopted a surviving daemon (its PTYs are preserved)
  18:18:01  pid 1611042  adopted surviving terminals after restart
  ```
  (`ps` confirms the new padi pair started 18:17:58-59; `/nix/var/nix/profiles/system` was relinked to generation 155 at 18:22:33.)

- `~/.cache/claude-cli-nodejs/<project>/mcp-logs-kolu/2026-07-30T16-20-11-669Z.jsonl` — the agent's last successful call, then its next one, which was the first request after that restart:
  ```
  18:44:11Z  Tool 'lifecycle_kill' completed successfully in 37ms
  23:15:47Z  Calling MCP tool: lifecycle_create
  23:15:47Z  surface-mcp: stdio transport is closed (the peer process exited or its stream ended); request not sent.
  ```

The agent read that message as "the MCP server is dead", stopped using MCP for the rest of the session, and fell back to `padi-tui`/`kaval-tui` for every terminal operation. A single retry would have worked.

## Why it matters

padi restarts are routine — an upgrade does it, and the restart deliberately preserves PTYs so nothing else notices. Every other client reconnects. For an agent driving kolu over MCP, one request per restart is destroyed, and the message it gets points at the wrong layer, so the natural response is to abandon MCP rather than retry.

## Relevant code

`packages/kolu-mcp/src/serve.ts:49-51` states the intended contract:

> `connect`: Produce a connected padi client. **Re-invoked after a transport drop** - MUST re-resolve + re-dial + re-gate, never cache a dead socket.

`packages/kolu-cli/src/mcp.ts` wraps that factory so an unreachable padi surfaces as a retryable error (`padi transport down: ... (retryable ...)`) - which is *not* the message observed above.

Filed from a session that hit this; the repro script is the deliverable.

srid:
padi restarts routinely — an upgrade does it — and the restart deliberately preserves PTYs so nothing else notices. `kolu mcp` came through alive too, but **the first padi-backed request after every restart was destroyed**, and every later one worked. The message it died with named the wrong layer:

```
stdio transport is closed (the peer process exited or its stream ended); request not sent.
```

An agent reads that on its own stdio channel and concludes the MCP server exited. In the field session behind #2082 it did exactly that: stopped calling MCP at 23:15 over a padi restart that had happened at 18:17, and drove terminals through `padi-tui`/`kaval-tui` for the rest of the run. A single retry would have worked.

### The cause is one dropped field

The adapter memoises one padi connection for reads and bespoke tools, and it only ever dropped that connection **inside `withClient`'s catch** — so the only way to discover a restart was to send a request into the dead socket and watch it fail. The redial machinery was all correct and all documented; what was missing was not *where* a restart heals but *when* the adapter finds out.

padi has been announcing it the whole time. `connectPadi` hands back a `DaemonConnection` whose `onClose` fires when the socket closes — and `connectKoluCliLocal` dropped that field on the floor projecting it onto `KoluCliConnection`. Carry it through and the corpse is discarded the instant padi goes away, so the next request dials fresh and nothing is spent.

```
padi exits ──▶ socket close ──▶ onClose ──▶ connection discarded (idle)
                                                    │
                        agent's next request ───────┴──▶ fresh dial ──▶ ✓
```

### What changed

| | Before | After |
| --- | --- | --- |
| How a restart is discovered | a request fails on the dead socket | the transport announces its close |
| Cost of a restart | exactly one destroyed request, every time | zero |
| Mid-flight death (the genuine race) | same message, naming the stdio link | names the *connection*, says the server is alive, says retry |

Two paths now drop a dead connection, and the lazy one is deliberately kept: `onClose` covers the ordinary restart, and the catch-side reset still covers a transport with no close signal to offer and the real race where the socket dies with a request already in flight. *A retry-and-replay would have closed that race too — and re-sent mutations against a fresh daemon generation, which is the two-clocks bug the surrounding code refuses on purpose. Nothing is queued and nothing is re-sent.*

`onClose` is **optional, and honestly so**. The local unix-socket arm supplies it because padi's contract carries it. The `--host` ssh arm dials through `dialAgentOnce`, and `AgentDial` has no close signal at all — so it omits the field and keeps today's behaviour rather than faking one. Teaching `@kolu/surface-remote` to report its close is the follow-up; it is not smuggled in here.

> The `resources/read` (`-32603` JSON-RPC error) vs `tools/call` (`isError` result) split the issue notes as fact #5 is **left alone** — that difference is MCP-spec-correct, not a defect.

### Proof

The existing restart leg could never have caught this: it calls *during* the gap, so the wasted request is spent on a call that was going to fail anyway (padi is down), which hides it. The new leg restarts padi across an **idle** gap and asserts the very first request afterwards lands — no retry, no warm-up. Red before, green after, on real spawned padis. Evidence comment to follow.

**Odu-impact verdict: `none`.** Grepped odu at its pinned kolu SHA (`2104ddea`): `src/cli/mcp.ts` passes `client: () => bClient` — the bare-client branch of `ClientOrConnection`, untouched by an added optional field on the owned-connection branch. Its freshness lives in `redialingAClient`, which dials `.ci/odu.sock` per call and so never holds a stale connection to invalidate. No pair PR for drishti either: `@kolu/surface-mcp` is not on that gate's package list, and nothing here touches one that is.

Closes #2082


srid:

**Status: applied.** Two Opus lenses read the diff independently and in parallel — lowy raised 8 findings, hickey 11. One reconcile-and-apply pass merged them into six commits. Four findings were the *same* finding reached from two directions (`≡`), which is what consensus looks like when nobody is comparing notes.

The two that mattered most were both `≡`, and neither was in the change I set out to review.

**The pusher was leaking sockets.** `dial()` returns a whole connection — client, disposer, close announcement — and the shared path keeps it whole, which is exactly why its identity guard works. The pusher path took it apart: a bare client, the disposer filed in a `WeakMap` keyed on the *client* object, and `onClose` dropped on the floor. When a factory hands back a stable client across dials — which `server.test.ts`'s own `concurrencySurface()` does, and which the new #2082 tests use — two concurrent `ensureAttached` dials make B's `WeakMap.set` overwrite A's disposer, and A's socket is never closed for the life of the process. That is the same "a projection quietly forgot a field" shape as #2082's root cause, reproduced inside the package fixing it. The pusher now owns whole connections; the leak has a failing-before/passing-after pin.

**One lifecycle, not three slots.** `sharedConn` / `closed` / `dialing` were three mutable places encoding one lifecycle, and this PR added a fourth writer on a third clock (the transport's own `onClose`). None of the contradictory combinations turned out to be reachable — but only because six functions all remember the same rules, which is not a property of the shape. And the encoding already had a hole: `closed` gated the *middle* of a dial but not its *entry*, so a request landing after teardown really opened a socket just to dispose it. One `ConnState` union (idle · dialing · live · closed) closes that for free and turns every guard into a tag test.

### Findings

| id | lens | finding | outcome |
| --- | --- | --- | --- |
| F3 ≡ F3 | lowy ∥ hickey | pusher sheds `onClose`/`dispose` into a client-keyed side table — real socket leak | applied `4666e478` |
| F1 + F2 | hickey | one lifecycle as three slots; `closed` doesn't gate dial entry | applied `efb0e79a` |
| F2 ≡ F7 | lowy ∥ hickey | `KoluMcpConnection` still a hand-copy — the export added by this PR was left unused | applied `a4989ffb` |
| F4 | lowy | **"ssh cannot observe its own close" is false** — see below | applied `98450440` |
| F5 | hickey | the "don't let the host conclude the server died" policy applied to 1 of 3 messages; born-dead throw fixed | applied `efb0e79a` |
| F4 | hickey | four prefix sites, three conventions; born-dead error was double-prefixed `surface-mcp: surface-mcp:` | applied `efb0e79a` |
| F8 ≡ F9 | lowy ∥ hickey | `ClientOrConnection<_S>` phantom type parameter | applied `24c4301e` |
| F8 | hickey | ssh arm still built from an inline literal — the shape the local extraction just removed | applied `24c4301e` |
| F7 | lowy | `BORN_DEAD_REDIALS` is a count where dial cost spans orders of magnitude | applied `efb0e79a` |
| F11 / F6 | hickey ∥ lowy | one narrative, twelve copies | applied `98450440` |
| F1 ≡ F10 | lowy ∥ hickey | third hand-rolled "hold one connection, drop on guarded `onClose`" | **recorded, not extracted** `d568feaf` |
| F6 | hickey | make the local arm's `onClose` required in type | **adjudicated: dropped** |

### lowy caught me stating a falsehood

I wrote, in six places, that the ssh `--host` transport *cannot observe its own close*. That is not true. The signal exists — `Connection.closed` at `surface-remote/src/session.ts:448`, and `sshConnector.ts:246-278`. What is missing is only a **field on `AgentDial` to carry it**. So the code asserted a limit of ssh where the truth was a missing projection at a hop — which is, precisely, the same omission that *was* #2082. The wording is corrected everywhere; `@kolu/surface-remote` is deliberately untouched, so the projection stays the follow-up.

### Recorded, not extracted

`@kolu/surface-daemon-supervisor`'s `holdConnection` (`endpoint.ts:995-1017`) already does the identity-guarded `onClose`→drop in almost the same words. With surface-mcp's slot and odu's `redialingAClient`, that is three hand-rolled copies of one mechanism — lowy names the receptacle `holdOneLink` with `@kolu/surface` as its home. A third instance is the moment to *name* the concept, not yet to graduate it, so it is now a row in the electricity ledger rather than a fourth home.

### Adjudicated (valve 2)

**hickey F6** — a `KoluCliLocalConnection` narrowing that makes the local arm's `onClose` required in type. *hickey:* the local arm always supplies it, so the type should say so. *Against:* this is the same ground as the `CloseSignal` sentinel union an adversarial verifier already judged a design preference rather than a defect, and the human explicitly chose optional-now / remote-as-follow-up. **Dropped** — recorded as an observation. It becomes worth revisiting when `AgentDial` grows the field, because then *both* arms supply it and the optionality has no remaining justification.

<sub>Verified before commit: `just check` clean · surface-mcp 76 passed · kolu-cli 19 passed, 4 skipped (daemon gate untouched) · `just fmt` no changes.</sub>

srid:
## Grok ⇄ Claude Code debate

✅ Consensus in 1 round(s), against base `321a2583`.

_No fix commits: Grok raised **no findings at any severity** and approved on the first round._

Grok 4.5 reviewed the full diff read-only and was pointed at the four places I most doubted the change — the bounded born-dead redial loop, the fresh `ConnState` union rewrite (attacked for regressions against the pre-union behaviour), the pusher’s new whole-connection ownership on the `stop()` / `unsubscribe` / retry paths, and whether the tests added across all three passes would pass against the un-fixed code. Its verdict, verbatim:

> The fix for #2082 is sound: onClose is threaded through named projections so the shared MCP connection slot drops eagerly on announcement, the ConnState union closes the post-teardown dial hole, and the bounded born-dead loop re-dials without replaying requests. Identity-guarded dropConn/onAnnouncedClose correctly ignore late or duplicate closes; the pusher’s whole-connection ownership fixes the WeakMap disposer leak without double-dispose on stop/unsubscribe/retry; and the new unit plus idle-restart e2e pins fail against the unfixed shapes they claim to cover. Deliberate choices (optional onClose, retained lazy catch, no request replay, holdOneLink deferred) hold under re-examination. No blocking, major, or minor defects found beyond what prior passes already fixed.

<sub>Two earlier passes had already landed fixes in the diff Grok reviewed — the architecture-first-principles pass and the lowy ∥ hickey lens review — so this was a fresh adversarial read of the *post-fix* tree, not a first look.</sub>

srid:

Two findings, both actioned in `3f3bb7892`. Ran with `--no-elegance` (the `/simplify` track had just swept this same tree).

**Rules — `prefer-ts-pattern`.** `dialShared`'s four-arm `switch` over the new `ConnState` union is dispatch-with-logic (the `idle` arm has a side effect), so it takes `match(...).exhaustive()`. Complied rather than disputed: `ts-pattern` is already a dependency of four other surface-family packages, so there is no graduation-fence argument, and I checked the two costs before agreeing — odu hydrates `surface-mcp` into its own `node_modules` and already carries `ts-pattern` directly, so the odu-impact verdict stays **`none`**; and the lockfile change does **not** move the `fetchPnpmDeps` hash, because `ts-pattern@5.9.0` was already in it for eleven packages and the *fetched set* is therefore unchanged. Verified by a full `nix build .#default`, which passes on the existing hash.

**Fact-check — a false claim in the electricity ledger.** The `holdOneLink` row claimed **four** hand-rolled implementations of the memoize-one-connection / identity-guarded-drop pattern, counting odu's `redialingAClient` as the fourth. It is not one. `redialingAClient` opens a fresh `.ci/odu.sock` for *every* upstream call and holds nothing — it has no slot to invalidate and pays a dial per call precisely so it never needs one. That is the **opposite** of this pattern.

The row now says three, and names odu as the genuine *alternative* it is: a consumer can buy the same freshness by making dials cheap enough not to memoize. That weakens this row's own test-③ population evidence, which is why it is stated rather than quietly deleted — the row exists to be honest about how strong the case for extraction is, and it is now one instance weaker than it claimed.

Worth naming the pattern: this is the **third** false claim caught in this branch, after a docs sentence promising a mid-flight failure meant no terminal was created (padi commits before writing the response) and six code sites asserting ssh "cannot observe its own close" (it can; only the `AgentDial` field to carry it is missing). All three were prose asserting behaviour the code does not have — the failure mode a comment-heavy change is most prone to, and the one a test cannot catch.

<sub>`just check` clean · surface-mcp 76/76 · kolu-cli 19 passed, 4 skipped (daemon gate off) · `nix build .#default` green.</sub>

srid:
## Evidence

The issue ships its own repro and calls it the deliverable, so that is the acceptance test. Run **on a clean pu box against real spawned padis**, same script, two trees — `321a2583` (master) and `3f3bb789` (this branch). Each run is fully isolated: its own state root, its own socket, no production daemon anywhere near it.

The script is the issue's `mcp-padi-restart-repro.mjs` step for step, with one change: it resolves `kolu`/`node`/`tsx` from a **worktree** rather than reading them off a running production padi, which is what lets the identical script run against both trees on a box where no production kolu exists.

### Before — `321a2583` (master)

```
=== STEP 4: restart padi across an IDLE gap (SIGTERM, then a fresh one on the same socket) ===
[padi:first] exited? yes (code=0 signal=null)
[padi:second] socket back up? true

=== STEP 5: THE FIRST padi-backed request after the restart ===
[mcp] alive? yes
resources/read  #1   → JSON-RPC error: {"code":-32603,"message":"surface stdio transport closed: unix socket
                       /tmp/…/padi.sock transport closed (SocketCloseError: 1000); the peer process exited
                       or its stream ended"}   <<< THE ONE #2082 IS ABOUT
  — following requests, to show whether it was transient or permanent —
resources/read  #2   → ok — {"contents":[{…,"text":"[]"}]}
resources/read  #3   → ok — {"contents":[{…,"text":"[]"}]}

=== RESULT ===
kolu mcp process: still alive
first padi-backed request after the restart: FAILED ❌
kolu mcp stderr: (empty)

VERDICT: FAIL — the restart destroyed a request
```

Every fact the issue reported, reproduced: the MCP process is **alive**, stderr is **empty**, only the **first** request dies, and the message it dies with talks about a peer process exiting — on the agent's own stdio channel.

### After — `3f3bb789` (this branch)

```
=== STEP 5: THE FIRST padi-backed request after the restart ===
[mcp] alive? yes
resources/read  #1   → ok — {"contents":[{…,"text":"[]"}]}   <<< THE ONE #2082 IS ABOUT
  — following requests, to show whether it was transient or permanent —
resources/read  #2   → ok — {"contents":[{…,"text":"[]"}]}
resources/read  #3   → ok — {"contents":[{…,"text":"[]"}]}

=== RESULT ===
kolu mcp process: still alive
first padi-backed request after the restart: SUCCEEDED ✅
kolu mcp stderr: (empty)

VERDICT: PASS — a restart costs no request
```

### The same thing, pinned as a test

The behaviour is now a permanent e2e leg rather than a script anyone has to remember to run — `mcp.e2e.test.ts` → *"restart across an IDLE gap: the FIRST padi-backed request after it succeeds (#2082)"*. It spawns a real padi, SIGTERMs it, respawns it on the same state root, and asserts the very first request afterwards lands, with no retry and no warm-up.

Worth saying why the **existing** restart leg could never have caught this: it calls *during* the gap, so the wasted request is spent on a call that was doomed anyway (padi is down). Only an **idle** gap exposes it.

Verified red-then-green at that level too — with the one-line `conn.onClose?.(…)` registration commented out on the box:

```
× restart across an IDLE gap: the FIRST padi-backed request after it succeeds (#2082)
  McpError: MCP error -32603: the connection to the served surface dropped while this request
  was in flight (surface stdio transport closed: unix socket …/padi.sock transport closed
  (SocketCloseError: 1000); the peer process exited or its stream ended). This MCP server is
  still running and has discarded the dead connection — retry, and the next request re-dials.
```

and green with it: **4/4 e2e legs pass** (unix socket · ssh stdio · restart mid-subscribe · restart across an idle gap).

Four surface-mcp unit pins cover the same ground without daemons, and all four go red with that single line removed.

<sub>Box: ephemeral Incus (`pu`), 32 cores / 125 GB, no production kolu present. Nothing was run against the author's machine.</sub>

srid:
## CI e2e — `3f3bb78#1`

Settled green across both platforms; no reruns, no errored nodes, nothing unposted.

| `x86_64-linux` | `kolu-ci-3` | 8 | 32 | 8 | 3m 35.6s |
| `aarch64-darwin` | `ci@petit` | 4 | 10 | 6 | 4m 47.3s |

<sub>Recipe-wall durations from the durable ledger `.ci/3f3bb78/runs/1.json`; parallelism from the last `e2e: workers=` line of each lane's `ci::e2e.log`. Whole-run wall clock 9m 29s.</sub>

PR #2120, opened 2026-08-08
Capture the first command's agent launch even under bash-preexec

### LINKED ISSUE #2119: bash-preexec/atuin swallows the first command's OSC 633;E mark — agents launched first-thing are never restore-resumable
## Symptom

After restarting kaval on `naiveintent` and running session restore, all terminals came back **but only some resumed their agents**. e.g. the `effect-blog` terminal restored as a bare bash shell even though a Claude Code session had been running in it. Of 10 saved active terminals, only 4 resumed (fix-2059-flatten-splits, olai, 2× xyne); effect-blog and hard-root came back agent-less.

Initial suspicion fell on the Effect migration (#2101, deployed the same day). **The migration is exonerated** — see timeline below.

## Root cause

**bash-preexec (pulled in by atuin in the user's `~/.bashrc`) breaks kolu's OSC 633;E command-capture hook for exactly the first command of every new bash terminal.** An agent launched as the first command — the standard flow, and *always* the case for kolu's own restore resume-writes — is never captured, so the terminal never earns a resumable `restoreTarget` and restores as a bare shell.

### The causal chain

1. Session restore resumes only terminals whose saved record has `restoreTarget: {kind: "exact"}` — `resumableTerminalIds()` in `packages/padi/src/session/resumable.ts`, consumed by `restoreSession()` in `packages/padi/src/session/sessionRestore.ts`.
2. That target is derived in `restoreTargetOf()` (`packages/padi/src/terminalWorkspace/fold.ts`) from the *observed* agent session **AND** the remembered launch command `memory.lastAgentCommand`. Agent *observation* still works (transcript watchers bind fine); the launch *command* does not.
3. `lastAgentCommand` has exactly one producer: the shell's preexec hook printing `OSC 633;E;<cmd>` (`OSC2_PREEXEC_FN` in `packages/integrations/pty/src/shell.ts`), parsed by kaval (`registerOscHandler(633, …)` in `packages/kaval/src/ptyHost.ts`) and folded via the `commandRun` channel.
4. kolu's hook rides a bash `DEBUG` trap installed at the end of kolu's wrapper rcfile. The user's `.bashrc` (sourced *before* that) loads **bash-preexec 0.6.0 + atuin 18.18.1**. bash-preexec defers its install to the **first prompt**, where it takes over the `DEBUG` trap. It does preserve kolu's trap by chaining it (`__bp_original_debug_trap` → `__kolu_preexec_dispatch` appears in `preexec_functions`), so steady-state capture recovers — but the install churn consumes bash-preexec's one-shot `__bp_preexec_interactive_mode` flag, so **the first user command finds the gate closed and no preexec fires**.
5. A terminal whose first command launches an agent TUI never shows another prompt → the launch line is lost for the terminal's whole life → `restoreTarget` stays `none` forever.

### Controlled repro (naiveintent, bash 5.3.15, bash-preexec 0.6.0, atuin 18.18.1)

Fresh kolu terminal, typed `echo t1` … `echo t4` with a prompt between each. kaval's log (`command run (OSC 633;E)` debug lines):

```
echo t2
echo t3
echo t4
```

`t1` is gone. Reproduces every time. Live interrogation of such a shell shows:

```
$ trap -p DEBUG
trap -- '__bp_preexec_invoke_exec "$_"' DEBUG          # bash-preexec owns the trap
$ echo "${preexec_functions[@]}"
__atuin_preexec starship_preexec_all __bp_original_debug_trap preexec
$ declare -f __bp_original_debug_trap
__bp_original_debug_trap () { __kolu_preexec_dispatch; }   # kolu's hook, chained — but too late for cmd #1
```

## Why only *some* terminals resumed

- **Resumed (olai, xyne ×2):** the user typed several shell commands (`cd`, `git pull`, …) before launching the agent — the launch was past the one-command window and got marked. Old kaval's log shows e.g. 14 marks for olai including `claude --dangerously-skip-permissions`.
- **Resumed (fix-2059):** zero marks its whole life, but its `exact` target was captured *before July 31* and `respawnActive()` in `sessionRestore.ts` deliberately re-seeds the saved `lastAgentCommand`/`restoreTarget` on every resume — so pre-existing targets survive restarts indefinitely, masking the capture loss.
- **Not resumed (effect-blog, hard-root):** agent launched as (one of) the first command(s). Zero `OSC 633;E` marks over their entire lifetime — **in the pre-migration kaval's log too**, which is the proof this predates #2101.
- The saved session currently on disk shows the smoking gun: terminals with Claude Code *live and detected right now* are persisted with `lastAgentCommand: null, restoreTarget: {kind:"none"}`.

## Timeline

| When | What |
|---|---|
| ~Apr 2026 | #366 replaces node-pty kernel-name polling with shell-integration hooks (macOS reports npm-shim CLIs as `node`; polling wasteful). |
| Apr 2026 | #677 makes the OSC 633;E preexec stash an agent-detection signal, noting the exact limitation that now bit: *“detection still depends on the preexec hook firing.”* |
| **Jul 31, 2026 ~10:11** | atuin + bash-preexec enter the user's home-manager config (generation 163; gen 160 has no atuin). Every kolu bash spawned since has the blind first command. |
| Aug 7 | Effect-migration deploy + kaval restart forces a full session restore — the first big restore since Jul 31 — and the silently-eroded `restoreTarget`s surface as “my agents didn't come back”. |

## Evidence (on naiveintent)

- `/run/user/1000/kaval-<hash>/kaval.log.old` — pre-restart kaval: 245 marks total; **0** for effect-blog / hard-root / fix-2059; 14 for olai incl. its claude launch.
- `/run/user/1000/kaval-<hash>/kaval.log` — post-restart: **zero** marks fleet-wide (every terminal's first command was the restore's own resume-write or an immediate agent launch) until the controlled experiment above.
- `~/.local/state/padi/config.json` — saved records as described.
- Home-manager generation bisect: gen 160 (Jul 31 09:39) no atuin; gen 163 (Jul 31 10:11) atuin present; `.bashrc` unchanged since (incl. across the Aug 7 deploy).

## Impact

- Any terminal where the agent is launched as the first command silently never becomes restore-resumable, on any box whose dotfiles load bash-preexec (atuin, ble.sh, OMB, etc. — a popular family).
- kolu's own restore resume-writes are themselves unmarked (first command), currently masked only by the `respawnActive` seed-carry.
- Recent-agents MRU and npm-shim agent detection (#677) lose the same signal in that window.

## Fix directions (to be designed)

1. **Cooperate with bash-preexec instead of fighting it**: kolu's rc tail runs *after* the user's `.bashrc`, so it can detect bash-preexec and register into `preexec_functions` / close the first-prompt gate, rather than relying on raw `DEBUG`-trap ownership.
2. **Emit from `PS0` instead of the `DEBUG` trap** (bash ≥ 4.4): `PS0` is expanded after readline reads the command and before execution, nobody else contends for it, and `history 1` carries the exact line.
3. **Server-side argv fallback**: when a live agent is observed but no `commandRun` mark ever arrived, read the foreground process argv (osfacts; `/proc/<pid>/cmdline` on Linux, `ps`/`KERN_PROCARGS2` on macOS) and synthesize the launch command — robust against *any* shell customization, at the cost of reopening the cross-platform question #366/#677 avoided.

## References

- #366 — Event-driven foreground process detection (title + process) — why kernel-name polling was abandoned (`node` on macOS, polling cost).
- #677 — Detect npm-shimmed codex and opencode — why OSC 633;E became the detection signal; documents the “preexec must fire” limitation.
- #2101 — Effect 4 migration — exonerated (rc generation `packages/kaval/src/initFiles.ts`, `shell.ts`, and the OSC 633 handler are unchanged by it; the miss pattern predates it in kaval's own logs).
- Code: `packages/integrations/pty/src/shell.ts` (`OSC2_PREEXEC_FN`, bash dispatch/arm), `packages/kaval/src/ptyHost.ts` (OSC 633 handler, `commandRunFanOut`), `packages/padi/src/terminalWorkspace/fold.ts` (`restoreTargetOf`), `packages/padi/src/session/resumable.ts`, `packages/padi/src/session/sessionRestore.ts` (`respawnActive` seed-carry).
- bash-preexec 0.6.0 (`__bp_install_after_session_init` deferred install; `__bp_preexec_interactive_mode` gating), atuin 18.18.1.

srid:
**Terminals whose agent was launched as the shell's first command silently never became restore-resumable** when the user's dotfiles load bash-preexec (atuin, ble.sh, oh-my-bash): the restore after a kaval restart brought them back as bare shells. Closes #2119 — full forensics there; the short version is that bash-preexec takes over the `DEBUG` trap at the first prompt and its dispatch gate stays closed for exactly the session's first command, which is precisely the agent-launch slot (kolu's own restore resume-writes are *always* the first command).

The fix moves the bash preexec emission off the `DEBUG` trap onto **`PS0`**, which bash expands after readline accepts a command and before it runs — the same preexec moment, but on a variable nothing in the ecosystem contends for. The command line comes from `builtin history 1`; a prompt-time baseline (`__kolu_hist_sync`) makes a `HISTCONTROL=ignorespace` skip structurally unable to replay the previous command. The old ready-flag guard machinery is deleted outright — PS0 only expands for commands readline actually accepted, so PROMPT_COMMAND hooks and readline widgets (`__fzf_history__` et al.) can no longer leak in by construction. _zsh is untouched: its native `add-zsh-hook preexec` composes with everything._

### Reproduced red-first, on a real PTY

`packages/kaval/src/shellPreexecCapture.test.ts` spawns a genuine interactive bash through the production spawn plan (`prepareShellInit` wrapper rc → `--rcfile` → node-pty) with a **vendored bash-preexec 0.6.0** in the temp-home `.bashrc`, and asserts kaval's own OSC 633 parse:

| scenario | old (DEBUG trap) | new (PS0) |
|---|---|---|
| first command, rc loads bash-preexec | ❌ mark lost | ✅ |
| later commands under bash-preexec | ✅ | ✅ |
| first command, plain rc | ✅ | ✅ |

_A pipe-based harness cannot express this — bash-preexec refuses to dispatch when stdout is not a tty — hence the PTY venue in kaval rather than kolu-pty's subprocess suite. The PS0 machinery itself (history baseline, ignorespace skip, prefix strip, user-PS0 append) is unit-tested in `shell.test.ts`, replacing the retired ready-flag suite._

> Bounds worth knowing: PS0 needs bash ≥ 4.4 (nix ships 5.x everywhere kolu runs; an older bash simply emits no marks, like an unwrapped fish shell), and a shell with history disabled emits none either. Terminals that already lost their launch line before this fix still have no saved restore target — relaunching the agent once in them repairs the record.

### Try it locally

```sh
nix run github:juspay/kolu/fuckbuggg
```


srid:
**CI e2e evidence** — run `8b071dd#2` (settled green, both platforms)

|---|---|---|---|---|---|
| x86_64-linux | kolu-ci-3 | 8 | 32 | 8 | 3m 38s |
| aarch64-darwin | ci@petit | 4 | 10 | 6 | 4m 22s |

PR #2102, opened 2026-08-04
Detect Xyne CLI sessions as a fifth agent (dock + restore)

srid:
**`xyne` ([Xyne CLI](https://github.com/xynehq/xyne-cli)) is now a first-class agent** next to Claude Code, Codex, OpenCode, and Grok — tile chrome, the dock state pip, and sleep/restore resume all light up when you run it. No config step: kolu watches `~/.xyne/agent/sessions` the same way it already watches each vendor's on-disk session tree.

### How it matches and derives state

```
foreground `xyne` (or OSC 633;E hint)
        │
        ▼
sessions/<encoded-cwd>/<ts>_<uuid>.jsonl   ← newest per cwd = session identity
  · header ({"type":"session"})  → sessionId · startedAt
  · tail newest model_change      → model
  · sidecar _summary.json         → title
  · recursive fs.watch + 1s poll floor → identity-flip rewake
```

Xyne's persisted transcript carries **no live phase events**, so `AgentInfo.state` is an honest `waiting` — the tile shows the Xyne badge with the session title and model, but no busy/needs-you motion until Xyne publishes a live signal. `taskProgress` and `contextTokens` are permanent honest nulls for the same reason.

Resume is `xyne --continue` (most recent in cwd) or `xyne --session <uuid>` (exact conversation — the id upstream's `--session` flag takes, [xynehq/xyne-cli#101](https://github.com/xynehq/xyne-cli/pull/101)), spliced by anyagent's existing `AGENT_RESUME` table — so a slept or cold-restored Xyne terminal wakes to the exact conversation.

### What's in the PR

- New `packages/integrations/xyne` (`kolu-xyne`) — `xyneAdapter` + unit tests
- `anyagent` kind / resume / STABLE_FLAGS / basename table; closed `AgentInfoSchema` arm in terminal-vocab; sensors register in padi
- Client `XyneIcon` + display tables; solid-statepip `GLYPH_XYNE` (the pip's per-brand identity mark)
- e2e `xyne.feature` with fixture writer (no real Xyne binary in CI)
- `KOLU_XYNE_DIR` test override, forwarded server→padi like the other agent dirs
- README, changelog, troubleshooting docs

> Transcript HTML export for Xyne is **out of scope** here — export refuses honestly (`NOT_FOUND`) until a kolu-side loader for the pi JSONL format lands.

### Try it locally

```sh
nix run github:juspay/kolu/xyne-cli
```

_Generated by Claude Code (model `kimi-k3`)._

PR #2124, opened 2026-08-08
One unreadable saved record must not cost the host every terminal

### LINKED ISSUE #2122: One undecodable saved terminal kills every live terminal on the host at boot
## What happened

A deploy on `pureintent`/`naiveintent` destroyed every surviving terminal on the host. The browser showed a stack of toasts:

> Terminal attach failed unexpectedly: no PTY with id 6a7f70f0-58e3-4c9c-9bd2-0cc175c15bc1
> Terminal attach failed unexpectedly: no PTY with id 94cb098e-11ad-4667-992e-f2eea7b992c4
> Terminal attach failed unexpectedly: no PTY with id fee433b8-37d9-4092-b377-3c5436812c9c

Those PTYs had been **successfully adopted one second earlier**. kolu killed them itself, and the saved session went from **18 terminals to 6** in the same second, so the restore card could not offer the lost ones back either.

## Root cause

Boot adoption decodes each saved terminal record with a **throwing** decode, and the boot's reaction to *any* reconcile failure is to kill the adopted kaval daemon and everything in it. So **one** record the running build cannot decode destroys **every** live terminal on the host.

The trigger here was an agent-kind the build did not know. The previous padi on that host was running the xyne-cli build (#2102, which adds a fifth `AgentKind`, `"xyne"`), and it persisted 18 terminals. The build deployed over it knows only four kinds, so the record carrying `restoreTarget.agent.kind: "xyne"` failed to decode:

```
{"level":50,"time":1786194511387,"msg":"surviving-session reconciliation failed — recycling the adopted daemon",
 "err":{"type":"SchemaError",
        "message":"Expected \"claude-code\" | \"codex\" | \"opencode\" | \"grok\"\n  at [\"restoreTarget\"][\"agent\"][\"kind\"]"}}
```

The error path starts at `["restoreTarget"]` (not `["terminals"][n]["restoreTarget"]`), which places it at the per-record decode — `adoptedAuthored` → `decodeAuthoredActive` (`Schema.decodeUnknownSync`) in `packages/padi/src/terminalEndpoint/local.ts:621-625`.

### The sequence, from `~/.local/state/padi/padi.log.4` on naiveintent

| time | event |
| --- | --- |
| 09:08:28 | old padi (xyne build) drains, persists **18 terminals** |
| 09:08:31.109 | new padi adopts the surviving kaval daemon |
| 09:08:31.334–386 | adopts **all 6 live PTYs**, including the three in the toasts |
| 09:08:31.387 | `SchemaError` on the `"xyne"` record → *"surviving-session reconciliation failed — recycling the adopted daemon"* |
| 09:08:31.770 | `stopping live daemon — boot policy = always recycle` — **all 6 PTYs die** |
| 09:08:32.340 | autosave writes **6 terminals** over the 18 on disk |

## Defects

1. **`adoptedAuthored` violates the repo's own `persisted-schema-stays-tolerant` policy.** It uses `Schema.decodeUnknownSync`, so a record it cannot decode throws. The sibling seed paths already do this right — `seedHandlelessTerminal` (same file) uses `decodeUnknownResult`, drops the one bad record with a warning, and keeps every other terminal. A record that fails to decode still has a **live PTY**, which should fall to the existing orphan adoption (`adoptLocalOrphan`) so the terminal survives with default chrome instead of being killed.

2. **The fail-closed recycle amplifies one bad record into total destruction** (`packages/padi/src/ptyHost/index.ts:406-422`). Its stated rationale is hidden unregistered PTYs — which does not require killing the terminals that *did* adopt cleanly. With a tolerant per-record boundary this arm becomes near-unreachable.

3. **The autosave clobbers the saved session between the reconcile failure and the park**, so the 12 unrecovered records were lost from the restore card too.

## Not an Effect-migration regression

The pre-#2101 code was `AuthoredActiveSchema.parse(record)` (zod, throwing) behind the identical fail-closed recycle, with the same four-kind enum. #2101 only changed the exception's name.

## Wider consequence

Any PR that adds an `AgentKindSchema` member is forward-incompatible with a rollback or a mixed-build host: an older build reading the newer build's session hits exactly this. With the tolerant boundary of (1) the cost drops from *every terminal on the host is killed* to *one terminal loses its resume target*.

srid:
Fixes #2122.

A deploy on a production host killed every terminal on it and left the browser showing a stack of **"Terminal attach failed unexpectedly: no PTY with id …"** toasts. The terminals had been **successfully adopted one second earlier** — kolu killed them itself, and then wrote a shrunken session over the full one so the restore card could not offer them back either.

## What actually happens

Boot adoption decodes each saved record with a **throwing** decode, and `ensureLocalEndpoint` answers *any* reconcile failure by recycling the adopted kaval daemon (`packages/padi/src/ptyHost/index.ts:406`). So **one** record the running build cannot read destroys **every** live terminal on the host, including the ones that decoded perfectly.

The trigger was going *backwards*: the previous padi on that host carried the xyne-cli build (#2102, which adds a fifth `AgentKind`) and persisted 18 terminals; the build deployed over it knows four kinds, so the record naming the unknown one threw.

```
{"level":50,"msg":"surviving-session reconciliation failed — recycling the adopted daemon",
 "err":{"type":"SchemaError",
        "message":"Expected \"claude-code\" | \"codex\" | \"opencode\" | \"grok\"\n  at [\"restoreTarget\"][\"agent\"][\"kind\"]"}}
```

The error path starts at `["restoreTarget"]` rather than `["terminals"][n][…]`, which places it at the per-record decode — `adoptedAuthored` → `decodeAuthoredActive`.

From `~/.local/state/padi/padi.log.4` on the affected host:

| time | event |
| --- | --- |
| 09:08:28 | old padi drains, persists **18 terminals** |
| 09:08:31.109 | new padi adopts the surviving kaval daemon |
| 09:08:31.334–386 | adopts **all 6 live PTYs** |
| 09:08:31.387 | `SchemaError` on the unknown-kind record |
| 09:08:31.770 | `stopping live daemon — boot policy = always recycle` — **all 6 PTYs die** |
| 09:08:32.340 | autosave writes **6 terminals** over the 18 |

**This is not an Effect-migration regression.** The pre-#2101 code was `AuthoredActiveSchema.parse(record)` (zod, throwing) behind the identical fail-closed recycle, with the same four-kind enum. #2101 only changed the exception's name.

## The fix

`adoptLocalTerminal` validates the whole record through `decodeUnknownResult` and returns `false` instead of throwing — the same tolerant boundary `seedHandlelessTerminal` in the same file already applies to the sleeping and parked seeds (the repo's `persisted-schema-stays-tolerant` policy). Because `SavedActiveTerminalSchema` is the whole record and the two projections beside it (`adoptedAuthored` / `adoptedSnapshot`) are strict subsets of those same fields, one successful gate makes both sub-decodes total by construction.

The caller routes a `false` to the **orphan** adoption the file already has: the PTY is alive either way, and "never reap a survivor" (F1) does not stop being true because the record beside it is unreadable. The terminal comes back as a live shell without its saved chrome; every other terminal comes back whole.

Two things fell out of writing it:

- **The id is validated rather than cast.** `reconcile` joins saved records to live PTYs on a raw string, so a saved id that is not a `TerminalId` reached the registry as a `record.id as TerminalId` — the one hole the orphan path had already closed.
- **The orphan adoption is now one named function** (`adoptSurvivorAsOrphan`) instead of an inline block, since two call sites need it. Its fail-closed kill for an unrepresentable *id* is unchanged, and the asymmetry is deliberate: an unusable id leaves kolu no way to hold the terminal at all, while an unusable record costs only the chrome beside a perfectly usable PTY.

## Tests

`packages/padi/src/terminalEndpoint/adoptTolerance.test.ts` reproduces the incident: three live PTYs, one whose saved record carries an agent kind this build's enum does not have.

Written first and confirmed **red on all 7** against unmodified `master`, failing with the production error verbatim:

```
AssertionError: promise rejected "SchemaError: Expected "claude-code" | "co… { …(3) }" instead of resolving
Caused by: SchemaError: Expected "claude-code" | "codex" | "opencode" | "grok"
```

They pin: the boot does not throw (so the recycle is never reached), every live PTY is adopted, **nothing is killed**, the undecodable record's terminal keeps its live pid and the daemon's cwd, the decodable records still ride through whole, the log names the record it could not read, and the converged session keeps all three.

The daemon mock serves empty streams for the per-terminal taps on purpose — an absent tap makes `adoptTerminal` reap the very PTY under test through its wiring-failure arm, which would hide the behaviour being pinned.

## Not covered here

Three things stay open, [recorded on the issue](https://github.com/juspay/kolu/issues/2122#issuecomment-5226395866) rather than folded in:

- **An undecodable SLEEPING record is still lost.** It has no PTY, so there is nothing to adopt; `seedHandlelessTerminal` already drops it tolerantly, and the converge then derives the session from the registry alone and writes it back — erasing the dormant terminal from disk rather than merely skipping it. This fix is what makes that path *reachable* (the boot used to die first), so it belongs in the record. Fixing it means having the converge carry forward records it could not decode — the same fail-safe `pairedDaemon.ts` already argues for — which changes its "the registry is the truth" invariant that the zest-incident tests pin, so it deserves its own review rather than riding along here.
- The fail-closed recycle is still all-or-nothing for a genuine `terminal.list` failure.
- The autosave can still write a partial registry over the saved session between a reconcile failure and the park — this is what turned 18 saved terminals into 6.

## Verification

- `packages/padi` unit lane: **73 files / 582 tests** green
- `just check` (typecheck across every package + biome lint): clean
- `just fmt`: applied

No surface-package changes, so the drishti pair gate and the odu-impact verdict do not apply.


srid:
## CI e2e metrics — `bd86725#1`

Full two-platform run, settled green (845.6s wall).

| `x86_64-linux` | `kolu-ci-4` | 8 | 32 | 8 | **219.7s** |
| `aarch64-darwin` | `ci@petit` | 4 | 10 | 6 | **288.9s** |

Durations are the whole `ci::e2e@<platform>` recipe wall from the durable ledger `.ci/bd86725/runs/1.json`, not Cucumber's internal timer. Worker counts are the resolved `e2e: workers=` line from each lane's log.

PR #2126, opened 2026-08-08
Effect 4.0.0-beta.106 — a rename that had to move two repos at once

srid:
Kolu has been on `effect@4.0.0-beta.103` since it was current. Three betas later, [beta.106](https://github.com/Effect-TS/effect/releases/tag/effect%404.0.0-beta.106) is out, and the interesting thing about the gap is how little of it Kolu had to react to — and which one thing it *couldn't* avoid.

## The forcing change

[beta.104](https://github.com/Effect-TS/effect/releases/tag/effect%404.0.0-beta.104) renamed the Schema error constructors to line up with their `Data` counterparts:

| before | after |
| --- | --- |
| `Schema.ErrorClass` | `Schema.Error` |
| `Schema.TaggedErrorClass` | `Schema.TaggedError` |
| `Schema.ErrorReviver` | `Schema.ErrorInstanceReviver` |

The old names are **gone**, not deprecated. Kolu spells `Schema.TaggedErrorClass` 37 times across 15 files — every wire-visible error in `@kolu/surface`, `@kolu/padi` and `kaval` is one — so the bump and the rename are one commit. None of the other renames in these betas (`Schedule.andThen`, `RateLimiter.makeSleep`, CLI `withHidden`) appear anywhere in the tree.

**The rename also lands in a repo Kolu doesn't own.** `osfacts-client` is grafted into the workspace from the `osfacts` npins pin, it is `.gitignore`d, and it declares three `Schema.TaggedErrorClass` errors of its own — so it broke the typecheck from a file no grep of this repo would ever surface. That fix is **juspay/osfacts#6** (green on both platforms), and the pin here advances to that commit so the graft and the workspace agree on one Effect version.

## What the new betas gave us, and what we took

Most of the new API has no home here — verified by sweep, not assumption: no `HttpClient` anywhere (so `withRateLimiter`'s new bounded 429 retries have nothing to attach to), no `Cron`, no `ExecutionPlan`, no Effect `Config`/`ConfigProvider`, no `HttpApi` (only `HttpRouter`), no multipart, no fast-check, no `File.seek`, and no Effect CLI.

**YAML parsing moves into Effect — but only where it belongs.** beta.104 added lightweight INI/YAML/TOML parsers under `effect/unstable/encoding`. The two governance checks now use `Yaml.parse`, which **drops the `yaml` dependency from `packages/tests` outright** (Effect was already a dependency there). Both parsers were run over the real inputs first — `pnpm-workspace.yaml` and the 87 KB `coverage-ledger.yaml` — and agree deep-equal, 4ms→2ms and 34ms→5ms.

`@kolu/solid-markdown` deliberately **keeps** the `yaml` package, because the same differential says the new parser is the wrong tool for *that* input. Front matter is user-authored, and against a corpus of all 312 front-matter blocks in this repo plus the cases `render.test.ts` pins, the new parser:

- throws on a multi-line plain scalar (`desc: one⏎  two`), which `yaml` folds — a front-matter block that renders as a table today would fall back to a raw code block;
- throws on the cyclic anchor `a: &a [*a]` that `render.test.ts` explicitly pins as "must still render a table";
- stops recognising a following ` #` comment once a plain scalar contains an apostrophe — `kolu's closed #1708 manager` keeps the comment as content where `yaml` truncates. Two real Atlas notes parse differently because of it.

(It is *better* on one axis — it parses the compact nested mappings in six `SKILL.md` files that `yaml` rejects — but "differently wrong on user content" is not a trade worth taking for a preview surface.)

**A `Stdio` adoption was tried and removed.** `kaval-tui attach` briefly asked Effect's new `Stdio` service whether stdin/stdout were terminals. The review gauntlet killed it and it is **not** in this diff — see the lens comment below. Short version: `NodeStdio.layer` is `Effect.sync(() => process.stdin.isTTY === true)`, a pass-through of the global it replaced; `main.ts` runs `cli()` at module load so no test can ever import it to substitute the service; and `attach.ts` already has `AttachTty`, a *tested* seam carrying the raw-mode and resize operations `Stdio` structurally cannot express. `packages/kaval-tui/src/main.ts` is byte-identical to master.

## The beta-assumption gate, honoured rather than satisfied

`packages/tests/governance/betaAssumptions.ts` exists so a version bump can't land quietly: every `BETA-ASSUMPTION(<tag>)` marker must carry the current pin's tag, so a bump turns all of them red until someone **re-measures the behaviour**. Re-stamping them is the easy way to defeat the mechanism, so each was actually re-measured against beta.106:

- the reactor, socket-redial and frame-limit assumptions by the named tests they cite — `@kolu/surface` is green at **61 files / 628 tests**;
- a failed `Effect.forkChild` child still does **not** fail its parent fiber;
- `Schema.optionalKey(Schema.Never)` still accepts a missing key and rejects any present value, explicit `undefined` included;
- the RPC pinger still `Effect.delay("5 seconds")` and opens its kill latch on the next unanswered tick — so `probeDaemonIdentity`'s derived 8 s still sits inside the band it was derived from.

**The gate had a blind spot, and this PR closes it.** Its regex validates the marker *line* only — so three markers here were re-stamped to beta.106 while the prose recording *what was measured* still said beta.103, and that shipped green. Code-police caught it. The prose is now re-verified (all of `frameLimit`'s quoted dist facts still hold verbatim in beta.106), and a second scan holds every `effect@<version>` written in prose under `packages/` to the pinned version. Proven in both directions: green as-is, and failing loudly when the exact original defect is re-introduced.

## One claim worth retracting

An earlier reading of beta.105's release note suggested Kolu's ~8 schema-decode sites would lose their human-readable errors, since "schema issues no longer format implicitly." Measured across both versions, that is **not** what happens: the change moved the **constructor** adapters to a generic `"Schema validation failed"` with the issue in `cause`, while `decodeUnknownSync` — which is what every one of those sites uses — still throws a `SchemaError` whose message is the path-annotated tree, byte-identical on beta.103 and beta.106. No migration was needed and none was done.

## Verification

- **CI green on both platforms** — `6a6c4c2#1`, 24 recipes per lane, 14m26s. All 48 required checks green.
- `@kolu/surface` unit lane — 61 files, 628 passed
- governance gate — `effect@4.0.0-beta.106 agreed across 68 pin sites, 11 beta-behavior assumptions stamped (3 evidence citations agreed)`
- no open code-scanning alerts on this PR

**Review gauntlet**: lens (lowy ∥ hickey) applied, simplify clean + 1 altitude fix, code-police 4 findings all actioned. The **agent-debate track has not run** — the `grok` CLI is unauthenticated on this machine (`grok login`), and the gauntlet forbids substituting another peer.

**Merge order**: juspay/osfacts#6 first, then re-point this pin at the merged commit (its `branch` field currently names the feature branch deliberately, so `npins update` can't silently walk it back to the pre-rename client).


srid:
## 🔍 Lens review — lowy ∥ hickey

**Status: applied.** Both lenses ran independently against `982f7a2...HEAD` and converged on the same finding, which changed the shape of this PR.

### The finding — `kaval-tui`'s `Stdio` adoption (reverted in `4de664ceb`)

This PR originally routed `cmdAttach`'s TTY gate through Effect's new `Stdio` service instead of `process.*.isTTY`. Both lenses said it should not stay, for reasons that survived checking:

**lowy (volatility / decomposition)** — the substitutability the service buys is speculative, not real. `NodeStdio.layer` is `Effect.sync(() => process.stdin.isTTY === true)`: a pass-through of the very global it replaced. Nothing exercises the substitution, and the `Stdio` requirement widened `program()`'s signature so every non-attach subcommand nominally depended on it.

**hickey (structural simplicity)** — the deciding point. `attach.ts` **already** has a seam for "talk to the terminal": `AttachTty`, which `attach.test.ts` drives with a fake, and which carries the raw-mode, size and resize operations `Stdio` structurally cannot express (its interface is `args`, `stdin`/`stdout`/`stderr`, and the two `isTerminal` predicates — no raw-mode primitive, no exit-safe synchronous write). Adding a second, narrower, untested seam beside a proven one is concept multiplication, not adoption.

**Author check that settled it:** `main.ts` runs `cli({…})` and `Effect.runPromiseExit(…)` at module load, so it cannot be imported by a test at all — which is why `cmdAttach` has no test today and every `kaval-tui` test targets an extracted module. So lowy's "keep it if you land the enabling test" escape hatch would have required extracting `cmdAttach` out of `main.ts` first: a refactor with no upgrade justification.

`packages/kaval-tui/src/main.ts` is now byte-identical to master.

### Deliberately not pursued

Hickey's stronger form — move the check onto `AttachTty.isTerminal()` so the file has exactly one terminal abstraction — is a genuine improvement, but it reshapes a tested interface for no upgrade reason. It belongs in its own PR, not in a dependency bump.

### Rest of the diff

Both lenses cleared it: the forced 37-site `TaggedErrorClass` → `TaggedError` rename, the two `effect/unstable/encoding` `Yaml.parse` swaps (with `@kolu/solid-markdown` correctly keeping the `yaml` package on the differential-testing evidence), the `npins` advance to the paired osfacts commit, and the re-measured BETA-ASSUMPTION re-stamps. No structural concerns.

srid:

Ran with `--no-elegance` (the simplify track already covered this tree). **4 findings, all actioned.**

| Pass | Finding | Action |
|---|---|---|
| Rules | `comment-the-non-obvious` — the `yaml` → `effect/unstable/encoding` swap was undocumented, in the very file that guards against effect drift | fixed in `28792dae5` |
| Fact-check | Three markers re-stamped to `beta.106` while the **evidence beneath them** still cited `beta.103` | fixed + re-verified in `a062987f8` |

### The fact-check finding, and why it mattered

`frameLimit.ts`, `mirrorRemoteSurface.ts` and `vocab.ts` each carry a `BETA-ASSUMPTION` marker followed by prose recording *what was measured*. I bumped the markers and left the prose — so each marker cited a measurement against a build nobody had run it on. That is the same silent version-drift the registry exists to catch, one line lower.

Fixed by re-verifying, not re-stamping. `frameLimit`'s quoted facts all still hold verbatim in beta.106's dist: `defaultMaxBufferSize = 16 * 1024 * 1024`, `isBufferSizeExceeded` still strictly-greater (a frame *exactly* at the cap is accepted), and `RpcServer.js` still answers `MaxBufferSizeExceeded` with `CloseEvent(1009, …)`. One claim was narrowed rather than copied forward — "the only `1009` in the dist" is now "the only `1009` **close code**", because there is a second textual hit inside a numeric table in `httpApiSwagger.js`.

### The structural half

The three edits matter less than the reason they were possible. `betaAssumptions.ts`'s regex validates the marker **line** and nothing else, so stale evidence one line below ships green forever — a blind spot in the gate itself, which would have compounded on every future bump.

A second scan now holds every `effect@<version>` written in prose under `packages/` to the pinned version. Proven both directions rather than asserted:

- **green as-is** — `11 beta-behavior assumptions stamped (3 evidence citations agreed)`
- **fails on the real defect** — re-introducing the exact `beta.103` line code-police found produces:

```
Effect beta-behavior assumptions are out of date (C3):
  packages/padi/src/vocab.ts:626 cites effect@4.0.0-beta.103, but the pin is now
  4.0.0-beta.106. This is the EVIDENCE under a marker: re-measure against
  4.0.0-beta.106 and update the prose (or drop the version from it), so a
  re-stamped marker never cites a measurement nobody took.
```

srid:
## CI e2e metrics — `72c08de#1`

Two-platform run settled **green** in 14m08s on the current head (osfacts pin now tracking `main`). Both lanes carried 24 recipes each; neither platform dropped.

|---|---|---|---|---|---|
| `x86_64-linux` | `kolu-ci-4` | 8 | 32 | 8 | **3m39s** |
| `aarch64-darwin` | `ci@petit` | 4 | 10 | 6 | **4m24s** |

Durations are the recipe-wall `durationMs` from the run ledger `.ci/72c08de/runs/1.json`, not Cucumber's internal timer; parallelism is the last `e2e: workers=` line of each lane's `ci::e2e.log`.

<sub>Third and final run for this branch. Supersedes `6a6c4c2#1` (pre-master-merge) and `5e6a1b5#1` (pin on `effect-client`); both were green on heads that are no longer current.</sub>

PR #2127, opened 2026-08-08
Back up state on every boot, restore from inside kolu (#1658)

srid:
Reopens and ships [#1658](https://github.com/juspay/kolu/issues/1658), extended to the post-W2.2 world (**two** state stores now, not one) and past its original scope with an in-app restore flow. The failure this exists for is not a half-written file — `conf`'s writes are atomic — but a bug persisting a **bad-but-valid value over good state**: the recorded incident where an out-of-band kaval restart emptied the registry and autosave persisted the emptied session over the real one, with no history on disk to recover from.

**Every boot now snapshots each `config.json` into a rotated `backups/` ring** (newest 10, byte-identical restarts deduped) *before* the `Conf` construction can rewrite it, plus a daily `unref`'d re-snapshot tick so a long-running daemon's newest backup never ages past a day. The ring is **one object** — `openStateBackupRing(configPath, log)` in `kolu-shared/state-backup` (the on-disk-state utility home, keeping the server↛padi seal intact) — owning snapshot · dedupe · prune · list · read · **undoable restore**: the restore verb takes the pre-restore undo snapshot itself (and *refuses* the restore when that snapshot fails), protects the member being restored from its own undo's prune, and hands `apply` only the parsed value so nothing downstream can re-read a member path mid-restore. The boot/daily snapshot path is deliberately **fail-soft** (log loudly, never block boot — the one sanctioned deviation from the fail-fast doctrine, argued in the issue); every restore-side path is fail-fast.

### Restore, per store

| Store | Wire | Semantics |
| --- | --- | --- |
| padi (per host) | `padiSurface` 5.2: `backups.list` / `backups.restore` (additive minor) | Snapshot's session re-spawned host-side through the **same `importSession` machinery as `session.import`** — freeze windows, parked-token idempotency, per-spawn compensation all inherited. No daemon restart. A remote host's ring lives on that box, reached through the per-host map client. |
| kolu-server | root RPCs `server/backups/list` / `restore` | **In-process, no restart**: the snapshot is validated by walking the *real* migration ladder on a scratch copy seeded from the value the ring read, `preferences`/`viewerMode` push through their cells' server-internal writers (clients update reactively), and the host pool **converges** onto the snapshot's fleet via the same `pool.add`/`remove` path the strip uses — with live membership consulted so an env-seeded (`KOLU_PADI_HOST`) host never counts as a failed add, same-phase hosts dialed concurrently, and a partial convergence **answered** as `{ hostFailures }` (warning toast) rather than a throw that would contradict what applied. |

`backups.list` summarizes each snapshot ("14 terminals" / "preferences · 3 hosts", with `empty` and `unreadable` as distinct facts) — after an incident the *newest* snapshots are often the corrupt ones, and the summary is how the good one is spotted. The ring is registered in the **upgrade-window shared-artifact registry** with an executed version+1 disposition proof (`inert-to-boot-restore-refuses`).

### UI

Command palette → **Debug → Restore state from backup**: one row per ring (kolu-server + each pool host's padi; unreachable hosts disabled — never offer an action the channel can't carry out), snapshots newest-first with an inline confirm that states the exact semantics. Remote padi stamps cross the foreign-clock fence (`clock.toLocal`). A discoverability tip, troubleshooting/operations/sessions doc sections, and a changelog entry ride along.

> **Design-philosophy accounting** — *fail-fast*: every restore path throws on any anomaly before applying anything; the one fail-soft path (the snapshot itself) is the safety-net exception #1658 argues, and it still logs loudly. *Electricity*: the ring is a bounded leaf algorithm, homed in `kolu-shared` (generic disk-state utilities), not a new package and not the drishti-gated `@kolu/surface`. *Reuse the source of truth*: padi restore rides `importSession` verbatim; server restore walks the real migration ladder and the pool's own membership verbs rather than a parallel writer. No `@kolu/surface*` API was touched, so the drishti pair-PR gate does not apply; odu-impact verdict: **none** — grounded by grep of a fresh odu checkout: zero `kolu-shared` / `@kolu/padi` / `state-backup` references outside `node_modules` (odu consumes only `@kolu/surface*`).

Reviewed by the full gauntlet (lowy∥hickey lens → grok debate to consensus → /simplify → code-police), each round committed on the branch — see the PR comments for the three reports. The architecture-first-principles step was skipped by explicit user opt-out for this run. Acceptance from the issue is pinned in `stateBackup.test.ts` (created / dedupe / prune / fail-soft), plus the full-ring restore-protection regression, the parseable-but-alien list guard, the scratch-ladder decode, two-way pool convergence, traversal refusal, and the padi twins.

### Try it locally

```sh
nix run github:juspay/kolu/backup-state
```


srid:

**Status: applied.** Two lenses read `029bcbc5d..HEAD` independently — lowy for
volatility-based decomposition, hickey for structural simplicity — and one
reconcile pass merged and implemented them. Every `fix` finding landed. No
direct contradictions. Three product-judgment items are parked for you below.

**Counts.** lowy: 15 findings (11 `fix`, 4 `drop`). hickey: 18 findings
(16 `fix`, 2 `drop`). Eight of those pair up as consensus (`≡`) — the same issue
seen through both lenses, applied once.

**Gate.** `just check` green (typecheck + biome across 1850 files), `just
test-unit` green workspace-wide, and the three directly-affected suites re-run
green: kolu-shared (19), padi `stateBackups` + `surface` (35), server
`stateBackups` + `state` (30).

### Commits

| SHA | Area |
|---|---|
| `494ef317d` | kolu-shared state-backup ring API |
| `34bec561f` | kolu-server restore path + the ring's boot altitude |
| `d60cb49de` | padi restore path |
| `20722c958` | State Backups dialog |

### Findings

| origin | title | outcome | commit |
|---|---|---|---|
| lowy L1 | Restore converges the LIVE pool off the PERSISTED set's baseline | applied | `34bec561f` |
| lowy L2 | Persisted key set hand-enumerated, with no compile-time tie to the schema | applied | `34bec561f` |
| lowy L3 ≡ hickey F3 | `configPath` promises genericity the ring does not deliver / writer template and reader regex kept in sync by hand | ≡ agreed | `494ef317d` |
| lowy L4 ≡ hickey P1 | The undo guarantee is a ceremony each consumer must remember / its outcome is discarded | ≡ agreed | `494ef317d`, `34bec561f`, `d60cb49de` |
| lowy L5 ≡ hickey P2 | A partial pool convergence is reported as a total failure | ≡ agreed | `34bec561f`, `20722c958` |
| lowy L6 ≡ hickey F2 | padi's state-file path derived in more than one place | ≡ agreed | `d60cb49de` |
| lowy L7 | The unreadable-snapshot convention is reimplemented per domain | applied | `494ef317d` |
| lowy L8 ≡ hickey F4 | The server arms the daily ticker at module scope; padi deliberately does not | ≡ agreed | `34bec561f` |
| lowy L9 | `implementKoluSurface`'s return type names a feature, not an axis | applied | `34bec561f` |
| lowy L10 | `kolu-shared`'s barrel is a topic bundle; its package doc is stale | applied | `494ef317d` |
| lowy L11 | Ring dir created at the ambient umask while the store it mirrors is 0600 | applied | `494ef317d` |
| lowy L12 | The ring orders by mtime rather than by the stamp it minted | observation (fix landed via hickey F6) | `494ef317d` |
| lowy L13 | `restorable` is a promise the list cannot keep | needs-human | — |
| lowy L14 | Ring depth is measured in boots, not in time | needs-human | — |
| lowy L15 | `snapshotStateFile` enumerates the ring twice per call | observation (incidentally resolved) | `494ef317d` |
| hickey F1 | `kolu-shared` fragments one ring into eight path-threaded exports | applied | `494ef317d` |
| hickey F5 | Three signals encode one dialog state machine | applied | `20722c958` |
| hickey F6 | `savedAtMs` and the file name are two representations of one instant | applied | `494ef317d` |
| hickey C1 | The two-store difference is braided across five dialog sites | applied | `20722c958` |
| hickey C2 | The inline confirm is hand-rolled beside the repo's canonical one | applied | `20722c958` |
| hickey C3 | `formatSize` duplicates the client's byte-display source of truth | applied | `20722c958` |
| hickey C4 | One "state backup entry" declared four times | applied in part | `34bec561f`, `d60cb49de` |
| hickey C5 | The two face modules are one algorithm written twice | applied (via F1) | `494ef317d`, `34bec561f`, `d60cb49de` |
| hickey P3 | `backups.restore` declares an answer nothing reads | applied | `d60cb49de` |
| hickey P4 | `startStateBackupTicker` returns a disarm no caller uses | applied | `494ef317d` |
| hickey P5 | padi restores a strict subset of what "state backup" names | needs-human | — |
| hickey P6 | The convergence baseline is read from persistence, written through the pool | observation | `34bec561f` |
| hickey P7 | Stale module doc pointer (`../stateBackup.ts`) | applied | `d60cb49de` |

### Contradictions

**None.** The two places the lenses could have collided did not:

- **lowy L1 vs hickey P6**, on the same convergence baseline. lowy: *"`currentHostKeys` is bound to `getPersistedHosts()`, but the verbs it gates are `pool.add`/`pool.remove`, which act on **live pool membership**. … an env-seeded host is **live but not persisted**."* hickey (dispositioned `drop`): *"Today the two agree … `getPersistedHosts` deliberately excludes `LOCAL_KEY`, which matches the snapshot's guest-only semantics, so the read is defensible and `pool.hosts()` (which includes local) would be *wrong* without a filter."* Compatible: the persisted list stays the baseline, and a live-membership probe was added for the ADD gate only.
- **lowy L12 (`drop`) vs hickey F6 (`fix`)**, on ordering by mtime. lowy: *"the mtime path is correct today; raising it so the cheap fix is on record if the ring grows"*, with the caution that *"a naive string sort puts `…Z-2.json` before `…Z.json` — i.e. marks the newer file as older"*. hickey: *"make the name the single source and delete the copy — `statSync` is still needed for `sizeBytes`, so this costs nothing."* Applied as hickey wrote it, with lowy's collision caution answered by a bump tie-break in the sort.

### Needs human

Three product/scope calls with no technical discriminator. Nothing was applied for these.

1. **lowy L13 — `restorable` is a promise the list cannot keep.** *"The summary reads the **raw** snapshot; the restore decodes it **after** walking the migration ladder. A snapshot that is valid JSON with a `hosts` array but fails `PersistedStateSchema` post-migration is listed as `state · N hosts`, rendered with a **Restore…** button, and then throws on click. … Cheap-list-vs-strict-restore is a defensible trade … so I am recording rather than prescribing."* Either `summarize` decodes each row, or the UI stops promising.

2. **lowy L14 — ring depth is measured in boots, not in time.** *"With a boot-triggered ring of ten, a user who restarts kolu ten times in an afternoon … has a history that is a few hours deep, and the ten newest snapshots all post-date the corruption."* Options named: a larger ring, a two-tier ring (newest N plus oldest M, or one per day for D days), or accepting the trade.

3. **hickey P5 — padi restores a strict subset of what "state backup" names.** *"`stateBackups.ts:90-110` restores only the `session` key, while the snapshot is the whole `config.json` … a user restoring after a corruption incident gets the session back but silently keeps the corrupt `activityFeed`. Whether that is the intended scope is a product judgment, not a structural one … the *structure* is fine, the *naming* over-promises."* Either restore the other keys through their `CellStore`s, or narrow the padi confirm copy.

### One judgment inside an applied fix

hickey P3 offered a "pick one, don't leave both". The reconcile pass took the
**drop** arm — `backups.restore`'s `output` removed, `importSession` back to
`Promise<void>`, `servePadi`'s discard wrapper gone — because the alternative
(seeding the active tile via `reportRestoredActive`) would seed from a dialog
that can restore a NON-active host's ring. Say the word if you want the other
arm instead.

### Adjudication (by the invoking /be-review run)

The three needs-human items were adjudicated by the run's author agent as follows — **all three dropped, no code change**:

1. **lowy L13** (`restorable` vs the restore gate) — **drop**. Both restore paths validate **before** applying anything (server: scratch-ladder decode precedes every write; padi: `importSession` decodes before `setSavedSession`), so the mismatch costs a loud, state-preserving error toast on a click, never a broken restore. Decoding all ten snapshots through the migration ladder on every dialog open buys UI politeness for real I/O; the cheap-list/strict-restore trade stands.
2. **lowy L14** (ring depth in boots, not time) — **drop / accepted trade**. The ring size and boot cadence are #1658's own specified scope; byte-identical dedupe already means only *distinct* states consume ring slots (rapid restarts don't flush history), and the daily tick bounds staleness from the other side. A two-tier ring is a follow-up if the class recurs, not this PR.
3. **hickey P5** (padi restores a subset of "state") — **drop**. Every decision point already speaks in session terms: the padi rows and confirms say "N terminals"/"session", and the docs say "padi **session** file". The umbrella palette name covers both stores, one of which (kolu-server) restores fully; widening padi restore to `activityFeed` (MRU lists that self-heal with use) adds a second writer path for negligible recovery value.

srid:
## Grok ⇄ Claude Code debate

**Consensus in 2 rounds** (grok `grok-4.5` reviewing, Claude Code `claude-fable-5` authoring), base `029bcbc5d`.

| --- | --- | --- |
| 1 | 65cf6395e | full-ring restore must not prune the member being restored (F1 blocking + F2/F3/F4) |

Round 2: grok verified every fix against the tree (independent repro of F1 now passing) and approved.

**Findings raised:**
- **F1** *(blocking)* — the pre-restore undo snapshot pruned a full ring and could delete the very member being restored; the server then re-read that path and threw ENOENT after the good snapshot was gone. Fixed: the prune protects the restore target, and `apply` receives only the parsed value — no consumer can re-read the member path mid-restore.
- **F2** *(major)* — `listWith` guarded only the read, so a parseable-but-alien member (JSON literal `null`) collapsed the whole enumeration. Fixed: one guard around read + summarize → `unreadable` row.
- **F3** *(minor)* — `list()`'s bare `statSync` could collapse the enumeration on a member that vanished between `readdir` and `stat`. Fixed: log + skip.
- **F4** *(nit)* — the server's apply ignored the raw value and re-opened the member path (the amplifier that turned F1 into a hard failure). Fixed by construction: `decodeStateBackup(raw, label)` seeds the scratch migration `Conf` from the value the ring read.

srid:

Rules + fact-check passes over `029bcbc5d..HEAD` (elegance skipped — /simplify ran separately). **3 findings, all applied** in 03ed90d9b:

- **fact-check** — a landed undo snapshot whose *prune* then threw was reported as `failed`, which would wrongly refuse an undoable restore; each unlink is now individually guarded and degrades to the documented one-over-size grace state.
- **prefer-ts-pattern** — the dialog's 3-variant summary union now renders via `match().exhaustive()`.
- **no-sync-blocking-on-the-serving-loop** — the ring's deliberately synchronous posture now carries its justifying carve-out comment (bounded 10-file ring, conf's own sync write discipline, the no-interleaving property both restore paths lean on).

Ruled out after checking: ring eviction math, restore's validate-before-apply ordering, path traversal on the wire-crossing file name, boot-snapshot-before-Conf ordering, concurrent-restore races, client double-restore.

srid:
## Evidence

Captured on a live dev kolu (chrome-devtools drive; `just dev-auto` on random ports per the dev-server discipline — production untouched). The scenario is #1658's own recovery shape: a padi ring member holding **2 terminals** exists, the live session is then **emptied** (both terminals closed — the saved session on disk is `null`), and the backup brings them back.

| Step | Capture |
| --- | --- |
| Palette → Debug → **Restore state from backup** — one row per ring (kolu-server + padi) | stores |
| The padi ring, newest first — "**2h ago · 2 terminals · 1 KB**" with the member file name | list |
| Inline confirm states the exact semantics before anything runs | confirm |
| **After Restore**: both terminals re-spawned live on the canvas, success toast, and the refreshed list now shows the pre-restore **undo snapshot at its head** ("just now · no session") — the restore is itself undoable, observed, not just promised | restored |
| Small viewport (390×844): dialog reachable, untruncated, full-width Restore button | mobile |

*Touch axis*: the dialog introduces no `isTouch()`-gated sizing — its action buttons are the shared `InlineConfirmButton` and full-width rows (≥24px targets on the small viewport capture), so a width resize covers the changed surface.

srid:
## CI e2e metrics — `69486ea#1`

| x86_64-linux | kolu-ci-2 | 8 | 32 | 8 | 275.4s |
| aarch64-darwin | ci@petit | 4 | 10 | 6 | 296.9s |

Full two-platform settle, 48/48 nodes green (one `ci::daemon@aarch64-darwin` red — a `reServeSurface` upstream-availability race in untouched `surface-remote` code — cleared on a single node rerun).

PR #2129, opened 2026-08-09
A lost connection is not a failed boot

srid:
Come back to a kolu tab that sat behind a fullscreen game and you were met with the red **kaval didn't start** card, stacked behind the honest "Reconnecting…" overlay — offering to restart a daemon that had never stopped. **kaval had been running for twelve hours with all ten PTYs alive.** The card cleared itself on reconnect, which made it alarming rather than merely wrong.

Nothing was broken. Two individually-correct mechanisms composed into a false claim:

| Mechanism | What it correctly does | What it contributed here |
| --- | --- | --- |
| `floorOnLiveness` (`@kolu/surface-map`, #1568) | Demotes a published `connected` entry over a dead link, so a green chip can never outlive the link that proves it | It demoted to **`warming`** — so a browser that merely **lost its socket** presented as a host that is **coming up** |
| The boot deadline (#1763) | Escapes a boot overlay held past its ceiling, so a wedged leg is never an eternal spinner | That reading is leg `daemon` under the **local 30s ceiling** |

The monotonic clock keeps advancing in a throttled tab, so half a minute of dropped socket was enough to certify a healthy daemon dead.

```
tab backgrounded → ws drops → floorOnLiveness demotes local entry: connected → warming
                                          ↓
                          warming + isLocalHost ⇒ leg "daemon", ceiling "local" (30s)
                                          ↓
                    monotonic clock advances (throttled tab still ticks)
                                          ↓
                       escapeSurface ⇒ { kind: "down", down: { state: "dead" } }
                                          ↓
                                  "kaval didn't start"
```

**This PR fixes it twice: once at the app, so the incident cannot recur, and once at the type, so the whole class cannot be spelled.**

## Part 1 — a floor on the observer

A boot deadline is a claim about the **server** — *"this leg was given its ceiling and never delivered"*. That claim is only ours to make while this browser can actually reach the server. `transportLive` joins the shared liveness base, and `resolveCanvasMode` downgrades an `accrue` frame to `clear` when it is false: the deadline neither fires nor accumulates.

The **mode passes through untouched**. The transport overlay already owns the screen (dimmed, click-through, scrollback still readable) — blanking the canvas on every drop would be a worse lie than the one being removed.

Two decisions worth reviewing, both load-bearing:

- **A separate fact, not the connected arm's `channelLive`** (= transport ∧ entry-connected). On the not-yet-connected arms the entry is *by construction* not connected, so `channelLive` is false there whether our link is up or down — precisely the two cases that must be told apart. This is also why flooring only the connected arm's pending gate would **not** have fixed the reported bug, which arrives via the *warming* arm.
- **`clear`, not `retain`.** `retain` holds the class anchor, so the outage's elapsed would survive the reconnect and the brief window where the socket is back but the snapshot has not landed would flash the same card anyway. Releasing it means observation restarts with a full fresh window.

> *Accepted edge:* a transport flapping faster than the ceiling keeps re-anchoring, so a genuinely wedged daemon behind a flapping link never escapes. That is the correct outcome — a browser that cannot hold a socket has no standing to certify a daemon dead, and the transport overlay is what the user sees meanwhile.

## Part 2 — the conflation becomes unspellable

The floor above is a rule. The lens review's verdict on it was that the in-PR half was the comment and the real fix was type-level, so here it is.

The defect underneath the incident is that **`floorOnLiveness`'s demotion was lossy**. It collapsed two facts with opposite consequences into one value:

- *"the publisher says this host is coming up"* — a real, self-healing campaign, worth putting a clock on;
- *"we cannot see the publisher at all"* — we know nothing, and may claim nothing.

A consumer that merely **spins** is fine either way. A consumer that **times** the entry — a deadline, an escalation, a "failed to start" verdict — can certify a healthy thing dead. A docstring is the wrong shape of fix for that; the next such consumer has to remember it.

So the floored state gets its own arm on the client-facing union:

```ts
type UnobservableEntry = {
  kind: "unobservable";
  membershipId: MembershipId;
  published: "warming" | "connected";   // the last word we actually heard
};
type ObservedEntryStatus<F, C> = EntryStatus<F, C> | UnobservableEntry;
type EntryState<F, C>          = ObservedEntryStatus<F, C> | { kind: "not-a-member" };
```

**The published `EntryStatus` — the wire type — is untouched.** The floor is a client-side projection applied after decode, so the blast radius stops at the client: `entries` and `state()` now carry `ObservedEntryStatus`, and nothing on the server or the schema moved. That is deliberate rather than convenient. The arm is a projection of *the consumer's own transport*, meaningless to anyone else, so `entryStatusSchema` has **no arm to encode it into** — a floored value cannot be republished, which is a stronger statement than the old note (#17) that it merely happened not to throw.

Three shape decisions, each doing work:

- **An arm, not a flag.** An optional `blind?: true` on `warming` is exactly the ignorable field `.claude/rules/conventions.md` forbids — it would leave the class alive under a nicer name. An arm is a compile error at every `.exhaustive()` in kolu *and* drishti.
- **`unobservable`, not `unreachable`.** It names *our* epistemic state, not the host's condition. "Unreachable" asserts something about a host that is usually perfectly fine while it is our socket that died — repeating the very conflation — and kolu's host vocabulary already spends that word on a genuinely `failed` host.
- **No `clockOffset`, no `connection`.** "Nothing green is paintable, no frozen live word keeps narrating" is a property of the arm's *shape*, not of two fields the floor remembers to clear. `failed` still passes the floor untouched: a post-mortem does not go stale like a live claim — which is why `published` has no `failed` inhabitant.

Ergonomics are the obvious objection, so `isSettling(state)` ships with it: one call, true for both unsettled arms, so a consumer that only wants a spinner pays nothing for the split. It is a `boolean` and can never be handed to something that times the entry — those still have to narrow the union, which is the point.

### Every call site, migrated deliberately

| Site | Policy while blind | Why |
| --- | --- | --- |
| `wire.ts` `connection()` | no fine word | the arm has no field for one |
| `useCanvasMode` / `canvasModeResolver` | its own `CanvasFacts` arm; **neutral `connecting`**, not `warming` | "this host is coming up" is the publisher's news, not ours |
| `hostChipTone` | muted + pulsing, **"unknown"**, `down: false`, tooltip names the last word heard | amber means "coming up"; red + strike means "unreachable" — both are claims |
| `daemonScan` | its own cause + copy | "padi is connecting" is a claim about a dial campaign we cannot observe |
| `HostDaemonChips` | no `PadiLink` claim at all | same "make no claim" the non-member arm makes |
| `diagnosticSnapshot` | names the last published word | this block is read *precisely* to diagnose this state |
| `useDaemonStatus` | unchanged | it reads `.kind === "connected"`, so it was already right |

The resolver's arm deliberately **shares** the accruing tag rather than declaring `clear` at its own return site: the blind verdict is owned by one place (Part 1's floor) because the floor and the AFP C6 earned-verdict exemption are two halves of one rule. A blind frame that cleared its own tag would never reach the exemption, so an already-earned card would lose its **Restart** button the instant the socket dropped.

Two things fell out of the migration that were latent before:

- **A fallback the new arm made visible.** The earned-verdict exemption fell back to *this frame's* tag when no boot identity was in hand — i.e. to a leg invented by the demotion (local ⇒ `daemon` ⇒ the dead card: this PR's own false claim, reached *through* the exemption). It was unreachable only because of an invariant in `bootDeadline.ts`. An exemption for an earned verdict must be able to **name the boot that earned it**; one that cannot now falls to the floor and makes no claim.
- **A stringly seam that described nothing.** `diagnosticSnapshot`'s `detailOf` took a structural `{ kind: string; reason? }` that compiled against every arm — and a failed entry's reason lives at `failure.reason`, so **every failed host in the pasted diagnostic block read `unknown`**. It takes the real union now, matched exhaustively.

## Verification

Red-on-current-code at three levels — including one that **cannot** be written against the old collapsed shape, because the two inputs produced byte-identical values:

- **`mapHarness.test.ts`** — the two floors now differ; `isSettling` covers both; the floored value is *unencodable*.
- **`canvasModeResolver.test.ts` / `useCanvasMode.test.ts`** — every arm that could reach an escape, the outage, the reconnect proving a full fresh window, the earned card surviving a blip, and an unearned verdict that cannot name its boot making no claim. The `useCanvasMode` harness now **models the floor** rather than letting a test hand-write the demoted kind, so `mapLive: false` spells the frame production actually makes.
- **`daemonScan.test.ts` / `hostChainGlance.test.ts`** — the chip and the scan panel keep the two apart.

`connectionOptionalKey.test.ts`'s two client-side pins were **rewritten, not deleted**: they used to assert the floored value spelled its missing `connection` as an absent key so an encode would not throw. There is no key left to spell, so they assert the stronger property instead — that the encode *does* throw.

## Consumers

**drishti — [srid/drishti#134](https://github.com/srid/drishti/pull/134)**, cut off current `origin/master`, CI green, pinned to this PR's final HEAD. Exactly **one** drishti site failed to compile (`DOT_TONE`, its single `Record<EntryState["kind"], …>`); eleven others would have compiled clean and painted a blind host as an amber "connecting…" campaign labelled "not configured". That PR absorbs the arm and closes those routes. It also carries kolu's `effect@4.0.0-beta.106` bump (#2126), which the new pin brings in transitively.

> *Gate note:* `.claude/rules/surface.md`'s `paths:` frontmatter lists surface, surface-app, surface-remote, surface-daemon and surface-daemon-supervisor — **not surface-map** — yet drishti consumes `@kolu/surface-map`'s `EntryState` in ten-plus files. The gate was honoured here on the merits; the path list looks like a real gap.

**odu — `none`.** Its tree at its pinned kolu SHA (`2104ddea`) has zero source references to `@kolu/surface-map`, `EntryState`, `EntryStatus`, or `floorOnLiveness`; the package appears only in `nix/overlay.nix`, `nix/env.nix`, `default.nix` and `ci/mod.just`, which hydrate it into `node_modules` as a transitive runtime dep. Nothing to break at the bump, nothing to adopt.

## Design philosophy

- **Fail fast — no fallbacks.** No knob, no degradation path. The optional-flag shape was rejected for exactly this reason, and the migration *deleted* a fallback rather than adding one.
- **Reuse the source of truth.** `transportLive` is the existing watchdog-backed `app.health().live`. `isSettling` exists so the split does not spawn a second copy of "is this in flight" in every consumer.
- **Volatility boundary.** The wire contract did not move; only the client-facing projection widened, at the one seam that already owned the floor.

Reported from the field; no issue was filed, so this PR is the record.

### Try it locally

```sh
nix run github:juspay/kolu/older-weight
```

_Generated by [`/be`](https://github.com/srid/agency) on Claude Code (models `claude-fable-5`, then `claude-opus-5` for the type-level half)._


srid:

**Status: `needs-human`** — 13 findings from two independent lenses reading the full diff
against `336ca71a6`. Seven applied across three commits, four recorded as observations,
and **two escalations** left for the author: one direct contradiction between the lenses,
and one boundary call they point opposite ways on.

| lens | findings | fix | drop |
|---|---|---|---|
| lowy | 7 | 5 (F1, F2, F3, F4's in-PR half, F5) | 2 (F6, F7) |
| hickey | 6 | 6 (F1–F6) | 0 |

Verified after applying: `just check` ✅ · `vitest run src/kaval/` 143/143 ✅ · `just fmt` ✅.
Nothing pushed, nothing merged.

### Commits

| SHA | area |
|---|---|
| `8bf7aff21` | the observer's pair leaves the surface-facts union |
| `193248bca` | name what the floor does to the campaign backstop |
| `b4db4deba` | floorOnLiveness says what its demotion erases |

Three area commits and no separate minor commit: every minor fix rewrote the exact
statements a major fix was already moving, so isolating them would have meant editing the
same lines twice. Each minor is attributed by its own bullet in the commit body.

### Per finding

| id | lens | finding | location | outcome |
|---|---|---|---|---|
| F1 | lowy | the observability floor is clock governance filed under surface facts | `canvasModeResolver.ts:198-213`, `:545-551` | **≡ agreed** — core applied `8bf7aff21`; the further relocation → **human** |
| F1 | hickey | `transportLive` rides the surface-facts union although `resolvePrecedence` never reads it | `canvasModeResolver.ts:198-220`, `:546` | **≡ agreed** — applied `8bf7aff21` |
| F2 | lowy | the floor is driven by a different liveness value than the floor it compensates for | `useCanvasMode.ts:71` vs `:98` | **≡ agreed** — applied `8bf7aff21` |
| F2 | hickey | one fact, two accessors, in one function | `useCanvasMode.ts:71` vs `:96-98` | **≡ agreed** — applied `8bf7aff21` |
| F3 | lowy | a cross-package invariant held by prose alone | `bootDeadline.ts:140-143` | ⚠️ **contradiction — escalated, not applied** |
| F3 | hickey | the soundness argument lives in another module and contradicts the return site it is about | `bootDeadline.ts:139-143` vs `canvasModeResolver.ts:432-436` | ⚠️ **contradiction — escalated, not applied** |
| F4 | lowy | electricity probe: the receptacle is right, its interface is lossy | `surface-map/src/client.ts:122-190` | observation (type change) + comment half applied `b4db4deba` |
| F4 | hickey | the escape decision is two overlapping `if`s with a re-tested conjunct | `canvasModeResolver.ts:509-556` | applied `8bf7aff21` (behaviour-preserving variant) |
| F5 | lowy | `EntryLivenessFacts`'s docstring no longer describes its members | `canvasModeResolver.ts:194-197` | applied `8bf7aff21` |
| F5 | hickey | the floor releases the campaign cell too, and the accepted-edge argument was written for the other one | `bootDeadline.ts:171-174`, `:62-81` | applied `193248bca` (option a) |
| F6 | lowy | the suspend-residual claim leans on a bound owned by another package | `bootDeadline.ts:36-40` | observation (lens disposition: drop) |
| F6 | hickey | the #2129 narrative is told four times | `canvasModeResolver.ts:30-45`, `:510-544`, both test preambles | applied `8bf7aff21`, `193248bca` |
| F7 | lowy | the test harness must hand-synchronise two knobs for one fact | `useCanvasMode.test.ts:16-27`, `:383-389` | observation — dissolved by F2's fix |

### What changed

**Both lenses, independently, on the same two things.** `transportLive` came off
`EntryLivenessFacts` and now arrives beside `exceeded` as `resolveCanvasMode`'s second
`observation` argument — so the union can no longer spell `entry: "connected" ∧
transportLive: false`, a cell reality cannot produce and which another file's correctness
argument depended on being impossible. And `useCanvasMode` reads
`const linkLive = padiMap.live()` **once**, feeding both the observability floor and the
connect-log's `link-down` reason. The floor no longer rides `app.health().live`, which is
transport ∧ every readiness cell's `liveWhen` — a future readiness cell on `koluSurface`
would have dropped it over a perfectly live socket and silently disabled the boot deadline
entirely (the #1763 no-escape class, re-opened from an unrelated module).

**One consequence worth your eye.** The pre-existing pin
*"a DEAD link makes the missing tail a link problem"* failed the moment the harness's two
liveness knobs became one — because it had been modelling `mapLive: false ∧
transportLive: true`, precisely the divergence F2 argues production cannot produce. It was
rewritten to reach the same frame by the only path that now exists: earn the ceiling over a
live link, then drop it (the AFP C6 exemption). More honest, and it is evidence *for* F2
rather than a weakening of the pin.

**Also applied:** the two-`if` escape became one `.exhaustive()` `match` on the observer's
pair (behaviour-identical — `retain` stays unfloored pending the escalation below); the
`CAMPAIGN_CEILING_MS` doc now states that the floor releases the campaign cell too and what
that costs on a flapping link, with the changelog line widened to cover both timers; the
#2129 story is told once in the resolver's module header and cited everywhere else; and
`floorOnLiveness`'s own docstring now says what its demotion erases.

### Observations (recorded, not applied)

- **lowy F6** — the suspend-residual claim leans on the `connectSurfaces` heartbeat/probe
  cadence, a bound owned by `@kolu/surface-app`. Lowy's own disposition is `drop` ("honest
  and hedged"), so the cheap half — naming the probe timeout — was deliberately not
  applied. Yours if you want the number in there.
- **lowy F7** — the two hand-synchronised harness knobs: recorded as F2's symmetry
  evidence, dissolved by F2's fix.
- **lowy F4, the type-level half** — the wanted home for *"did this `warming` come from the
  publisher or from the floor?"* is behind the framework boundary: `floorOnLiveness`
  demoting to an arm that **says** it was demoted (`{ kind: "warming", membershipId,
  observed: false }`, or a distinct `kind: "unobservable"`). Then `transportLive`
  disappears from kolu entirely and the `entry-contracts.mdx` paragraph describes a type
  instead of instructing a reader. That is a published-shape change needing
  `ref-surface-map.mdx` plus a paired drishti PR — a separate gated artifact, so it is
  recorded rather than absorbed here.

---

## ⚠️ Escalation 1 — direct contradiction: lowy F3 ∥ hickey F3

Same code (`bootDeadline.ts:139-143`), incompatible remedies. Not applied; the paragraph
is left verbatim (only `transportLive` → "the link fact", since the field name no longer
exists on the facts union).

**lowy F3 — keep `retain` unfloored, make the prose structural:**

> **Suggestion.** Applying F2 makes the claim structural rather than contingent — the
> floor and the demotion read one value, so they flip in the same tick and the
> unreachability is by construction. Then rewrite the paragraph to say *that*:
>
> ```
>  *     `retain` needs no floor BY CONSTRUCTION, not by convention: the floor reads the SAME
>  *     `padiMap.live()` that `floorOnLiveness` is handed, so a dead link demotes the entry off
>  *     the connected arm in the same tick the floor fires — and every `retain` return site is
>  *     on the connected arm. Two readings of one fact cannot disagree; two facts could.
> ```
>
> If F2 is not taken, this needs a test instead of a comment: pin a `retain` frame
> (`connected` + `recordsAwaited > 0`) with the link fact false and assert the anchor's
> behaviour, so the assumption fails loudly rather than silently.

**hickey F3 — floor `retain` too and delete the argument:**

> And it collides with the code it describes. `canvasModeResolver.ts:434-436` is
> `return f.channelLive ? clear({ kind: "empty" }) : retain({ kind: "connecting" });`, whose own
> comment says the `retain` arm handles "a dead-channel connecting … owned by the post-grace
> TRANSPORT overlay". On the connected arm `channelLive === transportLive ∧ entryConnected` with
> `entryConnected` true, so that `retain` fires **exactly when `transportLive` is false** — the
> very frame `bootDeadline.ts` calls unreachable. Both comments cannot be right.
>
> **Suggestion (apply verbatim).** Floor `retain` too, and delete the argument. If the claim is
> true, this is a provable no-op that buys the deletion of a cross-module invariant; if the claim
> is false, this is the fix. Either way the rule collapses to one sentence — *while unobservable,
> no anchor* — with no per-variant carve-out.

**Why neither was applied.** lowy keeps `retain` unfloored and asserts the frame is
unreachable by construction — a comment rewrite, zero behaviour change. hickey floors
`retain` and deletes the assertion — a no-op if the claim is true, a fix if it is false.
One of them is wrong about reachability, and the two comments in the tree state opposite
things about the same frame. Applying either would silently pick a side of a question
neither lens settled. Applying F2 *strengthens* lowy's premise (the floor and the demotion
now read one value), but it does not answer hickey's separate point that the `retain`
return site documents itself as handling exactly the frame the other comment calls
unreachable. lowy's own fallback — pin a `retain`-over-dead-link frame with a test — is the
third option and would settle it empirically.

## ⚠️ Escalation 2 — for the author: where the floor branch itself should live

Both lenses agree the field must leave the union (applied). They differ on how far the
floor itself should travel.

**lowy F1:**

> **Suggestion (behaviour-preserving; I traced all four branches).** Delete `transportLive` from
> `EntryLivenessFacts`, delete the floor branch from `resolveCanvasMode`, and move both halves
> into `bootDeadline.ts`, where the ceiling table, the anchors and the monotonic-clock rationale
> already live […] `canvasModeResolver.ts` reverts to its pre-diff shape

**hickey F1:**

> **Suggestion (apply verbatim).** Take it out of the union and put it beside the other
> observer-level input:
>
> ```ts
> export function resolveCanvasMode(
>   facts: CanvasFacts,
>   observation: { transportLive: boolean; exceeded: boolean },
> ): Precedence {
> ```

Hickey's shape was applied — it is the smaller move and the one compatible with the stated
rationale that the floor lives at the observer level in `resolveCanvasMode`. The residual
delta (relocating the branch into `recordBootFrame`) was not, because lowy's extra distance
rests partly on the "`clear` carries two provenances" argument the author's rationale
already answers, and because hickey's F3 and F4 both restructure the floor *inside*
`resolveCanvasMode`. Boundary taste on top of an agreed diagnosis — so it is yours.

## Adjudication by the author

Both escalations are settled; neither is outstanding.

**Escalation 1 — lowy F3 ∥ hickey F3 (the `retain` carve-out).** Settled by applying
hickey's remedy with lowy's reasoning, in `2ad556a61`. Both lenses were right about
different things: hickey that the `empty` site's `retain` fires exactly when the link
fact is false **in the type**, lowy that the frame is unreachable **in production**
(`floorOnLiveness` demotes the entry off the connected arm in the same tick, and after
F2 both halves read one `padiMap.live()`). So the floor now covers every accrual
variant — a provable no-op today that deletes a cross-package invariant previously held
by prose in a different module from the code it described. Pinned by test, so it fails
loudly if it ever stops being a no-op.

**Escalation 2 — should the floor branch move into `recordBootFrame`?** Settled as
**keep the current shape**. The floor returns a *mode* as well as a tag, so relocating
it wholly into the anchor writer would split the escape decision from the anchor write
across a call boundary. Recorded as a follow-up, not scope for this PR.

srid:
## Grok ⇄ Claude debate

Consensus after **2 rounds**, reviewed from `336ca71a6`.

| --- | --- | --- |
| 1 | 880a0da5e | A ceiling nobody watched is not a verdict — floor the clock, not just the frame |
| 2 | — | Peer verified all three findings resolved; no regressions; approved |

- **F1** — AFP C6 is ordered as `{accrual:'accrue', exceeded:true}` before the offline floor (`transportLive:false` → clear)
- **F2** — The module’s “ONE TELLING” of the observability floor says it “downgrades an `accrue` frame to `clear`” when `transportLive` is false
- **F3** — The pin “keeps painting the WORKSPACE behind the overlay” uses `connected` facts with `transportLive:false`

srid:

Run with `--no-elegance` (a `/simplify` pass had just landed as `e1be050fb`), scoped to `336ca71a6a999dd9690a4081d0f0f7b2b1b68d46..HEAD`.

| Pass | Findings | Outcome |
|---|---|---|
| Rules (7 built-in + 27 project) | 0 | Clean |
| Fact-check (correctness) | 2 | Both real, both fixed |

### Pass 1 — rule checklist: clean

All 7 built-in and all 27 `.agency/code-police.md` project rules returned no violation, each backed by a specific grep or enumeration. Worth naming a few that could plausibly have tripped:

- **`no-unbounded-growth`** — the new `watchedVerdict` map inherits the same prune/release lifecycle as the two existing anchor maps (it joined `EPISODE_CELLS`), so a departed host leaves nothing behind.
- **`dry-rule-of-three`** — the three per-host maps' release logic was *unified* into `EPISODE_CELLS` + `releaseEpisode()` by the simplify pass; that is the compliant direction, not a violation.
- **`no-thin-wrapper-functions`** — the test readers (`watchedExceeded` / `blindExceeded` / `modeOffline` / `tagOffline`) all clear the ≥3-caller carve-out.
- **`prefer-ts-pattern`** — the new observer dispatch is `match(...).exhaustive()`.
- **`no-overloaded-null`** — `watchedVerdict.get() ?? false` is one honest "no known exceedance" value, matching the rule's own carve-out.

One limitation recorded rather than hidden: **`no-vacuous-assertion`** was checked by tracing match-arm order and mock wiring, not by literal break-and-restore mutation (the pass is read-only).

### Pass 2 — fact-check: two findings, both fixed

Both were traced through the real source before being accepted, and the first was **reproduced red at the caller** (real anchor map, real resolver, moving clock) before any fix was written.

---

#### Finding 1 — an earned card could silently become a *different, unearned* one · **FIXED** (`b28ff47c1`)

The AFP C6 exemption held the **verdict** across a blip but recomputed the **surface** from live facts every frame — and the outage is the *same event* that moves those facts. `floorOnLiveness` demotes the entry off the `connected` arm on the tick the link dies, and the warming arm derives its leg from host-locality alone. So a card earned on the **`session`** leg became the local **`daemon`** escape: `down` / dead — the exact "kaval didn't start" claim this PR exists to remove, now reached *through* the exemption. It also contradicted the shipped copy in `remote-hosts.mdx` ("a card you were already shown stays put through a blip").

No existing test covered it: every "survives a blip" pin held `entryKind` constant across the outage, so none exercised the `connected` → demoted-`warming` transition where the leg identity actually changes.

**Fix.** The verdict is sampled-and-held, so the boot it is *about* is held with it — a new `BootIdentity` (`leg` + `ceiling`) cell beside `watchedVerdict`, released with the episode like every other cell, merged back over the frame's tag while blind. **Identity only**: the narration (`phase` / `log` / `logAbsence`) stays *this* frame's, because "kolu cannot see your machine" is worth saying while it is true — which is also what keeps the existing `logAbsence: "link-down"` pin honest.

#### Finding 2 — the demotion could arm the campaign anchor on a blind frame · **FIXED** (same commit)

Same root cause. For a remote host the demoted arm computed `leg: "provisioning"` where the earned leg was `session`, so `recordBootFrame` saw a first-ever `provisioning` leg and armed `campaignAnchors` on a frame with `transportLive === false` — contradicting this PR's own stated invariant ("no anchor ARMS or ADVANCES across frames observed while blind").

**Fix.** Recording the *merged* tag closes it without a new parameter: the same ceiling means no re-anchor, and the earned leg (not a demotion-invented `provisioning`) means no campaign cell armed on a frame nobody watched.

#### Follow-on hardening · `388fce813`

`earnedBoot` was initially optional — the silent-degradation path this repo forbids, since a caller that simply *forgot* it would compile and quietly restore the defect above. It is now **required**, so `undefined` has to be said out loud. Same shape as `floorOnLiveness`'s required `live`: the writer cannot forget to floor.

---

Everything else checked out: `bootDeadlineExceeded`'s sample-and-hold, `EPISODE_CELLS` / `releaseEpisode`, `pruneBootAnchors`'s mid-iteration delete, and every `.exhaustive()` dispatch. The two documented edges (a link flapping faster than the ceiling never escaping; `retain`-over-a-dead-link being production-unreachable but covered anyway) were correctly left alone.

**156/156 kaval pins green**, `just check` and `just fmt` clean.

srid:
### e2e timing — run `48b2c6a#1` (green, both platforms)

| `x86_64-linux` | `kolu-ci-2` | 8 | 32 | 8 | **2m 25s** (144626 ms) |
| `aarch64-darwin` | `ci@petit` | 4 | 10 | 6 | **4m 47s** (287177 ms) |

48 nodes ok · 0 failed · 0 errored, across both platforms.

*The linux e2e node needed one rerun: its first attempt died at 4m27s with a log that simply stops mid-scenario — no failure summary, no assertion, no step error — while `ci::home-manager@x86_64-linux` finished on the same box at the same second. That is a killed process under lane contention, not a test verdict; the rerun on the same box passed in less time than the attempt that died.*

srid:
## Evidence

A browser that has lost its socket to kolu-server cannot see the machine behind it. Before this PR it said so honestly for thirty seconds — and then, on the strength of a deadline it had no standing to run, decided the terminal daemon had failed to start and painted **kaval didn't start** behind the reconnect notice. The daemon had never stopped; the terminals were still running; the card offered to restart them.

Recorded by holding the browser offline for **45 seconds** — past the 30 s local boot ceiling — with a terminal already running, then letting the link come back.

### Before → after, same moment (~34 s into the outage)

| `336ca71` (pre-fix) | `388fce8` (this PR) |
|---|---|
|  |  |
| The red **kaval didn't start** card, with its **Restart kaval** button, stacked *behind* "Disconnected from server · Reconnecting…" — two surfaces contradicting each other over a machine that was fine. | Only the reconnect notice. The canvas holds the neutral connecting surface and makes no claim about what is running while it cannot see it. |

### The whole outage, after the fix



▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=pr2129-after.mp4

`echo terminals-are-running` → browser offline → socket severed → 45 s of nothing but **Reconnecting…** → back online → the link reopens, and the pane comes back **with its pre-outage scrollback intact**, then runs `echo the-daemon-never-stopped`. That last part is the point of the whole PR in one frame: the daemon the old card wanted to restart had been alive the entire time.



### The bug, reproduced on the pre-fix commit



▶ HD: https://juspay.github.io/video-evidence/evidence.html?repo=juspay/kolu&v=pr2129-before.mp4

Same scenario, same box, `336ca71a6` checked out: the card arrives on schedule ~30 s in.

### How it was captured

On an ephemeral `pu` box (`kolu-pr-2101-evidence`, 32-core NixOS), through the maintained Cucumber + Playwright step library under `KOLU_EVIDENCE=1`, one commit per clip:

```sh
KOLU_EVIDENCE=1 just test-quick features/link-outage.feature --name "^After"   # on 388fce8 — passes
KOLU_EVIDENCE=1 just test-quick features/link-outage.feature --name "^Before"  # on 336ca71 — the card appears
```

The scenario is built from existing steps (`the terminal is ready`, `I run …`, `the screen state should contain …`, `the WebSocket connection drops`, `the connection status should eventually be …`) plus four small ones for the part no existing step does: **holding** the link down. `reconnect_steps.ts` severs and lets the link re-dial at once (#410's round trip); #2129 needs the opposite, so the browser goes offline *first* (Chromium network emulation leaves a live WebSocket open, which is why `forceReconnect()` is still needed to sever) and every re-dial then fails for as long as the outage is held. The assertion polls once a second for the whole window rather than checking at the end — the pre-fix card appears at ~30 s and clears on reconnect, so a single end-of-window look would have missed it.

The scenario lives on the capture box only; it is not committed here, so the green CI on `388fce8` is the CI that shipped.

### One honest correction to the framing

The workspace does **not** keep painting behind the overlay, before or after this PR. The liveness floor demotes the local entry off `connected` the instant the link dies, so the canvas drops to the neutral connecting surface either way — and that is right ("we cannot see your machine"). What this PR changes is what happens *next*: hold that neutral surface, or escalate to a false verdict. The screenshots above are that difference and nothing else.

### Mobile / coarse pointer

Considered, not separately captured, and here is the reasoning rather than a silent skip: this change **removes** a surface on one code path and adds none. The card it stops rendering carried the only touch target in the frame (**Restart kaval**); the reconnect overlay that remains has no interactive element at all in its disconnected branch, so there is no hit target to measure against the 24 px floor and no layout that can clip at 390 px. Nothing here is gated on `isMobile` or `(pointer: coarse)`, and the diff touches only `packages/client/src/kaval/{bootDeadline,canvasModeResolver,useCanvasMode}.ts` — pure decision logic with no styling.


---

## Addendum — re-examined for the type-level half (`48b2c6a`)

The captures above were taken at `388fce8`, before the `unobservable` arm landed. **Every claim they make still holds**: the same scenario on `48b2c6a` produces the same page — only the reconnect notice, no **kaval didn't start** card, no offer to restart a daemon that never stopped. The videos were not re-shot, because the thing they are about did not move.

One thing *did* move on screen, and it is worth its own frame. The **host strip** used to borrow the word for "coming up" while it had nothing to say:

| `388fce8` — the floor demotes to `warming` | `48b2c6a` — the floor mints `unobservable` |
|---|---|
|  |  |
| **Amber**, pulsing, tooltip "connecting…" — a confident narration of a campaign nobody was watching, over a host that was fine the whole time. | **Muted grey**, still pulsing, tooltip *"status unknown — kolu is unreachable from this tab (last seen: connected)"*. No claim about the host; the last thing we actually heard, named. |

Same instant in both — 40 s into a genuine context-level browser outage, past the 30 s local ceiling. Amber is this vocabulary's word for *coming up*, and red + strike is its word for *unreachable*; a host we merely cannot see has earned neither. It is not painted as **failed** either: `down` stays false and the label keeps its normal weight.

When the link returns, the chip goes straight back to green — the entry was never anything but `connected`, and the arm was only ever a statement about us:



The canvas surface behind the reconnect overlay also changes identity — the neutral *warming* surface becomes the neutral *connecting* one, for the same reason the chip stopped being amber — but both are dimmed under the overlay the screenshots above already show, so there is nothing new to photograph there.

### How the addendum was captured

Same `pu` box (`kolu-pr-2101-evidence`), same harness, one commit per column, through a Playwright **context-level** offline hold (`context.setOffline(true)`) rather than a socket sever — a `forceReconnect()` re-dials in ~500 ms and never reaches the ceiling:

```sh
just test-quick features/blind-host.feature   # on 48b2c6a, then on 388fce8
```

srid:
### Consumer gate — [srid/drishti#134](https://github.com/srid/drishti/pull/134)

Paired PR is open, cut off current `origin/master`, and **pinned to this PR's post-merge HEAD `48b2c6a77`**. It absorbs the new `unobservable` arm at drishti's `entryStatusTone` seam and closes the eleven non-exhaustive sites that would otherwise have painted a blind host as an amber "connecting…" campaign.

> **Note on the gate's own path list.** `.claude/rules/surface.md`'s `paths:` frontmatter covers surface, surface-app, surface-remote, surface-daemon and surface-daemon-supervisor — **not surface-map**. Drishti consumes `@kolu/surface-map`'s `EntryState` in ten-plus files (`entryStatusTone.ts`, `HostDot.tsx`, `App.tsx`, `TabStrip.tsx`, `daemonHostGlance.tsx`), so the gate plainly *should* apply here; it was honoured on the merits. The path list looks like a real gap worth closing separately.

### Odu impact — `none`

Grepped odu's tree at its **pinned** kolu SHA (`2104ddea`):

```
$ grep -rnE "EntryState|EntryStatus|floorOnLiveness|isSettling|not-a-member|connectSurfaceMap|defineSurfaceMap|serveSurfaceMap" --include=*.ts --include=*.tsx .
(no matches)

$ grep -rn "@kolu/surface-map" .
nix/env.nix:10:  ODU_KOLU_SURFACE_MAP = pkgs.kolu-surface-map;
nix/overlay.nix:17:  kolu-surface-map = mkKoluPackage "surface-map";
default.nix:97:            ${pkgs.kolu-surface-map} @kolu/surface-map \
ci/mod.just:21:      "$ODU_KOLU_SURFACE_MAP" @kolu/surface-map \
```

Odu hydrates the package into `node_modules` as a transitive runtime dependency and never imports a single one of the changed exports. Nothing to break at the next pin bump, nothing to adopt — so no line item on [odu#43](https://github.com/juspay/odu/issues/43).

srid:
### Drishti gate — **green on both platforms**, pinned to this PR's final HEAD

[srid/drishti#134](https://github.com/srid/drishti/pull/134) @ `1de6910`, pinned to `48b2c6a77`, `MERGEABLE`:

```
24 ok · 0 failed · 0 errored · 0 skipped · 0 cancelled — OK
  ✔ ci::agent-boots@aarch64-darwin       ✔ ci::agent-boots@x86_64-linux
  ✔ ci::typecheck@both  ✔ ci::nix@both  ✔ ci::home-manager@both
  ✔ ci::test-daemon@x86_64-linux (skips on darwin by design)
```

The arm itself cost drishti one compile error (`DOT_TONE`) and eleven silent absorptions it closed on purpose. **The gate earned its keep on something else entirely**, and it is worth knowing about from this side: drishti's agent has its workspace manifest in *two* places — a committed `agent.package.json` the nix build copies, and a heredoc in `regenerate-agent-deps.sh` the lockfile is resolved against. Bumping one moved the lock while the build kept using the other, so a green typecheck and a green unit suite sat on top of an agent that could not boot. `ci::agent-boots` is the only thing in either repo that could have said so. Fixed there by deleting the duplicate.

PR #2133, opened 2026-08-09
A lifeless wire is unrepresentable, and the store merge stops guessing keys

srid:
A surface app's stale-tab handshake had two halves, and both were the app's to
remember. `connectSurface` handed back an `echo` you were expected to feed (by
threading `createServerLifecycle`'s `onProcessId` into `echo.remember`) and a
`status` accessor you were expected to render. Keep only `.client` — which
compiles — and you ship a tab that reconnects with no `?pid` at all, so
`rejectStaleProcess(null, live)` never rejects, the server's gate is dead code,
the wire's `retired` state is **unreachable**, and a page bound to a dead or
replaced server looks healthy indefinitely. That is [juspay/olai#61][olai61],
verified there empirically rather than by reading: with the echo removed, the new
e2e scenario times out on `Then the server rejected the stale tab`; with it wired,
the re-dial carries `?pid=<id>`, the server closes 4001, and the wire retires.

This makes both halves unrepresentable-by-omission — one by deleting the app's job
entirely, the other by moving it from an ignorable return value to a required
option.

## The design

**The framework owns the handshake, end to end.** `@kolu/surface`'s reserved
`system/identity` member — which every surface already answers — now carries a
`processId` alongside `startedAt`, minted once per process by `surfaceProcessId()`.
`createSurfaceSocket` probes it on every wire **open**, over the link's own
dispatch, and calls `echo.remember` itself. The URL thunk appends the result on the
next dial. No app code observes an identity; nothing returns an echo for an app to
feed; there is no step left to omit.

That is direction 2 from the brief, and it is the primary fix because it does not
merely make the mistake loud — it deletes the thing that could be mistaken. It also
closes the gap olai hit head-on: `system/identity` reported a start *time*, not the
id `rejectStaleProcess` compares, so olai had to grow an `identity.info` member of
its own. It reports both axes now, and an app needs no member of its own for either.

**One process id, one door.** Two ids in one server was the live hazard the moment
the client started echoing the framework's. So: `serveIdentity` stamps
`surfaceProcessId()` (no parameter to pass a different one through);
`gateStaleSocket` lost its `liveProcessId` argument and reads the same function;
and `rejectStaleProcess` — whose second argument *was* the way to point a gate at
an id the wire never reports — is deleted, its cases moved to `gateStaleSocket`'s
suite. surface-app's duplicate `identity.info` member, `serverIdentity()`,
`surfaceAppProbe()` and the `processId` injection on `surfaceAppServer()` are gone
with it. kolu's own `serverProcessId` (which stamps every log line) is now
`surfaceProcessId()`, so the log and the wire name one process.

**The terminal state is a required option, not an ignorable return value.**
`connectSurface` / `connectSurfaces` / `createSurfaceSocket` take
`retired: () => void`, with **no default**. This is direction 1; direction 3 (a
defaulted `reloadForUpdate`) is deliberately declined, because a silent default is
the disease this PR treats — it would restore exactly the property that let olai
compile. Requiring the handler does not make an app *render* anything (nothing at
the type level can), but it does make the terminal state impossible to be *unaware*
of: a wire that compiles has been asked what happens when it dies, and answered.

I chose a single required callback over the brief's `"reload" | (info) => void`
union: `reloadForUpdate` is already exported and matches `() => void`, so
`retired: reloadForUpdate` is the one-liner without a second spelling of the same
thing — and a magic string beside a handler is the kind of knob this repo's design
philosophy asks us not to grow.

**`SurfaceConnectionStatus`'s terminal arm is now `retired`, not `down`.** That
rename is the whole of olai's stated reason for reaching past the simple seam: its
`projectStatus` mapped the terminal wire to `down`, so "will never reconnect" read
as the transient state `reconnecting` already covers, and an indicator built on it
would say "reconnecting…" about a page that never will. It also collided with
`@kolu/surface-app`'s `ConnectionStatus.down`, which means the *opposite*. The wire
has one terminal state; it is now spelled the way the wire spells it. A
four-state indicator can be built on `conn.status()` alone.

## What a downstream app must now write

The whole of it — a compiling consumer with a live stale-tab handshake and a
visible terminal state:

```ts
const conn = await connectSurface({
  surface,
  url,
  retired: reloadForUpdate, // or: () => setRetired(true), and take the screen
});

export const client = conn.client;
export const status = conn.status; // connecting | live | reconnecting | retired
```

That is it. No echo, no `onProcessId`, no `identity.info` member to declare, no
`createServerLifecycle` unless you want the richer event (build-skew UI, an uptime
readout). The server side is `gateStaleSocket(ws, url)` — no id to pass.

## Migration notes for olai#61

Your PR gets smaller, and the two things it worked around are gone:

- **Delete the `identity.info` member** from `@olai/surface`, and `PROCESS_ID` from
  `server/src/identity.ts`. `system/identity` reports `processId` now, and
  `gateStaleSocket` (or a hand-rolled gate comparing `surfaceProcessId()`) reads the
  same value the client echoes. `listener.ts` needs no id threaded to it.
- **Delete the echo wiring.** `connection.echo.remember` fed through
  `createServerLifecycle`'s `onProcessId` is exactly the step this PR removes;
  `connectSurface` no longer returns an `echo`. Your e2e scenario
  (`Then the server rejected the stale tab`) should stay green with *less* code —
  and it now fails for a real regression rather than for a forgotten wire.
- **Add `retired:`** to your `connectSurface` call. Given `Restarted.tsx`, that is
  `retired: () => setRestarted(true)` — the same button, reached without a lifecycle.
- **You may not need `createServerLifecycle` at all.** Your `connectionOf` narrows
  four states out of the lifecycle event; `conn.status()` is those four states
  directly (`connecting` / `live` / `reconnecting` / `retired`), with the terminal
  one no longer disguised as `down`. The lifecycle's other restart arm
  (`restarted`, `transport: "open"` — a reconnect that landed on a *different*
  process) is unreachable once a stale-tab gate is installed: such a tab is closed
  at the handshake and arrives as `retired`. Keep the lifecycle if you want
  `serverProcessId()` for diagnostics or build-skew UI; drop it if the indicator is
  all you wanted. If you keep it, its `probe` is now
  `() => probeSurfaceIdentity(olai.rpc)` — no app-declared member.

## Consumers in this repo

kolu's `connectSurfaces` call takes `retired: recordWireRetired`, a leaf recorder
that stamps the wall clock into the diagnostic snapshot (`Debug → Diagnostic info`
gains a `server retired this tab:` line). kolu's user-facing recovery already rides
the same wire's terminal status through `rpc.ts`'s lifecycle, so this adds the
record rather than a second reload path. `rpc.ts`'s probe becomes
`probeSurfaceIdentity(surfaceApp.rpc)` and loses `onProcessId`.

The `@kolu/surface-app` **example** was itself an instance of the bug: its client
hand-built `websocketLink` → `createLiveSignal` → `surfaceClients`, so it dialled a
wire with no `pid` echo while its own server installed the gate. It uses
`connectSurfaces` now — the code people copy no longer teaches the omission.

## PART TWO: the store merge stops guessing an element key

Folded in at the human's direction — a second defect in the same layer, and the
same philosophy: never assume what the app did not declare.

`/surface/solid` merged every server value with a bare `reconcile(next)`,
which inherits Solid's `key: "id"` default: a field called `id` is read as an
element's IDENTITY, and the previous row objects are RECYCLED for the records the
diff thinks matched. On a payload whose elements have no top-level `id` — every
element's key reads `undefined`, so they all "match" — the object that held one
record is mutated to hold another. A reader that goes by position never notices; a
`<For>` keyed by reference, a per-row memo, or a component holding a row across
frames sees a row whose identity and whose fields disagree.

Every merge now passes `{ key: null }`, through ONE exported helper
(`unkeyedReconcile`) so `writeValue.ts` is the only module in the package that
imports `reconcile` at all — the question is asked in one place instead of at
every store write. That covers `createSubscription`, `createReactiveSubscription`
and `useCollectionDeltas` (which folds through the first), plus `useCell`'s three
sites. Per-key granularity inside a collection's dictionary is untouched: a
collection's `keySchema` is the DICTIONARY key and says nothing about identity of
array elements inside a value. A keyed merge, if one is ever wanted, would be
DECLARED on the member — which is a spec change with a call site, unlike
inheriting a field name from a library default.

**What I could not confirm, stated plainly.** The downstream diagnosis
(olai's live-dead worktree) attributes a VALUE corruption to this call — missing
and duplicated records in the store. I could not reproduce that: on Solid 1.9 both
spellings produce identical values across 4,000 randomised shape × edit pairs
(keyless rows, `id` rows, repeated-`id` rows, nested arrays), because `applyState`
re-applies the target's values positionally after its LCS placement. What the key
demonstrably changes is object IDENTITY — measured on a 3→4 mid-insert, the
default recycles 3 of 4 row objects and the object that was record `b` ends up
holding record `mid` — and that is a real defect with a real consumer-visible
consequence, so it is what the fix and the tests are stated against. The A/B in
that diagnosis is consistent with the identity mechanism reaching the DOM through
a reference-keyed reader; I have not proved that end of it, and this PR does not
claim to.

The four identity tests fail on `reconcile(next)` and pass on `{ key: null }` —
verified by reverting the one line and re-running. The value assertions beside
them pass either way and say so in the file, so the suite cannot be mistaken for a
value-diff reproduction. Writing them turned up a trap worth naming: a recycled
object is mutated IN PLACE, so a test that captures rows and reads their ids after
the next merge reads the ids they ENDED with and asserts nothing. The helper
records each row's id eagerly for exactly that reason.

## Drishti pair

[srid/drishti#135](https://github.com/srid/drishti/pull/135) — pinned to this
branch's FINAL HEAD (`5a47d7f`, the commit kolu CI settled green on for both platforms), typecheck clean, 391 tests pass. It drops the
`createProcessIdEcho` declaration and the `onProcessId` feed, swaps
`surfaceAppProbe` for `probeSurfaceIdentity`, drops `admin.processId` from the
router's return and the gate's arguments, and adds a `retired` handler in
drishti's logging idiom.

## ODU-IMPACT VERDICT: `none`

Grepped odu's tree at its pinned kolu SHA (`2104ddea`) for every changed export:
`connectSurface*`, `createSurfaceSocket`, `gateStaleSocket`, `acceptSurfaceSocket`,
`surfaceAppServer`, `serverIdentity`, `rejectStaleProcess`, `surfaceAppProbe`,
`surfaceAppSurface*`, `probeSurfaceIdentity`, `serveIdentity`, `ServedIdentity`,
`SurfaceConnectionStatus` — **zero hits**. The same grep covers part two: odu imports no `@kolu/surface/solid` seam at all (its consumption is `client`, `define`, `errors`, `link`, `links/*`, `loopback`, `peer-server`, `project`, `server`, `unix-socket`, plus `surface-mcp` and `surface-remote`), so the store-merge change reaches nothing there. odu imports `@kolu/surface` (client,
define, errors, link, links/*, loopback, peer-server, project, server,
unix-socket), `@kolu/surface-mcp` and `@kolu/surface-remote`; it consumes no
surface-app seam and reads no `system/identity` answer. The `ServedIdentitySchema`
field is additive-required on the wire, and odu ships its client and its servers
from one pin, so the bump carries no skew window either. Nothing to append to
[odu#43](https://github.com/juspay/odu/issues/43).

[olai61]: https://github.com/juspay/olai/pull/61


srid:
### e2e timing — `5a47d7f#1` (green, both platforms)

| `aarch64-darwin` | `ci@petit` | 4 | 10 | 6 | 291.9s |
| `x86_64-linux` | `kolu-ci-2` | 8 | 32 | 8 | 233.6s |

Durations are the recipe wall from the run ledger (`.ci/5a47d7f/runs/1.json`), not Cucumber's internal timer; parallelism is the last resolved `e2e: workers=` line in each lane's log.

PR #2135, opened 2026-08-10
A blocked worker with nobody listening is now unspellable

### LINKED ISSUE #1904: kolu MCP needs a standing subscription verb — the coordinator's watcher choreography drops worker reports at every seam

srid:
Closes #1904.

A coordinator agent drives worker agents in kolu terminals. Its only "tell me when" primitives were the blocking `wait_*` calls — each watching **one** terminal for **one** condition, and only while the call is open. So every coordinator hand-rolled the missing layer: background watcher agents looping the waits and reporting back. In one session that layer dropped reports at four seams, and a worker's MERGE-READY report went into the void until a human noticed by hand.

The fix is not a better watcher. It is that **there is nothing to arm**.

## The supervision edge *is* the subscription

A terminal already records who spawned it (`parentId`, refused if it would cycle), and kaval already serializes input into a terminal — a single-writer mailbox. Put those together and a worker going blocked delivers into its supervisor's mailbox *by construction*: dispatching the worker is what creates its report path.

That is the whole point. Every previous fix added a *listener*, and a listener is a separate act — so "dispatch a worker without arming its notification" stayed spellable, and a coordinator that forgot once dropped a report. Here the ask and the listener are the same fact, so forgetting is not expressible.

Guarded: **only agent terminals are written into.** A person's shell is never typed at — they already have the Dock, the canvas and OS notifications for this.

## What the design note asked for was already half-built

The plan-of-record prescribed firing when *agent-state is `awaiting`/`waiting` **and** raw PTY output has settled*, by re-conjoining `wait_agentState` and `wait_outputSettled`. Grounding that against the code turned up something better: **that conjunction shipped long ago** as padi's effective-finish (EF2), and it is exactly what the `urgency` cell's `finishedIds` has always meant. So nothing here recomputes it.

One correction to the note, which the shipped model had right and the note had wrong: `asking` (`awaiting_user`) is delivered **ungated**. An agent that declares it is blocked is definitionally not mid-output; the quiet window only ever guarded turn-*end*, which is where subagent flap actually lives.

What is genuinely new is the **edge**. `urgency` is a level — who is in each class *now* — and a notification needs a transition: who just *entered* one. That diff graduated out of the browser's own alert engine into `@kolu/terminal-vocab/attentionTransitions`, so a human at the canvas and a supervising agent in a PTY now fire on **one** definition rather than two that agreed by luck. The client keeps everything above the decision (its fire-once latch, the watched gate, the unseen-finished dot); padi keeps its own (the parent edge, the mailbox write).

## For a supervisor with no terminal

An MCP-only agent has no mailbox for the edge to reach, so it gets standing named subscriptions: `watch_open` → `watch_next` in a loop → `watch_close`.

The property that matters is what happens while nobody is asking. The `wait_*` tools are edge-triggered on a live call, so the gap between two waits is unobservable — that is seam 3. These buffer **in padi**, which outlives both the `kolu mcp` process and kaval, so the gap is not a hole; and they are keyed by a caller-chosen **name**, so re-opening after any restart reattaches instead of starting empty — the issue's own "by key, not by terminal instance". Overflow returns a `dropped` count rather than truncating, because silent truncation reads to a supervisor exactly like a calm workspace.

On the wire this is padi's existing pulse-then-requery idiom: instant `watch.open`/`drain`/`close` verbs beside a `watchPulse` doorbell. No handler is held open server-side for the hours a supervisor may idle, and a **lost pulse costs a wait, never an event** — the buffer is the authority.

## The four seams

| | |
| --- | --- |
| 1 — a kaval restart re-keyed every terminal id, orphaning watchers | Subscriptions key on a **name**, not a terminal instance; the default scope is *all terminals*, which no respawn invalidates. And a terminal leaving is now an **event** (`kind: "gone"`) — a retired id says so instead of going quiet. Note the underlying re-key is unchanged: it is a deliberate property of session restore that several parts of the repo design around, and making terminal ids stable across a recycle is its own architectural change, not a rider on this one. |
| 2 — `wait_*` said `closed` for terminals that were alive | `closed` was already the honest answer (feed lost ≠ terminal dead); the harm was that it *read* as death. It is now named as such where agents actually read it, and the new path can't produce the confusion at all — a dropped pulse returns `closed` **with the buffered events still queued**. |
| 3 — a watcher that completed left the lane unwatched | Killed twice: the edge needs no re-arming, and the subscription buffers across the gap. |
| 4 — coordinator restarts killed all watchers silently | There are no watcher processes. The queue lives in padi and is reattached by name. |

## Design philosophy

- **Fail fast** — a drain against an unopened name raises a declared `WatchSubscriptionNotFound` naming what *is* open, rather than answering `{events: []}`; an empty id list is refused at the schema; overflow is reported, never silent. Each of these is the same rule: a supervisor must never be able to read "nothing is subscribed" as "nothing is happening".
- **Reuse the source of truth** — no new conjunction (EF2 already had it), no new transition diff (graduated the client's), no new terminal read (`lookup` is the same `readOne` the collection serves).
- **Electricity** — the shared piece that moved is the pure *vocabulary* (a frame diff, zero I/O) into the package both consumers already depend on. The delivery and the registry stayed in padi, which owns the mailbox and the edge.

## Testing

32 new unit tests across the three new modules; 618 padi + 35 kolu-mcp tests green. The load-bearing ones are the guards — never into a human shell, never into a sleeping terminal, a root terminal delivers nowhere, a killed supervisor is a quiet no-op — plus the buffer properties: events arriving while nobody drains survive, a re-open keeps the queue, overflow reports its count, a timeout consumes nothing.

## Odu impact

`none` — verified by grepping odu's tree at its pinned kolu SHA for the changed exports. This PR touches `@kolu/padi`, `@kolu/kolu-mcp` and `@kolu/terminal-vocab`; odu consumes `@kolu/surface*` only, and no `@kolu/surface*` API changed here (so the drishti pair-PR gate does not apply either).


srid:
## Grok ⇄ Claude debate

Consensus after **3 rounds** against `8f0c0cf71`. All 7 findings resolved; approved.

| --- | --- | --- |
| 1 | d0565723d | sanitize intent before it reaches a PTY (F1, F3–F6; F2 partial) |
| 2 | 7d8573aa4 | truncate the nudge on graphemes, not code units (F7) |

The sharpest one was **F1**: `intent` is free-form multi-line markdown (the editor is a `<textarea>`, the schema enforces only non-emptiness, and `firstIntentLine` exists downstream precisely because intents contain newlines), and it was being interpolated raw into a PTY write terminated by `\r`. An ordinary two-line intent therefore submitted twice at a supervisor's prompt, with the tail arriving as its own instruction. **F7** was a regression the F1 fix introduced — the new length cap sliced UTF-16 code units, so an intent whose boundary landed inside an emoji emitted a lone surrogate; intents are emoji-bearing by design, so that was likely rather than a corner. Both are now pinned red-before/green-after.

One dispute, accepted by the reviewer: **F2** proposed skipping delivery when the supervisor's agent is `awaiting_user`. The doc overclaim it identified was real and fixed (an agent terminal *is* written to while you answer it, and the user docs now say so). The behavioral half was declined — it would invent a second silent drop of exactly the kind this change removes, `awaiting_user` means "the agent asked", not "a human is mid-keystroke", and it would suppress precisely the supervisor whose next decision depends on the news. The design-of-record line stays "never inject into a human's shell", now enforced against live state at the moment bytes move (F4).

### Findings

- **F1** — freeform `intent` interpolated into the PTY write with no sanitization
- **F2** — the human-protection guard is only "no live agent ⇒ skip"
- **F3** — `awaitWatchEvents`, the blocking half `watch_next` runs, had no tests
- **F4** — the agent guard read the observe-time frame while the write used the live handle
- **F5** — `drain` could hand back events at or below the acknowledged watermark
- **F6** — the single-line assertions never exercised the field that can introduce newlines
- **F7** — `ptySafe` truncated on UTF-16 code units, emitting a lone surrogate mid-emoji

srid:

Status: **applied** (with four items adjudicated by the author). lowy raised 15, hickey 15; **no direct contradictions**, so the debate valve never opened — the two lenses overlapped on four issues and agreed each time.

| Commit | Area |
| --- | --- |
| c60f05bac | the shared attention transition |
| 6d2270f47 | the standing-subscription registry |
| 61133bbfc | the settle-event fan-out |
| f467d5dd7 | minor polish |
| 5f716b32e | an undeliverable supervision edge is never silent (author adjudication of L1) |

The four both lenses reached independently (`≡`): the doorbell re-implemented padi's own `changePulseSource` and dropped its error seam (**L4≡F11** — lifted to a shared `pulseSource.ts` now backing all three pulse streams); two modules bypassed padi's pino seam for `console.error` (**L8≡F10**); the ack protocol's `cursor` named three different quantities across the API (**L5≡F10**, renamed `ackAfter`); and the per-event fan-out destroyed the frame the fold produced, so the mailbox sink could not batch and a kaval recycle wrote N separate submits into one supervisor (**F1≡L13** — sinks now take the whole frame, which also gets the PTY write off the reactor's recompute stack).

Two findings were real defects rather than structure: **F4**, `dropped` was zeroed by an acknowledgement that only describes seqs, so drops accruing after a reported batch vanished unreported — against the module's own "overflow is reported, never silent"; and **F2**, the doorbell counter lived on the record whose disappearance it had to announce, so a close with no prior pulses left a parked consumer waiting out its full timeout. Both fixed, both pinned.

### Adjudicated (routed to the human, decided by the author)

- **L1** *(major)* — no retention on the mailbox sink, so a **dormant** supervisor (slept/parked) never learns of a settle, and is not told on wake either. Real. Retention across a wake needs a delivery contract this module does not have (how long to hold, what flushes it, what if it never wakes), so it stays out of this PR — but the **silence** did not: that skip is now a `warn` and a stated limit in `packages/padi/README.md`, distinct from the by-design quiet skip over a human's live shell (5f716b32e). The fact itself still survives in `urgency` and in every standing subscription.
- **L9** — declined: the suggested single enforcement point does not compile (`Schema.isNonEmpty()` does not narrow to a non-empty tuple in this Effect version, so it would need a cast at the call site).
- **L12** — declined: `terminal-vocab`'s fourth subpath. Applying F3 gave that module a stateful export too, which argues for keeping its own door.
- **L10(b)** / **L14(b)** — declined: a new `unacked` wire field (additive scope) and a `watchRegistry` → `settleEventQueues` rename (naming taste; the current name matches the `watch.*` verbs and `watch_*` tools users actually see). Each finding's concrete half was applied.

srid:

Two passes (rules + fact-check) over the diff against `8f0c0cf71`, after the architecture, lens and debate rounds. Elegance skipped (`--no-elegance` — `/simplify` had just run). Four findings, all fixed in d9266ca51.

**Critical, from the fact-check pass** — an out-of-range `after` silently and permanently starved a subscription, and it was reachable through the *documented* recovery path. `seq` restarts at 0 on every padi boot; a supervisor is told to keep passing back the `ackAfter` it last saw and is given no way to detect a restart. So the ordinary sequence — padi restarts, the drain raises `WatchSubscriptionNotFound`, the agent re-opens and retries with the cursor it remembers — set a watermark no future event could climb past, and `accept()` discarded every settle for the rest of the daemon's life. No log, no `dropped` count: exactly the silent truncation `watchRegistry.ts`'s own header says the feature exists to remove. (The `accept()` watermark filter added earlier in this same gauntlet is what turned it from one-shot into permanent.) An acknowledgement beyond anything the daemon has emitted is now **ignored and warned about** rather than clamped — clamping would acknowledge events the caller never saw, where ignoring re-delivers at worst, and re-delivery is already the contract.

**`no-unbounded-growth`** — the per-subscription buffer was capped from the start; the *collection* was not, and `watch_open` is agent-callable. A caller minting a fresh name per call leaked one subscription per call for the daemon's life. Capped at 64, **refusing** rather than evicting: evicting somebody else's queue would blind a supervisor that did nothing wrong.

**`dry-rule-of-three`** — `watchNext.ts` restated padi's own name/watermark bounds. Deduping them surfaced a second defect the arg-schema pin caught immediately: reusing padi's already-*checked* `NonNegativeInt` buried the field's description in an `allOf` branch no MCP host reads — the exact annotate-first trap `wait.ts` documents. The **bound** is shared now; the annotation order stays the face's own, and both blurbs are pinned.

**`prefer-ts-pattern`** — `settleSentence`'s if/ternary over the 3-variant `SettleKind` union is now `match(...).exhaustive()`, so a fourth kind is a compile error rather than something that silently takes the "finished" wording.

Both passes specifically hunted a third instance of "a test asserts something production cannot produce" (that failure mode was caught twice earlier in this PR) across every new test file and every new `catch`/fallback site, and both came back clean. No caught error collapsing to empty, no unjustified fallback.

srid:
## e2e metrics — `2524a34#3`

| x86_64-linux | kolu-ci-3 | 8 | 32 | 8 | 3m 45s |
| aarch64-darwin | ci@petit | 4 | 10 | 6 | 4m 26s |

Durations are the recipe wall time from the run ledger; workers is the last resolved `e2e: workers=` line from each lane's log.

PR #2136, opened 2026-08-10
⌘⇧H opens on the host you came from

srid:
**Pressing the same chord twice now hops you back and forth between two machines.** ⌘⇧H used to open the host list on whichever machine happened to be first in the pool, so switching between two hosts meant reading the list and picking, every time. It now opens on the host you switched away from — making **⌘⇧H, Enter** the machine-level twin of **⌘K, Enter**, which has toggled between your last two terminals since 2.0.0.

The interesting part is that ⌘K's toggle was never a rule. It was a side effect of the root index dropping the active tile from Recent, which left the previous visit sitting at row 0 — a trick that does not generalise, because the Hosts list has to keep listing *every* machine, in pool order (a list that reshuffles under the cursor is not one you can learn). So rather than teach the host rows the same trick, the behaviour moved out of row ordering and into one named policy both kinds feed.

### One rule, two trails

`defaultSelectionIndex` (`palette/rootIndex.ts`) — *the most-recently-visited row of the leading kind, minus the row you are already on*:

```
⌘K root      terminals lead → active already excluded upstream → row 0   (unchanged)
⌘⇧K browse   terminals lead → active IS in the list → skipped → prev visit
⌘⇧H hosts    hosts lead → active skipped → highest switch-trail rank
any group    commands lead, no ranks, all tie at 0 → row 0               (unchanged)
```

"Leading kind" is what the list is a switcher *for*, and confining the search to it is also what lets the two trails keep their own units — a terminal's rank is a wall clock (`max(visit, activity)`), a host's is a small MRU ordinal — so they are never compared.

| | terminal rows | host rows |
| --- | --- | --- |
| Trail | `terminal/visitRecency` (existing) | `host/hostRecency` (new) |
| Recorded at | the activation choke point in `createViewState` | **an effect observing `activeHost`** |
| Persisted | `localStorage`, per device | `sessionStorage`, per tab — like the `activeHost` pref it shadows |

Observing the pref rather than instrumenting a writer is deliberate: a palette row, the selector strip, the mobile chip, and the membership reconcile's bounce to local all land on `activeHost`, so there is no switch path left to forget.

On the palette side, six hard-coded `setSelectedIndex(0)` resets (open, drill in, drill out, deep-link, path repair, query change) become one `resetSelection()`, and the empty-root Recent exclusion now reads the same `currentSelection()` the policy does — *the two were always the same idea ("never offer to go where you already are") spelled twice.*

> The e2e harness is single-host, so the host switcher itself cannot be driven there. `palette/hostSwitch.test.ts` runs the **real** `hostRootActions` rows through the **real** policy against the **real** trail — the join is where this feature would silently die — and it is falsifying: drop the `rankAt` stamp and it goes red. Live proof on a real second host rides in the evidence comment.

### Try it locally

```sh
nix run github:juspay/kolu/stormy-sleep
```


srid:

**Status: `applied`** — 18 findings (lowy 6, hickey 12), every one settled; 13 applied across three commits, 5 recorded as observations. Nothing went to the human.

Both lenses read the diff independently and landed on the same fault line from opposite directions: the change had correctly extracted *the rule*, but left the *units it compares* and the *obligation to invoke it* hand-held.

### What they agreed on

| | finding | outcome |
|---|---|---|
| **hickey F1** ≡ **lowy F6** | `rankAt` carried two units — wall-clock ms for a terminal row, a 1…20 MRU ordinal for a host row — with "never compare across kinds" living in three prose comments. hickey called it a fix; lowy traced all three comparisons, found each already kind-guarded, and filed it as a no-op with a trigger condition. | **applied hickey's fix**, which retires lowy's trigger: the host trail now carries a monotonic wall clock (`switchedAt`), exactly as `visitRecency` does, so `rankAt` is milliseconds everywhere and the three comments are deleted. The kind confinement stays, now meaning one thing (a ⌘K list is a *terminal* switcher) instead of two. |
| **lowy F4** ≡ **hickey F4** | The policy got DRY-ed; the *invocation* did not — eight `resetSelection()` call sites carrying a prose-only precondition ("call it AFTER `path`/`query` are written"), alongside two pre-existing effects enforcing the same invariant. Three mechanisms, one rule. | **applied**: one `createEffect(on([path, query, open]))`, zero call sites. A future `setPath` cannot forget the follow-up, and `filtered()` is read post-write by construction rather than by discipline. |
| **lowy F5** ≡ **hickey F5** | The tolerant-persisted-list parse spelled twice, and a third hand-rolled corrupt-value warn, next to the module whose entire job is validate-on-read. | **applied**: `parseTolerantList` hoisted into `persistedPref` beside the `parseBool` precedent; both trails route through it, and the warn reuses `defaultInvalidWarning`. |

### Solo findings applied

- **lowy F1** — the host trail owned its own *ranking policy*, against the rule `fleetActions.ts` states one file away ("lives here, not in the trail store, so visitRecency stays trail-only"). The tell was that `mru` — the thing the store owns — had no production consumer while `rankOf` did. `hostRankScore` now sits beside `terminalRankScore`; the store is a trail.
- **lowy F2** — the trail's correctness rested on `useActionContext` *remembering to construct it*, failing silently and plausibly if it ever stopped. Now an eager module-scope `createRoot`; the boot wire is deleted.
- **hickey F2** — the rule's typed-query clause lived in the palette while the module claiming to be THE rule had ten tests for the other clause and none for it. `defaultSelectionIndex(items, current, query)` now owns the whole rule, tested.
- **hickey F7** — both new test files shared one module singleton and apologised for it in comments. `createHostRecency()` is exported apart from the app instance; each case gets a fresh trail.
- **hickey F6** (the join test re-implemented `encodeHostKey` inside the very test meant to catch key mismatches) and **hickey F10** (imperative scan with two mutables and a `-1` sentinel → `reduce`).

### Three contradictions, adjudicated

1. **Eager module root** — lowy: fix (silent-failure risk). hickey: no-op (trades a phantom call for a module-load side effect). *Settled by in-repo precedent, not argument*: `wire.ts` already builds `encActiveHost` and `groundedActiveHost` at module scope, and `activeHost` itself is a module-scope `sessionStorage` pref — so the side effect hickey was avoiding is already unavoidable on any import of `wire`. **lowy applied.**
2. **Leading-kind inference** — hickey wanted a `within: ResultKind` argument computed from `path()`; lowy wanted the composition pinned by test. At root the caller genuinely *cannot* name the leading kind (root leads with terminals **or** hosts depending on whether Recent is empty), so the argument could not be supplied honestly. **lowy's version applied**, plus hickey's doc half — the empty-Recent host fall-through is now documented in `switcher.mdx`.
3. **`rankAt` units** — resolved by fixing it (row 1 above), which is why lowy's declined finding is marked superseded rather than dropped.

### Recorded as observations

**hickey F8** (`hostRootActions`'s singleton reach mirrors `terminalSwitchActionsForHost`'s — changing one alone reduces uniformity), **F11** (the two "never offer where you are" mechanisms serve visibility and highlight respectively; merging them is a UX change), **F12** (`terminalId` widened from the branded `TerminalId` — pre-existing, and tightening touches every fixture).

### One deliberate deviation from a lens patch

lowy's F4 patch guards the merged effect with a bare `mode().kind === "value"` return. Taken literally that regresses drilling *into* a value group — the old `drillInto` reset to row 0 unconditionally, while a bare mode guard leaves the parent level's index in place, so the worktree-agent picker would open on the wrong label. The shipped guard is scoped to *"only the query moved"*, which preserves both behaviours: typing a value never moves the highlight off the label you picked, and a scope change still re-lands it.

### Gates

`just fmt` clean · `just check` exit 0 · client unit suite **1363 passed** (154 files) · e2e **88 scenarios** green across `terminal-switcher`, `command-palette`, `deep-links`, `recent-agents`, `recent-repos`, `worktree-agent`, `theme`, `ports`.

srid:
## Grok ⇄ Claude Code debate

Consensus after **2 rounds**, base `0430c2ad6206c728191b4d430c4a930ac44672af`. Grok reviewed; Claude Code (the author) fixed or disputed each finding. Approved with no open findings at any severity.

| --- | --- | --- |
| 1 | fab86c656 | F1 fixed — the docs conflated a row's *rank* with the list's *paint order*; F2 disputed by experiment |

- **F1** — the "Where the highlight lands" paragraph claimed "most recent" is the same ranking each list already sorts by, which is true only of empty-root Recent. Hosts keeps **pool order** and Terminals browse keeps its **host-by-host order** — the `rankAt` split exists precisely so the trail moves the *highlight* without reshuffling the *rows*, and the old wording taught a reader to read a mid-list highlight as a bug. Rewritten to separate the two.
- **F2** — that the highlight owner being a `createEffect` (rather than `createRenderEffect`) leaves a frame where row 0 is painted before a non-zero default lands. **Disputed and accepted**: Solid's `createEffect` is post-*render*, not post-*paint* — its callbacks flush inside the same synchronous `runUpdates` cycle as the triggering write. Settled by modelling the real ⌘⇧H chain, including the second highlight pass caused by the lifecycle effect writing `path`; the DOM stand-in never holds a stale value. The suggested fix would also invert the phase ordering the highlight owner deliberately depends on (it is registered *after* the lifecycle effect so it observes the final `applyInitialPath` path). Grok's close: *"I conflated Solid `createEffect` with React `useEffect` … I have no counter-observation that would see a wrong row on screen."*

srid:

Scope: `git diff 0430c2ad6206c728191b4d430c4a930ac44672af...HEAD` (terminal-switcher / host-recency feature). Final gate — architecture-first-principles, lens-debate (lowy+hickey), an agent debate, and `/simplify` already ran and applied fixes on this tree. Elegance pass skipped (`--no-elegance`).

| Pass       | Issues found | Details |
| ---------- | ------------- | ------- |
| Rules      | 2             | 1 fixed, 1 declined with recorded reasoning |
| Fact-check | 1             | Confirmed, **fixed** |
| Elegance   | –             | Skipped (--no-elegance) |

**Violations or issues found.**

### Pass 1 — Rules

**`prefer-ts-pattern`FIXED** (commit `e7f3054c3`)
`packages/client/src/palette/rootIndex.ts`, `isCurrentRow` — a 3-branch `switch` dispatching different logic per `ResultKind` variant (`"terminal" | "host" | "command"`) is now `match(itemKind(item)).with(...).exhaustive()`, matching the project's established idiom for this exact shape of dispatch.

**`caught-error-must-not-collapse-to-empty`DECLINED**
`packages/client/src/host/hostRecency.ts` (`onInvalid: defaultInvalidWarning(STORAGE_KEY, [])`) and the equivalent shared path now used by `packages/client/src/terminal/visitRecency.ts`. A corrupt persisted switch trail collapses to `[]` via `console.warn` only.
Disposition: not silent — routes through `persistedPref.defaultInvalidWarning`, which names the offending key, the raw value, and the fallback it degraded to. The module's own doc deliberately calibrates this class as a *benign, recoverable* preference reset (same treatment as `activeHost`/`reattachAnnouncedAt`), distinct from actual user-data loss (the `comments` queue, which does toast because a wiped queue is lost data). A corrupt switch trail costs one wrong ⌘⇧H default that self-heals on the next switch. The diff didn't introduce the pattern — it moved `visitRecency`'s previously hand-rolled warn onto this same shared surface. Recorded in `e7f3054c3`'s commit message rather than fixed.

All other built-in and `.agency/code-police.md` project rules checked clean (each with a Checked-by grep/enumeration; full per-rule table omitted here for compactness, available on request).

### Pass 2 — Fact-check

**Confirmed, FIXED** (commit `1af6b5840`): stale `selectedIndex` when the active terminal/host changes out from under an open switcher.

- `packages/client/src/CommandPalette.tsx:825-839` — the default-highlight effect is `createEffect(on([path, filterQuery, () => props.open], (...) => setSelectedIndex(defaultSelectionIndex(filtered(), currentSelection(), query()))))`. Solid's `on()` only tracks the three listed deps; `currentSelection()` (line 428, reading `encActiveHost()`/`view.activeId()`) is read **untracked** inside the callback.
- The code assumes "current" only changes via the palette's own `execute()` path (which closes first), but two independent paths change it while the palette stays open:
  1. `packages/client/src/input/actions.ts:192` `cycleTerminalMru` (Ctrl+Tab) has no `focusScopeMarker`; `packages/client/src/CommandPalette.tsx:749` calls `e.stopPropagation()` (not `stopImmediatePropagation()`), and neither `CommandPalette.tsx` nor `ModalDialog.tsx` set `data-kolu-modal="true"` (confirmed via grep — zero hits) — so the global capture-phase shortcut dispatcher in `useShortcuts.ts` still fires and changes `view.activeId()` in the background while the palette's own Tab handling also runs.
  2. `packages/client/src/wire.ts:479` — the host-membership reconcile bounces `activeHost` to local if the active host leaves the pool, independent of any palette interaction.
- In both cases `filtered()` (a plain `createMemo`) correctly re-derives against the new selection, but `selectedIndex` does not — none of `path`/`filterQuery`/`open` changed. Enter then calls `execute(items[selectedIndex()])` against the fresh list at a stale index: can activate a different terminal/host than what was highlighted, can silently re-select the row the user is already on (contradicting the feature's own stated invariant in `website/src/content/docs/switcher.mdx` that this can never happen), or can jump elsewhere.

**Fix (applied in `1af6b5840`):** `currentSelection` is now a tracked dependency of the default-highlight effect — `on([path, filterQuery, () => props.open, currentSelection], …)` — so the highlight re-lands whenever the row you are on moves, whatever moved it.

**Left for a follow-up, deliberately:** the Ctrl+Tab double dispatch itself. Silencing it needs either `data-kolu-modal` on the palette (which would also swallow ⌘K, the chord that toggles it closed) or a narrowed `stopImmediatePropagation` — both changes to global key dispatch, well outside this diff, and both pre-existing. With the highlight tracking, the wrong-row consequence is gone regardless of which path moved the selection.

No other logic issues found in `hostRecency.ts`/`visitRecency.ts` (bounded lists, correct monotonic tie-break, per-entry-isolated persisted-schema tolerance, no unbounded growth) or in `persistedPref.ts`'s validation tightening.

### Gates (at `1af6b5840`)

`just check` exit 0 · client vitest **1362 passed / 154 files** · e2e **49 scenarios** green across `terminal-switcher`, `command-palette`, `worktree-agent`.

srid:
## Evidence

Driven against **two real machines** — this tab's local host plus a freshly provisioned box (`root@kolu-evidence-2136`, an ephemeral Incus container bound over ssh with a real padi closure copy). The e2e harness is single-host, so this is the only way the ⌘⇧H toggle can be shown at all.

An isolated `nix build .#default` instance on a random free port with its own `KOLU_STATE_DIR` / `KOLU_PADI_STATE_DIR` — production `kolu.service` was untouched (it was `inactive` throughout, and teardown killed the captured PID, never a pattern).

### The toggle, one keystroke pair each way

**On `local`** — the trail has only the boot host, so the highlight falls to the other machine (the documented fresh-tab fallback):



**Enter → now on `root@kolu-evidence-2136`. ⌘⇧H again** — and the highlight is on **`local`**, the machine we came from. Note it is the row *above* the active one: the list is still in **pool order**, only the highlight moved:



Enter hops back, and a third ⌘⇧H points at the remote again — the state read out of the live page at each step:

| step | active host | trail (`sessionStorage`) | highlighted row |
| --- | --- | --- | --- |
| open on local | `local` | `[local]` | `root@kolu-evidence-2136` *(fallback — nothing to come back to yet)* |
| Enter, ⌘⇧H | `remote:root@kolu-evidence-2136` | `[remote, local]` | **`local`** |
| Enter, ⌘⇧H | `local` | `[local, remote]` | **`root@kolu-evidence-2136`** |

Rows stayed in pool order (`local`, then the remote) at every step — the trail moved the highlight, never the list.

### Mobile / coarse pointer

390×844 with a real coarse pointer (`matchMedia('(pointer: coarse)').matches === true`, not a width resize): the Hosts list renders in the compact layout, the highlighted row is the same one, and both rows sit fully inside the viewport (measured `top`/`bottom`, nothing clipped).

srid:
## e2e timing — `be2300f#1`

| `x86_64-linux` | `kolu-ci-4` | 8 | 32 | 8 | 3m47s |
| `aarch64-darwin` | `ci@petit` | 4 | 10 | 6 | 4m53s |

Recipe-wall durations from the run ledger (`.ci/be2300f/runs/1.json`), parallelism from each lane's last `e2e: workers=` line.

